PR #96, opened 2019-01-18
Change table_collection to embed tables.

jeromekelleher:


petrelharp:
Gee, this change touches a lot of places. But, it makes memory management simpler?

jeromekelleher:
> Gee, this change touches a lot of places. But, it makes memory management simpler?

So the motivation here is to try and make things a bit simpler. Embedding the table structs in the Table collection is how I originally envisaged things working, but the complexity of the way I had things set up at the Python/C layer at the time meant that I was forced to use pointers instead. These issues have been resolved, so I think it would be good to go back to embedding the structs. The advantages are:

1. We have an ``indexes`` sub struct, and having to say ``tables.nodes->time`` and ``tables.indexes.insertion_order`` was annoyingly inconsistent. Now we say ``tables.nodes.time``.
2. We improve locality a bit and avoid an extra pointer dereference when we use the table collection to keep things organised (pretty marginal, though, I doubt it'd make much difference).

PR #97, opened 2019-01-18
LIGHT C++ wrapper around some types.

molpopgen:
WIP to provide C++ classes wrapping some of the lower-level types using smart pointers and related idioms.  The goal is safe/idiomatic handling of types holding bare pointers.

- [ ] Figure out a `meson.build` that works.


jeromekelleher:
This is very cool! 

I'm worried, though, that the resulting client code is a bit of a mishmash of C and C++ idioms and it's going to be difficult to give clear messages about how to write good programs with it. In particular, some stuff is "safe" (i.e., exceptions will be thrown and you don't need to check return values) and some isn't. I worry that the average tskit user (who I don't expect to be a particularly sophisticated C++ or C programmer) isn't going to be able to follow these distinctions and will end up not checking *any* errors/leaking lots of memory. 

What I'm wondering is, how much extra work would a full, idiomatic C++ wrapper be? For example, we'd have something like a ``NodeTable`` class which has a private ``tsk_node_table_t`` instance. This would have methods like ``add_row``, which would call the equivalent C function. Importantly, this would check the C return value and raise an exception if an error occurred. To me this would be a genuine gain over the C API, as the burden of error checking would be removed and the API would be conceptually cleaner and map directly into clean object-oriented C++ code.

I realise it would be a bit tedious to do and there would be a fair bit of repetition. I'd be happy to do a chunk of the legwork though if I had an example to work from.

@bhaller, have you any thoughts on this?

bhaller:
@jeromekelleher I didn't look at the diffs for what @molpopgen has done here, but if the idiom created here is "a mishmash" as you say, then I think your comment sounds spot on.  Handling pure C code is straightforward, if ugly; you know you need to be prepared for error returns from pretty much every call, and you act accordingly.  Handling pure C++ code is also straightforward (although I'm not enough of a C++ programmer to really fully understand the exception model in C++, which seems rather complicated and confusing; in my C++ code I use exceptions in only a very rudimentary manner).  But when the idiom is mixed, you have to check the documentation for every call to see whether you should expect it to raise or to give an error return or, perhaps, both in different circumstances.  That seems like the worst of both worlds.  But again, I didn't look at the diffs, so my opinion should be taken with a grain of salt; and I'm really not a very good C++ programmer, either, so in that respect, too, my opinion is of questionable value.  :->

molpopgen:
> I'm worried, though, that the resulting client code is a bit of a mishmash of C and C++ idioms and it's going to be difficult to give clear messages about how to write good programs with it. In particular, some stuff is "safe" (i.e., exceptions will be thrown and you don't need to check return values) and some isn't. I worry that the average tskit user (who I don't expect to be a particularly sophisticated C++ or C programmer) isn't going to be able to follow these distinctions and will end up not checking _any_ errors/leaking lots of memory.
> 

Hence the "WIP" label! :)

So, the goal here is the memory leaks, and I'm less concerned with throwing exceptions directly.  Thus, I replaced the struct I'd originally written with the unique_ptr.  

Basically, I treat this like one would treat the GSL: hold the types allocated by C in smart pointers, and then use the library "as normal".   Doing that is generally quite safe, as GSL types generally don't spawn other bare pointers, and access via the `data` members is rare.   You simply know not to free them or do other dumb things, just as with the vector class's `data()` function, which returns a bare pointer for compatibility with C API.  This is relevant because of the next point...

> What I'm wondering is, how much extra work would a full, idiomatic C++ wrapper be? For example, we'd have something like a `NodeTable` class which has a private `tsk_node_table_t` instance. This would have methods like `add_row`, which would call the equivalent C function. Importantly, this would check the C return value and raise an exception if an error occurred. To me this would be a genuine gain over the C API, as the burden of error checking would be removed and the API would be conceptually cleaner and map directly into clean object-oriented C++ code.
> 
> I realise it would be a bit tedious to do and there would be a fair bit of repetition. I'd be happy to do a chunk of the legwork though if I had an example to work from.

...which is that the tables collection contains many bare pointers that will be accessed.  Now, you have to start distinguishing *how* a NodeTable is being accessed, which is roughly equivalent to distinguishing different ways it was initialized.  If it is a node table that is part of a table collection, then that is a different wrapper from a node table allocated on its own.  Or, it is one wrapper with multiple different behaviors.

So, I think a lot of boiler plate is required.  The number of unit tests needed roughly doubles, and another dependency is required in order to unit tests the exceptions.

Anyhow, this is all largely experimental...

molpopgen:
Closing in favor of #98.

PR #98, opened 2019-01-21
Very rough C++ experiment for discussion.

jeromekelleher:
So here's roughly what I had in mind @molpopgen. I agree it's a fair bit of work, but I think the API we get at the end is quite nice. Basically exactly what you'd expect when coming from Python.

There's some details here that need working out about how the constructors work (the TableCollection version is leaking memory at the moment), but I'm sure it could be resolved.

What do you think? I'm not suggesting you do the work here BTW, I'd just value your opinion. I think it would be good to have a C++ API like this, but I'm not going to take on doing it without some expert advice, as I'm a total noob!

### DIFF
--- a/c/examples/cpp-experiment.cpp
+++ b/c/examples/cpp-experiment.cpp
@@ -1,0 +1,127 @@
+/* C++ wrapper experiement. */
+
+#include <iostream>
+#include <stdexcept>
+#include <sstream>
+
+#include <tskit/tables.h>
+
+using namespace std;
+
+void
+check_error(int val)
+{
+    if (val < 0) {
+        std::ostringstream o;
+        o << tsk_strerror(val);
+        throw std::runtime_error(o.str());
+    }
+}
+
+class NodeTable
+{
+
+    private:
+        tsk_node_table_t *table;
+        bool malloced_locally = true;
+
+    public:
+        NodeTable()
+        {
+            table = (tsk_node_table_t *) malloc(sizeof(*table));
+            if (table == NULL) {
+                throw std::runtime_error("Out of memory");
+            }
+            malloced_locally = true;
+            int ret = tsk_node_table_init(table, 0);
+            check_error(ret);
+        }
+
+        NodeTable(tsk_node_table_t *the_table)
+        {
+            /* FIXME: leaking memory here because the standard constructor has already been
+             * called. Must be a way to dealing with this though. */
+            table = the_table;
+            std::cout << "table constructor" << endl;
+            malloced_locally = false;
+        }
+
+        ~NodeTable()
+        {
+            if (malloced_locally && table != NULL) {
+                tsk_node_table_free(table);
+                free(table);
+            }
+        }
+
+
+        tsk_id_t add_row(tsk_flags_t flags, double time) /* and more params */
+        {
+            tsk_id_t ret = tsk_node_table_add_row(table, flags, time, TSK_NULL,
+                    TSK_NULL, NULL, 0);
+            check_error(ret);
+            return ret;
+        }
+
+        tsk_size_t get_num_rows(void)
+        {
+            return table->num_rows;
+        }
+        /* Etc */
+};
+
+class TableCollection
+{
+
+    private:
+        tsk_table_collection_t *tables;
+
+    public:
+        NodeTable nodes;
+
+        TableCollection(double sequence_length)
+        {
+            tables = (tsk_table_collection_t *) malloc(sizeof(*tables));
+            if (tables == NULL) {
+                throw std::runtime_error("Out of memory");
+            }
+            int ret = tsk_table_collection_init(tables, 0);
+            check_error(ret);
+            tables->sequence_length = sequence_length;
+            nodes = NodeTable(&tables->nodes);
+        }
+
+        ~TableCollection()
+        {
+            if (tables != NULL) {
+                tsk_table_collection_free(tables);
+                free(tables);
+            }
+        }
+
+        double get_sequence_length()
+        {
+            return tables->sequence_length;
+        }
+};
+
+
+int
+main()
+{
+    NodeTable nodes;
+
+    nodes.add_row(0, 1.0);
+    nodes.add_row(0, 2.0);
+    std::cout << "Straight table: num_rows = " << nodes.get_num_rows() << endl;
+
+    TableCollection tables = TableCollection(10);
+    std::cout << "Sequence length = " << tables.get_sequence_length() << endl;
+    tables.nodes.add_row(0, 1.0);
+    tables.nodes.add_row(0, 2.0);
+    tables.nodes.add_row(0, 3.0);
+    std::cout << "Via table collection: num_rows = " << tables.nodes.get_num_rows() << endl;
+
+    return 0;
+
+}
--- a/c/meson.build
+++ b/c/meson.build
@@ -1,7 +1,7 @@
-project('tskit', 'c', default_options: ['c_std=c99'])
+project('tskit', ['c', 'cpp'], default_options: ['c_std=c99'])
 
 kastore_dir=join_paths(meson.source_root(), 'kastore/c')
-add_global_arguments(['-I' + kastore_dir], language: 'c')
+add_global_arguments(['-I' + kastore_dir], language: ['c', 'cpp'])
 
 cc = meson.get_compiler('c')
 m_dep = cc.find_library('m', required: false)
@@ -85,6 +85,9 @@ executable('error_handling',
     sources: ['examples/error_handling.c'], link_with: [tskit_lib])
 executable('tree_iteration', 
     sources: ['examples/tree_iteration.c'], link_with: [tskit_lib])
+
+executable('cpp-experiment', 
+    sources: ['examples/cpp-experiment.cpp'], link_with: [tskit_lib], language: 'cpp')
 
 gsl_dep = dependency('gsl', required: false)
 if gsl_dep.found()
--- a/c/tskit/core.h
+++ b/c/tskit/core.h
@@ -5,6 +5,10 @@
 #ifndef __TSK_CORE_H__
 #define __TSK_CORE_H__
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #include <stdbool.h>
 
 #ifdef __GNUC__
@@ -219,5 +223,9 @@ size_t tsk_search_sorted(const double *array, size_t size, double value);
 
 #define TSK_UUID_SIZE 36
 int tsk_generate_uuid(char *dest, int flags);
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif


molpopgen:
Yes, so this goes down a long road.  It is doable, but does need some thought.

The reason for the leak is the following:

1. There is no [initialization list](https://en.cppreference.com/w/cpp/language/initializer_list) for the TableCollection constructors.  
2. Thus, the public NodeTable is [default initialized](https://en.cppreference.com/w/cpp/language/default_initialization), meaning that `NodeTable()` is called, which calls an allocation.
3. Then, line 91 replaces the allocated pointer with that from the private table collection.

The rule of thumb is that any non-trivial type should be dealt with via the initialization list.  The culprit is the default constructor for the NodeTable.   Something like the following will be safer:

```cpp
// Construct from a pre-allocated, pre-initialized node table
// The "malloced local" idiom will be FALSE
explicit NodeTable(tsk_node_table_t *); 

// This will steal the input pointer via a call to ptr_to_steal.release()
// malloced_local will be TRUE
// You will have to require that the underlying pointer be originally allocated with EITHER new or malloc,
// with no mixed use allowed, and, to be idiomatic, a preference for new.
explicit NodeTable(std::unique_ptr<tsk_node_table_t, void(*)(tsk_node_table_t *)> & ptr_to_steal);
```

The next memory leak is this one:

```cpp
// The auto-generated copy constructor
auto tcopy(tables);
```

molpopgen:
I'm happy to help out on something like this, but it will be a "heavy" task.  The other consideration is that you'll want to get any and all changes into the C API first.  The "double maintenance" burden will become high.

jeromekelleher:
> I'm happy to help out on something like this, but it will be a "heavy" task. The other consideration is that you'll want to get any and all changes into the C API first. The "double maintenance" burden will become high.

Excellent, thanks Kevin! I agree it's not trivial, but there are significant benefits to users so I think it's worth it. Also, I'm thinking of some places where a C++ API would be really nifty, like iterating over nodes in a tree in the various orders (zero-overhead abstraction would be great for this; very difficult to do well in C).

I agree the double maintenance is an issue, so I'm going to finish up the 1.0.0 changes for the C side of things first. After 1.0.0 is tagged I think we should be very reluctant indeed to make any changes to the documented API for user's sake, so perhaps the extra drag of the C++ layer is a good thing there.

molpopgen:
Here are some placeholders for posterity/discussion before I have to stop procrastinating:

1. The danger zone is errors happening in constructors.  When an exception is thrown, destructors only get called for properly-initialized objects.  This means that you can leak memory from a constructor. The solution is to use smart pointers prior to any work that can trigger an exception.  The current code makes note of this by way of an example.

2. We need to decide what operations are acceptable/allowed:
- [ ] Copy construction (`auto b(a)`)
- [ ] Assignment, or `decltype(a) b = a`, which occurs via `operator=`.
- [ ] Equality comparison, `a == b`, which is useful for testing after reading back from a file, etc.

3. The care needed for safe construction means that we can end up with a lot of private functions that are doing the hard work.  This clutters the header.  The alternative is:

```cpp
class NodeTableImpl; //forward-declaration

class NodeTable
{
   private:
        std::unique_ptr<NodeTableImpl> pimpl;  //pimpl = "pointer to implementation"

  public:

};
```

Then, `NodeTableImpl`, and any public functions depending on private data, get implemented in source files (.cc/.cpp) that become part of a `libtskitpp.so`.

This is optional.  Header-only libraries are all the rage these days, but we're not doing much fancy template wackery, so an old-school sofile seems appropriate.

There is some concern due to loss of inlining here, but we already have to call back to `libtskit.so`.

jeromekelleher:
This is all excellent, thanks @molpopgen! Re. the lack of inlining, I don't think this will matter at the level of the tables API. Nothing here is going to be called in a tight inner loop, so a bit of overhead makes no difference.

I think the place were perf will matter is on tree traversals in [pre,post,in]order. We can do this natively in C++ so can think about how to get top performance there when doing the Tree class.

molpopgen:
With LTO, the inlining can be brought back, in theory at least.  Client code will have to do that on their own, I think, and we need to work out the "rpath" stuff so that only the c++ lib needs linking to.

molpopgen:
Another decision is what testing framework to use.  All my stuff uses boost's test library, but it is a very heavy dependency. It is good, but I'll research an alternative.  This matters when testing for "this must throw an exception" and "this must not throw".

jeromekelleher:
> With LTO, the inlining can be brought back, in theory at least. Client code will have to do that on their own, I think, and we need to work out the "rpath" stuff so that only the c++ lib needs linking to.

I'm leaning towards recommending people just compile the code in to their own project. It's not much code and saves a lot of hassle, and is how people build bioinformatics programs in practise. 

> Another decision is what testing framework to use. All my stuff uses boost's test library, but it is a very heavy dependency. It is good, but I'll research an alternative. This matters when testing for "this must throw an exception" and "this must not throw".

What about GTest? I don't mind though, once it's reasonably simple to work with. The dependency thing doesn't really matter: it's only developers that need it. Once we can install it with apt-get it doesn't really matter (although it would be nice if we didn't have to pull in hundreds of megabytes each time for CI).

molpopgen:
> > With LTO, the inlining can be brought back, in theory at least. Client code will have to do that on their own, I think, and we need to work out the "rpath" stuff so that only the c++ lib needs linking to.
> 
> I'm leaning towards recommending people just compile the code in to their own project. It's not much code and saves a lot of hassle, and is how people build bioinformatics programs in practise.

Make sense.

> 
> > Another decision is what testing framework to use. All my stuff uses boost's test library, but it is a very heavy dependency. It is good, but I'll research an alternative. This matters when testing for "this must throw an exception" and "this must not throw".
> 
> What about GTest? I don't mind though, once it's reasonably simple to work with. The dependency thing doesn't really matter: it's only developers that need it. Once we can install it with apt-get it doesn't really matter (although it would be nice if we didn't have to pull in hundreds of megabytes each time for CI).

You mean [GoogleTest](https://github.com/google/googletest)?  IIRC, it doesn't support all of C++11, much less later.  It is also super-weird to install, following the tradition of many google C++ tools!  After some reading, I am leaning towards [catch2](https://github.com/catchorg/Catch2), which is a single header file that we can just drop in to the `tests` folder.  It is under the Boost license, which is also a plus.  They have a nice way of avoiding writing cumbersome fixtures for tests.

c/examples/cpp-experiment.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+        private:
+            tsk_node_table_t *table;
+            const bool allocated_locally;
+            using node_table_ptr = std::unique_ptr<tsk_node_table_t,void(*)(tsk_node_table_t*)>;
+
+            node_table_ptr make_empty_table()
+            {
+                node_table_ptr t(new tsk_node_table_t{},[](tsk_node_table_t* nt){delete nt;});
+                if (t.get() == NULL) {
```
jeromekelleher: I did a bit of digging and I don't think we need to worry about ``new`` returning NULL: it raises an exception if allocation fails.

c/examples/cpp-experiment.cpp
```diff
@@ -0,0 +1,215 @@
...
+                return t.release();
+            }
+
+        public:
+            explicit NodeTable(tsk_node_table_t *the_table) : table(allocate(the_table)), allocated_locally(the_table != table)
+            {
+                std::cout << "table constructor " << endl;
+            }
+
+            NodeTable(const NodeTable & other) : table(copy_construct_details(other)), allocated_locally(other.allocated_locally)
```
jeromekelleher: So, I think copy should always be a deep copy here. We allocate a new node_table_t and call node_table_copy() on it. Does this allow us to simplify things a bit here?

jeromekelleher:
I've played around with this for a couple of hours @molpopgen and have been learning a lot. I hadn't realised quite how tricky C++ object initialisation is...

c/examples/cpp-experiment.cpp
molpopgen: Yes, probably.  I was attempting to make this "copy-free":

```cpp
TableCollection foo;
auto n(foo.nodes);
```

If the constructor always makes a deep copy, then this will have the same effect:

```cpp
// Refer to the member table by reference
auto & n = foo.nodes;
```

molpopgen:
> I've played around with this for a couple of hours @molpopgen and have been learning a lot. I hadn't realised quite how tricky C++ object initialisation is...

There are vestiges of C here.  You have to make sure that all memory allocated can be freed for all possible exit paths out of a constructor.  Fortunately, smart pointers help a lot here.

c/examples/cpp-experiment.cpp
molpopgen: Yeah.  This actually brings up a whole other issue :).  I thought that the C++ interface should be "new and delete only", but then I wondered about the idea of a TableCollection "releasing" its data to a bare pointer, in the manner of `unique_ptr::release`.  In that case, it would have to be a "malloc/free" combo.

There is a [non-throwing](http://www.cplusplus.com/reference/new/nothrow/) way to call new, which returns `nullptr`.  I wonder if I left that `NULL` check in there when switching away from malloc?  It should be `nullptr` and not `NULL`, which makes me think I left some cruft around.

molpopgen:
One thing to keep in mind: the member tables may need to be marked as either `const` or `mutable`.  Preferably the former unless it makes certain other operations too tedious to implement.  We have to turn an attempt to move a `NodeTable` from a `TableCollection` into a compiler error.

Related: the implementation of move constructors will be a bit delicate.  I do not think that compiler-generated defaults will do the trick for the case where "allocated locally" is `false`.  In order to maximize utility, such constructors must not throw exceptions.  This will require me to do some reading--it is a tricky area.    See if there's a copy of [this](https://www.amazon.com/Effective-Modern-Specific-Ways-Improve/dp/1491903996) hanging around your office!  It will help.

jeromekelleher:
I've rebased this on master @molpopgen and added a rough outline of what I think the code organisation might look like. You'll need to ``reset --hard`` your branches because of the force push.

 I'm sure I've messed up some of the details with the namespace etc, so don't take anything as too literal. Does the overall organisation make sense to you? I'd like to split it up into separate hpp/cpp pairs following the C organisation, just so that we always know where things should be. Ideally I'd like to get all (or basically all, except say trivial accessors) implementation into the cpp files and out of the headers.

c/tskitpp.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <iostream>
+#include <stdexcept>
+#include <sstream>
+#include <memory>
+#include <cassert>
+
+#include <tskit.h>
+
+/* JK: this is just getting stuff working. Probably the namespace should be declared
+ * in each file or something? Looks like we can't have includes within a namespace
```
molpopgen: If you include stuff from `std` in another namespace, you get a new namespace, IIRC.   But yea, that's "not done".

molpopgen:
Looks like the next task is to get the remaining table classes started.  Then, we can copy the C example program and hack it to use the C++ interface.  While we are at it, we can have it use the C++11 rng libraries instead of GSL.

jeromekelleher:
Shall we close this @molpopgen? I don't think we're ever going to do a C++ interface now, the memory management is just too tricky and it's really not that big a deal to use the C API directly in C++.

molpopgen:
I agree.  I was never truly on board with the idea of a fully-fleshed-out set of classes.  To document things for folks curious in the future, the proper way to do this is with smart pointers and custom deleter functions.  For example:

```cpp
using table_collection_ptr = std::unique_ptr<tsk_table_collection_t, std::function<void(tsk_table_collection_t*)>>;

void
handle_tskit_return_code(int code)
{
    if (code < 0)
        {
            std::ostringstream o;
            o << tsk_strerror(code);
            throw std::runtime_error(o.str());
        }
}

table_collection_ptr
make_table_collection_ptr(double sequence_length)
{
    table_collection_ptr rv(new tsk_table_collection_t(),
                            [](tsk_table_collection_t* tables) {
                                tsk_table_collection_free(tables);
                                delete tables;
                            });
    int err = tsk_table_collection_init(rv.get(), 0);
    handle_tskit_return_code(err);
    rv->sequence_length = sequence_length;
    if (err != 0)
        {
            throw std::runtime_error("could not initialize tsk_table_collection");
        }
    return rv;
}
```

You are now exception-safe, and can store this pointer within classes easily.   An encapsulating class is non-copyable by default, and the constructors/operators needed to enable copy are simple to write.

Issue #29, opened 2018-12-19
Rename tree.length attribute to tree.span?


hyanwong:
At the moment the msprime/tskit API defines `SparseTree.length` to be the genomic length covered by a tree. It can be quite confusing when branches on a tree can have "lengths" too. I’ve been trying to use the word "span" instead of "length" to refer to the genome span covered by an edge or a tree. Would it be sensible to add the attribute .span as a preferred alternative to .length on a SparseTree (and perhaps change the description in other places in the documentation where we refer to extent over a genomic region, such as the "span" (i.e. length) of an edge).

PR #121, opened 2019-02-08
More flexible tree API

### LINKED ISSUE #24: Add ``start`` and ``stop`` arguments to TreeSequence.trees()
These should be seen as analogous to the ``start`` and ``stop`` arguments to Python's builtin ``range``, and be specified in tree sequence coordinates. We should initially ensure that start < stop, but in future we could allow start > stop, and enable reverse iteration in this way.

It would be best to do this with a low-level ``sparse_tree_seek()`` method to position the tree at the correct location.

### LINKED ISSUE #60: __eq__ support for SparseTree
It would be useful to have equality testing for trees on the same tree sequence. The semantics are already implement in C, so just need to hook into this code.

jeromekelleher:
We need better ways of getting at trees in a tree sequence. Specifically, I want to be able to go backwards through the trees in Python. Since we're nailing down the C API also we should try to finalise the capabilities now (even if some of them aren't as efficiently implemented as they might be).

This is a WIP to do that. Should 

- Close #60 

The low-level API is good I think, and the changes are welcome improvements. I need to think a bit more about what the high-level Python API will look like.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
             PYTHONPATH=python nosetests -v --with-coverage --cover-package tskit \
               --cover-branches --cover-erase --cover-xml \
               --cover-inclusive python/tests
-            python3 -m codecov -F python-tests
+            python3 -m codecov -F python_tests
 
       - run:
           name: Build Python package
@@ -97,7 +97,7 @@ jobs:
             gcov -pb tskit@sta/tskit_convert.c.gcno ../c/tskit/convert.c
             gcov -pb tskit@sta/tskit_stats.c.gcno ../c/tskit/stats.c
             cd ..
-            codecov -X gcov -F c-tests
+            codecov -X gcov -F c_tests
 
       - run:
           name: Valgrind for C tests.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
             PYTHONPATH=python nosetests -v --with-coverage --cover-package tskit \
               --cover-branches --cover-erase --cover-xml \
               --cover-inclusive python/tests
-            python3 -m codecov -F python_tests
+            python3 -m codecov -F python_tests -X gcov
 
       - run:
           name: Build Python package
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,7 @@ coverage:
     project:
       python-tests:
         target: 95%
-        flags: python-tests
+        flags: python_tests
       c-tests:
         target: 85%
-        flags: c-tests
+        flags: c_tests
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -192,13 +192,6 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
     for (ret = tsk_tree_last(&t); ret == 1; ret = tsk_tree_prev(&t)) {
         CU_ASSERT_EQUAL_FATAL(j - 1, t.index);
         ret = tsk_tree_equal(&t, &trees[t.index]);
-        if (ret != 0) {
-            printf("trees differ\n");
-            printf("REVERSE tree::\n");
-            tsk_tree_print_state(&t, stdout);
-            printf("FORWARD tree::\n");
-            tsk_tree_print_state(&trees[t.index], stdout);
-        }
         CU_ASSERT_EQUAL_FATAL(ret, 0);
         j--;
     }
@@ -282,6 +275,37 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
         ret = tsk_tree_equal(&t, &trees[t.index]);
         CU_ASSERT_EQUAL_FATAL(ret, 0);
     }
+
+    ret = tsk_tree_free(&t);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_tree_init(&t, ts, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    /* Calling next() on an uninitialised tree should be the same as first() */
+    j = 0;
+    while ((ret = tsk_tree_next(&t)) == 1) {
+        CU_ASSERT_EQUAL_FATAL(j, t.index);
+        ret = tsk_tree_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        j++;
+    }
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(j, num_trees);
+
+    ret = tsk_tree_free(&t);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_tree_init(&t, ts, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    /* Calling prev() on an uninitialised tree should be the same as last() */
+
+    j = num_trees;
+    while ((ret = tsk_tree_prev(&t)) == 1) {
+        CU_ASSERT_EQUAL_FATAL(j - 1, t.index);
+        ret = tsk_tree_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        j--;
+    }
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(j, 0);
 
     /* Free the trees. */
     ret = tsk_tree_free(&t);
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -34,6 +34,32 @@
  * Verification utilities.
  *======================================================*/
 
+static int
+check_trees_equal(tsk_tree_t *self, tsk_tree_t *other)
+{
+    int ret = 0;
+    int condition;
+    size_t N = self->num_nodes;
+
+    CU_ASSERT_FATAL(self->tree_sequence == other->tree_sequence);
+    condition = self->index == other->index
+        && self->left == other->left
+        && self->right == other->right
+        && self->sites_length == other->sites_length
+        && self->sites == other->sites
+        && memcmp(self->parent, other->parent, N * sizeof(tsk_id_t)) == 0;
+    /* We do not check the children for equality here because
+     * the ordering of the children within a parent are essentially irrelevant
+     * in terms of topology. Depending on the way in which we approach a given
+     * tree we can get different orderings within the children, and so the
+     * same tree would not be equal to itself. */
+    if (condition) {
+        ret = 1;
+    }
+    CU_ASSERT_FATAL(ret == tsk_tree_equals(self, other));
+    return ret;
+}
+
 static void
 verify_compute_mutation_parents(tsk_treeseq_t *ts)
 {
@@ -150,8 +176,8 @@ get_tree_list(tsk_treeseq_t *ts)
         CU_ASSERT_EQUAL_FATAL(ret, 0);
         ret = tsk_tree_copy(&trees[t.index], &t);
         CU_ASSERT_EQUAL_FATAL(ret, 0);
-        ret = tsk_tree_equal(&trees[t.index], &t);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = check_trees_equal(&trees[t.index], &t);
+        CU_ASSERT_EQUAL_FATAL(ret, 1);
         /* Make sure the left and right coordinates are also OK */
         CU_ASSERT_DOUBLE_EQUAL(trees[t.index].left, t.left, 1e-6);
         CU_ASSERT_DOUBLE_EQUAL(trees[t.index].right, t.right, 1e-6);
@@ -179,8 +205,8 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
     j = 0;
     for (ret = tsk_tree_first(&t); ret == 1; ret = tsk_tree_next(&t)) {
         CU_ASSERT_EQUAL_FATAL(j, t.index);
-        ret = tsk_tree_equal(&t, &trees[t.index]);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = check_trees_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 1);
         j++;
     }
     CU_ASSERT_EQUAL_FATAL(ret, 0);
@@ -190,8 +216,8 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
     j = num_trees;
     for (ret = tsk_tree_last(&t); ret == 1; ret = tsk_tree_prev(&t)) {
         CU_ASSERT_EQUAL_FATAL(j - 1, t.index);
-        ret = tsk_tree_equal(&t, &trees[t.index]);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = check_trees_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 1);
         j--;
     }
     CU_ASSERT_EQUAL_FATAL(ret, 0);
@@ -201,16 +227,16 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
     j = 0;
     for (ret = tsk_tree_first(&t); ret == 1; ret = tsk_tree_next(&t)) {
         CU_ASSERT_EQUAL_FATAL(j, t.index);
-        ret = tsk_tree_equal(&t, &trees[t.index]);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = check_trees_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 1);
         j++;
     }
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     CU_ASSERT_EQUAL_FATAL(j, num_trees);
     while ((ret = tsk_tree_prev(&t)) == 1) {
         CU_ASSERT_EQUAL_FATAL(j - 1, t.index);
-        ret = tsk_tree_equal(&t, &trees[t.index]);
-        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = check_trees_equal(&t, &trees[t.index]);
+        CU_ASSERT_EQUAL_FATAL(ret, 1);
         j--;
     }
     CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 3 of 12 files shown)


jeromekelleher:
This is quite a big increase in our power to move around the trees in Python, so pinging @tskit-dev/all for any thoughts. The PR adds:

- len(tree_sequence) returns the number of trees in tree sequence
- tree_sequence[index] returns a new tree instance at given index
- list(tree_sequence) returns a list of tree instances (inefficient, but handy)
- reversed(tree_sequence.trees()) efficiently iterates backwards over trees
- Manual method for iterating over trees like:

```python
tree = tskit.Tree(TreeSequence)
# Forwards iteration
while tree.next():
    # Do something with the tree
# Backwards iteration
while tree.prev():
    # Do something with tree
```

- Manual seeking methods (currently inefficient, but can be implemented efficiently in C if we use the underlying edge indexes):

```python
tree = tskit.Tree(ts)
tree.seek_index(len(ts // 2)) # Seek to the tree with middle index
tree.seek_position(ts.sequence_length / 2) # Seek to the tree in the middle of the sequence
```

Does these all look sensible to you?

hyanwong:
> Does these all look sensible to you?

These all seem very intuitive, and a great advance in the ability to deal with tree sequences. A quick thought: do we want to move to using len(ts) over ts.num_trees in the examples?

Does the seek_index() operator return an iterator? I'd want to be able to jump to (say) tree number 999 and iterate from then onwards. On that note, even better would be the ability to iterate over a slice a ts, either by (a) tree index or (b, and probably more commonly) over a genomic region. Could the seek_index() and seek_position() functions serve as a base for a function of this sort. Do you think it would be inelegant to override the slice operator for (a)?

We should probably warn in the docs about the inadvisability of doing list(ts) for large trees - could blow up your computer etc.

hyanwong:
Sorry - not reading hard enough. I see that tree.seek_index() changes the iterator position.

hyanwong:
On the topic of slicing, i'd like to be able to do

`
for tree in ts.trees(start=10, end=20):
`

which I guess could be an easy shorthand for your `seek_index()` code, if the default `end` value was len(ts), e.g. `for tree in ts.trees(start=len(ts)//2):`

This seems a bit neater to me. I guess other parameters could be used to iterate over trees sliced by position, but perhaps that risks parameter overload?

`
for tree in ts.trees(start_pos=0.2, end_pos=0.8):
`

jeromekelleher:
This is the bit I haven't figured out fully. I agree with what you're proposing above with start and end for ``trees``, I just want to make sure we're doing it in as Pythonic a way as possible. 

For slices, I think it'd be bad to use iterators there and would end up being quite confusing. We can stick with the ``ts[start:stop]`` notation is "handy but inefficient" narrative, allowing people to use this for small tree sequences. (This should work fine as is, btw)

hyanwong:
Agree with the use of `[]` slice notation as a "handy but inefficient" method.

What's not quite clear to me is the difference between `ts.trees()` and `tskit.Tree(TreeSequence)` - don't both return an iterator over the trees? With your new notation, might I not expect to be able to do:

`
tree = ts.trees()
tree.seek_index(20)
`

In which case, why have `tskit.Tree()` at all?

hyanwong:
Three final comments: 

1. `ts[99]` is useful, but often I'd want a tree at a specific genomic location, rather than a specific index. Not sure what is pythonic here, but if `ts[99]` is a shorthand for `ts.index(99)`, then maybe there should be a `ts.position()` too?

2. It seems as if you are moving to a world in which tree sequences are considered primarily as a, well, sequence of trees. Hence `ts[XX]` returns the XXth tree, and list(ts) returns the trees. So I might expect to iterate over the trees simply by doing `trees = iter(ts)` rather than `ts.trees()` or `tskit.Tree(ts)`. OTOH, `iter(ts)` wouldn't allow `start` and `stop` parameters, so maybe `ts.__iter()` should just be a shorthand for `ts.trees()` with the default set of parameters?

3. If we *do* use a parameterised way of accessing a range of trees, some thought should go into whether we name the parameters for the tree indexes `(begin, end)`, `(start, stop)` or `(left, right)`, and what we call the values for positions. Either way, I'd make sure that we refer to the range as a "span" rather than a "length" (see https://github.com/tskit-dev/tskit/issues/29).

jeromekelleher:
Thanks Yan, good points. There's an essential distinction between using ts.trees() and tskit.Tree(ts): one returns an iterator and the other doesn't. I need to write up how to use the Tree class for iteration in the tutorial. 

Specific points above:

1. Yes, we'll want a method to return the tree at a given position. Perhaps ``TreeSequence.at(position)``.
2. We can't have the ``__iter__`` method iterate efficiently if we want ``list(ts)`` to return something that makes sense. That's why ``trees()`` must be separate (as well as allowing for parameters).
3. Having thought about it, I think there's no point in having ``start``, ``stop`` parameters for ``.trees()`` in terms of indexes (when are you going to know you want the 100th to the 150th tree in a sequence?). So, we should just have ``start`` and ``stop`` in terms of coordinates. However, there are semantic difficulties with this (e.g., what does ``stop`` mean? Do we return the tree including this coordinate, or the last tree that doesn't include it?), which would need updates at the C level to resolve properly, so I think should be parked for now. I'll update the issue above with what I think should be done.

molpopgen:
On Mon, Feb 11, 2019, 3:00 PM Jerome Kelleher <notifications@github.com
wrote:

> Thanks Yan, good points. There's an essential distinction between using
> ts.trees() and tskit.Tree(ts): one returns an iterator and the other
> doesn't. I need to write up how to use the Tree class for iteration in the
> tutorial.
>
> Specific points above:
>
>    1. Yes, we'll want a method to return the tree at a given position.
>    Perhaps TreeSequence.at(position).
>    2. We can't have the __iter__ method iterate efficiently if we want
>    list(ts) to return something that makes sense. That's why trees() must
>    be separate (as well as allowing for parameters).
>    3. Having thought about it, I think there's no point in having start,
>    stop parameters for .trees() in terms of indexes (when are you going
>    to know you want the 100th to the 150th tree in a sequence?). So, we should
>    just have start and stop in terms of coordinates. However, there are
>    semantic difficulties with this (e.g., what does stop mean? Do we
>    return the tree including this coordinate, or the last tree that doesn't
>    include it?), which would need updates at the C level to resolve properly,
>    so I think should be parked for now. I'll update the issue above with what
>    I think should be done.
>
> For consistency, it seems that [start, stop) is the most natural
definition?

hyanwong:
> 3\. Having thought about it, I think there's no point in having `start`, `stop` parameters for `.trees()` in terms of indexes (when are you going to know you want the 100th to the 150th tree in a sequence?).

True, but similarly there's little reason to want to get the 50th tree using `ts[50]`. It would only be for testing purposes, or perhaps in internal code.

jeromekelleher:
> True, but similarly there's little reason to want to get the 50th tree using ts[50]. It would only be for testing purposes, or perhaps in internal code.

trees = [ts[j] for j in range(len(ts))]

jeromekelleher:
> definition?

I'm not following the question @molpopgen, what do you want defined? I've updated #24 to define the semantics of the range operation I'm thinking of.

jeromekelleher:
> True, but similarly there's little reason to want to get the 50th tree using ts[50]. It would only be for testing purposes, or perhaps in internal code.

To expand on this a bit, it's very easy to get the functionality required using the manual Tree API:
```python
tree = tskit.Tree(ts)
tree.seek_index(start)
while True:
    if tree.index == end:
         break
    # Do stuff
    tree.next()
```

If it's just for internal stuff and testing, there's no point in cluttering up the API with it when we can easily get the same effect.

molpopgen:
You were asking what stop should mean.  All other intervals are half open,
so it seems natural for stop to be treating the same way

hyanwong:
> On Mon, Feb 11, 2019, 3:00 PM Jerome Kelleher ***@***.*** wrote: Thanks Yan, good points. There's an essential distinction between using ts.trees() and tskit.Tree(ts): one returns an iterator and the other doesn't.

I'ma bit confused, but perhaps that means other users will be too. Which one returns an iterator? `ts.trees()` returns a generator with a __next__() method, which I loosely call an iterator. tskit.Trees(ts) returns an object which has a `.next()`, which makes it looks like a (Python 2) iterator with some extra functionality like `.prev()` and `seek_xx()`, a bit like the python2 `file` object, which is its own iterator (https://docs.python.org/2.4/lib/bltin-file-objects.html).

hyanwong:
> To expand on this a bit, it's very easy to get the functionality required using the manual Tree API:

Sorry, I meant, why even have `ts[50]` as a publicly available method: isn't indexing by tree position only going to be used for internal stuff and testing too? But perhaps it's so intuitive that this isn't really an issue.

jeromekelleher:
> You were asking what stop should mean.  All other intervals are half open, so it seems natural for stop to be treating the same way

Ah right, the rest of the text got a bit mangled by the GitHub UI. I completely agree, but I think we to modify the trees a little bit to make it truly useful, as described over in #24.

jeromekelleher:
> Sorry, I meant, why even have ts[50] as a publicly available method: isn't indexing by tree position only going to be used for internal stuff and testing too? But perhaps it's so intuitive that this isn't really an issue.

I think it's useful and intuitive. It's annoying to have to fiddle with generators sometimes when all you want is a specific tree. Being able to get a list of the trees is genuinely useful (not everyone is going to be dealing with million sample tree sequences all the time), so we should support it. That naturally means we need to support ``len()`` and ``__getitem__``.

jeromekelleher:
> I'ma bit confused, but perhaps that means other users will be too. Which one returns an iterator? ts.trees() returns a generator with a next() method, which I loosely call an iterator. tskit.Trees(ts) returns an object which has a .next(), which makes it looks like a (Python 2) iterator with some extra functionality like .prev() and seek_xx(), a bit like the python2 file object, which is its own iterator (https://docs.python.org/2.4/lib/bltin-file-objects.html).

Have you read the current material in the tutorial above? I haven't covered  the manual tree API yet, but hopefully the iterator side should be clear. I'm using the term iterator here specifically to mean that it's a Python 3 iterator object.

hyanwong:
It's clearly written, thanks. I guess I don't see why `tskit.Tree(ts)` is needed. Why can't I get a trees object using `trees = ts.trees()`, then as well as being able to iterate over it, call e.g. `trees.seek_index()` or even `trees.prev()`. After all, that's what the IOBase object in Python 3 does. You can call:

```
file = open("my file")
for l in file():
 print(l)
file.seek(0)
# etc
```

It seems confusing to me to have both a bespoke manual object, with obj.next(), *and* a commonly used iterator, with a `__next__()` method. Why not combine them so that, Pythonically, there's only one way to do it.

hyanwong:
> For consistency, it seems that [start, stop) is the most natural
> definition?

Not that I mind either way, but some of the docs use (left,right), e.g. https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.edge_diffs . This has the advantage that it's clearer (to me anyway) that we are talking about a length of genome and not e.g. tree indexes, or node times. But it has the disadvantage that it seems weird to specify e.g. `(left=20.5, right=10.5)` in order to get an iterator that goes right to left. But perhaps its confusing to do that anyway, given the previous discussion about half-open intervals, and which way they would be treated. For a RTL iterator over a region, it would be clearer to get the iterator and then do `reversed()` on it.

jeromekelleher:
> It seems confusing to me to have both a bespoke manual object, with obj.next(), and a commonly used iterator, with a __next__() method. Why not combine them so that, Pythonically, there's only one way to do it.

I've added documentation for this in the tutorial. Basically we have to have the Tree class because that's how the higher level operations work. We can choose to hide this by not documenting it, but I think the low-level operations are useful and sometimes more elegant.

jeromekelleher:
> Not sure if you are only talking about the openness of the interval, or also the names for the endpoints. If the latter as well as the former, note that some of the docs use the words [left,right) rather than [start, stop), e.g. https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.edge_diffs . These words have the advantage that it's clearer (to me anyway) that we are talking about a length of genome and not e.g. tree indexes, or node times. But it has the disadvantage that it seems weird to specify e.g. (left=20.5, right=10.5) in order to get an iterator that goes right to left. But perhaps its confusing to do that anyway, given the previous discussion about half-open intervals, and which way they would be treated. For a RTL iterator over a region, it would be clearer to get the iterator and then do reversed() on it.

Let's take this discussion to #24 please, we're not going to cover the iteration over the trees in a genomic range here.

hyanwong:
> Basically we have to have the Tree class because that's how the higher level operations work.

Ah, I see. It still seems a bit confusing to me to have two differently defined iterators over a tree sequence. Could you reduce the burden of maintaining 2 iterator classes, and also allow the low level functions to be used in both cases, if
* ts.trees() returns an iterable
* iter(ts.trees()) is identical to tskit.Tree(ts) [and perhaps iter(ts.trees(start, stop)) ==  tskit.Tree(ts, start, stop)]
* Tree.__next__() is basically identical to Tree.next() only it raises StopIteration rather than returning the null state. 
* Tree.__iter__ returns Tree.self

I'm no expert on the most Pythonic way to do this, though. Perhaps my suggestion overloads too much onto a single class.

gtsambos:
I don't fully understand all the technical details in this thread, but am chiming in with a few feelings:

> * len(tree_sequence) returns the number of trees in tree sequence
> * tree_sequence[index] returns a new tree instance at given index
> * list(tree_sequence) returns a list of tree instances (inefficient, but handy)
> * reversed(tree_sequence.trees()) efficiently iterates backwards over trees
> * Manual method for iterating over trees like:
> 
> ...
> 
> * Manual seeking methods (currently inefficient, but can be implemented efficiently in C if we use the underlying edge indexes):
> 
> ...

All of these look sensible and useful to me!

> 1. `ts[99]` is useful, but often I'd want a tree at a specific genomic location, rather than a specific index. Not sure what is pythonic here, but if `ts[99]` is a shorthand for `ts.index(99)`, then maybe there should be a `ts.position()` too?

This is my typical situation too - I think it would be very useful to be able to extract the tree at a given genomic location in a straightforward way.

Also, I've made much use of existing "handy but inefficient" tools in tskit (eg. the `draw()` methods)! I find them particularly useful while developing/demonstrating, when I'm typically only working with small-scale examples anyway.

jeromekelleher:
> Ah, I see. It still seems a bit confusing to me to have two differently defined iterators over a tree sequence. Could you reduce the burden of maintaining 2 iterator classes, and also allow the low level functions to be used in both cases, if

No, that's not it. There's only one "iterator" (in the Python sense), ts.trees(). The Tree class is a state machine, and not a Python iterator in any sense. Would you mind reading the updated tutorial that discussed the low-level code and see if this clears things up please?

jeromekelleher:
> This is my typical situation too - I think it would be very useful to be able to extract the tree at a given genomic location in a straightforward way.

Thanks for the feedback @gtsambos, I think we'll add the ``ts.at(position)`` to return the tree covering a given position.

jeromekelleher:
I'm having second thoughts about enabling ``tree_sequence[index]``, mainly because of the following, easily made, performance gotcha:
```python
# Fast!!
for tree in ts.trees():
    # do stuff
# Slow!!
for tree in ts():
    # do stuff
```
We can't avoid this if we want to support ``list(ts)`` returning something sensible. I've often found myself writing ``for tree in ts`` by mistake, so I think it's a probably something that would happen quite easily and lead to nasty and hard to trace performance issues. After people's feedback it seems clear that it's a lot less useful to be able to get trees by index than genomic location anyway, so I'm not sure anything other than ``ts[0]`` and ``ts[-1]`` would ever be used.

So, what if we get rid of the ``ts[index]`` and add a ``ts.aslist()`` method which returns the list of tree objects? This would fulfil the handy-but-slow requirement, but couldn't be used accidentally.

hyanwong:
> I'm having second thoughts about enabling `tree_sequence[index]`,

Seems a sensible idea to remove the indexing operator for this reason. We can presumably do `ts.first()` and `ts.last()` anyway. Incidentally, we could presumably then switch to an efficient iterator version of `for tree in ts:, were it not for the fact that there's a huge amount now written that uses `ts.trees()`, and also the need for genomic ranges (#24)

jeromekelleher:
> Seems a sensible idea to remove the indexing operator for this reason. We can presumably do ts.first() and ts.last() anyway.

Exactly

> Incidentally, we could presumably then switch to an efficient iterator version of for tree in ts:, were it not for the fact that there's a huge amount now written that uses ts.trees()`, and also the need for genomic ranges (#24)

We could, but I think it's better to keep things simple, since we have to have the ``.trees()`` version anyway.

hyanwong:
> No, that's not it. There's only one "iterator" (in the Python sense), ts.trees(). ... Would you mind reading the updated tutorial that discussed the low-level code and see if this clears things up please?

I did read the low-level docs beforehand, so it's probably just my lack for formal compsci training that is confusing me. From the docs, I gather that iterating over `ts.trees()` returns a single Tree object that gets updated (i.e. a new one isn't generated) on each iteration. Similarly, you can go through the entire tree sequence by taking a single Tree object and updating the same object by calling `.next()` on it. The two sound functionally identical to my untrained mind, and the (wrong) idea that the `Tree` object serves dual purpose as an iterator over a tree sequence is reinforced for people coming from Python 2, where the as the default iteration function in Py2 really is called `.next()`

jeromekelleher:
> did read the low-level docs beforehand, so it's probably just my lack for formal compsci training that is confusing me. From the docs, I gather that iterating over ts.trees() returns a single Tree object that gets updated (i.e. a new one isn't generated) on each iteration. Similarly, you can go through the entire tree sequence by taking a single Tree object and updating the same object by calling .next() on it. The two sound functionally identical to my untrained mind, and the (wrong) idea that the Tree object serves dual purpose as an iterator over a tree sequence is reinforced for people coming from Python 2, where the as the default iteration function in Py2 really is called .next()

OK, thanks, good to know. I'll have another go at explaining all this after getting rid of ``__getitem__``, etc.

jeromekelleher:
OK, have updated this a bit. Main changes:

- Remove the sequence operations from TreeSequence (no len(ts) or ts[index] now)
- Add TreeSequence.at(position) and TreeSequence.at_index(index)
- Renamed low-level seek_position() to seek()---rationale is that we'll be using genomic locations mostly for these, so keeping the API for dealing with locations as simple as possible.
- Removed the low-level state-machine Tree API from the tutorial and tried to make the class documentation rigorous. It'll cause more confusion than it's worth to try explaining it, and 99% of users will never see it.

hyanwong:
> * Renamed low-level seek_position() to seek()

This seems sensible. You could presumbly have `seek_index()` in the same way that a ts has `at_index()`

jeromekelleher:
> This seems sensible. You could presumbly have seek_index() in the same way that a ts has at_index()

That's how I've done it, yeah --- it's nice to keep things symmetric.

jeromekelleher:
OK, looks like this is a good platform to build on. I'll try and get to #24 soon so we can be sure it all fits together. Thanks for the feeback @hyanwong, very helpful.

petrelharp:
This is wonderful.

PR #125, opened 2019-02-15
Parsimony algorithm.

### LINKED ISSUE #99: Method to position mutations to explain genotypes
We want to make tskit usable for a wide variety of applications. One potentially large group of users is the viral and bacterial genomics community. In these applications there will often be  good estimate of a tree for the samples. We want to be able to represent  data by combining such a tree (converted to tree sequence form using [tsconvert.from_newick](https://github.com/tskit-dev/tsconvert/blob/225224648ccb1794a0a7395de99a0ed5c9f28fe1/tsconvert/newick.py#L97)) with the observed sequences. 

The key algorithm that we need is a method to place mutations on the tree for a given site. Luckily this is a well-studied classical phylogenetics question. Here is a rough prototype of the Sankoff algorithm:
```python
import msprime
import tskit
import numpy as np

class Site(object):
    def __init__(self):
        self.ancestral_state = None
        self.mutations = []

class Mutation(object):
    def __init__(self, node=None, derived_state=None):
        self.node = node
        self.derived_state = derived_state


def sankoff(tree, genotypes, alleles, distance_matrix=None):
    """
    Returns a Site object explaining the specified set of genotypes
    on the specified tree.

    Based on treatment from Clemente et al., https://doi.org/10.1186/1471-2105-10-51
    """
    num_alleles = np.max(genotypes) + 1
    if distance_matrix is None:
        distance_matrix = np.ones((num_alleles, num_alleles))
        np.fill_diagonal(distance_matrix, 0)
    S = np.zeros((num_alleles, tree.num_nodes))
    infinity = 1e7  # Arbitrary big value
    # Initialise the weights
    for allele, u in zip(genotypes, tree.tree_sequence.samples()):
        S[:, u] = infinity
        S[allele, u] = 0
    for p in tree.nodes(order="postorder"):
        for i in range(num_alleles):
            for child in tree.children(p):
                min_w = infinity
                for j in range(num_alleles):
                    min_w = min(min_w, distance_matrix[i, j] + S[j, child])
                S[i, p] += min_w

    site = Site()
    S_anc = [None for _ in range(tree.num_nodes)]
    for x in tree.nodes(order="preorder"):
        S_anc[x] = np.argmin(S[:, x])
        i = S_anc[x]
        min_cost = infinity
        for j in range(num_alleles):
            if x == tree.root:
                trans_cost = S[j, x]
            else:
                trans_cost = distance_matrix[i, j] + S[j, x]
            if trans_cost < min_cost:
                min_cost = trans_cost
                S_anc[x] = j
        if x == tree.root:
            site.ancestral_state = alleles[S_anc[x]]
        elif S_anc[x] != S_anc[tree.parent(x)]:
            # TODO track the mutation parent. Should be straightforward enough.
            site.mutations.append(Mutation(node=x, derived_state=alleles[S_anc[x]]))
    return site

# Try it out.
ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
tree = ts.first()

print("num_sites = ", ts.num_sites)
tables = ts.dump_tables()
tables.sites.clear()
tables.mutations.clear()

# print(tree.draw(format="unicode"))
for variant in ts.variants():
    site = sankoff(tree, variant.genotypes, variant.alleles)
    site_id = tables.sites.add_row(variant.site.position, ancestral_state=site.ancestral_state)
        tables.mutations.add_row(
            site=site_id, node=mutation.node, derived_state=mutation.derived_state)

ts2 = tables.tree_sequence()
assert np.array_equal(ts.genotype_matrix(), ts2.genotype_matrix())
# For simple infinite sites mutations we should recover exactly.
assert tables.sites == ts.tables.sites
assert tables.mutations == ts.tables.mutations
```

Some questions:

1. Does this belong in tskit, or should it go somewhere else?
2. Assuming it does belong in tskit, what should we call it? (Presumably a method of the Tree class?)

Any thoughts much appreciated!

jeromekelleher:
Closes #99 

WIP.


jeromekelleher:
I think this is ready for merging, so it would be good to get some review. The documentation on the new ``reconstruct`` method should explain what this is for, so I won't repeat it here.

I'm not entirely sure the interface is right (is "reconstruct" the right word?), so any thoughts on this would be appreciated.

The implementation is  fast: finding a parsimonious explanation for an arbitrary set of genotypes for a tree with 100K samples takes < 0.01 seconds.

Pinging @hyanwong and @petrelharp for thoughts/review.

jeromekelleher:
TODO: this needs to handle missing data.

hyanwong:
> finding a parsimonious explanation for an arbitrary set of genotypes for a tree with 100K samples takes < 0.01 seconds.

Wow. That's seriously impressive. I will try to look through this, but may have more questions than answers.

jeromekelleher:
Just thought of an issue here with not having a consistent mapping of genotype value to base pair. It doesn't matter for Fitch, but for Sankoff we need a transition matrix where we need a consistent mapping from 0 -> "A", 1 -> "C" etc for the matrix to make sense. Options are:

1) Insist on a single cost matrix and make it the users responsibility of knowing how the alleles are encoded, with the correct cost matrix.
2) Pass a list of m cost matrices, one per site, which can be of different sizes. This allows to have different matrices per site, but is a bit cumbersome. I guess we could have a higher level description of the matrix, using the list of `m` allele lists to translate?

hyanwong:
Presumably most sites will have SNP-type variants, but some will be indels etc, which could be accounted for by having different cost matrices.

I assume by 2. you mean having a list of matrices, with each site containing an index into the list? In other words, it's a bit like an additional tskit table, the "cost" table, which gives both the allele encoding and the cost of transition.

jeromekelleher:
> Presumably most sites will have SNP-type variants, but some will be indels etc, which could be accounted for by having different cost matrices.

Yes, exactly; it would be nice to have this generality.

> I assume by 2. you mean having a list of matrices, with each site containing an index into the list? In other words, it's a bit like an additional tskit table, the "cost" table, which gives both the allele encoding and the cost of transition.

Yes, that's a good way of looking at it.

petrelharp:
I may not be totally following here, but: you are wondering what about when the entries in the genotype matrix aren't consistently coded?  This seems overly accomodating, and easy to deal with on the user's side. If all mutations are ACGT, then it is not hard to readjust the genotype matrix to be consistent, if this has not been done already. In fact, this seems easier than trying to wrangle the list of per-site cost matrices. If the user has more than one type of mutation, they could separate out the genotype sub-matrices corresponding to different cost matrices and pass those in.

petrelharp:
> I'm not entirely sure the interface is right (is "reconstruct" the right word?), so any thoughts on this would be appreciated.

"Reconstruct" makes me think you're reconstructing the *tree* maybe. I think the more usual terminology for this is ["mapping" the mutations](https://www.ncbi.nlm.nih.gov/pubmed/12396587) on the tree? Maybe the method should be called `.map_mutations`?

petrelharp:
For the future: it would be natural to make this a method of a tree sequence, and give also a vector of positions, so that this happens for each site on the tree for its site.

jeromekelleher:
Thanks @petrelharp, I like ``map_mutations``, lets go with that.

jeromekelleher:
This is ready to go I think. Review would be appreciated here @petrelharp and @hyanwong!

jeromekelleher:
I'd like to get this one squared away soon, if possible. @hyanwong, would you mind taking a look through please and see if you agree with the high-level API and docs? This will be very useful in tsinfer, as we can get rid of the slow and crappy code for dealing with non-inference sites. Also, it has the first steps for dealing with missing data (missing genotype is -1, and genotypes are now signed).

hyanwong:
Right - will make this a priority today & tomorrow. Have to go for 2 hrs now but will be on this when I return.

petrelharp:
alert @bhaller [vargen will be changing](https://www.google.com/search?client=firefox-b-1-d&q=tskit):
```
Change genotypes from unsigned to signed to accomodate missing data
  (see :issue:`144` for discussion). This only affects users of the
  ``tsk_vargen_t`` class.
```

bhaller:
> alert @bhaller [vargen will be changing](https://www.google.com/search?client=firefox-b-1-d&q=tskit):
> 
> ```
> Change genotypes from unsigned to signed to accomodate missing data
>   (see :issue:`144` for discussion). This only affects users of the
>   ``tsk_vargen_t`` class.
> ```

Thanks for the heads-up.  Whenever things are in a state where it makes sense to pull over the new code from tskit into SLiM, I can coordinate with you on that; up to you.  (Well, until August 5th or so, and then I'll be busy until mid-September.)

docs/tutorial.rst
```diff
@@ -664,3 +664,119 @@ Running this example we get::
     >>> threads_example()
     100%|████████████████████████████████████████████████| 4045/4045 [00:09<00:00, 440.29it/s]
     Found LD sites for 4045 doubleton mutations out of 60100
+
+
+.. _sec_tutorial_parsimony:
+
+*********
+Parismony
```
hyanwong: "Parsimony"

python/_tskitmodule.c
```diff
@@ -8244,6 +8323,9 @@ static PyMethodDef Tree_methods[] = {
     {"get_newick", (PyCFunction) Tree_get_newick,
             METH_VARARGS|METH_KEYWORDS,
             "Returns the newick representation of this tree." },
+    {"map_mutations", (PyCFunction) Tree_map_mutations,
+            METH_VARARGS|METH_KEYWORDS,
+            "Returns the minimal ancestral state reconstruction for the specified genotypes." },
```
hyanwong: Do you mean "most parsimonious" rather than "minimal"?

python/requirements/development.txt
```diff
@@ -18,3 +18,4 @@ pyparsing
 pysam
 PyVCF
 python_jsonschema_objects
+biopython
```
hyanwong: I can't see biopython used anywhere in the PR - so is this needed? OTOH, I may be missing something.

python/tskit/trees.py
```diff
@@ -1425,6 +1429,50 @@ def get_parent_dict(self):
...
+        distinct characters mapped to integers in the range 0 to 63,
+        such that ``genotypes[0]`` contains the observation for sample 0 and so on.
+        Missing data can be specified for a sample using the value
+        ``tskit.MISSING_DATA`` (-1). At least one non-missing observation must be
+        provided.
+
+        The current implementation uses the Fitch parsimony algorithm to determine
+        the minimum number of state transitions required to explain the data.
+
+        The state reconstruction is returned as two-tuple, ``(ancestral_state,
```
hyanwong: Do we want to use the word "root state" here, rather than "ancestral state". The ancestral state might be though of for each node, rather than a general property of the tree (i.e. you could say there is an ancestral state in front of any transition on the tree, which could be different from the state at the root)

docs/examples.py
```diff
@@ -73,5 +73,39 @@ def moving_along_tree_sequence():
         print("Tree {} covers [{:.2f}, {:.2f})".format(tree.index, *tree.interval))
 
 
+def parsimony():
+    tree = msprime.simulate(6, random_seed=42).first()
+    colours = ["red", "blue", "green"]
+    genotypes = [0, 0, 0, 0, 1, 2]
+    node_colours = {j: colours[g] for j, g in enumerate(genotypes)}
+    ancestral_state, transitions = tree.map_mutations(genotypes)
```
hyanwong: Similarly, "root_state" might be preferable?

docs/examples.py
```diff
@@ -73,5 +73,39 @@ def moving_along_tree_sequence():
...
+    for transition in transitions:
+        print("\t", transition)
+    tree.draw("_static/parsimony2.svg", node_colours=node_colours)
+
+    tree = msprime.simulate(6, random_seed=42).first()
+    colours = ["red", "blue", "white"]
+    genotypes = [1, -1, 0, 0, 0, 0]
+    node_colours = {j: colours[g] for j, g in enumerate(genotypes)}
+    ancestral_state, transitions = tree.map_mutations(genotypes)
+    print("Ancestral state = ", ancestral_state)
```
hyanwong: Ditto

docs/examples.py
```diff
@@ -73,5 +73,39 @@ def moving_along_tree_sequence():
...
+    for transition in transitions:
+        print("\t", transition)
+    tree.draw("_static/parsimony1.svg", node_colours=node_colours)
 
-moving_along_tree_sequence()
+    tree = msprime.simulate(6, random_seed=42).first()
+    colours = ["red", "blue", "green", "white"]
+    genotypes = [-1, 0, 0, 0, 1, 2]
+    node_colours = {j: colours[g] for j, g in enumerate(genotypes)}
+    ancestral_state, transitions = tree.map_mutations(genotypes)
```
hyanwong: ditto

python/requirements/development.txt
jeromekelleher: I'm using it in the ``tests/test_parsimony.py`` to compare the Sankoff distances we compute with the Biopython version. It's a good sanity check.

python/tskit/trees.py
jeromekelleher: It's a good point, but we're using the term ``ancestral_state`` quite specifically in the library to mean the state inherited by all roots. I've changed the sentance to read 
```
The state reconstruction is returned as two-tuple, ``(ancestral_state,
        transitions)``, where ancestral_state is the state that
        has been assigned to the tree root(s).
```

This will hopefully clear things up a bit. I don't want to change the name to ``root_state``, because the link with ``Site.ancestral_state`` is deliberate and meaninful.

python/tskit/trees.py
```diff
@@ -1425,6 +1429,50 @@ def get_parent_dict(self):
...
+        list of ``tskit.StateTransition`` objects, which correspond to
+        mutations in the :ref:`sec_data_model`. Each element in this list
+        in an object with attributes ``node``, ``parent`` and ``state``,
+        which alternatively can be treated as a 3-tuple (it is implemented using
+        :func:`collections.namedtuple`). For each transition, ``node`` is the
+        tree node over which state transition occurs and ``state`` is the new
+        after this transition. When there are multiple transitions at returned,
+        ``parent`` contains the index of the previous state transition on the path
+        to root (see the :ref:`sec_mutation_table_definition` for more information
+        on the concept of mutation parents).
```
petrelharp: Misisng word: "and ``state`` is the new **state** after this transition".

Also, the mutation being "over" a node is a bit ambiguous, how about "``node`` is the tree node at the bottom of the branch on which the transition occurred"?

python/tskit/trees.py
petrelharp: What happens if the tree has more than one root?

python/requirements/development.txt
hyanwong: Ah - github just didn't load test_parsimony in my web browser window, so I didn't find it on initial search

c/tskit/trees.c
```diff
@@ -3639,6 +3639,191 @@ tsk_tree_prev(tsk_tree_t *self)
...
+                    stack_top++;
+                    stack[stack_top] = v;
+                }
+            } else {
+                stack_top--;
+                postorder_parent = parent[u];
+
+                /* Visit u */
+                if (! (node_flags[u] & TSK_NODE_IS_SAMPLE)) {
+                    /* Get the union of the child sets */
```
petrelharp: you mean "intersection"; and in line n+6 switch 'union' and 'intersection'.

petrelharp:
Looks great!

jeromekelleher:
Thanks @petrelharp! Quick question: after discussing this with Yan, I think it's better if we change the API to

```python
def map_mutations(self, genotypes, alleles):
      # do stuff
      return ancestral_state, mutations
```
Where alleles is a list of strings, and we now return the ancestral state as a string and return a list actual mutation objects rather than ``StateTransitions``. So, we map **exactly** to the site/mutation data model, rather than the slightly different integer encodings, which seems confusing to me now.

An advantage of this is that the ``cost_matrix`` we want to pass in then can either be a 2D numpy array mapping to the integer encodings of the alleles, OR a dictionary mappying allele pairs to costs. So, we could have:

```
alleles = ["A", "T"]
costs = {("A", "T"): 0.1, ("T", "A"): 0.2)}
# or 
costs = [
    [0, 0.1],
    [0.2, 0]]
```
This would be nice and general I think, and should be reasonably intuitive in either direction.

Any objections?

petrelharp:
That sounds much nicer!  I'm not sure that specifying a dictionary mapping of pairs is actually useful, though, unless you're thinking of cases where some transitions are not possible.

jeromekelleher:
> That sounds much nicer! I'm not sure that specifying a dictionary mapping of pairs is actually useful, though, unless you're thinking of cases where some transitions are not possible.

Great, I'll do that and merge. The advantage of the dictionary mapping pairs that the encoding 0->"A", 1->"C" etc can change from site to site (as it does with tskit's output). So, we can specify a single cost mapping for all sites without having to worry about permuting the numpy matrix each site. Anyway, we don't have to worry about that for now - the door is open to both approaches.

hyanwong:
> So, we can specify a single cost mapping for all sites without having to worry about permuting the numpy matrix each site.

This is a big advantage, in my book.

c/tskit/trees.c
jeromekelleher: Hah, yes, so I did!

PR #139, opened 2019-02-28
Performance tuning on trees

jeromekelleher:


jeromekelleher:
cc @molpopgen. I fixed the redundant code issue you mentioned, and while I was in there added some low-hanging performance tweaks. Turns out adding the restrict pointers basically makes no difference (annoyingly).

It looks like most of the time is being spent here:
```
       │         while (out >= 0 && out < num_edges && out_breakpoints[out_order[out]] == x) {                                                                                                                                                
  0.09 │ e0:   mov    0x48(%rsp),%eax                                                                                                                                                                                                         
  0.37 │       mov    0x1c(%rsp),%ebx                                                                                                                                                                                                         
       │       cmp    %eax,%ebx                                                                                                                                                                                                               
  0.28 │       mov    %ebx,%eax                                                                                                                                                                                                               
  0.09 │       not    %eax                                                                                                                                                                                                                    
  0.18 │       setl   %dl                                                                                                                                                                                                                     
       │       shr    $0x1f,%eax                                                                                                                                                                                                              
  0.37 │       and    %dl,%al                                                                                                                                                                                                                 
       │     ↓ je     4a1                                                                                                                                                                                                                     
  0.28 │       movslq %ebx,%rcx                                                                                                                                                                                                               
  0.09 │       mov    0x68(%rsp),%rbx                                                                                                                                                                                                         
  0.18 │       movslq (%rbx,%rcx,4),%rdx                                                                                                                                                                                                      
  0.18 │       mov    0x60(%rsp),%rbx                                                                                                                                                                                                         
 11.31 │       ucomis (%rbx,%rdx,8),%xmm0                                                             
```

which suggests that the majority of the time is being spent fetching edge information from memory, and the time spending updating the actual tree arrays is small. Probably the way to go to making this faster is to have two lists of edges in the correct orders in memory, which we traverse directly rather than through the index orders. This adds a bit of memory overhead and init time, but it should pay back with significantly faster tree update times. (This is what I'm doing in tsinfer, if I remember correctly.)

Anyway, not going to do this now, but interesting to think about for future reference.

molpopgen:
Yeah, it is impressive how much `x[y[i]]` can degrade performance.  In fwdpp, I keep a vector of `position, parent, child)` tuples, which I did based on our discussions a while back plus some profiling.

PR #140, opened 2019-02-28
Drop support for legacy Python 2.7

### LINKED ISSUE #137: Schedule for dropping Python 2.7 support
Python 2.7 will be officially [retired](https://pythonclock.org/) in less than a year. Top level package like numpy, pandas etc are also dropping 2.7 support by [2020](https://python3statement.org/). There seems little point in us keeping 2.7 support after that point. It's getting increasingly annoying to keep Python 2.7 features lying around, so the sooner we drop it the better IMO.

Is there any reason to *not* drop Python 2.7 support within the next few months @tskit-dev/all?

hugovk:
Fixes #137.

And upgrade Python syntax using https://github.com/asottile/pyupgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ language: python
 matrix:
   include:
     - os: linux
-      python: 2.7
-    - os: linux
       python: 3.6
     - os: osx
       language: generic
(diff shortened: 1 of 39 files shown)


petrelharp:
Gee, thanks for that @hugovk. I've looked through the changes and don't see anything wrong. Others with more experience should have a look, too, though.

jeromekelleher:
This is AMAZING, thank you so much @hugovk! I'll need to spend a bit of time going through this, but we'll definitely use it. You've made the decision to push this ahead a **lot** easier!

hugovk:
You're welcome!

It might be easier to check commit by commit.

jeromekelleher:
Merged: HOORAY!!!!!! Thanks so much again @hugovk, this was such an amazingly great surprise for you to help us out like this!

PR #143, opened 2019-03-06
Retain unary nodes with simplify()

gtsambos:
simplify() now takes a filter_unary argument (default: True). If False, all unary nodes are retained in the simplification.

The Python implementation in `python/tests/simplify.py` is now more or less finished; the tests in `python/tests/test_topology.py` are still under heavy development.

### DIFF
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -98,7 +98,7 @@ class Simplifier(object):
     """
     def __init__(
             self, ts, sample, reduce_to_site_topology=False, filter_sites=True,
-            filter_populations=True, filter_individuals=True):
+            filter_populations=True, filter_individuals=True, filter_unary=True):
         self.ts = ts
         self.n = len(sample)
         self.reduce_to_site_topology = reduce_to_site_topology
@@ -106,6 +106,7 @@ class Simplifier(object):
         self.filter_sites = filter_sites
         self.filter_populations = filter_populations
         self.filter_individuals = filter_individuals
+        self.filter_unary = filter_unary
         self.num_mutations = ts.num_mutations
         self.input_sites = list(ts.sites())
         self.A_head = [None for _ in range(ts.num_nodes)]
@@ -244,6 +245,7 @@ class Simplifier(object):
         All ancestry segments in S come together into a new parent.
         The new parent must be assigned and any overlapping segments coalesced.
         """
+
         output_id = self.node_id_map[input_id]
         is_sample = output_id != -1
         if is_sample:
@@ -260,6 +262,11 @@ class Simplifier(object):
                 if is_sample:
                     self.record_edge(left, right, output_id, ancestry_node)
                     ancestry_node = output_id
+                elif not self.filter_unary:
+                    if output_id == -1:
+                        output_id = self.record_node(input_id)
+                    self.record_edge(left, right, output_id, ancestry_node)
+
             else:
                 if output_id == -1:
                     output_id = self.record_node(input_id)
@@ -269,7 +276,11 @@ class Simplifier(object):
             if is_sample and left != prev_right:
                 # Fill in any gaps in the ancestry for the sample
                 self.add_ancestry(input_id, prev_right, left, output_id)
-            self.add_ancestry(input_id, left, right, ancestry_node)
+
+            if self.filter_unary:
+                self.add_ancestry(input_id, left, right, ancestry_node)
+            else:
+                self.add_ancestry(input_id, left, right, output_id)
             prev_right = right
 
         if is_sample and prev_right != self.sequence_length:
@@ -291,7 +302,8 @@ class Simplifier(object):
             x = self.A_head[edge.child]
             while x is not None:
                 if x.right > edge.left and edge.right > x.left:
-                    y = Segment(max(x.left, edge.left), min(x.right, edge.right), x.node)
+                    y = Segment(
+                        max(x.left, edge.left), min(x.right, edge.right), x.node)
                     S.append(y)
                 x = x.next
         self.merge_labeled_ancestors(S, parent)
@@ -408,6 +420,7 @@ class Simplifier(object):
             edges = all_edges[:1]
             for e in all_edges[1:]:
                 if e.parent != edges[0].parent:
+                    # print('input_id:', edges[0].parent)
                     self.process_parent_edges(edges)
                     edges = []
                 edges.append(e)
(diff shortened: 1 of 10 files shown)


python/tests/simplify.py
```diff
@@ -295,7 +309,11 @@ def process_parent_edges(self, edges):
             x = self.A_head[edge.child]
             while x is not None:
                 if x.right > edge.left and edge.right > x.left:
-                    y = Segment(max(x.left, edge.left), min(x.right, edge.right), x.node)
+                    if self.filter_unary:
+                        y = Segment(max(x.left, edge.left), min(x.right, edge.right), x.node)
+                    else:
+                        y = Segment(max(x.left, edge.left), min(x.right, edge.right),\
```
jeromekelleher: The recommended way to split this line is to use "hanging indents":
```python
y = Segment(
    max(x.left, edge.left), min(x.right, edge.right), 
    self.node_id_map[edge.child])
```

gtsambos:
Cool, thanks @jeromekelleher! Just to clarify: should flake8 return no warnings/errors at all? I'm only checking for line length, like this:
```
flake8 --max-line-length 89 tests/test_topology.py 
```

jeromekelleher:
> Cool, thanks @jeromekelleher! Just to clarify: should flake8 return no warnings/errors at all? I'm only checking for line length, like this:

Yep, flake8 should be perfectly happy with the code. It's useful to add the check to your git [pre-commit hook](https://githooks.com/). Here's the last few lines of my pre-commit:
```
# Bunch of other stuff from default hook.

# NOTE: the flake8 check MUST come before the git diff-index bit below!
exec python3 -m flake8 --max-line-length=89 python/tskit python/tests

# If there are whitespace errors, print the offending file names and fail.
exec git diff-index --check --cached $against --
```

python/tests/test_topology.py
```diff
@@ -2894,12 +2890,15 @@ def test_small_tree_filter_zero_mutations(self):
...
         self.assertEqual(ts.num_sites, 8)
         self.assertEqual(ts.num_mutations, 8)
-        tss, _ = self.do_simplify(ts, [4, 0, 1], filter_sites=True)
-        self.assertEqual(tss.num_sites, 5)
-        self.assertEqual(tss.num_mutations, 5)
-        tss, _ = self.do_simplify(ts, [4, 0, 1], filter_sites=False)
-        self.assertEqual(tss.num_sites, 8)
-        self.assertEqual(tss.num_mutations, 5)
+        for filt in [True, False]:
+            tss, _ = self.do_simplify(ts, [4, 0, 1], filter_sites=True,
```
jeromekelleher: This is a matter of taste, but I prefer to use hanging indents for this sort of thing:
```python
tss, _ = self.do_simplify(
    ts, [4, 0, 1], filter_sites=True, filter_unary=filt)
```
The reason I prefer hanging indents is that it's much more robust to code changes. For example, if we changed the name of the function from ``do_simplify`` to ``run_simplify``, we'd have to fix all the visual indents as well. If it was a hanging indent then no changes would be needed than a global search and replace.

For a large code-base like this, it's not a case of "if code changes are made" but "when"...

python/tests/simplify.py
```diff
@@ -156,6 +158,7 @@ def rewind_node(self, input_id, output_id):
         Remove the mapping for the specified input and output node pair. This is
         done because there are no edges referring to the node.
         """
+        print('rewind_node was called')
```
jeromekelleher: Stray print here.

python/tests/test_topology.py
gtsambos: Cool, fixed!

python/tests/simplify.py
gtsambos: Thanks. These were removed in the latest commit.

python/tests/simplify.py
```diff
@@ -291,7 +302,13 @@ def process_parent_edges(self, edges):
...
             while x is not None:
                 if x.right > edge.left and edge.right > x.left:
-                    y = Segment(max(x.left, edge.left), min(x.right, edge.right), x.node)
+                    if self.filter_unary:
+                        y = Segment(
+                            max(x.left, edge.left), min(x.right, edge.right), x.node)
+                    else:
+                        y = Segment(
+                            max(x.left, edge.left), min(x.right, edge.right),
+                            self.node_id_map[edge.child])
```
petrelharp: I don't see why this is working - could you explain? How do we know that `node_id_map` has been defined already for `edge.child`? Why do we need a different case for `filter_unary` and not?

gtsambos: > Why do we need a different case for filter_unary and not?

After thinking about it for a bit, I realised that we don't, so I've removed this complication in the latest commit. In a previous version I was using the priority queue of segments in a slightly different way, and the distinction was necessary then. The two cases should produce the same output now.

thanks for thinking about this so carefully @petrelharp!

jeromekelleher:
OK, I think this is ready to go! Thanks for some really excellent work @gtsambos, everything worked perfectly. The only change I had to make is to rename the feature ``keep_unary``, which was forced by the way the C library is set up (the default needed to be "no flags", and this wasn't possible if the flag was FILTER_UNARY --- I should have thought of this originally). I've pushed through all the changes needed including documentation, which may be a useful reference later if you're adding other features to simplify.

@petrelharp, would you mind taking one last look and merging if you're happy?

jeromekelleher:
Bumping this; @petrelharp, are you happy for us to merge?

gtsambos:
Hey @jeromekelleher, when I run the tests in `python/tests/test_topology` with your latest commit, I get a lot of errors. I think it might be something to do with [this line](https://github.com/tskit-dev/tskit/pull/143/files#diff-a89cffb041b1b7c33e7be6b595f785a3R1647), as there's no `map_nodes` argument in the Simplifier class. I'll investigate further and let you know.

gtsambos:
It looks to be something to do with the inputs you're allowed to make to the implementations of simplify in the other libraries that are compared with `compare_lib`. 
Here's an example error:
```
6200D-132482-M:python gtsambos$ python -m nose -vs tests/test_topology.py:TestSimplify.test_single_tree
test_single_tree (tests.test_topology.TestSimplify) ... ERROR

ERROR: test_single_tree (tests.test_topology.TestSimplify)
  File "/Users/gtsambos/Projects/tskit/python/tests/test_topology.py", line 2736, in test_single_tree
    self.verify_no_samples(ts)
  File "/Users/gtsambos/Projects/tskit/python/tests/test_topology.py", line 2676, in verify_no_samples
    ts, samples=ts.samples(), keep_unary=keep_unary)
  File "/Users/gtsambos/Projects/tskit/python/tests/test_topology.py", line 2632, in do_simplify
    map_nodes=True)
  File "/Users/gtsambos/Projects/tskit/python/tskit/trees.py", line 2968, in simplify
    keep_unary=keep_unary)
  File "/Users/gtsambos/Projects/tskit/python/tskit/tables.py", line 1647, in simplify
    keep_unary=keep_unary)
TypeError: function takes at most 5 arguments (6 given)

```

jeromekelleher:
The issue here is that your low-level C module is out of date and needs to be rebuilt @gtsambos, as I've updated the C code and this isn't compiled automatically. Try running 'make'; this should recompile everything and resolve the problems. (You might need to 'make clean' first).

gtsambos:
Ah, right 😅

jeromekelleher:
Merging this; thanks @gtsambos!

PR #151, opened 2019-03-18
Add gnn docs

hyanwong:


hyanwong:
Cocked up the squashing again. I'll submit this through a different PR

Issue #149, opened 2019-03-18
Draw edges in different colours on a tree


hyanwong:
At the moment we have `tree.draw(node_colours={})`, but I would like to be able to colour the edge above a node too. Are there any objections to me submitting something like this as a PR?

jeromekelleher:
That would be a nice feature, yeah. It would be good to think about how to do this in both unicode and SVG outputs though, but we'd probably need another dependency to do it properly with unicode (and it might get tricky). It'd be enough for now to just make sure that any new API is at least compatible with both cases, even if we only implement one.

hyanwong:
For colouring in text (would we want to colour the text for nodes too, using node_colours?), I guess we would use [colorama](https://pypi.org/project/colorama/). Would it be reasonable to assume that if colours _and_ unicode were used we would be happy to output ANSI escape character sequences into the text output, without warning the user?

jeromekelleher:
Sorry, I forgot we already had ``node_colours`` for SVG,  so ``edge_colours`` is a straightforward extension. Let's not worry about colour for the unicode trees for now (I remember looking into this, and it's a bit tricky; also adding new dependencies is something to be done with care).

hyanwong:
OK. I think `edge_colours` should be the colour of an edge *above a node*, - i.e. the map should be indexed by node numbers, not edge numbers. So should the parameter be called "edge_colours_above_node", or is it obvious?

jeromekelleher:
> OK. I think edge_colours should be the colour of an edge above a node

Agreed. I think the meaning is obvious all right.

hyanwong:
I think that for any colour specification (i.e. for nodes, edges, or mutations), specifying a colour of `None` should result in the appropriate item not being drawn. Can I alter the behaviour to do that? It's a way of being able to draw trees without e.g. the circles on the nodes (specify `node_colour = {n:None for n in tree.nodes()}`)

jeromekelleher:
> I think that for any colour specification (i.e. for nodes, edges, or mutations), specifying a colour of None should result in the appropriate item not being drawn. Can I alter the behaviour to do that? It's a way of being able to draw trees without e.g. the circles on the nodes (specify node_colour = {n:None for n in tree.nodes()})

This is different to what's done for labels, where ``None`` means the default and having a key missing means to not write the label at all. We can just replicate this behaviour for colours, right?

hyanwong:
[...]

PR #153, opened 2019-03-18
Add edge colouring

hyanwong:
And also add documentation for mutation colours etc.

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -37,7 +37,7 @@ NULL_NODE = -1
 
 def draw_tree(
         tree, width=None, height=None, node_labels=None, node_colours=None,
-        mutation_labels=None, mutation_colours=None, format=None):
+        mutation_labels=None, mutation_colours=None, format=None, edge_colours=None):
     # See tree.draw() for documentation on these arguments.
     if format is None:
         format = "SVG"
@@ -67,7 +67,8 @@ def draw_tree(
     td = cls(
         tree, width=width, height=height,
         node_labels=node_labels, node_colours=node_colours,
-        mutation_labels=mutation_labels, mutation_colours=mutation_colours)
+        mutation_labels=mutation_labels, mutation_colours=mutation_colours,
+        edge_colours=edge_colours)
     return td.draw()
 
 
@@ -89,7 +90,7 @@ class TreeDrawer(object):
 
     def __init__(
             self, tree, width=None, height=None, node_labels=None, node_colours=None,
-            mutation_labels=None, mutation_colours=None):
+            mutation_labels=None, mutation_colours=None, edge_colours=None):
         self._tree = tree
         self._num_leaves = len(list(tree.leaves()))
         self._width = width
@@ -100,6 +101,7 @@ class TreeDrawer(object):
         self._node_colours = {}
         self._mutation_labels = {}
         self._mutation_colours = {}
+        self._edge_colours = {}
 
         # Set the node labels and colours.
         for u in tree.nodes():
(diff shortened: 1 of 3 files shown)


hyanwong:
closes https://github.com/tskit-dev/tskit/issues/149

python/tskit/drawing.py
```diff
@@ -37,7 +37,7 @@
 
 def draw_tree(
         tree, width=None, height=None, node_labels=None, node_colours=None,
-        mutation_labels=None, mutation_colours=None, format=None):
+        edge_colours=None, mutation_labels=None, mutation_colours=None, format=None):
```
jeromekelleher: New parameters should go at the end in case any users are using these as position parameters.

python/tskit/trees.py
```diff
@@ -1004,8 +1004,17 @@ def draw(
         :param map node_labels: If specified, show custom labels for the nodes
             that are present in the map. Any nodes not specified in the map will
             not have a node label.
-        :param map node_colours: If specified, show custom colours for nodes. (Only
-            supported in the SVG format.)
+        :param map node_colours: If specified, show custom colours for the nodes
+            given in the map. (Only supported in the SVG format.)
+        :param map edge_colours: If specified, show custom colours for the edge
```
jeromekelleher: I think we should follow the same semantics as ``node_labels``: if the node isn't present, don't draw the edge, if it maps to None, use the default colour. I think the description would be clearer as "If specified, show custom colours for the edge joining each node in the map to its parent".

python/tskit/drawing.py
```diff
@@ -229,8 +233,12 @@ def draw(self):
                 labels.add(dwg.text(self._node_labels[u], (x[0] + dx, x[1] + dy)))
             if self._tree.parent(u) != NULL_NODE:
                 y = self._x_coords[v], self._y_coords[v]
-                lines.add(dwg.line(x, (x[0], y[1])))
-                lines.add(dwg.line((x[0], y[1]), y))
+                if self._edge_colours.get(u, None) is not None:
```
jeromekelleher: What does ``stroke=None`` do here if we let it go through? It would be nice to avoid the condition here, no point in doing fiddly drawing code twice.

python/tskit/drawing.py
hyanwong: Good point

python/tskit/trees.py
hyanwong: Re semantics, see discussion at 
https://github.com/tskit-dev/tskit/issues/149#issuecomment-474246468 . I will change the description accordingly

python/tskit/drawing.py
hyanwong: I'm trying to avoid producing an SVG with looks of stroke="black" commands as the default. I could code it up to create a dict `params = {}` or `params = {'stroke':'black'}`, then call `dwg.line(..., **params)` only once, if you prefer.

hyanwong:
I'm happy to squash this into a single commit once you have OKed it.

python/tskit/drawing.py
```diff
@@ -202,17 +207,25 @@ def draw(self):
...
         for u in self._tree.nodes():
             v = self._tree.get_parent(u)
             x = self._x_coords[u], self._y_coords[u]
-            colour = "black"
-            if self._node_colours.get(u, None) is not None:
-                colour = self._node_colours[u]
-            dwg.add(dwg.circle(center=x, r=3, fill=colour))
+            fill = self._node_colours.get(u, default_node_colour)
+            if fill is not None:
+                if fill == default_node_colour:
```
jeromekelleher: Might help to have a quick one-line comment here to explain why we're doing this. (I assume this is to minimise the size of the SVG, right?) A more compact way of expressing might be:
```
params = {} if fill == default_node_colour else {'fill': fill}

jeromekelleher: Very neat way of doing this BTW, I like it a lot.

hyanwong: Yes, that's right. Makes the SVG much cleaner. Happy to go with the one-line if-else. I assumed you didn't like these sort of python-esque contractions.

jeromekelleher:
Perfect, thanks @hyanwong!

Issue #167, opened 2019-04-05
Extra parameter to tree.draw() to scale to height of entire ts


hyanwong:
When drawing multiple trees from a ts, we might want the branch lengths to be comparable. AT the moment, trees are drawn to 100% of the height of the tree, not of the entire tree sequence.

I'm happy to code this up in a PR

jeromekelleher:
Good idea. This will make drawing a full tree sequence much easier. Something like ``global_height_scale=False`` as an extra parameter?

hyanwong:
I've coded it as `scale_to_ts=False`. But perhaps your name is better. Not quite sure how to write the test case - presumably we need a ts for testing that has different root heights for a few trees - we might have to invent one.

hyanwong:
Currently here: https://github.com/hyanwong/tskit/commit/6944b88c65a3fd02a646236e7835a73f669af8e1

Will create a PR once I've thought of the unit test

jeromekelleher:
Can you make the PR, it's easier to discuss then as we can see the diff clearly.

PR #168, opened 2019-04-05
Draw tree height flag

hyanwong:
Closes https://github.com/tskit-dev/tskit/issues/167


jeromekelleher:
Yeah, basically looks good, thanks. The name isn't great though (scale what to what aspect of the ts?).

Does this work for text trees as well?

hyanwong:
I haven't implemented it for text trees (height is probably difficult here).

Agree about the name. Actually, I can imagine cases where we want to compare 2 different tree sequences, and want to fix the scaling to be the same in both cases. So perhaps we should not just allow boolean values, but also allow floats, which we interpret as the deepest scaling point on the tree?

hyanwong:
So perhaps `height_scale="tree"` (default),  `height_scale="ts"` (an alternative), `height_scale=1234` (have a user-chosen number for the scale).

Not sure I like using strings like this, so other suggestions welcome.

hyanwong:
I've changed the param name to `max_timescale` - better suggestions welcome.

The docs refer to the `oldest_time` property, so may need changing depending on https://github.com/tskit-dev/tskit/pull/170

python/tskit/trees.py
```diff
@@ -1022,14 +1022,19 @@ def draw(
             joining each node in the map to its parent. As for ``node_colours``,
             unspecified edges take the default colour, and ``None`` values result in the
             edge being omitted. (Only supported in the SVG format.)
+        :param map max_timescale: If None (the default), scale the tree so that it fits
```
jeromekelleher: the type is float not map. I think we can call this ``time_scale``.

The scaling used to map tree sequence time values into image y-coordinates. If None or unspecified, this is set to the time of the oldest root in the tree.

jeromekelleher:
I think ``time_scale`` might be good, but I've realised there's a tricky issue. In some cases we might want to draw the tree in a 'compact' form, there the nodes aren't scaled by time at all but spaced equally along the y-dimension. This is what the text format does. How do we parameterise this so that we can support all these different usages in a consistent way?

Maybe strings **are** a good idea. What about something like, ``y_scale`` which is one of "tree_time", "tree_node", "ts_time", "ts_node"? For SVG it defaults to "tree_time" and text it defaults to "tree_node". The constant names aren't great, but hopefully ``y_scale`` is intuitive.

hyanwong:
> Maybe strings **are** a good idea. What about something like, `y_scale` which is one of "tree_time", "tree_node", "ts_time", "ts_node"? For SVG it defaults to "tree_time" and text it defaults to "tree_node". The constant names aren't great, but hopefully `y_scale` is intuitive.

I'm happy with this (and to code it up), but for our use case it would also be useful to be able to scale 2 tree sequences to the same height, which is why I suggested a numeric parameter. We could allow both numeric *and* string, which you might find a little ugly, but is reasonably clear if labelled as "y_scale", I think. Otherwise we could have another (numeric) parameter if `y_scale` is set to e.g. `user`, but that's also a bit clunky IMO. Let me know which you prefer.

jeromekelleher:
Excellent point @hyanwong, allowing a numeric value is also useful. I like your plan, let's go with that please. So, we have a new parameter ``y_scale`` which:

- If is equal to the string "tree_time" will set the y scale such that the oldest node in the tree is at the 'top' of the output image (any better way to say this?) and node positions are scaled relative to time. Default for SVG
- If it is ``tree_node`` will set the y scale such that nodes are equally spaced on the y axis (i.e., times are ignored) and the oldest node is at the top of the image (Default for text).
- If it's ``ts_node``, same as tree_node except we place nodes according to their global position in the tree sequence.
- If ``ts_time``, same as tree_time, except the time is scaled by the oldest node in the tree sequence rather than the oldest node in the current tree.
- If the value is not one of these strings, it must be a numeric value which is used to set the maximum time scale. That is, ``tree.draw(y_scale="tree_time")`` is equivalant to ``tree.draw(y_scale=max(tree.time(root) for root in tree.roots))``.

We don't have to implement all of these initially, but we should raise ValueErrors if choosing one that's currently unsupported so that people can't depend on faulty semantics.

How does this sound?

jeromekelleher:
This is exciting BTW --- ``TreeSequence.draw()`` should be pretty straightforward when we have these tools. Not having one has been been bugging me for *ages*!

hyanwong:
> Excellent point @hyanwong, allowing a numeric value is also useful. I like your plan, let's go with that please.

Right - this all sounds perfect. The only slight adjustment would be that I would have `ts_time` scaled to be the oldest root in the tree sequence (i.e. as returned by our new `max_root_time()` function)

hyanwong:
Also a minor thing, but are we ever likely to want to plot the trees rotated by 90 degrees? If so, `y_scale` might not be the best choice of name (although currently I haven't any other suggestion)

jeromekelleher:
OK, here's a rough draft that does what we want with SVG. This code gives:

```python

import msprime

ts = msprime.simulate(10, recombination_rate=0.1, random_seed=2)

for tree in ts.trees():
    print(tree.draw(format="unicode"))
    tree.draw(f"tree_level_{tree.index}.svg", y_scale="ts_level")
    tree.draw(f"tree_time_{tree.index}.svg", y_scale="ts_time")
```

Scaled by tree sequence time = 
tree_time_0
tree_time_1

Scaled by tree sequence level=
tree_level_0
tree_level_1

jeromekelleher:
The code is very messy at the moment, I'll clean it up before and add ts scaling for text trees as well before committing.

> Also a minor thing, but are we ever likely to want to plot the trees rotated by 90 degrees? If so, y_scale might not be the best choice of name (although currently I haven't any other suggestion)

Hmm. Good point. I'm not convinced about the API at the moment, it's quite clumsy, maybe there's a better way to do it.

hyanwong:
> The code is very messy at the moment, I'll clean it up before and add ts scaling for text trees as well before committing.

Ah - I didn't realise that you had code to do this already. Great. Is there anything more you need me to do for this PR?

jeromekelleher:
> Ah - I didn't realise that you had code to do this already. Great. Is there anything more you need me to 
do for this PR?

I'll push it through from here thanks. I'll ping you for a review though when it's ready.

jeromekelleher:
I've ended up adding two new arguments @hyanwong, ``tree_height_scale`` and ``max_tree_height``. Hopefully the logic will be clear from the documentation. I decided on 'tree_height' rather than 'y' because of the points you brought up about rendering these horizontally at some point.

For the text engine, we can now do this:
```python
ts = msprime.simulate(10, random_seed=2, recombination_rate=0.2)

trees = [tree.draw(format="unicode", max_tree_height="ts").splitlines() for tree in ts.trees()]
for j in range(len(trees[0])):
    line = " | ".join(tree[j] for tree in trees)
    print(line)
```
giving
```
                      |                       |                       |          21
                      |                       |                       |     ┏━━━━┻━━━━┓
                      |       20              |                       |     ┃         ┃
                      |  ┏━━━━┻━━━┓           |                       |     ┃         ┃
       19             |  ┃        19          |       19              |     ┃         19
  ┏━━━━┻━━━━┓         |  ┃    ┏━━━┻━━━┓       |   ┏━━━┻━━━┓           |     ┃       ┏━┻━━┓
  ┃         18        |  ┃    ┃       18      |   ┃       18          |     ┃       ┃    ┃
  ┃     ┏━━━┻━━┓      |  ┃    ┃     ┏━┻━━┓    |   ┃    ┏━━┻━━━┓       |     ┃       ┃    ┃
  ┃     ┃      ┃      |  ┃    ┃     ┃    ┃    |   ┃    ┃      17      |     17      ┃    ┃
  ┃     ┃      ┃      |  ┃    ┃     ┃    ┃    |   ┃    ┃    ┏━┻━━┓    |   ┏━┻━━┓    ┃    ┃
  ┃     ┃      16     |  ┃    ┃     ┃    ┃    |   ┃    ┃    ┃    ┃    |   ┃    ┃    ┃    ┃
  ┃     ┃    ┏━┻━┓    |  ┃    ┃     ┃    ┃    |   ┃    ┃    ┃    ┃    |   ┃    ┃    ┃    ┃
  ┃     ┃    ┃   15   |  15   ┃     ┃    ┃    |   ┃    ┃    ┃    15   |   ┃    15   ┃    ┃
  ┃     ┃    ┃  ┏┻┓   | ┏┻┓   ┃     ┃    ┃    |   ┃    ┃    ┃   ┏┻┓   |   ┃   ┏┻┓   ┃    ┃
  ┃     14   ┃  ┃ ┃   | ┃ ┃   ┃     14   ┃    |   ┃    ┃    14  ┃ ┃   |   14  ┃ ┃   ┃    ┃
  ┃   ┏━┻┓   ┃  ┃ ┃   | ┃ ┃   ┃   ┏━┻┓   ┃    |   ┃    ┃  ┏━┻┓  ┃ ┃   | ┏━┻┓  ┃ ┃   ┃    ┃
  13  ┃  ┃   ┃  ┃ ┃   | ┃ ┃   13  ┃  ┃   ┃    |   13   ┃  ┃  ┃  ┃ ┃   | ┃  ┃  ┃ ┃   13   ┃
┏━┻┓  ┃  ┃   ┃  ┃ ┃   | ┃ ┃ ┏━┻┓  ┃  ┃   ┃    | ┏━┻┓   ┃  ┃  ┃  ┃ ┃   | ┃  ┃  ┃ ┃ ┏━┻┓   ┃
┃  ┃  ┃  12  ┃  ┃ ┃   | ┃ ┃ ┃  ┃  ┃  12  ┃    | ┃  ┃   ┃  ┃  12 ┃ ┃   | ┃  12 ┃ ┃ ┃  ┃   ┃
┃  ┃  ┃ ┏┻┓  ┃  ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┏┻┓  ┃    | ┃  ┃   ┃  ┃ ┏┻┓ ┃ ┃   | ┃ ┏┻┓ ┃ ┃ ┃  ┃   ┃
┃  ┃  ┃ ┃ ┃  11 ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┃ ┃  11   | ┃  ┃   11 ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  ┃   11
┃  ┃  ┃ ┃ ┃ ┏┻┓ ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┃ ┃ ┏┻┓   | ┃  ┃  ┏┻┓ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  ┃  ┏┻┓
┃  10 ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃  10 ┃ ┃ ┃ ┃ ┃   | ┃  10 ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  10 ┃ ┃
┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃   | ┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃
2 0 9 1 7 8 5 6 3 4   | 3 4 2 0 9 1 7 8 5 6   | 2 0 9 5 6 1 7 8 3 4   | 1 7 8 3 4 2 0 9 5 6
```
If we use the default, or ``max_tree_height='tree'``, we get
```
       19             |       20              |       19              |          21
  ┏━━━━┻━━━━┓         |  ┏━━━━┻━━━┓           |   ┏━━━┻━━━┓           |     ┏━━━━┻━━━━┓
  ┃         18        |  ┃        19          |   ┃       18          |     ┃         19
  ┃     ┏━━━┻━━┓      |  ┃    ┏━━━┻━━━┓       |   ┃    ┏━━┻━━━┓       |     ┃       ┏━┻━━┓
  ┃     ┃      16     |  ┃    ┃       18      |   ┃    ┃      17      |     17      ┃    ┃
  ┃     ┃    ┏━┻━┓    |  ┃    ┃     ┏━┻━━┓    |   ┃    ┃    ┏━┻━━┓    |   ┏━┻━━┓    ┃    ┃
  ┃     ┃    ┃   15   |  15   ┃     ┃    ┃    |   ┃    ┃    ┃    15   |   ┃    15   ┃    ┃
  ┃     ┃    ┃  ┏┻┓   | ┏┻┓   ┃     ┃    ┃    |   ┃    ┃    ┃   ┏┻┓   |   ┃   ┏┻┓   ┃    ┃
  ┃     14   ┃  ┃ ┃   | ┃ ┃   ┃     14   ┃    |   ┃    ┃    14  ┃ ┃   |   14  ┃ ┃   ┃    ┃
  ┃   ┏━┻┓   ┃  ┃ ┃   | ┃ ┃   ┃   ┏━┻┓   ┃    |   ┃    ┃  ┏━┻┓  ┃ ┃   | ┏━┻┓  ┃ ┃   ┃    ┃
  13  ┃  ┃   ┃  ┃ ┃   | ┃ ┃   13  ┃  ┃   ┃    |   13   ┃  ┃  ┃  ┃ ┃   | ┃  ┃  ┃ ┃   13   ┃
┏━┻┓  ┃  ┃   ┃  ┃ ┃   | ┃ ┃ ┏━┻┓  ┃  ┃   ┃    | ┏━┻┓   ┃  ┃  ┃  ┃ ┃   | ┃  ┃  ┃ ┃ ┏━┻┓   ┃
┃  ┃  ┃  12  ┃  ┃ ┃   | ┃ ┃ ┃  ┃  ┃  12  ┃    | ┃  ┃   ┃  ┃  12 ┃ ┃   | ┃  12 ┃ ┃ ┃  ┃   ┃
┃  ┃  ┃ ┏┻┓  ┃  ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┏┻┓  ┃    | ┃  ┃   ┃  ┃ ┏┻┓ ┃ ┃   | ┃ ┏┻┓ ┃ ┃ ┃  ┃   ┃
┃  ┃  ┃ ┃ ┃  11 ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┃ ┃  11   | ┃  ┃   11 ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  ┃   11
┃  ┃  ┃ ┃ ┃ ┏┻┓ ┃ ┃   | ┃ ┃ ┃  ┃  ┃ ┃ ┃ ┏┻┓   | ┃  ┃  ┏┻┓ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  ┃  ┏┻┓
┃  10 ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃  10 ┃ ┃ ┃ ┃ ┃   | ┃  10 ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃  10 ┃ ┃
┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃   | ┃ ┏┻┓ ┃ ┃ ┃ ┃ ┃ ┃ ┃   | ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃
2 0 9 1 7 8 5 6 3 4   | 3 4 2 0 9 1 7 8 5 6   | 2 0 9 5 6 1 7 8 3 4   | 1 7 8 3 4 2 0 9 5 6
```

There's still a bit of work to be done with testing, as I haven't tried out all the corner cases and error conditions, but I'd like to use it a bit first and see how it works. Would you mind trying it out and seeing if it works for the cases you were looking at?

jeromekelleher:
I've added an experimental tree sequence visualisation using ``TreeSequence.draw_svg``. In retrospect I think it was a mistake to have the text and SVG drawings done using ``draw``, so probably the thing to do is deprecate this method and add ``draw_svg`` and ``draw_text`` methods to Tree and TreeSequence. The Tree.draw_* methods should be able to use the code for TreeSequence.draw_* methods, so we won't have too much duplication.

jeromekelleher:
Use like this:
```python

ts = msprime.simulate(15, recombination_rate=0.5, random_seed=2, mutation_rate=2)

ts.draw_svg("trees_time.svg", tree_height_scale="time", width=ts.num_trees * 300)
ts.draw_svg("trees_rank.svg", tree_height_scale="rank", width=ts.num_trees * 300)
```

hyanwong:
You could do like dendropy does, and have both format supported, with the documentation present in the `draw_svg`, `draw_text`, etc functions, and simply use a wrapper like:

```
def draw(self, format, **kwargs):
    if format=="svg":
        self.draw_svg(**kwargs)
    if format=="unicode":
        self.draw_unicode(**kwargs)
    ...
```

jeromekelleher:
The idea though is that the ``draw_svg`` and ``draw_text`` get their own arguments that are specific for the backend. For SVG we can then get rid of things like ``node_colour`` and instead have arbitrary SVG that we can associate with nodes (i.e., a dictionary that gets passed to the svgwrite ``add`` method), giving much finer control over the drawing. We can keep the ``Tree.draw`` method around indefinitely with the current interface, and then evolve the ``draw_svg`` and ``draw_text`` as we like without breaking people's code.

hyanwong:
> The idea though is that the `draw_svg` and `draw_text` get their own arguments that are specific for the backend.

That's what I mean. You define `draw_svg` and `draw_text` with their own differently labelled parameters, but you can also access them using `tree.draw(format='svg', ...)` where the `...` is different depending on the format argument. So no need to deprecate the tree.draw() format - it's only that the documentation needs to point to the `draw_svg()` and `draw_text()` methods so as to know what the possible parameters can be in each case.

jeromekelleher:
> That's what I mean. You define draw_svg and draw_text with their own differently labelled parameters, but you can also access them using tree.draw(format='svg', ...) where the ... is different depending on the format argument. So no need to deprecate the tree.draw() format - it's only that the documentation needs to point to the draw_svg() and draw_text() methods so as to know what the possible parameters can be in each case.

Yes, but this would mean that the new ``draw_svg`` function would have to support all the old parameters, or we'd break peoples code. We have a much better idea of how to structure this now, and a 'clean slate' will be very helpful to make something powerful and extensible.

hyanwong:
Oh, right, I see. You want to change the parameters we have already. Yes, if we want to do that I can see that `draw_svg` et al. is a simple way to retain backwards compatibility.

jeromekelleher:
> Oh, right, I see. You want to change the parameters we have already. Yes, if we want to do that I can see that draw_svg et al. is a simple way to retain backwards compatibility.

Yeah, I want to change the semantics away from specifying things like ``colour``, etc, and providing dictionaries of SVG entities instead e.g., ``nodes: {u: {'fill': 'red', 'r': 50} for u in tree.nodes()}`` would change the nodes to radius 50 with fill colour red. *Much* more flexible!

jeromekelleher:
The experimental new interfaces for tree drawing are pretty usable now, so I'm going to merge. Hopefully we can use them a bit and move quickly towards documenting and making them widely usable.

I think you're going to like the new SVG interface @hyanwong - it's really flexible! I think if we can more of the semantics of the trees encoded in there, we can open up some really cool possiblities like SVG animations and so on.

Also, we now have TreeSequence.draw_text and TreeSequence.draw_svg. Hooray!

PR #170, opened 2019-04-05
Simple access to oldest time in the ts

hyanwong:


python/tskit/trees.py
```diff
@@ -2242,6 +2242,22 @@ def num_migrations(self):
         """
         return self._ll_tree_sequence.get_num_migrations()
 
+    @property
+    def oldest_time(self):
+        """
+        Returns the oldest time for any of the edges in this tree sequence.
```
jeromekelleher: The name isn't quite right I think, as you might expect this to be max(tables.nodes.time) instead. What about ``max_parent_time``? "Returns the time of the oldest node that is a parent in an edge" (or something). Raises ValueError if there are no edges.

Probably simpler to define ``max_time`` to be honest. In practise the time needed to run np.max(tables.nodes.time) is going to be negligible (particularly for anything that small enough to draw, which is the application, right)?

python/tskit/trees.py
```diff
@@ -2242,6 +2242,22 @@ def num_migrations(self):
...
+
+        :return: The oldest time in this tree sequence.
+        :rtype: float
+        """
+        # Edges are guaranteed to be listed in parent-time order, so we can just get the
+        # last one, assuming there are any edges (otherwise return None). This is O(1)
+        # compared to looking for the maximum node time.
+        if self.num_edges:
```
jeromekelleher: Using Falseyness can have nasty gotchas so I strongly discourage it. Much more readable to explicitly state what you're guarding against here.

python/tskit/trees.py
```diff
@@ -2242,6 +2242,22 @@ def num_migrations(self):
...
+            l, r, parent, child = self._ll_tree_sequence.get_edge(self.num_edges - 1)
```
jeromekelleher: No point in using the low-level API here: 
```
edge = self.edge(self.num_edges - 1)
return self.node(edge.parent).time

Comment (hyanwong) [reply python/tskit/trees.py]: I get `AttributeError: 'TreeSequence' object has no attribute 'edge'`

File: python/tskit/trees.py (hunk 1 above)
Comment (hyanwong) [reply python/tskit/trees.py]: I see this as the oldest time in the *tree sequence*, by which I mean the sequence of trees, not the node table. I don't think of this necessarily the same as the oldest node.

File: python/tskit/trees.py (hunk 2 above)
Comment (hyanwong) [reply python/tskit/trees.py]: Good point

File: python/tskit/trees.py (hunk 1 above)
Comment (hyanwong) [reply python/tskit/trees.py]: I just asked Wilder what he thought. He agreed that he thought of the oldest time in a tree sequence as the oldest point in any tree in the ts. That might be distinct from the oldest node time, which might be better accessed through the property `oldest_node`.

Either way, if we are drawing trees, we want the oldest point in any of the trees, not (necessarily) the oldest node in the node table, although they are likely to be the same aside from pathological cases.

Comment (jeromekelleher) [reply python/tskit/trees.py]: A tree sequence is defined as the tables, so I would expect ``oldest_time`` to be the maximum node time. Fine to add another property, but it must have a descriptve name that is clearly distinct from max_node_time.

File: python/tskit/trees.py (hunk 3 above)
Comment (jeromekelleher) [reply python/tskit/trees.py]: Ah, I see that method is missing. Would you mind adding it? (Follow the definition of 'node' and change the definition of the ``edges`` iterator to use the new method)

File: python/tests/test_highlevel.py [hunk 4]
```diff
@@ -1234,6 +1234,14 @@ def test_get_time(self):
             for u in range(N):
                 self.assertEqual(ts.get_time(u), ts.node(u).time)
 
+    def test_oldest_time(self):
+        for ts in get_example_tree_sequences():
+            oldest = 0
+            for tree in ts.trees():
+                for root in tree.roots:
+                    oldest = max(oldest, tree.time(root))
+            assert oldest == ts.oldest_time
```
jeromekelleher: Must use unittest methods as we're using node, e.g., ``self.assertEqual(oldest, ts.oldest_time)``

python/tskit/trees.py
hyanwong: Right. GTG now. Will do later

python/tskit/trees.py
hyanwong: Fair enough. I guess since we can have unary / recombination nodes, it's not right to use e.g. `oldest_coalescence_time` or `max_tmrca`. Possibilities include `max_tree_depth`, `max_tree_time`. Or other suggestions (ping @awohns )

hyanwong: Actually, I quite like `oldest_root`, or perhaps (to make it clear that we aren't returning the root itself, but the time of the oldest root, `oldest_root_time`.

hyanwong:
Waiting on https://github.com/tskit-dev/tskit/pull/171 to provide the `ts.edge()` function

python/tskit/trees.py
jeromekelleher: ``oldest_root_time`` is good, but we would need to make sure we defined it as ``max(ts.node(u) for u in ts.samples())`` when there are zero edges (because of our definition of roots).

PR #171, opened 2019-04-05
Define a python edge(i) function

hyanwong:
As requested by @jeromekelleher . But the initialiser of the `Edge()` class is not defined with an `id_` parameter or an `id` attribute, unlike all the other SimpleContainer classes like `Node`, so I think this might need further work to make standard.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2295,8 +2295,7 @@ class TreeSequence(object):
         :rtype: iter(:class:`.Edge`)
         """
         for j in range(self.num_edges):
-            left, right, parent, child = self._ll_tree_sequence.get_edge(j)
-            yield Edge(left=left, right=right, parent=parent, child=child)
+            yield self.edge(j)
 
     def edgesets(self):
         # TODO the order that these records are returned in is not well specified.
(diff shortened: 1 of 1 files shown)


hyanwong:
Also, it appears to me that none of these accessor functions on a tree sequence (`ts.node()` & friends) seem to have unittests?

jeromekelleher:
Yes, they've were skipped (but the underlying low-level methods are tested pretty thoroughly). If you change the edges() iterator to use your new function this will take care of the testing issue. It's also a good idea to add the ID attribute --- but it has to be done in a backward compatible way (i.e., id_ would be a keyword parameter at the end of the positional list.)

hyanwong:
@jeromekelleher - I've changed the edges iterator but not added `id_` - would you like me to do so?

jeromekelleher:
Let's add the id please while we're here --- we'll just forget about it again otherwise.

jeromekelleher:
Turns out I need the tree scaling stuff for a presentation I'm working on. I'll push these PRs through --- thanks @hyanwong.

jeromekelleher:
Turns out adding an ID to the Edge class is a bit tricky, so I'm not going to bother with it.

PR #172, opened 2019-04-06
kc-distance with branch lengths

awohns:
Modified kc_distance to consider branch lengths via the lambda parameter (0 = only topological comparison, 1 = only branch lengths). Renamed to kc_distance_tree as it compares two trees.
Also added kc_distance_ts, which calcualtes kc_distance across a tree sequence.

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -29,6 +29,7 @@ import itertools
 import random
 import json
 import sys
+import math
 
 import numpy as np
 import msprime
@@ -120,11 +121,11 @@ def generate_segments(n, sequence_length=100, seed=None):
     return segs
 
 
-def kc_distance_tree(tree1, tree2, topo_v_age=0):
+def kc_distance_tree(tree1, tree2, lambda_param=0):
     """
     Returns the Kendall-Colijn distance between the specified pair of trees.
-    topo_v_age determines weight of topology vs branch lengths in calculating
-    the distance. Set topo_v_age at 0 to only consider topology, set at 1 to
+    lambda_param determines weight of topology vs branch lengths in calculating
+    the distance. Set lambda_param at 0 to only consider topology, set at 1 to
     only consider branch lengths. See Kendall & Colijn (2016):
     https://academic.oup.com/mbe/article/33/10/2735/2925548
     """
@@ -155,7 +156,57 @@ def kc_distance_tree(tree1, tree2, topo_v_age=0):
                         M[tree_index][pair_index] = tree.time(tree.root) - time
             if len(tree.children(u)) == 0:
                 M[tree_index][u + n] = tree.branch_length(u)
-    return np.linalg.norm((1 - topo_v_age) * (m[0] - m[1]) + topo_v_age * (M[0] - M[1]))
+    return np.linalg.norm((1 - lambda_param) *
+                          (m[0] - m[1]) + lambda_param * (M[0] - M[1]))
+
+
+def kc_distance_tree_simple(tree1, tree2, lambda_param=0):
+    """
+    Simplified version of the kc_distance_tree() function above.
+    Written without Python features to aid writing C implementation.
+    """
+    samples = tree1.tree_sequence.samples()
+    for sample1, sample2 in zip(samples, tree2.tree_sequence.samples()):
+        if sample1 != sample2:
+            raise ValueError("Trees must have the same samples")
+    if not len(tree1.roots) == len(tree2.roots) == 1:
+        raise ValueError("Trees must have one root")
+    k = samples.shape[0]
+    n = (k * (k - 1)) // 2
+    m = [np.ones(n + k), np.ones(n + k)]
+    M = [np.zeros(n + k), np.zeros(n + k)]
+    path_distance = [np.zeros(tree1.num_nodes), np.zeros(tree2.num_nodes)]
+    time_distance = [np.zeros(tree1.num_nodes), np.zeros(tree2.num_nodes)]
+    for tree_index, tree in enumerate([tree1, tree2]):
+        stack = [(tree.root, 0, tree.time(tree.root))]
+        while len(stack) > 0:
+            u, depth, time = stack.pop()
+            children = tree.children(u)
+            for v in children:
+                stack.append((v, depth + 1, tree.time(v)))
+            path_distance[tree_index][u] = depth
+            time_distance[tree_index][u] = tree.time(tree.root) - time
+            if len(tree.children(u)) == 0:
+                M[tree_index][u + n] = tree.branch_length(u)
+
+        for index, n1 in enumerate(samples):
+            for n2 in samples[index + 1:]:
+                mrca = tree.mrca(n1, n2)
+                pair_index = n1 * (n1 - 2 * k + 1) // -2 + n2 - n1 - 1
+                assert m[tree_index][pair_index] == 1
+                m[tree_index][pair_index] = path_distance[tree_index][mrca]
+                M[tree_index][pair_index] = time_distance[tree_index][mrca]
+
+    vT1 = 0
+    vT2 = 0
+    distance_sum = 0
+
+    for i in range(n + k):
+        vT1 = (m[0][i] * (1 - lambda_param)) + (lambda_param * M[0][i])
+        vT2 = (m[1][i] * (1 - lambda_param)) + (lambda_param * M[1][i])
+        distance_sum += (vT1 - vT2) ** 2
+
+    return math.sqrt(distance_sum)
 
 
 class TestKCMetric(unittest.TestCase):
@@ -168,9 +219,11 @@ class TestKCMetric(unittest.TestCase):
                 ts = msprime.simulate(n, random_seed=seed)
                 tree = ts.first()
                 self.assertEqual(kc_distance_tree(tree, tree), 0)
+                self.assertEqual(kc_distance_tree_simple(tree, tree), 0)
                 ts = msprime.simulate(n, random_seed=seed)
                 tree2 = ts.first()
                 self.assertEqual(kc_distance_tree(tree, tree2), 0)
+                self.assertEqual(kc_distance_tree_simple(tree, tree2), 0)
 
     def test_sample_2_zero_distance(self):
         # All trees with 2 leaves must be equal distance from each other.
@@ -178,11 +231,13 @@ class TestKCMetric(unittest.TestCase):
             tree1 = msprime.simulate(2, random_seed=seed).first()
             tree2 = msprime.simulate(2, random_seed=seed + 1).first()
             self.assertEqual(kc_distance_tree(tree1, tree2), 0)
+            self.assertEqual(kc_distance_tree_simple(tree1, tree2), 0)
 
     def test_different_samples_error(self):
         tree1 = msprime.simulate(10, random_seed=1).first()
         tree2 = msprime.simulate(2, random_seed=1).first()
         self.assertRaises(ValueError, kc_distance_tree, tree1, tree2)
+        self.assertRaises(ValueError, kc_distance_tree_simple, tree1, tree2)
 
     def validate_trees(self, n):
         for seed in range(1, 10):
@@ -190,6 +245,9 @@ class TestKCMetric(unittest.TestCase):
             tree2 = msprime.simulate(n, random_seed=seed + 1).first()
             self.assertAlmostEqual(
                 kc_distance_tree(tree1, tree2), kc_distance_tree(tree1, tree2))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree1, tree2),
+                kc_distance_tree_simple(tree1, tree2))
 
     def test_sample_3(self):
         self.validate_trees(3)
@@ -227,12 +285,24 @@ class TestKCMetric(unittest.TestCase):
                 kc_distance_tree(tree1, tree2), kc_distance_tree(tree1, tree2))
             self.assertAlmostEqual(
                 kc_distance_tree(tree2, tree1), kc_distance_tree(tree2, tree1))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree1, tree2),
+                kc_distance_tree_simple(tree1, tree2))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree2, tree1),
+                kc_distance_tree_simple(tree2, tree1))
             # compare to a binary tree also
             tree2 = msprime.simulate(n, random_seed=seed + 1).first()
             self.assertAlmostEqual(
                 kc_distance_tree(tree1, tree2), kc_distance_tree(tree1, tree2))
             self.assertAlmostEqual(
                 kc_distance_tree(tree2, tree1), kc_distance_tree(tree2, tree1))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree1, tree2),
+                kc_distance_tree_simple(tree1, tree2))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree2, tree1),
+                kc_distance_tree_simple(tree2, tree1))
 
     def test_non_binary_sample_10(self):
         self.validate_nonbinary_trees(10)
@@ -277,6 +347,10 @@ class TestKCMetric(unittest.TestCase):
             kc_distance_tree(tree_1, tree_2, 0), 0)
         self.assertAlmostEqual(
             kc_distance_tree(tree_1, tree_2, 1), 4.243, places=3)
+        self.assertAlmostEqual(
+            kc_distance_tree_simple(tree_1, tree_2, 0), 0)
+        self.assertAlmostEqual(
+            kc_distance_tree_simple(tree_1, tree_2, 1), 4.243, places=3)
 
     def test_10_samples(self):
         nodes_1 = io.StringIO("""\
@@ -377,6 +451,10 @@ class TestKCMetric(unittest.TestCase):
             kc_distance_tree(tree_1, tree_2, 0), 12.85, places=2)
         self.assertAlmostEqual(
             kc_distance_tree(tree_1, tree_2, 1), 10.64, places=2)
+        self.assertAlmostEqual(
+            kc_distance_tree_simple(tree_1, tree_2, 0), 12.85, places=2)
+        self.assertAlmostEqual(
+            kc_distance_tree_simple(tree_1, tree_2, 1), 10.64, places=2)
 
     def test_15_samples(self):
         nodes_1 = io.StringIO("""\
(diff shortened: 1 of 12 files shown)


python/tests/test_topology.py
```diff
@@ -48,6 +48,293 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+                            a, b = v, u
+                        pair_index = a * (a - 2 * k + 1) // -2 + b - a - 1
+                        assert m[tree_index][pair_index] == 1
+                        m[tree_index][pair_index] = depth
+                        M[tree_index][pair_index] = tree.time(tree.root) - time
+            if len(tree.children(u)) == 0:
+                M[tree_index][u + n] = tree.time(tree.parent(u))
+    return ((1 - lambda_param) * np.linalg.norm(m[0] - m[1])) + (lambda_param * np.linalg.norm(M[0] - M[1]))
+
+def kc_distance_ts(ts_1, ts_2, lambda_param = 0):
```
jeromekelleher: Let's keep it to just trees for the PR @awohns, we can think about the TS case later.

jeromekelleher:
What's the status with this PR @awohns? It's slipped under my radar for a bit, let's try and get it merged soon. Looks like the issue that the moment is some linting problems - try running flake8 locally.

Should also rebase to bring this up to date (ask Yan or me for guidance if you're not sure how to do this)

awohns:
Ok, I squashed my commits and rebased to tskit/master, hopefully that was all done correctly. The tests all pass and linting is fine now. Let me know if this is alright or if I should make some changes! 
Also, this is only the Python implementation, I haven't gotten around to the C one yet but it's on my todo list.

jeromekelleher:
Great, thanks. Can you grab me at some point to go through this with you please?

awohns:
Happy to walk you through this anytime!

python/tests/test_topology.py
```diff
@@ -49,6 +49,532 @@ def generate_segments(n, sequence_length=100, seed=None):
     return segs
 
 
+def kc_distance_tree(tree1, tree2, topo_v_age=0):
+    """
+    Returns the Kendall-Colijn distance between the specified
+    pair of trees. topology_vs_age determines weight of topology vs
```
brianzhang01: Discrepancy between `topo_v_age` and `topology_v_age`

python/tests/test_topology.py
```diff
@@ -49,6 +49,532 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+                        else:
+                            a, b = v, u
+                        pair_index = a * (a - 2 * k + 1) // -2 + b - a - 1
+                        assert m[tree_index][pair_index] == 1
+                        m[tree_index][pair_index] = depth
+                        M[tree_index][pair_index] = tree.time(tree.root) - time
+            if len(tree.children(u)) == 0:
+                M[tree_index][u + n] = tree.branch_length(u)
+    return ((1 - topo_v_age) * np.linalg.norm(m[0] - m[1])) + (
+        topo_v_age * np.linalg.norm(M[0] - M[1]))
```
brianzhang01: Are you sure this is equivalent to
```
np.linalg.norm((1 - topo_v_age) * (m[0] - m[1]) + topo_v_age * (M[0] - M[1]))
```
? Because that's what I understood the metric to be. From thinking about it quickly, I don't think the two expressions are the same, because you can't distribute `np.linalg.norm` aka Euclidean distance over a `+` operation.

python/tests/test_topology.py
```diff
@@ -49,6 +49,532 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+            children = tree.children(u)
+            for v in children:
+                stack.append((v, depth + 1, tree.time(v)))
+            for c1, c2 in itertools.combinations(children, 2):
+                for u in tree.samples(c1):
+                    for v in tree.samples(c2):
+                        if u < v:
+                            a, b = u, v
+                        else:
+                            a, b = v, u
```
brianzhang01: Alternatively,
```
a = min(u, v)
b = max(u, v)
```

python/tests/test_topology.py
```diff
@@ -49,6 +49,532 @@ def generate_segments(n, sequence_length=100, seed=None):
     return segs
 
 
+def kc_distance_tree(tree1, tree2, topo_v_age=0):
+    """
+    Returns the Kendall-Colijn distance between the specified
+    pair of trees. topology_vs_age determines weight of topology vs
+    branch lengths in calculating the distance.
```
brianzhang01: Maybe add a paper citation. This is the one I'm looking at: https://academic.oup.com/mbe/article/33/10/2735/2925548

python/tests/test_topology.py
awohns: Good catch

python/tests/test_topology.py
awohns: Yep!

python/tests/test_topology.py
awohns: Good idea

python/tests/test_topology.py
awohns: Just looked into this, you're right: they're generally not the same thing. They're only the same when ||A+B|| is orthogonal, and that only happens when the topologies between the two trees your comparing is the same. In practice the difference between the distributed version and the correct version is small in absolute terms, but I'll amend it to be correct. Great catch! Thanks Brian

python/tests/test_topology.py
```diff
@@ -49,6 +49,530 @@ def generate_segments(n, sequence_length=100, seed=None):
     return segs
 
 
+def kc_distance_tree(tree1, tree2, topo_v_age=0):
+    """
+    Returns the Kendall-Colijn distance between the specified pair of trees.
+    topo_v_age determines weight of topology vs branch lengths in calculating
+    the distance. See Kendall & Colijn (2016):
```
brianzhang01: Might be worth saying that 0 = all topology, 1 = all age

awohns: good point, changed.

jeromekelleher:
Great to have an extra pair of eyes on this, thanks @brianzhang01!

jeromekelleher:
@awohns, we should pick this up again soon and finish it up. Can you take a look and see what needs doing please?

awohns:
Just added a simple version of the python function (without numpy or itertools).
Will now start writing the C version

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+    """
+    samples = tree1.tree_sequence.samples()
+    for sample1, sample2 in zip(samples, tree2.tree_sequence.samples()):
+        if sample1 != sample2:
+            raise ValueError("Trees must have the same samples")
+    if not len(tree1.roots) == len(tree2.roots) == 1:
+        raise ValueError("Trees must have one root")
+    k = samples.shape[0]
+    n = (k * (k - 1)) // 2
+    m = [[1] * (n + k), [1] * (n + k)]
```
jeromekelleher: Might as well use np.zeros/np.ones here --- it's no closer to C to use this.

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+    M = [[0] * (n + k), [0] * (n + k)]
+    depth_dict = [[0] * tree1.num_nodes, [0] * tree2.num_nodes]
```
jeromekelleher: Using "_dict" is confusing here. Maybe "node_depth", and "node_time" so we know what they are mapping to?

jeromekelleher: Might also be helpful to change the terminology a bit: maybe "time_distance" and "path_distance", as each of these is measuring distance from the root, right?

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+                        assert m[tree_index][pair_index] == 1
+                        m[tree_index][pair_index] = depth
+                        M[tree_index][pair_index] = tree.time(tree.root) - time
+            if len(tree.children(u)) == 0:
+                M[tree_index][u + n] = tree.branch_length(u)
+    return np.linalg.norm((1 - topo_v_age) * (m[0] - m[1]) + topo_v_age * (M[0] - M[1]))
+
+
+def kc_distance_tree_simple(tree1, tree2, topo_v_age=0):
+    """
```
jeromekelleher: No point in duplicating the docstring here. Just say it's a simpler version of the one above, corresponding to the C version.

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+                        pair_index = a * (a - 2 * k + 1) // -2 + b - a - 1
+                        assert m[tree_index][pair_index] == 1
+                        m[tree_index][pair_index] = depth
+                        M[tree_index][pair_index] = tree.time(tree.root) - time
+            if len(tree.children(u)) == 0:
+                M[tree_index][u + n] = tree.branch_length(u)
+    return np.linalg.norm((1 - topo_v_age) * (m[0] - m[1]) + topo_v_age * (M[0] - M[1]))
+
+
+def kc_distance_tree_simple(tree1, tree2, topo_v_age=0):
```
jeromekelleher: I don't like the topo_v_age parameter. I think it would be better to use the original terminology of calling it alpha or  beta or whatever, and we can explain what this means in the documentation.

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+
+        for index, n1 in enumerate(samples):
+            for n2 in samples[index + 1:]:
+                mrca = tree.mrca(n1, n2)
+                pair_index = n1 * (n1 - 2 * k + 1) // -2 + n2 - n1 - 1
+                assert m[tree_index][pair_index] == 1
+                m[tree_index][pair_index] = depth_dict[tree_index][mrca]
+                M[tree_index][pair_index] = time_dict[tree_index][mrca]
+
+    vT1 = [(depth * (1 - topo_v_age)) + (topo_v_age * time)
```
jeromekelleher: You're still leaving a lot to do in the C implementation here. This should be a single loop, not using zip or list comprehensions or anything like that. ``for j in range()`` etc.

python/tests/test_topology.py
```diff
@@ -97,6 +98,615 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+def kc_distance_tree_simple(tree1, tree2, topo_v_age=0):
+    """
+    Returns the Kendall-Colijn distance between the specified pair of trees.
+    topo_v_age determines weight of topology vs branch lengths in calculating
+    the distance. Set topo_v_age at 0 to only consider topology, set at 1 to
+    only consider branch lengths. See Kendall & Colijn (2016):
+    https://academic.oup.com/mbe/article/33/10/2735/2925548
+    """
+    samples = tree1.tree_sequence.samples()
+    for sample1, sample2 in zip(samples, tree2.tree_sequence.samples()):
```
jeromekelleher: Don't need to bother with error checking here -- this is just duplicating code from above.

python/tests/test_topology.py
```diff
@@ -98,11 +98,11 @@ def generate_segments(n, sequence_length=100, seed=None):
     return segs
 
 
-def kc_distance_tree(tree1, tree2, topo_v_age=0):
+def kc_distance_tree(tree1, tree2, lambda_param=0):
```
jeromekelleher: ``lambda_param`` is awkward. The convention is usually to call it ``lambda_`` to avoid clashes with built-in symbols.

python/tests/test_topology.py
```diff
@@ -173,14 +170,19 @@ def kc_distance_tree_simple(tree1, tree2, topo_v_age=0):
...
-           for depth, time in zip(m[1], M[1])]
-    return math.sqrt(sum([(a - b) ** 2 for a, b in zip(vT1, vT2)]))
+                m[tree_index][pair_index] = path_distance[tree_index][mrca]
+                M[tree_index][pair_index] = time_distance[tree_index][mrca]
+
+    vT1 = 0 
+    vT2 = 0 
+    distance_sum = 0
+
+    for i in range(n + k):
```
jeromekelleher: Nice!

python/tests/test_topology.py
awohns: Got it, thanks for checking this out!

python/tests/test_topology.py
```diff
@@ -97,6 +98,618 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+def kc_distance_tree(tree1, tree2, lambda_param=0):
+    """
+    Returns the Kendall-Colijn distance between the specified pair of trees.
+    lambda_param determines weight of topology vs branch lengths in calculating
+    the distance. Set lambda_param at 0 to only consider topology, set at 1 to
+    only consider branch lengths. See Kendall & Colijn (2016):
+    https://academic.oup.com/mbe/article/33/10/2735/2925548
+    """
+    samples = tree1.tree_sequence.samples()
+    if not np.array_equal(samples, tree2.tree_sequence.samples()):
```
brianzhang01: Does .samples() give sorted samples by default? If not, we may need to sort them. If yes, I'm guessing this is why you didn't write `tree1.samples()`, maybe you could add a short comment that we want to get the samples via the tree_sequence so that things are sorted.

python/tests/test_topology.py
```diff
@@ -97,6 +98,618 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+        raise ValueError("Trees must have the same samples")
+    if not len(tree1.roots) == len(tree2.roots) == 1:
+        raise ValueError("Trees must have one root")
+    k = samples.shape[0]
+    n = (k * (k - 1)) // 2
+    m = [np.ones(n + k), np.ones(n + k)]
```
brianzhang01: I'm pretty sure it will also work to do
```
m = np.ones(2, n+k)
M = np.zeros(2, n+k)
```
and may be ever slightly faster

python/tests/test_topology.py
```diff
@@ -97,6 +98,618 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+    M = [np.zeros(n + k), np.zeros(n + k)]
+    for tree_index, tree in enumerate([tree1, tree2]):
+        stack = [(tree.root, 0, tree.time(tree.root))]
+        while len(stack) > 0:
+            u, depth, time = stack.pop()
+            children = tree.children(u)
+            for v in children:
```
brianzhang01: huh, this is clever, doing a DFS from the root. i might give this a try as well.

python/tests/test_topology.py
```diff
@@ -97,6 +98,618 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+    for sample1, sample2 in zip(samples, tree2.tree_sequence.samples()):
+        if sample1 != sample2:
+            raise ValueError("Trees must have the same samples")
+    if not len(tree1.roots) == len(tree2.roots) == 1:
+        raise ValueError("Trees must have one root")
+    k = samples.shape[0]
+    n = (k * (k - 1)) // 2
+    m = [np.ones(n + k), np.ones(n + k)]
+    M = [np.zeros(n + k), np.zeros(n + k)]
+    path_distance = [np.zeros(tree1.num_nodes), np.zeros(tree2.num_nodes)]
```
brianzhang01: interesting, is it this way because the trees can have different number of nodes due to polytomies?

python/tests/test_topology.py
```diff
@@ -97,6 +98,618 @@ def generate_segments(n, sequence_length=100, seed=None):
...
+        for seed in range(1, 10):
+            tree1 = msprime.simulate(n, random_seed=seed).first()
+            tree2 = msprime.simulate(n, random_seed=seed + 1).first()
+            self.assertAlmostEqual(
+                kc_distance_tree(tree1, tree2), kc_distance_tree(tree1, tree2))
+            self.assertAlmostEqual(
+                kc_distance_tree_simple(tree1, tree2),
+                kc_distance_tree_simple(tree1, tree2))
+
+    def test_sample_3(self):
```
brianzhang01: I use pytest parameterized tests which basically does this without the code duplication. however we're not using pytest. https://docs.pytest.org/en/latest/parametrize.html

python/tests/test_topology.py
jeromekelleher: ``.samples`` returns the samples in sorted order, yes.

python/tests/test_topology.py
jeromekelleher: Pytest looks like all right, but it'd be a big job moving over now. There's not a lot of motivation for taking this work on.

brianzhang01: Here's another solution I came across which would fit well, and doesn't require any extra packages: https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests.

From https://stackoverflow.com/questions/32899/how-do-you-generate-dynamic-parameterized-unit-tests-in-python

jeromekelleher: Hey, this is great feature I didn't know about, thanks @brianzhang01! I'll definitely be deploying that one.

python/tests/test_topology.py
awohns: Yes, while the samples have to be the same between the two trees, the internal nodes can be different

awohns:
> When the number of samples gets large, doing a top-down search from root may be more efficient (as I think you were doing in the Python code). I've played around with both approaches and have settled on a top-down search. Here, for the samples that have a MRCA at root, I think you end up climbing the path to root several times.

Yes I can see how top-down would be more efficient.

> I also used a trick where in my top-down search, instead of calling a function to get the samples underneath a node, I do a postorder traversal using recursion where subnodes of my node each compute their samples underneath. Then you can either concatenate these lists (as there shouldn't be duplicates) or use something like C++ std::merge if you want the sample list in sorted order.

So, at each node, this could find all the pairs of samples where that node is the MRCA. I suppose you just have to keep track of whether the node is from the left or right subtree?

jeromekelleher:
Looks good @awohns. Clang has found a couple of C bugs if you look at the CircleCI output, so they should be fixed.

Otherwise, looking good. All that's left to do is hook this up with the Python API so that we can call it from Python now and to write some docs. I'm happy to take this from here, or I can talk you through the remaining steps and  we get them done on Friday. I'd like to release 0.2.3 then, and it'd be nice to have this functionality in for the release.

jeromekelleher:
Also, can you [rebase](https://stdpopsim.readthedocs.io/en/latest/development.html#rebasing) this please?

jeromekelleher:
> When the number of samples gets large, doing a top-down search from root may be more efficient (as I think you were doing in the Python code). I've played around with both approaches and have settled on a top-down search. Here, for the samples that have a MRCA at root, I think you end up climbing the path to root several times.

This sounds really interested @brianzhang01 --- perhaps you could send us a Python version of your algorithm, (even better) open a PR on the Python implementations here with your ideas? This is certainly not an optimal way of doing KC distance, it would be good to iterate a bit on the implementation and share ideas.

brianzhang01:
I really just used the ideas demonstrated in the [Python version](https://github.com/tskit-dev/tskit/pull/172/files#diff-5941ccebfc6d10311b8d79e8bc986cffR146) and added my own optimization for getting the samples. Sounds like we're going to release this version of the C algorithm first, but I'd be willing to collaborate on a more efficient C version later. (I'm working in C++ with a TreeSequence representation similar to tskit, and using pybind11 for my bindings.)

awohns:
I've written some documentation for adding functions to tskit, do we want that in this PR?

jeromekelleher:
> I've written some documentation for adding functions to tskit, do we want that in this PR?

Excellent! No, let's open another PR for that.

PR #175, opened 2019-04-16
simplify() takes an input list of ancestral nodes

gtsambos:
The `Simplifier` object now takes an optional list of node IDs, `ancestors`.
If this is provided, `simplify()` will only retain nodes that are ancestors or samples, and will simplify the topology accordingly.

NB: This is still nowhere near ready. Here are some things I need to change before this is pulled in:

- ~Combine edges with overlapping endpoints (see comment below).~
- ~Ensure that edges are created in the right order (see comment below)~
- ~Add tests that are specific to this simplify method~
    - ~If a node has no ancestor, record no edges.~
- ~Remove the big `if: else` statement from `merge_labelled_ancestors` so that less code is repeated.~
- Make C version.

I can also envisage making a 'cleaner' version of this `simplify.py` script that structures the operations in these alternate versions of `simplify()` in a more general manner, particularly within the `merge_labelled_ancestors` method. As written, this method is still structured around the assumption that the most important distinction when deciding how to process nodes is whether the node is unary or not - this isn't the case for the two newer versions of `simplify()`.
However, this is probably best left for another pull request.

Commenting in @dvukcevic.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -1829,30 +1829,30 @@ test_map_ancestors_input_errors(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, NULL, 2, ancestors, 2, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, NULL, 2, ancestors, 2, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_PARAM_VALUE);
 
     /* Bad sample IDs */
     samples[0] = -1;
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
 
     /* Bad ancestor IDs */
     samples[0] = 0;
     ancestors[0] = -1;
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, &result);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);  
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, 0, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
 
-    /* Duplicate sample IDs */ 
-    ancestors[0] = 4; 
+    /* Duplicate sample IDs */
+    ancestors[0] = 4;
     samples[0] = 1;
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_DUPLICATE_SAMPLE);
 
-    /* Duplicate sample IDs */ 
-    ancestors[0] = 6; 
+    /* Duplicate sample IDs */
+    ancestors[0] = 6;
     samples[0] = 0;
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_DUPLICATE_SAMPLE);
 
     /* TODO more tests! */
@@ -1871,6 +1871,9 @@ test_map_ancestors_single_tree(void)
     tsk_edge_table_t result;
     tsk_id_t samples[] = {0, 1};
     tsk_id_t ancestors[] = {4, 6};
+    size_t i;
+    double res_left = 0;
+    double res_right = 1;
 
     tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges,
             NULL, NULL, NULL, NULL, NULL);
@@ -1879,18 +1882,13 @@ test_map_ancestors_single_tree(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-
-    // tsk_edge_table_print_state(&result, stdout);
 
     // Check we get the right result.
     CU_ASSERT_EQUAL(result.num_rows, 3);
-    size_t i;
     tsk_id_t res_parent[] = {4, 4, 6};
     tsk_id_t res_child[] = {0, 1, 4};
-    double res_left = 0;
-    double res_right = 1;
     for (i = 0; i < result.num_rows; i++) {
         CU_ASSERT_EQUAL(res_parent[i], result.parent[i]);
         CU_ASSERT_EQUAL(res_child[i], result.child[i]);
@@ -1921,7 +1919,7 @@ test_map_ancestors_no_edges(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 1, ancestors, 1, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 1, ancestors, 1, 0, &result);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     tsk_table_collection_free(&tables);
@@ -1947,7 +1945,7 @@ test_map_ancestors_samples_and_ancestors_overlap(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 4, ancestors, 1, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 4, ancestors, 1, 0, &result);
 
     // tsk_edge_table_print_state(&result, stdout);
 
@@ -1988,7 +1986,7 @@ test_map_ancestors_paper(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 3, ancestors, 3, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 3, ancestors, 3, 0, &result);
 
     // tsk_edge_table_print_state(&result, stdout);
 
@@ -2028,7 +2026,7 @@ test_map_ancestors_multiple_to_single_tree(void)
     ret = tsk_edge_table_init(&result, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 1, &result);
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 1, 0, &result);
 
     // tsk_edge_table_print_state(&result, stdout);
 
@@ -2462,7 +2460,7 @@ main(int argc, char **argv)
         {"test_map_ancestors_paper", test_map_ancestors_paper},
         {"test_map_ancestors_samples_and_ancestors_overlap",
             test_map_ancestors_samples_and_ancestors_overlap},
-        {"test_map_ancestors_multiple_to_single_tree", 
+        {"test_map_ancestors_multiple_to_single_tree",
         test_map_ancestors_multiple_to_single_tree},
         {"test_sort_tables_drops_indexes", test_sort_tables_drops_indexes},
         {"test_copy_table_collection", test_copy_table_collection},
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -1813,6 +1813,27 @@ test_provenance_table(void)
 }
 
 static void
+test_map_ancestors(void)
+{
+    int ret;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = {0, 1};
+    tsk_id_t ancestors[] = {4, 6};
+
+    tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges,
+            NULL, NULL, NULL, NULL, NULL);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_table_collection_map_ancestors(&tables, samples, 2, ancestors, 2);
+    // CU_ASSERT_EQUAL_FATAL(ret, 0); // causing errors but don't know why
+
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
 test_simplify_tables_drops_indexes(void)
 {
     int ret;
@@ -2216,6 +2237,7 @@ main(int argc, char **argv)
         {"test_load_tsk_node_table_errors", test_load_tsk_node_table_errors},
         {"test_simplify_tables_drops_indexes", test_simplify_tables_drops_indexes},
         {"test_simplify_empty_tables", test_simplify_empty_tables},
+        {"test_map_ancestors", test_map_ancestors},
         {"test_sort_tables_drops_indexes", test_sort_tables_drops_indexes},
         {"test_copy_table_collection", test_copy_table_collection},
         {"test_sort_tables_errors", test_sort_tables_errors},
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -1813,11 +1813,12 @@ test_provenance_table(void)
 }
 
 static void
-test_map_ancestors(void)
+test_map_ancestors_input_errors(void)
 {
     int ret;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
+    tsk_edge_table_t result;
     tsk_id_t samples[] = {0, 1};
     tsk_id_t ancestors[] = {4, 6};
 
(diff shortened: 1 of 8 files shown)


gtsambos:
Hi @jeromekelleher, here's some example output that illustrates some issues with my current code:
```
############################################################
#   Edges                                                  #
############################################################
id	left		right		parent	child
0	0.00000000	1.00000000	5	0
1	0.00000000	1.00000000	5	1
2	0.00000000	1.00000000	5	4
3	0.00000000	0.20324489	6	2
4	0.00000000	0.20324489	6	3
5	0.20324489	1.00000000	6	3
6	0.20324489	1.00000000	7	2
7	0.20324489	1.00000000	7	5
8	0.21961708	0.72082046	8	6
9	0.00000000	0.20324489	9	5
10	0.20324489	0.21961708	9	7
11	0.21961708	0.72082046	9	7
12	0.21961708	0.72082046	9	8
13	0.72082046	1.00000000	9	7
```
The topology represented by this edge table is correct, but you can see it has a few oddities that I need to sort out:

1. Some edges have the same parent/child nodes and overlapping endpoints, so should really be combined into one (eg. edges 4 and 5).

2. Some edges aren't in order (eg. edge 12 should be at the end of the table).

(2) is the bigger problem, and I think the best solution would be to change the order in which segments are processed. This is why I introduced a separate sorting order for Segments in my pseudo-code. I need to think a bit more about why this is happening (and also: why it seems to be specific to this version of simplify), but it might have something to do with the fact that some nodes will have a higher polytomy(?) after this type of simplication.

jeromekelleher:
This is excellent progress, thanks @gtsambos. I'll dig into it over the next few days and give detailed feedback.

jeromekelleher:
I had a quick play with this @gtsambos, but I'm afraid I'm having trouble getting it to run. Can you give me an example which will get me started please? I've tried this:
```python
import msprime
from tests.simplify import Simplifier

ts = msprime.simulate(10, recombination_rate=0.5, random_seed=2)
for tree in ts.trees():
    print(tree.draw(format="unicode"))

s = Simplifier(ts, ts.samples(), ancestors=[20])
tss, _ = s.simplify() 
s.print_state()

for tree in tss.trees():
    print(tree.draw(format="unicode"))
```
This gives the following output:
```

Input id: 10
Input id: 11
Input id: 12
Input id: 13
Input id: 14
Input id: 15
  File "tmp.py", line 13, in <module>
    tss, _ = s.simplify()
  File "/home/jk/work/github/tskit/python/tests/simplify.py", line 466, in simplify
    self.process_parent_edges(edges)
  File "/home/jk/work/github/tskit/python/tests/simplify.py", line 353, in process_parent_edges
    self.check_state()
  File "/home/jk/work/github/tskit/python/tests/simplify.py", line 493, in check_state
    assert x.right <= x.next.left
AssertionError
```

gtsambos:
Oh yes, sorry, I probably should have said - partly because of these issues, I've been running this script in a very particular way on a particular example, and the tables you get out will likely fail certain tests that need to pass before you can get a tree sequence output from it. I'm not exactly sure if that's what's going on here, but I'll post an example when I get in tomorrow.

gtsambos:
Hi @jeromekelleher - after all that fuss and thought, I think I've fixed both of the above problems in the latest commit 😅 Problem (2) ended up being just a consequence of problem (1), which was just an indentation error and is now resolved.

However, the error that you've shown above (thankyou!) is a separate issue; I think this is happening because some of your samples don't have any local ancestors in the list that you've inputted. I haven't made any code to cope with this scenario yet, but can focus on that now that the core algorithm seems to be working!

Also, it seems as though in this version of simplify, there's not much difference between sample and ancestor nodes. In both cases, we want to retain the node in the simplified tree sequence and record some edges. I've been making the following distinction between them:

- An ancestral node should only be tracked on intervals where a sample node descends from it.
- A sample node should be tracked on all intervals where it is defined in the original tree sequence.

Does this seem sensible?

jeromekelleher:
> An ancestral node should only be tracked on intervals where a sample node descends from it.

This definitely sounds sensible.

> A sample node should be tracked on all intervals where it is defined in the original tree sequence.

Not sure I follow this one. What do you mean where a sample node is defined? By definition a sample node is in every tree in the sequence (if only as an isolated root).

gtsambos:
> Not sure I follow this one. What do you mean where a sample node is defined? By definition a sample node is in every tree in the sequence (if only as an isolated root).

What if the sample node isn't in the final generation of the simulation though? If you've simulated a tree with `msprime` (or built one with `tsinfer`), you might then apply `simplify` with one of the simulated/estimated ancestors as a sample, right? You have the facility to 'remember' intermediary nodes like this in SLiM as well (although in that case, the node will appear in every tree in the sequence).

gtsambos:
Hi @jeromekelleher, could you please see whether the latest commit resolves the error that you reported in [this comment](https://github.com/tskit-dev/tskit/pull/175#issuecomment-483570800)? [I was wrong](https://github.com/tskit-dev/tskit/pull/175#issuecomment-484305568) about the cause - it wasn't actually to do with the ancestral nodes, but a subtle issue to do with one of the checks in the `check_state()` method.

I've resolved this for now by [disabling this assertion](https://github.com/tskit-dev/tskit/pull/175/files#diff-57057a768236761add9e1f82d3298620R491) when ancestral nodes are supplied, but to make sure this is a good idea, it would be helpful to have some clarification about the purpose that this check is supposed to serve in the original version of  `simplify`:

The assertion is that for each segment `x` in the list `A[node]` of ancestral segments, the following statement holds
```
assert x.right <= x.next.left
```
Is this check in there because if false, the node being processed is binary over the interval `(x.next.left, x.right)`?

jeromekelleher:
> Hi @jeromekelleher, could you please see whether the latest commit resolves the error that you reported in this comment? If so, I was wrong about the cause - it wasn't actually to do with the ancestral nodes, but a subtle issue to do with one of the checks in the check_state() method.

I had a quick look, and it's doing what I was expecting now: hooray!

> Is this check in there because if false, the node being processed is binary over the interval (x.next.left, x.right)?

This is asserting that there are no overlapping segments in the linked list representing ancestry for a given node. It looks like you're using this in a fundamentally different way though, as all the segments the list have the same interval. I'll need to think about this properly before I can give you any worthwhile feedback, and I'm afraid I'm quite busy right now (travelling next week and preparing for some talks). 

It looks like you've broken the back of it though, so perhaps a good thing to do now would be to come up with some test cases? I think it would be good to devise separate test cases for this functionality (i.e., a new TestCase class) that tries this out over a mixture of coalescent simulations, Wright Fisher simualtions (there's a simple WF simulator in the test cases) and hand-crafted examples. The ``TestSampleLists`` class might be a good model to follow for this. I think it would be too complicated if we followed the same pattern as we used for ``keep_unary``, as this functionality is quite different. Does that sound OK?

gtsambos:
> I'll need to think about this properly before I can give you any worthwhile feedback, and I'm afraid I'm quite busy right now (travelling next week and preparing for some talks).

No worries, and no hurry either! I also can't give this a lot of time over the next week or two - I have a big talk at the end of this week, and I'm in transit/on holiday in London next week, so I'll probably still be working on this when I get to Oxford 🇬🇧 

> Does that sound OK?

Yes, thanks for the useful pointers! I'll take a look at that `TestSampleLists` class.

petrelharp:
>   What if the sample node isn't in the final generation of the TS, though? If you've simulated a tree with msprime (or built one with tsinfer), many of the nodes in the resulting TS will only be related to the present-day samples over particular intervals, and so the haplotypes/history of those nodes will only be known on those intervals, rather than over the whole sequence.

So far, we've assumed that samples have to be *completely known* nodes. You're right, that one could imagine wanting to remove this restriction. However, suppose you asked to simplify a tree sequence, specifying an only partially-recorded node as a sample. I think you get an error because of the lack of the `is_sample` node flag, but besides that, it should work just fine - simplify works on tree sequences that don't coalesce all the way; and a node with no ancestors on a segment has nothing recorded for it in the edge table for that segment.

So, I think I'm saying that sample nodes should be tracked on the entire chromosome; even if the original tree sequence doesn't say where they inherit from on part of the chromosome, it'll work out fine.

gtsambos:
Hi @jeromekelleher, I think this is ready to go.

python/tests/map_to_ancestors.py
```diff
@@ -0,0 +1,399 @@
...
+"""
+
+import sys
+
+import numpy as np
+
+import tskit
+
+
+def overlapping_segments(segments):
```
jeromekelleher: We should import this from the other module - no point in duplicating this code.

python/tests/map_to_ancestors.py
```diff
@@ -0,0 +1,399 @@
...
+
+    while len(X) > 0:
+        left = right
+        X = [x for x in X if x.right > left]
+        if len(X) > 0:
+            right = min(x.right for x in X)
+            yield left, right, X
+
+
+class Segment(object):
```
jeromekelleher: Is this the same as the definition in the simplify module? If so, see last comment.

python/tests/map_to_ancestors.py
```diff
@@ -0,0 +1,399 @@
...
+        if output_id != -1:
+            num_edges = self.flush_edges()
+            if num_edges == 0 and not is_sample:
+                self.rewind_node(input_id, output_id)
+
+    def record_edge(self, left, right, parent, child):
+        """
+        Adds an edge to the output list.
+        """
+        if self.reduce_to_site_topology:
```
jeromekelleher: I don't think this is relevant here, is it?

gtsambos:
> Wouldn't it be simpler to add the AncestorMap class to the current simplify module? That would save all the copying and would make it clear what parts are shared and aren't.

Ah, I think I might have misunderstood the original instruction. Yep, I'll do this now.

python/tests/map_to_ancestors.py
gtsambos: Thanks, must have missed this - I think I've removed all other site/mutation related code.

gtsambos:
More streamlined version just pushed.
Most of the repetition is inside the actual `AncestorMap` class: the only method in this class that really differs from it's `Simplifier` equivalent is `merge_labelled_ancestors`: most of the other methods in are duplicates of the `Simplifier` version, or very similar versions where any code pertaining to sites/mutations has been cut out. 

It doesn't seem like there is much we can do about this though? I mean, I could just put the new `merge_labelled_ancestors` and `simplify` methods into the `Simplifier` class with a different name. But given the conversation we had with @petrelharp on Wednesday, it sounded like everyone was keen to keep this functionality outside of the `Simplifier` class in order to emphasise to users that it can't be used in place of one of the existing `simplify` techniques.

gtsambos:
Something else we could do to make it simpler is to remove the `filter_individual` and `filter_population` options, since they aren't super important?

jeromekelleher:
Great, thanks @gtsambos. The next thing to do I think is to remove the ``self.tables`` variable, and instead just output an edge table. This will remove a lot of the extra stuff, as we'll not need to create new nodes any more, and also the ``filter_x`` options no longer make sense as we're just outputting a set of edges.

jeromekelleher:
To make things sync up with the existing tests, we can add a function that takes the original tree sequence and the output edge table, and returns the corresponding tree sequence with the edges replaced by the output of ``map_ancestors``. This should be equivalent to what you're doing now (although you mind need to simplify this tree sequence to get rid of the redundant nodes)

gtsambos:
Cool, I'll do this now. Something

> To make things sync up with the existing tests, we can add a function that takes the original tree sequence and the output edge table, and returns the corresponding tree sequence with the edges replaced by the output of `map_ancestors`. This should be equivalent to what you're doing now (although you mind need to simplify this tree sequence to get rid of the redundant nodes)

I think it might be easier (and more intuitive for us to follow later) if I just modify the tests to look at the edge table output only. Simplifying the tree in addition could be a bit complicated, as it would also relabel the nodes in a different way.

On that point - did we also say that the output nodes now shouldn't be relabelled at all under this `AncestorMap` scheme? This would allow make it easier to compare with the original tree sequence.

jeromekelleher:
> I think it might be easier (and more intuitive for us to follow later) if I just modify the tests to look at the edge table output only. Simplifying the tree in addition could be a bit complicated, as it would also relabel the nodes in a different way.

OK, whichever is more straightforward.

> On that point - did we also say that the output nodes now shouldn't be relabelled at all under this AncestorMap scheme? This would allow make it easier to compare with the original tree sequence.

Yes, the node IDs remain the same now so that we can directly compare the output edge table with the original tree seque3nce.

gtsambos:
in the process of adding in some final flake8 tests and trying to rebase my history, I've screwed up something else in my local history ☠️ I'll try to sort this out now, but if not I may need your help with this tomorrow! Otherwise, I should be ready to start the C part of this work --

petrelharp:
re: merge conflicts - you may be interested in this writeup by Jerome: https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow
... if I get a bad merge conflict these days, I do what he suggests [here](https://stdpopsim.readthedocs.io/en/latest/development.html#when-rebasing-goes-wrong)

gtsambos:
Thanks @petrelharp !

Hey @jeromekelleher, I've just noticed that the travis-ci check has been running for hours and hours - I don't know what I've done, but on hindsight the latest commit was not a 'complete' enough one to be pushed up and I totally could have changed something in the C code that is making one of the checks go forever.

I need to have dinner now, but will look into getting rid of this once I've eaten!!

jeromekelleher:
No worries @gtsambos, Travis will be fine! Let's have a chat about rebasing today, if you're about.

gtsambos:
🎉 🎉 🎉

gtsambos:
@jeromekelleher: the codecov patch is failing, but I don't think there's much I can do about it because most of the lack of coverage is due to the `simplifier_check_state` and `simplifier_print_state` functions not being used.

jeromekelleher:
> @jeromekelleher: the codecov patch is failing, but I don't think there's much I can do about it because most of the lack of coverage is due to the simplifier_check_state and simplifier_print_state functions not being used.

OK, let's not worry about that then. Shall I put in the Python/C boilerplate then?

gtsambos:
> OK, let's not worry about that then. Shall I put in the Python/C boilerplate then?

That would feel very exciting!

jeromekelleher:
OK, that done @gtsambos: hooray! So, what do we need to do to merge this?

1. Fix failing test case (see the FIXME @gtsambos message in test_topology)
2. Write documentation (specifically, a docstring for TableCollection.map_ancestors)
3. Rebase and tidy up history a bit

Nearly there!

gtsambos:
Hey @jeromekelleher, when I run the tests after recompiling the C build with your changes, I get the following error a lot of times:
```
AttributeError: '_tskit.TableCollection' object has no attribute 'map_ancestors'
```
I assume there's something I need to reset/rebuild that I haven't done? Do you know what this could be?

jeromekelleher:
Yeah, you need to rebuild your low-level Python-C module (this is the _tskit.cpythonxxxx.so file). If you type 'make' in the Python directory this should do it. (This compilation is independent of the one you do in the C directory).

gtsambos:
ready to review now I think?

c/tskit/tables.c
```diff
@@ -5493,6 +5981,13 @@ tsk_table_collection_has_index(tsk_table_collection_t *self, tsk_flags_t TSK_UNU
         && self->indexes.edge_removal_order != NULL;
 }
 
+// bool
```
jeromekelleher: Why is this commented out?

python/tests/test_topology.py
```diff
@@ -3466,6 +3460,283 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+        self.assertEqual(all(tss.right), 1)
+
+    def test_single_tree_no_ancestors(self):
+        nodes = io.StringIO(self.nodes)
+        edges = io.StringIO(self.edges)
+        ts = tskit.load_text(nodes=nodes, edges=edges, strict=False)
+        tss = self.do_map(ts, samples=[2, 3], ancestors=[7])
+        self.assertEqual(tss.num_rows, 0)
+
+    # @unittest.skip("FIXME @gtsambos")
```
jeromekelleher: Delete this comment

jeromekelleher:
@petrelharp, can you take a quick look here please?

python/tests/test_topology.py
gtsambos: done

c/tskit/tables.c
gtsambos: I think this was always commented out? When I uncommented and tried to compile I got this error

```
_tables.c.o' -c ../tskit/tables.c
../tskit/tables.c:5985:1: error: no previous prototype for function 'tsk_edge_table_has_index' [-Werror,-Wmissing-prototypes]
tsk_edge_table_has_index(tsk_table_collection_t *self, tsk_flags_t TSK_UNUSED(options))
^
1 err
```

jeromekelleher: Hmm, it doesn't actually make sense anyway. Must have snuck through at some point. Can you delete it please?

python/tskit/tables.py
```diff
@@ -1663,6 +1663,28 @@ def simplify(
...
+        """
+        Returns an :class:`EdgeTable` instance displaying all relationships between
+        the nodes supplied in `samples` and `ancestors`.
+        Relationships between different nodes in `ancestors` are also shown.
+        The supplied nodes must be a subset of the node IDs in the tree sequence.
+        Other than this, the procedure is flexible; overlap between the lists of samples
+        and ancestors is permitted, and the nodes need not be ordered by time.
+        Node IDs are not changed in the output.
+        If none of the nodes in `ancestors` are ancestral to `samples` anywhere in
+        the tree sequence, an empty table will be returned.
```
petrelharp: This is great. Could you add another short paragraph saying precisely how the EdgeTable displays the relationships, @gtsambos? For instance, say that the IDs in the edge table refer to the original tree sequence; and also say when exactly an edge appears in the output edge table.

python/tests/test_topology.py
```diff
@@ -3466,6 +3460,282 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+                    while tree.interval[1] <= current_left:
+                        tree.next()
+                    # If there's just one child, check node is an ancestor
+                    # in the original tree.
+                    current_descendants = list(set(current_descendants))
+                    if len(current_descendants) == 1:
+                        self.assertTrue(
+                            tree.is_descendant(
+                                current_descendants[0],
+                                current_ancestor))
```
petrelharp: Hm. Shouldn't this check if current ancestor is the *most recent* one of the ancestors of the descendant?

python/tests/test_topology.py
```diff
@@ -3466,6 +3460,282 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+                    # Otherwise, for all combinations of descendants, check that the
+                    # current ancestor has the mrca as a descendant.
+                    else:
+                        pairs = itertools.combinations(current_descendants, 2)
+                        for pair in pairs:
+                            mrca = tree.mrca(pair[0], pair[1])
+                            self.assertTrue(tree.is_descendant(mrca, current_ancestor))
```
petrelharp: Another way to do both these checks (lines 3683-3694) would be to: for each of `current_descendants`, iterate up the tree until one of `ancestral_nodes` is reached, and check that is equal to `current_ancestor`. That would then check we've actually got the most recent ancestor for everyone instead of just an ancestor.

c/tests/test_tables.c
```diff
@@ -2216,6 +2454,14 @@ main(int argc, char **argv)
...
         {"test_simplify_tables_drops_indexes", test_simplify_tables_drops_indexes},
         {"test_simplify_empty_tables", test_simplify_empty_tables},
+        {"test_map_ancestors_no_edges", test_map_ancestors_no_edges},
+        {"test_map_ancestors_input_errors", test_map_ancestors_input_errors},
+        {"test_map_ancestors_single_tree", test_map_ancestors_single_tree},
+        {"test_map_ancestors_paper", test_map_ancestors_paper},
+        {"test_map_ancestors_samples_and_ancestors_overlap",
+            test_map_ancestors_samples_and_ancestors_overlap},
+        {"test_map_ancestors_multiple_to_single_tree",
+        test_map_ancestors_multiple_to_single_tree},
```
petrelharp: indentation. =)

python/tests/test_topology.py
```diff
@@ -3466,6 +3460,282 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+    """
+
+    def do_map(self, ts, ancestors, samples=None, compare_lib=True):
+        """
+        Runs the Python test implementation of map_ancestors.
+        """
+        if samples is None:
+            samples = ts.samples()
+        s = tests.AncestorMap(ts, samples, ancestors)
+        ancestor_table = s.simplify()
```
petrelharp: I was confused that the "map ancestors" method of `AncestorMap` is called `.simplify()`. Perhaps this should be called `.map_ancestors()` also? I'm trying to keep the tests readable.

c/tskit/tables.c
```diff
@@ -3892,17 +3892,504 @@ segment_overlapper_next(segment_overlapper_t *self,
...
+        assert(num_overlapping > 0);
+        if (is_ancestor || is_sample) {
+            for (j = 0; j < num_overlapping; j++) {
+                ret = ancestor_mapper_record_edge(self, left, right, X[j]->node);
+                if (ret != 0) {
+                    goto out;
+                }
+            }
+            ret = ancestor_mapper_add_ancestry(self, input_id, left, right,
+                input_id);
```
petrelharp: indendation

petrelharp:
This is totally wonderful!! I've had a good look through everything but the details of the C implementation. My main concern is that I don't think it's being completely verified; lmk if I missed something there. I also nitpicked the indentation in C, sorry. =)

gtsambos:
Great, thanks for looking at this @petrelharp! I'm preparing for a talk on Monday, but once that's done I'll have a look at some of those things you pointed out.
Re: the tests using the `verify` function - this was something @jeromekelleher and I spent quite a bit of time thinking about a while ago, but it may well be that there are more 'complete' ways of checking this. I also remember explicitly making some of the bespoke test cases to cover examples of the topologies that weren't covered by `verify`. Anyway, I'll review more closely at the start of next week.

petrelharp:
> Re: the tests using the verify function - this was something @jeromekelleher and I spent quite a bit of time thinking about a while ago, but it may well be that there are more 'complete' ways of checking this. 

I think my suggestion is pretty concrete; lmk if you want me to implement it.

> bespoke test cases

I love them!

c/tskit/tables.c
jeromekelleher: **cough** I prefer it the way @gtsambos has done it, and I think that's the way all the rest of the C code is done. Varying indentation based on the lengths of calling function names places a strong pressure to **not change** the names of things, or else you end up with a lot of wonky indentation.

petrelharp: oh is it? whoops. =)

python/tests/test_topology.py
gtsambos: After a bit of thought, yes, I think this is a better check. We were very fixated on the idea of doing a test involving MRCAs for some reason. I have changed this in the latest commit.

gtsambos:
Thanks again for your comments @petrelharp, I've made changes to all of these except the indentation.

python/tskit/tables.py
petrelharp: > For each sequence of edges that joins a sample node to an ancestral node

Hm, this is still a big ambiguous: if a tree has a -> b -> c, where a and b are ancestors and c is a sample, then there's a path from c to a but the output doesn't have a row for it.

How about, erm,
```
Returns an :class:`EdgeTable` describing how ``samples`` are descended from ``ancestors`` (and how ``ancestors`` are descended from each other). Each row ``parent, child, left, right`` in the output table indicates that ``child`` (which is either in ``samples`` or in ``ancestors``) has inherited the segment ``[left, right)`` from ``parent`` (which is in ``ancestors``) more recently than from any other node in ``ancestors``. Every segment of the genome on which each of ``samples`` has inherited from any of ``ancestors`` is described in this way, as well as any segment of the genomes of ``ancestors`` that has been inherited by any of ``samples``.  The node IDs in ``parent`` and ``child`` refer to the node table in the original tree sequence. In other words, the tree constructed at a position with the output edge table has the property that the parent of any node in ``samples``, if there is one, is the most recent ancestor of that node at that position out of all possible ancestors in ``ancestors``, and the same is true for any node in ``ancestors`` that is ancestral to any of ``samples``.
```

Is this right? Particularly about including edges with ancestors in them iff they are ancestral to samples?

This made me think of another case: what happens if a,b are samples and x,y are ancestors and there's a segment on which the relationship is x -> a -> y -> b?  Do we get
```
parent child
 x           a
 x           y
 y           b
```
...?

gtsambos: Thanks - have updated the description!

> This made me think of another case: what happens if a,b are samples and x,y are ancestors and there's a segment on which the relationship is x -> a -> y -> b? Do we get ...

Yep, that's right. I've added a new test case showing this behaviour (`test_single_tree_ancestors_descend_from_samples`)

gtsambos: (Maybe that test name should be changed, as it implies that the ancestors and samples are completely interchangeable which they actually aren't.) Eg. if in the above example you'd set samples = [a] and ancestors = [x, y], you'd only get the first row of that output.

gtsambos: Agh, just realised I should have looked at your example a bit more carefully.
You don't get this
```
parent child
 x           a
 x           y
 y           b
```

but this
```
parent child
 x           a
 a           y
 y           b
```

gtsambos: Hmm, just thinking that x -> a -> b is another distinct case that I need a distinct test for.
The `samples` and `ancestors` are almost the same - the only difference is that an `ancestor` will *only* be retained if it has a relevant descendant, while a `sample` will *always* be retained.

jeromekelleher:
OK, merging this. Well done @gtsambos!

python/tskit/tables.py
petrelharp: Hm, that means the documentation is not quite right. Can you perhaps fix it up?

jeromekelleher: Sorry, I didn't realise there was still unresolved issues here. We can fix this with a follow up PR though, right?

gtsambos: Yep, I'll open up another pull request with these small changes once I've made them.

petrelharp: thanks! I could do it, but I'm more certain that you know exactly what it does than I do.

PR #183, opened 2019-05-20
Use tskit.NULL rather than defining it again in drawing.py

hyanwong:
Seems sensible to use the general tskit.NULL id when drawing too


petrelharp:
Looks good to me; thanks!

jeromekelleher:
LGTM too!

PR #184, opened 2019-05-21
Updated GNN stat to use new definition.

jeromekelleher:
This fixes a bug in the logic of the GNN stat. This actually made the tests a little cleaner, as some special cases aren't necessary. I added some simple examples to make sure it's doing the right thing in the old pathological cases.

cc @hyanwong and @awohns

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -286,6 +286,15 @@ verify_genealogical_nearest_neighbours(tsk_treeseq_t *ts)
         samples, num_samples, sample_sets, sample_set_size, 2, 0, A);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
+    sample_sets[0] = samples;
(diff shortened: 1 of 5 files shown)


petrelharp:
Does it say somewhere what the new definition is?

jeromekelleher:
This will be in the tsinfer paper which we're revising now (we'll post the updated preep after final review). The update is very minor and just fixes a corner-case that @hyanwong spotted. 

Here's the definition in the meantime
Screenshot from 2019-05-22 08-14-53

jeromekelleher:
@awohns, I've incorporated your treewise GNN calculation in here temporarily to make sure it's computing GNN correctly. It's all checking out, so I'm confident it's correct.

I'll update this PR to add a ``treewise=False`` option to the ``genealogical_nearest_neighbours`` method which will optionally return all treewise GNNs by adding an extra dimension to the returned array.

Issue #8, opened 2018-11-17
simplify() can't cope with an np.array of standard integers


hyanwong:
Slightly surprisingly, I can't pass a numpy array of type np.int to ts.simplify() on my machine, as it defaults to np.int64, and is then downcast somewhere in simplify, causing the error:

`TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'`

jeromekelleher:
Closed in #185.

jeromekelleher:
See #192 for planned wider applications of the casting function.

PR #185, opened 2019-05-22
Fixes issue #8

hyanwong:
Solving https://github.com/tskit-dev/tskit/issues/8 seems a bit more convoluted than I had expected, but we need to allow passing `[]`, which by default is of dtype `np.float64`, and it seems as if numpy doesn't have any built-in bounds checking. I've created a function `to_np_int32` as I imagine we might want to use this in other python functions, as per the `set_columns` example by @petrelharp. But perhaps there's a neater solution.

As a bonus, this now catches errors like `ts.simplify([0.1,1.3])` which used to convert silently to ints, and it also returns a TypeError for `ts.simplify({})` rather than ValueError, which seems more correct to me. I've also added a few more test cases.

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -951,7 +951,7 @@ class TestNumpyInt32Conversion(unittest.TestCase):
             self.assertRaises(TypeError, tskit.tables.to_np_int32, bad_type)
 
     def test_overflow(self):
-        for bad_node in [np.iinfo(np.int32).min-1, np.iinfo(np.int32).max+1]:
+        for bad_node in [np.iinfo(np.int32).min - 1, np.iinfo(np.int32).max + 1]:
             self.assertRaises(  # Test plain array
                 OverflowError, tskit.tables.to_np_int32, [0, bad_node])
             self.assertRaises(  # Test numpy array
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -921,6 +921,51 @@ class TestBytePacking(unittest.TestCase):
         self.assertEqual(data, unpickled)
 
 
+class TestNumpyInt32Conversion(unittest.TestCase):
+    """
+    Tests the function that converts integer arrays to np.int32 for input to functions.
+    """
+    def test_arrays(self):
+        # Simple array
+        target = np.array([0, 1], dtype=np.int32)
+        for identical_array in [[0, 1], (0, 1), np.array([0, 1]), target]:
+            converted = tskit.tables.to_np_int32(identical_array)
+            # Use pickle to test exact equality including dtype
+            self.assertEqual(pickle.dumps(converted), pickle.dumps(target))
+        # Nested array
+        target = np.array([[0, 1], [2, 3]], dtype=np.int32)
+        for identical_array in [[[0, 1], [2, 3]], np.array([[0, 1], [2, 3]]), target]:
+            converted = tskit.tables.to_np_int32(identical_array)
+            self.assertEqual(pickle.dumps(converted), pickle.dumps(target))
+
+    def test_empty_arrays(self):
+        target = np.array([], dtype=np.int32)
+        converted = tskit.tables.to_np_int32([])
+        self.assertEqual(pickle.dumps(converted), pickle.dumps(target))
+        target = np.array([[]], dtype=np.int32)
+        converted = tskit.tables.to_np_int32([[]])
+        self.assertEqual(pickle.dumps(converted), pickle.dumps(target))
+
+    def test_bad_types(self):
+        for bad_type in [[0.1], ['string'], {}, [{}], np.array([0, 1], dtype=np.float)]:
+            self.assertRaises(TypeError, tskit.tables.to_np_int32, bad_type)
+
+    def test_overflow(self):
+        for bad_node in [np.iinfo(np.int32).min-1, np.iinfo(np.int32).max+1]:
+            self.assertRaises(  # Test plain array
+                OverflowError, tskit.tables.to_np_int32, [0, bad_node])
+            self.assertRaises(  # Test numpy array
+                OverflowError, tskit.tables.to_np_int32, np.array([0, bad_node]))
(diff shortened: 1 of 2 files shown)


python/tskit/tables.py
```diff
@@ -1815,3 +1817,23 @@ def unpack_strings(packed, offset, encoding="utf8"):
...
+    integer arrays (dtype=np.int64) we cast but check bounds to avoid wrap-around
+    conversion errors (numpy doesn't seem to provide this functionality)
+    """
+    int_array = np.array(int_array)
+    if int_array.size == 0:
+        return int_array.astype(np.int32)  # Allow empty arrays of any type
+    try:
+        return int_array.astype(np.int32, casting='safe')
+    except TypeError:
+        if (int_array >= np.iinfo(np.int32).min).all() and \
```
jeromekelleher: Cleaner to get a reference to np.iinfo(np.int32) in a variable, wouldn't need line break then.

hyanwong:
Is the right place to put my `to_np_int32()` function at the end of the `tables.py` file, and is the name OK?

Re: tests. These are done somewhat in the `test_samples_interface()` code, but I guess it would be useful to have a standalone set too.

python/tskit/tables.py
hyanwong: Good point. Done

python/tests/test_tables.py
```diff
@@ -921,6 +921,41 @@ def test_pickle_packing(self):
...
+
+    def test_bad_types(self):
+        for bad_type in [[0.1], ['string'], {}, [{}], np.array([0, 1], dtype=np.float)]:
+            self.assertRaises(TypeError, tskit.tables.to_np_int32, bad_type)
+
+    def test_overflow(self):
+        for bad_node in [np.iinfo(np.int32).min-1, np.iinfo(np.int32).max+1]:
+            self.assertRaises(
+                OverflowError, tskit.tables.to_np_int32, np.array([0, bad_node]))
+
```
jeromekelleher: Looks good. I think we just need one more test, where we try out the values what are just inside the allowable range to make sure the boundaries are handled correctly.

python/tskit/tables.py
```diff
@@ -1815,3 +1817,23 @@ def unpack_strings(packed, offset, encoding="utf8"):
...
+    """
+    int_array = np.array(int_array)
+    if int_array.size == 0:
+        return int_array.astype(np.int32)  # Allow empty arrays of any type
+    try:
+        return int_array.astype(np.int32, casting='safe')
+    except TypeError:
+        int32bounds = np.iinfo(np.int32)
+        if (int_array >= int32bounds.min).all() and (int_array <= int32bounds.max).all():
+            # This should raise the correct error on conversion from e.g. a float array
```
jeromekelleher: This comment is a bit unsettling - does it correct for them or not?

jeromekelleher:
> Is the right place to put my to_np_int32() function at the end of the tables.py file, and is the name OK?

Sure, this is as good a place as any and name is good.

python/tskit/tables.py
hyanwong: Yes, `np.array([1.0]).astype(np.int32, casting='same_kind')` raises 

```
TypeError: Cannot cast array from dtype('float64') to dtype('int32') according to the rule 'same_kind'
```

hyanwong: As confirmed in the test suite

python/tests/test_tables.py
hyanwong: Good point

hyanwong: Done

hyanwong:
I could also use `to_np_int32()` in `tables.nodes.set_columns` for the `flags` param, as flagged up by @petrelharp , but presumably there are a good number of places where we might want to apply this function, and I'm not the best placed to know where they all are, or how to compile a sensible list.

python/tskit/tables.py
jeromekelleher: I'm sure it's correct, it's just the comment implies that this hasn't been verified. In a few years someone will read through (probably me) and think that this function hasn't been tested and then waste a bunch of time retesting it. I'm just asking to change the comment so that it doesn't mislead like this.

jeromekelleher:
Flags is uint32, so we'd need a slightly different version there.  I think the to_int32 should be applied anywhere we referring to object IDs. These will be in the Tables API (node IDs, site IDs, etc), and where we compute stats (GNN, eg, takes lists of samples). We don't have to do it all at once, but it would be good to follow up with an issue saying this needs to be done if you don't want to do it here.

python/tskit/tables.py
hyanwong: Yes, fine, please change

hyanwong:
> We don't have to do it all at once, but it would be good to follow up with an issue saying this needs to be done if you don't want to do it here.

Right. I'll open an issue once this is merged.

python/tskit/tables.py
jeromekelleher: OK, I'll come back and fix the comment when I get a chance then

python/tskit/tables.py
```diff
@@ -1815,3 +1817,23 @@ def unpack_strings(packed, offset, encoding="utf8"):
     :rtype: list[str]
     """
     return [b.decode(encoding) for b in unpack_bytes(packed, offset)]
+
+
+def to_np_int32(int_array):
```
petrelharp: maybe this should be a private method? is there a common use case for tskit-users?

hyanwong: Good point. I can’t think of any reason to make this public. So I guess I should simply add an underscore to start of the function name?

jeromekelleher: I've given up on using underscores for this a while back, it's mostly pointless unless it's **really** something that you don't want people to use. The operational definition of private here is 'not listed in public documentation'.

If we want to keep the namespace clean we can modify what's exported in the ``msprime/__init__.py``, at the moment I think we just take everything, but we could selectively export functions and classes.

jeromekelleher:
Merging, thanks @hyanwong.

PR #187, opened 2019-05-25
very basic Fst implementation

GertjanBisschop:
added windowwise trees.calculate_Fst() and tree.Fst()

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1375,6 +1375,19 @@ class Tree(object):
     def __str__(self):
         return str(self.get_parent_dict())
 
+    def Fst(self, samples_p1, samples_p2):
+
+        p1 = [self.tmrca(u,v) for u,v in itertools.combinations(sample_p1, 2)]
+        p1_mean = np.mean(p1)
+        p2 = [self.tmrca(u,v) for u,v in itertools.combinations(sample_p2, 2)]
+        p2_mean = np.mean(p2)
+        within_av = np.mean([p1,p2])
+        between = [self.tmrca(u,v) for u,v in itertools.product(sample_p1, sample_p2)]
+        between_mean = np.mean(between)
+        data_av = np.mean([within_av,between_mean])
+        
+        return (data_av - within_av)/data_av
+
 
 def load(path):
     """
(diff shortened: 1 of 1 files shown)


petrelharp:
Gee, thanks for the contribution. Fst is totally something that people will want to compute. However, I'm not sure if this is how we want to do it. I don't think that this calculation corresponds to any of the many versions of Fst people calculate.

Notation: let
- `p1` and `p2` be a set of at least two samples each, nonoverlapping
- the window of the genome under consideration be of length `L`
- 

The easiest way to compute an Fst value from sequence in the window would be to let
- `A = (diversity(p1) + diversity(p2))/2`, where `diversity(p1)` is the mean density of nucleotide differences between pairs of samples from `p1`
- `B = divergence(p1, p2)`, where this is the mean density of nucleotide difference between a sample from `p1` and a sample from `p2`
- `Fst = 1 - A / B`

Since the mean density of nucleotide differences over a region between `x` and `y` is the mutation rate times the average distance in the tree between them, the "branch length" version of this is obtained by replacing "density of nucleotide differences" with "mean length in the tree". For instance, to compute `divergence([x],[y])` you would average, across the region, `2 * tree.tmrca(x,y) - tree.time(x) - tree.time(y)`.

This is almost what you have. But, you're (a) computing A and B separately for each tree, then (b) averaging (1 - A/B) across trees. This is different than computing A and B, separately, as averages across the entire window, then taking (1 - A/B).  What you do is closer to (but not the same as) a thing that people do sometimes also, which is to compute (1 - A/B) separately for each SNP then average these across SNPs.

So, I would implement windowed Fst just using the existing [`divergence`](https://github.com/tskit-dev/tskit/blob/master/python/tskit/stats.py#L175) function; you'd then get it for SNPs and for trees both for free. But, maybe you had a different definition of Fst in mind?

jeromekelleher:
To give you a bit of background on this @petrelharp, this Fst calculation is for a speciation workshop that @GertjanBisschop, @KLohse, @DRL and others are running in a few weeks. We want to get an efficient implementation of branch-length based Fst in here so that we can run windowed Fst scans for large tree sequences.

Re. the definition, your one seems a lot more attractive @petrelharp , since we can use the general machinery to do it. But, maybe the current one is more appropriate for this application. @Klohse, is this the difference you were mentioning between the Barton definition and everyone elses? We can definitely implement two definitions of Fst if we need to, we don't have to choose one.

KLohse:
Thanks a lot for your thoughts on this @petrelharp, super useful! I think there are two separate issues:

1) which definition of Fst we want to use. An alternative is: Fst =  (Tt -Tw)/Tt

where Tw= A in your notation and Tt is the average for pairwise Tmrca for a pairwise sample picked at random. For the case of two populations Tt = (A+B)/2 and so Fst = (B-A)/(B+A).
In the limit of large number of demes Tt -> B and so the two definitions become the same. I am not attached either one of the two definitions. Fst = 1 - A / B is simpler, Fst = (B-A)/(B+A) seems more popular with the speciation folk (who mainly think about two pops). We can pick either (or even better both) definitions as long as it's clear which one.

2) how to do the averaging in a window. @petrelharp is of course absolutely correct: we want to average A and B across the window and before computing Fst which is different than averaging Fst. It's great that we can use the existing  divergence function for this.

GertjanBisschop:
@petrelharp <https://github.com/petrelharp>, I was overcomplicating things. Obviously everything we need is already implemented and we can just use the divergence_matrix function. 

As mentioned in tree_stat_vector <https://github.com/tskit-dev/tskit/blob/2cca87badb75216807eb53e545bc91c01facdfc8/python/tskit/stats.py#L596> documentation, if we want determine Fst using overlapping windows then all that is left is to average the entries of the divergence matrix for those non-overlapping windows that build up a larger window (that does overlap with other larger windows), and then calculate Fst.

Thanks for all the great work!

Gertjan

jeromekelleher:
Thanks for this @GertjanBisschop. We'll implement an Fst function ASAP using the general framework @petrelharp and I are working on and let you know when it's ready to use. I'm going to close this PR.

PR #190, opened 2019-05-27
First pass at general treewise stats definition algs.

jeromekelleher:
These are not tested properly and mostly likely contain bugs.

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -38,21 +38,22 @@ import tskit
 import tests.tsutil as tsutil
 
 
-def windowed_tree_stat(ts, A, windows, stat_func):
-    tree = ts.first()
-    stat = stat_func(tree)
+def windowed_tree_stat(ts, stat, windows):
+    A = np.zeros((windows.shape[0] - 1, stat.shape[1]))
+    tree_breakpoints = np.array(list(ts.breakpoints()))
+    tree_index = 0
     for j in range(windows.shape[0] - 1):
         w_left = windows[j]
         w_right = windows[j + 1]
         while True:
-            t_left, t_right = tree.interval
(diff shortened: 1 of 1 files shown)


jeromekelleher:
cc @petrelharp

petrelharp:
An issue with "depolarization": under the previous method, divergence always agreed with [what you'd get from Hamming distance](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_tree_stats.py#L367). Even like [in this test case](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_tree_stats.py#L943) when the two samples differ, but neither has the ancestral allele. 

An even simpler example:
```
node    allele
ancestral  0
0               1
1               2
```
The two sequences are `1` and `2`; the site is biallelic and the sequenes differ at 100% of sites, so divergence should be 1.0, the different ancestral allele notwithstanding.

Recall that before, our function was `f(x) = (x[0] / n[0]) * (1 - x[1] / n[1])`, the probability that sample 0 carries the allele in question and sample 1 does not. Summed over all alleles (including the ancestral one) this gives you divergence.

But, now we want to only sum over *derived* alleles, so we've set `f(x) = (x[0] / n[0]) * (1 - x[1] / n[1]) + (1 - x[0] / n[0]) * (x[1] / n[1])`. However, in the situation above, when *both* alleles are derived, this gives twice the correct answer.

jeromekelleher:
Hmmm. Shall we merge this much and discuss later, or keep this open until we're sure the definition is correct?

petrelharp:
Peter to-do:
- unpolarize this code
- unpolarize the manuscript
- swap jerome's python functions in for mine
- write a test using weights

jeromekelleher:
Merging after in person code review.

PR #191, opened 2019-05-28
Define branch length over root to be zero.

### LINKED ISSUE #188: Define branch length over a root = 0?
We have to do annoying things when dealing with branch lengths, as we assume the parent exists. I think we should define the branch length as zero when the argument is a root. Can you see any pitfalls @petrelharp?

jeromekelleher:
Closes #188.


jeromekelleher:
Any objections to this @hyanwong? Seems like a straightforward improvement that's unlikely to break people's code (I doubt anyone depends on the current exception-throwing behaviour).

hyanwong:
I have no objections.

PR #194, opened 2019-05-29
implement statistics in C

petrelharp:
Work in progress.


petrelharp:
This is non-deterministic:
```
python3 -m nose tests/test_tree_stats.py:BranchLengthStatsTestCase.test_branch_Y_stats
```

petrelharp:
Proposal for the API:

Common inputs:
 - `windows` : either a increasing sequence of breakpoints or None (which means `[0, sequence_length)` or "sitewise" or "treewise"
 - `stat_type`: either `site` or `branch` or `node`

Output:
 - a `num_windows` x `m` array, where `m` is determined by the function

1. Defaults, for windows:
```
def parse_windows(windows, stat_type):
    # here's one way to get the common code into each function
    if windows is None:
        windows = [0.0, self.sequence_length]
    if windows == "treewise":
        windows = self.breakpoints()
    if windows == "sitewise":
       # do breakpoints that are midpoints of the windows
        x = self.tables.sites.position
        x.append(self.sequence_length)
        windows = x[:-1] + np.diff(x)/2
        windows[0] = 0.0
    # check validity
   return windows
```

2. Functions that compute one thing for each of a list of sets of samples:
Input:
- `sample_sets` - list of `m` sets of samples
- `windows`, `stat_type`
Output:
- either a `num_windows x m` array (if `stat_type` is `site` or `branch`), or
- a `num_windows x num_nodes` array, if `stat_type` is `node` and `m=1` (error otherwise)

3. Functions that compute one thing for each of a set of `k`-tuples of sets of samples:
Input:
- `sample_sets` - list of `p` sets of samples
- `indices` - list of length `m` of `k`-tuples of indices between `0` and `p-1` or None (in which case `p` should be the right thing for the funciton in question)
- `windows`, `stat_type`
Output:
- either a `num_windows x m` array (if `stat_type` is `site` or `branch`), or
- a `num_windows x num_nodes` array, if `stat_type` is `node` and `indices` is `None` (error otherwise)

petrelharp:
I spent far too long being tripped up by normalization in windowing.  Here's the upshot:

When returning windowed statistics, we should *always* return statistics that are *averaged* across the window. So, if you have two windows, with breakpoints `a < b < c`, and the value of the statistic on the two windows are `w[0]` and `w[1]`, then the value of the statistic on the window between `a` and `c` should be ``(b-a) * w[0] + (c - b) * w[1]) / (c - a)`.  This can be confusing because it's also natural to think about windows being additive - they *almost* are.

The exception is, of course, `sitewise` site statistics. Those are not returned for windows, so they don't count.  And, also, this means my proposal to just stick breakpoints between the positions of the sites doesn't work, we have to treat `windows="sitewise"` a little bit special. But, I think this is OK.

But: this is passing the tests, now! (I think - I didn't try all of them because of low power).

petrelharp:
Darn it, the tests **were** passing, until I had to merge in the changes.

python/tskit/trees.py
```diff
@@ -3347,15 +3347,8 @@ def divergence(self, sample_sets, indices=None, windows=None, stat_type="site"):
...
-        if stat_type == "node":
-            if indices is not None:
-                raise ValueError("cannot specify more than one statistic with nodes")
-        if indices is None:
-            if len(sample_sets) != 2:
-                raise ValueError("Must have 2 sample sets.")
-            indices = [(0, 1)]
-
-        windows = parse_windows(windows, stat_type)
+        sample_sets, indices = self.check_input(sample_sets, indices, 2)
```
petrelharp: The `check_input( )` function does this sort of thing. (although it doesn't do the "node" check, 'cause we don't have that implemented yet).

jeromekelleher:
> That was there for a reason:

I realised that afterwards! I ended up using ``isinstance(windows, str) and windows == "treewise"``, which is safe. Using ``windows is "treewise"`` is dangerous as the language doesn't guarantee that there aren't two copies of a string lying around in memory.

jeromekelleher:
Making progress here @petrelharp - the branch stats are all going through the generic C code now (with Python summary funcs, so still slow). The plan is to implement the site version of the general algorithm in C node and follow the same pattern. Should be easier, hopefully.

petrelharp:
Looking good. I made a python implementation of the node statistics. Not terribly well tested, yet.  Next on my list is to write an Fst function and do some more testing cleanup, and expanding coverage.

jeromekelleher:
OK, we're getting there @petrelharp! This adds a full-stack version of the diversity function for site and branch stats, and the C API is starting to take shape. Some comments/questions:

1. I left out the ``indexes`` parameter for the diversity function, as this seems like a pointless bit of complexity to me (which your documentation agrees with!). If someone wants to permute the columns in the  output, surely they just do it themselves afterwards?
2. I've added an option ``TSK_SPAN_NORMALISE`` to the low-level code, which gives us the option of normalising by the window span. This options is needed for sitewise windowing. My thinking on this is that, by default, we'll accumulate statistics in the windows. If SPAN_NORMALISE is set, we weight contributions by their span (if meaningful) and divide by window span afterwards. Perhaps a COUNT_NORMALISE would also make sense here, where we just take the simple average of stats in windows? SPAN_NORMALISE isn't implemented correctly right now as I wanted to see what you thought of the idea.
3. I've generalised the C API, and made the stats "mode" one of the options that can be specified. What do you think of "mode" rather than "stat_type"? Stat_type seems a bit redundant (we already know it's a stat), and "type" can be confusing in terms of float64, float32 etc. Mode seems good as it's nice and short, and we can enumerate the different modes and what they mean fairly concisely.
4. I'm having trouble figuring out how the stats tests are fitting together - are you in the middle of a cleanup?

petrelharp:
Hm. There's an awful lot of python glue for each summary function, like [for diversity](https://github.com/tskit-dev/tskit/pull/194/files#diff-4e46c6c971de423808a89bb94ca54779R6629). Maybe this isn't your long-term plan, but seems too bad given there is so little [at the C level](https://github.com/tskit-dev/tskit/pull/194/files#diff-a8f7b6e9a7a088188190be268161f93eR1603). 

Instead of defining separate methods for the `ll_tree_sequence` for each statistic, couldn't we just define a single method, like `ll_tree_sequence.named_general_stat( )` that you pass through a key that it uses to look up which function to use?  e.g., we could have a dict in tskit (`STAT_NAMES`) whose keys are stat names and whose values are an integer that tells the C code which summary function to use: so, [here](https://github.com/tskit-dev/tskit/pull/194/files#diff-993cce8b29b806d11aae4e2ead3c71ecR3294) we would 
```
            stat_name_flag = STAT_NAMES["diversity"]
            return self.ll_tree_sequence.named_general_stat(stat_name_flag,
                sample_set_sizes, flattened, windows, mode=stat_type)
```
and then `named_general_stat( )` would call the C function `tsk_named_general_stat( )` which would just be something like
```
int
tsk_treeseq_named_general_stat(tsk_treeseq_t *self, tsk_size_t stat_name,
        tsk_size_t num_sample_sets, tsk_size_t *sample_set_sizes, tsk_id_t *sample_sets,
        tsk_size_t num_windows, double *windows, double *result, tsk_flags_t options)
{
    return  tsk_treeseq_sample_count_stat(self,
        num_sample_sets, sample_set_sizes, sample_sets, NULL, 0,
        summary_funcs[stat_name],
        num_windows, windows, result, options);
}
```
... ?

petrelharp:
> 1.    I left out the indexes parameter for the diversity function, as this seems like a pointless bit of complexity to me (which your documentation agrees with!). If someone wants to permute the columns in the output, surely they just do it themselves afterwards?

Well, this was just there for consistency with all the other statistics. Which at least makes testing more convenient. Leaving it out is a good idea if it makes things simpler; so let's leave it out and see if we want to add it back in later.

> 2.    I've added an option TSK_SPAN_NORMALISE to the low-level code, which gives us the option of normalising by the window span. This options is needed for sitewise windowing. My thinking on this is that, by default, we'll accumulate statistics in the windows. If SPAN_NORMALISE is set, we weight contributions by their span (if meaningful) and divide by window span afterwards. Perhaps a COUNT_NORMALISE would also make sense here, where we just take the simple average of stats in windows? SPAN_NORMALISE isn't implemented correctly right now as I wanted to see what you thought of the idea.

I think that `SPAN_NORMALISE` is a good idea. At first it seemed like premature generalization to me, but of course we actually need it for `"sitewise"` statistics. By `COUNT_NORMALIZE` do you mean, like, taking the simple average over sites? So, it'd be the same thing as doing the normal windowing (which divides by window length), then multiplying by (window length) / (number of sites per window)? If so, I say we do *not* add this flag - it's unnecessary complexity  - but make it easy for people to get number of sites per window and make sure to write good documentation on how to do this.

> 3.    I've generalised the C API, and made the stats "mode" one of the options that can be specified. What do you think of "mode" rather than "stat_type"? Stat_type seems a bit redundant (we already know it's a stat), and "type" can be confusing in terms of float64, float32 etc. Mode seems good as it's nice and short, and we can enumerate the different modes and what they mean fairly concisely.

I like it. We should update the paper to reflect this also. It'll be a bit more awkward until we get used to it, but it emphasizes more the parallelism between different types, er, modes.

> 4.    I'm having trouble figuring out how the stats tests are fitting together - are you in the middle of a cleanup?

I have in my head how to do that cleanup, but haven't done it. I didn't want to do that rearrangement in case you were editing the tests. I'll do it later today.

python/tests/test_tree_stats.py
```diff
@@ -1204,20 +1204,24 @@ class GeneralStatsTestCase(StatsTestCase):
...
+        # print("Sample_sets = ", sample_sets)
+        # print("indexes", indices)
+        ts_fn(sample_sets, indices, windows, stat_type=self.stat_type)
+        # print("Got", ts_vals)
+        # JK: commenting this out as I'm not sure what it's testing for.
+        # self.assertEqual((len(windows) - 1, len(indices)), ts_vals.shape)
+        # leafset_args = [[sample_sets[i] for i in ii] for ii in indices]
+        # py_vals = np.column_stack([py_fn(*a, windows) for a in leafset_args])
+        # self.assertEqual(py_vals.shape, ts_vals.shape)
+        # self.assertArrayAlmostEqual(py_vals, ts_vals)
```
petrelharp: We use this to compare the tree sequence function (eg `ts_fn = ts.diversity( )`) to the simple python version defined here (eg `py_fn = PythonBranchStatCalculator.diversity( )`).

petrelharp:
General structure of `test_tree_stats.py`:
1. Python methods to compute everything, including boneheaded add-up-branch-lengths-on-all-the-trees ways, in classes like `PythonBranchLengthCalculator`.
2. Classes like `TestGeneralBranchStats` that test the general statistics methods (with weights and f's)
3. Classes that extend `SampleSetStatTestCase` to test the "sample set and indices" statistics (same structure of the tests for each of the three modes, so testing the three modes gets to re-use a lot of code).
4. Some specific trees that are easy to look at and tell what the right answer is.

I thought about merging (2) and (3) but it seemed like they were testing different enough things I kept them separate.

python/_tskitmodule.c
```diff
@@ -6655,6 +6655,8 @@ TreeSequence_diversity(TreeSequence *self, PyObject *args, PyObject *kwds)
     if (parse_stats_mode(mode, &options) != 0) {
         goto out;
     }
+    /* TODO this should always be span normalised? */
+    options |= TSK_STAT_SPAN_NORMALISE;
```
petrelharp: This was needed to make this correct.

I guess we need to decide if *all* stats have the `span_normalized` option. I guess so, but it's at the cost of making things more complex.

petrelharp: This was needed to make this correct.

I guess we need to decide if all stats have the span_normalized option. I guess so, but it's at the cost of making things more complex.

python/tests/test_tree_stats.py
```diff
@@ -254,9 +254,13 @@ class PythonNodeStatCalculator(object):
     def __init__(self, tree_sequence):
         self.tree_sequence = tree_sequence
 
-    def divergence(self, X, Y, windows=None):
+    def divergence(self, sample_sets, indices=None, windows=None):
```
petrelharp: I switched the interface on all these testing methods to mirror the real one (easier testing).

jeromekelleher:
Quick note here @petrelharp - it's fast!
```python
ts = tskit.load("large-example.trees")
print("samples = ", ts.num_samples)
print("sites   = ", ts.num_sites)
print("trees   = ", ts.num_trees)

before = time.perf_counter()
pi1 = ts.pairwise_diversity()
print("Old pi = ", pi1, "in {:.3f}s".format(duration))

before = time.perf_counter()
pi2 = ts.diversity([ts.samples()])
print("New pi = ", pi1, "in {:.3f}s".format(duration))
```
gives

```
$ python3 tmp.py
samples =  100000
sites   =  484395
trees   =  456789
Old pi =  40022.3295904093 in 0.574s
New pi =  40022.3295904093 in 0.356s
```

I was expecting the new version to be signficantly slower as I'd left a lot of low-hanging performance fruit lying around. My guess is that the old version uses the general tree code for generating the trees, which is a lot more complicated than the super-streamlined version used for these stats. Turns out this more than compensates for the extra overhead of function pointer dereferences and mallocs.

Anyway, it's fast enough for now, so no need to do any perf work before we release.

python/tests/test_tree_stats.py
jeromekelleher: I was just commenting it out temporarily to get things working - figured you'd sort it out.

jeromekelleher:
Collecting various comments in here

> Hm. There's an awful lot of python glue for each summary function, like for diversity. Maybe this isn't your long-term plan, but seems too bad given there is so little overhead per statistiic at the C level.

Unfortunately glue is inevitable with the Python C API. I've refactored things a bit to push common code for parsing windows and stuff into functions, which helps a lot. If functions have different parameters though, there's not much to be done other than just have different code for ingesting those parameters. I think what we have now is fine, and will scale to more stats.

> Well, this was just there for consistency with all the other statistics. Which at least makes testing more convenient. Leaving it out is a good idea if it makes things simpler; so let's leave it out and see if we want to add it back in later.

I think it'll be confusing for users, and any convenience we win now for ourselves in terms of testing we'll pay back many times by having to answer questions about what it's for and why you'd use it.

Also, I've gone with ``set_indexes`` in the low-level code because (a) it's a bit more descriptive and (b) we've been using "indexes" rather than "indices" elsewhere. I don't have a strong opinion about indexes vs indices, but we shouldn't mix them and indexes is in there already. Are you OK with changing the high-level code to use this also?

> I like it. We should update the paper to reflect this also. It'll be a bit more awkward until we get used to it, but it emphasizes more the parallelism between different types, er, modes.

Re mode: yes, great, let's propagate this terminology everywhere!

>  think that SPAN_NORMALISE is a good idea. At first it seemed like premature generalization to me, but of course we actually need it for "sitewise" statistics. By COUNT_NORMALIZE do you mean, like, taking the simple average over sites? So, it'd be the same thing as doing the normal windowing (which divides by window length), then multiplying by (window length) / (num...

This is my suggestion for the semantics:

- By default in all modes, we do a straightforward accumulation of statistics to bins. For site stats, this gives us the total value for all sites in the bin. For branch stats, it gives us the total value for each tree intersecting with the bin. With sitewise windows and mode=site, this gives us the per-site value. With treewise windows and mode=branch this gives us the per-tree value. 
- If SPAN_NORMALISE is set, we accumulate per bin and divide by the size of the bin. For treewise stats, the value added to the bin is weighted by the tree span intersection.
- If COUNT_NORMALISE is set, we accumulate per bin (NB, no span weighting for branch stats) and divide by the number of things we put into each bin. This will give you the per-tree average for the stat.

What do you think? The semantics are a mess at the moment, and this seems like a good way to pulling things apart. I don't think the COUNT_NORMALISE thing is trivial to reproduce when you think about mode=branch, and it seems useful to me.

python/_tskitmodule.c
jeromekelleher: see comments below on general normalisation semantics.

jeromekelleher:
BTW, just implemented divergence and it's also extremely fast. Hooray!

I'll get my head around the node stats next and we should be one the home straight then.

jeromekelleher:
Stuff from chatting:

1. 1-way functions don't need indexes, so we leave them out.
2. All other functions should have simular APIs so we'll be able to refactor these to use mostly common code in _tskitmodule.c.

petrelharp:
Also: by default we normalize, and have an option `unnormalized=False` that can be used to not do this.  I've updated [the docs](https://github.com/tskit-dev/tskit/pull/194/files#diff-b85c62cc704cd769f176f23cde1a8925) to explain this (and the rationale).

jeromekelleher:
I've implemented the node stats in C @petrelharp - unfortunately I broke some stuff in the tests, but hopefully the way forward is pretty clear now.

The main thing I've done is change the node stats to return a 3D array. It seemed easier to do this than to force the output of f to be one-dimensional and think through the forward-compatability issues at the C level. We would surely have wanted to make this generalisation pretty soon. The array indexes of the output are A[window, node, summary]. The Python algorithms all generalised nicely this way. I also updated the efficient algorithm to only incur the O(num_nodes) cost when we cross windows. This makes the (presumably) common case of lots of trees per window much more efficient.

There's still some issues though - the various algorithm versions aren't currently agreeing. I don't think the efficient algorithm matches the naive one in the unpolarised case, so it'd be great if you could take a look at that.

I've broken a bunch of tests unfortunately, as I started doing some refactoring but ran out of time. Sorry about that.

jeromekelleher:
Things are moving along here @petrelharp. I need to get an alpha version of this stuff shipped on Tuesday for the workshop I'm running next week, so here's a summary of what's left to do.

- Finish porting the remaining stats definitions (f4, f3, Y3)
- Generalise the code for manipulating input in trees.py, rename ``indices`` to ``indexes`` and add back default behaviour. Also rename ``stat_type`` to ``mode``
- Test/numpify ``divergence_matrix`` method (might just 'undocument for the beta).
- Add/port SFS code
- Port remaining test cases and clean up.
- Update docs to reflect recent changes

Longer term things to sort out:

- There's a few issues with node stats - we seem to basically get different answers whatever way we compute them.
- Settle on API and semantics for general node stats.
- Settle on windowing and normalisation semantics (see above for argument to decouple windowing and normalisation entirely)

petrelharp:
>  Policy for division by zero - return 0 or nan?

Definately `nan`. There's use cases when you want `0` here, but you should have to do it intentionally. Same reason missing values propagate by default in R. (too bad numpy doesn't have missing values!)

petrelharp:
>  Test/numpify divergence_matrix method (might just 'undocument for the beta).

Do we even need `divergence_matrix`? It would be nice to have *some* way to make it easy for people to get all combinations, but maybe that could be another argument? Like, `indices = "all"`?

petrelharp:
Other things (from email):

- [ ] Change the `windows` text-string options to just "trees" or "sites", which are just shortcuts. Using "sites" just gives the site positions + endpoints, so to get the current "sitewise" behavior they need to also pass `normalise=False` (with maybe a zero value at the end).

petrelharp:
> The main thing I've done is change the node stats to return a 3D array. 

Ok. Sounds good.

petrelharp:
Also:

- [ ] allow windows that don't start at 0.0 and end at sequence_length

I think this should be short-term, so we don't have to have a funny caveat about `windows="site"` that says it may or may not have a 0 at the start.

petrelharp:
Also:

- [] decide if the syntax is ``normalised = True`` or ``unnormalised = False`` (I think the first but I wrote the docs the other way)

c/tskit/trees.c
```diff
@@ -947,87 +984,1277 @@ tsk_treeseq_mean_descendants(tsk_treeseq_t *self,
...
+{
+    sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
+    const double *x = state;
+    double n;
+    size_t j;
+
+    for (j = 0; j < state_dim; j++) {
+        n = (double) args.sample_set_sizes[j];
+        /* TODO: what happens when n is zero ? */
+        result[j] = x[j] * (n - x[j]) / (n * (n - 1));
```
petrelharp: The denominator totally doesn't have to be in this function; and it seems like this is creating lots of extra multiplications and divisions. But I ran some (relatively small) checks and moving the `/ (n * (n - 1))` into `tsk_treeseq_diversity( )` had almost zero effect on run time. I'm surprised! I was all set to write a `column_normalize` function that we could use to do the normalizations, but perhaps we don't need to.

As for the `NaN`s, it looks like they are being correctly assigned. This is maybe [implementation dependent](https://www.gnu.org/software/libc/manual/html_node/Infinity-and-NaN.html). This is supposed to [throw an exception](https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html#FP-Exceptions) that I guess we are not trapping.

jeromekelleher:
> Do we even need divergence_matrix? It would be nice to have some way to make it easy for people to get all combinations, but maybe that could be another argument? Like, indices = "all"?

It seems like a nice bit of functionality - let's push it to an issue for follow up.

c/tskit/trees.c
jeromekelleher: Each call to ``f``  currently does something like four mallocs and a two or three function calls. I'd imagine a function call costs at least 10 - 100 times as much as a division. A malloc surely costs 10 times as much again.

So, the divisions will almost certainly never be significant perf-wise. Perhaps there are numerical reasons for pulling them out? (Or conversely, for *not* pulling them out? Do we want to accumulate large values and divide afterward?)

jeromekelleher:
> I think this should be short-term, so we don't have to have a funny caveat about windows="site" that says it may or may not have a 0 at the start. Edit: I think we should not have the funny caveat about windows="site" (so, don't return that zero at the start sometimes), but we could do that after the fact, by removing it before returning it if it is there, to get this alpha out there.

Agreed - I've opened an issue to track this for the real release (#202)

petrelharp:
I've got the test functions dealing with NaNs correctly - I think - but now I see what you mean about it being tricky: the other functions often return 0 when it should be NaN, because for instance if a ts has no Sites, then the f( ) never gets called in a site stat, and that's where the division by zero is. This *also* seems correct; I'm not sure which way to resolve (probably make the test "definitions" conform?).

jeromekelleher:
Doesn't look like you pushed the NaNs stuff @petrelharp. I've been working on it also, but it looks like we're mostly doing the right thing at the moment, so it's just making the tests see this and suppressing warnings.

jeromekelleher:
OK, going to merge this beast and follow up with a few minor changes needed for the alpha. FYI, I squashed down the commits a bit @petrelharp, but I didn't squash over your commits.

c/tskit/trees.c
petrelharp: Since it's all an overall normalization, the order of operations shouldn't matter for floating point error, I think: (x/N - y/N) and (x - y)/N are the same.  Ok, we'll leave well enough alone.

Issue #148, opened 2019-03-15
Modifying GNN to work for internal nodes


awohns:
The referenced line should be p=u if we want GNN to work for internal nodes. As it is now, the code automatically jumps up to the parent of u when looking for v

https://github.com/tskit-dev/tskit/blob/4d8a8c821f2d0b76c971d62a6bf54abc0e21beba/python/tests/tsutil.py#L716

jeromekelleher:
Can you come up with a simple example of this failing please @awohns?

hyanwong:
I've been chatting about this with Wilder. It depends a bit on what you expect a "nearest neighbour" to mean. Something like this gives an example:

```
> ts = msprime.simulate(4, random_seed=123)
> print(ts.first().draw(format="unicode"))
 6      
┏━┻━┓    
┃   5    
┃ ┏━┻┓   
┃ ┃  4   
┃ ┃ ┏┻┓  
2 0 1 3  

> set1 = [1,3]
> set2 = [0,2]
> ts.genealogical_nearest_neighbours([4],[set1, set2])
> # In this case the closest relatives to 4 are its immediate descendants, 1 and 3
> # so we might expect the GNN to be entirely in the first set (i.e. [1,0])
> ts.genealogical_nearest_neighbours([4],[set1, set2])
array([[0.66666667, 0.33333333]])
> # this is more like what we expect for "closest relatives"
> ts.genealogical_nearest_neighbours([3],[set1, set2])
array([[1., 0.]])
```

I guess it depends if by "neighbour" you also mean descendant.

hyanwong:
By the way, there is another small issue with the GNN calculation: if the focal node is not in any of the reference sets, then we probably shouldn't define _v_ as we do in the paper, which says "_v_ is the first node in _t_ on the path to root from _u_ such that C^t_v > 1". Instead in this case we should use C^t_v > 0. Or to put it more generally, the condition for C^t_v should use the Iversonian condition from the GNN definition in the paper.

C^t_v > [u∈􏰁U^K_j=1 R_j]

hyanwong:
The original issue flagged up here isn't solved - I think the issue probably needs reopening or addressing.

```
> ts = msprime.simulate(4, random_seed=123)
> print(ts.first().draw(format="unicode"))
 6      
┏━┻━┓    
┃   5    
┃ ┏━┻┓   
┃ ┃  4   
┃ ┃ ┏┻┓  
2 0 1 3  

> set1 = [1,3]
[...]

PR #197, opened 2019-06-05
Address #148

hyanwong:
If we ask for the GNN of an internal node then we need not go to the parent before looking for descendants. Note that the word "neighbour" in GNN might be a bit misleading here, but it is the most logical thing to do. Indeed, if we ask for the GNN of the root node there is no defined behaviour otherwise (as it has no parent).


jeromekelleher:
Excellent, thanks @hyanwong!

PR #219, opened 2019-06-11
fixed tests with multiple roots by change to path length

### LINKED ISSUE #205: Y stat definitions cannot handle multiple roots
Some tests are currently disabled for Y stats as the definitions do not work when the ts has multiple roots. These need to be updated.

@petrelharp, this seems like one for you?

petrelharp:
This fixes #205.


jeromekelleher:
Very elegant!

PR #220, opened 2019-06-11
Fst implementation (and doc fixes)

### LINKED ISSUE #199: Implement pairwise Fst
Having thought some more about the two alternative Fst definitions, we definitely want to general definition of Fst in terms of expected pairwise coalescence times:

Fst= (Tt-Tw)/(Tt)

Where Tt is the average for a pairwise sample taken at random. For the case of two populations we assume that we are equally likely to sample from within and between species, i.e. 
Tt=(Tb + Tw)/2 

Substituting gives:

Fst= (Tb-Tw)/(Tb+Tw)

@GertjanBisschop wrote a simple implementation of this using BranchStat.divergence_matrix:

def Fst(div_entry):
    A = np.mean((div_entry.diagonal()))
    B = div_entry[1,0]
    return (B-A)/(A+B)

def calculate_Fst(ts, samplesets, function=Fst):
    winlen = [0,ts.sequence_length]
    BranchStat = tskit.BranchLengthStatCalculator(ts)
    div_matrix = BranchStat.divergence_matrix(samplesets,winlen)
    Fst_result = list(map(function, div_matrix))
    return Fst_result

Hopefully this can be implemented in tskit to return a matrix of pairwise fst values for arbitrary sets of populations.

### LINKED ISSUE #214: Stats docs: correct sample_sets param docs
Stats use the string "– A list of lists of Node IDs, specifying the groups of individuals to compute diversity within." 

This doesn't apply to things like Y1, etc.

### LINKED ISSUE #217: Change indices to indexes in general stat docs
Currently used in the general stats overview.

petrelharp:
This is an implementation of Fst: the "1 - diversity/divergence" version. Since this is just using our diversity and divergence functions - it isn't a "general stat" itself, but a ratio of them - I have only tested one case (but one that people might care about - `windows="sites"`).

This closes #199.

Also:
closes #214 
closes #217 

In the process I ran into a question: what if there are no sites, what should be returned with `windows="sites"`? Currently it returns something with one row, but it should probably have zero rows.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -140,7 +140,7 @@ in units of crossovers and mutations per base pair, respectively).
 .. _sec_general_stats_sample_sets:
 
 ***********************
-Sample sets and indices
+Sample sets and indexes
 ***********************
 
 Many standard population genetics statistics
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Looks good, thanks @petrelharp. Will merge once CI passes (I rebased and removed a stray print).

petrelharp:
CI passed; merging. Thanks for cleaning up that print!

Issue #211, opened 2019-06-11
Document NaN behaviour in general stats


jeromekelleher:
NaNs are generated when division by zero happens in stats. Documnent this.

petrelharp:
Also make this consistent between C and python. The difficulty is that the `nan`s come from function evaluation, but if the function is never evaluated, it ends up being zero; whereas in python the divisions often come outside this, so even e.g. nodes that are never visited in a `node` statistic can end up with `0/0`.

petrelharp:
I've added a note about this in #221.

petrelharp:
fixed (could maybe be better tho') in cd3023920f0237f1599e15309c3f9cc95896a767

Issue #206, opened 2019-06-11
Node stat definitions for Y and f stats


jeromekelleher:
There are currently no definitions for Node and f stats in the tests. We need these for full testing. See also #204 and #205.

petrelharp:
fixed in cd3023920f0237f1599e15309c3f9cc95896a767

PR #221, opened 2019-06-12
Implement node statistics

### LINKED ISSUE #204: Node stats
Node stats are currently partially implemented in #194. Finish and test.

petrelharp:
There was a minor bug in the set-up. Fixed! :yay:

Closes #204 and #206 but disables a bunch of tests involving messy wf trees for which the python and C versions of node stats don't agree on where NaNs go: see #211.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1559,13 +1559,17 @@ tsk_treeseq_node_general_stat(tsk_treeseq_t *self,
 
     /* Set the initial conditions */
     for (j = 0; j < self->num_samples; j++) {
+        // total_weight is used in the next loop
+        weight_u = GET_2D_ROW(sample_weights, state_dim, j);
+        for (k = 0; k < state_dim; k++) {
+            total_weight[k] += weight_u[k];
+        }
+    }
+    for (j = 0; j < self->num_samples; j++) {
         u = self->samples[j];
         state_u = GET_2D_ROW(state, state_dim, u);
         weight_u = GET_2D_ROW(sample_weights, state_dim, j);
         memcpy(state_u, weight_u, state_dim * sizeof(*state_u));
-        for (k = 0; k < state_dim; k++) {
-            total_weight[k] += weight_u[k];
-        }
         ret = update_node_summary(u, result_dim, node_summary, state, state_dim,
             f, f_params, total_weight, polarised);
         if (ret != 0) {
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1564,12 +1564,12 @@ tsk_treeseq_node_general_stat(tsk_treeseq_t *self,
         for (k = 0; k < state_dim; k++) {
             total_weight[k] += weight_u[k];
         }
-    }
-    for (j = 0; j < self->num_samples; j++) {
         u = self->samples[j];
         state_u = GET_2D_ROW(state, state_dim, u);
         weight_u = GET_2D_ROW(sample_weights, state_dim, j);
         memcpy(state_u, weight_u, state_dim * sizeof(*state_u));
+    }
+    for (u = 0; u < (tsk_id_t) num_nodes; u++) {
         ret = update_node_summary(u, result_dim, node_summary, state, state_dim,
             f, f_params, total_weight, polarised);
         if (ret != 0) {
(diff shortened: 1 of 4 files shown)


c/tskit/trees.c
```diff
@@ -1559,13 +1559,17 @@ tsk_treeseq_node_general_stat(tsk_treeseq_t *self,
 
     /* Set the initial conditions */
     for (j = 0; j < self->num_samples; j++) {
-        u = self->samples[j];
-        state_u = GET_2D_ROW(state, state_dim, u);
+        // total_weight is used in the next loop
         weight_u = GET_2D_ROW(sample_weights, state_dim, j);
-        memcpy(state_u, weight_u, state_dim * sizeof(*state_u));
         for (k = 0; k < state_dim; k++) {
             total_weight[k] += weight_u[k];
```
petrelharp: the problem was that total_weight wasn't done being initialized by the time that update_node_summary was called on some of the nodes. grr!

petrelharp:
The modifications to the general node algorithm are very nice, btw.

python/tests/test_tree_stats.py
```diff
@@ -577,16 +564,19 @@ def test_single_tree(self):
         ts = msprime.simulate(6, random_seed=1)
         self.verify(ts)
 
+    @unittest.skip("inconsistent nan issues")
```
jeromekelleher: Hmm, looks like this is skipping nearly all the tests for every stat now, not just node stats. I guess this is a sign we should resolve the inconsistency before merging?

c/tskit/trees.c
jeromekelleher: Oops, sorry, my bad @petrelharp!

jeromekelleher:
Great that it was such a simple fix @petrelharp, hooray!

This PR is against the Fst branch rather than master though, so possibly the simplest thing to do is close this one and create a new PR against master?

petrelharp:
I merged the Fst branch and changed the base of this one to master. Will still need a rebase. But yes, let's get the `nan`s sorted.

The main question is whether the current C behavior for `nan`s is what we want. Here is what I think is going on: suppose that `f` returns `nan`. (In our current situations, it *always* returns nan because e.g. it divides by `n * (n-1)` and `n=1`; but there could be situations where it only sometimes returns nan.) So, there should be nan's everywhere... but, since we initialize the output to be all zeros, then for instance a window having no sites in it will remain 0 for a site statistic; likewise, a node that is in none of the trees in a window will remain 0.

So, I think the question comes down to what the *default* value should be, and the only two sensible options I can think of are `0` or `nan`.

Which one? Well, we *have* said that we want e.g. invariant sites to have f( ) = 0, so that we can think of site statistics as summing over *all* sites, including the invariant ones. If f() tells us the number of paths of some kind going through a node, then 0 is good. For other situations, it's maybe less good? For instance, the node-f4 statistic tells us how many paths of one kind *minus* how many paths of another kind go through the node. A zero obtained by cancelation is different than "no paths". But what we might really want to know here is how much of each window was the node actually in a tree. That is a quantity we can compute separately (using the function f() that asks if the node is above *any* of the samples or not).

After talking that through, I think the correct default depends on the situation, and no choice we make will be "best" in all situations. And, disambiguating the different cases can I think always be done easily with another statistic.

So: I vote to keep the behavior as-is (default = 0), documenting this and changing the python test functions to mimic it (this just requires moving the normalization inside the loops, I think).  How's that sound, @jeromekelleher?

jeromekelleher:
> So: I vote to keep the behavior as-is (default = 0), documenting this and changing the python test functions to mimic it (this just requires moving the normalization inside the loops, I think). How's that sound, @jeromekelleher?

Yes, agreed. Zero by default is surely the right behaviour in the vast majority of cases, so I think moving the norming within the loops in Python is the way forward.

petrelharp:
Ok, I'll get on this.

jeromekelleher:
Looks like you need to rebase against upstream/master here @petrelharp? You'll need o drop the first commit as this is already in there.

petrelharp:
Ok, I think this is ready to go, assuming the tests pass. :yay:

petrelharp:
o whoops, squashing commits...

jeromekelleher:
It's the "Fst implementation (and doc fixes)" that's causing the problems. If you 'drop' this and keep the other commits, I think it'll work fine.

petrelharp:
ok, for reals now.

jeromekelleher:
Looks ready to merge once CI passes @petrelharp?

petrelharp:
CI compiling is stricter than on my machine for some reason, so I didn't catch those warnings.

jeromekelleher:
This is fantastic, thanks @petrelharp! I'll push out another alpha tomorrow with this code in it.

PR #223, opened 2019-06-12
Merge pull request #1 from tskit-dev/master

brianzhang01:
Update my fork

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,7 @@ jobs:
             python setup.py egg_info
             python setup.py bdist_wheel
             pip install dist/*.tar.gz 
+            tskit --help
 
       - run:
           name: Compile C with gcc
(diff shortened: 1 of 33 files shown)


jeromekelleher:
Just FYI @brianzhang01, there's no real point in keeping your fork master in sync - it's much easier to base feature branches on upstream/master and rebase as need, as described [here](https://www.overleaf.com/project/58ad92aea7c44e52639510e4)

brianzhang01:
Ah okay, still figuring things out here. The Overleaf doesn't load for me (probably for good reason!); I think you may have meant to link [here](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow). So I just do `git checkout upstream/master` followed by `git checkout -b new_branch`, and that means it'll automatically copy over from whatever branch I was currently working on?

That's also nice because I think it saves me from doing rebasing if I only have one commit, whereas up to now I've always had to rebase because the pull request to my master counts as one commit.

jeromekelleher:
Aha, yes, bad copy-paste on my part! I've updated the link above in case anyone else tries to follow it.

Yes, by starting your feature branch based on upstream/master, most of the time you won't need to rebase at all if it's a small change and the upstream repo doesn't move too quickly.

PR #225, opened 2019-06-13
Add slice function

hyanwong:


jeromekelleher:
Very nice, thanks! I feel like we should have a numpified version which would be more efficient, as this will be slow when we have millions of edges. You're version is very clear though, so we should definitely keep it as a comparison for testing.

I'll take a look in a couple of days and see if the numpy version isn't too tricky.

hyanwong:
Right - not sure how many of these things you want in the C api too, so that's another option instead of the numpy route.

python/tskit/trees.py
```diff
@@ -3038,6 +3038,84 @@ def simplify(
...
+    def slice(
+            self, start=None, stop=None, rescale_coordinates=True, simplify=True,
+            record_provenance=True):
+        """
+        Truncate this tree sequence, returning a tree sequence that covers a reduced
+        genomic span. Sites not covered by the sliced region are thrown away.
+
+        :param float start: The leftmost genomic position, giving the (open) start point
+            of the kept region. Information along the genome prior to this point will
+            be discarded. If None, use the start position of the current tree sequence.
```
petrelharp: instead of `(open)` how about `giving the start point of the kept region (including sites at this point)`. And for `stop`, how about `giving the end point of the kept region (but not including sites at this point)`.

python/tskit/trees.py
```diff
@@ -3038,6 +3038,84 @@ def simplify(
...
+        :param float stop: The rightmost genomic position, giving the (closed) end point
+            of the kept region. Information along the genome at and past this point will
+            be discarded. If None, use the end position of the current tree sequence.
+        :param bool rescale_coordinates: Rescale genomic coordinates such that position
+            0 in the new tree sequence corresponds to position ``start`` in the old one.
+            All sites and tree intervals will thus change to reflect their new positions.
+            If ``False``, do not rescale: the resulting tree sequence will therefore
+            cover the same genomic length as the original, but will have tree and site
+            information missing for genomic regions outside the sliced span.
+            (Default: True)
```
petrelharp: I vote for not having this option at all, but if you do really want it, definately have the default as False. (Generally, I'd like to encourage people to use real coordinates; things are less confusing that way.)

python/tskit/trees.py
hyanwong: Yes, "open" and "closed" are a bit technical. But it's not just sites. This is applicable even without any mutations. So if I have a TS with no sites from 0.0 - 1.0, and I slice it down to 0.0 to 0.5, then there will be no tree at `sliced_ts.at(0.5)`

python/tskit/trees.py
hyanwong: It seems a little confusing to have a tree sequence with missing trees for certain sections. When simulating, I'm most likely to rescale. However, with real data, I agree that keeping original coordinates is less confusing. I'm happy to go with `False` as the default, but I'd definitely want to keep the possibility of chopping out the middle of a TS and saving it as an entirely separate object with new coordinates.

hyanwong: (I should say that this was the original motivation for coding up `slice`: I wanted to pick out the middle of a genomic segment I was using as a test case for `tsinfer`, and just focus on inferring that region. The coordinates in this case were irrelevant, and it would have been very confusing to plot test cases with empty regions to left and right of the focal region.)

petrelharp:
This *is* very nice. Here's bits of a numpy version:
```
def keep_with_offset(keep, data, offset):
    lens = np.diff(offset)
    return (data[np.repeat(keep, lens)], 
                np.concatenate([np.array([0],dtype=offset.dtype), 
                                           np.cumsum(lens[keep], dtype=offset.dtype)]))

edges = self.tables.edges
sites = self.tables.sites
mutations = self.tables.mutations
keep_edges = np.logical_or(edges.right >= start, edges.left < stop)
keep_sites = np.logical_and(sites.position >= start, sites.position < stop)
keep_mutations = keep_sites[mutations.site]
tables = self.dump_tables()
tables.edges.set_columns(
    left=np.fmax(start, edges.left[keep_edges]),
    right=np.fmin(stop,edges.left[keep_edges]),
    parent=edges.parent[keep_edges],
    child=edges.child[keep_edges])
new_as, new_as_offset = keep_with_offset(
                                             keep_sites, sites.ancestral_state, sites.ancestral_state_offset)
new_md, new_md_offset = keep_with_offset(
                                             keep_sites, sites.metadata, sites.metadata_offset)
tables.sites.set_columns(
    position=sites.position[keep_sites],
    ancestral_state=new_as,
    ancestral_state_offset=new_as_offset,
    metadata=new_md,
    metadata_offset=new_md_offset)
new_ds, new_ds_offset = keep_with_offset(
                                             keep_mutations, mutations.derived_state,
                                             mutations.derived_state_offset)
new_md, new_md_offset = keep_with_offset(
                                             keep_mutations, mutations.metadata, mutations.metadata_offset)
tables.mutations.set_columns(
    site=mutations.site[keep_mutations],
    node=mutations.node[keep_mutations],
    derived_state=new_ds,
    derived_state_offset=new_ds_offset,
    parent=mutations.parent[keep_mutations],
    metadata=new_md,
    metadata_offset=new_md_offset)
```

Hm, that ended up being a bit more involved (dtypes!) than I thought, but there you go.

python/tskit/trees.py
petrelharp: True. Can you suggest a better phrasing? "open" and "closed" are not familiar to too many people

python/tskit/trees.py
petrelharp: Oh, wait! Let me retract my statement - I read too quickly and thought `rescale` meant to make the new tree sequence have the same length as the old one. Apologies. Instead: I agree with this, but can we call it something else? Like `shift` or `shift_to_zero`?

hyanwong: I see. Yes, "rescale" might imply changing the meaning of a unit as well as shifting along, so not such a good choice of wording. How about `shift_coordinates=True/False` instead?

hyanwong: Or perhaps simply `reset_coordinates`

python/tskit/trees.py
hyanwong: "Inclusive" works OK, but the antonym "exclusive" is a little confusing, I think.

python/tskit/trees.py
hyanwong: Now renamed to to `reset_coordinates`, as that correctly implies shifting so that the leftmost tree is at 0.

python/tskit/trees.py
hyanwong: I've phrased around it instead: 
> "start: The leftmost genomic position, giving the start point
>  of the kept region. Tree sequence information along the genome prior to (but
>  not including) this point will be discarded."

and

> stop: The rightmost genomic position, giving the end point of the
> kept region. Tree sequence information at this point and further along the
> genomic sequence will be discarded.

python/tskit/trees.py
petrelharp: sounds good. thanks!

python/tskit/trees.py
petrelharp: oh, perfect.

petrelharp:
I was hoping you could slot in that numpy code and check it works, @hyanwong? Ping me if you'd like me to do something.

hyanwong:
> I was hoping you could slot in that numpy code and check it works, @hyanwong? Ping me if you'd like me to do something.

Oh, sorry! I didn't realise you were waiting on me. I'm not sure what the normal procedure is for having both a numpy and non-numpy version?

petrelharp:
Ah, good - glad I checked! We should probably use the numpy version in the production code (should be a lot faster) and your very nice, clear version in the testing code: so, you could move your code to a new method in the testing code, like
```
def slice(ts, start, stop, ...):
    ...
```
and then you can make a test that does something like
```
  for ts in get_tree_sequences():
     # choose start, stop randomly
     x = slice(ts, start, stop, ...)
     y = ts.slice(start, stop, ...)
     self.assertEqual(x, y)
```
This sort of computing the same thing in two different ways and checking they are equal is a very good way to check for correcness.

For instance, see how [here](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_highlevel.py#L1339) we compare the C version of simplify to a locally defined python version, applied to a bunch of tree sequences [here](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_highlevel.py#L1373).

hyanwong:
> We should probably use the numpy version in the production code (should be a lot faster) and your very nice, clear version in the testing code

Ah, I see. Right. Will get round to this when I have a few spare moments.

hyanwong:
@petrelharp I'm not sure that I can do `self.assertEqual(ts1, ts2)` can I? Is there an equality operator for tree sequences - it feels like there should be, but I'm getting things like:

    AssertionError: <tskit.trees.TreeSequence object at 0x1108e6eb8> != <tskit.trees.TreeSequence object at 0x110904080>

for tree sequences constructed with exactly the same params (e.g. simulated then sliced independently).

While I'm at it, I also can't do `ts2 = ts.copy()`, which seems a small oversight?

petrelharp:
It's true, that doesn't work; I was writing pseudocode (sorry). Instead,
```
self.assertEqual(x.tables.nodes, y.tables.nodes)
self.assertEqual(x.tables.nodes, y.tables.nodes)
self.assertEqual(x.tables.nodes, y.tables.nodes)
```

hyanwong:
Thanks @petrelharp - I guess my point was, should I code up an equality operator for tree sequences? If so, we might want something like `TreeSequence.equals(ts, include_provenance=True/False)` so that we can compare the sequence data itself, or the entire TS.

jeromekelleher:
> Thanks @petrelharp - I guess my point was, should I code up an equality operator for tree sequences? If so, we might want something like TreeSequence.equals(ts, include_provenance=True/False) so that we can compare the sequence data itself, or the entire TS.

There's an ``equals`` operator defined on TableCollections, so 
```
self.assertEquals(ts1.tables, ts2.tables)
```
will work as expected. Provenance makes this slightly tricky as the timestamps can differ even if the actual records are identical, so usually what I do is:
```
t1 = ts1.dump_tables()
t2 = ts2.dump_tables()
t1.provenances.clear()
t2.provenances.clear()
self.assertEqual(t1, t2)
```
It's a mild faff, but not a big deal.

hyanwong:
@jeromekelleher: any objections to me coding this up into a ts.tables.equals() function? It would seem useful.

jeromekelleher:
I dunno, I it seems a redundant given that the ``__eq__`` method is implemented and the only time it'll ever be used is for testing. I guess we could add a function to do it in ``tsutil`` (tables_equal, say) , which would be handy for testing?

petrelharp:
Either way; I'd probably just put it in tsutil or not at all myself.

hyanwong:
It seemed a bit strange to me not to be able to compare plain tree sequences to see if they described the same population (i.e. ignoring provenance). So I thought I should be able to do something that avoided having to copy then dump provenance, e.g (pseudocode)

    function equals(ts1, ts2, ignore_provenance) {
        return all([ts1.tables[n] == ts2.tables[n] for n in tablecollection.tablenames if n != 'provenance'])
    }

but I see that the `__eq__` function for table collections uses the low-level operators, so it's not so easy.

Also, I'm having second thoughts because ISTR that some of the orders for table rows are essentially arbitrary, aren't they? So (depending on how the table comparison operator works), it might be that the same ancestry (or site info) might not have the same row orders in the tables, and the `__eq__` function might be too restrictive.

hyanwong:
> ```
> t1 = ts1.dump_tables()
> t2 = ts2.dump_tables()
> t1.provenances.clear()
> t2.provenances.clear()
> self.assertEqual(t1, t2)
> ```

Just a minor observation: the suggestion above from @jeromekelleher doesn't test if the provenances contain the same data but with different timestamps. If we are comparing supposedly identical methods, which include creating provenance info, I guess we should be testing this too.

petrelharp:
Good points, @hyanwong. This seems like arguments in favor of not trying to do general equality testing here, but rather just testing whatever equality we need here for the slice function (our two version of the funciton should return all tables in the same order and everything).

hyanwong:
@petrelharp yes, I agree. But I wonder if it would be useful to be able to iterate over the tables in a TableCollection. That seems reasonable, given the idea of it being basically a collection of tables. I.e. is it reasonable to expect to be able to do

    for table in ts.tables:
        # do something with each table, depending on the table class
        # For comparison, we would want to ensure that tables were iterated over in deterministic order

One easy code for such an iterator might be to use logic along the lines of the following

    [n for n in sorted(dir(ts.tables)) if isinstance(n, tskit.BaseTable)]

jeromekelleher:
I think we're getting a bit too far from the original issue here - we don't really need this for this pr do we? Maybe open an issue to follow up?

hyanwong:
Yep - was wondering whether to open another issue on this - just wanted to check if it seemed sensible.

hyanwong:
@petrelharp re your numpy code, there's a missing bit of logic, as the mutations need to be inserted with the *new* site ids, not the old ones. I've hacked round this like:

        new_site_map = -np.ones(len(keep_sites), dtype=mutations.site.dtype)
        new_site_map[keep_sites] = np.arange(np.sum(keep_sites))
        ...
        tables.mutations.set_columns(
            site=new_site_map[mutations.site[keep_mutations]],
            ....

But that seems a bit of a hack. Can you think of a better way?

petrelharp:
Yes, I've figured out how to do this before! This is a job for `np.cumsum( )` - here is a version with the optional shift in also:
```
def keep_with_offset(keep, data, offset):
    lens = np.diff(offset)
    return (data[np.repeat(keep, lens)], 
                np.concatenate([np.array([0],dtype=offset.dtype), 
                                           np.cumsum(lens[keep], dtype=offset.dtype)]))

edges = self.tables.edges
sites = self.tables.sites
mutations = self.tables.mutations
keep_edges = np.logical_or(edges.right >= start, edges.left < stop)
keep_sites = np.logical_and(sites.position >= start, sites.position < stop)
site_map = np.cumsum(keep_sites, dtype='int32')
keep_mutations = keep_sites[mutations.site]
shift = start if reset_coordinates else 0.0
tables = self.dump_tables()
tables.edges.set_columns(
    left=np.fmax(start, edges.left[keep_edges]) - shift,
    right=np.fmin(stop, edges.left[keep_edges]) - shift,
    parent=edges.parent[keep_edges],
    child=edges.child[keep_edges])
new_as, new_as_offset = keep_with_offset(
     keep_sites, sites.ancestral_state, sites.ancestral_state_offset)
new_md, new_md_offset = keep_with_offset(
     keep_sites, sites.metadata, sites.metadata_offset)
tables.sites.set_columns(
    position=sites.position[keep_sites] - shift,
    ancestral_state=new_as,
    ancestral_state_offset=new_as_offset,
    metadata=new_md,
    metadata_offset=new_md_offset)
new_ds, new_ds_offset = keep_with_offset(
     keep_mutations, mutations.derived_state,
     mutations.derived_state_offset)
new_md, new_md_offset = keep_with_offset(
     keep_mutations, mutations.metadata, mutations.metadata_offset)
tables.mutations.set_columns(
    site=site_map[mutations.site[keep_mutations]],
    node=mutations.node[keep_mutations],
    derived_state=new_ds,
    derived_state_offset=new_ds_offset,
    parent=mutations.parent[keep_mutations],
    metadata=new_md,
    metadata_offset=new_md_offset)
```

hyanwong:
Ah - of course, thanks!

 (NB - you have a typo in the code (corrected in my PR) `right=np.fmin(stop, edges.left[keep_edges]) - shift,` should be `right=np.fmin(stop, edges.right[keep_edges]) - shift,`)

hyanwong:
Actually, I think `np.cumsum` produces an out-by-one error, doesn't it? At least, it seems to on my testing. E.g. the first `True` value in the array is given index 1, not index 0? Or am I misunderstanding?

hyanwong:
I've updated this with `np.cumsum() -1`, which passes the tests. I didn't use your `shift = start if reset_coordinates else 0.0` as I wanted to avoid adding 0.0 unnecessarily, in case it created strange floating point differences (although I guess it is usually optimised out).

I think the PR is ready for review now.

petrelharp:
Whoops yep; good catch. This is why I put the code in a comment, not in a PR. =) Thanks for fixing it.

python/tests/test_topology.py
```diff
@@ -27,17 +27,76 @@
...
+        tables.sequence_length = stop - start
+    if simplify:
+        tables.simplify()
+    if record_provenance:
+        # TODO add slice arguments here
+        # TODO also make sure we convert all the arguments so that they are
+        # definitely JSON encodable.
+        parameters = {
+            "command": "slice",
+            "TODO": "add slice parameters"
```
petrelharp: perhaps remove these TODOs?

python/tskit/trees.py
```diff
@@ -3070,6 +3070,117 @@ def simplify(
...
             return new_ts
 
+    def slice(
+            self, start=None, stop=None, reset_coordinates=True, simplify=True,
+            record_provenance=True):
+        """
+        Truncate this tree sequence, returning a tree sequence that spans a reduced
+        genomic interval. As with all tree sequence coordinates, the interval is treated
+        as half closed, that is, the leftmost point is included and rightmost point
+        excluded. Sites not covered by the sliced region are thrown away.
```
petrelharp: This could be a bit more descriptive: it doesn't just return "a" tree sequence; it's a particular one. How about
```
        Returns the tree sequence containing only information in this tree sequence about the
        genomic interval between ``start`` and ``stop``. As with all tree sequence coordinates,
        the interval is treated
        as half closed, that is, the leftmost point is included and rightmost point
        excluded. Sites not covered by the sliced region are thrown away.
```

python/tskit/trees.py
```diff
@@ -3070,6 +3070,117 @@ def simplify(
...
+
+        :param float start: The leftmost genomic position, giving the start point
+            of the kept region. Tree sequence information along the genome prior to (but
+            not including) this point will be discarded. If None, use the start position
+            of the current tree sequence.
+        :param float stop: The rightmost genomic position, giving the end point of the
+            kept region. Tree sequence information at this point and further along the
+            genomic sequence will be discarded. If None, use the end position of the
+            current tree sequence.
```
petrelharp: How about `If None, is set equal to the current tree sequence's ``sequence_length``.`

python/tskit/trees.py
```diff
@@ -3070,6 +3070,117 @@ def simplify(
...
+        """
+        Truncate this tree sequence, returning a tree sequence that spans a reduced
+        genomic interval. As with all tree sequence coordinates, the interval is treated
+        as half closed, that is, the leftmost point is included and rightmost point
+        excluded. Sites not covered by the sliced region are thrown away.
+
+        :param float start: The leftmost genomic position, giving the start point
+            of the kept region. Tree sequence information along the genome prior to (but
+            not including) this point will be discarded. If None, use the start position
+            of the current tree sequence.
```
petrelharp: This is always equal to 0, no? Shouldn't it be "If None, is set equal to 0."?

python/tskit/trees.py
```diff
@@ -3070,6 +3070,117 @@ def simplify(
...
+        :param float stop: The rightmost genomic position, giving the end point of the
+            kept region. Tree sequence information at this point and further along the
+            genomic sequence will be discarded. If None, use the end position of the
+            current tree sequence.
+        :param bool reset_coordinates: Reset the genomic coordinates such that position
+            0 in the returned tree sequence corresponds to position ``start`` in the
+            original one, and the returned tree sequence is of length ``stop``-``start``.
```
petrelharp: Shouldn't this say "has span stop - start" or "has sequence length stop - start"?

petrelharp:
Subtracting 0.0 does not cause floating point error, IIUC. I'd vote to keep my version, which has less duplicated code, but it's not a big deal. Otherwise, looks good (if the tests pass).

python/tests/test_topology.py
hyanwong: I've made a start at encoding parameters at https://github.com/tskit-dev/tskit/blob/eca54e39108423038a1b8f6112dab3b20a94e280/python/tskit/trees.py#L3108 and  https://github.com/tskit-dev/tskit/blob/eca54e39108423038a1b8f6112dab3b20a94e280/python/tskit/trees.py#L3179

But it seems like this general way of extracting and encoding params to a function using the `inspect` module should be abstracted so it can be re-used.

Also, if we are to implement this for `simplify()`, where there is a similar TODO left by @jeromekelleher , we need a way to serialize np.arrays in JSON, as we can pass a numpy array as a parameter to simplify. I haven't attempted to tackle this here, as `slice` doesn't require numpy arrays.

I'm not sure if I should open this up as another PR, and if so, whether *this* PR should include it or not.

python/tskit/trees.py
hyanwong: Actually, I find it a bit confusing to refer to two different TSs, and it's probably unnecessary. We could simply say:

        Truncate this tree sequence to include only information in the genomic interval
        between ``start`` and ``stop``.  Edges are truncated to this interval, and sites
        not covered by the sliced region are thrown away.

It's obvious from the return values that we return a TS rather than truncating in-place (which is not possible anyway, given that tree sequences are immutable)

python/tskit/trees.py
hyanwong: Fine.

python/tskit/trees.py
hyanwong: Good point!

python/tskit/trees.py

python/tests/test_topology.py
hyanwong: I've worked this up into something more general in https://github.com/tskit-dev/tskit/pull/243 - although not ready for review (e.g. no tests) there should be enough there to see the general approach.

petrelharp: It seems like `slice` has only a few relatively simple arguments. Why not just make a simple JSON dictionary of them for the provenance and postpone a more general solution? It is hard for me to see the right way to do it in general without working through some specific examples - as you've pointed out, there's some possibly serious issues lurking.

petrelharp: Hm, I see that's more or less what you've done.

hyanwong: Yes, I guess this can be marked as "resolved", on the assumption that https://github.com/tskit-dev/tskit/pull/243 (or something like it) will eventually get merged

petrelharp:
This is great; I'll let someone else have a look before merging.

python/tskit/trees.py
```diff
@@ -3070,6 +3071,124 @@ def simplify(
...
+        :param bool simplify: If True, simplify the resulting tree sequence so that nodes
+            no longer used in the resulting trees are discarded. (Default: True).
+        :param bool record_provenance: If True, record details of this call to
+            slice in the returned tree sequence's provenance information.
+            (Default: True).
+        :return: The sliced tree sequence.
+        :rtype: .TreeSequence
+        """
+        # First save the args used, to save in provenance
+        func_args, _, _, func_values = inspect.getargvalues(inspect.currentframe())
```
jeromekelleher: This provenance stuff seems like overkill here and a large fraction of the actual code in the method. Can we trim this out and replace it with a simple implementation which doesn't bother with the parameters? Looks like we have a general approach for this elsewhere, so let's do it all at once.

hyanwong: OK. I have deleted this and the corresponding tests, but I want to keep the test code around to reintroduce later, so I'm pasting it here, below:

```
    def test_slice_provenance(self):
        ts = msprime.simulate(5, random_seed=1, recombination_rate=2, mutation_rate=2)
        ts_sliced = ts.slice(0.4, 0.6)
        self.assertEqual(ts_sliced.num_provenances, 2)
        provenance = json.loads(ts_sliced.provenance(0).record)
        self.assertEquals(provenance['software']['name'], 'msprime')
        provenance = json.loads(ts_sliced.provenance(1).record)
        self.assertEquals(provenance['parameters']['command'], 'slice')
        self.assertEquals(provenance['parameters']['start'], 0.4)
        self.assertEquals(provenance['parameters']['stop'], 0.6)
        # Check that we don't record provenances if record_provenance=False
        ts_sliced = ts.slice(0.4, 0.6, record_provenance=False)
        self.assertEqual(ts_sliced.num_provenances, 1)
        provenance = json.loads(ts_sliced.provenance(0).record)
        self.assertEquals(provenance['software']['name'], 'msprime')
        # Check we raise an error if we pass something unserializable
        self.assertRaises(TypeError, ts.slice, 0.4, 0.6, record_provenance=lambda x: x)
```

jeromekelleher:
Great, this looks ready to merge to me. Can you do a quick rebase to bring it up to date please @hyanwong?

hyanwong:
Done

Issue #230, opened 2019-06-13
write segregating sites and Tajima's D functions


petrelharp:
The branch length version of segregating sites is total branch length, which we also want.

jeromekelleher:
What do we want to call this? Calling ``ts.segregating_sites(mode="branch")`` to get the total branch length feels bad. Maybe we should just have two methods, one for sites and one for branch_length?

petrelharp:
This was also bothering me. I think that just calling it "segregating sites" is the right way forward, though, because (a) we can provide a lookup  table #231 and (b) our convention so far is to name statistics after their "site" version. Which is sensible, I think. I've got some others' input, who agreed.

jeromekelleher:
Question: should the function be ``Tajimas_D`` or ``tajimas_D``? I think it should probably be the latter, as all other functions are lowercase, except for things like ``Fst``, ``Y2`` etc. These seem like reasonable exceptions since it's following standard notation. Capitalising Tajima seems less obviously like the right thing to do for me --- consistency is good for users.

Not a strong opinion though.

petrelharp:
>  except for things like Fst
I wondered about this also; but Tajima is definately capitalised; I'd think it would fall under the category of "things like Fst". Either is OK with me, but I'd vote to keep it capitalised.

jeromekelleher:
OK, let's keep it as Tajimas_D then.

jeromekelleher:
Closed in #237

PR #237, opened 2019-06-14
segregating sites and Tajima's D

petrelharp:
Here's an implementation of "segregating sites"; see #230. Probably needs some C-level tests but is all good otherwise.

Next: [Tajima's D](https://en.wikipedia.org/wiki/Tajima%27s_D).

### DIFF
--- a/python/tests/tsutil.py
+++ b/python/tests/tsutil.py
@@ -333,8 +333,8 @@ def generate_site_mutations(tree, position, mu, site_table, mutation_table,
     more than one mutation per edge.
     """
     assert tree.interval[0] <= position < tree.interval[1]
-    states = {"A", "C", "G", "T"}
-    state = random.choice(list(states))
+    states = ["A", "C", "G", "T"]
+    state = random.choice(states)
     site_table.add_row(position, state)
     site = site_table.num_rows - 1
     stack = [(tree.root, state, tskit.NULL)]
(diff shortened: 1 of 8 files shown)


petrelharp:
Screenshot from 2019-06-14 09-55-21

petrelharp:
Darn it, how do I make my local builds as strict as AppVeyor's? I don't see any of these warnings when I compile.

petrelharp:
Hm, issues seem to be in the Windows build. Anyone know what's going on with
```
ERROR: test_single_tree_jukes_cantor (tests.test_tree_stats.TestSiteDivergence)
  File "C:\projects\tskit\python\tests\test_tree_stats.py", line 655, in test_single_tree_jukes_cantor
    print(t.draw(format='unicode'))
  File "C:\Miniconda36-x64\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 17-23: character maps to <undefined>
```

python/tests/tsutil.py
```diff
@@ -333,8 +333,8 @@ def generate_site_mutations(tree, position, mu, site_table, mutation_table,
     more than one mutation per edge.
     """
     assert tree.interval[0] <= position < tree.interval[1]
-    states = {"A", "C", "G", "T"}
-    state = random.choice(list(states))
+    states = ["A", "C", "G", "T"]
+    state = random.choice(states)
```
petrelharp: The tests were not deterministic because of different orderings of sets, btw. This fixes that.

jeromekelleher: Dammit, got caught by that one before!

jeromekelleher:
This is great! Can you add a low-level test also please like [this one](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_lowlevel.py#L519)? This makes sure the Python C interface is properly exercised.

Happy to merge whenever.

petrelharp:
I've added Tajima's D and rebased/squashed. Should be good to go; merge if you like (assuming it still passes tests...)

jeromekelleher:
Looks great, merging!

Issue #238, opened 2019-06-16
additional restrictions on summary_func?


petrelharp:
It should probably be a guaranteed property of general statistics that if you compute a branch statistic before and after simplifying then you'll get the same answer. Unsimplified tree sequences can have dangling branches, which should have weight 0, so unpolarised statistics will assign value
```
  f(0) + f(total_weight)
```
to these branches. So, to ensure that this property holds, we need to require that
```
 f(0) = 0
 f(total_weight) = 0
```
This is the case for all statistics we've defined or discussed so far (e.g., for phenotypic covariance, we deifned the weights to sum to zero).

jeromekelleher:
Should we state this as a required property in the docs, or verify it in the code? I guess we should verify it for the Python interface for general stats anyway, as a useful debugging dev tool? It's not so obvious we should do it in C.

petrelharp:
Whoops - in fact, for unpolarised stats we only need that `f(0) + f(total_weight) = 0`. For polarised statistics we would want each term to be individually zero.

petrelharp:
I think we should document that both `f(0)` and `f(total_weight)` should both be zero, and verify it in python in general_stat. It's a very useful debugging tool, and putting it only in general stats will warn us, since we cross-check methods against general_stat anyhow.

jeromekelleher:
I agree - the general stats interface isn't high-performance anyway, so checking the function here makes total sense. (We should keep in mind people that are not us also, who are developing stats that **don't** end up being implemented in tskit)

jeromekelleher:
I think this still needs doing @petrelharp?

PR #239, opened 2019-06-16
phenotype correlation

petrelharp:
In the paper we've described a statistic that is "phenotype correlation", which is basically the correlation of a weight with the inheritance pattern of the site/branch/node. So far I've implemented covariance; doing correlation (dividing by `sqrt(p(1-p))`) won't be hard.  Just 'cause it sounds catchy I've called it "inheritance covariance", but should probably rename it "phenotype covariance" or even just "weight covariance".

In doing this I've had to make another route from python to c that parallels the "sample set" machinery. It's probably not quite the right API yet to be useful in a bunch of different situations.

To-do:
- implement correlation also
- write c tests
- write lowlevel python tests

Also see #238 which I realized while doing this.

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -477,6 +477,32 @@ verify_node_general_stat_errors(tsk_treeseq_t *ts)
 }
 
 
+typedef int one_way_weighted_method(tsk_treeseq_t *self,
+        tsk_size_t num_weights, double *weights,
+        tsk_size_t num_windows, double *windows, double *result, tsk_flags_t options);
+
+static void
+verify_one_way_weighted_func_errors(tsk_treeseq_t *ts, one_way_weighted_method *method)
+{
+    // we don't have any specific errors for this function
+    // but we might add some in the future
+    int ret;
+    tsk_size_t num_samples = tsk_treeseq_get_num_samples(ts);
+    double *weights = malloc(num_samples * sizeof(double));
+    double result;
+    size_t j;
+
+    for (j = 0; j < num_samples; j++) {
+        weights[j] = 1.0;
+    }
+
+    ret = method(ts, 0, weights, 0, NULL, &result, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_STATE_DIMS);
+
+    free(weights);
+}
+
+
 typedef int one_way_stat_method(tsk_treeseq_t *self,
         tsk_size_t num_sample_sets, tsk_size_t *sample_set_sizes, tsk_id_t *sample_sets,
         tsk_size_t num_windows, double *windows, double *result, tsk_flags_t options);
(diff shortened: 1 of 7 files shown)


jeromekelleher:
This is amazing @petrelharp, the summary func is so simple! Code looks great, the Python-C stuff is air-tight.

petrelharp:
Also to-do: center `W` for covariance. (maybe normalise too?)

c/tests/test_stats.c
```diff
@@ -477,6 +477,30 @@ verify_node_general_stat_errors(tsk_treeseq_t *ts)
...
+    double *weights = malloc(num_samples * sizeof(double));
+    double result;
+    size_t j;
+
+    for (j = 0; j < num_samples; j++) {
+        weights[j] = 1.0;
+    }
+
+    ret = method(ts, 0, weights, 0, NULL, &result, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_STATE_DIMS);
```
jeromekelleher: Leaking memory here - need to free weights.

c/tests/test_stats.c
```diff
@@ -1044,6 +1068,52 @@ test_paper_ex_diversity(void)
...
+    /* weights of 0 leads to 0 */
+    for (j = 0; j < num_samples; j++) {
+        weights[j] = 0.0;
+    }
+    ret = tsk_treeseq_inheritance_covariance(&ts, 1, weights, 0, NULL,
+            &result, TSK_STAT_SITE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_DOUBLE_EQUAL_FATAL(result, 0.0, 1e-6);
+
+    tsk_treeseq_free(&ts);
```
jeromekelleher: Also leaking here. It's a good idea to run these tests through valgrind, catches these problems up front.

c/tests/test_stats.c
petrelharp: heh. oh, right.

petrelharp:
I've implemented correlations, too. As expected, I (a) found some how-did-this-even-work-before errors, (b) found an whoops-I-was-changing stuff-back-in-python-without-meaning-to error, and (c) had to think harder about how to set things up. The main thing for (c) is that in computing correlations, we divide by `sqrt(p * (1 - p))`, which means we have to compute `p`, which means we need an additional column on the array of weights (that is all `1/n`s). This is no big deal - at the C level we have to copy the weight array anyhow (to standardise it), and required some adjustments to the testing set-up - but I suspect in the future we'll want to make this sort of thing generally easier somehow.

Hopefully, this is ready to go. I want to add at least one test where we check by hand what the answer is supposed to be, first, though, to avoid e.g. factors of 2 in common between the implementations.

Also to-do: rename functions. And, check the docs to make sure it says it's the *squared* correlation.

Renaming: I'm tending towards `trait_covariance` and `trait correlation`: thoughts?

petrelharp:
There's a significant omission here still - I forgot the `n-1` in the denominator. In the process of fixing this.

c/tskit/trees.c
```diff
@@ -1858,6 +1859,139 @@ tsk_treeseq_diversity(tsk_treeseq_t *self,
...
+
+    // center weights
+    for (j = 0; j < num_samples; j++) {
+        row = GET_2D_ROW(weights, num_weights, j);
+        for (k = 0; k < num_weights; k++) {
+            means[k] += row[k];
+        }
+    }
+    for (k = 0; k < num_weights; k++) {
+        means[k] /= (float) num_samples;
```
jeromekelleher: Cast to double rather than float - could have weird consequences otherwise.

c/tskit/trees.c
```diff
@@ -1858,6 +1859,139 @@ tsk_treeseq_diversity(tsk_treeseq_t *self,
...
+static int
+inheritance_correlation_summary_func(size_t state_dim, double *state, size_t TSK_UNUSED(result_dim),
+        double *result, void *TSK_UNUSED(params))
+{
+    const double *x = state;
+    double p;
+    size_t j;
+
+    p = x[state_dim - 1];
+    for (j = 0; j < state_dim - 1; j++) {
```
jeromekelleher: Is there any way p could be negative? Do we need to guard against this? Numpy returns a NaN if -1 is passed to sqrt, so many it's OK to just let it happen like we're doing with division by zero.

jeromekelleher: Stupid Jerome! I just read the first bit of the line again and realised why it was doing it. Maybe clearer to have an if statement for this?

c/tskit/trees.c
```diff
@@ -1858,6 +1859,139 @@ tsk_treeseq_diversity(tsk_treeseq_t *self,
...
+    // center and scale weights
+    for (j = 0; j < num_samples; j++) {
+        row = GET_2D_ROW(weights, num_weights, j);
+        for (k = 0; k < num_weights; k++) {
+            means[k] += row[k];
+            meansqs[k] += row[k] * row[k];
+        }
+    }
+    for (k = 0; k < num_weights; k++) {
+        means[k] /= (float) num_samples;
```
jeromekelleher: Cast to double here also.

c/tskit/trees.c
```diff
@@ -1858,6 +1859,139 @@ tsk_treeseq_diversity(tsk_treeseq_t *self,
...
+        sds[k] = sqrt(meansqs[k]);
+    }
+    for (j = 0; j < num_samples; j++) {
+        row = GET_2D_ROW(weights, num_weights, j);
+        new_row = GET_2D_ROW(new_weights, num_weights + 1, j);
+        for (k = 0; k < num_weights; k++) {
+            new_row[k] = (row[k] - means[k]) / sds[k];
+        }
+        // set final row to 1/n to compute frequency
+        new_row[num_weights] = 1.0 / (float) num_samples;
```
jeromekelleher: ditto here.

jeromekelleher:
> Renaming: I'm tending towards trait_covariance and trait correlation: thoughts?

Like. Trait is much clearer than inheritance.

petrelharp:
Fixed things up with comparing to actual covariances and correlations. Still got to

- [] change the name
- [] make the tests not take so long

petrelharp:
Maybe this is ready to go? See my solution to speeding up the stats tests (which were taking an hour): not  do all of them. =) (but, it's a good solution)

petrelharp:
BTW, I checked, and there was no bottleneck in the methods; it's just we were doing a *lot* of tests - like, 4 choices of window times five choices of sample set times three modes times twenty situations times lots of stats ... the "definition" methods for correlation and covariance seemed the slowest; and they'd take less than a second per tree sequence, but it added up.

jeromekelleher:
Looks great, thanks @petrelharp! Just needs a rebase and we're good to go I think (but, watch out for your other PR, which I think is also a commit of this branch? You'll need to rebase that as well, dropping all the leading commits.)

Re the random subsetting, that's a great solution. We should definitely make it deterministic though: there's nothing more frustrating than having a test that fails when you run all the tests, but passes when you run it by itself. All unit tests should run identically, regardless of what other tests are run.

petrelharp:
Ok, maybe this is good to go?

jeromekelleher:
Yep, looking good. Thanks @petrelharp!

Issue #189, opened 2019-05-27
Misaligned unicode from tree.draw() in jupyter notebook


brianzhang01:
Repro: https://github.com/brianzhang01/msprime_unicode/blob/master/msprime_unicode.ipynb

I have also observed this issue on Google Colab. The ASCII output is aligned fine (shown in notebook), and the SVG works too (not shown).

brianzhang01:
So apparently Unicode character width can either be 1 or 2 Latin characters of width. It seems like Jupyter notebook sets all Unicode characters to be rendered as width 2 by default. There may or may not be a way to alter this.
https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html#unicode-formatting

brianzhang01:
One approach would be to add an option into the Unicode drawer that converts regular ASCII characters into full-width versions: https://stackoverflow.com/questions/4622357/how-to-control-padding-of-unicode-string-containing-east-asia-characters

hyanwong:
I have had problems in the past with unicode font substitution. But I agree that having `format="unicode"` working in a Jupyter notebook would be nice, and a reason to keep the ascii format until this is working (although it's pretty easy to display SVG in a juptyer notebook, so I feel it isn't that strong an argument)

jeromekelleher:
I think this is a browser/font issue - when I view notebooks you've linked to they all look fine. My guess is that on Macs the browser doesn't use a proper fixed-width font when it sees unicode chars, but does for some reason with ascii. I'm guessing that you're using a Mac @brianzhang01?

jeromekelleher:
Here's what the trees look like for me on Vivaldi Linux:
![Screenshot from 2019-05-28 09-17-25](https://user-images.githubusercontent.com/2664569/58462328-8a925980-8129-11e9-881a-47f23d046fbb.png)

jeromekelleher:
Firefox linux:
![firefox](https://user-images.githubusercontent.com/2664569/58462411-c1686f80-8129-11e9-8e4c-30e20d457c48.png)

jeromekelleher:
Would you mind trying this on a few browsers on Macs @brianzhang01, so we can get a handle on how much of a problem this is? I know that these trees don't render properly on Android either.

If it is a browser font issue, I don't think there's much we can do about it, and all we can do is recommend that people use SVG trees in browser environments.

hyanwong:
[...]

PR #242, opened 2019-06-18
Refactor text drawing

jeromekelleher:


jeromekelleher:
Pinging @petrelharp, @hyanwong, @brianzhang01, @awohns, @gtsambos as tree-drawers.

There's some significant updates to the text drawing API here, and it would be good to get some feedback. Firstly, we have nice rendering of tree sequences: hooray!

```
4.94┊    18           ┊                 ┊                 ┊                 ┊                 ┊
    ┊  ┏━━┻━━┓        ┊                 ┊                 ┊                 ┊                 ┊
3.30┊  ┃     ┃        ┊    17           ┊    17           ┊      17         ┊        17       ┊
    ┊  ┃     ┃        ┊  ┏━━┻━━┓        ┊  ┏━━┻━━┓        ┊   ┏━━━┻━━━┓     ┊     ┏━━━┻━━━┓   ┊
2.03┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃    16        ┊  16       ┃     ┊     ┃       ┃   ┊
    ┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃  ┏━━┻━━┓     ┊ ┏━┻┓      ┃     ┊     ┃       ┃   ┊
1.03┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃  ┃     ┃     ┊ ┃  ┃     15     ┊    15       ┃   ┊
    ┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃  ┃     ┃     ┊ ┃  ┃   ┏━━┻━┓   ┊  ┏━━┻━┓     ┃   ┊
0.86┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃  ┃     ┃     ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃    14   ┊
    ┊  ┃     ┃        ┊  ┃     ┃        ┊  ┃  ┃     ┃     ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃   ┏━┻┓  ┊
0.40┊  ┃    13        ┊  ┃    13        ┊  ┃  ┃     ┃     ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃   ┃  ┃  ┊
    ┊  ┃  ┏━━┻━━┓     ┊  ┃  ┏━━┻━━┓     ┊  ┃  ┃     ┃     ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃   ┃  ┃  ┊
0.12┊  ┃  ┃    12     ┊  ┃  ┃    12     ┊  ┃  ┃    12     ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃   ┃  ┃  ┊
    ┊  ┃  ┃   ┏━┻━━┓  ┊  ┃  ┃   ┏━┻━━┓  ┊  ┃  ┃   ┏━┻━━┓  ┊ ┃  ┃   ┃    ┃   ┊  ┃    ┃   ┃  ┃  ┊
0.10┊  ┃  ┃   ┃   11  ┊  ┃  ┃   ┃   11  ┊  ┃  ┃   ┃   11  ┊ ┃ 11   ┃    ┃   ┊  ┃    ┃   ┃ 11  ┊
    ┊  ┃  ┃   ┃   ┏┻┓ ┊  ┃  ┃   ┃   ┏┻┓ ┊  ┃  ┃   ┃   ┏┻┓ ┊ ┃ ┏┻┓  ┃    ┃   ┊  ┃    ┃   ┃ ┏┻┓ ┊
0.09┊  ┃  ┃  10   ┃ ┃ ┊  ┃  ┃  10   ┃ ┃ ┊  ┃  ┃  10   ┃ ┃ ┊ ┃ ┃ ┃  ┃   10   ┊  ┃   10   ┃ ┃ ┃ ┊
    ┊  ┃  ┃ ┏━┻┓  ┃ ┃ ┊  ┃  ┃ ┏━┻┓  ┃ ┃ ┊  ┃  ┃ ┏━┻┓  ┃ ┃ ┊ ┃ ┃ ┃  ┃  ┏━┻┓  ┊  ┃  ┏━┻┓  ┃ ┃ ┃ ┊
0.02┊  9  ┃ ┃  ┃  ┃ ┃ ┊  9  ┃ ┃  ┃  ┃ ┃ ┊  9  ┃ ┃  ┃  ┃ ┃ ┊ ┃ ┃ ┃  9  ┃  ┃  ┊  9  ┃  ┃  ┃ ┃ ┃ ┊
    ┊ ┏┻┓ ┃ ┃  ┃  ┃ ┃ ┊ ┏┻┓ ┃ ┃  ┃  ┃ ┃ ┊ ┏┻┓ ┃ ┃  ┃  ┃ ┃ ┊ ┃ ┃ ┃ ┏┻┓ ┃  ┃  ┊ ┏┻┓ ┃  ┃  ┃ ┃ ┃ ┊
0.01┊ ┃ ┃ ┃ ┃  8  ┃ ┃ ┊ ┃ ┃ ┃ ┃  8  ┃ ┃ ┊ ┃ ┃ ┃ ┃  8  ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃  8  ┊ ┃ ┃ ┃  8  ┃ ┃ ┃ ┊
    ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┊ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊
0.00┊ 4 5 3 0 1 2 6 7 ┊ 4 5 3 0 1 2 6 7 ┊ 4 5 3 0 1 2 6 7 ┊ 3 6 7 4 5 0 1 2 ┊ 4 5 0 1 2 3 6 7 ┊
  0.00              0.14              0.22              0.76              0.84              1.00
```

The drawing is done by the ``drawing.TextTreeSequence`` class, which delegates most of the work to the TextTree class. The format of the time and span axis labels can be controlled using the ``time_label_format`` and ``position_label_format`` respectively. I think I've taken care of all the corner cases, and this feature is pretty solid now.

There's also a preliminary pass at drawing left-to-right trees using the ``orientation='left'`` option:
```
 ┏━━━━━━━━━━2
 ┃
8┫        ┏━0
 ┃  ┏━━━━5┫
 ┃  ┃     ┗━1
 ┗━7┫
    ┃  ┏━━━━3
    ┗━6┫
       ┗━━━━4
```
This basically works, but I still need to figure out how to create space for the labels along the branches. The reason I want to sort this out now is that I want to finalise the ``draw_text`` API, and to do this I need to understand the differences that will happen when we have horizontally rendered trees like this (for example, the time scales are much more complicated now because of the need to fit labels in the branches).

Questions:
1. Does this look good visually? Any ideas for tweaks?
2. Does the API seem sensible?
3. Can you think of any obvious extensions that **won't** fit into the API that's outlined here?

The plan would then be to mirror the structure of this API in the ``draw_svg`` code. SVG is a lot easier because you don't have a to worry about allocating space for things...

gtsambos:
> There's also a preliminary pass at drawing left-to-right trees using the orientation='left' option:

Yay, looks good Jerome! I was literally just thinking today that it would be useful to be able to draw rotated trees, as I've come up with some ancestry-related stepwise functions that are best conceptualised with time on the horizontal axis (I will show you at our next f2f meeting).

I think it would be more natural for these trees to be the other way around (ie. leaves on the left, root on the right). If so, then when you plot a sequence of left-aligned trees, the horizontal time axis will be the same for each tree. As it is, the same leaves on different trees will have different horizontal coordinates, which is a bit confusing given that they are all at time = 0. I also think this would be a bit more intuitive, since bigger (backwards-in)-time values would then be further to the right, just like on a number line.

jeromekelleher:
> As it is, the same leaves on different trees will have different horizontal coordinates, which is a bit confusing given that they are all at time = 0. I also think this would be a bit more intuitive, since bigger (backwards-in)-time values would then be further to the right, just like on a number line.

We'd definitely make sure that the nodes with the same time have the same x coordinates, which is true of the version above (but maybe your browser isn't rendering unicode correctly? This is a long-running issue, see #189).

Having alignment="right" as well as "left" would also be a logical option. Shouldn't be that hard to do.

jeromekelleher:
OK, this is looking like it's the right way to do things so I'm going to merge it. I'll open some issues tracking where we want to go with the tree drawing APIs once the SVG code has been cleaned up as well.

petrelharp:
this is wonderful.

brianzhang01:
So I'm on `tskit` 0.2.1 now and had my first spin with the new drawing functionality. However, I needed to call `ts.draw_text()` and `ts.first().draw_text()`, I couldn't just use `ts.first().draw()` to get the `orientaiton = "left"` option. Furthermore, neither of the `draw_text()` methods is currently documented [here](https://tskit.readthedocs.io/en/latest/python-api.html). Was this intended until we further clarify / finalize the API?

jeromekelleher:
The draw_text() and draw_svg() versions are experimental and aren't documented for the moment. I was planning to use them within the developer community for a bit before finalising the APIs and documenting fully. The long-term vision at the moment is that we'll keep the current ``Tree.draw()`` interface as a relatively simple interface and let draw_text and draw_svg be more tightly focused on text and SVG. In particular, we want to document the structure of the output SVG in draw_svg so that users can manipulate it for (e.g.) animations.

Probably we'll add the ``orientation`` argument to ``draw()`` at some point, but we'll need to implement it in ``draw_svg`` first.

Thanks for trying this stuff out --- please open issues if you have any feedback or spot any problems!

PR #243, opened 2019-06-19
Add function to JSONify command args

hyanwong:
This should allow us to encode the arguments passed e.g. to `simplify()` in the provenances data. This is currently [listed as a TO DO](https://github.com/tskit-dev/tskit/blob/57ff0abe4173611a396e9715632d544195880a75/python/tskit/trees.py#L3057).

I'm not sure if this is the right approach or not - it seems a bit magic to simply call (e.g. within `simplify()`)

    tables.provenances.add_row(record=provenance.provenance_command_JSON())

and have the whole thing constructed for you. Using `inspect.currentframe().f_back` within the new function, to get at the outer calling function, seems potentially fragile. Other possible ways to do this are welcome.

### DIFF
--- a/python/tskit/provenance.py
+++ b/python/tskit/provenance.py
@@ -153,7 +153,8 @@ def provenance_command_JSON(exclude_args=[]):
     """
     def serialize_numpy_array(arg):
         if type(arg) == np.ndarray:
-            return np.array_repr(arg)
+            # NB: we might also want to strip whitespace from the stringified array
+            return np.array_repr(arg, max_line_width=np.inf)
         return json.dumps(arg)
 
     exclude_args.append('self')
--- a/python/tskit/provenance.py
+++ b/python/tskit/provenance.py
@@ -25,10 +25,13 @@ Common provenance methods used to determine the state and versions
 of various dependencies and the OS.
 """
 import platform
+import inspect
+import collections
 import json
 import os.path
 
 import jsonschema
+import numpy as np
 
 import tskit.exceptions as exceptions
 import _tskit
@@ -132,3 +135,43 @@ def validate_provenance(provenance):
         jsonschema.validate(provenance, schema)
     except jsonschema.exceptions.ValidationError as ve:
         raise exceptions.ProvenanceValidationError from ve
+
+
+def provenance_command_JSON(exclude_args=[]):
+    """
+    This function can be called within a function and will encode the arguments used
+    to call that function into a JSON string, which is returned.
+
+    The exclude_args argument allows arguments to be excluded from the prevenance info
+    such as deprecated aliases for other arguments (the `self` argument of methods is
+    always excluded).
+
+    It can be used to e.g. serialize the arguments to functions like `simplify`
+    so they can be stored in a provenance field. It requires the arguments to be JSON
+    serializable (and uses np.array_repr to serialize numpy arrays, which are not
+    normally serializable by json).
+    """
+    def serialize_numpy_array(arg):
+        if type(arg) == np.ndarray:
+            return np.array_repr(arg)
+        return json.dumps(arg)
+
+    exclude_args.append('self')
+    caller = inspect.currentframe().f_back
+    command_name = inspect.getframeinfo(caller).function
+    command_args, _, _, command_vals = inspect.getargvalues(caller)
+    # Store params in an OrderedDict to retain the normal command order
+    parameters = collections.OrderedDict(command=command_name)
+    for a in command_args:
+        # Don't serialise the 'self' argument of class methods
+        if a not in exclude_args:
+            parameters[a] = command_vals[a]
+    provenance = get_provenance_dict(parameters)
+    try:
+        return json.dumps(provenance, default=serialize_numpy_array)
+    except TypeError as e:
+        err_str = "Cannot save provenance info when running `{}`:".format(command_name)
+        err_str += " some arguments are not convertable to JSON."
+        if 'record_provenance' in command_args:
+            err_str += " To avoid saving provenance info, set `record_provenance=False`"
+        raise TypeError(err_str) from e
--- a/python/tskit/provenance.py
+++ b/python/tskit/provenance.py
@@ -137,7 +137,17 @@ def validate_provenance(provenance):
         raise exceptions.ProvenanceValidationError from ve
 
 
-def provenance_command_JSON(exclude_args=[]):
+def dump_json_inc_numpy_arr(arg):
+    """
+    Numpy arrays and scalars aren't normally serializable - call tolist() on such objects
+    first. Note that we lose dtype information in this case, which may be unavoidable.
+    """
+    if isinstance(arg, (np.ndarray, np.generic)):
+        return json.dumps(arg.tolist())
+    return json.dumps(arg)
+
+
+def function_args_to_json(exclude_args=None):
     """
     This function can be called within a function and will encode the arguments used
     to call that function into a JSON string, which is returned.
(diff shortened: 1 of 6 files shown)


hyanwong:
One interesting thing to note with this approach is that any changes within the function to the parameters will be reflected in the provenance. So for instance,

    ts.slice(start=None, stop=4)

will not record `start: None`, but instead `start: 0`, since the code within the function sets start to zero if it is initially None:

    'parameters': {'command': 'slice', 'start': '0', 'stop': 4, 'reset_coordinates': True, 'simplify': True, 'record_provenance': True}

Although a little unexpected, this seems to reflect the philosophy of provenance data, which reflects the parameters used during runtime, rather than the values initially passed to the command e.g. where the docs say [Note that we include the random_seed here even though it was automatically generated.](https://tskit.readthedocs.io/en/latest/provenance.html#api-invocations)

python/tskit/trees.py
```diff
@@ -3041,11 +3040,13 @@ def simplify(
             sequence.
         :rtype: .TreeSequence or a (.TreeSequence, numpy.array) tuple
         """
-        tables = self.dump_tables()
+        tb = self.dump_tables()
```
hyanwong: have to use 'tb' not 'tables' as the variable so it doesn't get confused with the imported tskit.tables

python/tskit/tables.py
```diff
@@ -1581,7 +1581,7 @@ def simplify(
             filter_zero_mutation_sites=None,  # Deprecated alias for filter_sites
             reduce_to_site_topology=False,
             filter_populations=True, filter_individuals=True, filter_sites=True,
-            keep_unary=False):
+            record_provenance=True, keep_unary=False):
```
hyanwong: Adding `record_provenance` as the penultimate arg (not simply add to the list), so that it matches the argument order in ts.simplify()

jeromekelleher:
This is a neat idea @hyanwong, and I think we're going to have to do some magic like this or it'll be very error prone to do provenance properly. So I like the idea, and I think we should go with it. There'll be a fair bit of work figuring out all the corner cases and making sure that everything is correctly round tripped though.

I think it's a good thing that we record the function arguments as they are, rather than as they are transformed (.e.g, record start=None rather than start=0). Once the defaults don't change (and they won't), this is the same thing. Most of the time, we should be able to call function directly with the provenance dictionary as ``**kwargs`` and get back exactly the same answer. The random seed in msprime is a special case here because we are generating non-deterministic input as part of the function call.

It's debatable whether we should include the ``samples`` argument in ``simplify``. For large tree sequences, we could easily get into the situtation where converting to JSON takes longer than doing the simplification and the provenance is substantially bigger than the rest of the tree sequence. We definitely don't want to do that by default.

python/tskit/provenance.py
```diff
@@ -132,3 +135,44 @@ def validate_provenance(provenance):
         jsonschema.validate(provenance, schema)
     except jsonschema.exceptions.ValidationError as ve:
         raise exceptions.ProvenanceValidationError from ve
+
+
+def provenance_command_JSON(exclude_args=[]):
```
jeromekelleher: I think ``function_args_to_json`` would be more descriptive here.

python/tskit/provenance.py
```diff
@@ -132,3 +135,44 @@ def validate_provenance(provenance):
...
+    The exclude_args argument allows arguments to be excluded from the prevenance info
+    such as deprecated aliases for other arguments (the `self` argument of methods is
+    always excluded).
+
+    It can be used to e.g. serialize the arguments to functions like `simplify`
+    so they can be stored in a provenance field. It requires the arguments to be JSON
+    serializable (and uses np.array_repr to serialize numpy arrays, which are not
+    normally serializable by json).
+    """
+    def serialize_numpy_array(arg):
```
jeromekelleher: Pull this out into it's own function I'd say, so it can be tested independently.

python/tskit/provenance.py
jeromekelleher: You don't want to have a list as a function parameter default: [surprising things happen](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). ``None`` is safer.

jeromekelleher:
> This argues for coding something a bit special for numpy arrays. At the moment we produce text like array([1,2,3], dtype=int8) - i.e. missing the np. Or we could use .tolist() to turn arguments passed as a numpy array into a plain python list, losing the dtype.

Yes, this is tricky. Probably ``.tolist()`` is best if we want to support direct round-tripping,as even if we output ``np.array([1,2,3], dtype=np.int8)``, we would still have to have imported numpy as np, which some people won't do. I don't think the dtype information matters - we are roundtripping through JSON, and we should encode the data faithfully as JSON.

jeromekelleher:
> OK. In this case, using the current approach, we would need to either (a) call the magic provenance-creating-routine at the start of the function, before any variables-passed-as-arguments get overwritten, or (b) make sure that we don't overwrite argument variables (e.g. start)

I think calling the provenance encoding routine the start is a reasonable thing to do and an easy idiom to explain and follow. We can then write the provenance to the table at the end of the method, after the operation has definitely succeeded.

python/tskit/provenance.py
hyanwong: Ah - I didn't know that. Thanks for the tip.

hyanwong:
Requires https://github.com/tskit-dev/tskit/pull/244 to pass tests

hyanwong:
Something to note: there is no way in the provenances to work out if we called `ts.simplify()` or `ts.tables.simplify()`, since these are both stored as the same command-name 'simplify'. This is not really a problem if they both have *exactly* the same arguments. But it might seem odd from the outside. To ensure the parameters are saved the same way both times, I have taken out the substituting of `samples=None` to `samples = ts.samples()` in the simplify() version for a tree sequence, and just pass it straight through as `None` to TableCollection.simplify(), where it gets properly dealt with.

python/tskit/provenance.py
hyanwong: Done - no tests

python/tskit/provenance.py
hyanwong: Done

python/tskit/provenance.py
```diff
@@ -132,3 +135,44 @@ def validate_provenance(provenance):
...
+        if type(arg) == np.ndarray:
+            # NB: we might also want to strip whitespace from the stringified array
+            return np.array_repr(arg, max_line_width=np.inf)
```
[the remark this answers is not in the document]
hyanwong: Changed to `tolist()`

jeromekelleher:
I think we need to put this on the back burner for a bit @hyanwong --- there's tricky technical decisions to be made here, and it's better to have things as they are with clearly incomplete provenance (which people can't write code to rely on), than to ship something that we haven't thought through properly (which people could write code using, and we'd then subsequently have to support, substantially complicating things). There's a lot of other stuff going on that's on the critical path.

I'm sure you've made the major step in solving this problem here, but there's at least a week of solid work in going through all the gritty details and writing lots of tests to made sure it's all watertight.

hyanwong:
Yeah - happy to leave this PR hanging. I can add test cases if I have time, I guess, which are likely to be reused. Another example I just thought of is whether we should allow sets as input to `simplify(samples=...)`, or expect users to convert to a list / numpy array themselves, as sets don't serialize either.

jeromekelleher:
I think we can probably close this? The idea is simple enough to do, it's the downstream consequences that are tricky and I think we have bigger fish to fry than to be messing around with provenance.

jeromekelleher:
Using inspect to automagically generate provenance lead to a major memory leak in msprime (https://github.com/tskit-dev/msprime/issues/1899) so I'm going to have a strong bias against doing things this way from this point...

I think there's very little value in trying to track provenance systematically in tskit calls and we should basically give up on the idea.

hyanwong:
> Using inspect to automagically generate provenance lead to a major memory leak in msprime ([tskit-dev/msprime#1899](https://github.com/tskit-dev/msprime/issues/1899)) so I'm going to have a strong bias against doing things this way from this point...

Yep, a fair conclusion.

> I think there's very little value in trying to track provenance systematically in tskit calls and we should basically give up on the idea.

I'm not sure we should give up it entirely, especially for long-running programs like `tsinfer`, where you definitely don't want to be rerunning the command to check what parameters you ran it with. It could be really helpful to know what mismatch parameters / recombination rates were used to generate an inferred tree sequence, and this information is easy to lose track of, especially if it's an inferred tree sequence from somewhere else / off the internet.

jeromekelleher:
Sure - I agree for downstream programs it's a good idea to keep provenance. I just don't think that trying to keep provenance for things like ``simplify`` is worth it.

petrelharp:
FWIW, I've not used provenance for anything, really (and, in fact, am starting to stick stuff like this in metadata, as it's easier to get to).

PR #244, opened 2019-06-20
Make TableCollections iterable

hyanwong:
No test cases yet, I'll write them once the idea is approved

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -30,6 +30,7 @@ import random
 import string
 import unittest
 import warnings
+import itertools
 
 import numpy as np
 
@@ -1451,6 +1452,17 @@ class TestTableCollection(unittest.TestCase):
         d2 = t.asdict()
         self.assertEqual(set(d1.keys()), set(d2.keys()))
         # TODO test the fromdict constructor
+
+    def test_iter(self):
+        def test_iter(table_collection):
+            table_names = [
+                attr_name for attr_name in sorted(dir(table_collection))
+                if isinstance(getattr(table_collection, attr_name), tskit.BaseTable)]
+            for n in table_names:
+                yield n, getattr(table_collection, n)
+        ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
+        for t1, t2 in itertools.zip_longest(test_iter(ts.tables), ts.tables):
+            self.assertEquals(t1, t2)
 
     def test_equals_empty(self):
         self.assertEqual(tskit.TableCollection(), tskit.TableCollection())
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1506,6 +1506,20 @@ class TableCollection(object):
         title_line += "#"
         return line + "\n" + title_line + "\n" + line + "\n"
 
+    def __iter__(self):
+        """
+        Iterate over all the tables in this TableCollection, ordered by table name
+        (i.e. deterministically), returning a tuple of (table_name, table_object)
+        """
+        yield 'edges', self.edges
+        yield 'individuals', self.individuals
+        yield 'migrations', self.migrations
+        yield 'mutations', self.mutations
+        yield 'nodes', self.nodes
+        yield 'populations', self.populations
+        yield 'provenances', self.provenances
+        yield 'sites', self.sites
+
     def __str__(self):
         s = self.__banner("Individuals")
         s += str(self.individuals) + "\n"


hyanwong:
So now instead of @jeromekelleher 's [ts identify checking code](https://github.com/tskit-dev/tskit/pull/225#issuecomment-502347699) it's now possible to do

    assert all(t1==t2 for t1, t2 in zip(ts1.tables, ts2.tables) if t1[0] != 'provenances')

Which I think is a little cleaner. Or possibly, to avoid string definitions:

    assert all(t1==t2 for t1, t2 in zip(ts1.tables, ts2.tables) if not isinstance(t1[1], tskit.ProvenanceTable))

hyanwong:
Even better, we can easily check identity without the provenance timestamps having to be the same:

```
for t1, t2 in zip(ts1.tables, ts2.tables):
 if isinstance(t1[1], tskit.ProvenanceTable):
  assert all(t1[1].record==t2[1].record)
 else:
  assert t1==t2

```

python/tskit/tables.py
```diff
@@ -1505,6 +1505,17 @@ def __banner(self, title):
         title_line += "#"
         return line + "\n" + title_line + "\n" + line + "\n"
 
+    def __iter__(self):
+        """
+        Iterate over the tables in this TableCollection, ordered by table name
+        (i.e. deterministically), returning a tuple of (table_name, table_object)
+        """
```
jeromekelleher: Or, 
```python
yield "edges", self.edges
yield "nodes", self.nodes
...
```

to avoid the ``getattrs`` (which are surprisingly slow)?

hyanwong: I was trying to avoid actually hand-coding all the tables, in case new tables were added and we forgot to update this particular function. But happy with whatever strikes the correct balance between maintainability and speed.

jeromekelleher: Not planning on adding extra tables, and if we did, this is the absolute least of the problems we'd need to deal with! I bet the total number of characters in the simpler version is about the same as what you have currently anyway.

Using the current version in the tests ensures that an error will be thrown if we do add tables and forget to update this anyway, so it's good to have the automatic version as well.

hyanwong:
@jeromekelleher: do you think it is also worth including the table-check-without-timestamps code as e.g. ```

```
class TableCollection():
    def equal(tc_to_compare, ignore_timestamps=True):
        for t1, t2 in zip(ts1.tables, ts2.tables):
            if ignore_timestamps and isinstance(t1[1], tskit.ProvenanceTable):
                assert all(t1[1].record==t2[1].record)
            else:
                assert t1==t2
```

python/tskit/tables.py
```diff
@@ -1505,6 +1505,21 @@ def __banner(self, title):
...
+        # For speed we simply list these by hand. They are tested against an
```
jeromekelleher: We don't need this comment - it's perfectly obvious what the method is doing.

python/tests/test_tables.py
```diff
@@ -1497,6 +1497,17 @@ def test_asdict(self):
...
 
+    def test_iter(self):
+        def test_iter(table_collection):
+            table_names = [
+                attr_name for attr_name in sorted(dir(table_collection))
+                if isinstance(getattr(table_collection, attr_name), tskit.BaseTable)]
+            for n in table_names:
+                yield n, getattr(table_collection, n)
+        ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
+        for t1, t2 in zip(test_iter(ts.tables), ts.tables):
```
jeromekelleher: This won't spot if one iterator is longer than the other. Need to either make a list of it or use zip_longest.

jeromekelleher:
> @jeromekelleher: do you think it is also worth including the table-check-without-timestamps code in tskit? e.g.

No, to be honest. We're spending more time on making this stuff that we would save.

python/tests/test_tables.py
hyanwong: Well caught.

python/tskit/tables.py
hyanwong: Removed with https://github.com/tskit-dev/tskit/commit/04359b2d5bdba10c493746ba153505b3334a086f

jeromekelleher:
Looks good, thanks and ready to merge. Can you do a quick rebase to bring it up to date please?

hyanwong:
Done

Issue #192, opened 2019-05-28
Cast ID arguments to int32 using to_np_int32


jeromekelleher:
Any place we are providing a list of IDs as input to a function we should use the ``to_np_int32`` function defined in #185 to safely convert them.

These will be in the Tables API (node IDs, site IDs, etc), and where we compute stats (GNN, eg, takes lists of samples). 

Note there are still a number of functions that don't use the numpy C API (e.g., Tree.pairwise_diversity and TreeSequence.trees) but Python lists instead, which are overdue being updated (#64).

hyanwong:
I think we should generalise the casting function, so that it works for e.g. `np.uint8` too (I've just hit this when inputting genotypes: [-1, 0] was invisibly converted to [255, 0], which was hard to track down). The alteration below seems a trivial adjustment of  https://github.com/tskit-dev/tskit/blob/57ff0abe4173611a396e9715632d544195880a75/python/tskit/tables.py#L1822 - but I think the function should be taken out of tables.py and put somewhere more general. Any suggestions as to where?

```
def safe_np_int_conversion(int_array, dtype=np.int32):
    """
    A few functions require arrays of type np.int32. To allow passing standard numpy
    integer arrays (dtype=np.int64) we cast but check bounds to avoid wrap-around
    conversion errors (numpy doesn't seem to provide this functionality)
    """
    int_array = np.array(int_array)
    if int_array.size == 0:
        return int_array.astype(dtype)  # Allow empty arrays of any type
    try:
        return int_array.astype(dtype, casting='safe')
    except TypeError:
        bounds = np.iinfo(dtype)
        if np.all(int_array >= bounds.min) and np.all(int_array <= bounds.max):
            # Raises a TypeError when we try to convert from, e.g., a float.
            return int_array.astype(dtype, casting='same_kind')
        else:
            raise OverflowError("Cannot convert safely to {} type".format(dtype))
```

petrelharp:
Perhaps in a (new) `util.py`?

hyanwong:
I will do this. Assign it to me.

jeromekelleher:
Thanks! Done/.

jeromekelleher:
Does #245 cover *every* usage of numpy int arrays as input @hyanwong?

hyanwong:
[...]

PR #245, opened 2019-06-21
Make to_np_int32 generic for all dtypes

hyanwong:
Function to address #192

### DIFF
--- a/python/tests/test_util.py
+++ b/python/tests/test_util.py
@@ -1,0 +1,103 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Tskit Developers
+# Copyright (c) 2017 University of Oxford
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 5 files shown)


python/tskit/util.py
```diff
@@ -0,0 +1,60 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Module responsible for various utility functions used in other modules.
+"""
+
+import numpy as np
+
+
+def safe_np_int_cast(int_array, dtype, copy=True):
```
jeromekelleher: Probably want this to default to False I'd say - situations where we want to guarantee it's a copy are rare (and probably handled in the C code anyway).

python/tests/test_util.py
```diff
@@ -0,0 +1,103 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Tskit Developers
+# Copyright (c) 2017 University of Oxford
```
jeromekelleher: Can delete this line, as this file is new.

python/tskit/util.py
```diff
@@ -0,0 +1,60 @@
...
+        # Since this is a new numpy array anyway, it's always a copy, so economize by
+        # setting copy=False
+        copy = False
+    if int_array.size == 0:
+        return int_array.astype(dtype, copy=copy)  # Allow empty arrays of any type
+    try:
+        return int_array.astype(dtype, casting='safe', copy=copy)
+    except TypeError:
+        bounds = np.iinfo(dtype)
+        if np.all(int_array >= bounds.min) and np.all(int_array <= bounds.max):
```
jeromekelleher: More idiomatic here to catch and throw the error, then continue the function as normal. Emphasises what the normal path is:
```python 
if np.any(int_array < bounds.min) or np.any(int_array > bounds.max):
    raise OverflowError("Cannot convert safely to {} type".format(dtype))
if int_array.dtype.kind == 'i' and np.dtype(dtype).kind == 'u':
# As before
```

python/tests/test_util.py
hyanwong: You mean just the 2017 line?

python/tskit/util.py
hyanwong: Yes, nicer, I agree

python/tests/test_util.py
jeromekelleher: Yes, just the 2017 line.

jeromekelleher: Maybe strictly this should be "Copyright (c) 2019 Tskit Developers" as well, since it's a new file.

jeromekelleher:
Excellent, thanks @hyanwong. This is a very useful addition.

python/tests/test_util.py
jeromekelleher: Doesn't matter, ignore the last comments (didn't see it had been resolved).

hyanwong:
@jeromekelleher - did you see my comment above. There's a weirdness at e.g. https://github.com/tskit-dev/tskit/blob/4d6fdf05b68bff19cabeb79340ef0842ce3dfdf7/python/tskit/trees.py#L3258 where 3 lines above the array is formed as uint32, then cast into int32. I'm not sure why this is, but it seems wrong. There are a few further places in the same file where this happens too.

jeromekelleher:
I can't find the comment, but this is correct. It's two different arrays, right (sample_sets and sample_set_sizes)?

hyanwong:
Sorry, you are right. I'm being thick.

Issue #247, opened 2019-06-25
Any benefit of multiple sample_sets in allele frequency spectrum


jeromekelleher:
See also #203.

Working through the algorithms for the AFS it's not obvious to me that there's any real benefit perf-wise to providing multiple sample_set arguments. When we're computing the AFS values, the only thing we're actually sharing between the different sample sets is the ``parents`` array (the clinching argument for allowing other run concurrently is that we share some calculations across different output stats, thereby making it all more efficient).

When we allow for multiple sample sets in the AFS functions we are hit with choices about how we shape the output array. Firstly, to make the arrays rectangular, we need to make the frequency dimension equal to the size of the largest sample set, which is potentially wasteful if one sample set is much larger than the other. The second choice is which order the dimensions should go - it's not clear to me whether either ordering is better.

Given all this, I wonder if it's worth the trouble allowing for multiple sample sets when computing the AFS, particularly when you'd almost certainly be better off running several of these in parallel rather than the current vectorised version - the memory access patterns will be pretty nasty, I think.

Thoughts @petrelharp?

petrelharp:
Sounds good - only one sample set is fine. Although it's maybe worth thinking ahead to how the joint AFS will work.

jeromekelleher:
Good point - is the joint AFS you're thinking of like the ``get_joint_site_frequency_spectra`` [here](https://github.com/jeromekelleher/spg-chapter/blob/master/jupyter/inference-example.ipynb)? I.e., the indexes are given by ``num_samples_ancestral``, ``num_samples_not_ancestral`` for a single sample set?

If so, probably simplest to add another method ``joint_allele_frequency_spectrum``, to keep the output dimensions simple?

jeromekelleher:
Or more generally, would we want to give k sample sets as a parameter and return a k-d array for each window?

petrelharp:
> Or more generally, would we want to give k sample sets as a parameter and return a k-d array for each window?

Yes, this. How about this:
- `allele_frequency_spectrum(samples, ...)` and
- `joint_allele_frequency_spectrum(sample_sets, ...)` 
The first function is redundant, since `allele_frequency_spectrum(samples)` would equal `joint_allele_frequency_spectrum([samples])`, but not including the first function would lead to confusion in a number of ways.

jeromekelleher:
Done! I'll code this up over in #248

jeromekelleher:
Closed in #248 and #274.

Issue #203, opened 2019-06-11
sample frequency spectrum


jeromekelleher:
Add support for site frequency spectrum. Needs a specific C algorithm for efficiency.

I think we should call it the "sample_frequency_spectrum", since it's counting samples. site_frequency_spectrum is actively misleading I think, as well as leading to absurd function names like ``site_site_frequency_spectrum`` and ``branch_site_frequency_spectrum``. I'm open to other names though.

See also #196.

petrelharp:
Another name is "allele frequency spectrum", which is more what we are computing (the distribution of frequencies of alleles).

From email:
> Re the SFS, it wasn't obvious to me why you left out the total branch length over 0 samples - this is still a useful thing, right?

Isn't this always zero?

jeromekelleher:
> Isn't this always zero?

Not if your sample set size is < n - isn't it the branch length over "the complement of this sample set" then?

jeromekelleher:
> Another name is "allele frequency spectrum", which is more what we are computing (the distribution of frequencies of alleles).

I like it...

petrelharp:
> Not if your sample set is not all the samples - isn't it the branch length over "the complement of this sample set" then?

Oh, sorry - I misread that. Right, the `0`th entry of the AFS. Yes, that could be useful (although since it depends on what else is in the tree sequence, caution is needed).

jeromekelleher:
I've worked through the algorithm for the AFS @petrelharp, and the zeroth element works like you would expect **almost** all the time. The only time we get weird results where the obvious definition algorithm differs from the efficient edge-based algorithm is when we have edges lying around that are not ancestry to any sample (e.g., unsimplified forward sim). In this case, to output the same thing as the tree based definition algorithm we'd need to keep track of the count of **all samples** that are under each node as well as the sample sets, and only update the AFS when this is > 0. The definition of AFS[0] is then, "the span weighted total branch length along branches ancestral to *some* sample that is not in the current sample set."

This might not be worth the bother though, and we may be better off just setting AFS[0] = 0 always and documenting this. I do think we should keep things simple and make AFS[k] refer to the total branch length over k samples rather than trying to save space by making it k - 1 (or k+1, whichever). I think this would be a source of persistent off by one errors that would make people hate us forever.

petrelharp:
>  The only time we get weird results where the obvious definition algorithm differs from the efficient edge-based algorithm is when we have edges lying around that are not ancestry to any sample

[...]

PR #248, opened 2019-06-25
Allele frequency spectrum algorithm

jeromekelleher:
Updates the branch allele frequency spectrum algorithm. 

Will wait on input on #247 and #203 (on zeroth element) from @petrelharp before going further.


jeromekelleher:
I've added the basic algorithm for the joint branch AFS @petrelharp, and I think this is good to implement (although still not sure about the zeroth position in the AFS). 

I made a start on the site version, but then realised I didn't know what to do when there's multiple alleles. I assume we want to make it so we're counting everything that isn't ancestral (i.e., force a biallelic system)? Or do we want to add another dimension, one for each allele? Or do we want to specify the allele in question that we're getting the spectrum for?

petrelharp:
Well, to be consistent with everything else we can only use the information about which allele is ancestral if `polarised=True` (nb: we haven't actually put this option into any stats yet; this would be a good place to start). All other stats we are computing the "site" version as if we had the genotypes and nothing else (so, no info about backmutations or ancestral state). And, I think the AFS is the "distribution of allele frequencies", and at a site with three alleles, well, there are three alleles that ought to count. This also makes the AFS somewhat closer to what we compute based on the infinite-sites model.

I think what makes the most sense is to iterate over alleles, and add 1 to `AFS[k-1]` (or, `AFS[k]` if there's a zero slot) for each allele present in `k` copies.  So, if the haplotypes are the rows of
```
0011
0112
```
then the AFS would be, starting with singletons, `[2,3,0]`.  This differs from the usual thing in two ways: first, since both alleles at each biallelic site get counted, we're computing something more like the folded spectrum. The folded spectrum would be the `floor(n/2)`-vector `fAFS[k] = (AFS[k] + AFS[n-k])/2`. Second, the folded AFS computed this way would add up to something more than the number of polymorphic sites (it would add up to the number of alleles at polymorphic sites, over two). But, that's OK; just we need to make sure to document this.

So, I guess my proposal is, somewhat modified, this:

1. `polarised`: Add 1 to `AFS[k-1]` for each non-ancestral allele in `k` copies. Return for each window an `n-1`-vector, where `n` is the length of the sample set.
2. `unpolarised`: Add 1/2 to `AFS[min(k-1, n-k-1)]` for each allele in `k` copies. Return for each window a `floor(n/2)`-vector.

Alternatively, we could let the user "fold" it themselves, in which case the "unpolarised" version wouldn't really be unpolarised.

jeromekelleher:
I've had a go at implementing this @petrelharp --- it'd be great if you could try it out and see if it does what you expect. There's a few dangling issues about the zeroth AFS entry, but I haven't spent time working out what's happening here as we might not keep it.

Once we're all happy it's implementing the right thing I'll code it up in C.

Question: do we want the 'node' K-D joint AFS? (Please say 'no'!)

jeromekelleher:
I think I've fixed the conceptual problems here by only updating the AFS when at least one of the counts is > 0 - this made all the nasty problems in the corner cases go away. I haven't dealt with what should go in the AFS[0] in this case.

Coding up the C version is a real drudge exercise which made me think again about whether we can do this in the general framework. I think we could - discussion over in #257.

petrelharp:
> Question: do we want the 'node' K-D joint AFS? (Please say 'no'!)

Sure: `'no'`

Seriously, this seems fine to omit: it woudl probably return an array that is waaaay too big (as would `mode="site", windows="sites"`, incidentially). Although if it is sufficiently easy we should do it.

jeromekelleher:
I've updated this with the C version of the jAFS with all the requisite plumbing @petrelharp --- seems to work fine and the code duplication isn't too bad. There's a bunch of things that need to be ironed out, but I think the basic outline is good.

One thing it would be good to sort out: having different shapes for the arrays when we have polarised and not polarised is pretty tedious. Maybe we could leave this out? It'll use 2X as much memory as we need, true enough, but folding afterwards in Python would be so much simpler (we could add an option ``fold`` at the top level, and do the folding in numpy; it seems likely we can do this efficiently).

jeromekelleher:
Note to self: debugging the issue on Windows gave:
```
CASE 1: 1622122240
CASE 1: 1608453120
```
So ret is somehow getting assigned these very large values. Very weird...

jeromekelleher:
Doh, we weren't initialising ret. Wasn't being picked up by C tests because there's no tests on an empty ts/something with no sites. TODO: add test case where we create an empty ts and we run a bunch of stats on it.

petrelharp:
> Maybe we could leave this out? It'll use 2X as much memory as we need, true enough, but folding afterwards in Python would be so much simpler (we could add an option fold at the top level, and do the folding in numpy; it seems likely we can do this efficiently).

Totally. I support this option.

c/tskit/trees.c
```diff
@@ -1836,6 +1904,249 @@ tsk_treeseq_sample_count_stat(tsk_treeseq_t *self,
...
+    }
+    ret = get_allele_weights(site, counts, num_sample_sets, total_counts, &num_alleles,
+            &allele_counts);
+    if (ret != 0) {
+        goto out;
+    }
+
+    afs_size = 1;
+    for (j = 0; j < num_sample_sets; j++) {
+        afs_size *= result_dims[j];
```
petrelharp: shouldn't `afs_size` be passed in, at least so it isn't re-computed for every site?

jeromekelleher: Yes, that would probably help.

petrelharp:
Doing some spot checks, I see that with `polarised=False` there is a nonzero number in the `[0, 0]` entry, but that entry is zero if `polarised=True`. Maybe this is moot if we're going to fold at the python level.

petrelharp:
Maybe we should name this just `allele_frequency_spectrum`, without the `joint_`? Since there is no non-joint method, it's a bit redundant, and makes the method harder to find.

python/tests/test_tree_stats.py
```diff
@@ -2515,183 +2548,372 @@ class TestSitef4(Testf4, MutatedTopologyExamplesMixin):
...
-        self.assertEqual(sfs1.shape, sfs2.shape)
-        # print(sfs1)
-        # print(sfs2)
-        self.assertArrayAlmostEqual(sfs1, sfs2)
-        # print(sfs2.shape)
-
-
-class TestBranchSampleFrequencySpectrum(
-        TestSampleFrequencySpectrum, TopologyExamplesMixin):
+        # TODO implement span_normalise in the low-level code and iterate over this too
```
petrelharp: I wondered if we really need span_normalise... but, we do, because if we don't span normalise then the AFS computed in windows of dfferent lengths are not comparable.

c/tskit/trees.c
```diff
@@ -1836,6 +1904,249 @@ tsk_treeseq_sample_count_stat(tsk_treeseq_t *self,
...
+    double increment, sum, *afs, *allele_counts, *allele_count;
+    tsk_size_t *coordinate = malloc(num_sample_sets * sizeof(*coordinate));
+    bool polarised = !!(options & TSK_STAT_POLARISED);
+
+    if (coordinate == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    ret = get_allele_weights(site, counts, num_sample_sets, total_counts, &num_alleles,
+            &allele_counts);
```
petrelharp: pulling this out into `get_allele_weights` was a very good call.

petrelharp:
Here's a method to `fold` in just numpy. Not the most efficient, but a start. It was a little painful.
```
def fold(a):
    for axis in range(1, len(a.shape)):
        n = a.shape[axis]
        m = int(np.ceil(n/2))
        # note that modifying b and c *still modify a*
        b, c = np.split(a, [m], axis)
        if (n % 2) == 1:
            k = int(np.prod(a.shape) / n)
            ind = list(c.shape)
            ind[axis] = 1
            c = np.concatenate([np.zeros(ind), c], axis=axis)
        a = b + np.flip(c, axis)
    return a
```

petrelharp:
Top:

red is branch diversity from all samples
black are site diversity from all samples across indept replicate mutations

Bottom:

red and black are branch, site diversity respectively from independent subsamples of size 100 from the same instance (including same mutations)

Screenshot from 2019-07-13 13-56-09

Note that the branch diversity from a subsample tracks that for everyone quite well!

One more thing to add: comparison of branch and site with mutation rate cranked way up (to show it's the same).

petrelharp:
With 1000 individuals:
Screenshot from 2019-07-14 09-46-28

and with 10000 (which takes hours, mostly because of putting down mutations, I think, surprisingly?)
Screenshot from 2019-07-14 09-46-17

jeromekelleher:
Very nice @petrelharp! I think this is on the wrong PR though?

Re adding mutations: that could definitely be made faster.

jeromekelleher:
> Maybe we should name this just allele_frequency_spectrum, without the joint_? Since there is no non-joint method, it's a bit redundant, and makes the method harder to find.

I think we can add a top level ``allele_frequency_spectrum`` that takes a single sample set, and strips an output dimension. This would be the most user friendly option I think?

jeromekelleher:
If we're supporting ``fold`` @petrelharp, then it's probably best to do it at the C level. We already have the code written, and I think the memory management and so on would make more sense then. What's counter intuitive to me at the moment is the interplay between polarisation and folding. If ``fold`` is a top level option (independant of ``polarised``) which we pass through then I think it's all simple enough.

What do you think?

petrelharp:
> I think we can add a top level allele_frequency_spectrum that takes a single sample set, and strips an output dimension. This would be the most user friendly option I think?

Sure, why not.

I agree, it makes sense to do this in C especially since you've already got it done there. But what is the question about polarisation/folding?  My proposal was that if `polarised=False` then the result should be folded, because if you are unpolarised then the `k` and `n-k` bins are the same.

jeromekelleher:
> I agree, it makes sense to do this in C especially since you've already got it done there. But what is the question about polarisation/folding? My proposal was that if polarised=False then the result should be folded, because if you are unpolarised then the k and n-k bins are the same.

Well, I just find it confusing that  polarised=False changes the dimensions and it feels unnatural to fold as well as ignoring the contribution of the ancestral state. Plus, this the main reason that the AFS doesn't fit nicely into the general framework --- if folding was treated separately, it could be implemented as a second summary function. I would feel much cleaner to me if ``polarised`` and ``folded`` were two independent parameters. I'm not sure what an AFS with folded=False and polarised=False would actually *mean* though!

Hmm, I'm probably over thinking this; maybe we should just leave it as it is, so that polarised=False folds and changes the dimensions of the output.

petrelharp:
These are good points, but I think the definition of the folded AFS is an array with all the dimensions divided by two. I'm not opposed to having a separate `fold` parameter, (with default equal to `not polarised`) but let's see:

- `polarised=True, fold=True` would in fact be unpolarised, unless there's a distinction in the `[0]` slot?
- would `polarised=False, fold=False` in fact be polarised? or would the `[k]` and `[n-k]` slots be the same?

So if `polarised` and `fold` actually do something different (maybe just in the `[0]` slot?) then we could do this?

jeromekelleher:
> So if polarised and fold actually do something different (maybe just in the [0] slot?) then we could do this?

Maybe it does in the branch length case? I'm too confused to have an opinion at this point! I think we should shelve it - it's clear that the default for ``fold`` would be ``not polarised``, so we're not doing anything forwards incompatible here if we leave it out.

I've implemented the branch AFS in C now, so we should be almost there. Just a few wrinkles to work out with the zeroth element and documentation, and I think we're done.

petrelharp:
Ok, so propsal is to leave as-is: `polarised` returns a folded AFS. This seems good to me.

jeromekelleher:
> Ok, so propsal is to leave as-is: polarised returns a folded AFS. This seems good to me.

Cool. Apologies for the roundabout way of getting here!

jeromekelleher:
OK, merging this. There's still some outstanding issues, but let's get this much shelved away.

PR #251, opened 2019-07-01
linear regression

petrelharp:
The math is in the treestats paper; I think all the bits are there but with some errors yet.

I ended up duplicating a moderate amount of code, from the "weighted_stats" because this case is different to covariance and correlation (in that it has another argument, "covariates"). I haven't thought of a better way to do this yet - hampered by a lack of examples. So maybe it's fine; when/if we have more examples we can generalise appropriately.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1758,6 +1758,12 @@ typedef struct {
 } weight_stat_params_t;
 
 typedef struct {
+    tsk_size_t num_samples;
+    tsk_size_t num_covariates;
+    double *V;
+} covariates_stat_params_t;
+
+typedef struct {
     tsk_id_t *sample_sets;
     tsk_size_t num_sample_sets;
     tsk_size_t *sample_set_sizes;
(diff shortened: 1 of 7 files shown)


petrelharp:
This works, finally!  Mostly I had to track down some issues having to do with singular matrices.  NB: I ran the tests with `p=1.0`, which took awhile, but passed.

petrelharp:
ok, for realsies, without memory leaks.

petrelharp:
So, I added a check that our random checks are actually checking things. This turns out to be problematic, because, I guess, they aren't always. =) Thinking about a fix still.

python/tests/test_tree_stats.py
```diff
@@ -44,18 +44,29 @@
 np.random.seed(5)
 
 
-def subset_combos(*args, p=0.75):
+def subset_combos(*args, p=0.5, min_tests=3):
```
petrelharp: ok, so now this is accumulating a sorta random list of skipped combinations, that it will return to if there haven't been at least `min_tests` done. Also note that the random seed is now set *outside* this function, so the sequence of tests/don't test is different for each method (important).

petrelharp:
I'm hoping this is good, now. The docs could be better, but this is hard to explain, concisely.

c/tskit/trees.c
```diff
@@ -2012,6 +2018,113 @@ tsk_treeseq_trait_correlation(tsk_treeseq_t *self,
...
+    size_t i, j;
+
+    m = x[state_dim - 1];
+    for (i = 0; i < result_dim; i++) {
+        if ((m > 0.0) && (m < num_samples)) {
+            v = GET_2D_ROW(V, k, i);
+            a = x[i];
+            denom = m;
+            for (j = 0; j < k; j++) {
+                xx = x[result_dim + j];
```
jeromekelleher: This should be ``state_dim``, shouldn't it?

c/tskit/trees.c
```diff
@@ -2012,6 +2018,113 @@ tsk_treeseq_trait_correlation(tsk_treeseq_t *self,
...
+                a -= xx * v[j];
+                denom -= xx * xx;
+            }
+            if (denom < 1e-8) {
```
jeromekelleher: Hmmm; feels very arbitrary. We need a comment describing why this is done at the least. Should it be a tuning parameter, passed in to the top-level method?

c/tskit/trees.c
```diff
@@ -2012,6 +2018,113 @@ tsk_treeseq_trait_correlation(tsk_treeseq_t *self,
...
+                result[i] = 0.0;
+            } else {
+                result[i] = (a * a) / (denom * denom) / 2;
```
jeromekelleher: Is this the same as ``(a * a) / (2 * denom * denom)``? Or is it ``(2 * a * a) / (denom * denom)``. It'd be good for readability to either put in some brackets or reorganise it a bit.

python/tests/test_tree_stats.py
jeromekelleher: But this means that we'll have different tests run depending on the context - if we run one test on it's own, it'll be different to if we ran it with all other tests, or just one class. I've had this before, and it's horrible when trying to track down bugs. We should pass the random seed as a parameter. I don't think it's that important to have different test/don't test combinations for different methods, but we could come up with something that passes in a different seed which is a hash of the calling method or something like that.

python/tskit/trees.py
```diff
@@ -3448,6 +3449,8 @@ def trait_covariance(self, W, windows=None, mode="site", span_normalise=True):
             window (defaults to True).
         :return: A ndarray with shape equal to (num windows, num statistics).
         """
+        if W.shape[0] != self.num_samples:
+            raise ValueError("First trait dimension must be equal to number of samples.")
```
jeromekelleher: Is this not caught by the low-level code, or are you giving a more user-friendly error message? (More user friendly message is good!)

petrelharp: The second thing: the low level error is quite cryptic:
```
  File "/home/peter/projects/tskit-dev/tskit/python/tskit/trees.py", line 3584, in trait_regression
    mode=mode, span_normalise=span_normalise)
ValueError: object of too small depth for desired array
```
... although we should actually fix this for all the stats at once.

c/tskit/trees.c
petrelharp: Nope - here, `x[0], ..., x[result_dim - 1]` is the internal state corresponding to the traits, `W`, and `x[result_dim], ..., x[state_dim - 2]` is the state corresponding to the covariates, `Z`, and `x[state_dim - 1]` is the "tracked sample count" column. (But, maybe I should have called this `z` instead of `xx`.)

c/tskit/trees.c
petrelharp: Maybe this should be a `tol` parameter. In practice I was getting values of like 1e-15 here, so something like that would probably be fine, so this is rather conservative. 

But: denom is equal to the norm of the orthogonal projection of the genotype vector into the subspace spanned by the covariates.  A value of `denom` less than 1e-8 here means that the genotype is *very* close to being exactly in the span of the covariates.  For denom to be of size 1e-8 but the genotype *not* be in the span of the covariates, given that the genotype vector is binary, you'd need something like, say, sex as a covariate and a SNP that exactly segregates between sexes except for in one person out of a sample size of 100 million. So, yeah. `denom` is only anywhere close to zero in in either test data, where we've got like four samples and four covariates, or in cases where the genotype is very tightly correlated with a covariate, someplace that it's probably good to be conservative.

So: I'll add a comment?

c/tskit/trees.c
petrelharp: whoops, yes!

python/tskit/trees.py
petrelharp: Hm, actually, this is just in correlation and covariance also; checking like this is fine.

c/tskit/trees.c
jeromekelleher: What happens if we just let it pass through though? Do things blow up, or do we just get something quite large (I guess, since we're dividing by 1e-16?). What's the harm in getting back these values?

If we have to have a threshold for numerical reasons then we'll have to add it as a parameter in the C API; unlike Python, we can't just add new parameters later, unfortunately. If we hard code it in, in practise we'd end up having to add a second version of this function in with this parameter included.

c/tskit/trees.c
jeromekelleher: I see, thanks for clarifying

python/tests/test_tree_stats.py
petrelharp: Hm. I think it's important to have different combinations for different methods, because our different tests are testing different things, so if if *all* of the combinations skip one, then we're missing that out.

Your suggestion is good, but I don't know how to get different seeds in there. We can't just put in a seed in the call to `subset_combos( )` because then every derived class will use the same seed.  Hashing the call stack, maybe, but would the call stack really be the same when running all tests and running just the one you're interested in? Seems very tricky.

Before going down this rabbit hole, let me point out that if you want to re-find a failed test, you can just set `p=1.0` in the code (eg by uncommenting line 52 below).

c/tskit/trees.c
petrelharp: Actually, in practise we seem to get not-large values, because the numerator is small also. We don't want these because (a) we can't test them and (b) if the denominator is zero then the solution is non-unique, and the one we want to return here is 0.

I suppose we could add this as a parameter to the C API, although I really don't think anyone's going to want to change it, ever. I'm still voting to leave it out.

petrelharp:
I've got a solution to the nonrandom tests: see the last commit. It's using the same seed for verify( ) for each of the tests in the same name/mode combination, but that's probably ok.

python/tests/test_tree_stats.py
jeromekelleher: That's a good point - we can always reproduce specific combos by running them all on a particular test. OK, let's not worry too much about this, sorry for the noise.

c/tskit/trees.c
jeromekelleher: OK, let's leave it as it is then.

jeromekelleher:
Looks great @petrelharp - I think this is ready to merge after a rebase.

petrelharp:
Still not totally happy with the random checking, but maybe I'll figure it out later.

python/tests/test_tree_stats.py
petrelharp: oh, so you'd be ok with putting the seed back up to the top, rather than hashing? I'd prefer that.

jeromekelleher: Yeah, I guess. I have a feeling it'll come back to bite us, but I don't have a better suggestion so let's just set it at the module level and run the tests non-determinisitically.

petrelharp:
I've pushed this back to setting the seed globally. I say we merge this.

jeromekelleher:
Done!

Issue #209, opened 2019-06-11
Python C module no longer covered in CodeCov


jeromekelleher:
We're not getting coverage numbers here, need to update CI config.

PR #255, opened 2019-07-03
Numpy samples

jeromekelleher:


jeromekelleher:
TODO: check the codecov status for this and fixup if not OK (#209)

jeromekelleher:
Note to self: there's something weird happening with the compile step here where we don't seem to be compiling the Python module with coverage turned on. There's some debug cruft in here, so make sure to clear this up before merging.

jeromekelleher:
Merging this --- the low Python coverage is due to the addition of the low-level C module. Will follow up to bring up coverage there.

PR #261, opened 2019-07-11
working towards a slice that will work on intervals

andrewkern:
this is a baby step PR, aimed at getting the `slice()` function to work on multiple intervals. in talking with @petrelharp we figured the best start would be to implement this in a test, open a PR, and then move on. Thoughts on this inefficient implementation?


python/tests/test_topology.py
```diff
@@ -45,43 +45,52 @@ def slice(
...
+                        site.position - aRange[0], site.ancestral_state, site.metadata)
+                else:
+                    site_id = tables.sites.add_row(
+                        site.position, site.ancestral_state, site.metadata)
+                for m in site.mutations:
+                    tables.mutations.add_row(
+                        site_id, m.node, m.derived_state, m.parent, m.metadata)
     if reset_coordinates:
-        tables.sequence_length = stop - start
+        tables.sequence_length = stop[0] - start[0]  # this is not ideal
```
petrelharp: This should be `sum([y - x for x, y in zip(start, stop)])`, if we cut out all between-interval bits?

python/tests/test_topology.py
```diff
@@ -45,43 +45,52 @@ def slice(
...
-            tables.edges.add_row(
-                max(start, edge.left), min(stop, edge.right),
-                edge.parent, edge.child)
-    for site in ts.sites():
-        if start <= site.position < stop:
+        for aRange in zipRanges:
+            if edge.right <= aRange[0] or edge.left >= aRange[1]:
+                # This edge is outside the sliced area - do not include it
+                continue
             if reset_coordinates:
```
petrelharp: will `reset_coordinates` just shift everything to `start[0]` or snip out all between-interval spaces? I vote for the latter.

andrewkern: currently I'm not dealing with it at all. I figure that if there is more than one interval we probably don't want to allow a coordinate reset.

petrelharp:
Looks good, and brings up some API questions:

- As written, `start` and `stop` need to be lists/vectors. To avoid backwards-incompatible changes, we could check if `start` or `stop` are single numbers and if so replace them by e.g. `[start]`. It would also be nice to not have to remember the `[ ]` in the one-window case.
- what should `reset_coordinates` do? see notes in the code.

python/tests/test_topology.py
petrelharp: I'd say why not allow it? It wouldn't be hard to shift the coordinates to totally snip out those bits.

andrewkern:
the first item here is easy enough. I can alter that.

the second is a whole can of worms. what would be the point of resetting the coordinate space on a multiple interval slice? when would that be useful? not saying it isn't, I'm just not sure when we would want that at all.

petrelharp:
>  I'm just not sure when we would want that at all.

I guess that I'm not sure either. I'm just trying to make things match up with the previous, `.slice()` function.  An alternative would be to keep `slice` as is and make this function called `.mask()`. This seems redundant.

How about go ahead an implement without `reset_coordinates`, we can see how easy it is to add?

andrewkern:
worked on it this evening and think i have the `reset_coordinates` thing working. i also began fleshing out a test-- see `test_multi_interval_slice()`. obviously a bogus test still

python/tests/test_topology.py
```diff
@@ -45,43 +45,54 @@ def slice(
...
+    if np.ndim(start) == 0:
+        start = [start]
     if stop is None:
-        stop = ts.sequence_length
-
-    if start < 0 or stop <= start or stop > ts.sequence_length:
+        stop = [ts.sequence_length]
+    if np.ndim(stop) == 0:
+        stop = [stop]
+    zipRanges = list(zip(start, stop))
```
jeromekelleher: Style nitpick: can we keep to snake_case for consistency please?

jeromekelleher:
LGTM. I'm not sure I can see a way to doing it efficiently with numpy though - I think we'll have to do this in C. If this is something we want urgently for stdpopsim I can jump in - just let me know.

andrewkern:
okay i changed that camelCase name (sorry), rebased to squash commits, and force pushed. i do think we want this for stdpopsim so if you could jump in on the C implementation @jeromekelleher that would be amazing!

python/tests/test_topology.py
andrewkern: sorry... old habits die hard

jeromekelleher: No worries!

jeromekelleher:
OK, I'll have a go when I get a chance over the next few days. I'll push to this branch to keep it all the same PR.

petrelharp:
The numpy implementation should be good as long as there are not a large number of intervals. I suppose you're thinking of something that doesn't scale with the number of intervals? Oh, I see - it would be pretty straightforward using sortedness and edge insertion/removal indices. It still sounds tedious: I'm very tempted to say we should just see how the python implementation does with whatever we need in stdpopsim, though.

jeromekelleher:
I've made some updates here and started writing out a set of tests. I've realised that actually properly covering all the corner cases here properly will take a *lot* of effort, and I'm starting to wonder if this is the right approach in the first place.

First, I've reset the ``slice`` function back to its original definition, as it seemed that providing a list of intervals was more natural that two lists of start, stop values.

I then added a ``dice`` function, which seemed like a nice generalisation of slice. However, after making a start on probing the corner cases for this, I think it's the wrong thing to do:

- I think the UI is in the wrong direction: we should specify the intervals we want to **remove** rather than the ones we want to keep. This is what we actually want for stdpopsim anyway.
- I think ``reset_coordinates`` is a bad idea here. Firstly, there is a tricky numerical problems, where the sum of the interval lengths is not exactly equal to the right coordinate of the rightmost edge. I'm sure there'd be others as well if we tested this thoroughly. Secondly, why would we want to do this anyway? Surely we want to preserve the coordinates? For example, if we wanted to plot the tree sequence against the original recombination map, it'll be an error-prone mess if we squish all the coordinates together like this. There's no harm in having gaps with no topology in the middle of the tree sequence - they don't contribute any sites (if we remove them) and they don't affect the branch length statistics.

So, I propose we get rid of ``dice``, and have a new method which removes information from a given list of intervals. We did something similar for the tree sequences inferring by tsinfer for 1000 genomes etc where we snipped out the toplogy for centromeres (but that was only a single interval). Not sure what it should be called though...

Also, not convinced that ``slice`` is the best name for the current slice function in this light. We haven't released it, so we're free to change it if we want.

andrewkern:
I agree about `reset_coordinates` -- it's a bit esoteric.

`dice` could run either direction i suppose (erase or retain the intervals). should we just add an option to the function?

jeromekelleher:
> I agree about reset_coordinates -- it's a bit esoteric.

Great - I'm much happier about implementing/testing this if we don't have to do this.

> dice could run either direction i suppose (erase or retain the intervals). should we just add an option to the function?

That's true, you probably would want to do it in both directions for different applications. Is ``dice`` the right name, I wonder?

andrewkern:
one possible name for this function would be `subset_ranges` or simply `subset`. 

so what is the plan for this PR @jeromekelleher? do you want to merge a renamed `dice`?

jeromekelleher:
> so what is the plan for this PR @jeromekelleher? do you want to merge a renamed dice?

I'd like to add a function in this PR that'll support what we need for the analysis repo. I think there's two options:

- Have something like ``dice``, which can return a table collection with either a set of intervals removed or retained (but, without reset_coordinates).
- Make something that just deletes a set of intervals from the edges and sites. We could call this ``delete_intervals`` or ``delete``:

```python
def delete(self, intervals):
     """
     Returns a copy of this TableCollection with the tree topology and site data in the specified 
     list of genomic intervals deleted.  Intervals must be disjoint and sorted in increasing order.
      """
```

This seems clear and straightforward to implement, without an unnecessary complications. I guess we might want 'delete' for something later, but I can't see what.

jeromekelleher:
@petrelharp, have you any thoughts here? The ``delete`` option seems simplest to me and does the job we want.

jeromekelleher:
On a different note, maybe we should rename ``slice`` to ``extract_slice`` or something as well.

andrewkern:
@petrelharp and I talked more about this today. For the `analysis` repo I'm becoming less certain that we even want to `dice` the tree sequence, rather than just mask out the sites from analysis. If we were to use a nicely implemented `dice`, we would still have to simulate the full tree sequence for the chromosome, so there is no savings in computation. 

fodder for tomorrow's call.

jeromekelleher:
True enough - we should also think about what would happen to branch length statistics. I guess either way we'd want to mask out the windows that include these areas of the genome, so it's not going to make any difference whether we ``dice``/``delete`` it first or not.

andrewkern:
i've been playing with the `dice` function and added two more tests with more intervals to dice (10, 100). It seems to be very fast with this number of intervals. probably in fine shape for use with the stdpopsim analysis.

python/tskit/tables.py
```diff
@@ -1733,6 +1744,147 @@ def deduplicate_sites(self):
         self.ll_tables.deduplicate_sites()
         # TODO add provenance
 
+    def slice(
+            self, start=None, stop=None, reset_coordinates=True, simplify=True,
+            record_provenance=True):
+        """
+        Returns a copy of this set of tables which include only information in
+        the genomic interval between ``start`` and ``stop``.  Edges are truncated
+        to this interval, and sites not covered by the sliced region are discarded.
```
petrelharp: Should perhaps clarify that "The node and individual tables are not modified, so note that after this operation there may be nodes in the node table not ancestral to any samples (this does not affect most operations, however)."

petrelharp: Never mind; this is covered by the `simplify` option below.

petrelharp:
@andrewkern just did some speed tests on a tree sequence with >1M edges, and found that with 10 intervals it took <1sec, with 100 intervals it took 4sec and with 1000 intervals, 42sec. We will never have 1000 intervals, and we only have to do this once, so the current implementation seems good enough for now.

petrelharp:
LGTM except for documentation of `dice`.

petrelharp:
Well, let me discuss the name for a minute? Seems to me this could just be called `slice`, and have only one function - they do just the same thing, no? The name `dice` is nice but also risks initial confusion with regular polyhedra with numbered faces.

petrelharp:
> what would happen to branch length statistics

... or, any statistic that is span normalised. Good point. A nice solution to this would add a `mask` argument to the statistic, or even as a property of the tree sequence. But in the meantime at least it's easy to have windows with breakpoints aligning with masked regions and omit the masked-out ones.

jeromekelleher:
> Well, let me discuss the name for a minute? Seems to me this could just be called slice, and have only one function - they do just the same thing, no? The name dice is nice but also risks initial confusion with regular polyhedra with numbered faces.

Maybe this got lost above, but I don't think it should be called dice either, and I'm not sure 'slice' should be called slice as well. I'm voting for ``delete``, which **removes** a set of intervals. We rename ``slice`` to ``extract_slice``. We want to mask for stdpopsim, as we're finding the intervals we don't want, right?

petrelharp:
I think we'll have to square this away on Monday, and I've got a bit confused, but: is the proposal is to have *two* functions, one which *removes* intervals, and the other which *keeps* intervals? Obviously, there'd only have to be one implementation under the hood. This seems fine. If so, I think we should name the functions something close, e.g. `extract_slice` and `delete_slice`, so it's obvious they are in the same class of operations. (and, those names are fine with me)

jeromekelleher:
OK, I like ``extract_slice`` and ``delete_slice`` so long as (a) the argument is a list of intervals (not start, stop), and we remove the ``reset_coordinates`` argument. Reset_coordinates works fine for a single interval (i.e., the current ``slice`` function), but there are numerical precision issues with reset_coordinates that I don't feel like spending a day working through all the corner cases of.

So, that lead me to the proposal above: keep ``slice`` as it is, and offer a ``delete`` or ``mask_intervals`` or something function which takes a list of intervals and doesn't support ``reset_coordinates``.

hyanwong:
Just joining this conversation (I implemented the original `slice()`). Perhaps it's worth thinking of the 'reset coordinate' idea as trimming the left and right of a masked tree sequence. I could imagine defining `.rtrim()`, `.ltrim()`, and their joint `.trim()` on a ts (only the latter 2 would actually change the meaning of a position).

Then you could define `mask_intervals` as above, and redefine `slice()` to simply be a `mask_intervals` followed by `trim()`, with a warning if there are internal masked regions.

I haven't looks in detail about the proposal for 'masking' regions, but I do think there are two possibilities here: either there is a 'mask' defined, but the underlying edges stay the same (no data loss, simply data hiding), or the tree sequence is actually *lost* in the masked  intervals (i.e. edges are deleted or truncated). In the latter 'delete' case, there's the problem of what to do with edges that completely span the deleted region. If they are cut into two shorted edges, then this could lead to a large increase in # edges for small deleted regions. If they are kept, then the trees could look very weird in the deleted regions, and we'd need to make sure that the mask regions and the edge-defined regions never get out-of-sync.

Does that make sense?

jeromekelleher:
> Perhaps it's worth thinking of the 'reset coordinate' idea as trimming the left and right of a masked tree sequence. I could imagine defining .rtrim(), .ltrim(), and their joint .trim() on a ts (only the latter 2 would actually change the meaning of a position).

I think this is the right approach: let's take ``reset_coordinates`` out of these functions entirely, and make them separate functions. We can easily detect sections of the tree sequence where there are no edges, and we we can make a set of functions for alertering the coordinate space to remove these. So, we can follow up with other issues to track this.

Re masking/deleting, adding a permanent mask would be quite a bit of work here (where would we store it?), so I think deleting is the right approach for now. We completely remove any topology within the regions, including edges that span intervals.

hyanwong:
> We completely remove any topology within the regions, including edges that span intervals.

Right. We should note in the docs that this could easily end up bloating the number of edges, though (it's a bit like losing correlation between trees along the sequence, although we do retain the node numbering, I suppose)

hyanwong:
> I think deleting is the right approach for now.

I think if you are deleting, you should avoid the word 'mask' in the functions or their descriptions. To me, a mask is something you can layer on top, and then remove again, with no loss of data.

jeromekelleher:
> think if you are deleting, you should avoid the word 'mask' in the functions or their descriptions. To me, a mask is something you can layer on top, and then remove again, with no loss of data.

Agreed - we'll probably want to add the idea of a permanent mask at some point.

jeromekelleher:
@andrewkern - given that we can mask out the sequence data for now in the analysis repo and work ahead there, do you mind if we shelve this for a week or so? I don't want to rush through these changes and be saddled with something that's hard to maintain or causes lots of confusion in the long term.

petrelharp:
To me it sounded like when we talked through the ramifications that removing the bits of the tree sequence was going to be much easier than masking out bits of the sequence (just one line in the data generation rather than a bunch of different downstream pipelines). 

What's in a week - more time to think? I like our current proposal: implement `delete_intervals` that does not reset coordinates and also some `trim` functionality (but we do not depend on that). Are there concerns with this?

jeromekelleher:
> What's in a week - more time to think? I like our current proposal: implement delete_intervals that does not reset coordinates and also some trim functionality (but we do not depend on that). Are there concerns with this?

No, it's just I'm at SMBE this week so have limited time for GitHubbing. We'd need to cut a new alpha release as well to make this code available to stdpopsim, so I figured it was best to put it off for a little bit. 

If you want to push through the current proposal though, cutting a release is a 10 minute job.

petrelharp:
I'll fix this up soon; note to self: make it possible to include either the BED file name and/or the intervals in the provenance.

petrelharp:
Ok: the 1000 genomes mask has 109,666 distinct pieces (many of these are small, but 13,515 are still more than 1Kb). So, I wrote up a python version of the fast `dice`, that would scale to this many chunks. This would be straightforward to push over to C. However, I think it is also the *wrong* thing to do: cutting up our edges into 1e5 pieces is going to destroy the efficiency of the tree sequence, as @hyanwong said above.

So, rather than computing masked diversity using `dice`, I [used windowing in statistics computation](https://github.com/petrelharp/treestats_ms/blob/master/figures/plot_1kg_diversity.py) to do it. This was relatively painless and still very fast, at least for this dataset (seconds, 1000g, chr20).

So, here's my proposal: we do need functions to remove or keep big chunks of a tree sequence. But we shouldn't try to make these scale to lots and lots of chunks, and should implement this sort of masking at the output level: statistics, or genotypes. So, how about:

1. Keep `slice`, but renamed to `keep_slice`. 
2. Implement `delete_slices(starts, ends)`, which removes the bits between starts and ends, and does *not* do `reset_coordinates`.

Then, to deal with masking (this would take some work; to do separately):

3. Define a `Mask` object, which is a collection of intervals,
4. Let statistics computation take a `mask=` argument.
5. Let `genotype_matrix` and `vcf` take a `mask=` argument.

 Other notes:

a. The current implementation of `dice` needs to sort the tables after. If we want to worry about this we could implement the C version. I'll push my python version somewhere to this branch to see.
b. I think masking the trees is bad but masking the *sites* would be OK. So, we could provide a function that removes only the sites in lots of little chunks. But, I still vote not to do this, because it's harder to keep track of where this has been done; the proposal to have a Mask object is cleaner.

petrelharp:
The code is in `python/test_dice.py`. Sorry for the temporary clutter.

andrewkern:
I second the idea of defining a `Mask` class that would be a container for the intervals to be associated with statistics or the genotype_matrix

jeromekelleher:
I like all this @petrelharp, except I don't think we need an explicit Mask class --- it's just an (N, 2) numpy array, so it would be better to define it like this  and do the same thing for all intervals we take as input. So, ``delete_slices(intervals)`` takes a (N, 2) numpy array (or, more precisely, something that can be interpreted as a (N, 2) numpy array) describing N non-overlapping intervals.

petrelharp:
> I like all this @petrelharp, except I don't think we need an explicit Mask class --- it's just an (N, 2) numpy array, so it would be better to define it like this and do the same thing for all intervals we take as input. So, delete_slices(intervals) takes a (N, 2) numpy array (or, more precisely, something that can be interpreted as a (N, 2) numpy array) describing N non-overlapping intervals.

Oh, good point. Except that I think it should be `(2, N)` not `(N, 2)`, because numpy stores things rowwise, and we'll often want to do things with e.g. the vector of starts.

jeromekelleher:
> Oh, good point. Except that I think it should be (2, N) not (N, 2), because numpy stores things rowwise, and we'll often want to do things with e.g. the vector of starts.

It's a bit more intuitive to write ``[(start1, end1), (start2, end2), ...]`` though, isn't it? We can get the ``start`` and ``end`` arrays internally using numpy if we want to --- the numbers of intervals is never going to be big enough that this will matter performance wise.

hyanwong:
> * Keep `slice`, but renamed to `keep_slice`.
> * Implement `delete_slices(starts, ends)`,

If we are implementing a `trim()` function, why not simply have `slice(start, end)` (which can accept arrays of the same length for `start` and `end`, and does not change the coordinate system). For convenience, we could then simply rename the current "slice with reset" as something that clearly says what it does, such as `slice_and_trim()`?

petrelharp:
Sorry, @hyanwong - what is `trim`? Is that in a different issue that I've forgotten about?

hyanwong:
> Sorry, @hyanwong - what is `trim`? Is that in a different issue that I've forgotten about?

See above: https://github.com/tskit-dev/tskit/pull/261#issuecomment-513109946

Basically, `ltrim()` resets the zero point to the start of the trees on the left. Similarly `rtrim()` for the right, and `trim()` should do both. That seems reasonably self-explanatory to me?

jeromekelleher:
``trim`` etc are separate from this thread now @hyanwong: can you open an issue to track what they are for and what they do please? (I agree they are simple well defined ops, btw.)

jeromekelleher:
OK, I've updated this with my take:

1. Add ``keep_slices`` and ``delete_slices``, which both take a list of intervals as argument. Remove the ``reset_coordinates`` option entirely. It's pointless if we're implementing 'trim' etc.
2. Delete the ``slice`` method --- it's just ``keep_slices([[start, stop]])`` so doesn't seem like it's worth the hassle.
3. I've added some general interval handling functions which will be useful also for working with masks (I'm imagining we add a parameter ``mask=None`` to the stats functions which will use this infrastructure as well).

The tests are currently broken, as I've run out of time to fix them up.

Now that the semantics of ``keep_slices`` and ``delete_slices`` are clear, I'm tempted to rename them to ``keep(intervals)`` and ``delete(intervals)``. It's self-evident what they do from the signatures then, right? I'm not sure that the term ``slice`` is helping here, so at least i'd suggest renaming to ``keep_intervals(intervals)`` and ``delete_intervals(intervals)``, but then the ``_intervals`` suffix seems  redundant.

hyanwong:
Yes, sorry, I misunderstood (I thought that the suggested `keep` function also trimmed). I see now that `keep(intervals)` and `delete(intervals)` are simply complementary functions. I agree with dropping the `_slice` prefix. New issue for `trim()` opened at https://github.com/tskit-dev/tskit/issues/288.

Re: performing masked operations, is there a `jump()` or `skip()` function that allows the trees iterator to advance to a new genomic position a little further on? Or possibly a genome-coords equivalent to [`islice`](https://docs.python.org/2/library/itertools.html#itertools.islice)

jeromekelleher:
> Re: performing masked operations, is there a jump() or skip() function that allows the trees iterator to advance to a new genomic position a little further on? Or possibly a genome-coords equivalent to islice?

This is a great idea - we could add the ``mask=None`` argument to trees() to tell it to skip over trees in the mask.

jeromekelleher:
@petrelharp, @andrewkern, any objections to moving to ``delete(intervals)`` and ``keep(intervals)`` here? I'll finish this up ASAP if we're all happy with this direction.

petrelharp:
I vote for `delete_intervals( )` and `keep_intervals( )` since there's other things one could delete, etc.

hyanwong:
FWIW `delete_spans` is shorter to type, if you need a qualifier. Not sure what I prefer TBH.

petrelharp:
> FWIW delete_spans is shorter to type, if you need a qualifier. Not sure what I prefer TBH.

I'm good with that.

andrewkern:
> @petrelharp, @andrewkern, any objections to moving to `delete(intervals)` and `keep(intervals)` here? I'll finish this up ASAP if we're all happy with this direction.

happy with this. also like @hyanwong's suggestion of using `_spans` to disambiguate the object

jeromekelleher:
I'm going to go with ``delete_intervals(intervals)`` --- fewer words to explain!

hyanwong:
I'm coding up `trim()` and friends now. It seems like I should remove any sites that are in the trimmed regions. A function such as `remove_sites(self, site_ids, record_provenance=True)` seems a sensible addition to the tskit armoury, and might be useful for `delete_intervals` mightn't it (which is why I bring it up here)? I don't think this is already done is it? Any objections to my coding it up?

jeromekelleher:
> I'm coding up trim() and friends now. It seems like I should remove any sites that are in the trimmed regions. A function such as ts.remove_sites(self, site_ids, record_provenance=True) seems a sensible addition to the tskit armoury, and might be useful for delete_intervals mightn't it (which is why I bring it up here)? I don't think this is already done is it? Any objections to my coding it up

We're removing the sites here so I don't see any need for a remove_sites method right now.

jeromekelleher:
OK, hopefully this is ready for final review and merge.

jeromekelleher:
Nobody has complained, so I'm going to merge this.

petrelharp:
Summary: we now have

- `TableCollection.delete_intervals( )`
- `TableCollection.copy( )` :100: 

and some useful things in `util`: `pack_bytes`, `unpack_bytes`, `pack_strings`, `unpack_strings`, and a few interval operations.

:yay:!

jeromekelleher:
Also, TableCollection.keep_intervals(). I move the packing functions into util as it seemed like the right home for them.

Issue #262, opened 2019-07-16
Rename map_ancestors to map_to_ancestors?


jeromekelleher:
With the ``map_mutations`` method about to be added in #125, I wonder if the ``map_ancestors`` function is now a little confusingly named. Maybe ``map_to_ancestors`` would be a little more descriptive? Or perhaps ``trace_ancestors``?

Don't worry @gtsambos, if we decide to change the name I can make the changes!

gtsambos:
Either of those alternatives are fine by me! :)

petrelharp:
Another idea: `find_ancestors`?  When you write the paper, @gtsambos, are you going to want to call it "mapping"?

jeromekelleher:
We're providing the ancestors as input though, so 'find' seems slightly off.

gtsambos:
I'm not especially wedded to the word 'map'. I think 'find' is fine too, given that we don't a priori know which of/whether any of our provided `ancestors` are actually ancestral to the sample. Perhaps `find_ancestral_relationships` is more to the point, but a bit too long.

jeromekelleher:
How about ``link_ancestors``? We're creating a set of direct links from the samples to the ancestors, which we output via a set of edges.

gtsambos:
I like `link_ancestors`!

petrelharp:
What do you call it when you talk about it? Would you say "I have developed a method to link samples to their ancestors"?

I'd go for `link_to_ancestors`.

petrelharp:
Although note that this function has other uses: if passed no ancestors, then it should find all relationships between samples (right?).

gtsambos:
> Would you say "I have developed a method to link samples to their ancestors"?

I think I would feel comfortable with that! 😉 'Extract' is the word I use most often, to convey the fact that the information I'm finding is already in the tree sequence object I'm analysing. But I think 'link' conveys that idea well enough. 

> Although note that this function has other uses: if passed no ancestors, then it should find all relationships between samples (right?).

Nope, the function will fail if it is passed no ancestors. If you want to find relationships between all samples, you should just pass `samples` and `ancestors` the same list of nodes.

jeromekelleher:
OK, sounds like we have agreement. @gtsambos, can you make a call between ``link_ancestors`` and ``link_to_ancestors`` please? I'm happy with either.

gtsambos:
Let's go `link_ancestors` because it's shorter.

jeromekelleher:
[...]

PR #266, opened 2019-07-18
Updated docs for map_ancestors.

gtsambos:
This pull request continues the conversation about documentation for the `map_ancestors` feature developed in [this pull request](https://github.com/tskit-dev/tskit/pull/175).

cc: @petrelharp, @jeromekelleher

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1665,18 +1665,22 @@ class TableCollection(object):
 
     def map_ancestors(self, samples, ancestors):
         """
-<<<<<<< HEAD
         Returns an :class:`EdgeTable` instance describing a subset of the genealogical
         relationships between the nodes in``samples`` and ``ancestors``.
-=======
-        Returns an :class:`EdgeTable` instance describing how ``samples`` are descended
-        from ``ancestors``.
->>>>>>> 4526b9f... Updated docs for map_ancestors.
         Each row ``parent, child, left, right`` in the output table indicates that
         ``child`` has inherited the segment ``[left, right)`` from ``parent`` more
         recently than from any other node in these lists.
-        The following table shows which ``parent->child`` pairs will be shown in the output
-        of ``map_ancestors``:
+
+        In particular, suppose ``samples`` is a list of nodes such that ``time`` is 0
+        for each node, and ``ancestors`` is a list of nodes such that ``time`` is
+        greater than 0.0 for each node. Then each row of the output table will show
+        an interval ``[left, right)`` over which a node in ``samples`` has inherited
+        most recently from a node in ``ancestors``, or an interval over which one of
+        these ``ancestors`` has inherited most recently from another node in
+        ``ancestors``.
+
+        The following table shows which ``parent->child`` pairs will be shown in the
+        output of ``map_ancestors``:
 
         |Type of relationship      | Shown in output of ``map_ancestors``           |
         | ------------------------ | ---------------------------------------------- |
@@ -1685,11 +1689,6 @@ class TableCollection(object):
         |``sample1->sample2``      | Always                                         |
         |``sample->ancestor``      | Only if ``ancestor`` has a relevant descendant |
 
-        In other words, the tree constructed at a position with the output edge table has
-        the property that the parent of any node in ``samples``, if there is one, is the
-        most recent ancestor of that node at that position out of all possible ancestors
-        in ``ancestors``, and the same is true for any node in ``ancestors`` that is
-        ancestral to any of ``samples``.
         The difference between ``samples`` and ``ancestors`` is that information about
         the ancestors of a node in ``ancestors`` will only be retained if it also has a
         relevant descendant, while information about the ancestors of a node in
@@ -1698,12 +1697,12 @@ class TableCollection(object):
         of the inputted tree sequence.
 
         The supplied nodes must be non-empty lists of the node IDs in the tree sequence.
-        Other than this, the procedure is flexible; overlap between the lists of samples
-        and ancestors is permitted, and the nodes need not be ordered by time.
-        For instance, the same list of node IDs can be inputted to both ``samples`` and
-        ``ancestors`` to find all genealogical relationships within this list of nodes.
-        If none of the nodes in `ancestors` are ancestral to `samples` anywhere in
-        the tree sequence, an empty table will be returned.
+        The lists of ``samples`` and ``ancestors`` may overlap.
+        If ``samples`` and ``ancestors`` are the same list of nodes, the function will
+        find all genealogical relationships within this list.
+
+        If none of the nodes in ``ancestors`` or ``samples`` are ancestral to ``samples``
+        anywhere in the tree sequence, an empty table will be returned.
 
         :param list[int] samples: A list of node IDs to retain as samples.
         :param list[int] ancestors: A list of node IDs to use as ancestors.
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1665,25 +1665,43 @@ class TableCollection(object):
 
     def map_ancestors(self, samples, ancestors):
         """
+<<<<<<< HEAD
+        Returns an :class:`EdgeTable` instance describing a subset of the genealogical
+        relationships between the nodes in``samples`` and ``ancestors``.
+=======
         Returns an :class:`EdgeTable` instance describing how ``samples`` are descended
-        from ``ancestors`` (and how ``ancestors`` are descended from each other).
+        from ``ancestors``.
+>>>>>>> 4526b9f... Updated docs for map_ancestors.
         Each row ``parent, child, left, right`` in the output table indicates that
-        ``child`` (which is either in ``samples`` or in ``ancestors``) has inherited the
-        segment ``[left, right)`` from ``parent`` (which is in ``ancestors``) more
-        recently than from any other node in ``ancestors``. Every segment of the
-        genome on which each of ``samples`` has inherited from any of ``ancestors``
-        is described in this way, as well as any segment of the genomes of ``ancestors``
-        that has been inherited by any of ``samples``.  The node IDs in ``parent`` and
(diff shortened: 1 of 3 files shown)


gtsambos:
See also [this issue](https://github.com/tskit-dev/tskit/issues/269).

python/tskit/tables.py
```diff
@@ -1666,22 +1666,28 @@ def simplify(
     def map_ancestors(self, samples, ancestors):
         """
         Returns an :class:`EdgeTable` instance describing how ``samples`` are descended
-        from ``ancestors`` (and how ``ancestors`` are descended from each other).
+        from ``ancestors`` (and how these ``ancestors`` are descended from each other
```
jeromekelleher: This sentence doesn't quite parse for me.

python/tskit/tables.py
```diff
@@ -1666,22 +1666,33 @@ def simplify(
...
+        |``ancestor->sample``      | Always                                         |
+        |``ancestor1->ancestor2``  | Only if ``ancestor2`` has a relevant descendant|
+        |``sample1->sample2``      | Always                                         |
+        |``sample->ancestor``      | Only if ``ancestor`` has a relevant descendant |
+
+        In other words, the tree constructed at a position with the output edge table has
+        the property that the parent of any node in ``samples``, if there is one, is the
+        most recent ancestor of that node at that position out of all possible ancestors
+        in ``ancestors``, and the same is true for any node in ``ancestors`` that is
         ancestral to any of ``samples``.
```
petrelharp: I think that the table contradicts the text here? The text says that the parent of any sample is one of `ancestors`; but the table says that one sample can be a parent to another sample? [The tests](https://github.com/tskit-dev/tskit/blob/7f24d6adcbdfe8de5510c702895bfd55293da60f/python/tests/test_topology.py#L3695) suggest that *only* `ancestors` can appear in the `parents` column; so shouldn't the 3rd and 4th lines of the table be "Never"?

gtsambos: (sorry for the delay on this - have been busy with other things)

> The text says that the parent of any sample is one of ancestors; but the table says that one sample can be a parent to another sample?

I've changed the first sentence back to something quite vague now - I'm not sure that there is a clear way to describe everything that this function will and won't do in particular cases in a single, easy-to-parse sentence, so I'm hoping users will look at the table for this now.

> The tests suggest that only ancestors can appear in the parents column; 

Ah, thanks for pointing this out -- so this line of code is in the `verify` function, which was initially only being used to test situations where the samples were in the final generation of the simulation (ie. the last two rows of the table weren't being tested here). I have now changed the function, and expanded the tests using it so that all rows of the table are being tested.

> shouldn't the 3rd and 4th lines of the table be "Never"?

nup, see above. Btw, there are 'bespoke' tests covering the behaviour described in each line of the table.

1. [here](https://github.com/tskit-dev/tskit/blob/e55fe76a113ff13914bfbea900e889a862bec20b/python/tests/test_topology.py#L3672) (for example)
2. [here](https://github.com/tskit-dev/tskit/pull/266/files#diff-5941ccebfc6d10311b8d79e8bc986cffR3632)
3. [here](https://github.com/tskit-dev/tskit/pull/266/files#diff-5941ccebfc6d10311b8d79e8bc986cffR3652)
4. [here](https://github.com/tskit-dev/tskit/pull/266/files#diff-5941ccebfc6d10311b8d79e8bc986cffR3642)

gtsambos:
Btw, I feel that the description we have here may now getting a little unwieldy, to the point of obscuring what it might be most useful for. 

If it's okay, I might get rid of this [paragraph](https://github.com/tskit-dev/tskit/pull/266/files#diff-a89cffb041b1b7c33e7be6b595f785a3R1683-R1687)? I feel like it's just trying to put info from the table into a sentence, but it's kindof confusing and hard to interpret.

jeromekelleher:
Looks like you've done something weird with the history here @gtsambos (have a look at the network picture) --- merge is your enemy here, you want to use rebase every time.

petrelharp:
Ok! Let's get this sorted now. This is a super helpful function, but it's going to be much more useful if it's unambiguously documented.

You're right, the documentation *is* unwieldy. But I'm sure we can get something precise and concise, it's not *that* complicated. The table is not sufficient: what exactly it means is ambiguous (although it's totally helpful!).  Here, I'll try. [... tries...]

Ok, I've found that the text and the table still contradict each other. Over [here](https://github.com/tskit-dev/tskit/issues/262#issuecomment-512864290) you said that the function would fail if there were no ancestors. But, the table says that relationships between samples are always shown? I did a few small experiments, and it turns out that you can get all relationships between ``samples`` with *any* valid nodes passed in to `ancestors` whatsoever.  Passing in `ancestors = samples` gets you the same output as passing in `ancestors=[u]` where `u` is some node not ancestral to any of the samples. So, it looks like everything should work just fine if we allowed no ancestors at all. I can't quite tell where it throws the error when you try this, though: it looks to be around [here](https://github.com/tskit-dev/tskit/blob/master/python/_tskitmodule.c#L5522).

Now that I think I know exactly what it does, how's this (modified assuming that we can allow an empty argument to `ancestors`):

```
        Returns an :class:`EdgeTable` instance describing a subset of the genealogical
        relationships between the nodes in``samples`` and ``ancestors``.
        Each row ``parent, child, left, right`` in the output table indicates that
        ``child`` has inherited the segment ``[left, right)`` from ``parent`` more
        recently than from any other node in either of these lists, and only portions of the
        tree sequence ancestral to ``samples`` is described.
        The following table shows which ``parent->child`` pairs will be described in the output:

        |Type of relationship      | Shown in output of ``map_ancestors``           |
        | ------------------------ | ---------------------------------------------- |
        |``ancestor->sample``      | Always                                         |
        |``ancestor1->ancestor2``  | Only if ``ancestor2`` is ancestral to ``samples``|
        |``sample1->sample2``      | Always                                         |
        |``sample->ancestor``      | Only if ``ancestor`` is ancestral to ``samples``|

        The difference between ``samples`` and ``ancestors`` is that information about
        the ancestors of a node in ``ancestors`` will only be retained if it also has a
        relevant descendant, while information about the ancestors of a node in
        ``samples`` will always be retained.
        The node IDs in ``parent`` and ``child`` refer to the IDs in the node table
        of the input tree sequence.

        The lists of ``samples`` and ``ancestors`` may overlap. For instance, if ``ancestors``
        is empty, this will find all genealogical relationships between the ``samples``.
        If none of the nodes in either list are ancestral to `samples` anywhere in
        the tree sequence, an empty table will be returned.
```

> changed verify

This looks good! I think it verifies what we say here.

Almost there!

gtsambos:
> Almost there!

🎉 

> Over here you said that the function would fail if there were no ancestors. But, the table says that relationships between samples are always shown?

> So, it looks like everything should work just fine if we allowed no ancestors at all. I can't quite tell where it throws the error when you try this, though: it looks to be around here.

I don't think this is a contradiction, but perhaps I could have been clearer - I meant what you found by trying things out, and what is in the last paragraph of the docs for this function already -- that if `ancestors` is an empty list, the function will return an error. This happens because of [this condition here](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L4136). Initially I did allow `ancestors` to be an empty list (and actually, in earlier commits there were tests for this behaviour).

The requirement that `ancestors` is non-empty was a late change that I made on the advice of @jeromekelleher, because otherwise you can run into problems when you try to deallocate memory from the empty list of ancestors. I'm happy to look into modifying this if you think it is going to be confusing for the user, though maybe this is something to consider in a separate pull request. It wouldn't be hard to change it so that if the user passes an empty list into `ancestors`, `map_ancestors` internally copies the nodes from `samples` into `ancestors`.

For the moment, I think just making sure that the docs accurately describe current behaviour should be the main focus.

It's totally fine if the user's supplied list of `ancestors` doesn't happen to be ancestral to their supplied list of `samples` - see [this test](https://github.com/tskit-dev/tskit/pull/266/files#diff-5941ccebfc6d10311b8d79e8bc986cffR3586).

> You're right, the documentation is unwieldy. But I'm sure we can get something precise and concise, it's not that complicated. 

I sure hope so 😆 I guess part of why this description has been a bit painful for me to write is because in the use case I was designing this function for, `samples` are always from the most recent generation and `ancestors` never are. I expect that this is going to be the case for many users too, so I've added an extra sentence explicitly describing this particularly important use case upfront. Let me know what you think!

> Looks like you've done something weird with the history here

yep, oops  😅💀 I think this is fixed now!

gtsambos:
> yep, oops 😅💀 I think this is fixed now!

actually, looking at the graph, I think I've somehow done something even weirder now :/ I will look into it this afternoon. FYI I haven't made changes to anything other than `tables.py` and `test_topology.py`!

gtsambos:
Also, @jeromekelleher, is there some kind of command-line test I can do to check that the docstring is syntactically okay? I'm not very good at reStructured text yet so I seem to making a lot of errors that are hard for me to pick up by eye.

jeromekelleher:
> Also, @jeromekelleher, is there some kind of command-line test I can do to check that the docstring is syntactically okay? I'm not very good at reStructured text yet so I seem to making a lot of errors that are hard for me to pick up by eye.

If you go into the `docs`` directory and run ``make``, it'll build the docs locally and you can see how it looks there. 

> actually, looking at the graph, I think I've somehow done something even weirder now :/ I will look into it this afternoon. FYI I haven't made changes to anything other than tables.py and test_topology.py!

Yes, looks like you've gotten into a mess here! Perhaps [this](https://stdpopsim.readthedocs.io/en/latest/development.html#when-rebasing-goes-wrong) will help?

gtsambos:
Hi @jeromekelleher and @petrelharp, whatever I did to this branch looks well beyond my git surgery abilities 💀 💀 💀 so I just took the nuclear option and just made a new pull request for these changes over [here](https://github.com/tskit-dev/tskit/pull/284).

PR #267, opened 2019-07-18
Properly export constants

hyanwong:
At the moment `tskit.NULL` comes because we do `from _tskit import NULL` within trees.py, then in `__init.py__` we do `from tskit.trees import *`. This indirect import seems nasty to me. So I've changed this so that all constants are explicitly exported via a line in `__init.py__` instead.


jeromekelleher:
LGTM, thanks. will merge after #268.

PR #254, opened 2019-07-03
High-level doc structure for available stats.

jeromekelleher:
Adds a structured index of stats that we have implemented to the stats docs page. The structure is tentative - I think you're in a much better position to put some sensible structure and describe the high-level properties of the stats @petrelharp.

Also, I didn't address this point:

> Also distinguish those statistics for which E[site] = branch, from the 'derived' ones (currently just Fst and Tajima's D).

What do you think @petrelharp? If you want to update this before merging, you should be able to push commits to this branch. Otherwise, we can merge and create some issues to track the changes that need to be made.

PR #271, opened 2019-07-26
moved stats functions to stats.rst

petrelharp:
Here's the sphinx magic to move that over.  Other things that we could maybe do now:

1. I deprecated `pairwise_diversity` but should probably move that over to `stats.rst` also?
2. I also think we should move `genealogical_nearest_neighbors` and `mean_descendants` over to `stats.rst`; perhaps in a separate section? There's definately overlap between the topics; maybe the page gets called something like "Statistics and other genealogical descriptors"?
3. `mean_descendants` could be easily implemented as a node statistic, and we'd have use for the corresponding site and branch statistic, also. Maybe the most natural is that `mode="site", windows="sites", polarised=True` gives the derived allele frequency at each site. Given this, maybe a better name is `descendant_frequency` or just `frequency`?

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -26,10 +26,11 @@ Top level-classes
 
 
 .. autoclass:: tskit.TreeSequence()
-    :members:
+   :members:
+   :exclude-members: general_stat, sample_count_stat, diversity, divergence, allele_frequency_spectrum, segregating_sites, Fst, Tajimas_D, f2, f3, f4, Y1, Y2, Y3, trait_covariance, trait_correlation, trait_regression, parse_windows
 
 .. autoclass:: tskit.Tree
-    :members:
+   :members:
 
 
 +++++++++
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -235,23 +235,11 @@ That said, you should **not** rely on the specific behavior of whether ``0`` or 
 for "empty" cases like these: it is subject to change.
 
 
-.. Commenting these out for now as they are duplicates of the methods in the TreeSequence
-   and sphinx is unhappy.
+********************
+Statistics functions
+********************
 
-.. ********************
-.. Statistics functions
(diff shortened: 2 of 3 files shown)


petrelharp:
Hm: we could actually implement `mean_descendants` using `segregating_sites(..., mode="node", polarised=True)` if we had the `polarised` option.

petrelharp:
Ping @hyanwong @gtsambos about where to discuss genealogical summaries (perhaps in stats.rst?).

jeromekelleher:
I'm not sure I agree with completely removing the stats from the docs from the TreeSequence docs @petrelharp - they are still methods of the tree sequence, and people might expect to be able to search for them there.

Have you seen my alternative proposal for reorganising the stats docs here in #254? This tries to categorise the stats and gives links back to the technical documentation. I think this is good because it gives a high-level overview of what's available, and how they relate to each other.

jeromekelleher:
> Hm: we could actually implement mean_descendants using segregating_sites(..., mode="node", polarised=True) if we had the polarised option.

I'm fine with deprecating or even removing ``mean_descendants`` --- I don't think anyone is using it outside the BDI, and we should move over to the node based stats. It probably hasn't been included in a released version of tskit, so there's not much harm in removing it I think.

petrelharp:
>  Have you seen my alternative proposal for reorganising the stats docs here in #254? 

Whoops; sorry - missed that. It'd be nice to organize the many TreeSequence methods according to type, but maybe that's something that can be done in a cheatsheet-type page. Ok, I like that better.

> deprecating or even removing `mean_descendants`

I think we should definately keep the functionality! Just make it a node stat and rename it.

petrelharp:
Ok, I'll move my changes over to #254 and open a new issue for `mean_descendants`.

jeromekelleher:
Shall we close this PR then?

PR #272, opened 2019-07-26
Missing genotypes in variants and genotype_matrix

### LINKED ISSUE #270: Ancestral states when sample data is missing
Something I just realised, after implementing a unit test for tsinfer. If we have samples that are missing a left hand or right hand chunk of information, then my recent push (https://github.com/tskit-dev/tsinfer/pull/169) allows us to build a decent tree sequence, and a samples that has a particular region missing will simply have its node number unconnected to any other node in that region of the tree sequence (it will just be an unconnected point when plotted). But when we generate a haplotype from this tree sequence, there will be zeroes not tskit.MISSING_DATA for the missing regions. That’s because the ancestral state is zero in these regions, and since they aren’t linked by an edge to anywhere, they just get the ancestral state.

There are a few reasonable ways around this. One way (1) is to say that any completely unconnected nodes should always get a missing data flag. Another (2) is to say that for these sites, the ancestral state is tskit.MISSING_DATA, then place a mutation to ‘0’ above the root of the main tree. Yet another (3), which is more generic but has a wider impact is to set the ancestral state to tskit.MISSING_DATA for *all* sites, and always require a mutation to 0 to be placed above the main tree root.

I see this as a general problem that if we have a position in the tree sequence where there are multiple roots, there might be different ancestral states for each root.

jeromekelleher:
Here's the basic proposal:

- Add a flag ``isolated_samples_missing`` (or something) to ``variants``/``genotype_matrix``. 
- When this is true, for each site **after** we have generated all the genotypes in the usual way, update them to mark any samples that are 'isolated' as missing data (-1). An isolated sample is defined as a root that has no children, which I think covers all the cases where we would actually want to consider a sample as missing. (Note to self: there is a complication here when we are generating genotypes for non-sample nodes, in that we'll need to check if the root in question is one of the input list of nodes as well.)
- If we find at least one isolated root, we add a value to the end of the list of alleles to indicate this. In python this will be None and an C it'll be NULL. This will allow users to quickly check if there's any missing data, without needing to look at the genotypes.

There's some work to be done here, but I thought I'd share the basic idea first to see what people think before going any further with it. Pinging @hyanwong and @petrelharp for input.

Closes #270


hyanwong:
This seems like a reasonable way to deal with the issue to me. The only down side is that this way, the missingness criterion (i.e. the value to use for `isolated_samples_missing`) is not encoded in the tree sequence itself, but has to be recorded elsewhere. So e.g. when round-tripping data from a SampleData file, through tsinfer, and back to a variant matrix, I will have to remember whether the file contained missing data (and if so, ensure that  `isolated_samples_missing=True` for all potential variants).

jeromekelleher:
I think this is OK - for tsinfer derived tree sequences where we expect to deal with missing data we always set ``isolated_samples_missing=True``. If there is no missing data, then there'll be no missing data in the output genotypes.  

If client code doesn't know how to deal with missing data, by default any missing states are imputed as the ancestral state (which, I guess, is reasonable).

So, the missingness criteria *is* encoded in the tree sequence (samples with no edges), but how you deal with that missingness is determined by the user. If the user wants to know explicitly that the sites are missing when generating genotypes, they set ``isolated_samples_missing=True``. Otherwise, by default, we impute missing sites for samples marked as missing data to be the ancestral state. 

There's a different conversation to be had about how we deal with short regions of missing data *within* tsinfer (should we break edges and output them as missing, or should we impute the states using LS?), but that's something to discuss in the context of inference rather than tskit.

jeromekelleher:
Or, we could make this the default behaviour and add a flag ``impute_missing`` to fill in the missing data as ancestral state. I don't think it'll affect anyone to change the default behaviour here as this is pretty esoteric to people not developing on tsinfer. On reflection, this is probably the right thing to do - users should have to make a conscious decision to impute the missing data.

@petrelharp, any thoughts on this? I actually need this for something I'm working on now, so I'd like to push it through ASAP.

jeromekelleher:
... and of course this has implications for the stats. Currently most site based tests are broken as the default behaviour of ``genotype_matrix`` has changed.

petrelharp:
I think that making "isolated nodes = missing" is probably the right thing to do. I can think of all sorts of potential questions (what if there are *two* sister nodes in their own tree? why are they not missing?) but not questions that are likely to come up in practice. The case of ancestral nodes that aren't specified in the current tree is *much* more important, and these should obviously be missing.

(Note: the more obvious method of just sticking a "missing" mutation above the relevant samples should also work - because just because someone's genotype is unknown at a site doesn't mean we don't have a good guess at their location in the genealogy... but this we already have, right?)

hyanwong:
I agree that having isolated nodes as missing seems the most sensible default, FWIW

jeromekelleher:
Notes to self, TODO

- Add some specific tests for behaviour in the new module, i.e. haplotypes should raise exceptions in specific cases
- document changes to haplotypes, variants, genotype_matrix
- Add a section in the documentation laying out the rationale and how it works.

jeromekelleher:
I think this is ready to merge - final reviews anyone?

docs/data-model.rst
```diff
@@ -852,6 +852,34 @@ they occur, because these operations have the following properties:
 7. ``compute_mutation_parents`` fills in the ``parent`` information by using
     property (1).
 
+.. _sec_data_model_missing_data:
+
+************
+Missing data
+************
+
```
petrelharp: > datA

Isn't it "datá"?

petrelharp:
Looks good to me! I got confused for a bit about why we can't also denote missingness by setting the derived state of a mutation to "missing", but after thinking about it more, I like this approach.

jeromekelleher:
OK, merging.

PR #274, opened 2019-07-29
AFS and branch length cleanups

jeromekelleher:
This  should clears up any remaining issues with the AFS. It does feel weird not to do the same thing for the site AFS, but perhaps the same issues don't occur?  Any thoughts @petrelaharp?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -3759,11 +3759,10 @@ class TreeSequence(object):
         """
         Computes the allele frequency spectrum (AFS) in windows across the genome for
         with respect to the specified ``sample_sets``. See :ref:`sec_general_stats` for
-        details of ``windows``, ``mode``, ``span_normalise`` and ``polarised``.
+        details of ``windows``, ``mode``, ``span_normalise`` and ``polarised``,
+        and see :ref:`sec_tutorial_afs` for examples of how to use this method.
 
-        Please see the :ref:`sec_tutorial_afs` for examples of how to use this method.
-
-        Similar to other windowed stats, the first dimension in the returned array is
+        Similar to other windowed stats, the first dimension in the returned array
         corresponds to windows, such that ``result[i]`` is the AFS in the ith
         window. The AFS in each window is a k-dimensional numpy array, where k is
         the number of input sample sets, such that ``result[i, j0, j1, ...]`` is the
@@ -3775,13 +3774,14 @@ class TreeSequence(object):
         If a single sample set is specified, the allele frequency spectrum within
         this set is returned, such that ``afs[j]`` is the value associated with
         frequency ``j``. Thus, singletons are counted in ``afs[1]``, doubletons in
-        ``afs[2]``, and so on.
+        ``afs[2]``, and so on. The zeroth entry counts alleles or branches not
+        seen in the samples but that are polymorphic among the rest of the samples
+        of the tree sequence; likewise, the last entry counts alleles fixed in
+        the sample set but polymorphic in the entire set of samples. Please see
+        the :ref:`sec_tutorial_afs_zeroth_entry` for an illustration.
 
         .. warning:: Please note that singletons are **not** counted in the initial
-            entry in each AFS array (i.e., ``afs[0]``), but in ``afs[1]``. The
-            zeroth entry in the AFS can usually be omitted, but is significant
-            when subsets of all samples are provided as input. Please see
-            the :ref:`sec_tutorial_afs_zeroth_entry` for an illustration.
+            entry in each AFS array (i.e., ``afs[0]``), but in ``afs[1]``.
 
         If ``sample_sets`` is None (the default), the allele frequency spectrum
         for all samples in the tree sequence is returned.
@@ -3813,10 +3813,6 @@ class TreeSequence(object):
 
         "node"
             Not supported for this method (raises a ValueError).
-
-        The method does *not* currently count the number of fixed alleles
-        (for "site") or the length of branches above the root (for "branch"),
-        so the final entry is always zero.
 
         :param list sample_sets: A list of lists of Node IDs, specifying the
             groups of samples to compute the joint allele frequency
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -3789,14 +3789,15 @@ class TreeSequence(object):
         If more than one sample set is specified, the **joint** allele frequency
         spectrum within windows is returned. For example, if we set
         ``sample_sets = [S0, S1]``, then afs[1, 2] counts the number of sites that
-        at singletons within S0 and doubletons in S1.
+        at singletons within S0 and doubletons in S1. The dimensions of the
+        output array will be ``[num_windows] + [1 + len(S) for S in sample_sets]``.
 
-        If ``polarised`` is False (the default) the AFS will be *folded* such that
-        a frequency ``c`` is mapped to ``min(c, len(S) - c)``, where ``S`` is the sample
-        set in question. The dimensions of the output array will therefore be
-        ``[num_windows] + [1 + len(S) // 2 for S in sample_sets]``.
-        Otherwise, the dimensions of the output array will be
-        ``[num_windows] + [1 + len(S) for S in sample_sets]``.
+        If ``polarised`` is False (the default) the AFS will be *folded*, so that
+        the counts do not depend on knowing which allele is ancestral. If folded,
+        the frequency spectrum for a single sample set ``S`` has ``afs[j] = 0`` for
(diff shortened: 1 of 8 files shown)


petrelharp:
Thanks; I think there's two things still though:
1. We should do the same thing for site and branch.
2. We're still folding wrongly (discussed over [here](https://github.com/tskit-dev/tskit/issues/203#issuecomment-513871058) - there's a proposed solution there.)

petrelharp:
To be clear, the proposed solution is, when folding, to output an array with the *same* dimensions as unfolded, but with half the entries zero. My proposal is to fold so that the counts given are of the *minor* allele; so the algorithm would check if `k1+k2+...+km > (n1+n2+...+nm)/2`, and if so, set `AFS[n1-k1,n2-k2,...,nm-km] += AFS[k1,k2,...,km]` and then `AFS[n1-k1,n2-k2,...,nm-km] = 0`. Sorry for not figuring this out earlier.

jeromekelleher:
Sounds good @petrelharp. Keeping the dimensions the same is a major win in my book. I'll code this up and get back to you.

jeromekelleher:
I've updated the semantics of the site AFS here @petrelharp, would you mind taking a look please? The basic idea is not to include anything that's monomorphic across all samples. I've not pushed the changes through the C yet.

This brings up (yet another) subtle issue. In the branch AFS, we're excluding all branches that are ancestral to 0 samples, but including all branches ancestral to n samples (which is inconsistent with the site definition and to the ``segregating_sites`` function. Should we change this to also remove these branches that are ancestral to all samples?

I'm close to AFS'd out here - these corner cases are never ending!

jeromekelleher:
@petrelharp --- see the last commit: is this what you have in mind for folding? (I'll push through the rest of the changes, I just wanted to make sure I'm not on the wrong path first).)

petrelharp:
Er, no, that's not quite right. And that was extremely annoying to figure out, but I got it.
```
import numpy as np
import itertools

def fold(x, dims):
    """
    Folds the specified coordinates.
    """
    x = np.array(x)
    dims = np.array(dims)
    k = len(dims)
    n = np.sum(dims - 1) / 2
    s = np.sum(x)
    while s == n and k >= 0:
        k -= 1
        n -= (dims[k] - 1)/2
        s -= x[k]
    if s > n:
        x = dims - 1 - x
    assert np.all(x >= 0)
    return tuple(x)


def foldit(A):
    O = np.zeros(A.shape)
    dims = A.shape
    inds = [range(k) for k in dims]
    for ij in itertools.product(*inds):
        nij = fold(ij, dims)
        O[nij] += A[ij]
    return O


A = np.arange(12)
Af = np.array([11., 11., 11., 11., 11., 11.,  0.,  0.,  0.,  0.,  0.,  0.])
B = A.copy().reshape(3,4)
Bf = np.array([[11., 11., 11.,  0.],
               [11., 11.,  0.,  0.],
               [11.,  0.,  0.,  0.]])
C = A.copy().reshape(3,2,2)
Cf = np.array([[[11., 11.],
                [11., 11.]],
               [[11., 11.],
                [ 0.,  0.]],
               [[ 0.,  0.],
                [ 0.,  0.]]])
D = np.arange(9).reshape((3,3))
Df = np.array([[8., 8., 8.],
               [8., 4., 0.],
               [0., 0., 0.]])
E = np.arange(9)
Ef = np.array([8., 8., 8., 8., 4., 0., 0., 0., 0.])

assert np.all(foldit(A) == Af)
assert np.all(foldit(B) == Bf)
assert np.all(foldit(C) == Cf)
assert np.all(foldit(D) == Df)
assert np.all(foldit(E) == Ef)
```

jeromekelleher:
Note to self: need to do this after folding is sorted:

> This brings up (yet another) subtle issue. In the branch AFS, we're excluding all branches that are ancestral to 0 samples, but including all branches ancestral to n samples (which is inconsistent with the site definition and to the segregating_sites function. Should we change this to also remove these branches that are ancestral to all samples?

petrelharp:
>   This brings up (yet another) subtle issue. In the branch AFS, we're excluding all branches that are ancestral to 0 samples, but including all branches ancestral to n samples (which is inconsistent with the site definition and to the segregating_sites function. Should we change this to also remove these branches that are ancestral to all samples?

We should definately do the same thing for both branch and site. My inclination would be to *include* the branches/mutations ancestral to all samples in both (otherwise the last entry in the AFS is always 0?). Number of fixations seems like a nice thing to compute if they're actually in the tree sequence.

petrelharp:
> I'm close to AFS'd out here

seriously.

jeromekelleher:
> seriously.

Yah, what a can of worms...

jeromekelleher:
OK, that's it, I am done with the AFS (tempted rename it the FFS). As far as I know, site and branch are consistent now and should deal gracefully with arbitrarily weird stuff.

> We should definately do the same thing for both branch and site. My inclination would be to include the branches/mutations ancestral to all samples in both (otherwise the last entry in the AFS is always 0?). Number of fixations seems like a nice thing to compute if they're actually in the tree sequence.

I made this change @petrelharp, and some annoying corner cases bubbled up, which significantly reduced my will to live. So, I reverted and pushed the change through where we only update if 0 < all_samples_count < n in both site and branch AFS. This is, at least, consistent with the definition of segregating sites. 

Feel free to change it if you like though, I really don't have a strong opinion. It needs documenting now, but this will have to wait for another day I think.

jeromekelleher:
So yeah, I guess we merge this much and maybe get Konrad to try it out and see if he likes your folding mechanism? I can push out another alpha to make the code available.

petrelharp:
This is great. I agree, let's merge it. I've udpated the documentation. I would like in principle to get around to putting the fixed sites in; we'll see.

Hm, one thing: I thought we agreed to set the default to `polarised=True` for the AFS?

petrelharp:
Hm, that Travis error might be because I erroneously pushed to a new branch called "pull/274/head", and then deleted it, on the repository. :oh-no: :whoops-face:

python/tskit/trees.py
```diff
@@ -3812,6 +3813,10 @@ def allele_frequency_spectrum(
         "node"
             Not supported for this method (raises a ValueError).
 
+        The method does *not* currently count the number of fixed alleles
```
jeromekelleher: This is overstating it:
```python
ts = msprime.simulate(6, random_seed=2)
print(ts.draw_text())
afs = ts.allele_frequency_spectrum([[0, 2]], mode="branch", polarised=True)                                                  
print(afs)
```
gives

```
              10      ┊  
    ┊   ┏━━┻━━┓   ┊  
1.06┊   ┃     9   ┊  
    ┊   ┃   ┏━┻┓  ┊  
0.80┊   8   ┃  ┃  ┊  
    ┊ ┏━┻┓  ┃  ┃  ┊  
0.61┊ ┃  7  ┃  ┃  ┊  
    ┊ ┃ ┏┻┓ ┃  ┃  ┊  
0.08┊ ┃ ┃ ┃ ┃  6  ┊  
    ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊  
0.00┊ 3 0 2 1 4 5 ┊  
  0.00          1.00 

[[3.45905848 1.22532517 0.90607714]]

```
It's only stuff above the MRCA of **all** samples that's ignored, so the final entry will have values when the sample sets are a subset of all samples.

jeromekelleher:
> This is great. I agree, let's merge it. I've udpated the documentation. I would like in principle to get around to putting the fixed sites in; we'll see.

Sure, I'm happy to for this to be changed. See my comment about though: the final value of the AFS is used currently. 

> Hm, one thing: I thought we agreed to set the default to polarised=True for the AFS?

So, there's a technical issue with doing this at the C level: TSK_STAT_POLARISED is a bit flag and we can't default it to being set. Also, it seems a bit confusing to have polarised the default in some places but not in others. So, my take at this point would be to either leave things as they are, or to change the signature of this method to replace ``polarised=False`` with ``folded=False``, and document that folded == (not polarised).

petrelharp:
> So, my take at this point would be to either leave things as they are, or to change the signature of this method to replace polarised=False with folded=False, and document that folded == (not polarised).

Ok, let's leave things as they are.

python/tskit/trees.py
petrelharp: Good point; I've fixed this.

jeromekelleher:
> Ok, let's leave things as they are

Done! Merging; hooray no more FFS!

jeromekelleher:
Thanks for your help BTW @petrelharp  - I think I'd have just dropped this a while ago otherwise!

petrelharp:
Thank YOU for doing the work, and sorry again for misleading on the folding...

PR #279, opened 2019-07-31
Improve stat docs

petrelharp:
This is for discussion and iteration on the docs. I've fixed the links as pointed out by @molpopgen and as suggested on email:

> a Python example showing exactly what 'f' must do would be really useful.  Likewise, for sample_count_stat.  The docs state that f returns a 1d array, but it isn't clear (to me) if that array is of length 1 in the case of 1 sample set, but it seems like it must be, else it would have to be  higher dimension for > 1 sample set?

I made an example for sample counts; I was having a hard time coming up with a not-super-wierd example for general_stat; any suggestions?

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -276,6 +276,17 @@ pairwise) and Patterson's f statistics (which are four-way).
 - :meth:`.TreeSequence.Y3`
 - :meth:`.TreeSequence.Y2`
 
+------------------
+Trait correlations
+------------------
+
+These methods compute correlations and covariances of traits (i.e., an
+arbitrary vector) with allelic state, possibly in the context of a multivariate
+regression with other covariates (as in GWAS).
+
+- :meth:`.TreeSequence.trait_covariance`
+- :meth:`.TreeSequence.trait_correlation`
+
 ---------------
 General methods
 ---------------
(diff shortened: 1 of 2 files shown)


molpopgen:
Once I get a chance to play around a bit, maybe I'll come up with a "general" stat example.

petrelharp:
I thought about things like "sum of phenotype in subset of samples" or "produce of mean latitude and mean longitude of samples having this allele"...

molpopgen:
> I thought about things like "sum of phenotype in subset of samples" or "produce of mean latitude and mean longitude of samples having this allele"...

The z-score due to genotypes with this allele could be cool?

molpopgen:
> The z-score due to genotypes with this allele could be cool?

nevermind--this is complex b/c you'd have to know individuals?

petrelharp:
> The z-score due to genotypes with this allele could be cool?

what is that?

molpopgen:
> what is that?

(Mean phenotype of everyone with the mutation - population mean phenotype)/(std dev of phenotype in population).  

"Population" could mean "sample" here, too.

petrelharp:
That's [this])https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.trait_correlation) (which I see we forgot to link to in the stats docs...)

petrelharp:
Made that link.

jeromekelleher:
Shall we merge this much @petrelharp? If so, can you rebase and merge it please?

PR #284, opened 2019-08-02
Updated map_ancestors documentation - new version

gtsambos:
Same changes that I made in this [pull request](https://github.com/tskit-dev/tskit/pull/266) with a messed up git history.

cc @petrelharp, @jeromekelleher

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1682,6 +1682,8 @@ class TableCollection(object):
 
         The following table shows which ``parent->child`` pairs will be shown in the
         output of ``map_ancestors``.
+        A node is a relevant descendant on a given interval if it also appears somewhere
+        in the ``parent`` column of the outputted table.
 
         | Type of relationship     | Shown in output of ``map_ancestors``            |
         | ------------------------ | ----------------------------------------------- |
@@ -1697,10 +1699,12 @@ class TableCollection(object):
         The node IDs in ``parent`` and ``child`` refer to the IDs in the node table
         of the inputted tree sequence.
 
-        The supplied nodes must be non-empty lists of the node IDs in the tree sequence.
-        The lists of ``samples`` and ``ancestors`` may overlap.
-        If ``samples`` and ``ancestors`` are the same list of nodes, the function will
-        find all genealogical relationships within this list.
+        The supplied nodes must be non-empty lists of the node IDs in the tree sequence:
+        in particular, they do not have to be *samples* of the tree sequence. The lists
+        of ``samples`` and ``ancestors`` may overlap, although adding a node from
+        ``samples`` to ``ancestors`` will not change the output. So, setting ``samples``
+        and ``ancestors`` to the same list of nodes will find all genealogical
+        relationships within this list.
 
         If none of the nodes in ``ancestors`` or ``samples`` are ancestral to ``samples``
         anywhere in the tree sequence, an empty table will be returned.
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1665,27 +1665,45 @@ class TableCollection(object):
 
     def map_ancestors(self, samples, ancestors):
         """
-        Returns an :class:`EdgeTable` instance describing how ``samples`` are descended
-        from ``ancestors`` (and how ``ancestors`` are descended from each other).
-        Each row ``parent, child, left, right`` in the output table indicates that
-        ``child`` (which is either in ``samples`` or in ``ancestors``) has inherited the
-        segment ``[left, right)`` from ``parent`` (which is in ``ancestors``) more
-        recently than from any other node in ``ancestors``. Every segment of the
-        genome on which each of ``samples`` has inherited from any of ``ancestors``
-        is described in this way, as well as any segment of the genomes of ``ancestors``
-        that has been inherited by any of ``samples``.  The node IDs in ``parent`` and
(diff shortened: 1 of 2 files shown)


python/tskit/tables.py
```diff
@@ -1665,27 +1665,45 @@ def simplify(
...
+
+        The following table shows which ``parent->child`` pairs will be shown in the
+        output of ``map_ancestors``.
+
+        | Type of relationship     | Shown in output of ``map_ancestors``            |
+        | ------------------------ | ----------------------------------------------- |
+        | ``ancestor->sample``     | Always                                          |
+        | ``ancestor1->ancestor2`` | Only if ``ancestor2`` has a relevant descendant |
+        | ``sample1->sample2``     | Always                                          |
+        | ``sample->ancestor``     | Only if ``ancestor`` has a relevant descendant  |
```
petrelharp: vote to change "relevant descendant" to "descendant in samples or ancestors"

python/tskit/tables.py
```diff
@@ -1665,27 +1665,45 @@ def simplify(
...
+        the ancestors of a node in ``ancestors`` will only be retained if it also has a
+        relevant descendant, while information about the ancestors of a node in
+        ``samples`` will always be retained.
+        The node IDs in ``parent`` and ``child`` refer to the IDs in the node table
+        of the inputted tree sequence.
+
+        The supplied nodes must be non-empty lists of the node IDs in the tree sequence.
+        The lists of ``samples`` and ``ancestors`` may overlap.
+        If ``samples`` and ``ancestors`` are the same list of nodes, the function will
+        find all genealogical relationships within this list.
```
petrelharp: How about 
```
The supplied nodes must be non-empty lists of the node IDs in the tree sequence:
in particular, they do not have to be *samples* of the tree sequence. The lists of
``samples`` and ``ancestors`` may overlap, although adding a node from ``samples``
to ``ancestors`` will not change the output. So, setting ``samples`` and ``ancestors``
to the same list of nodes will find all genealogical relationships within this list.
```

petrelharp:
> I'm happy to look into modifying this if you think it is going to be confusing for the user, though maybe this is something to consider in a separate pull request. It wouldn't be hard to change it so that if the user passes an empty list into ancestors, map_ancestors internally copies the nodes from samples into ancestors.

It was confusing for me. Either doing the hack you suggest, or documenting the situation: you need to pass in *someone* for ancestors but since adding samples to the ancestors argument doesn't change anything, if you don't really want ancestors you can just put in a sample for ancestors. I've suggested a change.

> this description has been a bit painful for me to write is because in the use case I was designing this function for, samples are always from the most recent generation and ancestors never are.

Right! I think the "find close relatives" application will be maybe just as useful, though, so it's great that you've, er, killed two birds with one stone.

jeromekelleher:
I hate to be the git-pedant here @gtsambos, but you need to ``git fetch upstream`` and rebase again - if you look at the graph, your branch is forking off master a long way back and it needs to be at the tip. Upstream is moving quickly at the moment - we're trying to ship a bunch of stuff for 0.2.0 soon.

python/tskit/tables.py
gtsambos: this isn't quite true though - it will only keep a descendant in `ancestors` if that node is also parent to another edge in the table.

petrelharp: o right. =) so, shouldn't it say "has a descendant in samples"? or, "is ancestral to a sample"?

gtsambos:
> It was confusing for me. Either doing the hack you suggest, or documenting the situation: you need to pass in someone for ancestors but since adding samples to the ancestors argument doesn't change anything, 

I was hoping that this bit would indicate this behaviour, but I guess it doesn't explicitly say what the consequences will be if one of the lists is empty - I can amend

> The supplied nodes must be non-empty lists of the node IDs in the tree sequence.

gtsambos:
> I hate to be the git-pedant here @gtsambos, but you need to git fetch upstream and rebase again - if you look at the graph, your branch is forking off master a long way back and it needs to be at the tip. Upstream is moving quickly at the moment - we're trying to ship a bunch of stuff for 0.2.0 soon.

yup, done!

python/tskit/tables.py
gtsambos: 👍 this is in the latest commit.

python/tskit/tables.py
gtsambos: hmm, on its own that wording is also a bit confusing I think, because it might not be immediately ancestral to a sample - it could be ancestral to an ancestor that is ancestral to an ancestor... that is ancestral to a sample, for example.

I've added an extra sentence to say what I mean by 'relevant descendant', but I can also add something like 'has a descendant in samples on the given interval'

petrelharp:
> I was hoping that this bit would indicate this behaviour, but I guess it doesn't explicitly say what the consequences will be if one of the lists is empty - I can amend

you saw my suggestion, I hope?

gtsambos:
> you saw my suggestion, I hope?

yup it's in the latest commit! [here](https://github.com/tskit-dev/tskit/pull/284/commits/20adedf2285c5a5451eb741bbd6fbd50ac2f7d7f)

petrelharp:
Looks good to me! :yay: Thanks for getting this sorted.

jeromekelleher:
LGTM - can you push the merge button if you're happy please @petrelharp?

Issue #59, opened 2018-01-13
Add ``squash_edges`` function


jeromekelleher:
It's useful to be able to 'squash' redundant edges (i.e, two edges (l, x, p, c) and (x, r, p, c) squash into (l, r, p, c)), and we should add a public function the tables API for this. The C function already exists, so it's just a question of setting up the plumbing.

This would be a good place to starter issue for someone looking to figure out the connections between the Python APIs and C code.

gtsambos:
I'll attempt to take this on! Btw, there is a Python `squash_edges` function in `test_topology.py` over [here](https://github.com/tskit-dev/tskit/blob/d03ba177121c5c8ce710323fe6c7f94fda480364/python/tests/test_topology.py#L4016).

jeromekelleher:
Cool - the python version will be handy to write a few quick tests to compare with.

gtsambos:
Hmmm, I'm wondering whether the existing `squash_edges` function in C was made for some slightly different purpose to what we need it for here (link to code is [here](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L6762-L6786))?

- I think the first argument should be of type `tsk_edge_table_t *`? I've assumed this is a typo and fixed it
- Assuming that `edges` has type `edge_table_t *`, then all the things like `edges[k].parent` should actually be `edges->parent[k]` (I think?)
- This assertion `assert(edges[k - 1].parent == edges[k].parent)` suggests that the input into this function isn't a normal edge table but some truncated one, as most edge tables would have more than one distinct parent in them. The comparison function `cmp_edge_cl` that you defined [here](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L6747-L6756) also looks like it's been designed for this purpose.

Btw I know you're very busy @jeromekelleher! Just flagging these to indicate that this task might take a bit longer than we first thought

```
tsk_squash_edges(tsk_edge_t *edges, size_t num_edges, size_t *num_output_edges)
{
    int ret = 0;
    size_t j, k, l;
    tsk_edge_t e;


    qsort(edges, num_edges, sizeof(tsk_edge_t), cmp_edge_cl);
    j = 0;
    l = 0;
    for (k = 1; k < num_edges; k++) {
        assert(edges[k - 1].parent == edges[k].parent);
        if (edges[k - 1].right != edges[k].left || edges[j].child != edges[k].child) {
            e = edges[j];
            e.right = edges[k - 1].right;
[...]

PR #285, opened 2019-08-05
Squash_edges

gtsambos:
Adds a working `squash_edges` function (see [this issue](https://github.com/tskit-dev/tskit/issues/59)), links up with Python implementation.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -728,7 +728,6 @@ test_edge_table_squash(void)
 {
     int ret;
     tsk_table_collection_t tables;
-    tsk_edge_table_t *edges;
 
     const char *nodes_ex =
         "1  0       -1   -1\n"
@@ -755,16 +754,145 @@ test_edge_table_squash(void)
     parse_edges(edges_ex, &tables.edges);
     CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 4);
 
-    edges = &tables.edges;
-
-    ret = tsk_edge_table_squash(edges);
+    ret = tsk_edge_table_squash(&tables.edges);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     // Check output.
-    CU_ASSERT_EQUAL(edges->num_rows, 2);
+    CU_ASSERT_EQUAL(tables.edges.num_rows, 2);
 
     // Free things.
-    tsk_edge_table_free(edges);
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_edge_table_squash_multiple_parents(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+
+    const char *nodes_ex =
+        "1  0.000   -1    -1\n"
+        "1  0.000   -1    -1\n"
+        "1  0.000   -1    -1\n"
+        "1  0.000   -1    -1\n"
+        "0  1.000   -1    -1\n"
+        "0  1.000   -1    -1\n";
+    const char *edges_ex =
+        "5  10  5   3\n"
+        "5  10  5   2\n"
+        "0  5   5   3\n"
+        "0  5   5   2\n"
+        "4  10  4   1\n"
+        "0  4   4   1\n"
+        "4  10  4   0\n"
+        "0  4   4   0\n";
+/*
+            4       5  
+           / \     / \
+          0   1   2   3
+*/    
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 10;
+
+    parse_nodes(nodes_ex, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 6);
+    parse_edges(edges_ex, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 8);
+
+    ret = tsk_edge_table_squash(&tables.edges);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Check output.
+    CU_ASSERT_EQUAL(tables.edges.num_rows, 4);
+
+    // Free things.
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_edge_table_squash_empty(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+
+    const char *nodes_ex =
+        "1  0       -1   -1\n"
+        "1  0       -1   -1\n"
+        "0  0.253   -1   -1\n";
+    const char *edges_ex =
+        "";
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 10;
+
+    parse_nodes(nodes_ex, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 3);
+    parse_edges(edges_ex, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 0);
+
+    ret = tsk_edge_table_squash(&tables.edges);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Free things.
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_edge_table_squash_single_edge(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+
+    const char *nodes_ex =
+        "1  0   -1   -1\n"
+        "0  0   -1   -1\n";
+    const char *edges_ex =
+        "0  1   1   0\n";
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1;
+
+    parse_nodes(nodes_ex, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 2);
+    parse_edges(edges_ex, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 1);
+
+    ret = tsk_edge_table_squash(&tables.edges);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Free things.
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_edge_table_squash_bad_intervals(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+
+    const char *nodes_ex =
+        "1  0   -1   -1\n"
+        "0  0   -1   -1\n";
+    const char *edges_ex =
+        "0  0.6   1   0\n"
+        "0.4  1   1   0\n";
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1;
+
+    parse_nodes(nodes_ex, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 2);
+    parse_edges(edges_ex, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 2);
+
+    ret = tsk_edge_table_squash(&tables.edges);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_EDGES_CONTRADICTORY_CHILDREN);
+
+    // Free things.
     tsk_table_collection_free(&tables);
 }
 
@@ -2486,6 +2614,11 @@ main(int argc, char **argv)
         {"test_node_table", test_node_table},
         {"test_edge_table", test_edge_table},
         {"test_edge_table_squash", test_edge_table_squash},
+        {"test_edge_table_squash_multiple_parents", 
+            test_edge_table_squash_multiple_parents},
+        {"test_edge_table_squash_empty", test_edge_table_squash_empty},
+        {"test_edge_table_squash_single_edge", test_edge_table_squash_single_edge},
+        {"test_edge_table_squash_bad_intervals", test_edge_table_squash_bad_intervals},
         {"test_site_table", test_site_table},
         {"test_mutation_table", test_mutation_table},
         {"test_migration_table", test_migration_table},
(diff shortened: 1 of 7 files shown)


gtsambos:
Hi @jeromekelleher, I thought it would be best to make a pull request so you can see what I'm doing/so I waste time going off the track.

Here's what you suggested in the issue:

> 1.  Generalise the `squash_edges` function to take lists  that have more than one distinct parent value. This would require (a) adding `parent` as the first key in the comparison function (there should be lots of other comparison functions around to use as an example); (b) changing the for loop logic to deal with changing from one parent to another; (c) writing a couple of (very) simple test cases to cover this functionality (manually creating a few unsorted small lists of edges and testing the output is squashed appropriately).
> 2.  Add a a new method `tsk_edge_table_squash` that mallocs a list of edges, copies its contents into this list, calls `squash_edges`, calls `tsk_edge_table_clear()` and then runs `tsk_edge_table_add_row` for each of the new squashed edges.
> 3.  Set up the plumbing between `tsk_edge_table_squash` and `EdgeTable_squash` in `_tskitmodule.c`.

I think I'm going to do this in a different order bc the thing that's most confusing to me is the code that's already there. As a first step, I've fixed up `tsk_squash_edges` so that it works with the limited comparison function already there (the one that requires all parents to be the same), and I've made a function to test this more limited `tsk_squash_edges`. Could you please see if this looks okay? If so, I'll go on and do all those things you've suggested in step 1.

c/tskit/tables.c
```diff
@@ -6758,30 +6790,61 @@ cmp_edge_cl(const void *a, const void *b) {
...
+    qsort(edge_list, num_edges, sizeof(tsk_edge_t), cmp_edge_cl);
     j = 0;
     l = 0;
     for (k = 1; k < num_edges; k++) {
-        assert(edges[k - 1].parent == edges[k].parent);
-        if (edges[k - 1].right != edges[k].left || edges[j].child != edges[k].child) {
-            e = edges[j];
-            e.right = edges[k - 1].right;
-            edges[l] = e;
+        assert(edge_list[k-1].parent == edge_list[k].parent);
```
jeromekelleher: I'm not following the logic here --- we can duplicate the logic from the squash_edges function in ``test_topology.c``, right? The only difference here is that we're writing back to the same array rather than creating a second one. This makes it slightly trickier but I'm sure it's possible.

gtsambos:
Hi @jerome, thanks for looking at this! Are you free around this time tomorrow? I'm currently preparing a talk for Simon's group that's tomorrow morning, but very free after that. Will address these comments then also.

gtsambos:
Hi Jerome, so I still need to add some more Python and C tests of squash_edges, but I was wondering whether you could take a look at whether everything else looks like it's vaguely structured okay? (As per [this issue](https://github.com/tskit-dev/tskit/issues/289), I'm currently having some difficulties running Python tests.)

c/tskit/tables.c
```diff
@@ -1342,6 +1342,43 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
     return read_table_cols(store, read_cols, sizeof(read_cols) / sizeof(*read_cols));
 }
 
+int
+tsk_edge_table_squash(tsk_edge_table_t *self)
+{
+    int k;
+    int ret = 0;
+    tsk_edge_t e;
+    tsk_edge_t * edges;
```
jeromekelleher: Change to 
```
tsk_size_t num_output_edges;
tsk_edge_t *edges = malloc(self->num_rows * sizeof(tsk_edge_t));

if (edges == NULL) 
// raise errro
```

c/tskit/tables.c
```diff
@@ -1342,6 +1342,43 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
...
+    self->num_rows = num_output_edges;
+    for (k = 0; k < (int) num_output_edges; k++){
+        self->left[k] = edges[k].left;
+        self->right[k] = edges[k].right;
+        self->parent[k] = edges[k].parent;
+        self->child[k] = edges[k].child;
+     }
+
+     // Free edge list.
+     free(edges);
```
jeromekelleher: Change to 
```c
out:
    tsk_safe_free(edges);
    return ret
```

c/tskit/tables.c
```diff
@@ -1342,6 +1342,43 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
...
+        e.child = self->child[k];
+        edges[k] = e;
+    }
+
+    ret = tsk_squash_edges(edges, self->num_rows, &num_output_edges);
+    if (ret != 0) {
+        goto out;
+    }
+    tsk_edge_table_clear(self);
+    self->num_rows = num_output_edges;
```
jeromekelleher: Good to add an assertion to check that num_rows < max_rows here, just in case.

c/tskit/tables.c
```diff
@@ -6748,7 +6785,10 @@ static int
 cmp_edge_cl(const void *a, const void *b) {
     const tsk_edge_t *ia = (const tsk_edge_t *) a;
     const tsk_edge_t *ib = (const tsk_edge_t *) b;
-    int ret = (ia->child > ib->child) - (ia->child < ib->child);
+    int ret = (ia->parent > ib->parent) - (ia->parent < ib->parent);
+    if (ret == 0)  {
+        ret = (ia->child > ib->child) - (ia->child < ib->child);
+    }
```
jeromekelleher: This isn't right --- the conditions need to be nested I think.

c/tskit/tables.c
```diff
@@ -6769,19 +6810,30 @@ tsk_squash_edges(tsk_edge_t *edges, size_t num_edges, size_t *num_output_edges)
...
+            e.child = edges[j].child;
+
+            edges[j] = e;
+
             j = k;
             l++;
         }
     }
-    e = edges[j];
-    e.right = edges[k - 1].right;
```
jeromekelleher: This blows up when there's zero edges. Need to add a test case to cover this in C.

python/tskit/tables.py
```diff
@@ -657,6 +657,17 @@ def asdict(self):
...
 
+    def squash(self):
+        """
+        Sorts, then condenses the table into the smallest possible number of rows.
+        A single edge replaces any set of edges with the same parent and child node
+        such that the union of (left, right] intervals is a single interval.
+        The new edge will have the same parent and child node, a left coordinate
+        equal to the smallest left coordinate in the set, and a right coordinate
+        equal to the largest right coordinate in the set.
+        """
```
jeromekelleher: We should note whether the edges are in canonical order after this operation (must be, right?), linking to the sortedness requirements for edges in the docs.

gtsambos:
Hi @jeromekelleher, as long as this passes the build tests this should be ready for review

c/tests/test_tables.c
```diff
@@ -723,6 +723,168 @@ test_edge_table(void)
...
+    edges = &tables.edges;
+
+    ret = tsk_edge_table_squash(edges);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Check output.
+    CU_ASSERT_EQUAL(edges->num_rows, 2);
+
+    // Free things.
+    tsk_edge_table_free(edges);
```
jeromekelleher: Isn't this being double-freed?

c/tests/test_tables.c
```diff
@@ -723,6 +723,168 @@ test_edge_table(void)
     free(child);
 }
 
+static void
+test_edge_table_squash(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+    tsk_edge_table_t *edges;
```
jeromekelleher: I don't think we need the edges pointer here, it's just confusing things I think.

c/tests/test_tables.c
```diff
@@ -723,6 +723,168 @@ test_edge_table(void)
...
+    tables.sequence_length = 10;
+
+    parse_nodes(nodes_ex, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 3);
+    parse_edges(edges_ex, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 4);
+
+    edges = &tables.edges;
+
+    ret = tsk_edge_table_squash(edges);
```
jeromekelleher: Can use ``ret = tsk_edge_table_squash(&tables->edges);`` (or something similar)

c/tests/test_tables.c
```diff
@@ -723,6 +723,168 @@ test_edge_table(void)
...
+    tsk_edge_table_free(edges);
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_edge_table_squash_multiple_parents(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+    tsk_edge_table_t *edges;
```
jeromekelleher: Also don't need the pointer to the edge table here.

c/tskit/tables.c
```diff
@@ -1343,6 +1343,45 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
     return read_table_cols(store, read_cols, sizeof(read_cols) / sizeof(*read_cols));
 }
 
+int
+tsk_edge_table_squash(tsk_edge_table_t *self)
+{
+    int k;
+    int ret = 0;
+    tsk_edge_t e;
+    tsk_edge_t * edges;
```
jeromekelleher: We usually don't have a space here, ``tsk_edge_t *edges = NULL;``

c/tskit/tables.c
```diff
@@ -1343,6 +1343,45 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
...
+    edges = malloc(self->num_rows * sizeof(tsk_edge_t));
```
jeromekelleher: You must check if malloc failed: ``if (edges == NULL)...``.

python/tests/test_topology.py
```diff
@@ -4030,6 +4030,194 @@ def test_many_multiroot_trees(self):
...
+        squashed = ts.tables.edges
+        squashed.squash()
+        if compare_lib:
+            squashed_list = squash_edges(ts)
+            squashed_py = tskit.EdgeTable()
+            for e in squashed_list:
+                squashed_py.add_row(e.left, e.right, e.parent, e.child)
+            # Check the Python and C implementations produce the same output.
+            self.assertEqual(squashed_py, squashed)
+        return(squashed)
```
jeromekelleher: just ``return squashed``

gtsambos:
🤞 @jeromekelleher

c/tskit/tables.c
```diff
@@ -1343,6 +1343,44 @@ tsk_edge_table_load(tsk_edge_table_t *self, kastore_t *store)
...
+{
+    int k;
+    int ret = 0;
+    tsk_edge_t *edges = NULL;
+    edges = malloc(self->num_rows * sizeof(tsk_edge_t));
+    if (edges == NULL){
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    tsk_size_t num_output_edges;
```
jeromekelleher: This isn't mandated by C99, but older versions insisted on variable declarations coming before code. We've stuck to this in tskit as a generally good idea. So, the top of the function should look like:
```c
  int k;
  int ret = 0;
  tsk_edge_t *edges = NULL;
  tsk_size_t num_output_edges; 

  edges = malloc(self->num_rows * sizeof(tsk_edge_t));
```
By convention, we keep a space between the variable declaration bit at the start of the function and the actual code.

c/tskit/tables.c
```diff
@@ -6780,40 +6818,75 @@ static int
...
     return ret;
 }
 
 /* Squash the edges in the specified array in place. The output edges will
  * be sorted by (child_id, left).
  */
+
 int TSK_WARN_UNUSED
-tsk_squash_edges(tsk_edge_t *edges, size_t num_edges, size_t *num_output_edges)
+tsk_squash_edges(tsk_edge_t *edges, size_t num_edges, tsk_size_t *num_output_edges)
```
jeromekelleher: I should have spotted this before, but if we're changing one of these to ``tsk_size_t`` we should change both. Can you change ``num_edges`` to be a ``tsk_size_t`` please?

c/tskit/tables.c
```diff
@@ -6780,40 +6818,75 @@ static int
...
 {
     int ret = 0;
     size_t j, k, l;
-    tsk_edge_t e;
+
+    if (num_edges == 0 || num_edges == 1){
```
jeromekelleher: ``num_edges < 2`` requires only one comparison...

c/tskit/tables.c
```diff
@@ -6780,40 +6818,75 @@ static int
...
     l = 0;
     for (k = 1; k < num_edges; k++) {
-        assert(edges[k - 1].parent == edges[k].parent);
-        if (edges[k - 1].right != edges[k].left || edges[j].child != edges[k].child) {
-            e = edges[j];
-            e.right = edges[k - 1].right;
-            edges[l] = e;
+
+        if (edges[k-1].parent != edges[k].parent || edges[k-1].right != edges[k].left
+            || edges[j].child != edges[k].child) {
```
jeromekelleher: Indent this line one more step to show it's not part of the body:
```c
if (edges[k-1].parent != edges[k].parent || edges[k-1].right != edges[k].left
        || edges[j].child != edges[k].child) { 
     edges[l].left = edges[j].left;
...
```

c/tskit/tables.c
```diff
@@ -6780,40 +6818,75 @@ static int
...
+
+            edges[l].left = edges[j].left;
+            edges[l].right = edges[k-1].right;
+            edges[l].parent = edges[j].parent;
+            edges[l].child = edges[j].child;
+
             j = k;
             l++;
         }
```
jeromekelleher: else is on the same line as the brackets:
```c
} else {
```

c/tskit/tables.c
```diff
@@ -6780,40 +6818,75 @@ static int
...
     j = 0;
     l = 0;
     for (k = 1; k < num_edges; k++) {
-        assert(edges[k - 1].parent == edges[k].parent);
-        if (edges[k - 1].right != edges[k].left || edges[j].child != edges[k].child) {
-            e = edges[j];
-            e.right = edges[k - 1].right;
-            edges[l] = e;
+
+        if (edges[k-1].parent != edges[k].parent || edges[k-1].right != edges[k].left
```
jeromekelleher: space around arithmetic operators please: ``edges[k - 1].parent`` --- there's a bunch more here and below.

gtsambos:
🤞 🍀  @jeromekelleher

jeromekelleher:
Looks good @gtsambos; minor logic change in the algorithm, as discussed. Also, could you squash out the last few commits please? (Back as far as "Tests for squash_edges" seems appropriate)

gtsambos:
Hi @jeromekelleher, having spent the last little while thinking about this logic change, I'm not sure it makes things a whole lot clearer than what I had originally? The [latest commit](https://github.com/tskit-dev/tskit/pull/285/commits/f9653b55ce8f639756c0a5051d287c6aad23dbba) shows this change, but notice that it necessitates putting 'add edge' steps in several different parts of the function, which makes it harder to see what the core utility of the function is.

After spending some time fiddling around and breaking and fixing existing code, I don't think there's a way we can make this work without either repeating the 'add edge' code or repeating the 'bad edge interval' code.

gtsambos:
🍾 🎉

jeromekelleher:
Done; thanks @gtsambos!

Issue #73, opened 2018-07-02
use individual tables when writing out vcf files


petrelharp:
Currently `write_vcf` has a `ploidy` argument, and will assume that adjacent samples belong to the same individual. This should be deprecated after we provide a `make_individuals` method, and `write_vcf` should have an argument that toggles whether to write individuals or genomes (maybe: `output` either `monoploid` by default or `phased` or `unphased).

petrelharp:
Also we should have an option to output a diploid genotype matrix, summing together the rows of genotype matrix for the given individuals' nodes.

jeromekelleher:
> Also we should have an option to output a diploid genotype matrix, summing together the rows of genotype matrix for the given individuals' nodes.

So we output unphased "dosage" values?

petrelharp:
> So we output unphased "dosage" values?

Right. Well, we should output diploid genotypes based on individuals in a way that works well with scikit-allel.

jeromekelleher:
I'm going to close this after #73 is merged as this uses the individual table to generate the diploid/whateverploid individuals. We should open more specific issues for the other stuff raised here if we want them.

jeromekelleher:
Closed in #286.

Issue #30, opened 2018-12-19
document and/or change surprising effects of VCF position rounding


DrK-Lo:
I'm running replicate simulations, and in a subset of my simulations a few of the loci that I simulated in SliM do not have the expected allele frequencies and genotypes after overlaying neutral mutations with pyslim and outputting with write_vcf from msprime. In other words, the allele frequency at a position outputted from SliM does not match the allele frequency at that position after recapitation and overlaying neutral mutations, then outputting to VCF. 

I consulted with @petrelharp with this issue at: https://github.com/tskit-dev/pyslim/issues/25 where you can find files and code to reproduce the problem.

In this simulation, I had a selected mutation at 144367. After recaptiation and adding neutral mutations, mutated has a position at 144366.71192197, which rounds to 144367. 

In the VCF that there is a variant at position 144368, although the next position in the tree sequence is 144381.77609599 - we confirmed what has happened is that write_vcf has rounded positions, found two mutations at the same position, and bumped up the position of the selected mutation to be unique. 

This could be fixed in a couple ways. One way could be by being able to pass information about the mutations from pyslim to the vcf@info field.

jeromekelleher:
Closed in #286; alleles are supported properly now.

Issue #2, opened 2018-11-23
VCF should round to nearest integer


jeromekelleher:
The current VCF conversion code has an algorithm for ensuring that we don't output duplicate sites. This is a bad idea. The simplest thing to do is to round all positions to the nearest integer. It's up to the user how the interpret these. If they need to control this, then they're free to update the positions in the site table beforehand.

jeromekelleher:
Closed in #286.

PR #286, opened 2019-08-06
Write VCFs in Python

jeromekelleher:
There's lots of problems with our VCF output, and fixing them is really tedious when output is done in C. This changes to using numpy tricks to generate the VCF rather than low-level C. Hopefully we don't lose too much perf (will update with some numbers).

I'd propose we get rid of VCF support in the C library entirely then.

This is surely a win, right?


petrelharp:
> This is surely a win, right?

Yes!!!

jeromekelleher:
There is a significant perf hit in the current version: we've gone from 2.5 minutes to 10 minutes for a 100K sample tree sequence. It looks like most of the time is spent in unicode encoding though, which shouldn't be necessary.

jeromekelleher:
Latest updates bring this down to 4 minutes.  I'm willing to take a less than 2X hit here --- does it really matter if VCF output is slightly slower? It's still outputting at around 350M/s, which is as fast as most storage will be able to read it. Plus, I think a lot of the time is being spent in the header --- it's pretty stupid right now.

python/tskit/vcf.py
```diff
@@ -0,0 +1,87 @@
...
+    n = ts.get_sample_size() // ploidy
+    sample_names = ["msp_{}".format(j) for j in range(n)]
+    last_pos = 0
+    positions = []
+    for variant in ts.variants():
+        pos = int(round(variant.position))
+        if pos <= last_pos:
+            pos = last_pos + 1
+        positions.append(pos)
+        last_pos = pos
```
petrelharp: Here's a much faster way to do this bit:
```
positions = np.round(ts.tables.sites.position)
positions[1:] += np.cumsum(np.diff(positions) == 0)
```

jeromekelleher: Nice, thanks @petrelharp.

jeromekelleher:
TODO: rerun the perf check above, we were generating all **variants** in the header, which would massively slow things down. So, we should be more-or-less the same as the old C code now.

jeromekelleher:
> TODO: rerun the perf check above, we were generating all variants in the header, which would massively slow things down. So, we should be more-or-less the same as the old C code now.

It's down to 3.25 minutes now, of which ~3 seconds are spent in the header. So that's fast enough; job done. I'll update with some sane default behaviour and basic support for individuals soon.

jeromekelleher:
I think this is basically done except for a few more tests (need to check the location_transform and individual_names inputs) and to write some documentation. What do you think @petrelharp? Do you like the ``location_transform`` option? I think we need this flexibility, as there's the uber-bioinformatics confusion about whether we're using 1 or 0 based coordinates in VCF and with real data we're going to need to support both probably. The default policy is to just leave the coordinates as they are (which is what you want for simulated data, presumably).

I've thought a bit about how to include metadata as INFO fields, but I think we'll have to wait for proper metadata decoding within tskit for that. At least this does sane things site positions by default (rounds them), puts in the real allele values and uses individuals if they are present. This is a *big* improvement on what's currently available.

python/tskit/vcf.py
petrelharp: Whoops; I see that my proposal didn't actually do what I wanted. I see you aren't using it anyhow... but here's code that actually does the same thing as the `legacy_position_transform`:
```
def position_transform(positions, max_iter = 100):
    positions = np.round(positions).astype("int")
    for _ in range(max_iter):
        a = np.where(np.diff(positions) == 0)[0] + 1
        if len(a) == 0:
            break
        positions[a] += 1

# test it
def gen_positions(n, m):
    out = m * np.random.random(n)
    out.sort()
    return out

for _ in range(100):
    p1 = gen_positions(10, 20)
    p2 = p1.copy()
    transform(p1)
    legacy_position_transform(p2)
    assert(np.all(p1 == p2))
```

petrelharp:
>  Do you like the location_transform option? I think we need this flexibility, as there's the uber-bioinformatics confusion about whether we're using 1 or 0 based coordinates in VCF and with real data we're going to need to support both probably. The default policy is to just leave the coordinates as they are (which is what you want for simulated data, presumably).

Yes, I like it (including the default leave-as-is option!).  I provided some actually-correct numpy code for the "legacy" version.

petrelharp:
Looks good to me. Thanks for getting to this!

jeromekelleher:
> Yes, I like it (including the default leave-as-is option!). I provided some actually-correct numpy code for the "legacy" version.

Great, I'll finish this up and merge it so when I get a chance.

python/tskit/vcf.py
jeromekelleher: Hmm, that still doesn't do exactly the same thing. It doesn't matter anyway --- this is legacy behaviour, and for large n the time spent doing this is trivial.

jeromekelleher:
Closes #2 
Closes #30 
Closes  #73

jeromekelleher:
OK, going to merge this.

PR #291, opened 2019-08-08
Fixed typo in map_ancestors docs.

gtsambos:
Noticed that my documentation for map_ancestors wasn't rendering properly. I think this is just because I initially didn't really know what restructured text was, or fully understand that the documentation wasn't also compatible with Markdown syntax 😅Hopefully this fixes it!


petrelharp:
Whoops, thanks for the fix. You can tell that I just looked at the source and didn't render it myself. =) (note to self: always `make` the docs locally and read them that way...)

Issue #288, opened 2019-08-07
Implement `ltrim()`, `rtrim()`, and `trim()` to remove genomic areas with no edges


hyanwong:
From https://github.com/tskit-dev/tskit/pull/261: I could imagine defining `.ltrim()`, `.rtrim()`, and their joint `.trim()` on a ts, which basically reset genomic positions to crop off extremities of the ts which have no edges defined. E.g. `.ltrim()` shifts the ts so that the leftmost edge is at genomic position 0, `.rtrim()` simply redefines `ts.sequence_length` to be the max of `edges.right`. 

Only `.ltrim()` and `.trim()` would actually reset the coordinate system, and change the meaning of a genomic position.

PR #292, opened 2019-08-08
First pass at trim functions

hyanwong:
Closes https://github.com/tskit-dev/tskit/issues/288

No tests yet, as these will be easier once https://github.com/tskit-dev/tskit/pull/261 is committed, so that we can easily make tree sequences with no trees to left or right.

Also the documentation will need improving (e.g. document the TableCollection methods). I will do this once the general approach has been approved by @jeromekelleher .

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1858,14 +1858,6 @@ class TableCollection(object):
         :rtype: tskit.TableCollection
         """
 
-        def keep_with_offset(keep, data, offset):
-            # We need the astype here for 32 bit machines
-            lens = np.diff(offset).astype(np.int32)
-            return (data[np.repeat(keep, lens)],
-                    np.concatenate([
-                        np.array([0], dtype=offset.dtype),
-                        np.cumsum(lens[keep], dtype=offset.dtype)]))
-
         intervals = util.intervals_to_np_array(intervals, 0, self.sequence_length)
         if len(self.migrations) > 0:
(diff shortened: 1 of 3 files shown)


hyanwong:
Once https://github.com/tskit-dev/tskit/pull/261 is in, it would be rather easy to leave `slice` in as a convenience function. The documentation could simply say something like "`slice(start, stop): a convenient wrapper for `ts.keep_interval([(start, stop)]).trim()`.

But I'm equally happy not to have this.

hyanwong:
The code in `remove_sites` duplicates code in the proposed `keep_intervals()` function in https://github.com/tskit-dev/tskit/pull/261. But I think we do need to remove sites in `trim`, as we don't know beforehand if the TS to trim has come from `keep_intervals` or not. If not, we might end up with negative positions for some sites.

I suspect the `remove_sites` functionality is more generally useful too. I've coded it up here as a tables function, so it should be possible (although maybe less efficient) to use it in `keep_intervals`, if we want to reduce code duplication.

jeromekelleher:
This is quite a lot of code and will need a lot of testing. Can we push this back to after 0.2.0?

hyanwong:
Yes - I'd much prefer to prioritise getting 0.2.0 out. I've only used the trim functionality (which was in `slice()`) in one bit of non-reusable code anyway, so it's probably just me requiring this at the moment.

hyanwong:
Closing with the intent of opening a new PR, since the remove_sites functionality is now coded elsewhere

PR #294, opened 2019-08-09
Check for nonfinite edge values

### LINKED ISSUE #293: Stupid values in tables cause tables.tree_sequence() to hit an infinite loop
I inadvertently created a table collection containing an edge table with left=nan and right=nan. When I then tried to do `tables.tree_sequence()`, tskit seemed to enter an infinite loop from which I couldn't escape (I had to kill the process by hand).

Admittedly it's a stupid thing to have done. But should we try to exit more gracefully than this, or is it such a silly edge case that we shouldn't bother?

I'm happy to put in some assertions, but am not sure where they belong (in the C code, or in the python wrapper?)

hyanwong:
Fixes #293

Only currently testing edge values, so don't merge yet,


c/tskit/tables.c
```diff
@@ -5660,6 +5661,10 @@ tsk_table_collection_check_integrity(tsk_table_collection_t *self, tsk_flags_t o
             goto out;
         }
         /* Spatial requirements for edges */
+        if (isfinite(left) == 0 || isfinite(right) == 0) {
```
jeromekelleher: I don't think this is the right condition

hyanwong: Oh? Not sure what I should use instead...

jeromekelleher: Ah, no, it's right, it's just a bit confusing that way it's stated. What about 
```
if (! (isfinite(left) && isfinite(right))) 
```
Seems more readable.

hyanwong: Sure. Committed as an `--amend`

jeromekelleher:
Looks good, this is exactly right. So, ideally we'd check all the float values in the tables here with the C tests, and do a quick (non exhaustive) test in python/tests/test_tables.py to check that we're propagating the error up to the top level correctly.

hyanwong:
@jeromekelleher where do you want me to place `TSK_ERR_NODE_TIME_NONFINITE` in the list of errors in `core.h`. There isn't a set labelled `/* Node errors */` - I could create one but that would shift the numbering system up so that e.g. Mutation errors would start from 600 not 500 as they do at the moment. Do you care?

jeromekelleher:
> Added - or could these be considered "out of bounds" errors?

That's probably the most appropriate place yes, thanks.

hyanwong:
I'm having a minor problem understanding how to loop through the location values in C. In particular, I'm trying to do something like this (tables.c, line 5627), but obviously my syntax is wrong. What's the correct way to go through the k location values in each of j individuals?

```
    for (j = 0; j < num_individuals; j++) {
        for (k = 0; k < self->individuals.location_length; k++) {
            if (! isfinite(self->individuals.location[j][k])) {
                ret = TSK_ERR_INDIVIDUAL_LOCATION_NONFINITE;
                goto out;
        }
    }
```

jeromekelleher:
It's easier to avoid looping over the individuals and just look at whole position array in one go. Something like
```
for (j = 0; j < self->position_offset[self->num_rows]; j++) {
     // check self->position[j]
}
```
I might be off by one here though.

hyanwong:
I think this is done now, but `ci/circleci` is complaining about a memory leak. I can't see why that might be - I've copied all the code and each `tsk_treeseq_init()` is always matched with a `tsk_treeseq_free(&ts);` from what I can see. Any pointers to the problem gratefully received.

hyanwong:
@jeromekelleher: this is now done and ready for review.

jeromekelleher:
Looks like your editor is after reformatting ``test_trees.c``, leading to massive changes. Can you cut the diff down to the minimal changes here please?

hyanwong:
Whoops. I didn't notice that. Sorry. Not sure how that happened.

hyanwong:
Should be corrected now. Sorry about that.

c/tskit/core.c
```diff
@@ -198,6 +198,15 @@ tsk_strerror_internal(int err)
         case TSK_ERR_PROVENANCE_OUT_OF_BOUNDS:
             ret = "Provenance out of bounds";
             break;
+        case TSK_ERR_NODE_TIME_NONFINITE:
```
jeromekelleher: We're skipping migrations here, which have left, right and time float values IIRC. It's a bit painful to add two more errors just for this. Maybe `TSK_ERR_TIME_NONFINITE` and `TSK_ERR_COORDS_NONFINITE` and then these can be used in the migrations. Or, if you prefer to add two additional errors then that's fine too.

c/tskit/tables.c
```diff
@@ -5600,9 +5601,9 @@ tsk_table_collection_check_integrity(tsk_table_collection_t *self, tsk_flags_t o
 {
     int ret = TSK_ERR_GENERIC;
     tsk_size_t j;
-    double left, right, position;
+    double time, left, right, position;
     double L = self->sequence_length;
-    double *time = self->nodes.time;
+    double *times = self->nodes.time;
```
jeromekelleher: ``node_time`` here would be more consistent with other usage I think (but not a strong opinion, there's no right answer here)

c/tskit/tables.c
```diff
@@ -6005,7 +6029,7 @@ tsk_table_collection_build_index(tsk_table_collection_t *self, tsk_flags_t TSK_U
 {
     int ret = TSK_ERR_GENERIC;
     size_t j;
-    double *time = self->nodes.time;
+    double *times = self->nodes.time;
```
jeromekelleher: This has nothing to do with the changes in question --- please keep changes focused on a specific thing and avoid 'drive-by' changes like this. The singular usage is deliberate here and consistent elsewhere.

c/tskit/core.c
hyanwong: I've added migrations. The error is now `TSK_ERR_GENOME_COORDS_NONFINITE` so that we don't get confused with other sorts of coordinates (e.g. lat/long), and to make the location error completely clear and more general in the same way, I've changed `TSK_ERR_INDIVIDUAL_LOCATION_NONFINITE` to `TSK_ERR_SPATIAL_LOCATION_NONFINITE`.

https://github.com/tskit-dev/tskit/blob/7656bedfd29e46f450ed99f9fe17494e805d39ed/c/tskit/tables.c#L5759 has a `TODO it's a bit misleading to use the edge-specific errors here.` - but it seems like we are now happy using errors like `TSK_ERR_LEFT_LESS_ZERO` for genome coords in general. So should we remove the TODO, or maybe rename the error to something more verbose like `TSK_ERR_LEFT_GENOME_COORD_LESS_ZERO`?

c/tskit/tables.c
hyanwong: I've reverted back to `time` for an array of multiple times, as you had, and used `node_time` for the single instance variable. I'm not sure if this is what you want, given the comment on line 5605? Perhaps it should be the other way around?

c/tskit/tables.c
hyanwong: It seems (from below, line 6008) that you want to keep `time` (singular) for this array of times?

c/tskit/tables.c
```diff
@@ -5622,8 +5623,21 @@ tsk_table_collection_check_integrity(tsk_table_collection_t *self, tsk_flags_t o
         goto out;
     }
 
+    /* Individuals */
+    for (j = 0; j < self->individuals.location_offset[self->individuals.num_rows]; j++) {
+        if (! isfinite(self->individuals.location[j])) {
+            ret = TSK_ERR_SPATIAL_LOCATION_NONFINITE;
```
petrelharp: I do not think we should be throwing an error if spatial locations are not finite. For tskit, the spatial locations are just metadata, really, and we don't have any business enforcing anything there. (For instance, maybe someone wants to use inf as a special value...)

hyanwong: I defer to https://github.com/tskit-dev/tskit/issues/293#issuecomment-520066132

petrelharp: Whoops, sorry, missed that in the skim of this thread. =)

Ok, fine. We don't do anything with locations at the moment, but I can imagine using them in the future...

jeromekelleher: Yeah, the whole point of keeping locations as a column rather just metadata was that we might want to do calculations on them at some point, so we should enforce that they make sense here.

c/tskit/tables.c
jeromekelleher: This is fine --- as I say, there' no right answer here.

jeromekelleher:
Great stuff, thanks @hyanwong.

hyanwong:
Shall I remove the [TODO](https://github.com/tskit-dev/tskit/blob/7656bedfd29e46f450ed99f9fe17494e805d39ed/c/tskit/tables.c#L5759)? (https://github.com/tskit-dev/tskit/pull/294#discussion_r312732659)

jeromekelleher:
Let's not worry about it for now.

PR #302, opened 2019-08-13
Add individuals argument to write_vcf

### LINKED ISSUE #301: provide option to specify which individuals to output to VCF?
Currently, running `write_vcf( )` on a tree sequence that has individuals whose nodes are not all samples produced this error:
```
  _tskit.LibraryError: Cannot generate genotypes for non-samples unless missing data imputation is enabled
```
And, there is not actually an easy way to remove the non-sample individuals: even doing `ts.simplify(filter_individuals=True)` will not remove an individual if one of their nodes is retained through simplification.

For a MWE, run [this SLiM script](https://github.com/tskit-dev/pyslim/blob/master/tests/examples/recipe_nonWF.slim) and then
```
import pyslim, tskit
ts = pyslim.load("recipe_nonWF.trees")
rts = ts.recapitate(1e-2).simplify()
with open("test.vcf", "w") as f:
   rts.write_vcf(f)
# _tskit.LibraryError: Cannot generate genotypes for non-samples unless missing data imputation is enabled
```

Maybe the best fix would be to let the users decide which individuals to output. This is something that people will want to do anyhow.  We should also provide an easy way to tell if an individual has all sample nodes (e.g. by providing a `sample` property of the individual?).  Other options would be to (a) default to only outputting "sample" individuals, with a warning if this omits some, or (b) providing a method to get rid of the extra individuals. Maybe we want to do (b) anyhow, but I don't think we should do (a) because people will be confused about the discrepancy.

Hm: it occurs to me that in practice, "which nodes are in individuals" is probably in user's heads taking the place of "samples". Something to keep in mind.

petrelharp:
Closes #301 - how's this look? Seems pretty straightforward?

### DIFF
--- a/python/tskit/vcf.py
+++ b/python/tskit/vcf.py
@@ -50,10 +50,14 @@ class VcfWriter(object):
     file-like object.
     """
     def __init__(
-            self, tree_sequence, ploidy=None, contig_id="1", individual_names=None,
-            position_transform=None):
+            self, tree_sequence, ploidy=None, contig_id="1", individuals=None,
+            individual_names=None, position_transform=None):
         self.tree_sequence = tree_sequence
         self.contig_id = contig_id
+
+        if individuals is None:
+            individuals = np.arange(tree_sequence.num_individuals, dtype=int)
+        self.individuals = individuals
 
         self.__make_sample_mapping(ploidy)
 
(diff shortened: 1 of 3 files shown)


petrelharp:
If it looks good I can put in some tests...

python/tskit/vcf.py
```diff
@@ -50,17 +50,21 @@ class VcfWriter(object):
...
     def __init__(
-            self, tree_sequence, ploidy=None, contig_id="1", individual_names=None,
-            position_transform=None):
+            self, tree_sequence, ploidy=None, contig_id="1", individuals=None,
+            individual_names=None, position_transform=None):
         self.tree_sequence = tree_sequence
         self.contig_id = contig_id
 
+        if individuals is None:
+            individuals = list(range(self.tree_sequence.num_individuals))
```
jeromekelleher: ``np.arange(tree_sequence.num_individuals, dtype=int)`` would be a bit more efficient here.

jeromekelleher:
Yes, this is good I think. Thanks!

python/tskit/vcf.py
petrelharp: Still not sure whether to always numpy or only when necessary...

petrelharp:
Here are some tests; still should test that the actual correct sequence is output, somehow.

petrelharp:
I think this is good to go.

petrelharp:
ok for realz now

python/tskit/vcf.py
jeromekelleher: The rule I'm using at the moment is, use numpy *except* if the numpy version is hard to read and the perf implications aren't serious (the reason I didn't use your numpy version of the positions transform is that it didn't have exactly the same semantics, and it didn't seem that important, given limited time). Here, arange is exactly as readable as list(range()), so no reason not to use numpy.

jeromekelleher:
Looks good, thanks @petrelharp.

Issue #56, opened 2017-12-15
Implement __reversed__() methods for ts.nodes() (and others if appropriate)


hyanwong:
If this is easy, it would be useful to be able to iterate over e.g. nodes in a tree sequence in forward time as well as backwards time, like
`for n in reversed(ts.nodes()):`

I guess this might also be useful for ts.edges() and the like, but that might turn out to be difficult.

PR #304, opened 2019-08-14
Allow some basic iterators like nodes() and edges() to be reversed

hyanwong:
fixes https://github.com/tskit-dev/tskit/issues/56

I'm not sure if this is overly complicating things (so I haven't written unit tests yet), but it seems like a worthwhile addition. Iterating in reverse over migrations, mutations and edge_diffs is a little less easy to generalize, though (I can imagine the migrations version being useful).

I'm unsure of the right place to put the `_ReversibleIterator` class, though. See https://github.com/tskit-dev/tsinfer/pull/178

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1272,6 +1272,35 @@ class TestTreeSequence(HighLevelTestCase):
                     self.assertEqual(t3.interval, t2.interval)
                     self.assertEqual(t3.parent_dict, t2.parent_dict)
 
+    def test_sequence_iteration(self):
+        for ts in get_example_tree_sequences():
+            for table_name, _ in ts.tables:
+                sequence = getattr(ts, table_name)()
+                length = getattr(ts, "num_" + table_name)
+                # Test __iter__
+                for i, n in enumerate(sequence):
+                    self.assertEqual(i, n.id)
+                self.assertEqual(n.id, length - 1 if length else 0)
+                if table_name == 'mutations':
+                    # Mutations are not currently sequences, so have no len or idx access
+                    self.assertRaises(TypeError, len, sequence)
+                    if length != 0:
+                        with self.assertRaises(TypeError):
+                            sequence[0]
+                else:
+                    # Test __len__
+                    self.assertEqual(len(sequence), length)
+                    # Test __getitem__ on the last item in the sequence
+                    if length != 0:
+                        self.assertEqual(sequence[length - 1], n)  # +ive indexing
+                        self.assertEqual(sequence[-1], n)          # -ive indexing
+                    with self.assertRaises(IndexError):
+                        sequence[length]
+                    # Test reverse
+                    for i, n in enumerate(reversed(sequence)):
+                        self.assertEqual(i, length - 1 - n.id)
+                    self.assertEqual(n.id, 0)
+
 
 class TestFileUuid(HighLevelTestCase):
     """
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -296,7 +296,8 @@ class Migration(SimpleContainer):
     :ivar time: The time at which this migration occured at.
     :vartype time: float
     """
-    def __init__(self, left, right, node, source, dest, time):
+    def __init__(self, left, right, node, source, dest, time, id_=None):
+        self.id = id_
         self.left = left
         self.right = right
         self.node = node
@@ -430,7 +431,7 @@ def add_deprecated_mutation_attrs(site, mutation):
     """
     Add in attributes for the older deprecated way of defining
     mutations. These attributes will be removed in future releases
-    and are deliberately undocumented in version 0.5.0.
+    and are deliberately undocumented in tskit v0.2.2.
     """
     mutation.position = site.position
     mutation.index = site.id
@@ -2159,6 +2160,27 @@ class TreeIterator(object):
         return self.tree.tree_sequence.num_trees
 
 
+class SimpleContainerSequence(object):
+    """
+    Simple wrapper to allow arrays of SimpleContainers (e.g. edges, nodes) that have a
+    function allowing access by index (e.g. ts.edge(i), ts.node(i)) to be treated as a
+    python sequence, allowing forward and reverse iteration.
+    """
+    def __init__(self, getter, length):
+        self.getter = getter
+        self.length = length
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, index):
+        if index < 0:
+            index += len(self)
+        if index < 0 or index >= len(self):
+            raise IndexError("Index out of bounds")
+        return self.getter(index)
+
+
 class TreeSequence(object):
     """
     A single tree sequence, as defined by the :ref:`data model <sec_data_model>`.
@@ -2584,42 +2606,39 @@ class TreeSequence(object):
 
     def migrations(self):
         """
-        Returns an iterator over all the
+        Returns an iterable sequence of all the
         :ref:`migrations <sec_migration_table_definition>` in this tree sequence.
 
         Migrations are returned in nondecreasing order of the ``time`` value.
 
-        :return: An iterator over all migrations.
-        :rtype: :class:`collections.abc.Iterable`, :class:`Migration`
+        :return: An iterable sequence of all migrations.
+        :rtype: Sequence(:class:`.Migration`)
         """
-        for j in range(self._ll_tree_sequence.get_num_migrations()):
-            yield Migration(*self._ll_tree_sequence.get_migration(j))
+        return SimpleContainerSequence(self.migration, self.num_migrations)
 
     def individuals(self):
         """
-        Returns an iterator over all the
+        Returns an iterable sequence of all the
         :ref:`individuals <sec_individual_table_definition>` in this tree sequence.
 
-        :return: An iterator over all individuals.
-        :rtype: :class:`collections.abc.Iterable`, :class:`Individual`
+        :return: An iterable sequence of all individuals.
+        :rtype: Sequence(:class:`.Individual`)
         """
-        for j in range(self.num_individuals):
-            yield self.individual(j)
+        return SimpleContainerSequence(self.individual, self.num_individuals)
 
     def nodes(self):
         """
-        Returns an iterator over all the :ref:`nodes <sec_node_table_definition>`
+        Returns an iterable sequence of all the :ref:`nodes <sec_node_table_definition>`
         in this tree sequence.
 
-        :return: An iterator over all nodes.
-        :rtype: :class:`collections.abc.Iterable`, :class:`Node`
+        :return: An iterable sequence of all nodes.
+        :rtype: Sequence(:class:`.Node`)
         """
-        for j in range(self.num_nodes):
-            yield self.node(j)
+        return SimpleContainerSequence(self.node, self.num_nodes)
 
     def edges(self):
         """
-        Returns an iterator over all the :ref:`edges <sec_edge_table_definition>`
+        Returns an iterable sequence of all the :ref:`edges <sec_edge_table_definition>`
         in this tree sequence. Edges are returned in the order required
         for a :ref:`valid tree sequence <sec_valid_tree_sequence_requirements>`. So,
         edges are guaranteed to be ordered such that (a) all parents with a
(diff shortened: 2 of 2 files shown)


hyanwong:
I've obviously done this wrong because I don't have a `__next__` method on the iterator. Are any Python gurus here able to point me in the right direction to get `reversed(ts.nodes())` and friends working? Or perhaps it's not worth the effort.

jeromekelleher:
This is nice, and we could get it to work easily enough. However, I don't think we should unless someone is going to actually use it. I think a basic rule we need to have for features is, we don't add them unless either one of us (developers) needs it, or we've had a direct feature request from a user for it. 

These things can all be accessed efficiently enough using the Tables API and numpy, so I don't really see what we gain from it.

hyanwong:
There are certainly places in wilder's code where we iterate though nodes forward in time, and then backwards in time. But he generally uses the tables directly. I can definitely see reversing `nodes()` and `migrations()` being useful in stuff I write.

A more difficult thing would be the ability to iterate over edges in child time order, rather than parent time order. This is something I've had to hack round in the past.

hyanwong:
> This is nice, and we could get it to work easily enough. However, I don't think we should unless someone is going to actually use it. I think a basic rule we need to have for features is, we don't add them unless either one of us (developers) needs it, or we've had a direct feature request from a user for it.
> 
> These things can all be accessed efficiently enough using the Tables API and numpy, so I don't really see what we gain from it.

True, but pretty much *everything* can be accessed using the Tables API. In my mind the question is whether we are encouraging users to write easily readable code with the `tskit` object calls, or if we are trying to encourage them to go via the tables API. I often have the option of doing either, but try to pick the one that is simpler and more Pythonic. Hence why I'd prefer to use `reversed()` on my iterators, rather than look up how/whether I can efficiently implement my own version.

jeromekelleher:
Well if you have an application, then that's different --- the first rule applies. This is quite a simple change which is useful, so let's implement it. You should be able to model it on the [TreeIterator](https://github.com/tskit-dev/tskit/blob/master/python/tskit/trees.py#L1958) class which does this for ``reversed(ts.trees())``.

hyanwong:
I'll check with Wilder if he would use this too. But for the moment I'm having problems getting an iterator class to pass the tests with `write_vcf`: I think it must be doing something funky with the generators which I don't understand.

jeromekelleher:
Looks like you need to rebase, that might fix your problems with VCF tests.

jeromekelleher:
The rebasing is a side issue -- the diff includes changes the ``write_vcf`` function, which shouldn't be there (probably snuck in during a rebase). If you get rid of these it should work.

hyanwong:
Oh yes - well spotted. Sorry not to have seen that myself.

python/tskit/trees.py
```diff
@@ -1981,6 +1981,23 @@ def __next__(self):
...
+    def __init__(self, getter, getter_length):
+        self.getter = getter
+        self.getter_length = getter_length
+
+    def __iter__(self):
+        for j in range(self.getter_length):
+            yield self.getter(j)
+
+    def __reversed__(self):
+        for j in reversed(range(self.getter_length)):
```
jeromekelleher: I'm sure it does, but want to be certain: does ``reversed(range())`` do it without making an intermediate list?

hyanwong:
```
>>> reversed(range(10))
<range_iterator object at 0x10470ac00>
```
So yes. But at the moment the `_ReversibleIterator` doesn't allow `next()`, so we have lost some functionality here...

jeromekelleher: Can you give an example of next not doing the right thing here?

jeromekelleher: Right, I see the problem. You can't use 'yield' here, you have to return the next value from``__next``. This means you have to store the index and direction as state and increment/decrement explicity. See the TreeIterator, for example.

jeromekelleher: So, we need tests to make sure this is following the protocol correctly also.

jeromekelleher:
Where are we at with this @hyanwong? I think the changes needed are easy --- just get rid of the generators and add tests that the protocol is being correctly implemented.

hyanwong:
I guess if we want `__next__` behaviour, I'll need to store an index. Just wanted to check this was OK as an addition to the code.

hyanwong:
I've just amended the PR to use `__getitem__` and `__len__`, which automatically then allows this to be reversed. I think this is a nicer generalization, and more readable to boot. 

I'm not sure, however, if I should reset `self.index` within `__iter__()`, or only on `__init__()`?

Tests to follow.

python/tskit/trees.py
```diff
@@ -1981,6 +1981,33 @@ def __next__(self):
...
+    def __len__(self):
+        return self.getter_length
+
+    def __getitem__(self, i):
+        return self.getter(i)
+
+    def __next__(self):
+        try:
+            element = self[self.index]
+        except IndexError:
```
jeromekelleher: This depends on the getters always returning an index error, which I'm not sure we want to guarantee long term. It would be safer to check the bounds explicitly I think. What about negative indexes then? This stuff is quite subtle, we need to be 100% sure that we're implementing these protocols correctly.

python/tskit/trees.py
```diff
@@ -1981,6 +1981,33 @@ def __next__(self):
         return self.tree
 
 
+class _ReversibleIterator(object):
+    """
+    Simple class to provide a reversible iterator over getters like ts.node() & ts.edge()
+    """
+    def __init__(self, getter, getter_length):
```
jeromekelleher: ``length`` seems like a better name here.

python/tskit/trees.py
```diff
@@ -2421,8 +2448,7 @@ def individuals(self):
         :return: An iterator over all individuals.
         :rtype: iter(:class:`.Individual`)
         """
-        for j in range(self.num_individuals):
-            yield self.individual(j)
+        return _ReversibleIterator(self.individual, self.num_individuals)
```
jeromekelleher: This changes the semantics quite a lot though. Now, you could write 
```
ts.individuals()[0] 
```
It's not an iterator any more, it's a Sequence. Not that this is necessarily bad, it's just **different** and we need to spend time thinking through the consequences when we change the semantics of things.

hyanwong:
> I would prefer to keep the semantics of these functions as iterators, if just to reduce the amout of time we need to spend thinking about this now.

I suspect the way I coded it is the most useful, but happy to punt this (and any thinking required) until after 0.2.0 (is there a label or milestone for that)?

jeromekelleher:
https://github.com/tskit-dev/tskit/milestone/5

hyanwong:
Revisiting this, @awohns wants to be able to iterate over nodes and access all the edges for that node:

1. getting all the edges where that node is a *parent*, and always visiting *children before parents* (this is the normal order for a sorted edge table, so this is simply done using `itertools.groupby` on the edge table, grouping by parent)
2.  getting all the edges where that node is a *parent*, and always visiting *parents before children* (as for 1., but simply in reverse order - which is what this issue solves)
3.  getting all the edges where that node is a *child*, and always visiting *children before parents* (this requires re-ordering the edges by time of child node)
4.  getting all the edges where that node is a *child*, and always visiting *parents before children* (a reversed version of 3.)

These seem like generally useful ARG traversal methods, but 3. and 4. require some thought. Is this the sort of thing that we should code in an efficient way for end-users? It might be a simple modification of the sorting method on edge tables?

jeromekelleher:
After thinking about this for a bit, having the ``nodes``, ``edges`` etc return ``Sequences`` seems like a good idea. We can add a new class ``RowSequence`` (or something) which does what we want. However, we **could** just return the corresponding table object, which would basically do what we want anyway (right?). So,  ``TreeSequence.edges()`` would just ``return self.tables.edges``. The table object implements all the necessary stuff for iterating and accessing the rows by index. So, what's the point in implementing a whole bunch of new stuff?

Can you try this out please @hyanwong, and see how much it breaks?

>  getting all the edges where that node is a child, and always visiting children before parents (this requires re-ordering the edges by time of child node)
> getting all the edges where that node is a child, and always visiting parents before children (a reversed version of 3.)

These are separate issues, which would require a ``sort`` method on the EdgeTable or something. We should open a different issue to track this.

hyanwong:
> So, `TreeSequence.edges()` would just `return self.tables.edges`. The table object implements all the necessary stuff for iterating and accessing the rows by index. So, what's the point in implementing a whole bunch of new stuff?

Very good point. I agree. The only (slightly) confusing thing is that e.g. the tables.edges iterator returns a namedtuple of type `EdgeTableRow()`, so ts.edges() would return `EdgeTableRow()` objects too, rather than `tskit.trees.Edge()` objects as it does at present.

Perhaps we could make the `EdgeTableRow()` object redundant, and simply return an edge object all the time?

> Can you try this out please @hyanwong, and see how much it breaks?

Will do.

jeromekelleher:
> Perhaps we could make the EdgeTableRow() object redundant, and simply return an edge object all the time?

Yes, that would be ideal. There may be some additional fields we're filling in for the tree sequence objects though. Let's see how it works out.

hyanwong:
> Let's see how it works out.

Right. If we can re-use the SimpleContainer subclasses for the table rows too, I'd suggest extracting those into their own file (e.g. `containers.py`) which gets used both in trees.py and tables.py, as it's a little weird for them to stay in tree.py. But I'll check whether this is an appropriate solution first.

python/tskit/trees.py
hyanwong: We should never get a negative self.index, as we are starting at 0 and incrementing upwards on each `next` loop.

jeromekelleher: We don't need ``__next__`` at all, I think. Just implenent ``__len__`` and `__getitem__`` like [here](https://github.com/tskit-dev/tskit/blob/master/python/tskit/tables.py#L151) and delete all the other special methods. This should do everything we need.

Also need some tests of using negative indexes on the returned sequences then, of course.

hyanwong:
(I clicked the wrong button and accidentally closed this)

python/tskit/trees.py
hyanwong: OK. Happy to do that, although it does make this a breaking change, if anyone was previously relying on the `next()` functionality

python/tskit/trees.py
```diff
@@ -1986,6 +1987,23 @@ def __next__(self):
...
+    python Sequence object, allowing forward and reverse iteration.
+    """
+    def __init__(self, getter, length):
+        self.getter = getter
+        self.length = length
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, i):
```
jeromekelleher: Can you implement negative indexing please? Just copy the code from BaseTable.

jeromekelleher:
> Adding a call to ts.site(s) every time we call ts.mutation() might risk slowing things down a bit?

I wouldn't worry too much about that (there's not really much good reason for iterating over mutations without sites). What about circular references though? Mutation has a reference to its site, which presumably has a list of mutations? What do we do at the moment?

Maybe it is best to make an exception for mutations and just leave them as they are.

jeromekelleher:
Looks like this is more or less ready to go except we need to implement negative indexing. Can you finish this up please @hyanwong?

hyanwong:
Added the negative indexes. Guess I should squash too

python/tests/test_highlevel.py
```diff
@@ -1272,6 +1272,28 @@ def test_at(self):
                     self.assertEqual(t3.interval, t2.interval)
                     self.assertEqual(t3.parent_dict, t2.parent_dict)
 
+    def test_sequence_iteration(self):
```
jeromekelleher: I'm not seeing much in the way of testing negative indexes.

python/tests/test_highlevel.py
```diff
@@ -1272,6 +1272,28 @@ def test_at(self):
...
+        for ts in get_example_tree_sequences():
+            for table_name, _ in ts.tables:
+                sequence = getattr(ts, table_name)()
+                length = getattr(ts, "num_" + table_name)
+                # Test __iter__
+                for i, n in enumerate(sequence):
+                    self.assertEqual(i, n.id)
+                self.assertEqual(n.id, length - 1 if length else 0)
+                if table_name != 'mutations':
```
jeromekelleher: We should test something about the mutation table. Does len() fail, etc?

python/tests/test_highlevel.py
hyanwong: Added a trivial one for the last value. I guess that's enough?

python/tests/test_highlevel.py
hyanwong: Tested for failure of len() and indexing. Testing for access by index (square braces) is within a context manager.

hyanwong:
> Needs more tests.

Done now. Hopefully OK.

PR #305, opened 2019-08-15
Allow only some columns to be amended when setting all columns

hyanwong:
There are many cases in the test suites (a few changed in this commit), and I can think of some in the tskit code itself, where we want to alter only some columns in a table, but leave the rest of the columns intact. This commit allows us to do that without having to laboriously list all the column names by hand, which I find to be fragile.

The particular application I'm thinking of here is `ltrim()` (in another PR), which changes the positions of sites but leaves the other sites fields untouched. But I see `amend_columns()` as reducing code duplication all over the codebase.

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -1339,12 +1339,12 @@ class TestTablesToTreeSequence(unittest.TestCase):
 
         tables = ts.dump_tables()
         bad_coords = tables.edges.left + float('inf')
-        tables.edges.set_columns(**dict(tables.edges.asdict(), left=bad_coords))
+        tables.edges.amend_columns(left=bad_coords)
         self.assertRaises(_tskit.LibraryError, tables.tree_sequence)
 
         tables = ts.dump_tables()
         bad_coords = tables.edges.right + float('nan')
-        tables.edges.set_columns(**dict(tables.edges.asdict(), right=bad_coords))
+        tables.edges.amend_columns(right=bad_coords)
         self.assertRaises(_tskit.LibraryError, tables.tree_sequence)
 
     def test_bad_site_positions(self):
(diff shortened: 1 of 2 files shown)


jeromekelleher:
This is a good idea, but I think that most of the time we'll just want to set *one* column, so we should try to solve #12 instead. This would be much more intuitive and generally useful.

hyanwong:
But this *does* do that, I would say. And as an advantage you can set *any* number of columns just by using the right keyword:

```
tables = ts.dump_tables()
tables.edges.amend_columns(left=new_np_array)
```

hyanwong:
Oh, I see you mean it's more intuitive simply to set using `=` (and property magic) rather than using a function. I see.

hyanwong:
Well, if you don't mind *some* duplication, you can relatively easily use the `amend_columns()` function defined here to replace this:

```
    @individual.setter
    def individual(self, individual):
        self.set_columns(
            flags=self.flags, time=self.time, population=self.population,
            metadata=self.metadata, metadata_offset=self.metadata_offset,
            individual=individual)
```

with this

```
    @individual.setter
    def individual(self, individual):
        self.amend_columns(individual=individual)
```

so a little less fragile, but still an extra 3 lines for every property. on every table. Otherwise you probably want a custom descriptor.

jeromekelleher:
> Oh, I see you mean it's more intuitive simply to set using = (and property magic) rather than using a function. I see.

Exactly, yeah. I think we can do something with the descriptor protocol now, should be possible.

hyanwong:
I've just asked the internet: https://stackoverflow.com/questions/57512076/using-descriptors-in-a-superclass-to-avoid-code-duplication-in-subclasses

jeromekelleher:
I'm pretty sure I can get this working. Mind if I have a go?

hyanwong:
Please do. I suspect it might be useful to keep the `amend_columns` functionality in the background? And I guess it might be useful for changing e.g. the metadata & metadata_offset columns simultaneously (assuming this sort of functionality would be useful)

jeromekelleher:
I don't really like the ``amend_columns`` thing though --- it's confusing to have a different name for what is essentially ``set_columns`` but where we have unused columns defaulting to ``None``. That's the right way to update the API I think --- I totally agree that the operation is necessary though.

hyanwong:
Ah, well, if you want to change `set_columns` so that `None` values mean "leave this column as-is", rather than "set this column to Null", then that would be fine too, but I assumed you specifically wanted the latter behaviour.

jeromekelleher:
Yes, I'd forgotten that that was the semantics there. I'll look into the setter thing.

hyanwong:
Essentially obsoleted by https://github.com/tskit-dev/tskit/pull/307

PR #306, opened 2019-08-15
Remove code bloat in table.copy()

hyanwong:
I see no reason to keep repeating the table column names when we have a perfectly good function to do that for us for each table.

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -1949,3 +1949,8 @@ class TestBaseTable(unittest.TestCase):
         t = tskit.BaseTable(None, None)
         with self.assertRaises(NotImplementedError):
             t.asdict()
+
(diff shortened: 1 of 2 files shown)


hyanwong:
In fact, if we wanted to, we could remove all these `.copy()` methods in the subclasses and just place this in the BaseTable class, couldn't we?

```
    def copy(self):
        copy = self.__class__()
        copy.set_columns(**self.asdict())
        return copy
```

jeromekelleher:
Well spotted, thanks, this is great. The BaseTable way would be even better, if it works.

hyanwong:
It seems to work fine - squashed commit above. I guess this can be merged.

jeromekelleher:
Excellent, thanks.

PR #307, opened 2019-08-15
Generic column access.

### LINKED ISSUE #12: Method to set single column in table
It's tedious and error prone calling, ``x_table.set_columns(c1=x_table.c1, ..)`` when you only want to change one column.

We can do things like 
```python
node_table.flags = np.zeros_like(node_table.flags)
```
to set the full flags column in one go using some property magic. Does this seem like a good idea @petrelharp, @molpopgen?

jeromekelleher:
Closes #12 

See also #305

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -314,9 +314,10 @@ class CommonTestsMixin(object):
                 with self.assertRaises(ValueError):
                     setattr(table, col, bad_data)
 
-        # Try to access a column that isn't there.
+        # Try to read a column that isn't there. (We can always write to new attributes
+        # in Python, so there's nothing to test in that case.)
         with self.assertRaises(AttributeError):
-            table.no_such_column *= 1
+            _ = table.no_such_column
 
     def test_defaults(self):
         table = self.table_class()
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -245,7 +245,6 @@ class CommonTestsMixin(object):
                     self.assertRaises(ValueError, table.set_columns, **kwargs)
                     self.assertRaises(ValueError, table.append_columns, **kwargs)
 
-    @unittest.skip("Fix or remove when column setter done. #492")
     def test_set_read_only_attributes(self):
         table = self.table_class()
         with self.assertRaises(AttributeError):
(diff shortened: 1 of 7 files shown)


jeromekelleher:
This adds a list ``column_names`` to the class definition, which makes the rest of the definitions straightforward. We can tidy up things quite a bit with this --- I'll finish it up over the next few days.

hyanwong:
Very neat! One question: what if you ask for e.g. `table.nodes.column_names`? Will that not call the `__getattr__()` method, and then look for a `column_names` value in `self.ll_table`?

I'm assuming that this isn't a problem for methods - i.e. python knows not to run `__getattr__` when trying to get a handle to a method, like `table.nodes.set_columns`?

jeromekelleher:
``__getattr__`` only gets called when other lookup methods fail, so it's nice and neat for this. Anything defined in the normal way won't hit ``__getattr__``.

hyanwong:
Yep - should have looked up before commenting. This is very neat indeed. But (as I edited above) perhaps it's worth put a check in `__getattr__` for the existence of the name in `self.column_names` before making the call to `ll_table`. I guess you are happy simply to raise the error in the plain `getattr(self.ll_table, name)` call. I can see 2 minor advantages to raising it explicitly, neither of which are terribly important - (1) the error is raised in the most obvious place (2) it's never possible to access stuff via an attribute that doesn't exist in `self.column_names`, so read and write access fields are always guaranteed to match. Otherwise it might be possible to read from columns that are not writeable or listed in `asdict()`.

petrelharp:
Daaamn, very nice.

jeromekelleher:
This is ready for review, can you take a look please @hyanwong? 

Note it's not quite as pleasing to use as you might imagine---you have to be careful to set a new array or it might not have the effect you expect. We'd need to update the array in place to really make the most of things, but this is tricky for low-level memory management reasons and I don't want to get into it. I'm not sure I want to document this interface yet, as a consequence.

python/tests/test_tables.py
```diff
@@ -255,12 +237,87 @@ def test_set_read_only_attributes(self):
...
+            del input_data[list_col.name]
+            del input_data[offset_col.name]
+
+        for col, data in input_data.items():
+            for bad_data in [[], data[:-1]]:
+                with self.assertRaises(ValueError):
+                    setattr(table, col, bad_data)
+
+        # Try to access a column that isn't there.
+        with self.assertRaises(AttributeError):
```
hyanwong: Should we try getting and setting separately - if you test both together like this then one could not raise an error whereas the other could, and you wouldn't spot it.

python/tests/test_topology.py
```diff
@@ -3767,12 +3760,9 @@ def verify_parents(self, ts):
...
         tables = ts.tables
         self.assertTrue(np.array_equal(parent, tables.mutations.parent))
-        mutations = tables.mutations
-        mutations.set_columns(
-            site=mutations.site, node=mutations.node,
-            derived_state=mutations.derived_state,
-            derived_state_offset=mutations.derived_state_offset)
-        self.assertTrue(np.all(mutations.parent == tskit.NULL))
+        tmp = np.zeros_like(tables.mutations.parent) - 1
+        tables.mutations.parent = tmp
```
hyanwong: Why the extra tmp variable here?

hyanwong:
Does this PR make the following issue obsolete https://github.com/tskit-dev/tskit/issues/53 ?

python/tests/test_topology.py
jeromekelleher: Good point, I was being over paranoid.

jeromekelleher: Good point, I was being too paranoid.

jeromekelleher:
> Does this PR make the following issue obsolete #53 ?

No, that's a seperate issue. There we could do things like ``tables.nodes.flags[:] = 0``, which would be much more efficient. But this is tricky, as I said.

python/tests/test_tables.py
jeromekelleher: We can always set a new attribute on Python objects, so ``table.no_such_column = 1`` will not raise an error. I've update the tests to make it clear that we're just failing on the read side though.

hyanwong: Oh, good point

hyanwong:
I find "packset" a bit of a confusing term (I didn't understand what it did at first). I wonder if we call it "setpack" instead, then it will seem more like `set_columns()` etc? Or maybe a different name (e.g. `set_location()` - is there any reason why the user needs to know that it does a packing stage first?). Or maybe it's just me.

hyanwong:
This all looks *great*. I can probably find more places where the tests can be simplified now (e.g. my hacks at https://github.com/tskit-dev/tskit/blob/d6d1d3bb8aa1d26213d5d4e8fbb726f25472683a/python/tests/test_tables.py#L1342 (and the following lines 1355, 1363, 1373)

jeromekelleher:
Well, ``set_location`` is definitely confusing as ``tables.location`` gives you the packed version. I ended up with ``packset_x`` as ``pack_and_set_x`` is too long and cumbersome. ``set_pack_x`` seems wrong, because that sounds like ``set_and_pack_x``, and this is the opposite way around to what is actually done. 

This is a pretty technical thing, people will need to understand how the packing scheme works to use tables like this at all. Having a bit of esoteric terminology is OK I think.

There's a bunch of places the tests could be updated, but I don't have time to do it.

hyanwong:
Yeah, I don't really have much of a better solution? `set_packed_location`?

jeromekelleher:
> Yeah, I don't really have much of a better solution? set_packed_location? Other suggestions?

``set_packed_location`` implies that it's already been packed (which it hasn't). I think ``packset_location`` is fine --- @petrelharp, any opinions here?

hyanwong:
How about ‘set_location_columns’? That way it’s clear we are setting multiple related columns at once in a non-trivial way.

petrelharp:
Well, at first I wanted to say that just `.set_metadata()` would be fine, since that's what it's doing. I like the name `.packset_metadata()`, though; the downside is that it's longer. The upside and downside of `packset` is that it reminds the user of the underlying format. That's not always good, but I think it is good to be reminded in this case, because it's really easy to start doing things like `table.metadata[2]` to get the third metadata string, which is *wrong*.

In short, after thinking this through, I vote for `packset_X`.

Also, all this is wonderful!

jeromekelleher:
OK, thanks @petrelharp. Merging.

PR #312, opened 2019-08-20
Removed old pairwise_diversity code.

### LINKED ISSUE #215: Deprecate TreeSequence.pairwise_diversity
Change this function to use the new diversity method, and mark it as deprecated.

Remove low-level C code.

jeromekelleher:
Deprecate the Python TreeSequence.pairwise_diversity method.

Closes #215.

### DIFF
--- a/c/dev-tools/dev-cli.c
+++ b/c/dev-tools/dev-cli.c
@@ -158,29 +158,6 @@ print_ld_matrix(tsk_treeseq_t *ts)
 }
 
 static void
-print_stats(tsk_treeseq_t *ts)
-{
-    int ret = 0;
-    uint32_t j;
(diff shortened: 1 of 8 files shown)


petrelharp:
LGTM!

jeromekelleher:
Actually, it turns out there was an error in here and it wasn't getting tested properly. The old version had ``span_normalise=False``, effectively, and I think the documentation was incorrect. Can you have another quick look please @petrelharp and see if you agree with the docs?

We should schedule this function for removal in 0.3 I guess --- it's just confusing keeping it lying around, and I doubt anyone much is using it.

petrelharp:
Ah, yep, you are right. It looks good now.

Issue #229, opened 2019-06-13
write examples of stats computations for tutorial


petrelharp:

jeromekelleher:
I've made an example using divergence and diversity [here](https://github.com/DRL/SMBE-SGE-2019/blob/master/Session_1/3.Processing_simulation_results.ipynb). Maybe we could reuse some of this?

jeromekelleher:
Closed in #313.

Issue #200, opened 2019-06-10
Default behaviour for stats dimension


jeromekelleher:
It's a bit inconvenient stripping off empty dimensions from numpy arrays when we do things like
```python
ts.diversity()
```
I think it might be nice if we automatically strip off empty dimensions in **default cases**. So, in this example, we'd have 
```
ts.diversity() = A number
ts.diversity(windows=[0, ts.sequence_length]) = [A number]
ts.diveristy([ts.samples()]) = [A number]
ts.diversity(windows=[0, ts.sequence_length]) = [A number]
ts.diveristy([ts.samples()], windows=[0, ts.sequence_length]) = [[A number]]
```

I think this will generalise to other stats as well. There's no amibiguity as the dimension stripping only happens on default inputs, and if a user wants the extra dimensions they can simply specify these default inputs themselves.

This is seems like a user-friendly option, as it removes a lot of tedious ``stat[0, 0]`` accesses when you just want a single value. What do you think @petrelharp?

petrelharp:
Gee, that's a nice, elegant idea. I like it. But: this is tricky. I agree that all the `[0][0]`s are tedious and counterintuitive but switching output types is also counterintuitive.

I surveyed some bioinformatics users here (Anastasia), who thinks that (a) people aren't actually going to want to use the defaults all that often, because with the scale of data you have in tree sequences, you'll almost always want across windows and between populations, and (b) the different dimensionalities is confusing. I tend to agree, I think? 

So, I vote to leave it always 2D arrays? (Hm, this proposal was probably made with the node arrays (3D) in mind.)

I think that if we do want to make this change, the first test should be to write the documentation for it (in `docs/stats.rst`), and if it's too confusing to explain then to ditch it.

jeromekelleher:
Closed in #313.

PR #313, opened 2019-08-20
Strip empty first dimension for default windows.

jeromekelleher:
This implements a subset of #200. The idea is that, when we don't specify windows then it's a pain to have to write ``stat[0]`` to get what you actually want. This seems like a good usablilty feature to add, to me --- this is what you'd actually want the library to do, right?

More concretely, take the following example:
```
afs = ts.allele_frequency_spectrum(mode="branch")
print("afs: ", afs)
print("diversity", ts.diversity([ts.samples()], mode="branch"))

afs = ts.allele_frequency_spectrum(mode="branch", windows=[0, ts.sequence_length])
print("afs:", afs)
print("diversity",
    ts.diversity([ts.samples()], mode="branch", windows=[0, ts.sequence_length]))
```
gives
```
afs:  [0.  2.5 0.  0. ]
diversity [3.33333333]
afs: [[0.  2.5 0.  0. ]]
diversity [[3.33333333]]
```

In the first case, we don't specify any windows so we're only interested in a single window and we remove the empty first dimension.  In the second case, we explicitly specify the windows and so we keep the dimension in place.

I agree this is going to be a bit confusing to explain, but it seems worth it. The mistakes that are made will be less annoying, in the long run, than having to add an extra ``[0]`` to the end of each call when you just want one window. If, as @petrelharp says people are rarely interested in a single window, then they won't be affected by this default behaviour.

The reason I'm bringing this up now is because it **is** a long-term decision. Because we've specified the default value for windows, we can't change the default behaviour after we ship 0.2.0. So we're making the decision here that we'll **never** want to do this, which to me seems a real shame as it's quite neat and elegant (IMO). 

Because we don't have a default value for, e.g., ``diversity`` now we don't have to worry about the other half of #200 (stripping off the empty dimensions when we're only looking at one sample set in, e.g., ``diversity`` until later).

FWIW, the implementation is easy and it doesn't break many of the tests.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -14,9 +14,8 @@ and of the underlying trees (whose definition does not depend on the genome sequ
 Furthermore, these statistics have a common interface to easily compute
 (a) averaged statistics in windows along the genome,
 and (b) summary statistics of many combinations of sets of samples simultaneously.
-All methods return numpy arrays,
-whose rows correspond to the windows along the genome,
-and whose remaining dimensions are determined by the statistic.
+All methods return numpy arrays whose dimensions are
+determined by the parameters (see :ref:`sec_general_stats_output_dimensions`).
 
 .. warning:: :ref:`sec_data_model_missing_data` is not currently
    handled correctly by site statistics defined here, as we always
@@ -70,12 +69,6 @@ e.g., the sites of the SNPs.
 Windowing
 *********
 
-By default, statistics
-
-``windows = None``
-   This is the default, and equivalent to passing ``windows = [0.0, ts.sequence_length]``.
-   The output will still be a two-dimensional array, but with only one row.
-
 Each statistic has an argument, ``windows``,
 which defines a collection of contiguous windows along the genome.
 If ``windows`` is a list of ``n+1`` increasing numbers between 0 and the ``sequence_length``,
@@ -94,6 +87,14 @@ would be equivalent to averaging the rows of ``S``,
 obtaining ``((b - a) * S[0] + (c - b) * S[1]) / (c - a)``.
 
 There are some shortcuts to other useful options:
+
+``windows = None``
+   This is the default and computes statistics in single window over the whole
+   sequence. As the first returned array contains only a single
+   value, we drop this dimension as described in the :ref:`output dimensions
+   <sec_general_stats_output_dimensions>` section. **Note:** if you really do
+   want to have an array with a single value as the result, please use
+   ``windows = [0.0, ts.sequence_length]``.
 
 ``windows = "trees"``
    This says that you want statistics computed separately on the portion of the genome
@@ -191,7 +192,9 @@ Here are some additional special cases:
    If the statistic takes ``k`` inputs for ``k > 1``,
    and there are exactly ``k`` lists in ``sample_sets``,
    then this will compute just one statistic, and is equivalent to passing
-   ``indexes = [(0, 1, ..., k-1)]``.
+   ``indexes = (0, 1, ..., k-1)``. Note that this also drops the last
+   dimension of the output, as described in the :ref:`sec_general_stats_output_dimensions`
+   section.
    If there are not exactly ``k`` sample sets, this will throw an error.
 
 ``k=1`` does not allow ``indexes``:
@@ -203,16 +206,16 @@ Here are some additional special cases:
    were that allowed.)
 
 
-.. _sec_general_stats_output:
+.. _sec_general_stats_output_format:
 
-******
-Output
-******
+*************
+Output format
+*************
 
 Each of the statistics methods returns a ``numpy`` ndarray.
 Suppose that the output is name ``out``.
-In all cases, the number of rows of the output is equal to the number of windows,
-so that ``out.shape[0]`` is equal to ``len(windows) - 1``
+If ``windows`` has been specified, the number of rows of the output is equal to the
+number of windows, so that ``out.shape[0]`` is equal to ``len(windows) - 1``
 and ``out[i]`` is an array of statistics describing the portion of the tree sequence
 from ``windows[i]`` to ``windows[i + 1]`` (including the left but not the right endpoint).
 
@@ -229,6 +232,9 @@ from ``windows[i]`` to ``windows[i + 1]`` (including the left but not the right 
 
 The final dimension of the arrays in other cases is specified by the method.
 
+Note, however, that empty dimensions can optionally be dropped,
+as described in the :ref:`sec_general_stats_output_dimensions` section.
+
 A note about **default values** and **division by zero**:
 Under the hood, statistics computation fills in zeros everywhere, then updates these
 (since statistics are all **additive**, this makes sense).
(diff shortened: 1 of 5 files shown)


docs/stats.rst
```diff
@@ -70,6 +70,12 @@ e.g., the sites of the SNPs.
 Windowing
 *********
 
+By default, statistics
```
jeromekelleher: I started writing this, and then decided to see how easy it would be to implement before getting into it.

petrelharp:
You are probably right. I do worry that it's strange that specifying a non-default argument changes the output dimensions, but at least `None` is a different argument. Ok, go for it.

Re: the other half; I don't think we need to drop dimensions for a default `indexes` (when there is one); since `np.array([2.0])` acts a lot like `2.0`.  But, do we want to set up default indexes for other stats?

jeromekelleher:
> You are probably right. I do worry that it's strange that specifying a non-default argument changes the output dimensions, but at least None is a different argument. Ok, go for it.

Hooray! OK, I'll flesh this out and ping you.

> Re: the other half; I don't think we need to drop dimensions for a default indexes (when there is one); since np.array([2.0]) acts a lot like 2.0. But, do we want to set up default indexes for other stats?

Hmm, that's a good question. I'm not sure --- maybe this is an argument for dropping the default indexes value for Fst until after 0.2.0? The way I'm thinking about it at the moment is that post 0.2.0 we can possibly refine the semantics at bit such that the dimensions of your arguments determines the dimensions of your output. So, for example, we'd have
```
ts.diversity([ts.samples()])   ->   [x]
ts.diversity(ts.samples())     ->   x
```
Here, the user is giving clear information that they're only interested in one sample set in the second example, so we strip off the first dimension and just return the value. Possibly, the right behaviour then is for ``ts.diversity()`` to return ``x`` also.  I'm not sure how this interacts with the ``indexes`` argument, but I'm not sure I want to tie us down to something without thinking it through properly either.

I know you have reservations about playing fast and loose with the dimensions like this, so I'm not saying this is something I'd definitely want to do. But, I don't definitely **not** want to do it either, so keeping the door open by not specifying default values for things we're not sure about might be good.

petrelharp:
Re: indexes - sounds good; also see #315 in the meantime.

jeromekelleher:
OK, turns out we couldn't really punt this one down the road and need to deal with it now. Here's my proposal for how we strip off empty dimensions @petrelharp --- the new tests should explain what it's doing. If we think this is a good idea I'll tidy up the rest of the tests and fix up the documentation.

I haven't looked at the derived stats like Fst, Tajimas D and the covariance stats etc. These may need to be treated differently.

python/tskit/trees.py
```diff
@@ -3438,20 +3438,39 @@ def parse_windows(self, windows):
...
+        stat = method(*args, **kwargs, windows=windows)
+        if strip_dim:
+            stat = stat[0]
+        return stat
+
+    def __one_way_sample_set_stat(
+            self, ll_method, sample_sets, windows=None, mode=None, span_normalise=True,
+            polarised=False):
+        if sample_sets is None:
+            sample_sets = self.samples()
```
petrelharp: shouldn't this be
```
  sample_sets = [self.samples()]
```

petrelharp: oh wait, I see...

python/tskit/trees.py
```diff
@@ -3438,20 +3438,39 @@ def parse_windows(self, windows):
...
+        # First try to convert to a 1D numpy array. If it is, then we strip off
+        # the corresponding dimension from the output.
+        drop_dimension = False
+        sample_sets = np.array(sample_sets)
+        if len(sample_sets.shape) == 1:
```
petrelharp: This won't do what you want, here: `len(np.array([[0,1,2], [3,4]]).shape)` is 1. But I get the idea.

petrelharp: If we did more serious parsing of `sample_sets` we could throw a more infromative error than the somewhat inscrutible error we currently get when passing in `sample_sets = [0,1,2]`, which is I think "integers have no len()".

python/tests/test_tree_stats.py
```diff
@@ -5040,6 +5043,243 @@ def f(x):
...
+        self.assertArrayEqual(x.reshape(ts.num_nodes, 1), y)
+
+    def test_one_way_stat_windows(self):
+        ts = self.get_example_ts()
+        L = ts.sequence_length
+        N = ts.num_nodes
+
+        windows = [0, L / 4, L / 2, L]
+        A = ts.samples()[:5]
+        B = ts.samples()[5:]
```
petrelharp: should make these of unequal length to catch the parsing error

petrelharp:
Ok, so in words, we drop a dimension if

1) it is a one-way stat and `sample_sets` is actually just a list of samples
2) it is a k>1-way stat and `indexes` is actually just a k-tuple of indexes
3) it is a k>1-way stat and `indexes=None` and there are exactly `k` sample sets

This is nice and elegant. I like it. It also makes the interface a bit more confusing, but I say we go for it.

One wrinkle is that in the future we will probably have statistics that can take variaible numbers of indexes, so that e.g. `indexes = [(0, 1, 2), (0, 1)]` would be valid. (For instance, a multi-pop Fst.) I think we can just adjust things at that point, but we should make sure we don't make this impossible.

Do you have an opinion about dropping dimensions when `windows=None`?

python/tests/test_tree_stats.py
```diff
@@ -5040,6 +5043,243 @@ def f(x):
...
+        self.assertArrayEqual(x.reshape(ts.num_nodes, 1), y)
+
+    def test_two_way_stat_windows(self):
+        ts = self.get_example_ts()
+        L = ts.sequence_length
+        N = ts.num_nodes
+
+        windows = [0, L / 4, L / 2, L]
+        A = ts.samples()[:5]
+        B = ts.samples()[5:]
```
petrelharp: again

petrelharp:
> I haven't looked at the derived stats like Fst, Tajimas D and the covariance stats etc. These may need to be treated differently.

They should work the same, but we might need different code to make that happen...

jeromekelleher:
> This is nice and elegant. I like it. It also makes the interface a bit more confusing, but I say we go for it.

Excellent!

> Do you have an opinion about dropping dimensions when windows=None?

Yes, this is already implemented --- we drop the first dimension when windows is None. See the ``_default_windows`` tests for examples.

petrelharp:
> Yes, this is already implemented --- we drop the first dimension when windows is None. See the _default_windows tests for examples.

Oh, duh. =) Yes, looks great!

python/tskit/trees.py
jeromekelleher: It would be nice to improve this all right.

jeromekelleher: Hmm, this is going to be hard to do well. Kicking down the road.

jeromekelleher: See #322

jeromekelleher:
I think we're nearly there with this @petrelharp. I've tidied up the loose ends, and written some high-level docs. I think I just need to make another pass through the methods, probably linking back to some general descipription of the behaviour or parameters one-way and k-way statistics to document the behaviour in various cases.

What do you think?

jeromekelleher:
Note to self: need to update the AFS values in the tutorial, as these are now wrong (dimension stripping).

Also, does this close #229? They are examples --- if somewhat boring ones.

docs/stats.rst
```diff
@@ -241,6 +249,46 @@ For ``branch`` statistics, any windows with **no branches** will similarly remai
...
+3. In k-way stats, if the ``indexes`` argument is a 1D array of length k
+   we intepret this as specifying a single statistic and drop the last
+   dimension of the array. If ``indexes`` is None (the default) and
+   there are k sample sets, we compute the statistic from these sample sets
+   and drop the last dimension.
+
+Rules 2 and 3 can be summarised by "the dimensions of the input determines
+the dimensions of the output". Note that dropping these dimensions is
+**optional**: it is always possible to keep the full dimensions of the
+output arrays.
```
petrelharp: Maybe add
```
For instance, `ts.diversity()` will return a single number (the mean sequence diversity of all samples),
while `ts.diversity(samples=[ts.samples()], windows=[0.0, ts.sequence_length])` will return the same number in a `1x1` numpy array.  Similarly, `ts.divergence([[0], [1]], indexes=[(0, 1)], windows=[0.0, ts.sequence_length])` returns the same thing as `np.array([[ts.divergence([[0], [1]])]])`.
```

petrelharp: Or maybe not, this is covered very well in the tutorial.

python/tskit/trees.py
```diff
@@ -3440,20 +3440,46 @@ def parse_windows(self, windows):
                                  "the only allowed strings are 'sites' or 'trees'")
         return np.array(windows)
 
-    ############################################
-    # Statistics definitions
-    ############################################
+    def __run_windowed_stat(self, windows, method, *args, **kwargs):
```
petrelharp: nice

python/tskit/trees.py
```diff
@@ -3975,32 +4014,39 @@ def Fst(self, sample_sets, indexes=None, windows=None, mode="site",
...
-                     + 2 * divergences[:, :, i])
-            with np.errstate(divide='ignore', invalid='ignore'):
-                fst[:, :, i] -= 2 * (diversities[:, :, u] + diversities[:, :, v]) / denom
-        fst.shape = orig_shape
-        return fst
+        # TODO this should really be implemented in C (presumably C programmers will want
+        # to compute Fst too), but in the mean time implementing using the low-level
+        # calls has two advantages: (a) we automatically change dimensions like the other
+        # two-way stats and (b) it's a bit more efficient because we're not messing
+        # around with indexes and samples sets twice.
```
petrelharp: very nice, yes

petrelharp:
This looks great, especially the tutorial. I think this closes #229, also.

docs/stats.rst
jeromekelleher: Added, ta.

jeromekelleher:
TODO:

- Fix general stats interface like in #328 (also move around code like there, it's good to have the stats functions together).
- Update documentation. Need to describe behaviour  of, e.g., windows, concisely and precisely somewhere, and put in links to this in each function.

petrelharp:
Most (all?) of those TODOs are now done over in #330; maybe you want to merge that one in here before continuing.

PR #314, opened 2019-08-20
add restrictions to summary functions

### LINKED ISSUE #238: additional restrictions on summary_func?

petrelharp:
This would close #238.

But, I think we should allow summary functions that don't satisfy this; either we should throw a warning (not an error); or we should add a `strict` option that if it is False doesn't do the check.  I'm going to implement the second thing.

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -1123,7 +1123,8 @@ class TestDiversity(StatsTestCase, SampleSetStatsMixin):
         n = np.array([len(x) for x in sample_sets])
 
         def f(x):
-            return x * (n - x) / (n * (n - 1))
+            with np.errstate(invalid='ignore', divide='ignore'):
+                return x * (n - x) / (n * (n - 1))
 
         self.verify_definition(
             ts, sample_sets, windows, f, ts.diversity, diversity)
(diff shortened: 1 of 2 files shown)


petrelharp:
Here's `strict`. I hate to add another option, but we could totally want to use other functions sometimes.

python/tests/test_tree_stats.py
```diff
@@ -3350,6 +3331,14 @@ def test_bad_window_strings(self):
         with self.assertRaises(ValueError):
             ts.diversity([list(ts.samples())], mode="tree", windows="abc")
 
+    def test_bad_summary_function(self):
+        ts = self.get_tree_sequence()
+        W = np.ones((ts.num_samples, 3))
+        with self.assertRaises(ValueError):
+            _ = ts.general_stat(W, lambda x: x, 3, windows="sites")
```
jeromekelleher: We don't need to assign this to ``_`` here I think as we're executing a function. I did this in the other weird example because I wasn't sure what would happen if you accessed an attribute but didn't assign it to any varible.

python/tests/test_tree_stats.py
```diff
@@ -3310,19 +3305,22 @@ def get_tree_sequence(self):
                               mutation_rate=2, random_seed=1)
         return ts
 
+    def identity_f(self, ts):
```
jeromekelleher: Ah, I thought I commented on this already, seems to have gotten lost. We're redefining these functions a bunch of times --- I think it'd be better to just define them at the top-level stating what they're for. We don't use any state from the class anyway.

petrelharp:
Fixed those, thanks! Can merge?

hyanwong:
I'm still getting `tskit/python/tests/test_tree_stats.py:1502: RuntimeWarning: invalid value encountered in true_divide` - was there one that you missed @petrelharp ?

petrelharp:
> was there one that you missed @petrelharp ?

Hmph, looks like it.

Issue #250, opened 2019-06-29
Fst gives a NoneType error when no indices are given (the default)


gtsambos:
I have a simulation that does the right thing when I explicitly specify pairs of indices:

```
ts.Fst(sample_sets = samples_list,
      indexes = [(0,2), (0,1)])
```
outputs
```
array([[0.12508121, 0.12780193]])
```

but that gives the following error if I remove the `indexes` input (and so use the default value of `None`)

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-fbfb01290311> in <module>
----> 1 ts.Fst(sample_sets = samples_list)

/usr/local/lib/python3.7/site-packages/tskit/trees.py in Fst(self, sample_sets, indexes, windows, mode, span_normalise)
   3332         fst = np.repeat(1.0, np.product(divergences.shape))
   3333         fst.shape = divergences.shape
-> 3334         for i, (u, v) in enumerate(indexes):
   3335             denom = (diversities[:, :, u] + diversities[:, :, v]
   3336                      + 2 * divergences[:, :, i])

TypeError: 'NoneType' object is not iterable
```

jeromekelleher:
See also https://github.com/tskit-dev/tskit/issues/199#issuecomment-501467165

jeromekelleher:
I was looking into this @petrelharp, and it seems we don't have a lot of test coverage on Fst right now --- we're only testing the "site" version, essentially. Since this is a derived statistic it seems overboard to implement the full thing from scratch like the other stats, but we do need to get some sort of coverage here on "branch" and "node". Have you any suggestions for this?

petrelharp:
I guess we just need to do some "does this run" sort of checks. I'll have a look.

petrelharp:
Currently our default for `indexes=None` is `indexes=[(0, 1)]` for pairwise stats. We can't do this here because users might think that the single Fst value returned is the between-all-pops Fst value. I also don't want to just return the full matrix by default because it's not clear what order it should be in. So, I'm making the user specify.

Also, I ran into some issues with poorly-handled bad input to `safe_np_int_cast` that I've addressed.

jeromekelleher:
> Currently our default for indexes=None is indexes=[(0, 1)] for pairwise stats. We can't do this here because users might think that the single Fst value returned is the between-all-pops Fst value. I also don't want to just return the full matrix by default because it's not clear what order it should be in. So, I'm making the user specify.

[...]

PR #315, opened 2019-08-20
check for error when indexes=None

petrelharp:
We *do* have sensible behavior when `indexes=None`, but we weren't checking that this made sense given the sample_sets passed in. This does that.

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -3188,13 +3188,13 @@ class TestSampleSets(StatsTestCase):
     def test_non_samples(self):
         ts = self.get_example_ts()
         with self.assertRaises(exceptions.LibraryError):
-            ts.diversity([[10]])
+            ts.diversity([[ts.num_samples]])
 
         with self.assertRaises(exceptions.LibraryError):
-            ts.divergence([[10], [1, 2]])
+            ts.divergence([[ts.num_samples], [1, 2]])
 
         with self.assertRaises(ValueError):
-            ts.sample_count_stat([[10]], self.identity_f(ts), 1)
+            ts.sample_count_stat([[ts.num_samples]], self.identity_f(ts), 1)
 
     def test_span_normalise(self):
         ts = self.get_example_ts()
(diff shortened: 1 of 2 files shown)


petrelharp:
Recap here: currently for all sample set statistics with k > 1, `indexes=None` gives you the same as `indexes=[(0, 1, ..., k)]`. I think this makes sense, but only if there are exactly `k` statistics. The change for #250 made Fst agree with this convention.

Over in #313, @jeromekelleher  wrote:

> maybe this is an argument for dropping the default indexes value for Fst until after 0.2.0? The way I'm thinking about it at the moment is that post 0.2.0 we can possibly refine the semantics at bit such that the dimensions of your arguments determines the dimensions of your output. So, for example, we'd have
> ```
> ts.diversity([ts.samples()])   ->   [x]
> ts.diversity(ts.samples())     ->   x
> ```
> Here, the user is giving clear information that they're only interested in one sample set in the second example, so we strip off the first dimension and just return the value. Possibly, the right behaviour then is for ts.diversity() to return x also. I'm not sure how this interacts with the indexes argument, but I'm not sure I want to tie us down to something without thinking it through properly either.

So, the question is, do we want to remove default indexes entirely, for now, to think this through better, later?

jeromekelleher:
This is sensible behaviour, I like it.

> So, the question is, do we want to remove default indexes entirely, for now, to think this through better, later?

I'll have a look tomorrow morning and come back with a proposal.

petrelharp:
The natural proposal is that if `indexes=None` then we drop a dimension. I've talked this through with @attrna and I think it makes sense.  So, the output would have `len(output.shape)` equal to `(indexes is None) + (windows is None) + (mode == "node")`.

My main worry was: Suppose you have a script that computes statistics over a bunch of different chocies of windows, or different choices of indexes. You do not want to get outputs of different dimensions. But, this could not happen, unless you put `None` in your list of choices of windows, for some reason.

In fact, this "dropping dimensions" would be somewhat helpful. Suppose that I want to plot a horizontal line at the genome-wide divergence value:
```
div = ts.divergence([[0, 1], [2,3]])
ax.axhline(div[0][0])
```
But, even though I know that `div` will be a `1 x 1` array, I worry that it's actually got more than one value that I'm ignoring.  However, if we drop dimensions automatically, I'd get to do just
```
ax.axhline(div)
```
and I know that my `[0][0]` isn't dropping any information.

jeromekelleher:
This is a sensible update, so I'm going to merge it. I'll then bring #313 up to date with the full proposal for dimension stripping and we can see where we are from there.

PR #317, opened 2019-08-21
Add note about derived_state to compute_mutation_parents

hyanwong:


python/tskit/tables.py
```diff
@@ -1637,6 +1637,14 @@ def compute_mutation_parents(self):
         The ``parent`` of a given mutation is the ID of the next mutation
         encountered traversing the tree upwards from that mutation, or
         ``NULL`` if there is no such mutation.
+
+        .. note:: Setting new parents for mutations is likely to result in the
```
jeromekelleher: This is confusing --- **if** you're using this to merge together mutations that haven't been generated with consistent state patterns you **may** end up in this situation. Mushing together mutations that have been generated without regard for the derived state values is an esoteric application, and if you do this you need to understand what it is you're doing. I think this note will do more harm than good.

Something like 
```note: This method does not take into account the actual allelic values in any way, and therefore the table may not satisfy the mutation table requirements even if this method complete successfully.
```

hyanwong: I guess I can't see any applications (apart from floating point mutation values) where you have generated consistent state patterns but then want to run `compute_mutation_parents()`. But I guess there must be some.

Either way, I think you need to mention that the requirement breaking will be specifically in the `derived_state` column (I can't see any situations where other columns would ever become invalid after running `compute_mutation_parents()`). So it is only this column that will ever need checking for validity.

jeromekelleher: > I guess I can't see any applications (apart from floating point mutation values) where you have generated consistent state patterns but then want to run compute_mutation_parents(). But I guess there must be some.

We added the function for exactly this application in SLiM.

> Either way, I think we need to mention that the requirement breaking will be specifically in the derived_state column (I can't see any situations where other columns would ever become invalid after running compute_mutation_parents())

See my text above. It's GIGO: if the table that you have specified has inconsistent alleles the output also will. That's all we have to say. You're text above is far too specific to the application that you're currently working on.

petrelharp: I remember discussing removing this requirement, and thought we did remove it. It is not mentioned [in the documentation](https://tskit.readthedocs.io/en/latest/data-model.html#sec-mutation-requirements), and it does not appear to be checked for [in the code](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L5763). Am I missing something?

jeromekelleher: It's checked when we're generating variants or haplotypes --- we get an inconsistent states error being raised.

hyanwong: @petrelharp: it is mentioned in the docs - the last para in the [section you linked to](https://tskit.readthedocs.io/en/latest/data-model.html#sec-mutation-requirements).

I agree, however, that this requirement might be too strict. I can see a use case for having "null" mutations that don't change the state at all, but are there e.g. for statistical reasons.

petrelharp: Sorry, hadn't drunk enough tea yet. =)

I'm not actually sure we need this warning, because anyone using this method really should have read that section of the docs. But if we have it, then it should be a bit more specific, maybe:
```
note: This method does not check that in the resulting table all mutations result in a change of state, as required; see <link to the docs>.
```

jeromekelleher:
Can you update this with a less specific message please @hyanwong (or close)?

hyanwong:
> Can you update this with a less specific message please @hyanwong (or close)?

@jeromekelleher: are you happy with @petrelharp 's wording above?

PR #320, opened 2019-08-21
Allow Edges to return an ID

hyanwong:
This adds the `id` property to the Edge object, to match the Node object, and all the other objects returned from table rows. But it redefines the `Edge()` constructor (might that break people's code?), and in a number of the test suites I've hacked around the need to provide an ID for the edge by just putting an ID of `None` where we are constructing edges to then mess around with. I don't know if this is the right thing to do.


python/tskit/trees.py
```diff
@@ -171,7 +173,8 @@ class Edge(SimpleContainer):
         :meth:`.TreeSequence.node`.
     :vartype child: int
     """
-    def __init__(self, left, right, parent, child):
+    def __init__(self, id_, left, right, parent, child):
```
jeromekelleher: Put ``id_=None`` at the end here. This'll break less code.

hyanwong:
Yes - definitely push until after 0.2.0.

python/tskit/trees.py
hyanwong: OK. Note that in the tests these constructors are often called using positional params, so this will make it inconsistent with some of the `Node()` etc examples. But better that than breaking other people's code, I guess.

Presumably best to change the calls to use named parameters in the test cases too.

jeromekelleher: Yes, that's fine --- these things aren't particularly consistent anyway. There's no need to change existing tests if they're not actually using the ID parameter.

python/tests/simplify.py
```diff
@@ -192,13 +192,16 @@ def record_edge(self, left, right, parent, child):
             left = X[left_index]
             right = X[right_index]
         if child not in self.edge_buffer:
-            self.edge_buffer[child] = [tskit.Edge(left, right, parent, child)]
+            self.edge_buffer[child] = [tskit.Edge(
+                left=left, right=right, parent=parent, child=child, id_=None)]
```
jeromekelleher: ``id_`` defaults to None, so no point in changing this, right?

python/tests/simplify.py
```diff
@@ -192,13 +192,16 @@ def record_edge(self, left, right, parent, child):
...
         else:
             last = self.edge_buffer[child][-1]
             if last.right == left:
                 last.right = right
             else:
-                self.edge_buffer[child].append(tskit.Edge(left, right, parent, child))
+                self.edge_buffer[child].append(
+                    tskit.Edge(
+                        left=left, right=right, parent=parent, child=child, id_=None))
```
jeromekelleher: Ditto here.

python/tests/simplify.py
```diff
@@ -542,13 +545,16 @@ def record_edge(self, left, right, parent, child):
         Adds an edge to the output list.
         """
         if child not in self.edge_buffer:
-            self.edge_buffer[child] = [tskit.Edge(left, right, parent, child)]
+            self.edge_buffer[child] = [tskit.Edge(
+                left=left, right=right, parent=parent, child=child, id_=None)]
```
jeromekelleher: & here.

jeromekelleher: There's a few more of these --- I won't comment any more.

python/tskit/trees.py
```diff
@@ -171,17 +171,17 @@ class Edge(SimpleContainer):
...
     """
-    def __init__(self, left, right, parent, child):
+    def __init__(self, left, right, parent, child, id_=None):
+        self.id = id_
         self.left = left
         self.right = right
         self.parent = parent
         self.child = child
 
-    def __repr__(self):
```
jeromekelleher: Why delete this?

python/tests/simplify.py
hyanwong: Good point. I had forgotten about the default

python/tests/simplify.py
hyanwong: ditto

python/tskit/trees.py
hyanwong: The other SimpleContainer classes don't have an explicit __repr__ method (the rely on the base class). Having one here meant that we ended up not printing the id on display. So this seems redundant to me

hyanwong:
Good point about tests. I'll add some this weekend and remove the `_id=None` redundancies.

hyanwong:
By the way, where are the other places where we need the id integrated, as well as `edge_diffs()`? I thought I had found them, but there's a good chance I missed some.

python/tskit/trees.py
jeromekelleher: I guess it's probably unncessary. I put it in to make reading the edges easier (note the precision for left and right)

jeromekelleher:
There's the ``TreeSequence.edge()`` function anyway. That's probably it.

python/tskit/trees.py
hyanwong: Oh, good point about the precision. I'll reinstate with the id var added, then.

hyanwong:
I have added (rather minimal) tests, added the `__repr__` method for edges back in, and removed the unneeded params. It looks to me as if I had already changed that `TreeSequence.edge()` function as needed.

Travis (OS X) builds are failing now, but it seems unrelated:

```
ERROR: Cannot uninstall 'certifi'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

jeromekelleher:
Looks good, thanks @hyanwong. I'll merge when CI passes.

jeromekelleher:
CI failure is unrelated, #366.

jeromekelleher:
Can you rebase this please @hyanwong, so we're not pulling in fix for travis twice. Ready to merge then.

hyanwong:
Done

PR #323, opened 2019-08-22
Update C code to deal with missing data

hyanwong:
@jeromekelleher thought it might be simpler (and possibly more efficient) to code this up using in Python using `ts.genotype_matrix`, but by that time I had essentially done the C coding.

However, this code sets all the ancestral states, then whizzes through the trees again setting missing whenever it finds isolated nodes. So it might be not very efficient.

### DIFF
--- a/c/tskit/genotypes.c
+++ b/c/tskit/genotypes.c
@@ -66,12 +66,13 @@ tsk_hapgen_update_sample(tsk_hapgen_t * self, size_t sample_index, tsk_id_t site
 {
     int ret = 0;
     size_t index = sample_index * (self->num_sites + 1) + (size_t) site;
-
-    if (self->haplotype_matrix[index] == derived_state[0]) {
-        ret = TSK_ERR_INCONSISTENT_MUTATIONS;
-        goto out;
+    if (self->haplotype_matrix[index] != self->missing_data_character) {
+        if (self->haplotype_matrix[index] == derived_state[0]) {
+            ret = TSK_ERR_INCONSISTENT_MUTATIONS;
+            goto out;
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I'm not sure there's much point in doing this in C @hyanwong --- it's going to be a lot of work going through all the corner cases and making sure it's tested. Particularly when just doing it all in Python is much more straightforward --- see #326.

hyanwong:
Yeah, I had mostly coded this up before chatting to you. 

I do think there should be an efficient method for iterating over haplotypes that doesn’t require loading the entire genotype matrix into memory though. But this isn’t it.

jeromekelleher:
> I do think there should be an efficient method for iterating over haplotypes that doesn’t require loading the entire genotype matrix into memory though. But this isn’t it.

That's what the C code is doing, more-or-less.

hyanwong:
Really? It looks to me as if it makes a copy of the entire haplotype matrix: I’m assuming that requires it all to be in memory?

jeromekelleher:
> Really? It looks to me as if it makes a copy of the entire haplotype matrix: I’m assuming that requires it all to be in memory?

That's what I meant, it's making a full copy of the genotype matrix. I don't see a way around that, if you want to do it time-efficiently.

jeromekelleher:
I think we should close this and focus on #326; there's no point in wasting time getting this PR up to production quality when we're going to delete all the code soon anyway.

hyanwong:
Yes, sure. This temporary hack is being used by @marianne-aspbury at the moment, but she can simply use my branch.

PR #327, opened 2019-08-22
clarifying note about Z=None

### LINKED ISSUE #321: Remove None for Z in trait_regression
In ``trait_regression`` we currently have 
```python
       if Z is None:
            Z = np.ones((self.num_samples, 1))
        else:
            tZ = np.column_stack([Z, np.ones((Z.shape[0], 1))])
            if np.linalg.matrix_rank(tZ) == tZ.shape[1]:
                Z = tZ
```

but Z is not assigned a default value in the signature, and this behaviour on None isn't documented.

I suggest we remove the ``Z is None`` branch for now and revisit (maybe we'll want to do something with dimensions in the default case). What do you think @petrelharp?

petrelharp:
Closes #321. This has only changed the code by setting the default to `Z=None`.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -3712,7 +3712,8 @@ class TreeSequence(object):
                         W, windows=windows,
                         mode=mode, span_normalise=span_normalise)
 
-    def trait_regression(self, W, Z, windows=None, mode="site", span_normalise=True):
+    def trait_regression(self, W, Z=None, windows=None, mode="site",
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Looks good, thanks. Can you fix the linting issue and merge when CI is done please?

petrelharp:
Grrr, didn't have the pre-commit flake8 hook on that machine...

PR #328, opened 2019-08-22
Replace mean_descendants with a node stat

### LINKED ISSUE #273: implement mean_descendants as a node statistic
As discussed in #271, this is a node statistic. The differences to current node stats are that (a) it normalizes by the amount that the node is actually an ancestor; and (b) it is polarised (and `polarised=False` will give something that is always equal to 1). I propose that we make the behavior of (a) a parameter, and implement that normalization at the python level.

jeromekelleher:
Closes #273 

(I had to base this off #313, so ignore everything except the last commit here in the initial version)

I had a go at this @petrelharp, and managed to duplicate the original code using node stats and numpy operations, but I had to introduce two strict-failing summary functions to do it. It's not *totally* trivial to do, I think, and we'll end up having to do some of the work in C at least I think (by the time we'd have defined the operations in C we'd have done most of the work). So, this needs to be updated to follow the pattern of the rest of the API and do/not do the funky normalising it's got at the moment.

Given that there's a bit of work to do here, I suggest we push this out to beyond 0.2.0 and maybe put a bigger warning on the documentation stating that this **really** is preliminary and we reserve the right to change it any way we feel like.


jeromekelleher:
There's a few changes that I'll need to backport to #313 also --- turns out we weren't handling dimensions in the general stats properly.

python/tskit/trees.py
```diff
@@ -4244,13 +4135,136 @@ def f2(self, sample_sets, indexes=None, windows=None, mode="site",
...
+            proportion of the genome that `node` is an ancestor to anyone.
+
+        :param iterable reference_sets: A list of lists of node IDs.
+        :return: An array with dimensions (number of nodes in the tree sequence,
+            number of reference sets)
+        """
+        # An array of booleans per tree, true if node u is in the tree else false.
+        in_tree = self.sample_count_stat(
+            [self.samples()], lambda x: x + 1, output_dim=1, polarised=True,
+            strict=False, mode="node", windows="trees")[..., 0] > 1
```
petrelharp: I *think* that `ts.segregating_sites(..., mode="node")` does this.

petrelharp: Oh, darn: it *almost* does this, but doesn't count root nodes.

jeromekelleher: I played around with segregating_sites for a bit all right, but it was its behaviour around roots that made me give up. I don't see any around that.

python/tskit/trees.py
```diff
@@ -4244,13 +4135,136 @@ def f2(self, sample_sets, indexes=None, windows=None, mode="site",
...
+
+        C = self.sample_count_stat(
+            reference_sets, lambda x: x, output_dim=len(reference_sets), polarised=True,
+            strict=False, mode="node", windows="trees")
+        # Reshape so that we can multiply through by the span of each tree
+        tree_span = tree_span.reshape((tree_span.shape[0], 1, 1))
+        C *= tree_span
+
+        # To replicate the existing code, take the mean weighted by node span.
+        C = np.sum(C, axis=0)
```
petrelharp: I think the following does the same thing without having to do the `windows="trees"` stuff yourself:
```
denom = self.sample_count_stat(
            [self.samples()], lambda x: x > 0, output_dim=1, polarised=True,
            strict=False, mode="node", span_normalise=False)
numer = self.sample_count_stat(
        reference_sets, lambda x: x, output_dim=len(reference_sets), polarised=True, span_normalise=False,
        strict=False, mode="node")
C = numer / denom
```

petrelharp: This sort of normalization by `denom` is a pretty common node statistic operation that we should maybe even define as an option to those methods.

petrelharp: Or maybe (as we had said on this topic before) just provide a method that gives us the denominator. As a site statistic this'd give the number of segregating-or-fixed sites, and as a branch statistic would give us the total tree length plus any branches above the MRCA. So maybe it should be an option to `segregating_sites`, say, `include_fixed=False`.

petrelharp:
I've got this a good ways towards being a normal node statistic.  There's some API questions here, ping @hyanwong and @gtsambos:

1. Do we want to allow computation of the `branch` and `site` versions? For instance, the `site` version would ask for the mean allele frequency among each of the sample sets across derived alleles found in all samples. (This seems interesting, so yes?)

2. Should we rename `reference_sets` to `sample_sets` (I think yes, if the answer to (1) is yes).

3. Should we rename this based on the "site" version, like the other statistics? It could be "mean_segregating_frequency"?

This is somewhat in a different class than previous statistics because of the normalisation, but after talking it through, it's not feeling very different.

petrelharp:
Oh, and: I'm ok leaving this until after 0.2.0.

python/tests/test_tree_stats.py
```diff
@@ -801,8 +801,10 @@ def example_sample_sets(ts, min_size=1):
     yield [[s] for s in samples]
     if min_size == 1:
         yield [samples[:1]]
-    if ts.num_samples <= 2 and min_size >= 2:
```
petrelharp: I realized that this was never getting called.

python/tests/test_tree_stats.py
```diff
@@ -3441,8 +3441,7 @@ class TestGeneralSiteStats(StatsTestCase):
     def compare_general_stat(self, ts, W, f, windows=None, polarised=False):
         # Determine output_dim of the function
         M = len(f(W[0]))
-        py_ssc = PythonSiteStatCalculator(ts)
-        sigma1 = py_ssc.naive_general_stat(W, f, windows, polarised=polarised)
```
petrelharp: removing some cruft

python/tests/test_tree_stats.py
```diff
@@ -4391,7 +4390,6 @@ class BranchSampleSetStatsTestCase(SampleSetStatTestCase):
     def setUp(self):
         self.rng = random.Random(self.random_seed)
         self.stat_type = "branch"
-        self.py_stat_class = PythonBranchStatCalculator
```
petrelharp: only place this was mentioned; unused.

python/tests/test_tree_stats.py
```diff
@@ -5295,121 +5293,147 @@ def test_f3_windows(self):
         ts = self.get_example_ts()
         self.verify_three_way_stat_windows(ts, ts.f3)
 
-############################################
-# Old code where stats are defined within type
-# specific calculattors. These definititions have been
-# move to stat-specific regions above
-# The only thing left to port is the SFS code.
-############################################
```
petrelharp: we can get rid of this, yay!

python/tests/test_tree_stats.py
```diff
@@ -5295,121 +5293,147 @@ def test_f3_windows(self):
...
-                        w = f(x)
-                        for j in range(n_out):
-                            S[j] += w[j]
-            for j in range(n_out):
-                S[j] /= (end - begin)
-            out[j] = np.array([S])
-        return out
+    def verify(self, ts, reference_sets, windows=None):
+        C1 = mean_descendants(ts, reference_sets, mode=self.mode, windows=windows)
+        # C2 = tsutil.mean_descendants(ts, reference_sets, windows=windows)
```
petrelharp: to compare to the tsutil version we need to make that one windowed as well

petrelharp:
Gah; I found a "why did this ever work" error in the naive, test-only versions of two of the statistics, uncaught previously because we weren't looking at very many sample sets. Fixed now, but  I'll pull those changes off into another PR.

petrelharp:
I don't think we need to get this all sorted for 0.2.0, but if we're going to rename reference_sets to sample_sets, that'd be nice to do first, and that seems like an easy decision. (I vote yes.)

python/tskit/trees.py
```diff
@@ -4160,18 +4160,24 @@ def mean_descendants(self, reference_sets, windows=None):
...
+        sample_sets = reference_sets + [self.samples()]
         C = self.sample_count_stat(
-                    reference_sets, lambda x: x, output_dim=len(reference_sets),
+                    sample_sets, f, output_dim=len(sample_sets),
                     polarised=True, strict=False, mode="node", span_normalise=False,
                     windows=windows)
+        out_slice = tuple([slice(None)] * (len(C.shape) - 1) + [slice(0, -1)])
+        denom_slice = tuple([slice(None)] * (len(C.shape) - 1) + [-1])
+        out = C[out_slice]
+        denom = C[denom_slice].reshape(list(C.shape[:-1]) + [1])
```
petrelharp: This implementation is more opaque than [the previous implementation](https://github.com/tskit-dev/tskit/pull/328/commits/ab4dbd9aba581637f00c69ed1ded54540a93e065#diff-993cce8b29b806d11aae4e2ead3c71ecR4164) but is closer to how we'd do it in C.

jeromekelleher:
> I don't think we need to get this all sorted for 0.2.0, but if we're going to rename reference_sets to sample_sets, that'd be nice to do first, and that seems like an easy decision. (I vote yes.)

OK, great. Agreed on renaming to sample_sets, I'll get that in for 0.2.0.

hyanwong:
> Agreed on renaming to sample_sets

Just to check for this renaming: I'm assuming that we are never going to want to define sets that include non-sample nodes?

For example, there's Gil's idea of anonymising data by using the inferred ancestors as data points, rather than using the samples themselves. But it may be that this is done by redefining them as samples (especially now that we can have samples which are not complete sequences)

jeromekelleher:
> Just to check for this renaming: I'm assuming that we are never going to want to define sets that include non-sample nodes?

We might do I guess, but I think it's OK to have this slight inconsistency in favour of greater consistency across the various stats functions. We already allow for non-sample 'samples' in the ``samples`` argument to the ``variants`` method,  so it's not without precedent. 

In either case, it's the first positional parameter so the name isn't totally critical.

jeromekelleher:
OK, so the decision is to bring this stat up to date to make it more like the other general stats. This means that the API we're shipping in 0.2.0 for this function isn't going to be stable, but that's OK because we'll clearly document it as experimental.

I'll update this PR once the dust has settled a bit so it's a cleaner diff.

gtsambos:
fyi everyone: as this is my last day in Oxford I've been busy wrapping up some other things, so haven't had time to think about this properly yet - but I will soon!

jeromekelleher:
I've brought this up to date @petrelharp. We've reproduced the old behaviour successfully, but we want to have options for normalising by how long the node was in trees for. How would you like to proceed with this (possibly something it'd be better to talk over?)

jeromekelleher:
@awohns, you're the main user of mean_descendents at the moment. Can you comment on what normalisation options you'd like to see please?

awohns:
In addition to length of genome (which is default), I suppose the normalising by the size of the reference sets is the only option that comes to mind at the moment.

jeromekelleher:
> In addition to length of genome (which is default), normalising by the size of the reference sets is the only option that comes to mind at the moment.

Do you always want to normalise by the length of genome? I've gone round in circles in this one.

awohns:
Just reread the various threads on this. I agree with @petrelharp in #273, normalising by the span which the node is an ancestor to any sample should be an option, I think there are cases where you would want to normalise (for instance, if you're solely interested in which samples copy from an individual ancestor) and cases were you wouldn't (if you're comparing the relative contribution of a number of ancestors to the samples).

jeromekelleher:
(Basically I can't remember why I did this now, so happy to close unless someone else would like to have it in)

petrelharp:
This seems like a nice thing to do in terms of (a) making it more obvious what's going on with `mean_descendants`, (b) removing a bunch of code (since now `mean_descendants` would just use the general stat interface), and (c) making the interface consistent ("sample_sets"). However, it probably doesn't change much in practice, so it is not high priority. (Unless someone's come up with a use case for the `branch` or `site` mode stats here?)

jeromekelleher:
I agree, let's close. We can always reopen and pick up again later if someone really wants this.

PR #297, opened 2019-08-12
Stop "RuntimeWarning: invalid value encountered in true_divide"

hyanwong:
I was getting a little annoyed by "invalid value encountered in true_divide" warnings during the unit tests in some of the `TestDiversity` derived classes. This stops the errors, but perhaps we want to work out why they are happing in the first place. I guess this is a @petrelharp thing?

PR #330, opened 2019-08-23
Fix testing

petrelharp:
I discovered some errors in the testing setup while mucking around with #328, and have pulled those out. These changes depend on #313; ignore those commits.  I've run all the stats tests with p=1.0  to make sure things are good.  Note that there were no errors with the  actual implementations, only with two of our "naive" implementations.


petrelharp:
Oh, *this PR* actually closes #297.

PR #335, opened 2019-08-23
Fix TODO in computing mutation parents

hyanwong:


petrelharp:
I don't think this is right. Can you make a small example? And, I don't see how anything like this *could* work, because: consider the case that a mutation's parent is removed but that mutation's parent is not (or a longer such chain); to discover this you need to traverse trees. Well, I guess it could be done more simply than compute_mutation_parents, by traversing the tree of mutations only, but I don't know if this is worth the trouble?

hyanwong:
It works in this case because `keep_intervals` removes all mutations at a site. So if we remove one mutation, we are also compelled to remove the entire chain back to the root. At least, that's my take...

It's tested in e.g.

```
nosetests tests/test_topology.py:TestKeepIntervals.test_single_tree_jukes_cantor
```

which produces mutations with chains of parents, I think.

petrelharp:
Oh, right. Context, Peter, context. =)  Nice.

hyanwong:
It's easy to overthink this sort of thing. But a good sign of paranoia in general, I think.

jeromekelleher:
very neat, thanks @hyanwong. I'm assuming that this was tested pretty throughly first time around, so there's nothing else needed here.

Can you rebase to bring it up to date please?

hyanwong:
Rebased, and comment added to explain why we don't need to deal with the mutation tree (in case anyone is thinking of copying and pasting the code in a different context, when it might fail for reasons discussed above.

jeromekelleher:
OK, thanks. The appveyor failure is unrelated (#344), I'll merge once this is sorted out.

PR #342, opened 2019-08-27
Raise limit for offset cols to 2**32.

### LINKED ISSUE #341: Metadata columns artificially limited to 2G
The ``metadata_offset`` column in the C table API is a uint32_t (as are all other offset columns) but the metadata column is limited to storing < 2G of data. This is because we're using the same check for signed and unsigned limits, e.g., [here](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L235). 

Add a new function ``check_offset_overflow`` which will check the limits correctly in this case, and allow metadata columns to be 4G long.

jeromekelleher:
Closes #341

cc @petrelharp and @molpopgen. This takes some of the pressure off for metadata columns anyway.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -787,10 +787,10 @@ test_edge_table_squash_multiple_parents(void)
         "4  10  4   0\n"
         "0  4   4   0\n";
 /*
-            4       5  
+            4       5
            / \     / \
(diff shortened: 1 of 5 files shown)


jeromekelleher:
Build failure is unrelated: #344.

Issue #257, opened 2019-07-04
Redefine summary function to update rather than return


jeromekelleher:
While working through the changes needed to implement the AFS in C over in #248, I found myself copying and pasting a depressing amount of code, most of it doing basically the same thing as the general stats machinery.

It occured to me that if we changed the way the summary function works slightly, we should be able to reuse **all** the machinery when we're computing the jAFS and be able to specify a summary function like the other stats.

Currently, we're defining the f such that f(x) returns a 1D result vector, which we then add on to the current output window, w. That is, we do this in the naive case:
```python
    sigma = np.zeros((ts.num_trees, m))
    for tree in ts.trees():
        x = np.zeros((ts.num_nodes, k))
        x[ts.samples()] = w
        for u in tree.nodes(order="postorder"):
            for v in tree.children(u):
                x[u] += x[v]
        if polarised:
            s = sum(tree.branch_length(u) * f(x[u]) for u in tree.nodes())
        else:
            s = sum(
                tree.branch_length(u) * (f(x[u]) + f(total - x[u]))
                for u in tree.nodes())
          sigma[tree.index] = s 
```
whereas we'd now do:
```python
   sigma = np.zeros((ts.num_trees, m))
    for tree in ts.trees():
        x = np.zeros((ts.num_nodes, k))
        x[ts.samples()] = w
        for u in tree.nodes(order="postorder"):
            for v in tree.children(u):
                x[u] += x[v]
             f(tree.branch_length(u), x[u],  polarised, sigma[tree.index])
 ``` 
     
That is:
- We change the summary function to one that directly updates the output window by incrementing it
- We push the responsibility for dealing with polarisation down into the summary function (I guess, if you wanted to, you could have two different summary functions polarised & unpolarised for each stat). This is motivated by the AFS, where the polarised & unpolarised are pretty different and I can't see how we could efficiently do both in a general way. (It's not clear how this generalises to the site algorithm though, this is a bit different.)
- We pass in a constant multiplier factor so that branch length/span multiplied in as the values are computed.

The summary function will know about the expected output dimensions (an arbitrary n-D array)

This is a lot less elegant mathematically, but should be much more convenient and efficient computationally. The goal would then be that we do the minimal number of updates to the output window to compute the stats that we want (i.e. touching memory as little as possible).  We can also them put in arbitrary dimensioned arrays as the window elements, which should be general enough for most things!

[...]

PR #349, opened 2019-08-29
Branchstat algorithm

jeromekelleher:
Refactor of the branch/node stat algorithms. Changes

1. Store the value of f(x) during the branch_stat algorithm, which simplifies the logic a bit and reduces the number of calls of f substantially. (not fully implemented in C when the PR was opened)
2. Abstract out the details of running polarised vs unpolarised summary functions in C for node and branch stats. (It would be pleasing if this was possible for site stats also, but I don't see how as polarisation is much more central to how things are actually computed.)

Should be significant perf gains from this. I'll do some before/after comparisons when its done.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -988,19 +988,15 @@ update_state(double *X, size_t state_dim, tsk_id_t dest, tsk_id_t source, int si
 static inline int
 update_running_sum(double *s, size_t result_dim, double *X, size_t state_dim,
         general_stat_func_t *f, void *f_params, tsk_id_t u, tsk_id_t v,
-        const double *time, double *total_W,
-        bool polarised, int sign)
+        const double *time, int sign)
 {
     int ret = 0;
-    size_t k, m;
+    size_t m;
     double *X_u = GET_2D_ROW(X, state_dim, u);
     double branch_length;
-    double *total_minus_Xu = malloc(state_dim * sizeof(*total_minus_Xu));
     double *F = malloc(result_dim * sizeof(*F));
-    double *F_tmp = malloc(result_dim * sizeof(*F_tmp));
 
-    /* s += branch_length * (f(X[u]) + (not polarised) * f(total - X[u])) */
-    if (F == NULL || F_tmp == NULL || total_minus_Xu == NULL) {
+    if (F == NULL) {
         ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
@@ -1011,18 +1007,6 @@ update_running_sum(double *s, size_t result_dim, double *X, size_t state_dim,
         if (ret != 0) {
             goto out;
         }
-        if (! polarised) {
-            for (k = 0; k < state_dim; k++) {
-                total_minus_Xu[k] = total_W[k] - X_u[k];
-            }
-            ret = f(state_dim, total_minus_Xu, result_dim, F_tmp, f_params);
-            if (ret != 0) {
-                goto out;
-            }
-            for (m = 0; m < result_dim; m++) {
-                F[m] += F_tmp[m];
-            }
-        }
         branch_length = time[v] - time[u];
         for (m = 0; m < result_dim; m++) {
             s[m] += sign * branch_length * F[m];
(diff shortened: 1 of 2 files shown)


jeromekelleher:
This can close #257 I think --- getting the function to update now would be very minor perf tweak, probably not worth it. I haven't thought about whether this idea lets us abstract to the AFS yet. That would be nice.

jeromekelleher:
Benchmarks:

On this tree sequence:
```
$ python3 -m tskit info large-example.trees 
sequence_length:  100000000.0
trees:            456789
samples:          100000
individuals:      0
nodes:            500814
edges:            1941686
sites:            484395
mutations:        484395
migrations:       0
populations:      1
provenances:      1
```
and this code:
```python
import tskit
import time
import numpy as np

ts = tskit.load("large-example.trees")


print("mode", "n_sets", "time", sep="\t")
for mode in ["site", "branch", "node"]:
    for num_sample_sets in [1, 10, 100]:
        sample_sets = np.array_split(ts.samples(), num_sample_sets)
        before = time.perf_counter()
        x = ts.diversity(sample_sets=sample_sets, mode=mode)
        print(mode, num_sample_sets, "{:.3f}".format(duration), sep="\t")
```

On upstream/master we get:
```
$ python3 benchmark.py
mode    n_sets  time
site    1       0.344
site    10      0.557
site    100     2.413
branch  1       1.767
branch  10      2.949
branch  100     15.900
node    1       1.111
node    10      2.162
node    100     12.033
```
On this branch we get 
```
$ python3 benchmark.py
mode    n_sets  time
site    1       0.349
site    10      0.557
site    100     2.426
branch  1       0.647
branch  10      1.761
branch  100     11.555
node    1       0.570
node    10      1.654
node    100     10.541
```

So, we do get a decent perf jump on branch stats when we have a small number of sample sets. Not sure what's going on with the larger number of sample sets, will investigate.

jeromekelleher:
I don't think we're going to do much better here. When we have 100 sample sets, we get this from perf:
```
  44.93%  python3        _tskit.cpython-36m-x86_64-linux-gnu.so             [.] tsk_treeseq_general_stat                                                                                                                                      
  36.67%  python3        _tskit.cpython-36m-x86_64-linux-gnu.so             [.] diversity_summary_func                                                                                                                                        
  14.79%  python3        _tskit.cpython-36m-x86_64-linux-gnu.so             [.] npolarised_summary_func                                                                                                                                      
   0.65%  python3        [kernel]                                           [k] 0xffffffffb7c009e7                                                                                                                                            
   0.45%  python3        _tskit.cpython-36m-x86_64-linux-gnu.so             [.] tsk_treeseq_init                                                                                                                                              
   0.26%  python3        python3.6                                          [.] _PyEval_EvalFrameDefault           
```
When you dig in we're spending time in looping over the sample sets, as you might expect. We could probably squeeze a bit more perf here by using ``restrict``s, but they are tricky and I don't think we'll get much out of it.

The reason that node and branch stats are so much slower than site stats is because they're doing a lot more. The summary func is only called on mutation nodes not on *every* node. So, it's as good as it's going to get for now I think, and we should be pretty happy with this (it's friggin fast!).

jeromekelleher:
@petrelharp, any thoughts here? I'm going to merge this and write up the algorithm unless you object.

petrelharp:
Nice performance improvements!

> Abstract out the details of running polarised vs unpolarised summary functions in C for node and branch stats. (It would be pleasing if this was possible for site stats also, but I don't see how as polarisation is much more central to how things are actually computed.)

This is a nice way to do it. Doing it the same way for site stats would require passing a list of states (one for each derived allele) to a wrapped function, I think - we could work this out but it doesn't seem necessary.

c/tskit/trees.c
```diff
@@ -973,6 +973,8 @@ tsk_treeseq_check_windows(tsk_treeseq_t *self, size_t num_windows, double *windo
     return ret;
 }
 
+/* TODO make these functions more consistent in how the arguments are ordered */
```
petrelharp: :+1:

petrelharp:
This is very nice and clean. I like it.

Issue #326, opened 2019-08-22
Remove C support for haplotypes and implement in Python


jeromekelleher:
There's no real point in having the C haplotype generator any more when we can replace it with some numpy code that'll be more-or-less as fast. It'll look something like:

```python
H = np.empty((ts.num_samples, ts.num_sites), dtype=np.int8)

for var in ts.variants():
    alleles = np.array(ord(allele) for allele in var.alleles) # Throw an error if allele isn't single char
    H[:, var.site.id] = alleles[var.genotypes]
for h in H:
    # Do some numpy trickery like we use in VCF to change to a utf8 encoded string
    yield numpy_trickery(h)
```

Then we can deal with missing data much more simply and can get rid of a bunch of repetitious C code.

hyanwong:
This might be obvious, but for "real data", we might well want `haplotypes()` to return full sequences, including non-variable sites. Otherwise we may get what looks like a contiguous DNA sequence out of a `haplotypes()` method, but which is actually a pasting together of only the variable sites. I've stupidly just been bitten by this.

At the moment I'm doing this in simulations by adding *all sites* into the tree sequence, including non-variable ones. But perhaps we (a) want to warn when producing a text string with integer site positions that are not incrementing by 1 unit each site and/or (b) have a haplotype-rendering method that takes a reference sequence and fills in the missing sites with data from the reference.

Obviously these points are only really valid for tree sequences with integerized site positions.

hyanwong:
I think we need to distinguish between (a) tree sequences that using the (msprime default) infinite sites mutation model and (b) tree sequences that have sites at integer positions (and hence probably use [finite mutation models](https://github.com/tskit-dev/msprime/issues/553) ). There are two possible ways to output haplotypes:

* Each haplotype is output as a 1D numpy array. This makes sense in both (a) and (b): we might want to (efficiently if possible) iterate over samples, producing "haplotypes" that are essentially the columns of the `genotype_matrix()` (i.e. an array of ints).

* Each haplotype is output as a string (i.e. a concatenated series of characters), in something like [FASTA](https://github.com/tskit-dev/tskit/pull/350) format. This is trickier. Firstly, we probably want to make each allele of the same number of characters (or, as we do at present, restrict this sort of output to only having 1-character alleles). Secondly, we will usually want to indicate that there are gaps between the sites: this is definitely the case for (b) and may be the case for (a) if we do not have all the integer sites in the sites table.

[...]

PR #350, opened 2019-08-29
Writing fasta files from tree sequences

marianne-aspbury:
I've written some code to cover integrating fasta output from tree sequences into tskit, following from issue https://github.com/tskit-dev/tskit/issues/338 and have included various tests for it.
Happy to hear feedback on this (e.g. @jeromekelleher, @hyanwong).

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -31,6 +31,7 @@ import base64
 import warnings
 import functools
 import concurrent.futures
+import textwrap
 
 import numpy as np
 
@@ -2916,6 +2917,68 @@ class TreeSequence(object):
             sample_population = self.tables.nodes.population[samples]
             samples = samples[sample_population == population]
         return samples
+
+    def write_fasta(self, output, sequence_ids=None, wrap_width=60):
+        """
+        Writes haplotype data for samples in FASTA format to the
+        specified file-like object.
+
+        Default `sequence_ids` (i.e. the text immediately following ">") are
+        "tsk_{sample_number}" e.g. "tsk_0", "tsk_1" etc. They can be set by providing
+        a list of strings to the `sequence_ids` argument, which must equal the length
+        of the number of samples. Please ensure that these are unique and compatible with
+        fasta standards, since we do not check this.
+        Default `wrap_width` for sequences is 60 characters in accordance with fasta
+        standard outputs, but this can be specified. In order to avoid any line-wrapping
+        of sequences, set `wrap_width = 0`.
+
+        Example usage:
+
+        .. code-block:: python
+            with open("output.fasta", "w") as fasta_file:
+                ts.write_fasta(fasta_file)
+
+        This can also be achieved on the command line use the ``tskit fasta`` command,
+        e.g.:
+
+        .. code-block:: bash
+
+            $ tskit fasta example.trees > example.fasta
+
+        :param File output: The file-like object to write the fasta output.
+        :param list(str) sequence_ids: A list of string names to uniquely identify
+            each of the sequences in the fasta file. If specified, this must be a
+            list of strings of length equal to the number of samples which are output.
+            Note that we do not check the form of these strings in any way, so that it
+            is possible to output bad fasta IDs (for example, by including spaces
+            before the unique identifying part of the string).
+            The default is to output ``tsk_j`` for the jth individual.
+        :param int wrap_width: This parameter specifies the number of sequence
+            characters to include on each line in the fasta file, before wrapping
+            to the next line for each sequence. Defaults to 60 characters in
+            accordance with fasta standard outputs. To avoid any line-wrapping of
+            sequences, set `wrap_width = 0`. Otherwise, supply any positive integer.
+        """
+        # if not specified, IDs default to sample index
+        if sequence_ids is None:
+            sequence_ids = ["tsk_{}".format(j) for j in self.samples()]
+        if len(sequence_ids) != self.num_samples:
+            raise ValueError(
+                "sequence_ids must have length equal to the number of samples.")
+
+        wrap_width = int(wrap_width)
+        if wrap_width < 0:
+            raise ValueError("wrap_width must be a non-negative integer. "
+                             "You may specify `wrap_width=0` "
+                             "if you do not want any wrapping.")
+
+        for j, hap in enumerate(self.haplotypes()):
+            print(">", sequence_ids[j], sep="", file=output)
+            if wrap_width == 0:
+                print(hap, file=output)
+            else:
+                for hap_wrap in textwrap.wrap(hap, wrap_width):
+                    print(hap_wrap, file=output)
 
     def write_vcf(
             self, output, ploidy=None, contig_id="1", individuals=None,
--- a/python/tests/test_cli.py
+++ b/python/tests/test_cli.py
@@ -234,6 +234,32 @@ class TestTskitArgumentParser(unittest.TestCase):
         self.assertEqual(args.tree_sequence, tree_sequence)
         self.assertEqual(args.human, True)
 
+    def test_fasta_default_values(self):
+        parser = cli.get_tskit_parser()
+        cmd = "fasta"
+        tree_sequence = "test.trees"
+        args = parser.parse_args([cmd, tree_sequence])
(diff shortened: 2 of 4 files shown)


jeromekelleher:
Great stuff, thanks @marianne-aspbury. How about we go through this in person tomorrow?

marianne-aspbury:
Yep, sounds good, thanks

python/tskit/trees.py
```diff
@@ -2917,6 +2918,63 @@ def samples(self, population=None, population_id=None):
...
+    def write_fasta(self, output, sequence_ids=None, wrap_width=60):
+        """
+        Writes a fasta formatted file from tree sequence data to the
+        specified file-like object.
+
+        Default `sequence_ids` (i.e. the text immediately following ">") are
+        "tsk_{sample_number}" e.g. "tsk_0", "tsk_1" etc. They can be set by providing
+        a list of strings to the `sequence_ids` argument, which must equal the length
+        of the number of samples. Please ensure that these are unique and compatible with
+        fasta standards, since we do not check this (however, do not include the ">").
```
jeromekelleher: I think mentioning the ">" is confusing here; just leave it out I'd say.

python/tskit/trees.py
```diff
@@ -2917,6 +2918,63 @@ def samples(self, population=None, population_id=None):
...
+        Example usage:
+
+        .. code-block:: python
+            with open("output.fasta", "w") as fasta_file:
+                ts.write_fasta(fasta_file)
+
+        :param File output: The file-like object to write the fasta output.
+        :param list(str) sequence_ids: A list of string names to uniquely identify
+            each of the sequences in the fasta file. Ids will be printed after ">"
+            so you do not need to include this in the strings.
```
jeromekelleher: Delete this sentence I think --- mentioning ">" is confusing I think.

python/tskit/trees.py
```diff
@@ -2917,6 +2918,63 @@ def samples(self, population=None, population_id=None):
...
+            accordance with fasta standard outputs. To avoid any line-wrapping of
+            sequences, set `wrap_width = 0`. Otherwise, supply any positive integer.
+        """
+        # if not specified, IDs default to sample index
+        if sequence_ids is None:
+            sequence_ids = ["tsk_{}".format(j) for j in self.samples()]
+        if len(sequence_ids) != self.num_samples:
+            raise ValueError(
+                "sequence_ids must have length equal to the number of samples.")
+        if type(wrap_width) != int:
```
jeromekelleher: Good idea to check that these are integers, but this is a bit tricky (suppose the input was a numpy integer, e.g.). I'm tempted to let this fall through to ``text_wrap`` --- what happens if we provide a string or double to it?

python/tskit/trees.py
```diff
@@ -2917,6 +2918,63 @@ def samples(self, population=None, population_id=None):
...
+            raise TypeError("wrap_width must be a non-negative integer."
+                            "You may specify `wrap_width=0`"
+                            "if you do not want any wrapping.")
+        if wrap_width < 0:
+            raise ValueError("wrap_width must be a non-negative integer.")
+
+        for j, hap in enumerate(self.haplotypes()):
+            # if user doesn't want any wrapping need alt write method
```
jeromekelleher: As a matter of principle, it'd be nice to get rid of the duplicated code here. Also, I'm not sure about how well the "\n" s will work cross platform. This is a bit neater/safer (we avoid creating some intermediate strings here)
```
print(">", sequence_ids[j], sep="", file=output)
if wrap_width == 0:
     print(hap, file=output)
else:
    for hap_wrap in textwrap.wrap(hap, wrap_width):
        print(hap_wrap, file=output)
```

python/tskit/trees.py
jeromekelleher: I guess if we want to do it here, the canonical way would be to just convert to int:

```
wrap_width = int(wrap_width)
```
If it's not convertible to int, then it'll throw a type error as we want.

marianne-aspbury: Okay, fair enough, I just didn't think that the error output was very nice, it'll look like this if I get rid of the integer check.
image

python/tskit/trees.py
marianne-aspbury: Okay, thanks!

python/tskit/trees.py
marianne-aspbury: Ah this comment didn't load before, which approach do you think is better?

> 
> 
> I guess if we want to do it here, the canonical way would be to just convert to int:
> 
> ```
> wrap_width = int(wrap_width)
> ```
> 
> If it's not convertible to int, then it'll throw a type error as we want.

marianne-aspbury: For now I've just deleted the `raise TypeError()` part of the `write_fasta` function (but it's still in the tests). I can add something back in though

marianne-aspbury: now added `wrap_width = int(wrap_width)`

marianne-aspbury:
Added functionality for printing fastas on command line, pretty much copied directly from the vcf CLI code and all its tests.

hyanwong:
Sorry - I've not been following this. Note my reservations in https://github.com/tskit-dev/tskit/issues/326#issuecomment-524312997 about making sure that FASTA files (and haplotypes in general) output non-variable sites, at a minimum with a dot, but possibly filling in via a reference genome.

petrelharp:
I don't think we're ready to deal with invariant sites yet, @hyanwong - to do that we need to distinguish tree sequences with integer positions and a reference sequence from the current infinite-sites notions. I'm in favor of doing this, but it shouldn't get in the way of this effort.

jeromekelleher:
> I don't think we're ready to deal with invariant sites yet, @hyanwong - to do that we need to distinguish tree sequences with integer positions and a reference sequence from the current infinite-sites notions. I'm in favor of doing this, but it shouldn't get in the way of this effort.

Agreed. Let's consider what we can do as a stopgap for finite sites as part of #326 (without changing the default behaviour; we don't break people's code). We can then decide what the final semantics of the fasta function should be in #353 (tagged as a 0.2.3 release issue so we don't forget about it).

Let's get this PR finished up and merge first though.

hyanwong:
> I don't think we're ready to deal with invariant sites yet, @hyanwong

We've constructed tree sequences with a site at each integer position, but many sites without mutations. Isn't that the same as an invariant site? But I guess this might be though of as a bit of a hack.

python/tskit/trees.py
```diff
@@ -33,6 +33,7 @@
 import concurrent.futures
 
 import numpy as np
+import textwrap
```
jeromekelleher: textwrap is part of the standard library, so it belongs in the first group of imports.

python/tskit/trees.py
```diff
@@ -2917,6 +2918,68 @@ def samples(self, population=None, population_id=None):
             samples = samples[sample_population == population]
         return samples
 
+    def write_fasta(self, output, sequence_ids=None, wrap_width=60):
+        """
+        Writes a fasta formatted file from tree sequence data to the
```
jeromekelleher: This assumes that the user knows what a fasta is. How about something like "Writes haplotype data for samples in FASTA format to the specified file-like object".

python/tskit/trees.py
```diff
@@ -2917,6 +2918,68 @@ def samples(self, population=None, population_id=None):
...
+        # if not specified, IDs default to sample index
+        if sequence_ids is None:
+            sequence_ids = ["tsk_{}".format(j) for j in self.samples()]
+        if len(sequence_ids) != self.num_samples:
+            raise ValueError(
+                "sequence_ids must have length equal to the number of samples.")
+
+        wrap_width = int(wrap_width)
+        if wrap_width < 0:
+            raise ValueError("wrap_width must be a non-negative integer."
```
jeromekelleher: Need some spaces at the end of these strings here (look at the string that is printed out, the words will be mushed together)

marianne-aspbury:
> 
> 
> Looks great, thanks @marianne-aspbury! Some very minor comments above --- we're ready to merge after these are addressed and the branch is rebased to bring it up to date.

Thanks @jeromekelleher, I've fixed these now

PR #356, opened 2019-09-04
A little more clarification on edge_diffs()

hyanwong:
I have to think each time whether I can do `zip(ts.trees(), ts.edge_diffs())`, because normally diff iterators on a list of length `N` are of length `N-1`. Hopefully a short comment in the docs like this will make it clear, if it's not already.


petrelharp:
Looks good! Thanks.

PR #359, opened 2019-09-04
better error value on general stat

petrelharp:
The `strict` check was producing an inscrutible numpy-related error if the summary function did not return a numpy array. I've fixed this (so that functions can still return a list) and put a test for this behavior.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -125,7 +125,8 @@ shared by many statistics, which we describe in detail in the following subsecti
     What section(s) of the genome are we interested in?
 
 :ref:`sec_stats_span_normalise`
-    Should we normalise information in windows by their span along the genome?
+    Should the statistic calculated for each window be normalised by the length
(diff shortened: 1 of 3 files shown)


petrelharp:
This should be good to merge.

petrelharp:
I'm putting some other doc improvements on here too, ping @molpopgen?

docs/stats.rst
```diff
@@ -124,7 +124,8 @@ shared by many statistics, which we describe in detail in the following subsecti
     What section(s) of the genome are we interested in?
 
 :ref:`sec_stats_span_normalise`
-    Should we normalise information in windows by their span along the genome?
+    Should we the statistic calculated for each window be normalised by the length
```
jeromekelleher: Typo: Should *we* the

jeromekelleher:
Is this good to merge @petrelharp? I'm happy to fix the typo and rebase, just let me know.

PR #368, opened 2019-09-24
Remove object declaration from classes

hyanwong:
Now that we insist upon python 3, we can simply remove these, assuming that the increased brevity makes the code more readable.


petrelharp:
Makes sense, but I vote to leave it as-is?

hyanwong:
OK - looks like the preference is to keep `object`. Closing

PR #369, opened 2019-09-26
Simplify keep_intervals code

hyanwong:
Now we can simply use `delete_sites()` for most of the functionality in `keep_intervals()`, assuming that we want to reduce code duplication at the expense of a (slightly) less efficient algorithm.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1780,50 +1780,19 @@ class TableCollection(object):
         tables = self.copy()
         sites = self.sites
         edges = self.edges
-        mutations = self.mutations
         tables.edges.clear()
-        tables.sites.clear()
-        tables.mutations.clear()
         keep_sites = np.repeat(False, sites.num_rows)
-        keep_mutations = np.repeat(False, mutations.num_rows)
         for s, e in intervals:
             curr_keep_sites = np.logical_and(sites.position >= s, sites.position < e)
             keep_sites = np.logical_or(keep_sites, curr_keep_sites)
-            new_as, new_as_offset = keep_with_offset(
-                curr_keep_sites, sites.ancestral_state, sites.ancestral_state_offset)
-            new_md, new_md_offset = keep_with_offset(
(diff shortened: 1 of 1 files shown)


python/tskit/tables.py
```diff
@@ -1780,50 +1780,18 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
...
-        # Don't bother dealing with connected mutations as we are removing the whole site
-        tables.mutations.set_columns(
-            site=site_map[mutations.site[keep_mutations]],
-            node=mutations.node[keep_mutations],
-            derived_state=new_ds,
-            derived_state_offset=new_ds_offset,
-            parent=parent_map[mutations.parent[keep_mutations]],
-            metadata=new_md,
-            metadata_offset=new_md_offset)
+        tables.delete_sites(np.where(~keep_sites)[0], record_provenance=False)
```
jeromekelleher: I had to look up what the ~ operator does (bitwise complement). I think ``not keep_sites`` is probably easier to read and logically apt.

hyanwong: Oh, sure. I had forgotten that you could use `not` on a numpy array too.

hyanwong: Change made, commit amended

hyanwong:
Ah - I didn't test it! you *can't* do `(not np_array)`. I guess I should move to `np.not(np_array)`.

jeromekelleher:
This is behind the tip of master, so it'd be nice to rebase. It doesn't matter in this case (the commit in front is a trivial doc update), but it would be good to rebase off upstream/master as a matter of course when updating PRs.

jeromekelleher:
Merging, thanks.

hyanwong:
> This is behind the tip of master, so it'd be nice to rebase. It doesn't matter in this case (the commit in front is a trivial doc update), but it would be good to rebase off upstream/master as a matter of course when updating PRs.

I try to do that. Sorry I forgot this time. Too many other simultaneous commit pathways in my head!

PR #370, opened 2019-09-26
Document installing straight from github

hyanwong:
Just found out how to do this! I can't emphasis hard enough how much time and effort this will save my students.

My strong preference is to replace the previous suggestion of installing from a tarball with installing direct from github (as in this PR). Otherwise we could add the tarball method as an alternative underneath, but I think that's confusing.


jeromekelleher:
LGTM, useful solution.

PR #372, opened 2019-09-27
Change keep/del intervals to in-place

### LINKED ISSUE #371: Should table editing functions make copies or edit in-place?
I've just noticed that while `delete_intervals()` and `keep_intervals()` return a *copy* of the table collection, my new `delete_sites()` function edits the table object in-place instead, and does not return a copy.

I'm about to code up `trim()` on a table collection, and it would be useful to know what the correct thing to do here is.

A hybrid approach would be to have a parameter `copy=False|True` for all these table-editing functions. If we are chaining various editing functions together, it makes sense to me to allow the possibility of in-place editing, otherwise I'm guessing it will be a burden for large tree sequences?

hyanwong:
And add equivalent tree seq methods. Also alter documentation as appropriate.

Fixes #371

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -40,6 +40,30 @@ import tests.tsutil as tsutil
 import tests.test_wright_fisher as wf
 
 
+def ts_equal(ts_1, ts_2, compare_provenances=True):
+    """
+    Check equality of tree sequences, ignoring provenance timestamps (but not contents)
+    """
+    return tables_equal(ts_1.tables, ts_2.tables, compare_provenances)
+
+
+def tables_equal(table_collection_1, table_collection_2, compare_provenances=True):
+    """
+    Check equality of tables, ignoring provenance timestamps (but not contents)
+    """
+    for (_, table_1), (_, table_2) in zip(table_collection_1, table_collection_2):
+        if isinstance(table_1, tskit.ProvenanceTable):
+            if compare_provenances:
+                if np.any(table_1.record != table_2.record):
+                    return False
+                if np.any(table_1.record_offset != table_2.record_offset):
+                    return False
+        else:
+            if table_1 != table_2:
+                return False
+    return True
+
+
 def simple_keep_intervals(tables, intervals, simplify=True, record_provenance=True):
     """
     Simple Python implementation of keep_intervals.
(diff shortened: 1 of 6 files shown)


python/tests/test_topology.py
```diff
@@ -4647,16 +4665,17 @@ def example_intervals(self, tables):
...
     def do_keep_intervals(
             self, tables, intervals, simplify=True, record_provenance=True):
-        t1 = simple_keep_intervals(tables, intervals, simplify, record_provenance)
-        t2 = tables.keep_intervals(intervals, simplify, record_provenance)
-        t3 = t2.copy()
+        t1 = tables.copy()
+        simple_keep_intervals(t1, intervals, simplify, record_provenance)
+        t2 = tables.copy()
+        t2.keep_intervals(intervals, simplify, record_provenance)
         self.assertEqual(len(t1.provenances), len(t2.provenances))
```
jeromekelleher: Might as well use the new tables_equal function here.

python/tskit/tables.py
```diff
@@ -1721,89 +1720,76 @@ def delete_sites(self, site_ids, record_provenance=True):
...
-
-        Note that node IDs may change as a result of this operation,
-        as by default :meth:`.simplify` is called on the resulting tables to
-        remove redundant nodes. If you wish to keep node IDs stable between
-        this set of tables and the returned tables, specify ``simplify=True``.
-
-        See also :meth:`.keep_intervals`.
+        Delete all information from this set of tables which lies *within* the
+        specified list of genomic intervals. This is identical to the
+        :class:`TreeSequence` method of the same name but acts *in place* to alter
```
jeromekelleher: More direct here (and elsewhere) to say "This is identical to :meth:`TreeSequence.delete_intervals`, but acts *in place*..."

python/tests/test_topology.py
hyanwong: Good spot, thanks. There are probably other places where this can be done too, but I'll get round to that in another PR

hyanwong:
Both comments done. This is now 4 commits. Do you want them squashed?

jeromekelleher:
> Both comments done. This is now 4 commits. Do you want them squashed?

Please -- keeping simple changes like this atomic is helpful when reviewing the history.

hyanwong:
> Please -- keeping simple changes like this atomic is helpful when reviewing the history.

Done

jeromekelleher:
merging, thanks.

PR #373, opened 2019-09-27
Trim functions

### LINKED ISSUE #288: Implement `ltrim()`, `rtrim()`, and `trim()` to remove genomic areas with no edges

hyanwong:
Replaces #292 
Fixes #288

Test coverage not yet complete, as I'm not sure if we want the (now relatively trivial) `splice()` functionality, although it was present in a previous tskit alpha see (https://github.com/tskit-dev/tskit/pull/261#issuecomment-513201321). If people like having that shortcut, I can leave it in and add tests. Otherwise I'll pull it.

### DIFF
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -708,13 +708,20 @@ Schema section (TODO).
 Table transformation methods
 ============================
 
-The following methods operate *in place* on a :class:`.TableCollection`,
-transforming them while preserving information.
+In general, table methods operate *in place* on a :class:`.TableCollection`,
+directly altering the data stored within its constituent tables.
+
 In some applications, tables may most naturally be produced in a way that is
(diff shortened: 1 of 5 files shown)


hyanwong:
Temporarily closing until #372 is merged

PR #387, opened 2019-10-15
Return SimpleContainers as table rows where possible

hyanwong:
Fixes #304 in another manner (see https://github.com/tskit-dev/tskit/pull/304#issuecomment-542188917)

This is for review only. It replaces the TableRow namedtuples with their equivalent container classes where possible, which allows functions like `ts.edges()` and `ts.nodes()` to simply return the appropriate table, which (when iterated over) returns the correct classes, as before.

Two SimpleContainer classes are actually more complicated than the corresponding TableRow object. These are `Individuals` (which contains the nodes belonging to an individual, which is not data present in the individuals table) and `Sites` (which contains the mutations present in each site too). These have to retain their own "TableRow" objects, which I've renamed `SimpleIndividual` and `SimpleSite`.

Two SimpleContainer objects, `Edge` and `Migration`, have their `id_` parameter appear as the *last* rather than *first* parameter in the `__init__()` method. This annoyingly means we have to code a bespoke version of `__getitem__()` for getting the rows from those tables. Moving `_id` to the start of the parameter list for these SimpleContainers would make the order consistent and allow us to remove this bespoke code, but would be a breaking change in parameter order, so I haven't done it (yet).

The docstrings will need a little adjusting to point out that instead of returning iterators, functions like `ts.edges()` return the tables whose rows can be iterated over.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1262,6 +1262,18 @@ class TestTreeSequence(HighLevelTestCase):
                     self.assertEqual(t3.interval, t2.interval)
                     self.assertEqual(t3.parent_dict, t2.parent_dict)
 
+    def test_reversible_iterators(self):
+        for ts in get_example_tree_sequences():
+            for table_name, _ in ts.tables:
+                if table_name not in ['individuals', 'sites', 'mutations']:
+                    # ts.individuals(), ts.sites() and ts.mutations() are not currently
+                    # reversible, as they do not return the equivalent tables
+                    ts_iterator = getattr(ts, table_name)()
+                    n_rows = ts_iterator.num_rows
+                    for i, n in enumerate(reversed(ts_iterator)):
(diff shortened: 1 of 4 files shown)


hyanwong:
By the way, since we are presumably flagging this for the next tskit version, can we remove [`add_deprecated_mutation_attrs`](https://github.com/tskit-dev/tskit/blob/98ae384bf884f7de79196f42a0624ed431b34049/python/tskit/trees.py#L410)? That would remove the TODO on the `ts.mutations()` method and make that reversible too, leaving only `sites()` and `individuals()` as the odd ones out.

hyanwong:
After discussion with @jeromekelleher this seems a bit of a confusing way to go, so reverting to the previous idea of returning rows as sequences.

jeromekelleher:
Yes, thanks for exploring the idea here @hyanwong.

PR #390, opened 2019-10-16
Add trim() to tables and tree sequences

hyanwong:
fixes https://github.com/tskit-dev/tskit/issues/288

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -28,6 +28,7 @@ import unittest
 import itertools
 import random
 import json
+import sys
 
 import numpy as np
 import msprime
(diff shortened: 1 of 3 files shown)


python/tests/test_topology.py
```diff
@@ -4984,3 +4984,62 @@ def test_ts_single_tree_delete_middle(self):
         ts_keep = ts.delete_intervals([[0.25, 0.5]], record_provenance=False)
         ts_delete = ts.keep_intervals([[0, 0.25], [0.5, 1.0]], record_provenance=False)
         self.assertTrue(ts_equal(ts_keep, ts_delete))
+
+
+class TestTrim(unittest.TestCase):
+    """
+    Test the trimming functionality
+    """
```
jeromekelleher: This is a good start. You need to add some tests which include sites as well (not testing at all at the moment, right?)

The tests aren't very thorough either --- I could implement trim at the moment by just putting in a single edge from left to right, and I think the tests would pass. Need to test that the bits in the middle have been faithfully kept (I guess, comparing the trees is the easiest? parent_dict should be equal for each pair of trees in the before and after ts).

The coordinates are all very nice and clean right now too. I'm worried about the case where we have a value like 0.1 which doesn't represent exactly in binary. Should we force the left coordinate to be exactly zero, or do we accept the fact that there'll be a tiny gap at the start of the sequence? Or worse, could subtracting the left coordinate lead to a negative number?

There's all sorts of corner cases that need to be explored here.

python/tskit/trees.py
```diff
@@ -3286,6 +3286,50 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
...
+        :param bool record_provenance: If True, add details of this operation to the
+            provenance information of the returned tree sequence. (Default: True).
+        """
+        tables = self.dump_tables()
+        tables.rtrim(record_provenance)
+        return tables.tree_sequence()
+
+    def trim(self, record_provenance=True):
+        """
+        Returns a copy of this tree sequence with any empty regions on the right and left
```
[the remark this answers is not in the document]

python/tskit/tables.py
```diff
@@ -1791,6 +1791,85 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
...
+        Reset the coordinate system used in these tables, changing the left and right
+        genomic positions in the edge table such that the leftmost edge now starts at
+        position 0. This is identical to :meth:`TreeSequence.ltrim` but acts *in place*
+        to alter the data in this :class:`TableCollection`.
+
+        :param bool record_provenance: If ``True``, add details of this operation
+            to the provenance table in this TableCollection. (Default: ``True``).
+        """
+        leftmost = np.min(self.edges.left)
+        self.delete_sites(
```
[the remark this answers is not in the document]
jeromekelleher: Easiest thing to do is just error out if we have any migrations. They're rarely (if ever) used, so there's no point in complicating the logic here.

python/tests/test_topology.py
hyanwong: > This is a good start. You need to add some tests which include sites as well (not testing at all at the moment, right?)

Good point. I've just added mutations in. Sorry for the oversight.

> The tests aren't very thorough either --- I could implement trim at the moment by just putting in a single edge from left to right, and I think the tests would pass. Need to test that the bits in the middle have been faithfully kept (I guess, comparing the trees is the easiest? parent_dict should be equal for each pair of trees in the before and after ts).

Done in latest push

> The coordinates are all very nice and clean right now too. I'm worried about the case where we have a value like 0.1 which doesn't represent exactly in binary. Should we force the left coordinate to be exactly zero, or do we accept the fact that there'll be a tiny gap at the start of the sequence? Or worse, could subtracting the left coordinate lead to a negative number?

Where would the negative number come from? If we subtract the left of the leftmost edge, won't it simply make all the other coords not-quite-simple numbers?

Would it be enough to test with (say) 0.1 rather than 0.25 as I do at present?

python/tests/test_topology.py
```diff
@@ -4990,18 +4990,35 @@ class TestTrim(unittest.TestCase):
     """
     Test the trimming functionality
     """
+    def add_mutations(self, tables, position, ancestral_state, derived_states):
+        "Create a site at the specified position and randomly assign mutations to nodes"
+        site = tables.sites.add_row(position, ancestral_state)
+        for state in derived_states:
+            tables.mutations.add_row(
+                site, np.random.randint(tables.nodes.num_rows), state)
```
jeromekelleher: If you want multiple mutations, might as well use the jukes cantor function. This is adding randomness without setting a seed.

python/tests/test_topology.py
jeromekelleher: I don't know --- odd things happen at the boundaries when you have very small numbers. Try out a left coordinate of DBL_MIN or someting.

Rather than jamming everything into a handful of tests, write some more tests. Try out different corner cases in different named tests --- that way, a reader of the code can see what that specific test is for.

A 'verify' function here which takes a generic tree sequence as input and checks the output seems like a good idea. Simple specific tests are good, but there needs to be a more general checks to you can verify a bunch of different examples as well.

For example, you've not tested sites that are exactly on the boundaries here.

hyanwong: Yes, you are right, I should separate some of this stuff into separate tests.

python/tests/test_topology.py
hyanwong: Good point about the seed. I didn't notice the JC function

hyanwong: Ah, but I want to place the sites *at a specific position*, which is why I rolled my own. Then I can check if they have been deleted appropriately. The tsutil.jukes_cantor doesn't allow you to specify positions.

hyanwong:
I'm getting bogged down here - I can't see an obvious way to separate out the stuff in the unit tests without substantial code duplication. Will wait until I can chat in person before addressing the issues above.

python/tests/test_topology.py
jeromekelleher: Well, don't bother with randomly choosing the nodes then, put them over the samples. Either that or set a seed on np.random (non deterministic tests are bad news)

hyanwong: Yep, over samples is sensible. I agree about non-determinism.

python/tests/test_topology.py
```diff
@@ -4984,3 +4984,120 @@ def test_ts_single_tree_delete_middle(self):
...
+            tables.mutations.add_row(
+                site, i % tables.nodes.num_rows, state)
+
+    def test_failure_with_migrations(self):
+        ts = msprime.simulate(10, recombination_rate=2, random_seed=2)
+        ts = ts.keep_intervals([[0.1, 0.5]])
+        tables = ts.dump_tables()
+        tables.migrations.add_row(0, 1, 0, 0, 0, 0)
+        ts = tables.tree_sequence()
+        with self.assertRaises(ValueError):
```
jeromekelleher: I don't think this does what you intend --- I think only the first one will be caught.

hyanwong: Ah, you are right. I was being careless

hyanwong: replaced with 

```
        self.assertRaises(ValueError, ts.ltrim)
        self.assertRaises(ValueError, ts.rtrim)
        self.assertRaises(ValueError, ts.trim)
```

python/tskit/trees.py
```diff
@@ -3298,7 +3298,7 @@ def ltrim(self, record_provenance=True):
             provenance information of the returned tree sequence. (Default: True).
         """
         tables = self.dump_tables()
-        tables.rtrim(record_provenance)
+        tables.ltrim(record_provenance)
```
jeromekelleher: *Cough* --- your tests missed this...

jeromekelleher:
I've added some tests (and fixed a bug) which should hopefully illustrate what I'm talking about. Generic code to check the result and small individual tests to check a bunch of different scenarios. Looks like there's a bug in the case where we pass in zero edges. Implementing the actual code is a small fraction of the effort here; thinking through all the ways it can go wrong or be broken and devising tests to cover these is where the real work is.

We'll want to do something similar for rtrim. Plus, we need to test out the provenance code path so that we get test coverage here.

python/tskit/trees.py
hyanwong: Oh, that's embarrassing <blush>

hyanwong:
>I've added some tests (and fixed a bug) which should hopefully illustrate what I'm talking about. 

Thanks. Really helpful.

> Looks like there's a bug in the case where we pass in zero edges.

I think we should fail here with a useful error, along the lines of "Trimming a tree sequence with no edges would reduce the sequence length to zero, which is not allowed". I'll code this up.

hyanwong:
@jeromekelleher: I can squash the commits together once you have approved my changes.

python/tskit/tables.py
```diff
@@ -1791,6 +1791,91 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
...
+
+        :param bool record_provenance: If ``True``, add details of this operation
+            to the provenance table in this TableCollection. (Default: ``True``).
+        """
+        if self.migrations.num_rows > 0:
+            raise ValueError(
+                "You cannot trim a tree sequence containing migrations")
+        if self.edges.num_rows == 0:
+            raise ValueError(
+                "Trimming a tree sequence with no edges would reduce the sequence length"
```
jeromekelleher: Need a space after "length " here, or we get "lengthto" in the output.

python/tests/test_topology.py
```diff
@@ -4984,3 +4985,236 @@ def test_ts_single_tree_delete_middle(self):
...
+            self.assertAlmostEqual(
+                source_tree.interval[0], left + trimmed_tree.interval[0])
+            self.verify_sites(source_tree, trimmed_tree, left)
+
+    def verify_rtrim(self, source_ts, trimmed_ts):
+        self.assertEqual(source_ts.num_trees, trimmed_ts.num_trees + 1)
+        for j in range(trimmed_ts.num_trees):
+            source_tree = source_ts.at_index(j)
+            trimmed_tree = trimmed_ts.at_index(j)
+            self.assertEqual(source_tree.parent_dict, trimmed_tree.parent_dict)
```
jeromekelleher: Missing check for the intervals here. Should be identical, right? (Could probably use standard equality in the case?)

jeromekelleher: Also not checking the overall sequence length (should probably be checked above in ltrim as well)

python/tskit/tables.py
```diff
@@ -1791,6 +1791,91 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
...
+    def rtrim(self, record_provenance=True):
+        """
+        Reset the ``sequence_length`` property so that the sequence ends at the end of
+        the last edge. This is identical to :meth:`TreeSequence.rtrim` but acts
+        *in place* to alter the data in this :class:`TableCollection`.
+
+        :param bool record_provenance: If ``True``, add details of this operation
+            to the provenance table in this TableCollection. (Default: ``True``).
+        """
+        if self.migrations.num_rows > 0:
```
jeromekelleher: Could abstract the checks here into a single function, ``__check_trim_conditions`` or something. Saves duplicating these two checks and error messages.

python/tests/test_topology.py
hyanwong: Yep to testing for absolute equality in `verify_rtrim`.

Sequence length will have changed after trimming, so can't be (easily) verified in `verify_ltrim` or `verify_rtrim`. Easiest to understand if this is simply checked in each function (I'll add it)

hyanwong: Yes, for testing absolute equality in `rtrim`.

Will add something in to check seq length too

python/tskit/tables.py
hyanwong: Yes, makes sense

hyanwong: Whats the tskit convention between using double underscores and single underscores for private functions? E.g. higher up in BaseTable, a single underscore is used for `_check_required_args()`

jeromekelleher: Single underscore means something that will be used by subclasses, double underscore means it's private. Might not be 100% rigorous on this though.

Issue #246, opened 2019-06-24
Add 'time' order to tree.nodes()


jeromekelleher:
It is sometimes useful to iterate over nodes in the tree in 'time' order, where we guarantee that the time of each node visited is >= the time of the previous one. Add this to the options in the ``order`` argument.

jeromekelleher:
This is trivial to do:
```python
nodes = sorted([u for u in tree.nodes()], key=lambda u: tree.time(u))
```
the trick is to do it efficiently.

benjeffery:
Thought I'd spend an evening looking at this to see what it's like working with the codebase and as it had some friendly labels! I've implemented your trivial implementation above and an additional one using a priority queue from the `heapq` module. Code at https://github.com/benjeffery/tskit/commit/5feba1b4db5574436613e091b54470765f5e741d

To make it a fairer comparison between the methods, the order of traversal is the reverse of the one specified as this lets the priority queue method start at the root. Maybe there is a use case I am unaware of that strongly motivates the original order, but (unless `sample_lists=True`) you'd have to first traverse to get the samples then work backwards.

I generated some trees to traverse with `msprime.simulate(n)` and then traversed over the first 20 trees. 
![time](https://user-images.githubusercontent.com/8552/67529924-7beaa700-f6b5-11e9-87c3-808917297207.png)
Code for the plot at https://gist.github.com/benjeffery/8c06f75effdc020b09a7ae0e2ab99246
So the priority queue method is pretty slow - this is due to it not being quite the right algorithm as we are not taking advantage of the fact that we only have need for a monotonic priority queue. In this case to doing something like a radix heap (e.g. https://github.com/iwiwi/radix-heap) would be better, but I couldn't find anything to quickly drop into python.

But do we need to? The trivial method is faster than some of the existing methods, about a ~70-80% perf hit compared to `levelorder` and uses `num_nodes * 2 * 8bytes` of additional memory to do the sort.

Anyway, that's how far I got for now, happy to document and then make a PR if needed.

jeromekelleher:
This is great, thanks @benjeffery! We should definitely work this up as a PR.

I forget what the original motivation for having this ordering is, but I think it was from an application where we needed the nodes time-increasing order. Think of it as a stronger form of postorder. I can't really see what one would use time descending down the tree for (but, perhaps we should consider the names so that both are possible?).

[...]

Issue #384, opened 2019-10-15
Iterating over edges in a ts (or EdgeTable) by age of child, not parent


hyanwong:
For `tsdate`, @awohns needs to be able to iterate over all the edges that point to a particular child node in order of the time of that child node. Currently, edges are ordered by time of *parent* node (which is also useful). Iterating over edges by child node time as well as parent node time seems a generally useful thing to want to do if we are doing pre- or post- order traversal through the ARG.

Presumably we could implement a different in-place edge sort algorithm to do this, although it would make an invalidly-sorted edge table, which would need re-sorting normally before being turned into a tree sequence. Would this be seen as a bit hacky?

jeromekelleher:
I think implementing this as an argument to the ``TableCollection.sort()`` is the right approach. If no arguments are passed, then we sort in the canonical order. We can add an option called ``edge_order``, which defaults to "parent_time" but we can also specify as "child_time". Being able to sort into a non-canonical ordering is fine --- this is an advanced operation.

hyanwong:
Great. Perfect. @awohns - would you like to work this into a PR, as you'll need it for the "outside" part of the tsdate algorithm?

awohns:
Will do, thanks for raising the issue Yan!

hyanwong:
> Will do, thanks for raising the issue Yan!

Great. Thanks. By the way, you should be able to go in reverse order after the sort by simply doing `reversed(tables.edges)` so there's no need to implement an ascending vs descending option to the sort params.

hyanwong:
I'm wondering if it would be better to provide a parameter to the [`ts.edges()`](https://github.com/tskit-dev/tskit/blob/002133a2c2934bb207072431fc47fe61cdb34958/python/tskit/trees.py#L2532) method, see my comment [here](https://github.com/tskit-dev/tskit/pull/399#issuecomment-549875834)

jeromekelleher:
Can we close this @hyanwong, or is this something you think is still worth doing?

hyanwong:
After a bit of experimenting, it turns out that it's just as quick to access the edges in an arbitrary order using `ts.edge(i)` than is it to use the `edges()` iterator:

```
import msprime
ts = msprime.simulate(100000, length=10000, Ne=10000, recombination_rate=0.00001, random_seed=123)
```

In a jupyter cell:

```
%%timeit
for e in ts.edges():
    p = e.parent
```
[...]

PR #399, opened 2019-11-05
Add node traversal methods based on time ordering

benjeffery:
Addresses #246 using a very simple implementation as discussed. I went with `timeasc` and `timedesc` as the possible values for `order=` as they seemed to fit the existing values.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1621,6 +1621,23 @@ class TestTree(HighLevelTestCase):
             tree = next(ts.trees())
             self.verify_traversals(tree)
 
+            # To verify time-ordered traversal we can't use the method used for the
+            # other traversals above, it checks for one-to-one correspondence.
+            # As more than one ordering is valid for time, we do it separately here
+            for root in tree.roots:
+                time_ordered = tree.nodes(root, order="timeasc")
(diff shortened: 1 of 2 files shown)


benjeffery:
One thing I forgot to add - this treats each root separately, so if you call without a node and the tree has more than one root then the sort is only ordered for each root, not globally. This is how the other sort methods work, but I think it might make less sense for time ordering.

hyanwong:
Hi @benjeffery: thanks for getting involved with tskit development. If your head is in the ordering methods, you might like to take a stab at https://github.com/tskit-dev/tskit/issues/384 - there are definite use cases for visiting the *edges* of a tree sequence (rather than the nodes of a tree) in order of (a) child node times and (b) parent node times. In that issue I suggested simply re-ordering the edge table, but since we might want to keep the edge IDs consistent between the two methods, I wonder if providing an `order` argument to the [`ts.edges()`](https://github.com/tskit-dev/tskit/blob/002133a2c2934bb207072431fc47fe61cdb34958/python/tskit/trees.py#L2532) method might be better?

jeromekelleher:
This looks great, thanks @benjeffery. I think we'll need some more documentation on what the orders are, but I'll open an issue to track that.

PR #400, opened 2019-11-05
Perf improvments to node traversals

benjeffery:
Roughly halves time taken for node traversals with the `preorder`, `postorder` and `levelorder` methods.

See #246 for discussion as this came up in the context of adding a time-ordered node traversal.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1416,6 +1416,8 @@ class Tree(object):
         pop = stack.pop
         extend = stack.extend
         get_children = self.children
+        # Note: the usual style is to be explicit about what we're testing
+        # and use while len(stack) > 0, but this form is slightly faster.
         while stack:
             v = pop()
             extend(reversed(get_children(v)))
@@ -1429,6 +1431,8 @@ class Tree(object):
         extend = stack.extend
         get_children = self.children
         get_parent = self.get_parent
+        # Note: the usual style is to be explicit about what we're testing
+        # and use while len(stack) > 0, but this form is slightly faster.
         while stack:
             v = stack[-1]
             children = [] if v == parent else get_children(v)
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -1411,23 +1411,32 @@ def num_tracked_samples(self, u=None):
...
-        while len(stack) > 0:
-            v = stack.pop()
-            if self.is_internal(v):
-                stack.extend(reversed(self.children(v)))
+        stack = collections.deque([u])
+        # For perf we store these to avoid lookups in the tight loop
+        pop = stack.pop
+        extend = stack.extend
+        get_children = self.children
+        while stack:
```
jeromekelleher: Minor style quibble: we prefer to avoid 'Falsey', unless we're really using the fact that might evaluate to False in different contexts. Here, I find ``while len(stack) > 0`` to be much more readable: it's totally explicit, and the reader doesn't have to think "is this going to be False because stack is a boolean, or it might be None, or it might be a list of length zero?". It's pretty obvious here since the method is so short, but still I think it's easier to read when we're totally explicit about the conditional.

benjeffery: Completely agree - the intent is much clearer with `while len(stack) > 0`. 

I've used `while stack` here as you take ~10% perf hit from using the former as it has an extra python function call (2 calls for `postorder` if you also replace the `if children`). 

This level of optimisation might be overkill, given (as you said), that as soon as users start doing anything this part won't be the dominant term. Maybe a comment with the intent might be a good solution? Happy either way.

jeromekelleher: Aha, I see. I didn't think there'd be that much difference. You're right, the best thing to do here is add a comment saying we do this for perf.

benjeffery: Comments added, rebased to master.

PR #406, opened 2019-11-20
Change map_ancestors to link_ancestors

### LINKED ISSUE #262: Rename map_ancestors to map_to_ancestors?

hyanwong:
but leave `map_ancestors` in the python API as a deprecated (undocumented) alias. Fixes #262

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -3551,16 +3551,32 @@ class TestMapToAncestors(unittest.TestCase):
 
     def do_map(self, ts, ancestors, samples=None, compare_lib=True):
         """
-        Runs the Python test implementation of map_ancestors.
+        Runs the Python test implementation of link_ancestors.
         """
         if samples is None:
             samples = ts.samples()
         s = tests.AncestorMap(ts, samples, ancestors)
-        ancestor_table = s.map_ancestors()
+        ancestor_table = s.link_ancestors()
         if compare_lib:
-            lib_result = ts.tables.map_ancestors(samples, ancestors)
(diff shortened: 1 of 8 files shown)


python/tests/test_topology.py
```diff
@@ -3551,17 +3551,33 @@ class TestMapToAncestors(unittest.TestCase):
...
+    def test_deprecated_name(self):
+        # copied from test_single_tree_one_ancestor below
+        nodes = io.StringIO(self.nodes)
+        edges = io.StringIO(self.edges)
+        ts = tskit.load_text(nodes=nodes, edges=edges, strict=False)
+        samples = ts.samples()
+        ancestors = [8]
+        s = tests.AncestorMap(ts, samples, ancestors)
+        tss = s.link_ancestors()
+        lib_result = ts.tables.link_ancestors(samples, ancestors)
```
jeromekelleher: Not actually checking the deprecated alias here.

hyanwong: Doh - sorry. Went overboard with the search and replace. Have just changed it.

jeromekelleher:
Still broken, looks like it's missing a return statement.

hyanwong:
> Still broken, looks like it's missing a return statement.

Oh, really sorry - was trying to rush the fix before heading out. Will fix and test locally before pushing an update.

Issue #389, opened 2019-10-16
Given a tree sequence, draw marginal trees emphasising similarity between neighbouring trees


brianzhang01:
Currently we have two main tree sequence drawing methods: text and SVG. Each of these has a method to draw the entire tree sequence, which ends up drawing the marginal trees. Alternatively, one can iterate through the trees and call the appropriate draw method on the marginal trees.

However, the current methods of drawing do not emphasise the similarity between neighbouring trees. I have a proposal for this which is sketched out in [this repo](https://github.com/brianzhang01/tskit-viz). If you run `python sort.py`, you get the current output of drawing an `msprime`-generated tree sequence (displaying as a screenshot due to #189 ).

![Screen Shot 2019-10-16 at 4 28 31 PM](https://user-images.githubusercontent.com/7853915/66934230-143bb880-f032-11e9-9124-5e188362c3c8.png)

Going from the first to the second tree, all that has changed is that the edge from 2 to 14 has disappeared, and 2 has instead coalesced onto the edge from 15 to 20 to create a new node, 17. However, many other things are shifting around in the visualisation, obscuring the fundamental diff. Similarly, from the second to the third tree, all that has happened is that node 20 has been replaced by a node 19 of more recent time. The order of the samples could have stayed the same, but instead the nodes 18 and 17 are reversed in order, shifting the samples underneath.

Running the Python code (`python sort.py`), we get an ordering on the samples of each of the marginal trees that has much better display properties.
```
[0, 4, 1, 2, 7, 3, 5, 6, 9, 8]
[0, 4, 1, 2, 3, 7, 5, 6, 9, 8]
[0, 4, 1, 2, 3, 7, 5, 6, 9, 8]
```
If you imagine these sample orderings being used to draw the marginal trees, then the SPR (subtree pruning and regrafting) from tree 1 to 2 can be seen more clearly, and from tree 2 to 3, one can easily see that there is no change in topology, only a change in the height of the root node.

The operation I propose is to find the ordering of samples, consistent with the tree topology, that is minimum in terms of lexicographic ordering. The details of the implementation are in the [Github repo](https://github.com/brianzhang01/tskit-viz).

[...]

Issue #401, opened 2019-11-06
Document traversal orders


jeromekelleher:
The ``Tree.nodes`` iterator doesn't provide any description of what the various traversal orders actually are. Probably we should have a description list within the method to describe each one, possibly with links out to Wikipedia or somewhere.

brianzhang01:
Where specifically would the documentation be added in order to be viewable by a user? If it's straightforward it could be added to #411, though I'm already behind on getting the PR out.

jeromekelleher:
I think it can go directly into the docstring for the nodes function, so that the list will end up in [here](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.Tree.nodes)

Thanks @brianzhang01!

brianzhang01:
Cool, I'll add it to my PR. Feel free to assign me.

PR #411, opened 2019-11-21
Add minimum lexicographic order traversal for nodes

brianzhang01:
This is a first commit for #389 . When this is in, we can talk about changing the tree drawing code itself.

I hope tests pass! On my machine, I was getting one failing test.
```
ERROR: test_nonrectangular_input (tests.test_util.TestNumpyArrayCasting)
  File "/Users/brian/code/python/tskit/python/tests/test_util.py", line 113, in test_nonrectangular_input
    util.safe_np_int_cast(bad_input, dtype)
  File "/Users/brian/code/python/tskit/python/tskit/util.py", line 41, in safe_np_int_cast
    int_array = np.array(int_array)
ValueError: setting an array element with a sequence.

```
The test expects a `TypeError` but gets a `ValueError` instead. I'm using Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 16:30:03) and numpy 1.15.0.

### DIFF
--- a/python/tests/__init__.py
+++ b/python/tests/__init__.py
@@ -110,6 +110,30 @@ class PythonTree:
         for c in self.children(u):
             self._levelorder_nodes(c, l, level + 1)
 
+    def _minlex_postorder_nodes(self, u, l):
+        l.extend(self._minlex_postorder_nodes_helper(u)[1])
+
+    def _minlex_postorder_nodes_helper(self, u):
+        """
+        For a given input ID u, this function returns a tuple whose first value
+        is the minimum leaf node ID under node u, and whose second value is
+        a list containing the minlex postorder for the subtree rooted at node u.
+        The first value is needed for sorting, and the second value is what
+        finally gets returned.
+        """
+        children = self.children(u)
+        if len(children) > 0:
+            children_return = [self._minlex_postorder_nodes_helper(c) for c in children]
+            # sorts by first value, which is the minimum leaf node ID
+            children_return.sort()
+            minlex_postorder = []
+            for _, child_minlex_postorder in children_return:
+                minlex_postorder.extend(child_minlex_postorder)
+            minlex_postorder.extend([u])
+            return (children_return[0][0], minlex_postorder)
+        else:
+            return (u, [u])
+
     def nodes(self, root=None, order="preorder"):
         roots = [root]
         if root is None:
@@ -127,6 +151,8 @@ class PythonTree:
                 # Nested list comprehension flattens node_list in order
                 self._levelorder_nodes(u, node_list, 0)
                 node_list = iter([i for level in node_list for i in level])
+            elif order == "minlex_postorder":
+                self._minlex_postorder_nodes(u, node_list)
             else:
                 raise ValueError("order not supported")
             yield from node_list
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1888,9 +1888,8 @@ class TestTree(HighLevelTestCase):
             tree = next(ts.trees())
             self.verify_traversals(tree)
 
-            # To verify time-ordered traversal we can't use the method used for the
-            # other traversals above, it checks for one-to-one correspondence.
-            # As more than one ordering is valid for time, we do it separately here
+            # Verify time-ordered traversals separately, because the PythonTree
+            # class does not contain time information at the moment
             for root in tree.roots:
                 time_ordered = tree.nodes(root, order="timeasc")
                 t = tree.time(next(time_ordered))
@@ -1909,7 +1908,13 @@ class TestTree(HighLevelTestCase):
         t1 = tree
         t2 = tests.PythonTree.from_tree(t1)
         self.assertEqual(list(t1.nodes()), list(t2.nodes()))
-        orders = ["inorder", "postorder", "levelorder", "breadthfirst"]
+        orders = [
+            "inorder",
+            "postorder",
+            "levelorder",
+            "breadthfirst",
+            "minlex_postorder",
+        ]
         if tree.num_roots == 1:
             self.assertRaises(ValueError, list, t1.nodes(order="bad order"))
             self.assertEqual(list(t1.nodes()), list(t1.nodes(t1.get_root())))
@@ -1946,7 +1951,10 @@ class TestTree(HighLevelTestCase):
                         list(t2.nodes(root, order=test_order)),
                     )
                     all_nodes.extend(t1.nodes(root, order=test_order))
-                self.assertEqual(all_nodes, list(t1.nodes(order=test_order)))
+                # minlex_postorder reorders the roots, so this last test is
+                # not appropriate
+                if test_order != "minlex_postorder":
+                    self.assertEqual(all_nodes, list(t1.nodes(order=test_order)))
 
     def test_total_branch_length(self):
         # Note: this definition works when we have no non-sample branches.
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1564,12 +1564,82 @@ class Tree:
             yield v
 
     def _timeasc_traversal(self, u):
-        yield from sorted(self.nodes(u, order="levelorder"), key=self.time)
+        """
+        Sorts by increasing time but falls back to increasing ID for equal times.
+        """
+        yield from sorted(
+            self.nodes(u, order="levelorder"), key=lambda u: (self.time(u), u)
+        )
 
     def _timedesc_traversal(self, u):
+        """
+        Sorts by decreasing time but falls back to decreasing ID for equal times.
+        """
         yield from sorted(
-            self.nodes(u, order="levelorder"), key=self.time, reverse=True
+            self.nodes(u, order="levelorder"),
+            key=lambda u: (self.time(u), u),
+            reverse=True,
         )
+
+    def _minlex_postorder_traversal(self, u):
+        """
+        Postorder traversal that visits leaves in minimum lexicographic order.
+
+        Minlex stands for minimum lexicographic. We wish to visit a tree in such
+        a way that the leaves visited, when their IDs are listed out, have
+        minimum lexicographic order. This is a useful ordering for drawing
+        multiple Trees of a TreeSequence, as it leads to more consistency
+        between adjacent Trees.
+        """
+        # We skip perf optimisations here (compared to _preorder_traversal and
+        # _postorder_traversal) as this ordering is unlikely to be used in perf
+        # sensitive applications
+        stack = collections.deque([u])
+        parent = NULL
+
+        # We compute a dictionary mapping from internal node ID to min leaf ID
+        # under the node, using a first postorder traversal
+        min_leaf_dict = {}
+        while len(stack) > 0:
+            v = stack[-1]
+            children = [] if v == parent else self.children(v)
+            if children:
+                # The first time visiting a node, we push its children onto the stack.
+                # reversed is not strictly necessary, but it gives the postorder
+                # we would intuitively expect.
+                stack.extend(reversed(children))
+            else:
+                # The second time visiting a node, we record its min leaf ID
+                # underneath, pop it, and update the parent variable
+                if v != parent:
+                    # at a leaf node
+                    min_leaf_dict[v] = v
+                else:
+                    # at a parent after finishing all its children
+                    min_leaf_dict[v] = min([min_leaf_dict[c] for c in self.children(v)])
+                parent = self.get_parent(v)
+                stack.pop()
+
+        # Now we do a second postorder traversal
+        stack.clear()
+        stack.extend([u])
+        parent = NULL
+        while len(stack) > 0:
+            v = stack[-1]
+            children = [] if v == parent else self.children(v)
+            if children:
+                # The first time visiting a node, we push onto the stack its children
+                # in order of reverse min leaf ID under each child. This guarantees
+                # that the earlier children visited have smaller min leaf ID,
+                # which is equivalent to the minlex condition.
+                stack.extend(
+                    sorted(children, key=lambda u: min_leaf_dict[u], reverse=True)
+                )
+            else:
+                # The second time visiting a node, we pop and yield it, and
+                # we update the parent variable
+                parent = self.get_parent(v)
+                yield stack.pop()
 
     def nodes(self, root=None, order="preorder"):
         """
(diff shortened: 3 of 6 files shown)


brianzhang01:
I just commented on issue #389 . Here's a relevant excerpt for discussion.

> I've now decided that the right way to do this is to output a minimum lexicographic order by leaves, not samples, so this is resolved. However, another issue is that sometimes there can apparently be multiple roots for TreeSequence / Tree objects. Right now I only do minlex ordering within each root, and the roots are iterated over using for u in roots: (trees.py, nodes() function). Should we add separate logic to also go over the roots in a minlex order?

Also, for the numpy-related test that fails for me, we may want to consider broadening the test to be fine with either a ValueError or TypeError. The test is running on Mac OS X.

jeromekelleher:
This looks great, thanks @brianzhang01! We're currently tidying things up to release 0.2.3 tomorrow, so do you mind if we pick this up after that?

One thing that would be nice here is to make a test that explicitly shows what the ordering is for a specific tree, using something like [this](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_topology.py#L2655). We could add a test class called ``TestTraversalOrder`` to ``test_topology.py`` which tests this for an example tree, and just checks that we get the expected ordering.

We haven't seen the numpy failure on Travis, but sure, checking for a TypeError as well as ValueError would be fine (feel free to add a commit in here do it).

brianzhang01:
I just added the tests with an explicit tree. Yes of course, feel free to punt this until after the release.

I noticed while going through things that in the tests (test_highlevel.py:TestTree, test_traversals and verify_traversals), `timeasc` and `timedesc` are treated differently than the other tests because it's a loose criteria -- i.e. there can be multiple orderings that fulfill the criteria. This also meant that I didn't test these two orders in `test_topology.py`. However, we could consider sharpening the specification for `timeasc`, where it first tries to sort by increasing time and then by increasing node ID, if two times are the same. It shouldn't be a big change.

Also, there is still an open question from earlier about whether minlex_postorder should also try to go over the multiple roots in a tree in a minlex way, based on the minimum leaf under that root. I'd prefer to not do it because it's not really needed for the drawing code, but it would make more sense.

jeromekelleher:
> Also, there is still an open question from earlier about whether minlex_postorder should also try to go over the multiple roots in a tree in a minlex way, based on the minimum leaf under that root. I'd prefer to not do it because it's not really needed for the drawing code, but it would make more sense.

This would be just a case of sorting the roots before going through them though, wouldn't it? We might as well in that case.

> I noticed while going through things that in the tests (test_highlevel.py:TestTree, test_traversals and verify_traversals), timeasc and timedesc are treated differently than the other tests because it's a loose criteria -- i.e. there can be multiple orderings that fulfill the criteria. This also meant that I didn't test these two orders in test_topology.py. However, we could consider sharpening the specification for timeasc, where it first tries to sort by increasing time and then by increasing node ID, if two times are the same. It shouldn't be a big change.

Well spotted. I guess we should order by node ID as well, as you say. It's not a big change and it's good to be concrete about these things.

brianzhang01:
Cool, I'll work on adding both of these features. Just a few more notes on the timeasc / timedesc orderings. First of all, if I'm not mistaken, shouldn't sorting by node ID alone suffice to sort by time as well, since I thought the nodes were ordered by increasing time? Second, the current implementations of these are O(N log N) if N is the number of nodes in a tree, which is slower than the other traversal methods which are all O(N) I think. Instead of a full sort, one can also use a priority queue from root, but I think that would also end up being O(N log N).

Third, for timeasc / timedesc, we might also want to reorder the results of searching from each root. For instance, if there are 3 roots, timeasc / timedesc should maybe get all the nodes under all the roots and then sort by time / node ID. As a reminder, here's what the code for `nodes()` roughly looks like:
```
        methods = {
            "preorder": self._preorder_traversal,
            "inorder": self._inorder_traversal,
            "postorder": self._postorder_traversal,
            "levelorder": self._levelorder_traversal,
            "breadthfirst": self._levelorder_traversal,
            "timeasc": self._timeasc_traversal,
            "timedesc": self._timedesc_traversal,
            "minlex_postorder": self._minlex_postorder_traversal
        }
        iterator = methods[order]
        for u in roots:
            for v in iterator(u):
                yield v
```

So for minlex_postorder, and possibly for timeasc / timedesc, we need to add some logic in the nodes() function, not just in the specific methods for each traversal.

jeromekelleher:
> First of all, if I'm not mistaken, shouldn't sorting by node ID alone suffice to sort by time as well, since I thought the nodes were ordered by increasing time? 

No, it's not a requirement that nodes are in time order, they can be mixed up arbitrarily.

> Second, the current implementations of these are O(N log N) if N is the number of nodes in a tree, which is slower than the other traversal methods which are all O(N) I think. Instead of a full sort, one can also use a priority queue from root, but I think that would also end up being O(N log N).

Yes, but in practise it's surprisingly fast, see the plots in #246. We tried various clever algorithms which ended up being slower than the simple sort.

> Third, for timeasc / timedesc, we might also want to reorder the results of searching from each root. For instance, if there are 3 roots, timeasc / timedesc should maybe get all the nodes under all the roots and then sort by time / node ID. As a reminder, here's what the code for nodes() roughly looks like:

I'm not sure about this one. It's certainly more convenient to do it root-by-root, but I wonder if this is a particularly useful property for the other orderings. It's not going to complicate things much if we pull iterating over the roots into the ordered traversal methods, and if you **do** want to have separate ordered traversals for each root, then you can always do:
```
for root in tree.roots:
    for u in tree.nodes(u, order="timeasc"):
        # do your thing.
```

brianzhang01:
So a few changes just went in:

- Calling nodes() with minlex_postorder now first gets the minlex postorder ordering under each root, then sorts among the roots, before finally outputting a global minlex postorder.
- timeasc and timedesc traversals sort first using ascending / descending time, but fall back on ascending / descending node ID if two nodes have the same time.
- Got rid of some custom logic in test_highlevel.py:TestTree:test_traversals for the timeasc / timedesc cases. Because the sorting is now deterministic, we can instead use the verify_traversals function, I'm pretty sure.
- The TestTraversalOrder set of tests now includes tests for timeasc and timedesc in the method test_traversal_order (which tests all traversals), as well as a new test function test_minlex_postorder_multiple_roots designed to test the minlex postorder with multiple roots case.

I have left out the last thing we were discussing (whether timeasc and timedesc should be global, not just under each root) because it didn't seem like there was a strong argument for it.

python/tests/test_highlevel.py
```diff
@@ -1631,28 +1631,33 @@ def test_traversals(self):
...
-                    next_t = tree.time(u)
-                    self.assertGreaterEqual(next_t, t)
-                    t = next_t
-                time_ordered = tree.nodes(root, order="timedesc")
-                t = tree.time(next(time_ordered))
-                for u in time_ordered:
-                    next_t = tree.time(u)
-                    self.assertLessEqual(next_t, t)
+            # Verify time-ordered traversals separately, because the PythonTree
+            # class does not contain time information at the moment
```
brianzhang01: Note the reason given here

python/tests/test_highlevel.py
```diff
@@ -1686,7 +1691,10 @@ def verify_traversals(self, tree):
                         list(t1.nodes(root, order=test_order)),
                         list(t2.nodes(root, order=test_order)))
                     all_nodes.extend(t1.nodes(root, order=test_order))
-                self.assertEqual(all_nodes, list(t1.nodes(order=test_order)))
+                # minlex_postorder reorders the roots, so this last test is
+                # not appropriate
```
brianzhang01: Note the comment and the changed test

jeromekelleher:
Great, thanks @brianzhang01. @hyanwong, can you take a look and review here please?

python/tests/__init__.py
```diff
@@ -113,6 +113,22 @@ def _levelorder_nodes(self, u, l, level):
...
+    def _minlex_postorder_nodes(self, u, l):
+        l.extend(self._minlex_postorder_nodes_helper(u)[1])
+
+    def _minlex_postorder_nodes_helper(self, u):
+        children = self.children(u)
+        if children:
+            children_return = [self._minlex_postorder_nodes_helper(c) for c in children]
+            children_return.sort()
+            postorder = []
+            for element in children_return:
```
jeromekelleher: ``element`` isn't very helpful here, how about 
```python
nodes = []
for _, child_nodes in sorted_children:
    nodes.extend(child_nodes)
nodes.append(u)
```

python/tests/__init__.py
```diff
@@ -113,6 +113,22 @@ def _levelorder_nodes(self, u, l, level):
         for c in self.children(u):
             self._levelorder_nodes(c, l, level + 1)
 
+    def _minlex_postorder_nodes(self, u, l):
+        l.extend(self._minlex_postorder_nodes_helper(u)[1])
+
+    def _minlex_postorder_nodes_helper(self, u):
```
jeromekelleher: A quick docstring describing what this function returns would be helpful I think.

brianzhang01:
We've agreed for this PR to include a fix for #401 as well

brianzhang01:
I'll be going through this soon and will push an update in a bit, but I thought @benjeffery might be interested in staying looped in?

jeromekelleher:
Thanks @brianzhang01, it would be great to get this one squared away! One thing to note is that we've added some automated code formatting since this started. It's worth having a quick look at the developer [docs](https://tskit.readthedocs.io/en/latest/development.html#quickstart) so you can get ``pre-commit`` working.

Feel free to ping us here or on Slack (did you get an invite?) if you hit issues bringing the PR up to date.

python/tests/__init__.py
brianzhang01: I've opted for some longer names instead. Hopefully with the added docstring things are clear.

python/tskit/trees.py
```diff
@@ -1490,10 +1490,78 @@ def _levelorder_traversal(self, u):
...
+        parent = NULL
+        # For perf we store these to avoid lookups in the tight loop
+        pop = stack.pop
+        extend = stack.extend
+        get_children = self.children
+        get_parent = self.get_parent
+
+        # We store a dictionary from internal node ID to min leaf under the node
+        # First we compute this dictionary using a postorder traversal
+        min_leaf_dict = {}
```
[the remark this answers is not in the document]
brianzhang01: No, we can't compute it using `nodes(order="postorder")`, because the postorder underneath also needs to be minlex. However, I have gotten rid of the perf optimisations and added lots of comments. I forgot a lot of what the code was doing as well!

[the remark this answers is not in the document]
brianzhang01: If we want to make things even more readable, there is a recursive solution in tests/__init__.py, but I think we should still stick to an iterative solution here.

python/tskit/trees.py
```diff
@@ -1490,10 +1490,78 @@ def _levelorder_traversal(self, u):
...
+        stack.clear()
+        extend([u])
+        parent = NULL
+        # Note: the usual style is to be explicit about what we're testing
+        # and use while len(stack) > 0, but this form is slightly faster.
+        while stack:
+            v = stack[-1]
+            children = [] if v == parent else get_children(v)
+            if children:
+                children_min = [[min_leaf_dict[c], c] for c in get_children(v)]
```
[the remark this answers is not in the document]
brianzhang01: Thanks, that's great! I also added a `reverse=True`, which is required (tests fail without it, which means we have good tests).

python/tests/__init__.py
```diff
@@ -110,6 +110,30 @@ def _levelorder_nodes(self, u, l, level):
...
+        children = self.children(u)
+        if len(children) > 0:
+            children_return = [self._minlex_postorder_nodes_helper(c) for c in children]
+            # sorts by first value, which is the minimum leaf node ID
+            children_return.sort()
+            minlex_postorder = []
+            for _, child_minlex_postorder in children_return:
+                minlex_postorder.extend(child_minlex_postorder)
+            minlex_postorder.extend([u])
+            return (children_return[0][0], minlex_postorder)
```
brianzhang01, line 133: Note that we can't use sorted(children_return) above because we need to get the first element here as well

python/tskit/trees.py
```diff
@@ -1564,13 +1564,83 @@ def _levelorder_traversal(self, u):
...
+                # The first time visiting a node, we push onto the stack its children
+                # in order of reverse min leaf ID under each child. This guarantees
+                # that the earlier children visited have smaller min leaf ID,
+                # which is equivalent to the minlex condition.
+                stack.extend(
+                    sorted(children, key=lambda u: min_leaf_dict[u], reverse=True)
+                )
+            else:
+                # The second time visiting a node, we pop and yield it, and
+                # we update the parent variable
```
brianzhang01, line 1640: Note the copious comments that have been added to this function

python/tskit/trees.py
```diff
@@ -1582,10 +1652,45 @@ def nodes(self, root=None, order="preorder"):
...
+            - 'minlex_postorder': a usual postorder has ambiguity in the order in
+              which children of a node are visited. We constrain this by outputting
+              a postorder such that the leaves visited, when their IDs are
+              listed out, have minimum `lexicographic order
+              <https://en.wikipedia.org/wiki/Lexicographical_order>`__ out of all valid
+              traversals. This traversal is useful for drawing multiple trees of
+              a ``TreeSequence``, as it leads to more consistency between adjacent
+              trees. Note that internal non-leaf nodes are not counted in
+              assessing the lexicographic order.
+
```
brianzhang01, line 1689: I went with a bulleted list style. However, there is another style used in the documentation that I found: https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.allele_frequency_spectrum. Is it fine to keep it as a bulleted list? Is the documentation too verbose?

python/tskit/trees.py
```diff
@@ -1605,8 +1711,20 @@ def nodes(self, root=None, order="preorder"):
         roots = [root]
         if root is None:
             roots = self.roots
-        for u in roots:
-            yield from iterator(u)
+        if order == "minlex_postorder" and len(roots) > 1:
+            # we need to visit the roots in minlex order as well
+            # we first visit all the roots and then sort by the min value
```
brianzhang01, line 1716: This multi-root case is not documented prominently in the nodes() docstring, should we highlight it? It's unlikely that users would be using the function in the multiple roots case.

hyanwong:
Sorry - just seen that I was asked to review this in Dec. it seems like most is sorted, though. Do you need any more feedback @brianzhang01 ?

brianzhang01:
@hyanwong I think it's all right -- there's a lot here to be brought up to speed with. But if Ben wants to have a look I think that could be worthwhile as I've built off some of the edits he recently made.

I'll work on fixing the CI errors.

python/tskit/trees.py
hyanwong: Better too much detail than too little. The wikipedia pages are very helpful for tree traversal order explanations, so I'm happy those are in there.

hyanwong:
More generally, there are no tests for the order when there are multiple root trees under any of the other sorting methods. Should there be? I guess that would not be strictly part of this PR though?

brianzhang01:
I'm ready for another review pass on this, @jeromekelleher . I still haven't rebased but will do that once I get an OK on the PR.

jeromekelleher:
Sorry I've been slow to get to this @brianzhang01, will look tomorrow.

python/tskit/trees.py
jeromekelleher: Very nice, thanks!

python/tskit/trees.py
jeromekelleher: This is excellent, thanks @brianzhang01!

python/tskit/trees.py
jeromekelleher: I don't think we need to worry about it, as you say.

jeromekelleher:
When you're updating, you need to bring the kastore submodule up to date. I think the simplest way to do this is to run ``git submodule update``. It should disappear from your diff then.

Also, can you update the CHANGELOG to include this new feature please? (Probably easiest to do this after you've rebased to avoid conflicts, you can then add it to your squashed commit using ``git commit --amend``)

brianzhang01:
Thanks @jeromekelleher ! I'll hopefully rebase and squash and add a line to the CHANGELOG tonight and then you can merge at your convenience. Will ping on the Slack (which I've been invited to but not yet joined) or email if I have any questions.

You mentioned adding the new behaviour for "time_asc" / "time_desc" into the commit description. Here's a commit description I sketched out. For the first line, should I go more general and say "Update and document traversal orders for nodes in a Tree"? Similarly, should I update the PR description to something more general? For the CHANGELOG description, I will focus on the minimum lexicographic order and leave out the "time_asc" / "time_desc" change, unless told otherwise.

```
Add minimum lexicographic order traversal for nodes in a Tree

We add a new traversal order in the Tree.nodes() function called "minlex_postorder". "minlex" is short for "minimum lexicographic", and this traversal generates a postorder which out of all possible postorders, visits the leaves in minimum lexicographic order. This is helpful for drawing neighbouring trees in as consistent a way as possible, see #389.

The "time_asc" and "time_desc" traversal orders in Tree.nodes() have also been updated to sort first by time, then fall back to sorting by increasing or decreasing node ID respectively. This helps make the orderings deterministic.

With the updated orderings, we also add explicit tests for the traversal orders of pre-defined trees, including some corner cases like "inorder" with polytomies and "minlex_postorder" with multiple roots.

Finally, we add documentation for all the traversal orders, with links to Wikipedia as appropriate. This closes #401.
```

jeromekelleher:
Thanks @brianzhang01. The notes in the commit messages can be pretty cursory, it's just a short record for the future.

jeromekelleher:
Merged, thanks @brianzhang01 !

hyanwong:
@brianzhang01 - is there any reason why we don't have minlex_preorder? So that after visiting a node we visit its children in minlex order? I think that might be useful for my svg output.

brianzhang01:
@hyanwong Jerome suggested adding different traversals in the .nodes() function. I took a look and saw that the tskit drawing code only used postorder, and since we wanted to just replace that behaviour, I decided to only do minlex_postorder. If you want to add minlex_preorder, feel free to do so.

hyanwong:
OK, thanks.

PR #412, opened 2019-11-21
Use Sphinx currentmodule to rationalise & allow :attr: refs

hyanwong:
Some weird undocumented behaviour in sphinx means that `:attr:` doesn't work the same as `:meth:`, so while we can refer to

```:meth:`.TreeSequence.simplify` ```

we can't simply do ```:attr:`.TreeSequence.num_sites` ```

instead, we have to fully cite the path, e.g. 

```:attr:`tskit.TreeSequence.num_sites` ```

This leads to a load of sphinx warnings like `BDI/tskit/python/tskit/trees.py:docstring of tskit.Tree.seek:8: WARNING: py:attr reference target not found: TreeSequence.sequence_length`, and means that in the current docs, some links aren't working, e.g. see the broken link at https://tskit.readthedocs.io/en/latest/python-api.html#tskit.Tree.seek .

My suggestion to correct this is to use `.. currentmodule:: tskit` at the top of all relevant .rst files, which means we can then switch, and (optionally) drop the leading `.` from both method and attribute names. 

A further implication of the undocumented behaviour is that *within* a class such as `Tree`, we can't simply refer to attributes as ```:attr:`.num_roots` ```, but still have to qualify it as ```:attr:`tskit.Tree.num_roots` ```, or, with the `currentmodule` addition, simply ```:attr:`Tree.num_roots` ```. Since this is documentation *within* the Tree class, it looks weird to refer to **`Tree.num_roots`** written out in full in the docs for that class, and so I've used the sphinx tilde shortcut to remove the `Tree` prefix in these cases (i.e. ```:attr:`~Tree.num_roots` ```)

This PR implements the above changes in the files that cause the warnings, and means that we no longer get any warning output. For consistency, I've removed the (now optional) `.` that prefixes the class names too. However, I haven't done this for all the .py files, as it would be rather a large change - it's only been done in trees.py. If the approach is deemed correct, I suggest that I add the `currentmodule` directive to all the `.rst` files and remove the `.` prefix from the remaining python docstrings in tskit.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = #-W
+SPHINXOPTS    = -W
 SPHINXBUILD   = python3 -m sphinx
 SOURCEDIR     = .
 BUILDDIR      = _build
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -1,3 +1,4 @@
+.. currentmodule:: tskit
 .. _sec_data_model:
 
 ##########
@@ -20,6 +21,8 @@ store tree sequences on disk in the `Tree sequence file format`_ section.
 
 
 .. _sec_data_model_definitions:
+
+
 
 ***********
 Definitions
@@ -50,7 +53,7 @@ node
 individual
     In certain situations we are interested in how nodes (representing
     individual homologous genomes) are grouped together into individuals
-    (e.g., two nodes per diploid individual). For example, when we are working
+    (e.g. two nodes per diploid individual). For example, when we are working
     with polyploid samples it is useful to associate metadata with a specific
     individual rather than duplicate this information on the constituent nodes.
     See :ref:`sec_nodes_or_individuals` for more discussion on this point.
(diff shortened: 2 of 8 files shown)


hyanwong:
NB - this now allows us to turn on the `-W` switch for sphinx, which is currently commented out in `docs/Makefile`. I could add this to the PR if you like, @jeromekelleher

jeromekelleher:
This is a big improvement, thanks @hyanwong! Please do add the ``-W`` switch; things will decay pretty quickly otherwise...

hyanwong:
`-W` switch added and other files changed to be consistent (e.g. `.. currentmodule:: tskit` added at the top of .rst files that reference the python classes & methods. 

I've taken the executive decision to remove all leading `.` from the references to classes such as ```:class:`.TreeSequence` ``` from all docstrings. I've also *put in* the (more explicit) ```:data:`tskit.NULL` ``` rather than (as we previously had) ```:data:`.NULL` ```, as I think reading the `tskit` prefix is helpful here. Similarly with ```:func:`tskit.load_text` ```, as it helps having the `tskit` in there to point out it isn't a method of the current class.

jeromekelleher:
Superb, thanks. Will merge once CI is done.

PR #414, opened 2019-11-21
First pass at basic cli docs

### LINKED ISSUE #160: Document CLI
Need documentation for the CLI. Should be similar to the documentation used for the msprime cli.

hyanwong:
This is not at all hand-crafted, and therefore a bit confusing, but at least it's something. (Partially) fixes #160

Note that due to a bug in sphinx-argparse (which I've submitted a PR to), we have to comment out the entire cli subparser for FASTA output if we want to remove it from the CLI documentation. I've done this on the assumption that no-one is currently using FASTA output via the CLI anyway.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,0 +1,11 @@
+.. currentmodule:: tskit
+.. _sec_cli:
+
+======================
+Command line interface
+======================
+
+.. argparse::
+    :module: tskit.cli
+    :func: get_tskit_parser
+    :prog: python3 -m tskit
--- a/python/tskit/cli.py
+++ b/python/tskit/cli.py
@@ -152,6 +152,7 @@ def add_precision_argument(parser):
 
 def get_tskit_parser():
     top_parser = argparse.ArgumentParser(
+        prog="python3 -m tskit",
         description="Command line interface for tskit.")
     top_parser.add_argument(
         "-V", "--version", action='version',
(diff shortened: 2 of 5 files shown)


hyanwong:
Annoyingly, since I can't simply mark the fasta cli output for skipping, I've also had to temporarily disable the cli fasta unit tests. Can't see any other way round this, I'm afraid.

python/tskit/cli.py
```diff
@@ -187,15 +188,16 @@ def get_tskit_parser():
         help="Remove any duplicated mutation positions in the source file. ")
     parser.set_defaults(runner=run_upgrade)
 
-    parser = subparsers.add_parser(
+    # suppress fasta visibility pending https://github.com/tskit-dev/tskit/issues/353
+    """parser = subparsers.add_parser(
```
jeromekelleher: Probably better to just comment this out rather than making it a string that does nothing.

docs/cli.rst
```diff
@@ -0,0 +1,15 @@
...
+======================
+Command line interface
+======================
+
+.. argparse::
+    :module: tskit.cli
+    :func: get_tskit_parser
+    :prog: python3 -m tskit
+
+    fasta : @skip
```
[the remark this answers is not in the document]
hyanwong: No - does nothing. I left it in there for my own reference, but will delete.

python/tskit/cli.py
hyanwong: OK - happy to comment rather than stringify

jeromekelleher:
Just remembered --- I guess we're missing the sphinx_argparse module from the various requirements.txt files? (dev and rtd)

hyanwong:
Good point. I'll add.

hyanwong:
Actually, it's already in [development.txt](https://github.com/tskit-dev/tskit/blob/a5dd0af8fc4d9791a6136cb8e12c81e68ad7c99b/python/requirements/development.txt#L12). Where else does it need to go?

jeromekelleher:
Good, thanks. Let's see how RTD likes it then.

Issue #353, opened 2019-09-01
Finalise semantics of fasta output


jeromekelleher:
Pending some issues being worked out about what do when we have floating point site positions and integer site positions (see #326, #350)

hyanwong:
To address this issue, I do think it would be helpful to know what we plan to do about integer site positions. Do we have a consensus here? E.g. we could make a derived class, e.g. `FiniteSitesTreeSequence` as [suggested](https://github.com/tskit-dev/tskit/issues/146#issuecomment-527195817) by @petrelharp. If so we could store the site positions as true integers, or simply as floating point (and report them as integers). Alternatively, we could keep the same class, and have a flag in the metadata that tells us to interpret the existing floating point positions as integers. Other suggestions welcome too.

I'm hoping to get @dianalir to have a go at thinking about this soon, so comments very welcome, @jeromekelleher

jeromekelleher:
Integer site positions aren't necessary, I think. 

I think the semantics are: if there is no reference sequence present (which there never is, right now; see #146 for discussion) we output the haplotypes as computed by the ``haplotypes()`` method. If there is a reference sequence, we use this to fill in the gaps between the variant sites.

But for now, I think it's clear: we just output the haplotypes.

hyanwong:
OK. For the record, I think that it is quite misleading to concatenate together non-adjacent sites as adjacent letters in a FASTA file. The FASTA convention is (AFAIK) for letters to represent adjacent bases, and for potential gaps (of any length) to be represented by a `-`. So my eventual preference would be to output a set of non-adjacent sites something like:

`A-T-T-C-G-`... etc. 

I do think that would be the least misleading and most standards-conformant output. Of course, once we have integer sites (and/or reference sequences: I don't think they are mutually dependent), then in some circumstances we could omit the `-` character between sites.

Perhaps I'm being too pernicketty, though?

hyanwong:
(I forgot that I wrote up my thoughts on this in https://github.com/tskit-dev/tskit/issues/326#issuecomment-526819752)

jeromekelleher:
With non-integer site positions then there's no real meaning to 'between' them. We can add the padding behaviour later as an option, but I think the only reasonable thing to do for now is just output the haplotypes. It's up to the user to put in intermediate sites if they want them.

jeromekelleher:
[...]

PR #426, opened 2019-11-26
Output haplotypes with missing data  (and discuss indels)

hyanwong:
Ported from #425. This needs some discussion, and more tests adding, but I think it's the right way to go.

I think we also want to be able to (optionally) output `.` characters between the allele strings (see https://github.com/tskit-dev/tskit/issues/353#issuecomment-555930529 and discussion below that). I haven't implemented this yet, but it should be trivial to add the option to the `haplotypes()` function, and to this code.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -6,6 +6,13 @@ In development
 
 **New features**
 
+- Allow sites with missing data to be output by the `haplotypes` method, by
+  default replacing with ``-``. Errors are no longer raised for missing data
+  with `impute_missing_data=False`; the error types returned for bad alleles
+  (e.g. multiletter or non-ascii) have also changed from `_tskit.LibraryError`
+  to TypeError, or ValueError if the missing data character clashes
+  (:user:`hyanwong`, :pr:`426`).
+
 **Bugfixes**
 
 --------------------
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -549,12 +549,26 @@ class TestHaplotypeGenerator(unittest.TestCase):
         H = [h.replace("0", "A").replace("1", "T") for h in ts.haplotypes()]
         self.assertEqual(H, list(tsp.haplotypes()))
 
-    def test_multiletter_mutations(self):
+    def test_fails_multiletter_mutations(self):
         ts = msprime.simulate(10, random_seed=2)
         tables = ts.tables
         tables.sites.add_row(0, "ACTG")
         tsp = tables.tree_sequence()
-        self.assertRaises(exceptions.LibraryError, list, tsp.haplotypes())
+        self.assertRaises(TypeError, list, tsp.haplotypes())
+
+    def test_fails_deletion_mutations(self):
+        ts = msprime.simulate(10, random_seed=2)
+        tables = ts.tables
+        tables.sites.add_row(0, "")
+        tsp = tables.tree_sequence()
+        self.assertRaises(TypeError, list, tsp.haplotypes())
+
+    def test_nonascii_mutations(self):
+        ts = msprime.simulate(10, random_seed=2)
+        tables = ts.tables
+        tables.sites.add_row(0, chr(169))  # Copyright symbol
+        tsp = tables.tree_sequence()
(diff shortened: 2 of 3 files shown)


jeromekelleher:
OK, before we go any further with this, what's the point in supporting multi-letter alleles here? I.e., does anyone actually need it? Unless the answer is "Yes, I/someone else I know cannot do their work without this feature" then I vote to drop it.

We probably want a more general "alignments" method, if we're going down this road. I'm perfectly happy with the ``alignments`` method throwing an error if there's non-integer site positions. The semantics of indels etc become clearer then.

Outputting missing data from ``haplotypes`` is a good feature, we want this.

hyanwong:
FWIW, perf tests show that for large numbers of samples (1M), it is essentially the same speed as both the C version and the non-missing-data version.

jeromekelleher:
> We probably want a more general "alignments" method, if we're going down this road. I'm perfectly happy with the alignments method throwing an error if there's non-integer site positions. The semantics of indels etc become clearer then.

ps. even with an ``alignments`` method I don't think we should get into dealing with multi-letter alleles. It's a massive can of worms and would take weeks to work through all the details and corner cases. It's not on the critical path.

hyanwong:
> OK, before we go any further with this, what's the point in supporting multi-letter alleles here? 

A decent proportion of sites in real sequences are small indels. The question is whether we want to force people with real data to use the (yet to be implemented) finite sites version of a tree sequence, or whether we want to enable infinite sites tree sequences to output realistic-looking data.

I think we probably need an input from a real biologist here, e.g. @petrelharp ?

To summarize - is it useful to be able to output small indels as aligned haplotypes, e.g.

```
Sample1: .A.T.---.G.T
Sample2: .A.C.---.G.A
Sample3: .A.T.ATT.G.A
```

where the `.` represents non-variable sequence between variants, and hasn't been implemented yet.

hyanwong:
I will work this PR into a missing-data-only version.

Nevertheless I suspect that the multi-letter version will be useful for many biologists. I asked the HIV group, and about 2-5% of the variable sites in an HIV genome are in short indels, FWIW (mostly in 3bp lengths). For the SNPs in the 1000G, about 5% of the "standard" variants are short indels, so probably about 10% of the letters in a haplotype will be in an indel. The semantics for the haplotype output for indels seem very clear to me, but you are right that this is somewhat of a mouthful to bite off, so spinning it off into a different PR seems sensible

hyanwong:
p.s. do we want to allow single letter deletions (i.e. one allele = `""`)? The semantics here are quite clear, I would say

jeromekelleher:
I don't think we should keep the semantics of ``haplotypes`` as it is: output just the sites that are in the ts. The semantics are clear then. Indels etc can be dealt with later, when someone with an application asks for it.

We can discuss the ``alignments`` operation (which would be defined only on integer a TS with integer sites) separately.

There's nothing stopping people using tree sequences which contain indels, they just can't write out alignments at the moment. This is pretty minor IMO since we support VCF output.

hyanwong:
> We can discuss the `alignments` operation (which would be defined only on integer a TS with integer sites) separately.

Shall I move this discussion to another issue? It would IMO be reasonable to output multi-letter alleles from a non-integer-site TS.

hyanwong:
> I don't think we should keep the semantics of `haplotypes` as it is: output just the sites that are in the ts.

You mean we *should* keep the semantics? I'm unclear if you think we should allow e.g. `genotypes = [0, 1, 2, -1], alleles=["", "A", "G", None]` to be output as

```
-
A
G
-
```

I've coded this up in the latest push on this PR, as it seems well-defined and reasonable behaviour, but it may be that you would prefer the presence of a `""` in the allele list to throw an error.

hyanwong:
OK @jeromekelleher - I think this allows missing data in haplotypes, but not single letter deletions, and is thus ready for review/merge when CI has completed

jeromekelleher:
Needs a rebase, but looks to merge after that. 

Can you update the CHANGELOG with this new feature please, and note any consequences for backward compatability?

hyanwong:
Rebased and added the changelog - hope it's not too verbose.

Issue #146, opened 2019-03-13
Support for reference sequences


jeromekelleher:
For some applications it would be useful to know the reference sequence that a tree sequence coordinate space refers to. For example, with real data, we should (at a minimum) record the reference build (e.g., GRCh38) and the contig ID (e.g., chr22) associated with a tree sequence. Ideally, we would also like to be able to do things like:
```python
for site in ts.sites():
     print(site.position, site.ref_allele)
```

As well as situations in which a well-known canonical reference is available, we may also have a one-off reference sequence that we wish to record, e.g., in simulations.

To support this, I suggest adding a ``reference`` section to the file store, with some fields. Roughly, these might looks like:

```
reference/build         -- e.g. GRCh38
reference/contig       -- e.g. chr22
reference/id              -- md5 hash of the sequence, c.f. refget
reference/sequence -- Actual sequence information
```

The references section would be itself optional (keeping backward compatability), and some fields would probably be optional within this section (for example, ``reference/sequence`` would definitely be optional --- no point in storing GRCh38 chr1 over and over again). We can imagine having some mechanisms for automatically retrieving sequences using [refget](http://samtools.github.io/hts-specs/refget.html), but this isn't at all necessary for a basic implementation.

At a high-level, we should try to follow any upstream standards as closely as possible, e.g. GA4GH [refget](http://samtools.github.io/hts-specs/refget.html) and any others that are relevant.

Any thoughts @tskit-dev/all, @bhaller?

hyanwong:
This seems very sensible, but I wonder if the word "reference" is a bit too context-dependent. Perhaps "genome_reference" or something similar would be clearer?

bhaller:
In SLiM I'm calling it the "ancestral sequence".  I'm fine with "reference" or "genome_reference" or any such variant.  The rest of the proposal seems good to me.  It should be clarified, when this is documented, exactly what format reference/sequence is expected to be in – FASTA?  Is a FASTA header allowed?  Are newline characters allowed (\r, \n)?  Nucleotides only, presumably, not FASTA amino acid sequences?  But FASTA codes for ambiguous nucleotides of various sorts?  But I think SLiM will just dump an unbroken stream of ACGT into it, so that is all that I need to be supported.

jeromekelleher:
Thanks @bhaller. To clarify, it'll be a flat array of chars, such that ``reference[pos]`` is the reference allele at position ``pos``.

petrelharp:
[...]

PR #427, opened 2019-11-27
Multi letter alleles (e.g. indels) in haplotypes()

hyanwong:
There was discussion of this in the following PR: https://github.com/tskit-dev/tskit/pull/426. @jeromekelleher is keen not to open the multi-letter-allele can of worms, unless there is demand from users. @hyanwong thinks that (in the case of haplotype output) the semantics are reasonably clear, and being able to output haplotype strings containing indels will be needed by users quite soon.

Some of this is probably only relevant once [finite site tree sequences](https://github.com/tskit-dev/tskit/issues/146#issuecomment-559154270) have been implemented. But (IMO) if we think it is reasonable to concatenate widely spaced SNPs into a string of letters, it is also reasonable to concatenate multi-letter haplotypes from an infinite-sites TS, as long as each allele takes the same number of characters. This is what this PR does.

Also from that PR, a quick summary - do we think it useful to be able to output small indels as aligned haplotypes, e.g.

```
Sample1: .A.T.---.G.T
Sample2: .A.C.---.G.A
Sample3: .A.T.ATT.G.A
```

Note that the current code does not output the `.` between sites (an optional convention to represents non-variable regions between variants) - I put them there for clarity, and also because I think it might be a useful addition to address https://github.com/tskit-dev/tskit/issues/353#issuecomment-555930529

### DIFF
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -554,7 +554,14 @@ class TestHaplotypeGenerator(unittest.TestCase):
         tables = ts.tables
         tables.sites.add_row(0, "ACTG")
         tsp = tables.tree_sequence()
-        self.assertRaises(exceptions.LibraryError, list, tsp.haplotypes())
+        self.assertRaises(TypeError, list, tsp.haplotypes())
+
+    def test_nonascii_mutations(self):
+        ts = msprime.simulate(10, random_seed=2)
+        tables = ts.tables
+        tables.sites.add_row(0, chr(169))  # Copyright symbol
+        tsp = tables.tree_sequence()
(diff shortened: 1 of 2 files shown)


hyanwong:
p.s. I'm also opening this as a PR because I don't want to lose the work I've done on this, especially the fast function coded herein (`sitewise_allele_max_bytes()`) which speedily calculates the maximum byte length of all the alleles at a site, for all sites. I think this might be generally useful however we chose to deal with multi-letter alleles.

hyanwong:
Temporarily closing due to lack of bandwidth for dealing with finite sites / multiletter alleles. May be worth revisiting later

PR #429, opened 2019-11-28
Don't assume all treeseqs from simulations

hyanwong:
Don't imply that tree sequences are exclusively from simulation output.


hyanwong:
Amended to a a section link for tree seqs

jeromekelleher:
LGTM, thanks.

PR #430, opened 2019-11-29
User specified alleles

jeromekelleher:
This adds the ability to specify a fixed mapping for allelic values to genotypes. The motivation for this is that it can be quite annoying to have a different allele mapping for every site (especially when doing downstream things like haplotype matching). Allowing the user to specify a fixed mapping for this doesn't cover every possible situation, but it does make a very common one (fixed alphabet of ACGT) much simpler.

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -268,14 +268,8 @@ tsk_strerror_internal(int err)
         case TSK_ERR_MUTATION_PARENT_AFTER_CHILD:
             ret = "Parent mutation ID must be < current ID";
             break;
-        case TSK_ERR_TOO_MANY_ALLELES:
-            ret = "Cannot have more than 127 alleles";
-            break;
         case TSK_ERR_INCONSISTENT_MUTATIONS:
             ret = "Inconsistent mutations: state already equal to derived state";
-            break;
-        case TSK_ERR_NON_SINGLE_CHAR_MUTATION:
-            ret = "Only single char mutations supported";
             break;
         case TSK_ERR_UNSORTED_MUTATIONS:
             ret = "Mutations must be provided in non-decreasing site order";
@@ -363,10 +357,19 @@ tsk_strerror_internal(int err)
             ret = "Bad genotype value provided";
             break;
 
-        /* Missing data errors */
+        /* Genotype decoding errors */
+        case TSK_ERR_TOO_MANY_ALLELES:
+            ret = "Cannot have more than 127 alleles";
+            break;
+        case TSK_ERR_ZERO_ALLELES:
+            ret = "Must have at least one allele when specifying an allele map";
+            break;
         case TSK_ERR_MUST_IMPUTE_NON_SAMPLES:
             ret = "Cannot generate genotypes for non-samples unless missing data "
                     "imputation is enabled";
+            break;
+        case TSK_ERR_ALLELE_NOT_FOUND:
+            ret = "An allele was not found in the user-specified allele map";
             break;
 
         /* Distance metric errors */
--- a/c/tskit/genotypes.c
+++ b/c/tskit/genotypes.c
@@ -39,8 +39,18 @@
 void
 tsk_vargen_print_state(tsk_vargen_t *self, FILE *out)
 {
+    tsk_size_t j;
+
     fprintf(out, "tsk_vargen state\n");
+    fprintf(out, "tree_index = %d\n", self->tree.index);
     fprintf(out, "tree_site_index = %d\n", (int) self->tree_site_index);
+    fprintf(out, "user_alleles = %d\n", self->user_alleles);
+    fprintf(out, "num_alleles = %d\n", self->variant.num_alleles);
+    for (j = 0; j < self->variant.num_alleles; j++) {
+        fprintf(out, "\tlen = %d, '%.*s'\n", self->variant.allele_lengths[j],
+                self->variant.allele_lengths[j], self->variant.alleles[j]);
+
+    }
 }
 
 static int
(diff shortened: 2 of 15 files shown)


c/tskit/core.c
```diff
@@ -363,11 +357,20 @@ tsk_strerror_internal(int err)
             ret = "Bad genotype value provided";
             break;
 
-        /* Missing data errors */
+        /* Genotype decoding errors */
+        case TSK_ERR_TOO_MANY_ALLELES:
+            ret = "Cannot have more than 127 alleles";
```
petrelharp: below, I think 127 is `INT16_MAX`; maybe this should be a sprintf with that value?

c/tskit/genotypes.c
```diff
@@ -427,11 +515,20 @@ tsk_vargen_update_site(tsk_vargen_t *self)
             update_genotypes = tsk_vargen_update_genotypes_i8_traversal;
         }
     }
-
-    /* Ancestral state is always allele 0 */
-    var->alleles[0] = site->ancestral_state;
-    var->allele_lengths[0] = site->ancestral_state_length;
-    var->num_alleles = 1;
+    if (self->fixed_alleles) {
```
petrelharp: "fixed" has an unfortunate double connotation here - at first I was thinking about "fixed" as in "everyone has it", which is the same thing as "ancestral". Other option: `has_allele_map`?

python/tskit/trees.py
```diff
@@ -333,6 +336,11 @@ class Variant(SimpleContainer):
     no particular order, and the ordering should not be relied upon
     (but see the notes on missing data below).
 
+    The alleles encoding may also be provided by the user via the
```
petrelharp: it should be more clear this is the "second" way refered to above

python/tskit/trees.py
```diff
@@ -2890,14 +2898,25 @@ def haplotypes(self, impute_missing_data=False):
...
+        site are generated automatically, such that the ancestral state
+        is at the zeroth index and subsequent alleles are listed in no
+        particular order. This means that the encoding of alleles in
+        terms of genotype values can vary from site-to-site, which is
+        sometimes inconvenient. It is possible to specify a fixed mapping
+        from allele strings to genotype values using the ``alleles``
+        parameter. For example, if we set ``alleles=("A", "C", "G", "T")``,
+        this will map allele "A" to 0, "C" to 1 and so on (the
+        :data:`ALLELES_ACGT` constant provides a shortcut for this
+        common mapping).
```
petrelharp: agree: this is very much more helpful!

python/tskit/trees.py
```diff
@@ -2910,9 +2929,11 @@ def variants(self, as_bytes=False, samples=None, impute_missing_data=False):
...
         If :ref:`missing data<sec_data_model_missing_data>` is present
         at a given site, the genotypes array will contain a special value
-        ``tskit.MISSING_DATA`` (-1) to identify these missing samples.
-        See the :class:`Variant` class for more details on how missing
-        data is reported.
+        :data:`MISSING_DATA` (-1) to identify these missing samples,
+        and the ``alleles`` tuple will end with the value ``None``
+        (note that this is true whether we specify a fixed mapping
+        using the ``alleles`` parameter or not). See the :class:`Variant`
+        class for more details on how missing data is reported.
```
petrelharp: what happens if someone puts `None` in their `alleles` list?

petrelharp:
Very handy - I agree. I've gone through and commented.

c/tskit/core.c
jeromekelleher: It's either INT8_MAX or INT16_MAX, depending on whether we're using 16 bit genotypes. In practise, we'd need two error codes to distinguish these. Do you mean use sprintf so we're not hard coding the INT8_MAX? Quite awkward if so: these are statically allocated strings, and can't write to them without getting into parallelism issues etc.

c/tskit/genotypes.c
jeromekelleher: Good point, we should change this. I'll have a think about it.

python/tskit/trees.py
jeromekelleher: They'll get a TypeError. Are you thinking of the case where someone passes in an alleles list that they obtained from another call to ``variants``?

petrelharp: I just noticed that missing alleles are None, so it raised the possibility that someone would be confused and think they needed to manually insert it into their alleles list. Raising an error is good.

c/tskit/genotypes.c
hyanwong: Also perhaps change the PR name, so it's clearer when looking back at the git history?

python/tskit/__init__.py
```diff
@@ -39,6 +39,14 @@
 #: decreasing genomic coordinate values).
 REVERSE = _tskit.REVERSE
 
+#: The allele mapping where the strings "0" and "1" map to genotype
+#: values 0 and 1.
+ALLELES_01 = ("0", "1")
+
+#: The allele mapping where the four nucleotides A, C, G and T map to
+#: the genotype integers 0, 1, 2, and 3, respectively.
+ALLELES_ACGT = ("A", "C", "G", "T")
```
hyanwong: Do we also want to support `ALLELES_acgt` (lowercase). Probably not, as it's trivial to do so yourself, and uppercase is the norm (lowercase tends to be used for soft masked regions, although I don't really know the conventions well).

python/tskit/trees.py
```diff
@@ -322,9 +322,12 @@ class Variant(SimpleContainer):
...
-    actual observed state for a given sample. The first element of this
-    tuple is guaranteed to be the same as the site's ``ancestral_state`` value.
-    The list of alleles is also guaranteed not to contain any duplicates.
+    actual observed state for a given sample. The ``alleles`` tuple is
+    generated in one of two ways. The first (and default) way is for
+    ``tskit`` to generate the encoding on the fly as alleles are encountered
+    while generating genotypes. In this case, the first element of this
+    tuple is guaranteed to be the same as the site's ``ancestral_state`` value
+    and the list of alleles is also guaranteed not to contain any duplicates.
     However, allelic values may be listed that are not referred to by any
```
hyanwong: I think this is now a "(note that ...)" rather than a "However"

python/tskit/__init__.py
jeromekelleher: We can add it easily if someone wants it.

jeromekelleher:
Thanks for the comments @petrelharp and @hyanwong. Good catch on the fixed_alleles thing, that was a terrible choice of name! I've changed to "user alleles" throughout: better?

hyanwong:
Wow, this is a lot of work, and a very useful addition. There's quite a lot in here that I don't quite understand, but I've done my best to make useful comments.

A few other points:

1. As I understand it, it would be possible to output a fixed array for *all* sites (e.g. SNPs, indels, etc), as long as we listed all the possible variants at every site in the `alleles` string, is that right? E.g. the alleles string could be `['A', 'C', 'G', 'T', '', 'AT', 'TA' ...]`. Obviously that would get tedious for a large number of different variants, but this is a possibility, right? So would it be useful to have a function that uniqu-ifies both the `ancestral_state`s stored in the sites table, and the `derived_state`s store in the mutations table, and returns a list or set? That could then, if small enough, be passed as a alleles list, so that even slightly more complex variant patterns could be represented as a fixed list?

2. Will we need a way to get the 1D numpy array of integer encodings of the ancestral states over all sites, now that we can't assume it's all zeros?

jeromekelleher:
> As I understand it, it would be possible to output a fixed array for all sites (e.g. SNPs, indels, etc), as long as we listed all the possible variants at every site in the alleles string, is that right? E.g. the alleles string could be ['A', 'C', 'G', 'T', '', 'AT', 'TA' ...]. Obviously that would get tedious for a large number of different variants, but this is a possibility, right? So would it be useful to have a function that uniqu-ifies both the ancestral_states stored in the sites table, and the derived_states store in the mutations table, and returns a list or set? That could then, if small enough, be passed as an alleles list, so that even slightly more complex variant patterns could be represented as a fixed list?

Sure, nothing stopping a user doing this (so long as there's less than 128 of them). I don't think we need functions for it right now though. 

> Will we need a way to get the 1D numpy array of integer encodings of the ancestral states over all sites, now that we can't assume it's all zeros?

You can still get the ancestral state from the site object from variants, so getting the ancestral state index is just ``alleles.index(site.ancestral_state)``. Easy to reproduce, so I don't think we need API additions for this unless someone asks for it.

python/tskit/trees.py
```diff
@@ -2970,27 +2992,29 @@ def variants(self, as_bytes=False, samples=None, impute_missing_data=False):
...
             genotypes, or None for all samples. Default: None.
         :param bool impute_missing_data: If True, the allele assigned to any
             isolated samples is the ancestral state; that is, we impute
             missing data as the ancestral state. Default: False.
+        :param tuple alleles: A tuple of strings describing the encoding of
+            alleles to genotype values. At least one allele must be provided.
+            If duplicate alleles are provided, output genotypes will always be
+            encoded as the first occurance of the allele. If None (the default),
+            the alleles are encoded as they are encountered during genotype
+            generation.
```
petrelharp: maybe say "describing how genotypes are encoded as integers" (since "genotype values" is ambiguous)

petrelharp: or maybe "describing how genotypes are encoded as integers, when integers are used (e.g., in :meth:.genotype_matrix())"

petrelharp:
Looks good - "user_alleles" is ok, but I like it how the python interface just uses "alleles".

jeromekelleher:
> Looks good - "user_alleles" is ok, but I like it how the python interface just uses "alleles".

OK, cool. Will think about it again in the new year.

jeromekelleher:
Merging this --- ``user_alleles`` is only used internally and isn't part of the external C API @petrelharp. This is necessary for memory management and only used to distinguish the case where  a user specifies alleles and does not.

Issue #405, opened 2019-11-19
Calculate r LD coefficient


standard-aaron:
This is just a suggestion that i would assume is easy to implement -- is it possible to include methods for rmatrix() in addition to r2matrix()? 

(Why do I care about this? r rather than r2 is the relevant statistic if you are trying to calculate marginal effect sizes for a linked locus with multiple causal effects.)

--Aaron

jeromekelleher:
Thanks for the suggestion @35ajstern. Presumably, this is ``np.sqrt(r2_matrix)``, so seems like it's hardly worth the API duplication? (Perhaps I'm missing something though?)

35ajstern:
What if r is negative?

35ajstern:
This is obviously easy enough to calculate directly from the genotype matrix, but i'm guessing you guys have some secret sauce for how to optimize this calculation using the treeSeq -- hence my suggestion

jeromekelleher:
Ah yes, of course.

Well, it's actually pretty involved coding the LD methods up at the moment, but we're hoping to get a more general framework in place soon (with @apragsdale), so we'll definitely include this when we do. Unfortunately, I don't think it'll happen before that unless you'd like to volunteer some time.

35ajstern:
Sounds good -- might try to help with this if I have time soon. Feel free to close the issue!

apragsdale:
Hi Aaron, Jerome - yes, I've been (slowly) plugging away locally on getting a more general framework for two locus statistics implemented, tested, and working, but it's not quite ready to send in a pull request I think (Jerome, unless you think it's helpful to have open pull request that are actively in development and a ways away from being integrated - not entirely sure what the preferred approach is for that).

@35ajstern if you have some time soon and do want to help out, let's chat! I've also been a bit slow getting around to this, so more people working on it would be motivating..

jeromekelleher:
> two locus statistics implemented, tested, and working, but it's not quite ready to send in a pull request I think (Jerome, unless you think it's helpful to have open pull request that are actively in development and a ways away from being integrated - not entirely sure what the preferred approach is for that).

It depends on when you'd like to get feedback basically @apragsdale --- feel free to open a PR whenever you like.

Let's keep this issue open anyway, as a reminder that we want to implement r as well as r2.

benjeffery:
Closing for inactivity and labelling "future", please re-open if you plan to work on this.

PR #432, opened 2019-12-02
Generalized LD matrix calculator

apragsdale:
This is of course nowhere near complete, but I wanted to get this up here for some feedback and thoughts on this extension of the LD calculator.

Essentially, we want to be able to compute any two-locus statistic over pairs of mutations (or branches, in the future). Here is some python code to get the LD matrix for any function that operates on two-locus haplotype counts (like r^2 or r, or multi-population stats like r_1 x r_2). Population ids can be specified to tell us which samples to compute over, and the windows can defined over the whole tree sequence (if `None`) a subinterval of the sequence length, or multiple windows.

Edit: example of using these function
```
import msprime
from test_ld_matrix import *

population_configurations = [
    msprime.PopulationConfiguration(initial_size=1000, sample_size=10),
    msprime.PopulationConfiguration(initial_size=2000, sample_size=10)
]

demographic_events = [
    msprime.MassMigration(source=1, destination=0, time=50, proportion=1)
]


ts = msprime.simulate(Ne=1000, length=10000,
                      recombination_rate=2e-6, mutation_rate=2e-6,
                      population_configurations=population_configurations,
                      demographic_events=demographic_events)

# compute r^2 for all samples (combined pop 0 and 1) over whole sequence
A = get_ld_matrix(ts, r2_function)

# compute r^2 for samples from population 0
A = get_ld_matrix(ts, r2_function, population_ids=[0])

# compute r^2 for samples from population 1 for subinterval
A = get_ld_matrix(ts, r2_function, population_ids=[1], windows=[1000,2000])

# compute r_1 x r_2
A = get_ld_matrix(ts, r1_r2_function, population_ids=[0,1],
                  windows=np.linspace(0,ts.sequence_length, 11))
```

### DIFF
--- a/python/tests/test_ld_stats.py
+++ b/python/tests/test_ld_stats.py
@@ -40,27 +40,25 @@ import _tskit
 ##############################
 
 
-## notes 6/11:
+## notes 6/12:
 ## Question: Do we want to return the LD matrix over variants or over sites?
 ##  I think sites, and that's what the naive_ld_matrix returns
-##
-## Other notes: the second ld_matrix function is outdated, and cannot handle polarized
-##  or multiple derived mutations.
 ##
 ## Because the returned LD matrix might not include every variant, and variants may be
 ##  distributed over multiple windows, we probably need to also return information of
 ##  which index in each LD matrix corresponds to which mutation in the tree sequence.
 ##  Thus, probably want to also return list of lists of site id indexes.
 ##
-## Should windows be a list of non-overlapping intervals? Currently, set up so that we
-##  just define window edges, so all windows necessarily have zero separation. We might
-##  instead want to be able to compute LD matrices or LD scores from windows that are
-##  distant. So could look like `windows = [(100, 200), (500, 1000), (1000, 2000)]`
+## Should windows be a list of non-overlapping intervals? Previously, I set up so that we
+##  just define window edges, so all windows necessarily have zero separation. Now we
+##  instead are able to compute LD matrices or LD scores from windows that are
+##  not adjacent. So could look like `windows = [(100, 200), (500, 1000), (1000, 2000)]`
 
 
 def naive_ld_matrix(ts, function, sample_sets=None, windows=None, polarized=True):
     """
-    This function returns the LD matrix for a given two-locus stats function. This is
+    This function returns the LD matrix for a given two-locus stats function and a list
+    of site indexes corresponding to the indexes in the LD matrices. This is
     a simple implementation that loops through sites in the tree sequence, collects
     genotypes for observed variants at those sites, tallies two-locus genotype counts,
     and passes those counts to the stats function.
@@ -71,11 +69,10 @@ def naive_ld_matrix(ts, function, sample_sets=None, windows=None, polarized=True
     sample_sets: a list of lists of samples. Typically, we'd think of each list of
     samples as the samples that belong to each population we compute the statistics for.
 
-    windows: a list of window break points, so that there are len(windows)-1 total LD
-    matrices, if windows is given. If windows is None, we compute one LD matrix that
-    spans the entire sequence length. Note that windows does not need to span the entire
-    sequence length, but there will not be gaps between consecutive windows if
-    len(windows) is greater than two.
+    windows: a list of window intervals. If windows is None, we compute one matrix that
+    spans the entire sequence length. Note that windows does not need to cover the entire
+    sequence length, and there can be gaps between the windows.
+    For example: windows = [(100, 200), (500, 1000), (1000, 2000)]
 
     polarized: if True, we compute statistics over derived variants at each site. If
     False, we sum statistics over all variants, derived and ancestral, at each site.
@@ -86,32 +83,43 @@ def naive_ld_matrix(ts, function, sample_sets=None, windows=None, polarized=True
     genotypes are derived.
     """
     if windows is None:
-        windows = [0, ts.sequence_length]
-    else:
-        assert len(windows) > 1, "need to define at least one window interval"
-    windows = np.array(windows)
+        windows = [(0, ts.sequence_length)]
+    window_edges = np.array(windows)
 
-    assert np.all(windows[1:] - windows[:-1] > 0), "windows list must be increasing"
-    assert windows[0] >= 0 and windows[-1] <= ts.sequence_length
+    assert np.all(window_edges[:,0] < window_edges[:,1]), "windows must specify left/right endpoints"
+    assert np.all(window_edges[1:,0] >= window_edges[:-1,1] ), "windows cannot overlap"
+    assert window_edges[0][0] >= 0 and window_edges[-1][1] <= ts.sequence_length, "windows must be within sequence length"
 
     if sample_sets is None:
         sample_sets = [ts.samples()]
     ns = [len(ss) for ss in sample_sets]
 
-    num_window_mutations = [0 for w in range(len(windows)-1)]
+    num_window_mutations = [0 for _ in windows]
     site_indexes = {}
     for s in ts.sites():
         # get window index and snp index within that window
-        window_index = min(np.argwhere(s.position < windows)[0]) - 1
-        site_indexes[s.id] = (window_index, num_window_mutations[window_index])
-        num_window_mutations[window_index] += 1
+        window_index = np.where(
+            np.logical_and(
+                s.position > window_edges[:,0],
+                s.position < window_edges[:,1]
+            )
+        )[0]
+        if len(window_index) == 0:
+            continue
+        assert len(window_index) == 1
+        site_indexes[s.id] = (window_index[0], num_window_mutations[window_index[0]])
+        num_window_mutations[window_index[0]] += 1
 
     A = [np.zeros((num_mutations, num_mutations))
          for num_mutations in num_window_mutations]
 
     for v1 in ts.variants():
+        if v1.site.id not in site_indexes:
+            continue
         w1, m1 = site_indexes[v1.site.id]
         for v2 in ts.variants():
+            if v2.site.id not in site_indexes:
+                continue
             if v1.position >= v2.position:
                 w2, m2 = site_indexes[v2.site.id]
                 if w1 == w2: # sites fall within the same window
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Nice, I like it @apragsdale! Looks like it's not that different to the approach currently used in the LdCalculator, so it wouldn't be too hard to implement.

One thing I'd like to understand is what happens in the more general cases when we have many alleles. Is this something you've thought through? I

petrelharp:
Nice! Need to think more about generalization, but a minor suggestion: shouldn't we pass in `sample_sets` rather than population ids, like in the other stats? (This is probably on your list already, ok...)

apragsdale:
Thanks @jeromekelleher and @petrelharp! I mimicked the existing LdCalculator since computing a generalized LD matrix is just an extension of what already exists (and doing something simple may be a quick way to get something working for #405). It's also not a big leap to use this approach to compute common LD scores given a window around the focal SNP. @petrelharp I agree it makes sense to pass `sample_sets` rather than population IDs. That is an easy fix, and I'll work on both those things and push an update.

The more general LD calculation based on branches will require a different approach, similar to the branch weighting and updating in the single-site branch stat calculator. I thought it would make sense to modularize these efforts a bit, for a few reasons: 1. the algorithm for computing the LD matrix will be a bit different than for computing average or expected LD within distance bins (e.g. computing the LD decay curve), 2. it would keep the pull requests smaller and less sprawling, so the PR doesn't just sit here for a year and some of the simpler features are included sooner, and 3. I'm pretty new to contributing code like this, so smaller bites will be helpful for me :) Does this all sound reasonable?

apragsdale:
>One thing I'd like to understand is what happens in the more general cases when we have many alleles. Is this something you've thought through?

For the case with more than two alleles, I haven't thought hard about that. But it's worth thinking through.. I'll get back about that.

>Nice! Need to think more about generalization

The approach in this commit works for any function that takes allele and haplotype frequency counts, which are computed using tracked samples here. If we want to assign any arbitrary weight to the samples, I have something written up that can do that, but my Python implementation is painfully slower than using `tracked_samples`. But it does work! Do you have a specific generalization in mind for a function that takes something other than two-locus haplotype counts? I'll think about that a bit as well.

jeromekelleher:
SGTM @apragsdale, small chunks is a good way to go. We can get some initial code in the tests and can build things up from there. I'd like to see it generalised to arbitrary alleles all right though, and to harmonise with the existing general stats APIs as much as possible. We'll be deprecating the existing LdCalculator APIs, so basically ignore that in terms of APIs.

petrelharp:
Yes, sounds like a good plan. Here's one suggestion for making it apply to multiallelic sites: write it as a sum over alleles, where for each allele we compute something with that allele coded as `1` and not-that-allele coded as `0`. This is what we've done for the general stats.  Alternatively, there's some multiallelic generalizations of LD that involve distances between probability distributions (IIRC).

apragsdale:
Looking at the python implementations in `test_tree_stats.py`: If I understand right, `sample_sets` aren't passed, but instead you specify which samples to use and which population they belong to based on the `sample_weights`. Is that right? For example, we could specify if a sample belongs in group 1 or group 2 based on weights that are arrays of length 2, so that ```sample_weights = [[1, 0], [1, 0], [0, 1], [0, 0]]``` would mean the first two samples are in group 1, the third in group 2, and the fourth in neither. Or more simply, a list with 0s and 1s would specify which samples to include in the calculation of a statistic.

As for multi-allelic sites, I think it makes sense to write statistics as a sum over alleles. A lot of LD statistics require assigning labels to the alleles (for example, the sign of D and r depend on the labeling). The multi-allelic extension is straightforward, I think, if we label mutations as ancestral or derived (with possibly many derived, but a single ancestral), and summing over possible pairs of derived alleles at the two loci.

jeromekelleher:
> Looking at the python implementations in test_tree_stats.py: If I understand right, sample_sets aren't passed, but instead you specify which samples to use and which population they belong to based on the sample_weights. Is that right? For example, we could specify if a sample belongs in group 1 or group 2 based on weights that are arrays of length 2, so that sample_weights = [[1, 0], [1, 0], [0, 1], [0, 0]] would mean the first two samples are in group 1, the third in group 2, and the fourth in neither. Or more simply, a list with 0s and 1s would specify which samples to include in the calculation of a statistic.

That's the most general form, yeah. See [general_stat](https://github.com/tskit-dev/tskit/blob/master/python/tskit/trees.py#L3552) for the top-level interface. See also [sample_count_stat](https://github.com/tskit-dev/tskit/blob/master/python/tskit/trees.py#L3624) for a simple layer on top of this.

petrelharp:
Notes from 1/10/2020:

Proposal is to:

1. Have a weight matrix where each of `k` columns is a set of weights; what is kept track of is,
    for each pair of nodes across two trees, (or, pair of alleles at two loci,) the total weight 
    (a) of samples that are below both nodes; (b) of samples below the first node; (c) of samples below the second node.
     But, to make statistic definitions easier, we'll want to translate these to total weights in each of the four combinations of (below, not below) the two nodes.  Let `w[AB]` denote the total weight of samples having both alleles `A` and `B`; and similarly, `w[A], w[B]` are the total weight of samples having the two alleles. Here e.g., `w[AB]` will be a vector of length `k` (the number of weights).

2. Given these, we'll write e.g. `w[Ab] = w[B] - w[AB]`, i.e., the total weight of everyone with allele `A` and *not* having alelle `B`. This is simpler for multiallelic sites since we won't have to sum over all pairs of alleles *at each site*.

3. Summary functions will be of the form
    ` f(w[AB], w[Ab], w[aB])`
     where each argument is a `k`-vector. (also maybe given the total weight? or, four columns? same as we did before.)

4. Statistics are defined to be 
    `sum_{A} sum_{B} f(w[AB], w[Ab], w[aB])`,
    where the sums are over either (if unpolarized) all alleles at the two sites, or (if polarized) all derived alleles at the two sites.

For instance, "`D^2 = (pAB pab - pAb paB)^2 = AB,AB,ab,ab + Ab,Ab,aB,aB - 2 AB Ab aB ab`",
and so we can compute this by defining (writing `n[AB] = w[AB]` for the case of 1/0 weights and `n` the total weight):
```
  f () = ( n[AB] (n[AB] - 1) n[ab] (n[ab] - 1) - n[AB] n[Ab] n[aB] n[ab] ) / 2 n (n-1) (n-2) (n-3)
```

petrelharp:
Proposal: next we should work through a few more examples (list these?) and check that this definition seems good.

petrelharp:
As far as *output* (brainstorming):

- entire LD matrix for everything within some distance
- LD matrix for everything within some window (?)
- LD scores by SNP
- average LD by distance bin, in windows

jeromekelleher:
> Proposal: next we should work through a few more examples (list these?) and check that this definition seems good.

This sounds great @apragsdale and @petrelharp!

andrewkern:
gosh I had forgotten how cool this is @apragsdale!

apragsdale:
Some thoughts about the design of LD calculator functions and arguments. Right now I'm just focusing on computing the LD matrix and LD scores, and leaving the LD decay calculations for later.

Computing the LD matrix is necessarily a site statistic, while LD scores could be computed from sites or from branches of linked trees (in which case, it's a mixed site/branch statistic). At the moment, I've just tried to write some simple functions to focus on what options we should allow. Here's what I've got at the moment in the "naive" functions that have been updated with the most recent commit:

- `sample_sets`, which could/should be later generalized to sample weights
- `windows`, which currently is a list of window break points, and can be a subset of the full tree sequence length. I think I want to change this to let you pass a list of intervals, so that there can be gaps between consecutive windows.
- `polarized`, if True, sum over pairs of derived variants at the two sites. If False, sum over all variants.
- for LD scores, `focal_window_size`, which specifies the window around a focal that linked variants/branches contribute to the focal site's LD score.

I'd like to get your thoughts on how to handle windows. This would be a bit different from the windows option in the single site statistics functions, I think, since here I'm proposing to allow for windows to cover a subset of the full sequence length.

There is also the choice of how to handle windows with computing LD scores. For focal variants within some window, should the linked variants only also be taken from that window, or any variant within the focal_window_size, whether or not it falls within the focal site's window? I could imagine cases where one or the other would be wanted.

python/tests/test_ld_stats.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+##############################
+# LD matrix calculator
+##############################
+
+
+## notes 6/12:
+## Question: Do we want to return the LD matrix over variants or over sites?
+##  I think sites, and that's what the naive_ld_matrix returns
```
petrelharp, line 45: They're the same thing (well, in 1-to-1 correspondence, anyhow).

python/tests/test_ld_stats.py
```diff
@@ -0,0 +1,475 @@
...
+##
+## Because the returned LD matrix might not include every variant, and variants may be
+##  distributed over multiple windows, we probably need to also return information of
+##  which index in each LD matrix corresponds to which mutation in the tree sequence.
+##  Thus, probably want to also return list of lists of site id indexes.
+##
+## Should windows be a list of non-overlapping intervals? Previously, I set up so that we
+##  just define window edges, so all windows necessarily have zero separation. Now we
+##  instead are able to compute LD matrices or LD scores from windows that are
+##  not adjacent. So could look like `windows = [(100, 200), (500, 1000), (1000, 2000)]`
```
petrelharp, line 55: Oh, this is nice way of allowing either - how about if `windows` is just a single flat list, it is then interpreted as being the breakpoints between contiguous windows (that must span the whole genome) and converted to this form?

Should `windows` refer to both the rows *and* columns of the LD matrix? Or just one?

petrelharp: Oh, I see you're only doing the diagonal blocks as specified by the windows. Makes sense. But, should offdiagonals be allows? Would people want to compute, say, LD between a given SNP and all others in the genome?

Thinking a bit more about this - I think the LD matrix shold have `windows="sites"`, for consistency? And then be given another argument to specify *which* pairs of sites to output, if not all of them. I would have thought that we'd want something like `max_distance`, to get the diagonal band of the matrix, rather than filling out only blocks along the diagonal as this does (IIUC).  If you pass in different windows (ie, numbers, not "sites"), you'd get an nwindows x nwindows matrix whose (i,j)th entry is the average LD between SNPs in windows i and j (or, the corresponding thing integrated over the trees).

python/tests/test_ld_stats.py
```diff
@@ -0,0 +1,475 @@
...
+                                                 gts.count((1, 0)),
+                                                 gts.count((0, 1)))
+                                                for gts in sample_genotypes]
+                            A[w1][m1, m2] += function(two_locus_counts, ns, polarized)
+
+    for i, ld_mat in enumerate(A):
+        idx_upper = np.triu_indices(len(ld_mat), 1)
+        A[i][idx_upper] += A[i].T[idx_upper]
+
+    return A
```
petrelharp, line 151: nice and elegant! very nice!

python/tests/test_ld_stats.py
apragsdale: I was thinking of the case of multiple variants at a single site. But I do think we want to return LD over all variants at a site

petrelharp: Ah - more than two alleles? Sorry - thought you meant the objects returned by `ts.variants()`.  But yes - I agree.

python/tests/test_ld_stats.py
apragsdale: There are a few ways to set it up - and I agree that it would be nice to be more flexible (say, compute LD between this set and that set of sites or windows, like you say). I don't know if we want to require that windows cover the entire sequence length. Unlike the single site stats, two-locus stats are more expensive to compute, and will get more expensive to store a big N sites by M sites matrix.

I do like the idea of setting some focal window length to only get LD between pairs of variants within some distance, giving you a diagonal band.

There are definitely some more choices to be made about generalizing and argument options. These points you bring up are important to think about..

python/tests/test_ld_stats.py
apragsdale: Stack overflow :)

python/tests/test_ld_stats.py
apragsdale: Right, more than two alleles.. I was using "allele" interchangeably with "variant", but I see now that those are different in terms of what `ts.variants()` returns. Sorry for the confusion

benjeffery:
I believe this is superseded by recent work - reopen if needed.

PR #435, opened 2019-12-03
Clarify return type of tree.nodes()

hyanwong:


python/tskit/trees.py
```diff
@@ -1476,17 +1476,21 @@ def _timedesc_traversal(self, u):
...
-        Returns an iterator over the nodes in this tree. If the root parameter
-        is provided, iterate over the nodes in the subtree rooted at this
-        node. If this is None, iterate over all nodes. If the order parameter
+        Returns an iterator over the node IDs in this tree. If the root parameter
+        is provided, iterate over the node IDs in the subtree rooted at this
+        node. If this is None, iterate over all node IDs. If the order parameter
         is provided, iterate over the nodes in required tree traversal order.
 
+        .. note::
+            Unlike the identially named :meth:`TreeSequence.nodes` method, this
```
jeromekelleher: I don't think we need the "identically named" bit here (typo anyway). We could link to the bit in the tutorial where this is explained too: tskit.readthedocs.io/en/latest/tutorial.html#trees

hyanwong: OK, will remove

PR #436, opened 2019-12-04
Create tree.num_children() method

### LINKED ISSUE #431: Tree.num_children(node), or Tree.is_unary(node) function?
With tree sequences now allowing unary nodes, I find that I'm relatively regularly checking if a node is unary. I might be wrong, but it seems to me that it might be a little inefficient to keep doing `len(tree.children(u))` for this. If so, would it be sensible to simply report the number of children somehow, e.g. via a new method `Tree.num_children(u)`? Or (presumably even more efficiently, but much less generally), `Tree.is_unary(u)`?

hyanwong:
Fixes #431

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -13,6 +13,9 @@ In development
   to TypeError, or ValueError if the missing data character clashes
   (:user:`hyanwong`, :pr:`426`).
 
+- Access the number of children of a node in a tree directly using
+  ``tree.num_children(u)`` (:user:`hyanwong`, :pr:`436`).
+
 **Bugfixes**
 
 --------------------
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -8137,6 +8137,26 @@ out:
 }
 
 static PyObject *
+Tree_get_num_children(Tree *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    unsigned int num_children;
+    int node;
+    tsk_id_t u;
+
+    if (Tree_get_node_argument(self, args, &node) != 0) {
+        goto out;
+    }
+    num_children = 0;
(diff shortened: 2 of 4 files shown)


hyanwong:
Sorry - forgot to test. Will correct now.

python/CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ In development
   to TypeError, or ValueError if the missing data character clashes
   (:user:`hyanwong`, :pr:`426`).
 
+- Convenient fast access to the number of children of a node in a tree with a
```
jeromekelleher: Too much detail here: user's don't care how it's implemented.

python/_tskitmodule.c
```diff
@@ -8136,6 +8136,26 @@ Tree_get_mrca(Tree *self, PyObject *args)
     return ret;
 }
 
+static PyObject *
+Tree_get_num_children(Tree *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    size_t num_children;
```
jeromekelleher: could make this an unsigned int here and avoid the cast below.

python/tests/test_highlevel.py
```diff
@@ -1579,6 +1579,11 @@ def test_func(t, u):
                 self.assertEqual(l1, l2)
                 self.assertEqual(t.get_num_samples(u), len(l1))
 
+    def test_num_children(self):
+        for tree in self.get_tree():
+            for u in tree.nodes():
+                assert tree.num_children(u) == len(tree.children(u))
```
jeromekelleher: self.assertEqual

python/_tskitmodule.c
hyanwong: Good catch - copy-paste snafu. Done

python/CHANGELOG.rst
hyanwong: Shortened

PR #443, opened 2019-12-05
Add is_isolated and document use

### LINKED ISSUE #442: Add omit_missing to Tree.nodes()?
At the moment, the `Tree.nodes()` method lists all the nodes in a tree, even if the node is isolated (i.e. has missing data). It would be useful for me to somehow iterate through the nodes without counting the missing ones. A parameter to the `Tree.nodes()` iterator would work, or perhaps an easy way to check if a node is isolated? The only way I can think to do this now is to check `Tree.num_children(node) == 0 and Tree.parent(node) == tskit.NULL` - is that right?

hyanwong:
Fixes #442

### DIFF
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -1016,9 +1016,19 @@ Consider the following example:
 In this tree, node 4 is isolated, and therefore for any sites that are
 on this tree, the state that it is assigned is a special value
 ``tskit.MISSING_DATA``, or ``-1``, as long as there are no mutations above
-that node at that site. See the :meth:`TreeSequence.variants`
-method and :class:`Variant` class for more information on how missing
-data is represented.
+the node at that site. Note that, although isolated, because node 4
+is a sample node it is still considered as being present in the
+tree, meaning it will still returned by the :meth:`Tree.nodes` and
+:meth:`Tree.samples` methods. The :meth:`Tree.is_isolated` method can be used to
+identify nodes which are isolated samples:
+
+    >>> [u for u in tree.samples() if tree.is_isolated(u)]  # isolated samples in this tree
+    [4]
+    >>> [u for u in tree.nodes() if not tree.is_isolated(u)]  # topologically connected nodes
+    [0, 1, 2, 3, 5, 6, 7]
+
+See the :meth:`TreeSequence.variants` method and :class:`Variant` class for
+more information on how missing data is represented in variant data.
 
 
 .. _sec_text_file_format:
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -7857,3 +7857,58 @@ class TestTrim(unittest.TestCase):
         self.assertRaises(ValueError, ts.ltrim)
         self.assertRaises(ValueError, ts.rtrim)
         self.assertRaises(ValueError, ts.trim)
+
+
+class TestMissingData(unittest.TestCase):
+    """
+    Test various aspects of missing data functionality
+    """
+
+    # TODO tests for missing data currently sparse: more tests should go here
+
+    def ts_missing_middle(self):
+        # Simple ts with sample 0 missing a middle section
+        ts = msprime.simulate(4, mutation_rate=1, recombination_rate=4, random_seed=2)
+        tables = ts.dump_tables()
+        tables.edges.clear()
+        # mark the middle as missing
+        for e in ts.tables.edges:
+            if e.child == 0:
+                if e.left == 0.0:
+                    missing_from = e.right
+                elif e.right == 1.0:
+                    missing_to = e.left
+                else:
+                    continue  # omit this edge => node is isolated
+            tables.edges.add_row(e.left, e.right, e.parent, e.child)
+        # Check we have non-missing to L & R
+        self.assertTrue(0.0 < missing_from < 1.0)
+        self.assertTrue(0.0 < missing_to < 1.0)
+        return tables.tree_sequence(), missing_from, missing_to
+
+    def test_is_isolated(self):
+        ts, missing_from, missing_to = self.ts_missing_middle()
+        for tree in ts.trees():
+            if tree.interval[1] > missing_from and tree.interval[0] < missing_to:
+                self.assertTrue(tree.is_isolated(0))
+                self.assertFalse(tree.is_isolated(1))
+            else:
+                self.assertFalse(tree.is_isolated(0))
+                self.assertFalse(tree.is_isolated(1))
+            # A non-sample node is isolated if not in the tree
+            tree_nodes = set(tree.nodes())
+            for nonsample_node in np.setdiff1d(np.arange(ts.num_nodes), ts.samples()):
+                if nonsample_node in tree_nodes:
+                    self.assertFalse(tree.is_isolated(nonsample_node))
+                else:
+                    self.assertTrue(tree.is_isolated(nonsample_node))
+
+    def test_is_isolated_bad(self):
+        ts, missing_from, missing_to = self.ts_missing_middle()
+        for tree in ts.trees():
+            self.assertRaises(ValueError, tree.is_isolated, tskit.NULL)
+            self.assertRaises(ValueError, tree.is_isolated, ts.num_nodes)
+            self.assertRaises(ValueError, tree.is_isolated, -2)
+            self.assertRaises(TypeError, tree.is_isolated, None)
+            self.assertRaises(TypeError, tree.is_isolated, "abc")
+            self.assertRaises(TypeError, tree.is_isolated, 1.1)
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1216,6 +1216,19 @@ class Tree:
         """
         return len(self.children(u)) == 0
 
+    def is_isolated(self, u):
+        """
+        Returns True if the specified node is isolated in this tree: that is
+        it has no parents and no children. Sample nodes that are isolated
+        and have no mutations above them are used to represent
+        :ref:`missing data<sec_data_model_missing_data>`.
+
+        :param int u: The node of interest.
+        :return: True if u is an isolated node.
+        :rtype: bool
+        """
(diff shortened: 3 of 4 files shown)


docs/data-model.rst
```diff
@@ -997,9 +997,22 @@ Consider the following example:
 
 In this tree, node 4 is isolated, and therefore for any sites that are
 on this tree, the state that it is assigned is a special value
-``tskit.MISSING_DATA``, or ``-1``. See the :meth:`TreeSequence.variants`
-method and :class:`Variant` class for more information on how missing
-data is represented.
+``tskit.MISSING_DATA``, or ``-1``. Isolated nodes in a tree can be identified
+using the :meth:`Tree.is_isolated` method. Note that, although isolated, node
+4 above is still considered as being present in the tree. This means, for example,
```
jeromekelleher: need to insert "because it is a sample" here somewhere, to clarify.

docs/data-model.rst
```diff
@@ -997,9 +997,22 @@ Consider the following example:
 
 In this tree, node 4 is isolated, and therefore for any sites that are
 on this tree, the state that it is assigned is a special value
-``tskit.MISSING_DATA``, or ``-1``. See the :meth:`TreeSequence.variants`
-method and :class:`Variant` class for more information on how missing
-data is represented.
+``tskit.MISSING_DATA``, or ``-1``. Isolated nodes in a tree can be identified
```
jeromekelleher: might as well use the :data: reference here

docs/data-model.rst
```diff
@@ -997,9 +997,22 @@ Consider the following example:
...
+using the :meth:`Tree.is_isolated` method. Note that, although isolated, node
+4 above is still considered as being present in the tree. This means, for example,
+that the node is still returned by the :meth:`Tree.nodes` and :meth:`Tree.samples`
+methods. We can therefore identify isolated nodes, as well as nodes linked to the
+main tree topology, when iterating using these methods:
```
jeromekelleher: This is implying that we have one root: we often have many roots in forward simulations. Is there some way to rephrase that's not talking about "the tree topology"?

docs/data-model.rst
```diff
@@ -997,9 +997,22 @@ Consider the following example:
...
+
+    >>> [u for u in tree.nodes() if tree.is_isolated(u)]       # list isolated_nodes
+    [4]
+    >>> [u for u in tree.nodes() if not tree.is_isolated(u)]   # list nodes in topology
```
jeromekelleher: This is probably redundant, I think? These are non-isolated nodes, so pretty obvious, and avoids semantic issues with the description in your comment (see previous comment).

python/tskit/trees.py
```diff
@@ -912,6 +912,19 @@ def is_leaf(self, u):
         """
         return len(self.children(u)) == 0
 
+    def is_isolated(self, u):
+        """
+        Returns True if the specified node is isolated in this tree: that is
+        it has no parents and no children. Isolated nodes are a way of
```
jeromekelleher: Isolated **samples**

hyanwong: Good catch. Isolated internal nodes are another matter.

docs/data-model.rst
hyanwong: Yep, I was trying to find the right wording for "nodes in one of the conventional trees". Do you think it's reasonable to call these "nodes in a topology" (without implying that there is only one topology per tree), so e.g. 

> We can therefore identify isolated nodes, as well as nodes in a topology,

jeromekelleher: I don't think there's much point in talking about the negative --- isn't it enough to say that we're identifying the isolated nodes? The reader can reasonably infer that they also identify non-isolated nodes, which saves us getting into a confusing technical discussion of what non-isolated nodes are.

docs/data-model.rst
hyanwong: As you mentioned in https://github.com/tskit-dev/tskit/issues/442#issuecomment-562221369, it's possibly more useful to list the non-isolated nodes than the isolated ones. That's why I left this example in, really, even though it's a bit redundant. I could change he comment to simply "list non-isolated nodes" (although that makes it sound *really* redundant!)

jeromekelleher:
Still outstanding comments here @hyanwong, can you update please? I think we should keep the discussion to a minimum and not dwell too much on the relationship with missing data.

docs/data-model.rst
hyanwong: I still think it is useful to point out that if someone wants to print out the topologically connected nodes, they can use the `is_isolated` method. It's not going to be obvious to most people who deal with trees that, when looking at trees in a tree sequence, there will be nodes  which are not connected to the topology, but instead separate (isolated). Unless you know already about how missing data is encoded, this must come as a bit of a surprise. It certainly fooled me: I expected to be able to iterate over the nodes in a tree topology by doing `tree.nodes()`, and had to dig a bit to realise I had to actively exclude the "missing" ones.

Perhaps I'm confused by the work "missing". Nodes with "missing data" should not be thought of as nodes that are "missing" - they are still present in the output trees. It's simply that the STATE is missing (unknown).

All of which is to say that I find it useful to show an example of how to leave out "missing data" samples, so I've left the second example in, but commented that this gives "topologically connected nodes". Feel free to rephrase or delete as you see fit.

hyanwong:
I think this can be merged @jeromekelleher - I wondered whether https://github.com/tskit-dev/tskit/pull/462 made any difference to my additions to the "missing data" section, but I think not.

However, we might want to add something in the missing data section about `root_threshold` anyway.

jeromekelleher:
I think we probably want this function now that we've cleared up any ambiguity with the data model - would you mind bringing this up to date please @hyanwong?

hyanwong:
How is this?

python/tests/test_topology.py
```diff
@@ -7843,3 +7843,41 @@ def test_failure_with_migrations(self):
...
+                    missing_to = e.left
+                else:
+                    continue  # omit this edge => node is isolated
+            tables.edges.add_row(e.left, e.right, e.parent, e.child)
+        # Check we have non-missing to L & R
+        self.assertTrue(0.0 < missing_from < 1.0)
+        self.assertTrue(0.0 < missing_to < 1.0)
+        return tables.tree_sequence(), missing_from, missing_to
+
+    def test_is_isolated(self):
```
benjeffery, line 7875: I know this is a very simple method - but it's always good to test a couple of error cases, e.g. a negative index, a index bigger than the number of samples, an index of the wrong type just to check we get the right error messages.

hyanwong:
@benjeffery 's point about testing made me realise there is an ambiguity here. What about non-sample nodes? They are often missing from a particular tree. I had been assuming that the adjective "isolated" only ever applied to sample nodes, hence "isolated_as_missing", but I see that the test we use returns "is_isolated(n) = True" where n is an internal (non-sample) node which is missing from this tree, but present in others.

I'm conflicted here. It seems useful to be able to identify non-sample nodes that are missing from a tree, and to have a way to describe them. But using the term "isolated" to apply *only* to sample nodes makes it clearer what you are talking about, is less verbose, and also aligned with our much-debated `ISOLATED_AS_MISSING` notation.

python/tskit/trees.py
```diff
@@ -1210,6 +1210,19 @@ def is_leaf(self, u):
         """
         return len(self.children(u)) == 0
 
+    def is_isolated(self, u):
+        """
+        Returns True if the specified node is isolated in this tree: that is
+        it has no parents and no children. Sample nodes that are isolated
+        and have no mutations above them are commonly used to represent
```
jeromekelleher, line 1217: I would delete the "commonly" here .

jeromekelleher (commented):
One minor tweak and we're good to go.

benjeffery:
> @benjeffery 's point about testing made me realise there is an ambiguity here. What about non-sample nodes? They are often missing from a particular tree. I had been assuming that the adjective "isolated" only ever applied to sample nodes, hence "isolated_as_missing", but I see that the test we use returns `is_isolated(n) = True` where n is an internal (non-sample) node which is missing from this tree, but present in others. I guess if we only want to use "isolated" to refer to a sample node, we could check if the node is in the tree, and `is_isolated()` could error out if not.
> 
> I'm conflicted here. It seems useful to be able to identify non-sample nodes that are missing from a tree, and to have a way to describe them. But using the term "isolated" to apply _only_ to sample nodes makes it clearer what you are talking about, is less verbose, and also aligned with our much-debated `ISOLATED_AS_MISSING` notation.

I think retaining "isolated" to be a purely topological definition is correct, both samples and non-samples can be isolated. We want to retain a clear distinction with "missing" and saying only samples can be isolated muddies the waters for me.

jeromekelleher:
> We want to retain a clear distinction with "missing" and saying only samples can be isolated muddies the waters for me.

Agreed

python/tskit/trees.py
hyanwong: Sure - I just wanted to avoid implying that they are *always* used to represent missing data, given the recent discussion about `ISOLATED_AS_MISSING`. Have you got an idea for another adjective, or do you think simply "are used to represent missing data" is OK?

jeromekelleher: Just delete it I'd say - people can read the linked documentation for details.

hyanwong: OK. adjective deleted.

benjeffery:
@hyanwong Just one failing test from when non-samples couldn't be isolated, then I think this is good to go.

hyanwong:
> @hyanwong Just one failing test from when non-samples couldn't be isolated, then I think this is good to go.

Oh yes, sorry! Forgot that I put that sentinel in to remind me that a decision needed to be made. Now fixed.

hyanwong:
I think this is ready to merge @benjeffery ?

benjeffery (approved):
LGTM

benjeffery:
@hyanwong Force pushed to resolve conflict

jeromekelleher (approved):

PR #444, opened 2019-12-10
Add a 'span' property to the edge

hyanwong:
This seems like a nice convenience to me, but happy to money patch it into my own code instead if it's deemed too obvious.


hyanwong:
By the way, I see a comment in the SimpleContainers classes about using __slots__. I was wondering about this. Do you want me to add those in another PR?

jeromekelleher:
Sure, this is a handy addition. Let's not worry about slots for now, we'd need to do some perf analyses to see what the benefits are.

PR #445, opened 2019-12-12
Add __len__ to the tree iterator

hyanwong:
Trivial, but this allows me to use tqdm when iterating over trees without explictly having to specify the total, so probably worth it, unless there's good reason not to.


jeromekelleher:
Nice, clever idea. I don't think there's any side effects of adding len, so happy to merge when CI passes.

hyanwong:
Seems like a circle/ci internal issue. I'll re-try with a force push or something

PR #446, opened 2019-12-12
Simple test additions

hyanwong:
A few things I noticed when looking through some tests. In particular, it seems neat to use tuple unpacking when messing with table rows, so that you can do e.g.

```
for edge_row in ts.tables.edges:
    new_tables.edges.add_row(*edge_row)
```

without having to remember all the add_row parameters or their order, which I always found tedious. I didn't realise you could unpack namedtuples like this. So I added it to one of the tests, as a demo to remind myself, and to keep the tests checking that this method will keep on working (it also saves a line).

I've kept the three commits separate here, as they are all different sorts of thing, but I can squash them if needed.

### DIFF
--- a/python/tests/test_stats.py
+++ b/python/tests/test_stats.py
@@ -217,14 +217,13 @@ def naive_mean_descendants(ts, reference_sets):
     tree_iters = [ts.trees(tracked_samples=sample_set) for sample_set in reference_sets]
     for _ in range(ts.num_trees):
         trees = [next(tree_iter) for tree_iter in tree_iters]
-        left, right = trees[0].interval
-        length = right - left
+        span = trees[0].span
         for node in trees[0].nodes():
             num_samples = trees[0].num_samples(node)
             if num_samples > 0:
                 for j, tree in enumerate(trees):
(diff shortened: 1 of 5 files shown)


hyanwong:
NB - note that this also picked up the inconsistency in the order of definition of `ProvenanceTableRow`, so it's been a useful thing to look at.

hyanwong:
Oh, damn, this doesn't work, because the order in the ProvenanceTable is (deliberately, I think, so that `timestamp` can be left out) different in `ProvenanceTable.add_row()` to how it is listed in all other provenance table functions. So the "proper" way to set `add_row()` from a TableRow object is  the rather ugly `table.add_row(**row._asdict())`.

Hence I'm closing this PR for the moment and reassessing. I'll put the span changes in somewhere else. Sorry for the noise.

petrelharp:
Darn, that would be nice. At least `table.add_row(**row._asdict())` is better than remembering the order.

hyanwong:
Thinking this over, it should be easy to add a ‘copy_row’ generic function to the base table class which does basically this. Only another line or two of code. I’ll code it up if you like the idea @petrelharp.

PR #449, opened 2019-12-13
Add table.copy_row() function

hyanwong:
Perhaps slightly more convoluted than it needs to be, but the advantage is that now we check that the order of columns in any table is the same as in its `add_row` function, and have to override explicitly (and hopefully comment) if not - i.e. for the ProvenanceTable.

See e.g. comments around https://github.com/tskit-dev/tskit/pull/446#issuecomment-565308833

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -510,6 +510,22 @@ class CommonTestsMixin(object):
                 self.assertEqual(copy, table)
                 table = copy
 
+    def test_copy_row(self):
+        for num_rows in [0, 10]:
+            input_data = {col.name: col.get_input(num_rows) for col in self.columns}
+            for list_col, offset_col in self.ragged_list_columns:
+                value = list_col.get_input(num_rows)
+                input_data[list_col.name] = value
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This seems like a confusing addition to the API to me. Semantically, this is equivalent to ``add_row``, except it's got a different parameter type.  The problem we're solving doesn't seem like it's worth muddling the API for.

jeromekelleher:
We'd be better off thinking about a way to combine tables/slices of tables directly than to spend time on this. But it's a low priority right now, it's not stopping us from doing anything.

hyanwong:
Sure. I can just use the ‘**X._asdict()’ notation instead. Nothing urgent or necessary here. The main useful thing was checking that parameter order was consistent, and documenting why not in the ProvenanceTable, which seemed a possible future source of confusion. 

Shall I close this then?

jeromekelleher:
Please, yes. The ProvenanceTable thing is unfortunate, but unavoidable (you're right about the reason for it). As I say, the right way to do this is to join tables together and avoid the add_row stuff entirely.

petrelharp:
Thanks though - good thought.

PR #452, opened 2020-01-10
WIP: haplotype matching using Li and Stephens

jeromekelleher:
This PR is a WIP towards getting an initial version of a general purpose haplotype matching toolkit based on the Li and Stephens process into tskit. The initial commit would not be final and the API not documented, so we don't need to worry too much about designing long-terms stable APIs.

The initial goal is to get this into tskit now so I don't need to maintain this branch any more, and so we can apply the same ideas in tsinfer (if not actually use the same code, yet).


jeromekelleher:
OK, this is looking good enough to merge now. The code is roughly in the right shape I think, but the details and APIs will be refined gradually over the coming months as I test things out and do some performance tuning. Once it's ready for use, I'll make a high-level API suitable for general use and document it.

Issue #456, opened 2020-01-30
[feature proposal] Adaptor class for a networkx graph


winni2k:
It strikes me that an adaptor class for a tskit Tree to make it look like a networkx DiGraph would give easy access to networkx's substantial graph algorithm library (https://networkx.github.io/documentation/stable/reference/algorithms/index.html) to [perform](https://networkx.github.io/documentation/stable/reference/algorithms/traversal.html) the traversals described in the [tutorial](https://tskit.readthedocs.io/en/latest/tutorial.html#traversals) and much more. 

I think I could write something like that. Would there be interest in including such code in tskit?

jeromekelleher:
Very interesting idea @winni2k! I think we would definitely be interested in that.

Is there an interchange format that networkx will understand, or do we need to import networkx? I don't want to make networkx a hard requirement (but could be an ``extras_require``).

winni2k:
There are several interchange formats that networkx understands including python data structures such as dicts of lists, dicts of dicts, and numpy/scipy matrices (see [here](https://networkx.github.io/documentation/stable/reference/convert.html)). Networkx also understands several graph storage formats (see [here](https://networkx.github.io/documentation/stable/reference/readwrite/index.html)). This approach would require making a copy of a tree to then interrogate it. We would write a conversion function that converts a Tree object to a numpy/scipy adjacency matrix that can then be fed to [`networkx.from_numpy_matrix`](https://networkx.github.io/documentation/stable/reference/generated/networkx.convert_matrix.from_numpy_matrix.html). No imports of networkx necessary. Minimal code. But we would have to instruct users on how to call `networkx.from_numpy_matrix`. It looks like we could even export branch lengths as branch weights.

There's an alternative approach where we write a wrapper class for a Tree class that has the duck type of a `networkx.DiGraph` class. That would avoid copying the tree data, but would be harder to implement and keep in sync with the networkx library (more coupling). No import of networkx necessary either.

I think I'd prefer the first option because it is simpler.

molpopgen:
Great idea.

jeromekelleher:
The ``from_numpy_matrix`` option sounds perfect to me. Let's not worry the overhead about making a copy, it's unlikely we're going to be doing this for ginormous trees. So, how about we add a method to the tree class ``as_matrix()`` which will return a numpy adjacency matrix in a form that networkx understands, and give a short recipe in the docstring for how to import this into networkx?

[...]

PR #457, opened 2020-01-31
Create Tree.as_dict_of_dicts() method for conversion to networkx.Graph/DiGraph

winni2k:
This is a PR stub for the feature described in #456.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -111,6 +111,136 @@ Running this on the example above gives us::
 
    [(7, 0), (6, 1), (4, 2), (3, 2), (5, 1), (2, 2), (1, 2), (0, 2)]
 
+
+++++++++++++++++++++++++
+Traversals with networkx
+++++++++++++++++++++++++
+
+Traversals and other network analysis can also be performed using the sizeable
+`networkx <https://networkx.github.io/documentation/stable/index.html>`_
+library. This can be achieved by calling :meth:`Tree.as_dict_of_dicts` to
+convert a :class:`Tree` instance to a format that can be imported by networkx to
+create a graph::
+
+    import networkx as nx
+
+    g = nx.DiGraph(tree.as_dict_of_dicts())
+    print(sorted(g.edges))
+
+::
+
+    [(5, 0), (5, 1), (5, 2), (6, 3), (6, 4), (7, 5), (7, 6)]
+
+++++++++++++++++++
+Traversing upwards
+++++++++++++++++++
+
+We can revisit the above examples and traverse upwards with
+networkx using a depth-first search algorithm::
+
+    import networkx as nx
+
+    g = nx.DiGraph(tree.as_dict_of_dicts())
+    for u in tree.samples():
+        path = [u] + [parent for parent, child, _ in
+                      nx.edge_dfs(g, source=u, orientation="reverse")]
+        print(u, "->", path)
+
+giving::
+
+   0 -> [0, 5, 7]
+   1 -> [1, 5, 7]
+   2 -> [2, 5, 7]
+   3 -> [3, 6, 7]
+   4 -> [4, 6, 7]
+
++++++++++++++++++++++++++++++++++
+Calculating distances to the root
++++++++++++++++++++++++++++++++++
+
+Similarly, we can yield the nodes of a tree along with their distance to the
+root in pre-order in networkx as well::
+
+    import networkx as nx
+
+    g = nx.DiGraph(tree.as_dict_of_dicts())
+    for root in tree.roots:
+        print(nx.shortest_path_length(g, source=root).items())
+
+Running this on the example above gives us the same result as before::
+
+   [(7, 0), (6, 1), (4, 2), (3, 2), (5, 1), (2, 2), (1, 2), (0, 2)]
+
++++++++++++++++++++++++++++++++++++++++
+Finding nearest neighbors
++++++++++++++++++++++++++++++++++++++++
+
+If some samples in a tree are not at time 0, then finding the nearest neighbor
+of a sample is a bit more involved. Instead of writing our own traversal code
+we can again draw on a networkx algorithm.
+Let us start with an example tree with three samples that were sampled at
+different time points:
+
+.. image:: _static/different_time_samples.svg
+   :width: 200px
+   :alt: An example tree with samples taken at different times
+
+The generation times for these nodes are:
+
+.. code-block:: python
+
+    for u in tree.nodes():
+        print(u, tree.time(u))
+
+giving::
+
+    4 20.005398778263334
+    2 20.0
+    3 17.833492457579652
+    0 0.0
+    1 1.0
+
+Note that samples 0 and 1 are about 35 generations apart from each other even though
+they were sampled at almost the same time. This is why samples 0 and 1 are
+closer to sample 2 than to each other.
+
+For this nearest neighbor search we will be traversing up and down the tree,
+so it is easier to treat the tree as an undirected graph::
+
+    g = nx.Graph(tree.as_dict_of_dicts())
+
+When converting the tree to a networkx graph the edges are annotated with their
+branch length::
+
+    print(g.edges(data=True))
+
+giving::
+
+    [(4, 2, {'branch_length': 0.005398778263334236}),
+     (4, 3, {'branch_length': 2.171906320683682}),
+     (3, 0, {'branch_length': 17.833492457579652}),
+     (3, 1, {'branch_length': 16.833492457579652})]
+
+We can now use the "branch_length" field as a weight for a weighted shortest path
+search::
+
+    # a dictionary of dictionaries to represent our distance matrix
+    dist_dod = collections.defaultdict(dict)
+    for source, target in itertools.combinations(tree.samples(), 2):
+        dist_dod[source][target] = nx.shortest_path_length(
+            g, source=source, target=target, weight="branch_length"
+        )
+        dist_dod[target][source] = dist_dod[source][target]
+
+    # extract the nearest neighbor of nodes 0, 1, and 2
+    nearest_neighbor_of = [min(dist_dod[u], key=dist_dod[u].get) for u in range(3)]
+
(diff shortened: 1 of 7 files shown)


winni2k:
@jeromekelleher I think I have the basic functionality down now. Are there any tests you'd like to see? 

If not, then the next step is to remove all of those print statements, add some documentation and make an entry in the tutorial.

python/tests/test_highlevel.py
```diff
@@ -1660,6 +1661,36 @@ def test_newick(self):
...
+        self.assertSetEqual(set(tree.roots), roots)
+
+        leaves = {n for n in g.nodes if g.out_degree(n) == 0}
+        self.assertSetEqual(set(tree.leaves()), leaves)
+
+        for root in tree.roots:
+            print(root)
+            self.assertSetEqual(
+                set(u for u in tree.nodes() if tree.is_descendant(u, root) and
+                    u != root),
```
winni2k: Is it a bug that a sole root node is also considered a descendant by `Tree.is_descendant`?

python/tests/test_highlevel.py
```diff
@@ -1660,6 +1661,97 @@ def test_newick(self):
...
+    def verify_nx_algorithm_equivalence(self, tree, g):
+        for root in tree.roots:
+            print(root)
+            self.assertTrue(nx.is_directed_acyclic_graph(g))
+
+            # test descendants
+            self.assertSetEqual(
+                set(u for u in tree.nodes() if tree.is_descendant(u, root)),
+                set(nx.descendants(g, root)) | {root}
+            )
```
winni2k: Is it a bug that root nodes are considered descendants of themselves by `Tree.is_descendant`?

docs/tutorial.rst
```diff
@@ -111,6 +111,138 @@ Running this on the example above gives us::
...
++++++++++++++++++++++++++++++++++++++++
+
+If some samples in a tree are not at time 0, then finding the nearest neighbor
+of a sample is a bit more involved. Instead of writing our own traversal code
+we can again draw on a networkx algorithm.
+Let us start with an example tree with three samples that were sampled at
+different time points::
+
+.. image:: _static/different_time_samples.svg
+   :width: 200px
```
winni2k: There seems to be an error with the indentation here, but I can't figure it out. Any ideas? See https://circleci.com/gh/tskit-dev/tskit/1209?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

docs/tutorial.rst
```diff
@@ -111,6 +111,138 @@ Running this on the example above gives us::
...
+
++++++++++++++++++++++++++++++++++++++++
+Finding nearest neighbors
++++++++++++++++++++++++++++++++++++++++
+
+If some samples in a tree are not at time 0, then finding the nearest neighbor
+of a sample is a bit more involved. Instead of writing our own traversal code
+we can again draw on a networkx algorithm.
+Let us start with an example tree with three samples that were sampled at
+different time points::
```
jeromekelleher: Need to delete one of these ':' at the end, as this is causing the next block to be interpreted as a literal.

python/tskit/trees.py
```diff
@@ -1600,6 +1600,36 @@ def newick(self, precision=14, root=None, node_labels=None):
...
+            >>> nx.DiGraph(tree.as_dict_of_dicts())
+            >>> # undirected graphs work as well
+            >>> nx.Graph(tree.as_dict_of_dicts())
+
+        :return: Dictionary of dictionaries of dictionaries where the first key
+            is the source, the second key is the target of an edge, and the
+            third key is an edge annotation. At this point the only annotation
+            is the difference in "time" of source and target (in generations).
+        """
+        tables = self.tree_sequence.dump_tables()
```
jeromekelleher: I don't think we need to dump the tables here. How about
```python
dod = {}
for u in tree.nodes():
    dod[u] = something(tree.children(u))
    # Store the node, including the branch length?
```

python/tskit/trees.py
```diff
@@ -1600,6 +1600,36 @@ def newick(self, precision=14, root=None, node_labels=None):
...
+        dod = collections.defaultdict(dict)
+
+        for idx, data in enumerate(tables.nodes):
+            # Nodes at time 0 are always included in the tree
+            if data.time == 0:
+                dod[idx] = {}
+        for left, right, parent, children in tables.edges:
+            if left <= self.interval[0] and self.interval[1] <= right:
+                dod[parent][children] = {'time': self.time(parent) - self.time(children)}
```
jeromekelleher: We should probably call this ``branch_length`` rather than time I think.

python/tskit/trees.py
```diff
@@ -1600,6 +1600,36 @@ def newick(self, precision=14, root=None, node_labels=None):
             return self.__build_newick(root, precision, node_labels) + ";"
         return s
 
+    def as_dict_of_dicts(self):
```
jeromekelleher: I wonder would ``as_edgelist`` be simpler? This would just be 
```python
def as_edgelist(self):
    edgelist = []
    for u in tree.nodes():
          for v in tree.children(u):
              edgelist.append((u, v)) 
              # or (v, u), depending on which way we think edges go. I tend to think of them
              # as going from child to parent in a tree, but this isn't very common, I think.
```

Not sure whether this representation would allow us to include branch lengths though.

docs/tutorial.rst
```diff
@@ -111,6 +111,138 @@ Running this on the example above gives us::
...
+
+.. image:: _static/different_time_samples.svg
+   :width: 200px
+   :alt: An example tree with samples taken at different times
+
+.. code-block:: python
+
+    tables = tskit.TableCollection(sequence_length=1.0)
```
jeromekelleher: Would it be easier to do a quick msprime simulation rather than building a tree up using the tables API? Can always print out the tables via ``print(ts.tables)``

python/tskit/trees.py
winni2k: Sure!  That sounds a lot better.

python/tskit/trees.py
winni2k: I know a dict-of-dicts-of-dicts is a bit of a mouthfull, but including branch lengths (and other annotations) will allow weighted graph traversals, which I think could be quite useful?

docs/tutorial.rst
winni2k: Yup. I think standard pop size, followed by an exponentially expanding population with two samples in the standard part and one historical sample in the exponential expansion should probably do? Need to check.

python/tskit/trees.py
winni2k: Does an edge list allow orphan nodes (roots without children)? That seems to be a pretty common occurrence in the test data.

winni2k:
I've implemented all suggestions except for the `Tree.as_edgelist()` method. I think the dict_of_dicts approach is more flexible as it allows for branch length annotation, and I'm not sure the edge list would work with childless root nodes. See my comments above.

python/tests/test_highlevel.py
```diff
@@ -1660,6 +1660,139 @@ def test_newick(self):
...
+
+            # test MRCA
+            if tree.num_nodes < 20:
+                for u, v in itertools.combinations(tree.nodes(), 2):
+                    self.assertEqual(
+                        tree.mrca(u, v),
+                        nx.lowest_common_ancestor(g, u, v) or -1
```
jeromekelleher: I haven't seen the ``or -1`` before. What's this doing?

python/tskit/trees.py
jeromekelleher: Good points, let's stick with dict-of-dicts.

docs/tutorial.rst
```diff
@@ -111,6 +111,136 @@ Running this on the example above gives us::
...
+Finding nearest neighbors
++++++++++++++++++++++++++++++++++++++++
+
+If some samples in a tree are not at time 0, then finding the nearest neighbor
+of a sample is a bit more involved. Instead of writing our own traversal code
+we can again draw on a networkx algorithm.
+Let us start with an example tree with three samples that were sampled at
+different time points:
+
+.. image:: _static/different_time_samples.svg
```
jeromekelleher: Can we add the code for generating this tree to ``docs/examples.py`` please?

python/tskit/trees.py
```diff
@@ -1600,6 +1600,33 @@ def newick(self, precision=14, root=None, node_labels=None):
...
+            is the source, the second key is the target of an edge, and the
+            third key is an edge annotation. At this point the only annotation
+            is "branch_length", the length of the branch (in generations).
+        """
+        dod = {}
+        for parent in self.nodes():
+            dod[parent] = {}
+            for child in self.children(parent):
+                dod[parent][child] = {
+                    'branch_length': self.time(parent) - self.time(child)
```
jeromekelleher: Might as well make it ``dod[parent][child] = {'branch_length': self.branch_length(child)}`` just to make it totally explicit for readers of the code.

python/tests/test_highlevel.py
winni2k: The `or` operator returns the first argument if that argument is truthy. Otherwise it returns the second argument. It appears to be [standard python](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not).

This is a perl idiom that I must confess I am trying for the first time in python here. I just tried it on a whim to see if it works, and apparently it does. I agree that this usage is uncommon and am happy to replace it with something else if you prefer.

winni2k: Oh, and this works because an empty container is falsy. So I am asserting that the MRCA is `-1` if there is no lowest common ancestor.

jeromekelleher: Can you change this to something more explicit please? IMO, code is like jokes: if you have to explain it...

winni2k:
All done

jeromekelleher:
Great, thanks @winni2k. Can you rebase off upstream/master please? The PR is a little [behind](github.com/tskit-dev/tskit/network), and life is easier if the git history is linear.

winni2k:
🥁

PR #460, opened 2020-02-07
Slight clarification for missing data in map_mutations

hyanwong:
The use of missing data in `map_mutations` confused me when reading the docs. In particular, setting a sample to a state of `tskit.MISSING_DATA` does not result in a series of mutations that create a state of "missing data" in that sample (e.g. there is no "mutation to a missing data state" created). It simply results in that tip being ignored in the parsimony algorithm. So if the returned values for mutations and ancestral state are stuck on the tree, the state of that sample will instead be *imputed*. I think this should be mentioned somehow, as I would say that it's not that obvious to the end user.

I've made a small change to the docs that might help a bit, but it doesn't mention this imputation property, and I think it should. Suggestions for wording are welcome, and I'll add them to the PR.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1648,9 +1648,11 @@ class Tree(object):
         into the alleles list in the same manner as described in the
         :meth:`TreeSequence.variants` method. Thus, if sample ``j`` carries the
         allele at index ``k``, then we have ``genotypes[j] = k``.
-        Missing data can be specified for a sample using the value
-        ``tskit.MISSING_DATA`` (-1). At least one non-missing observation must be
-        provided. A maximum of 63 alleles are supported.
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Yes, good point. How about "Samples that have missing data will be taken into account by the parsimony algorithm, and these missing states  imputed in the returned mutation mapping. "

hyanwong:
I've thought about the wording. How about a separate more explicit paragraph like 

"Samples marked as missing an observation in the input ``genotypes`` need not be marked as "missing" (i.e. `:ref:'isolated <sec_missing_data>'`) in the tree. In this case, mapping the mutations returned by this method onto the tree will result in the missing observations being imputed to the most parsimonious state."

jeromekelleher:
SGTM

PR #466, opened 2020-03-06
Format Python code with black.

jeromekelleher:
This is an experiment to see what the tskit code would look like if we started using the [black](https://github.com/psf/black) python formatter. The idea is that black formats the code so that we don't have to. If we like this, then we would set up a GitHub hook to (e.g.) autoformat contributed code using black when a PR is made. This would make it easier for contributors, who wouldn't have to worry about living by pep8 rules any more.

Any thoughts? cc. @benjeffery, @petrelharp

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,13 +12,13 @@ jobs:
       - run:
           name: Checkout submodules
           command: |
-            git submodule update --init --recursive 
+            git submodule update --init --recursive
       - run:
           name: Install dependencies and set path
           command: |
             sudo apt-get install ninja-build libcunit1-dev valgrind clang doxygen libgsl-dev
             # Install meson to the system packages so we can run it as root
-            sudo pip install meson 
+            sudo pip install meson
             pip install --user -r python/requirements/development.txt
             pip install twine --user
             # Remove tskit installed by msprime
@@ -38,10 +38,11 @@ jobs:
           name: Lint Python
           command: |
             cd python
-            flake8 --max-line-length 89 tskit setup.py tests
+            flake8 *.py tskit/ tests/
+            black --check *.py tskit/ tests/
 
       - run:
-          name: Run Python tests 
+          name: Run Python tests
           command: |
             PYTHONPATH=python nosetests -v --with-coverage --cover-package tskit \
               --cover-branches --cover-erase --cover-xml \
@@ -50,7 +51,7 @@ jobs:
       - run:
           name: Upload Python coverage
           command: |
-            codecov -X gcov -F python_tests 
+            codecov -X gcov -F python_tests
             # Clean up reports so we don't upload them twice.
             rm -f coverage.xml
 
@@ -58,7 +59,7 @@ jobs:
             cd python
             gcov -pb -o ./build/temp.linux*/*.gcno _tskitmodule.c
             cd ..
-            codecov -X gcov -F python_c_tests 
+            codecov -X gcov -F python_c_tests
             # Clean up reports so we don't upload them twice.
             rm -f python/*.gcov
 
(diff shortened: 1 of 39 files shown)


benjeffery:
In general I really like the idea of a tool taking care of this for us. I don't like the double quotes or the hanging brackets, but I can get over that.
I'm unsure how a github hook would work - adding commits just for formatting doesn't seem quite right. A hook could reformat the commits with a rebase, contributors would then need to force push to update their PR (annoyingly the git error message they'd get on a normal pull would advise them to do pull and merge).

jeromekelleher:
I've had a look through the changes, and they're mostly fine. Some things look a little weird to me, but I'd happy live with it if it meant no more code formatting discussions!

I would want to add a few `#fmt off` and `fmt on` blocks in places where existing layout helps us to see the structure of test data, but that's about it.

Let's see what others think before I go to the trouble of doing this though.

molpopgen:
I like this.  I'm probably going to do this for my projects.  Right now, I rely on autopep8 run inside of vim, but it is a bit unreliable.

petrelharp:
I like the formatting, and I like reducing the barrier to new folks of getting their code formatted right. I was imagining a pre-commit hook, like git runs black before doing the commit? Is that what you were thinking, Jerome?  My only slight worry is that then maybe people don't learn to format code nicely themselves, but reducing the number of frustrating things for new folks is more than worth it, I think.

jeromekelleher:
> I like the formatting, and I like reducing the barrier to new folks of getting their code formatted right. 

Hooray! 

> I was imagining a pre-commit hook, like git runs black before doing the commit? Is that what you were thinking, Jerome? 

Yes, that's the best thing to do for regular tskit developers all right, and is one of the ways the black developers [recommend](https://black.readthedocs.io/en/stable/version_control_integration.html). A slightly fancier option would be to run black as a github action, and then tack on an extra commit to any PRs which fix any non-black formatting. Something like this is talked about [here](https://github.com/lgeiger/black-action/issues/1). This would really lower the barrier for contributors, as they wouldn't even need to install or run black (but at the expense of some extra commits in the history fixing their formatting). I'm not sure how hard this would be to set up, but it's a good example of the sorts of things we could look at doing with GitHub actions.

molpopgen:
We can also update the dev docs w/links to the various editor integrations.  I never format my own code in any language.  I've relied on `:!autopep8 -i %` for Python for ages, but `:Black` works a ton better.  (For the C/C++, `:ClangFormat`.)

benjeffery:
> A slightly fancier option would be to run black as a github action, and then tack on an extra commit to any PRs which fix any non-black formatting. 

This [limitation](https://github.community/t5/GitHub-Actions/Can-t-push-to-forked-repository-on-the-original-repository-s/m-p/35916/highlight/true#M2372) would mean a github action can't add a fixing commit to a PR from a contributors fork. :disappointed: 

[This approach](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/) looks very promising. It uses the `pre-commit` package to manage installing the hooks. It still means that contributors would need to run one command on install (`pre-commit install`) which we would put in the dev docs. It might even be an idea to put it in the `Makefile`.... :thinking:

petrelharp:
I think that doing `pip install black` is a pretty low barrier. We could request that black get run on PRs before merging, so they wouldn't have to do it right away.

jeromekelleher:
> I think that doing pip install black is a pretty low barrier. We could request that black get run on PRs before merging, so they wouldn't have to do it right away.

OK, let's take the low-tech solution for now. I'll update the PR and CI checks.

jeromekelleher:
OK, merging this PR as it's a beast. Have made issue #470 to follow up on docs and workflows.

Issue #470, opened 2020-03-09
Document Black formatting for Python code


jeromekelleher:
#466 changed the codebase to use Black formatting. We need to document this for developers.

As @benjeffery suggested [here](https://github.com/tskit-dev/tskit/pull/466#issuecomment-595997239) we should use the pre-commit package to automate this as much as possible.

Also nice to link to the [editor integration](https://black.readthedocs.io/en/stable/editor_integration.html#editor-integration) section to help people (I think the vim plugin is very nice).

molpopgen:
The vim plugin currently requires installing from git, which I did via pip3.  You get a "can't format" error otherwise, and there's already an issue ticket for this at black's GitHub.

jeromekelleher:
> The vim plugin currently requires installing from git, which I did via pip3. You get a "can't format" error otherwise, and there's already an issue ticket for this at black's GitHub.

I hit that one too. Hopefully it'll be cleared up soon.

benjeffery:
Adding to the list here, a "can docs build ok" pre-commit would have just saved me a CI pass.

PR #471, opened 2020-03-11
Pre commit checks (WIP)

benjeffery:
#470 
This work-in-progress PR adds pre-commit checks for code quality and style. If anyone has any suggestions for extra checks please comment. Current planned list:
- [x] black
- [x] flake8
- [x] common pre-commit
- [x] isort
- [x] bugbear
- [x] pyupgrade
- [ ] doc checks (both code, text and docstrings)
- [ ] C style
- [ ] integrate with CI


grahamgower:
Hey @benjeffery. Any plans to start adding type annotations? If so, static type checking could go here too.

jeromekelleher:
> Hey @benjeffery. Any plans to start adding type annotations? If so, static type checking could go here too.

No, we hadn't considered this. I've opened #472 to discuss.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+    -   id: debug-statements
+    -   id: mixed-line-ending
+-   repo: https://github.com/timothycrosley/isort.git
+    rev: 4.3.21-2
+    hooks:
+    -   id: isort
+        files: \.py$
+        # Following https://github.com/timothycrosley/isort/issues/694
+        args: [--multi-line=3, --trailing-comma, --force-grid-wrap=0, --use-parentheses,
+               --line-width=88, --project=tskit,_tskit,msprime]
```
jeromekelleher: msprime is third-party here I think, strictly speaking.

python/tests/test_dict_encoding.py
```diff
@@ -25,10 +25,9 @@
 """
 import unittest
 
+import _tskit as c_module
```
jeromekelleher: This isn't quite right - _tskit is being classified as third party?

python/tests/test_genotypes.py
```diff
@@ -22,18 +22,16 @@
...
-import unittest
 import itertools
 import random
+import unittest
 
-import numpy as np
 import msprime
-
+import numpy as np
+import tests.test_wright_fisher as wf
```
jeromekelleher: The ``tests`` module probably needs to be registered as local.

benjeffery:
> isort seems to be doing some weird stuff. How's your experience been with it so far?

Not that great - I'm still getting cases where it is disagreeing with `black`. I'm trying out https://github.com/asottile/reorder_python_imports instead.

.pre-commit-config.yaml
benjeffery: Yep, I thought so - however I'd seen it grouped with tskit in at one place so was seeing if that looked good throughout.

benjeffery:
`reorder_python_imports` doesn't let you specify `_tskit` as local. Seeing if a quick patch to add the option will do the trick.

benjeffery:
I've filed https://github.com/asottile/reorder_python_imports/pull/123 and will point the pre-commit config at my fork for now.

python/tests/test_genotypes.py
benjeffery: Looks fixed with `reorder_python_imports`

jeromekelleher:
Just wondering if the import sorting/categorising is worth the hassle @benjeffery? Could be a rabbit hole...

benjeffery:
> Just wondering if the import sorting/categorising is worth the hassle @benjeffery? Could be a rabbit hole...

Yeah, I was thinking that too. If the PR I have in now doesn't do the job I'll write it off as it is only a nice to have.

benjeffery:
`bugbear` threw up over a hundred issues which I have fixed - some of these are worth checking as they could be deeper things to fix. For example we had lots of "loop variable not used". For a lot of these it is correct to do so, but worth double checking we're not looping unnecessarily.

.flake8
```diff
@@ -1,6 +1,7 @@
 [flake8]
 # Based directly on Black's recommendations:
 # https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length
-max-line-length = 80
+max-line-length = 800
```
jeromekelleher: Seems like a long line! What's this based on?

python/tests/test_lowlevel.py
```diff
@@ -1832,7 +1837,7 @@ def f(mutations):
         u = ts.get_num_nodes()
         for bad_node in [u, u + 1, 2 * u]:
             self.assertRaises(_tskit.LibraryError, f, [(0.1, bad_node)])
-        for bad_pos in [-1, length, length + 1]:
+        for _bad_pos in [-1, length, length + 1]:
```
jeromekelleher: This looks like an error, we're not actually testing the bad_pos here .

python/tests/test_topology.py
```diff
@@ -4165,7 +4165,7 @@ def test_overlapping_unary_edges_internal_samples(self):
         trees = [{0: 2}, {0: 2, 1: 2}, {1: 2}]
         for t in ts.trees():
             self.assertEqual(t.parent_dict, trees[t.index])
-        for keep in [True, False]:
+        for _ in [True, False]:
```
jeromekelleher: This looks like an error too.

python/tests/test_tree_stats.py
```diff
@@ -2894,7 +2893,7 @@ def naive_branch_allele_frequency_spectrum(
     for j in range(num_windows):
         begin = windows[j]
         end = windows[j + 1]
-        for set_index, sample_set in enumerate(sample_sets):
+        for _set_index, _sample_set in enumerate(sample_sets):
             S = np.zeros(out_dim)
```
jeromekelleher: Not sure what's happening here, but looks like an error.

python/tests/test_tree_stats.py
```diff
@@ -3390,7 +3389,7 @@ def f(x):
             return x * (x < n)
 
         # Determine output_dim of the function
-        for mode in ("site", "branch", "node"):
+        for _mode in ("site", "branch", "node"):
             sigma1 = ts.sample_count_stat(sample_sets, f, 3, windows=windows)
```
jeromekelleher: Definite error, ``mode`` should be an arg to sample_count_stat

python/tskit/drawing.py
```diff
@@ -26,9 +26,8 @@
 import collections
 import numbers
 
-import svgwrite
 import numpy as np
-
+import svgwrite
```
jeromekelleher: svgwrite is third party.

python/tskit/stats.py
```diff
@@ -22,16 +22,15 @@
 """
 Module responsible for computing various statistics on tree sequences.
 """
-import threading
 import struct
 import sys
-
-import numpy as np
+import threading
```
jeromekelleher: threading is from the stdlib.

benjeffery:
I agree it is getting unwieldy. My approach will be to create a new PR for black+flake8 only, then rebase this PR. That way we don't lose the comments above.

python/tests/__init__.py
```diff
@@ -19,18 +19,16 @@
...
-from .simplify import *  # NOQA
+import tskit
 from . import tsutil
+from .simplify import *  # NOQA
 
-import tskit
+# TODO remove this code and refactor elsewhere.
 
 
-class PythonTree(object):
```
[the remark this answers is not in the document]
benjeffery: This is from `pyupgrade`. I don't think it can be configured to just skip that one.

.flake8
benjeffery: That's based on committing debug lines like an idiot. (Although in my defence `git diff master` was massive)

python/tests/__init__.py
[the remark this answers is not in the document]
jeromekelleher: I'm not sure pyupgrade is worth the bother then --- the only thing it's doing that looks useful is to change .format() calls into f strings. But this isn't that big a deal --- nothing wrong with calling .format IMO.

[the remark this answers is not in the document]
benjeffery: Interesting to find something we disagree on! I find without `(object)` after the basic classes the classes that inherit stand out more. Basically "brackets near classname = you can't understand this class without checking the superclass" is a rule in my mind.

I also find f-strings to be _much_ more readable.

Neither of these things are a religious conviction, but glad we're just picking a standard to end  discussing this stuff in every PR!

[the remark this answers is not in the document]
jeromekelleher: I'm willing to let go of both of these, neither are strong convictions. Let's keep pyupgrade!

python/tests/test_tree_stats.py
petrelharp: yep - holdover from a previous version

python/tests/tsutil.py
```diff
@@ -972,7 +972,7 @@ def mean_descendants(ts, reference_sets):
     last_update = np.zeros(ts.num_nodes)
     total_span = np.zeros(ts.num_nodes)
 
-    def update_counts(edge, sign):
+    def update_counts(edge, left, sign):
```
petrelharp: wow good catch

petrelharp:
Looking pretty good.

python/tests/tsutil.py
benjeffery: This was an interesting case - the linter said `left` wasn't being used... but it was! Via the closure and the fact a `for` loop "leaks" its loop variable to the outer scope!

.flake8
benjeffery: Fixed in the other PR

python/tests/test_tree_stats.py
benjeffery: Fixed

python/tests/test_tree_stats.py
benjeffery: Putting `mode` as an arg caused the test to fail. Needed changes to `denom` to match the shape of `sigma3`.

python/tskit/drawing.py
benjeffery: Fixed

python/tskit/stats.py
benjeffery: Fixed

python/tests/test_topology.py
benjeffery: Fixed

python/tests/test_lowlevel.py
benjeffery: Fixed

benjeffery:
>get the basic infrastructure for pre-commit and black in first, and then maybe follow up with other pre-commit checks one-by-one as follow up PRs?
As all the comment threads have resolved, I've split them into their own PRs. Once #475 is merged all the diffs will be clear and show just that tools effect. I'm closing this PR in favour of those.

python/tests/test_tree_stats.py
jeromekelleher: Good catch!

python/tests/tsutil.py
jeromekelleher: Wow, good work linter.

PR #475, opened 2020-03-12
Add pre-commit for black and flake8

benjeffery:
Extracting pre-commit work from #471 for just black and flake8 so it can be merged before the rest.
Not yet ready for review as I'm getting the github action working.


benjeffery:
@jeromekelleher As part of this PR I was going to move the pre-commit checks to a github action, but it seems I don't have permission to do this - the message is "Only repository collaborators may create workflows."

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,20 @@
+name: pre-commit
```
jeromekelleher: Maybe "lint" is a better name for this action? "pre-commit" is a bit confusing and just an implementation detail of how we're implementing the lint check.

docs/development.rst
```diff
@@ -33,11 +33,13 @@ Quickstart
...
 - Run the tests to ensure everything has worked: ``python -m nose -vs``. These should
   all pass.
-
-- Make your changes in a local branch, and open a pull request on GitHub when you
-  are ready. Please make sure that (a) the tests pass before you open the PR; and
-  (b) your code passes PEP8 checks (see :ref:`sec_development_code_style` below
-  for a git commit hook to ensure this happens automatically) before opening the PR.
+- Install the pre-commit checks: ``pre-commit install``
+- Make your changes in a local branch, on each commit checks will be run for code style
+  and common problems (see :ref:`sec_development_code_style`), sometimes these will
```
jeromekelleher: This should be two sentences I think?

jeromekelleher:
> @jeromekelleher As part of this PR I was going to move the pre-commit checks to a github action, but it seems I don't have permission to do this - the message is "Only repository collaborators may create workflows."

I don't really know how to fix this --- I've looked through the settings and can't find an easy way to add you as a collaborator. I had hoped to make you an admin of tskit-dev in a few weeks once you'd met the rest of the core developers (and then you'd have full admin rights for everything).

In the mean time, I guess we could merge this PR an I can follow up with whatever changes are needed to turn on the action?

docs/development.rst
petrelharp: also a bit more explanation would be helpful for people unfamiliar - e.g. say that the code changes are formatting only; that the code is guaranteed to do the same thing.

benjeffery: Agreed, I've also added quite a bit more text about how to run checks without committing, and committing without checks.

benjeffery:
In retrospect I think I am jumping the gun by adding the github action. I've reverted to using the current CI. Let's get this set of work wrapped up and schedule a CI/actions review for later?

jeromekelleher:
OK, merged, thanks @benjeffery. FYI I've used the "rebase and merge" option here (the "merge" option isn't available now because we've selected "enforce linear history").

benjeffery:
Looks like the only difference is that merging the PR didn't add a merge commit. Its a trade off of simpler history Vs recording each PR merge.

PR #478, opened 2020-03-13
Add pre-commit fixing of import order

benjeffery:
Adds `reorder_python_imports` which now works how we want it after https://github.com/asottile/reorder_python_imports/pull/123


python/tskit/tables.py
```diff
@@ -32,15 +32,14 @@
 import numpy as np
 
 import _tskit
+import tskit
+import tskit.provenance as provenance
+import tskit.util as util
 
 # This circular import is ugly but it seems hard to avoid it since table collection
```
benjeffery: Is this comment still relevant? I'm not seeing the issue with the import.

docs/examples.py
```diff
@@ -1,13 +1,14 @@
...
-import os
 import io
+import os
 import sys
 
 sys.path.insert(0, os.path.abspath("../python"))
 
 import numpy as np  # noqa: E402
+
 import msprime  # noqa: E402
```
benjeffery: Something about this being in the docs has made msprime part of the local set. I think it is the only "error" left. Not looking into it for now.

benjeffery: Fixed, wasn't being rearranged as was below code.

python/tskit/tables.py
jeromekelleher: No, I think it's fine. I used to work with someone who **really** hated circular imports so I've been apologetic about it since. It's nothing to worry about though, so yeah, let's dump the comment.

jeromekelleher:
Perfect! Just need to remove the out-of-date comment and we're good to merge.

python/tskit/tables.py

Issue #77, opened 2019-01-13
Change Python C module to use C99 style initialisers


jeromekelleher:
Change the struct initialisers in ``_tskitmodule.c`` to use the [recommended](
https://docs.python.org/3.6/extending/newtypes_tutorial.html#the-basics) c99 style initialisers. Eg, a Type definition would look like:

```C
static PyTypeObject LdCalculatorType = {
    PyVarObject_HEAD_INIT(NULL, 0)
    .tp_name = "_tskit.LdCalculator",
    .tp_doc = "LdCalculator objects.",
    .tp_basicsize = sizeof(LdCalculator),
    .tp_itemsize = 0,
    .tp_flags = Py_TPFLAGS_DEFAULT,
    .tp_new = PyType_GenericNew,
    .tp_init = (initproc) LdCalculator_init,
    .tp_dealloc = (destructor) LdCalculator_dealloc,
    .tp_methods = LdCalculator_methods,
    .tp_members = LdCalculator_members,
};
```

Also remove the ``-Wno-missing-field-initializers`` compiler flag from the Makefile and Circle CI compiler step. 

There may be issues with portability, so this should not be done until CI on Appveyor and Travis have been fully enables (see #41).

Good starter issue for someone looking to understand the Python C API a little better.

jeromekelleher:
We can't do this until Python 2 support is dropped because the Windows compiler for py27 doesn't support it.

jeromekelleher:
Good to get this done as part of the interchange infrastructure updates.

benjeffery:
Should any one else need to do this again here's the regex `^\s*(.*),\s*/\*\s*([a-z_]*)\s*\*/`  -> `.$2 = $1,`

PR #482, opened 2020-03-13
Add clang-format to pre commit

benjeffery:
[clang-format](https://clang.llvm.org/docs/ClangFormat.html) auto formats C code. This is very much a discussion, as the changes with these settings are drastic!


c/examples/api_structure.c
```diff
@@ -2,25 +2,24 @@
...
+  }
 
-int
-main(int argc, char **argv)
-{
-    int j, ret;
-    tsk_edge_table_t edges;
+int main(int argc, char **argv) {
+  int j, ret;
+  tsk_edge_table_t edges;
```
grahamgower: IMO having the function name start on a newline makes the code more grep friendly: `grep -n ^main *.c`.

And changing the indent settings to use 4 spaces will probably reduce this diff dramatically.

grahamgower:
Maybe its worth looking trying GNU indent? Its been around forever, and is very flexible. Something like `indent -kr -nut -i4 -psl` (but probably needs a few extra tweaks).

jeromekelleher:
OK, this is a good start! We'll need to craft the ``.clang-format`` a bit to get something reasonably close to what we have (I quite like the current style, as it happens...) but hopefully this is possible. @grahamgower is right that most of the changes are because of the indent size and the diff will look much less scary when we fix that.

jeromekelleher:
@molpopgen, you've used clang-format. Do you think we can configure it to get something like the current C style? I've always thought of this as a mixture of K&R and Linux kernel styles with a dash of pedantry around always using braces etc.

molpopgen:
A `.clang-format` file with the following is very close.  If I apply it to `tables.c`, the diff isn't that big, and several of the changes "fix" style inconsistencies within the file.

The options are here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

```
---
Language: Cpp
BasedOnStyle: GNU
SortIncludes:    false
AllowShortIfStatementsOnASingleLine: false
BreakBeforeBraces: Linux
TabWidth:        4
IndentWidth:     4
ColumnLimit:     89
SpaceBeforeParens:
    ControlStatements
SpacesInCStyleCastParentheses: false
SpaceAfterCStyleCast: true
IndentCaseLabels: true
```

benjeffery:
I've modified the configuration to use @molpopgen's file. Looks much better to me.

jeromekelleher:
> A .clang-format file with the following is very close. If I apply it to tables.c, the diff isn't that big, and several of the changes "fix" style inconsistencies within the file.

Thanks @molpopgen, this is great!

jeromekelleher:
I really like this too. I think the only thing I would gripe about is the visual indentation when breaking lines. Is there anyway to change to hanging indents?

molpopgen:
> I really like this too. I think the only thing I would gripe about is the visual indentation when breaking lines. Is there anyway to change to hanging indents?

Try adding this at the end and see if you like it:

```
AlignAfterOpenBracket: DontAlign
```

molpopgen:
@jeromekelleher  -- once you go down this road, it is time to install https://github.com/rhysd/vim-clang-format and then just `:ClangFormat` every now and then when editing.

molpopgen:
If you want to get a little edgy now, this will make the function parameter and arguments "black-like":

```
BinPackArguments: false
BinPackParameters: false
```

benjeffery:
I've added:
```
AlignAfterOpenBracket: DontAlign
BinPackArguments: false
BinPackParameters: false
```

molpopgen:
> I like it, on the whole. Functions with lots of args arg taking a lot of vertical space now, but I guess we can live with that.

Yes, I think this one could be "controversial", and it is due to one/both of these:

```
BinPackArguments: false
BinPackParameters: false
```

It matches how `black` treats Python, but it doesn't feel very "C" because we can't just fall back on `**kwargs`.

molpopgen:
> Looks like _tskitmodule got badly mangled also.

It seems that types in ALLCAPS are signalling something special to clang-format.  I can mess around a bit locally to see if I can sort that.

molpopgen:
This will restore macro alignment, if they are declared one after the other:

```
AlignConsecutiveMacros: true
```

molpopgen:
To restore `PyObject_HEAD` to being alone on a line:

```
StatementMacros: ['PyObject_HEAD']
```

There are possibly other such "statement macros" that could be added?

molpopgen:
To force a newline break after the return type in all function definitions:

```
AlwaysBreakAfterReturnType: AllDefinitions
```

Adding this removes a lot of diff'd lines.

molpopgen:
> It seems that types in ALLCAPS are signalling something special to clang-format. I can mess around a bit locally to see if I can sort that.

This is fixed by my last suggestion--the PyObject_HEAD is a corner case!

benjeffery:
> To restore `PyObject_HEAD` to being alone on a line:
> 
> ```
> StatementMacros: ['PyObject_HEAD']

What do you get for `clang-format --version`? I have `6.0.0-1ubuntu2` and it doesn't recognise this option. Tried to find it in the clang release notes and failed!

molpopgen:
I have 9 on Ubuntu.

python/_tskitmodule.c
```diff
@@ -11,8 +11,8 @@
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
  *
- * The above copyright notice and this permission notice shall be included in all
- * copies or substantial portions of the Software.
+ * The above copyright notice and this permission notice shall be included in
```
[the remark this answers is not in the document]
benjeffery: Fixed broken symlink

benjeffery:
I've added 
```
AlignConsecutiveMacros: true
StatementMacros: ["PyObject_HEAD"]
AlwaysBreakAfterReturnType: AllDefinitions
```
I think if the arguments of a function are one-per-line then the first arg should be on a new line, but I can't see an option for this, e.g.:
```
struct arg_int *num_samples1 = arg_int0("s",
        "sample-size",
        "<sample-size>",
        "Number of samples to keep in the simplified tree sequence.");
```
would be better as:
```
struct arg_int *num_samples1 = arg_int0(
        "s",
        "sample-size",
        "<sample-size>",
        "Number of samples to keep in the simplified tree sequence."
);
```
I find I miss the first arg otherwise, also better for merges.

molpopgen:
> I think if the arguments of a function are one-per-line then the first arg should be on a new line, but I can't see an option for this, e.g.:

This could be:

```
BinPackArguments: false
BinPackParameters: true
```

When it breaks on a set of arguments to a fxn, it looks like your 2nd example.

benjeffery:
Circle CI is failing as it doesn't have clang-format-9. I could pull in a PPA, if it is worth keeping the linting checks there too?

molpopgen:
The StatementMacros business first showed in in clang 8.  So we either punt on this particular formatting issue or pull in a newer clang via the PPA?  Thoughts, @jeromekelleher ?

jeromekelleher:
Thanks  for all this @molpopgen and @benjeffery, I think we're converging on a style definition pretty quickly. Some thoughts:

- We should be conservative about what version of clang-format we require. I would say let's keep the features we're using to whatever is available in Debian stable (v 7.0.1). This should resolve any CI problems as well.
- I don't think we should add this to ``pre-commit``. The vast majority of contributors are only going to be working with the Python code, so asking them to install ``clang-format`` for no reason would be annoying and off putting. Just checking as part of CI will be fine (anyone who is sophisticated enough to contribute to the C library won't be bothered by this).
- On a similar note, I think the simplest thing might be to just exclude _tskitmodule.c from the checks. Any additions to this need careful review anyway, and I don't mind keeping an eye on the style issues here (plus, this needs a c99 refactor anyway, see #77, which might resolve some of the stuff that's happening with clang-format now).
- On reflection, I like the black-ish one argument per line thing less, as vertical space is valuable in C code and we do have quite a few functions with lots of args. Can we do a hanging indent, with packing args onto as few lines as possible?  (I don't mind whether we put args on the first line or not)

How does all that sound?

benjeffery:
I've just found https://pypi.org/project/clang-format/ which installs v9 (what we want) of `clang-format` via `pip`. This means the C code checking will not require any changes to the install instructions, and we can have the right version everywhere.

Agree on excluding `_tskitmodule.c` and the style suggestion.

benjeffery:
Ah, it seems that package only supports linux :( 

Debian stable has `clang-format-7`. I'll change things to use this.

benjeffery:
Ok, just the CI to fix. Annoyingly `clang-format` doesn't give exit codes!

benjeffery:
Ok, just the last style tweaks I think.

.circleci/config.yml
```diff
@@ -16,7 +16,8 @@ jobs:
       - run:
           name: Install dependencies and set path
           command: |
-            sudo apt-get install ninja-build libcunit1-dev valgrind clang doxygen libgsl-dev
+            sudo apt-get install ninja-build libcunit1-dev valgrind clang doxygen
+            sudo apt-get install libgsl-dev
```
jeromekelleher: Any particular reason for install gsl-dev in a second apt-get?

c/examples/haploid_wright_fisher.c
```diff
@@ -25,8 +27,8 @@ simulate(tsk_table_collection_t *tables, int N, int T, int simplify_interval, gs
     tables->sequence_length = 1.0;
     parents = buffer;
     for (j = 0; j < N; j++) {
-        parents[j] = tsk_node_table_add_row(&tables->nodes, 0, T,
-                TSK_NULL, TSK_NULL, NULL, 0);
+        parents[j]
```
jeromekelleher: This looks weird to me - why is it putting the ``= tsk...()`` on the next line?

c/dev-tools/dev-cli.c
```diff
@@ -124,7 +124,11 @@ print_ld_matrix(tsk_treeseq_t *ts)
                 fatal_library_error(ret, "get_site");
             }
             printf("%d\t%f\t%d\t%f\t%.3f\n",
-                (int) sA.id, sA.position, (int) sB.id, sB.position, r2[k]);
+                (int) sA.id,
```
jeromekelleher: It's not obvious what the rules are for going into one-arg-per-line mode. Although in this case it's probably a good thing all right.

c/tests/test_convert.c
```diff
@@ -73,9 +80,15 @@ test_single_tree_newick_errors(void)
     size_t buffer_size = 1024;
     char newick[buffer_size];
 
-
-    tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges,
-            NULL, NULL, NULL, NULL, NULL);
+    tsk_treeseq_from_text(&ts,
+        1,
```
jeromekelleher: This is a case where we don't want one-arg-per line I think.

c/examples/haploid_wright_fisher.c
molpopgen: This is due to line lenght, I think. There are params one can add to allow slop in breaking due to lenght, kinda like what black allows.

.circleci/config.yml
benjeffery: It was either that or a line continuation id we have the same line length criteria in YAML. I've committed a line continuation in https://github.com/tskit-dev/tskit/pull/482/commits/30505576fc8c68841bff8faf6d4c778c62544656

jeromekelleher: Bit faster to do it all in one apt-get transaction so worth the line break.

benjeffery:
Opps, reverted `_tskitmodule.c` in https://github.com/tskit-dev/tskit/pull/482/commits/e043eb98342460e1b989a2492e7c7a92ac5953d4

molpopgen:
> How does all that sound?

This sounds good.  We should include the `.clang-format` in the relevant places so that people working on the C code have access to it.

c/tests/test_convert.c
molpopgen: I'm new to this option.  Not sure if there are ways to say "break if args are short" or not.

benjeffery:
Currently its at `c/.clang-format`. The tool goes up through the parent directories till it finds one so this should be the right place.

jeromekelleher:
OK, I think we should set 
```
BinPackArguments: true
￼BinPackParameters: true
```
and merge this ASAP. I think they use too much vspace in C code. Everything else is pretty much as  it was except for losing some alignment in macro definitions (which I can live without). 

We can always refine the style definition a little later with some tweaks but I think this is a solid basis for us to move forward one.

benjeffery:
```
BinPackArguments: true
BinPackParameters: true
```
set in https://github.com/tskit-dev/tskit/pull/482/commits/d58dc8987a3f9289a77c81c8a6b34e939d28722c

c/tskit/trees.h
```diff
@@ -36,22 +36,22 @@ extern "C" {
 
 #include <tskit/tables.h>
```
jeromekelleher: Let's just put a ``// clang-format off`` around all the macro definitions here. It's handy having alignment, makes it more readable. (I guess we can put in a note saying that clang-format 9 has the option we want?)

docs/development.rst
```diff
@@ -44,6 +45,11 @@ Quickstart
   code. To run the checks without committing use ``pre-commit run``. To bypass
   the checks (to save or get feedback on work-in-progress) use ``git commit
   --no-verify``
+- If you have modified C code then run this to make it conform to the project style::
+
+  $ sudo apt-get install clang-format-7
```
jeromekelleher: Do we absolutely need clang-format 7? I was hoping that anything >= 7 would work, and choosing an old version would 'just work' for the vast majority of people.

benjeffery: I've stuck with specifying it as `sudo apt-get install clang-format` on my latest "mint" install gives version 6! Even though you can `sudo apt-get install clang-format-9`.  No idea why the default is stuck at 6.

molpopgen:
Minimum version for the macro lists is 8.

c/tests/test_trees.c
```diff
@@ -4029,15 +3932,39 @@ static void
 test_simple_multi_tree(void)
 {
     tsk_id_t parents[] = {
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/482/commits/37c7f83c4568e6444d9a102de4f00580775674da 

BTW - you can do a multi-line comment by dragging the margin of the code.

c/tskit/trees.h
benjeffery: Also fixed in https://github.com/tskit-dev/tskit/pull/482/commits/37c7f83c4568e6444d9a102de4f00580775674da

I've left a note in `.clang-format` for when we upgrade.

docs/development.rst
jeromekelleher: Jeez, that's lame. Can we target >= 6 then, or are there some options that are different? I just ran version 6 on my mint machine and I didn't notice any issues (didn't run it on all the code though).

jeromekelleher: I'm just thinking of OSX users here --- they might think they absolutely have to have version 7 and it could be a major headache to to track down version 7 on whatever bolted on excuse for a software manager they happen to be using.

c/tskit/trees.h
jeromekelleher: Great, thanks. Can we do the same for the error macros in ``core.h`` please?

Nearly there...

docs/development.rst
benjeffery: Looks like >= 6 is ok. I had gone with 7 as it was on debian stable, but lets take pity on the OSX developer.

c/tskit/trees.h
benjeffery: Done

benjeffery:
Squashed and rebased!

benjeffery:
Spoke too soon CI doesn't like v6, fixing.

benjeffery:
CI is unhappy with v6 "Unable to locate package clang-format-6"

jeromekelleher:
> CI is unhappy with v6 "Unable to locate package clang-format-6" I suggest we revert to 7 for now, and check the OSX avaliability.

Presumably if we install ``clang-format`` in bionic (18.04) we'll get version 6? That's what's happening on our mint machines, right?

.github/workflows/lint.yml
```diff
@@ -6,8 +6,19 @@ on:
     branches: [master]
 
 jobs:
+  lint-c:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+    - name: install clang-format
+      run: sudo apt-get install clang-format clang-format-6
```
jeromekelleher: We should just be able to delete ``clang-format-6`` and it'll work, right?

benjeffery: You'd think so right! 
```
The following packages have unmet dependencies:
 clang-format : Depends: clang-format-6.0 (>= 6.0~) but it is not going to be installed
```

benjeffery: I'm an idiot and didn't read the error correctly, unlike 7, 8 and 9 it is 6.0.

benjeffery: Doesn't work with 6.0 either:
```
Te following packages have unmet dependencies:
 clang-format-6.0 : Depends: libllvm6.0 (= 1:6.0-1ubuntu2) but 1:6.0.1~svn334776-1~exp1~20190309042703.125 is to be installed
```
I'm firing up a "bionic minimal" VM to try.

jeromekelleher: Lame! clang-format-7 is fine, it's just for running CI. One of us will spot it quickly if something gets added that doesn't work on 6.

benjeffery: Fixed! The github image has a microsoft azure ppa that installs a cutom LLVM build.

jeromekelleher:
Merging! Thanks for slogging through this one @benjeffery, we got there eventually. Style discussions are always painful, and at least we've gotten this out of the way early.

Thanks a million for the help with clang-format @molpopgen, I really appreciate it!

PR #486, opened 2020-03-17
Refactor LS algorithm definitions.

jeromekelleher:
Update the LS haplotype matching algorithm to round after normalisation rather than before.

Related to https://github.com/tskit-dev/tsinfer/issues/246

This part of tskit is under development @benjeffery. The goal is to have a general purpose, very efficient haplotype matching library. The method hasn't been properly analysed or validated, so it's a "private" API for the moment. A paper will be written in 2020...

I've deliberately put in some badly formatted C code here to see what happens also.

### DIFF
--- a/c/tests/test_haplotype_matching.c
+++ b/c/tests/test_haplotype_matching.c
@@ -32,17 +32,10 @@
  * TestHMM
  ****************************************************************/
 
-static int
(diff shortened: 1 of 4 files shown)


jeromekelleher:
> I've deliberately put in some badly formatted C code here to see what happens also.

Works great!

PR #488, opened 2020-03-18
Python implementation of IBD methods.

gtsambos:
Pulling this in now so that others can see my continuing progress on IBD-calculating methods for `tskit` (cc @jeromekelleher for our call tomorrow). It's very basic so far but it should work:

**Command-line implementation:**
```
python3 tests/ibd.py test.trees
```
should return IBD segments shared between all pairs of samples.

```
python3 tests/ibd.py test.trees 0 1
```

should return IBD segments shared between samples 0 and 1 (ensure the nodes are ordered).

**Tests:**
```
python3 -m nose -vs tests/test_ibd.py
```
### IBD definitions

A pair of sample nodes that have both inherited the interval `(left, right)` from a single ancestor, `node`, are IBD over the interval `(left, right)`.

Further options:

 - `path_ibd`: If `True`, the IBD segment must be inherited by two samples from a common ancestor without being broken by recombination since the time of that ancestor. Equivalently, the path of the tree connecting the sample pair must be the same along the whole span of the interval.
 - `mrca_ibd`: If `True`, only IBD segments at the MRCA of the sample pair are recorded.

Cutoff criteria:

 - `min_length`: Only segments longer than `min_length` are returned.
 - `max_time`: Only segments from common ancestors who lived more recently than `max_time` are returned.

Note: currently, the only implemented version is `path_ibd=True`, `mrca_ibd=True`, `min_length=0`, `max_time=0`, but watch this space!

### DIFF
--- a/python/tests/ibd.py
+++ b/python/tests/ibd.py
@@ -24,6 +24,8 @@ Python implementation of the IBD-finding algorithms.
 """
 import argparse
 
+import numpy as np
+
 import tskit
 
 
@@ -96,15 +98,6 @@ class SegmentList:
             s = "[{}, ..., {}]".format(repr(self.head), repr(self.tail))
         return s
 
-    def __len__(self):
-        # Returns the number of segments in the list.
-        count = 0
-        seg = self.head
-        while seg is not None:
-            count += 1
-            seg = seg.next
-        return count
-
     def add(self, other):
         """
         Use to append another SegmentList, or a single segment.
@@ -143,22 +136,23 @@ class IbdFinder:
             self.samples = ts.samples()
         else:
             self.samples = samples
-        self.samples.sort()
+        if len(self.samples) == 0:
+            raise ValueError("The tree sequence contains no samples.")
+
+        self.sample_id_map = np.zeros(ts.num_nodes, dtype=int) - 1
+        for index, u in enumerate(self.samples):
+            self.sample_id_map[u] = index
         self.min_length = min_length
-        self.max_time = max_time
-        self.current_parent = self.ts.tables.edges.parent[0]
-        self.current_time = 0
+        if max_time is None:
+            self.max_time = 2 * ts.max_root_time
+        else:
+            self.max_time = max_time
         self.A = [None for _ in range(ts.num_nodes)]  # Descendant segments
         self.tables = self.ts.tables
-        self.parent_list = [[] for i in range(0, self.ts.num_nodes)]
-        for e in self.ts.tables.edges:
-            if (
-                len(self.parent_list[e.child]) == 0
-                or e.parent != self.parent_list[e.child][-1]
-            ):
-                self.parent_list[e.child].append(e.parent)
+
+        self.oldest_parent = self.get_oldest_parents()
+
         # Objects below are needed for the IBD segment-holding object.
-        self.sample_id_map = self.get_sample_id_map()
         self.num_samples = len(self.samples)
         self.sample_pairs = self.get_sample_pairs()
 
@@ -170,6 +164,18 @@ class IbdFinder:
         self.ibd_segments = {}
         for key in self.sample_pairs:
             self.ibd_segments[key] = None
+
+    def get_oldest_parents(self):
+        oldest_parents = [-1 for _ in range(self.ts.num_nodes)]
+        node_times = self.ts.tables.nodes.time
+        for e in self.ts.tables.edges:
+            c = e.child
+            if (
+                oldest_parents[c] == -1
+                or node_times[oldest_parents[c]] < node_times[e.parent]
+            ):
+                oldest_parents[c] = e.parent
+        return oldest_parents
 
     def add_ibd_segments(self, sample0, sample1, seg):
         index = self.find_sample_pair_index(sample0, sample1)
@@ -185,19 +191,6 @@ class IbdFinder:
             )
         else:
             self.ibd_segments[self.sample_pairs[index]].add(seg)
-
-    def get_sample_id_map(self):
-        """
-        Returns id_map, a vector of length ts.num_nodes. For a node with id i,
-        id_map[i] is the position of the node in self.samples. If i is not a
-        user-specified sample, id_map[i] is -1.
-        Note: it assumes nodes in the TS are numbered 0 to ts.num_nodes - 1
-        """
-        id_map = [-1] * self.ts.num_nodes
-        for i, samp in enumerate(self.samples):
-            id_map[samp] = i
-
-        return id_map
 
     def get_sample_pairs(self):
         """
@@ -240,167 +233,104 @@ class IbdFinder:
 
         return int(index)
 
-    def find_ibd_segments(self, min_length=0, max_time=None):
+    def find_ibd_segments(self):
         """
         The wrapper for the procedure that calculates IBD segments.
         """
 
         # Set up an iterator over the edges in the tree sequence.
-        mygen = iter(self.ts.edges())
-        e = next(mygen)
+        edges_iter = iter(self.ts.edges())
+        e = next(edges_iter)
+        parent_should_be_added = True
+        node_times = self.tables.nodes.time
 
         # Iterate over the edges.
         while e is not None:
-            # Process all edges with the same parent node.
-            S_head = None
-            S_tail = None
-            self.current_parent = e.parent
-            self.current_time = self.tables.nodes.time[self.current_parent]
-            if self.max_time is not None and self.current_time > self.max_time:
+
+            current_parent = e.parent
+            current_time = node_times[current_parent]
+            if current_time > self.max_time:
                 # Stop looking for IBD segments once the
                 # processed nodes are older than the max time.
                 break
 
-            # Create a list of segments, S, bookended by S_head and S_tail.
-            # The list holds all segments that are immediate descendants of
-            # the current parent node being processed.
-            while e is not None and self.current_parent == e.parent:
-                seg = Segment(e.left, e.right, e.child)
-                if S_head is None:
-                    S_head = seg
-                    S_tail = seg
-                else:
-                    S_tail.next = seg
-                    S_tail = S_tail.next
-                e = next(mygen, None)
+            seg = Segment(e.left, e.right, e.child)
 
-            # Create a list of SegmentList objects, SegL, bookended by SegL_head and
-            # SegL_tail. Each SegmentList corresponds to a segment s in S, and
-            # contains all segments of samples that descend from s.
-            SegL_head = SegmentList()
-            SegL_tail = SegmentList()
+            # Create a SegmentList() holding all segments that descend from seg.
+            list_to_add = SegmentList()
+            u = seg.node
+            if self.sample_id_map[u] != tskit.NULL:
+                list_to_add.add(seg)
+            else:
+                if self.A[u] is not None:
+                    s = self.A[u].head
+                    while s is not None:
+                        intvl = (
+                            max(seg.left, s.left),
+                            min(seg.right, s.right),
+                        )
+                        if intvl[1] - intvl[0] > 0:
+                            list_to_add.add(Segment(intvl[0], intvl[1], s.node))
+                        s = s.next
 
-            while S_head is not None:
-                u = S_head.node
-                list_to_add = SegmentList()
-                if u in self.samples:
-                    list_to_add.add(Segment(S_head.left, S_head.right, u))
-                else:
-                    if self.A[u] is not None:
-                        s = self.A[u].head
-                        while s is not None:
-                            intvl = (
-                                max(S_head.left, s.left),
-                                min(S_head.right, s.right),
-                            )
-                            if intvl[1] - intvl[0] > 0:
-                                list_to_add.add(Segment(intvl[0], intvl[1], s.node))
-                            s = s.next
+            if list_to_add.head is not None:
+                self.calculate_ibd_segs(current_parent, list_to_add)
 
-                # Add this list to the end of SegL.
-                if SegL_head.head is None:
-                    SegL_head = list_to_add
-                    SegL_tail = list_to_add
-                else:
-                    SegL_tail.next = list_to_add
-                    SegL_tail = list_to_add
+            # For parents that are also samples
+            if (
+                self.sample_id_map[current_parent] != tskit.NULL
+            ) and parent_should_be_added:
+                singleton_seg = SegmentList()
+                singleton_seg.add(Segment(0, self.ts.sequence_length, current_parent))
+                self.calculate_ibd_segs(current_parent, singleton_seg)
+                parent_should_be_added = False
 
-                S_head = S_head.next
+            # Move to next edge.
+            e = next(edges_iter, None)
 
-            # If we wanted to squash segments, we'd do it here.
-
-            # Use the info in SegL to find new ibd segments that descend from the
-            # parent node currently being processed.
-            if SegL_head.next is not None or self.current_parent in self.samples:
-                nodes_to_remove = self.calculate_ibd_segs(SegL_head)
-
-                # g. Remove elements of the list of ancestral segments A if they
-                # are no longer needed. (Memory-pruning step)
-                for n in nodes_to_remove:
-                    if self.current_parent in self.parent_list[n]:
-                        self.parent_list[n].remove(self.current_parent)
-                    if len(self.parent_list[n]) == 0:
-                        self.A[n] = []
-
-            # Create the list of ancestral segments, A, descending from current node.
-            # Concatenate all of the segment lists in SegL into a single list, and
-            # store in A. This is a list of all sample segments descending from
-            # the currently processed node.
-            seglist = SegL_head
-            self.A[self.current_parent] = seglist
-            seglist = seglist.next
-            while seglist is not None:
-                self.A[self.current_parent].add(seglist)
-                seglist = seglist.next
+            # Remove any processed nodes that are no longer needed.
+            if e is not None and e.parent != current_parent:
+                for i, n in enumerate(self.oldest_parent):
+                    if current_parent == n:
+                        self.A[i] = None
 
         return self.ibd_segments
 
-    def calculate_ibd_segs(self, SegL_head):
+    def calculate_ibd_segs(self, current_parent, list_to_add):
         """
-        Calculates all new IBD segments found at the current iteration of the
-        algorithm. Returns information about nodes processed in this step, which
-        allows memory to be cleared as the procedure runs.
+        Write later.
         """
 
-        list0 = SegL_head
-        # Iterate through all pairs of segment lists.
-        while list0.next is not None:
-            list1 = list0.next
-            while list1 is not None:
-                seg0 = list0.head
-                # Iterate through all segments with one taken from each list.
-                while seg0 is not None:
-                    seg1 = list1.head
-                    while seg1 is not None:
-                        left = max(seg0.left, seg1.left)
-                        right = min(seg0.right, seg1.right)
-                        if left >= right:
-                            seg1 = seg1.next
-                            continue
-                        nodes = [seg0.node, seg1.node]
-                        nodes.sort()
+        if list_to_add.head is None:
+            return []
 
-                        # If there are any overlapping segments, record as a new
-                        # IBD relationship.
-                        if right - left > self.min_length:
-                            self.add_ibd_segments(
-                                nodes[0],
-                                nodes[1],
-                                Segment(left, right, self.current_parent),
-                            )
+        if self.A[current_parent] is None:
+            self.A[current_parent] = list_to_add
 
+        else:
+            seg0 = self.A[current_parent].head
+            while seg0 is not None:
+                seg1 = list_to_add.head
+                while seg1 is not None:
+                    left = max(seg0.left, seg1.left)
+                    right = min(seg0.right, seg1.right)
+                    if left >= right:
                         seg1 = seg1.next
+                        continue
+                    nodes = [seg0.node, seg1.node]
+                    nodes.sort()
 
-                    seg0 = seg0.next
-                list1 = list1.next
-            list0 = list0.next
+                    # If there are any overlapping segments, record as a new
+                    # IBD relationship.
+                    if right - left > self.min_length:
+                        self.add_ibd_segments(
+                            nodes[0], nodes[1], Segment(left, right, current_parent),
+                        )
+                    seg1 = seg1.next
+                seg0 = seg0.next
 
-        # If the current processed node is itself a sample, calculate IBD separately.
-        if self.current_parent in self.samples:
-            seglist = SegL_head
-            while seglist is not None:
-                seg = seglist.head
-                while seg is not None:
-                    self.add_ibd_segments(
-                        seg.node,
-                        self.current_parent,
-                        Segment(seg.left, seg.right, self.current_parent),
-                    )
-                    seg = seg.next
-                seglist = seglist.next
-
-        # Specify elements of A that can be removed (for memory-pruning step)
-        processed_child_nodes = []
-        seglist = SegL_head
-        while seglist is not None:
-            seg = seglist.head
-            while seg is not None:
-                processed_child_nodes += [seg.node]
-                seg = seg.next
-            seglist = seglist.next
-        processed_child_nodes = list(set(processed_child_nodes))
-
-        return processed_child_nodes
+            # Add list_to_add to A[u].
+            self.A[current_parent].add(list_to_add)
 
 
 if __name__ == "__main__":
--- a/python/tests/ibd.py
+++ b/python/tests/ibd.py
@@ -28,6 +28,7 @@ import tskit
 import numpy as np
 import itertools
 import sys
+import argparse
 
 
 class Segment(object):
@@ -59,6 +60,27 @@ class Segment(object):
         return (self.node, self.left, self.right) < (other.node, other.left, other.right)
 
 
+class SegmentList(object):
+    """
+    A class representing a list of segments that are descended from a given ancestral node
+    via a particular child of the ancestor.
+    """
+    def __init__(self, head=None, tail=None, next=None):
+        self.head = head
+        self.tail = tail
+        self.next = next
+
+    def __str__(self):
+        s = "head={},tail={},next={}".format(
+            self.head, self.right, repr(self.next))
+        return s
+
+    def __repr__(self):
+        s = "[{}...]".format(repr(self.next.head))
+        return s
+
+
+
 class IbdFinder(object):
     """
     Finds all IBD relationships between specified samples in a tree sequence.
@@ -67,23 +89,28 @@ class IbdFinder(object):
     def __init__(
         self,
         ts,
-        samples,
-        min_length=0):
+        samples=None,
+        min_length=0,
+        max_time=None):
 
         self.ts = ts
-        self.samples = samples
+        if samples is None:
+            samples = ts.samples()
+        else:
+            self.samples = samples
         self.min_length = min_length
+        self.max_time = max_time
         self.current_parent = self.ts.tables.edges.parent[0]
-        self.A_head = [None for _ in range(ts.num_nodes)]
-        self.A_tail = [None for _ in range(ts.num_nodes)]
-        self.tables = tskit.TableCollection(sequence_length=ts.sequence_length)
+        self.current_time = 0
+        self.A = [[] for _ in range(ts.num_nodes)] # Descendant segments
+        self.tables = self.ts.tables
         # self.ibd_segments = dict.fromkeys(itertools.combinations(self.ts.samples(), 2), [])
 
 
-    def find_ibd_segments_of_length(self, min_length=0):
+    def find_ibd_segments(self, min_length=0, max_time=None):
         
         # 1
-        A = [[] for n in range(0, self.ts.num_nodes)]
+        # A = [[] for n in range(0, self.ts.num_nodes)]
         ibd_segments = dict.fromkeys(itertools.combinations(self.ts.samples(), 2), [])
         edges = self.ts.edges()  
         parent_list = self.list_of_parents() ## Needed for memory-pruning step
@@ -96,43 +123,53 @@ class IbdFinder(object):
         while e is not None:
 
             # 3a
-            S = []
+            S_head = None
+            S_tail = None
             self.current_parent = e.parent
+            self.current_time = self.tables.nodes.time[self.current_parent]
+            if self.max_time is not None and self.current_time > self.max_time:
+                # Stop looking for IBD segments once the
+                # processed nodes are older than the max time.
+                break
             
             # 3b
             while e is not None and self.current_parent == e.parent:
                 # Create the list S of immediate descendants of u.
-                S.append(Segment(e.left, e.right, e.child))
-                # if e.id < edges.num_rows - 1:
+                seg = Segment(e.left, e.right, e.child)
+                if S_head is None:
+                    S_head = seg
+                    S_tail = seg
+                else:
+                    S_tail.next = seg
+                    S_tail = S_tail.next
                 if e.id < self.ts.num_edges - 1:
                     e = next(mygen)
                     continue
                 else:
                     e = None
-                    # break
 
             # 3c
-            for seg in S:
+            while S_head is not None:
                 # Create A[u] from S.
                 # Do we still need to do the initialisation if the below is there??
-                u = seg.node
+                u = S_head.node
                 if u in self.ts.samples():
-                    A[self.current_parent].append([seg])
+                    self.A[self.current_parent].append([S_head])
                 else:
                     list_to_add = []
-                    for s in A[u]:
-                        l = (max(seg.left, s.left), min(seg.right, s.right))
+                    for s in self.A[u]:
+                        l = (max(S_head.left, s.left), min(S_head.right, s.right))
                         if l[1] - l[0] > 0:
                             list_to_add.append(Segment(l[0], l[1], s.node))
-                    A[self.current_parent].append(list_to_add)
+                    self.A[self.current_parent].append(list_to_add)
+                S_head = S_head.next
 
             # d. Squash
-            # A[self.current_parent] = self.squash(A[self.current_parent])
+            # A[self.current_parent] = self.squash(self.A[self.current_parent])
             
             # e. Process A[self.current_parent]
-            if len(A[self.current_parent]) > 1:
-                new_segs, nodes_to_remove = self.update_A_and_find_ibd_segs(
-                    A[self.current_parent], ibd_segments)
+            if len(self.A[self.current_parent]) > 1:
+                new_segs, nodes_to_remove = self.update_A_and_find_ibd_segs(ibd_segments)
                 
                 # e. Add any new IBD segments discovered.
                 for key, val in new_segs.items():
@@ -142,30 +179,30 @@ class IbdFinder(object):
                         else:
                             ibd_segments[key].append(v)
                             
-                # g. Remove elements of A[u] if they are no longer needed.
+                # g. Remove elements of self.A[u] if they are no longer needed.
                 ## (memory-pruning step)
                 for n in nodes_to_remove:
                     if self.current_parent in parent_list[n]:
                         parent_list[n].remove(self.current_parent)
                     if len(parent_list[n]) == 0:
-                        A[n] = []
+                        self.A[n] = []
                         
-            # Unlist the ancestral segments in A.
-            A[self.current_parent] = list(itertools.chain(*A[self.current_parent]))
+            # Unlist the ancestral segments in self.A.
+            self.A[self.current_parent] = list(itertools.chain(*self.A[self.current_parent]))
 
         # 4
         return ibd_segments
 
 
-    def update_A_and_find_ibd_segs(self, ancestral_segs, ibd_segments, mrca_ibd=False):
+    def update_A_and_find_ibd_segs(self, ibd_segments):
     
         new_segments = dict.fromkeys(itertools.combinations(self.ts.samples(), 2), [])
-        num_coalescing_sets = len(ancestral_segs)
+        num_coalescing_sets = len(self.A[self.current_parent])
         index_pairs = list(itertools.combinations(range(0, num_coalescing_sets), 2))
 
         for setpair in index_pairs:
-            for seg0 in ancestral_segs[setpair[0]]:
-                for seg1 in ancestral_segs[setpair[1]]:
+            for seg0 in self.A[self.current_parent][setpair[0]]:
+                for seg1 in self.A[self.current_parent][setpair[1]]:
 
                     if seg0.node == seg1.node:
                         continue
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Great, thanks @gtsambos, looking forward go going through this with you tomorrow.

python/tests/ibd.py
```diff
@@ -0,0 +1,348 @@
...
+                # Stop looking for IBD segments once the
+                # processed nodes are older than the max time.
+                break
+            
+            # 3b
+            while e is not None and self.current_parent == e.parent:
+                # Create the list S of immediate descendants of u.
+                S.append(Segment(e.left, e.right, e.child))
+                # if e.id < edges.num_rows - 1:
+                if e.id < self.ts.num_edges - 1:
```
jeromekelleher: Easier way to do this: ``e = next(mygen, None)``

python/tests/ibd.py
```diff
@@ -0,0 +1,348 @@
+# MIT License
+#
+# Copyright (c) 2018-2020 Tskit Developers
+# Copyright (c) 2015-2018 University of Oxford
```
jeromekelleher: This didn't exist until now, so change to
```
# Copyright (c) 2020 Tskit Developers
```

petrelharp:
This is great!! Let me know if I can help write some tests or something.

gtsambos:
Hi @jeromekelleher and @petrelharp (also cc @dvukcevic), this has come along a bit in the last few days after a few weeks of muddy-headedness! I have one or two relatively easy things I'll do over the next days -- cleaning and documenting the code a more, adding some other tests, making the CI tests work. Otherwise I think this should be ready to review quite soon, probably before next week.

> Start C-ifying the main algorithm. This is mainly a case of (a) changing the encoding of pairs (0, 0), (0, 1) etc to indexes 0, 1, ..., n**2 - 1; (b) changing the structure from creating temporary mappings that are moved between functions to global state that is modified in place by methods.

(b) was pretty easy to do; (a) has been much more complicated. I have created a new object, `SegmentList`, to hold linked lists of `Segment` objects. (Confusingly, most of the algorithm works with lists of SegmentLists -- ie. lists of lists of segments). For now, I have retained the representation of the output, `ibd_segments`, as a Python dictionary, with the keys being the sample pairs and the values being the list of IBD segments for each pair. However, in the C code this will be a struct array holding `SegmentList` objects. I've added some methods that mimic the procedure that will be needed to 'look up' the correct array element in C, even though this wasn't strictly necessary to make the Python code work. I'm happy to talk through this on a call or elsewhere.

> This is great!! Let me know if I can help write some tests or something.

Cool, thanks @petrelharp! I have a few others to add in myself; I'll let you know once I've done these so you can see if they look comprehensive enough.

jeromekelleher:
Sounds good @gtsambos, please do ping me if you want any input in the meantime.

petrelharp:
Me, too! Looking forward to it.

gtsambos:
ready to review ✨

gtsambos:
Oh yeah, I totally forgot that `discrete=True` won't work in the CI tests. Should I just comment these out until the new msprime release? I think that's the only reason why the CI is failing

benjeffery:
> Should I just comment these out until the new msprime release?

I think we can assume `discrete=True` will be in the next release so how about:
```python
if msprime._version.version >= '0.7.5':
    .... TEST
```

gtsambos:
okay will do, thanks @benjeffery!

jeromekelleher:
A good way to do this is to use ``@unittest.skipIf(msprime.__version__ <= 0.7.4)``

benjeffery:
Nice one! Wasn't aware of that decorator!

python/tests/test_ibd.py
```diff
@@ -0,0 +1,749 @@
...
+    Calculates IBD segments using both the 'naive' and sophisticated algorithms,
+    verifies that the same output is produced.
+    NB: May be good to expand this in the future so that many different combos
+    of IBD options are tested simultaneously (all the MRCA and path-IBD combos),
+    for example.
+    """
+    ts = treeSequence
+    ibd0 = ibd.IbdFinder(ts, samples=ts.samples())
+    ibd0 = ibd0.find_ibd_segments()
+    ibd1 = get_ibd_all_pairs(ts, path_ibd=True, mrca_ibd=True)
```
petrelharp, line 151: This would be a good place to test the `samples` argument, by also comparing output on a random subset of the samples.

petrelharp:
I'll have a read-through - unless someone else is in the middle of it already?

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+        self.ibd_segments[index] += seg
+        """
+        self.ibd_segments[self.sample_pairs[index]] + seg
+
+    def get_sample_id_map(self):
+        """
+        Returns id_map, a vector of length ts.num_nodes. For a node with id i,
+        id_map[i] is the position of the node in self.samples. If i is not a
+        user-specified sample, id_map[i] is -1.
+        Note: it assumes nodes in the TS are numbered 0 to ts.num_nodes - 1
```
petrelharp, line 195: This assumption is always be true.

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+                self.parent_list[e.child].append(e.parent)
+        # Objects below are needed for the IBD segment-holding object.
+        self.sample_id_map = self.get_sample_id_map()
+        self.num_samples = len(self.samples)
+        self.sample_pairs = self.get_sample_pairs()
+        """
+        Note: in the C code the object below should be a struct array.
+        Each item will be accessed using its index, which corresponds to a particular
+        sample pair. The mapping between index and sample pair is defined in the
+        find_sample_pair_index method further down.
```
petrelharp, line 174: Hm, this is in `N^2`-land, where I don't think we want to be? Maybe it'll be the case that most pairs have a segment, but also maybe not, and for big datasets this will be bad. So, in python you don't need to initialize all `N^2` empty dicts (but have to check below if the pair is already in `self.ibd_segemtns`). In C you can use an AVL tree of segment lists sorted on the pair of IDs to do the same thing.

python/tests/test_ibd.py
```diff
@@ -0,0 +1,749 @@
...
+
+        node_id = n.id
+        interval_list = []
+        if n.flags == 1:
+            continue
+
+        prev_dict = None
+        for t in ts.trees():
+
+            if len(list(t.nodes(n.id))) == 1 or t.num_samples(n.id) < 2:
```
petrelharp, line 53: Just
```
             if t.num_samples(n.id) < 2:
```
should be equivalent, no?

python/tests/test_ibd.py
```diff
@@ -0,0 +1,749 @@
...
+                    t, prev_dict, node_id
+                ):
+                    interval_list[-1] = (prev_int[0], current_int[1])
+                else:
+                    interval_list.append(current_int)
+
+            prev_dict = t.get_parent_dict()
+
+        for interval in interval_list:
+            if min_length == 0 or interval[1] - interval[0] > min_length:
```
petrelharp, line 75: I don't think you need the first condition here, either?

petrelharp:
Looking through... looking good... I'll say when I'm done. In the meantime, a comment: I see that you're doing the thing that I did on my last PR, which is to make the python implementation as much like the C implementation as possible, including building linked lists explicitly with pointers to the next one, etcetera.  BUT THEN, @jeromekelleher told me to be more pythonic, and leave that sort of thing up to python, thus leaving the basic structure of the algorithm more clear without having to worry about who's linked to who, etcetera. For instance, your `SegmentList` could probably just be a list, without defining a special class?

I think he's right - it helps to separate the logic  of the implementation from the fiddly bookkeeping details. But, on the other hand, I know for me it was helpful to do all this stuff in python before trying to do it in C. So, my suggestion is to at some point simplify these data structures to use more python - maybe now, but up to you.

gtsambos:
Thanks for looking at this so quickly @petrelharp! I had a call with @jeromekelleher  last night and we went over this together then as well. I'll comment on some of the specific things up above. About this:

> I see that you're doing the thing that I did on my last PR, which is to make the python implementation as much like the C implementation as possible

> For instance, your SegmentList could probably just be a list, without defining a special class?

I was trying to make this pretty C-like, yes. I think I'll keep my SegmentLists as they are, because I'm using them in a pretty involved way. In this IBD algorithm, we need to be able to add lists together/do other operations on them, rather than just cycling through elements of the list as we do in `simplify` and `link_to_ancestors`. Also, the output of the C-based algorithm will have to be an array of SegmentList objects (each one listing the IBD ancestors/segments for a particular sample pair), so it makes sense to me that we'd have also it in Python in order to compare the output between the two implementations. There are quite a few methods I'll have to define in C to do these things, to the point where I thought it would look a bit unrecognisable if I omitted them here.

Are there any things in this code other than the `SegmentList` objects that could be simplified, though?

python/tests/ibd.py
gtsambos: There's just a single dict, but yes, the number of keys in the dict is n choose 2 where the user has inputted n samples. In the call with @jeromekelleher last night we discussed initialising them with `None` instead of with empty SegmentLists, which will hopefully solve some of this big memory problem?

I think it makes sense to keep all sample pairs in the outputted dict, even if there are no IBD segments, because we'll have to do that in C as well. In C the output will be stored as a struct array, where we access the list of IBD segments corresponding to a particular pair using a predefined lexicographic ordering. It will be confusing and difficult to define an ordering based only on the sample pairs that happen to have non-empty IBD output, as that will differ from run to run, and it won't be known until the algorithm has been applied in any case.

petrelharp:
>  we need to be able to add lists together/do other operations on them, rather than just cycling through elements of the list 

Ah, ok. I haven't gotten there yet.

> Are there any things in this code other than the SegmentList objects that could be simplified, though?

I'm getting there! Had some important mucking about in the yard to do first, though.

gtsambos:
btw, some other things we talked about were

in `ibd.py`:

 - changing this [add method](https://github.com/tskit-dev/tskit/pull/488/files#diff-b380292c5148ec643106ade18321cbeaR110-R134) to the more explicit `SegmentList().add(...)` instead of using `+`
 - changing `SegL_head` and `SegL_tail` to be local variables inside the `find_ibd_segments` method
 - initialising `ibd_segments` with `None` rather than with `SegmentList()`

in `test_ibd.py`
 - Simplifying the 'naive' [get_ibd method](https://github.com/tskit-dev/tskit/pull/488/files#diff-82aa2791b41e10f6d5a3b3cdbeed01e2R17) to only test MRCA ibd (and so making the function clearer to follow)
- replacing `assert` statements with `self.assert` statements (look into defining a subclass of `unittest.TestCase` specific to these IBD tests)
 - Generating some weird random tree sequences with the forward-time DTWF simulator

python/tests/ibd.py
petrelharp: Well, we'll see, but I don't think you want to N^2 anything, including the size of your struct array. That's why I'm suggesting the AVL tree, which we have a predefined implementation of (see `avl.h`) and lets you do log(N) insertions and lookups easily: you don't have to worry about the ordering. It'll be easy enough to switch in later, I suppose, so maybe you should go forward with the current plan.

petrelharp:
Those sound good. Maybe I should wait 'til you've done those things to finish the careful read-through?

gtsambos:
sure :) hopefully will be done this week! I'd be really appreciative if you could look at the tests though, I'm not sure I put in as many wacky topologies as I could have

petrelharp:
The wacky topologies looked sufficient - my suggestion is the same as Jerome's - to do some N=10 discrete WF simulations to get more wacky topologies.

petrelharp:
Your existing `get_ibd` function is very nice - good use of `keep_unary`.  So, the idea is to rename it `path_ibd` and remove the `path_ibd` and `mrca_ibd` options? That makes sense. And, the `path_ibd=False` case will be even easier, since I think it's basically the non-sample nodes in `tree_sequence.simplify([sample0, sample1], keep_unary=False)`?

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+        Calculates all new IBD segments found at the current iteration of the
+        algorithm. Returns information about nodes processed in this step, which
+        allows memory to be cleared as the procedure runs.
+        """
+
+        list0 = self.SegL_head
+        # Iterate through all pairs of segment lists.
+        while list0.next is not None:
+            list1 = list0.next
+            while list1 is not None:
```
petrelharp, line 359: This looping over all pairs of lists is `N^2`, isn't it? Have you thought about how to avoid that?

gtsambos: Thankfully not I think (I hope!)! In a 'normal' tree seq there will actually be only one iteration of this loop -- there will only be 'iterations' when the currently processed node is polytomous (?). There is a separate list for each newly coalescing set of segments underneath the node. Say for example you're processing a node `n` and it is parent to 3 edges in the tree sequence. Then there will be 3 lists in the list of lists, L1 L2 and L3 say, and this loop will iterate over (L1, L2), (L1, L3), (L2, L3).

gtsambos:
also @petrelharp, let me know if you'd like to call to discuss any of this (if that seems like it would be an easier way to communicate)

python/tests/ibd.py
gtsambos: I have to say though, I do find it tricky to think about exactly *how* some of these procedures scale, though I can see that they are less than N^2. It would be good to spend more time pondering this so that in the manuscript we can give an explicit scaling factor. 

The  'loopiest' part of this algorithm is the looping over segments that happens inside this list loop you've highlighted here. At worst, if you were at the root of the trees, you could be comparing coalescing sets that each have n/2 segments in there, so a scaling factor of n^2/4. However, I don't think there's any way you could calculate IBD so far up a tree without doing that. 

A more sensible (though perhaps complicated?) way to assess how this scales would be to derive a scaling factor wrt IBD found at some depth in the tree.

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+                    S_head = seg
+                    S_tail = seg
+                else:
+                    S_tail.next = seg
+                    S_tail = S_tail.next
+                if e.id < self.ts.num_edges - 1:
+                    e = next(mygen)
+                    continue
+                else:
+                    e = None
```
petrelharp, line 298: A simpler loop is, I think:
```
for e in mygen:
   if e.parent != self.current_parent:
      break
   if S_head is None:
      ...
```

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+                    list_to_add + Segment(S_head.left, S_head.right, u)
+                else:
+                    s = self.A[u].head
+                    while s is not None:
+                        intvl = (max(S_head.left, s.left), min(S_head.right, s.right))
+                        if intvl[1] - intvl[0] > 0:
+                            list_to_add + Segment(intvl[0], intvl[1], s.node)
+                        s = s.next
+                self.add_to_chain(list_to_add)
+                S_head = S_head.next
```
petrelharp, line 316: Here I think in the first loop you iterate over edges, creating `S`. Then in the second loop you process these one at a time, and `S` isn't used again. That means (unless I'm' missing something) that you don't need two loops, even `S` at all?

python/tests/ibd.py
petrelharp: > At worst, if you were at the root of the trees, you could be comparing coalescing sets that each have n/2 segments in there, so a scaling factor of n^2/4

Right, good point - if we want pairwise IBD then we will output O(N^2) things in this case, no getting around it.

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+            list1 = list0.next
+            while list1 is not None:
+                seg0 = list0.head
+                # Iterate through all segments with one taken from each list.
+                while seg0 is not None:
+                    seg1 = list1.head
+                    while seg1 is not None:
+                        if seg0.node == seg1.node:
+                            seg1 = seg1.next
+                            continue
```
petrelharp, line 367: this isn't needed, as any two segments with the same node shouldn't overlap, so will `continue` five lines down anyhow?

python/tests/ibd.py
```diff
@@ -0,0 +1,485 @@
...
+                    while s is not None:
+                        intvl = (max(S_head.left, s.left), min(S_head.right, s.right))
+                        if intvl[1] - intvl[0] > 0:
+                            list_to_add + Segment(intvl[0], intvl[1], s.node)
+                        s = s.next
+                self.add_to_chain(list_to_add)
+                S_head = S_head.next
+
+            # If we wanted to squash segments, we'd do it here.
+
```
petrelharp, line 319: Gee, if you added a `[0, sequence_length)` segment to `segL` if `current_parent` is in `samples`, I think that'd simplify things, like you wouldn't have to treat the case separately [here](https://github.com/tskit-dev/tskit/pull/488/files#diff-b380292c5148ec643106ade18321cbeaR391)?

petrelharp:
Ok, very nice! Now that I've got a chance to sit down with it, I get it.

One suggestion to simplify things: currently, you comparing each pair of lists of lists to check for coalescences and at the end flatten them into `A[u]`. Instead, I think you could instead add each list of lists to `A[u]`, checking for coalescence with those already in `A[u]`? This might in fact avoid the need for `SegL` at all, since you could do something like this:
```
       def get_subsegments(self, u, left, right):
                list_to_add = SegmentList()
                if u in self.samples:
                    list_to_add + Segment(left, right, u)
                else:
                    s = self.A[u].head
                    while s is not None:
                        intvl = (max(left, s.left), min(right, s.right))
                        if intvl[1] - intvl[0] > 0:
                            list_to_add + Segment(intvl[0], intvl[1], s.node)
                        s = s.next
                return list_to_add
```
and then in `find_ibd_segments( )`:
```
            while e is not None and self.current_parent == e.parent:

                list_to_add = self.get_subsegments(e.child, e.left, e.right)
                self.calculate_ibd(self.A[e.child].head, list_to_add)
                self.A[e.parent] + list_to_add

                if e.id < self.ts.num_edges - 1:
                    e = next(mygen)
                    continue
                else:
                    e = None
```
... and then `calculate_ibd` looks for overlap between the two lists of segments it gets. Oh, I see a problem already - this doesn't allow for squashing of edges, which you're leaving space for. But maybe it'll give you an idea?

Also: one suggestion on naming: `add_to_chain` is something that manipulates `SegL`; but the names don't evoke this: maybe rename `SegL` (perhaps `current_chain`?) and name `add_to_chain` something to do with this (`add_to_current_chain`?). I think some of the other names could be made more consistent also.

petrelharp:
Let me know if the above is too vague to be useful - I could sketch it out in more detail (and probably find the wrong things while doing so).

gtsambos:
Thanks @petrelharp, and sorry I haven't answered your email yet either! These look like good suggestions -- a few other things have come up this week, I will think about these carefully over the next few days :)

python/tests/ibd.py
gtsambos: yup, good spot!

python/tests/ibd.py
gtsambos: Hmmm, yes, after thinking a bit I agree with you on this -- good thinking! It's not in the latest commit but it will be in the next one.

gtsambos:
> ... and then calculate_ibd looks for overlap between the two lists of segments it gets. Oh, I see a problem already - this doesn't allow for squashing of edges, which you're leaving space for. But maybe it'll give you an idea?

Yes, this was my concern -- squashing requires you to change the order of the segments, and basically requires you to do two whole traversals of all the segments in `SegL` -- once to squash, and then again to calculate IBD. I figure that the 'list of list' approach might make it easier to see which segments need to be compared with which after the first traversal, but I'll think a bit more about whether we can do without it

> Also: one suggestion on naming: add_to_chain is something that manipulates SegL; but the names don't evoke this: maybe rename SegL (perhaps current_chain?) and name add_to_chain something to do with this (add_to_current_chain?). I think some of the other names could be made more consistent also.

Yes, I'm happy to change these names if it makes everything a bit clearer (and if we end up deciding that `SegL` is necessary in the first place -- see above). Tbh I was changing these all the time while I was writing the code anyway!

petrelharp:
> Tbh I was changing these all the time while I was writing the code anyway!

I know how that goes!

Looks like you've got a bunch of new stuff in - let me know when I should have another look!

petrelharp:
Distracting question (sorry): currently, segments contain `[left, right)` and the identity of the sample node that inherits from that segment; have you thought about giving them a list (a set, really) of sample nodes that inherit from the segment?  This would dramatically reduce the number of segments you need to keep track of, but I'm not clear if it reduces code complexity or the number of operations substantially.

gtsambos:
> currently, segments contain [left, right) and the identity of the sample node that inherits from that segment; have you thought about giving them a list (a set, really) of sample nodes that inherit from the segment?

Hmm, I think this could make things quite a bit more complicated. It's important to keep in mind that even though lots of the same sample nodes might descend from the same ancestral node on lots of consecutive trees, the genomic interval that is inherited will often be different for the various samples, because each of their histories has involved slightly different recombination events. The number of 'node lists with the same ancestral interval' would be smaller than the number of segments, but it's not clear to me that it would be smaller by a non-scalar factor. It would defs increase the algorithm complexity somewhat, because we'd need to check for IBD relationships twice, in slightly different ways depending on whether we're comparing segments that are represented in the same node set, or in different ones. Also, any extension requiring squashing of intervals would be mindbogglingly complicated, since squashing relies on a pre-set ordering of the segments to be squashed, and we'd need to pull segments out of existing sets and into new ones...

Perhaps I'm just talking myself out of it at this point, but I think it might be better to retain individual segments for now!

> let me know when I should have another look!

will do!

petrelharp:
ah yes - makes sense. darn.

gtsambos:
> Instead, I think you could instead add each list of lists to A[u], checking for coalescence with those already in A[u]? This might in fact avoid the need for SegL at all, since you could do something like this:

After thinking about it a bit more, I think something like this *could* work even in the more complicated setting where there is squashing, but we'll need to have an extra method to remove a segment from an existing segment list and patch the list up. It would look something like this:

 - Loop through segments `s` in `S`
    - For each segment `s`, get the list of all subsegments, `SubS`
    - For each subsegment `sub` in `SubS`, do pairwise comparisons with each segment `a` in A[u].
       - If `a` and `sub` are squashable, remove `a` from `A[u]` and `sub` from `SubS`, and then add the squashed segment to the end of `S`.
       - Otherwise, if the intersection of `a` and `sub` introduces a new IBD relationship, record it. Add `sub` into `A`.

I'll play around with this over the weekend

petrelharp:
If you think it might be simpler, go for it!

gtsambos:
Hi @jeromekelleher and @petrelharp, the latest commit(s) represent a major clean-up/change to the algorithm based on Peter's comments from a few weeks ago. Namely, these ones:

> Instead, I think you could instead add each list of lists to A[u], checking for coalescence with those already in A[u]? This might in fact avoid the need for SegL at all, since you could do something like this:

> Here I think in the first loop you iterate over edges, creating S. Then in the second loop you process these one at a time, and S isn't used again. That means (unless I'm' missing something) that you don't need two loops, even S at all?

I am yet to add tests using the forward-time DTWF simulator, but once they are done and I've fixed any issues arising from them (hopefully not many!), I think I should start the C implementation 👨‍💻

gtsambos:
Hmm, would it okay to add the `packaging` module to `development.txt`? It looks smallish, and I think I need this do safely do comparisons of versions like we discussed [here](https://github.com/tskit-dev/tskit/pull/488#issuecomment-619913899). This is why the CI is currently failing btw.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+    def __init__(self, ts, samples=None, min_length=0, max_time=None):
+
+        self.ts = ts
+        # Note: samples *must* be in order of ascending node ID
+        if samples is None:
+            self.samples = ts.samples()
+        else:
+            self.samples = samples
+        self.samples.sort()
```
jeromekelleher, line 146: I don't think we should do this. The ordering of the samples is defined by the user and it would be unexpected to change this. Instead, we do something like
```python
self.sample_index_map = np.zeros(ts.num_nodes, dtype=int) - 1
for index, u in enumerate(self.samples):
    self.sample_index_map[u] = index
```
Then, when we want to know if something is a sample we see if ``self.sample_index_map[u] != tskit.NULL``.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+        # self.ibd_segments[index].add(seg)
+
+        if self.ibd_segments[self.sample_pairs[index]] is None:
+            self.ibd_segments[self.sample_pairs[index]] = SegmentList(
+                head=seg, tail=seg
+            )
+        else:
+            self.ibd_segments[self.sample_pairs[index]].add(seg)
+
+    def get_sample_id_map(self):
```
jeromekelleher, line 190: We don't need this function now I think, as we're storing the ``sample_index_map`` in the constructor (and this will reflect how we do it in C as well)

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+            if (
+                len(self.parent_list[e.child]) == 0
+                or e.parent != self.parent_list[e.child][-1]
+            ):
+                self.parent_list[e.child].append(e.parent)
+        # Objects below are needed for the IBD segment-holding object.
+        self.sample_id_map = self.get_sample_id_map()
+        self.num_samples = len(self.samples)
+        self.sample_pairs = self.get_sample_pairs()
+        self.PARENT_SHOULD_BE_ADDED = True
```
jeromekelleher, line 164: This and ``self.current_parent`` are only used in the ``find_ibd_segments`` function, so we don't need them to be instance variables.

jeromekelleher: Looks like the same is true for ``current_time``.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+            (self.num_samples) * (self.num_samples - 1) / 2
+            - (self.num_samples - i0) * (self.num_samples - i0 - 1) / 2
+            + i1
+            - i0
+            - 1
+        )
+
+        return int(index)
+
+    def find_ibd_segments(self, min_length=0, max_time=None):
```
jeromekelleher, line 244: I don't think these parameters are used now.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        The wrapper for the procedure that calculates IBD segments.
+        """
+
+        # Set up an iterator over the edges in the tree sequence.
+        mygen = iter(self.ts.edges())
+        e = next(mygen)
```
jeromekelleher, line 251:
```suggestion
        e = next(mygen, None)
```

jeromekelleher: ``mygen`` isn't particularly descriptive. Something like ``edges_iter`` would be better.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+        nodes_to_process = []
+
+        # Iterate over the edges.
+        while e is not None:
+            # Process all edges with the same parent node.
+            self.current_parent = e.parent
+            self.current_time = self.tables.nodes.time[self.current_parent]
```
jeromekelleher, line 258: Better to have a ``node_time`` variable which you declare outside the loop. Calling `` self.tables.nodes.time`` is expensive.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+            if self.max_time is not None and self.current_time > self.max_time:
```
jeromekelleher, line 259: A nice way to simplify this is to set ``self.max_time`` to double max at input. Then you don't have to check if it's none (and you'll need to do this for C anyway, as there's no ``None`` concept there).

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+                            nodes[1],
+                            Segment(left, right, self.current_parent),
+                        )
+                    seg1 = seg1.next
+                seg0 = seg0.next
+
+            # Add list_to_add to A[u].
+            self.A[current_parent].add(list_to_add)
+
+        # Calculate nodes processed in this step.
```
jeromekelleher, line 346: I'm not sure there's much point in returning this list, and it would be really tedious to do in C. All we're doing is returning a list consisting of the nodes in the input segment list, right?

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+                )
+                pn = self.calculate_ibd_segs(self.current_parent, singleton_seg)
+                self.PARENT_SHOULD_BE_ADDED = False
+                nodes_to_process += pn
+
+            # Move to next edge.
+            e = next(mygen, None)
+
+            # Remove any processed nodes that are no longer needed.
+            if e is None or e.parent != self.current_parent:
```
jeromekelleher, line 300: OK, so this is done each time we move on to a new parent. But what's it actually doing? It seems to me that we're only manipulating the ``self.parent_list`` array in this loop, so is it actually needed now?

jeromekelleher: The end effect is to set ``self.A[n]`` to the empty list, so is there some other way we could achieve this?

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+
+    def __lt__(self, other):
+        return (self.node, self.left, self.right) < (
+            other.node,
+            other.left,
+            other.right,
+        )
+
+
+class SegmentList:
```
jeromekelleher, line 70: Do we need this SegmentList structure now? It looks to me like we're only using linked lists of Segments now, so it would simplify things a lot if we didn't need this extra structure. Everything we can do to make things simpler and clearer here will knock multiple hours of effort and hundreds of lines off the C implementation!

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+        self.sample_pairs = self.get_sample_pairs()
+        self.PARENT_SHOULD_BE_ADDED = True
+
+        # Note: in the C code the object below should be a struct array.
+        # Each item will be accessed using its index, which corresponds to a particular
+        # sample pair. The mapping between index and sample pair is defined in the
+        # find_sample_pair_index method further down.
+
+        self.ibd_segments = {}
+        for key in self.sample_pairs:
```
jeromekelleher, line 172: I don't think there's much point in storing None for all the sample_pairs, right? We can just check to see if the key is already in the array or not.

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+        index = self.find_sample_pair_index(sample0, sample1)
+
+        # Note: the code below is specific to the Python implementation, where the
+        # output is a dictionary indexed by sample pairs.
+        # In the C implementation, it'll be more like
+        # self.ibd_segments[index].add(seg)
+
+        if self.ibd_segments[self.sample_pairs[index]] is None:
+            self.ibd_segments[self.sample_pairs[index]] = SegmentList(
+                head=seg, tail=seg
```
jeromekelleher, line 185:
```suggestion
        index = self.find_sample_pair_index(sample0, sample1)
        key = self.sample_pairs[index]
        if key not in self.ibd_segments[key]:
            self.ibd_segments[key] = SegmentList(head=seg, tail=seg)
```

Although, it would probably be better to map directly to the indexes, so that it'll be easier to compare with the C code (it's easy enough to convert the index to an actual tuple when printing out debug stuff.)

python/tests/test_ibd.py
```diff
@@ -0,0 +1,821 @@
...
+        verify_equal_ibd(ts)
+
+    # Finite sites
+    def sim_finite_sites(self, random_seed, dtwf=False):
+        seq_length = int(1e5)
+        positions = random.sample(range(1, seq_length), 98) + [0, seq_length]
+        positions.sort()
+        rates = [random.uniform(1e-9, 1e-7) for _ in range(100)]
+        r_map = msprime.RecombinationMap(
+            positions=positions, rates=rates, discrete=True
```
jeromekelleher, line 722: It would be simpler to use the shipping APIs for msprime than to do all this skipping (plus, we don't want to skip the tests - otherwise how do we know they are doing anything?). Use ``num_loci=seq_length`` to achieve the same result here. It is *definitely* possible to have discrete recombination breakpoints with msprime < 1.0

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+                # Stop looking for IBD segments once the
+                # processed nodes are older than the max time.
+                break
+
+            seg = Segment(e.left, e.right, e.child)
+
+            # Create a SegmentList() holding all segments that descend from seg.
+            list_to_add = SegmentList()
+            u = seg.node
+            if u in self.samples:
```
petrelharp, line 269: I'm confused here: why isn't `u` always `None`, since `node` wasn't given in the call to `Segment( )` above?

petrelharp: did you mean to have `node=e.parent`?

petrelharp: oh wait I'm being dumb sorry

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+    def calculate_ibd_segs(self, current_parent, list_to_add):
+        """
+        Write later.
+        """
+
+        if self.A[current_parent] is None:
+            self.A[current_parent] = list_to_add
+
+        else:
+            seg0 = self.A[current_parent].head
```
petrelharp, line 320: This is essentially doing
```
for seg0 in self.A[current_parent]:
  for seg1 in list_to_add:
     ...
```
right? it'd be a bit more clear written that way...

python/tests/ibd.py
```diff
@@ -0,0 +1,421 @@
...
+            for j in self.samples[(ind + 1) :]:
+                sample_pairs.append((i, j))
+
+        return sample_pairs
+
+    def find_sample_pair_index(self, sample0, sample1):
+        """
+        Note: this method isn't strictly necessary for the Python implementation
+        but is needed for the C implemention, where the output ibd_segments is a
+        struct array.
```
petrelharp, line 219: I'm still worried about the N^2 here. It might be easier all around, and avoid N^2, to have the output be a dictionary indexed by a sorted tuple of the sample IDs; and in C we could use an avl_tree, which does more or less the same thing.

petrelharp:
It looks a lot simpler, nice! I've got nothing else to add. I'm curious, though: where does this leave us for squashing edges - will require a refactor?

python/tests/ibd.py
jeromekelleher: We can't use an AVL tree here unfortunately @petrelharp  - the one we're using in msprime is GPL. So, I was thinking that the linear map would be simpler for now. I agree that given the list is likely to be sparse this N^2 implementation will be wasteful of memory. I doubt the time required to go through the output to find the non-null entries will end up being significant.

I agree that ultimately we will need to have some sort of sparse representation for the output, which will be keyed by the sorted pair (or equivalently, the index computed here). We will have to make some technology choices about which AVL tree/hash table implementation to use though, which I don't want to get into here.

gtsambos:
Hi Jerome, about this:

>  I'm not convinced the SegmentList structure is really needed now, for example. 

I think we could do without it (and am trying to mod it like this now), but getting rid of it comes at the cost of adding a couple of new `add_segment_to_list` methods that are all basically doing the same thing

gtsambos:
Hi @jeromekelleher, I'm just now thinking more deeply about if we were to get rid of SegmentLists. One issue is that the current output of this algorithm is currently SegmentLists (and it will have to be something like this, as this is basically what raw IBD information is). 

Updating these lists as the algorithm iterates isn't going to be quite as simple as before. Without having segment lists, we have two options:

 1. For each pair of samples, keep track of both the head and the tail segment. Each time a new IBD segment is found, add it to the tail. We'll need to make a struct array for the heads (indexed by the sample pair), and another for the tails (also indexed by the sample pair), and will have to update these array elements each time a new segment is found.

2. For each pair of IBD segments, we keep track of only the head segment. Everytime we want to add a new ibd segment, we will have to cycle through all of the segments starting from the first one. 

(1) seems a little more ideal to me because of its efficiency?

jeromekelleher:
OK, great, let's keep the SegmentLists then @gtsambos - just wanted to make sure we still needed them.

python/tests/ibd.py
petrelharp: Ok - that'll be easy enough to change, but I think we'll have to do it before applying this to the big datasets.

gtsambos:
Hi @petrelharp @jeromekelleher , sorry for not responding to a lot of this -- I've been very busy with other time-critical projects but have set aside this afternoon to go through these!

> I'm curious, though: where does this leave us for squashing edges - will require a refactor?

I think that as long as we keep SegmentLists, the 'refactoring' will mostly involve adding new code rather than modifying this existing code base -- most immediately, I can see that we'd need a method for removing a segment from a SegmentList and patching it up (as opposed to just adding them in, as we need here), and we'd also need a SegmentList-type object or set of objects to hold the edges once they have been squashed..

I haven't thought much about the AVL/hash table stuff, and to be honest I feel a little out of my depth -- I think I'll need to read and learn more about these before I'm able to have a proper conversation about it.

python/tests/ibd.py
gtsambos: See comment below for why I have retained SegmentLists.

python/tests/ibd.py
gtsambos: So, this list does serve a purpose here -- it's how we figure out which nodes' parents have been completely processed at a given step of the algorithm, and so which segments can be cleared from memory.  I'll think about whether there are better ways of doing this.

python/tests/ibd.py
gtsambos: Let's talk about this in person --

python/tests/ibd.py
gtsambos: Let's talk about this

python/tests/ibd.py
gtsambos: The reason why I haven't done this is that `self.A[current_parent]` and `list_to_add` are `SegmentList` objects rather than actual lists. So you can't cycle through them using `for`, you have to explicitly use the `next` attribute to get each segment in the sequence

gtsambos:
Hmmm, this build looks like it's failing for some reason unrelated to this pull request (something wrong with the `pysam` library?)

jeromekelleher:
Yep, Pysam thing is unrelated @gtsambos - see  #673

python/tests/ibd.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def find_ibd_segments(self):
+        """
+        The wrapper for the procedure that calculates IBD segments.
+        """
+
+        # Set up an iterator over the edges in the tree sequence.
+        edges_iter = iter(self.ts.edges())
+        e = next(edges_iter)
+        PARENT_SHOULD_BE_ADDED = True
```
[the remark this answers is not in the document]
gtsambos: Oh okay, I've changed this in the latest commit. I think I'm used to seeing boolean variables being capitalised, but it's no big deal to me either way

gtsambos:
> One minor change here @gtsambos, and we can merge then I think. As discussed, let's follow up with with an initial C implementation, not worrying too much about optimum data structures.

Great! I'll probably need to do some squashing and rebasing as well, but only once you're happy with the code itself.

python/tests/ibd.py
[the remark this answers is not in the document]
jeromekelleher: ALL_CAPS is usually reserved for constants, I think.

jeromekelleher:
OK, great! Can you rebase and squash please? Don't worry about the travis builds, we'll merge once they've been sorted out.

gtsambos:
Cool -- done

gtsambos:
great, thanks @benjeffery!

jeromekelleher:
Great, thanks @gtsambos. Merging this much, so we can move on with the next phase of C implementation.

PR #489, opened 2020-03-18
Add support for missing data in LS match haplotype.

jeromekelleher:
Seems like the simplest way to implement missingness in the LS copying process is to assume that missing data is equal to everything. This seems cleaner than just skipping over a site if it has h=-1, because it takes into account the relative recombination rates properly. Also, it's very easy to express in terms of the actual likelihood, forward matrix etc, whereas skipping a site would require special cases in the maths.

@hyanwong, any thoughts here? The plan is to port this in to tsinfer immediately.

### DIFF
--- a/c/tests/test_haplotype_matching.c
+++ b/c/tests/test_haplotype_matching.c
@@ -136,7 +136,7 @@ test_single_tree_exact_match(void)
     CU_ASSERT_EQUAL(path[2], 1);
 
     /* Should get the same answer at lower precision */
-    for (precision = 4; precision < 24; precision++) {
+    for (precision = 1; precision < 24; precision++) {
         ret = tsk_ls_hmm_set_precision(&ls_hmm, precision);
         CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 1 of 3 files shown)


hyanwong:
Yes, I hadn't thought of this idea, but I like it. As you say, it seems much cleaner. I can't see any immediate objections. In fact, it is more flexible because then we could (in theory) incorporate degrees of uncertainty/missingness ("we know it's likely to be a purine, but not sure which").

Only issue is how it interacts with the specified matching `mutation_rate` (or whatever we are calling it), especially if we have e.g. a GTR model of mutation, rather than a single param. But that's way down the line.

jeromekelleher:
> Yes, I hadn't thought of this idea, but I like it. As you say, it seems much cleaner. I can't see any immediate objections. In fact, it is more flexible because then we could (in theory) incorporate degrees of uncertainty/missingness ("we know it's likely to be a purine, but not sure which").

Oh, great idea. Yes, we have plenty space in < -1, can get all sorts of stuff in there.

PR #490, opened 2020-03-18
Make KC distance linear in size of KC vectors

daniel-goldstein:
Computing the KC distance requires building two vectors per tree, `m` and `M`, where each element records, for some leaves u & v, the distance of mrca(u, v) to the root in number of edges and time, respectively. The existing implementation was essentially the above explanation, computing the mrca for each pair, which for `n` leaves, takes `O(n^2 log(n))` time.

This PR implements a slightly different approach, where we do a preorder traversal of the tree, accumulating the depth and time from the root. For each internal node, we can take the Cartesian product of the node's children's leaves and assign each pair the current depth and time. We use `sample_lists` to iterate through the leaves without any extra traversal overhead, erroring out when `sample_lists` is not available. Since each leaf pair is considered once, the overall runtime is `O(n^2)`. This approach is similar to that used in the official R package for KC-distance, tree space.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -4282,146 +4282,6 @@ out:
     return ret;
 }
 
-int
-tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double *result)
-{
-    struct stack_elmt {
-        tsk_id_t node;
-        int path_depth;
-        double time_depth;
-    };
-
-    tsk_size_t num_nodes_self, num_nodes_other;
-    int ret = 0;
-    int stack_top = 0;
-    int path_depth, tree_index, pair_index, i;
-    double vT1, vT2, distance_sum, time_depth, root_time;
-    int *m[2], *path_distance[2];
-    double *M[2], *time_distance[2];
-    tsk_id_t N, u, v, mrca, n1, n2, num_samples, u_index;
-    tsk_tree_t *trees[2] = { self, other };
-    tsk_tree_t *tree;
-    const tsk_id_t *samples = self->tree_sequence->samples;
-    const tsk_id_t *other_samples = other->tree_sequence->samples;
-    const double *times;
-    const tsk_id_t *sample_index_map;
-    struct stack_elmt *stack = NULL;
-
-    memset(path_distance, 0, sizeof(path_distance));
-    memset(time_distance, 0, sizeof(time_distance));
-    memset(m, 0, sizeof(m));
-    memset(M, 0, sizeof(M));
-
-    if (tsk_tree_get_num_roots(self) != 1 || tsk_tree_get_num_roots(other) != 1) {
-        ret = TSK_ERR_MULTIPLE_ROOTS;
-        goto out;
-    }
-    if (self->tree_sequence->num_samples != other->tree_sequence->num_samples) {
-        ret = TSK_ERR_SAMPLE_SIZE_MISMATCH;
-        goto out;
-    }
-
-    num_samples = (tsk_id_t) self->tree_sequence->num_samples;
-    N = (num_samples * (num_samples - 1)) / 2;
-    num_nodes_self = self->num_nodes;
-    num_nodes_other = other->num_nodes;
-    stack = malloc(TSK_MAX(num_nodes_self, num_nodes_other) * sizeof(*stack));
-    m[0] = calloc((size_t)(N + num_samples), sizeof(m[0]));
-    m[1] = calloc((size_t)(N + num_samples), sizeof(m[1]));
-    M[0] = malloc((size_t)(N + num_samples) * sizeof(M[0]));
-    M[1] = malloc((size_t)(N + num_samples) * sizeof(M[1]));
-    path_distance[0] = malloc(num_nodes_self * sizeof(path_distance[0]));
-    path_distance[1] = malloc(num_nodes_other * sizeof(path_distance[1]));
-    time_distance[0] = malloc(num_nodes_self * sizeof(time_distance[0]));
-    time_distance[1] = malloc(num_nodes_other * sizeof(time_distance[1]));
-    if (stack == NULL || m[0] == NULL || m[1] == NULL || M[0] == NULL || M[1] == NULL
-        || path_distance[0] == NULL || time_distance[1] == NULL) {
-        ret = TSK_ERR_NO_MEMORY;
-        goto out;
-    }
-
-    for (i = 0; i < num_samples; i++) {
-        if (samples[i] != other_samples[i]) {
-            ret = TSK_ERR_SAMPLES_NOT_EQUAL;
-            goto out;
-        }
-        u = samples[i];
-        if (self->left_child[u] != TSK_NULL || other->left_child[u] != TSK_NULL) {
-            /* It's probably possible to support this, but it's too awkward
-             * to deal with and seems like a fairly niche requirement. */
-            ret = TSK_ERR_INTERNAL_SAMPLES;
-            goto out;
-        }
-    }
-
-    for (i = 0; i <= N + num_samples; i++) {
-        m[0][i] = 1;
-        m[1][i] = 1;
-    }
-
-    for (tree_index = 0; tree_index < 2; tree_index++) {
-        tree = trees[tree_index];
-        times = tree->tree_sequence->tables->nodes.time;
-        sample_index_map = tree->tree_sequence->sample_index_map;
-        stack_top = 0;
-        u = tree->left_root;
-        root_time = times[u];
-        stack[stack_top].node = u;
-        stack[stack_top].path_depth = 0;
-        stack[stack_top].time_depth = root_time;
-        while (stack_top >= 0) {
-            u = stack[stack_top].node;
-            path_depth = stack[stack_top].path_depth;
-            time_depth = stack[stack_top].time_depth;
-            stack_top--;
-            for (v = tree->left_child[u]; v != TSK_NULL; v = tree->right_sib[v]) {
-                stack_top++;
-                stack[stack_top].node = v;
-                stack[stack_top].path_depth = path_depth + 1;
-                stack[stack_top].time_depth = times[v];
-            }
-            path_distance[tree_index][u] = path_depth;
-            time_distance[tree_index][u] = root_time - time_depth;
-            u_index = sample_index_map[u];
-            if (u_index != TSK_NULL) {
-                M[tree_index][u_index + N] = times[tree->parent[u]] - times[u];
-            }
-        }
-        for (n1 = 0; n1 < num_samples; n1++) {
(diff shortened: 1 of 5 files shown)


jeromekelleher:
This looks great, thanks @daniel-goldstein. One basic question though: isn't the LeafTracker basically the same things as the SAMPLE_LISTS option for tsk_tree_t? For example, we use this in [genotypes.c](https://github.com/tskit-dev/tskit/blob/master/c/tskit/genotypes.c#L279) to quickly find the set of samples that descend from a given mutation when we're decoding genotypes. Ironically, though, it turns out that doing simple traversals like [this](https://github.com/tskit-dev/tskit/blob/master/c/tskit/genotypes.c#L351) are often *more* efficient than using the sample list feature. This is probably because the [algorithm for maintaining](https://github.com/tskit-dev/tskit/blob/master/c/tskit/trees.c#L3687) sample lists is a little stupid: we basically traverse upwards each time and rebuild the parts of the linked list on each update. I tried **really** hard to make this better though and failed --- it's very tricky. Also, though, we're only looking for sample lists for a small number of nodes per tree when decoding mutations. We'll be doing lots more for KC.

Anyway, the short version is, I don't think there's much point in having an extra LeafTracker class and we might as well just use the SampleList feature of the trees. The simplest thing to do is to raise an error if the tree doesn't have this feature enabled when kc_distance is called.

Looking forward to talking through the rest of the details on Friday.

daniel-goldstein:
AH ok, this makes sense. I thought we were discussing the use of something like the leaf tracker over the sideways links in the quintuply linked tree structure. Either way, glad I can pare down this PR then, not a big deal to change how we iterate through the leaves. Will discuss more tomorrow then.

daniel-goldstein:
Still got a few more tweaks, left. Running some benchmarks to see how KC performs.

daniel-goldstein:
Ran some benchmarks on holly comparing master and this branch. Below shows the runtime for a simple msprime simulation, no params except the number of samples and a fixed seed. Then run the KC distance of the first tree against itself.

Screen Shot 2020-03-26 at 11 39 53 AM

OOM errors start ~46k samples for both branches. Just under that, at 40k, this branch still runs at ~46s.

jeromekelleher:
Great work @daniel-goldstein. Is this ready for final review?

.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-        language_version: python3.6
+        language_version: python3
```
daniel-goldstein: Remove this

c/tskit/trees.c
```diff
@@ -4551,3 +4411,224 @@ tsk_diff_iter_next(tsk_diff_iter_t *self, double *ret_left, double *ret_right,
...
+{
+    tsk_safe_free(self->m);
+    tsk_safe_free(self->M);
+}
+
+static void
+update_kc_vectors_single_leaf(
+    tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u, double time)
+{
+    const tsk_id_t *sample_index_map = tree->tree_sequence->sample_index_map;
```
daniel-goldstein: This should take a tree sequence instead of a tree because it doesn’t use the tree and can then be reused in incremental algorithms where we have the sequence but no actual tree (same with similar helpers)

daniel-goldstein: Since it’s a static method on a tree sequence, do we give it the same name spacing tsk_update...?

python/tests/test_topology.py
```diff
@@ -232,45 +298,49 @@ def test_same_tree_zero_distance(self):
         for n in range(2, 10):
             for seed in range(1, 10):
                 ts = msprime.simulate(n, random_seed=seed)
-                tree = ts.first()
+                tree = next(ts.trees(sample_lists=True))
```
daniel-goldstein: I couldn’t find a place in Python to provide the sample_lists flag other than the trees iterator. Is there a better way to do this?

daniel-goldstein:
> Great work @daniel-goldstein. Is this ready for final review?

Thanks! I left a few comments I’d like your thoughts on. Otherwise good for review.

python/tests/test_topology.py
jeromekelleher: This is one way to do it. The other is 
```python
tree = tskit.Tree(ts, sample_lists=True)
tree.first()
```
I guess we should add kwargs to the ``first()`` etc methods.

jeromekelleher: The way you have is better than using the low-level API though.

.pre-commit-config.yaml
jeromekelleher: We need a PR to update this.

c/tskit/trees.c
```diff
@@ -4551,3 +4411,224 @@ tsk_diff_iter_next(tsk_diff_iter_t *self, double *ret_left, double *ret_right,
     self->tree_left = right;
     return ret;
 }
+
```
jeromekelleher: Let's stick in a banner here like the one above for Tree diff iterator

c/tskit/trees.c
jeromekelleher: Sure, let's pass in the tree sequence. I don't think we need to make it a method of tsk_treeseq, since it doesn't alter any state. The name is fine as it's static and won't be exported.

jeromekelleher: Might be worth inlining these helper functions too, as they are so simple.

c/tskit/trees.c
```diff
@@ -4551,3 +4411,224 @@ tsk_diff_iter_next(tsk_diff_iter_t *self, double *ret_left, double *ret_right,
...
+    kc_vecs->m[pair_index] = depth;
+    kc_vecs->M[pair_index] = time;
+}
+
+static void
+update_kc_vectors_all_pairs(tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u,
+    tsk_id_t v, int depth, double time)
+{
+    tsk_id_t leaf1_index, leaf2_index, leaf1, leaf2;
+    const tsk_id_t *samples = tree->tree_sequence->samples;
```
jeromekelleher: Let's try to squeeze a bit more perf out here. Something like
```c
    tsk_id_t leaf1_index, leaf2_index, leaf1, leaf2;
￼    const tsk_id_t *restrict samples = tree->tree_sequence->samples;
    const tsk_id_t *restrict left_sample = tree->left_sample;
    const tsk_id_t *restrict left_sample = tree->right_sample;
    const tsk_id_t *restrict sample_index_map = tree->tree_sequence->sample_index_map;
    double *restrict kc_m = c_vecs->m;
    double *restrict kc_M = c_vecs->M;

    leaf1_index = left_sample[u];
￼    while (true) {
￼        leaf1 = samples[leaf1_index];
￼        leaf2_index = left_sample[v];
￼        while (true) {
￼            leaf2 = samples[leaf2_index];
￼            // Put the logic for this helper function in here,referring only to local restrict pointers 
            update_kc_vectors_pair(tree, kc_vecs, leaf1, leaf2, depth, time);
￼            

             if (leaf2_index ==  right_sample[v]) {
￼                break;
￼            }
￼            leaf2_index = next_sample[leaf2_index];
￼        }
￼        if (leaf1_index == right_sample[u]) {
￼            break;
￼        }
￼        leaf1_index = tree->next_sample[leaf1_index];
￼    }
￼}

File: .pre-commit-config.yaml (hunk 1 above)
Comment (daniel-goldstein) [reply .pre-commit-config.yaml]: yup, it accidentally got dragged in with the rest of my modifications.

File: c/tskit/trees.c (hunk 2 above)
Comment (daniel-goldstein) [reply c/tskit/trees.c]: sounds good!

**Comment by daniel-goldstein** [comment]
Determined that a good chunk of time (~40%) is spent in moving data into the KC vectors. The leaf traversal order does not align well with the positions in the vectors (which are O(n^2) in space) and so gets very poor cache behavior. Leaving a comment in the code to explain this.

File: python/tests/test_topology.py [hunk 6]
```diff
@@ -168,6 +168,103 @@ def kc_distance(tree1, tree2, lambda_=0):
...
+        vT1 = (kc_vecs1.m[i] * (1 - lambda_)) + (lambda_ * kc_vecs1.M[i])
+        vT2 = (kc_vecs2.m[i] * (1 - lambda_)) + (lambda_ * kc_vecs2.M[i])
+        distance_sum += (vT1 - vT2) ** 2
+
+    return math.sqrt(distance_sum)
+
+
 def kc_distance_simple(tree1, tree2, lambda_=0):
     """
     Simplified version of the kc_distance() function above.
```
jeromekelleher: I think we should keep this version of kc_distance and replace the other one above. The idea is that we have a very naive, literal implementation of the metric as a base to check our more sophisticated algorithm against.  This version looks a bit simpler than the one above, right?

daniel-goldstein: I agree that the most straight-forward way to think about it is:
Cartesian product of leaves --> mrca --> depth/time
I think I could write that even a bit simpler than the two-step process that's being done here.

On a related note, I find the name `kc_distance_simple` to be kind of confusing, because I don't know whether to associate simple with naive or without advanced features. How about bare_bones or c_compatible?

jeromekelleher: Sure - how about ``naive_kc_distance``?

python/tests/test_topology.py
```diff
@@ -134,40 +134,130 @@ def kc_distance(tree1, tree2, lambda_=0):
...
-                        a = min(index1, index2)
-                        b = max(index1, index2)
-                        pair_index = a * (a - 2 * k + 1) // -2 + b - a - 1
-                        assert m[tree_index][pair_index] == 1
-                        m[tree_index][pair_index] = depth
-                        M[tree_index][pair_index] = tree.time(tree.root) - time
-            if len(tree.children(node)) == 0:
-                index = sample_index_map[node]
-                M[tree_index][index + n] = tree.branch_length(node)
+        for n1 in range(n):
```
jeromekelleher: If we're using this as the naive implementation, then let's use itertools to generate the sample pairs. We want two implementations:
1) naive_kc_distance which is the simplest, most Pythonic and readable code that we can come up with to compute the statistic
2) kc_distance which is mirror of the C implementation. (Maybe we should call this ``c_kc_distance`` for clarity?)

daniel-goldstein: Yup, I like that.

jeromekelleher:
Awesome, thanks @daniel-goldstein! Can you update the CHANGELOG with this (Performance improvements in kc_distance) please? We're good to merge then.

daniel-goldstein:
@jeromekelleher All set

jeromekelleher:
Thanks @daniel-goldstein, this is great stuff.

Issue #57, opened 2017-12-21
Metadata schemas and automatic decoding


jeromekelleher:
The metadata API allows you to store any byte string along with nodes, sites and mutations. This allows us to store pickled Python objects, for example, which will work fine. However, this isn't great for portability:

1. Obviously won't work on any other language than Python
2. Pickling/unpickling requires the class definition to be in the current namespace, which can get tricky. In practise, this would probably end up causing compatability headaches across code versions.

It would be better if we could use some third-party approach for encoding metadata so  that we can automatically decode the stored information into a Python object using a given schema. I'm thinking of things like JSON-schema and protobuf. If the schema is also stored in the HDF5 file, then the schema goes along with the data, making it much more likely the data can be interpreted correctly later.

I've done a little initial experimentation with this in #330, which shows how we can use JSON-schema, where the user is responsible for managing the schemas and so on. Here is a sketch of how it might work in the future if we put in some infrastructure for automatically decoding metadata into Python objects via a few pluggable hooks:

```python
# Your schema for custom metadata defined using json-schema
schema = """{
    "title": "Example Metadata",
    "type": "object",
    "properties": {
        "one": {"type": "string"},
        "two": {"type": "string"}
    },
    "required": ["one", "two"]
}"""

# Load this schema into an object builder, so that we make metadata objects
# Using https://github.com/cwacek/python-jsonschema-objects
builder = pjs.ObjectBuilder(json.loads(self.schema))
ns = builder.build_classes()
# Make a new metadata object, and assign it to a row in the nodes table.
metadata = ns.ExampleMetadata(one="node1", two="node2")
encoded = json.dumps(metadata.as_dict()).encode()
nodes.add_row(time=0.125, metadata=encoded)
# Load this table into a tree sequence, and store the schema so it can be retrieved later.
ts = msprime.load_tables(nodes=nodes, edges=edges, sequence_length=1)
ts.node_metadata_schema = schema
ts.dump("metadata-example.hdf5")

# Later, on another system, etc, we can load up the metadata and work with it.
ts = msprime.load("metadata-example.hdf5")
schema = json.loads(ts.node_metadata_schema)
builder = pjs.ObjectBuilder(json.loads(self.schema))
ns = builder.build_classes()
def decoder(encoded):
   return ns.ExampleMetadata.from_json(encoded.decode())

# If a decoder function is set, accesses to the ``node.metadata`` are intercepted and the decoder
# function is use to translate the raw bytes into a Python object.
[...]

PR #491, opened 2020-03-19
Metadata schema (WIP)

benjeffery:
https://github.com/tskit-dev/tskit/issues/57

I've got far enough with this that I'd appreciate some feedback - on how I've modified the C code particularly. I've concentrated on the "consumer" end of things. I've only worked on the `IndividualTable` as a proof-of-concept. The added test illustrates the functionality:
```python
    def test_metadata_schema(self):
        metadata_schema = {
            "encoding": "json",
            "schema": {
                "title": "Example Metadata",
                "type": "object",
                "properties": {"one": {"type": "string"}, "two": {"type": "number"}},
                "required": ["one", "two"],
            },
        }
        data = {"one": "val one", "two": 5}
        individuals = tskit.tables.IndividualTable(metadata_schema=metadata_schema)
        individuals.add_row(metadata=json.dumps(data).encode())
        self.assertDictEqual(individuals[0].metadata, data)
```
This can be extended to other encodings such as `struct` and `msgpack`. The added row is not yet validated against the schema, but that's the plan. Could also support `individuals.add_row(metadata=data)` by encoding too.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -566,6 +566,7 @@ tsk_individual_table_print_state(tsk_individual_table_t *self, FILE *out)
     fprintf(out, "metadata_length = %d\tmax= %d\tincrement = %d)\n",
         (int) self->metadata_length, (int) self->max_metadata_length,
         (int) self->max_metadata_length_increment);
+    fprintf(out, "metadata_schema = %.*s\n", self->metadata_schema_length, self->metadata_schema);
     fprintf(out, TABLE_SEP);
     /* We duplicate the dump_text code here because we want to output
      * the offset columns. */
@@ -653,8 +654,7 @@ tsk_individual_table_dump_text(tsk_individual_table_t *self, FILE *out)
     for (j = 0; j < self->num_rows; j++) {
         metadata_len = self->metadata_offset[j + 1] - self->metadata_offset[j];
         err = fprintf(out, "%d\t%d\t", (int) j, (int) self->flags[j]);
-        if (err < 0) {
-            goto out;
+        if (err < 0) {err = fprintf(out, "#metadata_schema");
         }
         for (k = self->location_offset[j]; k < self->location_offset[j + 1]; k++) {
             err = fprintf(out, "%.*g", TSK_DBL_DECIMAL_DIG, self->location[k]);
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -566,7 +566,6 @@ tsk_individual_table_print_state(tsk_individual_table_t *self, FILE *out)
     fprintf(out, "metadata_length = %d\tmax= %d\tincrement = %d)\n",
         (int) self->metadata_length, (int) self->max_metadata_length,
         (int) self->max_metadata_length_increment);
-    fprintf(out, "metadata_schema = %.*s\n", self->metadata_schema_length, self->metadata_schema);
     fprintf(out, TABLE_SEP);
     /* We duplicate the dump_text code here because we want to output
      * the offset columns. */
@@ -654,7 +653,8 @@ tsk_individual_table_dump_text(tsk_individual_table_t *self, FILE *out)
     for (j = 0; j < self->num_rows; j++) {
         metadata_len = self->metadata_offset[j + 1] - self->metadata_offset[j];
         err = fprintf(out, "%d\t%d\t", (int) j, (int) self->flags[j]);
-        if (err < 0) {err = fprintf(out, "#metadata_schema");
+        if (err < 0) {
+            goto out;
         }
         for (k = self->location_offset[j]; k < self->location_offset[j + 1]; k++) {
             err = fprintf(out, "%.*g", TSK_DBL_DECIMAL_DIG, self->location[k]);
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -2500,7 +2500,7 @@ IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *
         }
     }
     ret = tsk_individual_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -2969,7 +2969,7 @@ NodeTable_set_metadata_schema(NodeTable *self, PyObject *arg, void *closure)
         }
     }
     ret = tsk_node_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -3455,7 +3455,7 @@ EdgeTable_set_metadata_schema(EdgeTable *self, PyObject *arg, void *closure)
         }
     }
     ret = tsk_edge_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -3950,7 +3950,7 @@ MigrationTable_set_metadata_schema(MigrationTable *self, PyObject *arg, void *cl
         }
     }
     ret = tsk_migration_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -4407,7 +4407,7 @@ SiteTable_set_metadata_schema(SiteTable *self, PyObject *arg, void *closure)
         }
     }
     ret = tsk_site_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -4903,7 +4903,7 @@ MutationTable_set_metadata_schema(MutationTable *self, PyObject *arg, void *clos
         }
     }
     ret = tsk_mutation_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
@@ -5315,7 +5315,7 @@ PopulationTable_set_metadata_schema(PopulationTable *self, PyObject *arg, void *
         }
     }
     ret = tsk_population_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
-    if (ret < 0) {
+    if (ret != 0) {
         handle_library_error(ret);
         goto out;
     }
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -184,6 +184,9 @@ typedef struct {
     tsk_viterbi_matrix_t *viterbi_matrix;
 } ViterbiMatrix;
 
+/* A named tuple of metadata schemas for a tree sequence */
+PyTypeObject MetadataSchemas={};
+
 static void
 handle_library_error(int err)
 {
@@ -6583,6 +6586,24 @@ out:
 }
 
 static PyObject *
+TreeSequence_get_metadata_schemas(TreeSequence *self) {
+    PyObject *ret = NULL;
+
+    ret = PyStructSequence_New(&MetadataSchemas);
+    if (ret == NULL) {
+        goto out;
+    }
+    PyStructSequence_SetItem(
+        ret, 
+        0, 
+        PyBytes_FromStringAndSize(
+            self->tree_sequence->tables->nodes.metadata_schema, 
+            self->tree_sequence->tables->nodes.metadata_schema_length));
+out:
+    return ret;
+}
+
+static PyObject *
 TreeSequence_get_mutation(TreeSequence *self, PyObject *args)
 {
     int err;
@@ -8015,6 +8036,8 @@ static PyMethodDef TreeSequence_methods[] = {
         "Returns the number of unique nodes in the tree sequence." },
     {"get_num_samples", (PyCFunction) TreeSequence_get_num_samples, METH_NOARGS,
         "Returns the sample size" },
+    {"get_metadata_schemas", (PyCFunction) TreeSequence_get_metadata_schemas, METH_NOARGS,
+        "Returns the metadata schemas for the tree sequence tables"},
     {"get_samples", (PyCFunction) TreeSequence_get_samples, METH_NOARGS,
         "Returns the samples." },
     {"genealogical_nearest_neighbours",
@@ -10537,6 +10560,21 @@ PyInit__tskit(void)
     }
     Py_INCREF(&LsHmmType);
     PyModule_AddObject(module, "LsHmm", (PyObject *) &LsHmmType);
+
+    /* Metadata schemas namedtuple type*/
+    PyStructSequence_Field metadata_schemas_fields[] = {
+        {"node", "The node metadata schema"},
+        {NULL}
+    };
+    PyStructSequence_Desc metadata_schemas_desc = {
+        "MetadataSchemas",
+        "Namedtuple of metadata schemas for this tree sequence",
+        metadata_schemas_fields,
+        1
+    };
+    PyStructSequence_InitType(&MetadataSchemas, &metadata_schemas_desc);
+    Py_INCREF(&MetadataSchemas);
+    PyModule_AddObject(module, "MetadataSchemas", (PyObject*)&MetadataSchemas);
 
     /* Errors and constants */
     TskitException = PyErr_NewException("_tskit.TskitException", NULL, NULL);
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -242,6 +242,19 @@ make_metadata(const char *metadata, Py_ssize_t length)
 }
 
 static PyObject *
+make_Py_Unicode_FromStringAndLength(const char * str, size_t length) {
+    PyObject * ret;
+    /* Py_BuildValue returns Py_None for zero length */
+    if (length == 0) {
+        ret = PyUnicode_FromString("");
+    } else {
+        ret = Py_BuildValue("s#", str, length);
+    }
+    return ret;
+}
+    
+
+static PyObject *
 make_mutation(tsk_mutation_t *mutation)
 {
     PyObject *ret = NULL;
@@ -2465,7 +2478,7 @@ IndividualTable_get_metadata_schema(IndividualTable *self, void *closure)
     if (IndividualTable_check_state(self) != 0) {
         goto out;
     }
-    ret = PyBytes_FromStringAndSize(self->table->metadata_schema, self->table->metadata_schema_length);
+    ret = make_Py_Unicode_FromStringAndLength(self->table->metadata_schema, self->table->metadata_schema_length);
 out:
     return ret;
 }
@@ -2474,14 +2487,15 @@ static int
 IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *closure)
 {
     int ret = -1;
-    char *metadata_schema = "";
+    const char *metadata_schema = "";
     Py_ssize_t metadata_schema_length = 0;
     
     if (IndividualTable_check_state(self) != 0) {
         goto out;
     }
     if (arg != NULL) {
-        if (PyBytes_AsStringAndSize(arg, &metadata_schema, &metadata_schema_length) < 0) {
+        metadata_schema = PyUnicode_AsUTF8AndSize(arg, &metadata_schema_length);
+        if (metadata_schema == NULL) {
             goto out;
         }
     }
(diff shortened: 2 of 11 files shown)


c/tskit/tables.c
```diff
@@ -578,14 +582,31 @@ tsk_individual_table_get_row(
     return ret;
 }
 
+int
+tsk_individual_table_set_metadata_schema(
+    tsk_individual_table_t *self, const char *metadata_schema)
```
jeromekelleher: I think we should change the signature to include a length, otherwise this will break with unicode inputs. Since we'll be allowing free text within the schema for description fields, I think we need to allow for arbitrary unicode.

c/tskit/tables.c
```diff
@@ -578,14 +582,31 @@ tsk_individual_table_get_row(
...
+
 int
 tsk_individual_table_dump_text(tsk_individual_table_t *self, FILE *out)
 {
     int ret = TSK_ERR_IO;
     tsk_size_t j, k;
     tsk_size_t metadata_len;
     int err;
-
+    // TODO Feels like metadata schema should be here, but are there downstream tools
```
jeromekelleher: Hmm, not sure. It's just for debugging really and the print_state function will output the metadata schema there. This is something we'd need to consider in the context of round-tripping a tree sequence through text I think.

python/_tskitmodule.c
```diff
@@ -2351,6 +2351,45 @@ IndividualTable_get_metadata_offset(IndividualTable *self, void *closure)
...
+}
+
+static int
+IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *closure)
+{
+    int err;
+    int ret = -1;
+    char *metadata_schema = NULL;
+
+    if (!arg || !PyUnicode_Check(arg)) {
```
jeromekelleher: Can arg be NULL? I'd prefer the explicit ``arg == NULL`` way of saying this anyway.

jeromekelleher:
See [load_text](https://github.com/tskit-dev/tskit/blob/master/python/tskit/trees.py#L2119) for entry point in text parsing.

I'm not sure I care all that much about text to be honest, but we should consider it anyway.

hyanwong:
For research-quality data this seems good, but for messing around, it seems a bit tedious to *have* to specify a schema, if all you want to do is dump arbitrary metadata e.g. for testing purposes. Will there be a sensible default for the schema? For this sort of thing, I would also quite like the encoding to be taken care of, so I could just pass a (serializable) python dict as metadata, and the json encoding would deal with the rest:

```
individuals.add_row(metadata={"random":"anything"})
```

but then ideally I'd like to pull out the metadata and not have to explictly `json.load()` it, and I can't see how you could distinguish between metadata that has been passed in as a python object and metadata that has been explicitly json-ified by the user beforehand.

Also, just to second the idea of TreeSequence level metadata. In particular, given a tree sequence, I'd like to know whether time is measured in generations, years, thousands of years, "gene frequencies" (as in tsinfer), or is just some arbitrary order.

benjeffery:
@jeromekelleher Yeah, I've not done this right for unicode - will fix. For round tripping through text it is quite simple to have a table header be part of the dump?

@hyanwong I mentioned in #57 that we would think about auto-generating the schema being an option. Something like a `metadata_schema = generate_schema_from_example('struct', data)`. I also think we should consider `add_row` doing the validating and then encoding when a schema is present. I want to get the basic schema storage rock-solid first, then we can iterate on the high-level API.

jeromekelleher:
> @jeromekelleher For round tripping through text it is quite simple to have a table header be part of the dump?

OK, sounds good. I'll look forward to a proposal!

jeromekelleher:
> @jeromekelleher For round tripping through text it is quite simple to have a table header be part of the dump?

Ah yes, but this wouldn't include the top-level table collection metadata and schema. Then again, we don't include the top-level sequence_length attribute either, so I'm not sure we should worry about it. Text is very much "handy-to-have-for-tests-and-viz"  for me, but people shouldn't be encouraged to use it for real things.

benjeffery:
I've been reading over https://github.com/tskit-dev/msprime/issues/925 and https://github.com/popsim-consortium/stdpopsim/pull/438 thinking about how metadata is managed in a sequence of operations on a tree sequence by different tools.  

Say I want to add a new attribute to existing metadata - what happens if I specify an encoding that is different to the existing schema? As there is only one schema I can't. Given that "[Namespaces are one honking great idea -- let's do more of those!](https://www.python.org/dev/peps/pep-0020/)" is there an argument for namespacing the metadata? e.g. `individuals[0].metadata.msprime.attribute` where each namespace can have its own encoding and schema?

jeromekelleher:
> Say I want to add a new attribute to existing metadata - what happens if I specify an encoding that is different to the existing schema? As there is only one schema I can't. Given that "Namespaces are one honking great idea -- let's do more of those!" is there an argument for namespacing the metadata? e.g. individuals[0].metadata.msprime.attribute where each namespace can have its own encoding and schema?

I had imagined that well behaving code that processes a tree sequence would read in the metadata, update the schema to add new fields and then write it back out.  If all the encoding/decoding details are taken care of transparently by tskit, then the encoding shouldn't matter. How the new and old metadata is merged is definitely problematic though, and this does put a lot of trust in code that manipulates tree sequences doing the right thing.

In practise though, shouldn't tagging each field with the name of the program that produced it in the schema be sufficient? I'm not sure that we'll be doing that much mixing of metadatas, and that things like stdpopsim, msprime and pyslim are outliers which we can make sure do sensible things.

I like the idea of namespacing, but I would definitely resent all the extra ``tsinfer``'s I'd need to type when doing ``population.metadata.tsinfer.name``, or ``individual.metadata.tsinfer.id``. That to me is very much the nominal case, which we should make as fluent as possible.

jeromekelleher:
ps. I hadn't read the Zen of Python in a long time. Some real wisdom in there!

benjeffery:
Just discovered https://github.com/ijl/orjson - wow! Looks like this might be a good fit.

hyanwong:
`has strict JSON conformance in not supporting Nan/Infinity/-Infinity` - but I guess these might be OK if serialised in numpy? And it may not matter in metadata anyway.

benjeffery:
Yeah, that was the one rub. If it became a problem I maybe able to monkey-patch something in.

jeromekelleher:
Hmm, slightly worried about the orjson Rust dependency. Looks like there's good wheel support, but still. What would it actually buy us? Datetime and numpy array support looks nice, but not sure this is worth a complicated dependency...

benjeffery:
Yep, def needs to earn it's keep - one reason to bring it in might be performance, we'll be calling `dumps`/`loads` on every row. But lets bench it and see.

benjeffery:
On a similar note some of this might come in handy for validation: https://www.peterbe.com/plog/jsonschema-validate-10x-faster-in-python

benjeffery:
Ok, metadata is now validated and encoded on `add_row` and decoded (not validated) on `__get_item__` such that this now works for all tables:
```python
    def test_metadata_schema(self):
        metadata_schema = {
            "encoding": "json",
            "schema": {
                "title": "Example Metadata",
                "type": "object",
                "properties": {"one": {"type": "string"}, "two": {"type": "number"}},
                "required": ["one", "two"],
                "additionalProperties": False,
            },
        }
        data = {"one": "val one", "two": 5}
        individuals = tskit.tables.IndividualTable(metadata_schema=metadata_schema)
        individuals.add_row(metadata=data)
        self.assertDictEqual(individuals[0].metadata, data)
```
If you try to add invalid metadata you get an error like this:
```
jsonschema.exceptions.ValidationError: Additional properties are not allowed ("I really shouldn't be here" was unexpected
)

Failed validating 'additionalProperties' in schema:
    {'additionalProperties': False,
     'properties': {'one': {'type': 'string'}, 'two': {'type': 'number'}},
     'required': ['one', 'two'],
     'title': 'Example Metadata',
     'type': 'object'}

On instance:
    {"I really shouldn't be here": 6, 'one': 'val one', 'two': 5}
```

Next steps:

- Consider validation and encoding for bulk methods. Either by providing methods to encode and validate an array of metadata or by doing so inside `append_columns` and `set_columns` if those methods are passed non-bytes metadata columns.

- Currently written for simplicity, prob awful perf.

benjeffery:
Currently if `metadata_schema` is `None` then nothing is validated/decoded/encoded. This makes the code backwards compatible, but I wonder if it makes it too easy to do the wrong thing. Maybe an explicit `metadata_schema = { "encoding": "binary"}` would be better.

python/tskit/tables.py
```diff
@@ -30,6 +30,7 @@
 import warnings
 
 import numpy as np
+from jsonschema import validate
```
jeromekelleher, line 33: I prefer importing the module and calling functions on it usually, as then it's clear in the code what functions are defined locally and which ones from from outside. So, when I see ``jsonschema.validate``, it's totally obvious, but if I read ``validate`` I'm not so sure.

python/tskit/tables.py
```diff
@@ -259,12 +260,17 @@ def metadata_schema(self, metadata_schema):
     def metadata_schema(self):
         self.ll_table.metadata_schema = None
 
+    def validate_and_encode_metadata_row(self, metadata):
+        if self.metadata_schema is None:
+            return metadata
+        else:
+            validate(metadata, self.metadata_schema["schema"])
```
jeromekelleher, line 267: Be nice to define catch jsonschema's exception here and raise our own. We don't want jsonschema to become part of our API --- maybe we'll want to ditch it for something better later.

jeromekelleher: See exceptions.ProvenanceValidationError, eg.

jeromekelleher:
Looks great, let's talk through the details tomorrow.

petrelharp:
> one reason to bring it in might be performance, we'll be calling dumps/loads on every row

this might be an argument to not use json for big tree sequences where performance is an issue...

python/tskit/tables.py
benjeffery: Yep, I agree - editor put this one in and I haven't done a full-on `diff` check.
https://github.com/tskit-dev/tskit/pull/491/commits/a0c2750c320ee78f239a2c13b2307f4d10712283

benjeffery:
Bulk methods now take objects as metadata, e.g.:
```python
population_table.set_columns(metadata=[{'foo':1, 'bar':2}, {'foo':3, 'bar':4}])
```
Next step is for `table.metadata` to return decoded metadata.

python/tskit/tables.py
benjeffery: Done in https://github.com/tskit-dev/tskit/pull/491/commits/38de515732f6e037b237e0bb0d6808ae9ca33ce6

benjeffery:
@jeromekelleher and I had a chat about this PR, we decided too:

- Validation should fail if `metadata` is not passed to `add_row` and the table has a schema where `{}` would fail.
- Remove metadata validation/decoding/encoding from bulk methods on the table API
- Create a `tskit.metadata.MetadataSchema` class to manage validation, decoding/encoding.
- Add decode to `ts.node` et al.

We'll then defer other encoders and perf work to subsequent PRs

petrelharp:
> Validation should fail if metadata is not passed to add_row and the table has a schema where {} would fail.

So, will it be possible to say "X or nothing at all" in the schema, then? Currently we have mixes of metadata and no metadata when e.g. we add more mutations to a SLiM simulation with msprime. Adding SLiM metadata to all the msprime mutations would not be a bad thing, but we'll have to add the capability to msprime.mutate( ) to be able to do that, if it will fail tree sequence validation.

benjeffery:
The schema can specify, per attribute, if an attribute is required or not. Also configurable is if attributes not in the schema are allowed. I'll be adding some docs and examples to this as it's a very flexible system.

jeromekelleher:
Hey @grahamgower, would you mind taking a look over this proposal please? It'd be good to get your thoughts (especially since you're in the trenches with this stuff on stdpopsim!)

benjeffery:
@grahamgower this gist explains the situation and proposal: https://gist.github.com/benjeffery/e7c68bab9839259dbab9d888d2458fa3

grahamgower:
I'm not sure I can offer any useful feedback. Most of this discussion has gone over my head, as I'm not really familiar with tskit internals, and I'm a complete noob when it comes to encoding and json (I admit, I had to read the "what is a schema" bit on json-schema.org). The gist really helped though, thanks @benjeffery.

From the perspective of stdpopsim, I guess we'll largely pass through whatever metadata schemas are defined by msprime and SLiM/pyslim. A likely exception here will be the population metadata, which differs between the two simulators and so we'll need to replace the population metadata from one (or both) of the sources.

Regarding validation and performance, particularly for bulk methods, maybe having a "validate just one row" option would be a useful default? I imagine most code will pass consistently formatted metadata for each row, and just validating one (probably the first) will catch the most common problems.

hyanwong:
@awohns has been investigating different human genome builds (e.g. GRCh37 vs GRCh38). This would be good (indeed vital) to store somewhere in the tree sequence metadata. I mention it here @grahamgower as they will have different stdpopsim recombination maps, and ISTR there has been some discussion about how to store the reference genome identified in the tree sequences generated from stdpopsim? It would be nice to have a convention here - not sure if this is metadata territory or not, though.

grahamgower:
Yes, very good point @hyanwong! Absolutely a tree-sequence-wide metadata field, that is explicitly not bodged into the provenance table, is a very desirable feature!

petrelharp:
While we're collecting use cases, SLiM sometimes has a reference sequence (for the whole geome!) and a "generation" that says what the current time is in SLiM.

benjeffery:
@jeromekelleher I'm getting close to feature complete (but not documentation complete) for this PR. 
`TreeSequence` methods now returns container classes that lazily decode metadata.

The remaining part that I'm currently considering is where else in the Python API we need decoded metadata. I've been through `Tree` but it only returns `int`s unless I've missed something.

python/_tskitmodule.c
```diff
@@ -4198,6 +4360,44 @@ SiteTable_get_metadata_offset(SiteTable *self, void *closure)
...
+    
+    if (SiteTable_check_state(self) != 0) {
+        goto out;
+    }
+    if (arg != Py_None) {
+        if (PyBytes_AsStringAndSize(arg, &metadata_schema, &metadata_schema_length) < 0) {
+            goto out;
+        }
+    }
+    ret = tsk_site_table_set_metadata_schema(self->table, metadata_schema, metadata_schema_length);
```
jeromekelleher, line 4391: I don't think ``tsk_site_table_set_metadata_schema`` returns anything other than 0 on success, so can use the standard ``if (ret != 0)`` and don't bother setting ret = 0

Ah, but we're using ``ret`` as the Python signalling variable. I think it's best to split the concerns here, and use ``ret`` as the variable used for the Python return value and use maybe ``err = tsk_site_table_set_metadata_schema`` to track the return value of tskit API calls.

python/_tskitmodule.c
```diff
@@ -6303,6 +6585,32 @@ TreeSequence_get_site(TreeSequence *self, PyObject *args)
...
 
+static PyObject *
+TreeSequence_get_metadata_schemas(TreeSequence *self) {
+    PyObject *ret = NULL;
+    tsk_table_collection_t *tables = self->tree_sequence->tables;
+    ret = PyStructSequence_New(&MetadataSchemas);
+    if (ret == NULL) {
+        goto out;
+    }
+    PyStructSequence_SetItem(ret, 0, PyBytes_FromStringAndSize(
```
jeromekelleher, line 6596: These can fail, right? I think we'd need to check every one of these calls and clean up appropriately on errors.

I know it's tedious, but what about ``TreeSequence_get_node_metadata_schema()``, TreeSequence_get_individual_metadata_schema()``, etc? It's a bunch of boilerplate, but it's simple and easy to write in a bulletproof way.

python/tskit/metadata.py
```diff
@@ -0,0 +1,115 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Tskit Developers
+# Copyright (c) 2017 University of Oxford
```
jeromekelleher, line 4: This started in 2020, so the copyright should be 
```
# Copyright (c) 2020 Tskit Developers
```

python/tskit/metadata.py
```diff
@@ -0,0 +1,115 @@
...
+    @abc.abstractmethod
+    def decode(self, encoded_bytes: bytes) -> Any:
+        pass
+
+    @abc.abstractmethod
+    def encode(self, obj: Any) -> bytes:
+        pass
+
+
+class JSONCodec(MetadataCodec):
```
jeromekelleher, line 45: Tending to use e.g. ``JsonCodec`` elsewhere in the code, but I don't feel strongly about it.

python/_tskitmodule.c
```diff
@@ -2454,6 +2457,44 @@ IndividualTable_get_metadata_offset(IndividualTable *self, void *closure)
...
 
+static PyObject *
+IndividualTable_get_metadata_schema(IndividualTable *self, void *closure)
+{
+    PyObject *ret = NULL;
+
+    if (IndividualTable_check_state(self) != 0) {
+        goto out;
+    }
+    ret = PyBytes_FromStringAndSize(self->table->metadata_schema, self->table->metadata_schema_length);
```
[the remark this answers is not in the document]
benjeffery: I've gone back and forward on this as I didn't want to preempt other uses of the schema field e.g pickle to `bytes`. But schemas are somewhere we should be opinionated so I'm happy with unicode.  https://github.com/tskit-dev/tskit/pull/491/commits/7f4b9be2f14a6ddca6c06358d3591b0a6e21dd63

python/_tskitmodule.c
```diff
@@ -2454,6 +2457,44 @@ IndividualTable_get_metadata_offset(IndividualTable *self, void *closure)
...
+IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *closure)
+{
+    int ret = -1;
+    char *metadata_schema = "";
+    Py_ssize_t metadata_schema_length = 0;
+    
+    if (IndividualTable_check_state(self) != 0) {
+        goto out;
+    }
+    if (arg != Py_None) {
```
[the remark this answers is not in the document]
benjeffery: In my mind, something that is a simple getter/setter is a perfect candidate to be an attribute.  There's tests for the attribute here: https://github.com/benjeffery/tskit/blob/metadata/python/tests/test_tables.py#L662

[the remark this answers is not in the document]
benjeffery: You're right `None` isn't correct here, `del ll_table.metadata_schema` actually calls the setter with `NULL`. Fixed in https://github.com/tskit-dev/tskit/pull/491/commits/e53aada06ca7bbd25e7b402f60a99344bbbba8c8

python/_tskitmodule.c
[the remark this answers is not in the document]
benjeffery: https://github.com/tskit-dev/tskit/pull/491/commits/338a045e3f6586e4673076b4e2fd62ef05e9ef7f

python/_tskitmodule.c
```diff
@@ -2454,6 +2457,44 @@ IndividualTable_get_metadata_offset(IndividualTable *self, void *closure)
...
+        if (PyBytes_AsStringAndSize(arg, &metadata_schema, &metadata_schema_length) < 0) {
```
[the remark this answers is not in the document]
benjeffery: Done https://github.com/tskit-dev/tskit/commit/338a045e3f6586e4673076b4e2fd62ef05e9ef7f

python/_tskitmodule.c
```diff
@@ -2454,6 +2457,44 @@ IndividualTable_get_metadata_offset(IndividualTable *self, void *closure)
...
+    ret = PyBytes_FromStringAndSize(self->table->metadata_schema, self->table->metadata_schema_length);
+out:
+    return ret;
+}
+
+static int
+IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *closure)
+{
+    int ret = -1;
+    char *metadata_schema = "";
```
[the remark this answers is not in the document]
benjeffery: Also done in https://github.com/tskit-dev/tskit/commit/338a045e3f6586e4673076b4e2fd62ef05e9ef7f

python/_tskitmodule.c
benjeffery: The change to string changed how the string was made so just https://github.com/tskit-dev/tskit/pull/491/commits/d5881fa448ede1273e31612fd1e771ce57075e7d fixed this.

python/_tskitmodule.c
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/491/commits/338a045e3f6586e4673076b4e2fd62ef05e9ef7f I've added low-level tests in https://github.com/tskit-dev/tskit/pull/491/commits/e3f97bd90d3a0eb95dca5a2098c19f7b27174e99

python/tskit/metadata.py
benjeffery: Fixed https://github.com/tskit-dev/tskit/pull/491/commits/267ff13c7d0481f6f0a5d3d7134553d185395cb8

python/tskit/metadata.py
benjeffery: I was following the python way e.g. https://docs.python.org/3/library/json.html#json.JSONEncoder I'll change the others in the code?

benjeffery:
This PR is superseded by #542 and #543

PR #492, opened 2020-03-19
Add pre-commit for flake8-builtins

benjeffery:
Prevents shadowing built-ins as part of flake8.


jeromekelleher:
I'm kind of "meh" on this one, but the changes we need are pretty minor, so let's go for it.

benjeffery:
Yeah it isn't so relevant here, but considering we're setting a standard across all the repos for the future I think it's worth it.

PR #493, opened 2020-03-23
[C API] Metadata schema field on tables

benjeffery:
To split #491 up a bit I'm focusing on getting the C API right first. 
Proof-of-concept for the python is done in #491 meaning I'm pretty sure this is the (per-table) changes we want. Only `individuals` is modified for now. Once everyone is happy I'll do the other tables.
I intend to do top-level (per tree sequence) metadata in another PR.

- [x] Agree on pre-table changes
- [x] Individuals
- [x] Nodes
- [x] Edges (including adding metadata column (done in #496))
- [x] Sites
- [x] Mutations (including adding metadata column (done in #505)
- [x] Migrations
- [x] Populations

PR #496, opened 2020-03-24
Add metadata column to edges

benjeffery:
Part of the work in #493.

I realise I'm now three levels deep on PR requests, but at ~600 lines this seemed to warrant its own PR....

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -42,12 +42,15 @@
 #define DEFAULT_SIZE_INCREMENT 1024
 #define TABLE_SEP "-----------------------------------------\n"
 
+#define TSK_COL_OPTIONAL (1 << 0)
+
 typedef struct {
     const char *name;
     void **array_dest;
     tsk_size_t *len_dest;
     tsk_size_t len_offset;
     int type;
+    tsk_flags_t options;
 } read_table_col_t;
 
 typedef struct {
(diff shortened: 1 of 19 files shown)


benjeffery:
@jeromekeheller this is all ready for you to have a gander.

benjeffery:
@jeromekelleher This is all ready for you to have a gander.

benjeffery:
>  I know how tedious all that copy-pasting is...

Wasn't too bad as in doing it I got a good tour through the layers.

> One high-level point: do we want to increment FILE_VERSION_MAJOR for this? 

We need to at some point draw a line and from that point forward do everything we can to be backward compatible. We talked about this being 1.0 which seems to make sense as long as it lines up with how things are being used on the ground. If you think a breaking change would cause a lot of suffering already then we should start being backward compatible now.

I think optional columns and defaults is a separate issue (although linked in design considerations). Certainly one to consider for metadata on all tables. I'll have a play and see what this might look like for edges in the C.

jeromekelleher:
I was thinking about the optional columns thing this morning, and if we align it with the semantics of ``set_columns`` then it wouldn't be too bad to do. The idea would be to read in some pointers to columns using kastore, and any that are marked ``optional`` and missing would be NULL. The semantics what's allowed would then be taken care of automaticaly. It would need a bit of replumbing at the bottom level and we'd need to get rid of the idea of using the kastore memory directly, but that's probably of very little value anyway and a few memcpys won't make any difference.

Anyway, I don't think we should do that here. We can do something like tsk_table_collection_load_indexes for the edge table. In my mind everything we're adding to the file format in terms of metadata is optional, so I'd like to keep the file format version major the same. I've no problem bumping the version number if we do find something we need to change that'll break old code, though.

benjeffery:
@jeromekelleher So I've had a go at making metadata an optional column for edges. Basically the `metadata` and `metadata_offset` columns are `NULL` until the first row is added or row set appended that has metadata, at which point they are initialised. This saves the `uint32` per edge for empty metadata. This also means the files are backwards compatible.

I have only done the C side of things - wanted to get your thoughts before going further. It is probably easier to see these specific changes by just looking at https://github.com/tskit-dev/tskit/pull/496/commits/8ec7142c15f03d50ae2de5cffd71b28e1264c495.

jeromekelleher:
Looks good, thanks @benjeffery. I think we should simplify it a bit on this first pass though, and follow the same logic as the rest of the tables. Metadata/metadata_offset are never NULL, as we set them to default values if they are not specified. If we do the same thing when we read the data in from file, this should minimise the changes.

It would be simpler to just do something specific for the edge_table_t (see the bit where we're reading in the indexes, e.g) rather than making a generic optional/non-optional mechanism. However, there's no point in taking out the changes you've made to the rest of the columns as we'll need them soon.

Figuring out when we do and don't write out columns is something we can follow up on. I don't think storing NULL is the way to do it though, we should always have valid memory for each of the columns in each table as part of the API contract. (For metadata, we can do something simple like ``if metadata_length == 0 don't store metadata or metadata offets``)

benjeffery:
@jeromekelleher Ok, so this latest push creates an empty `metadata` and `metadata_offset` if none are found in the file. Note that this means that we keep the `uint32` extra per edge for now. Is this what you meant in your last comment?

c/tskit/tables.c
```diff
@@ -94,22 +97,32 @@ read_table_cols(kastore_t *store, read_table_col_t *read_cols, size_t num_cols)
...
+                *read_cols[j].len_dest = (tsk_size_t)(len - read_cols[j].len_offset);
+            } else if ((last_len + read_cols[j].len_offset) != (tsk_size_t) len) {
+                ret = TSK_ERR_FILE_FORMAT;
+                goto out;
+            }
+            if (type != read_cols[j].type) {
+                ret = TSK_ERR_FILE_FORMAT;
+                goto out;
+            }
+        } else if (!(read_cols[j].options && TSK_COL_OPTIONAL)) {
```
jeromekelleher: Did you mean boolean && or bitwise &? Probably bitwise?

jeromekelleher:
> @jeromekelleher Ok, so this latest push creates an empty metadata and metadata_offset if none are found in the file. Note that this means that we keep the uint32 extra per edge for now. Is this what you meant in your last comment?

Yes, that's the idea. It's probably more efficient to go through this in person though, let's talk it through later.

c/tskit/tables.c
benjeffery: Opps, fixed in https://github.com/tskit-dev/tskit/pull/496/commits/ace3d7b7de5568a04a00b4a2ff76b42f138983c4

benjeffery:
@jeromekelleher I've added the `metadata_malloced_locally` flag and added a python test I realised was missing. We talked about doing the same for migrations, I guess it would almost be a copy-paste of this?

python/_tskitmodule.c
```diff
@@ -366,10 +366,24 @@ make_node(tsk_node_t *r)
...
 {
-    return Py_BuildValue("ddii",
-            edge->left, edge->right, (int) edge->parent, (int) edge->child);
+    PyObject *ret = NULL;
+    PyObject* metadata = make_metadata(edge->metadata, (Py_ssize_t) edge->metadata_length);
+    if (metadata == NULL) {
+        goto out;
+    }
+    if (include_id) {
+    ret = Py_BuildValue("ddiiOi",
```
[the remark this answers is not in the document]
jeromekelleher: Indentation still wonky.

jeromekelleher:
I'm assuming that codecov is being pedantic also, and we've covered everything that we reasonably can?

benjeffery:
Thought I'd fixed that indent! Must have run `git-fubar` at some point.
Rebased and squashed. I think coverage is complaining about the patch as it adds a few lines that aren't covered as they are only run on OOM.

PR #497, opened 2020-03-25
more explanation of summary funcs

### LINKED ISSUE #398: f2, f3 and f4 stat definitions -- some ambiguity?
In my lab we are currently doing some checks that when we convert our tree sequences into genotype matrices and run them through ADMIXTOOLS or similar, we get the same values for f* statistics as we do when we calculate them directly on the tree sequence using the your new methods.

I got the following question from one of my colleagues:

> Can you confirm whether these statistics are calculated on population allele frequencies, or averages of randomly chosen individuals from each population?

My first instinct was the former, and this definition in @petrelharp and @jeromekelleher 's latest  preprint suggests that too:

Screen Shot 2019-11-01 at 6 35 14 pm

However, in API documentation for f2,f3 and f4 stats, this statement seems to suggest the other definition:

> Each is an average across randomly chosen set of four samples (a, b; c, d), one from each sample set:

It would be great if you could help us clarify this. Btw, I've also checked that when I rerun the f* stat calculations on the tree sequence, I get the same output, so if there is some randomness in the calculation it doesn't seem like the user can control it. This would be useful for the user to know, I think.

Also, btw, when you were coding these definitions up, did you also double check your f* site statistics against those produced with [scikit-allel](https://scikit-allel.readthedocs.io/en/stable/) or some other package? It may be useful for us to use the same package for our own statistic calculations, if so.

cc @mslick95

petrelharp:
Closes #398. But I am a bit concerned it is not actually clarifying anything.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -483,11 +483,14 @@ Summary functions
 
 For convenience, here are the summary functions used for many of the statistics.
 Below, :math:`x` denotes the number of samples in a sample set below a node,
-`n` denotes the total size of a sample set,
+`n` denotes the total size of a sample set, :math:`p = x / n`,
 and boolean expressions (e.g., :math:`(x > 0)`) are interpreted as 0/1.
 
 ``diversity``
    :math:`f(x) = \frac{x (n - x)}{n (n-1)}`
+
+   For an unpolarized statistic with biallelic loci, this calculates
+   :math:`2 p (1-p)`.
 
 ``segregating_sites``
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM. @gtsambos, any thoughts?

gtsambos:
Nice!

> I am a bit concerned it is not actually clarifying anything.

I think it does

petrelharp:
Shall we merge this, then?

jeromekelleher:
Done!

PR #505, opened 2020-03-30
Add metadata column to the migrations table

benjeffery:
Add migration metadata, note that like edges this is backward compatible as the columns are optional in the kastore.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -53,7 +53,7 @@ EdgeTableRow = collections.namedtuple(
 
 
 MigrationTableRow = collections.namedtuple(
-    "MigrationTableRow", ["left", "right", "node", "source", "dest", "time"]
+    "MigrationTableRow", ["left", "right", "node", "source", "dest", "time", "metadata"]
 )
 
 
@@ -761,7 +761,7 @@ class EdgeTable(BaseTable, MetadataMixin):
         self.ll_table.squash()
 
 
-class MigrationTable(BaseTable):
+class MigrationTable(BaseTable, MetadataMixin):
     """
     A table defining the migrations in a tree sequence. See the
     :ref:`definitions <sec_migration_table_definition>` for details on the columns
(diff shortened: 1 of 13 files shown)


benjeffery:
Looks like I broke the docs, will fix later as AFK.

benjeffery:
@jeromekelleher ready for a once over.

c/tskit/core.h
```diff
@@ -96,7 +96,7 @@ to the API or ABI are introduced, i.e., internal refactors of bugfixes.
 #define TSK_FILE_FORMAT_NAME          "tskit.trees"
 #define TSK_FILE_FORMAT_NAME_LENGTH   11
 #define TSK_FILE_FORMAT_VERSION_MAJOR 12
-#define TSK_FILE_FORMAT_VERSION_MINOR 1
+#define TSK_FILE_FORMAT_VERSION_MINOR 2
```
jeromekelleher: Probably not worth bothering with the minor version bump here? We've not released it so I think it's reasonable to merge together all the compatible changes between releases into one minor version bump. Doesn't seem much point in having lots of transient file versions.

python/tskit/tables.py
```diff
@@ -849,22 +880,50 @@ def set_columns(
         :type dest: numpy.ndarray, dtype=np.int32
         :param time: The time of each migration.
         :type time: numpy.ndarray, dtype=np.int64
+        :param metadata: The flattened metadata array. Must be specified along
+            with ``metadata_offset``. If not specified or None, an empty metadata
+            value is stored for each node.
```
jeromekelleher: s/node/migration

python/tskit/tables.py
```diff
@@ -878,9 +937,24 @@ def append_columns(self, left, right, node, source, dest, time):
...
```
jeromekelleher: ditto

c/tskit/core.h
benjeffery: Cool, as long one PR that needs to in a release increments should be good. I wasn't sure how many users we had who are on the bleeding edge.

benjeffery:
@jeromekelleher Ok, all fixed.

c/tskit/core.h
jeromekelleher: Trying to discourage bleeding edge users, but this is harmless anyway since the changes are backward/forward compatible.

jeromekelleher:
Merged, thanks. OK, that's the boring stuff out of the way!

PR #508, opened 2020-04-02
Error on unary nodes in KC distance

### LINKED ISSUE #487: kc_distance doesn't ignore unary nodes
We should presumably have the same KC distance regardless of unary nodes. But the example below shows that it doesn't work:

```python
# use record_full_arg to leave in some unary nodes
ts_unary = msprime.simulate(10, random_seed=1, recombination_rate=10, record_full_arg=True)

tree02 = ts_unary.at(0.2)
tree08 = ts_unary.at(0.8)
kc_unary = tree02.kc_distance(tree08)

ts_nounary = ts_unary.simplify()

tree02 = ts_nounary.at(0.2)
tree08 = ts_nounary.at(0.8)
kc_nounary = tree02.kc_distance(tree08)

assert kc_unary == kc_nounary  # currently fails
```

ping @daniel-goldstein as he's working on KC distance calcs at the moment.

daniel-goldstein:
The KC distance is only defined on full trees and unary nodes greatly affect distances. `kc_distance` now errors on trees with unary nodes. Users should run `simplify` on trees before comparing.

Resolves #487 

cc @jeromekelleher @hyanwong

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -4595,6 +4595,31 @@ test_unequal_samples_kc(void)
     tsk_tree_free(&other_t);
 }
 
+static void
+test_unary_nodes_kc(void)
+{
+    const char *nodes = "1  0   0\n"
+                        "1  0   0\n"
+                        "0  1   0\n"
+                        "0  2   0";
+    const char *edges = "0  1   2   0,1\n"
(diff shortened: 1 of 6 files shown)


jeromekelleher:
LGTM, thanks @daniel-goldstein!

PR #510, opened 2020-04-03
Ranking and unranking for leaf-labeled trees

daniel-goldstein:
This introduces a combinatorial method of ranking and unranking trees based on the labels of the leaves. Trees are considered unordered (ordering of the children under a node doesn't matter) and only leaf-labeled (IDs on internal nodes don't matter). This supports arbitrary polytomies and only considers the distinction between leaves and internal nodes, so does not care about samples. It does not support trees with multiple roots or trees with unary nodes.

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -1,0 +1,415 @@
+#
+# MIT License
+#
+# Copyright (c) 2020 Tskit Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for combinatorial algorithms.
+"""
+import itertools
+import unittest
+
+import tskit
+import tskit.combinatorics as comb
+from tskit.combinatorics import RankTree
+
+
+class TestCombination(unittest.TestCase):
+    def test_combination_with_replacement_rank_unrank(self):
+        for n in range(9):
+            for k in range(n):
+                nums = list(range(n))
+                combs = itertools.combinations_with_replacement(nums, k)
+                for exp_rank, c in enumerate(combs):
+                    c = list(c)
+                    actual_rank = comb.Combination.with_replacement_rank(c, n)
+                    self.assertEqual(actual_rank, exp_rank)
+                    unranked = comb.Combination.with_replacement_unrank(exp_rank, n, k)
+                    self.assertEqual(unranked, c)
+
+    def test_combination_rank_unrank(self):
+        for n in range(11):
+            for k in range(n):
+                nums = list(range(n))
+                for rank, c in enumerate(itertools.combinations(nums, k)):
+                    c = list(c)
+                    self.assertEqual(comb.Combination.rank(c, nums), rank)
+                    self.assertEqual(comb.Combination.unrank(rank, nums, k), c)
+
+    def test_combination_unrank_errors(self):
+        self.verify_unrank_errors(1, 1, 1)
+        self.verify_unrank_errors(2, 0, 1)
+
+    def verify_unrank_errors(self, rank, n, k):
+        with self.assertRaises(ValueError):
+            comb.Combination.unrank(rank, list(range(n)), k)
+
+
+class TestPartition(unittest.TestCase):
+    def test_rule_asc(self):
+        self.verify_rule_asc(1, [[1]])
+        self.verify_rule_asc(2, [[1, 1], [2]])
+        self.verify_rule_asc(3, [[1, 1, 1], [1, 2], [3]])
+        self.verify_rule_asc(4, [[1, 1, 1, 1], [1, 1, 2], [1, 3], [2, 2], [4]])
+        self.verify_rule_asc(
+            5,
+            [[1, 1, 1, 1, 1], [1, 1, 1, 2], [1, 1, 3], [1, 2, 2], [1, 4], [2, 3], [5]],
+        )
+        self.verify_rule_asc(
+            6,
+            [
+                [1, 1, 1, 1, 1, 1],
+                [1, 1, 1, 1, 2],
+                [1, 1, 1, 3],
+                [1, 1, 2, 2],
+                [1, 1, 4],
+                [1, 2, 3],
+                [1, 5],
+                [2, 2, 2],
+                [2, 4],
+                [3, 3],
+                [6],
+            ],
+        )
+
+    def verify_rule_asc(self, n, partitions):
+        self.assertEqual(list(comb.rule_asc(n)), partitions)
+
+    def test_partitions(self):
+        self.assertEqual(list(comb.partitions(0)), [])
+        for n in range(1, 7):
+            self.assertEqual(list(comb.partitions(n)), list(comb.rule_asc(n))[:-1])
+
+    def test_group_partition(self):
+        self.assertEqual(comb.group_partition([1]), [[1]])
+        self.assertEqual(comb.group_partition([1, 2]), [[1], [2]])
+        self.assertEqual(comb.group_partition([1, 1, 1]), [[1, 1, 1]])
+        self.assertEqual(comb.group_partition([1, 1, 2, 3, 3]), [[1, 1], [2], [3, 3]])
+
+
+class TestRankTree(unittest.TestCase):
+    def test_num_shapes(self):
+        for i in range(11):
+            all_trees = RankTree.all_unlabelled_trees(i)
+            self.assertEqual(len(list(all_trees)), comb.num_shapes(i))
+
+    def test_num_labellings(self):
+        for n in range(2, 8):
+            for tree in RankTree.all_unlabelled_trees(n):
+                tree = tree.label_unrank(0)
+                tree2 = tree.to_tsk_tree()
+                n_labellings = sum(1 for _ in RankTree.all_labellings(tree))
+                self.assertEqual(
+                    n_labellings, RankTree.from_tsk_tree(tree2).num_labellings()
+                )
+
+    def test_num_labelled_trees(self):
+        # Number of leaf-labelled trees with n leaves on OEIS
+        n_trees = [0, 1, 1, 4, 26, 236, 2752, 39208]
+        for i, expected in zip(range(len(n_trees)), n_trees):
+            actual = sum(1 for _ in RankTree.all_labelled_trees(i))
+            self.assertEqual(actual, expected)
+
+    def test_all_labelled_trees_3(self):
+        expected = ["(0,1,2)", "(0,(1,2))", "(1,(0,2))", "(2,(0,1))"]
+        actual = [t.newick() for t in RankTree.all_labelled_trees(3)]
+        self.assertEqual(expected, actual)
+
+    def test_all_labelled_trees_4(self):
+        expected = [
+            # 1 + 1 + 1 + 1 (partition of num leaves)
+            "(0,1,2,3)",
+            # 1 + 1 + 2
+            "(0,1,(2,3))",
+            "(0,2,(1,3))",
+            "(0,3,(1,2))",
+            "(1,2,(0,3))",
+            "(1,3,(0,2))",
+            "(2,3,(0,1))",
+            # 1 + 3
+            # partition of 3 = 1 + 1 + 1
+            "(0,(1,2,3))",
+            "(1,(0,2,3))",
+            "(2,(0,1,3))",
+            "(3,(0,1,2))",
+            # partition of 3 = 1 + 2
+            "(0,(1,(2,3)))",
+            "(0,(2,(1,3)))",
+            "(0,(3,(1,2)))",
+            "(1,(0,(2,3)))",
+            "(1,(2,(0,3)))",
+            "(1,(3,(0,2)))",
+            "(2,(0,(1,3)))",
+            "(2,(1,(0,3)))",
+            "(2,(3,(0,1)))",
+            "(3,(0,(1,2)))",
+            "(3,(1,(0,2)))",
+            "(3,(2,(0,1)))",
+            # 2 + 2
+            "((0,1),(2,3))",
+            "((0,2),(1,3))",
+            "((0,3),(1,2))",
+        ]
+        actual = [t.newick() for t in RankTree.all_labelled_trees(4)]
+        self.assertEqual(expected, actual)
+
+    def test_generate_trees_roundtrip(self):
+        n = 5
+        all_rank_trees = RankTree.all_labelled_trees(n)
+        all_tsk_trees = comb.all_trees(n)
+        for rank_tree, tsk_tree in zip(all_rank_trees, all_tsk_trees):
+            self.assertEqual(rank_tree, RankTree.from_tsk_tree(tsk_tree))
+
+    def test_all_labellings_roundtrip(self):
+        n = 5
+        rank_tree = RankTree.unrank((comb.num_shapes(n) - 1, 0), n)
+        tsk_tree = rank_tree.to_tsk_tree()
+        rank_tree_labellings = RankTree.all_labellings(rank_tree)
+        tsk_tree_labellings = comb.all_tree_labellings(tsk_tree)
+        for rank_t, tsk_t in zip(rank_tree_labellings, tsk_tree_labellings):
+            self.assertEqual(rank_t, RankTree.from_tsk_tree(tsk_t))
+
+    def test_unrank(self):
+        for n in range(6):
+            for shape_rank, t in enumerate(RankTree.all_unlabelled_trees(n)):
+                for label_rank, labelled_tree in enumerate(RankTree.all_labellings(t)):
+                    unranked = RankTree.unrank((shape_rank, label_rank), n)
+                    self.assertTrue(labelled_tree == unranked)
+
+        # The number of labelled trees gets very big quickly
+        for n in range(6, 10):
+            for shape_rank in range(comb.num_shapes(n)):
+                rank = (shape_rank, 0)
+                unranked = RankTree.unrank(rank, n)
+                self.assertTrue(rank, unranked.rank())
+
+                rank = (shape_rank, comb.num_labellings(shape_rank, n) - 1)
+                unranked = RankTree.unrank(rank, n)
+                self.assertTrue(rank, unranked.rank())
+
+    def test_unrank_errors(self):
+        self.verify_unrank_errors((-1, 0), 1)
+        self.verify_unrank_errors((0, -1), 1)
+        self.verify_unrank_errors((-1, 0), 2)
+        self.verify_unrank_errors((0, -1), 2)
+        self.verify_unrank_errors((-1, 0), 10)
+        self.verify_unrank_errors((0, -1), 10)
+
+        self.verify_unrank_errors((0, 1), 1)
+        self.verify_unrank_errors((1, 0), 2)
+        self.verify_unrank_errors((0, 1), 2)
+        self.verify_unrank_errors((2, 0), 3)
+        self.verify_unrank_errors((0, 1), 3)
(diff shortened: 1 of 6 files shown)


python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+        A tree with n leaves can be obtained by joining a tree with k < n
+        leaves with a tree of n - k leaves for all 1 < k < n - 1.
+        This, however, results in distinct trees that are isomorphic to each
+        other since you could swap the left and right subtrees to get an already
+        existing tree.
+
+        To eliminate this consideration moving forward, we say a tree
+        T = T_l ^ T_r is in its canonical orientation, which we will call
+        left-heavy, if num_leaves(T_l) >= num_leaves(T_r), and both T_l and T_r
+        are themselves left-heavy. A leaf is trivially canonically oriented.
```
daniel-goldstein, line 125: This is a bit problematic because tskit orders edges by putting the child with the smaller ID on the left. Since all samples are smaller than internal nodes, tskit will naturally make the trees "right-heavy" if e.g. a node has one child that's a leaf and one that's an internal node. It doesn't matter for any of the combinatorics, and we don't expect naturally created tskit trees to be canonical, but it could be confusing if you were unranking trees and printing them.

petrelharp:
This is pretty neat. Out of pure curiousity - what do you want these ranks for?

daniel-goldstein:
> This is pretty neat. Out of pure curiousity - what do you want these ranks for?

One thing we could use this for is to calculate the distribution of phylogenies that are represented in a tree. We could sample from populations/species and use the ranks of the subtrees comprised of those samples to create a distribution.

python/tests/test_combinatorics.py
jeromekelleher: > Since all samples are smaller than internal nodes,

This isn't necessarily true, as sample IDs can be anything. It's just that the canonical (post simplify) version of a tree sequence always has samples as 0 to n.

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+    def from_tsk_tree(tree):
+        return Leaf(tree.root)
+
+    def to_tsk_tree(self):
+        tables = tskit.TableCollection(1)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE)
+        return tables.tree_sequence().first()
+
+
+class Tree:
```
jeromekelleher, line 82: How about this alternative:
```python
class TreeNode:
    def __init__(self, children=list(), label=None):
         self.label = label
         self.children = children
         # Presumably we're building these from the leaves up? But maybe not, and that's
         # why you have a cached num_leaves value?
         self._num_leaves = None 
         self._cached_rank = None

def is_leaf(self):
      return len(self.children) == 0

def num_leaves(self):
    if self._num_leaves is None:
         if self.is_leaf():
             self._num_leaves = 1
         else:
             self._num_leaves = sum(child.num_leaves() for child in self.children)
    return self._num_leaves
```

Some rationale:

1. It seems inelegant to consider leaves and trees to be different types of things. They're really the same thing, except leaves have no children and the label is defined.
2. Using left and right is a little restrictive (maybe we'll want to define this on non-binary trees at some point). If we do keep it as left and right, I'd suggest ``left_child`` and ``right_child``. Code that uses lists of subtrees is much more elegant than left and right IMO though.
3. I was considering using a set for the children, but I think it's better to have a list in which canonically the children are ordered by the number of leaves they subtend (either in increasing or decreasing order).

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+
+    def num_leaves(self):
+        return self._num_leaves
+
+    ########################################################################
+    # Generators
+    ########################################################################
+
+    @staticmethod
+    def all_labeled_trees(n):
```
jeromekelleher, line 102: Possible source of friction here: British english prefers "labelled" vs "labeled". Since tskit is using British English in most places, I'd prefer "labelled".

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+        """
```
jeromekelleher, line 103: Can we change this to ``all_labelled_binary_trees``?

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+        Generate all unordered, leaf-labeled trees with n leaves.
+        """
+        for tree in Tree.all_unordered_trees(n):
+            yield from Tree.all_labelings(tree)
+
+    @staticmethod
+    def all_unordered_trees(n):
```
jeromekelleher, line 110: Why are the trees "unordered" here? What would an ordered tree look like, distinct from the it's labelling?

jeromekelleher, line 110: Change to ``all_unordered_binary_trees``

python/tests/test_combinatorics.py
jeromekelleher: If we go with the approach of each node having a list of children, we can make the canonical form listing the children in nondecreasing order of subtended leaves. This seems quite natural to me and would match up with tskit.

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+from collections import Counter
+from itertools import combinations
+
+import msprime
+
+import tskit
+import tskit.combinatorics as comb
+
+
+def L_tree(t):
```
jeromekelleher, line 17: It'd be helpful to give this a more meaningful name and a docstring. Also, isn't it a method of TreeNode?

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,668 @@
...
+        )
+
+    @staticmethod
+    def from_tsk_tree(tree):
+        def subtree(node_id):
+            if tree.is_leaf(node_id):
+                return Leaf(node_id)
+            left = tree.left_child(node_id)
+            right = tree.right_child(node_id)
+            if tree.num_samples(left) < tree.num_samples(right):
```
jeromekelleher, line 451: num_samples and num_leaves are not quite the same thing as we can have internal samples. I think for this work we can assume (or should enforce) that leaves === samples.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,352 @@
...
+    """
+    The cardinality of the set of unlabeled binary trees with n leaves,
+    under the conditions of tree isomorphism universal to this module.
+    """
+    if n <= 1:
+        return n
+
+    total = 0
+    for k in range(math.ceil(n / 2), n):
+        if k == n // 2:
```
jeromekelleher, line 187: Seems inelegant to check this each time through the loop when it can only happen once, right? Is what you're doing something like
```python
total = 0
if n % 2 == 1:
    k = n // 2 + 1 
    total = (S(k) * (S(k) + 1)) // 2
for k in range(n // 2, n):
    total +=   S(k) * S(n - k)
```

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,352 @@
...
+            total += (S(k) * (S(k) + 1)) // 2
+        else:
+            total += S(k) * S(n - k)
+    return total
+
+
+def L(shape_rank, n):
```
jeromekelleher, line 194: Docstring would be helpful here.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,352 @@
...
+        return True
+
+    left, right = canon_left_right(tree, u)
+    if is_symmetrical(tree, u):
+        return min(tree.samples(left)) < min(tree.samples(right))
+
+    return True
+
+
+def nCr(n, r):
```
jeromekelleher, line 284: Huh, annoying that numpy doesn't provide a definition of this. Looks like python 3.8+ has the [comb](https://docs.python.org/3.8/library/math.html?highlight=comb#math.comb) function.

Perhaps we should follow its API, and look to drop our own definition at point in the future?

daniel-goldstein:
@jeromekelleher Thanks for the great comments! I also agree lists are nicer to reason about than explicitly left/right (thinking n-tensors instead of 2-d matrices). To clarify, do you mean that every node could have any number of children, or simply abstract this to n-ary trees, where the number of children for each node is a parameter?

(A small edge case to remember is that this is undefined for trees with unary nodes since you could have an infinite number of internal nodes for each configuration).

jeromekelleher:
> To clarify, do you mean that every node could have any number of children, or simply abstract this to n-ary trees, where the number of children for each node is a parameter?

I was thinking about the n-ary case, where we provide the list of children explicitly as a parameter.

jeromekelleher:
We don't have to worry about anything other than binary trees right now though, so feel free to drop in asserts on ``len(self.children) == 2`` etc.

daniel-goldstein:
Great, I think that's a fine change. Arbitrary lengths on the children of each internal node would be combinatorially even more explosive, and I like seeing my tests run for numbers in the teens! :smile:

python/tskit/combinatorics.py
daniel-goldstein: my thoughts as well

daniel-goldstein: my thoughts as well

python/tests/test_combinatorics.py
daniel-goldstein: This should in fact be "unlabelled". All trees that we're talking about here are "unordered" in that permuting children of a node does not change the rank of the tree.

daniel-goldstein:
@jeromekelleher Still got a couple clean up items but could you take another look? Things look very different now.

jeromekelleher:
This looks great @daniel-goldstein, I'll need to go through the details with you to fully get what's going on.

One high-level thing though: perhaps we should add ``Tree.rank()`` as a method of Tree and ``unrank`` as a static method of Tree? Then we don't need to expose the ``tskit.combinatorics`` sub module to the user at all. That is, all the functionality stays in this module, and we just add a thin layer to the Tree class to call it.

jeromekelleher:
One thing that's not clear to me without running the code: are you talking about integer partitions (4 = 1 + 1 + 1 + 1, 1 + 1 + 2, ...) or set partitions ({1, 2, 3, 4} = 1|2|3|4, 12|3|4,...)?

daniel-goldstein:
> One thing that's not clear to me without running the code: are you talking about integer partitions (4 = 1 + 1 + 1 + 1, 1 + 1 + 2, ...) or set partitions ({1, 2, 3, 4} = 1|2|3|4, 12|3|4,...)? 

Integer partitions. I tried to use the word partition only when splitting up the number of leaves between the sub-trees. There are sort of set partitions when assigning labels but I found it pretty confusing to use it in both cases.

jeromekelleher:
> Integer partitions.

Cool. Let's use the [rule_asc](http://jeromekelleher.net/category/combinatorics.html) algorithm for generating them then - I'm fairly confident it's [correct](http://jeromekelleher.net/downloads/k06.pdf).

daniel-goldstein:
Overhaul #3 and I'm feeling much better about the structure here. @jeromekelleher the API should now be set (only parameters are the rank tuple and number of leaves), and can handle arbitrary polytomies with only a slight performance cost. I confirmed the number of trees up to num_leaves=7 on [OEIS](https://oeis.org/search?q=0%2C+1%2C+1%2C+4%2C+26%2C+236%2C+2752%2C+39208&sort=&language=english&go=Search). Also left TODOs on obvious fix ups and improvements.

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,472 @@
+#
+# IT License
```
jeromekelleher, line 2:
```suggestion
# MIT License
```

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,472 @@
...
+        for n in range(11):
+            for k in range(n):
+                nums = list(range(n))
+                for rank, c in enumerate(itertools.combinations(nums, k)):
+                    c = list(c)
+                    self.assertEqual(comb.Combination.rank(c, nums), rank)
+                    self.assertEqual(comb.Combination.unrank(rank, nums, k), c)
+
+
+class TestPartition(unittest.TestCase):
```
jeromekelleher, line 219: Would be nice leave ``rule_asc`` in it's original form, and put in a few tests against known values for it (up to 6, say). Then test separately that the ``partitions`` function is equal to ``list(rule_asc(n))[:-1]``.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+# place that can't just accept it from the start.
+def partition(n):
+    """
+    Ascending integer partitions of n.
+    """
+    if n > 0:
+        yield from rule_asc(n)
+
+
+def rule_asc(n):
```
jeromekelleher, line 762: Let's leave rule_asc as it is, and do the dropping in your partition function. Should give some indication where this came from also, it'll look mysterious to readers otherwise.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+
+
+# TODO I think we can use part-count form everywhere. Right now
+# there's a janky work-around of grouping the partition when
+# we needed in part-count form but it doesn't look like there's any
+# place that can't just accept it from the start.
+def partition(n):
+    """
+    Ascending integer partitions of n.
+    """
```
jeromekelleher, line 757: How about 
```
# last partition is guaranteed to be length 1 in ascending order.
return itertools.takewhile(lamba a: len(a) > 1, rule_asc(n))
```

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+        y = a[k] - 1
+        k -= 1
+        while x <= y:
+            a[k] = x
+            y -= x
+            k += 1
+        a[k] = x + y
+        # TODO nicer way to avoid the partition [n]?
+        if k == 0:
+            return
```
jeromekelleher, line 777: Mixing yield and return is bad for your health!

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+
+def set_minus(arr, subset):
+    return [x for x in arr if x not in set(subset)]
+
+
+# TODO I think we can use part-count form everywhere. Right now
+# there's a janky work-around of grouping the partition when
+# we needed in part-count form but it doesn't look like there's any
+# place that can't just accept it from the start.
+def partition(n):
```
jeromekelleher, line 754: I think ``partitions()`` is a bit more logical.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+Module for ranking and unranking trees. Trees are considered only
+leaf-labelled and unordered, so order of children does not influence equality.
+"""
+import heapq
+from functools import lru_cache
+
+import tskit
+
+
+def rank(tree):
```
jeromekelleher, line 33: Do we need these functions when they're methods of RankTree really?

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,798 @@
...
+
+
+class RankTree:
+    """
+    A tree class that maintains the topological ranks of each node in the tree.
+    This structure can be used to efficiently compute the rank of a tree of
+    n leaves and produce a tree given a rank.
+    """
+
+    def __init__(self, shape_rank=None, label_rank=None, children=None, label=None):
```
jeromekelleher, line 98: The boundary conditions don't feel quite right to me, and the behaviour seems quite complicated. How are you constructing RankTrees in practise? It's seems ugly that you have all these special cases for n = 1, when that should just fall out if the conditions have been set up properly. It would be good to make a single constructor with mandatory arguments.

daniel-goldstein: Ah yes the `n = 1` cases seemed to mostly be residue from my recursive implementations and were pretty easily resolved. The constructor is pretty messy and I found myself having two cases of constructing RankTrees that seemed at odds regarding the ranks (children is easy to make mandatory). When going from a tskit Tree, we need to compute the ranks as we construct the nodes from leaves to root. But when unranking, we already know the ranks of the children as part of the unranking process. One way to avoid this mess entirely could be to just make the ranks lazy, cached properties.

daniel-goldstein: The reason for doing it this way is that in order to label the trees you need the full tree of shape ranks, which I was trying to avoid recomputing.

python/tskit/combinatorics.py
daniel-goldstein: That's fair. I can consolidate the docstrings with the methods in `tskit.Tree`

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+        tables = tskit.TableCollection(seq_length)
+
+        def add_node(node):
+            if node.is_leaf():
+                assert node.label is not None
+                return node.label
+
+            child_ids = [add_node(child) for child in node.children]
+            # Arbitrarily set parent time +1 from their oldest child
+            max_child_time = max(tables.nodes.time[c] for c in child_ids)
```
jeromekelleher, line 252:
```suggestion
            max_child_time = max(node_time[c] for c in child_ids)
```

Making tables and arrays is expensive - they're actually all copies at the moment.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+            for child_id in child_ids:
+                tables.edges.add_row(0, seq_length, parent_id, child_id)
+
+            return parent_id
+
+        for _ in range(self.num_leaves):
+            tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        add_node(self)
+
+        # Have to sort for now because the of the first way
```
jeromekelleher, line 263: This comment doesn't quite parse.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+
+        return group_by(self.children, same_num_leaves)
+
+    def group_children_by_shape(self):
+        def same_shape(c1, c2):
+            return c1.num_leaves == c2.num_leaves and c1.shape_rank() == c2.shape_rank()
+
+        return group_by(self.children, same_shape)
+
+    def equal(self, other):
```
jeromekelleher, line 301: Could consider overriding ``__eq__`` and ``__neq__`` here. Seems like a good definition of equality.

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+
+# TODO This is called repeatedly in ranking and unranking and has a perfect
+# subtructure for DP. It's only every called on n in [0, num_leaves]
+# so we should compute a vector of those results up front instead of using
+# repeated calls to this function.
+# Put an lru_cache on for now as a quick replacement (cuts test time down by 80%)
+@lru_cache()
+def num_shapes(n):
+    """
+    The cardinality of the set of unlabelled binary trees with n leaves,
```
jeromekelleher, line 347: Is the "binary" bit still true?

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+    Ascending integer partitions of n, excluding the partition [n].
+    Since trees with unary nodes are uncountable, the partition of
+    leaves must be at least size two.
+    """
+    if n > 0:
+        # last partition is guaranteed to be length 1.
+        yield from itertools.takewhile(lambda a: len(a) > 1, rule_asc(n))
+
+
+def rule_asc(n):
```
jeromekelleher, line 691: A quick comment just saying where this is from would help. You need to think about people who are not me reading the code - help them out.

python/tests/test_combinatorics.py
```diff
@@ -0,0 +1,524 @@
...
+"""
+import itertools
+import unittest
+
+import tskit
+import tskit.combinatorics as comb
+from tskit.combinatorics import RankTree
+
+
+def all_labelled_trees(n):
```
jeromekelleher, line 34: Let's move these two functions into the ``combinatorics`` module and document them. They'll be really useful when explaining how this stuff works, and they're just generally cool.

I guess they should return a tskit.Tree instead of a RankTree though.

daniel-goldstein: Since we have to go through `RankTree` anyway, how about another generator function wrapper that converts `RankTree` to a tskit Tree? We can keep this name for the function that returns tskit Trees and use the other internally as this is currently being used.

python/tskit/combinatorics.py
daniel-goldstein: Nope.

python/tests/test_combinatorics.py
jeromekelleher: SGTM!

python/tskit/combinatorics.py
daniel-goldstein: 😮

daniel-goldstein: Not that I want to be making copies all the time, but doesn't this mean that `node_time` will remain empty?

python/tskit/combinatorics.py
daniel-goldstein: Preference for a link to put here?

jeromekelleher: http://jeromekelleher.net/generating-integer-partitions.html

python/tskit/combinatorics.py
```diff
@@ -0,0 +1,723 @@
...
+    @staticmethod
+    def from_tsk_tree(tree):
+        if len(tree.roots) != 1:
+            raise ValueError("Can't rank trees with multiple roots")
+
+        return RankTree.from_tsk_tree_node(tree, tree.root)
+
+    def to_tsk_tree(self):
+        seq_length = 1
+        tables = tskit.TableCollection(seq_length)
```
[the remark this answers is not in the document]
jeromekelleher: Ah, this is wrong. Ignore!

python/tests/test_combinatorics.py
daniel-goldstein: Ended up exposing three functions: all trees, all tree shapes, and all labellings of a given tree

daniel-goldstein:
@jeromekelleher Addressed and rebased

jeromekelleher:
Merging! This is amazing stuff, thanks @daniel-goldstein!

PR #515, opened 2020-04-07
Pin sphinx version

benjeffery:
Sphinx have just done a major release with lots of breaking changes - `breathe` has not yet caught up so we need to pin our sphinx version.

### DIFF
--- a/python/requirements/development.txt
+++ b/python/requirements/development.txt
@@ -17,7 +17,7 @@ pyparsing
 pysam
 python_jsonschema_objects
 PyVCF
-sphinx
+sphinx==2.4.4 #Pinned as breathe doesn't support 3.0 yet
 sphinx-argparse
 sphinx-issues
 sphinx_rtd_theme


jeromekelleher:
Thanks, good catch.

brianzhang01:
When I run `pip install -r python/requirements/development.txt` from a fresh anaconda environment with Python 3.7, I get:

ERROR: breathe 4.15.0 has requirement Sphinx>=3.0, but you'll have sphinx 2.4.4 which is incompatible.

Based on the discussion, I'm guessing I should ignore the message.

jeromekelleher:
Thanks @brianzhang01. 

Looks like breath has caught up @benjeffery? This will probably start cropping up in CI soon.

benjeffery:
Thanks for the report @brianzhang01 I'll dig into this now and get it all consistent.

benjeffery:
Some of our docs are not Sphinx3 compatible so this will take a bit longer than just unpinning the version.

jeromekelleher:
Sounds good. Perhaps we should just pin Breathe for now and return in a month or two when the dust has settled a bit?

benjeffery:
Yep, I dug into it to see if it was a small thing, but seems not. So now trying to pin `breathe` but still getting errors that I wasn't before !@>!>?%??.

benjeffery:
Completely re-cloning `tskit` fixed everything - so I had some odd state somewhere.
https://github.com/tskit-dev/tskit/pull/521

PR #516, opened 2020-04-07
Pin all dependencies in CI

benjeffery:
This PR makes the CI more deterministic by pinning all python dependencies. The only downside of this is that we will not be made aware when a new version of a dependency breaks things for users who aren't using pinned versions. The current system isn't great though as we get notified by CI breaking in unrelated PRs. We can ameliorate this by getting notification PRs with https://dependabot.com/ which I'll integrate in another PR.


jeromekelleher:
OK, let's do it, thanks.

PR #517, opened 2020-04-07
Add Dependabot

benjeffery:
Creates PRs for new versions of dependencies as they are published. Needs #516 to be merged first.


jeromekelleher:
Looks good. Do we need to set up something for dependabot? It's worth noting that this just handles the pip deps - conda dependencies can still break CI (probably not worth worrying about, just noting)

benjeffery:
You should have got a request on github to authorise the dependabot application. You're right dependabot doesn't (yet) cover conda. Will add when they do.

jeromekelleher:
I'm not seeing any messages from dependabot. Can you look into setting up linkage please? You should have the required access.

PR #522, opened 2020-04-09
Update setup.py to py3.6 minimum version

### LINKED ISSUE #520: Update setup.py to confirm python requirements
Setup.py currently states that we need python >= 3.4, but we've recently started using f-strings. 

Since Debian stable is currently on 3.7 and f-strings are pretty useful, I think we're OK to push our requirements up to 3.6. Happy to be convinced otherwise, although actually making sure that older versions of Python are supported is not trivial and pushes our CI requirements up a lot.

benjeffery:
Fixes #520 
I've also changed the travis config to test 3.7 and 3.8. I think the chance of breakage is _very_ low, but only cost is some travis CPU time as they run in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ language: python
 matrix:
   include:
     - os: linux
-      python: 3.6
+      python: 3.7
     - os: osx
       language: generic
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Dunno about getting the extra Python versions on Travis, this is going to really slow down CI (it's already the bottleneck most of the time)

benjeffery:
The travis docs made it seem like these run in parallel?  I've backed out that change. 3.8 failed due to `pysam` anyway.

PR #523, opened 2020-04-09
Add readme badges

### LINKED ISSUE #469: GitHub landing pages for Python and C
The current GitHub readme contains generic information about the tskit project. It would be nice if we also had specific landing pages for the C and Python libraries (i.e., README.md files), with links to these from the top-level README.

These should include any relevant badges (i.e, Black format, download counts, CI badges etc for Python; I guess just the Circle CI badge for C).

benjeffery:
Closes #469
Apparently appveyor needs a key from their settings page, so have skipped it for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,30 @@
 # tskit  <img align="right" width="145" height="90" src="https://github.com/tskit-dev/administrative/blob/master/tskit_logo.svg">
 
-[![PyPI version](https://img.shields.io/pypi/v/tskit.svg)](https://pypi.org/project/tskit/)
-[![Supported Python Versions](https://img.shields.io/pypi/pyversions/tskit.svg)](https://pypi.org/project/tskit/)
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Might be nice to split some of these out into the Python subdirectory, as this is the root of the Python package. Stuff like downloads all refers to the Python package.

jeromekelleher:
Is it a good idea to split these badges up between the C and Python directories (with links in the top-level README to these READMEs, I guess), or should we put it all into different sections of the top level README do you think @benjeffery?

I can imagine having badges that only apply to the C code in there at some point too, say things like coverity scan.

benjeffery:
@jeromekelleher I've updated this - I think it might be enough to close #469. Most people will land on the github front page first so I think putting both C and python there is better than having to click through.

jeromekelleher:
Yes, good call, thanks.

PR #530, opened 2020-04-11
Do not use kastore memory for columns.

### LINKED ISSUE #506: [C API] Copy columns out of kastore
Currently when a tsk_table_collection is read from file, it uses the kastore object as the backing memory. That is, we have a ``self->store`` pointer which we maintain for the lifetime of the table collection, and the table columns all point directly to the store's memory. This setup was motivated by a desire to use mmap within the kastore and therefore have zero copy semantics all the way down to the kernel for column memory. Turns out this isn't worth the complexity, as mmap use has a few major caveats and the files just aren't large enough to make this sort of optimisation worthwhile.

So, we should move to copying the columns out of the store. We should set this up so that we read the  columns into a set of local pointers, and then call (e.g.) ``tsk_node_table_set_columns`` with those pointers as arguments. This will take care of all the allocation behaviour, and make sure that the NULL column semantics accepted for set_columns is directly reflected also in the default columns from file. It's probably a straightforward change where we provide pointers to local variables rather than the table columns to ``read_table_cols``, and then call ``set_columns`` on those.

### LINKED ISSUE #527: tree sequence with edges has no edges on load
With the current HEAD, it seems that SLiM-produced tree sequences - which have edges - are not loaded with edges. For instance, running [this SLiM script](https://raw.githubusercontent.com/tskit-dev/pyslim/master/tests/examples/recipe_nonWF.slim):
```
wget https://raw.githubusercontent.com/tskit-dev/pyslim/master/tests/examples/recipe_nonWF.slim
slim recipe_nonWF.slim
tskit info recipe_nonWF.trees
```
tells me there are no edges:
```
sequence_length:  100.0
trees:            1
samples:          28
individuals:      24
nodes:            58
edges:            0
sites:            33
mutations:        40
migrations:       0
populations:      2
provenances:      1
```
but there **should** be, and kastore confirms it:
```
kastore dump recipe_nonWF.trees edges/parent | wc -l
# 93
```

I bisected this, and found that this got introduced somewhere in these commits:
```
f53ff61344dbe15184a065daf43600623c286dea
d73d85a3c1d8c5c2bc4f4a0174dd5b85fb8834a9
fa35ef554aa43290f1d021ef6792c0ad85be8baf
0161fb30ffb4f4a65a8240da004a374a2fb743fe
4b285a5a70b71ff1fa63c9c12f3044ab0d3abafb
cbeb1297962400a1004b519da78ed4aa5785d163
```

I can't tell *which* of these commits, though, because trying to load the tree sequence at any of these commits results in this error:
```
$ tskit info recipe_nonWF.trees 
  File "/home/peter/.local/bin/tskit", line 11, in <module>
    load_entry_point('tskit==0.2.4.dev0', 'console_scripts', 'tskit')()
  File "/home/peter/.local/lib/python3.7/site-packages/tskit-0.2.4.dev0-py3.7-linux-x86_64.egg/tskit/cli.py", line 281, in tskit_main
    args.runner(args)
  File "/home/peter/.local/lib/python3.7/site-packages/tskit-0.2.4.dev0-py3.7-linux-x86_64.egg/tskit/cli.py", line 54, in run_info
    ts = load_tree_sequence(args.tree_sequence)
  File "/home/peter/.local/lib/python3.7/site-packages/tskit-0.2.4.dev0-py3.7-linux-x86_64.egg/tskit/cli.py", line 48, in load_tree_sequence
    return tskit.load(path)
  File "/home/peter/.local/lib/python3.7/site-packages/tskit-0.2.4.dev0-py3.7-linux-x86_64.egg/tskit/trees.py", line 1805, in load
    return TreeSequence.load(path)
  File "/home/peter/.local/lib/python3.7/site-packages/tskit-0.2.4.dev0-py3.7-linux-x86_64.egg/tskit/trees.py", line 2385, in load
    ts.load(str(path))
_tskit.LibraryError: Bad offset provided in input array
```
So, my guess is that f53ff61344dbe15184a065daf43600623c286dea introduced this "bad offset" bug, and cbeb1297962400a1004b519da78ed4aa5785d163 fixed it but introduced a different bug.

I've inserted some `fprintf()`s into `tsk_edge_table_load` and `read_table_cols`, and verified that at [this point](https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L107) there are in fact edges being read in. I haven't gone further down the path to figure out where things are going wrong, though.

I know this is HEAD, but it's still not good to be silently getting bad tree sequences.

### LINKED ISSUE #528: [C API] Missing error check on table_x_copy
e.g. https://github.com/tskit-dev/tskit/blob/5f58b6186993cc5e635c621ee6e7c7a62aab8649/c/tskit/tables.c#L1410-L1413

jeromekelleher:
Seemed simplest to tackle this head on, as it's complicating quite a lot of what we're doing. It *should* close #527 (there's a test in here to verify it), but we really should get some files in from earlier file minor versions to make sure we're doing the right thing.

Closes #528
Closes #527
Closes #506

Not ready for merging yet though, as it needs some more C test coverage and there's something weird going on in one of the tests that needs figuring out.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -154,6 +154,10 @@ check_offsets(size_t num_rows, tsk_size_t *offsets, tsk_size_t length, bool chec
     int ret = TSK_ERR_BAD_OFFSET;
     size_t j;
 
+    if (offsets == NULL) {
+        return 0;
+    }
+
     if (offsets[0] != 0) {
         goto out;
     }
@@ -371,6 +375,9 @@ tsk_individual_table_copy(
     }
     ret = tsk_individual_table_set_columns(dest, self->num_rows, self->flags,
         self->location, self->location_offset, self->metadata, self->metadata_offset);
+    if (ret != 0) {
+        goto out;
+    }
     ret = tsk_individual_table_set_metadata_schema(
         dest, self->metadata_schema, self->metadata_schema_length);
 out:
@@ -543,13 +550,11 @@ out:
 int
 tsk_individual_table_free(tsk_individual_table_t *self)
 {
-    if (self->max_rows > 0) {
-        tsk_safe_free(self->flags);
-        tsk_safe_free(self->location);
-        tsk_safe_free(self->location_offset);
-        tsk_safe_free(self->metadata);
-        tsk_safe_free(self->metadata_offset);
-    }
+    tsk_safe_free(self->flags);
+    tsk_safe_free(self->location);
+    tsk_safe_free(self->location_offset);
+    tsk_safe_free(self->metadata);
+    tsk_safe_free(self->metadata_offset);
     tsk_safe_free(self->metadata_schema);
     return 0;
 }
@@ -730,24 +735,42 @@ static int
 tsk_individual_table_load(tsk_individual_table_t *self, kastore_t *store)
 {
     int ret = 0;
+    tsk_flags_t *flags = NULL;
+    double *location = NULL;
+    tsk_size_t *location_offset = NULL;
+    char *metadata = NULL;
+    tsk_size_t *metadata_offset = NULL;
     char *metadata_schema = NULL;
-    tsk_size_t metadata_schema_length;
+    tsk_size_t num_rows, location_length, metadata_length, metadata_schema_length;
 
     read_table_col_t read_cols[] = {
-        { "individuals/flags", (void **) &self->flags, &self->num_rows, 0, KAS_UINT32,
+        { "individuals/flags", (void **) &flags, &num_rows, 0, KAS_UINT32, 0 },
+        { "individuals/location", (void **) &location, &location_length, 0, KAS_FLOAT64,
             0 },
-        { "individuals/location", (void **) &self->location, &self->location_length, 0,
-            KAS_FLOAT64, 0 },
-        { "individuals/location_offset", (void **) &self->location_offset,
-            &self->num_rows, 1, KAS_UINT32, 0 },
-        { "individuals/metadata", (void **) &self->metadata, &self->metadata_length, 0,
-            KAS_UINT8, 0 },
-        { "individuals/metadata_offset", (void **) &self->metadata_offset,
-            &self->num_rows, 1, KAS_UINT32, 0 },
+        { "individuals/location_offset", (void **) &location_offset, &num_rows, 1,
+            KAS_UINT32, 0 },
+        { "individuals/metadata", (void **) &metadata, &metadata_length, 0, KAS_UINT8,
+            0 },
+        { "individuals/metadata_offset", (void **) &metadata_offset, &num_rows, 1,
+            KAS_UINT32, 0 },
         { "individuals/metadata_schema", (void **) &metadata_schema,
             &metadata_schema_length, 0, KAS_UINT8, TSK_COL_OPTIONAL },
     };
+
     ret = read_table_cols(store, read_cols, sizeof(read_cols) / sizeof(*read_cols));
+    if (ret != 0) {
+        goto out;
+    }
+    ret = check_offsets(num_rows, location_offset, location_length, true);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = check_offsets(num_rows, metadata_offset, metadata_length, true);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_individual_table_set_columns(
+        self, num_rows, flags, location, location_offset, metadata, metadata_offset);
     if (ret != 0) {
         goto out;
     }
@@ -884,6 +907,9 @@ tsk_node_table_copy(tsk_node_table_t *self, tsk_node_table_t *dest, tsk_flags_t 
     }
     ret = tsk_node_table_set_columns(dest, self->num_rows, self->flags, self->time,
         self->population, self->individual, self->metadata, self->metadata_offset);
+    if (ret != 0) {
+        goto out;
+    }
     ret = tsk_node_table_set_metadata_schema(
         dest, self->metadata_schema, self->metadata_schema_length);
 out:
@@ -1034,14 +1060,12 @@ out:
 int
 tsk_node_table_free(tsk_node_table_t *self)
 {
-    if (self->max_rows > 0) {
-        tsk_safe_free(self->flags);
-        tsk_safe_free(self->time);
-        tsk_safe_free(self->population);
-        tsk_safe_free(self->individual);
-        tsk_safe_free(self->metadata);
-        tsk_safe_free(self->metadata_offset);
-    }
+    tsk_safe_free(self->flags);
+    tsk_safe_free(self->time);
+    tsk_safe_free(self->population);
+    tsk_safe_free(self->individual);
+    tsk_safe_free(self->metadata);
+    tsk_safe_free(self->metadata_offset);
     tsk_safe_free(self->metadata_schema);
     return 0;
 }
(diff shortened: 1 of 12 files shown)


benjeffery:
This looks great, so much simpler to use `set_columns` now we're copying out of `kastore`

I've installed SLiM and confirmed that @petrelharp 's example now works for this branch, and understand how zero edges was happening before (think I should get a `if (ret != 0)` tattoo as a penance).

petrelharp:
Oh!! Thanks for the reminder, I think the code I am working on needs a `ret !=0` or two yet.

benjeffery:
The failing test is fixed by removing `TSK_NO_INIT_TABLES` in https://github.com/tskit-dev/tskit/pull/530/commits/b2e7536c4b09532ce4504e2a655254f479c32abc 

Is there any value in that flag anymore? I can't see it used anywhere else in the code, and now that `load_table` is using `set_columns` it really needs an `init`'d table.

I've also formatted the C to make the CI happy in https://github.com/tskit-dev/tskit/pull/530/commits/8ae100ccb7e2b372b9c757090e5c597013032ded

jeromekelleher:
Awesome, thanks @benjeffery! I'll finish this up first thing tomorrow morning.

jeromekelleher:
Some more updates here which should push the test coverage in C right up, as well as put some infrastructure in place for making all the "x, x_offset" columns optional. I still need to add some tests for handling the input of bad offsets, but it'll be ready for final review and merge then.

benjeffery:
Great, I think making the "both columns" error explicit is a good choice. 

Speaking of coverage - is it worth getting the coverage to ignore some of the lines that we can't easily hit like OOMs? I'd like to enforce 100% coverage then.

jeromekelleher:
> Speaking of coverage - is it worth getting the coverage to ignore some of the lines that we can't easily hit like OOMs? I'd like to enforce 100% coverage then.

If there's an easy way to do this, then great. I'm wary about *enforcing* 100% test coverage as it leads to weird code and tests sometimes, but having more insight into the catchable errors would be good.

It is *possible* to test all the OOM corner cases (I've done it for [kastore](https://github.com/tskit-dev/kastore/blob/master/c/malloc_tests.c)) but it's pretty tedious.

c/tskit/tables.c
```diff
@@ -154,6 +157,10 @@ check_offsets(size_t num_rows, tsk_size_t *offsets, tsk_size_t length, bool chec
     int ret = TSK_ERR_BAD_OFFSET;
     size_t j;
 
+    if (offsets == NULL) {
```
jeromekelleher, line 160: Not entirely happy with this - either refactor to put in a comment explaining why the early exit is needed.

benjeffery: > It is _possible_ to test all the OOM corner cases (I've done it for [kastore](https://github.com/tskit-dev/kastore/blob/master/c/malloc_tests.c)) but it's pretty tedious.

Doesn't look _too_ bad - it feels like this should be part of CUnit. It could run your test incrementing `fail_at` and asserting that a `CU_ASSERT_EQUAL_FATAL(ret, 0);` fails each time.

jeromekelleher: > Doesn't look too bad - it feels like this should be part of CUnit. It could run your test incrementing fail_at and asserting that a CU_ASSERT_EQUAL_FATAL(ret, 0); fails each time.

It would be good to have if you want to follow it up - it does need some special linkage wizardry to work too though. But I guess we could have a ``malloc_tests`` file? We don't actually to that much mallocing, which is a bonus.

jeromekelleher:
OK, I think this is ready to go. Can you have a look over please @benjeffery?

We probably won't be able to hit the test coverage threshold as there were a few incidental changes that have a lot of uncatchable errors, but I think we're covering the file format stuff pretty well now.

I hope you don't mind me squashing your "fixup!" commits, but I figured you wouldn't since they were marked as fixups.

c/tskit/tables.c
```diff
@@ -6931,42 +6996,38 @@ tsk_table_collection_dump_indexes(tsk_table_collection_t *self, kastore_t *store
...
+    if (edge_insertion_order != NULL) {
+        if (num_rows != self->edges.num_rows) {
+            ret = TSK_ERR_FILE_FORMAT;
+            goto out;
+        }
+        ret = tsk_table_collection_set_index(
+            self, edge_insertion_order, edge_removal_order);
+        if (ret != 0) {
+            goto out;
+        }
```
benjeffery, line 7030: These lines have no effect. Is their aim to prevent an unchecked `ret` if code is added after at a later date? The pattern in the rest of the file is not to have these, but I can see an argument for them.

c/tskit/tables.c
```diff
@@ -6977,71 +7038,62 @@ tsk_table_collection_load(
...
-    ret = tsk_provenance_table_load(&self->provenances, self->store);
+    ret = tsk_provenance_table_load(&self->provenances, &store);
     if (ret != 0) {
         goto out;
     }
-    ret = tsk_table_collection_load_indexes(self);
+    ret = tsk_table_collection_load_indexes(self, &store);
     if (ret != 0) {
         goto out;
     }
```
benjeffery, line 7093: Same thing here.

c/tskit/tables.c
```diff
@@ -6977,71 +7038,62 @@ tsk_table_collection_load(
...
-    ret = tsk_table_collection_check_offsets(self);
 out:
+    /* Any read errors will have already happened so we ignore any errors here. */
+    kastore_close(&store);
     return ret;
```
benjeffery, line 7097: I don't quite get this - in the case where there have been no read errors we're ignoring errors from `kastore_close`, which can still return errors, even when the mode is `KAS_READ`.

c/tskit/tables.c
```diff
@@ -6931,42 +6996,38 @@ tsk_table_collection_dump_indexes(tsk_table_collection_t *self, kastore_t *store
...
-        ret = has_insertion;
+    ret = read_table_cols(store, read_cols, sizeof(read_cols) / sizeof(*read_cols));
+    if (ret != 0) {
         goto out;
     }
-    has_removal = kastore_containss(self->store, removal_order);
-    if (has_removal < 0) {
-        ret = has_removal;
+    if ((edge_insertion_order == NULL) != (edge_removal_order == NULL)) {
+        ret = TSK_ERR_FILE_FORMAT;
```
benjeffery, line 7018: Shouldn't this be `TSK_ERR_BOTH_COLUMNS_REQUIRED`?

c/tskit/tables.c
```diff
@@ -3553,6 +3589,9 @@ tsk_population_table_print_state(tsk_population_table_t *self, FILE *out)
         (int) self->metadata_length, (int) self->max_metadata_length,
         (int) self->max_metadata_length_increment);
     fprintf(out, TABLE_SEP);
+    fprintf(out, "#metadata_schema#\n");
+    fprintf(out, "%.*s\n", self->metadata_schema_length, self->metadata_schema);
+    fprintf(out, "#end#metadata_schema\n");
```
benjeffery, line 3594: Ouch, well spotted. I clearly need some tests to check the round tripping of schemas.

c/tskit/tables.c
```diff
@@ -6592,44 +6662,42 @@ tsk_table_collection_print_state(tsk_table_collection_t *self, FILE *out)
 }
 
 int TSK_WARN_UNUSED
-tsk_table_collection_init(tsk_table_collection_t *self, tsk_flags_t options)
+tsk_table_collection_init(tsk_table_collection_t *self, tsk_flags_t TSK_UNUSED(options))
```
benjeffery, line 6665: Is the method signature kept the same for backwards compatibility? Or in case we need flags in future?

petrelharp: Both maybe, but more the latter.

c/tskit/tables.c
jeromekelleher: Well this is just a print_state which is a debugging tool, so I wouldn't worry too much about it.

c/tskit/tables.c
jeromekelleher: I'm glad you brought this up! I was thinking about this, and I think we should probably have explicit ret != 0 checks after every call even if it's at the end of the function. There's two good reasons for this:

1. As you say, if code gets added later we won't accidentally skip the check (and it's really easy to miss this in review, as the diffs often won't show enough context to spot the problem)
2. Also it helps us with figuring out where test coverage is missing. If we're testing error conditions on every other call except the last one, we could miss testing some important error conditions.

So, even though it's obviously stupid, I think it's the right thing to do. It might result in a tiny bit more code (but maybe not: the compiler might spot that it's redundant), but it's not important. Since you're in favour, I'm going to loop back and make the change on all the code I've just updated now.

c/tskit/tables.c
jeromekelleher: Hmm, yes. You're right, I'll update.

c/tskit/tables.c
```diff
@@ -116,13 +115,8 @@ read_table_cols(kastore_t *store, read_table_col_t *read_cols, size_t num_cols)
                 goto out;
             }
         } else if (!(read_cols[j].options & TSK_COL_OPTIONAL)) {
-            ret = TSK_REQUIRED_COL_NOT_FOUND;
+            ret = TSK_ERR_REQUIRED_COL_NOT_FOUND;
             goto out;
-        } else {
```
jeromekelleher, line 121: @benjeffery, I changed the semantics here slightly because we were getting inconsistent behaviour when the first column in a list was optional (so, everything else was expected to have 0 rows). Since we're already setting default values for the incoming columns, it seemed best to just use those values as the indicator and not to use the lengths at all. That's why we're now checking to see if metadata schema was set, rather than passing NULL, -1 through to set_metadata_schema.

jeromekelleher: I don't like this method though, it's too tricky. I think #537 will make things a lot simpler.

petrelharp:
Fantastic! :yay:

PR #531, opened 2020-04-12
Update miniconda URLs

benjeffery:
I'm attempting to fix travis builds that are failing for all PRs, trying out a few things here.


benjeffery:
Linux CI uses `wget`, OSX CI uses `curl`. `wget` was follwing the `301` redirect to the new miniconda location, `curl` was not! I've updated the URL for both.

PR #539, opened 2020-04-15
add sphinx viewcode ext

petrelharp:
This adds links in the documentation to the source code (not to github but to another page rendered in the same way containing the highlighted code). This is not so useful for functions that go directly to C but is useful for other functions: see below.  I tried this out because @mufernando mentioned he uses this feature in the scikit-allel docs all the time, why didn't we have it?  It seems to work.
Screenshot from 2020-04-15 10-41-34
Screenshot from 2020-04-15 10-41-42

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -119,6 +119,7 @@ TypedField.make_field = patched_make_field
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.intersphinx",
+    "sphinx.ext.viewcode",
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Very nice!

It's part of sphinx so shouldn't have any issues. I say we merge it and see how RTD likes it. We can always remove it if it breaks something.

benjeffery:
This is awesome - especially like that the source then links back to docs.

PR #548, opened 2020-04-20
Implement incremental KC Distance on tree sequences

daniel-goldstein:
A high-level implementation of the KC distance on Tree sequences. The tree sequence extension of KC walks a long both tree sequences, taking the KC distance between corresponding trees. The distance is the sum of the individual tree KC distances weighted by the fraction of the sequence spanned by the pair of trees.

The KC distance focuses on pairs of leaves, specifically the depth and time-from-root for the mrca of a given pair of leaves. To maintain the depth and time vectors used in KC incrementally, for each edge `e`, we must do an upward and downward traversal. While traversing up toward the root, we update the pairs of leaves where one leaf is in the subtree affected by `e` and one is not. Traversing down from `e`, we update all pairs of leaves where both leaves are in the subtree. Pairs where both leaves are outside of the subtree under `e` haven't been affected by the insertion/removal of that edge.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -3565,11 +3565,24 @@ tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_lengt
     return 0;
 }
 
+/* u must be a valid node in the tree. For internal use */
+static tsk_size_t
+tsk_tree_get_depth(tsk_tree_t *self, tsk_id_t u)
+{
+
+    tsk_id_t v;
+    tsk_size_t depth = 0;
+
+    for (v = self->parent[u]; v != TSK_NULL; v = self->parent[v]) {
+        depth++;
+    }
+
+    return depth;
+}
+
 int TSK_WARN_UNUSED
 tsk_tree_depth(tsk_tree_t *self, tsk_id_t u, tsk_size_t *depth_ret)
 {
-    tsk_id_t v;
-    tsk_size_t depth;
     int ret = 0;
 
     ret = tsk_tree_check_node(self, u);
@@ -3577,14 +3590,31 @@ tsk_tree_depth(tsk_tree_t *self, tsk_id_t u, tsk_size_t *depth_ret)
         goto out;
     }
 
-    depth = 0;
-    for (v = self->parent[u]; v != TSK_NULL; v = self->parent[v]) {
-        depth++;
-    }
-
-    *depth_ret = depth;
+    *depth_ret = tsk_tree_get_depth(self, u);
 out:
     return ret;
+}
+
+static int
+tsk_tree_node_root(tsk_tree_t *self, tsk_id_t u)
+{
+    tsk_id_t v = u;
+    while (self->parent[v] != TSK_NULL) {
+        v = self->parent[v];
+    }
+
+    return v;
+}
+
+/* TODO Should push this through the stack since this is a python method and
+ * it's here anyway */
+static double
+tsk_tree_branch_length(tsk_tree_t *self, tsk_id_t u)
+{
+    double *times = self->tree_sequence->tables->nodes.time;
+    tsk_id_t parent = self->parent[u];
+
+    return parent == TSK_NULL ? 0 : times[parent] - times[u];
 }
 
 static void
@@ -4331,12 +4361,11 @@ out:
     return ret;
 }
 
-int TSK_WARN_UNUSED
+int
 tsk_diff_iter_free(tsk_diff_iter_t *self)
 {
-    int ret = 0;
     tsk_safe_free(self->edge_list_nodes);
-    return ret;
+    return 0;
 }
 
 void
@@ -4447,7 +4476,7 @@ tsk_diff_iter_next(tsk_diff_iter_t *self, double *ret_left, double *ret_right,
  * ======================================================== */
 
 typedef struct {
-    int *m;
+    tsk_size_t *m;
     double *M;
     tsk_id_t n;
     tsk_id_t N;
@@ -4460,7 +4489,7 @@ kc_vectors_alloc(kc_vectors *self, tsk_id_t n)
 
     self->n = n;
     self->N = (n * (n - 1)) / 2;
-    self->m = calloc((size_t)(self->N + self->n), sizeof(int));
+    self->m = calloc((size_t)(self->N + self->n), sizeof(tsk_size_t));
     self->M = calloc((size_t)(self->N + self->n), sizeof(double));
     if (self->m == NULL || self->M == NULL) {
         ret = TSK_ERR_NO_MEMORY;
@@ -4491,7 +4520,7 @@ update_kc_vectors_single_leaf(
 
 static inline void
 update_kc_vectors_all_pairs(tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u,
-    tsk_id_t v, int depth, double time)
+    tsk_id_t v, tsk_size_t depth, double time)
 {
     tsk_id_t leaf1_index, leaf2_index, leaf1, leaf2, n1, n2, tmp, pair_index;
     tsk_treeseq_t *ts = tree->tree_sequence;
@@ -4500,7 +4529,7 @@ update_kc_vectors_all_pairs(tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u,
     const tsk_id_t *restrict right_sample = tree->right_sample;
     const tsk_id_t *restrict next_sample = tree->next_sample;
     const tsk_id_t *restrict sample_index_map = ts->sample_index_map;
-    int *restrict kc_m = kc_vecs->m;
+    tsk_size_t *restrict kc_m = kc_vecs->m;
     double *restrict kc_M = kc_vecs->M;
 
     leaf1_index = left_sample[u];
@@ -4536,18 +4565,19 @@ update_kc_vectors_all_pairs(tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u,
     }
 }
 
+struct kc_stack_elmt {
+    tsk_id_t node;
+    tsk_size_t depth;
+};
+
 static int
 fill_kc_vectors(tsk_tree_t *t, kc_vectors *kc_vecs)
 {
-    struct stack_elmt {
-        tsk_id_t node;
-        int depth;
-    };
-
-    int stack_top, depth;
+    int stack_top;
+    tsk_size_t depth;
     double time;
     const double *times;
-    struct stack_elmt *stack;
+    struct kc_stack_elmt *stack;
     tsk_id_t root, u, c1, c2;
     int ret = 0;
     tsk_treeseq_t *ts = t->tree_sequence;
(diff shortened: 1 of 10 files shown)


python/tests/test_topology.py
```diff
@@ -904,6 +903,365 @@ def test_internal_samples(self):
...
 
 
+class KCIncrementalTree(tsutil.SampleListTree):
+    """
+    Tracks the samples of each node in the tree and provides
+    convenience methods for incrementally updating values used
+    in KC distance.
+    """
+
+    def __init__(self, tree_sequence):
```
jeromekelleher, line 913: This is superfluous, isn't it? I.e., we can drop the init entirely.

python/tests/test_topology.py
```diff
@@ -904,6 +903,365 @@ def test_internal_samples(self):
...
+
+    _, edges_out, edges_in = edge_diffs
+    # Reset information about subtrees that were removed.
+    # Not strictly necessary since these values should all get
+    # overriden in the updates later, but good for making sure
+    # everything is being updated correctly.
+    for e in edges_out:
+        if inc_tree.is_sample(e.child):
+            reset_single_leaf(e.child)
+        inc_tree.visit_in_subtree_pairs(e.child, reset_pair)
```
jeromekelleher, line 1024: Clever, I like this structure of passing in the visit func. Nice approach.

daniel-goldstein:
I agree option 2 seems the better way forward. It saves a lot of duplication and would at the very least get most of the C implementation done before fine-tuning. Same goes for the perf, let’s hold off until it’s in C.

daniel-goldstein:
@jeromekelleher Changed to use the trees and edge diffs iterator at the same time. Do you want to get this in before moving on to the C implementation?

python/tests/test_topology.py
```diff
@@ -904,6 +903,398 @@ def test_internal_samples(self):
...
+                    for in_leaf in tree.leaves(edge_c):
+                        time = root_time - tree.time(p)
+                        update_kc_vectors_pair(
+                            kc, out_leaf, in_leaf, depth, time, sample_index_map
+                        )
+        c, p = p, tree.parent(p)
+        depth -= 1
+
+
+def node_depth(tree, u):
```
daniel-goldstein, line 1026: Would it be useful elsewhere if we added this to the `Tree` class?

jeromekelleher: Yes! ``Tree.depth(u)`` which returns the length of the path from u to root would be a useful operation I think. Would you mind following it up?

jeromekelleher:
> @jeromekelleher Changed to use the trees and edge diffs iterator at the same time. Do you want to get this in before moving on to the C implementation?

Thanks @daniel-goldstein, this is much nicer! I think we'll just keep this in the one PR for now if you don't mind - things might evolve a bit as we try it out in C, and it's easier keep the whole thing in my head if it's all in one diff. The PR isn't too big anyway.

We should follow up with the full-stack Tree.depth() operation first though, there's no point in reinventing this wheel.

daniel-goldstein:
Now stacked on #554

python/tests/test_topology.py
```diff
@@ -904,6 +903,400 @@ def test_internal_samples(self):
...
+
+def update_kc_incremental(tree, kc, edge_diffs, sample_index_map):
+    _, edges_out, edges_in = edge_diffs
+
+    # Update any depths/times changed by removed edges that might not be updated
+    # by inserted edges.
+    for e in edges_out:
+        update_in_subtree_pairs(tree, kc, e.child, sample_index_map)
+
+    for e in edges_in:
```
daniel-goldstein, line 969: I think this is the source of the initial slow down. For edges `e1` and `e2` in `edges_in`, if `e1` is a descendant of `e2`, then the set of pairs associated with `e1` is a subset of those associated with `e2`. The updates completed when processing `e1` would then get completely overwritten by those of `e2`. This is especially egregious in the case of the first tree, where this is doing `O(E * n^2)` updates. It should be enough (in the first tree's case) to only process those edges directly under the root, and doing so should run essentially the same algorithm as the single tree kc. This would then generalize to processing only those edges without another incoming edge on the path from that edge to the root.

daniel-goldstein: Decided that this should be a performance optimization in a second pass, not in this PR.

daniel-goldstein:
@jeromekelleher Made the changes we discussed and ready for review

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+        depth--;
+    }
+}
+
+static int
+update_in_subtree_pairs(tsk_tree_t *t, kc_vectors *kc, tsk_id_t u, double root_time)
+{
+    int stack_top;
+    tsk_size_t depth;
+    struct kc_stack_elmt *stack;
```
jeromekelleher, line 4808: It doesn't matter here because the first thing we do is malloc the stack, but I think it's still better to always initialise anything that gets `tsk_free``'d to NULL. We might add another malloced variable in the future, which gets alloc'd and tested first and then end up in an undefined state when we try to free whatever happens to be around.

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+            }
+        }
+    }
+
+out:
+    tsk_safe_free(stack);
+    return ret;
+}
+
+static int
```
jeromekelleher, line 4847: We should probably group these ``tsk_tree_x`` methods with the rest of tsk_tree_x functions within the file for consistency.

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+update_kc_incremental(tsk_tree_t *self, kc_vectors *kc, tsk_edge_list_t *edges_out,
+    tsk_edge_list_t *edges_in)
+{
+    int ret = 0;
+    tsk_edge_list_t *e_list;
+    tsk_edge_t *e;
+    tsk_id_t u;
+    double root_time, time;
+    double *times = self->tree_sequence->tables->nodes.time;
```
jeromekelleher, line 4878: Good idea to ``const`` this.

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+    tsk_treeseq_t *self, tsk_treeseq_t *other, double lambda_, double *result)
+{
+    int i;
+    tsk_id_t n;
+    double left, span, total;
+    tsk_tree_t t1, t2;
+    kc_vectors kc1, kc2;
+    tsk_diff_iter_t diff_iter1, diff_iter2;
+    tsk_treeseq_t *treeseqs[2] = { self, other };
+    tsk_tree_t *trees[2] = { &t1, &t2 };
```
jeromekelleher, line 4922: We can make this a bit simpler, can't we?
```c
tsk_tree_t trees[2];
tsk_tree_t kcs[2];
tsk_diff_iter_t diff_iters[2];

...
 for (i = 0; i < 2; i++) {
￼        ret = tsk_tree_init(&trees[i], treeseqs[i], TSK_SAMPLE_LISTS);
     ...
```

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+    kc_vectors *kcs[2] = { &kc1, &kc2 };
+    tsk_diff_iter_t *diff_iters[2] = { &diff_iter1, &diff_iter2 };
+    tsk_edge_list_t *t1_edges_out, *t1_edges_in, *t2_edges_out, *t2_edges_in;
+    double t1_left, t1_right, t2_left, t2_right;
+    int ret = 0;
+
+    for (i = 0; i < 2; i++) {
+        memset(trees[i], 0, sizeof(tsk_tree_t));
```
jeromekelleher, line 4930: Not convinced this is correct at the moment - isn't this writing ``sizeof(tsk_tree_t)`` bytes into a pointer sized variable?

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+        if (ret != 0) {
+            goto out;
+        }
+    }
+
+    total = 0;
+    left = 0;
+
+    ret = tsk_tree_first(&t1);
+    assert(ret == 1);
```
jeromekelleher, line 4960: Is this an assert or a possible error condition? I think things can go wrong depending on the input ts, so we should be checking for errors here.

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+        ret = kc_vectors_alloc(kcs[i], n);
+        if (ret != 0) {
+            goto out;
+        }
+    }
+
+    total = 0;
+    left = 0;
+
+    ret = tsk_tree_first(&t1);
```
jeromekelleher, line 4959: Just as easy to say ``&trees[0]`` here isn't it?

c/tskit/trees.c
```diff
@@ -4685,3 +4729,278 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+    assert(ret == 1);
+    ret = tsk_diff_iter_next(
+        &diff_iter1, &t1_left, &t1_right, &t1_edges_out, &t1_edges_in);
+    assert(ret == 1);
+    ret = update_kc_incremental(&t1, &kc1, t1_edges_out, t1_edges_in);
+    if (ret != 0) {
+        goto out;
+    }
+    while (tsk_tree_next(&t2)) {
```
jeromekelleher, line 4968: OK, looking at the rest of this, maybe is it simpler to have ``t1`` and ``t2`` than ``&trees[0]`` and ``&trees[1]``, so it's up to you whether we make this change. But we have to fix the memory overrun problem above.

python/_tskitmodule.c
```diff
@@ -7113,6 +7113,33 @@ TreeSequence_genealogical_nearest_neighbours(TreeSequence *self, PyObject *args,
...
+    TreeSequence *other = NULL;
+    static char *kwlist[] = {"other", "lambda_", NULL};
+    double lambda = 0;
+    double result = 0;
+    int err;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!d", kwlist,
+                &TreeSequenceType, &other, &lambda)) {
+        goto out;
+    }
```
jeromekelleher, line 7132: There's a ``TreeSequence_check_state`` function (or something) that we usually call here to make sure that the input TreeSequence object is correctly initialised. Probably overly paranoid, but should call it here for consistency.

python/tskit/trees.py
```diff
@@ -5370,6 +5370,21 @@ def genealogical_nearest_neighbours(self, focal, sample_sets, num_threads=0):
                 arrays = pool.map(worker, splits)
             return np.vstack(list(arrays))
 
+    def kc_distance(self, other, lambda_=0.0):
+        """
+        Returns an analog to the ``kc_distance`` between trees but for two tree
+        sequences. This takes the sum of the pairwise kc-distance between
+        corresponding trees on both sequences, each distance weighted by the
+        percentage of the sequence the trees span.
```
jeromekelleher, line 5378: ``percentage`` could be confusing here. How about ``fraction`` instead?

python/tskit/trees.py
```diff
@@ -5370,6 +5370,21 @@ def genealogical_nearest_neighbours(self, focal, sample_sets, num_threads=0):
                 arrays = pool.map(worker, splits)
             return np.vstack(list(arrays))
 
+    def kc_distance(self, other, lambda_=0.0):
+        """
+        Returns an analog to the ``kc_distance`` between trees but for two tree
```
jeromekelleher, line 5375: Might be better to be more direct here. Maybe "Returns the average :meth:`.Tree.kc_distance` between pairs of trees along the sequence. The average is weighted by the fraction of sequence length that each pair of trees overlap."

Or something like that.

c/tskit/trees.c
daniel-goldstein: I _think_ this is correct? `trees` is an array of pointers and `memset` takes a pointer to the memory of the given size, so this should be passing a pointer to the memory of the `tsk_tree_t` struct. If `trees` were an array of structs it would be `&trees[i]` which I think is the same thing right?

c/tskit/trees.c
daniel-goldstein: Yes this is a possible error condition. The same assert is made however in the input checking function above, should I throw an error there and keep this as an assert?

c/tskit/trees.c
daniel-goldstein: Either way I think I will make the change to just have an in-place array of structs because the end of the function doesn't get much messier with the change and it avoids this awkward duplication of pointers.

daniel-goldstein:
@jeromekelleher Addressed

c/tskit/trees.c
jeromekelleher: Yes, you're right, it is correct. I think it's good to have the array of structs all right though, it's less confusing.

c/tskit/trees.c
```diff
@@ -4685,3 +4751,255 @@ tsk_tree_kc_distance(tsk_tree_t *self, tsk_tree_t *other, double lambda, double
...
+    }
+    ret = tsk_tree_init(&t2, other, TSK_SAMPLE_LISTS);
+    if (ret != 0) {
+        goto out;
+    }
+
+    ret = tsk_tree_first(&t1);
+    if (ret != 1) {
+        goto out;
+    }
```
jeromekelleher, line 4786: Ah, OK, I see what you've done here. Let's not do a full tree generation pass just to check the inputs. We can easily run ``check_kc_distance_tree_inputs(&t1, &t2)`` as part of the main loop below. Generating the trees is an expensive operation.

c/tskit/trees.c
jeromekelleher: See the comment above - let's bring the validity checking step into this function. Shouldn't be much extra complexity and it'll save a fair bit of time.

c/tskit/trees.c
```diff
@@ -4612,76 +4642,347 @@ norm_kc_vectors(kc_vectors *self, kc_vectors *other, double lambda)
...
+    if (tsk_tree_get_num_roots(self) != 1) {
+        ret = TSK_ERR_MULTIPLE_ROOTS;
+        goto out;
+    }
+    if (!tsk_tree_has_sample_lists(self)) {
+        ret = TSK_ERR_UNSUPPORTED_OPERATION;
+        goto out;
+    }
+
+    num_samples = (tsk_id_t) self->tree_sequence->num_samples;
```
jeromekelleher, line 4661: Ah, so this is an issue. We can't go through the n samples for each tree to check if any of them are internal (an external sample in one tree could be internal in the next). We'll just have to relax the requirement, and raise an issue to make sure we follow up and document what happens when we run KC on trees with internal samples. (It all probably works just fine, we'd just need to document what happens and put in a few test cases to make sure it's doing something sensible).

daniel-goldstein: I get that it would be nice to allow it on internal samples but wouldn't this be the same behavior as in the single tree case? The whole thing fails if any individual comparisons would fail. I guess to fix it long term the question is does KC operate on samples or leaves?

jeromekelleher: I think we need to think of KC as working on samples rather than leaves. At the moment, we're assuming a 1-1 correspondence between the two, but I don't think it's hard to tweak the definition so that it makes sense for arbitrary samples (might even do it out of the box).

Anyway, for now, let's just get rid of this check in the TS version because we don't want to incur this O(n) cost for each tree, write a test to make sure things don't blow up when we have internal samples, and then follow up with an issue to make sure that we define KC properly on arbitrary samples.

daniel-goldstein: Ah ok, sounds good!

jeromekelleher:
I think this is ready to merge @daniel-goldstein?

daniel-goldstein:
Yes, it is. Forgot to tag you @jeromekelleher

jeromekelleher:
Done! Can you open an issue flagging the KC-internal-samples issue please @daniel-goldstein?

daniel-goldstein:
Done #575  :) @jeromekelleher

frezaei98:
Hi @jeromekelleher  and @daniel-goldstein 

In calculating KC metric for tree sequence, did you consider polytomies?
I mean: if a tree has polytomies, how calculate KC metric?

jeromekelleher:
Polytomies are captured directly by the metric itself (see the [paper](https://arxiv.org/abs/1507.05211)), although the interpretation is somewhat arguable. This is discussed in the tsinfer and ARGneedle papers, I believe (@hyanwong may have further details)

hyanwong:
Yes, the KC metric does allow polytomies (although not unary nodes). It's not clear, however, that the treatment of polytomies will result in a metric that is useful for your specific use-case. For example, Pier Palamara (ARGneedle PI) is very skeptical of using KC for this reason: he thinks it doesn't penalise polytomies enough.

One way of thinking about this is to ask how the metric with a polytomy compares to the average of the metrics for all possible binary resolutions of that polytomy. Usually, I think that the average is higher than the with-polytomy case. However, it's not clear that a metric exists that allows these to be the same.

You can randomly resolve the polytomies if you like, using [Tree.split_polytomies](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.Tree.split_polytomies), then recalculate the KC metric, which will give you an alternative (noisier) metric.

PR #550, opened 2020-04-21
Added nexus output and test

saunack:


python/tskit/trees.py
```diff
@@ -5442,3 +5442,50 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
             " method instead"
         )
+
+    def nexus(self, precision=14, node_labels=None):
+        """
+        Returns a nexus encoding <https://en.wikipedia.org/wiki/Nexus_file>`_
+        of this tree.
+        :param int precision: The numerical precision with which branch lengths are
+            printed.
+        :param dict node_labels: If specified, show custom labels for the nodes
```
jeromekelleher, line 5452: Let's leave this parameter out, for now. I think we keep the newick strings having their Node IDs as labels. We can add in support for the TRANSLATE blocks in a follow up.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,50 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+            that are present in the map. Any nodes not specified in the map will
+            not have a node label.
+        :return: A newick representation of this tree.
```
jeromekelleher, line 5455: A nexus representation of this tree sequence.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,50 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+        :rtype: str
+        """
+        sample_tree = self.first()
```
jeromekelleher, line 5458: Things are bit simpler if we use [samples](https://tskit.readthedocs.io/en/latest/data-model.html#definitions) rather than leaves. Then, we have 
```
node_labels = {u: str(u) for u in self.samples()}
```
then we don't need to look at the first tree at all.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,50 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+            node_labels = leaf_labels
+        # if node_labels is None:
+        # leaf_labels = {u: str(u + 1) for u in sample_tree.leaves()}
+        # else:
+        # leaf_labels = {u: node_labels[u] for u in sample_tree.leaves()}
+
+        # print(self.first().draw(format='unicode'))
+        s = "#NEXUS\n"
+        # writing the labels of the nodes in the trees
+        s = s + "BEGIN TAXA;\n"
```
jeromekelleher, line 5470: ``s += "BEGIN TAXA;\n"`` is a bit nicer.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,50 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+        s = s + "\tTAXLABELS\n"
+
+        for leaf in sample_tree.leaves():
+            s = s + f"\t\t{leaf}\n"
+        s = s + "\t\t;\n"
+        s = s + "END;\n"
+
+        s = s + "BEGIN TREES;\n"
+        s = s + "\tTRANSLATE\n"
```
jeromekelleher, line 5479: As I said, let's leave out the translate block for now.

python/tests/test_nexus.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_binary_example(self):
+        ts = msprime.simulate(
+            sample_size=4, recombination_rate=2, random_seed=self.random_seed
+        )
+        return ts
+
+    def test_binary_tree(self):
+        ts = self.get_binary_example()
+        self.verify_nexus_topology(ts)
```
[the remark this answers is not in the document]
saunack: Are tests for multi-rooted trees or subtrees appropriate for TreeSequences?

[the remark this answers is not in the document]
jeromekelleher: No, we don't need to worry about those here.

python/tests/test_phylo_formats.py
```diff
@@ -146,3 +147,83 @@ def test_single_node_label(self):
...
+        for e in ts.edgesets():
+            if len(e.children) > 2:
+                found = True
+                break
+        self.assertTrue(found)
+        return ts
+
+    def test_nonbinary_example(self):
+        ts = self.get_nonbinary_example()
+        self.verify_nexus_topology(ts)
```
jeromekelleher, line 229: We also want to test the single tree example:
```
def test_single_tree(self):
     ts = msprime.simulate(10, random_seed=2)
     self.verify_nexus_topology(ts)
```

It might be nice to think about how we can avoid duplicating code within this test module. Perhaps we could do a test mixin, or even just factor out the ``get_X_example`` functions. We do want a few more test examples in any case.

python/tests/test_phylo_formats.py
```diff
@@ -146,3 +147,83 @@ def test_single_node_label(self):
...
+        nexus_treeseq = Nexus.Nexus(nexus)
+
+        interval = [0.0, 0.0]
+        for tree, nexus_tree in zip(treeseq.trees(), nexus_treeseq.trees):
+            name = nexus_tree.name
+            split_name = name.split("_")
+            self.assertEqual(len(split_name), 2)
+            start = float(split_name[0][4:])
+            end = float(split_name[1])
+            self.assertAlmostEqual(interval[1], start)
```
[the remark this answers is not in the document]
saunack: Should we match the intervals with the individual tree's intervals or check if the nexus output spans the whole interval?

python/tskit/trees.py
```diff
@@ -5442,3 +5442,39 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+
+        :param int precision: The numerical precision with which branch lengths
+            and tree positions are printed.
+        :param dict node_labels: If specified, show custom labels for the nodes
+            that are present in the map. Any nodes not specified in the map will
+            not have a node label.
+        :return: A nexusu representation of this tree.
+        :rtype: str
+        """
+        leaf_labels = {u: str(u + 1) for u in self.samples()}
```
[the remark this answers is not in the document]
saunack: The nexus format contains the nexus outputs in the TREE block, which use these labels by default.
Instead, I could use the default label itself and use the translate block to convert the node ids to the ones used by newick. The code could then be used for custom translations too.

python/tests/test_phylo_formats.py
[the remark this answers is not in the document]
jeromekelleher: Just matching the trees intervals is fine, this is tested heavily elsewhere.

python/tskit/trees.py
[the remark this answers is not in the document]
jeromekelleher: Ah, I see, and then 1 to n is used to index into the TRANSLATE block? OK, then we want to make sure that the labels in the newick trees are 1 to n (u + 1 won't do it, as the samples are not necessarily 0 to n - 1), and use the TRANSLATE block to map these to their actual node IDs, as the default behaviour.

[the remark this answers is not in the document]
jeromekelleher: There is a distinction between "leaves" and "samples" here too which we want to be careful with. Samples are special nodes we are interested in. Samples are not necessarily leaves, and leaves are not necessarily samples. https://tskit.readthedocs.io/en/latest/data-model.html#definitions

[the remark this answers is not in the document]
saunack: In that case, I should follow the code in TestNewick class and use tree.leaves().
In which case are the leaves not labelled 0 to n-1? I might need to change the code accordingly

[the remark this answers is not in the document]
jeromekelleher: We are more interested in samples than leaves, so it would be better to use samples. Have a look in test_topology for examples in which leaves are not 0 to n - 1, there's all sorts of weird topologies that can happen that are different to the basic msprime simualtion.

saunack:
I translated the tree samples manually so that they are now labelled from 1 to n. The tests also seem to be working fine.
Any more tests which need to be added? I have added all the tests from TestNewick after appropriate modifications (the only ones left are the ones which use node_labels).
If this looks fine, I can go ahead and add user-defined translate blocks too, if that is a required feature

python/tests/test_phylo_formats.py
```diff
@@ -94,18 +75,48 @@ def get_multiroot_example(self):
...
 
     def test_binary_tree(self):
-        ts = self.get_binary_example()
+        ts = super().get_binary_example()
         for t in ts.trees():
             self.verify_newick_topology(t)
 
     def test_multiroot(self):
-        ts = self.get_multiroot_example()
+        ts = super().get_multiroot_example()
```
jeromekelleher, line 119: super() isn't needed here, you can call ``self.`` still.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,44 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+        """
+        Returns a nexus encoding <https://en.wikipedia.org/wiki/Nexus_file>`_
+        of this tree. Trees along the sequence are listed sequentially in
+        the TREES block. The tree spanning the interval :math:`[x, y)`` is
+        given the name "tree_x_y". Spatial positions are written at the
+        specified precision.
+
+        :param int precision: The numerical precision with which branch lengths
+            and tree positions are printed.
+        :return: A nexus representation of this tree.
```
jeromekelleher, line 5456: Still not accurate I'm afraid: "A nexus representation of this tree sequence."

python/tskit/trees.py
```diff
@@ -5442,3 +5442,44 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+
+        for leaf_id in leaf_labels.values():
+            s += f"\t\t{leaf_id}\n"
+        s += "\t\t;\n"
+        s += "END;\n"
+
+        s += "BEGIN TREES;\n"
+        s += "\tTRANSLATE\n"
+        for label in leaf_labels:
+            s += "\t\t" + leaf_labels[label] + " " + str(label) + ",\n"
```
jeromekelleher, line 5475: f strings will make this cleaner: 
```python
s += f"\t\t{leaf_labels[label]} {label}, \n"
```
Looks a bit like Java at the moment.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,44 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+        """
+        leaf_labels = {}
+        for i, u in enumerate(self.samples()):
+            leaf_labels[u] = str(i + 1)
+
+        s = "#NEXUS\n"
+        s += "BEGIN TAXA;\n"
+        s += "\tTAXLABELS\n"
+
+        for leaf_id in leaf_labels.values():
```
jeromekelleher, line 5467: Your depending on the order of iteration here, which isn't guaranteed in older versions of Python. Seems clearer to be explicit about the order that your assuming here by iterating over samples:
```
for u in self.samples():
     s += f"\t\t{leaf_labels[u]}\n"
```

python/tskit/trees.py
```diff
@@ -5442,3 +5442,44 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+            s += f"\t\t{leaf_id}\n"
+        s += "\t\t;\n"
+        s += "END;\n"
+
+        s += "BEGIN TREES;\n"
+        s += "\tTRANSLATE\n"
+        for label in leaf_labels:
```
jeromekelleher, line 5474: Same here I guess, although it's less obvious that order matters here.
```
for u in self.samples():
    s += f"\t\t{leaf_labels[u]} {u}, \n"
```

saunack:
There's also an option for specifying rooted and unrooted trees in Nexus. Should I use that too in the nexus output?

saunack:
Also, a peculiar behavior that I have noticed. The BioNexus package doesn't actually do anything with the BEGIN TAXA block; it's just stored in another class. Even if I write incorrect labels there, it does not matter since the parsing for the nexus trees uses only the data in the TRANSLATE block and the TREE block. If the BioNexus package does not itself check for consistency across the TAXA and TRANSLATE block, should I do it instead?

jeromekelleher:
> There's also an option for specifying rooted and unrooted trees in Nexus. Should I use that too in the nexus output?

If it's part of the standard I guess we should state that the trees are rooted, I suppose.

> Also, a peculiar behavior that I have noticed. The BioNexus package doesn't actually do anything with the BEGIN TAXA block; it just stores it in another class. Even if I write incorrect labels there, it does not matter since the parsing for the nexus trees uses only the data in the TRANSLATE block and the TREE block. If the BioNexus package does not itself check for consistency across the TAXA and TRANSLATE block, should I be doing it instead?

Hmm. We could try parsing with the [nexus](https://pypi.org/project/python-nexus/) package to see if it reads the TAXA block? No harm in testing against two different libraries, in case one interprets the standard differently to the other.

But yes, I guess we should make sure that the labels are correctly output. You should be able to use the output from BioPython to do this?

saunack:
The python-nexus package that you mentioned also doesn't check for consistency between different blocks.
And yeah, I should be able to do it using BioPython.
Although now I doubt if we should check for it. Can you go through page 15 and page 42 of [this paper](informatics.nescent.org/w/images/8/8b/NEXUS_Final.pdf) on Nexus and let me know your views on it? From what I gathered, it is not necessary to write all taxons there and using numbers as taxlabels is not preferred.

python/tskit/trees.py
```diff
@@ -5442,3 +5442,45 @@ def newick_trees(self, precision=3, breakpoints=None, Ne=1):
...
+        for i, u in enumerate(self.samples()):
+            leaf_labels[u] = str(i + 1)
+
+        s = "#NEXUS\n"
+        s += "BEGIN TAXA;\n"
+        s += f"\tDIMENSIONS NTAX={self.num_samples};\n"
+        s += "\tTAXLABELS\n"
+
+        for u in self.samples():
+            s += f"\t\t{u}\n"
```
jeromekelleher, line 5469: So looks like this isn't allowed in the spec:
> Taxon names are single NEXUS words. They must not correspond to another taxon name or number; thus, 1 is not a valid name for the second taxon listed. The standard defines no limit to their length, although individual programs might impose restrictions. 

We could have ``tsk_{u}`` as the taxon name? It's not totally clear to me that they are ruling out numbers here though, it's confusing.

jeromekelleher:
This is surprisingly complicated! I don't know enough about this to make calls on it really - @hyanwong, would you mind reviewing here please? In particular, can you clear up what the TAXLABELS and TRANSLATE blocks actually mean, and what we should use them for?

python/tskit/trees.py
hyanwong: Re numbers, the relevant section may be this:

> Taxa, characters, and trees possess an inherent order, determined by their order of definition. This allows them to be referred to by name (if they have been given names) or by number.
Thus, in the following matrix:
MATRIX
taxon_A 001020101
taxon_B 101001022
taxon_C 001010201;
the second taxon could be referred to as taxon_B or as 2. Characters, taxa, and trees can be
referred to by their name or number

hyanwong: @saunack: I used a `TRANSLATE` block to convert to zero-based numbering, FWIW.

hyanwong:
@saunack: I'm not sure why we want to define a TAXA block at all. We don't have to, do we? Then all this goes away.

The main sticking point here is whether you can have a TRANSLATE command without a TAXON block - i.e. can the first column of the TRANSLATE table simply refer to a taxon number, even if it hasn't been labelled in a TAXON block. I think this is OK - certainly at the moment we should be applying TRANSLATE to *all* the nodes, not just the sample ones, and as I read it, you are only putting the samples into the TAXA block anyway: i.e. there will be internal nodes that are TRANSLATEd but not in the TAXA block.

python/tskit/trees.py
hyanwong: I think we should do away with the TAXA block entirely. It's not required by the spec, as far as I can see. In the spec they give, for example:

```
#NEXUS
BEGIN TREES;
TREE best= (fish, (frog, (snake, mouse)) ) ;
END;
```

saunack:
No, it's not required. Everything works well without it too.
All the problems go away then. Thanks @hyanwong 
Although I do not translate all the nodes in the tree. Is it necessary since we're just checking for isomorphism? Traversing the tree from the leaf to the root and checking node lengths along the way should be fine, I guess

hyanwong:
> Although I do not translate all the nodes in the tree. Is it necessary

The reason I used a TRANSLATE block is that, AFAIK, zero is not a valid NEXUS taxon number. I think this caused problems when reading these nexus files into R. So I moved all the taxon numbers up by 1 using the TRANSLATE block. But then we have the issue that strictly, according to one version of the spec, we need to translate to *names* not numbers, and we are back to the problem of defining a name for each taxon (including internal nodes). Personally, I think we should treat the spec as allowing numbers as the second column in the translate block, otherwise it opens up a can of worms.

hyanwong:
Perhaps we should add a flag to nexus conversion that says essentially "leave as 0-based node numbers" (this breaks the spec so may cause problems for some nexus readers) or "translate to 1-based node numbers" (it is unclear if this breaks the TRANSLATE block spec or not). What do you think @jeromekelleher ?

jeromekelleher:
> Perhaps we should add a flag to nexus conversion that says essentially "leave as 0-based node numbers" (this breaks the spec so may cause problems for some nexus readers) or "translate to 1-based node numbers" (it is unclear if this breaks the TRANSLATE block spec or not). What do you think @jeromekelleher ?

The trouble is our sample IDs aren't necessarily 0,.., n - 1, so just adding one doesn't really work and makes things even more confusing. I don't think we should break the spec - can we not to something like making the names ``[f"tsk_{u}" for u in samples]`` and translate them to node numbers?

hyanwong:
> The trouble is our sample IDs aren't necessarily 0,.., n - 1, so just adding one doesn't really work and makes things even more confusing. 

I wasn't meaning just sample IDs. Surely we want *all* the IDs (including internal nodes) in the trees, otherwise we can't compare tree-to-tree? This would make *all* node IDs equal to n+1, which is surely fine?

jeromekelleher:
Oh, so we label all internal nodes as well. That makes sense.

But isn't there still an issue issue that Nexus thinks numbers refer to (1 based) indexes into the Taxa list?  I think your approach above is the correct one:

> Personally, I think we should treat the spec as allowing numbers as the second column in the translate block, otherwise it opens up a can of worms.

Can you illustrate this with a quick example?

hyanwong:
So as I understand it, we don't require a TAXA list (there is an example in the spec without one), and indeed it's not even clear from the spec that internal nodes can be labelled as taxa. So *if* we were using a 1-based system for node labels in the trees, everything would be grand - the numbers refer to positions in the (non-existent) TAXA list. However, we will have a node (could be internal or a tip) with the label `0`. I have previously found that this causes problems with nexus parsers. However, there is the option to pass a `TRANSLATE` block before the tree definitions, of the form `TRANSLATE tree_label1 taxa_label1, tree_label2 taxa_label2 ...` So previously, to allow a tree with a node labelled 0, I have set

```
TRANSLATE
0 1,
1 2,
2 3
...
```

simply to keep the nexus parsers happy. This means that the tree can continue to have a label `0` on a node, but for nexus purposes it is taken as referring to the first item in the taxon list, were a taxon list to exist.

The only problem now is that if read strictly, you might think that _`taxa_label1`_, _`taxa_label2`_ (i.e. the 2nd "column") in the TRANSLATE block above have to be labels, not indexes into the list. However, the labels and indexes are treated as interchangeable in the rest of the spec, and the proposed automatic nexus parser is specifically coded to allow the second column in the translate block to be an index or a label, so I deem this an acceptable work-around.

jeromekelleher:
Great, thanks @hyanwong. It's very tricky having this interchangability between labels and numbers, so wouldn't it be better if we took away any possibility of ambiguity by defining our samples using the TAXA list like
```
TAXA
    tsk_0, tsk_1,...
```
where the taxon label is ``[f"tsk_{u}" for u in ts.samples()]``? And we then consistently refer to the nodes using these labels?

It's horribly confusing having offset-by-one as the default, and I think this is the only way to both resolve the potential ambiguity between our node IDs and the taxon IDs and to avoid this nasty source of potential errors.

Let's not worry about the few extra bytes that this costs - efficiency is not a concern here.

hyanwong:
> wouldn't it be better if we took away any possibility of ambiguity by defining our samples using the TAXA list like

That's fine for the samples, but I don't know if we want to (or can) do that for the non-sample nodes, as these aren't really taxa. I do think we should leave the internal node numbers in there. I suppose we could use the TRANSLATE block, and list the samples only, in which case if the sample IDs are (say) 10, 11, 12, then we would have blocks like

```
TAXA tsk_10, tsk_11, tsk_12
TRANSLATE 10 1, 11 2, 12 3
```

And we would need to work out how to map the internal nodes and add them to the translate block too, at least for the nodes whose tskit ID lies between 0 and the number of listed taxa. Otherwise a `1` in the newick will be taken to refer to the first listed taxon (i.e. tsk_10). So either way we will need to do numerical remapping, I fear.

jeromekelleher:
I'm not convinced that internal nodes can't be thought of as taxa - suppose we have two closely related species and we know the name of the ancestral species. Does the spec specifically say that taxa can only be leaf nodes?

Perhaps we could try doing

```
taxa = [f"tsk_{u}" for u in ts.num_nodes()]
```

and see if the parsers choke on this? There's too much ambiguity if we don't stringify our node IDs.

hyanwong:
The only reference I can find to internal nodes is this: 

> The label of the node is a NEXUS token that is either a taxon's defined name, a taxon's number, a taxon's label from the translation table, or a clade's defined name. The label is optional for internal nodes that are not observed taxa; it is not optional for terminal nodes. Internal nodes that have no label are represented implicitly by the parentheses containing the list of subclades.

jeromekelleher:
> Given that we could have terminal nodes that are not samples (e.g. in an ancestors TS), perhaps we should indeed list all nodes, including internal ones, as TAXA. It might be useful to have a different TAXA prefix for samples, though?

Good idea - how about ``["tsk_{node.flags}_{node.id}" for node in ts.nodes()]``? This will usually be zero or 1, but can contain more information if it's there. (Let's not worry about other node information for now - we can follow this up with a better encoding scheme if this works.)

@saunack, would you mind trying this idea out to see if BioPython will round-trip these node labels faithfully? It'll look something like
```python
node_labels = {node.id: f"tsk_{node.flags}_{node.id}" for node in self.nodes()}
taxa_block = "TAXA " + ",".join(node_labels[node.id] for node in self.nodes()) + "\n"
# Pass node_labels in as the argument to tree.newick()
```

hyanwong:
> `taxa_block = "TAXA " + ",".join`
I don't think you join TAXA by a comma, BTW. I happen to know one of the principal authors of a well-used nexus parsing library, so I'm just chatting to him about conformance.

jeromekelleher:
> I don't think you join TAXA by a comma, BTW. 

Just going by your example above: 
> TAXA tsk_10, tsk_11, tsk_12
> TRANSLATE 10 1, 11 2, 12 3

> BTW. I happen to know one of the principal authors of a well-used nexus parsing library, so I'm just chatting to him about conformance.

Great, thanks.

hyanwong:
> Just going by your example above:

Oops. My bad :embarrassed:. Corrected now!

hyanwong:
Apparently PAUP doesn't (didn't?) like having taxa as internal labels, but I can't see anything in the spec against it. And we aren't likely to want to read tskit output into PAUP, I guess.

jeromekelleher:
Excellent, consensus!

saunack:
> @saunack, would you mind trying this idea out to see if BioPython will round-trip these node labels faithfully? It'll look something like
> 
> ```python
> node_labels = {node.id: f"tsk_{node.flags}_{node.id}" for node in self.nodes()}
> taxa_block = "TAXA " + ",".join(node_labels[node.id] for node in self.nodes()) + "\n"
> # Pass node_labels in as the argument to tree.newick()
> ```

This works without any issues. To be clear though, I haven't used the TRANSLATE block.

jeromekelleher:
OK, great! This is ready to merge @saunack, thanks very much.

I updated the PR with a few extra tests, making sure we deal with multiple trees correctly and that we're mapping from ID labels in a one-to-one manner. I also updated the docs a bit.

hyanwong:
I'm not sure it it's a good idea to omit the TRANSLATE block? It does pass BioPython import, and maybe some others, but I think it might not be spec-compliant? ISTR it broke import into R, for example. It would be trivial to add, this surely @jeromekelleher ?

hyanwong:
Mark Holder, who I consulted said that without the TRANSLATE command:
> oddly, n+1 will be a default alias for n in your 0-based tree

Probably worth importing a small example into R to check.

jeromekelleher:
I've just merged @hyanwong  - would you mind checking it out in R, and adding any extra stuff you think it needs to be standards compliant?

saunack:
Since I've renamed all the nodes to tsk_flag_id, shouldn't this no longer be a problem?
The newick format also uses these labels for printing the trees so if it was to be read by some other program, it should work fine too.
Although I may wrong due to the compliance part, which I don't know anything about.

hyanwong:
@saunack - do you mean you have renamed them *in the trees*? Or simply in the TAXA block. If in the trees themselves, that's fine, of course, although it obviously bloats the file size, and it's not so obvious to pull a tree out of the text file and compare it to the `Tree.newick()` output.

jeromekelleher:
> do you mean you have renamed them in the trees? Or simply in the TAXA block. If in the trees themselves, that's fine, of course, although it obviously bloats the file size, and it's not so obvious to pull a tree out of the text file and compare it to the Tree.newick() output.

Yes, it's in the trees. I don't think we really care about file size. It doesn't matter that Tree.newick() has a different default format.

hyanwong:
Oh, well that's fine then. Sorry for the noise. I just assumed you were re-using the `Tree.newick()` output and sticking it into the nexus file.

PR #554, opened 2020-04-23
Add depth method to tree class

daniel-goldstein:
Now `t.depth(u)` returns the depth of node `u` in tree `t`, where depth is defined as the number of nodes on the path from `u` to the root, not including `u`.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -3565,6 +3565,28 @@ tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_lengt
     return 0;
 }
 
+int TSK_WARN_UNUSED
+tsk_tree_depth(tsk_tree_t *self, tsk_id_t u, tsk_size_t *depth_ret)
+{
+    tsk_id_t v;
+    tsk_size_t depth;
+    int ret = 0;
+
+    ret = tsk_tree_check_node(self, u);
+    if (ret != 0) {
+        goto out;
+    }
+
+    depth = 0;
+    for (v = self->parent[u]; v != TSK_NULL; v = self->parent[v]) {
+        depth++;
+    }
+
+    *depth_ret = depth;
+out:
+    return ret;
+}
+
 static void
 tsk_tree_check_state(tsk_tree_t *self)
 {
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -8765,6 +8765,26 @@ out:
     return ret;
 }
 
+static PyObject *
+Tree_depth(Tree *self, PyObject *args)
+{
+    PyObject *ret = NULL;
+    tsk_size_t depth;
+    int node, err;
+
+    if (Tree_get_node_argument(self, args, &node) != 0) {
+        goto out;
+    }
+    err = tsk_tree_depth(self->tree, node, &depth);
+    if (ret != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = Py_BuildValue("I", (unsigned int) depth);
+out:
+    return ret;
+}
+
 static bool
 Tree_check_sample_list(Tree *self)
 {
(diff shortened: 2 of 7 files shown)


daniel-goldstein:
@jeromekelleher Should I be using a `uint32_t` (or some other int type) for depth and cast it to an `unsigned int` when building the python value?

c/tskit/trees.c
```diff
@@ -3565,6 +3565,30 @@ tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_lengt
     return 0;
 }
 
+int TSK_WARN_UNUSED
+tsk_tree_depth(tsk_tree_t *self, tsk_id_t u, unsigned int *depth_ret)
```
jeromekelleher, line 3569: Let's make this a ``tsk_size_t``, for consistency.

c/tskit/trees.c
```diff
@@ -3565,6 +3565,30 @@ tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_lengt
...
+    tsk_id_t parent;
+    unsigned int depth;
+    int ret = 0;
+
+    ret = tsk_tree_check_node(self, u);
+    if (ret != 0) {
+        goto out;
+    }
+
+    parent = self->parent[u];
```
jeromekelleher, line 3580: A bit shorter:
```c
depth = 0; 
for (v = self->parent[u]; v != TSK_NULL; v = self->parent[v]) {
    depth++;
}
```
I think ``parent`` is a slightly misleading variable name as it's a iteration variable and using ``v`` for it is a bit more idiomatic. I don't feel strongly about it though, no biggie.

python/_tskitmodule.c
```diff
@@ -8765,6 +8765,26 @@ Tree_get_children(Tree *self, PyObject *args)
...
+
+    if (Tree_get_node_argument(self, args, &node) != 0) {
+        goto out;
+    }
+    err = tsk_tree_depth(self->tree, node, &depth);
+    if (ret != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = Py_BuildValue("I", depth);
```
jeromekelleher, line 8783: As we're changing depth to a (tsk_size_t), it'll be good to cast this to an unsigned int now, like
```c
ret = Py_BuildValue("I", (unsigned int) depth);
```
we play a bit loosey-goosey with the types here, as the Python C API doesn't have converter strings for specified sizes like uint32_t etc.

python/tskit/trees.py
```diff
@@ -952,6 +952,17 @@ def time(self, u):
         """
         return self._ll_tree.get_time(u)
 
+    def depth(self, u):
+        """
+        Returns the number of nodes above the specified node on
```
jeromekelleher, line 957: The word "node" is a bit overused here, how about "Returns the number of nodes on the path from ``u`` to root.`` This seems quite clear and concise.

jeromekelleher:
@petrelharp, do you agree with this definition of Tree.depth(u):

> Returns the number of nodes on the path from ``u`` to root.

I think the PR is uncontroversial otherwise, no need to look over it.

jeromekelleher:
Actually, now that I look at it that definition is wrong! Maybe "Returns the number of nodes on the path from u to root, not including u.  Thus, the depth of u is 0 if u is a root."

daniel-goldstein:
Good catch! In the code I used the wording "above u" but I think it's better to just be more explicit and say not including u.

jeromekelleher:
> Good catch! In the code I used the wording "above u" but I think it's better to just be more explicit and say not including u.

"Above" is tricky, because a lot of people think of trees with the root at the bottom (I don't know *where* they got that from!), so it's good to avoid it if you can. Sometimes it just makes things unreadable if you don't choose a direction though.

petrelharp:
> Returns the number of nodes on the path from u to root, not including u. Thus, the depth of u is 0 if u is a root.

I agree with this, except that it should say "from u to **a** root". (probably what you meant to write).

Issue #26, opened 2018-05-30
Would be nice if SVG trees could have the tip labels rotated by 90 degrees 


hyanwong:
At the moment, for large trees with custom labels, labels from SparseTree.draw(format="svg") are bound to overlap. An option to rotate the labels by 90 degrees would be helpful, and probably very easy to code.

hyanwong:
We can now do this via SVG styles (`text {transform: rotate(90deg)}`), and https://github.com/tskit-dev/tskit/issues/589 would deal with the padding issue, so I'm closing this.

PR #555, opened 2020-04-24
Svg corrections

### LINKED ISSUE #467: SVG IDs not unique and being cached by Jupyter
Hi everyone! I keep finding that when I use `draw_svg` several times in the same Jupyter notebook, I get odd (wrong) drawings that are always wrong in the same way. An example is below:

Screen Shot 2020-03-09 at 5 04 00 pm

In the notebook shown above, I have two difference tree sequence objects, `ts0` and `ts1`, that I wish to draw. Using `draw_svg` with `ts0` first produces the correct output, but when I try to draw `ts1` in the following cell, you can see that the first few trees are just copied over from `ts0`. 

Without digging into it further I'm not totally sure what's going on, but I'm wondering whether `draw_svg` is inadvertently adding something to some global object each time it is called, without checking to see if it already exists beforehand and clearing it?

hyanwong:
Fixes #467 . I have left it as 3 separate commits in case these should be separated out, as they do different but related (and dependent) things.

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1448,11 +1448,11 @@ class TestDrawSvg(TestTreeDraw):
         svg2 = ts.at_index(1).draw()
         # if not scaled to ts, node 3 is at a different height in both trees, because the
         # root is at a different height. We expect a label looking something like
-        # <text x="10.0" y="XXXX">3</text> where XXXX is different
+        # <g transform="translate(10.0 XXXX)"><text>3</text></g> where XXXX is different
         str_pos = svg1.find(">3<")
-        snippet1 = svg1[svg1.rfind("<", 0, str_pos) : str_pos]
+        snippet1 = svg1[svg1.rfind("<g", 0, str_pos) : str_pos]
         str_pos = svg2.find(">3<")
-        snippet2 = svg2[svg2.rfind("<", 0, str_pos) : str_pos]
+        snippet2 = svg2[svg2.rfind("<g", 0, str_pos) : str_pos]
         self.assertNotEqual(snippet1, snippet2)
 
         svg1 = ts.at_index(0).draw(max_tree_height="ts")
@@ -1462,9 +1462,9 @@ class TestDrawSvg(TestTreeDraw):
         self.verify_basic_svg(svg1)
         self.verify_basic_svg(svg2)
         str_pos = svg1.find(">3<")
-        snippet1 = svg1[svg1.rfind("<", 0, str_pos) : str_pos]
+        snippet1 = svg1[svg1.rfind("<g", 0, str_pos) : str_pos]
         str_pos = svg2.find(">3<")
-        snippet2 = svg2[svg2.rfind("<", 0, str_pos) : str_pos]
+        snippet2 = svg2[svg2.rfind("<g", 0, str_pos) : str_pos]
         self.assertEqual(snippet1, snippet2)
 
     def test_draw_simple_ts(self):
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -30,6 +30,7 @@ import unittest
 import xml.etree
 
 import msprime
+import xmlunittest
 
 import tests.tsutil as tsutil
 import tskit
@@ -44,18 +45,20 @@ class TestTreeDraw(unittest.TestCase):
         ts = msprime.simulate(10, random_seed=1, mutation_rate=1)
         return next(ts.trees())
 
-    def get_nonbinary_tree(self):
+    def get_nonbinary_ts(self):
         demographic_events = [
             msprime.SimpleBottleneck(time=0.1, population=0, proportion=0.5)
         ]
-        ts = msprime.simulate(
+        return msprime.simulate(
             10,
             recombination_rate=5,
             mutation_rate=10,
             demographic_events=demographic_events,
             random_seed=1,
         )
-        for t in ts.trees():
+
+    def get_nonbinary_tree(self):
+        for t in self.get_nonbinary_ts().trees():
             for u in t.nodes():
                 if len(t.children(u)) > 2:
                     return t
@@ -139,6 +142,60 @@ class TestTreeDraw(unittest.TestCase):
         tables = tskit.TableCollection(sequence_length=1)
         ts = tables.tree_sequence()
         return next(ts.trees())
+
+    def get_simple_ts(self):
+        """
+        return a simple tree seq that does not depend on msprime
+        """
+        nodes = io.StringIO(
+            """\
+        id      is_sample   population      individual      time    metadata
+        0       1       0       -1      0.00000000000000
+        1       1       0       -1      0.00000000000000
+        2       1       0       -1      0.00000000000000
+        3       1       0       -1      0.00000000000000
+        4       0       0       -1      0.02445014598813
+        5       0       0       -1      0.11067965364865
+        6       0       0       -1      1.75005250750382
+        7       0       0       -1      2.31067154311640
+        8       0       0       -1      3.57331354884652
+        9       0       0       -1      9.08308317451295
+        """
+        )
+        edges = io.StringIO(
+            """\
+        id      left            right           parent  child
+        0       0.00000000      1.00000000      4       0
+        1       0.00000000      1.00000000      4       1
+        2       0.00000000      1.00000000      5       2
+        3       0.00000000      1.00000000      5       3
+        4       0.79258618      0.90634460      6       4
+        5       0.79258618      0.90634460      6       5
+        6       0.05975243      0.79258618      7       4
+        7       0.90634460      0.91029435      7       4
+        8       0.05975243      0.79258618      7       5
+        9       0.90634460      0.91029435      7       5
+        10      0.91029435      1.00000000      8       4
+        11      0.91029435      1.00000000      8       5
+        12      0.00000000      0.05975243      9       4
+        13      0.00000000      0.05975243      9       5
+        """
+        )
+        sites = io.StringIO(
+            """\
+        position      ancestral_state
+        0.01          A
+        """
+        )
+        mutations = io.StringIO(
+            """\
+        site   node    derived_state    parent
+        0      4       T                -1
+        """
+        )
+        return tskit.load_text(
+            nodes, edges, sites=sites, mutations=mutations, strict=False
+        )
 
 
 class TestFormats(TestTreeDraw):
@@ -364,7 +421,7 @@ class TestDrawTextErrors(unittest.TestCase):
                 t.draw_text(orientation=bad_orientation)
 
 
-class TestDrawTextExamples(unittest.TestCase):
+class TestDrawTextExamples(TestTreeDraw):
     """
     Verify that we get the correct rendering for some examples.
     """
@@ -954,42 +1011,7 @@ class TestDrawTextExamples(unittest.TestCase):
         self.verify_text_rendering(t.draw_text(), tree)
 
     def test_simple_tree_sequence(self):
-        nodes = io.StringIO(
-            """\
-            id      is_sample   population      individual      time    metadata
-            0       1       0       -1      0.00000000000000
-            1       1       0       -1      0.00000000000000
-            2       1       0       -1      0.00000000000000
-            3       1       0       -1      0.00000000000000
-            4       0       0       -1      0.02445014598813
-            5       0       0       -1      0.11067965364865
-            6       0       0       -1      1.75005250750382
-            7       0       0       -1      2.31067154311640
-            8       0       0       -1      3.57331354884652
-            9       0       0       -1      9.08308317451295
-        """
-        )
-        edges = io.StringIO(
-            """\
-            id      left            right           parent  child
-            0       0.00000000      1.00000000      4       0
-            1       0.00000000      1.00000000      4       1
-            2       0.00000000      1.00000000      5       2
-            3       0.00000000      1.00000000      5       3
-            4       0.79258618      0.90634460      6       4
-            5       0.79258618      0.90634460      6       5
-            6       0.05975243      0.79258618      7       4
-            7       0.90634460      0.91029435      7       4
-            8       0.05975243      0.79258618      7       5
-            9       0.90634460      0.91029435      7       5
-            10      0.91029435      1.00000000      8       4
-            11      0.91029435      1.00000000      8       5
-            12      0.00000000      0.05975243      9       4
-            13      0.00000000      0.05975243      9       5
-        """
-        )
-        ts = tskit.load_text(nodes, edges, strict=False)
-
+        ts = self.get_simple_ts()
         ts_drawing = (
             "9.08┊    9    ┊         ┊         ┊         ┊         ┊\n"
             "    ┊  ┏━┻━┓  ┊         ┊         ┊         ┊         ┊\n"
@@ -1065,41 +1087,7 @@ class TestDrawTextExamples(unittest.TestCase):
         )
 
     def test_max_tree_height(self):
-        nodes = io.StringIO(
-            """\
-            id      is_sample   population      individual      time    metadata
-            0       1       0       -1      0.00000000000000
-            1       1       0       -1      0.00000000000000
-            2       1       0       -1      0.00000000000000
-            3       1       0       -1      0.00000000000000
-            4       0       0       -1      0.02445014598813
-            5       0       0       -1      0.11067965364865
-            6       0       0       -1      1.75005250750382
-            7       0       0       -1      2.31067154311640
-            8       0       0       -1      3.57331354884652
-            9       0       0       -1      9.08308317451295
-        """
-        )
-        edges = io.StringIO(
-            """\
-            id      left            right           parent  child
-            0       0.00000000      1.00000000      4       0
-            1       0.00000000      1.00000000      4       1
-            2       0.00000000      1.00000000      5       2
-            3       0.00000000      1.00000000      5       3
-            4       0.79258618      0.90634460      6       4
-            5       0.79258618      0.90634460      6       5
-            6       0.05975243      0.79258618      7       4
-            7       0.90634460      0.91029435      7       4
-            8       0.05975243      0.79258618      7       5
-            9       0.90634460      0.91029435      7       5
-            10      0.91029435      1.00000000      8       4
-            11      0.91029435      1.00000000      8       5
-            12      0.00000000      0.05975243      9       4
-            13      0.00000000      0.05975243      9       5
-        """
-        )
-        ts = tskit.load_text(nodes, edges, strict=False)
+        ts = self.get_simple_ts()
         tree = (
             "   9   \n"
             " ┏━┻━┓ \n"
@@ -1134,16 +1122,46 @@ class TestDrawTextExamples(unittest.TestCase):
                 t.draw_text(max_tree_height=bad_max_tree_height)
 
 
-class TestDrawSvg(TestTreeDraw):
+class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestCase):
     """
     Tests the SVG tree drawing.
     """
 
     def verify_basic_svg(self, svg, width=200, height=200):
+        prefix = "{http://www.w3.org/2000/svg}"
         root = xml.etree.ElementTree.fromstring(svg)
-        self.assertEqual(root.tag, "{http://www.w3.org/2000/svg}svg")
+        self.assertEqual(root.tag, prefix + "svg")
         self.assertEqual(width, int(root.attrib["width"]))
         self.assertEqual(height, int(root.attrib["height"]))
+
+        # Verify the class structure of the svg
+        root_group = root.find(prefix + "g")
+        self.assertIn("class", root_group.attrib)
+        self.assertRegexpMatches(
+            root_group.attrib["class"], r"\b(tree|tree-sequence)\b"
+        )
+        if "tree-sequence" in root_group.attrib["class"]:
+            trees = root_group.find(prefix + "g")
+            self.assertIn("class", trees.attrib)
+            self.assertRegexpMatches(trees.attrib["class"], r"\btrees\b")
+            first_tree = trees.find(prefix + "g")
+            self.assertIn("class", first_tree.attrib)
+            self.assertRegexpMatches(first_tree.attrib["class"], r"\btree\b")
+        else:
+            first_tree = root_group
+        # Check that we have edges, symbols, and labels groups
+        groups = first_tree.findall(prefix + "g")
+        self.assertGreater(len(groups), 0)
+        for group in groups:
+            self.assertIn("class", group.attrib)
+            cls = group.attrib["class"]
+            self.assertRegexpMatches(cls, r"\b(edges|symbols|labels)\b")
+            if "symbols" in cls or "labels" in cls:
+                # Check that we have nodes & mutations subgroups
+                for subgroup in group.findall(prefix + "g"):
+                    self.assertIn("class", subgroup.attrib)
+                    subcls = subgroup.attrib["class"]
+                    self.assertRegexpMatches(subcls, r"\b(nodes|mutations)\b")
 
     def test_draw_file(self):
         t = self.get_binary_tree()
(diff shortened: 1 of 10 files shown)


hyanwong:
No tests added, but there were none in this area before, either, AFAIK.

hyanwong:
By the way, I haven't done as @jeromekelleher suggested and "put the id(ts) as prefix to all the values", because it is perfectly possible that someone might try to draw the same tree (or tree sequence) in a notebook but with different parameters, e.g. for height and width. We *need* an id for the trees in a svg tree sequence, because the id is used to position the tree along the x-axis. But if we use IDs that are the same for a given tree seq then we hit the same problem of duplicate SVG ids for what should be different drawings.

Instead, I have a small function to generate a (pretty much) unique UID suffix for the trees in a ts. I could increase the number of characters in the UID, but at the moment it generates one of ~10^14 possible identifiers, so I think the chances of clashing are low.

We needn't worry about the obscureness of the notation, I think, as it's an internal thing only. If we want to target a particular SVG drawing (tree or tree sequence) my new `svg_id` parameter allows a user-defined id to be passed in, which is then a handle to the correct image.

Ping @benjeffery as I think he knows about this stuff.

hyanwong:
Just modified the last commit so that we are guaranteed all svg tree ids in a TS are unique. Overkill, but hey...

hyanwong:
I'm being stupid here. If we place the trees in a drawn tree sequence in their own group, rather than using SVG `<def>`s, then we don't need a unique ID at all for each tree, and can get rid of all that messy UID stuff. This is much more sensible, as we aren't reusing the tree plots anyway (which is the point of `<def>`). So I've re-written the last commit to remove any need for IDs at all. I've checked and this does solve #467

petrelharp:
Oh good! That makes much more sense.

hyanwong:
While I'm on it, I would also like to be able to rotate the text labels by 90 degrees (see https://github.com/tskit-dev/tskit/issues/26). It would be nice to allow this by simply changing the CSS, but it turns out to be a little complex (see [here](https://stackoverflow.com/questions/61425345/rotate-svg-text-around-central-text-position-using-css/61430437)). The simplest solution would be to wrap all text elements in a `<g>` tag, so for *each label*, rather than outputting

`<text class="n14" x="70.15" y="129.62">14</text>`

we would use the more verbose

`<g transform="translate(70.15 129.62)"><text class="n14">14</text></g>`

But then we could style specific types of labels with e.g.

`.labels .nodes .sample text {transform:rotate(90deg)}`

I think the extra size in the svg is probably worth it. Opinions @benjeffery / @jeromekelleher ?

python/tskit/drawing.py
```diff
@@ -166,7 +166,19 @@ def remap(original_map, new_key, none_value):
 
 class SvgTreeSequence:
     """
-    Draw a TreeSequence in SVG.
+    An SVG representation of a tree sequence. The elements in the tree sequence are
+    placed into different `SVG groups <https://www.w3.org/TR/SVG2/struct.html#Groups>`_
```
jeromekelleher, line 170: This documentation isn't exposed to the user (it's a private internal class), so there's no point in elaborating here. Need to move this to the ``draw_svg`` function.

python/tskit/drawing.py
```diff
@@ -258,12 +273,77 @@ def __init__(
 
 class SvgTree:
     """
-    An SVG representation of a single tree.
+    An SVG representation of a single tree. The elements in the tree are placed into
+    different `SVG groups <https://www.w3.org/TR/SVG2/struct.html#Groups>`_ for easy
```
jeromekelleher, line 277: Ditto here on the docs.

python/tskit/drawing.py
```diff
@@ -258,12 +273,77 @@ def __init__(
...
+
+    .. code-block:: css
+
+        .tree .labels .nodes text:not(.sample) {visibility: hidden}
 
-    TODO should provide much more SVG structure which we document fully
-    to that the SVG elements can be manipulated directly by the user.
-    For example, every edge should be given an SVG ID so that it can
-    be referred to and modified.
+    You can also change the format of various items: the following styles will colout the
```
jeromekelleher, line 324: s/colout/colour

python/tskit/drawing.py
```diff
@@ -280,8 +360,10 @@ def __init__(
         node_label_attrs=None,
         mutation_attrs=None,
         mutation_label_attrs=None,
+        svg_id=None,
```
jeromekelleher, line 363: I thought we were moving away from using SVG IDs now? It doesn't seem to be used anyway, so let's remove it unless you have some specific plan.

jeromekelleher:
> I think the extra size in the svg is probably worth it. Opinions @benjeffery / @jeromekelleher ?

I don't think we need to worry too much about SVG size. If we have a useful feature we want, and it needs less than quadratic space then I say we add it.

python/tskit/drawing.py
hyanwong: Yes, thanks. I wasn't sure where to put them, so dumped them there for the time being. I'll hoick them out and put them in the proper place.

python/tskit/drawing.py
hyanwong: While I'm at it, can I make most of the params (perhaps after `size`) keyword only? Another param that might be useful is `styles` to embed your own stylesheet info into the top of the svg, and take advantage of these nice new features without having to hand-edit the svg or embed it in an html file. Do you want me to add this to the current PR @jeromekelleher , or submit it as another?

jeromekelleher: Let's do any further svg extensions  as a follow up. Don't worry about the interfaces within the drawing.py module, these aren't exposed anyway. Making things kw only is probably a good idea in draw_svg etc.

jeromekelleher:
Also, can you update the CHANGELOG to highlight this new feature (& bugfix)?

benjeffery:
`<g>` are generally a good thing in SVG. I wouldn't worry about adding them. For example when doing animations changes to `<g>` generally give a much better frame rate than other elements.

Also def needs tests. At a bare minimum comparing the raw html string output to known good. This lets us update deps etc. with confidence.

python/tskit/drawing.py
hyanwong: @benjeffery - re styles, I was imagining being able to pass e.g. `.nodes .n10 {stroke: red}` to the draw_svg function somehow, which would create

`<svg... ><style>.nodes .n10 {stroke: red}</style>...</svg>`

I don't think that can be done as an attribute on the svg tag, can it?

hyanwong:
> `<g>` are generally a good thing in SVG. I wouldn't worry about adding them. For example when doing animations changes to `<g>` generally give a much better frame rate than other elements.

I agree. Do you want me to add the `<g>` on text as a separate PR, or roll it into this one?

> Also def needs tests. At a bare minimum comparing the raw html string output to known good. This lets us update deps etc. with confidence.

happy to add a known string test. Will start coding that up now.

benjeffery:
> happy to add a known string test. Will start coding that up now.
Parsing the XML is _much_ better as it ignores whitespace and attribute changes.
Using `xmlunittest` is an option:
```python
from xmlunittest import XmlTestCase
class TestSVG(XmlTestCase):
    test_case.assertXmlEquivalentOutputs(xml_string, expected_xml_string)
```

python/tskit/drawing.py
benjeffery: Nope, you'd need another param. If a use case here is stand-alone SVG then this makes sense - if it is just for HTML then the caller can add css themselves.

hyanwong: Yes, I do think that some people will want to create stand-alone SVGs. If we add a "styles" parameter, this would save them the necessity of hand-editing the SVG, which I think is worth it, especially if we are advising people that this is a simple way to e.g. hide certain labels etc.

hyanwong:
@benjeffery I think this is ready to review. It seems to be failing appveyor because the appveyor config is not reading `requirements/CI/requirements.txt` and so is not installing `xmlunittest`. Not sure why Appveyor is set up like this: do you know, and also how to fix it?

benjeffery:
@hyanwong  Appveyor installs from python\requirements\conda-minimal.txt

benjeffery:
@hyanwong actually this is just for testing right? Can be pip installed in the appveyor config.

hyanwong:
Yes, it's just for testing, which is why I put it in CI and requirements/development.txt only. If we are testing on appveyor, shouldn't we use one of those? But I could put it in the appveyor conf.

benjeffery:
There are already a few of these types of imports in the appveyor config. This is because we use conda to install the deps in windows-land. Installing the whole of `development.txt` just exposes us to windows install issues for no point.

hyanwong:
> There are already a few of these types of imports in the appveyor config. This is because we use conda to install the deps in windows-land. Installing the whole of `development.txt` just exposes us to windows install issues for no point.

Ah, OK. I'll add it then

benjeffery:
One more thought - I think we should deprecate the old `draw` method.

python/tskit/trees.py
```diff
@@ -4024,10 +4202,124 @@ def trim(self, record_provenance=True):
...
+                .tree .labels .nodes text:not(.sample) {visibility: hidden}
+
+            See :meth:`Tree.draw_svg` for further details.
+
+        :param str path: The path to the file to write the output. If None, do not write
+            to file.
+        :param size: A tuple of (width, height) giving the width and height of the
+            produced SVG drawing in abstract user units (usually interpreted as pixels on
+            display). If None, defaults of height=:math:`200` and
+            width=:math:`200 \times num_trees` are used.
```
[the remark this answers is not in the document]
hyanwong: Oh yes, thanks for catching. Not sure this is the best way to display it, but couldn't think of a better

python/tskit/trees.py
```diff
@@ -4024,10 +4202,124 @@ def trim(self, record_provenance=True):
...
+        node_label_attrs=None,
+        mutation_label_attrs=None,
+        root_svg_attributes=None,
+        style=None,
+    ):
+        """
+        Return an SVG representation of a tree sequence.
+
+        When working in a Jupyter notebook, use the ``IPython.display.SVG`` function
+        to display the SVG output from this function inline in the notebook.
```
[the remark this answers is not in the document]
hyanwong: LGTM

python/tests/test_drawing.py
```diff
@@ -1134,16 +1137,44 @@ def test_max_tree_height(self):
...
 
 
-class TestDrawSvg(TestTreeDraw):
+class TestDrawSvg(TestTreeDraw, XmlTestCase):
     """
     Tests the SVG tree drawing.
     """
 
-    def verify_basic_svg(self, svg, width=200, height=200):
+    def verify_basic_svg(self, svg, width=200, height=200, new_routine=True):
```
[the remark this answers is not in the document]
hyanwong: Good point. I thought the old drawing code was independent, but now that I look at it, the old code is also a wrapper to the `SvgTree` function, so both should pass, and you are right. I'll remove this.

hyanwong:
> One more thought - I think we should deprecate the old `draw` method.

But only for format="svg". How do we do that?

python/tskit/drawing.py
```diff
@@ -467,19 +506,23 @@ def draw(self):
                     labels = self.right_labels
                 else:
                     labels = self.left_labels
-            # TODO add ID to node label text.
             # TODO get rid of these manual positioning tweaks and add them
```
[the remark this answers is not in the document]
hyanwong: Oh, sorry, I added the thumbsup, but misunderstood. The offsets aren't in a transform at the moment: we still have right_labels, mid_labels, & left_labels. It would be better to add classes, I think, but I don't want to fix this right now, so am leaving it as-is. Hence the manual-positioning TODO still stands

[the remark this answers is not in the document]
benjeffery: OK I get it now!

python/tests/test_drawing.py
```diff
@@ -1134,16 +1137,44 @@ def test_max_tree_height(self):
...
+                trees = root_group.find(prefix + "g")
+                self.assertIn("class", trees.attrib)
+                self.assertRegexpMatches(trees.attrib["class"], r"\btrees\b")
+                first_tree = trees.find(prefix + "g")
+                self.assertIn("class", first_tree.attrib)
+                self.assertRegexpMatches(first_tree.attrib["class"], r"\btree\b")
+            else:
+                first_tree = root_group
+            # Check that we have edges, symbols, and labels groups
+            for group in first_tree.findall(prefix + "g"):
```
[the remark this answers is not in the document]
hyanwong: Fixed with most recent commit

benjeffery:
> But only for format="svg". How do we do that?
I was thinking to deprecate the whole of `draw` in favour of `draw_text` and `draw_svg`.

hyanwong:
> But only for format="svg". How do we do that?
> I was thinking to deprecate the whole of `draw` in favour of `draw_text` and `draw_svg`.

Do you want to do this in this PR or a later change?

benjeffery:
Later - I'll add an issue!

hyanwong:
PS - @benjeffery : I guess I should squash all these commits before you merge?

benjeffery:
Yep, squash to what ever makes sense to you, also needs a rebase.

hyanwong:
> There are also a couple of uncovered code paths in the tests that would be nice to hit

I caught a couple, but couldn't understand why it was complaining about on line 423 though.

python/tests/test_drawing.py
```diff
@@ -1483,3 +1514,156 @@ def test_tree_height_scale(self):
         for bad_scale in [0, "", "NOT A SCALE"]:
             with self.assertRaises(ValueError):
                 ts.draw_svg(tree_height_scale=bad_scale)
+
+    def test_known_svg_tree(self):
+        tree = self.get_nonbinary_tree(sample_size=3)
+        svg = tree.draw_svg(
+            root_svg_attributes={"id": "AB"}, style=".edges {stroke: red}"
+        )
+        expected_svg = b"""<?xml version="1.0" encoding="utf-8" ?>
```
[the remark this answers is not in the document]
jeromekelleher: We'll need to keep the trees in a file then as well - they're msprime simulations and not guaranteed to stay the same over time or across platforms. But we can keep the text descriptions in ``data/examples`` or something, which will work fine.

python/tskit/drawing.py
```diff
@@ -178,14 +188,26 @@ def __init__(
...
+        root_svg_attributes=None,
+        style=None,
     ):
         self.ts = ts
         if size is None:
             size = (200 * ts.num_trees, 200)
+        if root_svg_attributes is None:
+            root_svg_attributes = {}
+        if max_tree_height is None:
+            max_tree_height = "ts"
```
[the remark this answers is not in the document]
jeromekelleher: Bitter experience has shown that putting in hard coded values as the default in complicated functions is a Bad Idea. The problem is that you have no way of knowing whether the user specifically asked for "ts" or just said "give me the default". Unless you are **certain** that this will always be the appropriate default, in the presence of all possible future arguments and their possible value assignments, using ``None`` as the default is the way to go.

python/tskit/drawing.py
```diff
@@ -200,34 +222,34 @@ def __init__(
...
+        for svg_tree, tree in zip(svg_trees, ts.trees()):
+            svg_tree.root_group["transform"] = f"translate({x} {y})"
+            trees.add(svg_tree.root_group)
             ticks.append((x, tree.interval[0]))
             x += tree_width
         ticks.append((x, ts.sequence_length))
 
-        dwg = self.drawing
-
         # # Debug --- draw the tree and axes boxes
```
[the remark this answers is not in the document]
jeromekelleher: It is quite handy - maybe a debug flag is a good idea.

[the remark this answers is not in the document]
jeromekelleher: But we don't have to deal with this here I think.

python/tskit/trees.py
```diff
@@ -1194,10 +1194,191 @@ def draw_text(self, orientation=None, **kwargs):
...
+        """
+        Return an SVG representation of a single tree.
+
+        When working in a Jupyter notebook, use the ``IPython.display.SVG`` function
+        to display the SVG output from this function inline in the notebook::
+
+            >>> SVG(tree.draw_svg())
+
+
+        .. note::
```
jeromekelleher, line 1223: I don't think we need a "note" block for this, it's most of the documentation.

python/tskit/trees.py
```diff
@@ -1194,10 +1194,191 @@ def draw_text(self, orientation=None, **kwargs):
...
+            not have a label.
+        :type node_labels: dict(int, str)
+        :param mutation_labels: If specified, show custom labels for the
+            mutations (specified by ID) that are present in the map; any mutations
+            not present will not have a label.
+        :type mutation_labels: dict(int, str)
+        :param node_attrs: Set custom attributes for specific node symbols. Each key in
+            the `node_attrs` map is a node id; the corresponding value must be a
+            dictionary specifying additional parameters passed to the
+            :meth:`svgwrite.drawing.Drawing.circle` method.
```
jeromekelleher, line 1336: I'm not sure we want to tie ourselves either to a specific shape for nodes/mutation or to the ``svgwrite`` package long term. I'd much prefer to keep it generic and defined in terms of SVG.

The simpler thing to do there is to just drop these ``x_attrs`` parameters from the API for now.

python/tskit/trees.py
```diff
@@ -4024,10 +4206,127 @@ def trim(self, record_provenance=True):
...
+        """
+        Return an SVG representation of a tree sequence.
+
+        When working in a Jupyter notebook, use the ``IPython.display.SVG`` function
+        to display the SVG output from this function inline in the notebook::
+
+            >>> SVG(tree.draw_svg())
+
+
+        .. note::
```
jeromekelleher, line 4235: Same comment about the "note" block

python/tskit/trees.py
```diff
@@ -4024,10 +4206,127 @@ def trim(self, record_provenance=True):
...
+            produced SVG drawing in abstract user units (usually interpreted as pixels on
+            display). If None, defaults to a width of
+            :math:`200 \\times \\textrm{num_trees}` and a height of :math:`200`.
+        :type size: tuple(int, int)
+        :param str tree_height_scale: Control how height values for nodes are computed.
+            If this is equal to ``"time"``, node heights are proportional to their time
+            values (this is the default). If this is equal to ``"log_time"``, node
+            heights are proportional to their log(time) values. If it is equal to
+            ``"rank"``, node heights are spaced equally according to their ranked times.
+        :param str,float max_tree_height: The maximum tree height value for each tree in
```
jeromekelleher, line 4266: Why do we want to add this parameter to TreeSequence.draw? I don't think it makes any sense to have different Y-scales per tree. We want to have a Y-axis showing time at some point, and having this option would break that.

jeromekelleher:
Looks great, thanks @hyanwong! I think we should scale the ambition back a little bit, and remove some of the parameters from the documented API for now (particularly the x_attrs params), so that we can think about how to structure things a bit more generally.

I'm also hesitant about documenting the exact pixel-sizes of the layouts, as I view the current hard-coded approach as a temporary hack until a nice relative layout scheme can be devised. The guiding principle here is, don't commit us to supporting anything long-term unless we're absolutely sure that it's a good long-term solution that we'll never want to change.

python/tests/test_drawing.py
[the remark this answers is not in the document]
hyanwong: Can we just use one of the hard coded examples like the one in test_simple_tree_sequence, and add some mutations by hand? Then we aren’t dependent on msprime for these external svgs.

python/tskit/drawing.py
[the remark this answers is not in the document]
hyanwong: So should I remove the commented out bit or not?

python/tskit/trees.py
hyanwong: ISWYM. Shall I simply undocument them, but leave them hidden in there?

python/tskit/trees.py
hyanwong: It was in before, but happy to remove. The only reason is you might want to compare the topology between trees in a ts without caring about the heights, and so scale all trees to the same height so that the short ones are still readable. So height=rank might make sense here.

jeromekelleher: Let's remove it for now to simplify things here.

python/tskit/trees.py
jeromekelleher: Yeah, just undocument. That way we can use them and develop them a bit before commiting to a long-term API

hyanwong:
> I'm also hesitant about documenting the exact pixel-sizes of the layouts, 

Are we committing to keeping those sizes though? Could we not say “Currently the defaults are xxx but this may be subject to change in future releases”?

jeromekelleher:
> Are we committing to keeping those sizes though? Could we not say “Currently the defaults are xxx but this may be subject to change in future releases”?

No, but documentation like this has a nasty habit of going out of date because we forget about it. I wouldn't put in any details that we aren't confident are stable. It's not an important detail that users need to know.

python/tskit/trees.py
hyanwong: Removed from the `svg_draw()` exposed function, although I left it in the underlying one for simplicity.

python/tskit/trees.py

python/tests/test_drawing.py
```diff
@@ -1177,7 +1208,7 @@ def test_draw_file(self):
     def test_draw_defaults(self):
         t = self.get_binary_tree()
         svg = t.draw()
-        self.verify_basic_svg(svg)
+        self.verify_basic_svg(svg, new_routine=False)
```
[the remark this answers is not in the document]
hyanwong: Removed

python/tests/test_drawing.py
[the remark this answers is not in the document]
hyanwong: Changed to the hard-coded example, with a single mutation

python/tskit/drawing.py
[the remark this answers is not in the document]
hyanwong: I've added an extra TODO comment to point out that the commented out part should be activated with a debug flag.

hyanwong:
This is all done. If you are happy @jeromekelleher and @benjeffery, I'll squash the last commit which addresses the last concerns into the one before it, check if it needs rebaseing, and then someone can merge.

benjeffery:
A thought on customising I had today (not for this PR) - one way to do it would be to have callback params that allow functions to be specified that return HTML for nodes and edges etc. The callback would be called with the node/edge/mutation data and the default HTML entity. The callback then either makes a new HTML entity, or modifies the default and returns it.

hyanwong:
An observation that any styling passed in to one SVG on a Jupyter notebook will affect *all* the trees in the notebook (as per a comment I saw: "scoped CSS has been removed from the specs, so any `<style>` tags in the document will be hoisted to the global scope"). The only way to stop this is to given an ID to the SVG and target the style at that SVG. We should probably give an example of this in the docs, as it's a bit unexpected to the uninitiated.

benjeffery:
@hyanwong A quick added sentence on the style param docs should do it.

jeromekelleher:
This is great, thanks @hyanwong, let's get this one merged in ASAP. Looks like we're basically there.

benjeffery:
Yep, think it is just the doc mention of style hoisting.

hyanwong:
Added note about styles and squashed the PR changes into the last commit. Should be OK to merge I guess @benjeffery

benjeffery:
@hyanwong I see you've just made some changes, is this ready for a once over as am keen to merge this so the others can proceed?

hyanwong:
It was probably accidental. I did squash the last PR corrections into the previosu commit, but it should otherwise be the same. I think I'm confused as to what I did. Can you pinpoint any changes?

hyanwong:
It's frustrating working on 5 interrelated PRs at once. Especially as I can't always do them on the same computer at the moment. So I never remember which local branches are pushed, etc etc. I think I've whinged to you before that I can't really get git to suit my workflow (which tends to be to not complete anything if I can see something shiny elsewhere)

hyanwong:
@benjeffery it's definitely ready for a final look, and hopefully a merge, though. I did have to rebase too, as Daniel & Sunack's changes went in before this was sorted.

benjeffery:
Ok, I'll give it the once over. Are you happy for me to make any small changes myself as might be late by the time I am done? (will leave anything big till the morning, although don't expect that to be the case!)

Lets talk about git workflow on a call - would be good if we can make this easier for you.

hyanwong:
Wow, you're working late. I had gone to bed. Sorry!

PR #562, opened 2020-04-28
Add python 3.8 to travis

benjeffery:
I have a local test failure when using 3.8. Seeing if the CI also hits this. We should test 3.8 on at least one of the CI's anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ language: python
 matrix:
   include:
     - os: linux
+      python: 3.8
+    - os: linux
       python: 3.7
     - os: osx
       language: generic
@@ -41,7 +43,7 @@ install:
       conda install -c conda-forge gxx_linux-64;
     fi
   - conda install --yes --file=python/requirements/conda-minimal.txt
-  - conda install -c bioconda --yes pysam
+  #- conda install -c bioconda --yes pysam
   # codecov requires a more recent version of certifi, but we can't uninstall the old one
   # see https://github.com/tskit-dev/tskit/issues/366
   - pip install --ignore-installed certifi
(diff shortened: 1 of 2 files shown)


benjeffery:
Note that all the travis builds proceed in parallel, so this won't slow things down.

benjeffery:
The problem was a broken test that silently passed in 3.6/7.

python/tests/test_tables.py
```diff
@@ -1320,7 +1320,7 @@ def test_incompatible_edges(self):
         tables2 = ts2.dump_tables()
         tables2.edges.set_columns(**tables1.edges.asdict())
         # The edges in tables2 will refer to nodes that don't exist.
-        self.assertRaises(_tskit.LibraryError, tables1.sort())
```
jeromekelleher, line 1323: Jeez, how did this pass! The 
```
with self.assertRaises(xx):
    do_thing()
```
notation is no much better.

.travis.yml
```diff
@@ -41,7 +43,7 @@ install:
       conda install -c conda-forge gxx_linux-64;
     fi
   - conda install --yes --file=python/requirements/conda-minimal.txt
-  - conda install -c bioconda --yes pysam
+  #- conda install -c bioconda --yes pysam
```
jeromekelleher, line 46: How are we able to do without pysam? Ah, I see, we're probably getting it from pip. Well, that'll be brittle I'm guessing, and it's a heck of a big compile if they don't have binary wheels built.

Might be easier to just bump to 3.8 on Appveyor and leave the travis config alone? We're already excluding a bunch of awkward stuff on Windows anyway.

benjeffery: This line is commented out (was planning to remove but wanted the discussion) because pysam via conda doesn't support 3.8. You're right that it is coming in from pip later on.
Build log on linux:
`Using cached pysam-0.15.4-cp38-cp38-manylinux2010_x86_64.whl (12.0 MB)`
And OSX:
`Downloading pysam-0.15.4-cp36-cp36m-macosx_10_6_intel.whl`

So I think this is OK.

python/tests/test_tables.py
benjeffery: It seems `None` as the callable just skips the test!! Can see why 3.8 changed that!

jeromekelleher:
OK, great. Merging.

PR #570, opened 2020-05-02
Grouped svg

hyanwong:
This implements the idea by @benjeffery of grouping SVG components using the tree hierarchy, which should allow subtrees to be styled and even moved using SVG. It is a continuation of #555, and it could be argued that it should be rolled into that, as it will require the documentation proposed in #555 to be changed. 

To see how it might work, try the following example:

```
ts = msprime.simulate(20, mutation_rate=2, random_seed=1)
# Colour all descendant edges & mutations descended from parent node 37
SVG(ts.first().draw_svg(size=(400, 400), style=".mut {fill: grey} g.p37 .edge {stroke: red} g.p37 .mut {fill: red}"))
```
Screenshot 2020-05-02 at 19 36 52

The implementation means that we have to abandon the idea of putting edges, symbols, & text in different groups. Instead we ensure that text & symbols are above edges within each group by putting edges as the first group item. This seems to work OK: for instance

```
ts = msprime.simulate(20, mutation_rate=2, random_seed=1)
# Colour all nodes yellow to show they are in from of black edge lines
SVG(ts.first().draw_svg(size=(400, 400), style=".node circle {r:6px; fill: yellow; stroke: black}"))
```

Screenshot 2020-05-02 at 19 38 00

### DIFF
--- a/python/tests/data/svg/tree.svg
+++ b/python/tests/data/svg/tree.svg
@@ -1,70 +1,47 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<svg baseProfile="full" height="200" id="XYZ" version="1.1" width="200" xmlns="http://www.w3.org/2000/svg" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xlink="http://www.w3.org/1999/xlink" baseProfile="full" height="200" id="XYZ" version="1.1" width="200">
   <defs>
-    <style type="text/css">
-<![CDATA[.edges {stroke: blue}]]>    </style>
+    <style type="text/css"><![CDATA[.axis {font-weight: bold}.tree, .axis {font-size: 14px; text-anchor:middle;}.edge {stroke: black; fill: none}.node > circle {r: 3px; fill: black; stroke: none}.tree text {dominant-baseline: middle}.mut > text.lft {transform: translateX(0.5em); text-anchor: start}.mut > text.rgt {transform: translateX(-0.5em); text-anchor: end}.root > text {transform: translateY(-0.8em)}.leaf > text {transform: translateY(1em)}.node > text.lft {transform: translate(0.5em, -0.5em); text-anchor: start}.node > text.rgt {transform: translate(-0.5em, -0.5em); text-anchor: end}.mut {fill: red; font-style: italic}]]></style>
+    <style type="text/css"><![CDATA[.edges {stroke: blue}]]></style>
   </defs>
   <g class="tree t0">
-    <g class="edges" fill="none" stroke="black">
-      <path class="p9 c4" d="M 64.0 169.6231433344194 V 30.0 H 100.0"/>
-      <path class="p4 c0" d="M 46.0 170.0 V 169.6231433344194 H 64.0"/>
-      <path class="p4 c1" d="M 82.0 170.0 V 169.6231433344194 H 64.0"/>
-      <path class="p9 c5" d="M 136.0 168.29406477810417 V 30.0 H 100.0"/>
-      <path class="p5 c2" d="M 118.0 170.0 V 168.29406477810417 H 136.0"/>
-      <path class="p5 c3" d="M 154.0 170.0 V 168.29406477810417 H 136.0"/>
-    </g>
-    <g class="symbols">
-      <g class="nodes">
-        <circle class="n9" cx="100.0" cy="30.0" r="3"/>
-        <circle class="n4" cx="64.0" cy="169.6231433344194" r="3"/>
-        <circle class="n0 sample" cx="46.0" cy="170.0" r="3"/>
-        <circle class="n1 sample" cx="82.0" cy="170.0" r="3"/>
-        <circle class="n5" cx="136.0" cy="168.29406477810417" r="3"/>
-        <circle class="n2 sample" cx="118.0" cy="170.0" r="3"/>
-        <circle class="n3 sample" cx="154.0" cy="170.0" r="3"/>
-      </g>
-      <g class="mutations" fill="red">
-        <rect class="m0 s0 n4" height="6" transform="translate(-3 -3)" width="6" x="64.0" y="99.8115716672097"/>
-      </g>
-    </g>
-    <g class="labels" dominant-baseline="middle" font-size="14">
-      <g class="nodes">
-        <g text-anchor="start">
-          <g transform="translate(141.0, 163.29406477810417)">
-            <text class="n5">5</text>
-          </g>
+    <g class="node n9 root" transform="translate(100.0 30.0)">
+      <g class="m0 node n4 p9 s0" transform="translate(-36.0 139.623)">
+        <g class="leaf node n0 p4 sample" transform="translate(-18.0 0.376857)">
+          <path class="edge" d="M 0 0 V -0.376857 H 18.0"/>
+          <circle cx="0" cy="0" r="1"/>
+          <text>0</text>
         </g>
-        <g text-anchor="middle">
-          <g transform="translate(100.0, 25.0)">
-            <text class="n9">9</text>
-          </g>
-          <g transform="translate(46.0, 190.0)">
-            <text class="n0 sample">0</text>
-          </g>
-          <g transform="translate(82.0, 190.0)">
-            <text class="n1 sample">1</text>
-          </g>
-          <g transform="translate(118.0, 190.0)">
-            <text class="n2 sample">2</text>
-          </g>
-          <g transform="translate(154.0, 190.0)">
-            <text class="n3 sample">3</text>
-          </g>
+        <g class="leaf node n1 p4 sample" transform="translate(18.0 0.376857)">
+          <path class="edge" d="M 0 0 V -0.376857 H -18.0"/>
+          <circle cx="0" cy="0" r="1"/>
+          <text>1</text>
         </g>
-        <g text-anchor="end">
-          <g transform="translate(59.0, 164.6231433344194)">
-            <text class="n4">4</text>
-          </g>
+        <path class="edge" d="M 0 0 V -139.623 H 36.0"/>
+        <circle cx="0" cy="0" r="1"/>
+        <text class="rgt">4</text>
+        <g class="mut m0 s0" transform="translate(0 -69.8116)">
+          <rect height="6" transform="translate(-3 -3)" width="6" x="0" y="0"/>
+          <text class="rgt">0</text>
         </g>
       </g>
-      <g class="mutations" font-style="italic">
-        <g text-anchor="start"/>
-        <g text-anchor="end">
-          <g transform="translate(59.0, 103.8115716672097)">
-            <text class="m0 s0 n4">0</text>
-          </g>
+      <g class="node n5 p9" transform="translate(36.0 138.294)">
+        <g class="leaf node n2 p5 sample" transform="translate(-18.0 1.70594)">
+          <path class="edge" d="M 0 0 V -1.70594 H 18.0"/>
+          <circle cx="0" cy="0" r="1"/>
+          <text>2</text>
         </g>
+        <g class="leaf node n3 p5 sample" transform="translate(18.0 1.70594)">
+          <path class="edge" d="M 0 0 V -1.70594 H -18.0"/>
+          <circle cx="0" cy="0" r="1"/>
+          <text>3</text>
+        </g>
+        <path class="edge" d="M 0 0 V -138.294 H -36.0"/>
+        <circle cx="0" cy="0" r="1"/>
+        <text class="lft">5</text>
       </g>
+      <circle cx="0" cy="0" r="1"/>
+      <text>9</text>
     </g>
   </g>
 </svg>
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1315,86 +1315,86 @@ class Tree:
 
             >>> SVG(tree.draw_svg())
 
-
-
-        The elements in the tree are placed into
-        different `SVG groups <https://www.w3.org/TR/SVG2/struct.html#Groups>`_ for
-        easy styling and manipulation. Both these groups and their component items
-        are marked with SVG classes so that they can be targetted. This allows
-        individual components of the drawing to be hidden, styled, or otherwise
+        The elements in the tree are grouped according to the structure of the tree,
+        using `SVG groups <https://www.w3.org/TR/SVG2/struct.html#Groups>`_. This allows
+        easy styling and manipulation of elements and subtrees. Elements in the SVG file
+        are marked with SVG classes so that they can be targetted, allowing
+        different components of the drawing to be hidden, styled, or otherwise
         manipulated. For example, when drawing (say) the first tree from a tree
         sequence, all the SVG components will be placed in a group of class ``tree``.
         The group will have the additional class ``t0``, indicating that this tree
         has index 0 in the tree sequence. The general SVG structure is as follows:
 
-        * The *tree* group (classes ``tree`` and ``tN`` where `N` is the tree index).
-          This contains the following three groups:
+        Each tree is contained in a group of class ``tree``. Additionally, this group
+        has a class ``tN`` where `N` is the tree index.
 
-          * The *edges* group (class ``edges``), containing edges. Each edge has
-            classes ``pX``, and ``cY`` where `X` and `Y` are the ids of the parent
-            and child nodes.
-          * The *symbols* group (class ``symbols``), containing two subgroups:
+        Within the ``tree`` group there is a nested hierarchy of groups corresponding
+        to the tree structure. Any particular node in the tree will have a corresponding
+        group containing child groups (if any) followed by the edge above that node, a
+        node symbol, and (potentially) text containing the node label. For example, a
+        simple two tip tree, with tip node ids 0 and 1, and a root node id of 2 will have
+        a structure similar to the following:
 
-            * The *node symbols* group (class ``nodes``) containing a
-              `circle <https://www.w3.org/TR/SVG2/shapes.html#CircleElement>`_ for
-              each node. Each node symbol has a class ``nX`` where ``X`` is the node
-              id. Symbols corresponding to sample nodes are additionally labelled
-              with a class of ``sample``.
-            * The *mutation symbols* group (class ``mutations``) containing a
-              `rectangle <https://www.w3.org/TR/SVG2/shapes.html#RectElement>`_ for
-              each mutation. Each mutation symbol has classes ``mX``, ``sY``, and
-              ``nZ`` where `X` is the mutation id, `Y` is the site id, and `Z` is
-              the id of the node above which the mutation occurs.
+        .. code-block::
 
-          * The *labels* group (class ``labels``) containing two subgroups:
-
-            * The *node labels* group (class ``nodes``) containing text for each
-              node. Each `text <https://www.w3.org/TR/SVG2/text.html#TextElement>`_
-              element in this group corresponds to a node, and has the same set of
-              classes as its equivalent node symbol (i.e. ``nX`` and potentially
-              ``sample``)
-            * The *mutation labels* group (class `mutations`) containing containing
-              text for each mutation. Each
-              `text <https://www.w3.org/TR/SVG2/text.html#TextElement>`_ element in
-              this group corresponds to a mutation, and has the same set of classes
-              as its equivalent mutation symbol (i.e. ``mX``, ``sY``, and ``nZ``)
+            <g class="tree t0">
+              <g class="node n2 root">
+                <g class="node n1 p2 sample leaf">
+                  <path class="edge" ... />
+                  <circle />
+                  <text>Node 1</text>
+                </g>
+                <g class="node n0 p2 sample leaf">
+                  <path class="edge" ... />
+                  <circle />
+                  <text>Node 0</text>
+                </g>
+                <path class="edge" ... />
+                <circle />
+                <text>Root (Node 2)</text>
+              </g>
+            </g>
 
         The classes can be used to manipulate the element, e.g. by using
         `stylesheets <https://www.w3.org/TR/SVG2/styling.html>`_. Style strings can
         be embedded in the svg by using the ``style`` parameter, or added to html
         pages which contain the raw SVG (e.g. within a Jupyter notebook by using the
-        IPython HTML() function). As a simple example, the following style string
-        will hide all labels:
+        IPython ``HTML()`` function). As a simple example, passing the following
+        string as the ``style`` parameter will hide all labels:
 
         .. code-block:: css
 
-            .tree .labels {visibility: hidden}
+            .tree text {visibility: hidden}
 
         You can also change the format of various items: the following styles will
-        display the symbols of the *sample* nodes only in blue, rotate the sample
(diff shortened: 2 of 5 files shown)


hyanwong:
Note that since the label and the symbol are now quite tightly linked, the mutation labels naturally default to being displayed in red text, matching the symbols. I left this in, as it seemed possibly a better viz idea anyway, but we can style them back if necessary.

petrelharp:
Sorry, basic question: how do we get the `SVG( )` command you're using to do this?

hyanwong:
> Sorry, basic question: how do we get the `SVG( )` command you're using to do this?

Oh, sorry, this is in a jupyter notebook:

```
import msprime
from IPython.display import SVG
```

But the exported svg file should be viewable in a browser like chrome, in which case, outside of Jupyter, simply do

```
ts.first().draw_svg("output.svg", size=(400, 400), style="circle.node {r:6px; fill: yellow; stroke: black}")
```

to get `output.svg` which you can open in a browser

hyanwong:
Even on a medium-sized TS (200 samples) this only increases the file size by a quarter or less, so I would say it's worth it. Not sure whether to add it to #555 though - thoughts @benjeffery ?

petrelharp:
Wow, this looks great. I haven't looked into the prevoius, non-hierarchical implementation, but... pooking around at the SVG, here's a suggestion: seems like the mutation labels should be in the same `g` as the mutations themselves? that might make label placement easier also or at least more elegant?

hyanwong:
> seems like the mutation labels should be in the same `g` as the mutations themselves?

Yep, fair point. I was wary of increasing the number of groups any more than I had already, but, in for a penny, in for a pound, as they say here. That could also apply to the node + node label, I guess. In fact, it probably will mean slightly smaller files, as I won't need to duplicate the class names: I can just put them on the enclosing group.

More likely than changing individual mutations, it seems likely that people might want to do something with *all* the mutation labels, or *all* the symbols, but the classes allow us to do that anyway.

hyanwong:
More improvements just committed, which give much greater flexibility. Moreover, they result in a svg file size which is *smaller* (in bytes) than the current versions on master, because of using transformations on the groups, which leads to integer positioning. I have also extracted most of the hard-coded positioning hacks and placed them as css transformations, which can therefore be easily overridden. 

The following Jupyter notebook script demos a few examples:

```
import msprime
from IPython.display import display, SVG, HTML
ts = msprime.simulate(20, mutation_rate=2, random_seed=1)


display(
    HTML("<h1>Default setting</h1>"),
    SVG(ts.first().draw_svg(size=(400, 400))),

    HTML("<h1>Demos styling parts of the tree</h1>"),
    HTML("<p>Here we give an #id to the SVGs (e.g. EX1) so that our style changes only apply to one specific tree</p>"),

    HTML("<h3>1) Edges are hidden behind node & mutation symbols</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX2"},
        style="#EX2 .node {r:6px; fill: yellow; stroke: black}")),

    HTML("<h3>2) Target a specific node via id (n35)</h3>"),
    HTML("<p>note the child selector, '>', otherwise all descendants of 35 would be changed</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX3"},
        style="#EX3 .n35 > .node {r:6px; fill: yellow; stroke: black}")),

    HTML("<h3>3) Target all children of a node</h3>"),
    HTML("<p>Here we color edges. Targetting can be done by node id or e.g. as here, by id of a mutation (m4) above the node</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX4"},
        style="#EX4 .m4 .edge {stroke: red;} #EX4 .mut {fill: gray}")),

    HTML("<h3>4) Hide internal node labels via css</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX5"},
        style="#EX5 :not(.sample) > text {visibility: hidden;}")),

    HTML("<h3>5) Hide internal node labels and rotate leaf labels</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX6"},
        style="#EX6 .leaf > text {transform: translateY(10px) rotate(90deg); text-anchor: start} #EX6 :not(.sample) > text {visibility: hidden;}")),

    HTML("<h3>6) Highlight an edge via child and parent node ids</h3>"),
    HTML("<p>Using both child and parent is useful when plotting an entire ts</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX7"},
        style="#EX7 .p34.n33 > .edge {stroke-width:3; stroke: green}")),
)
```

benjeffery:
Looking good - haven't done a line by line yet, but glad this seems to be working well. I'm trying to think of use cases where this structure is worse than the previous, can you think of anything that is harder this way?

As for PRs, as #555 is close to done I say don't combine them.

hyanwong:
> Looking good - haven't done a line by line yet, but glad this seems to be working well. I'm trying to think of use cases where this structure is worse than the previous, can you think of anything that is harder this way?

I don't think it restrricts anythign but some stuff may be a little more complicated, mainly because of the way I've structured it. E.g. if you want to target a particular node *but not its children* you need to remember to use the "child only" selector, e.g. `.n1 > node` not simply `n1 node` to target the node symbol Also the way of selecting text on nodes but not on mutations might be thought a little obscure - I chose not to give (yet another) class to the node labels, as you can simply look for a label *next to* a node with `.node + text`.

> As for PRs, as #555 is close to done I say don't combine them.

Right. Howvever, if we are thinking of going in the direction of this PR, I wonder if I should remove the documentation about css classes from #555 @benjeffery , as it will be obsoleted by the changes here?

hyanwong:
> I'm trying to think of use cases where this structure is worse than the previous, can you think of anything that is harder this way?

Oh, another point is that I am now using styles to position the text relative to the node, rahter than hard-coding offsets. This opens up the problem that, if the SVG is embedded in an html document, the styles will apply over the entire document. We *could* anticipate this and assign each SVG a unique ID (if one wasn't provided by the user) and target all the styles to that ID, which would stop the styles "spilling out", as it were. Or we could just document that this is expected to happen given the SVG specs, and provide an example like the one above for people to follow if they expect to be embedding SVGs.

hyanwong:
Also 2 more ideas on which I'd quite like your opinion, @benjeffery (and possibly @jeromekelleher). 

1) I wonder if we should by default put the derived state and/or  in the mutation label, e.g. on the other side of the branch from the text? The example below is maybe a little over-complicated (`A` changed to `T` at site `0` (mutation id `#0`), but I imagine that for "real" users, the state change might be more relevant than the mutation id

Screenshot 2020-05-04 at 09 17 16


2) I am wanting to colour all the edges under a *mutation* in a colour. I had a sudden thought that we could treat the mutation like a unary node, and instead of drawing a line from a parent node directly to the child node, drawing through the (say) 2 mutations between them, we could draw it from the parent node to the first mutation, then another line from that mutation to the next mutation, and then finally the last line to the child node. Like a node, each mutation would also create a hierarchical group that could then be targetted. Does this sound sensible, or a silly overkill?

benjeffery:
> I wonder if I should remove the documentation about css classes from #555 @benjeffery , as it will be obsoleted by the changes here?

Not much point, just change them with this PR.

> This opens up the problem that, if the SVG is embedded in an html document, the styles will apply over the entire document.

I haven't looked at the exact way you're doing the text positioning but could you assign a unique class name and qualify all the styles that way?

> I wonder if we should by default put the derived state and/or in the mutation label

I like this, but I think we should nail down the customisation API first as otherwise we'll end up with loads of flags  for options like this. This example would be easy to do with a callback to customise mutation label, and could be included as an example?

> I had a sudden thought that we could treat the mutation like a unary node, and instead of drawing a line from a parent node directly to the child node

I like this, separate PR? (I know there are few piling up, but generally smaller and a few is better, although not sure I'm following my own advice here with metadata!)

hyanwong:
> > This opens up the problem that, if the SVG is embedded in an html document, the styles will apply over the entire document.
> 
> I haven't looked at the exact way you're doing the text positioning but could you assign a unique class name and qualify all the styles that way?

Yes, I could. The question is whether we want to do this invisibly to the user, or get them do do it explicitly. It seems a bit ugly taking their SVG styles and modifying them on the fly.

> I like this, but I think we should nail down the customisation API first as otherwise we'll end up with loads of flags for options like this. This example would be easy to do with a callback to customise mutation label, and could be included as an example?

Also easy to do using the `mutation_labels` dict which we already have. It's really a question of what we want to be the default.

> > I had a sudden thought that we could treat the mutation like a unary node, and instead of drawing a line from a parent node directly to the child node
> 
> I like this, separate PR?

OK. Will add it to this PR though, as it's integral to this one, I think.

benjeffery:
> > > This opens up the problem that, if the SVG is embedded in an html document, the styles will apply over the entire document.
> > 
> > 
> > I haven't looked at the exact way you're doing the text positioning but could you assign a unique class name and qualify all the styles that way?
> 
> Yes, I could. The question is whether we want to do this invisibly to the user, or get them do do it explicitly. It seems a bit ugly taking their SVG styles and modifying them on the fly.

Ahh, I misunderstood you meant you as a user, rather than internal to tskit.

petrelharp:
> This opens up the problem that, if the SVG is embedded in an html document, the styles will apply over the entire document.

Shouldn't the css be embedded in the SVG? Then (I'd think?) that it wouldnt affect other SVG in the same document? -  https://www.w3.org/TR/SVG/styling.html#StyleElementExample

petrelharp:
> I wonder if we should by default put the derived state and/or in the mutation label, e.g. on the other side of the branch from the text?

Gee, that'd be nice, but I'll bet it gets too busy very quickly. If there's an easy way to say "please give alleles instead of mutation ID" that might be useful?

> Like a node, each mutation would also create a hierarchical group that could then be targetted. Does this sound sensible, or a silly overkill?

Nice idea. So, it'd be easy to color all samples with a particular mutation, for instance. But I suppose that's already pretty easy, by looking up the mutation's node.  I'm leaning towards "overkill", but not for a good reason...

hyanwong:
> Shouldn't the css be embedded in the SVG? Then (I'd think?) that it wouldnt affect other SVG in the same document? - 

It is embedded in the SVG, but I'm afraid your gut thought is wrong - stylesheets are always taken as applying to the entire document, even if embedded deep within an element. Or as https://www.w3.org/TR/SVG/styling.html#StyleSheetsInHTMLDocuments says, "any HTML ‘style’ element used in an SVG document must also apply its style sheet to the (entire) document". Also see multiple questions on Stack Overflow, e.g. see the end of the answer at https://stackoverflow.com/questions/20078032/inline-svg-embedded-css-interpreted-at-page-level

hyanwong:
> I'm leaning towards "overkill", but not for a good reason...

The specific reason I wanted this is to target the *fraction* of the edge that lies below a particular mutation. E.g. for focal mutation 5, I would like a picture like the one below:

Screenshot 2020-05-05 at 12 13 16

Not the one that we previously would have been stuck with:

Screenshot 2020-05-05 at 12 09 43

A minor different, but an improvement, I think. The change is coded up in https://github.com/tskit-dev/tskit/pull/570/commits/9f19cddd178cc3cf90f013eada0819082f745a81. It means that the styles defined at the top of this issue will need changing, however. I'll post equivalents below, although I think the class labels might need a bit or work to make styling easier (I should chat about this with @benjeffery )

hyanwong:
Current equivalent style commands with the new commit:

```
import msprime
from IPython.display import display, SVG, HTML
ts = msprime.simulate(20, mutation_rate=2, random_seed=1)


display(
    HTML("<h1>Default setting</h1>"),
    SVG(ts.first().draw_svg(size=(400, 400))),

    HTML("<h1>Demos styling parts of the tree</h1>"),
    HTML("<p>Here we give an #id to the SVGs (e.g. EX1) so that our style changes only apply to one specific tree</p>"),

    HTML("<h3>1) Edges are hidden behind node & mutation symbols</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX2"},
        style="#EX2 .node > .edge + * {r:6px; fill: yellow; stroke: black}")),

    HTML("<h3>2) Target a specific node via id (n35)</h3>"),
    HTML("<p>note the child selector, '>', otherwise all descendants of 35 would be changed</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX3"},
        style="#EX3 .node.n35 > .edge + * {r:6px; fill: yellow; stroke: black}")),

    HTML("<h3>3) Target all children of a node</h3>"),
    HTML("<p>Here we color edges. Targetting can be done by node id or e.g. as here, by id of a mutation (m4) above the node</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX4"},
        style="#EX4 .m4 * .edge {stroke: red;} #EX4 .mut {fill: gray}")),

    HTML("<h3>4) Hide internal node labels via css</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX5"},
        style="#EX5 .node:not(.sample) > text {visibility: hidden;}")),

    HTML("<h3>5) Hide internal node labels and rotate leaf labels</h3>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX6"},
        style="#EX6 .leaf > text {transform: translateY(10px) rotate(90deg); text-anchor: start} #EX6 .node:not(.sample) > text {visibility: hidden;}")),

    HTML("<h3>6) Highlight an edge via child and parent node ids</h3>"),
    HTML("<p>Using both child and parent is useful when plotting an entire ts</p>"),
    SVG(ts.first().draw_svg(
        size=(400, 400),
        root_svg_attributes={"id": "EX7"},
        style="#EX7 .p35.n31 > .edge {stroke-width:3; stroke: green}")),
)
```

petrelharp:
>  It is embedded in the SVG, but I'm afraid your gut thought is wrong

Hah, thanks for explaining.  I should have googled some more.  It does seem pretty unfortunate to have one tree's svg affecting other ones: that'll cause all sorts of bad things happen when including more than one tree on a page, no? You could add a random id at the top level and modify the css to make the properties only apply to *this* SVG?

benjeffery:
I don't think rewriting the CSS is trivial. I'd prefer a note to the user in the docs that they should qualify their CSS if needed.

python/tskit/trees.py
```diff
@@ -1194,10 +1194,155 @@ def draw_text(self, orientation=None, **kwargs):
...
+            (specified by ID) that are present in this map; any nodes not present will
+            not have a label.
+        :type node_labels: dict(int, str)
+        :param mutation_labels: If specified, show custom labels for the
+            mutations (specified by ID) that are present in the map; any mutations
+            not present will not have a label.
+        :type mutation_labels: dict(int, str)
+        :param dict root_svg_attributes: Additional attributes, such as an id, that will
+            be embedded in the root ``<svg>`` tag of the generated drawing.
+        :param str style: A
```
petrelharp, line 1328: should say here that this stuff applies to the entire doc it's in

petrelharp:
> The specific reason I wanted this is to target the fraction of the edge that lies below a particular mutation. E.g. for focal mutation 5, I would like a picture like the one below:

Damn, that's really nice. I say go for it.

petrelharp:
So: how would we add a `font_size` argument? This is specified in the css, so the only good way I can think of to do it without affecting other trees in the same document is to add a top-level ID that qualifies all the embedded css. How about (a) adding a `root_id` argument to the method, which if not specified is randomly generated, and (b) inserting it as a qualifier to the embedded css [here](https://github.com/tskit-dev/tskit/pull/570/files#diff-39c96ebaf61930e719f9425c17063ceeR276)? I'm not sure what you are saying, @benjeffery, but I'm not proposing dynamically rewriting the css, so if a user specifies css they've got to do the right qualifying themselves.

Sorry I'm a bit slow on all this, hope I'm not making too much noise.

python/tskit/trees.py
hyanwong: I've put this as a Note in #555: 

```
        ..note::
             A feature of SVG style commands is that they apply not just to the contents
             within the <svg> container, but to the entire file. Thus if an SVG file is
             embedded in a larger document, such as an HTML file (e.g. when an SVG
             is displayed inline in a Jupyter notebook), the style will apply to all SVG
             drawings in the notebook. To avoid this, you can tag the SVG with a unique
             SVG using ``root_svg_attributes={'id':'MY_UID'}``, and prepend this to the
             style string, as in ``#MY_UID .tree .edges {stroke: gray}``.
```

hyanwong:
> So: how would we add a `font_size` argument? 

I guess by `style = ".tree {font-size: 16pt}"` (which as you say will affect all trees in the notebook) or `style = "#MY_UID .tree {font-size: 16pt}"`, which will require you to supply `'id':MY_UID` to the `root_svg_attributes` parameter when drawing the SVG.

> This is specified in the css, so the only good way I can think of to do it without affecting other trees in the same document is to add a top-level ID that qualifies all the embedded css. How about (a) adding a `root_id` argument to the method, which if not specified is randomly generated, and (b) inserting it as a qualifier to the embedded css [here](https://github.com/tskit-dev/tskit/pull/570/files#diff-39c96ebaf61930e719f9425c17063ceeR276)? 

Right: for the styles we create ourselves, we can add the ID. But unless we have a `font_size` argument, the user will *always* be stuck with the pre-defined font size for a tree, unless they explicitly override it using the `styles=` param, when they would have to hand-insert the ID anyway! In other words, all the trees in a notebook will have the same (default) style anyway, so we don't really care that the styles from one is (under the hood) overriding the styles from other. It's only when you want to change *one* tree that you start seeing problems, in which case you are specifying your own `style` string anyway, and will have to insert an ID if you don;t want the styles to spill over.

> I'm not sure what you are saying, @benjeffery, but I'm not proposing dynamically rewriting the css, so if a user specifies css they've got to do the right qualifying themselves.

The problem here is that we then need to somehow return the ID we autogenerated, so that the user can use it. Which seems a bit messy to me. Why not just let them define it themselves?

hyanwong:
> Damn, that's really nice. I say go for it.

Thanks @petrelharp . The one issue here is that @jeromekelleher mentioned he'd prefer to split this into smaller incremental PRs (e.g. (1) one for nesting nodes, and a further one (2) for nesting mutations as well as nodes). I'm happy to do this, but not so happy to document the first behaviour, when it will then be obsoleted by the second. I suspect the way you would use css to target a node in (1) vs (2) would be quite different, so documenting (1) would end up being a waste of time if we are going to go with (2) in the long term. Not sure what I should do, other that perhaps remove the documentation for the time being, until we decide on a scheme?

petrelharp:
Right, I agree with what you're saying, @hyanwong.  I think we *should* be able to, in the future, do things like add a `font_size` argument to `draw_svg( )`, so we need a way to deal with this.

> The problem here is that we then need to somehow return the ID we autogenerated, so that the user can use it. Which seems a bit messy to me. Why not just let them define it themselves?

My proposal lets them define it themselves if they want to use it, but does not require that they do so. Something like:
```
def draw_svg(self, ..., root_id=None):
   # ...
   if root_id is None:
       root_id = np.random.randint(10000000)
```
Does this make sense?

petrelharp:
> @jeromekelleher mentioned he'd prefer to split this into smaller incremental PRs 

This is good in general, but if splitting it would make much extra work, then it's not the right thing to do...

jeromekelleher:
I'm confused about why targeting mutations make targeting nodes obsolete - surely we'd want to do both? We might want to style subtrees when there are no mutations present. My suggestion is that we focus on nodes in the this PR and then follow up with another PR for mutations. There's too much going on at once otherwise.

benjeffery:
> I think we _should_ be able to, in the future, do things like add a `font_size` argument to `draw_svg( )`, so we need a way to deal with this.

The way to deal with this is to have a uid _class_ we apply and use if we need to do stuff like this. Leave the ID to be selected and used by the user.

hyanwong:
> I'm confused about why targeting mutations make targeting nodes obsolete - surely we'd want to do both? 

Yes, definitely both. But the css syntax (and hence documentation) needed to style a node or a mutation will be quite different in the first case ("mutations in a small group associated with a node") vs the second ("mutations define a large encapsulating group"). This will also change the way we would identify a particular node via css. I just don't want to document one way of styling nodes and mutations, only to have to change it a day later to another way.

> We might want to style subtrees when there are no mutations present. My suggestion is that we focus on nodes in the this PR and then follow up with another PR for mutations. 

Yes, happy to do this, but as I said, I don't want to add documentation to this PR which will then be entirely obsoleted by the mutations change. I could simply remove all the styling documentation in this PR, as it will be no longer accurate anyway.

jeromekelleher:
> Yes, happy to do this, but as I said, I don't want to add documentation to this PR which will then be entirely obsoleted by the mutations change. I could simply remove all the styling documentation in this PR, as it will be no longer accurate anyway.

Let's not mention mutations at all, here. Does that solve the problem?

hyanwong:
> Let's not mention mutations at all, here. Does that solve the problem?

Yes, I can remove the description of mutations and the examples from the docs, and from the test suite.

hyanwong:
Done. I will move the mutation grouping stuff to another PR.

hyanwong:
@benjeffery if you are in SVG mode from checking #555 you might want to check this too. It's ready for review, I reckon, although an intermediate step on the way to the full mutation grouping idea, which I haven't submitted yet (just working through edge cases at the root)

benjeffery:
As background got merged this is now rebased to master so only the new changes are visible. I had to force push, so you'll need to `git reset`.

hyanwong:
By the way, when importing these SVGs into Inkscape (and presumably other SVG editors) the nested groupings become a pain. Fortunately Inkscape has a "deep ungroup" function which does recursive ungrouping, but it's in the "extra-plugins" menu, not next to "group" and "ungroup". Something to mention in a graphics vignette perhaps, but to specific for the general docs

python/tests/data/svg/tree.svg
```diff
@@ -2,69 +2,48 @@
...
-      </g>
-    </g>
-    <g class="labels" dominant-baseline="middle" font-size="14">
-      <g class="nodes">
-        <g text-anchor="start">
-          <g transform="translate(141.0, 163.29406477810417)">
-            <text class="n5">5</text>
-          </g>
+    <g class="node n9 root" transform="translate(100.0 30.0)">
+      <g class="m0 node n4 p9 s0" transform="translate(-36.0 139.6231433344194)">
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/570/commits/d8842dfad4be2655b28f3d5b08fa56b6af31ee45

benjeffery:
@hyanwong, @jeromekelleher  I'm now happy with this PR, we have a follow up hot on its heels to add some additional testing and the mutation-hierarchy-level feature.

jeromekelleher:
Looks great! Can we get a few simple unit tests for ``rnd``, just to document the behaviour and make sure it doesn't diverge by accident in the future?

Happy to merge after that.

jeromekelleher:
Looks like mergify isn't merging because codecov hasn't reported on the PR. Probably simplest to just merge manually. This is quite a bit behind the tip of master, so needs a rebase.

hyanwong:
I'll rebase now

benjeffery:
Codecov status last night stated they had degraded service.

jeromekelleher:
Merged! Thanks @hyanwong and @benjeffery, this is a huge step forward for viz!

PR #578, opened 2020-05-06
Improve KC incremental algorithm by using top-down approach

daniel-goldstein:
The incremental algorithm for KC involves doing a traversal down the subtree under each incoming edge. Doing so results in redundant updates as the subtrees for edges closer to the root fully cover pairs of samples in lower down subtrees. This takes an especially high toll when inserting edges into the first tree as this performs an `O(n^2)` pass per edge, resulting in an `O(n^3)` algorithm on the first tree.

To fix this, we process edges in reverse order (processing those closer to the root first). When traversing the subtree under an edge, we mark nodes as dirty if the edge is being removed, and clean if the edge is being inserted. If the child of an incoming edge is already clean, then all pairs associated with that edge have already been updated from a previous edge and we do not need to process it again. This algorithm ensures that edges being processed have sample lists that are pairwise disjoint, and therefore the total cost of processing the pairs of samples under all the edges is `O(n^2)`.

@jeromekelleher @hyanwong

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -917,19 +917,22 @@ def ts_kc_distance_incremental(ts1, ts2, lambda_=0):
     t1_vecs = KCVectors(ts1.num_samples)
     t2_vecs = KCVectors(ts2.num_samples)
 
+    t1_depths = np.zeros(ts1.num_nodes)
+    t2_depths = np.zeros(ts2.num_nodes)
+
     edge_diffs_iter_1 = ts1.edge_diffs()
     tree_iter_1 = ts1.trees(sample_lists=True)
     t1, t1_diffs = next(tree_iter_1), next(edge_diffs_iter_1)
-    update_kc_incremental(t1, t1_vecs, t1_diffs, sample_maps[0])
+    update_kc_incremental(t1, t1_vecs, t1_diffs, sample_maps[0], t1_depths)
     for t2, t2_diffs in zip(ts2.trees(sample_lists=True), ts2.edge_diffs()):
-        update_kc_incremental(t2, t2_vecs, t2_diffs, sample_maps[1])
+        update_kc_incremental(t2, t2_vecs, t2_diffs, sample_maps[1], t2_depths)
         while t1_diffs[0][1] < t2_diffs[0][1]:
             span = t1_diffs[0][1] - left
             total += norm_kc_vectors(t1_vecs, t2_vecs, lambda_) * span
 
             left = t1_diffs[0][1]
             t1, t1_diffs = next(tree_iter_1), next(edge_diffs_iter_1)
-            update_kc_incremental(t1, t1_vecs, t1_diffs, sample_maps[0])
+            update_kc_incremental(t1, t1_vecs, t1_diffs, sample_maps[0], t1_depths)
         span = t2_diffs[0][1] - left
         left = t2_diffs[0][1]
         total += norm_kc_vectors(t1_vecs, t2_vecs, lambda_) * span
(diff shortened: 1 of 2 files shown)


daniel-goldstein:
Some benchmarks that compare the naive implementation of KC on sequences (calling `Tree.kc_distance` on each pair of trees), the current incremental algorithm on master, and the optimized incremental algorithm on this branch. All benchmarks were run with a 1mb msprime simulation with the only other parameters being sample size and recombination rate.

Under instances of relatively few trees, the start up cost of doing an `O(n^2)` pass for every edge on the first tree dominates the current incremental implementation.

Screen Shot 2020-05-05 at 6 02 04 PM

The optimization implemented here ultimately runs the same traversal as `Tree.kc_distance` for the first tree, so it makes sense that we see them perform similarly.

As the recombination rate rises and the number of trees becomes the dominating factor, both incremental implementations perform better than the naive version, but the optimized incremental algorithm is still consistently faster.

Screen Shot 2020-05-05 at 6 02 23 PM

Screen Shot 2020-05-05 at 6 02 40 PM
Note the sample sizes tested here are different from the first two trials due to the naive version running very slowly.

hyanwong:
This is *great*, a very sensible approach and really good perf plots. Thanks @daniel-goldstein . It looks like we should simply merge this as-is.

python/tests/test_topology.py
```diff
@@ -917,44 +917,50 @@ def ts_kc_distance_incremental(ts1, ts2, lambda_=0):
     t1_vecs = KCVectors(ts1.num_samples)
     t2_vecs = KCVectors(ts2.num_samples)
 
+    t1_clean_nodes = np.zeros(ts1.num_nodes, dtype=bool)
```
jeromekelleher, line 920: ``clean`` isn't quite the right word I think. What we're marking is unaffected subtrees, isn't it? But the usage here is the opposite, where we're traversing into the clean subtrees.

It would be good to think the terminology through again, and add a few comments there to make it clear what's happening. (I think this is going to be a recurring pattern in top-down incremental algs, so I want to make sure we get it right first time.)

daniel-goldstein:
Ok I now think this approach is wrong, and you're right @jeromekelleher there is a pattern here that is more general for top-down incremental algorithms. Every top-down algorithm is going to have some per-node state that trickles down from parents to children. When an edge is removed, a subtree is "blocked off" in a way. And when an edge is inserted, the subtree is unblocked and information can trickle down into it. I think a general top-down algorithm would look something like this:
1. For each outgoing edge, set the child's state to null (blocked).
2. For each incoming edge `(child, parent)` in top-down order
    1. If `parent` is the root, set the state of `parent` to whatever the initial or root state is
    2. Update `child`'s state based off `parent`'s state
        1. Recur on the children of `child` _whose states are not null_.

In KC our per-node state is the depth of that node. If the state of a leaf were only its depth we'd be done, but that's not the case. In KC, the state of a leaf is `O(n)` and is its relationships with every other leaf. So once we reach a leaf, we need only perform a single traversal back up to the root, where along the way we take the pairs of that leaf and every other leaf (the same thing we are currently doing in `update_in_out_subtree_pairs`). This way we:
- don't need to throw away any edges
- only ever perform the upwards traversal for leaves that are part of an SPR
- never perform the upwards traversal for the same leaf more than once, keeping the algorithm `O(n^2)`

What do you think?

jeromekelleher:
Sounds good @daniel-goldstein. The way I'm thinking about it that we have *detached subtrees* (or, more concisely, roots). So, then a child that was a root is attached to a parent it (a) is no longer a root; and (b) we take some action based on being attached into another subtree.

Does this make sense do you think?

daniel-goldstein:
Ah yes, detached makes a lot of sense

python/tests/test_topology.py
```diff
@@ -917,89 +917,96 @@ def ts_kc_distance_incremental(ts1, ts2, lambda_=0):
...
-
-def update_in_out_subtree_pairs(tree, kc, edge, sample_index_map):
+        v = stack.pop()
+        if tree.is_leaf(v):
+            update_kc_pairs_with_leaf(tree, kc, v, sample_index_map, depths)
+        else:
+            c = tree.left_child(v)
+            while c != -1:
+                # Terminate iteration at nodes that are currently considered
+                # roots by the edge diffs
```
daniel-goldstein, line 986: This is a little not obvious but is because we're iterating over the completed tree as opposed to the edge diffs as their being inserted so this is a way to see where we are in the insertion, assuming this is done top-down. Any better way of doing that?

daniel-goldstein:
Updated algorithm to use what could be a more standard "top-down" approach. The per-node state is depth from the root. When an edge is removed, the root of the newly detached tree is set to have depth 0. The depth of each node is propagated down subtrees that have been attached or detached. When a leaf is reached in the downward traversal, a single upward traversal is performed to update the KC vector pairs of that leaf with every other leaf in the tree. The performance difference compared to the previous optimized algorithm at the start of this PR is marginal.

python/tests/test_topology.py
```diff
@@ -917,89 +917,96 @@ def ts_kc_distance_incremental(ts1, ts2, lambda_=0):
...
-            u = e.child
+    # Update state of detached subtrees.
+    for e in reversed(edges_out):
+        u = e.child
+        depths[u] = 0
+
+        # Only update detached subtrees that remain detached. Otherwise,
+        # they must be reattached by an incoming edge and will be
+        # updated below.
+        if tree.parent(u) == tskit.NULL:
```
daniel-goldstein, line 954: This is a bit of a shortcut instead of always updating the detached subtree. I'm kind of uneasy about it because it leaves the depths on the subtree in a bit of a bad state until it is ultimately reinserted, but we do take a bit of a performance hit otherwise for what feels like avoidable work.

jeromekelleher:
Looks good, thanks @daniel-goldstein. I'll need you to talk me through the changes in a bit more detail, so I'm fully with you.

jeromekelleher:
Sorry, you just lost the CI race with #584 @daniel-goldstein. Can you rebase again please? Ready to merge then.

PR #581, opened 2020-05-06
Use same root branch for ts

hyanwong:
When plotting a single tree, I do not think we should show a root branch in a tree unless there is a mutation above the root in that tree. At the moment we show a root branch if there is a mutation above the branch in any tree in the ts (which also means we have to iterate through the entire ts when plotting a single tree)

For a whole ts, however, we should plot the root branch for all trees if any tree has a mutation above the root. This minor change does that too.


hyanwong:
Closed for the moment, as tests will rely on #570

PR #582, opened 2020-05-06
compare indiv table rows and test

petrelharp:
@mufernando noticed that we can't compare IndividualTableRows because of the numpy arrays. And, it turned out this wasn't being tested anywhere. There is [discussion about the right way to do this](https://github.com/python-attrs/attrs/issues/435) in attrs, but currently we just have to implement ``__eq__``. The precise way to do this will [change soon](https://www.attrs.org/en/stable/changelog.html), but no big deal.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -39,11 +39,27 @@ import tskit.util as util
 attr_options = {"slots": True, "frozen": True, "auto_attribs": True}
 
 
-@attr.s(**attr_options)
+# note: soon attrs will deprecate cmp; we just need to change this argument to eq
+@attr.s(cmp=False, **attr_options)
 class IndividualTableRow:
     flags: int
     location: np.ndarray
     metadata: bytes
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return False
+        else:
+            return all(
+                (
+                    self.flags == other.flags,
+                    np.array_equal(self.location, other.location),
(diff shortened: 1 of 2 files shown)


python/tskit/tables.py
```diff
@@ -39,12 +39,27 @@
 attr_options = {"slots": True, "frozen": True, "auto_attribs": True}
 
 
-@attr.s(**attr_options)
+@attr.s(cmp=False, **attr_options)
```
jeromekelleher, line 42: Hmm, better to use ``eq=False`` then, since ``cmp`` is being deprecated?

petrelharp: You'd think so, but on my machine I have attrs 19.1.0, in which the `eq` parameter does not exist.

petrelharp:
Judging from that attrs github issue, I think this is fine.

python/tskit/tables.py
jeromekelleher: I see. So cmp will be deprecated and we'll have to deal with that at some point. Maybe add a quick comment to that effect?

petrelharp:
Whoops - I missed some test coverage; now I should have got it?

python/tskit/tables.py
petrelharp: done

petrelharp:
I'll merge when the checks finish unless someone says otherwise.

PR #584, opened 2020-05-06
Remove non-threadsafe stacks from tree struct.

### LINKED ISSUE #459: Remove non-threadsafe stacks from tsk_tree_t
tsk_tree_t current has two stack variables which are used in one or two functions (making them non-threadsafe). Remove these and malloc the stacks each time instead.

jeromekelleher:
Closes #459

@daniel-goldstein, can you have a quick check through this please?

### DIFF
--- a/c/tskit/convert.c
+++ b/c/tskit/convert.c
@@ -46,6 +46,7 @@ typedef struct {
     size_t precision;
     tsk_flags_t options;
     char *newick;
+    tsk_id_t *traversal_stack;
     tsk_tree_t *tree;
 } tsk_newick_converter_t;
 
@@ -55,7 +56,7 @@ tsk_newick_converter_run(
 {
     int ret = TSK_ERR_GENERIC;
     tsk_tree_t *tree = self->tree;
-    tsk_id_t *stack = self->tree->stack1;
+    tsk_id_t *stack = self->traversal_stack;
     const double *time = self->tree->tree_sequence->tables->nodes.time;
     int stack_top = 0;
     int label;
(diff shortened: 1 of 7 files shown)


c/tskit/convert.c
```diff
@@ -46,6 +46,7 @@ typedef struct {
     size_t precision;
     tsk_flags_t options;
     char *newick;
+    tsk_id_t *traversal_stack;
```
daniel-goldstein, line 49: Is there any reason to be concerned about thread safety for the newick converter?

jeromekelleher: No, but we want to get rid of the stacks that are in the ``tsk_tree_t`` structure, as that was a Bad Idea in the first place.

daniel-goldstein: Looks good then!

jeromekelleher:
Aha, testing on Windows to the rescue! Turns out I had made the stacks 1 too small in the MRCA case  (which was correct before, but I had dismissed the + 1 as pointless), and this problem only appeared in the Python testing on Windows.

So, it's worth testing on Windows even if you don't really expect many people to use the software on Windows, just for a bit of platform diversity.

daniel-goldstein:
Interesting! I also figured that the + 1 was pointless. Why is it necessary?

daniel-goldstein:
Oh I see, the root node pushes its parent, null, onto the stack

jeromekelleher:
> Oh I see, the root node pushes its parent, null, onto the stack. Would be good to have a test for the mrca of two isolated samples, that would expose the bug.

Excellent call, done.

PR #587, opened 2020-05-06
Grouped svg mutations

hyanwong:
This follows on from #570 and picks up the idea of creating an SVG group for each mutation too (see https://github.com/tskit-dev/tskit/pull/570#issuecomment-624011266 for justification).

It seems like this is a good idea to me. The main thing to sort out is how we assign classes to the groups. At the moment I assign the parent node id (or "root") plus the child node id to all groups, including mutation groups. Mutation groups then have an additional 'mut" class and the mutation number and site number, whereas node groups have a "node" class. An example would be something like this (assuming an example with mutation 0 over a root node id 8 with 2 daughters, id 4 and 5)

```xml
<g class="tree t0">
  <g class="root n8 mut m0 s0">  <!-- a mutation group identifying mutation 0 over node 8 -->  
    <g class="root n8 node">   <!-- a node group identifying node 8-->
      <g class="p8 n4 mut m1 s1">  <!-- a mutation group identifying mutation 1 over child node 4 -->
      ... further mutation or node groups, including the node group for node 4 and descendants
      </g>
      <g class="p8 n5 mut m2 s2">  <!-- a mutation group identifying mutation 2 over child node 5 -->
      ... further mutation or node groups, including the node group for node 5 and descendants
      </g>
      <path class="edge" ...> <!-- the line joining the root node to mutation 0 -->
      <circle ...> <!-- the root node symbol, doesn't strictly need a class, as `edge + *` will do -->
      <text ...> <!-- the root node label, always a text node -->
    </g>
    <path class="edge" ...> <!-- the line above mutation 0 -->
    <rect ...> <!-- the mutation symbol for mutation 0 -->
    <text ...> <!-- the label for mutation 0, always a text node -->
  </g> <!-- close mutation group -->
</g> <!-- close tree group -->
```

This is all quite complicated to describe, but logical. I think it allows us to target anything we might want, but it could be easier to explain if, for instance, the circles and rects are given a class "symbol", rather than targetted because they always follow an element of class "edge".

Probably easiest if I discuss this with someone (e.g. @benjeffery ) on a Jitsi call?

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -473,12 +473,14 @@ class SvgTree:
             self.node_attrs[u] = {}
             if node_attrs is not None and u in node_attrs:
                 self.node_attrs[u].update(node_attrs[u])
+            self.add_class(self.node_attrs[u], "sym")  # class 'sym' for symbol
             label = ""
             if node_labels is None:
                 label = str(u)
             elif u in node_labels:
                 label = str(node_labels[u])
             self.node_label_attrs[u] = {"text": label}
+            self.add_class(self.node_label_attrs[u], "lab")  # class 'lab' for label
             if node_label_attrs is not None and u in node_label_attrs:
                 self.node_label_attrs[u].update(node_label_attrs[u])
 
@@ -494,6 +496,7 @@ class SvgTree:
                 }
                 if mutation_attrs is not None and m in mutation_attrs:
                     self.mutation_attrs[m].update(mutation_attrs[m])
+                self.add_class(self.mutation_attrs[m], "sym")  # class 'sym' for symbol
                 label = ""
                 if mutation_labels is None:
                     label = str(m)
@@ -502,6 +505,7 @@ class SvgTree:
                 self.mutation_label_attrs[m] = {"text": label}
                 if mutation_label_attrs is not None and m in mutation_label_attrs:
                     self.mutation_label_attrs[m].update(mutation_label_attrs[m])
+                self.add_class(self.mutation_label_attrs[m], "lab")
 
         self.draw()
 
@@ -622,7 +626,12 @@ class SvgTree:
             edge_height = dy / (len(self.node_mutations[focal]) + 1)
             offset_y = edge_height
 
-        # Add mutation symbols + labels
+        # Add mut group for each mutation, and give it these classes for css targetting:
+        # "mut"
+        # "nA":           where A == focal node id
+        # "pB" or "root": where B == parent id (or "root" if the focal node is a root)
+        # "mX":           where X == mutation id
+        # "sY":           where Y == site id
         for m in reversed(self.node_mutations[focal]):
             mutation_classes = ["mut", f"m{m.id}", f"s{m.site}"]
             grp = grp.add(
@@ -636,7 +645,7 @@ class SvgTree:
                     g=grp, edge_dxy=(edge_x, edge_height), node=focal, mutation=m.id
                 )
             )
-            # after the first sideways move, all further movements go downwards
+            # after the first sideways line of an edge all further movements go downwards
             offset_x = edge_x = 0
             offset_y = edge_height
 
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -260,7 +260,6 @@ class SvgTreeSequence:
         dwg = svgwrite.Drawing(size=self.image_size, debug=True, **root_svg_attributes)
         self.drawing = dwg
         dwg.defs.add(dwg.style(SvgTree.standard_style))
-        self.node_labels = {u: str(u) for u in range(ts.num_nodes)}
         if node_labels is None:
             node_labels = {u: str(u) for u in range(ts.num_nodes)}
         if style is not None:
@@ -273,7 +272,6 @@ class SvgTreeSequence:
         else:
             axis_top_padding = 5
             tick_len = (5, 5)
-        self.node_labels = {u: str(u) for u in range(ts.num_nodes)}
         if root_branch is None:
             root_branch = any(
                 any(tree.parent(mut.node) == NULL for mut in tree.mutations())
@@ -619,14 +617,11 @@ class SvgTree:
         offset_y = dy
         if v == NULL:
             classes.append(f"root")
-            if self.root_branch:
-                # FIXME this is pretty crappy for spacing mutations over a root.
-                root_branch_len = self.root_branch_length
-            else:
-                root_branch_len = 0
             edge_x = 0
-            edge_height = root_branch_len / (len(self.node_mutations[focal]) + 1)
-            offset_y = dy - root_branch_len + edge_height
+            edge_height = self.root_branch_length / (
+                len(self.node_mutations[focal]) + 1
+            )
+            offset_y = dy - self.root_branch_length + edge_height
         else:
             classes.append(f"a{v}")
             edge_x = offset_x
@@ -653,7 +648,8 @@ class SvgTree:
                 )
             )
             # after the first sideways line of an edge all further movements go downwards
-            offset_x = edge_x = 0
+            offset_x = 0
+            edge_x = 0
             offset_y = edge_height
 
         # Add a new group for each node, and give it these classes for css targetting:
@@ -663,13 +659,12 @@ class SvgTree:
         # "sample":       a class present if the focal node is a sample
         # "leaf":         a class present if the focal node is a leaf
         classes.append("node")
-        ind = self.tree.tree_sequence.node(focal).individual
-        pop = self.tree.tree_sequence.node(focal).population
-        if ind != NULL:
-            classes.append(f"i{ind}")
-        if pop != NULL:
-            classes.append(f"p{pop}")
-        if self.tree.is_sample(focal):
+        focal_node = self.tree.tree_sequence.node(focal)
+        if focal_node.individual != NULL:
+            classes.append(f"i{focal_node.individual}")
+        if focal_node.population != NULL:
+            classes.append(f"p{focal_node.population}")
+        if focal_node.is_sample():
             classes.append("sample")
         if self.tree.is_leaf(focal):
             classes.append("leaf")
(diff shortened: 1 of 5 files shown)


benjeffery:
We just had a call where we thought the symbols and labels should have a class as the HTML entity may change and shouldn't be used as a target.

hyanwong:
I wonder if we should also add the population id and individual id to the node group (perhaps only if these are not `tskit.NULL`)? The obvious notation would be

```
<g class="pPOPNUM iIND_NUM ... p8 n4 mut m1 s1">
```

but `pID` for population would conflict with `pID` used to indicate parent ID. My inclination is to change the notation for parent nodes. Perhaps `aID` for "ancestorID" or `mID` for "motherID" (and `m` is the letter before `n`, so we would have `mA nB` for parent A + daughter B.

petrelharp:
> I wonder if we should also add the population id and individual id to the node group (perhaps only if these are not tskit.NULL)?

That'd make implementing the `.draw()` method even easier! Maybe `pop<POP ID>` and `ind<IND ID>`?

jeromekelleher:
The more semantic info we're including the better, I guess. Seems like there's no harm in it? We can always provide methods to trim out the info later if it turns out to add too much weight.

hyanwong:
I've rebased, implemented populations and individuals for nodes ("p<POP_ID>" and "i<IND_ID>", with parent nodes being "a<PARENT_NODE_ID" - other alternative letters welcome, although given the number of times these are repeated, I've stuck to single letters). So this is also ready for review now - the final piece of my SVG jigsaw! 

For discussion is Peter's suggestion about whether to by default create a UID for the SVG, to stop user-generated styles spilling into different SVGs in the same document (which I agree is very unexpected, and can play havoc with custom SVGs in Jupyter notebooks). However, I can't think of an elegant way to do this automatically: the easiest way is to require the user to specify an ID in the style specification, which would mean they would have to extract any autogenerated ID anyway, unless we defined some sort of placeholder. Alternatively, we could require the style specifications to be an array of strings, and we automatically prefix each string with "#UID " when writing to the SVG file, but messing with the user-supplied styles like this seems a little wrong.

Finally, I was also thinking that it would be useful to embed the number of mutations, sites, nodes, and edges per tree as metadata into the SVG somehow, as these are useful for animation. However, the most obvious way to add them as [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/data-*) is SVG2 only, and not supported by svgwrite. We could possible define our own namespace for this (see the top answer at https://stackoverflow.com/questions/15532371/do-svg-docs-support-custom-data-attributes) but that seems quite heavyweight. Perhaps @benjeffery has a better solution?

When all seems OK, I'll write some docs and amend the commit with those.

benjeffery:
For the UID issue - I don't think we want to be too clever here with re-writing styles etc. I think a note in the docs to add a UID to both your styles and top-level attributes is fine. 

All browsers support data attributes: https://caniuse.com/#feat=mdn-svg_attributes_data and `svgwrite` can be persuaded to support any attribute you like:

```python
import svgwrite
svgwrite.data.full11.elements['g'].valid_attributes = frozenset(
    list(svgwrite.data.full11.elements['g'].valid_attributes) + 
    ['data-whatever']
)
svgwrite.data.full11.attributes['data-whatever'] = svgwrite.data.types.SVGAttribute(
    'data-whatever', anim=False, types=frozenset(['number']), const=[]
)
d = svgwrite.Drawing()
d.add(d.g(data_whatever=5))

You need the first two lines just after the import.

File: python/tskit/drawing.py [hunk 1]
```diff
@@ -253,10 +254,15 @@ def __init__(
             root_svg_attributes = {}
         if max_tree_height is None:
             max_tree_height = "ts"
+        if root_svg_attributes is None:
+            root_svg_attributes = {}
         self.image_size = size
         dwg = svgwrite.Drawing(size=self.image_size, debug=True, **root_svg_attributes)
         self.drawing = dwg
         dwg.defs.add(dwg.style(SvgTree.standard_style))
+        self.node_labels = {u: str(u) for u in range(ts.num_nodes)}
```
jeromekelleher, line 263: self.node_labels is being set twice, I think.

python/tskit/drawing.py
```diff
@@ -577,38 +606,89 @@ def assign_x_coordinates(self, tree, x_start, width):
...
+        dwg = self.drawing
+        grp = curr_svg_group
+        v = self.tree.parent(focal)
+        classes = [f"n{focal}"]
+        offset_x = dx
+        offset_y = dy
         if v == NULL:
-            classes.add(f"root")
+            classes.append(f"root")
+            if self.root_branch:
```
jeromekelleher, line 622: I don't think this if statement does anything - you've computed the ``self.root_branch_length`` above so there's no need for the local variable.

python/tskit/drawing.py
```diff
@@ -577,38 +606,89 @@ def assign_x_coordinates(self, tree, x_start, width):
...
+                    transform=f"translate({rnd(offset_x)} {rnd(offset_y)})",
+                )
+            )
+            ret.append(
+                SvgGroupInfo(
+                    g=grp, edge_dxy=(edge_x, edge_height), node=focal, mutation=m.id
+                )
+            )
+            # after the first sideways line of an edge all further movements go downwards
+            offset_x = edge_x = 0
```
jeromekelleher, line 656: Avoid these multiple assignments.

python/tskit/drawing.py
```diff
@@ -577,38 +606,89 @@ def assign_x_coordinates(self, tree, x_start, width):
...
+            offset_y = edge_height
+
+        # Add a new group for each node, and give it these classes for css targetting:
+        # "node"
+        # "nA":           where A == focal node id
+        # "aB" or "root": where B == parent id (or "root" if the focal node is a root)
+        # "sample":       a class present if the focal node is a sample
+        # "leaf":         a class present if the focal node is a leaf
+        classes.append("node")
+        ind = self.tree.tree_sequence.node(focal).individual
```
jeromekelleher, line 666: Not much point in looking up the node twice 
```
focal_node = self.tree.tree_sequence.node(focal)
if focal_node.individual != NULL:
   # etc
```

python/tskit/drawing.py
```diff
@@ -617,73 +697,59 @@ def draw(self):
...
+                    stack.extend(self.add_node(curr.g, c, dx, dy))
 
             # Add edge first => below
-            if v != NULL:
-                self.add_class(self.edge_attrs[u], "edge")
-                pv = node_x_coord_map[v], node_y_coord_map[v]
-                dx = pv[0] - pu[0]
-                dy = pv[1] - pu[1]
+            dx, dy = curr.edge_dxy
+            if dx == dy == 0:
```
jeromekelleher, line 717: Ouch, I didn't even realise that python let you do this. ``if dx == 0 and dy == 0``

jeromekelleher: I doubt this does what you expect, either, so we're probably missing some test coverage. This should mean (dx == dy) == 0 -> (True | False) == 0 .

hyanwong: I'm pretty sure a==b==c does the expected thing due to [comparison operator chaining](https://docs.python.org/3/reference/expressions.html#comparisons); "unlike C, expressions like a < b < c have the interpretation that is conventional in mathematics"

So 0==0==1 is False, and all works as expected. Personally I find this syntax clearer, but then my head is rarely in C mode.

benjeffery: I prefer `if dx == 0 and dy == 0` as it is unambiguous for both "C thinkers" and "mathematics thinkers".

hyanwong: Sure, happy to switch: just pointing out that it's not unreasonable.

hyanwong: (by the way, does that mean you also dislike `a<b<c`, which I fund much clearer than breaking up the test into 2 halves, like you have to do in C)?

jeromekelleher: You're right, looks like ``0 == 0 == 0`` is True rather than False. But, it's not obvious, so I would be against using it. Stuff like ``a < b < c`` is great, on the other hand, as the meaning is unambiguous (to my eyes anyway).

hyanwong:
I'm not paying much attention to this for the next week or two, but just to note that the main point of this PR, which assigns part of the line that represents an edge to a `mut` SVG group, means that when you try to animate an SPR between two trees, it's hard to find corresponding edges to transform, because the single line representing an edge may be broken down into pieces.

However, I *think* it might be possible to have the best of both worlds here, by drawing the line from a true node (not a mutation) upwards through the mutations and back to the previous node. I've coded this up in f365d51 but it will need a little thought/discussion, hopefully in a few weeks time.

jeromekelleher:
Hi @hyanwong, we're hoping to get 0.3.0 out in a couple of weeks and we should get this finished first, I guess? How much is left to do here?

hyanwong:
So I stopped developing this @jeromekelleher because I realised that it might make our lives more difficult if we want to map one tree to an adjacent tree in the tree sequence. With the "mutation makes an SVG group" idea, we no longer maintain some identical groups when moving to a new tree, so e.g. animating SPRs becomes more problematic.

I think we need some discussion on the right approach here. @benjeffery normally has useful things to say.

jeromekelleher:
I think it's probably OK to leave this go until the next release, unless you think there are backwards incompatible changes that will need to be made later to accommodate mutations? Do you foresee the structure of the SVG changing?

hyanwong:
Well, *if* we want to create a new subtree group under each mutation (i.e. as implemented in this PR) then yes, the SVG structure will be quite different. Sorry

jeromekelleher:
> Well, if we want to create a new subtree group under each mutation (i.e. as implemented in this PR) then yes, the SVG structure will be quite different. Sorry.

Hmmm. Well, I'm going to make this your call then. We do want to ship in the next few weeks, so it would be good to have a think about it. 

We could version the SVG output, so at least we could tell users about this, and this would let us make changes without having to worry *too* much about compatability?

jeromekelleher:
I've tagged this for 0.3.1, as it's pretty clear we're not going to get it sorted for 0.3.0. I think we'll just have to mark this SVG structure as "in development" and warn people to expect breakage if they depend on the structure.

hyanwong:
Yes - I think we might want not to implement SVG groups for mutations anyway. I'm debating whether to close this issue and try a new approach. We should probably have a more general discussion some time.

hyanwong:
Closing for the time being, as we are unsure if grouped mutations will be quite as flexible as we had hoped

PR #590, opened 2020-05-07
AFS docs

petrelharp:
I got confused about the output of some AFS, and worked out what's going on.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -106,6 +106,67 @@ using these methods directly, so they should be preferred.
 - :meth:`TreeSequence.sample_count_stat`
 
 
+.. _sec_stats_afs:
+
++++++++++++++++++++++++++
+Allele frequency spectrum
++++++++++++++++++++++++++
+
+We compute allele frequency spectra (AFS), including windowed and joint spectra,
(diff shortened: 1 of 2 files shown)


petrelharp:
To be clear - I'm asking for input mostly on whether what I've written is clear, not whether what I describe is the right thing to compute for the AFS - we've discussed that extensively elsewhere (although it might not be), but in any case is a question for a different thread.

molpopgen:
I think that this reads very well.  I'd say that this section would need a careful reading by anyone, but that's the nature of the problem.  Pictures may help, but I am totally fine with what you have written here.

petrelharp:
> and I think that the treatment of the data is correct.

*phew!* =)

petrelharp:
> I made my comments in the main thread for the PR.

I don't see any comments?

molpopgen:
> I don't see any comments?

hmmm--reload?  I just said that I liked it ;)

molpopgen:
This is the comment: https://github.com/tskit-dev/tskit/pull/590#issuecomment-625944221

petrelharp:
Oh, I saw that one. =)

PR #591, opened 2020-05-07
Update test_minimal_cpp.cpp

molpopgen:
use the C++ header for assert.


molpopgen:
this is a pedantic change that I made when looking something up.  C++ programs should include C headers `<foo.h>` as `<cfoo>`.

jeromekelleher:
Thanks @molpopgen, very helpful!

PR #595, opened 2020-05-08
Add a trivial python example to add metadata to a ts

hyanwong:
Also practise (US) to practice (UK, noun)

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -24,7 +24,7 @@ the keys and types of those properties are specified along with optional
 long-form names, descriptions
 and validations such as min/max or regex matching for strings. See
 :ref:`sec_metadata_example` below. Names and descriptions can assist
-downstream users in understanding and using the metadata. It is best practise to
+downstream users in understanding and using the metadata. It is best practice to
 populate these fields if your files will be used by any third-party, or if you wish to
 remember what they were some time after making the file!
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I don't think this is terribly helpful, unfortunately - it's much simpler to use examples where we're building up a table collection from scratch, rather than bulk-editing an existing one. This will just confuse people, I think.

hyanwong:
> I don't think this is terribly helpful, unfortunately

Yeah, I sort-of agree, which is why I asked in Slack about alternatives.

> it's much simpler to use examples where we're building up a table collection from scratch

Yes, but the *idea* of creating a table is quite a step for many people, I think. Also, I think that most people interested in metadata will want to add metadata to an existing TS they have created already (or which they have a pipeline for creating). Perhaps we could show an example of adding individuals to a vanilla msprime TS instead, and associating those individuals with samples?

I certainly found it pretty hard to understand how the metadata schemas worked without a python example to copy and paste. Just the plain schema details are not enough, I feel. Unfortunately this does mean having to get to grips with the idea of tables somehow, which is already a bit of a jump. Not sure I have any bright ideas for how to make this easy, I'm afraid @jeromekelleher .

benjeffery:
This isn't a metadata specific concern - maybe a doc/tutorial section on "modifying existing tree sequences?"

jeromekelleher:
I think there is one of those already for updating tree sequences.

I think we should wait until the dust settles a bit on the metadata stuff before trying to write tutorials like this. It's a very limited number of people we need to communicate the details of "how do you make metadata schemas work" vs the vast majority of people who **use** them and have no idea about the schemas. We should be focusing on the users, not developers right now.

hyanwong:
I don't mind too much if it is about updating existing or creating new ones. I just want some python code to use to help me understand how metadata works. I'm not getting this from the documentation in metadata.rst, so I would hope to have something there to tell me firstly how I use a schema (i.e. python commands), rather than (just) an example of a schema format. Otherwise I'm completely lost before I even get to the schema description part.

petrelharp:
Hm: looking at the docs on metadata again, I think a re-ordering would help: start off the [Working with Metadata](https://tskit.readthedocs.io/en/latest/tutorial.html#working-with-metadata) tutorial with how to *access* metadata, and then have a subsequent subsection on *modifying* it (including modifying the schemas). Right now the first example actually modifies the table's schema, which is not what most users want to do. The stuff on modifying schemas should maybe just go in the main [metadata docs](https://tskit.readthedocs.io/en/latest/metadata.html).

I think this example is useful; there is a section on [bulk modification](https://tskit.readthedocs.io/en/latest/tutorial.html#metadata-for-bulk-table-methods) which would be a good place for this example (with some more words, I think).

benjeffery:
I've added an issue to track metadata doc suggestions: https://github.com/tskit-dev/tskit/issues/603

benjeffery:
@hyanwong I propose closing this PR as #603 will fix up the metadata docs?

PR #596, opened 2020-05-08
more flags to check integrity

### LINKED ISSUE #593: More fine-grained integrity check flags would be helpful
I'd like to make SLiM call `tsk_table_collection_check_integrity()` in each generation, to ensure that bugs like https://github.com/MesserLab/SLiM/issues/62 don't slip through the cracks again.  However, the set of flags presently supported by `tsk_table_collection_check_integrity()` don't provide fine-grained enough control.  For example:

- While simulation is running in SLiM, we have no population table set up.  We therefore always fail the check `population >= num_populations` and get a `TSK_ERR_POPULATION_OUT_OF_BOUNDS` error.  This check is always run (there is no flag controlling it), so there's no way for us to turn it off.  So at present we can't call `tsk_table_collection_check_integrity()` at all.

- Some of what `TSK_CHECK_MUTATION_ORDERING` controls would be useful to us.  However, mutation ordering itself is not maintained while SLiM is running; we do a sort immediately before simplifying, to fix that up.  So dividing `TSK_CHECK_MUTATION_ORDERING` into the parts we want, versus the sort-order check we don't want, would be useful.

There are probably other cases as well.  @petrelharp says he will have a go at this, since he understands both the tskit side and the SLiM side of this issue.  Once the necessary changes are made in tskit, they would be pulled into SLiM's copy of the code.  SLiM now has a stub method, `SLiMSim::CheckTreeSeqIntegrity()`, where a call to `tsk_table_collection_check_integrity()` is commented out; that stub is just waiting for this fix, and then can be enabled with the appropriate flags.

petrelharp:
Closes #593.

This changes the meaning of `TSK_CHECK_MUTATION_ORDERING`; so that what was `TSK_CHECK_MUTATION_ORDERING` before is now `TSK_CHECK_MUTATION_ORDERING | TSK_CHECK_MUTATION_PARENTS`, but that flag is not being used on its own in the tskit code, and I doubt elsewhere (@molpopgen?), so I think it's fine to change.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -2898,6 +2898,320 @@ test_column_overflow(void)
     tsk_table_collection_free(&tables);
 }
 
+static void
+test_table_collection_check_integrity(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1;
+
+    /* nodes */
+    ret = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, INFINITY, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_TIME_NONFINITE);
+
+    ret = tsk_node_table_clear(&tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, 0.0, 0, TSK_NULL, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, TSK_NO_CHECK_POPULATION_REFS);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_POPULATION_OUT_OF_BOUNDS);
+
+    ret = tsk_node_table_clear(&tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, 0.0, TSK_NULL, 0, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INDIVIDUAL_OUT_OF_BOUNDS);
+
+    ret = tsk_node_table_clear(&tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, 0.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 0);
+    ret = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, 1.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    /* edges */
+    ret = tsk_edge_table_add_row(&tables.edges, 0.0, 1.0, TSK_NULL, 0, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NULL_PARENT);
+
+    ret = tsk_edge_table_clear(&tables.edges);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_edge_table_add_row(&tables.edges, 0.0, 1.0, 2, 0, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
+
+    ret = tsk_edge_table_clear(&tables.edges);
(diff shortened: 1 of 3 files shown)


molpopgen:
I'm not using these flags anywhere, so changes here won't affect me.

jeromekelleher:
LGTM @petrelharp, splitting up CHECK_MUTATION_ORDERING like you've done makes sense.

I'm less convinced about the wisdom of adding ``TSK_CHECK_POPULATIONS``, since it's so easy to work around. Why not just add the populations you're referring to into the population table and take advantage of the integrity checking? This will have to be done at the end anyway, when the TS is output. 

I'm uneasy about it because it's making a special case  for checking population reference integrity, for no really good reason - why not have opt-ins for all the other reference checks as well? This will also change the default behaviour from "check all references" to "check all references, except populations", which seems weird, confusing and liable to lead to nasty bugs in the future.

petrelharp:
> I'm uneasy about it because it's making a special case for checking population reference integrity, for no really good reason.

Well, I wondered about this also. @bhaller, what do you think? It is a little onerous to keep the population table updated, since we can't just look at the extant populations, but have to also check if we need extra population table entries for populations that no longer exist (so, maybe there's a node that refers to population 3, but due to removing  populations, maybe there is no population 3 any more; so when writing tables SLiM just adds an "empty" population 3 to the table, and leaves it up to the user to deal with this).  I forget if there's an "add population table" function, but if there was, it could be called before the check, and that seems not a big deal?  

It is a bit odd to have references to populations that don't exist in the tables- although, I thought through this all again just now and I think it's the right way to go: we could arrange it so that deleting a population then adding another that has the same ID as the first resulted in distinct entries in the population table... but, in the current scheme, a node having population k means "population that was number k at the node's birth time", which is pretty easy to cross-reference to information written out from SLiM; if we did something fancier, this would be more confusing.

bhaller:
Hmm.  I'm not really a fan of shifting over to trying to keep an up-to-date population table at all times in SLiM; that's a pretty major change, with the potential for all sorts of glitches (like the user re-using the same subpop id and how we deal with that), for no benefit given that the current scheme works perfectly.  I'm also not a fan of building and then throwing away a population table in every generation just to do this integrity check – potentially slow for large models, and really just a waste of time since all of that work would be getting done just to satisfy the integrity checker.  It also seems a little problematic to be running an integrity check, not on the actual data structures we're using in SLiM, but on patched-up, temporary data structures put together for the purpose of making the integrity checker happy.

@jeromekelleher asks: "why not have opt-ins for all the other reference checks as well?"  I think that was meant rhetorically, but actually that sounds like the right approach to me.  Make all checks opt-in, and subdivide the existing checks in whatever way provides the control that client code needs.  If "check all references" is expected to be something most clients want, provide a predefined value for all of those flags ORed together, for convenience.  If there is a worry about breaking existing code that calls this function, then keep the existing function with its existing semantics, and define a new function with a different name that has the new all-opt-in semantics, and make the old function simply call the new function with the appropriate flags set, preserving backward compatibility.

molpopgen:
Another option is for slim to have its own integrity checker, perhaps using the tskit version as a starting point.  The current function is a "check all the things" implementation, and it seems reasonable for clients to write their own replacements if they want something different.  The checks aren't that hard to write, and they won't drift since slim includes the tskit code.

bhaller:
That's certainly another option, if it seems best.

petrelharp:
No need to fork this code, we can do it in tskit. I'm happy to stick in flags for the other things. That'll change the existing behavior even more, but I'm not sure we have to worry about it (and now's the time, anyhow)?

molpopgen:
My comment about forking was in response to the idea that disabling some checks, like population table integrity, may not be desired in tskit itself.  This is the usual trade off between a general purpose library and specialized client needs.

jeromekelleher:
I've thought this through and have a proposal that will hopefully keep everyone happy. How about we put in flags to **opt out** of reference integrity checking for specific tables? I'm definitely against the idea of making the user opt-in to reference integrity checks, mainly because this would make it much harder for me to be user that tskit itself is safe (this is the main point of these checks - guaranteeing safety at load time so that we don't have to check things later while doing processing). By default ``check_integrity`` must do the basic checks to guarantee that all references can be followed safely.

However, there's no harm in providing the option to opt *out* of these checks, where the users is definitely saying "I'm responsible for managing these references". So, we can add something like ``TSK_NO_CHECK_POPULATION_REFS`` which will tell us to not check population references. This is seems like the right approach to me anyway, since it should be exceptional things that you want to opt out of, rather than being forced to opt in to a long list of things (making it much more likely that something is missed).

We don't need to add these for all the difference reference types right now, since the pattern is fairly clear (but, no harm in just doing it either in the interest of just getting it done.)

bhaller:
@jeromekelleher sure, sounds great.  :->

jeromekelleher:
Bumping this @petrelharp - do you think we could finish this out in the next few days? We can just keep it to an opt-out for population reference integrity, and a test for this flag. I think that'll cover the SLiM requirement then, right?

petrelharp:
How's that? I added plenty of C tests.  Note that this no longer changes previous implications of flags.

c/tskit/tables.c
```diff
@@ -6438,6 +6438,8 @@ tsk_table_collection_check_edge_ordering(tsk_table_collection_t *self)
  * TSK_CHECK_SITE_ORDERING       Check that sites are in nondecreasing position order.
  * TSK_CHECK_SITE_DUPLICATES     Check for any duplicate site positions.
  * TSK_CHECK_MUTATION_ORDERING   Check mutation ordering contraints for a tree sequence.
+ * TSK_NO_CHECK_MUTATION_PARENTS Check contraints on mutation 'parent' column.
```
jeromekelleher, line 6441: Missing "Do not" in the description

c/tskit/tables.h
```diff
@@ -636,7 +636,9 @@ typedef struct {
 #define TSK_CHECK_SITE_ORDERING (1 << 2)
 #define TSK_CHECK_SITE_DUPLICATES (1 << 3)
 #define TSK_CHECK_MUTATION_ORDERING (1 << 4)
-#define TSK_CHECK_INDEXES (1 << 5)
```
jeromekelleher, line 639: Might be easier to follow if we kept the NO_CHECKs grouped together. Also no real reason to change the value of CHECK_INDEXES here - we should leave it as (1<<5).

c/tests/test_tables.c
```diff
@@ -2809,6 +2809,318 @@ test_column_overflow(void)
...
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_RIGHT_GREATER_SEQ_LENGTH);
+
+    ret = tsk_migration_table_clear(&tables.migrations);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_migration_table_add_row(
+        &tables.migrations, 0.6, 0.5, 1, 0, 1, 1.5, NULL, 0);
+    CU_ASSERT_FATAL(ret >= 0);
+    ret = tsk_table_collection_check_integrity(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_EDGE_INTERVAL);
```
jeromekelleher, line 3121: :heart: great tests!

petrelharp:
There; how's that.

jeromekelleher:
Looks good, you've got a memory leak in the tests though.

petrelharp:
Should be good now.

jeromekelleher:
Looks good, thanks @petrelharp. I've asked the magic pixie to merge it!

PR #601, opened 2020-05-11
Struct codec

benjeffery:
Implementation of the `struct` codec for metadata. Supports encoding an arbitrary nesting of variable-length arrays and fixed-key-set objects up to the python stack depth. Only overhead is array length as keys and types are not stored in the encoded bytes but read from the schema.

TODOs:
- [x] Schema: Enforce all object keys must be required and no additional
- [x] Schema: Enforce `binaryFormat` strings by regex
- [x] ~~Schema: Enforce `index` to be unique and numeric~~
- [x] Schema: No union types
- [x] JSONSchema `null` type handling
- [x] Investigate replacing `index` with alphabetic key sort
- [x] ~~`array` type in JSONSchema allows heterogeneous arrays, support?~~ (No! Forbid in schema)
- [x] Force endian-ness
- [x] ~~Support variable length strings?~~ Deferring to another PR if wanted. #613 
- [x] ~~Perf pass~~ deferring to another PR #614 
- [x] Docs
- [x] Use iterator in decode
- [x] Support padding
- [x] Reinstate index as an optional parameter
- [x] Allow specifying array length format
- [x] Add string encoding parameter
- [x] Test with SLiM metadata
- [x] Extract encode/decode functions and test individually
- [x] Don't catch all types, but error on unsupported.
- [x] Add note about `p` type needing an extra byte
- [ ] File issues about numpy encoding, variable length strings, efficient `enum` encoding, bytes, and compression.

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -23,10 +23,12 @@
 """
 Tests for metadata handling.
 """
+import collections
 import io
 import json
 import os
 import pickle
+import struct
 import tempfile
 import unittest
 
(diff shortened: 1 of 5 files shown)


petrelharp:
I'm not parsing exactly what's allowed here - will SLiM's metadata fit? It sounds like yes except maybe the mutation metadata, that is a [variable-length list of fixed structs](https://github.com/tskit-dev/pyslim/blob/master/pyslim/slim_metadata.py#L43)

benjeffery:
The only thing that is not allowed are objects that have missing keys or extra keys, so a variable length list of fixed objects that all have all the keys is fine.

benjeffery:
@jeromekelleher Just needs docs. So I don't have to rewrite docs after changes maybe review the code now?

python/tests/test_metadata.py
```diff
@@ -400,3 +401,422 @@ def decode(self, encoded):
...
+    def test_missing_binary_format(self):
+        schema = {
+            "codec": "struct",
+            "type": "number",
+        }
+        with self.assertRaisesRegex(
+            exceptions.MetadataSchemaValidationError,
+            "number type must have binaryFormat set",
+        ):
+            metadata.MetadataSchema(schema)
```
jeromekelleher, line 822: This is excellent! Some additional tests would be good

- It would be nice to have examples in which we compare the encoded value to the known bytes representation. For example, a ``c`` encoding of 1, should have bytes value ``b[1]``. We could just compare directly to the expected output of struct.pack, but also good to have the raw bytes in there for a few examples as a sanity check.
- Likewise, we want to check the expected encoding for a few more complicated examples. For instance, we want to make sure that we're definitely not putting in extra padding bytes when we have, say a "b" and "Q" value packed next to each other.
- What happens when the user tries to specify endianness? Nice to see an explicit test for this.
- What happens when we try to decode garbage? Any potential trip hazards here?
- We want to capture the truncation of long strings behaviour and test for this. What happens when someone tries to encode too large an integer?

I think it'd be good to split things up into a few test classes here. What you have are mostly ``TestStructRoundTrip``, and we want to also have ``TestStructInterface`` and ``TestStructEncoding`` as well, I guess.

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
         return data
 
 
+def binary_format_validator(validator, types, instance, schema):
```
jeromekelleher, line 126: This feels slightly strange to be yielding an exception instance. I guess this is the interface defined by upstream? Be good to note this.

python/tests/test_metadata.py
jeromekelleher: Ah, also: strings. Are these strictly utf8? Can we put in unicode in there? How is the encoding decided? Is the length of the field the number of bytes, or the encoded width of the string? Need some tests around this.

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
...
+META_SCHEMA["properties"]["type"] = {"$ref": "#/definitions/simpleTypes"}
+META_SCHEMA["definitions"]["root"]["properties"]["type"] = {
+    "$ref": "#/definitions/simpleTypes"
+}
+# No hetrogeneous arrays
+META_SCHEMA["properties"]["items"] = {"$ref": "#/definitions/root"}
+META_SCHEMA["definitions"]["root"]["properties"]["items"] = {
+    "$ref": "#/definitions/root"
+}
+# binaryFormat matches regex
```
jeromekelleher, line 155: Nice! Use the schema to encode structiness. I like it.

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
...
+    "type": "string",
+    "pattern": r"^[cbB\?hHiIlLqQfd]|\d*[sp]$",
+}
+StructCodecSchemaValidator.META_SCHEMA = META_SCHEMA
+
+
+class StructCodec(AbstractMetadataCodec):
+    """
+    Codec that encodes data using struct. Note that this codec has extra restrictions
+    Namely that object keys must be fixed (all present and no extra); each entry should
```
jeromekelleher, line 170:
```suggestion
    Codec that encodes data using struct. Note that this codec has extra restrictions:
    object keys must be fixed (all present and no extra); each entry should
```

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
...
+        elif isinstance(obj, list) or isinstance(obj, tuple):
+            return [cls.order_by_index(v, False) for v in obj]
+        else:
+            return obj
+
+    def __init__(self, schema: Mapping[str, Any]) -> None:
+        # This codec requires that all properties are required and additional ones
+        # not allowed. Rather than get schema authors to repeat that everywhere
+        # we add it here
+        def enforce_fixed_properties(obj):
```
jeromekelleher, line 198: Be helpful to note here why we can't do this with the metaschema.

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
...
+
+        def make_decode(sub_schema):
+            """
+            Create a function that can decode objects of this schema
+            """
+            if sub_schema["type"] == "array":
+                element_decoder = make_decode(sub_schema["items"])
+                array_length_size = struct.calcsize("<Q")
+
+                def array_decode(bytes_, offset):
```
jeromekelleher, line 251: Just a comment really. I find ``bytes_`` a bit awkward and would tend to use ``buff`` or ``buffer`` for variables like this. No biggie either way.

python/tskit/metadata.py
```diff
@@ -120,6 +123,191 @@ def decode(self, data: bytes) -> bytes:
...
+                    for key, sub_decoder in sub_decoders.items():
+                        obj, offset = sub_decoder(bytes_, offset)
+                        ret[key] = obj
+                    return ret, offset
+
+                return obj_decode
+
+            elif sub_schema["type"] == "string":
+                size = struct.calcsize("<" + sub_schema["binaryFormat"])
+                return lambda bytes_, offset: (
```
jeromekelleher, line 279: These lambdas and private functions are getting a bit cryptic. Maybe we could make these into a set of small module level functions instead, like ``_struct_codec_decode_string`` or something? This would also make it easier to do a few small unit tests on them.

python/tskit/metadata.py
```diff
@@ -146,15 +334,20 @@ def __init__(self, schema: Optional[dict]) -> None:
...
             try:
-                codec_instance = codec_registry[codec](schema)
+                codec_cls = codec_registry[schema["codec"]]
             except KeyError:
                 raise exceptions.MetadataSchemaValidationError(
                     f"Unrecognised metadata codec '{schema['codec']}'. "
                     f"Valid options are {str(list(codec_registry.keys()))}."
                 )
+            codec_instance = codec_cls(schema)
+            # If the codec has modified the schema on __init__, get it back.
```
jeromekelleher, line 346: This is strange - need a bit more explanation of what's going on in the comment here.

jeromekelleher:
Oh yeah, another nice test: take some SLiM metadata encoded as bytes (can just put the value directly into the source), craft a schema to decode it and compare against the known value. Should be easy enough to get an example and the known values using pyslim.

This is exciting!!! Genuine interoperability across programs and platforms!!!

benjeffery:
From a call today - for interop need both padding and order - use `index` to force ordering. Also array length format should be specifiable.
Extract decoders and test individually.

benjeffery:
For string type add encoding to schema (default utf-8).

python/tests/test_metadata.py
benjeffery: I've added quite a few tests for strings, alongside adding the `stringEncoding` and `nullTerminated` parameters.

python/tskit/metadata.py
benjeffery: Comment added in https://github.com/tskit-dev/tskit/pull/601/commits/0fb1e66b68b8a060eb49ffaff527170cbc6a66f1

python/tskit/metadata.py
benjeffery: Note added in https://github.com/tskit-dev/tskit/pull/601/commits/09cbea7cde5ced87bb5229f921f39cf32ca56de9

python/tskit/metadata.py
benjeffery: Buffer is better, renamed.

python/tskit/metadata.py
benjeffery: Broken up and individually tested.

python/tskit/metadata.py
benjeffery: Refactored to make intent clear in https://github.com/tskit-dev/tskit/pull/601/commits/0617b6087274df382489b3b47e181f2e4505e823

benjeffery:
I've managed to get some SLiM generated metadata decoded in https://github.com/tskit-dev/tskit/pull/601/commits/b73c80ff8aecc040d12ca192c054c57d7829f0e3

However there is an issue - as detailed at https://github.com/tskit-dev/pyslim/blob/master/pyslim/slim_metadata.py#L66 SLiM uses the length of the metadata buffer to encode the length of the mutations array. This is not supported in the current metaschema. We could support this by adding a `noLengthEncoding` or similarly-named parameter to the `array` type in the schema, which causes the codec to read elements until the buffer is exhasuted.

benjeffery:
@jeromekelleher I think this is ready for another pass. I at least need a break from it to be able to come back and tell if it is any good!

Some aspects we might like to make issues for, which I think are best not rolled into this PR as it is already large!

1. Variable length strings - easily supported but have an irreducible perf hit due to an extra `pack/unpack` call. Any structure with a variable string in cannot be decoded in one chunk.
2. Enumerated types - could support these with the schema defining the values and the codec using the minimum number of bytes to represent them. Note that some of the SLiM metadata is already of this type, e.g. nucleotide, sex.
3. Storage of `bytes`. There is no way in the schema to round trip a python `bytes` string. Also not possible in JSON: `TypeError: Object of type 'bytes' is not JSON serializable` We could add a type `bytes` to the metaschema, or add `encoding: None` to the string type.
4. Support for numpy arrays.
5. Compression - the codec could support compression.

petrelharp:
Ping @bhaller: https://github.com/tskit-dev/tskit/pull/601#issuecomment-631820502

benjeffery:
> since the choice is between extending the metadata schema....

I've already extended the schema, see the SLiM tests in the diff of this PR.

bhaller:
> > since the choice is between extending the metadata schema....
> 
> I've already extended the schema, see the SLiM tests in the diff of this PR. 

Hi @petrelharp @benjeffery .  Great; that seems like the right choice to me – we don't want to change the SLiM metadata format unless we really have to, as it would just introduce headaches for backward compatibility etc.

I would be interested to have a bit of background about this whole thing.  At the top, it says it "fixes #601" but if I click on that link it just takes me back to this PR; so this PR apparently fixes itself.  :->  Not sure what's going on with that.  Also at the top it says "Implementation of the struct codec for metadata."  I guess "the struct codec" is a Python thing?  What will this ultimately provide to the user?  Anyway, no need to spend time explaining to me now; I imagine all will become clear to me once the doc for this has been written.  :->

benjeffery:
@bhaller some docs here: https://tskit.readthedocs.io/en/latest/metadata.html

petrelharp:
Nice solution. Minor suggestion: how about `lengthFromBuffer` rather than `noLengthEncodingExhaustBuffer`?

benjeffery:
I wanted to make it long and scary looking as it should only be used in specific legacy situations like this.

bhaller:
> I wanted to make it long and scary looking as it should only be used in specific legacy situations like this.

Why is that?  It seems pretty natural to allow the number of entries to be determined by the length of the buffer; why wouldn't new users of this stuff want to utilize this option?

benjeffery:
> Why is that? It seems pretty natural to allow the number of entries to be determined by the length of the buffer; why wouldn't new users of this stuff want to utilize this option?

There's a couple of reasons.

Firstly an array with this option needs to be the last object read from the buffer. It's easy to configure this wrong - the error message you get in such a situation is not obvious and even worse bad data could be read back silently without error if the user sets this option on two arrays. 

Secondly a key use-case for the metadata schemas is adding metadata to a tree sequence that already has some, say in a downstream tool or user code. In this scenario the schema is read, modified to add the new fields and then written back. If the existing schema has a `noLengthEncodingExhaustBuffer` array then either the option needs to be removed or the `index` of the array (and all it's parent objects) forced to be last, this is brittle and not straight forward.

@jeromekelleher This second point reminded me of a thought. Maybe we should force the top-level type to be `object` as it will mean extra metadata can be added by other tools, without breaking code that was designed for the un-modified metadata, and without the other tool needing to pick a key for the original tool's metadata. One of the cool things about this codec is that the binary representation of an object with one property is identical to the representation of that property, so no change is needed for SLiM, for example, for this to be enforced. For example accessing metadata like this:
```python
tables.mutations[0].metadata[index]
```
and
```python
tables.mutations[0].metadata['SLiM'][index] 
```
Is just a change of schema, not binary representation. I think I just convinced myself we should force this, happy to talk it over tomorrow.

bhaller:
Thanks for the explanation, makes sense.

python/tskit/metadata.py
```diff
@@ -120,6 +128,310 @@ def decode(self, data: bytes) -> bytes:
...
+        return lambda buffer: {
+            key: sub_decoder(buffer) for key, sub_decoder in sub_decoders.items()
+        }
+
+    @classmethod
+    def make_string_decode(cls, sub_schema):
+        f = "<" + sub_schema["binaryFormat"]
+        size = struct.calcsize(f)
+        encoding = sub_schema.get("stringEncoding", "utf-8")
+        nullTerminated = sub_schema.get("nullTerminated", False)
```
jeromekelleher, line 303: nitpick: we're using snake_case for Python variables elsewhere.

benjeffery:
> We're not seeing codecov reports - can you confirm the diff coverage is 100% here @benjeffery?

They are at the top of the thread, but yes 100% https://github.com/tskit-dev/tskit/pull/601#issuecomment-626432057

python/tskit/metadata.py
benjeffery: Heh, too much JSON! Fixed in https://github.com/tskit-dev/tskit/pull/601/commits/f65bc76d411bd231877dc3c8dc45510479e381d7

benjeffery:
@jeromekelleher Agreed about the docs, the initial structure has broken down with all the extra parameters that were added. Shall I go ahead to squash, merge and file the 5 issues at https://github.com/tskit-dev/tskit/pull/601#issuecomment-632226469 , plus an issue for metadata docs pass? (Sorry this PR has a high R0!)

jeromekelleher:
Yes, please go ahead and squash/merge this one and file issues for any stuff that's remaining.

PR #604, opened 2020-05-11
Plot labels without decimal places if all integers

### LINKED ISSUE #597: Remove decimal places from ts.draw_svg axis for discrete recomb maps
We should just plot integer numbers on the tree sequence x axis, not e.g. 10.00

hyanwong:
Fixes #597. Tests to follow

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1750,6 +1750,15 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestCase):
         svg = ts.draw_svg()
         self.verify_basic_svg(svg, width=200 * ts.num_trees)
 
+    def test_draw_integer_breaks_ts(self):
+        r_map = msprime.RecombinationMap.uniform_map(1000, 0.001, num_loci=1000)
+        r_map = msprime.RecombinationMap.uniform_map(1000, 0.005, num_loci=1000)
+        ts = msprime.simulate(5, recombination_map=r_map, random_seed=1)
+        svg = ts.draw_svg()
+        self.verify_basic_svg(svg, width=200 * ts.num_trees)
+        for b in ts.breakpoints():
+            self.assertNotEqual(svg.find(f">{b:.0f}<"), -1)
+
     def test_draw_even_height_ts(self):
         ts = msprime.simulate(5, recombination_rate=1, random_seed=1)
         svg = ts.draw_svg(max_tree_height="tree")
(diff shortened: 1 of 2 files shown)


python/tskit/drawing.py
```diff
@@ -291,13 +291,20 @@ def __init__(
         tree_x = self.treebox_x_offset
         break_x = self.treebox_x_offset
```
jeromekelleher, line 293: A bit less code:
```python
breakpoints = ts.breakpoints(as_array=True)
integer_breakpoints = np.all(breakpoints == np.floor(breakpoints))
```

Looks good otherwise.

hyanwong: Yeah, I started writing this by using the numpy breakpoints array, but then I wondered if we would ever want to plot a tick label that wasn't at a breakpoint (e.g. evenly spaced axis ticks), so I separated the concept of a label from the actual value dumped into the label. Happy to revert to using the breakpoints array, though.

jeromekelleher: OK, well why not go through the list of labels afterwards then? Interspersing the logic for checking whether they are integers or not with the code for putting the labels into the list makes it more likely that this check will be forgotten later, if we do change the code. So,
```python
integer_breakpoints = all(round(label) == label for _, _, label in xticks)
```

hyanwong: Done in latest push, and tests added

python/tests/test_drawing.py
```diff
@@ -1750,6 +1750,15 @@ def test_draw_simple_ts(self):
         svg = ts.draw_svg()
         self.verify_basic_svg(svg, width=200 * ts.num_trees)
 
+    def test_draw_integer_breaks_ts(self):
+        r_map = msprime.RecombinationMap.uniform_map(1000, 0.001, discrete=True)
```
jeromekelleher, line 1754: ``discrete=True`` is msprime 1.0 and hasn't been released. Need to use the ``num_loci=1000`` argument instead.

hyanwong: Yep, sorry. Fixed.

jeromekelleher:
Needs a rebase. Would you mind checking the [network](https://github.com/tskit-dev/tskit/network) tab on these please? When things are moving quickly you basically need to rebase the PR at each update - that gives us a chance to merge while it's at the tip of master.

hyanwong:
> Needs a rebase. Would you mind checking the [network](https://github.com/tskit-dev/tskit/network) tab on these please? When things are moving quickly you basically need to rebase the PR at each update - that gives us a chance to merge while it's at the tip of master.

Damn, sorry. I had meant to do this but forgot. Hopefully Ben's automation will mean you don't have to keep an eye on this sort of thing yourself in the future

jeromekelleher:
Merged, thanks @hyanwong.

Issue #598, opened 2020-05-08
KC Distance seg faults on non-sample leaves


daniel-goldstein:
Likely stems from the synonymous use of leaf and sample everywhere, which is not very consistent or correct. Related to #575

jeromekelleher:
I think the right thing to do here is to change our definition of KC to be in terms of samples instead of leaves. Should be straightforward enough.

daniel-goldstein:
Fixed by #610

PR #610, opened 2020-05-12
Update KC metric to work on samples not leaves

### LINKED ISSUE #575: Enable internal samples for KC distance
Internal samples are not currently supported in `Tree.kc_distance`, as it assumes leaves are synonymous with samples. However, the metric should show the similarity between trees with respect to the relationships between samples. This is a pretty straightforward change since the metric just focuses on pairs of nodes and their mrca, but it's worth noting the difference between leaves and samples and the effect on the metric in the docs.

daniel-goldstein:
KC distance previously assumed that samples == leaves. However, the metric should measure the similarities of trees based on the relationships between the samples. This changes the metric to exclusively consider samples, not leaves. This means that internal samples no longer throw an exception and subtrees without any samples in them do not contribute to the KC distance.

It also makes more sense to focus on solely samples because the set of leaves is not consistent across trees in a tree sequence. Internal nodes that are roots in some trees and not in others affect how many leaves there are per tree. The sample set, however, is consistent across the sequence.

Resolves #575 #598

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -4529,7 +4529,7 @@ kc_vectors_free(kc_vectors *self)
 }
 
 static inline void
-update_kc_vectors_single_leaf(
+update_kc_vectors_single_sample(
     tsk_treeseq_t *ts, kc_vectors *kc_vecs, tsk_id_t u, double time)
 {
     const tsk_id_t *sample_index_map = ts->sample_index_map;
@@ -4543,25 +4543,19 @@ static inline void
 update_kc_vectors_all_pairs(tsk_tree_t *tree, kc_vectors *kc_vecs, tsk_id_t u,
     tsk_id_t v, tsk_size_t depth, double time)
 {
-    tsk_id_t leaf1_index, leaf2_index, leaf1, leaf2, n1, n2, tmp, pair_index;
-    tsk_treeseq_t *ts = tree->tree_sequence;
-    const tsk_id_t *restrict samples = ts->samples;
+    tsk_id_t sample1_index, sample2_index, n1, n2, tmp, pair_index;
     const tsk_id_t *restrict left_sample = tree->left_sample;
     const tsk_id_t *restrict right_sample = tree->right_sample;
     const tsk_id_t *restrict next_sample = tree->next_sample;
-    const tsk_id_t *restrict sample_index_map = ts->sample_index_map;
     tsk_size_t *restrict kc_m = kc_vecs->m;
     double *restrict kc_M = kc_vecs->M;
 
-    leaf1_index = left_sample[u];
-    while (true) {
-        leaf1 = samples[leaf1_index];
-        leaf2_index = left_sample[v];
-        while (true) {
-            leaf2 = samples[leaf2_index];
-
-            n1 = sample_index_map[leaf1];
-            n2 = sample_index_map[leaf2];
+    sample1_index = left_sample[u];
+    while (sample1_index != TSK_NULL) {
+        sample2_index = left_sample[v];
+        while (sample2_index != TSK_NULL) {
+            n1 = sample1_index;
+            n2 = sample2_index;
             if (n1 > n2) {
                 tmp = n1;
                 n1 = n2;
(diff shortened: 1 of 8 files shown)


c/tskit/trees.c
```diff
@@ -4789,20 +4771,21 @@ check_kc_distance_tree_sequence_inputs(tsk_treeseq_t *self, tsk_treeseq_t *other
...
+update_kc_pair_with_sample(tsk_tree_t *self, kc_vectors *kc, tsk_id_t sample,
     tsk_size_t *depths, double root_time)
 {
     tsk_id_t c, p, sib;
     double time;
     tsk_size_t depth;
     double *times = self->tree_sequence->tables->nodes.time;
 
-    for (c = leaf, p = self->parent[leaf]; p != TSK_NULL; c = p, p = self->parent[p]) {
+    for (c = sample, p = self->parent[sample]; p != TSK_NULL;
```
jeromekelleher, line 4782: Eek, didn't notice the comma operator before. I'd prefer to avoid this - presumably it's the same thing if we maintain the c variable within the loop body?

python/tests/test_topology.py
```diff
@@ -220,24 +217,24 @@ def update_kc_vectors_single_leaf(kc_vecs, u, time, sample_index_map):
...
+    sample1_index = c1_left
     while True:
         c2_left = tree.left_sample(c2)
         c2_right = tree.right_sample(c2)
-        leaf2_index = c2_left
+        sample2_index = c2_left
         while True:
-            leaf1 = leaves[leaf1_index]
-            leaf2 = leaves[leaf2_index]
+            leaf1 = samples[sample1_index]
```
jeromekelleher, line 229: s/leaf/sample presumably?

python/tests/test_topology.py
```diff
@@ -220,24 +217,24 @@ def update_kc_vectors_single_leaf(kc_vecs, u, time, sample_index_map):
...
+            leaf2 = samples[sample2_index]
             update_kc_vectors_pair(kc_vecs, leaf1, leaf2, depth, time, sample_index_map)
```
jeromekelleher, line 231: Can't you pass ``sample1_index`` and ``sample2_index`` in directly here, and avoid using ``sample_index_map`` in update_kc_vectors_pair? Looks like we're going index -> node ID -> index here.

python/tests/test_topology.py
```diff
@@ -267,14 +264,13 @@ def c_kc_distance(tree1, tree2, lambda_=0):
     Written without Python features to aid writing C implementation.
     """
     samples = tree1.tree_sequence.samples()
+    if len(samples) != len(tree2.tree_sequence.samples()):
```
jeromekelleher, line 267: tree2.tree_sequence.num_samples is a bit more efficient.

python/CHANGELOG.rst
```diff
@@ -12,6 +12,11 @@ In development
 
 **New features**
 
+- The Kendall-Colijn tree distance metric now operates on samples instead of
```
jeromekelleher, line 15: Is there any potential breakage here? I think we were assuming before that leaves == samples, so it's probably the same thing ultimately, except we didn't support internal samples before.

python/tests/test_topology.py
daniel-goldstein: Definitely, fixed this in C and forgot to do the same in python.

python/CHANGELOG.rst
daniel-goldstein: > Is there any potential breakage here?

No I don't believe so.



> I think we were assuming before that leaves == samples, so it's probably the same thing ultimately, except we didn't support internal samples before.

This is pretty much right. We were enforcing that samples were a subset of leaves, and breaking on leaves that were not samples. So any previously working code must have had samples == leaves. There's "anti-breakage" in that code with non-sample leaves that used to break will now silently pass. The difference between leaves and samples are in the docs, I'm unsure whether it should be reiterated here.

jeromekelleher: We weren't actually enforcing non-sample leaves before through, were we? I.e., I don't think it would have thrown an error if there were leaves that were not samples (as opposed to samples that were not leaves). So, we didn't actually change very much, I think (need to be quite precise about what has changed).

daniel-goldstein: You're right we were not. If we're just sticking to enforcement, we could just say "now permits internal samples".

jeromekelleher: I think that's the right approach - rather than saying we're doing something different now, say we're generalising what was there before.

PR #618, opened 2020-05-15
Implement get_left_child

### LINKED ISSUE #615: Minlex ordering means node label positioning no longer correct
As we decide which side of a branch or node to put the label using `tree.left_sib(u)` (https://github.com/tskit-dev/tskit/blob/bbb020fcebc7fd00a700cd55b55ba8d2c4c8ec04/python/tskit/drawing.py#L626) which no longer always agrees with the minlex ordering.

hyanwong:
Fixes #615. It was logically easier (and probably more generally useful) for me to implement a get_left_child rather than a get_left_sib method, so to position labels we just check to see if a node is the left child of its parent.

Will stick a test in if you think this is a reasonable approach, @jeromekelleher .

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -252,6 +252,7 @@ class TestClosestLeftNode(TestTreeDraw):
 
         m1 = drawing.get_left_neighbour(tree, "minlex_postorder")
         m2 = get_left_neighbour(tree, "minlex_postorder")
+        np.testing.assert_array_equal(m1, m2)
 
     def test_2_binary(self):
         ts = msprime.simulate(2, random_seed=2)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM. A couple of quick tests will be fine, no need for anything elaborate.

hyanwong:
If you are now happy with #570 I'll wait for that to be merged, then sort this out with a simple test or two, as that PR will change the drawing code needed and the test writing.

jeromekelleher:
OK, sounds good. I you want #570 merged, give a quick rebase and we'll get it in next.

hyanwong:
Rebased to master - I noticed that the previous verification method was missing "minlex_postorder": when I add it in, the tests fail, which I wasn't expecting, so I've stuck in the code, but commented it out and opened a review. Do you have any inkling of why this fails @jeromekelleher (I presume you wrote the test)?

python/tests/test_drawing.py
```diff
@@ -252,6 +252,10 @@ def verify(self, tree):
 
         m1 = drawing.get_left_neighbour(tree, "minlex_postorder")
         m2 = get_left_neighbour(tree, "minlex_postorder")
+        # The following should work, but doesn't
+        # np.testing.assert_array_equal(m1, m2)
```
jeromekelleher, line 256: This should work (m1 == m2), but the closest_left_node comparison won't work because that's based on the tree order.

hyanwong:
This can be merged now I think, @jeromekelleher.

PR #620, opened 2020-05-15
Graft tree sequences

petrelharp:
I've done the initial copy-over of some nice work that @mufernando has done on creating a `graft` method (see https://github.com/mufernando/graft ). I'll let him explain the rationale here.


petrelharp:
Wait - never mind - he'll create the PR.

PR #626, opened 2020-05-19
Set style="fill: XXX" rather than fill="XXX"

### LINKED ISSUE #625: Node colours not appearing Tree.draw SVG
When I use the `node_colours` option on `Tree.draw`, the resulting SVG contains the the appropriate `fill="blue"` or `fill="red"` but are not being drawn correctly, seen below.

Screen Shot 2020-05-19 at 11 00 19 AM

@hyanwong Do you have an idea of what's happening here?

hyanwong:
Fixes #625


hyanwong:
Thanks for sorting the tests Ben. I had intended to do this at the weekend, so you saved me some work there!

PR #639, opened 2020-05-23
Force top-level of metadata to be an object

### LINKED ISSUE #634: Enforce that metadata is an object at top level
For interoperability it would be best if metadata was always an object at the top-level, enforce this in the schema - see https://github.com/tskit-dev/tskit/pull/601#issuecomment-632358574 for details.

benjeffery:
Fixes #634

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -344,6 +344,15 @@ class TestMetadataModule(unittest.TestCase):
         with self.assertRaises(ValueError):
             metadata.parse_metadata_schema(json.dumps({"codec": "json"})[:-1])
 
+    def test_bad_top_level_type(self):
+        for bad_type in ["array", "boolean", "integer", "null", "number", "string"]:
+            schema = {
+                "codec": "json",
+                "type": bad_type,
+            }
+            with self.assertRaises(exceptions.MetadataSchemaValidationError):
+                metadata.MetadataSchema(schema)
+
     def test_null_codec(self):
         ms = metadata.MetadataSchema(None)
         self.assertEqual(str(ms), "")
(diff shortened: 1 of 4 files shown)


python/tests/test_metadata.py
```diff
@@ -1417,37 +1552,46 @@ def test_mutation(self):
...
-                "boundsY0": {"type": "number", "binaryFormat": "d", "index": 6},
-                "boundsY1": {"type": "number", "binaryFormat": "d", "index": 7},
-                "boundsZ0": {"type": "number", "binaryFormat": "d", "index": 8},
-                "boundsZ1": {"type": "number", "binaryFormat": "d", "index": 9},
-                "migrationRecCount": {
-                    "type": "integer",
-                    "binaryFormat": "d",
-                    "index": 10,
-                },
+                "SLiM": {
```
jeromekelleher, line 1563: Do we actually need this here? This is already an object without forcing the top level key, isn't it?

benjeffery: The only table that needs this is `mutations` as otherwise its top-level type is `array`. I thought it looked better to be consistent across tables. I think we need to think about how this works with multiple tools adding to metadata - otherwise we'll end up with name clashes and people renaming existing entries to things like `old_X`.

jeromekelleher: I don't think we should recommend every program having one key "programname" - that would be really tedious and error prone.

We don't need to get into this here anyway, as these are just test cases. Unless it's actually needed to make the test cases pass, lets delete the top-level key. For the mutations table the key should be something more informative than "SLiM" - this doesn't tell the read much about what's actually in there. We can make recommentations on how metadata should be structured when multiple tools are expected to update it later.

benjeffery: Yep, I don't think a top-level "programme" key is right either in the general case. I've removed the top-level except for mutations where it is now `stacked_mutation_array`.

Issue #534, opened 2020-04-15
Tree-sequence-level metadata


benjeffery:
As discussed in #491 it would be useful to have top-level metadata, using the same schema and encodings as row-level metadata. #491 is already a large PR so this will be done separately.

hyanwong:
By the way, would we ever want *tree*-specific metadata @benjeffery ? There's really no place for this. We would need a separate "tree" table.

jeromekelleher:
We'd need a clear and compelling justification for this - everything is fundamentally built around the idea that individual trees aren't important, it's the nodes and edges.

hyanwong:
Yes, I generally agree. The recent foray into BEAST trees (thanks a *lot* for looking into this @jeromekelleher) was my initial thought for this, but since that didn't really pan out, I can't think of any immediate justification at the moment, other than people are used to annotating specific trees with additional information. In general I agree that we should force them to think otherwise, though

hyanwong:
For easy reference, the discussion in https://github.com/tskit-dev/tskit/issues/495 details some of the uses of ts-level metadata, and some discussion of how we might want it to work.

benjeffery:
Once #534 is complete, if the user really wanted per-tree metadata it could be stored as an array in the tree-sequence metadata. As this is not a common use-case I think that should suffice.

jeromekelleher:
Good call @benjeffery, that's a very pragmatic idea.

PR #642, opened 2020-05-24
Add tree sequence metadata and metadata_schema to C API

benjeffery:
C API part of #534. Adds `metadata` and `metadata_schema` fields to table collection, with accessors on tree sequence. These store arbitrary bytes and are optional in the file format.


c/tskit/tables.c
```diff
@@ -7000,6 +7051,44 @@ tsk_table_collection_read_format_data(tsk_table_collection_t *self, kastore_t *s
...
     self->file_uuid[TSK_UUID_SIZE] = '\0';
+
+    ret = kastore_containss(store, "metadata");
+    if (ret != 1 && ret != 0) {
+        ret = tsk_set_kas_error(ret);
+        goto out;
+    }
+    if (ret == 1) {
+        ret = kastore_gets_int8(
+            store, "metadata", (int8_t **) &metadata, (size_t *) &metadata_length);
```
benjeffery, line 7062: The cast here and on line 7079 are effectively what is happening in the table metadata already, but I think it isn't _strictly_ right. It is practically fine as it works in our tests, but it's my understanding that C99 only asks that `CHAR_BIT => 8` not `CHAR_BIT==8`

benjeffery:
Looks like I could get a little more coverage here, setting back to draft.

benjeffery:
Added a test on the TS methods. Looks like the rest of the coverage misses are hard error paths. Ready for review.

c/tskit/tables.c
jeromekelleher: Yeah, that is a potential issue. Probably we should just put in a ``assert(CHAR_BIT == 8)`` in the code somewhere central. It's not worth getting worried about more than that though.

c/tskit/tables.c
```diff
@@ -7000,6 +7051,45 @@ tsk_table_collection_read_format_data(tsk_table_collection_t *self, kastore_t *s
     }
     memcpy(self->file_uuid, uuid, TSK_UUID_SIZE);
     self->file_uuid[TSK_UUID_SIZE] = '\0';
+
+    ret = kastore_containss(store, "metadata");
```
jeromekelleher, line 7055: Good catch, this was potentially masking errors. I'd say using ``if (ret < 0) `` is *slightly* more idiomatic for detecting the error condition, but it's not worth holding the PR up for.

python/tests/test_file_format.py
```diff
@@ -825,8 +827,9 @@ def verify_missing_fields(self, ts):
         with kastore.load(self.temp_file) as store:
             all_data = dict(store)
         for key in all_data.keys():
-            # We skip this key as it is optional
-            if "metadata_schema" not in key:
+            # We skip these keys as it is optional
+            if "metadata_schema" not in key and key != "metadata":
+                print(key)
```
jeromekelleher, line 832:
```suggestion
```

benjeffery: Facepalm! So close to a "hole in one" PR. Ah well, will try again next time!

c/tskit/tables.c
benjeffery: Agreed, should be that.

benjeffery:
@jeromekelleher did you see my comment above about the casting to `char`?

benjeffery:
Oops just seen the reply, ignore!

jeromekelleher:
I seem to have messed the PR up here  - I thought you were off today @benjeffery so I force pushed a couple of changes. Looks like I broke stuff.

Crap, I think I pulled in kastore updates by ``git commit -a``.

jeromekelleher:
I made the mess, I'll fix it up!

benjeffery:
I'm off, but can quickly push the branch back to how it was?

jeromekelleher:
> I'm off, but can quickly push the branch back to how it was?

Na, thanks, it's an easy fix.

c/tskit/tables.c
jeromekelleher: opened isssue to track, #645.

PR #650, opened 2020-05-26
Implement algorithm for calculating species tree topology distributions.

daniel-goldstein:
This implements both per-tree and incremental algorithms for computing exact tree topology distributions on tree sequences.

As discussed in my talk last week, this uses a DP approach to compute all possible species topology distributions at each node in the tree, starting from leaves up to roots. The `TopologyTree` class is used to store topology distributions at each node. This approach does not support internal samples, and enforces that samples be a subset of leaves (leaves that are not samples are ignored). This does, nicely, support multiply-rooted trees. As there can't be species topologies that span different roots, we compute the topologies for each root using `TopologyTree` and "add" the resulting distributions together.

Since this approach needs only a single upward traversal, the incremental algorithm is straightforward. We maintain a reference to the `TopologyTree` for each node and reconstruct those `TopologyTree`s on the upward traversals when inserting/removing edge diffs.

Should we follow the pattern of adding wrapper methods on `Tree` and `TreeSequence` (and move the docstrings over) to not have to expose the combinatorics module? If so, should we lift the functions over entirely or make the `Tree{Sequence}` methods one-liners?

cc @jeromekelleher @hyanwong

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -23,11 +23,18 @@
 """
 Test cases for combinatorial algorithms.
 """
+import collections
+import io
 import itertools
 import unittest
 
+import msprime
+import numpy as np
+
+import tests.test_wright_fisher as wf
 import tskit
 import tskit.combinatorics as comb
+from tests import test_stats
 from tskit.combinatorics import RankTree
 
 
@@ -351,6 +358,24 @@ class TestRankTree(unittest.TestCase):
                 self.assertTrue(tree.is_canonical())
                 self.assertEqual(tree, reconstructed)
 
+    def test_from_unary_tree(self):
+        tables = tskit.TableCollection(sequence_length=1)
+        c = tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        p = tables.nodes.add_row(time=1)
+        tables.edges.add_row(left=0, right=1, parent=p, child=c)
+
+        t = tables.tree_sequence().first()
+        with self.assertRaises(ValueError):
+            RankTree.from_tsk_tree(t)
+
+    def test_to_tsk_tree_errors(self):
+        alpha_tree = RankTree.unrank((0, 0), 3, ["A", "B", "C"])
+        out_of_bounds_tree = RankTree.unrank((0, 0), 3, [2, 3, 4])
+        with self.assertRaises(ValueError):
+            alpha_tree.to_tsk_tree()
+        with self.assertRaises(ValueError):
+            out_of_bounds_tree.to_tsk_tree()
+
     def test_rank_errors_multiple_roots(self):
         tables = tskit.TableCollection(sequence_length=1.0)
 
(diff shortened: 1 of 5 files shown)


python/tests/test_combinatorics.py
```diff
@@ -23,8 +23,13 @@
 """
 Test cases for combinatorial algorithms.
 """
+import io
 import itertools
 import unittest
+from collections import Counter
```
jeromekelleher, line 29: I prefer to just import the module and then use ``collections.Counter`` in the code. This makes it obvious where the ``Counter`` class comes from. Otherwise, you see a ``Counter`` class in the code and you're not quite sure if it's the standard library's one, or a home grown version.

python/tskit/combinatorics.py
```diff
@@ -26,11 +26,221 @@
...
+        key = get_population
+
+    topology_tree = np.full(ts.num_nodes, None, dtype=object)
+    parent = np.full(ts.num_nodes, -1)
+
+    def update_state(tree, u):
+        stack = [u]
+        while len(stack) > 0:
+            v = stack.pop()
+            children = []
```
jeromekelleher, line 66: Might as well use the high-level tree way here:
```python
children = tree.children(v)
for c in children:
     # etc
```

python/tskit/trees.py
```diff
@@ -833,6 +833,37 @@ def unrank(rank, num_leaves):
...
+        Internal samples are not supported.
+
+        Produces a dictionary from a ``frozenset`` of labels to
+        ``collections.Counter`` of topologies, where each tree topology is represnted
+        by its rank (see :ref:`sec_combinatorics`). This represents the
+        topologies for a particular set of labels and the number of times
+        each topology appears when subsampling samples from ``tree``.
+
+        :param tree tskit.Tree: The tree to compute the topologies for.
+        :param key: A function of the form ``key(u)`` where ``u`` is a
```
jeromekelleher, line 852: Let's change ``key`` to ``sample_sets`` so that we're using the same concepts as the stats API.

python/tests/test_combinatorics.py
```diff
@@ -413,3 +425,302 @@ def test_is_symmetrical(self):
         self.assertFalse(three_leaf_asym.is_symmetrical())
         six_leaf_sym = RankTree(children=[three_leaf_asym, three_leaf_asym])
         self.assertTrue(six_leaf_sym.is_symmetrical())
+
+
+class TestCountTopologies(unittest.TestCase):
+    def verify_topologies(self, ts, expected=None):
+        populations = [pop.id for pop in ts.populations()]
+        topologies = [t.count_topologies() for t in ts.trees()]
+        inc_topologies = list(ts.count_topologies())
```
daniel-goldstein, line 434: Comment here explaining subsets of populations topologies

daniel-goldstein:
@jeromekelleher Addressed the changes we discussed. Created a `TopologyCounter` which is indexable by a tuple of sample set indexes that's returned from `count_topologies`, and a `PartialTopologyCounter` that handles most of the heavy lifting in the algorithm. With these in place there wasn't much point to a `TopologyTree`, so the per-node state in the counting topologies methods are `TopologyCounter`s.

python/tskit/trees.py
```diff
@@ -833,6 +833,55 @@ def unrank(rank, num_leaves):
...
+        :param list sample_sets: A list of lists of Node IDs, specifying the
+            groups of individuals to compute the statistic with.
+            Defaults to all samples grouped by population.
+        :rtype: tskit.TopologyCounter
+        :raises ValueError: If nodes in ``sample_sets`` are invalid or are
+            internal samples.
+        """
+        if sample_sets is None:
+            sample_sets = [
+                list(self.tree_sequence.samples(population=pop.id))
```
jeromekelleher, line 879: Seems a waste to convert to list here - a numpy array should be fine, right?

python/tests/test_combinatorics.py
```diff
@@ -420,3 +433,331 @@ def test_is_symmetrical(self):
...
+        ]
+
+        pop_configurations = [
+            msprime.PopulationConfiguration(sample_size=size) for size in sample_sizes
+        ]
+        return msprime.simulate(
+            population_configurations=pop_configurations,
+            migration_matrix=migration_matrix,
+            recombination_rate=0.1,
+        )
```
jeromekelleher, line 763: As there's a lot going on in the PartialTopologyCounter, it'd be nice to see a few simple tests for it here. Nothing exhaustive, just a handful of simple examples demonstrating that it does what it's supposed to.

daniel-goldstein: Definitely, I think the tests definitely make it easier to pick apart.

python/tests/test_combinatorics.py
```diff
@@ -420,3 +433,469 @@ def test_is_symmetrical(self):
...
+        with self.assertRaises(IndexError):
+            list(ts.count_topologies(sample_sets))
+
+    def test_standard_msprime_migrations(self):
+        for num_populations in range(2, 5):
+            samples = [5] * num_populations
+            ts = self.simulate_multiple_populations(samples)
+            self.verify_topologies(ts)
+
+    def test_complex_dtwf_topologies(self):
```
jeromekelleher, line 851: This is good, but I was thinking more about the *forwards time* trees that we get out of the local Wright-Fisher simulator. Something like [these three tests](https://github.com/tskit-dev/tskit/blob/master/python/tests/test_stats.py#L293) would be really good. We want to see, in particular, how it deals with the nasty raggedy trees that a WF simulation produces.

python/tskit/trees.py
```diff
@@ -833,6 +833,57 @@ def unrank(rank, num_leaves):
...
+        :param list sample_sets: A list of lists of Node IDs, specifying the
+            groups of nodes to compute the statistic with.
+            Defaults to all samples grouped by population.
+        :rtype: tskit.TopologyCounter
+        :raises ValueError: If nodes in ``sample_sets`` are invalid or are
+            internal samples.
+        """
+        if sample_sets is None:
+            sample_sets = [
+                np.array(self.tree_sequence.samples(population=pop.id))
```
jeromekelleher, line 881: We don't need this cast, do we? I'm sure tree_sequence.samples returns a numpy array.

python/tests/test_combinatorics.py
daniel-goldstein: Those tests work great just drop in. Cool also to test against all sample set partitions!

python/tskit/combinatorics.py
```diff
@@ -24,13 +24,244 @@
...
+
+class TopologyCounter:
+    """
+    Contains the distributions of embedded topologies for every combination
+    of the sample sets used to generate the ``TopologyCounter``. It is
+    indexable by a combination of sample set indexes and returns a
+    ``collections.Counter`` whose keys are topology ranks
+    (see :ref:`sec_tree_ranks`), and whose counts represent the probability
+    of that topology if one randomly selected an sample from each
+    sample set pertaining to that ``Counter``.
```
petrelharp, line 135: the last part of this sentence doesn't quite make sense to me?

python/tests/test_combinatorics.py
```diff
@@ -420,3 +445,516 @@ def test_is_symmetrical(self):
...
+            migration_matrix=migration_matrix,
+            population_configurations=population_configurations,
+            demographic_events=demographic_events,
+            random_seed=2,
+            model="dtwf",
+        )
+
+        self.verify_topologies(ts)
+
+    def test_forward_time_wright_fisher(self):
```
jeromekelleher, line 922: Be a little nicer to split this up into three tests, which are named. That way we can see the different types of forward sim topology that we're testing against. Big monolithic tests are hard to parse quicky for the reader.

python/tskit/trees.py
```diff
@@ -833,6 +833,57 @@ def unrank(rank, num_leaves):
         """
         return combinatorics.RankTree.unrank(rank, num_leaves).to_tsk_tree()
 
+    def count_topologies(self, sample_sets=None):
+        """
+        Calculates the distribution of embedded topologies for every combination
+        of the sample sets in ``sample_sets``. ``sample_sets`` defaults to all
+        samples in the tree grouped by population. In this case, the topologies
+        produced represent evolutionary topologies of the populations in this
+        tree.
```
petrelharp, line 842: I think you can delete this last sentence - I don't think it's adding anything. If you keep it, don't use "evolutionary" - there's nothing about evolution here - perhaps you mean "genealogical"?

python/tskit/combinatorics.py
petrelharp: After reading the documentation for `count_topologies` I know what this does - maybe just refer to that method?

python/tskit/trees.py
```diff
@@ -833,6 +833,57 @@ def unrank(rank, num_leaves):
         """
         return combinatorics.RankTree.unrank(rank, num_leaves).to_tsk_tree()
 
+    def count_topologies(self, sample_sets=None):
+        """
+        Calculates the distribution of embedded topologies for every combination
+        of the sample sets in ``sample_sets``. ``sample_sets`` defaults to all
+        samples in the tree grouped by population. In this case, the topologies
```
petrelharp, line 840: do the sample sets have to be nonoverlapping? (this affects the math below)

python/tskit/trees.py
```diff
@@ -833,6 +833,57 @@ def unrank(rank, num_leaves):
...
+        populations' samples in this tree.
+
+        The counts for each topology in the :class:`tskit.TopologyCounter`
+        are absolute counts that we would get if we were to select all
+        combinations of samples from the relevant sample sets.
+        For sample sets :math:`[s_0, s_1, ..., s_n]`, the total number of
+        topologies for those sample sets is equal to
+        :math:`|s_0| * |s_1| * ... * |s_n|`, so the counts in the counter
+        ``topology_counter[0, 1, ..., n]`` should sum to
+        :math:`|s_0| * |s_1| * ... * |s_n|`.
```
petrelharp, line 861: nice job making it concrete!

python/tskit/trees.py
```diff
@@ -5840,6 +5891,28 @@ def kc_distance(self, other, lambda_=0.0):
         """
         return self._ll_tree_sequence.get_kc_distance(other._ll_tree_sequence, lambda_)
 
+    def count_topologies(self, sample_sets=None):
+        """
+        Generates the same distribution of topologies as :meth:`Tree.count_topologies`
+        but for every tree in a tree sequence. For use on a tree sequence this method is
```
petrelharp, line 5897: How exactly is this combined across all the trees? Is it added together by tree? Weighted by span?

petrelharp:
Looks good! I had some clarifying comments on the docs, and am happy to look again if you want, but don't have to.

python/tskit/trees.py
daniel-goldstein: Yes they do, good point.

python/tskit/trees.py
daniel-goldstein: They are not combined, it produces a generator that yields the per-tree result. Can clarify

daniel-goldstein:
Thanks @petrelharp!

python/tskit/trees.py
daniel-goldstein: Thanks!

jeromekelleher:
OK, great, I'm going to mark this for merging.

@daniel-goldstein, can you open issues to track anything we've left dangling/unfinished here please?

benjeffery:
@mergifyio refresh

benjeffery:
Travis AWOL again! Kicking it.

PR #653, opened 2020-05-27
Document enumerating tree topologies in combinatorics module

### LINKED ISSUE #643: Document tree rank/unrank
#510 merged support for ranking and unranking trees. It added a stub ``combinartorics.rst`` file, which we need to fill out.

This file should explain the basic approach, and what the ranks mean. A lot of the content can probably be pulled out of the current documentation within docstrings. There's currently a few long, detailed docstrings that explain how things work - this content can be moved into the rst file.

daniel-goldstein:
This fills out the top-level documentation for the combinatorics module. It addresses the two key pieces of functionality: ranking/unranking trees and enumerating all trees, and how the two are intimately connected. Also provides a short example of using ranks to count tree topologies in a tree sequence.

Resolves #643


docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
 .. _sec_combinatorics:
 
-=====================================
-Ranking and Unranking Tree Topologies
-=====================================
-TODO
+=============
+Combinatorics
+=============
+This module implements a combinatorial approach to counting topologies
```
jeromekelleher, line 6: We're not implementing it as an externally visible module - can you change the intro to talking about tskit in general rather than the combinatorics module?

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+of leaf-labelled trees. It provides methods for enumerating
+all possible trees with a given number of labelled leaves, as well as
+converting back and forth between a tree and its position, or rank, in
+the enumeration of all possible topologies.
+
++++++++++++++++++++++++++++++++++
+Ranking/Unranking Topologies
++++++++++++++++++++++++++++++++++
+The enumeration of leaf-labelled trees discussed below allows for an
```
jeromekelleher, line 15: Which trees? It feels like there used to be a sentence before this. It would be good to briefly discuss what kind of trees we're talking about, I guess.

jeromekelleher: Maybe helpful here to show all the trees of 4 (or something) just as an intuitive way into things?

jeromekelleher: The concept of rank should be clear then.

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+    ``shape rank`` and ``label rank``.
+
+unrank
+    The action of reconstructing a leaf-labelled tree from its ``rank``.
+
+++++++++++++++++++++++
+Enumerating Topologies
+++++++++++++++++++++++
+
+Below we outline the approach for enumerating all unique topologies
```
jeromekelleher, line 50: There's a lot of "Belows" and stuff. How about  "Enumerating all unique topologies of leaf-labelled trees with n leaves consists of two steps".

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+    :width: 32%
+
+Note that the right subtree is symmetrical, so we must always give the smaller
+label to the left leaf in that subtree to avoid redundant labellings.
+
+This *canonical labelling* removes any redundant trees and allows us the
+following enumeration algorithm, which is presented below in both the
+binary and general case.
+
+Examining first the binary case, let :math:`T = (T_j, T_k)`, with left subtree
```
jeromekelleher, line 150: At this point, I'm not sure that the algorithm is what we want. Probably best to keep this level of detail for a more technical writeup. I guess what we're trying to get across here in these docs is, *what* it does, not *how* it does it.

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+    :math:`n` leaves.
+
+label rank
+    For a tree :math:`T`, the label rank is the position of the labelling of
+    :math:`T`'s leaves in the enumeration of all unique ways the leaves of
+    :math:`T` could be labelled.
+
+rank
+    The rank of a leaf-labelled tree is considered to be the tuple of its
+    ``shape rank`` and ``label rank``.
```
petrelharp, line 41: Usually, a rank is a single number, so at first I thought you meant that the rank is the rank of the tuple in the enumerations of all such tuples; but you actually mean the tuple, right? Perhaps this is worth explaining, e.g., adding "(Note: this 'rank' is a pair of numbers, not a single number, but we call it a rank because it can be used to order labeled trees.)"

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+
+++++++++++++++++++++++
+Enumerating Topologies
+++++++++++++++++++++++
+
+Below we outline the approach for enumerating all unique topologies
+of leaf-labelled trees with :math:`n` leaves. This takes part in two
+steps. First, we must consider how to generate unique tree shapes.
+Then, for each shape we must consider how to uniquely assign labels
+to the leaves. Note that this enumeration is not limited to binary trees.
```
petrelharp, line 54: ".. takes part.." --> "This happens in two steps."  ?

petrelharp: Also, instead of "Note that this enumeration is not limited to binary trees", maybe you should say, more explicitly, "This is an enumeration of all general rooted trees, not only binary trees." (oh: the word "rooted" is important, too)

petrelharp: You should probably say early on that all your trees are rooted.

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+Enumerating Tree Shapes
+***********************
+Let :math:`S_n` denote the set of all unlabelled trees, or tree shapes
+of :math:`n` leaves.
+We can see that :math:`S_1` has one element: a single leaf.
+
+A tree with :math:`n > 1` leaves can be obtained by joining at least two trees whose number
+of leaves sum to :math:`n`. This maps very closely to the concept of integer partitions.
+We first compute all integer partitions* of `n`, and then for each partition :math:`P`,
+compute the cross product of all :math:`k`-leaf trees for all :math:`k \in P`.
```
petrelharp, line 66: instead of "cross product", how about "combinations"?

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+***********************
+Enumerating Tree Shapes
+***********************
+Let :math:`S_n` denote the set of all unlabelled trees, or tree shapes
+of :math:`n` leaves.
+We can see that :math:`S_1` has one element: a single leaf.
+
+A tree with :math:`n > 1` leaves can be obtained by joining at least two trees whose number
+of leaves sum to :math:`n`. This maps very closely to the concept of integer partitions.
+We first compute all integer partitions* of `n`, and then for each partition :math:`P`,
```
petrelharp, line 65: shouldn't you say "all nondecreasing integer partitions", and explain what "nondecreasing" means?

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+.. note::
+    Using non-decreasing integer partitions enforces a *canonical orientation* on the tree
+    shapes, where children under a node are ordered by the number of leaves below them.
+    This is important because it prevents us from repeating trees that are topologically
+    the same but whose children are ordered differently.
+
+However, this method has the potential to double-count trees when we encounter
+partitions with repeated numbers of leaves. To remove these redundant trees we
+reduce the set of trees so that if a subtree has the same number of leaves as
+its left sibling, it must have an equal or greater shape rank.
```
petrelharp, line 90: very nice solution!

daniel-goldstein:
> This is great @daniel-goldstein! I think we can keep the detailed algorithmic desciptions out of here (we're explaining _what_ not _how_).
> 
> What we're missing is references to the actual methods. I think it would be good to have a table (or something) right at the top of the page that list does something like
> 
> Tree.rank | Return the rank of a tree
> Tree.unrank | Returns a tree with a given rank
> all_labelled_trees | Returns a generator over all labelled trees with n leaves.
> 
> Where the function name is a click-through link to the documentation.
> 
> I think a short demonstration in a section at the end would be really helpful too.

Ok great, that makes sense. I'll save the algorithmic details for a technical write-up (but will incorporate the helpful feedback!)

docs/combinatorics.rst
petrelharp: So, I guess that what you're doing is saying the subtrees must be ordered by `(size, rank)`? So the rank of a tree among trees with `n` leaves is the same as the rank of its ordered list of subtrees: `((size_1, rank_1), ..., (size_n, rank_n))` among all such ordered lists?

daniel-goldstein: Yes, for unlabelled trees. The definition is extended when you add labels. Precisely, it's

https://github.com/tskit-dev/tskit/blob/72ac9d2472f645163c0675c4413ddc3cb6e709b6/python/tskit/combinatorics.py#L264-L269

docs/combinatorics.rst
```diff
@@ -1,6 +1,202 @@
...
+
+In the case of binary trees, this simplifies to:
+Let :math:`P = [j, k]` be a partition of :math:`n`. If :math:`j < k`,
+then the trees generated by :math:`P` are isomorphic to :math:`S_j \times S_k`.
+If :math:`j = k`, then the trees generated are isomorphic to all 2-element
+combinations with replacement of trees from :math:`S_j`.
+This generalizes to the following algorithm run per partition:
```
petrelharp, line 97: Should say what this is an algorithm to do (to enumerate binary, rooted trees with n leaves, right?).  Is this something you implement? If not, maybe you don't need it, or else say that it's for illustration.

docs/combinatorics.rst
daniel-goldstein: Do you want to keep the generating functions in the combinatorics module, or put them under a different namespace? That would affect whether the module needs to be visible.

docs/combinatorics.rst
petrelharp: Agree - I don't think this is an enumeration algorithm? I see what the ordering *is*, but not everything you say below comes in a natural order, as far as I can tell.

Oh, wait: actually, the ordering itself is a bit confusing, as it's defined implicitly in terms of the orderings of smaller-n labelings? You've said how to compare labelings with different labels; but to rank labelings you need to also compare labellings with the same labels. Maybe inserting a definition along the lines of the one above would be good. I guess that this is that the rank of an labeling of a tree is its rank among sorted tuples `((size_1, shape_1, labeling_rank_1), ... (size_1, shape_1, labeling_rank_1))`, where `labeling_rank_1` is the rank of that particular labeling of labels [1...n] to the subtree with shape `shape_1` and size `size_1` (which is less than `n`)?

I may not have been very clear there. I am guessing that the labeling
```
  ┃  
 ┏┻━┓
 ┃  ┃
┏┻┓ ┃
┃ ┃ ┃
1 2 0
```
comes before
```
  ┃  
 ┏┻━┓
 ┃  ┃
┏┻┓ ┃
┃ ┃ ┃
0 2 1
```
because the smallest (and thus earliest) subtree has the smaller label in the first labeling?  Is this true?

daniel-goldstein: @petrelharp I *think* I understand what you're asking and yes, there is a piece missing here. The way in which labellings are created is directly dependent on how you choose combinations of labels when splitting labels up between subtrees. This uses a lexicographic ordering for choose that comes with `itertools.combinations` to select labels to assign to subtrees, so for the trees you mention it would be ordered (0, (1, 2)), (1, (0, 2)), (2, (0, 1))

petrelharp: Ah-ha - that's equivalent to what I said, but simpler. So, you're saying that labelings are ordered lexicographically, basically, with the grouping of labels determined by the subtrees, and the ordering of these groups determined by subtree order as described before.

docs/combinatorics.rst
jeromekelleher: We can make them visible externally by exporting from ``__init__.py``, like 
```
from . combinatorics import all_tree_topologies 
```
 etc. (Maybe this is a better name than ``all_unlabelled_trees``?)

daniel-goldstein: > We can make them visible externally by exporting from `__init__.py`, like

Sounds good (I'm flattered my functions get top-level)

> etc. (Maybe this is a better name than all_unlabelled_trees?)

I think we've been referring to topologies without further specification as the leaf-labelled trees. Based on what I have now, we could do:

```
all_tree_topologies(n) (right now this is all_trees)
all_tree_shapes(n)
all_tree_labellings(t)
```

or could switch it to something like:

```
all_labelled_tree_topologies(n)
all_tree_topologies(n)
all_tree_labellings(t)?
```

daniel-goldstein:
Thanks @jeromekelleher @petrelharp! I've incorporated your feedback and moved the algorithmic discussion out of this doc and saved it for later. Went short and sweet here with laying out the current functionality and briefly explaining what these weird ranks are.

docs/combinatorics.rst
```diff
@@ -1,6 +1,116 @@
...
+:meth:`Tree.rank`                     Return the rank of this tree.
+:meth:`Tree.unrank`                   Return a Tree given its rank and
+                                      a number of leaves.
+:func:`tskit.all_tree_topologies`     Return a generator over all
+                                      leaf-labelled trees of n leaves.
+:func:`tskit.all_tree_shapes`         Return a generator over all
+                                      tree shapes of n leaves.
+:func:`tskit.all_tree_labellings`     Return a generator over all
+                                      labellings of the given tree's shape.
+=================================     =====================================
```
daniel-goldstein, line 24: @jeromekelleher Went with this naming pattern but happy to change it up if you have a better suggestion.

docs/combinatorics.rst
```diff
@@ -1,6 +1,116 @@
...
 
-=====================================
-Ranking and Unranking Tree Topologies
-=====================================
-TODO
+=============
+Combinatorics
+=============
+tskit uses a combinatorial approach to identify unique topologies of
+rooted, leaf-labelled trees. It provides methods for enumerating
```
petrelharp, line 8: whoops - one more caveat - "rooted, leaf-labeled trees without unary nodes"?

petrelharp:
Gee, I'm worried I gave you the wrong idea! A bunch of that stuff you had written earlier was very good, and very helpful! It explained what exactly the rank is, which was really neat!  It just needed a bit of adjustment, I think. I'd vote to put some of it back in - at least, it'd be nice to have a concrete definition of how the rank is computed?

daniel-goldstein:
> Gee, I'm worried I gave you the wrong idea! A bunch of that stuff you had written earlier was very good, and very helpful! It explained what exactly the rank is, which was really neat! It just needed a bit of adjustment, I think. I'd vote to put some of it back in - at least, it'd be nice to have a concrete definition of how the rank is computed?

No worries! I figured it'd be easier to overcorrect and then add things back from my other commit, but I am having trouble finding the line at which it becomes overly technical. Maybe just a small bit about how shapes are computed from partitions (perhaps with the accompanying diagram I had) and then combinations of labels selected in a lexicographic order?

docs/combinatorics.rst
jeromekelleher: I was thinking of making "topologies" synonymous with "tree shapes" , but I think you're probably right. How about ``all_tree_topologies`` -> ``all_trees``, and then keep the other two as they are? Usually when someone thinks of a tree in this context it'll be labelled.

daniel-goldstein:
@petrelharp Added a commit with what I think are the important bits about enumerating shapes and labellings that contribute to a complete ordering of topologies.

docs/combinatorics.rst
```diff
@@ -1,6 +1,116 @@
...
+:meth:`Tree.rank` method.
+
+.. code-block:: python
+
+    ranks = [t.rank() for t in tskit.all_tree_topologies(num_leaves=3)]
+    print("Ranks of 3-leaf trees:", ranks)
+
+::
+
+   Ranks of 3-leaf trees: [(0, 0), (1, 0), (1, 1), (1, 2)]
```
jeromekelleher, line 75: Out of interest, is there a bijection of these tuples to integers? It would be nicer if we could just define the rank as an integer.

docs/python-api.rst
```diff
@@ -487,6 +487,18 @@ using a schema. See :ref:`sec_metadata`, :ref:`sec_metadata_api_overview` and
 
 .. _sec_stats_api:
 
+*************
+Combinatorics
+*************
+The following are generators for fully enumerating unique tree topologies.
```
jeromekelleher, line 493: A link to ``:meth:`Tree.rank` and tree.unrank would be helpful here also.

docs/combinatorics.rst
```diff
@@ -1,6 +1,207 @@
...
+
+All tree shapes with four leaves, and the partitions that generate
+them, are:
+
+.. image:: _static/four_leaf_tree_shapes.png
+   :alt: All four-leaf tree shapes and their generating partitions
+
+Note that the middle column reflects all tree shapes of three leaves
+in the right subtree!
+
```
daniel-goldstein, line 156: Probably should mention here that the order of partitions is using ascending compositions?

docs/combinatorics.rst
jeromekelleher: It would be *super* nice if the rank was a direct index into ``list(all_trees(n))``

docs/combinatorics.rst
jeromekelleher: That fitted in naturally to your earlier exposition, I think (which I think we should include as another section at the end). Can link to my thesis, if someone wants to know what an ascending composition is, I guess?

docs/combinatorics.rst
daniel-goldstein: _This_ has been bugging me for a month. A naive bijection would essentially be:
```
sum(num_labellings(s, n) for s in range(shape_rank)) + label_rank
```

but I have not thought of a way yet without running through all preceding shapes, because depending on the number of symmetries shapes can be labelled in varying numbers of ways.

daniel-goldstein:
> I was thinking of making "topologies" synonymous with "tree shapes" , but I think you're probably right. How about all_tree_topologies -> all_trees, and then keep the other two as they are? Usually when someone thinks of a tree in this context it'll be labelled.

I should have waited. I previously had it as `all_trees` 😄 

I think we overlapped in review/new commit @jeromekelleher. To clarify, do you mean that the section I just added "Enumerating Topologies" is sufficient in place of your suggested "Formal definition"? Or are there additional details you think I should bring back?

docs/combinatorics.rst
jeromekelleher: OK, let's not get caught up on it. If we figure out a bijection we can change things later.

jeromekelleher:
> I think we overlapped in review/new commit @jeromekelleher. To clarify, do you mean that the section I just added "Enumerating Topologies" is sufficient in place of your suggested "Formal definition"? Or are there additional details you think I should bring back?

Sorry, you're right. I was commenting on the version just before you put in the section. It looks great now. I do like ``all_trees`` though, it's neat and this is the main function that people will want to use when exploring this stuff.

docs/combinatorics.rst
jeromekelleher: Sorry, I was commenting out of place here. If you like you can stick in a link to my summary page on [ascending compositions](http://jeromekelleher.net/generating-integer-partitions.html) as a reference.

daniel-goldstein:
Great, updated/squashed/pushed.

docs/combinatorics.rst
daniel-goldstein: Whoops sorry forgot this comment.

docs/combinatorics.rst
petrelharp: I dunno - having a 2d index is kinda nicer because it makes it easier to separate tree shape from labeling?

petrelharp:
Let me know when/if you want me to take another read through.

docs/combinatorics.rst
daniel-goldstein: How often would you want to just know about the shape? (labelling means nothing on its own)

daniel-goldstein:
@petrelharp Ready whenever you can take a look!

docs/combinatorics.rst
```diff
@@ -1,6 +1,210 @@
...
+leaves can be obtained by joining at least two trees whose number of
+leaves sum to :math:`n`.
+This maps very closely to the concept of integer partitions.
+Each tree shape of :math:`n` leaves can be represented by taking a
+nondecreasing integer partition of :math:`n` (elements of the partition
+are sorted in nondecreasing order) and recursively partitioning its
+elements. The order in which we select partitions of :math:`n` is
+determined by the efficient
+`rule_asc <http://jeromekelleher.net/generating-integer-partitions.html>`_
+algorithm for generating them.
```
petrelharp, line 149: Why not insert your nice explanation from before of the order that the integer partitions come in?  (I think you had such a description?)

docs/combinatorics.rst
```diff
@@ -1,6 +1,210 @@
...
+standard lexicographic ordering to choose labels. See how the trees
+are sorted by the order in which the left leaf's label was chosen.
+
+.. note::
+    There is a caveat here regarding symmetry, similar to that of repeating
+    tree shapes. Symmetrical trees run the risk of creating redundant labellings
+    if all combinations of labels were exhausted. To prevent redundant labellings
+    we impose a *canonical labelling*. In the case of two symmetrical subtrees,
+    the left subtree must receive the minimum label from the label set. Notice
+    how this is the case in the right subtrees above.
```
petrelharp, line 204: Gee, also, I think you had a good description of what the order on labellings is, before? It'd be nice to put that back in.

petrelharp:
Well, it's still missing a lot of the interesting technical detail that says exactly what the ordering is - what's your reasoning for leaving that out?

docs/combinatorics.rst
jeromekelleher: A bijection would give you both - you could move back and forth between the two representations. I can't see anyone being particularly interested in the actual rank values beyond defining the actual tree. If this rank was also its index in ``all_trees()``, then so much... neater.

jeromekelleher:
Looks great @daniel-goldstein, let's go through it later.

jeromekelleher:
OK, let's get this merged. We can follow up with more detail later, but this is enough to give some information for now.

PR #654, opened 2020-05-27
removed stray print

petrelharp:
This printed something every time an array was decoded; whoops? ping @benjeffery when you return to make sure it wasn't reminding you of something?


benjeffery:
Thanks for catching that.

PR #623, opened 2020-05-18
Union of tree sequences

mufernando:
If we have two tree sequences which share part of their past histories, we might want to graft them together (also see #381). One use case could be that one population split into two and you might want to simulate their branches independently as a way of parallelising the simulations.

Here, @petrelharp and I implemented a method to `graft` two tables, a **base** table and its **sister** (details of dev process can be found in [this repo](https://github.com/mufernando/graft)). The main rationale is:

- First, identify which nodes are shared between tables, and encode them in a `node_map`;
- If the tree sequences for a different number of generations (e.g., in a forward simulator) , it is possible the times along the nodes in the `node_map` don't match. For that reason, we also implemented a method to `add_time` to the node times and migration times of a Table Collection;
- Now, we can copy the **base** as a **new** table and start adding the non-shared bits of the **sister** table to it;
- All the nodes of the **sister** table not in the `node_map` are added to **new** -- these are the target nodes;
    - These nodes are assigned new individuals;
    - The population of all target nodes are considered new, and are added to **new** population table;
- Edges connecting target nodes are added to **new**;
- Mutations falling in any of the target nodes are also added;
    - At first pass all these mutations are assigned new sites, which are later deduplicated;
    - We do not assign a parent mutation, but late recompute them;
- Migrations within **base** are kept, and migrations within **sister** are added, with nodes and populations being remapped;
- A new row is added to the provenance table -- the way it's currently done is NOT final;
- At last, we can sort, deduplicate sites, and compute the parent mutations;

We implemented some tests, which rely on simplification. The idea is that if you simulate a history of one population splitting into two, simplify independently on the nodes of population 0 and 1, and graft them together you should get equivalent TreeSequences. 

The implementation is not finalised, this is a slow -- but likely correct -- solution to the problem. 

We would like any input on implementation/interface, but most importantly there are two points of improvement:

- We cannot currently test how the MigrationTable is grafted, because our tests rely on simplification and MigrationTables are not dealt with by `simplify` (see [here](https://github.com/tskit-dev/tskit/issues/20))
- We need help figuring out how to deal with the ProvenanceTable of the grafted tree sequence.

PR #663, opened 2020-06-01
method to subset and reorder table collection

mufernando:
Here I implemented a numpy version of a method to subset a table collection.

There are probably many uses for this, but most importantly:

- If you subset on a permutation of all existing nodes, you can reorder the Nodes, Individuals, and Populations Tables to your liking. I think @hyanwong was interested in this use?
- For the new `graft` method we need a better way of checking for equivalency between Table Collections, and now we can quickly do so by subsetting.
    - We now can get the non-overlapping bits of two tree sequences using this method, so it could be used to rethink the way `graft` works.

@petrelharp described the motivation and approach [here](https://github.com/tskit-dev/tskit/pull/623#issuecomment-631655756):

The proposed subset-and-reorder method is: tables.subset_nodes(nodes) modifies the tables to retain exactly those:

- nodes that are listed in nodes and in the order listed
- mutations whose node entry is in nodes
- sites with remaining mutations
- edges whose parent and child entries are in nodes
- individuals referred to by the nodes in nodes, and in the order encountered in nodes
- only populations referred to by nodes in nodes, and in the order encountered in nodes
- migrations whose nodes are in nodes

I also implemented a new function to `subset_ragged_col` that subsets ragged cols based on a list of indices (or boolean mask). I added it to `util.py` but not sure that is the right place.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7799,6 +7799,164 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
     return tsk_table_collection_truncate(self, &start);
 }
 
+int TSK_WARN_UNUSED
+tsk_table_collection_subset(
+    tsk_table_collection_t *self, tsk_id_t *nodes, tsk_size_t num_nodes)
+{
+    int ret = 0;
+    tsk_id_t k, i, new_ind, new_pop, new_parent, new_child, new_node;
+    tsk_node_t node;
+    tsk_individual_t ind;
+    tsk_population_t pop;
+    tsk_edge_t edge;
+    tsk_mutation_t mut;
+    tsk_site_t site;
+    tsk_id_t *node_map = NULL;
+    tsk_id_t *individual_map = NULL;
+    tsk_id_t *population_map = NULL;
+    tsk_id_t *site_map = NULL;
+    tsk_id_t *mutation_map = NULL;
+    tsk_table_collection_t tables;
+
+    ret = tsk_table_collection_copy(self, &tables, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_table_collection_clear(self);
+    if (ret != 0) {
+        goto out;
+    }
+
+    node_map = malloc(tables.nodes.num_rows * sizeof(*node_map));
+    individual_map = malloc(tables.individuals.num_rows * sizeof(*individual_map));
+    population_map = malloc(tables.populations.num_rows * sizeof(*population_map));
+    site_map = malloc(tables.sites.num_rows * sizeof(*site_map));
+    mutation_map = malloc(tables.mutations.num_rows * sizeof(*mutation_map));
+    if (node_map == NULL || individual_map == NULL || population_map == NULL
+        || site_map == NULL || mutation_map == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    memset(node_map, 0xff, tables.nodes.num_rows * sizeof(*node_map));
+    memset(individual_map, 0xff, tables.individuals.num_rows * sizeof(*individual_map));
+    memset(population_map, 0xff, tables.populations.num_rows * sizeof(*population_map));
+    memset(site_map, 0xff, tables.sites.num_rows * sizeof(*site_map));
+    memset(mutation_map, 0xff, tables.mutations.num_rows * sizeof(*mutation_map));
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
+        ret = tsk_node_table_get_row(&tables.nodes, nodes[k], &node);
+        if (ret < 0) {
+            goto out;
+        }
+        new_ind = TSK_NULL;
+        if (node.individual != TSK_NULL) {
+            if (individual_map[node.individual] == TSK_NULL) {
+                tsk_individual_table_get_row(&tables.individuals, node.individual, &ind);
+                ret = tsk_individual_table_add_row(&self->individuals, ind.flags,
+                    ind.location, ind.location_length, ind.metadata,
+                    ind.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+                individual_map[node.individual] = ret;
+            }
+            new_ind = individual_map[node.individual];
+        }
+        new_pop = TSK_NULL;
+        if (node.population != TSK_NULL) {
+            if (population_map[node.population] == TSK_NULL) {
+                tsk_population_table_get_row(&tables.populations, node.population, &pop);
+                ret = tsk_population_table_add_row(
+                    &self->populations, pop.metadata, pop.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+                population_map[node.population] = ret;
+            }
+            new_pop = population_map[node.population];
+        }
+        ret = tsk_node_table_add_row(&self->nodes, node.flags, node.time, new_pop,
+            new_ind, node.metadata, node.metadata_length);
+        if (ret < 0) {
+            goto out;
+        }
+        node_map[node.id] = ret;
+    }
+
+    // edges
+    for (k = 0; k < (tsk_id_t) tables.edges.num_rows; k++) {
+        tsk_edge_table_get_row(&tables.edges, k, &edge);
+        new_parent = node_map[edge.parent];
+        new_child = node_map[edge.child];
+        if ((new_parent != TSK_NULL) && (new_child != TSK_NULL)) {
+            ret = tsk_edge_table_add_row(&self->edges, edge.left, edge.right, new_parent,
+                new_child, edge.metadata, edge.metadata_length);
+            if (ret < 0) {
+                goto out;
+            }
+        }
+    }
+
+    // mutations and sites
+    i = 0;
+    for (k = 0; k < (tsk_id_t) tables.sites.num_rows; k++) {
+        tsk_site_table_get_row(&tables.sites, k, &site);
+        while ((i < (tsk_id_t) tables.mutations.num_rows)
+               && (tables.mutations.site[i] == site.id)) {
+            tsk_mutation_table_get_row(&tables.mutations, i, &mut);
+            new_node = node_map[mut.node];
+            if (new_node != TSK_NULL) {
+                if (site_map[site.id] == TSK_NULL) {
+                    ret = tsk_site_table_add_row(&self->sites, site.position,
+                        site.ancestral_state, site.ancestral_state_length, site.metadata,
+                        site.metadata_length);
+                    if (ret < 0) {
+                        goto out;
+                    }
+                    site_map[site.id] = ret;
+                }
+                new_parent = TSK_NULL;
+                if (mut.parent != TSK_NULL) {
+                    new_parent = mutation_map[mut.parent];
+                }
+                ret = tsk_mutation_table_add_row(&self->mutations, site_map[site.id],
+                    new_node, new_parent, mut.derived_state, mut.derived_state_length,
+                    mut.metadata, mut.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+                mutation_map[mut.id] = ret;
+            }
+            i++;
+        }
+    }
+
+    /* TODO: Subset of the Migrations Table. The way to do this properly is not
+     * well-defined, mostly because migrations might contain events from/to
+     * populations that have not been kept in after the subset. */
+    if (tables.migrations.num_rows != 0) {
+        ret = TSK_ERR_MIGRATIONS_NOT_SUPPORTED;
+        goto out;
+    }
+
+    // provenance (new record is added in python)
+    ret = tsk_provenance_table_copy(
+        &tables.provenances, &self->provenances, TSK_NO_INIT);
+    if (ret < 0) {
+        goto out;
+    }
+
+out:
+    tsk_safe_free(node_map);
+    tsk_safe_free(individual_map);
+    tsk_safe_free(population_map);
+    tsk_safe_free(site_map);
+    tsk_safe_free(mutation_map);
+    tsk_table_collection_free(&tables);
+    return ret;
+}
+
 static int
 cmp_edge_cl(const void *a, const void *b)
 {
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -2410,6 +2410,43 @@ int tsk_table_collection_simplify(tsk_table_collection_t *self, tsk_id_t *sample
     tsk_size_t num_samples, tsk_flags_t options, tsk_id_t *node_map);
 
 /**
+@brief Subsets and reorders a table collection according to an array of nodes.
+
+@rst
+Reduces the table collection to contain only the entries referring to
+the provided list of nodes, with nodes reordered according to the order
+they appear in the ``nodes`` argument. Specifically, this subsets and reorders
+each of the tables as follows:
+
+1. Nodes: if in the list of nodes, and in the order provided.
+2. Individuals and Populations: if referred to by a retained node,
+   and in the order first seen when traversing the list of retained nodes.
+3. Edges: if both parent and child are retained nodes.
+4. Mutations: if the mutation's node is a retained node.
+5. Sites: if any mutations remain at the site after removing mutations.
+6. Migrations: if the migration's node is a retained node.
+
+Retained edges, mutations, sites, and migrations appear in the same
+order as in the original tables.
+
+If ``nodes`` is the entire list of nodes in the tables, then the
+resulting tables will be identical to the original tables, but with
+nodes (and individuals and populations) reordered.
+
+.. note:: Migrations are currently not supported by susbset, and an error will
+    be raised if we attempt call subset on a table collection with greater
+    than zero migrations.
+@endrst
+
+@param self A pointer to a tsk_table_collection_t object.
+@param nodes An array of num_nodes valid node IDs.
+@param num_nodes The number of node IDs in the input nodes array.
+@return Return 0 on success or a negative value on failure.
+*/
+int tsk_table_collection_subset(
+    tsk_table_collection_t *self, tsk_id_t *nodes, tsk_size_t num_nodes);
+
+/**
 @brief Set the metadata
 @rst
 Copies the metadata string to this table collection, replacing any existing.
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -6257,6 +6257,38 @@ out:
 }
 
 static PyObject *
+TableCollection_subset(TableCollection *self, PyObject *args)
+{
+    int err;
+    PyObject *ret = NULL;
+    PyObject *nodes = NULL;
+    PyArrayObject *nodes_array = NULL;
+    npy_intp *shape;
+    size_t num_nodes;
+
+    if (!PyArg_ParseTuple(args, "O", &nodes)) {
+        goto out;
+    }
+    nodes_array = (PyArrayObject *) PyArray_FROMANY(nodes, NPY_INT32, 1, 1,
+            NPY_ARRAY_IN_ARRAY);
+    if (nodes_array == NULL) {
+        goto out;
+    }
+    shape = PyArray_DIMS(nodes_array);
+    num_nodes = shape[0];
+
+    err = tsk_table_collection_subset(self->tables, PyArray_DATA(nodes_array), num_nodes);
+    if (err != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = Py_BuildValue("");
+out:
+    Py_XDECREF(nodes_array);
+    return ret;
+}
+
+static PyObject *
 TableCollection_sort(TableCollection *self, PyObject *args, PyObject *kwds)
 {
     int err;
@@ -6397,6 +6429,8 @@ static PyMethodDef TableCollection_methods[] = {
     {"link_ancestors", (PyCFunction) TableCollection_link_ancestors,
         METH_VARARGS|METH_KEYWORDS,
         "Returns an edge table linking samples to a set of specified ancestors." },
+    {"subset", (PyCFunction) TableCollection_subset, METH_VARARGS,
+        "Subsets the tree sequence to a set of nodes." },
     {"sort", (PyCFunction) TableCollection_sort, METH_VARARGS|METH_KEYWORDS,
         "Sorts the tables to satisfy tree sequence requirements." },
     {"equals", (PyCFunction) TableCollection_equals, METH_VARARGS,
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -265,6 +265,18 @@ class TestTableCollection(LowLevelTestCase):
         del edges
         self.assertEqual(tc.edges.num_rows, 2)
 
+    def test_subset_bad_args(self):
+        ts = msprime.simulate(10, random_seed=1)
+        tc = ts.tables.ll_tables
+        with self.assertRaises(TypeError):
+            tc.subset(np.array(["a"]))
+        with self.assertRaises(ValueError):
+            tc.subset(np.array([[1], [2]], dtype="int32"))
+        with self.assertRaises(TypeError):
+            tc.subset()
+        with self.assertRaises(_tskit.LibraryError):
+            tc.subset(np.array([100, 200], dtype="int32"))
+
 
 class TestTreeSequence(LowLevelTestCase, MetadataTestMixin):
     """
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -36,6 +36,7 @@ import msprime
 import numpy as np
 
 import _tskit
+import tests.test_wright_fisher as wf
 import tests.tsutil as tsutil
 import tskit
 import tskit.exceptions as exceptions
@@ -2363,3 +2364,168 @@ class TestBaseTable(unittest.TestCase):
         t = tskit.BaseTable(None, None)
         with self.assertRaises(NotImplementedError):
             t.set_columns()
+
+
+class TestSubsetTables(unittest.TestCase):
+    """
+    Tests for the TableCollection.subset method.
+    """
+
+    def get_msprime_example(self, sample_size=10, seed=1234):
+        M = [[0.0, 0.1], [1.0, 0.0]]
+        population_configurations = [
+            msprime.PopulationConfiguration(sample_size=sample_size),
+            msprime.PopulationConfiguration(sample_size=sample_size),
+        ]
+        ts = msprime.simulate(
+            population_configurations=population_configurations,
+            migration_matrix=M,
+            length=2e5,
+            recombination_rate=1e-8,
+            mutation_rate=1e-7,
+            record_migrations=False,
+            random_seed=seed,
+        )
+        # adding metadata and locations
+        ts = tsutil.add_random_metadata(ts, seed)
+        ts = tsutil.insert_random_ploidy_individuals(ts, max_ploidy=1)
+        return ts.tables
+
+    def get_wf_example(self, N=5, ngens=2, seed=1249):
+        tables = wf.wf_sim(N, N, seed=seed)
+        tables.sort()
+        ts = tables.tree_sequence()
+        # adding muts
+        ts = tsutil.jukes_cantor(ts, 1, 10, seed=seed)
+        ts = tsutil.add_random_metadata(ts, seed)
+        ts = tsutil.insert_random_ploidy_individuals(ts, max_ploidy=2)
+        return ts.tables
+
+    def get_examples(self, seed):
+        yield self.get_msprime_example(seed=seed)
+        yield self.get_wf_example(seed=seed)
+
+    def verify_subset_equality(self, tables, nodes):
+        sub1 = tables.copy()
+        sub2 = tables.copy()
+        tsutil.py_subset(sub1, nodes, record_provenance=False)
+        sub2.subset(nodes, record_provenance=False)
+        self.assertEqual(sub1, sub2)
+
+    def verify_subset(self, tables, nodes):
+        self.verify_subset_equality(tables, nodes)
+        subset = tables.copy()
+        subset.subset(nodes, record_provenance=False)
+        # adding one so the last element always maps to NULL (-1 -> -1)
+        node_map = np.repeat(tskit.NULL, tables.nodes.num_rows + 1)
+        indivs = []
+        pops = []
+        for k, n in enumerate(nodes):
+            node_map[n] = k
+            ind = tables.nodes[n].individual
+            pop = tables.nodes[n].population
+            if ind not in indivs and ind != tskit.NULL:
+                indivs.append(ind)
+            if pop not in pops and pop != tskit.NULL:
+                pops.append(pop)
+        ind_map = np.repeat(tskit.NULL, tables.individuals.num_rows + 1)
+        ind_map[indivs] = np.arange(len(indivs), dtype="int32")
+        pop_map = np.repeat(tskit.NULL, tables.populations.num_rows + 1)
+        pop_map[pops] = np.arange(len(pops), dtype="int32")
+        self.assertEqual(subset.nodes.num_rows, len(nodes))
+        for k, n in zip(nodes, subset.nodes):
+            nn = tables.nodes[k]
+            self.assertEqual(nn.time, n.time)
+            self.assertEqual(nn.flags, n.flags)
+            self.assertEqual(nn.metadata, n.metadata)
+            self.assertEqual(ind_map[nn.individual], n.individual)
+            self.assertEqual(pop_map[nn.population], n.population)
+        self.assertEqual(subset.individuals.num_rows, len(indivs))
+        for k, i in zip(indivs, subset.individuals):
+            ii = tables.individuals[k]
+            self.assertEqual(ii, i)
+        self.assertEqual(subset.populations.num_rows, len(pops))
+        for k, p in zip(pops, subset.populations):
+            pp = tables.populations[k]
+            self.assertEqual(pp, p)
+        edges = [
+            i
+            for i, e in enumerate(tables.edges)
+            if e.parent in nodes and e.child in nodes
+        ]
+        self.assertEqual(subset.edges.num_rows, len(edges))
+        for k, e in zip(edges, subset.edges):
+            ee = tables.edges[k]
+            self.assertEqual(ee.left, e.left)
+            self.assertEqual(ee.right, e.right)
+            self.assertEqual(node_map[ee.parent], e.parent)
+            self.assertEqual(node_map[ee.child], e.child)
+            self.assertEqual(ee.metadata, e.metadata)
+        muts = []
+        sites = []
+        for k, m in enumerate(tables.mutations):
+            if m.node in nodes:
+                muts.append(k)
+                if m.site not in sites:
+                    sites.append(m.site)
+        site_map = np.repeat(-1, tables.sites.num_rows)
+        site_map[sites] = np.arange(len(sites), dtype="int32")
+        mutation_map = np.repeat(tskit.NULL, tables.mutations.num_rows + 1)
+        mutation_map[muts] = np.arange(len(muts), dtype="int32")
+        self.assertEqual(subset.sites.num_rows, len(sites))
+        for k, s in zip(sites, subset.sites):
+            ss = tables.sites[k]
+            self.assertEqual(ss, s)
+        self.assertEqual(subset.mutations.num_rows, len(muts))
+        for k, m in zip(muts, subset.mutations):
+            mm = tables.mutations[k]
+            self.assertEqual(mutation_map[mm.parent], m.parent)
+            self.assertEqual(site_map[mm.site], m.site)
+            self.assertEqual(node_map[mm.node], m.node)
+            self.assertEqual(mm.derived_state, m.derived_state)
+            self.assertEqual(mm.metadata, m.metadata)
+        self.assertEqual(tables.migrations, subset.migrations)
+        self.assertEqual(tables.provenances, subset.provenances)
+
+    def test_ts_subset(self):
+        nodes = np.array([0, 1])
+        for tables in self.get_examples(83592):
+            ts = tables.tree_sequence()
+            tables2 = ts.subset(nodes, record_provenance=False).dump_tables()
+            tables.subset(nodes, record_provenance=False)
+            self.assertEqual(tables, tables2)
+
+    def test_subset_all(self):
+        # subsetting to everything shouldn't change things
+        # except the individual ids in the node tables if
+        # there are gaps
+        for tables in self.get_examples(123583):
+            tables2 = tables.copy()
+            tables2.subset(np.arange(tables.nodes.num_rows))
+            tables.provenances.clear()
+            tables2.provenances.clear()
+            tables.individuals.clear()
+            tables2.individuals.clear()
+            tables.nodes.clear()
+            tables2.nodes.clear()
+            self.assertEqual(tables, tables2)
+
+    def test_random_subsets(self):
+        rng = np.random.default_rng(1542)
+        for tables in self.get_examples(9412):
+            for n in [2, tables.nodes.num_rows - 10]:
+                nodes = rng.choice(np.arange(tables.nodes.num_rows), n, replace=False)
+                self.verify_subset(tables, nodes)
+
+    def test_empty_nodes(self):
+        for tables in self.get_examples(8724):
+            subset = tables.copy()
+            subset.subset(np.array([]), record_provenance=False)
+            self.assertEqual(subset.nodes.num_rows, 0)
+            self.assertEqual(subset.edges.num_rows, 0)
+            self.assertEqual(subset.populations.num_rows, 0)
+            self.assertEqual(subset.individuals.num_rows, 0)
+            self.assertEqual(subset.migrations.num_rows, 0)
+            self.assertEqual(subset.sites.num_rows, 0)
+            self.assertEqual(subset.mutations.num_rows, 0)
+            self.assertEqual(subset.provenances, tables.provenances)
--- a/python/tests/tsutil.py
+++ b/python/tests/tsutil.py
@@ -539,6 +539,76 @@ def compute_mutation_parent(ts):
     return mutation_parent
 
 
+def py_subset(tables, nodes, record_provenance=True):
+    """
+    Naive implementation of the TableCollection.subset method using the Python API.
+    """
+    if np.any(nodes > tables.nodes.num_rows) or np.any(nodes < 0):
+        raise ValueError("Nodes out of bounds.")
+    full = tables.copy()
+    # there is no table collection clear in the py API
+    tables.nodes.clear()
+    tables.individuals.clear()
+    tables.populations.clear()
+    tables.edges.clear()
+    tables.migrations.clear()
+    tables.sites.clear()
+    tables.mutations.clear()
+    # mapping from old to new ids
+    node_map = {}
+    ind_map = {tskit.NULL: tskit.NULL}
+    pop_map = {tskit.NULL: tskit.NULL}
+    for old_id in nodes:
+        node = full.nodes[old_id]
+        if node.individual not in ind_map and node.individual != tskit.NULL:
+            ind = full.individuals[node.individual]
+            new_ind_id = tables.individuals.add_row(
+                ind.flags, ind.location, ind.metadata
+            )
+            ind_map[node.individual] = new_ind_id
+        if node.population not in pop_map and node.population != tskit.NULL:
+            pop = full.populations[node.population]
+            new_pop_id = tables.populations.add_row(pop.metadata)
+            pop_map[node.population] = new_pop_id
+        new_id = tables.nodes.add_row(
+            node.flags,
+            node.time,
+            pop_map[node.population],
+            ind_map[node.individual],
+            node.metadata,
+        )
+        node_map[old_id] = new_id
+    for edge in full.edges:
+        if edge.child in nodes and edge.parent in nodes:
+            tables.edges.add_row(
+                edge.left,
+                edge.right,
+                node_map[edge.parent],
+                node_map[edge.child],
+                edge.metadata,
+            )
+    if full.migrations.num_rows > 0:
+        raise ValueError("Migrations are currently not supported in this operation.")
+    site_map = {}
+    mutation_map = {tskit.NULL: tskit.NULL}
+    for i, mut in enumerate(full.mutations):
+        if mut.node in nodes:
+            if mut.site not in site_map:
+                site = full.sites[mut.site]
+                new_site = tables.sites.add_row(
+                    site.position, site.ancestral_state, site.metadata
+                )
+                site_map[mut.site] = new_site
+            new_mut = tables.mutations.add_row(
+                site_map[mut.site],
+                node_map[mut.node],
+                mut.derived_state,
+                mutation_map.get(mut.parent, tskit.NULL),
+                mut.metadata,
+            )
+            mutation_map[i] = new_mut
+
+
 def algorithm_T(ts):
     """
     Simple implementation of algorithm T from the PLOS paper, taking into
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2494,3 +2494,23 @@ class TableCollection:
         currently indexed this method has no effect.
         """
         self.ll_tables.drop_index()
+
+    def subset(self, nodes, record_provenance=True):
+        """
+        Modifies the tables in place to contain only the entries referring to
+        the provided list of nodes, with nodes reordered according to the order
+        they appear in the list. See :meth:`TreeSequence.subset` for a more
+        detailed description.
+
+        :param list nodes: The list of nodes for which to retain information. This
+            may be a numpy array (or array-like) object (dtype=np.int32).
+        :param bool record_provenance: Whether to record a provenance entry
+            in the provenance table for this operation.
+        """
+        nodes = util.safe_np_int_cast(nodes, np.int32)
+        self.ll_tables.subset(nodes)
+        if record_provenance:
+            parameters = {"command": "subset", "nodes": nodes.tolist()}
+            self.provenances.add_row(
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -4470,6 +4470,45 @@ class TreeSequence:
         tables.trim(record_provenance)
         return tables.tree_sequence()
 
+    def subset(self, nodes, record_provenance=True):
+        """
+        Returns a tree sequence modified to contain only the entries referring to
+        the provided list of nodes, with nodes reordered according to the order
+        they appear in the ``nodes`` argument. Specifically, this subsets and reorders
+        each of the tables as follows:
+
+        1. Nodes: if in the list of nodes, and in the order provided.
+        2. Individuals and Populations: if referred to by a retained node,
+           and in the order first seen when traversing the list of retained nodes.
+        3. Edges: if both parent and child are retained nodes.
+        4. Mutations: if the mutation's node is a retained node.
+        5. Sites: if any mutations remain at the site after removing mutations.
+        6. Migrations: if the migration's node is a retained node.
+
+        Retained edges, mutations, sites, and migrations appear in the same
+        order as in the original tables.
+
+        If ``nodes`` is the entire list of nodes in the tables, then the
+        resulting tables will be identical to the original tables, but with
+        nodes (and individuals and populations) reordered.
+
+        To instead subset the tables to a given portion of the *genome*, see
+        :meth:`.keep_intervals`.
+
+        **Note:** This is quite different from :meth:`.simplify`: the resulting
+        tables contain only the nodes given, not ancestral ones as well, and
+        does not simplify the relationships in any way.
+
+        :param list nodes: The list of nodes for which to retain information. This
+            may be a numpy array (or array-like) object (dtype=np.int32).
+        :param bool record_provenance: If True, add details of this operation to the
+            provenance information of the returned tree sequence. (Default: True).
+        :rtype: .TreeSequence
+        """
+        tables = self.dump_tables()
+        tables.subset(nodes, record_provenance)
+        return tables.tree_sequence()
+
     def draw_svg(
         self,
         path=None,
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -3212,6 +3212,147 @@ test_table_collection_check_integrity(void)
     tsk_table_collection_free(&tables);
 }
 
+static void
+test_table_collection_subset(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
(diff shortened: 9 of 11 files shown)


petrelharp:
To other viewing this issue - we've got a fair bit of work to tidy this up, so comments appreciated especially on the big picture, but we know that it's not ready to go yet.

Maybe the biggest question in my head is what to call this method? It has two main use cases, really: subsetting, and *reordering*, so maybe the name should reflect reordering also.

petrelharp:
To avoid confusion: right now, there are two implementations, one (`.subset()`) in numpy, the other (`.subset_nodes()`) in C. They should be equivalent (except for bugs, of which there are some currently), and it will be interesting to compare them.

python/tests/test_tables.py
```diff
@@ -2272,32 +2272,30 @@ class TestSubsetTables(unittest.TestCase):
...
-            mutation_rate=1e-8,
-            record_migrations=True,
-            random_seed=self.seed,
-        )
-        if mutations_per_branch > 0:
-            ts = tsutil.insert_branch_mutations(ts, mutations_per_branch)
-        return ts
-
-    def verify_subset_equality(self, tables, subset, nodes):
+    def verify_subset_equality(self, tables, nodes):
```
petrelharp, line 2275: this is verifying both implementations agree; probably we will stick with the C version, and will put the numpy version in here, to compare against.

python/tests/test_tables.py
```diff
@@ -2272,32 +2272,30 @@ class TestSubsetTables(unittest.TestCase):
...
+        sub1 = tables.copy()
+        sub2 = tables.copy()
+        sub1.subset(nodes)
+        sub2.subset_nodes(nodes)
+        sub1.provenances.clear()
+        sub2.provenances.clear()
+        for x, y in [
```
petrelharp, line 2282: gee but they don't agree right now, this is debugging code

python/tskit/tables.py
```diff
@@ -2498,18 +2471,35 @@ def drop_index(self):
...
-        Subsets a TableCollection in place and retains (i) the nodes listed in
-        `nodes` and in the order listed, (ii) only mutations whose node entry
-        are in `nodes`, (iii) the sites of the remaining mutations, (iv) the
-        edges whose parent and child are in `nodes`, (v) individuals referred
-        to by the nodes in `nodes` and in the order listed, (vi) only
-        populations referred to by the nodes in `nodes` and in the order
-        listed, and (vii) the migrations whose nodes are in `nodes`. Note that
-        if `nodes` is a permutation of all nodes in the TableCollection, then
-        the subset operation simply reorders the Nodes, Individuals, and
-        Population Tables in the order encountered in `nodes`.
```
petrelharp: see if you like this documentation?

jeromekelleher:
The main big-picture question for me would be, doesn't simplify do this already? We started out calling simplify ``subset``, and that was its first application before we thought about using it for forwards simulation.

hyanwong:
> The main big-picture question for me would be, doesn't simplify do this already? We started out calling simplify `subset`, and that was its first application before we thought about using it for forwards simulation.

A good point. Just a quick thought: simplify won't reorder internal nodes arbitrarily (but I'm not sure if the new method will do that either). Also simplify doesn't reorder populations / individuals AFAIK, but I guess it could be extended to do that.

petrelharp:
> The main big-picture question for me would be, doesn't simplify do this already?

Simplify does a lot more.  As Yan points out, simplify only lets you reorder samples; but much of the structure above the samples might be changed (nodes removed, edges squashed, etc).  I've put a reminder of this [in the docs](https://github.com/tskit-dev/tskit/pull/663/files#diff-a89cffb041b1b7c33e7be6b595f785a3R2498); should I make this confusing point even more explicit?  If you passed in just the list of sample nodes to subset in an msprime tree, you'd get back the trivial forest where no-one is related to anyone else. Maybe putting a note like this into the docs would help clarify:
```
Note that *only* nodes listed are included: so if full trees are to be retained in the tree sequence, all ancestral node IDs must be included in ``nodes``, not just the samples.
```

jeromekelleher:
Ah, that makes sense now. Sure, seems like a useful operation then.

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        (and individuals and populations) reordered.
+
+        **Note:** This is quite different from :meth:`.simplify`: the resulting
+        tables contain only the nodes given, not ancestral ones as well, and
+        does not simplify the relationships in any way.
+
+        :param list[int] nodes: A list of node IDs to retain in the tables.
+        """
+        if nodes.shape[0] == 0:
+            raise ValueError("Nodes cannot be empty.")
```
petrelharp, line 2505: Actually, I think it should return an empty tree sequence? (and, the code below should be adjusted to do this if nodes is empty, if it doesn't already).

jeromekelleher:
Ping me when you'd like detailed feedback - I'll mute this thread for now.

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        if np.max(nodes) >= self.nodes.num_rows:
+            raise ValueError("One of the nodes is not in the TableCollection.")
```
petrelharp, line 2507: I think the way we usually say this is something like "Node out of bounds". Should also check that all nodes are >= 0?

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        tables = self.copy()
+        n = tables.nodes
+        # figuring out which individuals to keep
+        old_inds = n.individual[nodes]
+        indiv_to_keep, i = np.unique(old_inds, return_index=True)
+        # maintaining order in which they appeard
+        indiv_to_keep = indiv_to_keep[np.argsort(i)]
```
petrelharp, line 2514: very nice

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        tables contain only the nodes given, not ancestral ones as well, and
+        does not simplify the relationships in any way.
+
+        :param list[int] nodes: A list of node IDs to retain in the tables.
+        """
+        if nodes.shape[0] == 0:
+            raise ValueError("Nodes cannot be empty.")
+        if np.max(nodes) >= self.nodes.num_rows:
+            raise ValueError("One of the nodes is not in the TableCollection.")
+        tables = self.copy()
```
petrelharp, line 2508: I think we should do `self.clear()` at this point.

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        n = tables.nodes
+        # figuring out which individuals to keep
+        old_inds = n.individual[nodes]
+        indiv_to_keep, i = np.unique(old_inds, return_index=True)
+        # maintaining order in which they appeard
+        indiv_to_keep = indiv_to_keep[np.argsort(i)]
+        # removing -1
+        indiv_to_keep = indiv_to_keep[indiv_to_keep != tskit.NULL]
+        # subsetting individuals table
+        if indiv_to_keep.shape[0] == 0:
```
petrelharp, line 2518: We shouldn't need this special case (esp if we clear the tables above).

mufernando: I think `set_columns` complains about it being empty.

python/tskit/util.py
```diff
@@ -183,6 +183,31 @@ def unpack_arrays(packed, offset):
     return ret
 
 
+def subset_ragged_col(packed, offset, subset, col="metadata"):
```
petrelharp, line 186: This is a nice, elegant way of doing what we need. I wonder if should be public, though? Is it too specific to our use case? If it doesn't need to be public, we could call it `_subset_ragged_col`  (and then we wouldn't have to test it). I am honestly not sure. 

I do think the argument `col` should be called `type`, and the method should be called `subset_ragged_column`.

petrelharp, line 186: This is a nice, elegant way of doing what we need! But, should it be public? Or is it too specific to our use case? I honestly don't know. It's a pretty thin wrapper around pack/unpack, but maybe a useful enough one? The return type is a bit odd. Hm - what if it returned a tuple, instead, wouldn't that work? So then we could do e.g. 
```
            self.individuals.set_columns(
                i.flags[indiv_to_keep],
                *util.subset_ragged_col(
                    i.location, i.location_offset, indiv_to_keep, "location"
                ),
                *util.subset_ragged_col(i.metadata, i.metadata_offset, indiv_to_keep),
            )
```
It would mean we couldn't use keywords, which are nice, but the return value would be more pythonic?  If we don't make this public (so, call it `_subset_ragged_col`) then we wouldn't have to test it.

In any case, I think the argument `col` should be called `name`, and the method should be `subset_ragged_column` (ie, with 'column' spelled out).

petrelharp: Oh, and I think we should not provide a default to `col` - there's not a natural default, so the user should specify.

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+        if pop_to_keep.shape[0] == 0:
+            self.populations.clear()
+        else:
+            p = tables.populations
+            self.populations.set_columns(
+                **util.subset_ragged_col(p.metadata, p.metadata_offset, pop_to_keep)
+            )
+        # mapping of ind/pop id in full tables to subset tables
+        ind_map = {ind: i for i, ind in enumerate(indiv_to_keep)}
+        pop_map = {pop: j for j, pop in enumerate(pop_to_keep)}
```
petrelharp, line 2544: perhaps set `ind_map[tskit.NULL] = tskit.NULL`? (also, you should use NULL, not -1)

petrelharp: and, with that in, I wonder if `np.fromiter(map(ind_map.get, old_inds))` would be more efficient? not quite as clear, though.

petrelharp: Or, wouldn't it be simpler to use the same strategy as `node_map` below, with an array instead of a dict?

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+            flags=n.flags[nodes],
+            population=new_pops,
+            individual=new_inds,
+            time=n.time[nodes],
+            **util.subset_ragged_col(n.metadata, n.metadata_offset, nodes),
+        )
+        # mapping node ids in full to subsetted table
+        # making the node map +1 bc last will be mapping -1 to -1
+        node_map = np.repeat(tskit.NULL, tables.nodes.num_rows + 1)
+        node_map[-1] = tskit.NULL
```
petrelharp, line 2559: this line is unnecessary

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+                    "ancestral_state",
+                ),
+                **util.subset_ragged_col(s.metadata, s.metadata_offset, keep_sites),
+            )
+            site_map = np.repeat(-1, tables.sites.num_rows)
+            site_map[keep_sites] = np.arange(self.sites.num_rows)
+            site_map = site_map.astype("int32")
+            mutation_map = np.repeat(tskit.NULL, tables.mutations.num_rows)
+            mutation_map[keep_muts] = np.arange(np.count_nonzero(keep_muts))
+            mutation_map = mutation_map.astype("int32")
```
petrelharp, line 2623: you should be able to put the dtype in the initial line (np.repeat)?

python/tskit/tables.py
```diff
@@ -2470,3 +2470,172 @@ def drop_index(self):
...
+            # adding tskit.NULL to the end to map -1 -> -1
+            mutation_map = np.concatenate(
+                (mutation_map, np.array([tskit.NULL], dtype="int32"))
+            )
```
petrelharp, line 2627: do like you did above - just initialize the map to be one longer

python/tskit/tables.py
mufernando: right, we don't need that

python/tskit/tables.py
mufernando: that is not a method of TableCollection. also, that is redundant because we use `set_columns` for all tables.

python/tskit/tables.py
mufernando: no, because the individual and population tables could be empty, in which case numpy version would fail

mufernando: also, according to [this](https://stackoverflow.com/questions/35215161/most-efficient-way-to-map-function-over-numpy-array/35216364) not clear if you gain much from using `np.fromiter(map)` compared to a comprehension.

python/tskit/util.py
mufernando: changed the function and argument names. the problem with returning a tuple is that `set_columns` only takes keyword arguments.

petrelharp: If you put the arguments in the right order, you don't need to specify the keywords?

mufernando:
> I don't think the `subset_ragged_columns` is working as it should because it is not reordering.

fixed this and added metadata to the tests

mufernando:
Speed comparisons for the C and Numpy versions:

Here for one tree of 4M nodes, I tested the speed for different sizes of subset:
Screen Shot 2020-06-03 at 2 23 52 PM

Here for trees of varying sizes, I tested the speed to get a same sized subset (40 nodes):
Screen Shot 2020-06-03 at 2 24 27 PM

mufernando:
add a test with random locations

python/tests/test_tables.py
```diff
@@ -2272,19 +2272,231 @@ class TestSubsetTables(unittest.TestCase):
...
+            pack = tskit.pack_arrays
+        else:
+            unpack = tskit.unpack_bytes
+            pack = tskit.pack_bytes
+        unpacked = unpack(packed, offset)
+        unpacked = [unpacked[i] for i in subset]
+        return pack(unpacked)
+
+    def np_subset(self, tables, nodes, record_provenance=True):
+        """
```
petrelharp, line 2299: We don't want to duplicate documentation. Just leave this at "Numpy version of TableCollection.subset, for testing purposes."

python/tests/test_tables.py
```diff
@@ -2272,19 +2272,231 @@ class TestSubsetTables(unittest.TestCase):
...
+        tables contain only the nodes given, not ancestral ones as well, and
+        does not simplify the relationships in any way.
+
+        :param list[int] nodes: A list of node IDs to retain in the tables.
+        """
+        if nodes.size == 0:
+            raise ValueError("Nodes cannot be empty.")
+        if np.max(nodes) >= tables.nodes.num_rows or np.any(nodes < 0):
+            print(nodes)
+            print(np.all(nodes > 0))
```
petrelharp, line 2328: remove print statements

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+            ind = tables.nodes[n].individual
+            pop = tables.nodes[n].population
+            if ind not in indivs and ind != tskit.NULL:
+                indivs.append(ind)
+            if pop not in pops and pop != tskit.NULL:
+                pops.append(pop)
+        ind_map = np.repeat(tskit.NULL, tables.individuals.num_rows + 1)
+        ind_map[indivs] = np.arange(len(indivs), dtype="int32")
+        pop_map = np.repeat(tskit.NULL, tables.populations.num_rows + 1)
+        pop_map[pops] = np.arange(len(pops), dtype="int32")
```
petrelharp, line 2491: How about instead:
```
        ind_map = np.repeat(tskit.NULL, tables.individuals.num_rows + 1)
        pop_map = np.repeat(tskit.NULL, tables.populations.num_rows + 1)
        num_inds = 0
        num_pops = 0
        for k, n in enumerate(nodes):
            node_map[n] = k
            ind = tables.nodes[n].individual
            pop = tables.nodes[n].population
            if ind not in indivs and ind != tskit.NULL:
                ind_map[ind] = num_inds
                num_inds += 1
            if pop not in pops and pop != tskit.NULL:
                pop_map[pop] = num_pops
                num_pops += 1
```

petrelharp: oh wait never mind, you use pops and indivs below! =)

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+    def test_subset_all(self):
+        # subsetting to everything shouldn't change things
+        ts = self.get_msprime_mig_example(
+            Ne=100, sample_size=10, mutations_per_branch=2
+        )
+        tables = ts.tables
+        tables2 = tables.copy()
+        tables2.subset_nodes(np.arange(ts.num_nodes))
+        # except that you need one extra row in the provenances
+        self.assertEqual(tables.provenances.num_rows + 1, tables2.provenances.num_rows)
```
petrelharp, line 2568: how about do record_prov=False and not need to clear the provenances?

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+
+    def test_subset_permutation(self):
+        # reversing twice should get us back where we started
+        ts = self.get_msprime_mig_example(
+            Ne=100, sample_size=10, mutations_per_branch=2
+        )
+        tables = ts.tables
+        tables2 = tables.copy()
+        tables2.subset_nodes(np.arange(ts.num_nodes - 1, -1, -1))
+        tables.provenances.clear()
```
petrelharp, line 2581: same here?

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+        self.assertEqual(tables, tables2)
+
+    def test_random_subsets(self):
+        np.random.seed(self.seed)
+        for num_muts in [0, 10]:
+            ts = self.get_msprime_mig_example(
+                Ne=100, sample_size=10, mutations_per_branch=num_muts
+            )
+            ts = tsutil.add_random_metadata(ts, self.seed)
+            # adding locations
```
petrelharp, line 2596: don't need this comment, really

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+        tables2.provenances.clear()
+        self.assertEqual(tables, tables2)
+
+    def test_random_subsets(self):
+        np.random.seed(self.seed)
+        for num_muts in [0, 10]:
+            ts = self.get_msprime_mig_example(
+                Ne=100, sample_size=10, mutations_per_branch=num_muts
+            )
+            ts = tsutil.add_random_metadata(ts, self.seed)
```
petrelharp, line 2595: how about put the random metadata and individuals up in get_mig_example?

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,374 @@ def test_set_columns_not_implemented(self):
...
+        5. Sites: if any mutations remain at the site after removing mutations.
+        6. Migrations: if the migration's node is a retained node.
+
+        If ``nodes`` is the entire list of nodes in the tables, then the resulting
+        tables will be identical to the original tables, but with nodes
+        (and individuals and populations) reordered.
+
+        **Note:** This is quite different from :meth:`.simplify`: the resulting
+        tables contain only the nodes given, not ancestral ones as well, and
+        does not simplify the relationships in any way.
```
petrelharp: How about adding:
```
To subset the tables to a portion of the genome, see :meth:`.keep_intervals`.
```

petrelharp:
Looks good to me!  Questions for others:

- How's the `subset( )` name? We might want to subset by other things in the future, but that could be done by adding an argument `individuals=`, for instance.

- Should the numpy version remain in the testing code? It's nice to have the independent implementation, but it's a kinda bulky one, so maybe it doesn't need to be here? It seems OK to me, if slightly more than is necessary.

hyanwong:
> * How's the `subset( )` name? We might want to subset by other things in the future, but that could be done by adding an argument `individuals=`, for instance.

What alternatives are there? E.g. is `reduce` any better? I think it's hard to decide until we have a list of candidates.

> * Should the numpy version remain in the testing code? It's nice to have the independent implementation, but it's a kinda bulky one, so maybe it doesn't need to be here? It seems OK to me, if slightly more than is necessary.

I don't think we care too much about bulking up the testing code, do we? Better to have it *somewhere*, for future ref (it might not be that obvious to dig it up from the history)

petrelharp:
> What alternatives are there? E.g. is reduce any better? I think it's hard to decide until we have a list of candidates.

I'm looking for some. But if "subset" works, no need to look hard! I like subset better than reduce. slice_nodes occurs to me, but doesn't do it for me either. 'reorder' maybe, but then that doesn't connote the subsetting operation.

python/tests/test_tables.py
mufernando: it seems the default in other tskit functions is `record_provenance=True` so I wanted to keep it consistent.

petrelharp: That should dfeinately be the default! I was saying that in this call to subset, you ask it to not record provenance.

mufernando:
I think we are ready for your more detailed comments, @jeromekelleher!

jeromekelleher:
> I think we are ready for your more detailed comments, @jeromekelleher!

Thanks @mufernando! Sorry I haven't had a chance to look at this yet, and am away tomorrow, so it'll probably be next week before I have a look.

petrelharp:
A thought here: we should probably note in the docs that if the nodes are divided into disjoint subsets A and B, then tables.subset(A) and tables.subset(B) don't quite contain everything in the original tables, because they'd miss out edges between A and B.

jeromekelleher:
Sorry, I being so slow to look here - I promise I'll do it tomorrow. In the meantime, here's an example that came up from an msprime user who wanted to simulate populations with different recombination maps:
```python
L = 1000
T = 1
map0 = msprime.RecombinationMap.uniform_map(rate=0.01, length=L, discrete=True)
map1 = msprime.RecombinationMap.uniform_map(rate=0.02, length=L, discrete=True)
population_configurations = [
    msprime.PopulationConfiguration(),
    msprime.PopulationConfiguration()
]

# Simulate population 0
ts_pop0 = msprime.simulate(
    samples=[msprime.Sample(population=0, time=0)] * 10,
    population_configurations=population_configurations,
    recombination_map=map0, end_time=T)
# Simulate population 1
ts_pop1 = msprime.simulate(
    samples=[msprime.Sample(population=1, time=0)] * 10,
    population_configurations=population_configurations,
    recombination_map=map1, end_time=T)

t0 = ts_pop0.tables
t1 = ts_pop1.tables

# Insert the nodes for t1 into t0
N = len(t0.nodes)
t0.nodes.append_columns(
    flags=t1.nodes.flags,
    time=t1.nodes.time,
    population=t1.nodes.population,
    individual=t1.nodes.individual,
    metadata=t1.nodes.metadata,
    metadata_offset=t1.nodes.metadata_offset)

# Add the edges for t1 into t0
t0.edges.append_columns(
    left=t1.edges.left,
    right=t1.edges.right,
    parent=t1.edges.parent + N,
    child=t1.edges.child + N)

t0.sort()
joint_ts = t0.tree_sequence()

# Now run the simulation of the joint history.
final_ts = msprime.simulate(
    from_ts=joint_ts,
    population_configurations=population_configurations,
    # Migrate everyone to population 0
    demographic_events=[
        msprime.MassMigration(time=T, source=1, dest=0)
    ],
    recombination_map=map0)
```

The idea is we simulate the two populations independently, then merge the two tree sequences together and then simulate their shared ancestry. This is quite similar to what's going on here, right? Should we also provide an operation to ``add`` two tree sequences together?

petrelharp:
> This is quite similar to what's going on here, right? Should we also provide an operation to add two tree sequences together?

I think you meant to post this example over in #623?  In which we have decided (I think) to call the method `union()`, not `graft()`?

jeromekelleher:
> I think you meant to post this example over in #623? In which we have decided (I think) to call the method union(), not graft()?

You're right, I did. Crud. I've moved the example over there for reference.

jeromekelleher:
I will definitely comment on this PR first thing tomorrow!

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
 }
 
+static int TSK_WARN_UNUSED
+node_table_copy_row(tsk_node_table_t *source, tsk_node_table_t *dest, tsk_id_t row,
+    tsk_id_t *population_map, tsk_id_t *individual_map)
+{
+    int ret = 0;
+    tsk_id_t pop, ind;
+    tsk_size_t metadata_length;
+    if (row >= (tsk_id_t) source->num_rows) {
```
jeromekelleher, line 7809: Need to check if ``row < 0`` as well

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+        ret = TSK_ERR_NODE_OUT_OF_BOUNDS;
+        goto out;
+    }
+    pop = source->population[row];
+    if (pop >= 0) {
```
jeromekelleher, line 7814: pop != TSK_NULL is clearer, I think.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
     return tsk_table_collection_truncate(self, &start);
 }
 
+static int TSK_WARN_UNUSED
+node_table_copy_row(tsk_node_table_t *source, tsk_node_table_t *dest, tsk_id_t row,
+    tsk_id_t *population_map, tsk_id_t *individual_map)
+{
+    int ret = 0;
+    tsk_id_t pop, ind;
+    tsk_size_t metadata_length;
```
jeromekelleher, line 7808: Actually, it'd save a bit of trouble if you did
```c
ret = node_table_get_row(source, row, &node);
if (ret < 0) {
    goto out;
}
```
This will do the bounds checking for you, and save some tedium in dealing with the offsets.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+int TSK_WARN_UNUSED
+tsk_table_collection_subset(
+    tsk_table_collection_t *self, tsk_id_t *nodes, size_t num_nodes)
+{
+    int ret = 0;
+    tsk_id_t k, n, m, ind, pop, mut, site;
+    tsk_id_t *node_map, *individual_map, *population_map, *site_map, *mutation_map;
+    tsk_table_collection_t tables;
+
+    node_map = calloc(self->nodes.num_rows, sizeof(tsk_id_t));
```
jeromekelleher, line 7989: Since we're memsetting all of these straight afterwards, might as well use ``malloc`` here.

jeromekelleher: Also, better do use sizeof(*node_map), e.g.
```c
node_map = malloc(self->nodes.num_rows * sizeof(*node_map));
```
this is a bit more robust and less error prone.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_table_collection_subset(
+    tsk_table_collection_t *self, tsk_id_t *nodes, size_t num_nodes)
+{
+    int ret = 0;
+    tsk_id_t k, n, m, ind, pop, mut, site;
+    tsk_id_t *node_map, *individual_map, *population_map, *site_map, *mutation_map;
```
jeromekelleher, line 7986: So it's not essential here as we're mallocing these pointers before we do anything else, but I think it's still better to explicitly inititialise these variables to NULL. We might update the code later and put in (say) an error check before the mallocs and then everything would explode. So,
```
tsk_id_t *node_map = NULL;
tsk_id_t *individual_map = NULL; // etc
```

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    if (ret < 0) {
+        goto out;
+    }
+out:
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_table_collection_subset(
+    tsk_table_collection_t *self, tsk_id_t *nodes, size_t num_nodes)
```
jeromekelleher, line 7982: I'd say ``tsk_size_t`` is more appropriate for ``num_nodes``.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    if (ret != 0) {
+        goto out;
+    }
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
+        n = nodes[k];
+        node_map[n] = k;
+        ind = tables.nodes.individual[n];
+        if (ind >= 0) {
```
jeromekelleher, line 8019: I think ind != TSK_NULL is a bit clearer. Ditto below.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+            if (individual_map[ind] < 0) {
```
jeromekelleher, line 8020: ``if individual_map[ind] == TSK_NULL`` - more concrete, clearer to the reader.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+        ret = node_table_copy_row(
+            &tables.nodes, &self->nodes, n, population_map, individual_map);
+        if (ret < 0) {
+            goto out;
+        }
+    }
+
+    // edges
+    for (k = 0; k < (tsk_id_t) tables.edges.num_rows; k++) {
+        m = node_map[tables.edges.parent[k]];
```
jeromekelleher, line 8049: I'm not sure we're getting much from the ``copy_row`` function here. What about
```c
edge_table_get_row(&tables.edges, k, &edge);
new_parent = node_map[edge.parent];
new_child = node_map[edge.child];
if (new_parent != TSK_NULL and new_child != TSK_NULL) {
    ret = edge_table_add_row(&self->edges,
          edge->left, edge->right, new_parent, new_child, edge->metadata, edge->metadata_length);
    if (ret < 0) {
        goto out;
    }
```

This actually makes life quite a bit simpler, because we don't need to test the ``edge_table_copy_row`` function now.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+        goto out;
+    }
+    ret = tsk_table_collection_clear(self);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
+        n = nodes[k];
```
jeromekelleher, line 8016: I find the notation ``n`` for a node confusing. We use variables like ``u`` and ``v`` to refer to nodes elsewhere, and keep ``n`` for things like "this is a fixed number".

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+            ret = edge_table_copy_row(&tables.edges, &self->edges, k, node_map);
+            if (ret < 0) {
+                goto out;
+            }
+        }
+    }
+
+    // mutations and sites
+    mut = 0;
+    for (site = 0; site < (tsk_id_t) tables.sites.num_rows; site++) {
```
jeromekelleher, line 8061: I think it would be simpler to use the ``get_row`` functions here and skip the helpers, like
```c
for (site_id = 0; ...) {
     tsk_site_table_get_row(&tables.sites, site_id, &site);
     while ((mutation_id < (tsk_id_t) tables.mutations.num_rows)
             && (tables.mutations.site[mutation_id] == site)) {
             tsk_mutation_table_get_row(&tables.mutations, mutation_id, &mutation)
             new_node = node_map[mutation.node];
             if (new_node != TSK_NULL) {
                   if (site_map[node.site] == TSK_NULL) {
                          ret = tsk_site_table_add_row(&tables.sites, site.position, /* etc */);                
                          if (ret < 0) {
                                goto out;
                          }
                         site_map[site] = ret;
                    }
                    // etc
             }
}

File: c/tskit/tables.h [hunk 33]
```diff
@@ -2409,6 +2409,24 @@ completes.
 int tsk_table_collection_simplify(tsk_table_collection_t *self, tsk_id_t *samples,
     tsk_size_t num_samples, tsk_flags_t options, tsk_id_t *node_map);
 
+/**
+@brief Subsets a table collection by a set of nodes.
+
+@rst
+Reduces the table collection, retaining only the nodes given in ``nodes``,
```
jeromekelleher, line 2416: This is a bit terser than the Python docs. C programmers will want precision too. I think we want something a more like the version we have on the python TreeSequence documentation.

python/_tskitmodule.c
```diff
@@ -6256,6 +6256,38 @@ TableCollection_link_ancestors(TableCollection *self, PyObject *args, PyObject *
...
+    shape = PyArray_DIMS(nodes_array);
+    num_nodes = shape[0];
+
+    err = tsk_table_collection_subset(self->tables, PyArray_DATA(nodes_array), num_nodes);
+    if (err != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = Py_BuildValue("");
+out:
```
jeromekelleher, line 6286: Very nice, this is perfect. We need some tests of this in test_lowlevel.py though, to make sure that the error conditions are handled correctly. We'll want to call the new method with, e.g., the incorrect number of arguments and bad types for numpy to make sure these error conditions are handled.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_table_collection_clear(self);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
```
jeromekelleher, line 8015: So if the first thing we do here is 
```
      ret = tsk_node_table_get_row(&tables.nodes, nodes[k], &node);
      if (ret < 0) {
           goto out;
      }
```
we do two things: 1) we check input, to make sure it's a valid node reference; and (2) we get the actual node object.

jeromekelleher: We'll definitely want some tests to make check that we error out when we pass in bad node refs.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,317 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    node_map = calloc(self->nodes.num_rows, sizeof(tsk_id_t));
+    individual_map = calloc(self->individuals.num_rows, sizeof(tsk_id_t));
+    population_map = calloc(self->populations.num_rows, sizeof(tsk_id_t));
+    site_map = calloc(self->sites.num_rows, sizeof(tsk_id_t));
+    mutation_map = calloc(self->mutations.num_rows, sizeof(tsk_id_t));
+    if (node_map == NULL || individual_map == NULL || population_map == NULL
+        || site_map == NULL || mutation_map == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
```
jeromekelleher, line 7998: The behaviour is currently undefined if any of these mallocs fail. That's because the ``tables`` variable is uninitialised, and we call table_collection_free on it below. We need to either memset the tables to zero, or run the table collection copy *first* before the mallocs (assuming they've been set to NULL).

python/tests/test_tables.py
```diff
@@ -2266,3 +2266,336 @@ def test_set_columns_not_implemented(self):
...
+            unpack = tskit.unpack_arrays
+            pack = tskit.pack_arrays
+        else:
+            unpack = tskit.unpack_bytes
+            pack = tskit.pack_bytes
+        unpacked = unpack(packed, offset)
+        unpacked = [unpacked[i] for i in subset]
+        return pack(unpacked)
+
+    def np_subset(self, tables, nodes, record_provenance=True):
```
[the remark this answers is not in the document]
mufernando: I agree this naive version would be the best as a test, but note we have the `verify_subset` which is more obviously correct. There we go row by row for each table.

I don't think we need the `np_subset` in the test to make it robust, but we decided to keep it there in case we want to have a look at it in the future. Importantly, with the numpy version we discovered that even though the numpy version is a little easier to write, the C version is waay faster.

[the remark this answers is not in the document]
petrelharp: > Importantly, with the numpy version we discovered that even though the numpy version is a little easier to write, the C version is waay faster.

This was a good exercise (for me at least) - naively, I thought that the numpy version would be easier to write and just as fast; in fact, it was harder to write and slower. (see the graph above) Maybe we could make it just as fast with additional work, but that would defeat the purpose. I lean towards leaving it in but would support removing it if you have a stronger opinion.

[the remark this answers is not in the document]
jeromekelleher: Which graph is this @petrelharp? 

Sure, we can keep the numpy version in, but I'd still like to see the naive dumb-as-rocks Python row-by-row version. I agree that the verify function does this, but it's not quite the same thing to check something as it is to write it down and actually do it. We can always compare all three versions to each other. In terms of making sure that the ragged column stuff is done properly, I think this is really the best and simplest way to do it.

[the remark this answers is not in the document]
petrelharp: Took me a minute to find, but [here](https://github.com/tskit-dev/tskit/pull/663#issuecomment-638340711) it is.

I'm not yet convinced that the naive version would make the testing more robust here, but I'll discuss with @mufernando.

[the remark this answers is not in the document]
jeromekelleher: > Took me a minute to find, but here it is.

Thanks - that is an impressive difference. Writing non-simple things in numpy is hard!

> I'm not yet convinced that the naive version would make the testing more robust here, but I'll discuss with @mufernando.

It's not just about testing robustness @petrelharp, it's about clarity of what the test is actually *doing*. We need to be able to read these tests in 5 years time and understand how they work - hence, my push for a simple, direct, Python implementation.

We'd have written it in the time it takes to write these comments...

c/tskit/tables.c
mufernando: ok, so memset tables to zero right after declaration

jeromekelleher: Yeah, either that or run tsk_table_collection_copy before the mallocs. I think this is guaranteed to init the object.

mufernando:
I think I addressed all of your comments, @jeromekelleher 

I was fighting with codecov bc it says the loops over mutations and migrations are never covered by our tests, but I'm pretty sure they are -- all our examples have mutations and migrations which seem to be correctly subsetted...

petrelharp:
I can take a look first if you like, @jeromekelleher.

jeromekelleher:
> I can take a look first if you like, @jeromekelleher.

Sure, that'd be great, thanks @petrelharp.  Or are you procrastinating from something...

c/tskit/tables.h
```diff
@@ -2409,6 +2409,36 @@ completes.
...
+they appear in the ``nodes`` argument. Specifically, this subsets and reorders
+each of the tables as follows:
+
+1. Nodes: if in the list of nodes, and in the order provided.
+2. Individuals and Populations: if referred to by a retained node,
+   and in the order first seen when traversing the list of retained nodes.
+3. Edges: if both parent and child are retained nodes.
+4. Mutations: if the mutation's node is a retained node.
+5. Sites: if any mutations remain at the site after removing mutations.
+6. Migrations: if the migration's node is a retained node.
```
petrelharp, line 2427: Mabye add afterwards: "Retained edges, mutations, sites, and migrations appear in the same order as in the original tables."

python/tests/test_lowlevel.py
```diff
@@ -233,9 +233,9 @@ def test_simplify_bad_args(self):
         ts = msprime.simulate(10, random_seed=1)
         tc = ts.tables.ll_tables
         with self.assertRaises(TypeError):
-            tc.simplify()
+            tc.subset()
```
petrelharp, line 236: whoops! did you mean to copy this test?

petrelharp: Oh, I see you did, below.

python/tests/test_tables.py
```diff
@@ -2266,3 +2268,169 @@ def test_set_columns_not_implemented(self):
...
+        rng = np.random.default_rng(9385)
+        ts = self.get_msprime_mig_example(Ne=100, sample_size=10, seed=12396)
+        tables = ts.tables
+        for n in [2, ts.num_nodes - 10]:
+            nodes = rng.choice(np.arange(ts.num_nodes), n, replace=False)
+            self.verify_subset(tables, nodes)
+
+    def test_wf_examples(self):
+        tables = wf.wf_sim(5, 5, seed=65634)
+        tables.sort()
```
petrelharp, line 2417: why the sort?

python/tskit/trees.py
```diff
@@ -4401,6 +4401,41 @@ def trim(self, record_provenance=True):
...
+        they appear in the ``nodes`` argument. Specifically, this subsets and reorders
+        each of the tables as follows:
+
+        1. Nodes: if in the list of nodes, and in the order provided.
+        2. Individuals and Populations: if referred to by a retained node,
+           and in the order first seen when traversing the list of retained nodes.
+        3. Edges: if both parent and child are retained nodes.
+        4. Mutations: if the mutation's node is a retained node.
+        5. Sites: if any mutations remain at the site after removing mutations.
+        6. Migrations: if the migration's node is a retained node.
```
petrelharp, line 4417: same note as in the C docs

python/tests/test_tables.py
```diff
@@ -2266,3 +2268,169 @@ def test_set_columns_not_implemented(self):
...
+
+    def test_random_subsets(self):
+        rng = np.random.default_rng(9385)
+        ts = self.get_msprime_mig_example(Ne=100, sample_size=10, seed=12396)
+        tables = ts.tables
+        for n in [2, ts.num_nodes - 10]:
+            nodes = rng.choice(np.arange(ts.num_nodes), n, replace=False)
+            self.verify_subset(tables, nodes)
+
+    def test_wf_examples(self):
```
petrelharp, line 2415: How about move the WF code to `self.get_wf_example`, and then in all the tests you can do something like this:
```
def get_examples():
   yield self.get_msprime_mig_example(Ne=100, sample_size=10, seed=12396)
   yield self.get_wf_example()

def test_subset_all(self):
    for ts in self.get_examples():
```

That way we'd be running the tests on more than one example, and would separate out the example generation from the tests.

python/tests/test_lowlevel.py
mufernando: oops, fixed it

petrelharp:
Looks good - just a few comments. I don't understand codecov either.

python/tests/test_tables.py
mufernando: I'm not sure, but all the other examples with the WF simulator do that. I think it has to do with the Edges Tables requirements.

python/tests/test_tables.py
mufernando: good idea!

mufernando:
I think I addressed all the issues. Not sure about codecov, @jeromekelleher. I don't understand how it works -- sometimes it complains sometimes it doesn't. Pretty sure the parts about mutations and migrations in the C code are covered by our tests, though.

python/tests/test_tables.py
petrelharp: Ah, well, since we're just working with tables, not tree sequences, let's *not* sort!

petrelharp:
Yep, over to you, @jeromekelleher (except my note about removing the sort() above).

python/tests/test_tables.py
mufernando: We need tree sequences for some of the tsutil functions (to add metadata, etc).

petrelharp: Hm: those don't need the tree sequence either, but ok!

mufernando: what do you mean? they take in tree sequences and dump the tables internally.

petrelharp: Right - they could be rewritten to operate on tables instead of tree sequences, since that's really what's going on. But, I'm not suggesting you do that.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    memset(mutation_map, 0xff, tables.mutations.num_rows * sizeof(*mutation_map));
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
+        nid = nodes[k];
+        ret = tsk_node_table_get_row(&tables.nodes, nid, &node);
+        if (ret < 0) {
+            goto out;
+        }
+        iid = tables.nodes.individual[nid];
```
jeromekelleher, line 7854: No need to look up the tables here - it's  more efficient to use the ``node_t`` struct that you've created. So, here,
```c
if (node.individual != TSK_NULL) {
     if (individual_map[node.individual] == TSK_NULL) {
```
which is also more readable.

jeromekelleher: Ah, I see, you were also handling the case there the individual is NULL. How about,
```c
new_ind_id = TSK_NULL;
if (node.individual != TSK_NULL) {
     // as before, then
     new_ind_id = individual_map[node.individual];
}

File: c/tskit/tables.c [hunk 44]
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+                tsk_individual_table_get_row(&tables.individuals, iid, &ind);
+                ret = tsk_individual_table_add_row(&self->individuals, ind.flags,
+                    ind.location, ind.location_length, ind.metadata,
+                    ind.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+                individual_map[iid] = ret;
+            }
+            iid = individual_map[iid];
```
jeromekelleher, line 7866: This isn't very readable - iid is referring to two different IDs at different points in the loop. How about ``new_ind_id`` or something?

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+        }
+        pid = tables.nodes.population[nid];
```
jeromekelleher, line 7868: Can use the same pattern here,
```c
new_pop_id = TSK_NULL;
if (node.population != TSK_NULL) {
    // etc
    new_pop_id = population_map[node.population];
}

File: c/tskit/tables.c [hunk 46]
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+                population_map[pid] = ret;
+            }
+            pid = population_map[pid];
+        }
+        ret = tsk_node_table_add_row(&self->nodes, node.flags, node.time, pid, iid,
+            node.metadata, node.metadata_length);
+        if (ret < 0) {
+            goto out;
+        }
+        if (ret != k) {
```
jeromekelleher, line 7886: Why would ret not be equal to k? If this is checking for the presence of a bug it should be an assert (but, I think it's unnecessary?, could just assert this at the end of the loop once?

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    }
+    memset(node_map, 0xff, tables.nodes.num_rows * sizeof(*node_map));
+    memset(individual_map, 0xff, tables.individuals.num_rows * sizeof(*individual_map));
+    memset(population_map, 0xff, tables.populations.num_rows * sizeof(*population_map));
+    memset(site_map, 0xff, tables.sites.num_rows * sizeof(*site_map));
+    memset(mutation_map, 0xff, tables.mutations.num_rows * sizeof(*mutation_map));
+
+    // nodes, individuals, populations
+    for (k = 0; k < (tsk_id_t) num_nodes; k++) {
+        nid = nodes[k];
```
jeromekelleher, line 7849: I wouldn't bother saving this variable, just run  ``tsk_node_table_get_row(&tables.nodes, nodes[k], &node);`` and use ``node.id`` everywhere later. This is exactly as efficient (possibly more, as you're reducing the number of variables in the stack, and therefore better for cache).

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+        if (ret != k) {
+            goto out;
+        }
+        node_map[nid] = ret;
+    }
+
+    // edges
+    for (eid = 0; eid < (tsk_id_t) tables.edges.num_rows; eid++) {
+        tsk_edge_table_get_row(&tables.edges, eid, &edge);
+        pid = node_map[tables.edges.parent[eid]];
```
jeromekelleher, line 7895: We don't need to look up the table here, how about 
```c
new_parent = node_map[edge.parent];
new_child = node_map[edge.parent];
```
It's more readable than pid and eid.  (What's p, and e? Which set of tables is it an ID in?)

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    }
+
+    // mutations and sites
+    mid = 0;
+    for (sid = 0; sid < (tsk_id_t) tables.sites.num_rows; sid++) {
+        tsk_site_table_get_row(&tables.sites, sid, &site);
+        while ((mid < (tsk_id_t) tables.mutations.num_rows)
+               && (tables.mutations.site[mid] == sid)) {
+            tsk_mutation_table_get_row(&tables.mutations, mid, &mut);
+            nid = node_map[mut.node];
```
jeromekelleher, line 7913: new_node, etc.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+            ret = tsk_edge_table_add_row(&self->edges, edge.left, edge.right, pid, cid,
+                edge.metadata, edge.metadata_length);
+            if (ret < 0) {
+                goto out;
+            }
+        }
+    }
+
+    // mutations and sites
+    mid = 0;
```
jeromekelleher, line 7907: ``mutation_id`` is better - is this the ``mid`` point of a line?

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+    }
+
+    // migrations
+    for (gid = 0; gid < (tsk_id_t) tables.migrations.num_rows; gid++) {
+        tsk_migration_table_get_row(&tables.migrations, gid, &mig);
+        nid = node_map[mig.node];
+        if (nid != TSK_NULL) {
+            sid = mig.source;
+            did = mig.dest;
+            if (sid != TSK_NULL) {
```
jeromekelleher, line 7947: Migrations are guaranteed to be non-null, so you can skip this check.

c/tskit/tables.c
```diff
@@ -7799,6 +7799,182 @@ tsk_table_collection_clear(tsk_table_collection_t *self)
...
+                sid = population_map[sid];
+            }
+            if (did != TSK_NULL) {
```
jeromekelleher, line 7950: Actually, what if this is a population we haven't already seen before? You would need to check the map, I guess, and update if necessary. I don't think I'd bother though: unless you have an application using migrations, I would just raise an error if there's stuff in the migration table. I don't think it's really used much. This is what we do in simplify fwiw (there's an error code for "we don't support migrations" which could be reused).

python/tests/tsutil.py
```diff
@@ -539,6 +539,85 @@ def compute_mutation_parent(ts):
...
+                node_map[edge.child],
+                edge.metadata,
+            )
+    for mig in full.migrations:
+        if mig.node in nodes:
+            tables.migrations.add_row(
+                mig.left,
+                mig.right,
+                node_map[mig.node],
+                pop_map.get(mig.source, tskit.NULL),
```
jeromekelleher, line 596: I don't think this works - see comments above in the C. I think we should just not bother with migrations.

jeromekelleher:
We also need to bump up test coverage quite a bit. The C tests don't have any sites or mutations, so that's not being covered. Neither are migrations (but we should just drop that anyway, and test that we behave correctly when we try to subset a including them).

Need to add a test for the TreeSequence.subset method also. This can just be verifying that it returns the same thing as ``ts.dump_tables().subset()``

c/tskit/tables.c
mufernando: Right, the dest population has to be in the map, but it is possible the source isn't. I didn't know source and dest had to be non-NULL. My idea was that in this case the source pop would be NULL in the subset tables. Alternatively we could remove migrations with source pop not seen in any of the `nodes` from the subset. I think we are so close to being able to support the Migrations Table in this operation that we should just go for it.

jeromekelleher: I think either of them could be missing from the map. Suppose you have a node that starts in one population then goes for a long journey through hundreds of populations. In that case the majority of the populations will not have been seen before. So, I think we'd have to append all these populations to the map.

There's a fair bit of testing to be done in this... It would be a lot quicker the other way.

petrelharp: Ah, of course - I'd intended them to be NULL in this case, but of course it doesn't make sense for them to be NULL. I agree that throwing an error is the thing to do.

c/tskit/tables.c
petrelharp: > could just assert this at the end of the loop once?

good call, yes

c/tskit/tables.c
petrelharp: Whoops - i meant to throw an error if the source or dest turns out to be unmapped.

mufernando: I decided to remove support for migrations altogether...

mufernando:
thanks for the revisions, @jeromekelleher! I added some mutations to the C side, but codecov isn't being rerun. Can we explicitly ask it to be run again?

benjeffery:
Codecov isn't reporting as valgrind is reporting a memory leak: https://circleci.com/gh/tskit-dev/tskit/2088?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mufernando:
yeah, just fixed that. but either way codecov is not in the list of pending checks.

benjeffery:
Codecov will only appear after circle ci is successfully completed, the coverage is measured in the circle CI run.

jeromekelleher:
Looking great, thanks @mufernando! Apologies for all the nitpicking, but I think it was worth the effort. The code is super-clear and clean now.

Just need to get codecov happy and squash the commits now I think.

mufernando:
Thank you, @jeromekelleher and @petrelharp for the patience and help along the way!

I am not sure how to bump up codecov at this point. The parts of the code not being tested are some ifs on the return of `add_row` lines. What is the proper way of dealing with this?

jeromekelleher:
You've tested all the conditions that can reasonably happen @mufernando, so we're ready to merge. Can you squash the commits please?

mufernando:
woot woot!

thanks, you all!

PR #664, opened 2020-06-02
more general error on fail to write

petrelharp:
@mufernando discovered that on his system (macOS), this test actually returns "Operation not permitted" instead of "Permission denied". This should cover that situation as well.  @mufernando, can you see if it passes that test, now?


mufernando:
It passes all the tests now. Thanks!

petrelharp:
Minor, but easy to fix!

Issue #603, opened 2020-05-11
Improve metadata docs


benjeffery:
I've started an issue on this as there has been some discussion and wanted a central place for people to make suggestions, rather than on other PRs.

e.g.
https://github.com/tskit-dev/tskit/pull/595#issuecomment-626731968

benjeffery:
`struct` codec parameters need more explaining for sure.

benjeffery:
`struct` default array length is `L` not `Q`.

bhaller:
Hi folks.  Here's a read-through from a naive user who only just barely grasps what any of this means, and is pretty much Python-illiterate, so you can just ignore my comments if they are too clueless.  :->

Typo: "A good to guide to"

Regarding: "The metadata for each entity is described by a schema for each entity type."  I feel like at this point it needs to be clarified who defines the schema, and whether it is fixed for each entity type or not.  A sentence like "The metadata schema for each entity type is typically defined by the software that is generating the tree sequence; it will vary among software packages depending upon the information they have and wish to preserve.  A user of tskit could also, in principle, define their own metadata schema in order to associate different or additional information with entities, but more typically they will work with the metadata schema they receive from the tree-sequence-generating software they use."  Something like that, addressing where the schema comes from, who defines it, whether it is always the same for a given entity, whether it can be changed, etc.  What the usage pattern is that surrounds all this.

Typo: "by any third-party" (no hyphen since "third party" is a noun, not an adjective; as an adjective, "third-party software", it gets a hyphen :->)

Re: "It is possible to define a custom codec using tskit.register_metadata_codec(), however this should only be used when necessary as downstream users of the metadata will not be able to decode it without the custom codec."  I'm not sure I understand this point.  If you use a custom codec, the underlying data will still be binary, and could still be decoded with a JSON or struct codec that matches the format of the data, no?  It just can't be decoded by whatever the custom codec is unless the downstream user has that custom codec?  Or maybe there can only be one codec declared for a given metadata schema?  Might need clarification.  What I'm driving at is: one might want to define a custom codec that decodes metadata into some kind of spiffy custom Python objects that do fancy things, but the underlying binary data might also be representable with a struct codec that matches the format, allowing access to the metadata without all the spiffy extra features provided by the custom codec.  Is this sort of dual nature allowed?  Or am I completely misunderstanding what is meant by a "custom codec"?

[...]

PR #666, opened 2020-06-03
Add high-level python API for tree-sequence metadata

### LINKED ISSUE #534: Tree-sequence-level metadata

benjeffery:
Resolves #534 
As we are doing a metadata docs pass in #603 I have only done minimal docs.

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -2109,6 +2109,103 @@ class TestTableCollection(unittest.TestCase):
         self.assertEqual(len(tree.parent_dict), 0)
 
 
+class TestTableCollectionMetadata(unittest.TestCase):
+
+    metadata_schema = metadata.MetadataSchema(
+        {
+            "codec": "json",
+            "title": "Example Metadata",
+            "type": "object",
+            "properties": {
+                "one": {"type": "string"},
+                "two": {"type": "number"},
+                "three": {"type": "array"},
+                "four": {"type": "boolean"},
+            },
+            "required": ["one", "two", "three", "four"],
+            "additionalProperties": False,
+        },
+    )
+
+    def metadata_example_data(self, val=0):
+        return {
+            "one": "val one",
+            "two": val,
+            "three": list(range(val, val + 10)),
+            "four": True,
+        }
+
+    def test_set_metadata_schema(self):
+        tc = tskit.TableCollection(1)
+        metadata_schema2 = metadata.MetadataSchema({"codec": "json"})
+        # Default is no-op metadata codec
+        self.assertEqual(str(tc.metadata_schema), str(metadata.MetadataSchema(None)))
+        # Set
+        tc.metadata_schema = self.metadata_schema
+        self.assertEqual(str(tc.metadata_schema), str(self.metadata_schema))
+        # Overwrite
+        tc.metadata_schema = metadata_schema2
+        self.assertEqual(str(tc.metadata_schema), str(metadata_schema2))
+        # Remove
+        tc.metadata_schema = ""
+        self.assertEqual(str(tc.metadata_schema), str(metadata.MetadataSchema(None)))
+        # Set after remove
+        tc.metadata_schema = self.metadata_schema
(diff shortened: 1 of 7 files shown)


python/tests/test_tables.py
```diff
@@ -2109,6 +2109,107 @@ def test_sequence_length_longer_than_edges(self):
...
+                "three": {"type": "array"},
+                "four": {"type": "boolean"},
+            },
+            "required": ["one", "two", "three", "four"],
+            "additionalProperties": False,
+        },
+    )
+
+    def metadata_example_data(self):
+        try:
```
jeromekelleher, line 2131: I don't like this as it makes the tests stateful and the test data will depend on what other tests are run on this class. Also adding instance variables to a TestCase is a bit weird. Either put in some constants, or make ``val`` a parameter.

python/tskit/tables.py
```diff
@@ -1884,6 +1884,30 @@ def sequence_length(self, sequence_length):
     def file_uuid(self):
         return self.ll_tables.file_uuid
 
+    @property
+    def metadata_schema(self) -> metadata.MetadataSchema:
+        """
```
jeromekelleher, line 1889: Docstring comments aren't indented like this elsewhere.

jeromekelleher: Ditto other new methods here.

python/tskit/tables.py
```diff
@@ -1884,6 +1884,30 @@ def sequence_length(self, sequence_length):
...
+            The :class:`tskit.MetadataSchema` for this TableCollection.
+        """
+        return metadata.parse_metadata_schema(self.ll_tables.metadata_schema)
+
+    @metadata_schema.setter
+    def metadata_schema(self, schema: metadata.MetadataSchema) -> None:
+        # Check the schema is a valid schema instance by roundtipping it.
```
jeromekelleher, line 1896: "roundtripping".

docs/tutorial.rst
```diff
@@ -853,6 +853,27 @@ Both the schema and the metadata can be retrieved from a :class:`tskit.TreeSeque
...
+Top-level metadata can be set on a :class:`tskit.TableCollection` and read from a
+:class:`tskit.TreeSequence`:
+
+.. code-block:: python
+
+    tables = tskit.TableCollection(sequence_length=1)
+    tables.metadata_schema = tskit.MetadataSchema(
+        {
+            "codec": "json",
+            "properties": {"author": {"type": "string"}, "version": {"type": "number"}},
```
jeromekelleher, line 865: I'm not blown away by this example - it's not really showing the crucial stuff that we want people to associate with a tree sequence so that people can understand what it means. Examples are important, and people have a very strong habit of copying them and assuming that they represent best practise. 

But, maybe we should just put this off until we do the proper treatment of standardised tree sequence metadata keys (which we don't want to get bogged down in now). I'd say just delete the example.

python/tests/test_tables.py
benjeffery: I made it a param.

python/tskit/tables.py
benjeffery: Dammit, asleep at the wheel today! Fixed.

python/tskit/tables.py
benjeffery: Fixed.

docs/tutorial.rst
benjeffery: Yeah, should have thought more about that one. Deleted for now.

benjeffery:
@jeromekelleher Fixed up - given the number of this PR I got off lightly. :imp:

jeromekelleher:
Not an Iron Maiden fan then? I keep singing "6,66 the PR of the Beast" to myself. Classic.

benjeffery:
Travis AWOL again - will look into this as happening a lot.

PR #672, opened 2020-06-06
Mutation time

### LINKED ISSUE #513: Add MutationTable.time column
Currently we have no way of recording the precise age of a mutation. All that we know is that it occured somewhere on a branch. Most of the time this isn't an issue, but there are situtations in which we need to record this precise time (see, https://github.com/tskit-dev/msprime/pull/946, e.g.).

We can add a MutationTable.time column easily enough. If the mutation time is *not* specified (the default) then we set it to the time of the mutation's node. If multiple mutations are added on the same branch without times being specified, then they will have the same time (oh well, no big deal I guess).

In terms of the file format, we make this optional in the same way that Edge metadata is optional (it would be simplest to call ``set_columns`` to implement this though). It might be best to do #506 first,  therefore.

As this is backward compatible, we don't need to increment the file version number.

We will break some C code since ``mutation_table_add_row`` et al will need a new parameter, which most naturally comes between the current ``parent`` and ``derived_state``. I guess providing TSK_NULL indicates you want the default for this. Since we're already breaking some C code, this seems fine.

@benjeffery, any thoughts here?

### LINKED ISSUE #692: Mutation time - backwards compatibility
As it stands when #672 is merged we break compatibility in the following ways:

1. File format (not too bad as #672 has a `tskit upgrade` cli support)
2. `python MutationTable.add_row` every call to this function will need updating.
3. `python MutationTable.[set|append]_columns` every call to these functions will need updating.
4. `cpython LightweightTableCollection.from_dict` will fail when passed dicts made in, for example, msprime 0.7.4 or newer.
5. C API extra time args to mutation table methods (no way around this)

2 and 3 _could_ be avoided by allowing a default mutation-time of zero as on these methods. This would result in an invalid tree-sequence _but_ `TableCollection.tree_sequence` could catch mutation time errors, call `compute_mutation_times` and build the tree sequence again, logging out a warning that mutation times were set.

4 Could be avoided in a similar way, by detecting the absence of mutation times in the dict and calling `compute_mutation_times` again.

Note that currently `compute_mutation_times` spreads mutations evenly across edges (except for those above a root node where the mutation is placed at the same time as the root)

In summary we could make this release more or less backward compatible (python-wise). My concern is that users should be aware that they have placeholder mutation times, as otherwise they could end up using them in some analysis as if they were "proper".

Tagging some people who may have input here: @jeromekelleher @petrelharp @hyanwong @molpopgen @bhaller

benjeffery:
Fixes #513
Fixes #692 

The mutation table now has a mandatory `time` column. For a tree sequence to be valid, these times have to be non-null, finite, older than their node, younger than parent mutation and younger than the parent node of their node, this is all checked in `tsk_table_collection_check_mutation_times` which is called by `tsk_table_collection_check_integrity`.

There is a new function `tsk_table_collection_compute_mutation_times` which creates times that satisfy these constraints by equally spreading out mutations along an edge that share a site. Along with a `tskit upgrade` addition to add times.

UPDATE:
Following discussion in #692 I'm changing this PR to be backwards compatible with mutation times being optional.

TODOs:
- [x]  C tests for new functions
- [x]  C docs for new functions
- [x] Python tests for new functions
- [x] Python docs for new functions
- [x]  Add mutation.time to C Python
- [x]  Add `compute_mutation_times` to C Python
- [x]  Add mutation.time to Python
- [x]  Fix up python tests somehow (monkey patch `msprime` to call  `compute_mutation_times` as a temporary measure?)
- [x]  ~Add random spreading option to  `compute_mutation_times`~ Will move to follow-up PR.
- [x] File format version number bump
- [x] `tskit upgrade` implementation and tests.
- [x] Specifc NAN as default
- [x] file test with missing column
- [x] test with mixed nans and values
- [x] test that NAN macro `isnan` in python
- [x] test no times in dict repr
- [x] test for `__eq__` `__neq__` on `Mutation` class

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -6,14 +6,8 @@ In development.
 
 **Breaking changes**
 
-- Mutations now have a double-precision floating-point ``time`` column. As this is a
-  mandatory column the file format version has been bumped to 13.0. Pre-existing
-  tree sequences can be upgraded using the ``tskit upgrade`` command-line utility,
-  which will assign times to mutations by spreading them evenly along the edges on
-  which they occur (using ``tsk_table_collection_compute_mutation_times``. For a tree
-  sequence to be considered valid it must meet new criteria for mutation times, see
-  :ref:`sec_mutation_requirements`. Mutation methods such as ``add_row`` now have an
-  additional argument.
+- ``tsk_mutation_table_add_row`` has an extra ``time`` argument. If the time
+  is unknown ``TSK_UNKNOWN_TIME`` should be passed.
   (:user:`benjeffery`, :pr:`672`)
 
 - Change genotypes from unsigned to signed to accommodate missing data
@@ -44,11 +38,13 @@ In development.
 
 **New features**
 
-- Add ``time`` column to mutations, along with new function
-  ``tsk_table_collection_compute_mutation_times`` and new flag to
+- Mutations now have an optional double-precision floating-point ``time`` column.
+  If not specified, this defaults to a particular NaN value (``TSK_UNKNOWN_TIME``)
+  indicating that the time is unknown. For a tree sequence to be considered valid
+  it must meet new criteria for mutation times, see :ref:`sec_mutation_requirements`.
+  Add ``tsk_table_collection_compute_mutation_times`` and new flag to
   ``tsk_table_collection_check_integrity``:``TSK_CHECK_MUTATION_TIME``.
   (:user:`benjeffery`, :pr:`672`)
-
 
 - Add ``metadata`` and ``metadata_schema`` fields to table collection, with accessors on
   tree sequence. These store arbitrary bytes and are optional in the file format.
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -6,6 +6,16 @@ In development.
 
 **Breaking changes**
 
+- Mutations now have a double-precision floating-point ``time`` column. As this is a
+  mandatory column the file format version has been bumped to 13.0. Pre-existing
+  tree sequences can be upgraded using the ``tskit upgrade`` command-line utility,
+  which will assign times to mutations by spreading them evenly along the edges on
+  which they occur (using ``tsk_table_collection_compute_mutation_times``. For a tree
+  sequence to be considered valid it must meet new criteria for mutation times, see
+  :ref:`sec_mutation_requirements`. Mutation methods such as ``add_row`` now have an
+  additional argument.
+  (:user:`benjeffery`, :pr:`672`)
+
 - Change genotypes from unsigned to signed to accommodate missing data
   (see :issue:`144` for discussion). This only affects users of the
   ``tsk_vargen_t`` class. Genotypes are now stored as int8_t and int16_t
@@ -33,6 +43,12 @@ In development.
   (:user:`benjeffery`, :pr:`493`)
 
 **New features**
+
+- Add ``time`` column to mutations, along with new function
+  ``tsk_table_collection_compute_mutation_times`` and new flag to
+  ``tsk_table_collection_check_integrity``:``TSK_CHECK_MUTATION_TIME``.
+  (:user:`benjeffery`, :pr:`672`)
+
 
 - Add ``metadata`` and ``metadata_schema`` fields to table collection, with accessors on
   tree sequence. These store arbitrary bytes and are optional in the file format.
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -129,6 +129,29 @@ verify_compute_mutation_parents(tsk_treeseq_t *ts)
 }
 
 static void
+verify_compute_mutation_times(tsk_treeseq_t *ts)
+{
+    int ret;
+    size_t size = tsk_treeseq_get_num_mutations(ts) * sizeof(tsk_id_t);
+    tsk_id_t *time = malloc(size);
+    tsk_table_collection_t tables;
+
+    CU_ASSERT_FATAL(time != NULL);
+    ret = tsk_treeseq_copy_tables(ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    memcpy(time, tables.mutations.time, size);
+    /* Make sure the tables are actually updated */
+    memset(tables.mutations.time, 0, size);
+
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(memcmp(time, tables.mutations.time, size), 0);
+
+    free(time);
+    tsk_table_collection_free(&tables);
+}
+
+static void
 verify_individual_nodes(tsk_treeseq_t *ts)
 {
     int ret;
@@ -3140,12 +3163,12 @@ test_single_tree_bad_mutations(void)
     const char *sites = "0       0\n"
                         "0.1     0\n"
                         "0.2     0\n";
-    const char *mutations = "0   0  1  -1\n"
-                            "1   1  1  -1\n"
-                            "2   4  1  -1\n"
-                            "2   1  0  2\n"
-                            "2   1  1  3\n"
-                            "2   2  1  -1\n";
+    const char *mutations = "0   0  1  -1  0\n"
+                            "1   1  1  -1  0\n"
+                            "2   4  1  -1  1\n"
+                            "2   1  0  2   0\n"
+                            "2   1  1  3   0\n"
+                            "2   2  1  -1  0\n";
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
     tsk_flags_t load_flags = TSK_BUILD_INDEXES;
@@ -3281,6 +3304,27 @@ test_single_tree_bad_mutations(void)
     CU_ASSERT_EQUAL(ret, TSK_ERR_MUTATION_PARENT_AFTER_CHILD);
     tsk_treeseq_free(&ts);
     tables.mutations.parent[2] = TSK_NULL;
+
+    /* time < node time */
+    tables.mutations.time[2] = 0;
+    ret = tsk_treeseq_init(&ts, &tables, load_flags);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[2] = 1;
+
+    /* time > parent mutation */
+    tables.mutations.time[4] = 0.5;
+    ret = tsk_treeseq_init(&ts, &tables, load_flags);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[4] = 0;
+
+    /* time > parent node */
+    tables.mutations.time[0] = 1.5;
+    ret = tsk_treeseq_init(&ts, &tables, load_flags);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[0] = 0;
 
     /* Check to make sure we've maintained legal mutations */
     ret = tsk_treeseq_init(&ts, &tables, load_flags);
@@ -3714,12 +3758,12 @@ test_single_tree_compute_mutation_parents(void)
     const char *sites = "0       0\n"
                         "0.1     0\n"
                         "0.2     0\n";
-    const char *mutations = "0   0  1  -1\n"
-                            "1   1  1  -1\n"
-                            "2   4  1  -1\n"
-                            "2   1  0  2\n"
-                            "2   1  1  3\n"
-                            "2   2  1  -1\n";
+    const char *mutations = "0   0  1  -1  0\n"
+                            "1   1  1  -1  0\n"
+                            "2   4  1  -1  1\n"
+                            "2   1  0  2   0\n"
+                            "2   1  1  3   0\n"
+                            "2   2  1  -1  0\n";
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
 
@@ -3748,10 +3792,6 @@ test_single_tree_compute_mutation_parents(void)
 
     /* Compute the mutation parents */
     verify_compute_mutation_parents(&ts);
-
-    /* Verify consistency of individuals */
-    verify_individual_nodes(&ts);
-    tsk_treeseq_free(&ts);
 
     /* Bad site reference */
     tables.mutations.site[0] = -1;
@@ -3813,6 +3853,124 @@ test_single_tree_compute_mutation_parents(void)
     CU_ASSERT_EQUAL(ret, 0);
     CU_ASSERT_EQUAL(tsk_treeseq_get_num_sites(&ts), 3);
     CU_ASSERT_EQUAL(tsk_treeseq_get_num_mutations(&ts), 6);
+    tsk_treeseq_free(&ts);
+
+    tsk_treeseq_free(&ts);
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_single_tree_compute_mutation_times(void)
+{
+    int ret = 0;
+    const char *sites = "0       0\n"
+                        "0.1     0\n"
+                        "0.2     0\n"
+                        "0.3     0\n";
+    const char *mutations = "0   0  1  -1  3\n"
+                            "1   1  1  -1  3\n"
+                            "2   4  1  -1  8\n"
+                            "2   1  0  2   4\n"
+                            "2   1  1  3   2\n"
+                            "2   2  1  -1  4\n"
+                            "3   6  1  -1  10\n";
+    /*          6          */
+    /*          6          */
+    /*         / \         */
+    /*        /   \        */
+    /*       2     \       */
+    /*      /       5      */
+    /*     4       / \     */
+    /*    0 1,3,4 5   \    */
+    /*   0   1   2     3   */
+
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    tables.sequence_length = 1;
+    parse_nodes(single_tree_ex_nodes, &tables.nodes);
+    CU_ASSERT_EQUAL_FATAL(tables.nodes.num_rows, 7);
+    tables.nodes.time[4] = 6;
+    tables.nodes.time[5] = 8;
+    tables.nodes.time[6] = 10;
+    parse_edges(single_tree_ex_edges, &tables.edges);
+    CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 6);
+    parse_sites(sites, &tables.sites);
+    parse_mutations(mutations, &tables.mutations);
+    CU_ASSERT_EQUAL_FATAL(tables.sites.num_rows, 4);
+    CU_ASSERT_EQUAL_FATAL(tables.mutations.num_rows, 7);
+    tables.sequence_length = 1.0;
+
+    ret = tsk_table_collection_build_index(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    /* Check to make sure we have legal mutations */
+    ret = tsk_treeseq_init(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_sites(&ts), 4);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_mutations(&ts), 7);
+
+    /* Compute the mutation times */
+    verify_compute_mutation_times(&ts);
+
+    /* Verify consistency of individuals */
+    verify_individual_nodes(&ts);
+    tsk_treeseq_free(&ts);
+
+    /* Bad random param */
+    ret = tsk_table_collection_compute_mutation_times(&tables, (double *) 1, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_BAD_PARAM_VALUE);
+
+    /* Bad site reference */
+    tables.mutations.site[0] = -1;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_SITE_OUT_OF_BOUNDS);
+    tables.mutations.site[0] = 0;
+
+    /* Bad site reference */
+    tables.mutations.site[0] = -1;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_SITE_OUT_OF_BOUNDS);
+    tables.mutations.site[0] = 0;
+
+    /* mutation sites out of order */
+    tables.mutations.site[0] = 2;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_UNSORTED_MUTATIONS);
+    tables.mutations.site[0] = 0;
+
+    /* sites out of order */
+    tables.sites.position[0] = 0.11;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_UNSORTED_SITES);
+    tables.sites.position[0] = 0;
+
+    /* Bad node reference */
+    tables.mutations.node[0] = -1;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
+    tables.mutations.node[0] = 0;
+
+    /* Bad node reference */
+    tables.mutations.node[0] = (tsk_id_t) tables.nodes.num_rows;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
+    tables.mutations.node[0] = 0;
+
+    /* Mutations not ordered by site */
+    tables.mutations.site[2] = 0;
+    ret = tsk_table_collection_compute_mutation_times(&tables, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_UNSORTED_MUTATIONS);
+    tables.mutations.site[2] = 2;
+
+    ret = tsk_treeseq_init(&ts, &tables, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_sites(&ts), 4);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_mutations(&ts), 7);
     tsk_treeseq_free(&ts);
 
     tsk_treeseq_free(&ts);
@@ -5350,7 +5508,7 @@ test_deduplicate_sites_errors(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = tsk_site_table_add_row(&tables.sites, 2, "TT", 2, "MM", 2);
     CU_ASSERT_EQUAL_FATAL(ret, 1);
-    ret = tsk_mutation_table_add_row(&tables.mutations, 0, 0, -1, "T", 1, NULL, 0);
+    ret = tsk_mutation_table_add_row(&tables.mutations, 0, 0, -1, 0, "T", 1, NULL, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = tsk_node_table_add_row(&tables.nodes, 0, 0, TSK_NULL, TSK_NULL, NULL, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
@@ -5732,6 +5890,8 @@ main(int argc, char **argv)
             test_single_tree_simplify_null_samples },
         { "test_single_tree_compute_mutation_parents",
             test_single_tree_compute_mutation_parents },
+        { "test_single_tree_compute_mutation_times",
+            test_single_tree_compute_mutation_times },
         { "test_single_tree_is_descendant", test_single_tree_is_descendant },
         { "test_single_tree_map_mutations", test_single_tree_map_mutations },
         { "test_single_tree_map_mutations_internal_samples",
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -3792,6 +3792,7 @@ test_single_tree_compute_mutation_parents(void)
 
     /* Compute the mutation parents */
     verify_compute_mutation_parents(&ts);
+    tsk_treeseq_free(&ts);
 
     /* Bad site reference */
     tables.mutations.site[0] = -1;
@@ -3855,7 +3856,6 @@ test_single_tree_compute_mutation_parents(void)
     CU_ASSERT_EQUAL(tsk_treeseq_get_num_mutations(&ts), 6);
     tsk_treeseq_free(&ts);
 
-    tsk_treeseq_free(&ts);
     tsk_table_collection_free(&tables);
 }
 
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -287,6 +287,18 @@ tsk_strerror_internal(int err)
         case TSK_ERR_UNSORTED_MUTATIONS:
             ret = "Mutations must be provided in non-decreasing site order";
             break;
+        case TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE:
+            ret = "Mutations must have a time equal to or greater than that of their "
+                  "node";
+            break;
+        case TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION:
+            ret = "Mutations must have a time equal to or less than that of their "
+                  "parent mutation";
+            break;
+        case TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE:
+            ret = "Mutations must have a time less than the parent node of "
+                  "the edge they are on";
+            break;
 
         /* Sample errors */
         case TSK_ERR_DUPLICATE_SAMPLE:
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -288,16 +288,16 @@ tsk_strerror_internal(int err)
             ret = "Mutations must be provided in non-decreasing site order";
             break;
         case TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE:
-            ret = "Mutations must have a time equal to or greater than that of their "
-                  "node";
+            ret = "A mutation's time must be >= the node time, or be marked as "
+                  "'unknown'";
             break;
         case TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION:
-            ret = "Mutations must have a time equal to or less than that of their "
-                  "parent mutation";
+            ret = "A mutation's time must be <= the parent mutation time (if known), or "
+                  "be marked as 'unknown'";
             break;
         case TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE:
-            ret = "Mutations must have a time less than the parent node of "
-                  "the edge they are on";
+            ret = "A mutation's time must be < the parent node of the edge on which it "
+                  "occurs, or be marked as 'unknown'";
             break;
 
         /* Sample errors */
@@ -636,4 +636,16 @@ tsk_round(double x, unsigned int ndigits)
         z = z / pow1;
     }
     return z;
+}
+
+/* As NANs are not equal, use this function to check for equality to TSK_UNKNOWN_TIME */
+bool
+tsk_is_unknown_time(double val)
+{
+    union {
+        uint64_t i;
+        double f;
+    } nan_union;
+    nan_union.f = val;
+    return nan_union.i == TSK_UNKNOWN_TIME_HEX;
 }
--- a/c/tskit/core.h
+++ b/c/tskit/core.h
@@ -34,7 +34,9 @@
 extern "C" {
 #endif
 
+#include <math.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <limits.h>
 
 #ifdef __GNUC__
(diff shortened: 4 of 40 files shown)


jeromekelleher:
I've had a quick look over, looks exactly right to me.

benjeffery:
@jeromekelleher 
Just the `tskit upgrade` code to go now, running into an issue. I need to load the tables from a kastore, bypassing validation. I'm trying something like this:
```python
    with kastore.load(src) as store:
        data = dict(store)
    data["format/version"] = numpy.asarray([13, 0], dtype=numpy.uint32)
    data["mutations/time"] = np.full(data["mutations/node"].shape, -1, dtype=np.float64)
    table_dict = collections.defaultdict(collections.defaultdict)
    for key, value in data.items():
        if "/" in key:
            table, col = key.split("/")
            if col != "metadata_schema":
                table_dict[table][col] = value
        else:
            table_dict[key] = value
    # TODO Reinstate schemas
    tables = tskit.TableCollection.fromdict(table_dict)
    tables.compute_mutation_times()
    tables.tree_sequence().dump(dest)

However I'm getting:
```
  File "/home/benj/projects/tskit/python/tskit/tables.py", line 467, in set_columns
    self.ll_table.set_columns(
TypeError: Cannot cast array data from dtype('uint8') to dtype('int8') according to the rule 'safe'
```
As it seems the python code is using `NPY_INT8` and the C code `KAS_UINT8` for byte columns. This looks like an error?

**Comment by benjeffery** [comment]
I've gotten around this for now with a cast: https://github.com/tskit-dev/tskit/pull/672/commits/9b6553d97ec7e2cd08234f796c5b745a3de365ff

**Comment by benjeffery** [comment]
Almost ready for review - let me do a line-by-line check first.

**Comment by benjeffery** [comment]
@jeromekelleher Ready for review. Sorry it took a while, keep finding tests to add and bits to tweak!

File: c/tskit/tables.c [hunk 1]
```diff
@@ -6435,6 +6449,101 @@ tsk_table_collection_check_edge_ordering(tsk_table_collection_t *self)
...
+        if (tk < M) {
+            right = TSK_MIN(right, edges.right[O[tk]]);
+        }
+
+        /* Tree is now ready. We look at each site on this tree in turn */
+        while (site < (tsk_id_t) sites.num_rows && sites.position[site] < right) {
+            while (mutation < mutations.num_rows && mutations.site[mutation] == site) {
+                mutation_time = mutations.time[mutation];
+                if (mutation_time == TSK_NULL) {
+                    ret = TSK_ERR_NULL_TIME;
```
petrelharp, line 6512: Gee, I at least think of TSK_NULL as the integer label -1, not a float. Should this instead check for negative times, and if not, say that they must be nonnegative?

c/tskit/tables.c
```diff
@@ -7566,9 +7686,10 @@ tsk_table_collection_deduplicate_sites(
         goto out;
     }
     /* Check everything except site duplicates (which we expect) and
-     * edge indexes (which we don't use) */
+     * edge indexes and mutation timesc (which we don't use) */
```
petrelharp, line 7689: "timesc"

docs/data-model.rst
```diff
@@ -384,6 +385,9 @@ The ``site`` column is an integer value defining the ID of the
 The ``node`` column is an integer value defining the ID of the
 first :ref:`node <sec_node_table_definition>` in the tree below this mutation.
 
+The  ``time`` column is a double precision floating point value recording when the
+mutation happened. This is the same time dimension as that for nodes.
```
petrelharp, line 389: How about just "recording how long ago the mutation happened", without the reference to the node's "time dimension"?

docs/data-model.rst
```diff
@@ -813,6 +820,15 @@ of this fact to compute the ``parent`` column of a mutation table, if all
 other information is valid.
 
 
+Computing mutation times
+------------------------
+
+In the case where the method generating a tree sequence does not generate mutation
+times, valid times can be provided by :meth:`TableCollection.compute_mutation_parents`.
+If all other information is valid this method will evenly distribute mutations at a
+site along their shared edges.
```
petrelharp, line 829: To clarify which sort of "evenly", how about "will assign times to the mutations by placing them at evenly spaced intervals along their edge (for instance, a single mutation on an edge between a node at time 1.0 and a node at time 4.0 would be given time 2.5; while two mutations on that edge would be given times 2.0 and 3.0)."

petrelharp:
Wow, that hits a lot of places! I've read through, and it looks great. Thanks!

Note: we should next ask sort tables to use the time column.

python/tests/__init__.py
```diff
@@ -20,11 +20,113 @@
...
+
+        tables_dict = lwt.asdict()
+        tables_dict["mutations"]["time"] = np.full(
+            tables_dict["mutations"]["node"].shape, -1, np.float64
+        )
+        tables = tskit.TableCollection.fromdict(tables_dict)
+        tables.provenances.add_row(json.dumps(provenance_dict))
+
+        tables.build_index()
+        tables.compute_mutation_times()
```
petrelharp, line 123: Is the idea that when msprime adds in times itself that this won't be needed? When that happens, a bunch of the checks that times can be exactly recomputed by compute_mutation_times will fail; maybe it's better to not assume that times coming in from mutate are regularly spaced?

benjeffery: Yeah, I copied this test structure from mutation parents. I thought mutation times might be configurable in msprime and that this test could specify what it wanted. But this this test isn't doing much meaningful testing though.

c/tskit/tables.c
benjeffery: I think checking for non-negative is the right thing to do, so will switch it to that.

python/tests/__init__.py
petrelharp: I'm not sure, but my guess is that adding mutations regularly along branches would be much harder in msprime than putting them uniformly (since you'd have to know the future to do regularly spaced). And, I can't think of a reason anyone would want that.

python/tests/__init__.py
```diff
@@ -20,11 +20,113 @@
...
+import sys
+
+import msprime
+import numpy as np
 
 import tskit
 from . import tsutil
 from .simplify import *  # NOQA
 
+if msprime.__version__ <= "0.7.4":
```
jeromekelleher, line 33: Is this robust to people running the development version of msprime? It'd be worth running the tests in a virtualenv to see. Quite a few devs are running the git version of msprime now.

python/tests/test_highlevel.py
```diff
@@ -743,6 +746,16 @@ def test_compute_mutation_parent(self):
             parent = ts.tables.mutations.parent
             self.assertTrue(np.array_equal(parent, before))
 
+    def test_compute_mutation_time(self):
+        for ts in get_example_tree_sequences():
```
jeromekelleher, line 750: This will fail, I think, when msprime generates real times.

python/tests/test_topology.py
```diff
@@ -5858,6 +5864,142 @@ def test_many_multiroot_trees_recurrent_mutations(self):
...
+    Tests that mutation time is correctly specified, and that we correctly
+    recompute it with compute_mutation_times.
+    """
+
+    seed = 42
+
+    def verify_times(self, ts):
+        time = tsutil.compute_mutation_times(ts)
+        tables = ts.tables
+        print(time - tables.mutations.time)
```
jeromekelleher, line 5878: Stray print here

python/tests/tsutil.py
```diff
@@ -539,6 +582,48 @@ def compute_mutation_parent(ts):
...
+    the times evenly over the edge.
+
+    :param TreeSequence ts: The tree sequence to compute for.  Need not
+        have a valid mutation time column.
+    """
+    tables = ts.dump_tables()
+    mutations = tables.mutations
+
+    mutations_above_node = collections.defaultdict(list)
+    for mut_idx, mut in enumerate(mutations):
```
jeromekelleher, line 598: I think mut_idx is available as ``mut.id``, so can simplify this a bit

python/tskit/cli.py
```diff
@@ -87,6 +87,11 @@ def run_upgrade(args):
             "will remove these duplicate positions. However, this will result in loss "
             "of data from the original file!"
         )
+    except OSError as e:
+        if "file signature not found" not in str(e):
+            raise e
+        else:
+            tskit.fix_legacy_kastore(args.source, args.destination)
```
jeromekelleher, line 94: Hmmm, shouldn't this be done in ``tskit.load_legacy``? We want this function to load kastore files and fix them up as well, not just run the upgrade in the CLI.

python/tskit/formats.py
```diff
@@ -252,6 +260,30 @@ def load_legacy(filename, remove_duplicate_positions=False):
     return ts
 
 
+def fix_legacy_kastore(src, dest):
+    with kastore.load(src) as store:
```
jeromekelleher, line 264: This should go into ``load_legacy``, so we can load a pre version 13. Should go something like
```
data = dict(store)
data["mutations/time"] = np.zeros(data["mutations/node"].shape,  dtype=np.float64)
tables = tskit.TableCollection.fromdict(**data) # or something
tables.compute_mutation_times()
return tables.tree_sequence()
```

python/tskit/tables.py
```diff
@@ -1367,22 +1371,23 @@ def _text_header_and_rows(self):
...
-                "{}\t{}\t{}\t{}\t{}\t{}".format(
-                    j, site[j], node[j], derived_state[j], parent[j], md
+                "{}\t{}\t{}\t{}\t{}\t{}\t{}".format(
+                    j, site[j], node[j], time[j], derived_state[j], parent[j], md
                 ).split("\t")
             )
         return headers, rows
 
-    def add_row(self, site, node, derived_state, parent=-1, metadata=None):
+    def add_row(self, site, node, time, derived_state, parent=-1, metadata=None):
```
jeromekelleher, line 1390: I'd have been inclined to put this after derived_state as an optional value with default 0 to avoid some breakage. I agree this is conceptually the right place for it though. Since you've gone through the pain of updating the test suite it's probably best to keep it as it is. I doubt we'll break *that* much downstream code...

python/tskit/tables.py
```diff
@@ -2213,6 +2227,23 @@ def compute_mutation_parents(self):
         self.ll_tables.compute_mutation_parents()
         # TODO add provenance
 
+    def compute_mutation_times(self):
+        """
+        Modifies the tables in place, computing valid values for the ``time`` column of
+        the mutation table. For this to work, the node and edge tables must be
+        valid, and the site and mutation tables must be sorted (see
+        :meth:`TableCollection.sort`).
```
jeromekelleher, line 2235: And indexed

c/tskit/tables.c
benjeffery: Ok didn't realise node times could be negative, I will remove `TSK_ERR_NULL_TIME` as an error completely, and -1 will just be a default where needed.

benjeffery:
> 1. I'm a bit concerned about the breakage caused by the Python MutationTable.add_row. Maybe it is the right thing to do to force users to come up with a time that makes sense rather than just calling compute_mutation_times, though.

My assumption is that there will be stats methods etc. that rely on mutation times so the actual change isn't "we added a parameter to `add_row`" but "we added a property that needs to make sense" so breakage is desired to bring people's attention to it. Using `compute_mutation_times` needs to be an active choice.

> 2. I don't think we should check the parent time in the table collection, and that should be done at tree sequence load time because it's a property of the tree topologies.

I also wasn't sure about this as it felt wrong to need to build indexes to check a table collection, but I wasn't clear on the distinction between "valid tables" and "valid tree sequence".

python/tskit/formats.py
benjeffery: I thought so too but was dissuaded by the comment in load_legacy: "This method is only intended to be used to read old format HDF5 files. "

c/tests/test_trees.c
```diff
@@ -3819,6 +3873,120 @@ test_single_tree_compute_mutation_parents(void)
...
+    ret = tsk_treeseq_init(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_sites(&ts), 4);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_mutations(&ts), 7);
+
+    /* Compute the mutation times */
+    verify_compute_mutation_times(&ts);
+
+    /* Verify consistency of individuals */
```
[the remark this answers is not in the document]
benjeffery: I didn't get it either was in original code: https://github.com/tskit-dev/tskit/blob/master/c/tests/test_trees.c#L3752 but I'll delete.

c/tskit/tables.c
```diff
@@ -6435,6 +6449,101 @@ tsk_table_collection_check_edge_ordering(tsk_table_collection_t *self)
...
+                    ret = TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE;
+                    goto out;
+                }
+                if (mutations.parent[mutation] != TSK_NULL
+                    && mutations.time[self->mutations.parent[mutation]]
+                           < mutation_time) {
+                    ret = TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION;
+                    goto out;
+                }
+                if (parent[mutations.node[mutation]] != TSK_NULL
```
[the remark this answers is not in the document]
benjeffery: We can check `TSK_ERR_TIME_NONFINITE`, `TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE` and `TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION` without building the trees, only `TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE` needs them. 
But I would argue that we check them at tree sequence init time. I initially put them here as mutation parents are checked here - but I now understand that ordering but mutation parents is a key sorting criteria, whereas time isn't.

c/tskit/tables.c
benjeffery: `TSK_ERR_NULL_TIME` removed in https://github.com/tskit-dev/tskit/pull/672/commits/06e1d3787436bec2823f8d92b18cfbd79715c0bc

c/tskit/tables.c
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/79fb3a06a109d1b5e37812859c26d33357d38a82

docs/data-model.rst
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/6296a8cee08b755e0fba43390e144bd96768c23e

docs/data-model.rst
benjeffery: Yeah an example really helps. Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/1514c8f4be60fad61b2bda4db12f056d5e6f7c65

python/tskit/formats.py
jeromekelleher: That comment is a remnant from when all the files were HDF5 and we just referred to the files as "HDF5" as a matter of course. I think load_legacy should cope with any old file format and return a tree sequence, making any choices that need to be made to bring it up to date.

c/tskit/tables.c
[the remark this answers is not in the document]
jeromekelleher: I think it's good to check any stuff that we can check as part of the table validation step, up to the point where we need the indexes and have to build trees.

jeromekelleher:
> Using compute_mutation_times needs to be an active choice.

I agree, you've made the right choice here.

python/tests/__init__.py
benjeffery: Tests now don't use msprime times, only msprime topology in https://github.com/tskit-dev/tskit/pull/672/commits/93e65c6ddc6f4768c1ac07576e62979bb5a04604

c/tests/test_trees.c
```diff
@@ -128,6 +128,29 @@ verify_compute_mutation_parents(tsk_treeseq_t *ts)
...
+    size_t size = tsk_treeseq_get_num_mutations(ts) * sizeof(tsk_id_t);
+    tsk_id_t *time = malloc(size);
+    tsk_table_collection_t tables;
+
+    CU_ASSERT_FATAL(time != NULL);
+    ret = tsk_treeseq_copy_tables(ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    memcpy(time, tables.mutations.time, size);
+    /* Make sure the tables are actually updated */
+    memset(tables.mutations.time, 0xff, size);
```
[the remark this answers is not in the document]
benjeffery: Lazy copy paste, fixed in https://github.com/tskit-dev/tskit/pull/672/commits/ce98ac966a587761a7336c88af839c099cb9f832

c/tests/test_trees.c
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/5d94b62f05000940d4840cf2ed01e853ddd1ab8f

c/tskit/tables.c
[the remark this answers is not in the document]
petrelharp: I agree. I was vaguely worried about this also.

c/tskit/tables.c
petrelharp: We do? Oh, that is vaguely familiar. Happliy, our [documentation agrees](https://tskit.readthedocs.io/en/latest/data-model.html#node-requirements).

c/tskit/tables.c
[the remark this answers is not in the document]
benjeffery: In https://github.com/tskit-dev/tskit/pull/672/commits/42b8a7b32ac8c07ade900a7731d766f3aaa2fc85 I've moved the parent check into the tree iteration in `tsk_treeseq_init_trees`.

c/tskit/tables.c
```diff
@@ -7730,6 +7852,109 @@ tsk_table_collection_compute_mutation_parents(
     return ret;
 }
 
+int TSK_WARN_UNUSED
+tsk_table_collection_compute_mutation_times(
+    tsk_table_collection_t *self, double *TSK_UNUSED(options))
+{
+    int ret = 0;
+    const tsk_id_t *I, *O;
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/9ee98a7ed31d70db151b0b980c79497278086384

c/tskit/tables.c
```diff
@@ -7730,6 +7852,109 @@ tsk_table_collection_compute_mutation_parents(
...
+    const tsk_id_t M = (tsk_id_t) edges.num_rows;
+    tsk_id_t tj, tk;
+    tsk_id_t *parent = NULL;
+    tsk_size_t *numerator = NULL;
+    tsk_size_t *denominator = NULL;
+    tsk_id_t u;
+    double left, right;
+    tsk_id_t site;
+    /* Using unsigned values here avoids potentially undefined behaviour */
+    uint32_t j, mutation, first_mutation;
```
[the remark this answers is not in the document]
benjeffery: `tsk_table_collection_compute_mutation_parents` also used that type, so fixed in both: https://github.com/tskit-dev/tskit/pull/672/commits/8add75efdc43ce83b6c2365dd6fec3573fbbfd37

c/tskit/tables.c
```diff
@@ -7730,6 +7852,109 @@ tsk_table_collection_compute_mutation_parents(
...
+            }
+            /* Go over the mutations again assigning times. As the sorting requirements
+               guarantee that parents are before children, we assign oldest first */
+            for (j = first_mutation; j < mutation; j++) {
+                u = mutations.node[j];
+                numerator[u]++;
+                if (parent[u] == TSK_NULL) {
+                    /* This mutation is above a root */
+                    mutations.time[j] = nodes.time[u];
+                } else {
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/6f508802895082d39a90f1e926a6ccbbe0f08bfb

c/tskit/tables.c
```diff
@@ -7730,6 +7852,109 @@ tsk_table_collection_compute_mutation_parents(
     return ret;
 }
 
+int TSK_WARN_UNUSED
+tsk_table_collection_compute_mutation_times(
+    tsk_table_collection_t *self, double *TSK_UNUSED(options))
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/b61c34c1905b254ac68d6b56d86810cb31cde528

docs/data-model.rst
```diff
@@ -1129,18 +1145,19 @@ Mutation text format
 ====================
 
 The mutation text format must contain the columns ``site``,
-``node`` and ``derived_state``. The ``parent`` and ``metadata`` columns
+``node`` and ``derived_state``. The ``time``, ``parent`` and ``metadata`` columns
 may also be optionally present (but ``parent`` must be specified if
-more than one mutation occurs at the same site). See the
+more than one mutation occurs at the same site). If ``time`` is absent ``-1`` will be
```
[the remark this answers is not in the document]
benjeffery: https://github.com/tskit-dev/tskit/pull/672/commits/b98ba5aac7d8fc8c29108b9ea14422b869cba4eb

python/CHANGELOG.rst
```diff
@@ -6,6 +6,14 @@ In development
 
 **Breaking changes**
 
+- Mutations now have a double-precision floating-point ``time`` column. As this is a
+  mandatory column the file format major version has been bumped to ``13.0``.
```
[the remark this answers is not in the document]
benjeffery: https://github.com/tskit-dev/tskit/pull/672/commits/54d2161e30d00c488d9c584e70a53b4e35451118

python/tests/__init__.py
```diff
@@ -20,11 +20,113 @@
...
+    # As mutation time is introduced into tskit but not msprime we need to patch up
+    # msprime to return times
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/54d2161e30d00c488d9c584e70a53b4e35451118

python/tests/__init__.py
benjeffery: The tests fail with a dev version of msprime. I coded up https://github.com/benjeffery/msprime/tree/mutation_time_hack to fix this on the msprime end, but discovered the tskit tests already fail with a dev msprime see below, so maybe not worth the hack for now.
```
FAIL: test_draw_integer_breaks_ts (tests.test_drawing.TestDrawSvg)
  File "/home/benj/projects/tskit/python/tests/test_drawing.py", line 1774, in test_draw_integer_breaks_ts
    self.assertNotEqual(svg.find(f">{b:.0f}<"), -1)
AssertionError: -1 == -1
-------------------- >> begin captured logging << --------------------
msprime.ancestry: DEBUG: Setting initial model {'name': 'hudson'}
msprime.ancestry: DEBUG: Running simulation until maximum: inf
--------------------- >> end captured logging << ---------------------

FAIL: test_integer_edges_all_sites (tests.test_topology.TestReduceTopology)
  File "/home/benj/projects/tskit/python/tests/test_topology.py", line 6526, in test_integer_edges_all_sites
    ts = self.get_integer_edge_ts(5, 10)
  File "/home/benj/projects/tskit/python/tests/test_topology.py", line 6514, in get_integer_edge_ts
    self.assertEqual(int(edge.left), edge.left)
AssertionError: 6 != 6.016224464401603
-------------------- >> begin captured logging << --------------------
msprime.ancestry: DEBUG: Setting initial model {'name': 'hudson'}
msprime.ancestry: DEBUG: Running simulation until maximum: inf
--------------------- >> end captured logging << ---------------------

FAIL: test_integer_edges_one_site (tests.test_topology.TestReduceTopology)
  File "/home/benj/projects/tskit/python/tests/test_topology.py", line 6519, in test_integer_edges_one_site
    ts = self.get_integer_edge_ts(5, 10)
  File "/home/benj/projects/tskit/python/tests/test_topology.py", line 6514, in get_integer_edge_ts
    self.assertEqual(int(edge.left), edge.left)
AssertionError: 6 != 6.016224464401603
-------------------- >> begin captured logging << --------------------
msprime.ancestry: DEBUG: Setting initial model {'name': 'hudson'}
msprime.ancestry: DEBUG: Running simulation until maximum: inf
--------------------- >> end captured logging << ---------------------
```

python/tests/test_highlevel.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/93e65c6ddc6f4768c1ac07576e62979bb5a04604

python/tests/test_topology.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/93e65c6ddc6f4768c1ac07576e62979bb5a04604

python/tests/tsutil.py
benjeffery: Sadly it's a `MutationTableRow` so no `id`.

c/tskit/tables.c
```diff
@@ -7730,6 +7765,114 @@ tsk_table_collection_compute_mutation_parents(
     return ret;
 }
 
+int TSK_WARN_UNUSED
+tsk_table_collection_compute_mutation_times(tsk_table_collection_t *self, double *random)
```
jeromekelleher, line 7769: I think we probably want to have an options parameter here as well - we'll regret it if we don't.

python/tskit/formats.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/f5cf3d2184e834f5e07b8777866b399d37d82ac3

python/tskit/tables.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/9d255cd75c2be687535eb6333f64a09e4aa597ce

c/tskit/tables.c
benjeffery: https://github.com/tskit-dev/tskit/pull/672/commits/bbf936f6b40e7e38c4e831dddafe59042a34eba5

benjeffery:
Ok, I think all comments are addressed.
Tests will pass once this is rebased onto #687, but I don't want to invalidate all the comments till they are resolved.

python/tests/__init__.py
jeromekelleher: Didn't see this comment sorry - GitHub is annoying! Once you comment on a PR it keeps updating on every tiny change that's made, so I mute it. But it appears this mutes the responses too.

jeromekelleher: These tests are pretty trivial, it's just using some deprecated msprime APIs around discrete sites. Nothing to get  worried about.

python/tests/tsutil.py
jeromekelleher: ah.

c/tskit/tables.c
```diff
@@ -2542,8 +2542,14 @@ tsk_mutation_table_append_columns(tsk_mutation_table_t *self, tsk_size_t num_row
     } else {
         memcpy(self->parent + self->num_rows, parent, num_rows * sizeof(tsk_id_t));
     }
-    memcpy(self->time + self->num_rows, time, num_rows * sizeof(double));
-
+    if (time == NULL) {
+        /* If time is NULL, set all times to NAN which is the default */
```
jeromekelleher, line 2546: We need to be careful here - need to make sure it's a non-signalling NaN, or else weird things will happen on different platforms.

c/tskit/core.c
```diff
@@ -289,15 +289,15 @@ tsk_strerror_internal(int err)
             break;
         case TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE:
             ret = "Mutations must have a time equal to or greater than that of their "
-                  "node";
+                  "node or equal to NAN";
```
jeromekelleher, line 292: It's pedantic, but nothing is equal to NaN AFAIK. More correct to say "... of their or be a NaN value."

c/tskit/tables.c
benjeffery: My understanding here from https://en.wikipedia.org/wiki/IEEE_754 is that anything with exponent == 0xFF and fraction != 0 is a NaN and that the highest fraction bit (bit 22) is `1` for a quiet NaN and `0` for signalling. I think the best thing to do is put in some asserts (hopefully compile time) that 0xFF*8 is a non signalling NaN.

benjeffery:
@jeromekelleher In a lot of places we check for table equality, this will fail for mutations that have the default value. My current plan is to override equality methods to make NAN==NAN, we'll see how that goes.

benjeffery:
This then gets more complex - we'd like two mutations that have "default" time to be equal. It would be simplest to keep using `memcmp` for table equality so the C and Python default times both need to be the same bit field. Thankfully it seems that (bitfield-wise) `math.nan` == `np.nan` ==  C `NAN`. For the record this is `011111111111100000...zeros...0` which for python is hard-coded [here](https://github.com/python/cpython/blob/e9684fac5a158be9806304a676e619857520a4dc/Python/dtoa.c#L282) for numpy hard-coded [here](https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/npy_math.h#L40) (it's the same value even though the hex is different as it is cast to double after). The C99 spec only specifies that the `NAN` macro is a quiet nan, but we can assert that it matches.

Some python operations generate NANs with a different sign bit to `math.nan` but I don't think that is a problem as we don't want those NANs to be equal. The operations that generate NANs with the same sign bit are problematic as they will appear equal to the default so will be valid but are likely due to error, ensuring no NANs before times are used will be essential.

 So the way forward is to fill the C array with `NAN` and not `0xFF` by default.

c/tskit/tables.c
jeromekelleher: We could do a loop here and assign manually to a NAN macro? Might be simplest in this case. We can use the NAN value defined in GNU C, and make sure it's defined otherwise.

jeromekelleher:
> So the way forward is to fill the C array with NAN and not 0xFF by default.

Good, agreed. I think we can be pragmatic here and use memcmp when checking for equality. It's fine if we say the tables are only equal if they both contain exactly the *same* NAN.

The Python default times can be set in the Python C API, so we should be able to use the same macro everywhere to set the default value. Come to think of it, it'd be better if we defined our own NaN value for TSK_MUTATION_TIME_UNKNOWN so that we don't get hit by different platforms using different bit values for the NAN macro. We would want a set of tables generated on Windows to be the same as one generated on Linux.

jeromekelleher:
We could be even stricter actually, and only consider this one specific NaN, TSK_MUTATION_TIME_UNKNOWN, as acceptable input, and any other NaN throws an error. We'd probably need to do some fiddling like making a union to that we can compare the actual bits, but it should work.

What do you think? It would solve some problems.

benjeffery:
I think that's right - to have a macro with a specific NaN. The fraction bits are free to use and not set usually - we could define our own bit pattern for TSK_MUTATION_TIME_UNKNOWN so that adding a row with np.nan or any other NaN would fail?

jeromekelleher:
> I think that's right - to have a macro with a specific NaN. The fraction bits are free to use and not set usually - we could define our own bit pattern for TSK_MUTATION_TIME_UNKNOWN so that adding a row with np.nan or any other NaN would fail?

That seems like a good idea. No room for ambiguity then, but NaNs will still propagate in a sensible way if people try to do calculations on them directly. Excellent.

benjeffery:
Cool, we'll go with:
```python
>>> struct.unpack(">d", b'\x7f\xf8tskit!')
(nan,)
```

benjeffery:
@jeromekelleher ok, I think this is ready for review. We have Python mutation equality methods so I needed to bring the default value through to Python. I considered replacing the NAN with None at the CPython layer, but think that's potentially confusing and bad perf.

jeromekelleher:
Great, thanks @benjeffery. I'll go through ASAP.

c/CHANGELOG.rst
```diff
@@ -6,6 +6,13 @@ In development.
 
 **Breaking changes**
 
+- Mutations now have a double-precision floating-point ``time`` column. This is
+  optional in the file format if it is missing ``NAN`` is used for all mutations.
```
jeromekelleher, line 10:
```suggestion
- Mutations now have an optional double-precision floating-point ``time`` column. If not specified, this defaults to a particular NaN value indicating that the time is unknown.
```

jeromekelleher: Also, I guess this isn't a breaking change any more - move to new features?

c/tskit/core.c
```diff
@@ -287,6 +287,18 @@ tsk_strerror_internal(int err)
         case TSK_ERR_UNSORTED_MUTATIONS:
             ret = "Mutations must be provided in non-decreasing site order";
             break;
+        case TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE:
+            ret = "Mutations must have a time equal to or greater than that of their "
```
jeromekelleher, line 291: This doesn't quite parse (it's implying the time must be greater than TSK_MUTATION_UNKNOWN_TIME). How about "A mutation's time must be >= the node time, or be marked as 'unknown'".

This error message will be shown in Python and C, so we don't want to use the C macro.

The other messages aren't quite right either.

c/tskit/core.h
```diff
@@ -67,6 +69,33 @@ extern "C" {
 #define TSK_DBL_DECIMAL_DIG (DBL_DIG + 3)
 #endif
 
+/* We define a specific NAN value for default mutation time which indicates
+ * the time is unknown. We use a specific value so that if mutation time is set to
+ * a NAN from a computation we can reject it.
+ */
```
jeromekelleher, line 75: We should note here where we got the bit pattern from - people will wonder where we plucked this specific value from otherwise.

c/tskit/core.h
```diff
@@ -67,6 +69,33 @@ extern "C" {
...
+#define TSK_MUTATION_UNKNOWN_TIME_HEX 0x7FF874736B697421ULL
+static inline double
+__tsk_nan_f(void)
+{
+    const union {
+        uint64_t i;
+        double f;
+    } nan_union = { TSK_MUTATION_UNKNOWN_TIME_HEX };
```
jeromekelleher, line 83: This is safe, I think, but I had to look it up. Maybe it's a bit more explicit if we use this:
```c
const union {
￼        uint64_t i;
￼        double f;
￼    } nan_union = { .i=TSK_MUTATION_UNKNOWN_TIME_HEX };
```

c/tskit/core.h
```diff
@@ -67,6 +69,33 @@ extern "C" {
...
+{
+    union {
+        uint64_t i;
+        double f;
+    } nan_union;
+    nan_union.f = val;
+    return nan_union.i == TSK_MUTATION_UNKNOWN_TIME_HEX;
+}
+#define TSK_MUTATION_UNKNOWN_TIME __tsk_nan_f()
+#define TSK_IS_MUTATION_UNKNOWN_TIME(val) __tsk_is_nan_f(val)
```
jeromekelleher, line 97: Should we just make this a function altogether? ``is_mutation_unknown_time(val)`` seems a bit more natural. Or, to put it another way, is there any particular reason for making this a function-like macro rather than just function?

python/CHANGELOG.rst
```diff
@@ -6,6 +6,12 @@ In development
 
 **Breaking changes**
 
+- Mutations now have a double-precision floating-point ``time`` column. This is
```
jeromekelleher, line 9: Same comment as the C changelog.

python/tests/test_util.py
```diff
@@ -23,13 +23,26 @@
...
 import itertools
+import math
 import pickle
 import unittest
 
 import numpy as np
 
 import tests.tsutil as tsutil
 import tskit.util as util
+from tskit import MUTATION_UNKNOWN_TIME
```
jeromekelleher, line 34: I can see the rationale for MUTATION_UNKNOWN_TIME in C, but it seems a bit clunky in Python. Maybe we could use ``UNKNOWN_TIME`` in Python? I guess in general, if we every need to specify another unknown time we'd use this value, so is there any avantage to TSK_MUTATION_UNKNOWN_TIME vs TSK_UNKNOWN_TIME in C either?

python/tskit/tables.py
```diff
@@ -1388,22 +1392,31 @@ def _text_header_and_rows(self):
...
 
-    def add_row(self, site, node, derived_state, parent=-1, metadata=None):
+    def add_row(
+        self,
+        site,
+        node,
+        derived_state,
+        parent=-1,
+        metadata=None,
+        time=tskit.MUTATION_UNKNOWN_TIME,
```
jeromekelleher, line 1418: My only worry with this is how it'll render on RTD. In the best case, it'll render as "NaN", which will give the false impression that any NaN can be provided as the default value.

I think it's a bit safer to make the default ``None``, and just transform it before calling the ll_table version. The perf hit is trivial - anyone calling ``add_row`` in Python has said goodbye to decent perf anyway.

python/tskit/trees.py
```diff
@@ -366,11 +370,33 @@ def __init__(
...
         self.derived_state = derived_state
         self.parent = parent
         self._encoded_metadata = encoded_metadata
         self._metadata_decoder = metadata_decoder
 
+    def __eq__(self, other):
+        # We need to remove metadata and the decoder so we are just comparing
+        # the encoded metadata, along with the other attributes.
+        # We also need to remove time as have to compare to unknown time.
+        other_ = {**other.__dict__}
```
jeromekelleher, line 383: Maybe ``other_ = copy.copy(other)`` is a bit more readable?

c/CHANGELOG.rst
benjeffery: Still the breaking change to `add_row` but the rest is moved in https://github.com/tskit-dev/tskit/pull/672/commits/676d83276d06a979f2a3f5e840c974928e193fd4

c/tskit/core.c
benjeffery: Fixed https://github.com/tskit-dev/tskit/pull/672/commits/6180ad53b8098de42fa9b6c01ef77a0448d36d26

c/tskit/core.h
benjeffery: Fixed https://github.com/tskit-dev/tskit/pull/672/commits/f42d90850f749822050bc896ed2a76cd28b911c4

c/tskit/core.h
benjeffery: Fixed https://github.com/tskit-dev/tskit/pull/672/commits/ba7157c861afb08210407550bfc73a9bc84beeb7

c/tskit/core.h
benjeffery: I think I wanted to keep it all together - but yes no need to be a macro. Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/9b742c8bc1e5fb298dff844af0b56a9975ea605f

python/CHANGELOG.rst
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/532f50baa5f427a5b8b80058ba00ca0a29f52f63

python/tests/test_util.py
benjeffery: Agreed, reads better as just unknown_time Changed in https://github.com/tskit-dev/tskit/pull/672/commits/202d5df081c31732a47e8fc344678e6b7d828a17

python/tskit/tables.py
benjeffery: Ha, good catch it rendered as `<MagicMock id='140509585067552'>` !
Fixed in https://github.com/tskit-dev/tskit/pull/672/commits/f2a1686148665737c0857c31a6190aaa5ed9e01f

python/tskit/trees.py
benjeffery: Needs to be `other_ = copy.copy(other).__dict__` but is clearer, changed in https://github.com/tskit-dev/tskit/pull/672/commits/bc4041b63c354897d8f4c0d7538bd80592880beb

benjeffery:
@jeromekelleher Addressed comments. When I squash I'll keep the first commit separate as that was the initial proposal.

jeromekelleher:
LGTM, let's merge!

benjeffery:
unnamed

Squashed to 2 commits!

jeromekelleher:
Let's remember how hard this was the next time we start talking about adding new columns to the tables!

PR #676, opened 2020-06-10
Refer to CoC in dev and point to inclusive code.

jeromekelleher:
This seems like a good place to mention the code of conduct and to set expectations for inclusive code and documentation.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -12,6 +12,13 @@ opening an `issue <http://github.com/tskit-dev/tskit/issues>`_ or
 `pull request <http://github.com/tskit-dev/tskit/pulls>`_ if you
 see any problems.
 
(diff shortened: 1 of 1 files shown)


petrelharp:
"gender neutral" yes; "racially neutral" is less clear, because (a) why would our docs refer to race?  (b) which definition of  race, anyhow? Oh, like if the documentation treats eg white europeans as the default 'population' for comparison? Or contrasts them to 'more diverse world populations' or something like that? Good point, yes. But how about 
  " and in particular to be neutral regarding gender, race, and geographic origin."
... ?

jeromekelleher:
I was thinking more about things like whitelist/blacklist etc, and followed the headings in Chromium's docs that are linked. Very happy to change though.

PR #679, opened 2020-06-15
C implementation of IBD-finding code.

gtsambos:
At the moment, just contains skeleton code to show the expected structure.

@jeromekelleher @petrelharp @dvukcevic

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -2852,6 +2852,147 @@ test_link_ancestors_multiple_to_single_tree(void)
 }
 
 static void
+test_ibd_finder(void)
+{
+    int ret;
+    // int j;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = { 0, 1 };
+    tsk_ibd_finder_t ibd_finder;
+
+    tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges, NULL, NULL,
+        NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 2, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    tsk_ibd_finder_free(&ibd_finder);
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_ibd_finder_multiple_trees(void)
+{
+    int ret;
+    // int j;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_ibd_finder_t ibd_finder;
+    // tsk_segment_t *seg;
+
+    tsk_treeseq_from_text(&ts, 2, multiple_tree_ex_nodes, multiple_tree_ex_edges, NULL,
+        NULL, NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    tsk_ibd_finder_free(&ibd_finder);
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_ibd_finder_min_length_max_time(void)
+{
+    int ret;
+    // int j;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_ibd_finder_t ibd_finder;
+    // tsk_segment_t *seg;
+
+    tsk_treeseq_from_text(&ts, 2, multiple_tree_ex_nodes, multiple_tree_ex_edges, NULL,
+        NULL, NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.5, 3.0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    tsk_ibd_finder_free(&ibd_finder);
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_ibd_finder_errors(void)
+{
+    int ret;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_id_t samples1[] = { 0, 0 };
+    tsk_id_t samples2[] = { -1, 1 };
+    tsk_id_t samples3[] = { 0 };
+    tsk_ibd_finder_t ibd_finder;
+
+    tsk_treeseq_from_text(&ts, 2, multiple_tree_ex_nodes, multiple_tree_ex_edges, NULL,
+        NULL, NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Duplicate sample IDs
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples1, 2, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_DUPLICATE_SAMPLE);
+    tsk_ibd_finder_free(&ibd_finder);
+
+    // Invalid sample IDs
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples2, 2, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
+    tsk_ibd_finder_free(&ibd_finder);
+
+    // Only 1 sample
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples3, 1, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NO_SAMPLE_PAIRS);
+    tsk_ibd_finder_free(&ibd_finder);
+
+    // Bad length or time
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.0, -1);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_PARAM_VALUE);
+
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_ibd_finder_odd_topologies(void)
+{
+    int ret;
+    // int j;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+    tsk_id_t samples[] = { 0, 1 };
+    tsk_id_t samples1[] = { 0, 2 };
+    tsk_ibd_finder_t ibd_finder;
+
+    tsk_treeseq_from_text(
+        &ts, 1, odd_tree1_ex_nodes, odd_tree1_ex_edges, NULL, NULL, NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // Multiple roots.
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 2, 0, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_ibd_finder_free(&ibd_finder);
+
+    // Parent is a sample.
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples1, 2, 0, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_GENERIC); // Currently fails
+    tsk_ibd_finder_free(&ibd_finder);
+
+    tsk_table_collection_free(&tables);
+    tsk_treeseq_free(&ts);
+}
+
+static void
 test_simplify_tables_drops_indexes(void)
 {
     int ret;
@@ -4627,6 +4768,11 @@ main(int argc, char **argv)
         { "test_link_ancestors_multiple_to_single_tree",
             test_link_ancestors_multiple_to_single_tree },
         { "test_sort_tables_offsets", test_sort_tables_offsets },
+        { "test_ibd_finder", test_ibd_finder },
+        { "test_ibd_finder_multiple_trees", test_ibd_finder_multiple_trees },
+        { "test_ibd_finder_min_length_max_time", test_ibd_finder_min_length_max_time },
+        { "test_ibd_finder_odd_topologies", test_ibd_finder_odd_topologies },
+        { "test_ibd_finder_errors", test_ibd_finder_errors },
         { "test_sort_tables_drops_indexes", test_sort_tables_drops_indexes },
         { "test_sort_tables_edge_metadata", test_sort_tables_edge_metadata },
         { "test_sort_tables_no_edge_metadata", test_sort_tables_no_edge_metadata },
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -2855,20 +2855,42 @@ static void
 test_ibd_finder(void)
 {
     int ret;
-    // int j;
+    int j, k;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
     tsk_id_t samples[] = { 0, 1 };
     tsk_ibd_finder_t ibd_finder;
+    double true_left[] = { 0.0 };
+    double true_right[] = { 1.0 };
+    tsk_id_t true_node[] = { 4 };
+    tsk_segment_t *seg = NULL;
 
+    // Read in the tree sequence.
     tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges, NULL, NULL,
         NULL, NULL, NULL, 0);
     ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 2, 0.0, 0.0);
+    // Run ibd_finder.
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 1, 0.0, 0.0);
+
+    // Check the output.
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
+    for (j = 0; j < (int) ibd_finder.num_pairs; j++) {
+        tsk_ibd_finder_get_ibd_segments(&ibd_finder, j, &seg);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        k = 0;
+        while (seg != NULL) {
+            CU_ASSERT_EQUAL_FATAL(seg->left, true_left[k]);
+            CU_ASSERT_EQUAL_FATAL(seg->right, true_right[k]);
+            CU_ASSERT_EQUAL_FATAL(seg->node, true_node[k]);
+            k++;
+            seg = seg->next;
+        }
+    }
+
+    // Free.
     tsk_ibd_finder_free(&ibd_finder);
     tsk_table_collection_free(&tables);
     tsk_treeseq_free(&ts);
@@ -2878,21 +2900,41 @@ static void
 test_ibd_finder_multiple_trees(void)
 {
     int ret;
-    // int j;
+    int j, k;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
-    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_id_t samples[] = { 0, 1, 0, 2 };
     tsk_ibd_finder_t ibd_finder;
-    // tsk_segment_t *seg;
+    double true_left[2][2] = { { 0.0, 0.7 }, { 0.7, 0.0 } };
+    double true_right[2][2] = { { 0.7, 1.0 }, { 1.0, 0.7 } };
+    double true_node[2][2] = { { 4, 5 }, { 5, 6 } };
+    tsk_segment_t *seg = NULL;
 
+    // Read in the tree sequence.
     tsk_treeseq_from_text(&ts, 2, multiple_tree_ex_nodes, multiple_tree_ex_edges, NULL,
         NULL, NULL, NULL, NULL, 0);
     ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.0, 0.0);
+    // Run ibd_finder.
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 2, 0.0, 0.0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
+    // Check the output.
+    for (j = 0; j < (int) ibd_finder.num_pairs; j++) {
+        ret = tsk_ibd_finder_get_ibd_segments(&ibd_finder, j, &seg);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        k = 0;
+        while (seg != NULL) {
+            CU_ASSERT_EQUAL_FATAL(seg->left, true_left[j][k]);
+            CU_ASSERT_EQUAL_FATAL(seg->right, true_right[j][k]);
+            CU_ASSERT_EQUAL_FATAL(seg->node, true_node[j][k]);
+            k++;
+            seg = seg->next;
+        }
+    }
+
+    // Free.
     tsk_ibd_finder_free(&ibd_finder);
     tsk_table_collection_free(&tables);
     tsk_treeseq_free(&ts);
@@ -2902,21 +2944,44 @@ static void
 test_ibd_finder_min_length_max_time(void)
 {
     int ret;
-    // int j;
+    int j, k;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
-    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_id_t samples[] = { 0, 1, 1, 2, 2, 0 };
     tsk_ibd_finder_t ibd_finder;
-    // tsk_segment_t *seg;
+    double true_left[3][1] = { { 0.0 }, { -1 }, { -1 } };
+    double true_right[3][1] = { { 0.7 }, { -1 }, { -1 } };
+    double true_node[3][1] = { { 4 }, { -1 }, { -1 } };
+    tsk_segment_t *seg = NULL;
 
+    // Read in the tree sequence.
     tsk_treeseq_from_text(&ts, 2, multiple_tree_ex_nodes, multiple_tree_ex_edges, NULL,
         NULL, NULL, NULL, NULL, 0);
     ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.5, 3.0);
+    // Run ibd_finder.
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 3, 0.5, 3.0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
+    // Check the output.
+    for (j = 0; j < (int) ibd_finder.num_pairs; j++) {
+        ret = tsk_ibd_finder_get_ibd_segments(&ibd_finder, j, &seg);
+        CU_ASSERT_TRUE_FATAL((ret == 0) || (ret == -1));
+        if (ret == -1) {
+            continue;
+        }
+        k = 0;
+        while (seg != NULL) {
+            CU_ASSERT_EQUAL_FATAL(seg->left, true_left[j][k]);
+            CU_ASSERT_EQUAL_FATAL(seg->right, true_right[j][k]);
+            CU_ASSERT_EQUAL_FATAL(seg->node, true_node[j][k]);
+            k++;
+            seg = seg->next;
+        }
+    }
+
+    // Free.
     tsk_ibd_finder_free(&ibd_finder);
     tsk_table_collection_free(&tables);
     tsk_treeseq_free(&ts);
@@ -2928,8 +2993,7 @@ test_ibd_finder_errors(void)
     int ret;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
-    tsk_id_t samples[] = { 0, 1, 2 };
-    tsk_id_t samples1[] = { 0, 0 };
+    tsk_id_t samples[] = { 0, 1, 2, 0 };
     tsk_id_t samples2[] = { -1, 1 };
     tsk_id_t samples3[] = { 0 };
     tsk_ibd_finder_t ibd_finder;
@@ -2939,24 +3003,23 @@ test_ibd_finder_errors(void)
     ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    // Duplicate sample IDs
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples1, 2, 0.0, 0.0);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_DUPLICATE_SAMPLE);
-    tsk_ibd_finder_free(&ibd_finder);
-
     // Invalid sample IDs
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples2, 2, 0.0, 0.0);
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples2, 1, 0.0, 0.0);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
     tsk_ibd_finder_free(&ibd_finder);
 
     // Only 1 sample
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples3, 1, 0.0, 0.0);
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples3, 0, 0.0, 0.0);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NO_SAMPLE_PAIRS);
     tsk_ibd_finder_free(&ibd_finder);
 
     // Bad length or time
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 3, 0.0, -1);
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 2, 0.0, -1);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_PARAM_VALUE);
+    tsk_ibd_finder_free(&ibd_finder);
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 2, -1, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_PARAM_VALUE);
+    tsk_ibd_finder_free(&ibd_finder);
 
     tsk_table_collection_free(&tables);
     tsk_treeseq_free(&ts);
@@ -2970,7 +3033,7 @@ test_ibd_finder_odd_topologies(void)
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
     tsk_id_t samples[] = { 0, 1 };
-    tsk_id_t samples1[] = { 0, 2 };
+    // tsk_id_t samples1[] = { 0, 2 };
     tsk_ibd_finder_t ibd_finder;
 
     tsk_treeseq_from_text(
@@ -2979,14 +3042,14 @@ test_ibd_finder_odd_topologies(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     // Multiple roots.
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 2, 0, 0);
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples, 1, 0, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     tsk_ibd_finder_free(&ibd_finder);
 
-    // Parent is a sample.
-    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples1, 2, 0, 0);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_GENERIC); // Currently fails
-    tsk_ibd_finder_free(&ibd_finder);
+    // // Parent is a sample.
+    // ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, samples1, 1, 0, 0);
+    // CU_ASSERT_EQUAL_FATAL(ret, 0); // This currently produces wrong output
+    // tsk_ibd_finder_free(&ibd_finder);
 
     tsk_table_collection_free(&tables);
     tsk_treeseq_free(&ts);
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -25,6 +25,7 @@
 #include "testlib.h"
 #include <tskit/tables.h>
 
+#include <float.h>
 #include <unistd.h>
 #include <stdlib.h>
 
@@ -2862,7 +2863,6 @@ ibd_finder_init_and_run(tsk_ibd_finder_t *ibd_finder, tsk_table_collection_t *ta
     if (ret != 0) {
         goto out;
     }
-
     ret = tsk_ibd_finder_set_min_length(ibd_finder, min_length);
     if (ret != 0) {
         goto out;
@@ -2871,7 +2871,6 @@ ibd_finder_init_and_run(tsk_ibd_finder_t *ibd_finder, tsk_table_collection_t *ta
     if (ret != 0) {
         goto out;
     }
-
     ret = tsk_ibd_finder_run(ibd_finder);
     if (ret != 0) {
         goto out;
@@ -2901,8 +2900,7 @@ test_ibd_finder(void)
     ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    // Run ibd_finder.
-    ret = ibd_finder_init_and_run(&ibd_finder, &tables, samples, 1, 0.0, 0.0);
+    ret = ibd_finder_init_and_run(&ibd_finder, &tables, samples, 1, 0.0, DBL_MAX);
 
     // Check the output.
     CU_ASSERT_EQUAL_FATAL(ret, 0);
@@ -2948,7 +2946,7 @@ test_ibd_finder_multiple_trees(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     // Run ibd_finder.
-    ret = ibd_finder_init_and_run(&ibd_finder, &tables, samples, 2, 0.0, 0.0);
+    ret = ibd_finder_init_and_run(&ibd_finder, &tables, samples, 2, 0.0, DBL_MAX);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     // Check the output.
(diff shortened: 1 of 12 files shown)


jeromekelleher:
This all looks good to me @gtsambos. To make the memory management practical, I think a good way to structure things would be to have client code that looks something like this:
```c
tsk_ibd_finder_t ibd_finder;
tsk_segment_t *seg;

int ret = tsk_ibd_finder_alloc(&ibd_finder, /* other params... */);
// error check ret
ret  = tsk_ibd_finder_run(&ibd_finder, /* other params... */);
// error check ret
for (j = 0; j < ibd_finder.num_pairs; j++) {
    ret = ibd_finder_get_pair_from_index(&ibd_finder, j, &a, &b);
    // error check ret;
    printf("IBD segs for (%d, %d) = ", a, b);
    for (seg = ibd_finder.ibd_segments[j]; seg != NULL; seg = seg->next) {
          /* Not sure if these are the correct fields, this is just to give a rough idea */
          printf("(%f, %f, %d)", seg->left, seg->right, seg->node);
    }
    printf("\n");
}
ibd_finder_free(&ibd_finder);
```

So, we'd make the ``tsk_ibd_finder`` part of the public API (so the struct is defined in tables.h), and have client code access some of the fields inside this struct directly. This is not ideal, and in particular the big ``ibd_segments`` array will use a lot of memory. However, it will give us working code and a way to move forward. 

Once this version is working, we can investigate more memory efficient sparse ways to represent the results, and can think about a final documented C API then.

Does this make sense?

gtsambos:
Yes, thanks Jerome. I'll start by making a very rudimentary `ibd_finder` struct that doesn't do anything (perhaps just holding the sample pairs), and try to get this client code working

gtsambos:
Hi @jeromekelleher, the structure you've now described in the [comment above](https://github.com/tskit-dev/tskit/pull/679#issuecomment-644006159) is now implemented in `test_tables.c` as `test_ibd_finder`. I'm still a bit hazy on how the user will implement these methods directly themselves (a PyObject???), but maybe we can talk about that later?

gtsambos:
hey @jeromekelleher (or @benjeffery), is there more detailed log output from the CircleCI builds that we can see somewhere? This makes it sound like perhaps there is...

```
Full log written to /home/circleci/tskit/build-gcc/meson-logs/testlog.txt
 ```

I'm having some trouble fixing the CircleCI failure in this PR, because the ninja build is only failing in CircleCI -- all of my local ninja builds are fine and aren't displaying this error.

**Comment by petrelharp** [comment]
Maybe you have to be admin to see it, or something? Anyhow, once logged in to circleCI I can see it: it failed on "Run C tests", saying:
```
#!/bin/bash -eo pipefail 
ninja -C build-gcc test 
 

ninja: Entering directory `build-gcc'
 
[0/1] Running all tests.
 
1/9 core               OK             0.02s
 
2/9 tables             FAIL           0.13s (killed by signal 11 SIGSEGV)
 
3/9 trees              OK             0.18s
 
4/9 genotypes          OK             0.13s
 
5/9 convert            OK             0.13s
 
6/9 stats              OK             0.18s
 
7/9 haplotype_matching OK             0.43s
 
8/9 file_format        OK             0.33s
 
9/9 minimal_cpp        OK             0.15s
 

 
Ok:                 8   
 
Expected Fail:      0   
 
Fail:               1   
 
Unexpected Pass:    0   
 
Skipped:            0   
 
Timeout:            0   
 

 

 
The output from the failed tests:
 

 
2/9 tables             FAIL           0.13s (killed by signal 11 SIGSEGV)
 

 
--- command ---
 
08:59:46 /home/circleci/tskit/build-gcc/test_tables
 
--- stdout ---
 

 

 
     CUnit - A unit testing framework for C - Version 2.1-3
 
     http://cunit.sourceforge.net/
 

 

 
Suite: tskit
 
  Test: test_node_table ...passed
 
  Test: test_edge_table ...passed
 
  Test: test_edge_table_copy_semantics ...passed
 
  Test: test_edge_table_squash ...passed
 
  Test: test_edge_table_squash_multiple_parents ...passed
 
  Test: test_edge_table_squash_empty ...passed
 
  Test: test_edge_table_squash_single_edge ...passed
 
  Test: test_edge_table_squash_bad_intervals ...passed
 
  Test: test_edge_table_squash_metadata ...passed
 
  Test: test_site_table ...passed
 
  Test: test_mutation_table ...passed
 
  Test: test_migration_table ...passed
 
  Test: test_individual_table ...passed
 
  Test: test_population_table ...passed
 
  Test: test_provenance_table ...passed
 
  Test: test_table_size_increments ...passed
 
  Test: test_table_collection_simplify_errors ...passed
 
  Test: test_table_collection_metadata ...passed
 
  Test: test_simplify_tables_drops_indexes ...passed
 
  Test: test_simplify_empty_tables ...passed
 
  Test: test_simplify_metadata ...passed
 
  Test: test_link_ancestors_no_edges ...passed
 
  Test: test_link_ancestors_input_errors ...passed
 
  Test: test_link_ancestors_single_tree ...passed
 
  Test: test_link_ancestors_paper ...passed
 
  Test: test_link_ancestors_samples_and_ancestors_overlap ...passed
 
  Test: test_link_ancestors_multiple_to_single_tree ...passed
 
  Test: test_ibd_finder ...
 
-------
 

 
Full log written to /home/circleci/tskit/build-gcc/meson-logs/testlog.txt
 
FAILED: meson-test 
 
/usr/local/bin/meson test --no-rebuild --print-errorlogs
 
ninja: build stopped: subcommand failed.
 
 
Exited with code exit status 1 

CircleCI received exit code 1 
```

**Comment by jeromekelleher** [comment]
Running it through valgrind is the first thing to try anyway @gtsambos - until that's running without problems things can fail in unpredictably different ways across systems.

**Comment by gtsambos** [comment]
Thanks @petrelharp and @jeromekelleher -- I noticed the valgrind errors and was working on that too, but thought the ninja problem might have been independent of it. I'll see what happens when I fix up the conditional jumps.

**Comment by benjeffery** [comment]
Circle CI failed due to a network issue. I re-ran and it has passed.

**Comment by gtsambos** [comment]
Hallelujah! Thanks @benjeffery! Btw, this pull request is still a few commits away from being ready to review -- I just wanted to sort out the CI issues before putting the final bells and whistles in.

**Comment by gbinux** [comment]
Hi, I am new here and but worked on IBD calculation code in C for NEWICK format trees for my rotation project. Considering the inefficiency of this format, I really hope the code can be rewritten to work with tskit tree sequence format. @gtsambos,  I am excited to know from @jeromekelleher that you have been working on this for thesis work. I have spent some time thinking about the algorithm for better efficiency. I have got some idea and would like to know whether the same algorithm has been already implemented. Would you like to chat about it?

**Comment by gtsambos** [comment]
Hi @gbinux, I'm glad to hear that you're excited, and I certainly hope this proves more efficient than what is currently possible with Newick trees!  I'd be interested in chatting (perhaps alongside @jeromekelleher or my supervisor @dvukcevic?), but I'm currently trying to wind up a few different projects and feel a bit swamped right now. Would it be okay if I got back to you in a week or two once this PR is done?

> Circle CI failed due to a network issue. I re-ran and it has passed.

btw @benjeffery, it looks like this problem has recurred. I'm still tidying some things up and anticipate needing to make a few more commits -- how about I just ping you when they are done so that you don't have to keep rerunning CircleCI manually?

**Comment by benjeffery** [comment]
I think a rebase might pull in a fix for gcov. Best to rebase often anyway.

**Comment by gbinux** [comment]
> Hi @gbinux, I'm glad to hear that you're excited, and I certainly hope this proves more efficient than what is currently possible with Newick trees! I'd be interested in chatting (perhaps alongside @jeromekelleher or my supervisor @dvukcevic?), but I'm currently trying to wind up a few different projects and feel a bit swamped right now. Would it be okay if I got back to you in a week or two once this PR is done?

Hi @gtsambos, thanks for sharing your current status. I completely understand that it is not easy to work on multiple projects at the same time. Just let me know when you feel comfortable to chat.  Looking forward to talking with you about IBD.

**Comment by gtsambos** [comment]
hi @jeromekelleher, I think this is ready to have a proper look at now. There are a few minor issues that are yet to be fixed; these include

 - the code that frees some memory as the algorithm runs: I have some questions for you about this
 - how to increase coverage so that codecov passes? More or less all of the code that remains untested is in the return statements for logical errors that shouldn't happen (and so are tricky to test)
 - user interface, as we've discussed before

While I was reviewing this code, I also spotted another fringe case in which this algorithm will not work yet. It's a logical flaw that pertains to both the Python and the C code at this point, so I need to think a bit before I fix it, but I expect it will be a fairly minor fix. I've flagged this case in one of the C tests.

**Comment by gtsambos** [comment]
one more push to hopefully fix the CircleCI issue (again)?

**Comment by gtsambos** [comment]
> Circle CI failed due to a network issue. I re-ran and it has passed.

fyi @benjeffery and @jeromekelleher, this is happening again, and the PR has been fully rebased so don't know exactly what is going on

**Comment by benjeffery** [comment]
I've kicked circle CI and it seems to have passed.

**Comment by gtsambos** [comment]
so @jeromekelleher, in light of #782 I think I'm going to drop my memory pruning step for now, and put a similar method to yours in here in a later PR. I'm going to do a quick look over of this now for test coverage and fringe topologies, but once that's done, I think it's ready to review? Will ping you when I'm there

**Comment by jeromekelleher** [comment]
Totally agree @gtsambos, sounds like an excellent plan. Ping me when it's ready for the Python-C stuff and I'll put in a commit for that.

**Comment by gtsambos** [comment]
ready @jeromekelleher

File: c/tskit/tables.h [hunk 1]
```diff
@@ -2816,6 +2868,8 @@ int tsk_table_sorter_free(struct _tsk_table_sorter_t *self);
 
 /** @} */
 
+int tsk_table_collection_check_integrity(
```
jeromekelleher, line 2871: Looks like this is incidental? The function should be delcared above, I think you can delete this.

jeromekelleher (commented):
Looks good, thanks @gtsambos. Can you squash the commits down to one please? This'll make it easier for me to add on stuff.

Once that's done, I'll add the python-c layer stuff. Might be next week before I get to it, though.

c/tskit/tables.h
gtsambos: Thanks, I must have accidentally put this in during one of my rebasing adventures. Fixed and squashed now.

c/tests/test_tables.c
```diff
@@ -2851,6 +2851,147 @@ test_link_ancestors_multiple_to_single_tree(void)
...
+    tsk_ibd_finder_t ibd_finder;
+
+    tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges, NULL, NULL,
+        NULL, NULL, NULL, 0);
+    ret = tsk_treeseq_copy_tables(&ts, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_ibd_finder_init_and_run(&ibd_finder, &tables, 1.0, samples, 2, 0.0, 0.0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
```
jeromekelleher, line 2871: We're not really testing anything here - can we check the known result for a couple of sample pairs?

c/tskit/tables.c
```diff
@@ -5341,6 +5334,547 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    self->samples = samples;
+    self->num_samples = num_samples;
+    self->sequence_length = sequence_length;
+    self->num_nodes = tables->nodes.num_rows;
+    self->tables = tables;
+    self->num_pairs = num_samples * (num_samples - 1) / 2;
+    self->min_length = min_length;
+    self->max_time = max_time;
+    if (max_time == 0) {
+        self->max_time = 2 * self->tables->nodes.time[self->num_nodes - 1];
```
jeromekelleher, line 5506: This seems a bit arbitrary - how about DBL_MAX? This would be a good default, I think.

c/tskit/tables.c
```diff
@@ -5341,6 +5334,547 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+        goto out;
+    }
+
+    // Mallocing and callocing.
+    num_nodes_alloc = 1 + tables->nodes.num_rows;
+    self->ancestor_map_head = calloc(num_nodes_alloc, sizeof(tsk_segment_t *));
+    self->ancestor_map_tail = calloc(num_nodes_alloc, sizeof(tsk_segment_t *));
+    self->ibd_segments_head = calloc(self->num_pairs, sizeof(tsk_segment_t *));
+    self->ibd_segments_tail = calloc(self->num_pairs, sizeof(tsk_segment_t *));
+    self->pair_index = -1;
```
jeromekelleher, line 5531: I don't think we need this variable, see comments below.

c/tskit/tables.c
```diff
@@ -5341,6 +5334,547 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    index0 = self->sample_id_map[sample0];
+    index1 = self->sample_id_map[sample1];
+    pair_index = ((float) self->num_samples) * ((float) self->num_samples - 1) / 2
+                 - ((float) self->num_samples - (float) index0)
+                       * ((float) self->num_samples - (float) index0 - 1) / 2
+                 + (float) index1 - (float) index0 - 1;
+
+    // Check the result is an integer before converting it.
+    assert(ceilf(pair_index) == pair_index);
+    self->pair_index = (int) pair_index;
```
jeromekelleher, line 5638: We should *return* the pair index here, not store it in an instance variable.

c/tskit/tables.c
```diff
@@ -5341,6 +5334,547 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+            //     }
+            // }
+        }
+    }
+
+out:
+    return ret;
+}
+
+int TSK_WARN_UNUSED
```
jeromekelleher, line 5792: I don't think this should be part of the external API. Mark it as static and put it further up in the file (static functions must be defined before they are used)

c/tskit/tables.c
```diff
@@ -5341,6 +5334,547 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+                if (l < r) {
+                    if (r - l > self->min_length) {
+                        if (seg0->node < seg1->node) {
+                            ret = ibd_finder_find_sample_pair_index(
+                                self, seg0->node, seg1->node);
+                        } else {
+                            ret = ibd_finder_find_sample_pair_index(
+                                self, seg1->node, seg0->node);
+                        }
+                        if (ret != 0) {
```
jeromekelleher, line 5834: So, here we'd do
```c
if (ret < 0) {
    goto out;
}
pair_index = ret;
 ret = ibd_finder_add_output(elf, l, r, current_parent, pair_index);

File: c/tskit/tables.h [hunk 8]
```diff
@@ -2733,6 +2767,48 @@ int tsk_table_collection_link_ancestors(tsk_table_collection_t *self, tsk_id_t *
...
+/* This is the current IBD C interface */
+int tsk_ibd_finder_free(tsk_ibd_finder_t *self);
+int tsk_ibd_finder_init_and_run(tsk_ibd_finder_t *ibd_finder,
+    tsk_table_collection_t *tables, double sequence_length, tsk_id_t *samples,
+    tsk_size_t num_samples, double min_length, double max_time);
+int tsk_ibd_finder_run(tsk_ibd_finder_t *ibd_finder);
+int tsk_ibd_finder_calculate_ibd(tsk_ibd_finder_t *self, tsk_id_t current_parent);
+void ibd_finder_print_state(tsk_ibd_finder_t *self, FILE *out);
+
+/* JK Here's the interface we want:
```
jeromekelleher, line 2779: I think we can simplify things quite a bit by changing the public interface. I've made a sketch of what this might look like here.

python/_tskitmodule.c
```diff
@@ -6547,6 +6547,9 @@ TableCollection_link_ancestors(TableCollection *self, PyObject *args, PyObject *
     return ret;
 }
 
+// IBD: add entry for find_ibd as a method of the TableCollection Python class.
```
jeromekelleher, line 6550: Delete

python/tests/test_lowlevel.py
```diff
@@ -303,6 +303,53 @@ def test_union_bad_args(self):
...
+        for bad_float in ["sdf", None, {}]:
+            with self.assertRaises(TypeError):
+                tc.find_ibd([0, 1], min_length=bad_float)
+            with self.assertRaises(TypeError):
+                tc.find_ibd([0, 1], max_time=bad_float)
+
+    def test_idb_output_no_recomb(self):
+        ts = msprime.simulate(10, random_seed=1)
+        tc = ts.tables.ll_tables
+        segs = tc.find_ibd([0, 1, 2, 3])
```
jeromekelleher, line 326: I've changed the interface here a bit. Rather than each value in the dictionary being a list of (left, right, node) tuples, I've changed to the values being dictionaries with keys "left", "right" and "node" which map to numpy arrays of the values. This is for efficiency at the C layer, and I think it'll end up being more useful later anyway when we use these for calculations.

Arguably we should make these values an ``IbdSegments`` class rather than plain dicts, but I'm not sure it matters much. 

What do you think? Will this break much of your existing Python code?

c/tskit/tables.c
```diff
@@ -5341,6 +5334,549 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+            if (node_times[*p] < node_times[(int) input_edges->parent[j]]) {
+                self->oldest_parent[c] = input_edges->parent[j];
+            }
+            assert(&self->oldest_parent[c] != NULL);
+        }
+    }
+}
+
+static int
+ibd_finder_init(tsk_ibd_finder_t *self, double sequence_length, tsk_id_t *samples,
```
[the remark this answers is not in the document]
gtsambos: ah, somehow I didn't realise that `sequence_length` was also an attribute of the table collection, I thought we'd have to do something annoying like find the max of all the `right` coordinates. Yes, we won't need this parameter then

c/tskit/tables.c
jeromekelleher: Hmm, although it's not clear to me how we'll map back from arbitrary sample pairs to their indexes in the input list here. It's probably simplest to do a linear scan of the list of pairs for now, and come with something better later.

python/tests/test_lowlevel.py
gtsambos: So, just to be clear, this means that if a user wanted to see what IBD segments corresponds to sample pair `(u, v)`, they'd do something like this?
```
ibd_uv = ibd_output[(u, v)]
for i in range(len(ibd_uv)):
    print(ibd_uv['left'][i], ibd_uv['right'][i], ibd_uv['node'][i])
```
I think if this makes the C layer more efficient then we should do it, but it might be nice to have a Python function that converts this back into the more interpretable format.

> Will this break much of your existing Python code?

I could change the Python code, but it might be easier to just convert the existing Python output into this new format within the `verify_ibd` testing function.

> Arguably we should make these values an IbdSegments class rather than plain dicts, but I'm not sure it matters much.

Yeah, I was originally thinking of this output as a class of its own too, but I'm also not particularly attached to the idea!

Thanks for looking through this btw! I'll go over the other bits in more detail tomorrow.

c/tskit/tables.c
gtsambos: Yeah, this is something I didn't think of in the call -- the function that defines the mapping between the 2D sample pairs and the ID index is going to be a lot more complicated in this new setting of arbitrary sample pairs. You could still come up with a canonical way of ordering them, but actually finding the right index for a given pair will inevitably involve some scan through the list.

Maybe it would be worth retaining these existing `ibd_finder_find_sample_pair_index` functions for whenever the user does just want to calculate IBD for all sample pairs (which, as I said in the call, I think would be a sensible default to assume)? This linear scanning for pair indices is going to be suboptimal in exactly that case.

python/tests/test_lowlevel.py
jeromekelleher: I think it's better to think of them as numpy arrays so we can do vector operations on them, so 

```python
ibd_uv = ibd_output[(u, v)]
left = ibd_uv["left"] # Although nicer if it was ``ibd_uv.left``
right =  ibd_uv["right"]
print("mean length = ", np.mean(right - left))
```

It's a classic struct-of-arrays vs array-of-structs thing. We've been tending to go for struct-of-arrays in tskit, so I'd tend to favour this approach for now. Let's see how it goes when we start working with some big tree sequences.

c/tskit/tables.c
gtsambos: My thought where that (assuming the nodes are ordered by backwards-time), `self->tables->nodes.time[self->num_nodes - 1]` should be the time of the oldest node in the tree, so setting `max_time` to be twice this should safely ensure it is never exceeded. I'm happy to change it to DBL_MAX though!

jeromekelleher:
I'm going to mute this thread for now @gtsambos - can you @ me when you'd like me to take another look please?

gtsambos:
Actually @jeromekelleher, even though I'm still working on this, I'd really appreciate if you could look at the changes I've made to the Python code in `ibd.py`. This has now been changed to incorporate some of the major things you suggested above, including
 - inputting sample pairs instead of just samples
 - various extra checks of the input (there are a few extra ways in which an inputted sample pair might be dodgy even if the samples in the pair are okay)
 - returning the ibd output in a dict holding dicts of numpy arrays

At the moment I'm working on updating the C code to be consistent with these changes, where it's necessary.

gtsambos:
> various extra checks of the input (there are a few extra ways in which an inputted sample pair might be dodgy even if the samples in the pair are okay)

About this: I currently have some checks for 'dodginess' in the Python code, see the new method `check_sample_pairs`. However, I'm not sure whether it's best to also test these in the C code, or whether these should be checked before the ibd_finder code is even invoked. In particular:
 - Pairs with duplicate ids `(u, u)` are disallowed
 - Duplicate pairs `(u, v), (u, v)` or `(u, v), (v, u)`  are disallowed. (It wouldn't be too hard to modify the code to allow such duplicates, but at the cost of making the pair look-up take twice as long as it currently does)

python/tests/ibd.py
```diff
@@ -126,22 +126,37 @@ def add(self, other):
 class IbdFinder:
     """
     Finds all IBD relationships between specified samples in a tree sequence.
+    One of either sample_pairs or samples must be provided
     """
 
-    def __init__(self, ts, samples=None, min_length=0, max_time=None):
+    def __init__(
+        self, ts, sample_pairs=None, samples=None, min_length=0, max_time=None
```
jeromekelleher, line 133: Let's keep it simple for now and just accept ``sample_pairs``. We can add the all-pairs behaviour later at the top level.

jeromekelleher (commented):
Some minor points @gtsambos.

- let's not worry about dodgy inputs for now
- Use numpy in the Python implementation (``self.sample_pairs = np.array(sample_pairs)``, then check if it's a 2D array of ints of the right shape).

c/tskit/tables.h
gtsambos: I now have all the functions you've suggested, but have retained the `ibd_finder_init_and_run` to streamline the code inside the C tests. We don't need to use it in the interface itself.

gtsambos:
Hi @jeromekelleher, this is ready to check assuming that I've cleared the CI checks. Other than the comments above, the main things I'm unsure of are these:

> I've changed the interface here a bit. Rather than each value in the dictionary being a list of (left, right, node) tuples, I've changed to the values being dictionaries with keys "left", "right" and "node" which map to numpy arrays of the values. This is for efficiency at the C layer, and I think it'll end up being more useful later anyway when we use these for calculations.

This has now been changed in the Python mock-up, but I'm not sure if I also need to change the output in the C code to reflect this?

About the sample pair stuff: In the Python mock-up, I've made it so that the user inputs either a 1D numpy array of sample nodes or a 2D array of pairs. If the user inputs a 1D array, all pairs of these samples are used, and the original procedure for calculating the pair index is used. If the user inputs a 2D array, only those specific pairs are used, and the pair indices are found by a linear search through the user's inputted list.

In the C code, only the 2D procedure is currently possible, but I would like to retain the 1D option because if you are interested in IBD between all possible pairs of samples, the pair index calculation can be done much more efficiently. What would be a sensible way to enable this? I was thinking that it might be better to do it outside of the low level C code, maybe like this:

[user provides an inputted tree sequence and a list]

If the list is 1D, treat the list as a list of samples.
    - Calculate all possible pairs of samples using itertools.
    -  Pass this list of pairs into `ibd_finder_init.
        - Set a (new) boolean attribute of ibd_finder_t called `calculate_all_pairs` to `true`. 
        - Inside `calculate_ibd_segs`, add conditional statements like these
```
if (self->calculate_all_pairs) {
   ret = ibd_finder_find_sample_pair_index(self, seg0->node, seg1->node); // efficient
} else {
   ret = ibd_finder_find_sample_pair_index2(self, seg0->node, seg1->node); // not efficient
}
```

python/tests/ibd.py
gtsambos: Okay, cool. This addresses one of the points I made below.

gtsambos:
Also, I discovered some extra cases where this isn't doing what I'd expect -- they are 'fringe' topological cases where some of the samples are parents of other samples. I will look at this later.

gtsambos:
Hmm, only the low level tests are failing now. I'll see if there's an easy fix; if not, I'll look in more detail tomorrow

jeromekelleher:
Re this one:
``` 
tc = ts.tables.ll_tables
```
need to update this to 
```
tc = ts.tables._ll_tables
```
as this has changed while you're PR has been ongoing.

jeromekelleher:
@gtsambos - I'd really prefer not to put in special cases at this point. We'll make whatever changes we need to make the pair-index lookup efficient. There's no point, in my opinion, in optimising the n^2 code path because it's just never going to even be runnable  (n = 10^6). Let's  keep it as simple as possible for now, and get the interface settled so that we can merge. It's easy to follow up with smaller updates that tweak performance, but I want to get the actual functionality in there as soon as possible so that we can start working with it and can finalise the interface.

gtsambos:
> @gtsambos - I'd really prefer not to put in special cases at this point.

okay, thanks @jeromekelleher. I've updated `_tskitmodule.c` and commented out the 'special code' for the time being. Once this passes CI this should be ready.

gtsambos:
🥳 🥳 🥳 ready to go!

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    }
+
+    // Do we need the below? Seems like no?
+    // memset(self->sample_id_map, 0xff, self->tables.nodes.num_rows * sizeof(tsk_id_t));
+    ret = ibd_finder_init_samples(self);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // TODO: feels weird not having a return statement for this?
```
jeromekelleher, line 5616: It looks fine - it doesn't have any error conditions, so no need for return value.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+// out:
+//     return (int) pair_index;
+// }
+
+static int
+ibd_finder_find_sample_pair_index2(
+    tsk_ibd_finder_t *self, tsk_id_t sample0, tsk_id_t sample1)
+{
+    int i = 0;
+    while (i < (int) self->num_pairs) {
```
jeromekelleher, line 5741: Eek! What about:
```c
// change the return type to tsk_id_t so we don't have to cast back and forth to int

tsk_id_t i, s0, s1;

for (i = 0; i < (tsk_id_t) self->num_pairs; i++) {
    s0 = self->pairs[2 * i];
    s1 =self->pairs[2 * i + 1];
    if ((s0 == sample0 && s1 == sample1) || (s0 == sample1 && s1 == sample0)) {
        break;  
    }
    ret = -1;
    if (i < (tsk_id_t) self->num_pairs) {
          ret = tsk_id_t;
    }
    return ret;
} 
```

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    }
+}
+
+static int TSK_WARN_UNUSED
+ibd_finder_calculate_ibd(tsk_ibd_finder_t *self, tsk_id_t current_parent)
+{
+    int ret = 0;
+    int j, pair_index;
+    tsk_segment_t *seg, *seg0, *seg1;
+    double l, r;
```
jeromekelleher, line 5763: ``left`` and ``right`` are a bit better than ``l`` and ``r``.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+
+out:
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_ibd_finder_get_ibd_segments(
+    tsk_ibd_finder_t *self, tsk_id_t pair_index, tsk_segment_t *ret_ibd_segments_head)
+{
+    int ret = 0;
```
jeromekelleher, line 5829: There's no need to cast pair_index to an int here, you can do
```
if ((pair_index < 0) || (pair_index >= (tsk_id_t) self->num_pairs)) {
```

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    }
+    self->segment_queue_size = 0;
+
+out:
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_ibd_finder_get_ibd_segments(
+    tsk_ibd_finder_t *self, tsk_id_t pair_index, tsk_segment_t *ret_ibd_segments_head)
```
jeromekelleher, line 5827: I think it would be better do this slightly differently:
```c
int TSK_WARN_UNUSED
￼tsk_ibd_finder_get_ibd_segments(
￼    tsk_ibd_finder_t *self, tsk_id_t pair_index, tsk_segment_t **ret_ibd_segments_head)
{
     // error checking as before 
     *ret_ibd_segments_head = self->ibd_segments_head[pair_index]
}
```
Notice that we're passing in a pointer-to-a-pointer here. Client code would then look like

```c
segment_t *seg;

ret = tsk_ibd_finder_get_ibd_segments(ibd_finder, 0, &seg);
if (ret..) // .. blah blah

while (seg != NULL) {
     // process IBD segment
    seg = seg->next;
}
```

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+            fprintf(out, "(%f,%f->%d)", u->left, u->right, u->node);
+        }
+        fprintf(out, "\n");
+    }
+}
+
+int TSK_WARN_UNUSED
+tsk_ibd_finder_run(tsk_ibd_finder_t *self)
+{
+
```
jeromekelleher, line 5898: extra space

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    int ret = 0;
+    size_t j;
+    tsk_id_t current_parent, u;
+    size_t num_edges = input_edges->num_rows;
+    tsk_segment_t *seg;
+    tsk_segment_t *s;
+    double intvl_l, intvl_r, current_time;
+    bool parent_should_be_added = true;
+
+    if (num_edges > 0) {
```
jeromekelleher, line 5909: Do you actually need the ``num_edges > 0`` check here?

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+            // seg.
+            u = seg->node;
+            if (self->is_sample[u]) {
+                ret = ibd_finder_enqueue_segment(self, seg->left, seg->right, seg->node);
+                if (ret != 0) {
+                    goto out;
+                }
+            } else {
+                if (self->ancestor_map_head[u] != NULL) {
+                    s = self->ancestor_map_head[u];
```
jeromekelleher, line 5932: for loop is a bit more concise:
```c
for (s = self->ancestor_map_head[u]; s != NULL; s = s->next) {
}
```
￼Also looks like you don't need the ``if (self->ancestor_map_head[u] != NULL) {`` condition either.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+                    goto out;
+                }
+                ret = ibd_finder_calculate_ibd(self, current_parent);
+                if (ret != 0) {
+                    goto out;
+                }
+                parent_should_be_added = false;
+            }
+
+            // if (j != num_edges - 1 && input_edges->parent[j + 1] != current_parent) {
```
jeromekelleher, line 5979: This smells a bit - it's fine leaving commented out code when developing, but once something is ready to be merged it should be cleaned up, removing stuff like this.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+            // }
+        }
+    }
+
+out:
+    ibd_finder_print_state(self, stdout);
+    return ret;
+}
+
+// Define a function that frees memory from the ibd_finder.
```
jeromekelleher, line 5998: This comment is redundant

c/tskit/tables.h
```diff
@@ -2733,6 +2767,52 @@ int tsk_table_collection_link_ancestors(tsk_table_collection_t *self, tsk_id_t *
...
+int tsk_ibd_finder_run(tsk_ibd_finder_t *ibd_finder);
+void ibd_finder_print_state(tsk_ibd_finder_t *self, FILE *out);
+
+// new additions:
+int tsk_ibd_finder_set_min_length(tsk_ibd_finder_t *self, double min_length);
+int tsk_ibd_finder_set_max_time(tsk_ibd_finder_t *self, double max_time);
+int tsk_ibd_finder_get_ibd_segments(tsk_ibd_finder_t *ibd_finder, tsk_id_t pair_index,
+    tsk_segment_t *ret_ibd_segments_head);
+/* JK Here's the interface we want:
+
```
jeromekelleher, line 2784: We should clean this up - no need to keep these comments lying around.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,638 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+
+    //     // Initialise the sample id map.
+    //     self->sample_id_map[samples[j]] = j;
+    // }
+out:
+    return ret;
+}
+
+static void
+ibd_finder_init_oldest_parents(tsk_ibd_finder_t *self)
```
jeromekelleher, line 5530: Are we using this now? If not, we should delete it. It's easy enough to reinstate later and having code that doesn't do anything lying around is a bad idea.

c/tskit/tables.c
jeromekelleher: https://en.wikipedia.org/wiki/Code_smell

c/tskit/tables.h
```diff
@@ -18,7 +18,7 @@
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
  * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT O©R OTHERWISE, ARISING FROM,
```
jeromekelleher, line 21: Something's gone a little screwy with the copyright symbols here (elsewhere too I think)

c/tskit/tables.h
```diff
@@ -688,6 +688,40 @@ typedef struct _tsk_table_sorter_t {
...
+    struct _tsk_segment_t *next;
+    tsk_id_t node;
+} tsk_segment_t;
+
+// Define a struct called ibd_finder_t.
+
+typedef struct {
+    tsk_id_t *pairs;
+    size_t num_pairs;
+    tsk_id_t *samples;
```
jeromekelleher, line 705: We're not using the ``samples`` and ``num_samples`` any more are we? We should remove any unused values from the struct.

python/tests/simplify.py
```diff
@@ -1,12 +1,12 @@
 # MIT License
 #
-# Copyright (c) 2018-2020 Tskit Developers
-# Copyright (c) 2015-2018 University of Oxford
+# Copyright(c) 2018 - 2020 Tskit Developers
```
jeromekelleher, line 3: Lots of unwanted formatting changes here - your editor has gone power mad!

jeromekelleher (commented):
Looks good @gtsambos. I think the C code needs a bit of a cleanup and there's some random changes to various files that need to be reverted. Once that's done I'll update the Python-C interface to input a 2D array of sample pairs, and we should be ready to merge then I think.

c/tskit/tables.h
gtsambos: oh, I was keeping them because they are needed to calculate pair indices the old-fashioned efficient way when you are considering all sample pairs. I can take them out for now though.

python/tests/simplify.py
gtsambos: I don't know what's happened here, eek 🤦

gtsambos: I think I might have accidentally used clang-format on the Python code without realising I was in the wrong directory. Anyway, will fix

c/tskit/tables.c
gtsambos: Thanks. I was doing all this int casting deliberately, because I didn't actually realise you could compare for the equality of `tsk_id_t` objects like this! I assumed that for user-defined structs, you'd have to explicitly construct a method to check for equality like `__eq__` in Python. I'm glad that at least some things are easier in C!

I've changed this to be halfway between your suggested function and my original one -- I've kept the return type as an int because it is supposed to be an index, rather than a node ID.  Hopefully it looks a bit neater now though.

python/_tskitmodule.c
```diff
@@ -6632,6 +6632,133 @@ TableCollection_union(TableCollection *self, PyObject *args, PyObject *kwds) {
...
+    npy_intp *shape;
+    static char *kwlist[] = {"samples", "min_length", "max_time", NULL};
+
+    memset(&ibd_finder, 0, sizeof(ibd_finder));
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|dd", kwlist, &samples,
+                &min_length, &max_time)) {
+        goto out;
+    }
+    /* NOTE: This will be a 2D array when we read in the pairs explicity */
+    samples_array = (PyArrayObject *) PyArray_FROMANY(samples, NPY_INT32,
```
gtsambos, line 6740: @jeromekelleher I changed this to be a 2D array -- I needed to do this to ensure the low level tests passed the CI.

gtsambos:
Thanks @jeromekelleher, hopefully ready this time 🤞 

btw:

> Once that's done I'll update the Python-C interface to input a 2D array of sample pairs

I think I did this myself in the last round of commits -- see my comment

jeromekelleher:
I've made a pass through and tidied things up a little bit in terms of the C API @gtsambos, and I think it's all looking good there. However, it looks like your tests in ``test_ibd.py`` aren't actually calling into the C code, to me. We'd need to hook this in somehow so that we're running the library code as well as the Python prototypes (we use the ``compare_lib=True`` pattern when testing simplify). 

Once the tests are all hooked up and we're getting identical results from the Python and C versions, that I think we're ready to merge.

gtsambos:
Hi @jeromekelleher, I've now tidied up the Python tests a little and added a `compare_lib` option so the Python and C output can be compared. There are a few different issues:

First, I'm having a bit of trouble getting this to work with the `min_length` and `max_time` arguments -- I've added a few extra lines to the method in `tables.py` so that these arguments can at least be used, but I'm getting a `tskit.GenericError` when I actually try to specify non-generic values. I'm wondering whether I need to explicitly convert these into something like a `PyArrayObject` in `_tskitmodule.c` (like you did for the sample pairs)?

Another problem is that I get a `tskit.GenericError` whenever the output is empty. I've realised that I actually didn't have any C tests covering this case (!) so I'll add one in, as I'm not sure whether this is an issue with the C code itself, or whether there's some problem trying to convert the C output back into valid Python output.

gtsambos:
> Once the tests are all hooked up and we're getting identical results from the Python and C versions, that I think we're ready to merge.

In addition, there are some test failures resulting from genuine discrepancy between the methods. I think these are all because my C code is inadvertently returning all IBD segments instead of just the most recent ones (related to the concept of `mrca_ibd` that we discussed way earlier in the year). I have flagged these all as expected failures for now, but am looking into a fix

gtsambos:
I think this is ready now @jeromekelleher! I'm just about to push up a squashed version of the PR, feel free to have a look!

There is one test that I've flagged as an expected failure that I think I will fix later, if it's okay by you. It is a test case that involved a particular topology of ancient samples, so not a situation that I'll encounter in my demonstrations or analyses. The Python and C implementations are returning the same result, it's just a little different from what I would expect. (I will have to make further changes to this code anyway when your memory-saving procedure comes in a later PR.)

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+
+    // Allocate the heaps used for small objects.
+    ret = tsk_blkalloc_init(&self->segment_heap, 8192);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // Mallocing and callocing.
+    num_nodes_alloc = 1 + tables->nodes.num_rows;
+    self->ancestor_map_head = calloc(num_nodes_alloc, sizeof(tsk_segment_t *));
```
jeromekelleher, line 5528: It's a bit safer to say
```c
self->ancestor_map_head = calloc(num_nodes_alloc, sizeof(*self->ancestor_map_head));
```
As then we don't have to remember what the types are.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    }
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_ibd_finder_set_max_time(tsk_ibd_finder_t *self, double max_time)
+{
+    int ret = 0;
+
+    if (max_time == 0) {
```
jeromekelleher, line 5568: I'm not sure we really need this default at 0 - wouldn't it be simpler to just not call ``tsk_ibd_finder_set_max_time`` if you want the default?  Calling with DBL_MAX is fine for your helper function in the tests (probably a bit more readable, anyway).

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+{
+    int ret = 0;
+    int j, pair_index;
+    tsk_segment_t *seg, *seg0, *seg1;
+    double left, right;
+
+    if (self->ancestor_map_head[current_parent] == NULL) {
+        for (j = 0; j != (int) self->segment_queue_size; j++) {
+            seg = &self->segment_queue[j];
+            // if (current_parent == 4) {
```
jeromekelleher, line 5640: Dead code

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+        }
+
+        // Stop if the processed node's time exceeds the max time.
+        current_parent = input_edges->parent[j];
+        current_time = self->tables->nodes.time[current_parent];
+        if (current_time > self->max_time) {
+            goto out;
+        }
+
+        // if (current_parent == 4) {
```
jeromekelleher, line 5792: Dead code

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+        // }
+
+        // Extract segment.
+        seg = tsk_ibd_finder_alloc_segment(
+            self, input_edges->left[j], input_edges->right[j], input_edges->child[j]);
+        // Create a SegmentList holding all of the sample segments descending from
+        // seg.
+        u = seg->node;
+        if (self->is_sample[u]) {
+            // printf("hereee");
```
jeromekelleher, line 5805: More prints - can you have a look through the diff for any more of these please?

python/tskit/tables.py
```diff
@@ -2795,3 +2797,18 @@ def union(
             self.provenances.add_row(
                 record=json.dumps(provenance.get_provenance_dict(parameters))
             )
+
+    def find_ibd(self, samples, max_time=None, min_length=0):
+        """
```
jeromekelleher, line 2802: Better to have no docstring. If this is included, then this function will be included on read the docs (which we don't want it to be, yet.

python/tskit/tables.py
```diff
@@ -2795,3 +2797,18 @@ def union(
...
+        Docstring for method goes in here.
+        """
```
jeromekelleher, line 2804:
```python
def find_ibd(self, samples, max_time=None, min_length=None):
     max_time = sys.float_info.max if max_time is None else max_time
     min_length = 0  if min_length is None else min_length
     return self._ll_tables.find_ibd(samples, max_time=max_time, min_length=min_length)
```
It's better to avoid hard-coding the default values of parameters into the signature unless we're very sure they'll never change.

jeromekelleher (commented):
Looks great, thanks @gtsambos! Once final round of tweaks and we're good to merge I think.

c/tskit/tables.c
```diff
@@ -5388,6 +5381,489 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+    if (tail == NULL) {
+        x = tsk_ibd_finder_alloc_segment(self, left, right, output_id);
+        if (x == NULL) {
+            ret = TSK_ERR_NO_MEMORY;
+            goto out;
+        }
+        self->ancestor_map_head[input_id] = x;
+        self->ancestor_map_tail[input_id] = x;
+    } else {
+        // if (tail->right == left && tail->node == output_id) {
```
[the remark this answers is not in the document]
gtsambos: Oh, because it leads to incorrect IBD output (you end up with the `path_ibd=False` version of IBD, which we agreed a while ago should not be the default). We want to record distinct IBD segments for intervals that have been separated by a recombination event before being passed to the samples, even if the common ancestor of both intervals is the same. I should have actually removed this instead of commenting it out, though

c/tskit/tables.c
gtsambos: I initially though this as well, but using `DBL_MAX` in the test file kept producing errors like these:
```
../tests/test_tables.c:2873:21: error: use of undeclared identifier 'DBL_MAX'
    if (max_time != DBL_MAX ) {
                    ^
../tests/test_tables.c:2910:74: error: use of undeclared identifier 'DBL_MAX'
    ret = ibd_finder_init_and_run(&ibd_finder, &tables, samples, 1, 0.0, DBL_MAX );
```

gtsambos:
Thanks @jeromekelleher, slightly revised PR just pushed now. I've added a few comments to yours above ⬆️

c/tskit/tables.c
benjeffery: I think you need `#include <float.h>` for that macro?

gtsambos: oh, I just realised I probably need one of these in test_tables.c now:
```
#include <stdio.h>
#include <stdbool.h>
#include <stdint.h>
```
let me go through and find that

gtsambos: is there a way I can actually see what is in these files? I don't know which one of these DBL_MAX is in

benjeffery: There's a list here: https://en.cppreference.com/w/c/types/limits `DBL_MAX` should be in `float.h`

benjeffery: Opps that might be c++.

gtsambos: is it in `float.h`? I'm going to try that now

gtsambos: Hmm, I might have to do this tomorrow @jeromekelleher -- this change is causing some of my C tests to fail and I can't obviously see why

gtsambos: (thanks @benjeffery, only just saw your comments now!)

benjeffery: You just get used to github live-updating when it stops... been there!

gtsambos:
Hi @jeromekelleher, would it be okay if we pressed ahead with this PR and flagged the `max_time` thing above as an issue to be fixed later? I've spent quite a bit of time trying and failing to get the tests working with that small alteration, and given how many deadlines I have for various things next week, I think it would be a better use of my time to focus on those just for now.

My latest problem is that `sys.float_info.max` in `tables.py` doesn't appear to be interpreted as equivalent to `DBL_MAX` in the C code, so most of my Python tests are failing when I change the `max_time` default away from what it currently is in this PR

jeromekelleher:
Sounds good @gtsambos - I'll fix up the max_time issue now and merge.

jeromekelleher (approved):
Looks great, thanks @gtsambos! Looking forward to seeing what it can do!

Issue #665, opened 2020-06-03
Windowed Genealogical Nearest Neighbours (GNN)


awohns:
@hyanwong, Gil and I have recently discussed details of a "windowed" GNN, which would use either span- or time-based windows to calculate the nearest neighbours of a set of focal nodes. Here's a proposal of what we hope to implement.

**Background:** The currently-implemented GNN function definition is the following:
`genealogical_nearest_neighbours(focal, sample_sets, num_threads=0)`

- focal is "A list of 𝑛 nodes whose GNNs should be calculated."
- sample_sets is "A list of 𝑚 lists of node IDs."

The function returns "An 𝑛 by 𝑚 array of focal nodes by GNN proportions. Every focal node corresponds to a row. The numbers in each row corresponding to the GNN proportion for each of the passed-in reference sets. Rows therefore sum to one."

As detailed in the mathematical definition of a GNN [here](https://www.nature.com/articles/s41588-019-0483-y), the GNN is calculated on a per-tree basis, and then normalised by the span of each tree.

**Proposal**
We aim to generalise the GNN function to calculate the GNN using two types of windows: span-based windows (for example, each tree in the tree sequence) and time-based windows. 
- _Positional GNN_ using span-based windows: Calculate the GNN for each focal node as before, but return GNN values at each window.
- _Time-Aware GNN_ using time intervals: Calculate the GNN for each focal node as before, but note which time bin _v_ falls into (using notation from the equation linked above). 

This change will _not_ modify the per-tree calculation of the GNN.

**New Function Definition**
`genealogical_nearest_neighbours(focal, sample_sets, windows, time_windows, num_threads=0)`
- windows is a list of _r_ increasing_ list of span-based breakpoints between the windows to compute the statistic in
- time_windows is a list of _s_ increasing list of time-based breakpoints which delineate time intervals to compute the statistic in

If `windows` and `time_windows` are None, the function returns an 𝑛 by 𝑚 array exactly as in the current implementation. If `windows` is not None, an _r_ by _m_ by _n_ array of focal nodes by GNN proportions by windows is outputted. Every focal node corresponds to a 2d array of _r_ by _m_. For each focal node at each window, a 1d array correspond to the GNN proportion for each of the passed-in reference sets at that positional window, and should all sum to one. Similarly, if `time_windows` is not None, an _s_ by _m_ array for each focal node is returned. If both `windows` and `time_windows` is used, a 4d, _r_ by _s_ by _m_ by _n_ array is outputted.

[...]

PR #683, opened 2020-06-16
Windowed Genealogical Nearest Neighbours

awohns:
Python implementation of a windowed genealogical nearest neighbours statistic, as discussed in #665.

Function definition is: 
`windowed_genealogical_nearest_neighbours(ts, focal, reference_sets, windows=None, time_windows=None, span_normalise=True, time_normalise=True)`

If the function is called without any optional arguments, the output is a 2d array of _n_ by _m_ where _n_ are the focal nodes whose GNNs are being calculated and _m_ is the number of reference sets at which to calculate GNNs. This identical to the output from  ts.genealogical_nearest_neighbours(). Passing arguments to `windows` and `time_windows` each increase the dimensions of the output by 1. If only `windows` is used, the output is a 3d array of _r_ by _n_ by _m_, where _r_ is the number of windows. If only `time_windows` is used, the output is a 3d array of _s_ by _n_ by _m_, where _s_ is the number of time windows. If both `windows` and `time_windows` are used, the output is a 4d array of _r_ by _s_ by _n_ by _m_.

To be consistent with the tree stats already implemented in tskit, `windows` should be a list of n+1 increasing numbers beginning with 0 and ending with the sequence_length. The `time_windows`, on the other hand, do not have to extend to the oldest root node. There can be time intervals with no information, in such intervals the output is NaN.

I aimed for `span_normalise` to operate in the same way as it does in the previously implemented tree stats, i.e. divide by the window’s span if this argument is true. `time_normalise` works in a similar fashion. If true, the values at each time interval are divided by the spans of the parent nodes found in each time interval. If neither `span_normalise` nor `time_normalise` is used, the sum of the returned ndarray is 1. If only `span_normalise` is used, summing over reference sets and time_windows (if used) will yield 1 at each window. If only `time_normalise` is used, summing over reference sets and windows (if used) will yield 1 at each time_window. If both `span_normalise` and `time_normalise` are used, summing over reference sets will yield 1 at each `window, time_window `array value.

I'd be very grateful for any comments, particularly from the stats master @petrelharp!

### DIFF
--- a/python/tests/test_stats.py
+++ b/python/tests/test_stats.py
@@ -390,6 +390,122 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
     return A
 
 
+def parse_time_windows(ts, time_windows):
+    if time_windows is None:
+        time_windows = [0.0, ts.max_root_time]
+    return np.array(time_windows)
+
+
+def windowed_genealogical_nearest_neighbours(
+    ts,
+    focal,
+    reference_sets,
+    windows=None,
+    time_windows=None,
+    span_normalise=True,
+    time_normalise=True,
+):
+    """
+    genealogical_nearest_neighbours with support for span- and time-based windows
+    """
+    reference_set_map = np.full(ts.num_nodes, tskit.NULL, dtype=int)
+    for k, reference_set in enumerate(reference_sets):
+        for u in reference_set:
+            if reference_set_map[u] != tskit.NULL:
+                raise ValueError("Duplicate value in reference sets")
+            reference_set_map[u] = k
+
+    windows_used = windows is not None
+    time_windows_used = time_windows is not None
+    windows = ts.parse_windows(windows)
+    num_windows = windows.shape[0] - 1
+    time_windows = parse_time_windows(ts, time_windows)
+    num_time_windows = time_windows.shape[0] - 1
+    A = np.zeros((num_windows, num_time_windows, len(focal), len(reference_sets)))
+    K = len(reference_sets)
+    parent = np.full(ts.num_nodes, tskit.NULL, dtype=int)
+    sample_count = np.zeros((ts.num_nodes, K), dtype=int)
+    time = ts.tables.nodes.time
+    norm = np.zeros((num_windows, num_time_windows, len(focal)))
+
+    # Set the initial conditions.
+    for j in range(K):
+        sample_count[reference_sets[j], j] = 1
+
+    window_index = 0
+    for (t_left, t_right), edges_out, edges_in in ts.edge_diffs():
+        for edge in edges_out:
+            parent[edge.child] = tskit.NULL
+            v = edge.parent
+            while v != tskit.NULL:
+                sample_count[v] -= sample_count[edge.child]
+                v = parent[v]
+        for edge in edges_in:
+            parent[edge.child] = edge.parent
+            v = edge.parent
+            while v != tskit.NULL:
+                sample_count[v] += sample_count[edge.child]
+                v = parent[v]
+
+        # Update the windows
+        assert window_index < num_windows
+        while windows[window_index] < t_right and window_index + 1 <= num_windows:
+            w_left = windows[window_index]
+            w_right = windows[window_index + 1]
+            left = max(t_left, w_left)
+            right = min(t_right, w_right)
+            span = right - left
+            # Process this tree.
+            for j, u in enumerate(focal):
+                focal_reference_set = reference_set_map[u]
+                delta = int(focal_reference_set != tskit.NULL)
+                p = u
+                while p != tskit.NULL:
+                    total = np.sum(sample_count[p])
+                    if total > delta:
+                        break
+                    p = parent[p]
+                if p != tskit.NULL:
+                    scale = span / (total - delta)
+                    time_index = np.searchsorted(time_windows, time[p]) - 1
+                    if 0 <= time_index < num_time_windows:
+                        for k in range(len(reference_sets)):
+                            n = sample_count[p, k] - int(focal_reference_set == k)
+                            A[window_index, time_index, j, k] += n * scale
+                        norm[window_index, time_index, j] += span
+            assert span > 0
+            if w_right <= t_right:
+                window_index += 1
+            else:
+                # This interval crosses a tree boundary, so we update it again
+                # in the next tree
+                break
+
+    # Reshape norm depending on normalization selected
+    # Return NaN when normalisation value is 0
+    if span_normalise and time_normalise:
+        reshaped_norm = norm.reshape((num_windows, num_time_windows, len(focal), 1))
+    elif span_normalise and not time_normalise:
+        norm = np.sum(norm, axis=1)
+        reshaped_norm = norm.reshape((num_windows, 1, len(focal), 1))
+    elif time_normalise and not span_normalise:
+        norm = np.sum(norm, axis=0)
+        reshaped_norm = norm.reshape((1, num_time_windows, len(focal), 1))
+
+    with np.errstate(invalid="ignore", divide="ignore"):
+        A /= reshaped_norm
+    A[np.all(A == 0, axis=3)] = np.nan
+
+    # Remove dimension for windows and/or time_windows if parameter is None
+    if not windows_used and time_windows_used:
+        A = A.reshape((num_time_windows, len(focal), len(reference_sets)))
+    elif not time_windows_used and windows_used:
+        A = A.reshape((num_windows, len(focal), len(reference_sets)))
+    elif not windows_used and not time_windows_used:
+        A = A.reshape((len(focal), len(reference_sets)))
+    return A
+
+
 class TestGenealogicalNearestNeighbours:
     """
     Tests the TreeSequence.genealogical_nearest_neighbours method.
@@ -432,14 +548,22 @@ class TestGenealogicalNearestNeighbours:
         A2 = tsutil.genealogical_nearest_neighbours(ts, focal, reference_sets)
         A3 = ts.genealogical_nearest_neighbours(focal, reference_sets)
         A4 = ts.genealogical_nearest_neighbours(focal, reference_sets, num_threads=3)
+        A5 = windowed_genealogical_nearest_neighbours(ts, focal, reference_sets)
         assert np.array_equal(A3, A4)
         assert A1.shape == A2.shape
         assert A1.shape == A3.shape
         assert np.allclose(A1, A2)
         assert np.allclose(A1, A3)
+        mask = ~np.isnan(A5)
+        assert np.sum(mask) > 0 or ts.num_edges == 0
+        assert np.allclose(A1[mask], A5[mask])
+        assert np.allclose(A5[mask], A2[mask])
+        assert np.allclose(A5[mask], A3[mask])
+
         if ts.num_edges > 0 and all(ts.node(u).is_sample() for u in focal):
             # When the focal nodes are samples, we can assert some stronger properties.
             assert np.allclose(np.sum(A1, axis=1), 1)
+            assert np.allclose(np.sum(A5, axis=1), 1)
         return A1
 
     def test_simple_example_all_samples(self):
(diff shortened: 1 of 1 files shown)


petrelharp:
That sounds great. Thinking through this out loud here - tell me if you agree: `span_normalize` controls whether the statistic is extensive along the genome or not: if `span_normalize=False`, then the value for [0, sequence_length) is the sum of the values of the sub-windows. The same thing is true for `time_normalize`, right? If `time_normalize=False`, then the value for a time_window of [0, 100) would be equal to the sum of the values for time_windows [0, 50) and [50, 100)?

 If so, I agree this is the right thing to toggle. I wonder if `time_normalize` is the right name, though, since users might at first assume that it divides by the length of time in the window.  I'm torn, though, because it does have to do with the time direction, as above. What do you think?

I'll look at the actual code a bit later. =)

petrelharp:
After looking at the code, a few more thoughts: First, very nice and elegant [normalization solution](https://github.com/tskit-dev/tskit/pull/683/files#diff-834587237cd20ab6f39a961ee7deefbcR477). I think this is just the way we want to do things more generally!

Here's why: One thing I wanted to check is that if time_normalize=False, that span_normalize=True is just dividing by the span of the window. This is not true... but, only when the focal node isn't in the same tree as any of the reference sets - ie, when we have missing data. This is what we should be doing with missing data in other statistics (normalizing by the non-missing span). So: very nice! And, with time_windows, the interpretation of time_nomralize=True for, say, diversity would be to see where along the genome the SNPs lie weighted by contribution to diversity, out of only those in that particular time bin. Which seems useful.

And, about the name?  From the code, clearly time_normalize and span_normalize are doing analogous things, so naming them to sound the same is a good idea. And, as you say, time_normalize makes it so that the sum in each time bin is the same, so it's a good name. So maybe we just want to put a big note in the docs that time_normalize doesn't divide by the length of the time bin, since any other name won't emphasize the parallel between the two types of normalization?

So: I like it all a lot, despite the worry that time_normalize connotes dividing by the length of the time window.

python/tests/test_stats.py
```diff
@@ -381,6 +381,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    genealogical_nearest_neighbours with support for span- and time-based windows
+    """
+    reference_set_map = np.zeros(ts.num_nodes, dtype=int) - 1
+    for k, reference_set in enumerate(reference_sets):
+        for u in reference_set:
+            if reference_set_map[u] != -1:
+                raise ValueError("Duplicate value in reference sets")
+            reference_set_map[u] = k
+
+    windows_parsed = ts.parse_windows(windows)
```
jeromekelleher, line 409: It's a bit simpler if you just say
```
windows = ts.parse_windows(windows)
```
there's no real reason to keep the original value lying around, and it's a potential source of errors.

python/tests/test_stats.py
```diff
@@ -381,6 +381,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    num_windows = windows_parsed.shape[0] - 1
+    time_windows_parsed = parse_time_windows(ts, time_windows)
```
jeromekelleher, line 411: Ditto here for ``time_windows``

python/tests/test_stats.py
```diff
@@ -381,6 +381,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+                p = u
+                while p != tskit.NULL:
+                    total = np.sum(sample_count[p])
+                    if total > delta:
+                        break
+                    p = parent[p]
+                if p != tskit.NULL:
+                    scale = span / (total - delta)
+                    time_index = np.searchsorted(time_windows_parsed, time[p]) - 1
+                    if time_index < num_time_windows and time_index >= 0:
```
jeromekelleher, line 462: ``if 0 <= time_index < num_time_windows`` is a bit more elegant.

python/tests/test_stats.py
```diff
@@ -381,6 +381,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+                        for k, _reference_set in enumerate(reference_sets):
```
jeromekelleher, line 463: ``for k in range(len(reference_sets))`` is a bit clearer.

jeromekelleher (commented):
This looks great Wilder! I haven't thought deeply about the normalisation, but I think the vast majority of what you have is right. If you want to plough ahead with the C version, then I'd say go for it.

ps. You have a stray binary file in the diff.

awohns:
Thanks for your very helpful comments @petrelharp!

>  `span_normalize` controls whether the statistic is extensive along the genome or not: if `span_normalize=False`, then the value for [0, sequence_length) is the sum of the values of the sub-windows. The same thing is true for `time_normalize`, right? If `time_normalize=False`, then the value for a time_window of [0, 100) would be equal to the sum of the values for time_windows [0, 50) and [50, 100)?

Yes, that's correct! 
Minor point: I was mistaken in my initial post, if both normalisations are _not_ used, then the value for each focal node summed over all windows, time_windows, and reference sets will be equal to the length of the sequence.

> After looking at the code, a few more thoughts: First, very nice and elegant [normalization solution](https://github.com/tskit-dev/tskit/pull/683/files#diff-834587237cd20ab6f39a961ee7deefbcR477). I think this is just the way we want to do things more generally!

Great!

> Here's why: One thing I wanted to check is that if time_normalize=False, that span_normalize=True is just dividing by the span of the window. This is not true... but, only when the focal node isn't in the same tree as any of the reference sets - ie, when we have missing data. This is what we should be doing with missing data in other statistics (normalizing by the non-missing span). So: very nice! 

Ah, I hadn't thought of this, thanks for spotting it!

> And, with time_windows, the interpretation of time_nomralize=True for, say, diversity would be to see where along the genome the SNPs lie weighted by contribution to diversity, out of only those in that particular time bin. Which seems useful.

I see, would be interesting to look at this with real data. 

> If so, I agree this is the right thing to toggle. I wonder if `time_normalize` is the right name, though, since users might at first assume that it divides by the length of time in the window. I'm torn, though, because it does have to do with the time direction, as above. What do you think?
> And, about the name? From the code, clearly time_normalize and span_normalize are doing analogous things, so naming them to sound the same is a good idea. And, as you say, time_normalize makes it so that the sum in each time bin is the same, so it's a good name. So maybe we just want to put a big note in the docs that time_normalize doesn't divide by the length of the time bin, since any other name won't emphasize the parallel between the two types of normalization?

Yes I see what you mean. It's normalising by the span of genome in the time bin, just as span_normalise is normalising by the span of genome in the window (which isn't always the length of the window, as you noted earlier). I see how it could be misleading if people were to focus more on the time part of the name, so perhaps a prominent note is the best way forward.

> So: I like it all a lot, despite the worry that time_normalize connotes dividing by the length of the time window.

Great, thanks again!

python/tests/test_stats.py
awohns: I kept the original around so I could reshape the output on lines 477-487. I'll rewrite to store a boolean if `windows` and `time_windows` are used.

jeromekelleher: Ah, I see. It's probably good idea to store the ``output_dims`` explicitly at the top then, so it's clear what you're deriving this from.

hyanwong:
It would be nice to get this PR into tskit in the near future. Am I right that it is mainly lacking a C implementation? The test suite is all there, right? And the approach has been agreed by @petrelharp ? If so, this seems like quite an easy thing to tick off, although I guess @awohns doesn't have time to do it at the moment, so we should look for other volunteers.

petrelharp:
My former self sounded happy with it...

awohns:
Hi all, I just brought this PR up to date and added the suggestions from @jeromekelleher. 
A few thoughts:
1. The one failing test represents a decision point. In the windowed GNN, I return nans when the normalisation values are 0. Genealogical nearest neighbors doesn't allow division by zero in the normalisation ([see here](https://github.com/tskit-dev/tskit/blob/20c423f055f03b84c98980e6e99febba9a6217d9/python/tests/test_stats.py#L380)). Hence the output is different between the two (shown here by the failing test on an empty tree sequence). What do others think is the best choice here?
2. I have a C version of windowed GNN in the works, but it will take a bit of time to PR. @hyanwong and others have mentioned that they would like a working version of windowed GNN to play around with. If this PR looks ok (after the decision point), shall we merge it, and make this python version of windowed GNN available for use (perhaps undocumented until the C version)? Or is it better to wait for the C version.
3. @petrelharp previously mentioned expanding the stats framework to incorporate time windows. Is this something you're still interested in doing, Peter? If so, I'd be happy to help!

petrelharp:
> @petrelharp previously mentioned expanding the stats framework to incorporate time windows. Is this something you're still interested in doing, Peter? If so, I'd be happy to help!

Yes, I'm interested, but can't work on it right now. (happy to help if you'd like to start, though?) As far as this PR goes, we should just make it forwards-compatible with how we will do time windows, and I think it is.

hyanwong:
> Hi all, I just brought this PR up to date and added the suggestions from @jeromekelleher.

Great, thanks a lot @awohns 
 
> 2. I have a C version of windowed GNN in the works, but it will take a bit of time to PR. @hyanwong and others have mentioned that they would like a working version of windowed GNN to play around with. If this PR looks ok (after the decision point), shall we merge it, and make this python version of windowed GNN available for use (perhaps undocumented until the C version)? Or is it better to wait for the C version.

I don't see why we should wait until a C version. I say merge after we've decided on point 1.

jeromekelleher:
> Genealogical nearest neighbors doesn't allow division by zero in the normalisation (see here). Hence the output is different between the two (shown here by the failing test on an empty tree sequence). What do others think is the best choice here?

Probably nans is the right thing to do. We should make it consistent with the rest of the stats API - what do these output when we have a division by zero on normalisation? (Nans, I suspect)

petrelharp:
I agree: `nan`.

jeromekelleher:
Also, no need for C version.

awohns:
Ok, I've changed the test to accommodate the fact that windowed-GNN returns nans when normalisation is zero (and GNN effectively does not) and made the function available as a `tree-sequence` method. I've also written documentation for the function. Let me know if there are issues with the documentation and if it's ok to have the function stuck into `trees.py` or if it should live elsewhere

jeromekelleher:
Sorry @awohns, I think were some crossed wires here. I don't think we want a new function ``windowed_xxx``, it would be better to update the existing gnn function to have the functionality we want.

So, I was imagining we merge your Python version as part of the test suite, and then do the C updates required to change the library function as a follow up. Does that sound reasonable?

hyanwong:
Is this ready to merge @awohns ? Does it need a review? It would be nice to get this in.

awohns:
I think it's good to go but it hasn't been merged yet because of the activity around the last release, is that right @jeromekelleher and @benjeffery?

benjeffery:
I didn't know it was ready! Will take a look at it.

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    reference_sets,
+    windows=None,
+    time_windows=None,
+    span_normalise=True,
+    time_normalise=True,
+):
+    """
+    genealogical_nearest_neighbours with support for span- and time-based windows
+    """
+    reference_set_map = np.zeros(ts.num_nodes, dtype=int) - 1
```
benjeffery, line 405: Helps readability to be specific about what the `-1`s are:
```suggestion
    reference_set_map = np.full(ts.num_nodes, tskit.NULL, dtype=int)
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    for k, reference_set in enumerate(reference_sets):
+        for u in reference_set:
+            if reference_set_map[u] != -1:
```
benjeffery, line 408:
```suggestion
            if reference_set_map[u] != tskit.NULL:
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+
+    windows_used = windows is not None
+    time_windows_used = time_windows is not None
+    windows = ts.parse_windows(windows)
+    num_windows = windows.shape[0] - 1
+    time_windows = parse_time_windows(ts, time_windows)
+    num_time_windows = time_windows.shape[0] - 1
+    A = np.zeros((num_windows, num_time_windows, len(focal), len(reference_sets)))
+    K = len(reference_sets)
+    parent = np.zeros(ts.num_nodes, dtype=int) - 1
```
benjeffery, line 420:
```suggestion
    parent = np.full(ts.num_nodes, tskit.NULL, dtype=int)
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    norm = np.zeros((num_windows, num_time_windows, len(focal)))
+
+    # Set the initial conditions.
+    for j in range(K):
+        sample_count[reference_sets[j], j] = 1
+
+    window_index = 0
+    for (t_left, t_right), edges_out, edges_in in ts.edge_diffs():
+        for edge in edges_out:
+            parent[edge.child] = -1
```
benjeffery, line 432:
```suggestion
            parent[edge.child] = tskit.NULL
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+            v = edge.parent
+            while v != -1:
```
benjeffery, line 434:
```suggestion
            while v != tskit.NULL:
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+                sample_count[v] -= sample_count[edge.child]
+                v = parent[v]
+        for edge in edges_in:
+            parent[edge.child] = edge.parent
+            v = edge.parent
+            while v != -1:
```
benjeffery, line 440:
```suggestion
            while v != tskit.NULL:
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+        while windows[window_index] < t_right and window_index + 1 <= num_windows:
+            w_left = windows[window_index]
+            w_right = windows[window_index + 1]
+            left = max(t_left, w_left)
+            right = min(t_right, w_right)
+            span = right - left
+            # Process this tree.
+            for j, u in enumerate(focal):
+                focal_reference_set = reference_set_map[u]
+                delta = int(focal_reference_set != -1)
```
benjeffery, line 455:
```suggestion
                delta = int(focal_reference_set != tskit.NULL)
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+                        for k in range(len(reference_sets)):
+                            n = sample_count[p, k] - int(focal_reference_set == k)
+                            A[window_index, time_index, j, k] += n * scale
+                        norm[window_index, time_index, j] += span
+            assert span > 0
+            if w_right <= t_right:
+                window_index += 1
+            else:
+                # This interval crosses a tree boundary, so we update it again
+                # in the for the next tree
```
benjeffery, line 475:
```suggestion
                # in the next tree
```

python/tests/test_stats.py
```diff
@@ -384,7 +384,123 @@ def naive_genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    if not windows_used and time_windows_used:
+        A = A.reshape((num_time_windows, len(focal), len(reference_sets)))
+    elif not time_windows_used and windows_used:
+        A = A.reshape((num_windows, len(focal), len(reference_sets)))
+    elif not windows_used and not time_windows_used:
+        A = A.reshape((len(focal), len(reference_sets)))
+    return A
+
+
+class TestGenealogicalNearestNeighbours(unittest.TestCase):
```
benjeffery, line 503: Don't think you need the inheritance
```suggestion
class TestGenealogicalNearestNeighbours:
```

python/tests/test_stats.py
```diff
@@ -426,14 +542,20 @@ def verify(self, ts, reference_sets, focal=None):
...
+        A5 = windowed_genealogical_nearest_neighbours(ts, focal, reference_sets)
         assert np.array_equal(A3, A4)
         assert A1.shape == A2.shape
         assert A1.shape == A3.shape
         assert np.allclose(A1, A2)
         assert np.allclose(A1, A3)
+        if not np.any(np.isnan(A5)):
+            assert np.allclose(A1, A5)
+            assert np.allclose(A5, A2)
+            assert np.allclose(A5, A3)
```
benjeffery, line 554: Might as well check the non-nans.
```suggestion
        mask = ~np.isnan(A5)
        assert np.sum(mask) > 0 or ts.num_edges == 0
        assert np.allclose(A1[mask], A5[mask])
        assert np.allclose(A5[mask], A2[mask])
        assert np.allclose(A5[mask], A3[mask])

```

python/tests/test_stats.py
```diff
@@ -653,6 +775,284 @@ def test_empty_ts(self):
...
+                [[[np.nan, np.nan]], [[np.nan, np.nan]]],
+            ],
+            equal_nan=True,
+        )
+
+    def test_span_normalise(self):
+        ts = self.get_two_tree_ts()
+        sample_sets = [[0, 1], [2]]
+        focal = [0]
+        windows = ts.sequence_length * np.random.uniform(size=5)
```
benjeffery, line 952: Is this OK without a seed?

benjeffery (commented):
Few minor comments, thanks for picking this up!

python/tests/test_stats.py
awohns: I think it's ok to have some variability?

awohns: This is actually a general question I have, I thought all tests should be deterministic, but I recall another issue where @jeromekelleher suggested leaving seeds out

benjeffery: Well as long as there is no pathological case where things fail, and almost all of the possible combos are a meaningful test then its fine.

petrelharp: I vote to set a seed, so that if it *does* fail then we'll be able to debug it. If it's not being adequately tested with a single seed, then we need more tests.

jeromekelleher: All unit tests should have seeds -- the test suite should be fully deterministic.

awohns:
Seems like an unrelated test: `testNodeGeneticRelatedness::test_many_trees_sequence_length` failed, not sure why

hyanwong:
It's associated with a `numpy.ufunc` error. I think @benjeffery sorted out something with a similar error this morning.

benjeffery:
Hmm, not sure why this is happening intermittently, and only on circle. Will look into it Monday.

benjeffery:
@awohns CI seems to be happy now.

jeromekelleher:
Ready to merge?

awohns:
Yep this is good to go I think!

jeromekelleher (approved):

hyanwong:
While I remember, this is the nicest way to get a gnn plot from the result, which we could put in a tutorial:

```python
focal_node
sample_sets = {ts.population(p).metadata["name"]: ts.samples(p) for p in [0,1,2]}  # compare to populations 0, 1, 2 
gnn = windowed_genealogical_nearest_neighbours(ts, focal, list(sample_sets.values()), windows=windows)
windows=np.linspace(0, ts.sequence_length, num_windows+1)
plt.stackplot(windows[1:], gnn[:,0,:].T, labels= sample_sets.keys(), step='pre')
plt.xlim(left=0)
```

PR #691, opened 2020-06-29
Changed simplify_segment_t to tsk_segment_t.

gtsambos:
Just a minor fix for a TODO spotted in tables.c:

> TODO: This [simplify_segment_t] should be renamed to tsk_segment_t when we move to tskit. msprime can then #include this and also use it. msprime needs a different segment definition, which it can continue to call 'segment_t' as it doesn't export a C API.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4526,16 +4526,12 @@ table_sorter_free(table_sorter_t *self)
  * segment overlapper
  *************************/
 
-/* TODO: This should be renamed to tsk_segment_t when we move to tskit.
- * msprime can then #include this and also use it. msprime needs a different
- * segment definition, which it can continue to call 'segment_t' as it
- * doesn't export a C API. */
-typedef struct _simplify_segment_t {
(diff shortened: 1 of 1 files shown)


gtsambos:
Hmm. I don't know why clang-format is failing, as all the tests are fine on my local machine

c/tskit/tables.c
```diff
@@ -4530,12 +4530,12 @@ table_sorter_free(table_sorter_t *self)
  * msprime can then #include this and also use it. msprime needs a different
```
jeromekelleher, line 4530: We should delete this comment - it'll just be confusing now that the TODO is a DONE!

jeromekelleher:
Re clang-format, I think you just need to run it on the code. The alignment has changed slightly (clang-format isn't run automatically by pre-commit because we don't want people who are just working with the Python code to have to install it. See [here](https://tskit.readthedocs.io/en/latest/development.html#sec-development-c-code-style) for details.

PR #702, opened 2020-07-01
Add summary tables to Tree and TreeSequence docs.

grahamgower:
I often find myself forgetting the name of a `Tree` or `TreeSequence` method/attribute, or to which class it belongs. And as both classes have many methods, simply finding the correct place in the docs is more time consuming than it needs to be. So this change inserts `Methods` and `Attributes` summary tables at the top of each of these classes.

autodocsumm

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -124,6 +124,7 @@ extensions = [
     "breathe",
     "sphinxarg.ext",
     "sphinx_issues",
+    "autodocsumm",
 ]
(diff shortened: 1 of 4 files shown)


benjeffery:
This is great and one of my pain points with the docs - been meaning to look into how to do this!

jeromekelleher:
Yes, this is awesome, thanks @grahamgower!

PR #709, opened 2020-07-08
Refactor table collection integrity checks and document.

### LINKED ISSUE #592: [C API] Document tsk_table_collection_check_integrity()
In the course of tracking down a bug in SLiM having to do with an uncaught (by SLiM) overflow of a tskit table (didn't check the pesky return value!), @molpopgen wrote a little snippet of debugging code, which I modified into this form:

			std::size_t debug_edge_index = 0;
			
			for( ; debug_edge_index < tables_.edges.num_rows ; ++debug_edge_index)
			{
				if ((tables_.edges.parent[debug_edge_index] >= tables_.nodes.num_rows) || (tables_.edges.parent[debug_edge_index] < 0))
				{
					std::cout << "parent index out of range: " << tables_.edges.parent[debug_edge_index] << ", " << tables_.nodes.num_rows << std::endl;
					break;
				}
				if ((tables_.edges.child[debug_edge_index] >= tables_.nodes.num_rows) || (tables_.edges.child[debug_edge_index] < 0))
				{
					std::cout << "child index out of range: " << tables_.edges.child[debug_edge_index] << ", " << tables_.nodes.num_rows << std::endl;
					break;
				}
			}

It caught the problem and provided the key information needed to track down the bug.  It's not rocket science, obviously, but it got me to thinking that it would be useful for tskit to have a "consistency check" function that would check all sorts of things about a table collection: that all indices are within range, that things that ought to correspond with each other do correspond with each other, etc.  That the table collection is, in every respect one can think of, well-formed.  Client code like SLiM, fwdpp, etc., could call that check as often as desired to make sure that nothing untoward has occurred.  I'll tag @petrelharp here as well since he was in the previous discussion / bug-hunt.  :->

### LINKED ISSUE #649: [C API] Remove CHECK_OFFSETS and make part of minimal integrity checks
Currently option 0 means check the integrity of object references, and to check offsets we need to specify CHECK_OFFSETS separately to check whether the offsets in ragged columns are valid. I don't see a good reason for this, and the basic default checks should make sure that the tables at least contain valid memory.

Should do this as part of the process of documenting this API.

See also #596, #592

jeromekelleher:
This PR documents the tsk_table_collection_check_integrity function, documents it and refactors the set of options a bit.

As part of the process of documenting the options for this function I realised that they were a confusing mishmash, and I spent a long time trying to figure some simple and clear semantics. We had made special cases of the mutation parents and mutation times basically for the sake of ``compute_mutation_parents``, but it's actually much simpler to just set the values to NULL beforehand.

We do change the semantics of mutation time a little bit here too. Mixing known and unknown times at a site is fine, but we can have situations where a string of unknown time mutations separate mutations with times, and we need to check these times don't conflict.

Hopefully the whole thing is a bit easier to follow now, and maybe even a little bit more efficient too.

@benjeffery, @molpopgen, would you mind taking a look at the documentation and see if it makes sense to you as C API users? (Any other comments would be much appreciated too, of course!)

Closes #649
Closes #592


benjeffery:
Will take a look at this later today.

c/tskit/tables.h
```diff
@@ -2558,7 +2555,42 @@ int tsk_table_collection_build_index(tsk_table_collection_t *self, tsk_flags_t o
...
+    Check contraints on the ordering of mutations. Any non-null
+    mutation parents and known times are checked for ordering
+    constraints.
+TSK_CHECK_INDEXES
+    Check that the table indexes exist, and contain valid edge
+    references.
+TSK_CHECK_ALL
+    All checks needed to define a valid tree sequence that
+    can be determined without building the individual
+    trees.
```
benjeffery, line 2590: Might be nice to list out the flags this turns on ?

jeromekelleher: Maybe, but I was thinking that this might get out of date as we add more. Anyway, it's all the TSK_CHECK* above, so hopefully that'll be clear.

c/tskit/tables.h
```diff
@@ -2558,7 +2555,42 @@ int tsk_table_collection_build_index(tsk_table_collection_t *self, tsk_flags_t o
...
 @rst
-TODO: document me. https://github.com/tskit-dev/tskit/issues/592
+
+Checks the integrity of this table collection. The default checks (i.e., with
+options = 0) guarantee the integrity of memory and references within the
+table collection. All spatial values (along the genome) are checked
+to see if they are finite values and within the required bounds. Time values
+are checked to see if they are finite or marked as unknown.
+
+To check if a set of tables
```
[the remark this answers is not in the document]
jeromekelleher: It's probably better to say this twice though, I'll update.

c/tskit/tables.h
benjeffery: "Defensive documentation" :smile: You're prob right the marginal utility of listing is outweighed by the probability of us forgetting to change it.

jeromekelleher:
Thanks for the review @benjeffery. Clarifying this made we realise that we can fix an annoying problem with the tree sequence object, in that we can now detect any bad topologies at load time. There's no good reason not to I think, other than potentially breaking people's code. But, if it depends on this, then I'm sure it must be broken anyway.

jeromekelleher:
I considered changing TSK_CHECK_ALL to include an actual check of the trees too. In a way, this would be quite handy because we're actually depending on the trees in ``compute_mutation_parents`` and so on. We may not even be checking for these bad topologies at the moment in there.

I backed out of it because it seemed like this would imply another iteration over the trees at tsk_treeseq_t load time, because we would still have to iterate over them once to count the trees. Maybe it's not worth worrying about this, but I figured it's better to just let this one go and get the code shipped.

benjeffery:
Great, I think moving the contradictory children check is the right thing.

As for getting ALL to check the trees - I considered check_integrity returning the number of trees meaning that - `tsk_treeseq_init_trees` wouldn't need to iterate over the trees. But on checking `tsk_treeseq_init_trees` also calculates breakpoints in a separate loop. From a cursory inspection I couldn't tell why this is done in a separate loop?

jeromekelleher:
> As for getting ALL to check the trees - I considered check_integrity returning the number of trees meaning that - tsk_treeseq_init_trees wouldn't need to iterate over the trees. But on checking tsk_treeseq_init_trees also calculates breakpoints in a separate loop. From a cursory inspection I couldn't tell why this is done in a separate loop?

It's a good idea, but I think it would complicate the interface a bit if ``check_integrity`` returned 0 on success sometimes and the number of trees other times. Maybe it's not that bad though, if we have TSK_CHECK_TREES instead of TSK_CHECK_ALL, which returns the number of trees. It would save a pass through the edges at load time, which must add up for big tree sequences.

It wouldn't take long to do, and it probably is the right thing to do, so we keep all the validation checks in one place. What do you reckon?

We need to know the number of trees so we can malloc stuff on a per-tree basis. I don't think there's any way around that.

jeromekelleher:
Note this would change the semantics a bit, so that TSK_CHECK_TREES would imply all the other checks, and not be the OR of the bits. We can't check the trees without all the other things being safe first. Would need a little bit of a refactor but not that much.

benjeffery:
> We need to know the number of trees so we can malloc stuff on a per-tree basis. I don't think there's any way around that.
Ah of course.  I guess there would be a perf hit to mallocing a reasonably sized buffer and then realloc'ing to double it every time you hit the end? I don't yet have a good intuition for how long tree iteration takes compared to that kind of operation. 

As for the return value - we have other operations (e.g. `add_row`) using positive returns, and if it only does this on `TSK_CHECK_TREES` then none of the existing `!= 0` need to change. It never felt right to have the checks split across two functions and this is a way to do it without an extra pass.

jeromekelleher:
Ready for another look @benjeffery

c/tests/test_tables.c
```diff
@@ -3291,7 +3305,7 @@ test_table_collection_check_integrity(void)
     CU_ASSERT_FATAL(ret >= 0);
     ret = tsk_table_collection_check_integrity(&tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-    ret = tsk_table_collection_check_integrity(&tables, TSK_CHECK_MUTATION_TIME);
+    ret = tsk_table_collection_check_integrity(&tables, TSK_CHECK_MUTATION_ORDERING);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION);
 
     /* Note that we do not test for TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE as
```
benjeffery, line 3311: This comment is out of date - the test for TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE should be added here now.

molpopgen:
Just found this in my notifications.  Taking a look.

molpopgen:
@jeromekelleher -- I think this all looks sensible.  I'll admit that I've not dug into the details of this API before, though.

Issue #708, opened 2020-07-08
Edge metadata is not sorted when edges are.


benjeffery:
#496 added metadata to edges, however `table_sorter_sort_edges` doesn't sort the metadata, leading to edges' metadata being mixed.

jeromekelleher:
Good catch, thanks @benjeffery!

jeromekelleher:
Closed in #722

PR #711, opened 2020-07-08
Add example of sorting the edge table with a custom C++ function.

molpopgen:
See #708

### DIFF
--- a/c/examples/cpp_sorting_example.cpp
+++ b/c/examples/cpp_sorting_example.cpp
@@ -1,0 +1,102 @@
+#include <cstddef>
+#include <vector>
+#include <algorithm>
+#include <stdexcept>
+#include <sstream>
+#include <iostream>
+#include <type_traits>
+#include <tskit.h>
+
+static void
+handle_tskit_return_code(int code)
+{
+    if (code != 0) {
+        std::ostringstream o;
+        o << tsk_strerror(code);
+        throw std::runtime_error(o.str());
+    }
+}
+
+struct edge_plus_time {
+    double time;
+    tsk_id_t parent, child;
+    double left, right;
+};
+
+int
+sort_edges(tsk_table_sorter_t *sorter, tsk_size_t start)
+{
+    if (sorter->tables->edges.metadata_length != 0) {
+        throw std::invalid_argument(
+            "the sorter does not currently handle edge metadata");
+    }
+    if (start != 0) {
+        throw std::invalid_argument("the sorter requires start==0");
+    }
+
+    std::vector<edge_plus_time> temp;
+    temp.reserve(static_cast<std::size_t>(sorter->tables->edges.num_rows));
+
+    auto edges = &sorter->tables->edges;
+    auto nodes = &sorter->tables->nodes;
+
+    for (tsk_size_t i = 0; i < sorter->tables->edges.num_rows; ++i) {
+        temp.push_back(edge_plus_time{ nodes->time[edges->parent[i]], edges->parent[i],
+            edges->child[i], edges->left[i], edges->right[i] });
+    }
+
+    std::sort(begin(temp), end(temp),
+        [](const edge_plus_time &lhs, const edge_plus_time &rhs) {
+            if (lhs.time == rhs.time) {
+                if (lhs.parent == rhs.parent) {
+                    if (lhs.child == rhs.child) {
+                        return lhs.left < rhs.left;
+                    }
+                    return lhs.child < rhs.child;
+                }
+                return lhs.parent < rhs.parent;
+            }
+            return lhs.time < rhs.time;
+        });
+
+    for (std::size_t i = 0; i < temp.size(); ++i) {
+        edges->left[i] = temp[i].left;
+        edges->right[i] = temp[i].right;
+        edges->parent[i] = temp[i].parent;
+        edges->child[i] = temp[i].child;
+    }
+
+    return 0;
+}
+
+int
+main(int argc, char **argv)
+{
+    if (argc != 3) {
+        std::cerr << "Usage: " << argv[0] << " input.trees output.trees\n";
+        std::exit(0);
+    }
+    const char *infile = argv[1];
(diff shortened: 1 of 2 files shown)


c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+    const auto *edges = &sorter->tables->edges;
+    const auto *nodes = &sorter->tables->nodes;
+
+    // NOTE: this future-proofs the type against changing to int64_t
+    // for tsk_id_t.  It will also look odd to a C programmer,
+    // but the edges->parent[0] need not be valid at run time.
+    // Rather, it is checking what the return type of the expression
+    // would be, which is tsk_id_t& (with the & being due to decltype
+    // not returning the decay type).
+    const auto num_nodes
```
jeromekelleher, line 49: Maybe I'm being dense here, but why not use ``tsk_size_t num_nodes = nodes->num_rows``? Most of the point of the typedef is that allows us to change to 64 bit later, if we want. (The rest of the point is readability, but some would argue against that and say it's better to be explicit about what the types actually are.)

Likewise for the other types - I think it'd be more readable to use tsk_size_t than type inference.

c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+    for (decltype(sorter->tables->edges.num_rows) i = 0;
+         i < sorter->tables->edges.num_rows; ++i) {
+        if (edges->parent[i] >= num_nodes || edges->child[i] >= num_nodes) {
+            throw std::invalid_argument("edge table contains invalid node ids");
+        }
+        temp.push_back(edge_plus_time{ nodes->time[edges->parent[i]], edges->parent[i],
+            edges->child[i], edges->left[i], edges->right[i] });
+    }
+
+    if (temp.size() != static_cast<std::size_t>(edges->num_rows)) {
```
jeromekelleher, line 62: I don't think this can actually happen, can it?

c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // Rather, it is checking what the return type of the expression
+    // would be, which is tsk_id_t& (with the & being due to decltype
+    // not returning the decay type).
+    const auto num_nodes
+        = static_cast<std::remove_reference<decltype(edges->parent[0])>::type>(
+            nodes->num_rows);
+
+    for (decltype(sorter->tables->edges.num_rows) i = 0;
+         i < sorter->tables->edges.num_rows; ++i) {
+        if (edges->parent[i] >= num_nodes || edges->child[i] >= num_nodes) {
```
jeromekelleher, line 55: We don't have to worry about his within the sorter routine, as this will have been checked already by ``tsk_table_collection_check_integrity`` (unless we call sorter_run() with TSK_NO_CHECK_INTEGRITY)

c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+    tsk_table_collection_t tables;
+    auto ret = tsk_table_collection_load(&tables, infile, 0);
+    handle_tskit_return_code(ret);
+
+    tsk_table_sorter_t sorter;
+    ret = tsk_table_sorter_init(&sorter, &tables, 0);
+    handle_tskit_return_code(ret);
+    sorter.sort_edges = sort_edges;
+    try {
+        ret = tsk_table_sorter_run(&sorter, NULL);
```
jeromekelleher, line 112: Cool, so exceptions will propagate up through the stack? That's very handy.

c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+                        return lhs.left < rhs.left;
+                    }
+                    return lhs.child < rhs.child;
+                }
+                return lhs.parent < rhs.parent;
+            }
+            return lhs.time < rhs.time;
+        });
+
+    // NOTE: at this point, exceptions are no longer
```
jeromekelleher, line 80: I don't think we need the comment here. This example is for documentation, so we want it to be nice and concise.

c/examples/cpp_sorting_example.cpp
```diff
@@ -0,0 +1,121 @@
...
+struct edge_plus_time {
+    double time;
+    tsk_id_t parent, child;
+    double left, right;
+};
+
+int
+sort_edges(tsk_table_sorter_t *sorter, tsk_size_t start)
+{
+    // if (sorter->tables->edges.metadata != nullptr) {
```
jeromekelleher, line 29: Better to say ``sorter->tables->edges.metadata_length != 0``, as this will always work.

molpopgen: Ah, that makes sense.  The exception was triggered by output from msprime.

c/examples/cpp_sorting_example.cpp
molpopgen: Yes, they'll propagate.  C++ is responsible for the stack here.

c/examples/cpp_sorting_example.cpp
molpopgen: Typically, one would prefer to just use `auto` here, following the [AAA](https://herbsutter.com/2013/06/13/gotw-94-special-edition-aaa-style-almost-always-auto/) rule of thumb ("code for the interface, not the implementation"). However, that leads to cast warnings later, but those would go away if the node check during copy disappears.

c/examples/cpp_sorting_example.cpp
molpopgen: Yeah, so that's a little tricky.  I'm removing the check, but invalid ids can then fail silently.

jeromekelleher: It can't though - tsk_table_collection_check_integrity has already checked this during ``sorter_init``, unless you explicitly tell it not to.

Issue #705, opened 2020-07-02
make sort stricter


petrelharp:
With `subset`, we can reorder tables. But we still can't easily take two sets of completely equivalent tables and put them in the same order. From [the requirements](https://tskit.readthedocs.io/en/latest/data-model.html#valid-tree-sequence-requirements):

- nodes we can order using subset
- pops and indivs will be ordered by earliest node by subset
- edge order is undefined among parents of the same time
- mutation order is undefined among mutations of the same time
- migration order is undefined among migrations of the same time

The ordering could be total by adding `parent` to the edge sort key, and `node` to the mutation and migration sort keys.

This would be nice to do since when doing `union` #623 we would like to make sure that when subsetting on the shared nodes the tables are equal. It'd probably be useful for other testing.

Table sorting is a bottleneck in forwards sims, so perhaps we want to make the "strict sort" optional; and if `strict=True` we first sort with the above additions to the keys, then do `subset(range(num_nodes))`.  This would be guaranteed to always get the same order no matter the original order of the tables.

jeromekelleher:
Sounds good @petrelharp, I'm all for making a canonical form. I agree we can't introduce these extra restrictions as requirements. I doubt it would make any difference performance-wise, but I don't think we can start sorting populations and individuals by default because this will surely break people's code.

I think we should use "canonical" as the term rather than strict, to emphasise we're putting tables into a canonical form.

mufernando:
Well, it seems the sorting of edges surpass the requirements, because in practice they are being sorted by time, then parent node id, child node id and lastly the left coordinate:

https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L4325-L4344

So we only need to do add sorting of mutations by nodes if time is tied.

petrelharp:
Ok, great! We should make this clear in the documentation.

mufernando:
Ah, also curious about #627, which changes the sorting API.

Should we go off of that?

jeromekelleher:
> Ah, also curious about #627, which changes the sorting API.

Hah, good plot @mufernando. Yes, it would probably be best to base your code off that as I've made some changes there. Better still, why don't I just finish that PR...

mufernando:
Some notes on a convo @petrelharp and I had earlier today:

1. The proposed stricter sorting would make the tables canonical given the order of the nodes.
2. It's not possible to make the tables canonical without specifying the order of at least the leaves.
[...]

PR #715, opened 2020-07-09
canonical sorting

mufernando:
Currently, sort does not put a Table Collection in canonical form (given the order of nodes) -- see #705 .

Here I am working on adding mutation time, mutation parent and node to the mutation sorting key.

(still very early, working on a python impl of sort first!)

### DIFF
--- a/python/tests/tsutil.py
+++ b/python/tests/tsutil.py
@@ -730,7 +730,7 @@ def orig_site_keys(i, tables):
     return tables.sites.position[i]
 
 
-def orig_mut_keys(i, tables, sorted_sites):
+def orig_mut_keys(i, tables, sorted_sites, **kwargs):
     orig_site = tables.mutations.site[i]
     return sorted_sites.index(orig_site)
 
--- a/python/tests/tsutil.py
+++ b/python/tests/tsutil.py
@@ -680,6 +680,111 @@ def compute_mutation_times(ts):
     return times
 
 
+def disorder_ts(ts, seed):
+    """
+    Randomizes the order the Edge, Site and Mutation Tables rows.
+    :param TreeSequence ts: The tree sequence to disorders the tables from.
+        Need not have a valid mutation time column.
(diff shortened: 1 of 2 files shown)


python/tests/tsutil.py
```diff
@@ -680,6 +680,111 @@ def compute_mutation_times(ts):
...
+        randomised_mutations.extend(s.mutations)
+    rng.shuffle(randomised_mutations)
+    for m in randomised_mutations:
+        tables.mutations.add_row(
+            site=site_id_map[m.site],
+            node=m.node,
+            derived_state=m.derived_state,
+            parent=m.parent,
+            metadata=m.metadata,
+            # time=m.time,
```
jeromekelleher, line 718: Time should be working, right?

jeromekelleher:
Looks great @mufernando, this is exatctly the right approach. Ping me when you'd like some feedback.

jeromekelleher:
@mufernando - is this something that we want to get in for the C API release, or will having it done for the next Python release be enough? We'd really like to get the C API shipped soon, and I think just getting the sorting requirement done will be enough for that. Is that OK, or do you want the ``canonicalise`` (or whatever) method to do into the C API in time for the next release too? (The only practical reason for this is if you want to use it in SLiM, I think)

mufernando:
I don't need this in SLiM and probably won't be able to get to this over the next week or two, so we can wait for the next release.

benjeffery:
@mufernando Just pinging as this PR has been inactive for a while - do you plan to pick it up?

mufernando:
I don't think I'll have the time to work on this for a while. But it seems the work I did with validating the C sort with a Python implementation could be incorporated. Is it bad to leave this hanging here?

benjeffery:
It's no problem to leave it - just useful for planning releases to know which PRs are still active.

benjeffery:
Superseded by #1108

PR #717, opened 2020-07-11
Fix edge metadata sorting and prevent processing edge metadata in other functions.

### LINKED ISSUE #708: Edge metadata is not sorted when edges are.

benjeffery:
Fixes #708 

Note that this is stacked on #712 so the diff looks bigger than it is.

To test I have made `parse_edges` return metadata by default. This is a little odd - but as one row is many edges in parse edges there doesn't seem to be a clean way to specify per-edge metadata there anyway.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -757,6 +757,7 @@ test_edge_table_copy_semantics(void)
     int ret;
     tsk_treeseq_t ts;
     tsk_table_collection_t t1, t2;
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Gah, CircleCI won't work now as it has an older version of msprime in its cache. I think the simplest thing to do is reopen this PR from my fork.

PR #726, opened 2020-07-20
Sort by mutation time and check sort in integrity

### LINKED ISSUE #651: Add requirement that mutations are sorted by time
> A though: once this is implemented, we should have `sort_tables` sort mutations by time, which satisfies the current requirement that parents come before children, but would make the ordering produced by sort_tables less ambiguous (not totally, if in discrete time, though). Perhaps we should make "sorted by time" a requirement of the mutation table, also.

_Originally posted by @petrelharp in https://github.com/tskit-dev/tskit/issues/513#issuecomment-634211466_

benjeffery:
Fixes #651 

Needs python test work now.

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -286,7 +286,8 @@ tsk_strerror_internal(int err)
             ret = "Inconsistent mutations: state already equal to derived state";
             break;
         case TSK_ERR_UNSORTED_MUTATIONS:
-            ret = "Mutations must be provided in non-decreasing site order";
+            ret = "Mutations must be provided in non-decreasing site order and "
+                  "non-increasing time order within each site";
             break;
         case TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE:
             ret = "A mutation's time must be >= the node time, or be marked as "
@@ -299,6 +300,10 @@ tsk_strerror_internal(int err)
         case TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE:
             ret = "A mutation's time must be < the parent node of the edge on which it "
                   "occurs, or be marked as 'unknown'";
+            break;
+        case TSK_ERR_MUTATION_TIME_HAS_BOTH_KNOWN_AND_UNKNOWN:
+            ret = "Mutation times must either be all marked 'unknown', or all be known "
+                  "values";
             break;
 
         /* Sample errors */
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4343,9 +4343,12 @@ cmp_mutation(const void *a, const void *b)
     const tsk_mutation_t *ib = (const tsk_mutation_t *) b;
     /* Compare mutations by site */
     int ret = (ia->site > ib->site) - (ia->site < ib->site);
+    /* Within a particular site sort by time if known, then ID. This ensures that
+     * relative ordering within a site is maintained */
+    if (ret == 0 && !tsk_is_unknown_time(ia->time) && !tsk_is_unknown_time(ib->time)) {
+        ret = (ia->time < ib->time) - (ia->time > ib->time);
+    }
     if (ret == 0) {
-        /* Within a particular site sort by ID. This ensures that relative ordering
-         * within a site is maintained */
         ret = (ia->id > ib->id) - (ia->id < ib->id);
     }
     return ret;
@@ -6658,7 +6661,8 @@ tsk_table_collection_check_mutation_integrity(
     int ret = 0;
     tsk_size_t j;
     tsk_id_t parent_mut;
-    double mutation_time, parent_mutation_time;
+    double mutation_time;
+    double last_known_time = INFINITY;
     const tsk_mutation_table_t mutations = self->mutations;
     const tsk_id_t num_nodes = (tsk_id_t) self->nodes.num_rows;
     const tsk_id_t num_sites = (tsk_id_t) self->sites.num_rows;
@@ -6666,6 +6670,7 @@ tsk_table_collection_check_mutation_integrity(
     const double *node_time = self->nodes.time;
     const bool check_mutation_ordering = !!(options & TSK_CHECK_MUTATION_ORDERING);
     bool unknown_time;
+    bool unknown_times_seen = false;
 
     for (j = 0; j < mutations.num_rows; j++) {
         /* Basic reference integrity */
@@ -6696,12 +6701,23 @@ tsk_table_collection_check_mutation_integrity(
                 goto out;
             }
         }
+        /* Check known/unknown times are not both present */
+        if (unknown_time) {
+            unknown_times_seen = true;
+        } else if (unknown_times_seen) {
+            ret = TSK_ERR_MUTATION_TIME_HAS_BOTH_KNOWN_AND_UNKNOWN;
+            goto out;
+        }
 
         if (check_mutation_ordering) {
+            /* Check site ordering and reset time check if needed*/
             if (j > 0) {
                 if (mutations.site[j - 1] > mutations.site[j]) {
                     ret = TSK_ERR_UNSORTED_MUTATIONS;
                     goto out;
+                }
+                if (mutations.site[j - 1] != mutations.site[j]) {
+                    last_known_time = INFINITY;
                 }
             }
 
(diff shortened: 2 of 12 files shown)


c/tskit/tables.c
```diff
@@ -6742,6 +6747,23 @@ tsk_table_collection_check_mutation_integrity(
                     parent_mut = mutations.parent[parent_mut];
                 }
             }
+            
+            /* Check time ordering, we do this after the time checks above, so that
+               more specific errors trigger first */
+            if (!tsk_is_unknown_time(last_known_time) &&
```
jeromekelleher, line 6753: It's a bit confusing to have to check ``last_known_time`` is known - could we avoid this by setting ``last_known_time`` to -1 initially or something?

jeromekelleher: We've also already checked for ``!unknown_time`` just above - why not keep this in the same block?

benjeffery: Negative times are valid, so using unknown seemed the best way to say "no known time has been seen for this site".

jeromekelleher: Aha, so they are. What about -inf then? This is always less than everything, isn't it, and we check for ``isfinite(time)`` before this.

benjeffery: I've used positive infinity (parents before children right?) in https://github.com/tskit-dev/tskit/pull/726/commits/3057d9897f39cd252170a256005f2a41b03cc1e2

c/tskit/tables.c
```diff
@@ -6742,6 +6747,23 @@ tsk_table_collection_check_mutation_integrity(
...
+                !unknown_time &&
+                j > 0 &&
+                mutations.site[j - 1] == mutations.site[j] &&
+                mutation_time > last_known_time
+            ){
+                    ret = TSK_ERR_UNSORTED_MUTATIONS;
+                    goto out;
+            }
+
+            if (!unknown_time || (j>0 && mutations.site[j - 1] != mutations.site[j])) {
```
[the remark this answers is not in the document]
benjeffery: I've modified the tests to cover this sloppy thinking! Fixed in https://github.com/tskit-dev/tskit/pull/726/commits/3057d9897f39cd252170a256005f2a41b03cc1e2

c/tskit/tables.c
```diff
@@ -6719,6 +6724,10 @@ tsk_table_collection_check_mutation_integrity(
                 }
             }
 
+            if (j > 0 && mutations.site[j - 1] != mutations.site[j]) {
```
jeromekelleher, line 6727: I'm nitpicking here, but we could avoid the j > 0 comparison and be a bit more cache friendly if we put this check into the (j > 0) block around 6706. It probably doesn't matter, but we do run this every time we create a tree sequence object, so it makes sense to keep it as tight as we can.;

benjeffery: Nah, not a nickpit! Moved in https://github.com/tskit-dev/tskit/pull/726/commits/79dd9ecd10227631a7dec75c722fb109bb98cc9b

benjeffery:
@jeromekelleher RFR! Happy to hear suggestions for further tests or docs I missed.

c/tskit/core.c
```diff
@@ -286,7 +286,9 @@ tsk_strerror_internal(int err)
             ret = "Inconsistent mutations: state already equal to derived state";
             break;
         case TSK_ERR_UNSORTED_MUTATIONS:
-            ret = "Mutations must be provided in non-decreasing site order";
+            ret = "Mutations must be provided in non-decreasing site order and "
+                  "decreasing "
```
jeromekelleher, line 290: Join lines here

benjeffery:
Actually, hold fire - seems windows wants to put unknown first and `glibc` wants to put it last.

docs/data-model.rst
```diff
@@ -653,7 +653,8 @@ Furthermore,
 For simplicity and algorithmic efficiency, mutations must also:
 
 - be sorted by site ID;
-- when there are multiple mutations per site, parent mutations must occur
+- when there are multiple mutations per site, mutations should be ordered by
+  decreasing time, if known, and parent mutations must occur
```
jeromekelleher, line 657: s/decreasing/nonincreasing

Adjacent times can be equal, right? Also in the changelong.

python/tests/test_highlevel.py
```diff
@@ -754,6 +754,7 @@ def test_compute_mutation_time(self):
                 np.allclose(python_time, tables.mutations.time, rtol=1e-15, atol=1e-15)
             )
             # Check we have valid times
+            tables.sort()
```
jeromekelleher, line 757: Why do we need to sort here? This would be a bug in compute_mutation_times, wouldn't it?

benjeffery:
So for the current `cmp_mutation` in this PR if you have three mutations with times 4, UNK, 5 (and hence ids 0,1,2) then you get:
5<4, 4<U, U<5 which is inconsistent. With glibc this results in 4,5,U and windows does U,4,5. We could fix this by saying unknown times are either + or - infinity. BUT that might mean we broke any ordering that was present due to the "parents first" rule. I'm not sure how to fix this as check for a parent chain in `cmp_mutation` sounds expensive.

python/tests/test_highlevel.py
benjeffery: This is failing on the example tree from `insert_branch_mutations` looking into it.

benjeffery: Massive confusion till I realised this test uses a non-fixed random seed! :facepalm:

jeromekelleher: Dammit, I hate it when that happens! :smile:

jeromekelleher:
Is this because of stable and unstable sorting algs in the libc implementations @benjeffery?

python/tests/test_highlevel.py
benjeffery: Here's the badly sorted data that's come out of `compute_mutation_times`. Just a single site. I'm trying to get my head round what's going on. https://gist.github.com/benjeffery/fe7727ae93bbe79fa5b3509428097465

benjeffery: I don't think this is a bug in `compute_mutation_times` The order going in is just "sites and parents before children" but once times are applied (using the node times) the old order is no longer valid.

c/tskit/core.c
benjeffery: https://github.com/tskit-dev/tskit/pull/726/commits/4427cd24a7551634867d4d2656f3938389a3fa0c

docs/data-model.rst
benjeffery: https://github.com/tskit-dev/tskit/pull/726/commits/4427cd24a7551634867d4d2656f3938389a3fa0c

c/tskit/tables.c
```diff
@@ -4343,9 +4343,12 @@ cmp_mutation(const void *a, const void *b)
     const tsk_mutation_t *ib = (const tsk_mutation_t *) b;
     /* Compare mutations by site */
     int ret = (ia->site > ib->site) - (ia->site < ib->site);
+    /* Within a particular site sort by time if known, then ID. This ensures that
+     * relative ordering within a site is maintained */
+    if (ret == 0 && !tsk_is_unknown_time(ia->time) && !tsk_is_unknown_time(ib->time)) {
```
jeromekelleher, line 4348: I can't see how this wouldn't completely determine the sort order - if either time is unknown we should fall back on comparing by ID.

benjeffery: The problem here is that our `cmp` gives inconsistent orderings (5<4, 4<U, U<5) in my example above as one is by time  and the other two by id. The cycle has to be broken somehow.

benjeffery: @petrelharp I think you might be helpful here as I'm not sure how to fix this one! This comment hopefully explains the situation: https://github.com/tskit-dev/tskit/pull/726#issuecomment-662429215

jeromekelleher:
Ugh, this is nasty. The only simple way out of this I can see is to disallow mixed known and unknow times at a site. That way we should never have to compare unknown times with anything. It's the edgiest of edge-cases, in any case (who is actually going to do this?).

@petrelharp, any thoughts here?

It's great this has come up now - testing works!

jeromekelleher:
@petrelharp, we could use your input here. Do you object to disallowing mixed known and unknown times at a site? Otherwise, we get into undefined behaviour because or sort keys are ill defined, as Ben has shown here.

python/tests/test_highlevel.py
jeromekelleher: So I think it is a bug in compute_mutation_times if it doesn't return a well-formed mutation table. If we have to, we could do a site-wise sort, once all the times have been generated? (Or if this is too awkward, we could do a full TableCollection.sort(), but this is a very heavy weapon.)

benjeffery: In https://github.com/tskit-dev/tskit/pull/726/commits/30a3134d8c6ff5258fec539d5f260dd99361f503 I've used `tsk_table_collection_sort` but only if it is needed. If this turns out to be a bottleneck we can put more effort in, but I'm not sure it is worth the complexity without a pressing use case.

petrelharp:
I agree: disallow mixed missing/nonmissing times. I don't see a good way around it and don't know if a reason we'd need to have mixed times. Ps Sorry for the delay, I am on semi vacation.

benjeffery:
No worries, thanks for having a look. We can allow mixed known/unknown as long as `sort` isn't called. Or is it might be cleaner to have it be a table-level requirement.

jeromekelleher:
> We can allow mixed known/unknown as long as sort isn't called. Or is it might be cleaner to have it be a table-level requirement.

Let's enforce it as a table-level requirement. It's just hassle for us otherwise.

python/tests/test_highlevel.py
benjeffery: Need to change this to use a bookmark pointing to the end of the edge table to skip edge sorting.

benjeffery: Skip edge sorting in https://github.com/tskit-dev/tskit/pull/726/commits/36ce487bba0a3b9aa45be0f013840ab690f1a21b

benjeffery:
Seems to be some issues with coverage reporting as C lines I know are covered from python are not showing as such.

mufernando:
nice work on this @benjeffery, thanks!

PR #730, opened 2020-07-23
Allow top-level metadata type to be union of null, object #659

### LINKED ISSUE #659: Allow top-level metadata type to be a union of null, object.
As discussed in https://github.com/tskit-dev/pyslim/pull/82 empty metadata is a common occurrence when a tool is adding to an existing tree sequence. To add metadata a tool adds a schema which means default "na" values must be added to all existing rows. 
We can change the metaschema to allow a union of null,object at the top-level of the schema to support a subset of rows having empty metadata. The encoders and decoders will special-case this usage.

benjeffery:
Fixes #659

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -61,8 +61,9 @@ META_SCHEMA = replace_root_refs(META_SCHEMA)
 META_SCHEMA["definitions"]["root"] = copy.deepcopy(META_SCHEMA)
 META_SCHEMA["codec"] = {"type": "string"}
 META_SCHEMA["required"] = ["codec"]
-# For interoperability reasons, force the top-level to be an object
-META_SCHEMA["properties"]["type"] = {"enum": ["object"]}
+# For interoperability reasons, force the top-level to be an object or union
+# of object and null
+META_SCHEMA["properties"]["type"] = {"enum": ["object", ["object", "null"]]}
 TSKITMetadataSchemaValidator.META_SCHEMA = META_SCHEMA
 
 
@@ -140,7 +141,10 @@ def binary_format_validator(validator, types, instance, schema):
     # Non-composite types must have a binaryFormat
     if (
         validator.is_type(instance, "object")
-        and (instance.get("type") not in (None, "object", "array", "null"))
+        and (
+            instance.get("type")
+            not in (None, "object", "array", "null", ["object", "null"])
+        )
         and "binaryFormat" not in instance
     ):
         yield jsonschema.ValidationError(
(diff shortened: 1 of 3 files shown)


python/tskit/metadata.py
```diff
@@ -61,8 +61,9 @@ def replace_root_refs(obj):
 META_SCHEMA["definitions"]["root"] = copy.deepcopy(META_SCHEMA)
 META_SCHEMA["codec"] = {"type": "string"}
 META_SCHEMA["required"] = ["codec"]
-# For interoperability reasons, force the top-level to be an object
-META_SCHEMA["properties"]["type"] = {"enum": ["object"]}
+# For interoperability reasons, force the top-level to be an object or union
+# of object and null
```
jeromekelleher, line 65: It's not obvious why this isn't ``{"enum": ["object", "null"]}``, why not just a union of object and null?

benjeffery: `{"enum": ["object", "null"]}` would allow `object` and `null` but wouldn't allow the union `["object", "null"]`

jeromekelleher: I thought that's what we wanted? Metadata you decode is either an object or null?

benjeffery: Yes, so we need to specify `["object", "null"]` as the `type` of the top-level. `{"enum": ["object", "null"]}` would not allow this as only `"object"` or `"null"` would be valid specifications.

jeromekelleher: I'm being thick here, I know, I still don't get it. Probably easiest to talk me through it in person!

grahamgower: This is the meta-schema, which defines what's allowed in the schema. The schema then defines what's allowed in the metadata. If you want either "object" or "null" allowed in the metadata, the schema says union of "object" and "null", and then the meta-schema must allow the schema to say unions are allowed for "object" and "null" (not just "object" or "null").

grahamgower: (I recently read up on Guido's PEG parser stuff coming in python 3.9 (https://medium.com/@gvanrossum_83706/peg-parsing-series-de5d41b2ed60), so this thread caught my eye...)

jeromekelleher: Riiiighhhtt. OK, I think I get it now. Thanks @grahamgower. In any case, if both of you are happy this is right then I'm happy too!

benjeffery: Thanks for clearing that up @grahamgower! I admit the meta schema has had me head scratching quite a bit...

PR #736, opened 2020-07-29
Set version for 0.3.0.beta1 release

benjeffery:


jeromekelleher:
For [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-releases) version numbering this should be ``0.3.0b1`` rather than ``0.3.0.b1``. Setuptools will normalise to 0.3.0b1, so that there's a slight difference between the package version and the ``__version__`` value:
```
>>> import tskit
>>> tskit.__version__
'0.3.0.b1'
>>> 
```
and the package is ``tskit-0.3.0b1.tar.gz`` (this is also the version string from pypi).

I don't think it'll make any difference for the beta though.

benjeffery:
I realised this at the time, but went with the convention specified in the comment in `_version.py` as setuptools was still happy.

jeromekelleher:
We should update the comment there to avoid confusion I think - ``0.3.0.devN`` is correct for [development releases](https://www.python.org/dev/peps/pep-0440/#developmental-releases) but pre-releases have a different convention. :shrug:

Issue #750, opened 2020-08-06
tables.tree_sequence() does not check if nodes from an individual are contiguous


hyanwong:
In the [docs](https://tskit.readthedocs.io/en/latest/data-model.html#node-requirements) it says "For simplicity and algorithmic efficiency, all nodes referring to the same (non-null) individual must be contiguous.". But this isn't checked when converting tables to a tree sequence (see below). I assume this could lead to bugs.

```
ts = msprime.simulate(4, random_seed=1)
tables = ts.tables
tables.individuals.add_row()
tables.individuals.add_row()
ind = tables.nodes.individual
ind[ts.samples()] = [0, 1, 0, 1]  # make nodes from same individual non-contiguous
tables.nodes.individual = ind
new_ts = tables.tree_sequence()
print(new_ts.individual(0)).nodes

array([0, 2], dtype=int32)
```

petrelharp:
I vaguely remember deciding that we should enforce this, and the only individual-producing code I can think of - which is SLiM - will ensure this. So, that seems good. And, I still think it's a good idea. It could cause someone's script that makes up individuals produce a bad tree sequence, so this should definately be added to  the CHANGELOG under "possibly breaking changes".

And, btw, this requirement means that adding individuals to - say - an msprime simulation could require re-ordering the nodes, so it's a good thing we have `subset` to do that now.

petrelharp:
p.s. I can't think of anywhere this *would* currently lead to bugs, so that's good anyhow.

jeromekelleher:
This does seem like a bug if we're not enforcing the property and we're using it internally I'm not sure that we are though, it seems like a pretty big oversight if we forgot to test it.

petrelharp:
I'll have a look through the code - we don't use node.individual in very many places, so it won't be too bad to check.

petrelharp:
Nope - we really don't seem to be relying on this property anywhere in the C code. The code that finds which nodes go to which individual (`tsk_treeseq_init_individuals`) would be a lot simpler, in fact, if we did rely on nodes from the same individual being adjacent.

The downside to requiring this is that adding individuals to an msprime simulation (for instance) would then require reordering of nodes instead of just filling in the individual column.

hyanwong:
So should I delete the comment in the docs then?

jeromekelleher:
> So should I delete the comment in the docs then?

Yes please - we've clearly gone to the trouble of implementing and testing this without assuming adjacency, and the docs must have been forgotten about.

PR #751, opened 2020-08-06
Check that individuals have the same pop, time, & contiguous nodes

### LINKED ISSUE #749: Valid TS should not allow nodes from multiple populations within the same individual
At the moment, it appears as if it is valid to create nodes within the same individual that belong to different populations (code below). I think we should ban this, and raise an error when tables with this property are turned into a tree sequence. Does this seem sensible? I don't think that having a single individual containing two sample nodes each belong to a different population makes much sense.

```
ts = msprime.simulate(4, random_seed=1)
tables = ts.tables
tables.individuals.add_row()
tables.individuals.add_row()
ind = tables.nodes.individual
pop = tables.nodes.population
ind[ts.samples()] = np.arange(ts.num_samples)//2
tables.nodes.individual = ind
pop = np.full(ts.num_nodes, tskit.NULL, dtype=tables.nodes.population.dtype)
pop[0] = 0  # only sample 0 has a population, but 
tables.nodes.population = pop
new_ts = tables.tree_sequence()
print(str(new_ts.tables.nodes))

id	flags	population	individual	time	metadata
0	1	0	0	0.00000000000000	
1	1	-1	0	0.00000000000000	
2	1	-1	1	0.00000000000000	
3	1	-1	1	0.00000000000000	
4	0	-1	-1	0.17986860883729	
5	0	-1	-1	1.97752050080185	
6	0	-1	-1	2.69754599651364	

```

hyanwong:
Fixes #749 and #750

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -336,6 +336,15 @@ tsk_strerror_internal(int err)
         case TSK_ERR_METADATA_DISABLED:
             ret = "Metadata is disabled for this table, so cannot be set";
             break;
+        case TSK_ERR_INDIVIDUAL_NODES_NOT_CONTIGUOUS:
+            ret = "Nodes belonging to the same individual must be contiguous";
+            break;
+        case TSK_ERR_MULTIPLE_POPS_FOR_INDIVIDUAL:
+            ret = "Nodes belonging to one individual must be in the same population";
+            break;
+        case TSK_ERR_MULTIPLE_TIMES_FOR_INDIVIDUAL:
+            ret = "Nodes belonging to one individual must occur at the some time";
+            break;
 
         /* Limitations */
         case TSK_ERR_ONLY_INFINITE_SITES:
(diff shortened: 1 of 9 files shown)


hyanwong:
NB: I had to remove 2 tests which assumed it was reasonable to allow interleaved individuals.

jeromekelleher:
If it's OK to have interleaved individuals now, then I don't see any reason to make this requirement stricter. We should just remove the out-of-date note from the docs.

petrelharp:
> We should just remove the out-of-date note from the docs.

I'm OK with that also, although I feel like we'll be making our lives in the future easier if we do enforce this requirement. But possibly breaking people's code is bad, soo...

hyanwong:
I'm ambivalent here, but I think it would be good to sort out this inconsistency (and #750) before v0.3.0 if possible.

hyanwong:
Closing to re-work

PR #756, opened 2020-08-08
Make TableCollection methods more findable

petrelharp:
I was having trouble finding the TableCollection methods, so I added a summary of these methods, and rearranged the structure of the docs about Tables to be - I think? - more intuitive. Previously, the docs talked through each of the tables and eventually got to table collections; now it goes "you already know about tree sequences: most of those operations are actually on table collections; and ok, here's how the underlying tables work".

Suggestions welcome.

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -133,7 +133,7 @@ Tables and Table Collections
 The information required to construct a tree sequence is stored in a collection
 of *tables*, each defining a different aspect of the structure of a tree
 sequence. These tables are described individually in :ref:`the next section
-<sec_tables_docs>`. However, these are interrelated, and so many operations work
+<sec_tables_api_tables>`. However, these are interrelated, and so many operations work
 on the entire collection of tables, known as a ``TableCollection``.
 The :class:`TableCollection` and :class:`TreeSequence` classes are
 deeply related. A ``TreeSequence`` instance is based on the information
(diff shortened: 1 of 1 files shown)


molpopgen:
I like the auto summary.  One thing that has been bugging me is the description of mutability.  If you have a tree sequence, calling the functions that change it do not raise an error (like the various `add_row` methods).  Rather, they just silently do nothing.

petrelharp:
> Rather, they just silently do nothing.

Oh, that seems fixable, with some python magic.

docs/python-api.rst
```diff
@@ -126,9 +126,41 @@ using the :ref:`Tables API <sec_tables_api>`.
...
+Many of the ``TreeSequence`` methods that return a modified tree sequence
+are in fact wrappers around a corresponding ``TableCollection`` method
+that modifies a copy of the origin tree sequence's table collection.
+
+.. autoclass:: tskit.TableCollection(sequence_length=0)
+    :autosummary:
+    :members:
+
+
+.. _sec_tables_docs:
```
jeromekelleher, line 159: Can we change the lable to ``sec_tables_api_tables`` for consistency please? I've tried to make it reasonably consistent elsewhere (but not entirely successfully). The current label doesn't really fit with any conventions though.

jeromekelleher:
> One thing that has been bugging me is the description of mutability. If you have a tree sequence, calling the functions that change it do not raise an error (like the various add_row methods). Rather, they just silently do nothing.

I've opened a new issue for this @molpopgen, #760. Short version is, it's not that straightforward and we'll need to it properly at some point anyway.

Issue #721, opened 2020-07-15
Compile warnings on clang >= 6


jeromekelleher:
Because of an upstream bug in clang >= 6, we get warnings with floating point operations like ``isnan`` and ``isfinite``, 
```
FAILED: tskit@sha/tskit_trees.c.o                               
clang -Itskit@sha -I. -I.. -Isubprojects/kastore -I../subprojects/kastore -Xclang -fcolor-diagnostics -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -g -fPIC -Wall -Wextra -Werror
 -Wpedantic -W -Wmissing-prototypes -Wstrict-prototypes -Wconversion -Wshadow -Wpointer-arith -Wcast-align -Wcast-qual -Wwrite-strings -Wnested-externs -fshort-enums -fno-common -MD -MQ 'tsk
it@sha/tskit_trees.c.o' -MF 'tskit@sha/tskit_trees.c.o.d' -o 'tskit@sha/tskit_trees.c.o' -c ../tskit/trees.c
../tskit/trees.c:284:28: error: implicit conversion loses floating-point precision: 'const double' to 'float' [-Werror,-Wconversion]
                if (!isnan(mutation_time[mutation])              
                     ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~                                                                             
/usr/include/math.h:644:46: note: expanded from macro 'isnan'
#  define isnan(x) __MATH_TG ((x), __isnan, (x))                                                                                
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~                
/usr/include/math.h:559:16: note: expanded from macro '__MATH_TG'
   ? FUNC ## f ARGS                             \               
     ~~~~~~~~~ ^~~~                                              
1 error generated.                                  
[8/52] Compiling C object 'tskit@sha/tskit_tables.c.o'.          
FAILED: tskit@sha/tskit_tables.c.o                                                                                              
clang -Itskit@sha -I. -I.. -Isubprojects/kastore -I../subprojects/kastore -Xclang -fcolor-diagnostics -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -g -fPIC -Wall -Wextra -Werror
[...]

PR #758, opened 2020-08-10
Move fix for clang bug to C file.

jeromekelleher:
This moves the workaround for the clang bug (#721) from the header file to a C file, to address very good points made by @bhaller. The fix is now local to C code, so we don't need to check for C++ any more, and library users will no longer have this workaround foisted on them unwittingly.

@molpopgen, do you see any problems with doing this?

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -32,6 +32,7 @@
 #include <float.h>
 #include <math.h>
 
+#define _TSK_WORKAROUND_FALSE_CLANG_WARNING
 #include <tskit/tables.h>
 
 #define DEFAULT_SIZE_INCREMENT 1024
--- a/c/tests/testlib.h
+++ b/c/tests/testlib.h
@@ -31,6 +31,7 @@
 #include <stdlib.h>
 
 #include <CUnit/Basic.h>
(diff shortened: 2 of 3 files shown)


c/tskit/tables.c
```diff
@@ -34,6 +34,27 @@
 
 #include <tskit/tables.h>
 
+#if defined(__clang__)
+/* Work around bug in clang >= 6.0, https://github.com/tskit-dev/tskit/issues/721
+ * (note: fixed in clang January 2019)
+ * This workaround does some nasty fiddling with builtins, but since it's defined
+ * here it shouldn't apply anywhere outside this C file and it should affect
```
bhaller, line 41: *shouldn't

bhaller:
@petrelharp I guess this should be pulled into SLiM once this PR is approved, so I'm tagging you to put it on your radar.  :->

jeromekelleher:
Ended up having to take a different approach here because we used isnan etc in the test suite as well. Same basic outcome though.

Thanks for taking a look @bhaller.

The 0.99.4 version should be tagged in the next day or so, so hopefully we can sync up on that for SLiM.

molpopgen:
This should be fine.

jeromekelleher:
@benjeffery, this look good to you?

PR #759, opened 2020-08-10
Fixes for newer doxygen versions.

jeromekelleher:
Might as well get this in before 0.99.4 @benjeffery, since the doxygen was technically malformed (but older versions didn't seem to mind).


benjeffery:
Great, LGTM.

PR #761, opened 2020-08-10
Allow mixture of known and unknown as long as they are on different sites

### LINKED ISSUE #748: Mixed missing and non-missing mutation times clarification
#726 added a requirement that we cannot mix known and unknown mutation times. However, I think this is a bit too strict, because we only need to insist that known and unknown mutation times can't be mixed *within a site* to resolve the ordering problems.

Should we relax to the checking we don't mix known and unknown times within a site? We could bundle this into the 0.99.4 C API release we need to do anyway (#747).

cc @petrelharp, @mufernando

benjeffery:
FIxes #748 by relaxing the constraint on mutation times to allow mixtures of known/unknown as long as they are on different sites.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -6701,13 +6701,6 @@ tsk_table_collection_check_mutation_integrity(
                 goto out;
             }
         }
-        /* Check known/unknown times are not both present */
-        if (unknown_time) {
-            unknown_times_seen = true;
-        } else if (unknown_times_seen) {
-            ret = TSK_ERR_MUTATION_TIME_HAS_BOTH_KNOWN_AND_UNKNOWN;
-            goto out;
-        }
 
         if (check_mutation_ordering) {
             /* Check site ordering and reset time check if needed*/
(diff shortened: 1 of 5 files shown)


c/tskit/tables.c
```diff
@@ -6701,7 +6701,11 @@ tsk_table_collection_check_mutation_integrity(
                 goto out;
             }
         }
-        /* Check known/unknown times are not both present */
+
+        /* Check known/unknown times are not both present on a site*/
+        if (j > 0 && mutations.site[j - 1] != mutations.site[j]) {
+            unknown_times_seen = false;
```
jeromekelleher, line 6707: Should probably do this below where we set ``last_known_time = INFINITY;``. We're inherently depending on the site ordering being correct anyway, so we might as will avoid the extra comparison.

benjeffery: I had it there initially - but then the check can only happen if `check_mutation_ordering` is set.

jeromekelleher: We'll only catch mixed stuff at a site if all the mutations for a given site are adjacent, so we depend on this ordering being true anyway.

benjeffery: Cool, I've moved it and changed the test.

PR #768, opened 2020-08-13
Only build wheels once on release

benjeffery:
Currently they are done twice.


benjeffery:
Manually merging as this is an action config change.

PR #770, opened 2020-08-15
err whoopsie

petrelharp:
@bhaller noticed this minor whoopsie: the output is from `fprintf`, which returns the number of characters written. The other `X_dump_text` methods have `< 0` correctly.


benjeffery:
Gah! my bad I think! Still getting back in the C groove.

Issue #727, opened 2020-07-20
Generating ms-output from tree sequence


SBNoor:
I have tree sequence recording from SLiM. Is there a way in msprime to generate the MS output directly from msprime?

jeromekelleher:
Putting it into 0.3.0 for now, but we can push back to 0.3.1 if needs be.

yunusbb:
Ability to output in ms format is very much anticipated function. This can make tons of old scripts and pipelines usable. Would be great if someone can add this ms-format output?

saurabhbelsare:
Hi, 
I'm working on a function to write out ms format output from tree sequence, creating a function similar to the existing write_vcf function. Hope this approach is fine.

jeromekelleher:
That would be excellent, thanks @saurabhbelsare!

yunusbb:
Wow, great! Thanks a lot, Saurabh!

petrelharp:
Say, @yunusbb and @SBNoor - @saurabhbelsare is working on this over in #854, and we'd like to make sure this would do what you need for your pipelines. Care to elaborate what sort of ms format you use? e.g. do you get the newick trees (with `-T`)? is `num_replicates > 1`?

PR #772, opened 2020-08-18
Add macs conversion from msp

winni2k:
Move the macs and newick CLI commands over from msprime. See https://github.com/tskit-dev/msprime/pull/1124. This closes https://github.com/tskit-dev/msprime/issues/1120.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -4302,6 +4302,25 @@ class TreeSequence:
         s += "END;\n"
         return s
 
+    def to_macs(self):
+        """
+        Return a `macs encoding <https://github.com/gchen98/macs>`_
+        of this tree sequence.
+
+        :return: The macs representation of this TreeSequence as a string.
+        :rtype: str
+        """
+        n = self.get_sample_size()
+        m = self.get_sequence_length()
+        output = [f"COMMAND:\tnot_macs {n} {m}"]
+        output.append("SEED:\tASEED")
+        for variant in self.variants(as_bytes=True):
+            output.append(
+                f"SITE:\t{variant.index}\t{variant.position / m}\t0.0\t"
+                f"{variant.genotypes.decode()}"
+            )
+        return "\n".join(output) + "\n"
+
     def simplify(
         self,
         samples=None,
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -118,6 +118,9 @@ SVG drawing improvements and many others.
 - Add ``tree.as_dict_of_dicts()`` function to enable use with networkx. See
   :ref:`sec_tutorial_networkx` (:user:`winni2k`, :pr:`457`).
 
+- Add ``tree_sequence.to_macs()`` function to convert tree sequence to MACS
+  format (:user:`winni2k`, :pr:`727`)
+
 **Bugfixes**
 
 - :issue:`453` - Fix LibraryError when ``tree.newick()`` is called with large node time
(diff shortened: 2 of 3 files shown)


jeromekelleher:
Thanks @winni2k. I'm not sure there's much point in offering these as parts of the ``tskit`` cli though (it's surely a pretty niche requirement). I wonder if it would be better to add a method ``as_macs`` (or ``to_macs``, maybe?) to the TreeSequence class. This would go well with the proposed ``as_ms`` method (#1120). It also makes me wonder if we should call the current ``nexus`` method ``as_nexus`` - this would give a nice convention for converting to other formats. What do you think @winni2k?

Certainly we don't need the CLI ``newick`` command anyway, I think, so we can just drop that.

winni2k:
Ah, dang, I missed the link to #727. 

If we have a nexus converter (where?) then the newick conversion is certainly obsolete. I also see we have newick conversion in ts_convert (https://github.com/tskit-dev/tsconvert/blob/master/tsconvert/newick.py#L136). There's also a `to_ms` method in ts_convert? https://github.com/tskit-dev/tsconvert/blob/master/tsconvert/newick.py#L105 

We are using `to_newick` in ts_convert, so maybe go with `to_*`? 

Removing the newick conversion from tskit now...

jeromekelleher:
Great, thanks @winni2k. I didn't explain myself very well the last time, but I was suggesting that we get rid of the ``to_macs`` CLI command and instead add a ``to_macs`` (or ``as_macs`` - I'm still on the fence on this one!) *method* to the TreeSequence class. I don't think we need the CLI end point, but a method could be useful to people. It's basically the same code, it's just the tests will live in a different module (I guess the ``test_highlevel.py`` file).

I wouldn't worry too much about what things are called in tsconvert - that's never really gotten off the ground and was never brought up to a release.

winni2k:
Ah, conversion commands on the CLI would clutter the tskit interface? 

I suppose the correct thing to do then would be to move the CLI command to ts_convert, but have the method on the TreeSequence class. I'll add the method. 

Naming: How about following the example of pandas dataframe conversion on this? https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html

winni2k:
Do you have a link for the macs format so that I can write something sensible in the method documentation?

winni2k:
Also, what about `write_macs`? We have the `write_vcf` method...

jeromekelleher:
> Naming: How about following the example of pandas dataframe conversion on this?

Yes! Perfect - I like having someone else to follow. Let's call them ``to_X``. We can make things more consistent later, let's just get the ``to_macs`` method in for now.

I think a link to the [source repo](https://github.com/gchen98/macs) for macs is good enough in terms of docs.

python/tskit/trees.py
```diff
@@ -4295,6 +4295,34 @@ def nexus(self, precision=14):
         s += "END;\n"
         return s
 
+    def to_macs(self, path_or_buf=None):
```
jeromekelleher, line 4298: We're not testing the ``path_or_buf`` argument here, so I'd say let's not bother with it. Just return a string - we can add this type of argument later, if we need it (and should try to be consistent across the formats).

jeromekelleher:
Great, thanks @winni2k - can you squash the commits please? Good to merge then.

jeromekelleher:
Excellent, thanks @winni2k! Sorry, I forgot one thing - can you add this to the python CHANGELOG in the "New features" section please (I should add a tickbox for this in a PR template, I keep forgetting about it!)

winni2k:
Of course! How about this?

jeromekelleher (approved):

jeromekelleher:
Perfect! Thanks @winni2k !

PR #778, opened 2020-08-19
Fix bugs in missing data handling.

### LINKED ISSUE #776: Segfault in ``samples`` argument to TreeSequence.variants with missing data
Reproduce:
```python
   def test_empty_ts_incomplete_samples(self):
        tables = tskit.TableCollection(1.0)
        tables.nodes.add_row(tskit.NODE_IS_SAMPLE, 0)
        tables.nodes.add_row(tskit.NODE_IS_SAMPLE, 0)
        tables.sites.add_row(0.5, "A")
        ts = tables.tree_sequence()
        variants = list(ts.variants(samples=[0])) # segfault
```
This is because the sample_index_map isn't checked when we mark missing data.

This will only get triggered when there's missing data and someone asks for genotypes for a subset of the samples.

### LINKED ISSUE #777: Mutations over isolated samples should not be missing data
Currently if a mutation exists over an isolated sample, it's incorrectly decoded as having missing data when it should have the corresponding allele.

See #716 for more discussion.

jeromekelleher:
This fixes some bugs in missing data handling. I've chosen to regard #777 where mutations over samples were decoded as missing data as a straight-up bug rather than something we need to see as a breaking change. Also spotted another howler in there while I was coding this up.

This needs more testing, but I thought I'd get some feedback on whether this was the right thing to do before doing that.

Closes #776
Closes #777

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -135,6 +135,12 @@ SVG drawing improvements and many others.
 - :issue:`453` - Fix LibraryError when ``tree.newick()`` is called with large node time
   values (:user:`jeromekelleher`, :pr:`637`).
 
+- :issue:`777` - Mutations over isolated samples were incorrectly decoded as
+  missing data. (:user:`jeromekelleher`, :pr:`778`)
+
+- :issue:`776` - Fix a segfault when a partial list of samples
+  was provided to the ``variants`` iterator. (:user:`jeromekelleher`, :pr:`778`)
+
 **Deprecated**
 
 - The ``sample_counts`` feature has been deprecated and is now
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -12,6 +12,14 @@
   from the MRCAs of samples in the simplified tree sequence back to the roots
   in the input tree sequence (:user:`jeromekelleher`, :issue:`775`, :pr:`782`).
 
+**Bugfixes**
+
+- :issue:`777` - Mutations over isolated samples were incorrectly decoded as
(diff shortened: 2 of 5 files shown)


python/CHANGELOG.rst
```diff
@@ -123,6 +123,12 @@ SVG drawing improvements and many others.
 - :issue:`453` - Fix LibraryError when ``tree.newick()`` is called with large node time
   values (:user:`jeromekelleher`, :pr:`637`).
 
+- :issue:`777` - Mutations over isolated samples were incorrectly decoded as
+  missing data. (:user:`jeromekelleher`, :pr:`778`)
+
+- :issue:`776` - Fix a segfault when partial a partial list of samples
```
hyanwong, line 129: "partial" is duplicated as a typo here

hyanwong:
Looks good to me. Lots more tests, not so much extra non-test code.

I see that in the Python `haplotypes()` method we say 

>If instead it is set to True, missing data will be imputed such that all
>        isolated samples are assigned the ancestral state (unless they have
        mutations directly above them, in which case they will take the most recent
        derived  mutational state for that node)

So we are already doing something similar there (and this implies that the previous genotypes() behaviour should be thought of as buggy). I wonder if this means we can simplify the python code?

benjeffery (approved):
LGTM - glad to get this in for 0.3.0

jeromekelleher:
(Note to self, check what happens when we have multiple mutations over an isolated sample)

jeromekelleher:
And close #796

PR #780, opened 2020-08-19
Document the edge order in edge_diffs

hyanwong:
It's useful to know if e.g. we can group by parent id in the edges_in and edges_out lists (which are, indeed, lists, not tuples, as documented)


python/tskit/trees.py
```diff
@@ -3446,14 +3446,20 @@ def edge_diffs(self):
...
         covering the interval (hence ``edges_out`` will always be empty for the
-        first tree). The ``edges_in`` value is a tuple of edges that were just
+        first tree). The ``edges_in`` value is a list of edges that were just
         inserted to construct the tree convering the current interval. The
-        iterator terminates after the final interval in the tree sequence (i.e.
-        it does not report a final removal of all remaining edges); the number
-        of iterations is thus always equal to the number of trees in the tree
-        sequence.
+        edges in each ``edges_in`` list are returned in the
+        :ref:`order required for a valid tree sequence <sec_edge_requirements>`
```
jeromekelleher, line 3454: I don't think this is true - we're not returning edges in the table order, it's the order given by the indexes. I think it's also not in order of ID, it's in order of parent node time.

jeromekelleher (changes requested):
Good to make the docs more precise, but this isn't actually true I'm afraid.

python/tskit/trees.py
hyanwong: Ah - I wasn't exactly sure. But are the edges for a given parent node always going to be returned consecutively?

hyanwong: (NB: @jeromekelleher - this is important for detecting polytomies duding edge diff iteration)

jeromekelleher: Here's where it's done: https://github.com/tskit-dev/tskit/blob/master/c/tskit/tables.c#L7203

So it looks like the sort is ``(left, time[parent], parent, child)``. This does guarantee that all edges for a given parent are adjacent, and children are ordered by node ID.

hyanwong: Thanks. I'll adjust my wording. I think the inverse order is used for edges_out?

jeromekelleher: Yes, it's the same order but reversed.

PR #781, opened 2020-08-20
Ancestral nodes from a founding generation can now be retained in simplify

### LINKED ISSUE #775: Add ``keep_input_roots`` option to simplify
Following from the discussion in #716, it seems like a lot of the confusion surrounding missing data is due to simplify not having quite the right semantics for the forward simulation case. For forward simulations, we really do want to know about the branches back to the original generation, but these are removed by simplify as it gets rid of any topology above the roots. So, I think we want to have an option to simplify called ``keep_input_roots`` which preserves the roots of the trees in the input tree sequence. The easy way to do this is something like
```python
        all_roots = set()
        for tree in ts.trees():
            for root in tree.roots:
                all_roots.add(root)        
        ts.simplify(samples + list(all_roots))
```

This doesn't do *quite* what we'd want though, I think. Consider this example
```
1.76┊    14       ┊    14       ┊     14      ┊    14       ┊             ┊  
    ┊  ┏━━┻━┓     ┊  ┏━━┻━━━┓   ┊   ┏━━┻━━┓   ┊  ┏━━┻━━┓    ┊             ┊  
1.51┊  ┃    ┃     ┊  ┃     13   ┊   ┃    13   ┊  ┃    13    ┊      13     ┊  
    ┊  ┃    ┃     ┊  ┃    ┏━┻━┓ ┊   ┃    ┏┻━┓ ┊  ┃   ┏━┻━┓  ┊    ┏━━┻━━┓  ┊  
1.27┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊   ┃    ┃  ┃ ┊  ┃  12   ┃  ┊    ┃    12  ┊  
    ┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊   ┃    ┃  ┃ ┊  ┃  ┏┻┓  ┃  ┊    ┃    ┏┻┓ ┊  
1.21┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊   ┃    ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊   11    ┃ ┃ ┊  
    ┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊   ┃    ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊  ┏━┻━┓  ┃ ┃ ┊  
0.64┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊  10    ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊  ┃   ┃  ┃ ┃ ┊  
    ┊  ┃    ┃     ┊  ┃    ┃   ┃ ┊  ┏┻━┓  ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊  ┃   ┃  ┃ ┃ ┊  
0.49┊  ┃    9     ┊  ┃    9   ┃ ┊  ┃  ┃  ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊  ┃   ┃  ┃ ┃ ┊  
    ┊  ┃  ┏━┻━┓   ┊  ┃  ┏━┻┓  ┃ ┊  ┃  ┃  ┃  ┃ ┊  ┃  ┃ ┃  ┃  ┊  ┃   ┃  ┃ ┃ ┊  
0.31┊  ┃  ┃   8   ┊  ┃  ┃  8  ┃ ┊  ┃  ┃  8  ┃ ┊  ┃  ┃ ┃  8  ┊  ┃   8  ┃ ┃ ┊  
    ┊  ┃  ┃ ┏━┻┓  ┊  ┃  ┃ ┏┻┓ ┃ ┊  ┃  ┃ ┏┻┓ ┃ ┊  ┃  ┃ ┃ ┏┻┓ ┊  ┃  ┏┻┓ ┃ ┃ ┊  
0.26┊  ┃  ┃ ┃  7  ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  
    ┊  ┃  ┃ ┃ ┏┻┓ ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  ┃  ┃ ┃ ┃ ┃ ┊  
0.03┊  6  ┃ ┃ ┃ ┃ ┊  6  ┃ ┃ ┃ ┃ ┊  6  ┃ ┃ ┃ ┃ ┊  6  ┃ ┃ ┃ ┃ ┊  6  ┃ ┃ ┃ ┃ ┊  
    ┊ ┏┻┓ ┃ ┃ ┃ ┃ ┊ ┏┻┓ ┃ ┃ ┃ ┃ ┊ ┏┻┓ ┃ ┃ ┃ ┃ ┊ ┏┻┓ ┃ ┃ ┃ ┃ ┊ ┏┻┓ ┃ ┃ ┃ ┃ ┊  
0.00┊ 0 5 1 2 3 4 ┊ 0 5 1 2 4 3 ┊ 0 5 1 2 4 3 ┊ 0 5 1 3 2 4 ┊ 0 5 2 4 1 3 ┊  
  0.00          0.18          0.37          0.75          0.91          1.00 
```
when we simplify down for samples = [0, 5] (plus the roots) we get
```
1.76┊  3    ┊   3   ┊     3 ┊  
    ┊  ┃    ┊  ┏┻━┓ ┊       ┊  
1.51┊  ┃  2 ┊  ┃  2 ┊  2    ┊  
    ┊  ┃    ┊  ┃    ┊  ┃    ┊  
0.03┊  4    ┊  4    ┊  4    ┊  
    ┊ ┏┻┓   ┊ ┏┻┓   ┊ ┏┻┓   ┊  
0.00┊ 0 1   ┊ 0 1   ┊ 0 1   ┊  
  0.00    0.18    0.91    1.00 
```

(If this looks weird in your browser, try a different one. Curse the modern web and lack of reliable monospace fonts!)

The weirdness in how the trees are drawn here is because the old roots are samples, but if we marked them as non-samples the odd topologies would still be there. What do we do with nodes that are roots in one part of the tree sequence but are internal nodes elsewhere? We probably don't want to leave them out of the edges, as this would be misleading, so this means we kind-of have to mark the input roots as special beforehand. If we just want to mark the nodes at the top of the trees, then in principle this information is available in simplify at the end of the algorithm, but it's not obvious to me how to make use of it.

Not sure what to do here. Perhaps it would be best if the forward simulators made a pass through the trees first to get the roots, like the example above? In practise, this would just be a list of the first generation individuals who have any descendants. The cost would be to (a) build the index required for iterating over the trees, which is significant; and (b) a single pass through the trees and roots, which is probably not that big a cost, unless it's a large simulation and done very early (and so there's lots of roots).

Any thoughts @molpopgen @petrelharp @bhaller?

gtsambos:
Closes #775 (I think?)

### DIFF
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -108,6 +108,7 @@ class Simplifier:
         filter_populations=True,
         filter_individuals=True,
         keep_unary=False,
+        founders=None,
     ):
         self.ts = ts
         self.n = len(sample)
@@ -140,6 +141,11 @@ class Simplifier:
         self.position_lookup = None
         if self.reduce_to_site_topology:
             self.position_lookup = np.hstack([[0], position, [self.sequence_length]])
+        self.founders = founders
+        if self.founders is not None:
+            self.keep_founders = True
+        else:
+            self.keep_founders = False
 
     def record_node(self, input_id, is_sample=False):
         """
@@ -263,6 +269,8 @@ class Simplifier:
         """
         output_id = self.node_id_map[input_id]
         is_sample = output_id != -1
+        if self.keep_founders:
+            is_founder = input_id in self.founders
         if is_sample:
             # Free up the existing ancestry mapping.
             x = self.A_tail[input_id]
@@ -281,6 +289,11 @@ class Simplifier:
                     if output_id == -1:
                         output_id = self.record_node(input_id)
                     self.record_edge(left, right, output_id, ancestry_node)
+                elif self.keep_founders and is_founder:
+                    if output_id == -1:
+                        output_id = self.record_node(input_id)
+                    self.record_edge(left, right, output_id, ancestry_node)
+
             else:
                 if output_id == -1:
                     output_id = self.record_node(input_id)
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -4550,6 +4550,7 @@ class TestSimplify(unittest.TestCase):
         filter_populations=True,
         filter_individuals=True,
         keep_unary=False,
+        founders=None,
     ):
         """
         Runs the Python test implementation of simplify.
@@ -4563,6 +4564,7 @@ class TestSimplify(unittest.TestCase):
             filter_populations=filter_populations,
             filter_individuals=filter_individuals,
             keep_unary=keep_unary,
+            founders=founders,
         )
         new_ts, node_map = s.simplify()
         if compare_lib:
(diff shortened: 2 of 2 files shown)


gtsambos:
This can be closed now that #775 has been resolved by #782.

PR #782, opened 2020-08-20
Keep path to ancient in simplify

jeromekelleher:
This is an initial prototype showing that we can solve #775 if we snip out the "used up" ancestry as we progress through simplify. This is initially just a proof of concept, as it does things in exactly the opposite way as would be intended.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4566,7 +4566,8 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, tsk_bookmark_t *start)
         if (start->sites == self->tables->sites.num_rows
             && start->mutations == self->tables->mutations.num_rows) {
             skip_sites = true;
-        } else {
+        } else if (start->sites != 0
+                   || start->mutations != 0) {
             ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
             goto out;
         }
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4547,6 +4547,7 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, tsk_bookmark_t *start)
 {
     int ret = 0;
     tsk_size_t edge_start = 0;
+    bool skip_sites = false;
 
     if (start != NULL) {
         if (start->edges > self->tables->edges.num_rows) {
@@ -4555,10 +4556,17 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, tsk_bookmark_t *start)
         }
         edge_start = start->edges;
 
-        /* For now, if migrations, sites or mutations are non-zero we get an error.
-         * This should be fixed: https://github.com/tskit-dev/tskit/issues/101
-         */
-        if (start->migrations != 0 || start->sites != 0 || start->mutations != 0) {
+        if (start->migrations != 0) {
+            ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
+            goto out;
+        }
+        /* We only allow sites and mutations to be specified as a way to
+         * skip sorting them entirely. Both sites and mutations must be
+         * equal to the number of rows */
+        if (start->sites == self->tables->sites.num_rows
+            && start->mutations == self->tables->mutations.num_rows) {
+            skip_sites = true;
+        } else {
             ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
             goto out;
         }
@@ -4574,13 +4582,15 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, tsk_bookmark_t *start)
             goto out;
         }
     }
-    ret = tsk_table_sorter_sort_sites(self);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_table_sorter_sort_mutations(self);
-    if (ret != 0) {
-        goto out;
+    if (!skip_sites) {
+        ret = tsk_table_sorter_sort_sites(self);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_table_sorter_sort_mutations(self);
+        if (ret != 0) {
+            goto out;
+        }
     }
 out:
     return ret;
@@ -4697,6 +4707,7 @@ typedef struct {
     /* When reducing topology, we need a map positions to their corresponding
      * sites.*/
     double *position_lookup;
+    int64_t edge_sort_offset;
 } simplifier_t;
 
 static int
@@ -5434,11 +5445,13 @@ simplifier_print_state(simplifier_t *self, FILE *out)
 
     fprintf(out, "--simplifier state--\n");
     fprintf(out, "options:\n");
-    fprintf(
-        out, "\tfilter_unreferenced_sites: %d\n", !!(self->options & TSK_FILTER_SITES));
-    fprintf(out, "\treduce_to_site_topology  : %d\n",
+    fprintf(out, "\tfilter_unreferenced_sites   : %d\n",
+        !!(self->options & TSK_FILTER_SITES));
+    fprintf(out, "\treduce_to_site_topology : %d\n",
         !!(self->options & TSK_REDUCE_TO_SITE_TOPOLOGY));
-    fprintf(out, "\tkeep_unary  : %d\n", !!(self->options & TSK_KEEP_UNARY));
+    fprintf(out, "\tkeep_unary              : %d\n", !!(self->options & TSK_KEEP_UNARY));
+    fprintf(out, "\tkeep_input_roots        : %d\n",
+        !!(self->options & TSK_KEEP_INPUT_ROOTS));
 
     fprintf(out, "===\nInput tables\n==\n");
     tsk_table_collection_print_state(&self->input_tables, out);
@@ -5740,6 +5753,25 @@ out:
     return ret;
 }
 
+static void
+simplifier_map_mutations(
+    simplifier_t *self, tsk_id_t input_id, double left, double right, tsk_id_t output_id)
+{
+    mutation_id_list_t *m_node;
+    double position;
+    tsk_id_t site;
+
+    m_node = self->node_mutation_list_map_head[input_id];
+    while (m_node != NULL) {
+        site = self->input_tables.mutations.site[m_node->mutation];
+        position = self->input_tables.sites.position[site];
+        if (left <= position && position < right) {
+            self->mutation_node_map[m_node->mutation] = output_id;
+        }
+        m_node = m_node->next;
+    }
+}
+
 static int TSK_WARN_UNUSED
 simplifier_add_ancestry(
     simplifier_t *self, tsk_id_t input_id, double left, double right, tsk_id_t output_id)
@@ -5770,6 +5802,7 @@ simplifier_add_ancestry(
             self->ancestor_map_tail[input_id] = x;
         }
     }
+    simplifier_map_mutations(self, input_id, left, right, output_id);
 out:
     return ret;
 }
@@ -5881,11 +5914,11 @@ simplifier_init(simplifier_t *self, tsk_id_t *samples, size_t num_samples,
     }
     memset(
         self->node_id_map, 0xff, self->input_tables.nodes.num_rows * sizeof(tsk_id_t));
-    ret = simplifier_init_samples(self, samples);
+    ret = simplifier_init_sites(self);
     if (ret != 0) {
         goto out;
     }
-    ret = simplifier_init_sites(self);
+    ret = simplifier_init_samples(self, samples);
     if (ret != 0) {
         goto out;
     }
@@ -5895,6 +5928,7 @@ simplifier_init(simplifier_t *self, tsk_id_t *samples, size_t num_samples,
             goto out;
         }
     }
+    self->edge_sort_offset = TSK_NULL;
 out:
     return ret;
 }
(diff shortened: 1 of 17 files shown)


python/tests/simplify.py
```diff
@@ -318,6 +318,7 @@ def process_parent_edges(self, edges):
                         max(x.left, edge.left), min(x.right, edge.right), x.node
                     )
                     S.append(y)
+                    self.used_up_ancestry[edge.child].append((y.left, y.right))
```
jeromekelleher, line 321: This is where we'd modify x if we really did this - we'd dynamically snip out the section ``y.left, y.right`` from the x segment. The details would be a bit tedious, but it's not hard.

petrelharp: We even *did* that in the initial version of simplify, but I don't think it was quite done, so I don't know if looking at that code would give us a head start on implementation...

jeromekelleher: Ah, OK. Former us thought it was a good idea too, intriguing!

python/tests/simplify.py
```diff
@@ -432,8 +433,57 @@ def finalise_references(self):
...
+                    num_intersecting += 1
+                # num_intersecting += 1
+                # print("\t", ancestry[gap])
+            if num_intersecting > 0:
+                # print("ROOT")
+                assert self.node_id_map[input_id] == -1
+                output_id = self.record_node(input_id)
+                # print(input_id, ancestry, gaps)
+                for gap in gaps:
+                    num_intersecting += 1
```
gtsambos, line 472: Is this count of intersections being used anywhere beyond line 466? It looks like as soon as this for loop is exited, you'll flush edges and then return to the start of the node loop in line 441, when `num_intersecting` will be rest to 0 again.

gtsambos (commented):
Okay, I can see what you're trying to do here @jeromekelleher! I'll start with the more technical comments before they leave my brain -- doing it all in one step at the end of the simplification processes introduces some inefficiencies that you wouldn't have if you were keeping track of the uncoalesced segments dynamically. For, instance, you calculate the list of uncoalesced segments `gaps` anew independently for each node, but they are highly dependent on each other -- every interval of uncoalesced ancestry that you 'find' as you progress further up the tree sequence is an interval that can be removed from 1 or more of the nodes lower in the tree.

Here's a sketch of how I'd approach this dynamically:

There'd be some array of lists of uncoalesced segments, say `self.uncoalesced_segs`, with one array element per node.
As you loop over edges `(L, R, P, C)` in the edge table: 
1.  Add the interval `(L, R)` to `self.uncoalesced_segs[u]` if one of the following is true:
  - `u` is a sample
  - `u` is the parent node `P`
2. Remove the intersection of the interval `(L, R)` from segment(s) in `self.uncoalesced_segs[C]`
3. Do all the normal `simplify` things (recording new edges, nodes etc)

gtsambos:
Does this make sense?

Also, I was trying to think about how this PR is different from what I've been working on in #679 and #488. I think the main difference is that here, each segment has a state of being either coalesced or uncoalesced, and although this state changes as you work through `simplify` -- an uncoalesced segment becomes coalesced as soon as it has an ancestor somewhere further up in the tree -- it is a 'global' property of the segment, in that it doesn't depend on the node's relationship to all other nodes, but only to its most closely related node. This is different from the IBD work, where the state of coalescence is specific/relative to every other sample node.

gtsambos:
Also also, I still feel this is a very complicated change to make just to solve the problem in #775. However, there would be other interesting information that you'd get for 'free' by doing it this way (assuming there aren't other costs to algorithmic speed etc). For instance, you'd be able to see how the total length of segments of uncoalesced ancestry changes as you move further back in time. I imagine this would have interesting/somewhat predictable theoretical properties, and would be very sensitive to aspects of demography

gtsambos:
Hmm, so it occurred to me that my approach in the [review comment](https://github.com/tskit-dev/tskit/pull/782#pullrequestreview-473234834) might partly negates one of the advantages [you saw](https://github.com/tskit-dev/tskit/issues/775#issuecomment-677600885) for this change -- it'll take up some memory, and though the amount of memory should decrease as the algorithm continues, it might not offset the amount of memory that you'll be saving by trimming some things out of  the list of ancestral segments `A`

jeromekelleher:
This makes sense @gtsambos, but I'm actually not suggesting that we store the ancestry gaps like I did in this quick hack. My proposal is to snip out the used up segments of ancestry as we go through the algorithm. I'm having a go at it at the moment, let's see what it looks like.

jeromekelleher:
Here's a working version of simplify with ancestry snipping and a ``keep_input_roots`` option. Implementing snipping out the ancestry was a bit tedious, as expected, but the code isn't very complicated and it's hard to imagine it slowing things down in any real sense (unless the segment chains get really fragmented), and it could allow us to save some memory.

Once this was done, the ``keep_input_roots`` option was easy, with a couple of tricky wrinkles:

1. Annoyingly, the edges we output aren't guaranteed to be in the right order in the output table. In particular, if a root is internal in another tree along the sequence, then when we write out the root edges for it later, they will not be adjacent in the table. I don't think there's much we can do about this, other then mitigate the full sort it implies (either sort part of the table, or else merge the edges intelligently). But, I guess this won't ever happen in the forward simulation case (right?), so maybe it's not so critical that we handle it well, we just have to detect reliably when the sort is needed.
2. There's probably some subtleties with how we handle mutations on the new edges - I haven't thought this through properly yet.

What do you think @petrelharp - is this worth pushing through for the forthcoming SLiM release? I can spent maybe another day on it, if you think it's worth it, and should get most stuff sorted in that much time.

jeromekelleher:
@gtsambos - I've updated this with a good version of the ancestry snipping. This could be useful for what you're working on?

petrelharp:
>  What do you think @petrelharp - is this worth pushing through for the forthcoming SLiM release? 

I've discussed with @bhaller and we think it is worthwhile - if this works out, then we would ditch all the confusing "first generation" stuff, which would be quite nice for the users (and make the code nicer).

petrelharp:
I've had a close read of this - looks good! - and see what the problem with sorting is. I haven't thought of a solution, but you're right: for forwards simulations all the extra roots are (usually) at the same time, so no sorting will be necessary.  (This isn't true if you add a new history-less population partway through, but that's quite uncommon, I think.)

bhaller:
> I've had a close read of this - looks good! - and see what the problem with sorting is. I haven't thought of a solution, but you're right: for forwards simulations all the extra roots are (usually) at the same time, so no sorting will be necessary. (This isn't true if you add a new history-less population partway through, but that's quite uncommon, I think.)

Well, I don't know how common it is, but it is certainly legal in SLiM, so we will need to think about how to handle it; I guess we'll need a flag to force the sort to happen in that generation, or something?

petrelharp:
> Well, I don't know how common it is, but it is certainly legal in SLiM, so we will need to think about how to handle it; I guess we'll need a flag to force the sort to happen in that generation, or something?

The proposal isn't to leave it up to the user, but rather to have simplify not do more sorting than is necessary. The nice thing here is that in the usual case in SLiM, there won't be any necessary sorting. For instance: we can check if the youngest root is no older than the last parent in the edge table; if so, no sorting is necessary.

jeromekelleher:
The ancestry snipping process has basically no effect on performance as far as I can tell. Running simplify on a 1.8 gig unsimplified tree sequence with 100k samples (200 generations using the haploid_wright_fisher example program). The difference between the two runs here is < 1%, so I'm not going to get worried about that right now. Likewise RAM hasn't really changed (but I haven't tried to free the used-up ancestry segments - that could be worthwhile).

This doesn't include the mapping roots bit (or any mutations) - I just wanted to get a like-for-like benchmark just crunching through lots of topology. Looks good so far.

**BEFORE**
```
$ /usr/bin/time python3 benchmark-simplify.py 
loaded in 10.364779717987403 s
nodes =  20100000
edges =  40000000
Simplified in  6.4478405870031565
nodes =  659800
edges =  3121843
15.06user 3.02system 0:17.75elapsed 101%CPU (0avgtext+0avgdata 8191192maxresident)k
0inputs+0outputs (0major+3155214minor)pagefaults 0swaps
```
**AFTER**
```
jk@tempest$ /usr/bin/time python3 benchmark-simplify.py 
loaded in 10.335074374976102 s
nodes =  20100000
edges =  40000000
Simplified in  6.466902140993625
nodes =  659800
edges =  3121843
14.95user 3.18system 0:17.73elapsed 102%CPU (0avgtext+0avgdata 8205248maxresident)k
0inputs+0outputs (0major+3158301minor)pagefaults 0swaps
```

petrelharp:
Wow, that's great. Let me know when you want me to have a look, or to come up with some more tests.

jeromekelleher:
Will do @petrelharp - I'm going to spend another couple of hours polishing it up, and it'd be great if  you could take a good look then.

jeromekelleher:
OK, I think this is ready for review now. I'd like to bump up the C test coverage a bit (we're not hitting a bunch of the awkward corner cases) but I'm not sure what's the best way to get at them. Should probably craft a "simplest...." example to do it. Probably the easiest way is to make a really fragmented small ts...

Anyway, ready for review. Good to get your eyes on this @petrelharp @benjeffery @gtsambos!

python/tests/test_topology.py
```diff
@@ -5414,6 +5616,143 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+                self.assertEqual(new_node.population, input_node.population)
+                self.assertEqual(new_node.individual, input_node.individual)
+                self.assertEqual(new_node.metadata, input_node.metadata)
+                # This should only be marked as a sample if it's an
+                # element of the samples list.
+                self.assertEqual(new_node.is_sample(), input_root in samples)
+                # Find the MRCA of the samples below this root.
+                root_samples = list(tree_with_roots.samples(root))
+                mrca = functools.reduce(tree_with_roots.mrca, root_samples)
+                if mrca != root:
```
petrelharp, line 5660: nice. this is a very good test.

python/tests/test_topology.py
```diff
@@ -5414,6 +5616,143 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+                    u = new_to_input_map[mrca]
+                    root_path = []
+                    while u != tskit.NULL:
+                        root_path.append(u)
+                        u = input_tree.parent(u)
+                    input_sites = {site.position: site for site in input_tree.sites()}
+                    new_sites = {
+                        site.position: site for site in tree_with_roots.sites()
+                    }
+                    positions = set(input_sites.keys()) & set(new_sites.keys())
```
petrelharp, line 5677: shouldn't these be the same, because of `filter_sites=False`?

python/tests/test_topology.py
```diff
@@ -5414,6 +5616,143 @@ def test_many_trees_recurrent_mutations_internal_samples(self):
...
+                    for position in positions:
+                        self.assertTrue(left <= position < right)
+                        new_site = new_sites[position]
+                        # We assume the metadata contains a unique key for each mutation.
```
petrelharp, line 5681: I don't see where these metadata keys were inserted?

python/tests/test_topology.py
petrelharp: Oh, never mind, I see why. But I guess that we could make sure we don't skip sites by doing
```
                    input_sites = {site.position: site for site in input_tree.sites()
                            if site.position >= left and site.position < right}
                    new_sites = {
                        site.position: site for site in tree_with_roots.sites()
                        if site.position >= left and site.position < right
                    }
                    self.assertEqual(set(input_sites.keys()), set(new_sites.keys()))
                    positions = input_sites.keys()
```

petrelharp: (I stuck this in my PR to this branch.)

python/tests/test_topology.py
petrelharp: Oh, I guess it's in `insert_branch_mutations`, which are the only ones with multiple mutations? Ok, good.

python/tests/tsutil.py
```diff
@@ -1288,3 +1290,27 @@ def genealogical_nearest_neighbours(ts, focal, reference_sets):
     L[L == 0] = 1
     A /= L.reshape((len(focal), 1))
     return A
+
+
+def coiterate(ts1, ts2, **kwargs):
```
petrelharp, line 1295: very nice

python/tests/test_topology.py
jeromekelleher: Yeah, I stuck some unique metadata into ``insert_branch_mutations``. We should test that this assumption is true here though, as it'll be easy to forget about this.

gtsambos:
> Anyway, ready for review. Good to get your eyes on this @petrelharp @benjeffery @gtsambos!

> @gtsambos - I've updated this with a good version of the ancestry snipping. This could be useful for what you're working on?

hooray, exciting! I'll look over tonight. And yes, if this works as well as it look like it does in your tests, it should be able to replace the awkward "deleting bits of ancestral segments" stuff that's currently giving me some grief in #679.

gtsambos:
Also, what's the best way to submit revisions to this (like @petrelharp has done) if I want to suggest something? Do I fork your repo, make my changes to the branch and then submit a PR as usual (?)

petrelharp:
I did `git fetch origin pull/782/head:new_simplify; git checkout new_simplify`, did edits, pushed to my fork of tskit-dev/tskit, and made a pull request (um, from my fork's page, I think?).

jeromekelleher:
> I did git fetch origin pull/782/head:new_simplify; git checkout new_simplify, did edits, pushed to my fork of tskit-dev/tskit, and made a pull request (um, from my fork's page, I think?).

You need to make a PR against this branch in my fork (jeromekelleher:simplify-keep-path-to-ancient). If you go to my fork and open a PR, it should give you the options to select the base branch and the PR branch.

Thanks!

jeromekelleher:
(Oops, I was quoting @petrelharp above who had already solved the problem!)

python/tests/test_topology.py
jeromekelleher: Good idea.

c/tskit/tables.c
```diff
@@ -4555,10 +4556,17 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, tsk_bookmark_t *start)
...
-         * This should be fixed: https://github.com/tskit-dev/tskit/issues/101
-         */
-        if (start->migrations != 0 || start->sites != 0 || start->mutations != 0) {
+        if (start->migrations != 0) {
+            ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
+            goto out;
+        }
+        /* We only allow sites and mutations to be specified as a way to
+         * skip sorting them entirely. Both sites and mutations must be
+         * equal to the number of rows */
```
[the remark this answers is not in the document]
jeromekelleher: Done.

jeromekelleher:
TODO 

- Update changelog
- Bump up C test coverage with a hand crafted example to cover awkward cases in snip ancestry.

benjeffery (approved):
I've been through this without deep-diving the algorithm, looks really good.

petrelharp:
> Bump up C test coverage with a hand crafted example to cover awkward cases in snip ancestry.

Ping if you'd like me to do this one.

jeromekelleher:
> Ping if you'd like me to do this one.

Thanks, but I think I know how to do it, so I'll have a go soon.

jeromekelleher:
Huh - turns out the reason we couldn't get test coverage on the squash loop is that it never gets called, ever. So, we can delete it! Gotta love the test suite here, it's totally awesome.

jeromekelleher:
OK, that's as much coverage as we're going to get - let's merge!

jeromekelleher:
Actuall, crud, no, let's not merge. Forgot to update the changelog.

gtsambos:
I had a brief look at this last night in bed, it looks really nice :) I'm going to look a bit more closely now at how you are dealing with the removal/modification of the removed segments in the ancestral map, but a general 👍  from me

gtsambos (commented):
hey, awesome job @jeromekelleher, this looks great! So far I've focused on the Python code, and specifically on the topology-related stuff (haven't thought much about the implications for remapping of mutations) -- I think I'm going to need a bit more time to look at the C code here in detail, as it just takes a while to parse what's going on here, but I think you should go ahead and merge the PR. Thanks also for cleaning up that test suite of simplify-esque examples -- there are defs a lot of overlap between those that it's good to consolidate.

python/tests/simplify.py
```diff
@@ -303,6 +321,51 @@ def merge_labeled_ancestors(self, S, input_id):
...
                 self.rewind_node(input_id, output_id)
 
+    def extract_ancestry(self, edge):
+        S = []
+        x = self.A_head[edge.child]
+
+        x_head = None
+        x_prev = None
+        while x is not None:
+            if x.right > edge.left and edge.right > x.left:
```
gtsambos, line 331: What happens if this condition is false ie. if there is *no* overlap between `edge` and `x`? In this case presumably you would want `x` to remain unchanged in `A`, but it looks as if you remove it regardless of whether this `if` condition is true or not (lines 365-366 further down). I couldn't see a test that addressed this fringe case explicitly but I'll have another look now

gtsambos: Oh actually nevermind, I realise that's the case your lines 357 - 3461 are covering here

PR #786, opened 2020-08-21
Use namedtuple for interval

### LINKED ISSUE #784: Create a namedtuple for interval values
There are several places in the tskit codebase where we return an interval of `[left, right]`. It seems sensible to return a `namedtuple` instead, so we can either say `interval[0]` or `interval.left`. Any reason why not to do this?

hyanwong:
Fixes #784

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -645,8 +645,8 @@ class TestTreeSequence(HighLevelTestCase):
                 children[edge.parent].add(edge.child)
             while tree.interval[1] <= left:
                 tree = next(trees)
-            self.assertTrue(left >= tree.interval[0])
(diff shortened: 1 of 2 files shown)


hyanwong:
It might be nice (although certainly not vital) to add a `.span` property too:

```
BaseInterval = collections.namedtuple("BaseInterval", ["left", "right"])
class Interval(BaseInterval):
    @property
    def span(self):
        return self.right-self.left
```

Overkill?

jeromekelleher:
I don't see any harm, once it doesn't clash with the namedtuple behaviour.

benjeffery (approved):
Great, thanks for the extra tests. Happy to have `span` later as a separate PR.

PR #787, opened 2020-08-21
Add include_terminal param

### LINKED ISSUE #783: Extra param in edge_diffs to output all the final edges
I use the `edge_diffs()` iterator quite a lot. Occasionally I would like to do something with all the edges that remain at the end of the iteration. There might be a workaround, but the easiest thing to me would be if we could call something like `edge_diffs(include_terminal_edges=True)` which would report one extra iteration, listing the final `edges_out` that would be removed to clear the edges at the end of the sequence. Any objections to this?

hyanwong:
Fixes #783

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -378,16 +378,16 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
 }
 
 static void
-verify_tree_diffs(tsk_treeseq_t *ts)
+verify_tree_diffs(tsk_treeseq_t *ts, tsk_flags_t options)
 {
-    int ret;
+    int ret, valid_tree;
     tsk_diff_iter_t iter;
     tsk_tree_t tree;
     tsk_edge_list_node_t *record;
     tsk_edge_list_t records_out, records_in;
-    size_t num_nodes = tsk_treeseq_get_num_nodes(ts);
-    size_t j, num_trees;
-    double left, right;
+    tsk_size_t num_nodes = tsk_treeseq_get_num_nodes(ts);
+    tsk_size_t j, num_trees;
+    double lft, rgt;
     tsk_id_t *parent = malloc(num_nodes * sizeof(tsk_id_t));
     tsk_id_t *child = malloc(num_nodes * sizeof(tsk_id_t));
     tsk_id_t *sib = malloc(num_nodes * sizeof(tsk_id_t));
@@ -400,17 +400,17 @@ verify_tree_diffs(tsk_treeseq_t *ts)
         child[j] = TSK_NULL;
         sib[j] = TSK_NULL;
     }
-    ret = tsk_diff_iter_init(&iter, ts);
+    ret = tsk_diff_iter_init(&iter, ts, options);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = tsk_tree_init(&tree, ts, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-    ret = tsk_tree_first(&tree);
-    CU_ASSERT_EQUAL_FATAL(ret, 1);
+    valid_tree = tsk_tree_first(&tree);
+    CU_ASSERT_EQUAL_FATAL(valid_tree, 1);
     tsk_diff_iter_print_state(&iter, _devnull);
 
     num_trees = 0;
-    while ((ret = tsk_diff_iter_next(&iter, &left, &right, &records_out, &records_in))
-           == 1) {
+    while (
+        (ret = tsk_diff_iter_next(&iter, &lft, &rgt, &records_out, &records_in)) == 1) {
         tsk_diff_iter_print_state(&iter, _devnull);
         num_trees++;
         /* Update forwards */
@@ -420,9 +420,11 @@ verify_tree_diffs(tsk_treeseq_t *ts)
         for (record = records_in.head; record != NULL; record = record->next) {
             parent[record->edge.child] = record->edge.parent;
         }
-        /* Now check against the sparse tree iterator. */
-        for (j = 0; j < num_nodes; j++) {
-            CU_ASSERT_EQUAL(parent[j], tree.parent[j]);
+        if (valid_tree) {
+            /* Now check against the sparse tree iterator. */
+            for (j = 0; j < num_nodes; j++) {
+                CU_ASSERT_EQUAL(parent[j], tree.parent[j]);
+            }
         }
         /* Update backwards */
         for (record = records_out.tail; record != NULL; record = record->prev) {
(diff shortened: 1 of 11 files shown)


c/tskit/trees.c
```diff
@@ -4371,7 +4371,8 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
  * ======================================================== */
 
 int TSK_WARN_UNUSED
-tsk_diff_iter_init(tsk_diff_iter_t *self, tsk_treeseq_t *tree_sequence)
+tsk_diff_iter_init(
+    tsk_diff_iter_t *self, tsk_treeseq_t *tree_sequence, bool include_terminal)
```
jeromekelleher, line 4375: Let's make this a ``tsk_flags_t options`` argument - this is the approach taken elsewhere in the API, as it's more flexible and breaks less code when new options are added.

python/tests/test_highlevel.py
```diff
@@ -655,6 +655,19 @@ def test_edge_diffs(self):
         for ts in get_example_tree_sequences():
             self.verify_edge_diffs(ts)
 
+    def test_edge_diffs_include_terminal(self):
+        for ts in get_example_tree_sequences():
+            edges = set()
+            i = 0
+            for _, e_out, e_in in ts.edge_diffs(include_terminal=True):
```
jeromekelleher, line 662: We need to test the value of left and right here as well. What are they?

python/tskit/trees.py
```diff
@@ -3457,15 +3457,17 @@ def edge_diffs(self):
...
-        (i.e. it does not report a final removal of all remaining edges); the
-        number of iterations is thus always equal to the number of trees in the
-        tree sequence.
-
+        :param bool include_terminal: If False (default), the iterator terminates
+            after the final interval in the tree sequence (i.e. it does not
+            report a final removal of all remaining edges), so the number
+            of iterations will be equal to the number of trees in the tree
+            sequence. If True, there is one more iteration, with the last
+            ``edges_out`` value reporting all the edges contained in the final
```
jeromekelleher, line 3465: What's be value of left and right here?

c/tskit/trees.c
```diff
@@ -4499,10 +4503,10 @@ tsk_diff_iter_next(tsk_diff_iter_t *self, double *ret_left, double *ret_right,
...
 
         right = sequence_length;
-        if (self->insertion_index < self->num_edges) {
+        if (self->insertion_index < (tsk_id_t) self->num_edges) {
             right = TSK_MIN(right, edges->left[insertion_order[self->insertion_index]]);
         }
-        if (self->removal_index < self->num_edges) {
+        if (self->removal_index < (tsk_id_t) self->num_edges) {
             right = TSK_MIN(right, edges->right[removal_order[self->removal_index]]);
         }
```
jeromekelleher, line 4511: We need some C tests of this to make sure we're not touching any uninitialised memory. Easiest thing to do is to run through the iterator setting a ``parent`` array as usual, and then check at the end that the parent array is all -1 if TSK_INCLUDE_TERMINAL has been set.

jeromekelleher (changes requested):
Looks good, but needs more tests. Definitely on the C side we want to exercise this on a few different topology types - what happens on an empty ts? An ordinary ts with several trees? We have to think through all the corner cases.

python/tskit/trees.py
hyanwong: Good point. By default it is `right==left==sequence_length`, which seems reasonable to me, so I've documented this and added to to the high_level tests.

c/tskit/trees.c
hyanwong: Just done this in https://github.com/tskit-dev/tskit/pull/787/commits/b4a3f1d9d26c44b8330bb9bf5276ee7e7c621093 - I've not written many C tests before, so let me know if anything is wrong / unconventional.

hyanwong:
> Looks good, but needs more tests. Definitely on the C side we want to exercise this on a few different topology types - what happens on an empty ts? An ordinary ts with several trees? We have to think through all the corner cases.

Can we not test an empty ts, multi-tree ts etc at the python level? We have a function `get_example_tree_sequences()` in test_highlevel.py which can return various pathological tree sequences like this. Actually, having looked at it, I see it doesn't test a ts with no edges, so I'll open a PR for that.

jeromekelleher:
> Can we not test an empty ts, multi-tree ts etc at the python level? We have a function get_example_tree_sequences() in test_highlevel.py which can return various pathological tree sequences like this. 

No, because we can't run valgrind when Python is running and would therefore not detect memory access issues (which are non-deterministic and differ across platforms, so you can't just say "it doesn't segfault, so it's fine").

There's loads of tests checking for empty tree sequences across the suite - the test_highlevel.py approach isn't how we're doing things any more.

hyanwong:
> No, because we can't run valgrind when Python is running and would therefore not detect memory access issues (which are non-deterministic and differ across platforms, so you can't just say "it doesn't segfault, so it's fine").

Ah, OK. Apart from your suggestion above for all `parent[j] == -1`, I'm not familiar with the way you might check memory issues here. But now that I have implemented that, perhaps I can just add `test_multiroot_diff_iter` and `test_empty_diff_iter` functions to test_trees.c?

> There's loads of tests checking for empty tree sequences across the suite - the test_highlevel.py approach isn't how we're doing things any more.

Added anyway: I think there's no harm, and a function to produce a whole set of different semi-pathological tree sequences seems a useful thing to have around.

jeromekelleher:
> Ah, OK. Apart from your suggestion above for all parent[j] == -1, I'm not familiar with the way you might check memory issues here. But now that I have implemented that, perhaps I can just add test_multiroot_diff_iter and test_empty_diff_iter functions to test_trees.c?

Once the tests are implemented valgrind is run on them automatically, checking for these errors.

python/_tskitmodule.c
```diff
@@ -9869,7 +9874,8 @@ TreeDiffIterator_init(TreeDiffIterator *self, PyObject *args, PyObject *kwds)
     }
     memset(self->tree_diff_iterator, 0, sizeof(tsk_diff_iter_t));
     err = tsk_diff_iter_init(self->tree_diff_iterator,
-            self->tree_sequence->tree_sequence);
+            self->tree_sequence->tree_sequence,
+            include_terminal);
```
benjeffery, line 9878: Shouldn't this be `options`?

c/tests/test_trees.c
```diff
@@ -380,76 +380,94 @@ verify_tree_next_prev(tsk_treeseq_t *ts)
...
-        num_trees++;
-        /* Update forwards */
-        for (record = records_out.head; record != NULL; record = record->next) {
-            parent[record->edge.child] = TSK_NULL;
-        }
-        for (record = records_in.head; record != NULL; record = record->next) {
-            parent[record->edge.child] = record->edge.parent;
-        }
-        /* Now check against the sparse tree iterator. */
+    for (include_terminal = 0; include_terminal <= 1; include_terminal++) {
```
benjeffery, line 399: For functions with options I've been using the pattern at https://github.com/tskit-dev/tskit/blob/master/c/tests/test_tables.c#L782 as I think it is clearer what's going on.

benjeffery (changes requested):
Think I found am incorrect variable usage in the python C layer, and one comment about test style.

python/_tskitmodule.c
hyanwong: Oh, very good catch. Thanks.

c/tests/test_trees.c
hyanwong: Sorry - is that the right link @benjeffery - I can't see anything about options there?

c/tests/testlib.c
```diff
@@ -48,7 +48,21 @@ const char *single_tree_ex_mutations
       "2    2     1   -1\n"
       "2    3     1   -1\n";
 
-/* Example from the PLOS paper */
+/*** Example from the PLOS paper ***/
```
jeromekelleher, line 51: Fantastic! Thanks for putting these in, that's a major bonus.

hyanwong: No probs. I just need to sort out the C linting (not set up on my system) and Ben's last comment, and this should be done, ready for 0.3.1.

c/tskit/trees.c
jeromekelleher: Nope, these look ideal, thanks (modulo issues @benjeffery points out)

c/tests/testlib.c
benjeffery: I can c lint if needed.

c/tests/test_trees.c
benjeffery: Yeah, that's the right link - basically call the test function with each option. In the link:
```
    test_edge_table_with_options(0);
    test_edge_table_with_options(TSK_NO_METADATA);
```

hyanwong: Oh, I see. Sorry - I was being dense. I agree, much clearer.

c/tests/testlib.c
hyanwong: Thanks @benjeffery - it would save me quite a lot of hassle if you could lint the C, so please do if that's OK. I've just pushed the commit with the "options" parameter, which is much better (thanks for the tip). I can also squash the commits into 1, or I guess you can, it it's easy (the first one, f9a4a972d346862c5ba22e4673b856a149256c01) is a separate PR, though, so perhaps leave that)

benjeffery: Done, force pushed.

hyanwong: Thanks a lot, Ben!

benjeffery (changes requested):
Looking good, thanks @hyanwong! We just need a changelog before merge.

hyanwong:
> Looking good, thanks @hyanwong! We just need a changelog before merge.

Done, thanks. Note that this includes the change in https://github.com/tskit-dev/tskit/pull/788 as the first commit in the PR, so that can be PR can be closed if this is merged.

jeromekelleher:
Let's hold off on this until our bugfix releases have been shipped.

hyanwong:
I think this is ready to merge @benjeffery ?

benjeffery (approved):
LGTM, lets merge.

jeromekelleher (approved):

benjeffery:
@hyanwong I've just force pushed to resolve the conflict and add the change to the C changelog too as this is a change to both APIs.

PR #788, opened 2020-08-21
Change diff iter types to tsk_ equivalents

hyanwong:
As discussed on slack. I don't think any of these are exposed to Python, so conversion to/from python equivalents shouldn't catch us out here.


jeromekelleher (approved):
LGTM, thanks.

But let's keep this one queued up for post release.

hyanwong:
Included with #787

PR #794, opened 2020-08-26
Deprecate impute_missing_data and TSK_IMPUTE_MISSING_DATA

benjeffery:
Fixes #716 
I have made the arguments to the functions that take the new argument all keyword-only. This didn't break any of our test code and will only break those users that are using the arguments positionally. While this break could be avoided, I think it is best as it allows us to remove the old argument eventually with out more subtle breakage.

PR #797, opened 2020-08-26
missing data model clarification

### LINKED ISSUE #795: clarify missing data model in docs
The documentation needs to say that having a mutation over an isolated sample makes it not missing. I can have a go at this.

petrelharp:
I don't *think* this got hit in #794 or #778.

### DIFF
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -993,7 +993,7 @@ is ``7``.
 Missing data
 ************
 
-Missing data is encoding in tskit using the idea of *isolated samples*.
+Missing data is encoded in tskit using the idea of *isolated samples*.
 A sample's genotype is missing at a position if it is *isolated* and if it has
 no mutations directly above it at that position. An isolated sample is a sample
 node (see :ref:`sec_data_model_definitions`) that has no children and no
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -993,15 +993,18 @@ is ``7``.
 Missing data
 ************
 
-Missing data is encoding in tskit using the idea of *isolated samples*. An
-isolated sample is a sample node (see :ref:`sec_data_model_definitions`)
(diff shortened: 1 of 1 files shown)


docs/data-model.rst
```diff
@@ -993,15 +993,18 @@ is ``7``.
...
-Missing data is encoding in tskit using the idea of *isolated samples*. An
-isolated sample is a sample node (see :ref:`sec_data_model_definitions`)
-that has no children and no parent, in a particular tree. This is encodes
-the idea that we don't know anything about a given sample over a specific
-interval. This definition covers the standard idea of missing data in
-genomics (where we do not know the sequence of a given contemporary sample at
-some site, for whatever reason), but also more generally the idea that
-we may not know anything about large sections of the genomes of ancestral
-samples.
+Missing data is encoding in tskit using the idea of *isolated samples*.
```
benjeffery, line 996:
```suggestion
Missing data is encoded in tskit using the idea of *isolated samples*.
```

docs/data-model.rst
```diff
@@ -993,15 +993,18 @@ is ``7``.
...
+A sample's genotype is missing at a position if it is *isolated* and if it has
```
benjeffery, line 997: Should we use "site" rather than position?

petrelharp: Gah, a good question. Could be either, but I think that "position" is the position along the genome, and so that's where the genotype is (or is not known), while "site" is a special object that might also be associated with the position. For instance, ts.site( ) doesn't get you genotypes, but ts.variants( ) does (although the site is one of the things returned). That's why I wrote 'position'?

benjeffery: Cool, I'm happy with position as you can only have one site per position anyway.

jeromekelleher (approved):
LGTM!

PR #798, opened 2020-08-27
Add Tree.draw_tikz() to output tex/tikz drawing code.

grahamgower:


grahamgower:
Ping @gtsambos.

python/tskit/drawing.py
```diff
@@ -221,6 +246,153 @@ def remap_style(original_map, new_key, none_value):
...
+            },
+            right node text/.style = {
+                node text,
+                above right=.1,
+            },
+            leaf node text/.style = {
+                node text,
+                below=.1,
+            },
+            % User customisation goes here. $user_style_text
```
jeromekelleher, line 308: Did you mean to have $user_style_text on a comment line?

jeromekelleher (commented):
Looks good to me @grahamgower, but I just had a quick scan.

python/tskit/drawing.py
grahamgower: Yeah, I made an arbitrary choice about how to deal with something annoying in the required syntax. The `options` part of the tikzpicture (the bit inside `[` and `]`) doesn't seem to allow empty lines. If `$user_style_text` is on its own line, then this is a syntax error when the user doesn't specify anything. So I just prepend "\n" to the style when the user *does* specify something. This is ugly, and I'm sure we can do better.

jeromekelleher: Sounds good - needs a bit of explanation somewhere, though, as this is the sort of thing someone would "fix" while doing a pass through the code.

gtsambos:
fyi @grahamgower, I'm forking this now so I can play around with getting mutations in here

gtsambos:
Hi @grahamgower, I've created a PR against your fork (let me know if you have trouble with this). The additional code puts mutations on the plots, like this: 

image

So far the code assumes mutation times aren't known -- but the case when they are shouldn't be difficult either, will look into it later this weekend

grahamgower:
Thanks, that looks great!

grahamgower:
and, TreeSequences...
s

gtsambos:
very nice 👀  I'd like to get a handle on the TreeSequence-wide `draw_tikz` method too, perhaps I could do some minor fixups on that to help me learn my way around it? I'm thinking scaling all the trees according to the max root time, and perhaps also adding axis labels

grahamgower:
Yeah, adding the scaling and axes would be really useful. Just give me a moment, I've changed to use a `\pic`, and need to commit that.

grahamgower:
Ok, have at it @gtsambos! Now we can do silly things like this:
```python
import msprime
ts = msprime.simulate(10, recombination_rate=1, mutation_rate=1, random_seed=1)
ts.draw_tikz(
    "/tmp/s.tex",
    tree_height_scale="rank",
    standalone=True,
    preamble="\\usetikzlibrary{3d, shapes.geometric}",
    style="""%
        every pic/.append style={canvas is zy plane at x},
        mutationlabels/.append style = {
            opacity=0,
        },
        mutations/.append style = {
            star,
            star point ratio=2.5,
        },\
        """
)
```
s1

grahamgower:
With bigger sample sizes (n=1000, with 19 trees) I get: `! TeX capacity exceeded, sorry [main memory size=5000000]`. Maybe we can structure things differently to improve this, but it's likely we're not going to be able to scale as well as for svg.

grahamgower:
The solution for large drawings is to compile with `lualatex`, rather than `pdflatex`. It took a minute or two, but was actually quicker than calling draw_svg(). I guess this is beyond the limit of what's sensible.
```
msprime.simulate(1000, recombination_rate=1, mutation_rate=1, random_seed=1)
```
```
$ ls -lh /tmp/s.{tex,pdf,svg}
-rw-r--r-- 1 grg grg 4,5M Aug 29 10:15 /tmp/s.pdf
-rw-r--r-- 1 grg grg  16M Aug 29 10:23 /tmp/s.svg
-rw-r--r-- 1 grg grg 1,5M Aug 29 10:14 /tmp/s.tex
```

gtsambos:
> With bigger sample sizes (n=1000, with 19 trees) I get: ! TeX capacity exceeded, sorry [main memory size=5000000]. Maybe we can structure things differently to improve this, but it's likely we're not going to be able to scale as well as for svg.

> The solution for large drawings is to compile with lualatex, rather than pdflatex. It took a minute or two, but was actually quicker than calling draw_svg(). I guess this is beyond the limit of what's sensible.

I wouldn't be too worried about this -- I doubt people will be using `draw_tikz` on this kind of scale.

jeromekelleher:
Yeah, I'm 100% not worried about rendering large trees within tex. This is fundamentally for printed page output, so if it's too big to go on a printed page we don't care about it.

benjeffery:
@grahamgower does this need tests to be mergable? Or are there still open questions about the implementation?

grahamgower:
Probably tests and docs are the biggest things outstanding. And making sure the API is consistent enough with the other drawing functions would be a good idea.

grahamgower:
Rebased. I'll take another look though this in the coming week.

jeromekelleher:
This is really useful and a great addition, but I'm gone off the idea of incorporating it into tskit. Our experience with the SVG viz is that it quickly turns into a lot of code, and it's quite hard to test. In this case, we'd need to pull in a tex engine to CI and for all of our client side dependencies. Keeping consistency between the various viz APIs is a pain also.

So, I suggest spinning it out into an independent project? ``tstikz`` or something?

benjeffery:
> So, I suggest spinning it out into an independent project? `tstikz` or something?

I think this is probably the right thing to do, and that maybe even the svg code is hefty enough not be in tskit, but that is another discussion!

hyanwong:
Does this belong in https://github.com/tskit-dev/tsviz ?

benjeffery:
> Does this belong in https://github.com/tskit-dev/tsviz ? Perhaps we could move the PR over there and eventually merge it?

Sounds good to me!

benjeffery:
Closing this as we decided to move it to tsviz - have opened https://github.com/tskit-dev/tsviz/issues/9 so this PR code doesn't get lost.

PR #799, opened 2020-08-27
Shift language on error and docs to replace 'impute'

benjeffery:
Very minor, but I realised our error message still referenced imputation, and would be confusing.


jeromekelleher (approved):
Good catch, thanks.

PR #802, opened 2020-08-27
Fix OSX wheels test deps

benjeffery:
We need a better way of managing all the different deps - or to remove these tests as we did in msprime, will do after release.


benjeffery:
Merging manually as mergify doesn't support merges to actions confog changes.

PR #803, opened 2020-08-27
Update trees.py

### LINKED ISSUE #739: Clarify docs for map_mutations
At the moment, it's unclear to me from the [docs for map_mutations](https://github.com/tskit-dev/tskit/blob/2aca42cfb9960512e8fcc6c418aa44d43c4f6607/python/tskit/trees.py#L2210) whether:
a. A single returned mutation with have the `parent` attribute set to tskit.NULL (I would hope so)
b. The mutations are returned such that parents mutations occur before children in the list
c. The `parent` attribute is an ID into *the 0-based list of returned mutations* (I can't see what else it could be, but the docs could be clearer here, I think.

I'm happy to reword the docs, if my assumptions (a) (b) and (c) are all correct.

hyanwong:
Fixes #739

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2250,14 +2250,15 @@ class Tree:
 
         The state reconstruction is returned as two-tuple, ``(ancestral_state,
         mutations)``, where ``ancestral_state`` is the allele assigned to the
-        tree root(s) and ``mutations`` is a list of :class:`Mutation` objects.
+        tree root(s) and ``mutations`` is a list of :class:`Mutation` objects,
+        ordered as :ref:`required in a mutation table<sec_mutation_requirements>`.
         For each mutation, ``node`` is the tree node at the bottom of the branch
         on which the transition occurred, and ``derived_state`` is the new state
-        after this mutation. When multiple mutations are returned the ``parent``
-        property contains the index of the previous mutation on the path to root
-        (see the :ref:`sec_mutation_table_definition` for more information on the
-        concept of mutation parents). All other attributes of the :class:`Mutation`
-        object are undefined and should not be used.
+        after this mutation. The ``parent`` property contains the index in the
+        returned list of the previous mutation on the path to root, or ``tskit.NULL``
+        if there are no previous mutations (see the :ref:`sec_mutation_table_definition`
+        for more information on the concept of mutation parents). All other attributes
+        of the :class:`Mutation` object are undefined and should not be used.
 
         .. note::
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -2250,14 +2250,15 @@ def map_mutations(self, genotypes, alleles):
...
+        ordered as :ref:`required in a mutation table<sec_mutation_requirements>`.
         For each mutation, ``node`` is the tree node at the bottom of the branch
         on which the transition occurred, and ``derived_state`` is the new state
-        after this mutation. When multiple mutations are returned the ``parent``
-        property contains the index of the previous mutation on the path to root
-        (see the :ref:`sec_mutation_table_definition` for more information on the
-        concept of mutation parents). All other attributes of the :class:`Mutation`
-        object are undefined and should not be used.
+        after this mutation. The ``parent`` property contains the index in the
+        returned list of the previous mutation on the path to root, or ``tskit.NULL``
```
jeromekelleher, line 2258: It's not the index in the returned list, it's the parent mutation's ID (i.e., it's index in the mutation table)

python/tskit/trees.py
```diff
@@ -2250,14 +2250,15 @@ def map_mutations(self, genotypes, alleles):
...
+        if there are no revious mutations (see the :ref:`sec_mutation_table_definition`
```
jeromekelleher, line 2259: typo, "revious"

jeromekelleher (changes requested):
Some inaccuracies here.

python/tskit/trees.py
hyanwong: I don't think it can be, can it (which is why I wanted to clarify this). `map_mutations` has no concept of a pre-existing mutation table, does it? After all, the mutations that it returns have not yet been added to the table, so they don't really have an ID yet?

hyanwong: [Here](https://github.com/tskit-dev/tsinfer/blob/ab5cddc9f498f28eb6034b1c16a1178f36aafa4e/tsinfer/inference.py#L637) for instance, I explicitly use `mutation.parent` as an index into the returned list, which needs to be converted into an id in the mutation table.

jeromekelleher: Sorry, my bad, I didn't look at the context. I'll look at this again next week - to much going on right now!

hyanwong: No hurry!

benjeffery (approved):
Makes sense to me, I don't think this is contentious so I'll merge.

PR #810, opened 2020-08-28
Fix windows wheel test

benjeffery:
Fixes a braindead error I made that slipped through testing as it only errored if OSX wheel building finishes first.


benjeffery:
Manual merge to github actions config.

PR #811, opened 2020-08-28
Set .span property for Interval

hyanwong:


hyanwong:
Tests are very minimal, but I can't think of any corner cases, really.

jeromekelleher:
It'd be nice to do a *bit* more testing. To pass these tests, all I'd need to do is have
```python
@property
def span(self):
     return 1.0
```

hyanwong:
Like this? I simply changed it to test a few trees.

python/tests/test_highlevel.py
```diff
@@ -2363,12 +2363,19 @@ def test_seek(self):
...
-        tree = ts.first()
-        self.assertEqual(tree.interval[0], 0)
-        self.assertEqual(tree.interval.left, 0)
-        self.assertEqual(tree.interval[1], 1)
-        self.assertEqual(tree.interval.right, 1)
+        ts = msprime.simulate(10, recombination_rate=1, random_seed=1)
+        self.assertGreater(ts.num_trees, 1)
+        breakpoints = list(ts.breakpoints())
+        self.assertEqual(breakpoints[0], 0)
+        self.assertAlmostEqual(breakpoints[-1], ts.sequence_length)
```
jeromekelleher, line 2370: Should be exactly equal, shouldn't they? We'd only expect numerical differences if we've performed some calculations on the values.

jeromekelleher (commented):
Looks good, thanks @hyanwong. Minor comment.

benjeffery (commented):
Sorry @hyanwong I missed the notifications on your push. Looks good modulo JK's comment.

python/tests/test_highlevel.py
hyanwong: Fine. I wasn't sure how the breakpoints were calculated, but if the last one is simply passed through from the seq length, then that should be fine. I'll change it.

hyanwong:
Should be done now

jeromekelleher (approved):

PR #812, opened 2020-08-28
More precise description of mutation order

hyanwong:
From reading of the sites iterator / accessor, I think this is right


hyanwong:
@petrelharp - I see that the changelog says "Table sorting orders mutations by non-increasing time per-site, which is also a requirement for a valid tree sequence (:user:`benjeffery`, :pr:`672`).", but the docs (data-model.rst) still say "The :meth:`TableCollection.sort` method ensures that mutations are sorted according site ID, but does not at present enforce that mutations occur after their parent mutations." - I guess this need deleting?

petrelharp:
Oh, yes! good catch.

Issue #684, opened 2020-06-18
Index required for efficient random access for trees


jeromekelleher:
Current TreeSequence.at() and other methods to seek to a particular tree work by iterating along the trees. This isn't very efficient, if we want to get a tree in the middle of the sequence.

Can we seek efficiently using the current indexes, or do we need to build some sort of interval tree?

jeromekelleher:
Related: #24 and #4

grahamgower:
The intervals are non-overlapping, so an interval tree should be unnecessary---just binary search the sorted list of start coordinates.

jeromekelleher:
They are overlapping though, aren't they? We *should* be able to get them from the existing indexes though, by binary searching the in-order and out-order indexes...

jeromekelleher:
See also #685

jeromekelleher:
I've thought about this again, and no, I can't see anyway of doing this efficiently without a different type of index. The indexes we have tell us the order in which edges go in and out in a left-to-right traversal, basically by sorting on the left  and right coordinates. Since the edges for a given tree are *not* adjacent within these lists in general, then binary searching for a given position will only give you the edges that start/end closest to that point. You can have long edges (say) spanning the entire interval which are not inserted/removed anywhere near this point along the genome.

So, we will need some sort of data structure that'll support interval overlap queries if we're going to implement this (and edge finding in general, #685) efficiently.

benjeffery:
> They are overlapping though, aren't they? We _should_ be able to get them from the existing indexes though, by binary searching the in-order and out-order indexes...

To do this you'd need to find the first in edge `i` that has `i.right > target`, then you also find the first out edge `o` that has `o.right > i.left`. Then you could continue normal tree iteration from that point. 
I'm not sure this is so simple as binary search though as in is not sorted by right and vice-versa.

benjeffery:
Heh, seems we wrote at the same time.

grahamgower:
Oh, yes. Sorry for the misunderstanding. So you possibly want nested container lists. https://academic.oup.com/bioinformatics/article/23/11/1386/199545

jeromekelleher:
> To do this you'd need to find the first in edge i that has i.right > target, then you also find the first out edge o that has o.right > i.left. Then you could continue normal tree iteration from that point.

[...]

PR #815, opened 2020-08-28
Random split polytomy

### LINKED ISSUE #809: Randomly resolve polytomies
I've finally got some working code to randomly resolve polytomies in a tree sequence. It seems to work for the admittedly small sample of inferred trees that I have tried. It tries to be clever by not resolving per tree, but per polytomy (hence if an identical polytomy spans several trees, it will only resolve the polytomy once, creating fewer edges than a per-tree approch, which should mean it scales to larger sample sizes better). I'm posting it here so that:

a. I don't lose it (!)
b. We can decide if we want something like this in the base `tskit` library, as a method on a tree sequence.
c. @hyl317 can use it if he wants, for ARGweaver compatibility (although his current solution may be faster)

Obviously it needs a fair bit of tidying up. It's also slow, but there's also considerable scope for optimization, I think. It requires the PR I made at #787

```python
import collections
import itertools

import numpy as np
import tskit

def resolve_polytomy(parent_node_id, child_ids, new_nodes_by_time, rng):
    """
    For a polytomy and list of child node ids, return a list of (child, parent) tuples,
    describing a bifurcating tree, rooted at parent_node_id, where the new_nodes_by_time
    have been used to break polytomies. All possible topologies should be equiprobable.
    """
    assert len(child_ids) == len(new_nodes_by_time) + 2
    edges = [[child_ids[0], None], ]  # Introduce a single edge that will be deleted later
    edge_choice = rng.integers(0, np.arange(1, len(child_ids) * 2 - 1, 2))
    tmp_new_node_lab = [parent_node_id] + new_nodes_by_time
    assert len(edge_choice) == len(child_ids) - 1
    for node_lab, child_id, target_edge_id in zip(tmp_new_node_lab, child_ids[1:], edge_choice):
        target_edge = edges[target_edge_id]
        # print("target", target_edge)
        # Insert to keep edges in time order of parent
        edges.insert(target_edge_id, [child_id, node_lab])
        edges.insert(target_edge_id, [target_edge[0], node_lab])
        target_edge[0] = node_lab
    # We need to re-map the internal nodes so that they are in time order
    real_node = iter(new_nodes_by_time)
    edges.pop() # remove the unary node at the top
    node_map = {c: c for c in child_ids}
    # print("orig_edges", edges)
    # print("parent IDs to allocate", new_nodes_by_time)
    # last edge should have the highest node
    node_map[edges[-1][1]] = parent_node_id
    for e in reversed(edges):
        # edges should be in time order - the oldest one can be give the parent_node_id
        if e[1] not in node_map:
            node_map[e[1]] = next(real_node)
        if e[0] not in node_map:
            node_map[e[0]] = next(real_node)
        e[0] = node_map[e[0]]
        e[1] = node_map[e[1]]
    # print("mapped edges", edges)
    assert len(node_map) == len(new_nodes_by_time) + len(child_ids) + 1
    return edges


def resolve_polytomies(ts, *, epsilon=1e-10, random_seed=None):
    """
    For a given parent node, an edge in or an edge out signifies a change in children
    Each time such a change happens, we cut all existing edges with that parent,
    and add the previous portion in to the new edge table. If, previously, there were
    3 or more children for this node, we break the polytomy at random
    """
    rng = np.random.default_rng(seed=random_seed)

    tables = ts.dump_tables()
    edges_table = tables.edges
    nodes_table = tables.nodes
    # Store the left of the existing edges, as we will need to change it if the edge is split
    existing_edges_left = edges_table.left
    # Keep these arrays for handy reading later
    existing_edges_right = edges_table.right
    existing_edges_parent = edges_table.parent
    existing_edges_child = edges_table.child
    existing_node_time = nodes_table.time

    edges_table.clear()

    edges_for_node = collections.defaultdict(set)  # The edge ids dangling from each active node
    nodes_changed = set()

    for interval, e_out, e_in in ts.edge_diffs(include_terminal=True):
        for edge in itertools.chain(e_out, e_in):
            if edge.parent != tskit.NULL:
                nodes_changed.add(edge.parent)

        pos = interval[0]
        for parent_node in nodes_changed:
            child_edge_ids = edges_for_node[parent_node]
            if len(child_edge_ids) >= 3:
                # We have a previous polytomy to break
                parent_time = existing_node_time[parent_node]
                new_nodes = []
                child_ids = existing_edges_child[list(child_edge_ids)]
                remaining_edges = child_edge_ids.copy()
                left = None
                max_time = 0
                for edge_id, child_id in zip(child_edge_ids, child_ids):
                    max_time = max(max_time, existing_node_time[child_id])
                    if left is None:
                        left = existing_edges_left[edge_id]
                    else:
                        assert left == existing_edges_left[edge_id]
                    if existing_edges_right[edge_id] > interval[0]:
                        # make sure we carry on the edge after this polytomy
                        existing_edges_left[edge_id] = pos

                # ADD THE PREVIOUS EDGE SEGMENTS
                dt = min((parent_time - max_time)/(len(child_ids)*2), epsilon)
                # Each broken polytomy of degree N introduces N-2 extra nodes, each at a time
                # slighly less than the parent_time. Create new nodes in order of decreasing time
                new_nodes = [nodes_table.add_row(time=parent_time - (i * dt))
                             for i in range(1, len(child_ids) - 1)]
                # print("new_nodes:", new_nodes, [tables.nodes[n].time for n in new_nodes])
                for new_edge in resolve_polytomy(parent_node, child_ids, new_nodes, rng):
                    edges_table.add_row(left=left, right=pos, child=new_edge[0], parent=new_edge[1])
                    # print("new_edge: left={}, right={}, child={}, parent={}".format(
                    #     left, pos, new_edge[0], new_edge[1]))
            else:
                # Previous node was not a polytomy - just add the edges_out, with modified left
                for edge_id in child_edge_ids:
                    if existing_edges_right[edge_id] == pos:  # this edge has just gone out
                        edges_table.add_row(
                            left=existing_edges_left[edge_id],
                            right=pos,
                            parent=parent_node,
                            child=existing_edges_child[edge_id],
                        )

        for edge in e_out: 
            if edge.parent != tskit.NULL:
                # print("REMOVE", edge.id)
                edges_for_node[edge.parent].remove(edge.id)
        for edge in e_in:
            if edge.parent != tskit.NULL:
                # print("ADD", edge.id)
                edges_for_node[edge.parent].add(edge.id)            

        # Chop if we have created a polytomy: the polytomy itself will be resolved
        # at a future iteration, when any of the edges move in or out of the polytomy
        while nodes_changed:
            node = nodes_changed.pop()
            edge_ids = edges_for_node[node]
            # print("Looking at", node)

            if len(edge_ids) == 0:
                del edges_for_node[node]
            # if this node has changed *to* a polytomy, we need to cut all of the
            # child edges that were previously present by adding the previous segment
            # and left-truncating
            elif len(edge_ids) >= 3:
                # print("Polytomy at", node, " breaking edges")
                for edge_id in edge_ids:
                    if existing_edges_left[edge_id] < interval[0]:
                        tables.edges.add_row(
                            left=existing_edges_left[edge_id],
                            right=interval[0],
                            parent=existing_edges_parent[edge_id],
                            child=existing_edges_child[edge_id],
                        )
                    existing_edges_left[edge_id] = interval[0]
    assert len(edges_for_node) == 0

    tables.edges.squash()
    tables.sort() # Shouldn't need to do this: https://github.com/tskit-dev/tskit/issues/808

    return tables.tree_sequence()
```

The code can be tested using something like this:

```
import io
import collections
import tqdm
import time

import msprime
import tsinfer

### Check equiprobable

nodes_polytomy_4 = """\
id      is_sample   population      time
0       1       0               0.00000000000000
1       1       0               0.00000000000000
2       1       0               0.00000000000000
3       1       0               0.00000000000000
4       0       0               1.00000000000000
"""
edges_polytomy_4 = """\
id      left            right           parent  child
0       0.00000000      1.00000000      4       0,1,2,3
"""

poly_ts = tskit.load_text(
    nodes=io.StringIO(nodes_polytomy_4),
    edges=io.StringIO(edges_polytomy_4),
    strict=False,
)

trees = collections.Counter()
for seed in tqdm.trange(1, 100000):
    ts2 = resolve_polytomies(poly_ts, random_seed=seed)
    trees.update([ts2.first().rank()])
print(trees)

### Time on 10000 tip inferred TS and check

ts_old = msprime.simulate(10000, recombination_rate=100, mutation_rate=10, random_seed=123)
sd = tsinfer.SampleData.from_tree_sequence(ts_old, use_times=False)
ts = tsinfer.infer(sd)
print(f"{ts.num_samples} tips ; {ts.num_trees} trees")
start = time.time()
ts2 = resolve_polytomies(ts, random_seed=1)
print("Time (s):", time.time()-start)
for tree in ts2.trees():
    for node in tree.nodes():
        assert tree.num_children(node) < 3
```

hyanwong:
Fixes #809

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -976,3 +976,38 @@ class TestCountTopologies:
         ts = tables.tree_sequence()
         samples = ts.samples()
         self.verify_topologies(ts, sample_sets=[samples[:10], samples[10:]])
+
+
+class TestPolytomySplitting:
+    rng = np.random.default_rng(seed=123)
+
+    def test_randomly_resolve_polytomy(self):
+        relations = comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5, 6], self.rng)
+        assert len(relations) == 6
+        parents = [r[0] for r in relations]
+        children = [r[1] for r in relations]
+        assert set(parents) == set(range(4, 7))
+        assert set(children) == set(range(0, 7)) - {4}
+        # all internal nodes (including the root) are parents twice only
+        for internal_node in [4, 5, 6]:
+            assert parents.count(internal_node) == 2
+        # all nodes except root are children once only
+        for child_node in [0, 1, 2, 3, 5, 6]:
+            assert children.count(child_node) == 1
+
+    def test_noop(self):
+        relations = comb.randomly_resolve_polytomy(2, [0, 1], [], self.rng)
+        assert {tuple(r) for r in relations} == {(2, 0), (2, 1)}
+
+    def test_bad_intermediate_nodes(self):
+        for bad in ["Hi", None]:
+            with pytest.raises(TypeError):
+                comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], bad, self.rng)
+
+    def test_too_many_intermediate_nodes(self):
+        with pytest.raises(ValueError):
+            comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5, 6, 7], self.rng)
+
+    def test_too_few_intermediate_nodes(self):
+        with pytest.raises(ValueError):
+            comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5], self.rng)
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -26,6 +26,7 @@ Test cases for combinatorial algorithms.
 import collections
 import io
 import itertools
+import random
 
 import msprime
 import numpy as np
@@ -978,36 +979,30 @@ class TestCountTopologies:
         self.verify_topologies(ts, sample_sets=[samples[:10], samples[10:]])
 
 
-class TestPolytomySplitting:
-    rng = np.random.default_rng(seed=123)
+class TestTreeNode:
+    """
+    Tests for the TreeNode class used to build simple trees in memory.
+    """
 
-    def test_randomly_resolve_polytomy(self):
-        relations = comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5, 6], self.rng)
-        assert len(relations) == 6
-        parents = [r[0] for r in relations]
-        children = [r[1] for r in relations]
-        assert set(parents) == set(range(4, 7))
-        assert set(children) == set(range(0, 7)) - {4}
-        # all internal nodes (including the root) are parents twice only
-        for internal_node in [4, 5, 6]:
-            assert parents.count(internal_node) == 2
-        # all nodes except root are children once only
-        for child_node in [0, 1, 2, 3, 5, 6]:
-            assert children.count(child_node) == 1
+    @pytest.mark.parametrize("n", [2, 3, 5, 10])
+    def test_random_binary_tree(self, n):
+        # Note this doesn't check any statistical properties of the returned
+        # trees, just that a single instance returned in a valid binary tree.
+        rng = random.Random(32)
+        labels = range(n)
+        root = comb.TreeNode.random_binary_tree(labels, rng)
 
-    def test_noop(self):
-        relations = comb.randomly_resolve_polytomy(2, [0, 1], [], self.rng)
-        assert {tuple(r) for r in relations} == {(2, 0), (2, 1)}
-
-    def test_bad_intermediate_nodes(self):
-        for bad in ["Hi", None]:
-            with pytest.raises(TypeError):
-                comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], bad, self.rng)
-
-    def test_too_many_intermediate_nodes(self):
-        with pytest.raises(ValueError):
-            comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5, 6, 7], self.rng)
-
-    def test_too_few_intermediate_nodes(self):
-        with pytest.raises(ValueError):
-            comb.randomly_resolve_polytomy(4, [0, 1, 2, 3], [5], self.rng)
+        stack = [root]
+        num_nodes = 0
+        recovered_labels = []
+        while len(stack) > 0:
+            node = stack.pop()
+            num_nodes += 1
+            if node.label is not None:
+                assert len(node.children) == 0
+                recovered_labels.append(node.label)
+            for child in node.children:
+                assert child.parent == node
+                stack.append(child)
+        assert len(recovered_labels) == n
+        assert sorted(recovered_labels) == list(labels)
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -23,12 +23,14 @@
 """
 Test cases for the supported topological variations and operations.
 """
+import collections
 import functools
 import io
 import itertools
 import json
 import math
 import random
+import re
 import sys
 import unittest
 
@@ -7923,3 +7925,165 @@ class TestExampleTrees:
 
         assert ts.node(ts.first().root).time == branch_length
         assert ts.kc_distance(topological_equiv_ts) == 0
+
+
+class TestPolytomySplitting:
+    """
+    Test the ability to randomly split polytomies
+    """
+
+    # A complex ts with polytomies - in the first 2 trees the polytomy involves
+    # the same children, so should be resolved in one go
+    #
+    # 1.00┊    6      ┊      6    ┊       6   ┊           ┊      6    ┊
+    #     ┊ ┏━┳┻┳━┓   ┊   ┏━┳┻┳━┓ ┊    ┏━━╋━┓ ┊           ┊   ┏━┳┻┳━┓ ┊
+    # 0.50┊ 5 ┃ ┃ ┃   ┊   5 ┃ ┃ ┃ ┊    5  ┃ ┃ ┊      5    ┊   ┃ ┃ ┃ ┃ ┊
+    #     ┊ ┃ ┃ ┃ ┃ . ┊   ┃ ┃ ┃ ┃ ┊ . ┏┻┓ ┃ ┃ ┊ . ┏━┳┻┳━┓ ┊ . ┃ ┃ ┃ ┃ ┊
+    # 0.00┊ 0 2 3 4 1 ┊ 0 1 2 3 4 ┊ 0 1 2 3 4 ┊ 0 1 2 3 4 ┊ 0 1 2 3 4 ┊
+    #   0.00        0.20        0.40        0.60        0.80        1.00
+    nodes_polytomy_44344 = """\
+    id      is_sample   population      time
+    0       1           0               0.0
+    1       1           0               0.0
+    2       1           0               0.0
+    3       1           0               0.0
+    4       1           0               0.0
+    5       0           0               0.5
+    6       0           0               1.0
+    """
+    edges_polytomy_44344 = """\
+    id      left     right    parent  child
+    0       0.0      0.2      5       0
+    1       0.0      0.8      5       1
+    2       0.0      0.4      6       2
+    3       0.4      0.8      5       2
+    4       0.0      0.6      6       3,4
+    5       0.0      0.6      6       5
+    6       0.6      0.8      5       3,4
+    7       0.8      1.0      6       1,2,3,4
+    """
+
+    def ts_polytomy_4(self):
+        return tskit.Tree.generate_star(4).tree_sequence
+
+    def ts_polytomy_44344(self):
+        return tskit.load_text(
+            nodes=io.StringIO(self.nodes_polytomy_44344),
+            edges=io.StringIO(self.edges_polytomy_44344),
+            strict=False,
+        )
+
+    # NB: testing statistical properties is nontrivial in a unit test - we need to run
+    # repeats to check expectations, which can be slow
+    @pytest.mark.slow
+    def test_equiprobable(self):
+        n_tops_4 = 15  # 4-tomy has 15 poss. resolutions
+        n_tops_3 = 3  # 3-tomy has 3 poss. resolutions
+        # Simplify the example tree to remove isolated node 0
+        ts = self.ts_polytomy_44344().simplify(np.arange(1, 5), keep_unary=True)
+        assert ts.num_trees == 4
+
+        num_tree_topologies = [collections.Counter() for _ in range(ts.num_trees)]
+        start_seed = 123
+        n = 2700  # This may take some time, e.g. 5 secs
+        for seed in range(start_seed, start_seed + n):
+            ts_split = ts.split_polytomies(random_seed=seed * 10).simplify(
+                keep_unary=False
+            )
+            # Tree 0 should have been concatenated into tree 1 by simplification
+            for tree in ts_split.trees():
+                num_tree_topologies[tree.index].update([tree.rank()])
+        for i, n_topologies in enumerate([n_tops_4, n_tops_3, n_tops_4, n_tops_4]):
+            assert len(num_tree_topologies[i]) == n_topologies
+            # If equal probabilities, max difference
+            freqs = [c / n for c in num_tree_topologies[i].values()]
+            # all counts freqs should be roughly 1/nth of the trees
+            assert min(freqs) > 1 / n_topologies * 0.8
+            assert max(freqs) < 1 / n_topologies * 1.2
+
+    def test_simple_examples(self):
+        ts = self.ts_polytomy_4().split_polytomies(random_seed=12)
+        for tree in ts.trees():
+            for node in tree.nodes():
+                assert tree.num_children(node) < 3
+
+    def test_complex_examples(self):
+        ts = self.ts_polytomy_44344().split_polytomies(random_seed=12)
+        for tree in ts.trees():
+            for node in tree.nodes():
+                assert tree.num_children(node) < 3
+        # Tree 0 should have same resolution (same internal nodes) as tree 1
+        t = ts.at_index(0)
+        n0 = {n for n in t.nodes() if not (t.is_leaf(n) or t.parent(n) == tskit.NULL)}
+        t = ts.at_index(1)
+        n1 = {n for n in t.nodes() if not (t.is_leaf(n) or t.parent(n) == tskit.NULL)}
+        assert n0 == n1
+        # Tree -2 should not have same resolution as tree -1 (all internal nodes should
+        # differ) as the root has changed, even though the children are the same
+        t = ts.at_index(-2)
+        n0 = {n for n in t.nodes() if not (t.is_leaf(n) or t.parent(n) == tskit.NULL)}
+        t = ts.at_index(-1)
+        n1 = {n for n in t.nodes() if not (t.is_leaf(n) or t.parent(n) == tskit.NULL)}
+        assert len(n0 & n1) == 0
+
+    def test_nonbinary_simulation(self):
+        demographic_events = [
+            msprime.SimpleBottleneck(time=1.0, population=0, proportion=0.95)
+        ]
+        ts = msprime.simulate(
+            20,
+            recombination_rate=10,
+            mutation_rate=5,
+            demographic_events=demographic_events,
+            random_seed=7,
+        )
+        n_poly = 0
+        for e in ts.edgesets():
+            if len(e.children) > 2:
+                n_poly += 1
+        assert n_poly > 3
+        ts_binary = ts.split_polytomies(random_seed=11)
+        for tree in ts_binary.trees():
+            for node in tree.nodes():
+                assert tree.num_children(node) < 3
+
+    def test_bad_method(self):
+        with pytest.raises(ValueError, match="Method"):
+            self.ts_polytomy_4().split_polytomies(method="something_else")
+
+    def test_epsilon_for_nodes(self):
+        with pytest.raises(
+            ValueError, match="not small enough to create new nodes under node 4"
+        ) as exc_info:
+            self.ts_polytomy_4().split_polytomies(epsilon=1, random_seed=12)
+        m = re.search(r"must be < ([-\d.e]+).", exc_info.value.args[0])
+        assert m is not None
+        suggested_epsilon = float(m.group(1)) * 0.999
+        self.ts_polytomy_4().split_polytomies(epsilon=suggested_epsilon, random_seed=12)
+
+    def test_epsilon_for_mutations(self):
+        tables = self.ts_polytomy_4().dump_tables()
+        root_time = tables.nodes.time[-1]
+        assert root_time > 0.1
+        site = tables.sites.add_row(position=0.5, ancestral_state="0")
+        mut_diff = 0.01
+        tables.mutations.add_row(
+            site=site, time=root_time - mut_diff, node=0, derived_state="1"
+        )
+        ts = tables.tree_sequence()
+        with pytest.raises(
+            _tskit.LibraryError,
+            match="not small enough to create new nodes below a polytomy",
+        ):
+            ts.split_polytomies(epsilon=mut_diff * 2, random_seed=13)
+        # A 4-tomy creates 2 new nodes => epsilon of ~ 1/3 of mut_diff should work
+        self.ts_polytomy_4().split_polytomies(epsilon=mut_diff / 3, random_seed=13)
+
+    def test_provenance(self):
+        ts = self.ts_polytomy_4()
+        ts_split = ts.split_polytomies(random_seed=14)
+        record = json.loads(ts_split.provenance(ts_split.num_provenances - 1).record)
+        assert record["parameters"]["command"] == "split_polytomies"
+        ts_split = ts.split_polytomies(random_seed=12, record_provenance=False)
+        record = json.loads(ts_split.provenance(ts_split.num_provenances - 1).record)
+        assert record["parameters"]["command"] != "split_polytomies"
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -30,7 +30,6 @@ import itertools
 import json
 import math
 import random
-import re
 import sys
 import unittest
 
@@ -7927,13 +7926,24 @@ class TestExampleTrees:
         assert ts.kc_distance(topological_equiv_ts) == 0
 
 
+def num_leaf_labelled_binary_trees(n):
+    """
+    Returns the number of leaf labelled binary trees with n leaves.
+
+    TODO: this would probably be helpful to have in the combinatorics
+    module.
+
+    https://oeis.org/A005373/
+    """
+    return int(np.math.factorial(2 * n - 3) / (2 ** (n - 2) * np.math.factorial(n - 2)))
+
+
 class TestPolytomySplitting:
     """
     Test the ability to randomly split polytomies
     """
 
-    # A complex ts with polytomies - in the first 2 trees the polytomy involves
-    # the same children, so should be resolved in one go
+    # A complex ts with polytomies
     #
     # 1.00┊    6      ┊      6    ┊       6   ┊           ┊      6    ┊
     #     ┊ ┏━┳┻┳━┓   ┊   ┏━┳┻┳━┓ ┊    ┏━━╋━┓ ┊           ┊   ┏━┳┻┳━┓ ┊
@@ -7963,8 +7973,8 @@ class TestPolytomySplitting:
     7       0.8      1.0      6       1,2,3,4
     """
 
-    def ts_polytomy_4(self):
-        return tskit.Tree.generate_star(4).tree_sequence
+    def tree_polytomy_4(self):
+        return tskit.Tree.generate_star(4)
 
     def ts_polytomy_44344(self):
         return tskit.load_text(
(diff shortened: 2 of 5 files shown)


hyanwong:
Currently only a draft, as there are basically no tests, rather a long method name (but I feel we need the word "random" in there, since AFAIK this is the only routine in tskit which involves randomness), and I only have a horribly inefficient way to check the time of mutations above child nodes. I'm sure the latter could be optimised - suggestions for approaches are welcome.

benjeffery:
Great stuff - haven't reviewed the code yet, but one question:is where we get the RNG on the C side (if indeed we intend to implement this C-side).

hyanwong:
What RNG is used in msprime on the C side? If we do want a C side implementation, I assume we just copy that. Having said which, profiling the code suggests (surprisingly) that the majority of the time is spent in `table.add_row()`, so a C implementation would mainly be to allow C users to access the functionality.

benjeffery:
> What RNG is used in msprime on the C side? 
GSL is used, which we can't use here due to licensing restrictions (It's GPL were MIT), hence it being an issue. One idea we had for another method was to pass in a random array - but I'm guessing here we don't know how many random numbers will be needed? In the other use case it was known.

grahamgower:
PCG family of random generators could be an option: https://www.pcg-random.org/download.html

hyanwong:
> PCG family of random generators could be an option: https://www.pcg-random.org/download.html

Numpy [currently uses](https://numpy.org/doc/stable/reference/random/index.html?highlight=random#quick-start) PCG64 as its default PRNG, which is good enough for me!

python/tskit/tables.py
```diff
@@ -2781,3 +2782,204 @@ def union(
...
+        for interval, e_out, e_in in ts.edge_diffs(include_terminal=True):
+            pos = interval[0]
+            prev_tree = None if pos == 0 else next(tree_iter)
+
+            for edge in itertools.chain(e_out, e_in):
+                if edge.parent != tskit.NULL:
+                    nodes_changed.add(edge.parent)
+
+            if check_mutations and prev_tree is not None:
+                # This is grim. There must be a more efficient way.
```
hyanwong, line 2885: @brianzhang01 - can you think of a nicer way to find the oldest mutation above a node?

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/815/)

python/tskit/tables.py
hyanwong: I just thought - we can probably do this by pre-sorting a copy of the mutations - sorting first by node, then by time (if there are no mutation times, then we don't need to do this anyway). The slight problem is that we might have mutations that are oldest above a node, but not on the current tree. We should be able to mask those out, however, by taking the list of mutations for this node and restricting to only the sites in this tree.

hyanwong:
@benjeffery - I've revised this after discussion with Jerome, so that the code is placed in `combinatorics.py` and doesn't clog up `trees.py`. I've also removed the ability to modify the epsilon value, and now we just error out if epsilon is too large. There's a helpful error message if this is because of node times, and a less helpful one (at the end of the function) which traps messages about mutation times and raises an error mentioning epsilon. I'm not sure this is the right way to do it, but it avoids having to explicitly check for mutation times, which was a pain in the previous version. The RNG is the one from numpy, so porting this to C might be less obvious. The only "method" is "random" for the moment, but I think it's worth having that argument in as it emphasises the fact that by default there is a stochastic component to this algorithm.

I've added Jerome as a reviewer, as much of this was hist suggestion for reworking, but I don't know how much he wants to check over it all?

benjeffery:
I've just started reviewing.

python/tests/test_topology.py
```diff
@@ -7951,3 +7953,163 @@ def test_star_branch_length(self):
...
+    def ts_polytomy_44344(self):
+        return tskit.load_text(
+            nodes=io.StringIO(self.nodes_polytomy_44344),
+            edges=io.StringIO(self.edges_polytomy_44344),
+            strict=False,
+        )
+
+    # Statistical test coded here for reference
+    # @unittest.skip("Testing statistical properties is nontrivial in a unit test")
+    def test_equiprobable(self):
```
jeromekelleher, line 8006: I don't think we can do this as a unit test as it'll be too slow and cause problems. We'll need some sort of validation suite or something. I'm not sure what to do with the code for now though.

@benjeffery, any thoughts?

jeromekelleher (commented):
It basically looks good, although I haven't really grokked the algorithm. This is quite a tricky thing, which I would imagine needs quite a lot of validation and testing to make sure it has the right properties in all situations.

My inclination would be to merge this ASAP but "undocument" it ahead of the upcoming release. That way, we can use it internally for our own projects without committing ourselves to any particular API or semantics, and also give us a chance to consider how we'll validate stuff like this which we can't just do simple unit tests on.

python/tskit/combinatorics.py
```diff
@@ -720,6 +721,208 @@ def is_symmetrical(self):
...
+    ):
+        for edge in itertools.chain(e_out, e_in):
+            if edge.parent != tskit.NULL:
+                nodes_changed.add(edge.parent)
+        pos = interval[0]
+        for parent in nodes_changed:
+            if len(edges_from_node[parent]) >= 3:
+                child_edges = list(edges_from_node[parent])
+                # We have a previous polytomy to break
+                parent_time = tables.nodes.time[parent]
```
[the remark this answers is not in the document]
hyanwong: Ah - I wondered if accessing the attribute was costly (I had a variable like this before, but took it out as it seemed redundant). I'll put it back in, thanks.

python/tskit/combinatorics.py
```diff
@@ -720,6 +721,208 @@ def is_symmetrical(self):
...
+    allowed_methods = ["random"]
+    if epsilon is None:
+        epsilon = 1e-10
+    if method is None:
+        method = "random"
+    if method not in allowed_methods:
+        raise ValueError(f"method must be chosen from {allowed_methods}")
+    rng = np.random.default_rng(seed=random_seed)
+    tables = ts.dump_tables()
+    # Store existing table data before clearing
```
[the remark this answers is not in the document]
hyanwong: If I create `existing_edges = tables.edges.copy()`, is it still worth creating `existing_edges_right = existing_edges.right` for use in an inner loop? Or is accessing `existing_edges.right` just as performant?

python/tests/test_topology.py
hyanwong: For the moment I can uncomment the skip and say "To be put in validation suite". How is this sort of simulation testing done in msprime?

benjeffery (commented):
I've not dug in deep to the algo - but here are a few comments.

python/tests/test_topology.py
hyanwong: I didn't know about `@pytest.mark.slow` - seems like a good compromise.

python/tskit/combinatorics.py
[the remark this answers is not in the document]
jeromekelleher: Currently ``existing_edges.right`` does create a full copy of the array data, so if it's used in an inner loop it's worth making the array all right.

hyanwong:
I've made the review changes suggested by @benjeffery and @jeromekelleher . Do you want me to undocument this too (how? I could put the documentation into the `combinatorics.split_polytomy()` function, which I don't think currently gets exported to the docs).

benjeffery:
What's been the process before? Do we wait for the C port to document? Seems it is still useful in current form - maybe with a warning about perf.

hyanwong:
> What's been the process before? Do we wait for the C port to document? Seems it is still useful in current form - maybe with a warning about perf.

I don't think it's the performance that Jerome is worried about. I think it's the complexity of the algorithm. He doesn't want to document it without testing more extreme cases, I would guess.

Having said which, I have been running this algorithm on a variety of large simulated and inferred tree sequences without issue so far, although I haven't been doing explicit statistical tests for equiprobability on the resolved trees.

jeromekelleher:
> What's been the process before? Do we wait for the C port to document? Seems it is still useful in current form - maybe with a warning about perf.

It's not so much about perf as semantics - I'd need to see quite a lot more testing of how this works on various tree sequence topologies to be assured that it's actually dong the right thing in all cases. Things like nodes sequentially jumping out of a large polytomy tree-by-tree - how would this be resolved? How *should* it be resolved? Maybe we're happy that marginally we're resolving individual trees such that we're sampling from the space of binary trees, but it's not obvious that what we're doing is the correct thing across *all* trees in the sequence. What does it really mean, anyway, to resolve a polytomy that spans multiple trees, but has different nodes in each. This is not a straightforward problem and it needs real research to solve it, and I don't want to give the impression that we've *done* this research and that people should trust the code to do the right thing in all cases (which, is the goal: people should trust tskit).

So, this discussion makes me wonder: what if instead of trying to randomly resolve a whole tree sequence, we just did it tree-by-tree instead? I.e., the Tree class had a method ``resolve_polytomies`` that returned a Tree with the same interval, with all the poloytomies broken. The underlying tree sequence would consist of the same set of nodes, and would only have edges spanning the interval in question (so, if the tree is in the middle of the sequence, it will have missing data on both sides of it).

This would get rid of the tricky semantic questions, and make this much easier to test and validate. I'm pretty sure this fulfils your use case @hyanwong, since you're doing pretty non-standard stuff with the tree metrics, anyway (although it means you have to coiterate tree-by-tree, but that's OK: we could promote [this function](https://github.com/tskit-dev/tskit/blob/main/python/tests/tsutil.py#L1287) to be part of the public API. The performance impact of doing KC on a set of trees vs tree sequences would be entirely swamped by the cost of resolving the trees, anyway.

benjeffery:
@jeromekelleher Just a note for future in case this happens to someone else. Yan was not getting coverage reports due to following the repo on his personal circle CI account: https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-

hyanwong:
> What does it really mean, anyway, to resolve a polytomy that spans multiple trees, but has different nodes in each. 

That should never happen. As soon as any of the nodes underneath a polytomy change (or the nod itself changes), we do a brand-new random resolution.

jeromekelleher:
> That should never happen. As soon as any of the nodes underneath a polytomy change (or the node itself changes), we do a brand-new random resolution. The complexity in the algorithm is simply spotting whether the edges under a polytomy have changed or not, when we move to a new tree (as the change could be in a completely different part of the tree)

Then why bother trying to do this at the tree sequence level? It seems like this would a lot simpler if we just regarded it on a marginal tree basis.

hyanwong:
> Then why bother trying to do this at the tree sequence level? It seems like this would a lot simpler if we just regarded it on a marginal tree basis.

Because we have very large number of samples, and hence a change to a new tree will probably not involve any changes to a polytomous node. I expect many of these polytomies to span a large number of trees without changing their children at all (the tree change could be lower down in the subtree, or in a different subtree). If we go tree-by-tree we re-break these every time.

python/tskit/combinatorics.py
```diff
@@ -720,6 +721,207 @@ def is_symmetrical(self):
...
+    existing_edges_child = tables.edges.child.copy()  # Read only copy for efficiency
+    node_time = tables.nodes.time
+    tables.edges.clear()
+
+    edges_from_node = collections.defaultdict(set)  # Active descendant edge ids
+    nodes_changed = set()
+
+    for interval, e_out, e_in in ts.edge_diffs(include_terminal=True):
+        for edge in itertools.chain(e_out, e_in):
+            if edge.parent != tskit.NULL:
```
jeromekelleher, line 829: ``edge.parent`` can't be NULL if these are unmodified edges from the tree sequences

Same below

hyanwong: Oh, good point. I always think that the root has a edge with a null parent, but that's not how we do it.

jeromekelleher:
I'm nervous about his code because it's a complicated algorithm that I don't understand, that we've only tested on a handful of topologies, haven't extensively validated and proved no theoretical properties about. The fact that we've jumped over the simpler case of doing this tree-by-tree makes me more nervous still.

Surely it makes sense to start with the simpler case of single-tree random resolution and then *if* this doesn't do what's needed we extend to the tree sequence case? We have more pressing things to do than worry about the precise details of the splitting semantics in tree sequences  - but if you want to add functions to the library then that's precisely what we have to do.

hyanwong:
Well, I'm mainly trying to get this in so that I can test the inferred trees from tsinfer. Why not include it but undocumented it, as you previously suggested? It seems like quite a lot more coding work to me to implement a Tree.split_polytomy() method, and applying such a thing treewise will also mean that operations on the resulting ts might be quite a lot slower, as there could be loads more edges, and it's already slow calculating the KC on the trees.

If it's not in tskit, then I guess I'll just include the code as-is in the tsinfer / tsdate testing codebase, but the reason for including this that I had hoped to slim that down a bit.

hyanwong:
I've just been thinking about how this would be done on a tree-by-tree basis, @jeromekelleher, and I don't think the algorithm would be much simpler, would it? You would loop over the trees, and every time you came across a node with >2 children, you would "cut" the edges leading to that node into 3 parts, one before the tree, one over the tree span, and another after the current tree. So you would either have to co-iterate over the edge diffs and the trees, and build up a set of new edges as you go, or do somewhat like I'm doing now and iterate over only the edge diffs and keep track of the number of children for each node using the diffs. It's all going to involve keeping track of currently used edges, in a slightly convoluted way. I think the algorithm would end up being almost as intricate.

Alternatively, you could simply cut up *all* the trees into separate edges, split the polytomies, and squash them together again, but for large trees this would be very painful, wouldn't it?

I feel like I'm missing an obvious way to do this, though. Do you or @benjeffery have any better suggestions?

jeromekelleher:
I was suggesting adding a ``Tree.split_polytomies`` method that just returns one tree:

> I.e., the Tree class had a method resolve_polytomies that returned a Tree with the same interval, with all the poloytomies broken. The underlying tree sequence would consist of the same set of nodes, and would only have edges spanning the interval in question (so, if the tree is in the middle of the sequence, it will have missing data on both sides of it). 

So, there's only one "real" tree in the underlying tree sequence (or ts.num_trees is at most 3). This is simpler because it completely avoids having to worry about the global semantics of the splitting operations, you just think about the local tree. That's something that's pretty well established, we understand how to do and we can easily document.

For your application, I guess you'd do something like
```python
for interval, source_tree,  inferred_tree in tskit.coiterate(source_ts, inferred_ts):
      distances = np.array(num_random_resolutions)
      for j in range(num_random_resolutions):
          resolved_tree = inferred_tree.resolve_polytomies()
          distances[j] = source_tree.kc_distanc(resolved_tree)
     mean_distance = np.mean(distances)
     # Now weight by either the interval's span or the number of sites in the interval and add to the 
     # total
```

This is nice and simple, and easy to describe.

hyanwong:
Oh, I see. So I never get a bifurcating tree sequence out of it at all. You are right, that would be easier. But then I wouldn't be able to store the resulting ts at all, unless I output all the trees as newick, or something. It's also convenient to be able to re-iterate over the split tree sequence a few times, to run various other tests on the bifurcating trees (e.g. using the dendropy  or R libraries library to get additional distance metrics, etc.).

If it's turning out to be a pain getting this in, then I guess I'm just going to stick with packaging it separately, TBH. I feel it would be too much work figuring out how to construct the individual trees when I have something that I'm happy with at the moment.

jeromekelleher:
@hyanwong - see my proposal over in #1013. Having done some reading on this (there's a significant combinatorial literature on it), it's not obvious to me that your algorithm is correct either - what are you basing your approach on?

hyanwong:
Thanks @jeromekelleher . The algorithm is the one used in Dendropy (now corrected, see e.g. https://github.com/jeetsukumaran/DendroPy/commit/7131bf0d2b3623667934581897d3cdf42cdadb65), I've discussed it in more detail with Mark Holder, and also (briefly, as an aside) in a email discussion with Emmanual Paradis, the author of the Ape phylogenetic package which provides this functionality in R. I can paste the correspondence below if you like. I'm sure it's correct.

hyanwong:
p.s. I've just noticed that the algorithm I use is that labelled 2.1 in Furnas (1984): https://link.springer.com/article/10.1007/BF01890123

jeromekelleher:
Fair enough, but the random binary tree algorithm is still buried heavily in implementation detail. I think the approach in #1013 is a lot easier to follow and verify.

So, my preferred approach here would be to use the single tree approach defined in #1013, combined with the tests and UI refinements you have here. We can extend to a "tree sequence" resolution in a follow up if that turns out to be necessary. Is this OK? If you agree I'll add some commits onto this PR to do this, so that the tree sequence approach is still in the git history, should we need it.

hyanwong:
OK. I think it's worth testing the algorithm for equiprobability, though: as you say, translating these correctly isn't always easy. My version has the benefit that you can allocate all the random numbers you need in one go, which is handy for python, but not an issue for C. I think Furnas (1984) shows that it has nice scaling properties (which is presumably why it was picked for other phylogenetic programs), but that might be true of Remy's method too (do you have a citation for this in the phylogenetic literature, as not everyone has access to TAOCP?).

jeromekelleher:
The reference is [here](http://www.cs.utsa.edu/~wagner/knuth/fasc4a.pdf), page 16. It's a linear time algorithm (I don't see how one can do much better?) and has been proved to sample uniformly from the space of labelled binary trees.

jeromekelleher:
I've updated this @hyanwong, can you take a look please?

I think it's mostly ready to go, but I think we should probably change the semantics of ``epsilon`` to do the following. If it is not specified, we space the node times out evenly along the branches of the randomly generated binary tree, such that epsilon we space out the nodes more evenly. We can easily do this now, and it will do what people want most of the time, I'd imagine.

If you really want a bunch of super-short branch lengths then you can specify epsilon manually.

hyanwong:
> I've updated this @hyanwong, can you take a look please?

Will do. I haven't followed the logic exactly yet, through.

> I think it's mostly ready to go, but I think we should probably change the semantics of `epsilon` to do the following. If it is not specified, we space the node times out evenly along the branches of the randomly generated binary tree, such that epsilon we space out the nodes more evenly. We can easily do this now, and it will do what people want most of the time, I'd imagine.

I started out thinking about an adaptive epsilon like this, but it's hard to do because of mutation times, I think? There's no easy way to get the time of the oldest mutation above a node, without actually just going through the mutations one by one. I think we said that would probably not be too much of a hit, but it seemed easier to just require epsilon to be set manually.

jeromekelleher:
> I started out thinking about an adaptive epsilon like this, but it's hard to do because of mutation times, I think? There's no easy way to get the time of the oldest mutation above a node, without actually just going through the mutations one by one. I think we said that would probably not be too much of a hit, but it seemed easier to just require epsilon to be set manually.


It's easy enough to get the sites and mutations - we just iterate through them all for the tree. We're already doing a O(n) traversal of the nodes, so there's no point in worrying about efficiency there.

python/tskit/combinatorics.py
```diff
@@ -28,12 +28,133 @@
...
+
+    @staticmethod
+    def random_binary_tree(leaf_labels, rng):
+        """
+        Based on the description of Remy's method of generating "decorated"
+        random binary trees in TAOCP 7.2.1.6. This is not a direct
+        implementation of Algorithm R, because we are interested in
+        the leaf node labellings.
+
+        The pre-fasicle text is available here, page 16:
```
hyanwong, line 58: I've never heard of a fasicle?! Looks like it has an extra c in the spelling anyway!

hyanwong:
> > I started out thinking about an adaptive epsilon like this, but it's hard to do because of mutation times, I think? There's no easy way to get the time of the oldest mutation above a node, without actually just going through the mutations one by one. I think we said that would probably not be too much of a hit, but it seemed easier to just require epsilon to be set manually.
> 
> It's easy enough to get the sites and mutations - we just iterate through them all for the tree. We're already doing a O(n) traversal of the nodes, so there's no point in worrying about efficiency there.

OK, so we'll presumably have to add that logic and space the nodes out evenly between the oldest mutation above a child, or, if there are no mutation times, the oldest child node time?

python/tskit/combinatorics.py
```diff
@@ -28,12 +28,133 @@
...
+        parameters = {"command": "split_polytomies"}
+        tables.provenances.add_row(
+            record=json.dumps(tskit.provenance.get_provenance_dict(parameters))
+        )
+    try:
+        tables.sort()
+        ts = tables.tree_sequence()
+    except tskit.LibraryError as e:
+        msg = str(e)
+        if msg.startswith(
```
hyanwong, line 141: If we are checking on the time of children or mutations above children, then I think it would be better to raise these errors when we know we are going to violate the time constraints, as then we can say which nodes or mutations are causing the problems, which is much more helpful to the user.

I didn't do this for mutation times because I removed the code for checking mutation times, but if we are going to reinstate it, we should raise the error there.

brianzhang01 (commented):
I had these comments queued up from a review I started but didn't finish. Hope the comments still show up (I don't even know what they are at this moment, haven't figured out how to view them).

python/tskit/combinatorics.py
```diff
@@ -28,12 +28,133 @@
...
+        """
+        nodes = [TreeNode(label=leaf_labels[0])]
+        for label in leaf_labels[1:]:
+            n = TreeNode(label=label)
+            x = rng.choice(nodes)
+            internal = TreeNode(parent=x.parent, children=[x, n])
+            if x.parent is not None:
+                index = x.parent.children.index(x)
+                x.parent.children[index] = internal
+            rng.shuffle(internal.children)
```
hyanwong, line 69: I must admit that I don't quire understand why this extra shuffle is in there? Won't the first `rng.choice` make the selection random anyway?

hyanwong: (by the way, I think this is essentially the same edge-addition algorithm that I originally implemented)

brianzhang01:
Having read through the discussion...
- I am partial towards Jerome's suggestion to resolve individual trees on-the-fly. It's simpler and the KC metric operates tree-wise and then aggregates, so it would make for conceptually simpler KC code as well. The maintenance burden would be lower as more people would understand the code. Examples of maintenance include non-random resolution and different distributions of randomly resolving trees (e.g. rtree vs. rcoal resolution in ape: https://www.rdocumentation.org/packages/ape/versions/5.4-1/topics/rtree, current behavior is rtree)
- It seems like Yan's function would still have a nice place to live: tsinfer. Besides metrics, what is nice about having and serializing a tree sequence with polytomies resolved?
- The coiterate function in tsutil looks useful for metrics purposes
- For what it's worth, I currently have a KC with random resolution of polytomies that more-or-less resolves trees-on-the-fly and computes the KC between two trees at a particular position. Then I sample many positions and compute a Monte Carlo-style average. It's written in my own C++ codebase. The point here is that to cut down on the time of computing a tree-sequence-wide KC, this might be an option, but you'd want to measure the effects of the Monte Carlo approximation.

hyanwong:
> Having read through the discussion...
> 
> * I am partial towards Jerome's suggestion to resolve individual trees on-the-fly. It's simpler and the KC metric operates tree-wise and then aggregates, so it would make for conceptually simpler KC code as well. 

But I don't that's what the KC metric on a tree sequence does, @brianzhang01. Daniel coded up an an incremental algorithm, didn't he? I'm afraid that I still think that it's a bit nuts to re-resolve the polytomy every time anything anywhere on the tree changes: it doesn't feel like a natural thing to do, because the tree boundaries are really nothing to do with the polytomy itself. I suspect that it also won't scale to large trees, where pretty much every base could cause a tree change *somewhere* in the TS, probably miles away from the polytomy you are trying to resolve?

python/tests/test_combinatorics.py
```diff
@@ -976,3 +977,30 @@ def test_forward_time_wright_fisher_simplified(self):
...
         self.verify_topologies(ts, sample_sets=[samples[:10], samples[10:]])
+
+
+class TestTreeNode:
+    """
+    Tests for the TreeNode class used to build simple trees in memory.
+    """
+
+    @pytest.mark.parametrize("n", [2, 3, 5, 10])
+    def test_random_binary_tree(self, n):
```
brianzhang01, line 988: Maybe add a comment that we are checking for a valid binary tree on n leaves, and not checking anything about the randomness

python/tskit/combinatorics.py
```diff
@@ -28,12 +28,133 @@
...
+    """
+    Simple linked tree class used to generate tree topologies.
+    """
+
+    parent = attr.ib(default=None)
+    children = attr.ib(factory=list)
+    label = attr.ib(default=None)
+
+    @staticmethod
+    def random_binary_tree(leaf_labels, rng):
```
brianzhang01, line 51: Should we document the arguments, leaf_labels and rng?

python/tskit/combinatorics.py
```diff
@@ -28,12 +28,133 @@
...
+    allowed_methods = ["random"]
+    if epsilon is None:
+        epsilon = 1e-10
+    if method is None:
+        method = "random"
+    if method not in allowed_methods:
+        raise ValueError(f"Method must be chosen from {allowed_methods}")
+
+    tables = tree.tree_sequence.dump_tables()
+    tables.keep_intervals([tree.interval], simplify=False)
```
brianzhang01, line 105: This seems not ideal for a long tree sequence, but I guess there's no easy way around it? Instead of O(E) where E is the total number of edges, seems like there should be an O(K + log(E)) algorithm where K is the number of edges to keep, via stabbing queries? I'm not that familiar with the tskit backend though.

python/tskit/trees.py
```diff
@@ -2370,6 +2370,67 @@ def kc_distance(self, other, lambda_=0.0):
...
+        or "polytomy" --- is resolved into successive bifurcations. New nodes are
+        inserted at times fractionally less than than the time of node ``u``
+        (controlled by the ``epsilon`` parameter).
+
+        If the ``method`` is ``"random"`` (currently the only option, and the default
+        when no method is specified), then for a node with :math:`n` children, the
+        :math:`(2n - 3)!!` possible bifurcating topologies are produced with equal
+        probability.
+
+        .. fixme:: Looks like this function is wrong
```
brianzhang01, line 2393: ?

python/tskit/trees.py
```diff
@@ -2370,6 +2370,67 @@ def kc_distance(self, other, lambda_=0.0):
...
+
+        The returned :class`.Tree` will have the same genomic span as this tree,
+        and node IDs will be converved (that is, node ``u`` in this tree will
```
brianzhang01, line 2396: converted

brianzhang01: The idea of node IDs being converted isn't obvious to me, could use another sentence of documentation. So if I had leaves 0, 1, 2, and a polytomy 3 that is the parent of all three leaves, then in the returned tree I would have 0, 1, 2, 3, 4? Where 3 is the "converted" node ID and 4 is an extra node from breaking the polytomy?

python/tskit/trees.py
```diff
@@ -2370,6 +2370,67 @@ def kc_distance(self, other, lambda_=0.0):
...
+        tree sequence that contains only one non-degenerate tree, that is, where
+        edges cover only the interval spanned by this tree.
+
+        .. note::
+            A tree sequence :ref:`requires<sec_valid_tree_sequence_requirements>` that
+            parents be older than children and that mutations are younger than the
+            parent of the edge on which they lie. If ``epsilon`` is not small enough,
+            compared to the distance between a polytomy and its oldest child (or oldest
+            child mutation) these requirements may not be met. In this case an error is
+            raised, recommending a smaller epsilon value be used.
```
brianzhang01, line 2407: An alternative to epsilon would be to use https://numpy.org/doc/stable/reference/generated/numpy.nextafter.html. That should give finer resolution than trying to add a small floating point value.

hyanwong: I didn't know about nextafter. This is nice, but I'm cautious about using it, because we have been bitten in the past by creating nodes fractionally smaller than something, and then not being able to wedge in extra stuff later on (e.g. mutations). I can't see why you might want to do this in this case, but I'm still wary. Anyway, Jerome's suggestion of defaulting to an even spread of times seems good to me.

brianzhang01:
> But I don't that's what the KC metric on a tree sequence does, @brianzhang01. Daniel coded up an an incremental algorithm, didn't he? I'm afraid that I still think that it's a bit nuts to re-resolve the polytomy every time anything anywhere on the tree changes: it doesn't feel like a natural thing to do, because the tree boundaries are really nothing to do with the polytomy itself. I suspect that it also won't scale to large trees, where pretty much every base could cause a tree change _somewhere_ in the TS, probably miles away from the polytomy you are trying to resolve?

I see, I just skimmed #548 a bit. I guess I meant that the clearest mathematical definition of tree-sequence KC is as an average over the tree. But OK, if you want to use Daniel's function with polytomy breaking then I guess you would need a resolved tree-sequence.

> Yes, I didn't know about this. It's great! Should it be extended to multiple tree sequences, do you think?

I can't think of a clear use case off the top of my head.

python/tskit/combinatorics.py
jeromekelleher: We choose a node x to modify randomly in the first step, adding a new internal node which will be the parent of x and the n'th leaf. We must randomise the order of [x, n] because the leaf would aways be either the left or right child, which would (I think) mean we don't generate all possible trees.

I'll add a few comments to clarify, as it's a simple algorithm and worth explaining for future reference.

hyanwong: I think it *does* generate all possible trees, even if the child order isn't shuffled. At least, it still did in my tests. And from reading the Knuth explanation, it doesn't seem necessary (I think?)

python/tskit/trees.py
jeromekelleher: What's the correct number here? (2n - 3)!! is clearly very wrong indeed, and (2n - 3)! is just wrong.

hyanwong: Good point, sorry. Don't know how that snuck in. https://www.jstor.org/stable/2412810?seq=2 : 
Screenshot 2020-11-19 at 23 25 37

jeromekelleher: Ah, [here](https://oeis.org/A005373/a005373.pdf) it is. We're missing a bunch of terms..

hyanwong: I beat you to it :)

brianzhang01: I think these two expressions are the same, actually, by the conventions of double factorial. https://en.wikipedia.org/wiki/Double_factorial

hyanwong: Oh yes. I must have looked that up and forgotten I had done so! Thanks @brianzhang01 . The double factorial thing is nice and concise.

jeromekelleher:
Thanks for the comments @hyanwong and @brianzhang01, this is all very helpful. I'll have to leave this for a few days, so will respond to your individual comments when I can get backt o it.

However, I'm pretty solidly against the idea of doing resolutions tree-sequence-wide now as i'm *sure* there's a bunch of complexities we won't have considered, and we don't need it right now.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-815/)

python/tests/test_combinatorics.py
jeromekelleher: Done, good call.

python/tskit/combinatorics.py
jeromekelleher: Done. I've not done the formal sphinx params as this isn't part of the public API.

python/tskit/combinatorics.py
jeromekelleher: This is a great point @brianzhang01. Currently this operation is O(E) as you say, and it's one of the things we could make more efficient if we had implemented stabbing queries. See #684 and other [related issues](https://github.com/tskit-dev/tskit/projects/5).

For this application, I think the current approach is fine, but we would like to speed it up at some point.

python/tskit/trees.py
jeromekelleher: I agree with Yan here - if folks want to have a super-fine epsilon then they can provide it as a parameter, but as a default nextafter is a bit tricky. In practise 1e-10 is a reasonable default, I think, leaving enough space in between nodes for mutations (if they are needed) while still being very small.

I've changed my mind on the auto-spacing after thinking more about it @hyanwong - it'll be quite hard to test, and actually you probably *do* want super short branches on the randomly resolved polytomies.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-815/)

jeromekelleher:
This is ready to go, I think. Here's some evaluation code, for the record.

```python
import tskit
import io

import numpy as np
import seaborn as sns
import pandas as pd


def num_leaf_labelled_binary_trees(n):
    """
    Returns the number of leaf labelled binary trees with n leaves.

    https://oeis.org/A005373/
    """
    return int(np.math.factorial(2 * n - 3) / (2 ** (n - 2) * np.math.factorial(n - 2)))


for n in range(2, 7):
    N = num_leaf_labelled_binary_trees(n)
    print(f"Running n = {n}, {N} topologies")
    ranks = []
    tree = tskit.Tree.generate_star(n)
    for seed in range(1000 * N):
        split_tree = tree.split_polytomies()
        ranks.append(split_tree.rank())
    df = pd.DataFrame({"rank": ranks})
    assert len(df["rank"].unique()) == N
    ax = sns.countplot(x="rank", data=df)
    plt.savefig(f"n={n}.png")
    plt.clf()
```

n=3

n=4

n=5

jeromekelleher (approved):

hyanwong:
Thanks @jeromekelleher - did you remove the extra shuffle? You can probably use your tests above to convince yourself that it isn't needed.

jeromekelleher:
> Thanks @jeromekelleher - did you remove the extra shuffle? You can probably use your tests above to convince yourself that it isn't needed.

No, I left it in despite not being able to see any difference in having it from the tests. According to Knuth, there are 4n - 2 ways of creating a binary tree with n labelled leaves from one with n - 1. But, we're only choosing from 2n - 1 when we select the focal node x, so it seems to me we have to have the shuffle to get the right number.

As I say, I couldn't see any difference in the tests, but maybe there's something subtle in the leaf labellings that we're missing - we're only ever testing on the ordered labels, so maybe something would show up if we spent more time on that. But, there didn't seem much point to me, since the shuffle can't do any harm and there's not actually any good reason from the algorithm description's perspective for leaving it out.

hyanwong:
Sorry, just picking up a comment I made above @jeromekelleher - if possible it's *really* helpful to give some feedback about (a) what epsilon value to use, and (b) where the problem is when epsilon is too large. For node times (but not mutation times) we should be able to do that during the split algorithm, rather than right at the end of the function, when we have lost the information about where the problem is.

jeromekelleher:
Can you open an issue to track this please @hyanwong?

hyanwong:
Just to follow up. Switching to splitting polytomies per tree rather than per polytomous node now means that splitting polytomies is now taking longer than the inference process itself (and with better tree sequence, I suspect it will end up taking up more and more of the time, so I am still pretty keen on considering a tree-sequence-wide polytomy resolution algorithm.

hyanwong:
> Can you open an issue to track this please @hyanwong?

Opened at #1088

PR #822, opened 2020-08-31
check equality for missing times in mutation table rows

### LINKED ISSUE #821: mutation row with missing time not equal to itself
For instance:
```
t = tskit.TableCollection(sequence_length=10)
t.sites.add_row(position=1, ancestral_state='0')
t.mutations.add_row(site=0, node=0, derived_state='1', time=tskit.UNKNOWN_TIME)
assert(t.mutations[0] == t.mutations[0])
```

Arguably, this is correct, because NAN != NAN, but I think this isn't what we want.

petrelharp:
Closes #821, although not in the most elegant way.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -197,7 +197,9 @@ To include the changes that the hooks made, ``git add`` any
 files that were modified and run ``git commit`` (or, use ``git commit -a``
 to commit all changed files.)
 
-If you would like to run the checks without committing, use ``pre-commit run``.
+If you would like to run the checks without committing, use ``pre-commit run``
+(but, note that this will *only* check changes that have been *staged*;
+do ``pre-commit run --all`` to check unstaged changes as well).
 To bypass the checks (to save or get feedback on work-in-progress) use
 ``git commit --no-verify``
 
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -857,6 +857,57 @@ class TestIndividualTable(unittest.TestCase, CommonTestsMixin, MetadataTestsMixi
         self.assertEqual(list(t[0].location), [0])
         self.assertEqual(list(t[1].location), [1, 2, 3])
 
+    def test_missing_time_equal_to_self(self):
+        t = tskit.TableCollection(sequence_length=10)
+        t.sites.add_row(position=1, ancestral_state="0")
+        t.mutations.add_row(site=0, node=0, derived_state="1", time=tskit.UNKNOWN_TIME)
+        self.assertEqual(t.mutations[0], t.mutations[0])
+
+    def test_various_not_equals(self):
+        args = {
+            "site": 0,
+            "node": 0,
+            "derived_state": "a",
+            "parent": 0,
+            "metadata": b"abc",
+            "time": 0,
+        }
+        a = tskit.MutationTableRow(**args)
+        self.assertNotEqual(a, [])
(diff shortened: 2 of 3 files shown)


petrelharp:
Also a small clarifying note about pre-commit.

... which is not working for me, for some reason: running `pre-commit run` does everything as expected, but then `git commit` fails with a bunch of flake8 errors beginning with
```
python/tskit/metadata.py:161:13: F541 f-string is missing placeholders
python/tskit/util.py:111:26: E203 whitespace before ':'
python/tskit/util.py:129:31: E203 whitespace before ':'
python/tskit/util.py:189:23: E203 whitespace before ':'
python/tskit/util.py:207:36: E203 whitespace before ':'
```
The only way to get it to commit is to use `--no-verify`. I've uninstalled/reinstalled the pre-commit hooks, my flake8 is up to date, I've looked for other hooks... no idea what's going on. Any ideas?

python/tests/test_tables.py
```diff
@@ -1061,9 +1061,17 @@ def test_simple_example(self):
         t = tskit.MutationTable()
```
jeromekelleher, line 1061: Looks good, thanks @petrelharp. It'd be nice to add a quick test to check this in isolation, since it's subtle enough to miss in the first place. Something like ``test_missing_time_equal_to_self`` and just using the example you had in the issue.

jeromekelleher (commented):
LGTM, thanks @petrelharp. I think we should wait on @benjeffery before merging though (he'll be back in a few days).

python/tskit/tables.py
```diff
@@ -112,6 +112,21 @@ class MutationTableRow:
     metadata: bytes
     time: float
 
+    def __eq__(self, other):
+        return (
```
jeromekelleher, line 116: Looks like you need an ``isinstance(other, MutationTableRow)`` here and return False otherwise.

docs/development.rst
```diff
@@ -197,7 +197,9 @@ To include the changes that the hooks made, ``git add`` any
 files that were modified and run ``git commit`` (or, use ``git commit -a``
 to commit all changed files.)
 
-If you would like to run the checks without committing, use ``pre-commit run``.
+If you would like to run the checks without committing, use ``pre-commit run``
```
benjeffery, line 200: Maybe we should suggest `pre-commit run --all` then the caveat is unneeded?

benjeffery, line 200: Maybe we should suggest `pre-commit run --all` then the caveat is unneeded?

benjeffery (approved):

benjeffery:
Opps I meant to comment not approve - just that one fix to the eq method.

Thanks for finding and fixing my omission @petrelharp!

python/tests/test_tables.py
petrelharp: done.

docs/development.rst
petrelharp: good idea

petrelharp:
> Thanks for finding and fixing my omission @petrelharp!

It's /our/ omission, isn't it? But, thanks for getting everything so wonderfully in line that the dangling ends are so minor!

petrelharp:
I've added more tests. I think this is good to go.

jeromekelleher (approved):
LGTM. I'll let @benjeffery push the green button though, in case he sees anything.

PR #823, opened 2020-09-01
test for time with keep_roots

petrelharp:
There appears to be a bug related to mutation times and `simplify(keep_input_roots=True)`. This PR modifies a test so that it triggers the problem: simplify changes the node the mutation is on, if that node is retained as an input root.  Here's a simple example:
```
import tskit

t = tskit.TableCollection(sequence_length=1)
t.nodes.add_row(time=0, flags=1)
t.nodes.add_row(time=1, flags=0)
t.nodes.add_row(time=2, flags=0)
t.edges.add_row(parent=1, child=0, left=0, right=1)
t.edges.add_row(parent=2, child=1, left=0, right=1)
t.sites.add_row(position=0, ancestral_state='0')
t.mutations.add_row(site=0, derived_state='1', time=2, node=2)

print(t.mutations)
# id	site	node	time	derived_state	parent	metadata
# 0	0	2	2.0	1	-1	

t.simplify(keep_input_roots=True)
print(t.mutations)
# id	site	node	time	derived_state	parent	metadata
# 0	0	0	2.0	1	-1	
```
The `node` should be `1` afterwards, not `0`, and doing `t.tree_sequence()` throws
```
_tskit.LibraryError: A mutation's time must be < the parent node of the edge on which it occurs, or be marked as 'unknown'
```

I'll see if I can track this down now...

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -6458,7 +6458,8 @@ simplifier_insert_input_roots(simplifier_t *self)
                     if (ret != 0) {
                         goto out;
                     }
-                    simplifier_map_mutations(self, input_id, x->left, x->right, x->node);
+                    simplifier_map_mutations(
+                        self, input_id, x->left, x->right, output_id);
                 }
                 x = x->next;
             }
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -482,7 +482,7 @@ class Simplifier:
                 while x is not None:
                     if x.node != output_id:
                         self.record_edge(x.left, x.right, output_id, x.node)
-                        self.map_mutations(x.left, x.right, input_id, x.node)
+                        self.map_mutations(x.left, x.right, input_id, output_id)
                     x = x.next
                 self.flush_edges()
                 root_time = self.tables.nodes.time[output_id]
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -3223,24 +3223,27 @@ class TestSimplifyExamples(TopologyTestCase):
             keep_input_roots=True,
         )
 
-    def test_map_mutations_with_and_without_roots(self):
+    def test_place_mutations_with_and_without_roots(self):
         nodes_before = """\
         id      is_sample   time
         0       1           0
         1       0           1
+        2       0           2
         """
         edges_before = """\
         left    right   parent  child
         0       2       1       0
+        0       2       2       1
         """
         sites = """\
         id  position    ancestral_state
         0   1.0         0
         """
         mutations_before = """\
-        site    node    derived_state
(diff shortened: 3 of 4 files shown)


c/tskit/tables.c
```diff
@@ -6458,7 +6458,7 @@ simplifier_insert_input_roots(simplifier_t *self)
                     if (ret != 0) {
                         goto out;
                     }
-                    simplifier_map_mutations(self, input_id, x->left, x->right, x->node);
+                    simplifier_map_mutations(self, input_id, x->left, x->right, output_id);
```
petrelharp, line 6461: doh!

petrelharp:
I guess this slipped through the cracks because assigning a mutation to the root node instead of the node-above-the-root doesn't change anything except for the genotype of the node-above-the-root-node.I only saw it because mutation.time caused it to throw an error!

c/tskit/tables.c
jeromekelleher: Double Doh!

jeromekelleher:
Dammit! Great catch, thanks @petrelharp. Looks like we need to get a bugfix release out for this.

Can you add your example above in as an explicit standalone test? I think it's good to keep simple tests around that show these bugs, just in case we ever regress.

petrelharp:
I've fixed up the python version of simplify to agree, and added tests for above-the-root mutations here (and, elsewhere, since I've added them to tsutil.insert_branch_mutations, miraculously not breaking anything!). So... maybe this is ready to go? If others and/or CI don't spot anything?

petrelharp:
> Can you add your example above in as an explicit standalone test?

Well, I was going to, but instead I modified [this test](https://github.com/tskit-dev/tskit/pull/823/files#diff-5941ccebfc6d10311b8d79e8bc986cffR3226) to do pretty much exactly that. I think it's a simple and stand-along enough test?

petrelharp:
Gah, as noted elsewhere my pre-commit is broken...

jeromekelleher:
> Well, I was going to, but instead I modified this test to do pretty much exactly that. I think it's a simple and stand-along enough test?

Sure, sounds good. The tests are pretty strong now you've added mutations over the root to the insert mutations method.

jeromekelleher:
Ready to go, if you could squash, please

benjeffery:
Added to 0.3.1 which is shaping up to be a bugfix release.

petrelharp:
Squashed; should be ready.

jeromekelleher (approved):

jeromekelleher:
Thanks for cleaning up my mess @petrelharp! We'll get this bundled into a C release in the next few days so you can pull it into SLiM.

PR #833, opened 2020-09-03
Fix docs for sphinx 3

benjeffery:
We've pinned sphinx+breathe for a few months, things seemed to have settled now, only minimal changes needed. `autodocsumm` now needs to be pinned however, as its new version errors out.

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -26,9 +26,9 @@ sequences, for example using :func:`tskit.load`.
 The ``TreeSequence`` class
 ++++++++++++++++++++++++++
 
-.. autoclass:: tskit.TreeSequence()
-    :autosummary:
+.. autoclass:: TreeSequence()
     :members:
+    :autosummary:
 
 
 ++++++++++++++++++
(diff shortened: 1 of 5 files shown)


docs/python-api.rst
```diff
@@ -26,9 +26,9 @@ sequences, for example using :func:`tskit.load`.
 The ``TreeSequence`` class
 ++++++++++++++++++++++++++
 
-.. autoclass:: tskit.TreeSequence()
```
jeromekelleher, line 29: I think the brackets were deliberate here because it hides the messy low-level details of how these objects are constructed from the user (who doesn't need to know about this). Is this no longer possible in sphinx 3?

jeromekelleher (commented):
Looks good, but it's a significant change if we need to drop all of the custom signatures.

docs/python-api.rst
benjeffery: Ah ok, I get it. Reverted to brackets.

benjeffery:
@jeromekelleher I've re-added the brackets and checked the output. Should be ok now.

jeromekelleher (approved):

Issue #27, opened 2018-07-10
Sort using mutation.parent property


petrelharp:
Currently `sort_tables` doesn't ensure mutations occur after their parents. It probably should, but this behvaior is documented, so it's not major.

This is a way that tables can be out of order that `sort_tables` does not fix, so we should probably address it, although we don't have a use case that produces tables out of order in this way.

The only way I can think of to do this is to add a field to the mutation struct used for sorting that is "number of hops to the root", computed by following `mutation.parent` indices until you hit a `-1`.

hyanwong:
I've just hit this, when trying to turn a biallelic site into a multiallelic site, for testing purposes. I tried simply adding extra mutations onto nodes in the tree at that site, but fell foul of the mutation order requirement, which I can't easily see how to fix.

hyanwong:
Ping @Chris1221 as an issue that could probably be tackled along with https://github.com/tskit-dev/tskit/issues/117 (I would find it useful if this issue were solved)

Chris1221:
Great, thank you @hyanwong, I'll take a look in the next little while. :)

petrelharp:
This'd be nice to fix, but for your tests, @hyanwong - did you know that if you add the new mutations without any parents, then call `compute_mutation_parents`, this will fill them in for you (using order in the table)?

hyanwong:
@petrelharp thanks - the problem here is with mutations with order dependence (i.e. one mutation on the same lineage as another at the same site). I don't think `compute_mutation_parents` works in this case.

@Chris1221 could add the following code as a unit test for fixing it, perhaps?

```
import tskit
import msprime

ts = msprime.simulate(4, mutation_rate=2)
tables = ts.dump_tables()
tables.mutations.add_row(
    site=0, node=ts.first().root, derived_state="2", parent=tskit.NULL)
tables.sort()  # Does not sort mutations within a site
tables.build_index()  # Apparently needed to compute mutation parents
tables.compute_mutation_parents()  # Currently fails with _tskit.LibraryError
```

hyanwong:
NB - once fixed, remember to remove the docs noting this problem, on line 653 of `docs/data-model.rst`

Chris1221:
Thanks @hyanwong, I was thinking of asking for an example of the failure -- this will help me tremendously to get started.

jeromekelleher:
Closing this, as the semantics of sort are pretty well nailed down. Please reopen if you disagree.

[...]

PR #836, opened 2020-09-03
Mutations are now sorted properly.

hyanwong:


hyanwong:
Just correct the docs

docs/data-model.rst
```diff
@@ -656,10 +656,9 @@ For simplicity and algorithmic efficiency, mutations must also:
   **before** their children (i.e. if a mutation with ID :math:`x` has
   ``parent`` with ID :math:`y`, then we must have :math:`y < x`).
 
-Violations of these sorting requirements are detected at load time.
-The :meth:`TableCollection.sort` method ensures that mutations are sorted
-according site ID, but does not at present enforce that mutations occur
-after their parent mutations.
+Violations of these sorting requirements are detected at load time. The
+:meth:`TableCollection.sort` method will ensure that these sortedness
```
jeromekelleher, line 660: Well, not quite - what if all the mutation times are unknown?

hyanwong: Doesn't `sort` then sort so that parents appear before children? I assumed this had been fixed too.

jeromekelleher: No, it has no way of knowing. It's just stable in that it doesn't rearrange things.

hyanwong: Oh, sorry, I though that had been fixed too. So https://github.com/tskit-dev/tskit/issues/27 is still an active issue I guess. Is there even a solution (as this is a bit annoying!)

petrelharp: Perhaps this should say
```
but does not enforce that mutations occur after their parent mutations if mutation times are unknown.
```

If the `parent` attribute is correct then it could be used to ensure that the sortedness criteria are satisfied even without times. I think we didn't do this because we didn't have a use case where parent mutation IDs are known but out of order.

hyanwong: So is there any way to change the sorting code so that the example I gave in https://github.com/tskit-dev/tskit/issues/27#issuecomment-562062932 actually works (i.e. sort in the correct way to allow `compute_mutation_parents` to do its stuff?

petrelharp: (refreshing my memory of the discussion in #651...)  Well, if you put a mutation.time in that example, then the sorting code *will* do what you want. Otherwise, I that sorting mutations to respect ordering on the trees will require constructing the trees, which we'd really like to avoid doing. Do you have a use case that this is making things difficult for?

jeromekelleher: I don't think we should change the sorting order @hyanwong - we went through it all in great detail the first time around and we ended up with a compromise that works well. There's no reason to change it now I think.

jeromekelleher:
Can we close this @hyanwong ?

PR #850, opened 2020-09-11
Fix type in documentation for tskit.Mutation

brianzhang01:
Currently in the [documentation](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.Mutation), the type of node shows up as float and the type of time is absent. It surprised me that node would be a float and it turned out to be a typo in the documentation.


petrelharp:
yep, good catch! thanks!!

PR #851, opened 2020-09-15
Make simplify use mostly keyword-only args

### LINKED ISSUE #846: Make all except first arg to simplify() keyword-only
Now that we are embracing keyword-only args, we should impose this in `simplify()`, as there are lots of similar-sounding options, with no obvious order. I.e. we should define

```
    def simplify(
        self,
        samples=None,
        *,
        filter_zero_mutation_sites=None,  # Deprecated alias for filter_sites
        reduce_to_site_topology=False,
        filter_populations=True,
        filter_individuals=True,
        filter_sites=True,
        keep_unary=False,
        keep_input_roots=False,
    )
```

Also see https://github.com/tskit-dev/tskit/issues/374 . This clearly breaks backwards compatibility, so I have opened it as an issue in case we want to discuss.

hyanwong:
Fixes #846

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,9 +4,11 @@
 
 **Breaking changes**
 
-- All but the first argument to ``TreeSequence.simplify``, ``TableCollection.simplify``
-  must be keyword arguments, not positional. This allows the order & number of arguments
-  to be made the same. (:user:`hyanwong`, :issue:`374`, :issue:`846`, :pr:`851`)
+- Change several methods (``simplify()``, ``trees()``, ``Tree()``) so most parameters
+  are keyword only, not positional. This allows reordering of parameters, so
+  that deprecated parameters can be moved, and the parameter order in similar functions,
+  e.g. ``TableCollection.simplify`` and ``TreeSequence.simplify()`` can be made
+  consistent (:user:`hyanwong`, :issue:`374`, :issue:`846`, :pr:`851`)
 
 
 **Features**
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,6 +2,13 @@
 [X.X.X] - XXXX-XX-XX
 --------------------
 
+**Breaking changes**
+
+- All but the first argument to ``TreeSequence.simplify``, ``TableCollection.simplify``
+  must be keyword arguments, not positional. This allows the order & number of arguments
+  to be made the same. (:user:`hyanwong`, :issue:`374`, :issue:`846`, :pr:`851`)
+
+
 **Features**
 
 - Genomic intervals returned by python functions are now namedtuples, allowing ``.left``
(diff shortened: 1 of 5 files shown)


benjeffery:
@hyanwong I just force-pushed to add tests for this and the other kwarg only methods. This is mainly so we don't add positional arguments in error in future.

hyanwong:
Do we also want to make `ts.trees()` kwonly (and presumably most of the args to Tree.__init__)?

hyanwong:
I've added the kw-only stuff to `.trees()` as well, @benjeffery (in a separate commit), as Jerome [gave that the thumbs up](https://github.com/tskit-dev/tskit/issues/846#issuecomment-691297407)

python/CHANGELOG.rst
```diff
@@ -2,6 +2,13 @@
 [X.X.X] - XXXX-XX-XX
 --------------------
 
+**Breaking changes**
+
+- Change several methods (``simplify()``, ``trees()``, ``Tree()``) so most parameters
+  are keyword only, not positional. This allows sensible & standard order of arguments,
```
jeromekelleher, line 8: This second sentence will confuse people.

jeromekelleher (commented):
Looks good. I'm slightly nervous about making ``tracked_samples`` kw-only, though, as I have a feeling there's a fair bit of coding using that. Are we sure all the documentation examples are ok with this?

I guess tsinfer is a good yardstick - if we don't break any code in there in things like the validation suite, then maybe it's OK. Still, breaking peoples code gratuitously for the sake of neatness isn't good.

hyanwong:
> Looks good. I'm slightly nervous about making `tracked_samples` kw-only, though, as I have a feeling there's a fair bit of coding using that. Are we sure all the documentation examples are ok with this?

I could shift the asterisk so that `tracked_samples` can be positional? Whatever you prefer.

jeromekelleher:
> I could shift the asterisk so that tracked_samples can be positional? Whatever you prefer.

This is probably safer, I think. No point in breaking code for the sake of it.

hyanwong:
> > I could shift the asterisk so that tracked_samples can be positional? Whatever you prefer.
> 
> This is probably safer, I think. No point in breaking code for the sake of it.

Done, with new changelog. Ready to merge, I think.

jeromekelleher:
Great, thanks. It would be nice to test that ``trees([0])`` is the same thing as ``trees(tracked_samples=[0])``, just in case that asterisk gets moved by accident at some point.

hyanwong:
> Great, thanks. It would be nice to test that `trees([0])` is the same thing as `trees(tracked_samples=[0])`, just in case that asterisk gets moved by accident at some point.

Added `test_tracked_samples_is_first_arg`

benjeffery (approved):
LGTM

benjeffery:
Just as I was about to merge I noticed the tests were not quite right, fixed and force pushed.

Issue #31, opened 2019-01-10
Record provenance for tableollection operations


jeromekelleher:

hyanwong:
See https://github.com/tskit-dev/tskit/pull/243 for an initial PR suggesting a universal function that does this from TreeSequences. It should be relatively easy to modify it for a TableCollection method.

PR #853, opened 2020-09-16
Put record_provenance in table.simplify

### LINKED ISSUE #374: How to make simplify() consistent between tree seqs & tables?
At the moment, `tables.simplify()` does not take a `record_provenance` parameter. It should. However, `ts.simplify()` *does* take such a parameter, as the penultimate one (before the `keep_unary` argument).

If we add this param as the last param of `tables.simplify()`, it breaks less code, but it means that the tree_sequence and tables versions have parameters in different orders, which is *horrible*.

There's no easy solution here that I can see. My preferred option is for `ts.simplify()` to match the other tree sequence methods and have `record_provenance` as the last parameter, but that's a breaking change.

Suggestions?

hyanwong:
Fixes #374. Also places deprecated aliases at end of the argument list

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -107,7 +107,7 @@ def simple_keep_intervals(tables, intervals, simplify=True, record_provenance=Tr
     tables.build_index()
     tables.compute_mutation_parents()
     if simplify:
-        tables.simplify()
+        tables.simplify(record_provenance=False)
     if record_provenance:
         parameters = {"command": "keep_intervals", "TODO": "add parameters"}
         tables.provenances.add_row(
@@ -2833,6 +2833,7 @@ class TestSimplifyExamples(TopologyTestCase):
             samples=samples,
             filter_sites=filter_sites,
             keep_input_roots=keep_input_roots,
+            record_provenance=False,
         )
         if debug:
             print("before")
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2226,13 +2226,14 @@ class TableCollection:
         self,
         samples=None,
         *,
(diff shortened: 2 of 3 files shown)


jeromekelleher (approved):
LGTM

benjeffery:
@hyanwong Do we have an issue for the ongoing work that is marked as TODO in this PR?

hyanwong:
You mean the provenance stuff? It's all full of TODOs like this (I'm just copying the other methods here). I made a rough PR at https://github.com/tskit-dev/tskit/pull/243 (but then left it for you to do properly :) ) and there's a minimally descriptive PR at https://github.com/tskit-dev/tskit/issues/31 .

benjeffery:
Ah cool! I'll pick that up soon, need to do a wave of provenance work to flesh out the API.

hyanwong:
Yep. It would be helpful, if possible, to provide a generic way of doing this. E.g. in tsinfer we want to record provenance by accessing the command name and parameters, and currently we are omitting the parameters ( https://github.com/tskit-dev/tsinfer/issues/332 )

hyanwong:
Also, I've taken the (non-executive) decision to place deprecated-but-documented parameters at the end of the parameter list, as long as it is possible to change their position without changing the behaviour (i.e. if they are keyword-only anyway). I guess that's OK @benjeffery & @jeromekelleher ?

benjeffery:
> Also, I've taken the (non-executive) decision to place deprecated-but-documented parameters at the end of the parameter list, as long as it is possible to change their position without changing the behaviour (i.e. if they are keyword-only anyway). I guess that's OK @benjeffery & @jeromekelleher ?

Yep, if it doesn't break anything like you say I agree.

Issue #820, opened 2020-08-30
pre-commit hooks don't run clang-format


grahamgower:
Is this to avoid having it as a developer requirement, for python-only developers?

jeromekelleher:
Exactly - we figured we'd confuse/put off more people by making this a hard requirement for Python development. I agree it's annoying for us C coders though. I wonder is there any way of optionally running it?

grahamgower:
Well part of the problem, as I see it, is that the `pre-commit` tool wants to live at ~/.git/hooks/pre-commit, and is not a shell script. This means (a) one can't trivially add a clang-format line (or anything else) to one's own pre-commit hooks, and (b) any changes that are made will be overwritten when the pre-commit tool wants to update.

benjeffery:
Pre-commit only runs checks on changed files - it may be that a python-only change wouldn't need it to be installed. Might be worth looking into if we think it's a pain point.

grahamgower:
Is it possible to move the pre-commit tool's pre-commit script to ~/.git/hooks/pre-commit-pre-commit.py (or something slightly more sane), and then have a regular shell script in ~/.git/hooks/pre-commit that calls this? Then one could whack `clang-format` into the pre-commit shell script, and the pre-commit tool's python script could be updated without affecting this.

benjeffery:
> Is it possible to move the pre-commit tool's pre-commit script to ~/.git/hooks/pre-commit-pre-commit.py (or something slightly more sane), and then have a regular shell script in ~/.git/hooks/pre-commit that calls this? Then one could whack `clang-format` into the pre-commit shell script, and the pre-commit tool's python script could be updated without affecting this.

Not without modifying the pre-commit python module we're using. Let me test if adding `clang-format` to our pre-commit config forces python devs to install it.

benjeffery:
I've looked into this @grahamgower and it's not possible to easily move pre-commit's script for all users as it is installed by pre-commit itself. You can do this on a local repo though if you want. Adding clang-format to the pre-commit config also is problematic for python-only devs. Sorry there isn't an immediately clean solution here!

grahamgower:
Thanks @benjeffery, I did end up doing this for my local repo. Here's what I ended up with.
```
$ cat .git/hooks/pre-commit
#!/bin/sh

.git/hooks/pre-commit_pre-commit.py

clang-format --dry-run -Werror c/tskit/* c/tests/*.c c/tests/*.h
if [ $? != "0" ]; then
[...]

PR #856, opened 2020-09-22
Update pre-commit packages

benjeffery:
Whilst looking at #820 I updated our pre-commit dependencies. 

- Black now formats docstrings :+1: and behaviour round trailing commas has been improved.  
- flake8 has new errors about visually ambiguous variable names. It wasn't a lot of changes so changed the code instead of disabling the error.

### DIFF
--- a/python/tests/__init__.py
+++ b/python/tests/__init__.py
@@ -83,35 +83,35 @@ class PythonTree:
             v = self.right_sib[v]
         return ret
 
-    def _preorder_nodes(self, u, l):
-        l.append(u)
+    def _preorder_nodes(self, u, node_ist):
+        node_ist.append(u)
         for c in self.children(u):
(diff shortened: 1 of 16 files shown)


python/tests/__init__.py
```diff
@@ -83,35 +83,35 @@ def children(self, u):
             v = self.right_sib[v]
         return ret
 
-    def _preorder_nodes(self, u, l):
-        l.append(u)
+    def _preorder_nodes(self, u, li):
```
jeromekelleher, line 86: how about ``node_list`` here? makes it pretty clear

jeromekelleher (approved):
Very nice, a good batch of improvements. One minor suggestion for improving variable name in the production code, but merge away as you see fit.

PR #867, opened 2020-09-28
Convert tests to pytest

benjeffery:
I've tagged a load of people for comments, ignore if you're not interested or busy!

For some time now the `nose` homepage has had this message:

"Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2."

https://pytest.org/ seems to be the most capable and popular replacement, for example it gives better error messages than `nose`. Our test suite is already compatible, but is not idiomatic pytest. The main change for that is to use plain `assert`s rather than `self.assert*`.  This PR shows what that would look like - I'm a fan of the cleaner style. One could go further and use `pytest`'s fixtures and class-less tests.

This is a big change to the test code, but also a change to people's workflow. I don't think it will be disruptive however.

PS I haven't converted `self.assertAlmostEqual` as that is a bit more work.

PPS This PR speeds up circle CI from ~22 min to 6 min! (Circle CI gives you 22 cores! Who knew!)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,9 +123,8 @@ jobs:
       - run:
           name: Run Python tests
           command: |
-            PYTHONPATH=python nosetests -v --with-coverage --cover-package tskit \
-              --cover-branches --cover-erase --cover-xml \
-              --cover-inclusive python/tests
+            PYTHONPATH=python pytest --cov=python/tskit  --cov-report=xml --cov-branch \
+            -n `nproc` python
 
       - run:
           name: Upload Python coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ install:
 
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      nosetests -vs --processes=`sysctl -n hw.ncpu` --process-timeout=5000 tests;
+      pytest -n `sysctl -n hw.ncpu` python;
     else
-      nosetests -vs --processes=`nproc` --process-timeout=5000 tests;
+      pytest -n `nproc` python;
     fi
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@ build_script:
   - cmd: python setup.py build_ext --inplace
   # Install some modules needed for tests
   - cmd: python -m pip install -r requirements/appveyor-pypi.txt
-  - cmd: python -m nose -vs --processes=%NUMBER_OF_PROCESSORS% --process-timeout=5000
+  - cmd: python -m pytest -n %NUMBER_OF_PROCESSORS%
 
 after_test:
   - cmd: python setup.py bdist_wheel
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -382,25 +382,32 @@ Tests
 -----
 
 The tests are defined in the ``tests`` directory, and run using
-`nose <http://nose.readthedocs.io/en/latest/>`_. If you want to run
+`pytest <https://docs.pytest.org/en/stable/>`_. If you want to run
 the tests in a particular module (say, ``test_tables.py``), use:
 
 .. code-block:: bash
 
-    $ python3 -m nose -vs tests/test_tables.py
+    $ python3 -m pytest tests/test_tables.py
 
 To run all the tests in a particular class in this module (say, ``TestNodeTable``)
 use:
 
 .. code-block:: bash
 
-    $ python3 -m nose -vs tests/test_tables.py:TestNodeTable
+    $ python3 -m pytest tests/test_tables.py::TestNodeTable
 
 To run a specific test case in this class (say, ``test_copy``) use:
 
 .. code-block:: bash
 
-    $ python3 -m nose -vs tests/test_tables.py:TestNodeTable.test_copy
+    $ python3 -m pytest tests/test_tables.py::TestNodeTable::test_copy
+
+You can also run tests with a keyword expression search. For example this will
+run all tests that have ``TestNodeTable`` but not ``copy`` in their name:
+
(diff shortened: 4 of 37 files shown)


gtsambos (approved):
Looks good to me @benjeffery  👍  I remember wondering why tskit was using nose when I first got involved. The pytest syntax looks just as easy to use/understand, and if it lessens the time I spend anxiously waiting for CI to pass, all the better! I think it would be a good idea to also modify the developer docs as soon as possible, because as you say, it's a major change to the development workflow that will immediately affect everyone with an open PR.

petrelharp:
This looks fine, I'm in favor of it if it's the right way to go. Thanks!  But, how non-idiomatic pytest is it after this PR? Are fixtures and classless tests more idiomatic (I don't have a good picture of what that'd look like)?

hyanwong:
Good work @benjeffery - looks fiddly to get it all working. I always have to remind myself to use self.assertXXX rather than the more idiomatic Python native call, and had noticed the nose obsoletion issue so I like this. And the time improvement can't be a bad thing at all (although I don't understand why). Doe we still get the nice output colours with pytest?

How come self.assertAlmostEqual still works? Is this just a patch that pytest provides? Numpy provides an almost-equal thing that could probably be used as an almost drop-in replacement in some cases.

jeromekelleher:
Looks good to me. I'm not hugely enthusiastic about actually doing the change, but we will need to grasp this nettle at some point so now may be as good a time as any.

Do test classes still need to be subclasses of unittest to work? Does test discovery depend on this, or can we have a bunch of classes in there called "TestX", and pytest will find and execute functions called ``test_X`` in them? There's a lot of tests that don't do any setup/teardown, so it'd be nice to make them not subclasses of unittest.TestCase if we could. I'm assuming that the pytest idiom is to not use unittest at all?

I'm not sure it's worth the effort going through and making old tests pytest-idiomatic, it would end up being weeks of work and not actually get us anywhere. I'm happy with the idea of having slightly un-idiomatic tests, while still taking advantage of pytest's features.

benjeffery:
> I think it would be a good idea to also modify the developer docs as soon as possible,

Fixed in https://github.com/tskit-dev/tskit/pull/867/commits/2ae694d11d70b98d7c164eed550dd57f842722e9

> But, how non-idiomatic pytest is it after this PR? Are fixtures and classless tests more idiomatic (I don't have a good picture of what that'd look like)?

Fixtures are pretty cool if not used too "magically" they basically let you define things your tests use and re-use. They get passed to the test as arguments. For example if we had:
```python
@pytest.fixture(scope="module")
def simple_tree():
    msprime.simulate(10, mutation_rate=1, random_seed=42)
```
`simple_tree` would be called once per test file. Then any test that has `simple_tree` as an argument would be called with that single value. This saves a lot of time and duplication. For mutable things you'd set the `scope` to `function`. Scope can also be `class`. There are also defined fixtures for common tasks: https://docs.pytest.org/en/stable/fixture.html

Classless tests are just that - you just define a function `test_whatever` and it will be run,  I like grouping tests by class though, so I'd use sparingly.

> And the time improvement can't be a bad thing at all (although I don't understand why).

`nose` didn't let us measure coverage in parallel, pytest does.

> Do we still get the nice output colours with pytest?

Yep

> How come self.assertAlmostEqual still works?

That's part of the python `unittest` module which pytest still works with. All the `self.assert*` can still be used.

> Do test classes still need to be subclasses of unittest to work? Does test discovery depend on this, or can we have a bunch of classes in there called "TestX", and pytest will find and execute functions called test_X in them? 

No, subclassing isn't needed. The discovery follows this method:
 -  Search for test_*.py or *_test.py files
 -  Find "test" prefixed test functions or methods outside of class
 -  Find "test" prefixed test functions or methods inside "Test" prefixed classes (without an __init__ method), or subclasses of `unittest.TestCase`

> I'm assuming that the pytest idiom is to not use unittest at all?

Pretty much - although it is certainly not awkward to use it.

> I'm not sure it's worth the effort going through and making old tests pytest-idiomatic, it would end up being weeks of work and not actually get us anywhere. I'm happy with the idea of having slightly un-idiomatic tests, while still taking advantage of pytest's features.

Exactly, this PR was mostly made with an automated tool https://github.com/pytest-dev/unittest2pytest. Lets not make busy work! We can use pytest features moving forward and refactor old stuff if we happen to be editing it.

jeromekelleher:
Sounds good, thanks @benjeffery. I think it'd be good to make any test classes that don't make use of unittest features plain classes if we can, but maybe this is best left to a follow-up.

benjeffery:
> I think it'd be good to make any test classes that don't make use of unittest features plain classes if we can, but maybe this is best left to a follow-up.

Done in https://github.com/tskit-dev/tskit/pull/867/commits/3ddf6e9e73668eb2a03c45f0a235bb1673f40936

jeromekelleher (approved):

jeromekelleher:
Ah, looks like you need to port over some ``@unittest.skipIf(IS_WINDOWS)`` logic. There's a bunch of things that we either can't do on windows or can't be bothered making the tests more obscure for the sake of keeping windows happy.

benjeffery:
> Ah, looks like you need to port over some `@unittest.skipIf(IS_WINDOWS)` logic. There's a bunch of things that we either can't do on windows or can't be bothered making the tests more obscure for the sake of keeping windows happy.

Fixed in https://github.com/tskit-dev/tskit/pull/867/commits/5298020396b22192797dfefb11d60beccd02cb63

benjeffery:
I've added a `--skip-slow` option to the tests. This gets the test suite wallclock time down to 20 seconds on my desktop machine.
:racehorse:

Issue #870, opened 2020-09-29
Add indexes to the dict encoding


jeromekelleher:
Currently we don't write out the indexes to the dict encoding, which means that we rebuild then each time we pass tree sequence data through the low-level encoding (see #864 , #862). This is really quite inefficient for large tables and is probably causing some significant inefficiencies for msprime.simulate and other applications where we're shuffling around data through the dict encoding.

Issue #657, opened 2020-06-01
File objects for Python dump/load


jeromekelleher:
#599 added support for dumping and loading from streams in the C code. We want to add this functionality to the Python api also. This can largely follow the example set in the kastore code.

jeromekelleher:
This is a good one to do in combination with #14.

jeromekelleher:
Notes from #609:

Semantics should follow what we've done in kastore, where the Python C interface is adapted to use a fileno. This would mean that we should be able to work with any kind of fileobj other than a BytesIO.

There's a choice to be made in whether we rename the parameter from path to file (we could keep path as a deprecated alias keyword only arg).

Issue #14, opened 2018-06-29
Provide TableCollection.dump( ) method


petrelharp:
Currently to write out a table collection, you have to make it into a tree sequence first, which then just calls `table_collection_dump()` under the hood. Unless there's a good reason not, it'd be convenient sometimes to dump the table collection directly.

jeromekelleher:
I think we can do this now and should pick it up.

PR #871, opened 2020-09-29
Add nbytes to Table and TableCollection.

### LINKED ISSUE #54: Add ``total_bytes`` attribute to Table objects
It'll be useful to be able to track the actual size in bytes of tables. Add an attribute to return the total amount of memory consumed by each table.

jeromekelleher:
It's pretty useful to have a summary of the number of bytes used by a table collection. This uses a definition based on the total bytes used to encode the data, which seems like as good a definition as any. Trying to capture the actual memory used seems like a fiddly waste of time.

However, this currently doesn't work because of a few limitations:

- We can't dump a TableCollection #14 
- We can't dump to a File object #657 (This isnt' strictly necessary, but testing would be a lot handier if we could avoid making a TemporaryDirectory, etc)
- We don't have any information about the indexes in the dict encoding #870

Closes #54

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -35,13 +35,13 @@ import pickle
 import platform
 import random
 import re
-import shutil
 import tempfile
 import unittest
 import uuid as _uuid
 import warnings
 
 import attr
+import kastore
 import msprime
 import networkx as nx
 import numpy as np
@@ -360,17 +360,10 @@ class TestMRCACalculator:
                         assert mrca == sv.get_mrca(j, k)
 
 
-class HighLevelTestCase(unittest.TestCase):
+class HighLevelTestCase:
     """
     Superclass of tests on the high level interface.
     """
-
-    def setUp(self):
-        self.temp_dir = tempfile.mkdtemp(prefix="tsk_hl_testcase_")
-        self.temp_file = os.path.join(self.temp_dir, "generic")
-
-    def tearDown(self):
-        shutil.rmtree(self.temp_dir)
 
     def verify_tree_mrcas(self, st):
         # Check the mrcas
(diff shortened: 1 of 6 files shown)


benjeffery:
_repr_html_ needs nbytes when this is all working.

benjeffery:
@jeromekelleher Are you happy for me to pick this up and push commits to this PR?

jeromekelleher:
Yes please @benjeffery, that would be great, thanks. I think we can define the nbytes as "the total number of bytes required for the values in the dict encoding". I don't think there's any point in including the keys. Therefore, the nbytes for a TreeSequence is the same as its underlying TableCollection.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/871/)

benjeffery:
@jeromekelleher Think I'm happy with this now. It has a few test cleanups included that I came across.

benjeffery (approved):
Approving as it is your PR and mergify needs a non-owner approval.

python/tests/test_highlevel.py
```diff
@@ -486,7 +480,7 @@ def verify_trees(self, ts):
             assert not (st2 != st1)
             left, right = st1.get_interval()
             breakpoints.append(right)
-            self.assertAlmostEqual(left, length)
+            assert left == approx(length)
```
jeromekelleher, line 483: I had a good look at this, and I think it'd be better if we used the pattern ``x == pytest.approx(y)``. We're using ``pytest.`` for a bunch of other things, and I don't think there's a good reason for making an exception here. It's more obvious to me what it's actually doing as well, when I see  ``x == pytest.approx(y)`` in isolation (i.e., if I was coming in fresh to this code without reading the imports)

jeromekelleher (commented):
Nice! See comment above about pytest patterns, though

python/tests/test_highlevel.py
benjeffery: Yep, looks like something local. Fixed.

benjeffery:
Fixed up, rebased and squashed.

PR #873, opened 2020-09-29
0.3.2 prep

benjeffery:
I added a changelog for tsk_bug_assert, as it is part of the public API intended for use in downstream code. Borderline case though, so happy to remove if we don't want to promote it.


jeromekelleher (approved):
LGTM!

PR #878, opened 2020-09-30
Remove out-of-date comment

### LINKED ISSUE #877: Check attrs minimum version
Since this [update](https://github.com/tskit-dev/tskit/commit/30fb6d22f81c8cf9d98a163a24f3a24b1f9a8516#r42844819) we might need a new version of attrs.

Also the comment in question needs updating.

benjeffery:
Fixes #877

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -45,7 +45,6 @@ from tskit import UNKNOWN_TIME
 attr_options = {"slots": True, "frozen": True, "auto_attribs": True}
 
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
What about the minimum version of attrs, though? I think we backed out of doing this change at the time because of compatibilty worries.

benjeffery:
> What about the minimum version of attrs, though? I think we backed out of doing this change at the time because of compatibilty worries.

It's all good, we've used `eq` on a line lower down for a while and `setup.py` has the right version already.

jeromekelleher (approved):

PR #880, opened 2020-09-30
Isolated lwt_interface code.

jeromekelleher:
Here's a proposal for isolating the LightweightTableCollection code, and making it a bit easier to include in subprojects and to inherit tests. The idea is that (e.g.) msprime would soft-link the ``lwt_interface`` directory into its project root, and it should be able to include the tests in the same way that tskit does.

If it all works and we like it then I'll tidy up the code.

The idea is to also have a Cython example in the lwt_interface directory, which will serve as something we can test as well as documentation.


jeromekelleher:
Oh yeah, pytest FTW here! I doubt I could have pulled off the tricks to get the tests running here using nose.

jeromekelleher:
Great, looks like this is all working. Just tried it out in msprime, and it seems to work well. Should make maintenance as we update the table definitions much easier.

Going to get some feedback from @benjeffery first, then finish this up I think.

python/lwt_interface/README.md
```diff
@@ -0,0 +1,12 @@
+# LightweightTableCollection interface
+
+The files in this directory define the LightweightTableCollection
+interface used to safely interchange table collection data between 
+difference compiled instances of the tskit C library. This is a
```
petrelharp, line 5: "different compiled"

benjeffery (approved):
This is great - seems like you've managed to de-duplicate really cleanly.

jeromekelleher:
I think we should probably just merge this, as I don't know when I'll get to finish it up @benjeffery. Would you mind opening issues to track

- Finishing documentation for this
- Running clang-format on _tskitmodule.c (turns out it's fine if you just turn off clang-format in a few key places). Good to do this in conjunction with update to C99

Issue #275, opened 2019-07-30
compute covariance matrix and PCA


petrelharp:
To compute a covariance matrix, or do PCA, we currently have to export the genotype matrix. It would be nice to do this using the statistics framework, and totally do-able. It is also possible to find PCs without computing the covariance matrix first. This would be a nice compact but very useful contribution for someone to take on.

daniel-trejobanos:
I am curious about this, specially how to compute the PCA without the covariance matrix? do you have a paper explaining this?

petrelharp:
I have not worked out the details, but the basic idea is that the PCs are the eigenvectors of the genetic covariance matrix, and modern iterative methods (like Krylov methods) exist to find the eigenvectors of a matrix A without ever computing A explicitly, but rather finding the result of multiplying A by some random vectors. This paper: https://www.ncbi.nlm.nih.gov/pubmed/26924531 does something like this. In our situation, A = G^T G, where G is the genotype matrix (possibly normalized), and so we can use our general statistics to quickly compute u^T A v for vectors u and v.

That's the general idea. I have not worked out the details, so it's possible there's something very tricky in there, but I'm happy to help work it through.

hyanwong:
Also see performance issues in https://github.com/tskit-dev/tskit/issues/1743

hanbin973:
I think we can close this issue with 0f7fa20 ?

petrelharp:
🎉 Closed in #3008 and other places!

PR #898, opened 2020-10-08
Add function for covariance matrix

brieuclehmann:
An implementation of covariance matrix calculation for sample leaves in a tree sequence. The covariance between a pair of leaves in a single tree can be calculated as the time of their most recent common ancestor to the time of the root. The covariance across a tree sequence is a weighted sum of these tree covariances, with weights given by the span of the tree.

The incremental implementation uses similar reasoning as the KC distance implementation #548 . We maintain the covariance between all pairs of leaves as follows. For each edge e, perform an upward and downward traversal. While traversing up toward the root, update the pairs of leaves where one leaf is in the subtree affected by e and one is not. Traversing down from e, update all pairs of leaves where both leaves are in the subtree. Pairs where both leaves are outside of the subtree under e haven't been affected by the insertion/removal of that edge.

Addresses part of #275

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2753,8 +2753,8 @@ out:
 }
 
 static int
-relatedness_summary_func(size_t state_dim, const double *state, 
-    size_t result_dim, double *result, void *params)
+relatedness_summary_func(size_t state_dim, const double *state, size_t result_dim,
+    double *result, void *params)
 {
     sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
     const double *x = state;
@@ -2778,9 +2778,9 @@ relatedness_summary_func(size_t state_dim, const double *state,
 
 int
 tsk_treeseq_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
-    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets, tsk_size_t num_index_tuples,
-    const tsk_id_t *index_tuples, tsk_size_t num_windows, const double *windows, double *result,
-    tsk_flags_t options)
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
+    tsk_size_t num_index_tuples, const tsk_id_t *index_tuples, tsk_size_t num_windows,
+    const double *windows, double *result, tsk_flags_t options)
 {
     int ret = 0;
     ret = check_sample_stat_inputs(num_sample_sets, 2, num_index_tuples, index_tuples);
@@ -2795,7 +2795,7 @@ out:
 }
 
 static int
-Y2_summary_func(size_t TSK_UNUSED(state_dim), const double *state, size_t result_dim, 
+Y2_summary_func(size_t TSK_UNUSED(state_dim), const double *state, size_t result_dim,
     double *result, void *params)
 {
     sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2752,6 +2752,7 @@ out:
     return ret;
 }
 
+static int
 relatedness_summary_func(size_t state_dim, const double *state, 
     size_t result_dim, double *result, void *params)
 {
@@ -2769,7 +2770,7 @@ relatedness_summary_func(size_t state_dim, const double *state,
 
     for (k = 0; k < state_dim; k++) {
         num += args.sample_set_sizes[k];
-    } 
+    }
 
     if (num != sumx) {
         c = 1;
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2760,27 +2760,17 @@ relatedness_summary_func(size_t state_dim, const double *state,
     const double *x = state;
     tsk_id_t i, j;
     size_t k;
-    int c = 0;
     double sumx = 0;
     double meanx;
-    double num = 0;
 
     for (k = 0; k < state_dim; k++) {
         sumx += x[k];
     }
 
-    for (k = 0; k < state_dim; k++) {
-        num += args.sample_set_sizes[k];
-    }
-
-    if (num != sumx) {
-        c = 1;
-    }
     meanx = sumx / (double) state_dim;
     for (k = 0; k < result_dim; k++) {
         i = args.set_indexes[2 * k];
         j = args.set_indexes[2 * k + 1];
-        // result[k] = x[i] * x[j] * c;
         result[k] = (x[i] - meanx) * (x[j] - meanx);
     }
     return 0;
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2759,7 +2759,7 @@ relatedness_summary_func(size_t state_dim, const double *state,
     const double *x = state;
     tsk_id_t i, j;
     size_t k;
-    double c = 0;
+    int c = 0;
     double sumx = 0;
     double num = 0;
 
@@ -2767,13 +2767,17 @@ relatedness_summary_func(size_t state_dim, const double *state,
         sumx += x[k];
     }
 
-    for (k = 0; k < result_dim; k++) {
+    for (k = 0; k < state_dim; k++) {
         num += args.sample_set_sizes[k];
     } 
 
     if (num != sumx) {
         c = 1;
     }
+
+    // printf("num = %0.7f\n", num);
+    // printf("sumx = %0.7f\n", sumx);
+    // printf("c = %d\n", c);
 
     for (k = 0; k < result_dim; k++) {
         i = args.set_indexes[2 * k];
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2752,8 +2752,57 @@ out:
     return ret;
 }
 
+relatedness_summary_func(size_t state_dim, const double *state, 
+    size_t result_dim, double *result, void *params)
+{
+    sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
+    const double *x = state;
+    tsk_id_t i, j;
+    size_t k;
+    double c = 0;
+    double sumx = 0;
+    double num = 0;
+
+    for (k = 0; k < state_dim; k++) {
+        sumx += x[k];
+    }
+
+    for (k = 0; k < result_dim; k++) {
+        num += args.sample_set_sizes[k];
+    } 
+
+    if (num != sumx) {
+        c = 1;
+    }
+
+    for (k = 0; k < result_dim; k++) {
+        i = args.set_indexes[2 * k];
+        j = args.set_indexes[2 * k + 1];
+        result[k] = x[i] * x[j] * c;
+    }
+    return 0;
+}
+
+int
+tsk_treeseq_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets, tsk_size_t num_index_tuples,
+    const tsk_id_t *index_tuples, tsk_size_t num_windows, const double *windows, double *result,
+    tsk_flags_t options)
+{
+    int ret = 0;
+    ret = check_sample_stat_inputs(num_sample_sets, 2, num_index_tuples, index_tuples);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_treeseq_sample_count_stat(self, num_sample_sets, sample_set_sizes,
+        sample_sets, num_index_tuples, index_tuples, relatedness_summary_func,
+        num_windows, windows, result, options);
+out:
+    return ret;
+}
+
 static int
-Y2_summary_func(size_t TSK_UNUSED(state_dim), const double *state, size_t result_dim,
+Y2_summary_func(size_t TSK_UNUSED(state_dim), const double *state, size_t result_dim, 
     double *result, void *params)
 {
     sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2762,6 +2762,7 @@ relatedness_summary_func(size_t state_dim, const double *state,
     size_t k;
     int c = 0;
     double sumx = 0;
+    double meanx;
     double num = 0;
 
     for (k = 0; k < state_dim; k++) {
@@ -2775,15 +2776,12 @@ relatedness_summary_func(size_t state_dim, const double *state,
     if (num != sumx) {
         c = 1;
     }
-
-    // printf("num = %0.7f\n", num);
-    // printf("sumx = %0.7f\n", sumx);
-    // printf("c = %d\n", c);
-
+    meanx = sumx / (double) state_dim;
     for (k = 0; k < result_dim; k++) {
         i = args.set_indexes[2 * k];
         j = args.set_indexes[2 * k + 1];
-        result[k] = x[i] * x[j] * c;
+        // result[k] = x[i] * x[j] * c;
+        result[k] = (x[i] - meanx) * (x[j] - meanx);
     }
     return 0;
 }
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -1334,6 +1334,17 @@ test_paper_ex_divergence(void)
 }
 
 static void
+test_paper_ex_relatedness_errors(void)
+{
+    tsk_treeseq_t ts;
+
+    tsk_treeseq_from_text(&ts, 10, paper_ex_nodes, paper_ex_edges, NULL, paper_ex_sites,
+        paper_ex_mutations, paper_ex_individuals, NULL, 0);
+    verify_two_way_stat_func_errors(&ts, tsk_treeseq_relatedness);
+    tsk_treeseq_free(&ts);
+}
+
+static void
 test_paper_ex_Y2_errors(void)
 {
     tsk_treeseq_t ts;
@@ -1679,6 +1690,7 @@ main(int argc, char **argv)
         { "test_paper_ex_Y1", test_paper_ex_Y1 },
         { "test_paper_ex_divergence_errors", test_paper_ex_divergence_errors },
         { "test_paper_ex_divergence", test_paper_ex_divergence },
+        { "test_paper_ex_relatedness_errors", test_paper_ex_relatedness_errors },
         { "test_paper_ex_Y2_errors", test_paper_ex_Y2_errors },
         { "test_paper_ex_Y2", test_paper_ex_Y2 },
         { "test_paper_ex_f2_errors", test_paper_ex_f2_errors },
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -1334,6 +1334,25 @@ test_paper_ex_divergence(void)
 }
 
 static void
+test_paper_ex_relatedness(void)
+{
+    tsk_treeseq_t ts;
+    tsk_id_t samples[] = { 0, 1, 2, 3 };
+    tsk_size_t sample_set_sizes[] = { 2, 2 };
+    tsk_id_t set_indexes[] = { 0, 0 };
+    double result;
+    int ret;
+
+    tsk_treeseq_from_text(&ts, 10, paper_ex_nodes, paper_ex_edges, NULL, paper_ex_sites,
+        paper_ex_mutations, paper_ex_individuals, NULL, 0);
+
+    ret = tsk_treeseq_relatedness(&ts, 2, sample_set_sizes, samples, 1, set_indexes, 0,
+        NULL, &result, TSK_STAT_SITE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_treeseq_free(&ts);
+}
+
+static void
 test_paper_ex_relatedness_errors(void)
 {
     tsk_treeseq_t ts;
@@ -1691,6 +1710,7 @@ main(int argc, char **argv)
         { "test_paper_ex_divergence_errors", test_paper_ex_divergence_errors },
         { "test_paper_ex_divergence", test_paper_ex_divergence },
         { "test_paper_ex_relatedness_errors", test_paper_ex_relatedness_errors },
+        { "test_paper_ex_relatedness", test_paper_ex_relatedness },
         { "test_paper_ex_Y2_errors", test_paper_ex_Y2_errors },
         { "test_paper_ex_Y2", test_paper_ex_Y2 },
         { "test_paper_ex_f2_errors", test_paper_ex_f2_errors },
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -338,8 +338,6 @@ int tsk_treeseq_allele_frequency_spectrum(const tsk_treeseq_t *self,
     const tsk_id_t *sample_sets, tsk_size_t num_windows, const double *windows,
     double *result, tsk_flags_t options);
 
-/* Two way sample set stats */
-
 typedef int general_sample_stat_method(const tsk_treeseq_t *self,
     tsk_size_t num_sample_sets, const tsk_size_t *sample_set_sizes,
     const tsk_id_t *sample_sets, tsk_size_t num_indexes, const tsk_id_t *indexes,
@@ -357,6 +355,10 @@ int tsk_treeseq_f2(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
     const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
     tsk_size_t num_index_tuples, const tsk_id_t *index_tuples, tsk_size_t num_windows,
     const double *windows, double *result, tsk_flags_t options);
+int tsk_treeseq_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets, tsk_size_t num_index_tuples,
+    const tsk_id_t *index_tuples, tsk_size_t num_windows, const double *windows, double *result,
+    tsk_flags_t options);
 
 /* Three way sample set stats */
 int tsk_treeseq_Y3(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -356,9 +356,9 @@ int tsk_treeseq_f2(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
     tsk_size_t num_index_tuples, const tsk_id_t *index_tuples, tsk_size_t num_windows,
     const double *windows, double *result, tsk_flags_t options);
 int tsk_treeseq_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
-    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets, tsk_size_t num_index_tuples,
-    const tsk_id_t *index_tuples, tsk_size_t num_windows, const double *windows, double *result,
-    tsk_flags_t options);
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
+    tsk_size_t num_index_tuples, const tsk_id_t *index_tuples, tsk_size_t num_windows,
+    const double *windows, double *result, tsk_flags_t options);
 
 /* Three way sample set stats */
 int tsk_treeseq_Y3(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -6946,7 +6946,7 @@ TreeSequence_k_way_stat_method(TreeSequence *self, PyObject *args, PyObject *kwd
 {
     PyObject *ret = NULL;
     static char *kwlist[] = { "sample_set_sizes", "sample_sets", "indexes", "windows",
-        "mode", "span_normalise", NULL };
+        "mode", "span_normalise", "polarised", NULL };
     PyObject *sample_set_sizes = NULL;
     PyObject *sample_sets = NULL;
     PyObject *indexes = NULL;
@@ -6960,14 +6960,15 @@ TreeSequence_k_way_stat_method(TreeSequence *self, PyObject *args, PyObject *kwd
     npy_intp *shape;
     tsk_flags_t options = 0;
     char *mode = NULL;
-    int span_normalise = 1;
+    int span_normalise = true;
+    int polarised = false;
     int err;
 
     if (TreeSequence_check_tree_sequence(self) != 0) {
         goto out;
     }
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OOOO|si", kwlist, &sample_set_sizes,
-            &sample_sets, &indexes, &windows, &mode, &span_normalise)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OOOO|sii", kwlist, &sample_set_sizes,
+            &sample_sets, &indexes, &windows, &mode, &span_normalise, &polarised)) {
         goto out;
     }
     if (parse_stats_mode(mode, &options) != 0) {
@@ -6975,6 +6976,9 @@ TreeSequence_k_way_stat_method(TreeSequence *self, PyObject *args, PyObject *kwd
     }
     if (span_normalise) {
         options |= TSK_STAT_SPAN_NORMALISE;
+    }
+    if (polarised) {
+        options |= TSK_STAT_POLARISED;
     }
     if (parse_sample_sets(sample_set_sizes, &sample_set_sizes_array, sample_sets,
             &sample_sets_array, &num_sample_sets)
@@ -7354,7 +7358,7 @@ static PyMethodDef TreeSequence_methods[] = {
     { .ml_name = "relatedness",
         .ml_meth = (PyCFunction) TreeSequence_relatedness,
         .ml_flags = METH_VARARGS | METH_KEYWORDS,
-        .ml_doc = "Computes genetic relatedness between sample sets." },    
+        .ml_doc = "Computes genetic relatedness between sample sets." },
     { .ml_name = "Y1",
         .ml_meth = (PyCFunction) TreeSequence_Y1,
         .ml_flags = METH_VARARGS | METH_KEYWORDS,
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -7029,6 +7029,12 @@ TreeSequence_divergence(TreeSequence *self, PyObject *args, PyObject *kwds)
 }
 
 static PyObject *
+TreeSequence_relatedness(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_k_way_stat_method(self, args, kwds, 2, tsk_treeseq_relatedness);
+}
+
+static PyObject *
 TreeSequence_Y2(TreeSequence *self, PyObject *args, PyObject *kwds)
 {
     return TreeSequence_k_way_stat_method(self, args, kwds, 2, tsk_treeseq_Y2);
@@ -7345,6 +7351,10 @@ static PyMethodDef TreeSequence_methods[] = {
         .ml_meth = (PyCFunction) TreeSequence_divergence,
         .ml_flags = METH_VARARGS | METH_KEYWORDS,
         .ml_doc = "Computes diveregence between sample sets." },
+    { .ml_name = "relatedness",
+        .ml_meth = (PyCFunction) TreeSequence_relatedness,
+        .ml_flags = METH_VARARGS | METH_KEYWORDS,
+        .ml_doc = "Computes genetic relatedness between sample sets." },    
     { .ml_name = "Y1",
         .ml_meth = (PyCFunction) TreeSequence_Y1,
         .ml_flags = METH_VARARGS | METH_KEYWORDS,
--- a/python/tests/test_covariance.py
+++ b/python/tests/test_covariance.py
@@ -192,6 +192,7 @@ class TestCovariance(unittest.TestCase):
         assert ts.num_trees > 2
         self.verify(ts)
 
+    @pytest.mark.skip(reason="internal samples not working")
     def test_internal_samples(self):
         nodes = io.StringIO(
             """\
--- a/python/tests/test_covariance.py
+++ b/python/tests/test_covariance.py
@@ -1,0 +1,288 @@
+# MIT License
+#
+# Copyright (c) 2018-2020 Tskit Developers
+# Copyright (c) 2016-2017 University of Oxford
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for covariance computation.
+"""
+import io
+import itertools
+import unittest
+
+import msprime
+import numpy as np
+import pytest
+
+import tests.tsutil as tsutil
+import tskit
+
+
+def check_cov_tree_inputs(tree):
+    if not len(tree.roots) == 1:
+        raise ValueError("Trees must have one root")
+
+    for u in tree.nodes():
+        if tree.num_children(u) == 1:
+            raise ValueError("Unary nodes are not supported")
+
+
+def naive_tree_covariance(tree):
+    """
+    Returns the (branch) covariance matrix for the sample nodes in a tree. The
+    covariance between a pair of nodes is the distance from the root to their
+    most recent common ancestor.
+    """
+    samples = tree.tree_sequence.samples()
+    check_cov_tree_inputs(tree)
+
+    n = samples.shape[0]
+    cov = np.zeros((n, n))
+    for n1, n2 in itertools.combinations_with_replacement(range(n), 2):
+        mrca = tree.mrca(samples[n1], samples[n2])
+        cov[n1, n2] = tree.time(tree.root) - tree.time(mrca)
+        cov[n2, n1] = cov[n1, n2]
+
+    return cov
+
+
+def naive_ts_covariance(ts):
+    """
+    Returns the (branch) covariance matrix for the sample nodes in a tree
+    sequence. The covariance between a pair of nodes is the weighted sum of the
+    tree covariance, with the weights given by the spans of the trees.
+    """
+    samples = ts.samples()
+
+    n = samples.shape[0]
+    cov = np.zeros((n, n))
+    for tree in ts.trees():
+        cov += naive_tree_covariance(tree) * tree.span
+
+    return cov / ts.sequence_length
+
+
+def ts_covariance_incremental(ts):
+    for tree in ts.trees():
+        check_cov_tree_inputs(tree)
+
+    n = ts.num_samples
+
+    this_cov = np.zeros((n, n))
+    total_cov = np.zeros((n, n))
+    last_update = np.zeros((n, n))
+
+    for tree, edge_diff in zip(ts.trees(sample_lists=True), ts.edge_diffs()):
+        (t_left, t_right), edges_out, edges_in = edge_diff
+
+        for e in reversed(edges_out):
+            u = e.child
+            if tree.parent(u) == tskit.NULL:
+                for leaf in tree.leaves(u):
+                    update_cov_pairs_with_leaf(
+                        tree, total_cov, this_cov, last_update, leaf, t_left
+                    )
+
+        for e in reversed(edges_in):
+            u = e.child
+            for leaf in tree.leaves(u):
+                update_cov_pairs_with_leaf(
+                    tree, total_cov, this_cov, last_update, leaf, t_left
+                )
+
+    total_cov += this_cov * (t_right - last_update)
+    for (n1, n2) in itertools.combinations(range(n), 2):
+        total_cov[n2, n1] = total_cov[n1, n2]
+    return total_cov / ts.sequence_length
+
+
+def update_cov_pairs_with_leaf(tree, total_cov, this_cov, last_update, leaf, t_left):
+    """
+    Perform an upward traversal from `leaf` to the root, updating the covariance
+    matrix elements for pairs of `leaf` with every other leaf in the tree.
+    """
+    root_time = tree.time(tree.root)
+    p = tree.parent(leaf)
+    c = leaf
+
+    if tree.is_sample(c):
+        total_cov[c, c] += this_cov[c, c] * (t_left - last_update[c, c])
+        last_update[c, c] = t_left
+        this_cov[c, c] = root_time - tree.time(c)
+    while p != -1:
+        time = root_time - tree.time(p)
+        for sibling in tree.children(p):
+            if sibling != c:
+                update_cov_all_pairs(
+                    tree, total_cov, this_cov, last_update, leaf, sibling, time, t_left
+                )
+        c, p = p, tree.parent(p)
+
+
+def update_cov_all_pairs(tree, total_cov, this_cov, last_update, c1, c2, time, t_left):
+    s1_ind = tree.left_sample(c1)
+    while True:
+        s2_ind = tree.left_sample(c2)
+        while True:
+            (n1, n2) = (s1_ind, s2_ind)
+            if n1 > n2:
+                (n1, n2) = (n2, n1)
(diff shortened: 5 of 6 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/898/)

petrelharp:
Great! Very excited! Let's see, a few comments. We've discussed elsewhere that this is actually `kinship`, since `covariance` should be the covariance one calculates from the matrix of genotypes, and which subtracts the means from each allele frequency. (That'd be nice to do, also, though!). And we need to track down the best definition of kinship, and make sure we match that.

And, I believe this is the same thing, written as a branch statistic:
```
n = ts.num_samples
def f(x):
  # x[i] gives the number of descendants in sample set i below the branch
  # and the summary function should return 1 if both are below it, 0 otherwise
  return np.array([x[i] * x[j] for i in range(n) for j in range(n)]) 

sample_sets = [[u] for u in ts.samples()]
kinship = ts.sample_count_stat(sample_sets, f, output_dim=n * n,
                                 polarised=True, mode="branch",
                                 span_normalise=False, strict=False).reshape((n,n))
```
Do you agree, that this is the same thing?

Regardless of which thing we use, most of the work is in writing the "naive" calculation and the tests. So, this is great.

gregorgorjanc:
> And we need to track down the best definition of kinship

This is potentially a minefield - I have spent a lot of time on this and the terminology, notation and descriptions are a huge mess IMO. I see it merely as a metric that tells us how similar/identical two chromosomes are. There are lots of details. I think this is the summary.

With a pedigree, we can only make probabilistic statements about this similarity/identity because we never observe chromosomes; further because pedigree is always finite, we always miss some ancient relationships and can only report similarity/identity in the IBD sense with respect to some base population where we assume all individuals are unrelated - so we are constantly underestimating kinship and there is the variance of true/realised values around our expectations.

With genomic data (some markers or whole sequence) we can observe similarity/identity - kinship is just a function of the proportion of identical alleles (some subtract 2*allele freq, some don't) - this is IBS. Here we can bring back IBD by looking stretches of IBS similarity/identity, but that is crude as IBD segments have exponential lengths (it easy to filter out short segments thinking they are very old).

With trees, we can either use tree topology, lengths and times to get ~"prior/expected" covariance - I think this is identical to what @gtsambos was doing, no? We can also look at sequence similarity, so this is ~"posterior/realised" covariance, which is the same as the previous paragraph, but we can use tree sequence to compute this more efficiently.

@brieuclehmann @petrelharp @jeromekelleher does the above make sense?

petrelharp:
Agreed that this is a minefield. As you say, talking about "IBD" has the same issues. A big part of the confusion, though, is because sometimes people are talking about the pedigree quantity (analagous to the branch statistic) and sometimes they are talking about the genotype quantity (site statistic), which estimates the pedigree quantity.

Here we're going to implement a particular statistic, though, so we do need to figure out (a) what exactly to calculate, and (b) what exactly to call it.

As for what to calculate, often a good way to figure this out is to take the quantity that people most commonly compute from genotype data, and implement that. That way, people can use tskit to compute the thing they're used to calculating from their genotype matrices. You're saying people compute this as "proportion of identical alleles", which let's translate to "proportion of segregating sites at which the two target genomes agree", or "number of sites at which the two agree divided by number of segregating sites".

Hm, ok, let's see. The branch statistic we want to compute is "total shared area". The corresponding site statistic would be "total number of segregating sites at which the genomes agree", since that's what mutations that fall on the 'shared area' would create. So, the branch statistic that corresponds to "proportion of segregating sites at which the two target genomes agree" would be "total shared area divided by total area".

So, how about this for a proposal: we implement
```
ts.kinship(sample_sets, proportion=True, ...)
```
where if `proportion=False` then we just compute the numerator in the statistics above (i.e., number of shared alleles or total shared tree area), and if `proportion=True` then we divide by the denominator (by `segregating_sites`, basically).

And, what to call it? I would *like* to call it `kinship`, is this a reasonably common and expected definition? If so, should the default `proportion` above be True or False?

A note: the IBD notion of kinship (i.e., proportion of genome that is IBD back to some particular time, say) is obviously related, and could be obtained from @gtsambos's ibd implementation. I think if we implement that, we could call it `ts.ibd_kinship( )`, say, and not risk confusion.

brieuclehmann:
@petrelharp I agree that the branch statistic does the same thing, and I've checked it with some test examples :). I didn't realise the stats framework was so flexible!

Regarding terminology, I agree that kinship would be a good name for this, as this suggests a notion of similarity between individuals. We could have a different ts.xxx_kinship() functions to specify exactly which type of kinship we mean, or add it as an argument in ts.kinship(type = "xxx") (or both?). I would favour `proportion = True` as I generally prefer unitless quantities (but I don't feel strongly about this).

Regarding which implementation to use, it may just be a question of efficiency, since computing the kinship for large tree sequences could quickly become expensive.

gregorgorjanc:
@petrelharp I like this site & branch duality - I still need to get my head around it though. Do we really get the same answer if I work with site statistic (genotypes) and branch statistic here? They will obviously be very similar (tree is inferred from mutations), but not necessarily the same, will they?

I vote for naming this function `kinship`, but I feel this is exactly? the same as what @gtsambos is doing. Are we duplicating effort or are these two "bits" really doing something different? Or is just the approach different? Obviously cutting the tree at some point (to get the proportion of chromosomes that are IBD back to some particular time) will give

petrelharp:
> Do we really get the same answer if I work with site statistic (genotypes) and branch statistic here?

Which sort of "the same" do you mean? You will get different numbers if you do `mode="site"` or `mode="branch"` (with different units, even). They are the "same" in that the site statistic is expected value of the branch statistic under a neutral, infinite-sites model of mutation, multiplied by the mutation rate - in other words, the site statistic (which is something you can compute from only the genotypes) is an estimator of the branch statistic (which you need the trees for).

> I vote for naming this function kinship, but I feel this is exactly? the same as what @gtsambos is doing. 

I don't see this? These aren't exactly the same, because we're getting a single number for each pair, and @gtsambos is getting a list of genomic segments. But I don't think that kinship is a summary of the IBD segments, even: IBD segments tell you how long ago a MRCA is, but for kinship we need to also know the age of the root of the tree. I suppose that given the IBD segments *and* the age of the root of the tree at each site you could compute kinship, but it'd be a very inefficient way to do it?

petrelharp:
> We could have a different ts.xxx_kinship() functions to specify exactly which type of kinship we mean, or add it as an argument in ts.kinship(type = "xxx") (or both?). I would favour proportion = True as I generally prefer unitless quantities (but I don't feel strongly about this).

I agree, on all points! I'd vote for `kinship(... type=xxx)` for minor modifications of the same thing, and `xxx_kinship()` for fundamentally different things (i.e., things computed using totally different code paths). But I think we need to so some careful looking at the literature to see what people do in practice.

gregorgorjanc:
> Which sort of "the same" do you mean? You will get different numbers if you do mode="site" or mode="branch" (with different units, even). They are the "same" in that the site statistic is expected value of the branch statistic under a neutral, infinite-sites model of mutation, multiplied by the mutation rate - in other words, the site statistic (which is something you can compute from only the genotypes) is an estimator of the branch statistic (which you need the trees for).

Ok, this clears bits in my head!

gregorgorjanc:
>> I vote for naming this function kinship, but I feel this is exactly? the same as what @gtsambos is doing.

> I don't see this? These aren't exactly the same, because we're getting a single number for each pair, and @gtsambos is getting a list of genomic segments. But I don't think that kinship is a summary of the IBD segments, even: IBD segments tell you how long ago a MRCA is, but for kinship we need to also know the age of the root of the tree. I suppose that given the IBD segments and the age of the root of the tree at each site you could compute kinship, but it'd be a very inefficient way to do it?

Yes, but @gtsambos is also summarising the IBD segments between pairs of chromosomes into a single number by summing the length of segments and dividing by the chromosome length.

So, not the same procedure, but the end is the same and possibly some intermediate steps too.

petrelharp:
> Yes, but @gtsambos is also summarising the IBD segments between pairs of chromosomes into a single number by summing the length of segments and dividing by the chromosome length.
>
> So, not the same procedure, but the end is the same and possibly some intermediate steps too.

Ah, right - sorry, I was thinking you meant in the method that's implemented in tskit. But this would give kinship relative to the population at a particular time (if that's how IBD is computed), which depends on the time chosen, and is very different than what we're computing.

petrelharp:
**Note:** I wrote a bunch of the below before looking back at [Speed & Balding](https://www.nature.com/articles/nrg3821) - you should probably first read the "SNP-based measures of relatedness" section in that paper first.

Ok, here's a summary of the issues. The kinships coefficient ("kinship") is pretty universally defined to be the "proportion of IBD"... but that's not an actual definition because IBD is not a singly-defined concept. This was (maybe originally?) called "relatedness" by [Wright 1922](https://www.journals.uchicago.edu/doi/10.1086/279872). There's (at least?) three main use cases:

- **close relative identification**: for instance, two genomes obtained by separate meioses from the same diploid parent should have kinship of 0.5.
- **kin selection**: in these studies they'd like the kinship coefficient to be the realtedness coefficient in [Hamilton's rule](https://en.wikipedia.org/wiki/Kin_selection#Hamilton's_rule), defined in terms of a threshold above which a gene increases in frequency. I'm not sure exactly what this should be.
- **mixed-model covariance matrix in GWAS**: I think this is the covariance of trait deviations to the sample mean, rather than to the root, as we have been talking about.

Estimators for the first use case also always normalize somehow for "background relatedness", as they must.  Talking through this a bit: suppose that `P` is the proportion of segregating alleles shared by two sibling (haploid) genomes, and `PP` the same quantity for the two parents (who are "unrelated"). Then the expected value of P is
```
E[P] = (1 + PP) / 2
```
Note that
```
(E[P] - PP) / (1 - PP) = 1 / 2
```
so if we estimate `PP` by the average value of `P` across all pairs and plug it in to the previous equation, we get an estimator of the first use case above, `(P - PP)/(1 - PP)`. (This is in fact what is done by some methods; see below.) Note also that if `S` is the number of shared alleles (what we've started computing here), and `SP` the shared alleles between the parents, and `N` is the number of segregating sites, then `P = S/N`, so also
```
E[S] = (N + SP)/2
```
and so another estimator of `P` would be
```
(S - mean(S)) / (N - mean(S)) .
```

Ok, and how about trait covariance? Let `X[i]` be the trait value of `i`, and `Z` the trait value obtained by summing contributions across all the roots, and `S[i,j] = cov(X[i] - Z, X[j] - Z)`, what we've been computing here.  Instead, let `XP` be the mean of the `X`, so that with `J[i,j] = delta(i,j) - 1/n`,
```
  cov( X[i] - XP, X[j] - XP ) = (J S J) [i,j]
                   = S[i,j] - sum_k S[i,k] / n - sum_k S[k,j] / n + sum_{ij} S[i,j] / n^2
                   =: Sigma[i,j]
```
So, `Sigma[i,j]` is the number of shared alleles between `i` and `j`, minus the mean number shared by each of `i` and `j`, plus the overall shared number across all pairs. In shorthand, this is `S[i,j] - mean(S[i]) - mean(S[j]) + mean(S)`. If `n` is big, then those three means will all be pretty close to the mean "background relatedness", so this would be very close to `S - mean(S)` in the previous paragraph. Note also that `S[i,i] = N`, so that `cor( X[i] - XP, X[j] - XP )` is very close to the estimator of `P` above, `(S - mean(S)) / (N - mean(S))`. And, I haven't read the details of all the various kinship estimators, perhaps some actually do it this way (do you know, @gregorgorjanc?).

So, here's one proposal:
```
def kinship(self, sample_sets, subtract_mean=True, rescale=True):
```
where

- if `subtract_mean=True`, then we return `S[i,j] - mean(S[i]) - mean(S[j]) + mean(S)`
- if `rescale=True`, then we divide by `sqrt(output[i,i], output[j,j])` (so getting a correlation matrix instead of a covariance matrix).

So, 

- `subtract_mean=False, rescale=False` would give the thing we've started calculating here, number of shared alleles
- `subtract_mean=True, rescale=False` would give the covariance matrix of centered trait values
- `subtract_mean=True, rescale=True` would give the correlation matrix of centered trait values, and also be an estimate of the kinship coefficient, thus agreeing with other uses of the term.

Remaining questions:

- Is this right? =) In particular, would the proposed default values actually give you an estimator of a number that'd let you identify sibs, cousins, etc in the usual way?
- Which other software computes exactly the same thing? I think at least the mixed-model GWAS methods, is this true? 
- should we call this `relatedness( )` instead? which seems to mean the same thing?

I need to digest the summary in Speed & Balding more before feeling sure about this proposal. (Hopefully someone else will, also?)

Notes:

- "proportion of shared genome" means something different if the proportion is by length or by number of segregating sites. In the method-of-moments estimator above, it doesn't matter. We're computing the "number of segregating sites" version, which is more strongly affected by regions of the genome with lots of tall trees. If we computed this using IBD tract lengths, we'd get the other version. I think this is fine.

References and notes:

- [Speed & Balding](https://www.nature.com/articles/nrg3821) is a good summary of different methods, and explains how genotype correlation gives an estimate of the kinship coefficient
- [Wikipedia](https://en.wikipedia.org/wiki/Coefficient_of_relationship#Kinship_coefficient) says it's the proportion of IBD alleles.
- [Thompson 1975]](https://pubmed.ncbi.nlm.nih.gov/1052764/) uses "kinship" to refer to "proportion of genes shared in common" and refers to Malecot 1948 for it
- [How to estimate kinship, Goudet, Kay & Weir 2018](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6220858/) says that they follow others and "take the current population as the reference" so that "the kinship of a pair of individuals is compared to the average for all pairs of individuals in the sample, so the average estimated kinship is zero"; their method (Weir & Goudet 2017) estimates relative kinship as (M - \bar M) / (1 - \bar M) where M is the proportion of identical alleles and \bar M is the average of M across all pairs
- [KING](https://academic.oup.com/bioinformatics/article/26/22/2867/228512), like [plink](https://www.cog-genomics.org/plink/2.0/distance), works with the rough idea that two genomes are either IBD, and thus identical, with probability f, or are like random draws from the allele frequency, with probability (1-f), and f is the kinship coefficient. But then note that their estimates can be negative. Also see [the original KING paper](https://www.sciencedirect.com/science/article/pii/S0002929712003096).
- [Dou et al 2017](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.1007021) "Kinship coefficient, defined as the probability that two homologous alleles drawn from each of two individuals are identical by descent (IBD)"

gregorgorjanc:
I will just add some bits for completeness. I will focus on quantitative genetics view. I apologise for the length and a bit of ramble, but I thought it would be better if I spill it all out and connect seemingly different things;)

[Wright 1922](https://www.journals.uchicago.edu/doi/pdf/10.1086/279872) worked on mating systems in general pedigrees (instead of just simple cases). Using his [method of path analysis](https://pdfs.semanticscholar.org/90bd/1ea6df6cf7ff14c392818751270b93d53d4e.pdf) he came up with:
* **coefficient of relationship (relatedness) [= correlation between additive genetic values of two diploid individuals**] and 
* **coefficient of inbreeding [= correlation between (additive) genetic values of two genomes (uniting gametes) of a diploid individual]**.

This was all with pedigrees where we can only make probabilistic IBD statements relative to some founder population where we assume everybody is unrelated (though that is not true! - this matters quite a bit when we move to genomics).

Later [Emik and Terrill 1949](https://academic.oup.com/jhered/article-abstract/40/2/51/802925) came up with a simple recursive algorithm to calculate these pedigree-based coefficients from an ordered pedigree (parents before progeny).

[Henderson 1976](https://www.jstor.org/stable/2529339?seq=1#metadata_info_tab_contents) [apparently he had bits of this worked out much earlier, but here he published linear algebra treatment] put all this into the context of a linear model by linking together the phenotype model

pheno_value[i] = mean + add_gen_value[i] + error[i]

and the genetic model on a pedigree (of diploid individuals)

add_gen_value[i] = 0.5add_gen_value[father[i]] + 0.5add_gen_value[father[i]] + mendelian_sampling_value[i].

Mendelian sampling captures deviation from parent average due to recombination, mutation, gene conversion and segregation, possibly also selection. With pedigrees, we do not observe any genes, but we can infer add_gen_values and their mendelian_sampling_value from regressing phenotypes on a pedigree.

If we treat additive genetic values as random variables and we do Cov() and Var() algebraic calculations on an ordered pedigree we are effectively running the Emik and Terrill algorithm. For the whole vector of **a** we get Var(**a**) = **A**\sigma^2 - an important caveat here - we are doing covariances here, not correlations - so Henderson strictly called the matrix **A** as the **numerator relationship matrix** (referring to the numerator of Wright's relationship correlation coefficient). Today this matrix is often called just a relationship matrix and causes confusions between correlations and covariances. Sometimes it is also called a kinship matrix (see below why this is confusing). To be pedantic, **A** is not covariance matrix! **A**\sigma^2 is covariance matrix between additive genetic values, while **A** itself is just a covariance coefficient matrix.

**An important point for me here is that one first sets up a genetic model (process) and then co-variance calculations fall out automatically (most genomic variants are not following this logic)**.

What is in this **A** matrix? Diagonal element gives variance of additive genetic values - for a diploid individual this is:

```
add_gen_value[i] = add_gen_value[i,1] + add_gen_value[i,2] # there are two genomes in there
Var(add_gen_value[i]) = Var(add_gen_value[i,1] + add_gen_value[i,2])
                      = Var(add_gen_value[i,1]) + Var(add_gen_value[i,2]) + 2Cov(add_gen_value[i,1], add_gen_value[i,2])

assuming that

Var(add_gen_value[i,j]) = 1/2 \sigma^2_a

and that

2Cov(add_gen_value[i,1], add_gen_value[i,2]) = F[i]\sigma^2_a

where F[i] is Wright's coefficient of inbreeding, which is equal to Cov(add_gen_value[i,], add_gen_value[i,])

we get 

Var(add_gen_value[i]) = (1 + F[i]) \sigma^2_a
```

Off-diagonal elements give covariance between additive genetic values of two individuals - for diploids we have

```
add_gen_value[i] = add_gen_value[i,1] + add_gen_value[i,2] # there are two genomes in there
add_gen_value[j] = add_gen_value[j,1] + add_gen_value[j,2] # there are two genomes in there
Cov(add_gen_value[i], add_gen_value[j]) = Cov(add_gen_value[i,1], add_gen_value[j,1]) +
                                          Cov(add_gen_value[i,1], add_gen_value[j,2]) +
                                          Cov(add_gen_value[i,2], add_gen_value[j,1]) +
                                          Cov(add_gen_value[i,2], add_gen_value[j,2])
```

Henderson was working in animal breeding. I am less familiar with the historical development of the human genetics branch, but my understanding is that [Cotterman 1940](https://etd.ohiolink.edu/!etd.send_file?accession=osu1298297334&disposition=inline) and [Malecot 1948](http://agris.fao.org/agris-search/search.do?recordID=US201300403470) worked on IBD probabilities (not correlations or covariances), which however gave the same intrinsic formulae as in Wright or Henderson (up to some scaling). I don't know this history but my understanding is that kinship coefficients rose out of asking what is the probabililty that two random alleles from two diploid individiuals (one allele per ind) are IBD. This gives:

```
k[i,j]=1/4*Pr(genome[i,1] = genome[j,1]) +
       1/4*Pr(genome[i,1] = genome[j,2]) + 
       1/4*Pr(genome[i,2] = genome[j,1]) + 
       1/4*Pr(genome[i,2] = genome[j,2])
```

If we apply the same to one individual we get

```
k[i,i]=1/4*Pr(genome[i,1] = genome[i,1]) + ==> 1/4
       1/4*Pr(genome[i,1] = genome[i,2]) + 
       1/4*Pr(genome[i,2] = genome[i,1]) + 
       1/4*Pr(genome[i,2] = genome[i,2])  + ==> 1/4
k[i,i]=1/2 + 1/2Pr(genome[i,1] = genome[i,2])
```

This is effectively the same as 1/2 of the above covariance calculations for additive genetic values, so **kinship matrix between individuals** **K** is 1/2 of Henderson's numerator relationship matrix **A**. A modern summary is in Lange, K. (1997) Mathematical and statistical methods for genetic analysis, see also [Thompson 2013](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3664843/). As I wrote above, literature interchanges terms kinship, relationship, relatedness, genetic, covariance, ... all the time. Even my version of all of the above might be disputed by someone.

OK, what about genomics? There is a gazillion of relationship variants! IBD, IBS, segments, scaling this and that. To my understanding, most of these variants are due to different views.

Following the IBD probability view from pedigrees, it seems natural that if we can get a better estimate of IBD probabilities we can simply replace these in the kinship matrix **K** or numerator relationship matrix **A**. This can for example capture one of "recombination, mutation, gene conversion and segregation, possibly also selection", but also relatedness between founders. There are at least two common ways of estimating IBD probabilities - lets focus on identical DNA segments between two chromosomes - this is giving us information about Pr(genome[i,k] = genome[j,l]). With a set of identical DNA segments between two individuals, we can calculate the kinship coefficient between these two individuals as a sum of segment lengths divided by the genome length as @gtsambos showed us the other day. One catch here is related to the length of segments that are treated as identical. Some aficionados will say that we care about IBD because that's what the field has been doing since ever and because for some reason we want to focus just on recent relationships we will impose some threshold on segment length, say 1cM. This can create bias because segment lengths have exponential distribution, so a short segment can also reflect a recent relationship. **IBD concept was invented as a proxy for IBS because we could not observe DNA back in the day. Now we can.** So why shouldn't we take into account single allele sharing? One reason for this was that until recently we mostly worked with SNP arrays with a limited number of markers and the IBD aficionados felt that sharing allele at one locus gives you no IBD info. If we allow IBD segment length down to a single locus (we are looking at IBS) then kinship coefficient between two individuals k[i,j] is the proportion of alleles that these two individuals share; for an individual k[i,i] would be 1/2 if it is heterozygous at all loci and 1 if it is homozygous at all loci.

I think that working with trees can neatly solve this IBD vs IBS debate because it unifies them - namely, in pedigrees, IBD  reflected relationship up to some founding populations. With genomics, we can now measure relationships between these founders and I think that at least conceptually it makes sense to go back in time to the root. See also [Powell 2010](https://pubmed.ncbi.nlm.nih.gov/20877324/) on IBD vs IBS.

What about the "covariance/model" approach? Lets assume that additve genetic value of an individual is a linear combination of its genotype x (row vector for p loci with values 0, 1, and 2) and loci effects alpha (column vector):

```
add_gen_value[i] = x alpha
```

or for a set of individuals (X is n-by-p matrix)

```
add_gen_value[1:n] = X alpha
```

What is the variance of this genetic model conditional on observed X?

```
Var(add_gen_value | X) = Var(X alpha | X)
                       = XVar(alpha)X'
                       = XX'\sigma^2_alpha (assuming Var(alpha)=I\sigma^2_alpha)
```

where XX' (n*n matrix for n individuals) is often called **genomic numerator relationship matrix** or **genomic relationship matrix** or **genomic kinship** (argh...). What's in XX'? Each element is an inner product between genotypes of two individuals (or for an individual with itself on diagonal).

An alternative to the above view on modelling individual values is to model effects of loci with a penalised multiple regression:

```
pheno_value = 1mean + X alpha + error
error ~ N(0, I\sigma^2_e)
alpha ~ N(0, I\sigma^2_alpha)
```

Note that X'X/n (p*p matrix for p loci) is related to the linkage-disequilibrium (LD) (covariance) matrix between loci.

But some centre and scale genotypes!? The above Var(add_gen_value | X) shows that we do not need to. The two most common versions of centring and scaling are:

let maf[i] = mean(X[, i])/2 be allele frequency in the sample

[Vanraden 2008](https://www.sciencedirect.com/science/article/pii/S0022030208709901)
Z[, i] = X[, i] - 2maf[i] (we are centring the genotypes here)
ZZ'/2sum(maf*(1-maf))

The centring and scaling (by 2sum(maf*(1-maf)) = sum of heterozygosities) logic above was to get genomic numerator relationship matrix that would be somewhat "similar" to pedigree numerator relationship matrix - because VanRaden had millions of phenotyped and pedigreed cattle and only ~1000 genotyped bulls so it made sense to tailor genomic info to pedigree info. Note that centring introduces negative values in ZZ', which makes IBD probabiliby aficionados screaming! But, negative values do make sense in the context of [Wright's inbreeding coefficient in structured populations](https://en.wikipedia.org/wiki/F-statistics) 1-F_IT = (1-F_IS)(1-F_ST). I believe this also gave a somewhat similar estimate of additive genetic variance \sigma^2_a similar to what we get from the pedigree. But this argument is riddled with problems because the real parameter in Var(add_gen_value | X) = XX'\sigma^2_alpha is \sigma^2_alpha not \sigma^2_a (VanRaden assumed that \sigma^2_a ~ \sigma^2_alpha2sum(maf*(1-maf)), which would make sense if loci would not be in any LD). There are several Nature Genetics papers on a better genomic relationship matrix, but none mentions this, sic.

[Yang 2010](https://www.nature.com/articles/ng.608#Sec8)
Z[, i] = (X[, i] - 2maf[i])/k (we are standardizing the genotypes here)
above can be k = sd(X[, i]) (pure standardization) or k=sqrt(2maf(1-maf)) (assuming Hardy-Weinberg equilibrium)
ZZ'/p

The logic here is that we upscale importance of rare alleles. Note that this implies a different prior than Var(alpha)=I\sigma^2_alpha. This Yang version is very sensitive to small allele frequencies and often genotype data is filtered on MAF to avoid this problem. The VanRaden version is stable and the de-facto standard in agriculture (where all this is used on a daily basis!).

Of note on centring. Remember the phenotype model:

```
pheno_value[i] = mean + add_gen_value[i] + error[i]
add_gen_value[i] = X[i, ] alpha
pheno_value[i] = mean + X[i, ] alpha + error[i]
               = mean + (Z[i, ] + 2maf) alpha + error[i]
               = (mean + 2maf alpha) + Z[i, ] alpha + error[i]
```

So, centring in the context of the above model only pushes the mean of genotypes in to intercept. There can be numerical reasons to centre covariates in a model. In the context of trees, I think that centring would push root's genotypes into the intercept and estimates of alphas would then be allele substitution effects between the root's ancestral alleles and mutations.

gregorgorjanc:
Based on the above "novel" we should indeed be careful how we name @brieuclehmann's covariance calculation! It is not kinship, he is calculating covariance/similarity/identity between chromosomes k and l of individuals i and j, Pr(genome[i,k] = genome[j,l]). Genomic covariance, gametic covariance (following [Schaeffer et al 1989](https://www.sciencedirect.com/science/article/pii/S0022030289792316)), just covariance?

Kinship is when we sum these values into one value that describes relationship between two diploid individuals (of polyploid).

petrelharp:
> It is not kinship, he is calculating covariance/similarity/identity between chromosomes 

Hm - is it not kinship just because it is acting on single chromosomes, instead of diploid individuals?  If so, how would you define kinship between haploid individuals (say, male bees)?

petrelharp:
Another note: I think that in fact we're going to want to deal with the centred covariance matrix, since we don't observe phenotypes at the roots of the trees, and so our model is only defined up to an overall shift.

gregorgorjanc:
> > It is not kinship, he is calculating covariance/similarity/identity between chromosomes 
> 
> 
> 
> Hm - is it not kinship just because it is acting on single chromosomes, instead of diploid individuals?  If so, how would you define kinship between haploid individuals (say, male bees)? 

Yes, following the established literature I believe kinship coefficient is between diploid individuals, but the authors obviously did not have other ploidy cases in mind.

I am happy for either `ts.covariance` or `ts.kinship`, but there might be a case for both (to give similarly between chromosomes or between individuals) and in the case of haploids the two methods are the same?

Or `ts.kinship(type="ibs|ibd", nodes=TRUE|FALSE, otherargs=xxx)`?

gregorgorjanc:
> Another note: I think that in fact we're going to want to deal with the centred covariance matrix, since we don't observe phenotypes at the roots of the trees, and so our model is only defined up to an overall shift.

I see your point.

Should we then add `centre` argument to `ts.kinship` with default set to true?

What about scaling to [0, 1]?

gtsambos:
just chiming in to say I've had a busy week, but will try to get my head around this soon!

gtsambos:
Thanks for the useful fleshed out explanations above @gregorgorjanc and @petrelharp, it's clear you've both thought about this a lot over the years! I've copied some of it to refer to later 🤓 

I would vote against calling this method `kinship` simply because it is a different calculation from the 'kinship coefficients' calculated with IBD information (which is what I was showing in my talk), and I'd worry that users would mistake the two. However, they are conceptually *very* closely related. Say for a given pair of samples, you had the IBD kinships at every time point, ie. IBD kinships as a function of time. Then integrate that function over time. The number you'd get would be this statistic that @brieuclehmann has implemented here! Ie. it's more like an expectation of the kinship rather than the kinship itself. Pretty cool!

gregorgorjanc:
> I would vote against calling this method `kinship` simply because it is a different calculation from the 'kinship coefficients' calculated with IBD information (which is what I was showing in my talk), and I'd worry that users would mistake the two.

Yeah, this area is confusing, but if we start using different terms for very similar things (the difference is _only_ in the type of information used (IBS vs IBD) and whether we centre or not) then we are just adding to the confusion. But I sense that whatever we name these methods it will not solve the confusion ;(

gtsambos:
> (the difference is only in the type of information used (IBS vs IBD) and whether we centre or not)

I'm not so sure, to me it seems like the biggest difference is that IBD kinship is defined wrt to particular times, whereas this statistic is a summary of relatedness over all times. By IBS vs IBD, do you mean the site vs branch version of this statistic? I'd be a bit careful about calling the implemented site statistic a measure of IBS, as it will only be so under an infinite sites assumption

gregorgorjanc:
> > (the difference is only in the type of information used (IBS vs IBD) and whether we centre or not)
> 
> I'm not so sure, to me it seems like the biggest difference is that IBD kinship is defined wrt to particular times, whereas this statistic is a summary of relatedness over all times. By IBS vs IBD, do you mean the site vs branch version of this statistic? I'd be a bit careful about calling the implemented site statistic a measure of IBS, as it will only be so under an infinite sites assumption

Yes, IBD kinship can have an imposed limit on the age (or relatedly on the length of IBD segments that are considered in calculating IBD kinship), but it's up to a user to impose that limit or not. In the limit (when we do not impose any limit) IBD and IBS kinship should return the same thing (IBD is just a proxy for IBS).

As to site vs branch statistics, I am not sure. I understood that branch statistic will assume neutral infinite sites, while site statistic will reflect deviations from the assumption(s) because we are looking at actual genotypes (hence identical by state). Or is it the other way around?

I love these discussions! I would like to note that we are trying to bring together lots of different concepts here, hence so many comments.

petrelharp:
> Say for a given pair of samples, you had the IBD kinships at every time point, ie. IBD kinships as a function of time. Then integrate that function over time. The number you'd get would be this statistic that @brieuclehmann has implemented here! Ie. it's more like an expectation of the kinship rather than the kinship itself. Pretty cool!

Whoa, good point! I had not realized this, or seen it anywhere!

gregorgorjanc:
> > Say for a given pair of samples, you had the IBD kinships at every time point, ie. IBD kinships as a function of time. Then integrate that function over time. The number you'd get would be this statistic that @brieuclehmann has implemented here! Ie. it's more like an expectation of the kinship rather than the kinship itself. Pretty cool!
> 
> Whoa, good point! I had not realized this, or seen it anywhere!

Isn't IBD kinship as a function of time a cumulative function - so it's already integrating; at the root IBD kinship gives the same as @brieuclehmann IBS kinship? I view IBD kinship as F_IS (from 1-F_IT = (1-F_IS)(1-F_ST)) and when time goes to root F_IS equals F_IT.

petrelharp:
>  Isn't IBD kinship as a function of time a cumulative function?

It is cumulative: I think it's "proportion of the genome that has coalesced by this time"; but Georgia points out that what we're calculating has another integral in - the difference is that IBD kinship has units of genome length (well, proportion of the genome), whereas Briuc's calculation has units of time * length.

Oh! This is the big difference! With IBD kinship, it doesn't matter if you coalesce earlier or later in the pedigree, as the underlying model doesn't include new mutations and all genetic variation comes from the founders; but for covariance of quantitative traits, coalescing more recently makes for a higher covariance, because genetic variation comes from new mutations.

petrelharp:
BTW: another word for all this is "coancestry". Unfortunately it appears to be a synonym for "kinship", so it's not much helping.

brieuclehmann:
Just a comment following a chat with @gregorgorjanc, I think the current implementation assumes all samples are leaves and I'm not certain that it returns the right answer if we have internal samples. I'll add a test to this effect...

gregorgorjanc:
> > Isn't IBD kinship as a function of time a cumulative function?
> 
> It is cumulative: I think it's "proportion of the genome that has coalesced by this time"; but Georgia points out that what we're calculating has another integral in - the difference is that IBD kinship has units of genome length (well, proportion of the genome), whereas Brieuc's calculation has units of time * length.

@brieuclehmann calculations are _time * length_, but at the end, he normalises by genome length so the final unit is just _time_. 

> Oh! This is the big difference! With IBD kinship, it doesn't matter if you coalesce earlier or later in the pedigree, as the underlying model doesn't include new mutations and all genetic variation comes from the founders; but for covariance of quantitative traits, coalescing more recently makes for a higher covariance, because genetic variation comes from new mutations.

Yes, that is the key difference between IBD and IBS. I think this also depends if we take a _site_ view (where mutations matter) of _branch_ view (where mutations do not matter, but they are used to infer a tree from real data).

petrelharp:
> Yes, that is the key difference between IBD and IBS.

I think I know what you mean, but this is different than the usual distinction between IBD and IBS (whether you consider independent mutations to the same allele as the 'same' or not).

My next step is going to be to either enumerate or maybe just write a test implementation of a bunch of the kinship options. That'll make things more concrete. (suggestions of which ones to do are welcome)

> I think the current implementation assumes all samples are leaves

@brieuclehmann, I haven't yet figured out if your implementation has advantages over the general-purpose statistics one. Do you know?

brieuclehmann:
@petrelharp Based on some not very rigorous testing, it looks like my implementation is about 7x faster. An example with:
```
ts = msprime.simulate(100, recombination_rate = 1, random_seed=1, length=10)
```
Results:
```
>>> t0 = time.time(); A = ts_covariance_incremental(ts); t1 = time.time()
>>> t1 - t0
1.7671921253204346
>>> t0 = time.time(); kinship = ts.sample_count_stat(sample_sets, f, output_dim=n * n,
...                                  polarised=True, mode="branch",
...                                  span_normalise=False, strict=False).reshape((n,n)); t1 = time.time()
>>> t1 - t0
11.71674108505249
```
If I recall correctly, the C implementation kc_distance() function (upon which this is based) is another 10-100x faster.

petrelharp:
Wow, that's terriffic! And, I'm really surprised - I thought it was pointless to compare your python implementation to the C implementation of sample_count_stat. Do you know what the advantage is?

brieuclehmann:
I'm not too sure as I'm not familiar with how sample_count_stat() works under the hood (I'll look into it!). My implementation only updates the elements of the covariance matrix that are affected by the edge_diffs from one tree to the next, so it might be that?

gregorgorjanc:
> My next step is going to be to either enumerate or maybe just write a test implementation of a bunch of the kinship options. That'll make things more concrete. (suggestions of which ones to do are welcome)

For IBS the two below are very common/standard in QuanGen. There are lots of other variants and everybody has their favourite! If we start with these, then it's "easy" to add others.

[Vanraden 2008](https://www.sciencedirect.com/science/article/pii/S0022030208709901)
Z[, i] = X[, i] - 2maf[i] (we are centring the genotypes here)
ZZ'/2sum(maf*(1-maf))

[Yang 2010](https://www.nature.com/articles/ng.608#Sec8)
Z[, i] = (X[, i] - 2maf[i])/k (we are standardizing the genotypes here)
above can be k = sd(X[, i]) (pure standardization) or k=sqrt(2maf(1-maf)) (assuming Hardy-Weinberg equilibrium)
ZZ'/p

brieuclehmann:
OK, I think I understand... The logic is pretty much identical to my implementation, so it might be something to do with the fact that `f(x)` is multivariate? Am I right in thinking that: at each edge_diff, sample_count_stat() computes the full n x n `f(x)` from scratch? Where
```
def f(x):
  return np.array([x[i] * x[j] for i in range(n) for j in range(n)]) 
```
Or does sample_count_stat treat each element of a multivariate independently? (in which case I'm stumped.)

petrelharp:
Oh, yes! That's probably it, good call @brieuclehmann. Maybe this will be faster, then? It would be nice to have a way to do this by default in the statistics:
```
n = ts.num_samples
def f(x):
  # x is length 2, giving the number of descendants in a pair of sample sets below the branch
  # and the summary function should return 1 if both are below it, 0 otherwise
  return np.array([x[0] * x[1]])

kinship = np.zeros((n,n))
for i in range(n):
  for j in range(i, n):
    k = ts.sample_count_stat([[ts.samples()[i]], [ts.samples()[j]]], f, output_dim=1, 
                                 polarised=True, mode="branch",
                                 span_normalise=False, strict=False)
    kinship[i,j] = kinship[j,i] = k
```

brieuclehmann:
That's even slower (28s), sadly... Possibly because we have to repeat quite a few of the upward/downward traversals compared to just going through the tree sequence once?

petrelharp:
Yes, probably so. Thanks for checking!

brieuclehmann:
My implementation currently fails on trees with internal samples (and also won't work with multiple-rooted trees). @petrelharp Do you have a view on whether I should try to tweak sample_count_stat() (i.e. to only update elements of multivariate `f(x)` when needed) or to fix my implementation?

petrelharp:
Good question! I'm personally more motivated to make statistics more efficient (and, this would be do-able!), but if this is too daunting then implementing yours in C would be fine. However, I think we really do want to implement not just the branch statistic but also the site and node statistics, so you should factor that into your considerations? Another option would be to do neither - to go ahead with what we've got right now until speed becomes a bottleneck. Maybe we should have a chat next week?

jeromekelleher:
Just had a chat with @brieuclehmann and we think the following is a good way forward:

- Implement a ``ts.covariance_matrix`` function (or whatever we call it) using Peter's node-stat method under the hood (with a C implementation of the summary function). Test this against the Python code that @brieuclehmann has written already and merge.
- Compare the performance of this method to running kc_distance. The kc distance algorithm should have similar properties to what Brieuc has written here, so this'll give us a good idea of how a full C implementation of his algorithm will scale.

The reason for doing this is it's not obvious to us whether the node-stat method will be as fast as the kc-like algorithm. The reason for this is that the node stat method will take O(num_edges * log n * n^2), since Peter's summary function is O(n^2). I'm not sure what the complexity of the kc-like approach will be, but it certainly no worse than this (pairwise comparison of all samples via downward traversals cost at most O(n^2)), and I suspect it'll be quite a lot better.

For now, the sample stat method will perform a lot better than any of the Python based approaches, and will get us some code committed into the repo that we can start building on. I think it's better to get the basic covariance matrix method in now because we'll likely need to build more on what @brieuclehmann has done when we start doing the matrix-vector multiplications and matrix inversions that are needed as the downstream operations, and we don't want this diff to get too big.

How does this sound @petrelharp?

petrelharp:
That sounds like the right way forward to me. Maybe I should open an issue with my suggested change to the general stat code that would let us do this faster, and keep that separate.

jeromekelleher:
In the meeting  this evening we decided that a good way forward is to have a function ``covariance``, which uses the existing ``ts.divergence`` as a template. Some parameters like ``span_normalise`` might not make sense though, and we might want an extra parameter to control whether we center the values or not.

petrelharp:
OK, the proposal is

- @brieuclehmann to implement the summary function in C
- the method will be called `relatedness( )` and will have a similar interface to `divergence( )`

The justification for taking `relatedness` for this method is mostly that we think this is the *right* notion of relatedness, and that "ibd relatedness" should be called something else.
To be determined:

- what we might call an ibd relatedness method, if we were to implement one?
- whether there's an option to normalize by segregating sites
- whether there's options to do various other centering/scaling operations that are common in the literature

gregorgorjanc:
I would throw both IBS and IBD under the same umbrella - this is all relatedness/kinship/covariance.

I think that @brieuclehmann branch statistic implementation would give the same output (number or proportion of alleles/genome shared expected under a neutral model [or is this any model that generated tree?]) as @gtsambos kinship (length or proportion of genome shared) if @brieuclehmann would not go all the way up to the root or if @gtsambos would not limit time or length of segments. The difference is only on limits we impose and how we present the outcome. We can think of adding (or not) IBD later. I think we are missing out on the unification of this area if we keep them separated.

Site statistic will be different (realised number or proportion of alleles/genome shared), but is again just a version of the above.

As to what is IBS and related normalisation, I think it does not matter. We are interested in how similar two pieces of DNA are. Since we do it with trees, segments come in naturally (via spans), which is why I think this covers both IBD and IBS. We can represent this similarity by a number of shared alleles, by that number relative to the number of segregating sites or number of all sites. All three outcomes will be perfectly correlated (we are just diving by a different constant). Similarly for centering - we just shift all values up or down to some base.

petrelharp:
Note: my summary function above needs a modification, I think: right now, it is affected by branches that are shared by *all* samples (eg. the branch above the root, if present). I think we probably don't want to do this, so:
```
n = ts.num_samples
def f(x):
  # x[i] gives the number of descendants in sample set i below the branch
  # and the summary function should return 1 if both are below it, 0 otherwise
  return np.array([x[i] * x[j] * (x[n] != n) for i in range(n) for j in range(n)]) 

sample_sets = [[u] for u in ts.samples()] + [ts.samples()]
relatedness = ts.sample_count_stat(sample_sets, f, output_dim=n * n,
                                 polarised=True, mode="branch",
                                 span_normalise=False).reshape((n,n))
```
(note we can also remove the "strict=False" now).

petrelharp:
In an effort to make things more concrete, here's implementations of four of the things that (I think) we've been talking about. (I'm not certain they're correct, either!) Note that they're all using the same underlying summary function, they just differ in site/branch and in whether they're normalized to be a proportion or not. To be clear, I *don't* think we should define methods named the things below. I think that these things should probably be returned by a single method with different options, appropriately named.
```
def relatedness_matrix(ts, sample_sets, **kwargs):
    # NOTE: I'm outputting a matrix here just for convenience; the proposal
    # is that the tskit method *not* output a matrix, and use the indices argument
    n = len(sample_sets)
    num_samples = sum(map(len, sample_sets))
    def f(x):
      # x[i] gives the number of descendants in sample set i below the branch
      return np.array([x[i] * x[j] * (sum(x) != num_samples) for i in range(n) for j in range(n)]) 

    return ts.sample_count_stat(sample_sets, f, output_dim=n * n, **kwargs
                                     ).reshape((n,n))

def tree_covariance(ts, sample_sets):
    # if X[i] = (sum of all phenotypes across sample_sets[i])
    # where phenotypes are the sum of white noise on the trees back to the root,
    # then this gives cov[X[i], X[j]]
    return relatedness_matrix(ts, sample_sets, mode="branch", span_normalise=False, polarised=True)

def mutation_covariance(ts, sample_sets):
    # if X[i] = (sum of all mutations across genotypes in sample_sets[i])
    # where mutations are N(0,1), then this gives cov[X[i], X[j]]
    # Equivalently, gives the total number of pairwise derived allele matches.
    return relatedness_matrix(ts, sample_sets, mode="site", span_normalise=False, polarised=True)


def pairwise_proportion_shared(ts, sample_sets, **kwargs):
    K = relatedness_matrix(ts, sample_sets, mode="site", polarised=False)
    # lazily, just take "segregating site" to mean segregating among *all* samples
    S = ts.segregating_sites(mode="site")
    P = K / S
    for i, A in enumerate(sample_sets):
        for j, B in enumerate(sample_sets):
            P[i,j] /= len(A) * len(B)
    return K / S

def prob_shared_alleles(ts, sample_sets):
    # K[i,j] is the probability that two alleles drawn from sample_sets[i] and [j],
    # with replacement, at a random segregating site, agree
    return pairwise_proportion_shared(ts, sample_sets, mode="site")

def prop_shared_branches(ts, sample_sets):
    # K[i,j] is the probability that a random location chosen proportion to area
    # on the portion of the tree sequence segregating among samples is ancestral
    # to both of a random draw from sample_sets[i] and [j], with replacement
    return pairwise_proportion_shared(ts, sample_sets, mode="branch")
```

gregorgorjanc:
> P = K / S
>     for i, A in enumerate(sample_sets):
>         for j, B in enumerate(sample_sets):
>             P[i,j] /= len(A) * len(B)
>     return K / S

There is an error here - you calculate P = K / S, modify it, but then you return K / S.

What is this modification of P doing?

petrelharp:
Ah thanks - I meant to return `P` (and will edit in that fix).  This is converting the amount of pairwise sharing to the probabiilty a randomly chosen pair shares. For instance, `prob_shared_alleles` is (I believe) the "allele-sharing" metric of relatedness in Equation (6) of Speed & Balding.

c/tskit/trees.c
```diff
@@ -2734,6 +2734,56 @@ tsk_treeseq_divergence(tsk_treeseq_t *self, tsk_size_t num_sample_sets,
...
+        sumx += x[k];
+    }
+
+    for (k = 0; k < result_dim; k++) {
+        num += args.sample_set_sizes[k];
+    } 
+
+    if (num != sumx) {
+        c = 1;
+    }
```
petrelharp, line 2759: maybe the problem is floating-point error? A maybe easier way to do this (that might help?) would be to leave off `sums` and `num` and do
```
     for (k = 0; k < state_dim; k++) {
        if (x[k] != args.sample_set_sizes[k]) {
            c = 1;
        }
    }
```

brieuclehmann:
This latest commit includes a c-like implementation of the summary function in python and fixed a bug in the C implementation.
The c-like implementation in python passes the tests and looks to me identical to the c implementation, which is still failing... Based on a few printfs, it doesn’t look like it’s to do with floating point errors, but I can’t spot the bug!

jeromekelleher:
@brieuclehmann that should sort out the ``polarised`` issue. I skipped the internal samples test just to make sure that everything else is passing OK. We should decide  how to fix this properly before merging the PR

brieuclehmann:
Thank you @jeromekelleher ! The (failure of the) internal samples test raises a slightly thorny issue related to how we should 'center' the covariance, as discussed previously with @petrelharp and @gregorgorjanc . Here is the tree sequence for that test, the samples are nodes (1, 2, 3, 5).
```
1.00┊   7     ┊         ┊         ┊  
    ┊ ┏━┻━┓   ┊         ┊         ┊  
0.80┊ ┃   ┃   ┊     8   ┊         ┊  
    ┊ ┃   ┃   ┊   ┏━┻━┓ ┊         ┊  
0.70┊ ┃   ┃   ┊   ┃   ┃ ┊   6     ┊  
    ┊ ┃   ┃   ┊   ┃   ┃ ┊ ┏━┻━┓   ┊  
0.50┊ ┃   5   ┊   5   ┃ ┊ ┃   5   ┊  
    ┊ ┃ ┏━┻┓  ┊  ┏┻━┓ ┃ ┊ ┃ ┏━┻┓  ┊  
0.40┊ ┃ ┃  4  ┊  4  ┃ ┃ ┊ ┃ ┃  4  ┊  
    ┊ ┃ ┃ ┏┻┓ ┊ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┏┻┓ ┊  
0.20┊ ┃ ┃ ┃ 3 ┊ ┃ ┃ ┃ 3 ┊ ┃ ┃ ┃ 3 ┊  
    ┊ ┃ ┃ ┃   ┊ ┃ ┃ ┃   ┊ ┃ ┃ ┃   ┊  
0.10┊ ┃ 1 2   ┊ ┃ 2 1   ┊ ┃ 1 2   ┊  
    ┊ ┃       ┊ ┃       ┊ ┃       ┊  
0.00┊ 0       ┊ 0       ┊ 0       ┊  
  0.00      0.20      0.80      1.00 
```
The source of the failure is that the naive implementation calculates shared branch areas up to the roots of the entire tree sequence, i.e. nodes (7, 8, 6), whereas the sample_count_stat implementation calculates shared branch areas up to the mrca of the sample nodes, i.e. node 5. It's not obvious to me which is 'correct' - maybe we want to cover these different options with an extra argument?

jeromekelleher:
I think the ``sample_count_stat`` version is probably doing the right thing here, so we should just change the naive implementation. We agonised endlessly about this sort of thing when doing the general stats, so I think just following the framework's conventions is probably OK.

Unless @petrelharp disagrees?

brieuclehmann:
Sorry, I hadn't spotted that 8 is the mrca in the second tree. `sample_count_stat` includes the shared branch areas up to 8 in the second tree, but only up to 5 in trees 1 and 3.

gregorgorjanc:
> Sorry, I hadn't spotted that 8 is the mrca in the second tree. `sample_count_stat` includes the shared branch areas up to 8 in the second tree, but only up to 5 in trees 1 and 3.

OK, then the difference between taking into account oldest ancestors versus MRCA for each tree (`sample_count_stat` way) will be just a scale difference. I vote for the `sample_count_stat` way since that is the default elsewhere. It might be that this is not a good default, but kick this can further down the road.

One argument for taking into account all ancestors would be that the associated variance component from a mixed model (where this relatedness matrix would be used) would not map to a well-defined time-point in the past across all trees (as it does *if* we use pedigrees with the same depth for all individuals), but ancestors and MRCA are or different ages anyway.

brieuclehmann:
I’ve adjusted the tests to have mutations (so that I’m testing against the naive genotype covariance rather than branch covariance). There’s still the question of rescaling: it turns out that for `polarised = False`, we want to divide by 2p (where p is the number of segregating sites), since that’s the total number of alleles. It’s not clear to me if there’s a neat way to do this rescaling within the summary function / sample_count_stat ...

jeromekelleher:
This is a heads-up here @brieuclehmann that you might have some merge problems when you update because of #939 

It shouldn't affect you much, but you might need to add a few ``const``s in to make things consistent. I'm happy to do the update if this'll help.

It would be good to get a version of this PR in soon though - the codebase is moving quickly at the moment and PRs get out of date quickly. We can always file some issues to track anything that isn't quite worked out yet.

jeromekelleher:
Is this more-or-less ready to go then @brieuclehmann? We can open an issue to track the ``proportion`` and documentation issues sepearately.

You have some lint issues to fix here I think.

brieuclehmann:
Just a minor tweak to be made to remove `polarised` in the denominator. Also, that's very much a hack at the moment (see last commit) but I'm not sure where best to do the division by 2.

Re linting, I think it's because I don't have clang-format installed...

jeromekelleher (approved):

jeromekelleher:
Looks great, thanks @brieuclehmann! Can you open issues to track any remaining things that need to be resolved please?

petrelharp:
Two minor things:
- I think the denominator should always be 2, and should be incorporated into the summary function.
- I think the C and python methods should be consistently named: one is `relatedness`, the other `genotype_relatedness`; I think we decided to do the second thing.

These can be done in follow-up. We also want more testing in the follow-up.

jeromekelleher:
OK, great, I'm going to merge this. Thanks @brieuclehmann, this is a bit step forward!

Issue #891, opened 2020-10-02
Add _tskitmodule.c to clang-format


benjeffery:
See also #868 and #77

benjeffery:
Closed by #899

PR #899, opened 2020-10-08
Add clang-format to pre-commit

### LINKED ISSUE #868: Add make target for clang-format.
As in msprime, to save devs needing to remember which files to format.

benjeffery:
I've found a way to skip `clang-format` if it isn't found on the system.

Fixes #868, #891, #820

@petrelharp Mergify is configured to require 1 approving review - so while @jeromekelleher is away I'll tag you for reviews ok?

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -608,7 +608,7 @@ On Debian/Ubuntu, these can be installed using
 
 .. code-block:: bash
 
-    $ sudo apt install libcunit1-dev ninja-build meson clang-format
+    $ sudo apt install libcunit1-dev ninja-build meson clang-format-6.0
 
 (A more recent version of meson can alternatively be installed using ``pip``, if you wish.)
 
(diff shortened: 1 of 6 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/899/)

docs/development.rst
```diff
@@ -627,14 +627,8 @@ Code style
...
 C code is formatted using
 `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_
-with a custom configuration.
-To ensure that your code is correctly formatted, you can run
-
-.. code-block:: bash
-
-    clang-format -i c/tskit/* c/tests/*.c c/tests/*.h
-
-before submitting a pull request.
```
petrelharp, line 637: I think it's nice to leave in the DIY command? pre-commit checks a lot of things and can produce a lot of output, so it's less overwhelming to be able to easily run them separately sometimes.  Unless this is out of date somehow?

petrelharp: It could say "If you have clang-format installed, you can run the check manually as follows..."

petrelharp (approved):
LGTM, with the minor suggestion (or not)

docs/development.rst
mufernando: I agree with Peter. But just noticed that when you run clang-format in `c/tskit/trees.c` it changes sth that it shouldnt.

benjeffery: >  But just noticed that when you run clang-format in c/tskit/trees.c it changes sth that it shouldnt.
What version of `clang-format`? It is not stable across releases. CI and this pre-commit use 6.0.

mufernando: hah, thanks! I had 5.0.1

petrelharp: Oh, well if it depends precisely on the version, maybe we should just say "You can run clang-format yourself, but note that the results depend on clang-format's version, and we use a specific version in the pre-commit checks."

petrelharp: Or even "We don't recommend you run clang-format yourself, because the results depends on clang-format's version."

benjeffery: I've re-added a manual running snippet and added a note about the version.

benjeffery:
Merging manually due to github action changes blocking mergify.

PR #909, opened 2020-10-17
Support file objects (streams, sockets etc.)

### LINKED ISSUE #657: File objects for Python dump/load

benjeffery:
Closes #657 

WIP! Using PR for testing on WIN/OSX.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -1,13 +1,62 @@
+# MIT License
+#
+# Copyright (c) 2018-2020 Tskit Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Configuration and fixtures for pytest. Only put test-suite wide fixtures in here. Module
+specific fixtures should live in their modules.
+
+To use a fixture in a test simply refer to it by name as an argument. This is called
+dependancy injection. Note that all fixtures should have the suffix "_fixture" to make
+it clear in test code.
+
+For example to use the `ts` fixture (a tree sequence with data in all fields) in a test:
+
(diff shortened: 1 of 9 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/909/)

benjeffery:
There seems to be an issue with codecov not showing files, so haven't checked coverage yet. Will look into it.

python/tests/conftest.py
```diff
@@ -1,4 +1,33 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Configuration and fixtures for pytest. Only put test-suite wide fixtures in here. Module
+specific fixtures should live in their modules.
```
jeromekelleher, line 24: This is great! Can you just put in either a quick example of how to use these fixtures, or perhaps a link to the documentation from pytest for how this this is done?

jeromekelleher (approved):
LGTM! I haven't pored over the details of the tricky stuff, but I assume this is derived from what we did in kastore, so I don't need to?

python/tests/conftest.py
benjeffery: https://github.com/tskit-dev/tskit/pull/909/commits/c30dde114a74293c4ea7c956304dbdb08ce46780

benjeffery:
> I assume this is derived from what we did in kastore, so I don't need to?

Yep, a lot of this was copied over.

python/tests/conftest.py
```diff
@@ -53,9 +72,11 @@ def simple_ts():
     return msprime.simulate(2, random_seed=42)
 
 
-# A tree sequence with data in all fields
 @fixture(scope="session")
 def ts():
```
jeromekelleher, line 76: This is great, but I worry a bit about attaching meaning to the (very commonly used) variable ``ts`` here. I can see this being confusing for people. What if we adopt a convention like having a suffix or prefix of ``fixture`` for these parameters? That might make things a bit more obvious.

python/tests/conftest.py
```diff
@@ -100,5 +121,8 @@ def ts():
 
 
 @fixture(scope="session")
-def multi_ts():
-    return [msprime.simulate(i, random_seed=42) for i in range(2, 100, 5)]
+def multi_ts(ts):
```
jeromekelleher, line 124: This is a bit vague, as I'd imagine we'll want all sorts of different fixtures like this. What about ``replicate_ts_fixture`` which will return a list of msprime replicate sims?

python/tests/conftest.py
benjeffery: Yeah, this kind of magic can be really confusing if you aren't aware what it going on. It seemed a bit much to put a suffix on, but when I think about the reader of the code it really does help. I've added them in https://github.com/tskit-dev/tskit/pull/909/commits/b0b566b26220b9a9214b378c3163b5586b341814

python/tests/conftest.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/909/commits/b0b566b26220b9a9214b378c3163b5586b341814

python/tests/conftest.py
jeromekelleher: I agree, it feels a bit wordy in the code, but I think it'll help people who aren't used to using pytest.

jeromekelleher:
LGTM, merge away whenever!

PR #897, opened 2020-10-07
table collection equivalence

mufernando:
WIP #896

PR #917, opened 2020-10-20
Change tsk_table_collection_equals to take options.

jeromekelleher:
In favour of the ``tsk_table_collection_equals_with_options`` introduced in #897.

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -29,6 +29,7 @@ import json
 import math
 import pickle
 import random
+import time
 import unittest
 import warnings
 
@@ -2395,7 +2396,7 @@ class TestTableCollection:
         t2.populations.clear()
         assert t1 == t2
 
-    def test_equals_with_options(self):
+    def test_equals_options(self):
         pop_configs = [msprime.PopulationConfiguration(5) for _ in range(2)]
         migration_matrix = [[0, 1], [1, 0]]
         t1 = msprime.simulate(
@@ -2421,21 +2422,20 @@ class TestTableCollection:
         t1.metadata_schema = tskit.MetadataSchema({"codec": "json", "type": "object"})
         t1.metadata = {"hello": "world"}
         assert not t1.equals(t2)
-        assert t1.equals(t2, ignore_top_level_metadata=True)
+        assert t1.equals(t2, ignore_ts_metadata=True)
         assert not t2.equals(t1)
-        assert t2.equals(t1, ignore_top_level_metadata=True)
+        assert t2.equals(t1, ignore_ts_metadata=True)
         t2.metadata_schema = t1.metadata_schema
         assert not t1.equals(t2)
-        assert t1.equals(t2, ignore_top_level_metadata=True)
+        assert t1.equals(t2, ignore_ts_metadata=True)
         assert not t2.equals(t1)
-        assert t2.equals(t1, ignore_top_level_metadata=True)
+        assert t2.equals(t1, ignore_ts_metadata=True)
 
-        # testing both
         t1.provenances.add_row("random stuff")
         assert not t1.equals(t2)
-        assert not t1.equals(t2, ignore_top_level_metadata=True)
+        assert not t1.equals(t2, ignore_ts_metadata=True)
         assert not t1.equals(t2, ignore_provenance=True)
-        assert t1.equals(t2, ignore_top_level_metadata=True, ignore_provenance=True)
+        assert t1.equals(t2, ignore_ts_metadata=True, ignore_provenance=True)
 
         t1.provenances.clear()
         t2.metadata = t1.metadata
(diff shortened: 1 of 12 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/917/)

jeromekelleher:
It would be good to get some feedback here at this point @benjeffery @petrelharp @mufernando. Basically, I've added an "options" argument to all the tables and made metadata comparisons optional now. I've also added an option for ignoring provenance timestamps, which will be handy when comparing simulation outputs.

Questions:

1. Do we want the ``TSK_CMP_NO_TS_METADATA`` option now? Seems like just saying ``TSK_CMP_NO_METADATA`` and not ignoring all the metadata is probably what we want, most of the time?
2. Should we also have an option to ignore the metadata schemas? If so, how does this interact with the ``TSK_CMP_NO_METADATA`` flag? Probably the simplest is the have TSK_CMP_NO_METADATA just ignore metadata and TSK_CMP_NO_METADATA_SCHEMA just ignore the schema?
2. What do we do at the Python level? Perhaps it would be better to have 

```python
def equals(self, other, metadata=True, provenance=True, provenance_timestamp=True):
```

rather than the current ``ignore_metadata=False`` or (worse) ``no_metadata=False``?

petrelharp:
>     1. Do we want the `TSK_CMP_NO_TS_METADATA` option now? Seems like just saying `TSK_CMP_NO_METADATA` and not ignoring all the metadata is probably what we want, most of the time?

Well, in the use case of union'ing SLiM tree sequences, we do want to just ignore top-level metadata, so... yes?
 
>     2. Should we also have an option to ignore the metadata schemas? If so, how does this interact with the `TSK_CMP_NO_METADATA` flag? Probably the simplest is the have TSK_CMP_NO_METADATA just ignore metadata and TSK_CMP_NO_METADATA_SCHEMA just ignore the schema?

Well, ignoring metadata should ignore schemas, but I don't think we should ignore schemas if we compare metadata. It would maybe be useful to be able to ask to compare equivalency of schemas rather than equiality at the bytes level, but that's another can of worms (to do it in C anyhow). I'd say skip it unless we actually need it?

>     3. What do we do at the Python level? Perhaps it would be better to have
> 
> ```python
> def equals(self, other, metadata=True, provenance=True, provenance_timestamp=True):
> ```
> 
> rather than the current `ignore_metadata=False` or (worse) `no_metadata=False`?

Sounds good, plus maybe `top_level_metadata=metadata`?

mufernando:
This all looks good to me. I agree that we need `TSK_CMP_NO_TS_METADATA` for top-level only metadata/schema.

jeromekelleher:
I think this is ready for final review here. I ended up going back to the original "IGNORE_X" names, as the semantics got confusing when we changed back and forth between positive and negative in Python and C. "Ignore" is a bit longer, but I think it's quite clear what the flags go then.

Speaking of semantics - are we all happy that ``ignore_metadata`` implies ``ignore_ts_metadata``, and the latter has no effect when ``ignore_metadata=True``?

python/tests/test_tables.py
```diff
@@ -2540,6 +2539,88 @@ def test_sequence_length_longer_than_edges(self):
         assert len(tree.parent_dict) == 0
 
 
+class TestEqualityOptions:
+    def test_equals_provenance(self):
+        t1 = msprime.simulate(10, random_seed=42).tables
+        t2 = msprime.simulate(10, random_seed=42).tables
```
benjeffery, line 2545: Windows is having trouble here as the resolution of `datetime.datetime.now()` is about 1ms(!!)
```suggestion
        t1 = msprime.simulate(10, random_seed=42).tables
        time.sleep(0.1)
        t2 = msprime.simulate(10, random_seed=42).tables
```

python/tskit/tables.py
```diff
@@ -2146,14 +2180,58 @@ def __str__(self):
...
+        :param bool ignore_metadata: If True *all* metadata and metadata schemas
+            will be excluded from the comparison. This includes the top-level
+            tree sequence and constituent table metadata (default=False).
+        :param bool ignore_ts_metadata: If True the top-level tree sequence
+            metadata and metadata schemas will be excluded from the comparison.
+            If ``ignore_metadata`` is True, this parameter has no effect.
+        :param bool ignore_provenance: If True the provenance tables are
+            not included in the comparison.
+        :param bool ignore_timestamps: If True the provenance timestamp column
+            is ignored in the comparision.
```
benjeffery, line 2216: Add a note that there is no effect if `ignore_provenance` it is there in the c doc and for `ignore_ts_metadata`

benjeffery (approved):
I agree with the semantics we have in this PR now. Just one minor doc suggestion and a broken windows test that is easily (if somewhat unsatisfyingly) fixed.

python/tskit/tables.py
jeromekelleher: Good point, well spotted.

mufernando:
LGTM!

jeromekelleher:
OK, great, let's get this one in then.

Issue #844, opened 2020-09-10
const-correct public C API


castedo:
Following up on discussion with @molpopgen and @jeromekelleher regarding possibility of making tskit 1.0 const-correct (const-correct in the API being use of "const" in pointer parameter types declaration when API functions do not change the data pointed to).

I can't think of any downside for the callers of the API to have those parameter types get const'ed and I don't believe it will break any users (unless they are doing something unusual like keeping pointers to functions and assigning API functions to those function pointers).

In the public API, this is convenient to API users who use const often (e.g. @molpopgen and many C++ programmers). It is less annoying too. They don't have to cast away const-ness when calling the API. 

However, for tskit implementors this can become a pain. This page has a nice description of the trade-off and uses the term 
"_const poisoning_":
https://wiki.sei.cmu.edu/confluence/display/c/DCL00-C.+Const-qualify+immutable+objects

I can suggest initially only focusing on making the public API const-correct and not worrying about the broader use of const in the implementation of tskit. **If** there is a const issue introduced by this const type change, one can just blindly do something like this:
```
int public_api_func(double *values)
{
   ... code not "speaking in const" ...
}
```
goes to
```
int public_api_func(double const *_values)
{
    double *values = (double*)_values;
   ... same code not "speaking in const" ...
}
```
And if there is no const issue, then no need to do this.

In a sense, this is putting the burden of casting away const-ness on the implementors of tskit rather than users who call it. Which seems to be the right place to put that burden.

benjeffery:
Yep, I agree this is a good thing to do. I've caught issues before by "constifiying" some C++ code. I think it also helps the API become a bit more self-documenting.

jeromekelleher:
OK, added to the C 1.0 milestone.

molpopgen:
While I am in favor of this in general, it could be a big load of work that goes nowhere.  In my experience, it is hard to `const`ify a code base after the fact.

C allows 3 distinct uses of `const` in a function prototype.  If the goal is to document to the user that their
input data are guaranteed to be unmodified by the function, only one of the 3 uses satisfies that goal and is also
compatible with a C++ compiler. 

Personally, I prefer `foo(const double * x)` to `foo(double * const x)`.  Given that msprime uses the GSL, and the former convention is used there, it may be confusing to have different flavors of const in play for people new to looking at the code.
For the latter case, the user's input data are still mutable.

```c
#include <stdlib.h>
#include <stdio.h>
#include <assert.h>

[...]

PR #919, opened 2020-10-21
constify C Table API

castedo:
As a first stage for #844 I am focusing on the Table API which does not depend on other parts of the C API.

I will probably have this "constification" of the Table API ready today.

I'm not sure how long it will take the do the rest of the C API yet. But this topic branch focused on only the Table API could be merged before the rest of the C API is constified.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -148,7 +148,8 @@ out:
 
 /* Checks that the specified list of offsets is well-formed. */
 static int
-check_offsets(size_t num_rows, tsk_size_t *offsets, tsk_size_t length, bool check_length)
+check_offsets(
+    size_t num_rows, const tsk_size_t *offsets, tsk_size_t length, bool check_length)
 {
     int ret = TSK_ERR_BAD_OFFSET;
     size_t j;
@@ -207,7 +208,7 @@ out:
 
 static int
 write_metadata_schema_header(
-    FILE *out, char *metadata_schema, tsk_size_t metadata_schema_length)
+    FILE *out, const char *metadata_schema, tsk_size_t metadata_schema_length)
 {
     const char *fmt = "#metadata_schema#\n"
                       "%.*s\n"
@@ -367,8 +368,8 @@ out:
 }
 
 int TSK_WARN_UNUSED
-tsk_individual_table_copy(
-    tsk_individual_table_t *self, tsk_individual_table_t *dest, tsk_flags_t options)
+tsk_individual_table_copy(const tsk_individual_table_t *self,
+    tsk_individual_table_t *dest, tsk_flags_t options)
 {
     int ret = 0;
 
@@ -391,8 +392,8 @@ out:
 
 int TSK_WARN_UNUSED
 tsk_individual_table_set_columns(tsk_individual_table_t *self, tsk_size_t num_rows,
-    tsk_flags_t *flags, double *location, tsk_size_t *location_offset,
-    const char *metadata, tsk_size_t *metadata_offset)
+    const tsk_flags_t *flags, const double *location, const tsk_size_t *location_offset,
+    const char *metadata, const tsk_size_t *metadata_offset)
 {
     int ret;
 
@@ -408,8 +409,8 @@ out:
 
 int
 tsk_individual_table_append_columns(tsk_individual_table_t *self, tsk_size_t num_rows,
-    tsk_flags_t *flags, double *location, tsk_size_t *location_offset,
-    const char *metadata, tsk_size_t *metadata_offset)
+    const tsk_flags_t *flags, const double *location, const tsk_size_t *location_offset,
+    const char *metadata, const tsk_size_t *metadata_offset)
 {
     int ret;
     tsk_size_t j, metadata_length, location_length;
@@ -486,7 +487,7 @@ out:
 
 static tsk_id_t
 tsk_individual_table_add_row_internal(tsk_individual_table_t *self, tsk_flags_t flags,
-    double *location, tsk_size_t location_length, const char *metadata,
+    const double *location, tsk_size_t location_length, const char *metadata,
     tsk_size_t metadata_length)
 {
     tsk_bug_assert(self->num_rows < self->max_rows);
@@ -507,7 +508,7 @@ tsk_individual_table_add_row_internal(tsk_individual_table_t *self, tsk_flags_t 
 
 tsk_id_t
 tsk_individual_table_add_row(tsk_individual_table_t *self, tsk_flags_t flags,
-    double *location, tsk_size_t location_length, const char *metadata,
+    const double *location, tsk_size_t location_length, const char *metadata,
     tsk_size_t metadata_length)
 {
     int ret = 0;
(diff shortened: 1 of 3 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/919/)

castedo:
There are a few more functions in the Table API that I am still investigating. Specifically the `_get_row(` functions which set (return) row structures. At the moment two of these row structure types, namely `tsk_individual_t` and `tsk_site_t` look like they maybe can be used to *modify* the contents of the table those rows "point" to. Depending on the intended semantics, thos `_get_row` functions maybe should be constafied too.

benjeffery:
Thanks for picking this up @castedo! To avoid large conflicts we can merge as you have chunks ready.

castedo:
Ready for review now.

There is a type change which I'd like to bring to attention. The type change is part of making a number of other functions const-correct, namely the `_get_row(` functions setting (returning) `tsk_individual_t` and `tsk_site_t`.

The high-level semantic clarficiation I've coded is the following:

that the eight `tsk_..._table_get_row(` functions are intended to get a "row" from the table and NOT to use that "row" struct to then make changes to the original `tsk_..._table_t`. In other words, the row returned is conceptually a const iterator/pointer for read-only access.

In the current `main` branch the only ambiguity is with `tsk_individual_table_get_row` and `tsk_site_table_get_row` because the "row" types returned contain non-const pointers. In this PR I have changed those pointers to be const, as summarized in code here:

```
typedef struct {
...
    const double *location;
...
    const tsk_id_t *nodes;
...
} tsk_individual_t;

typedef struct {
...
    const tsk_mutation_t *mutations;
...
} tsk_site_t;
```

This is consistent with the `const char *` data members.

castedo:
FYI, the unsuccessful check is due to `Too long with no output (exceeded 10m0s): context deadline exceeded` after running `pytest`. I suspect it's some hiccup on the part of CircleCI. `pytest` passed fine on my local machine.

benjeffery:
> FYI, the unsuccessful check is due to `Too long with no output (exceeded 10m0s): context deadline exceeded` after running `pytest`. I suspect it's some hiccup on the part of CircleCI. `pytest` passed fine on my local machine.

This is a flakey test introduced by #909. I'll re-run circle CI.

castedo:
If you can't get enough of reviewing const parameter changes, endulge some more on #920! :smile: 
Take this PR now and #920 later or just take #920 which includes everything in this PR. They are commits touching separate files.

benjeffery (approved):
Yikes, thanks for this @castedo - looks like a lot of work. 
All these changes make sense to me. 

I'll ping @molpopgen for a quick look -  ignore if you're too busy with teaching.

castedo:
>  looks like a lot of work.

It's so much easier to suggest an API should use `const` vs actually doing the work to make it reality. :laughing:

Originally I thought this work would be tweaking a few function declarations. But I ended up deciding I should actually confirm that const parameters really don't get changed. And in confirming that, I basically was figuring out what parameters should be const for all the functions that get called. It was kind of fun figuring out what the code does! I keep hoping I'm going to spot an interesting bug.

c/tskit/tables.c
```diff
@@ -391,7 +392,7 @@ tsk_individual_table_copy(
 
 int TSK_WARN_UNUSED
 tsk_individual_table_set_columns(tsk_individual_table_t *self, tsk_size_t num_rows,
-    tsk_flags_t *flags, double *location, tsk_size_t *location_offset,
+    tsk_flags_t *flags, const double *location, tsk_size_t *location_offset,
```
jeromekelleher, line 395: Shouldn't all these pointer parameters be const (i.e., not just ``location``)?

c/tskit/tables.c
```diff
@@ -927,8 +930,8 @@ tsk_node_table_copy(tsk_node_table_t *self, tsk_node_table_t *dest, tsk_flags_t
 
 int TSK_WARN_UNUSED
 tsk_node_table_set_columns(tsk_node_table_t *self, tsk_size_t num_rows,
-    tsk_flags_t *flags, double *time, tsk_id_t *population, tsk_id_t *individual,
-    const char *metadata, tsk_size_t *metadata_offset)
+    tsk_flags_t *flags, const double *time, const tsk_id_t *population,
```
jeromekelleher, line 933: flags should be const here I think

jeromekelleher: Looks like there's a good few of these - I won't comment on any more of them

c/tskit/tables.h
```diff
@@ -92,14 +92,14 @@ typedef struct {
     tsk_flags_t flags;
     /** @brief Spatial location. The number of dimensions is defined by
      * ``location_length``. */
-    double *location;
+    const double *location;
```
jeromekelleher, line 95: It's not so clear to me what the ``const`` is doing here. Is this documenting the fact that the memory for this object is managed by the library and should be read-only to the client?

jeromekelleher (commented):
I think this a big improvement and a significant boost for library correctness and documentation - thanks very much for taking this on @castedo! I have a few questions and I think there's a some minor inconsistencies in places, but I think we should do this.

I was worried about const poisoning, but it doesn't seem to have affected our test code at all, and we have compiler flags on very high paranoia settings so that seems fine.

jeromekelleher:
> I keep hoping I'm going to spot an interesting bug.

Keep looking, I'm sure there's some there! :wink:

c/tskit/tables.h
castedo: "managed" in the sense of read-only to the client, yes.

The main motivation for adding `const` to `double *location` was due to:
```
tsk_individual_table_get_row(const tsk_individual_table_t *self, tsk_id_t index, tsk_individual_t *row);
```
The array of locations is part of the state of `tsk_individual_table_t`. And I made that 1st parameter `const`. So if one calls `tsk_individual_table_get_row` to get a `tsk_individual_t` "row" _into_ that table, that "row" should not be able to change the state of the table.

But there is a design decision there. Looking over the code it seemed to me the intention is that these "row" objects are conceptually read-only copies of the data found in the table. In terms of literal memory they might not be entirely copied memory, but conceptually they are copies of logical state.

jeromekelleher: The row-objects returned by ``x_table_get_row`` are definitely read-only and will remain that way, so it would be great to constify these to make this formal. Read/write row access will always be tricky because of the ragged columns, so when we want to make row-wise editing possible we should do it via a ``x_table_set_row(table, index, *params)``, where params are the same as ``x_table_add_row``. That way we can do the repacking of the ragged columns transparently (if needed).

c/tskit/tables.c
castedo: Makes sense to me too to have these be const too. I suspect I skipped them since they were listed as "Undocumented methods" in the header file. But seeing now what all has been updated, it does seem like it will be less confusing to just have everything in the file be using const when appropriate rather.

I'll go through and update more.

castedo: Good catches, does not make sense why I omitted to const those. Fixes in this commit now. I'll comment on another section which I've skipped. And inclined to skip for now.

castedo:
I've resolved remaining comments. This should be ready to merge. Looks like the build fail is another timeout and not real fail.

However, related to the comments is two parts of the header file I have not "constified". One is this section of "Undocumented methods":
https://github.com/tskit-dev/tskit/blob/c238d5707908f6c1af824a4c3150525fc4c6c3b3/c/tskit/tables.h#L2787
which looks like there could be some clarification on what params are or are not left unmodified. However, when reading the implementation code, it wasn't quickly clear to me.

And then there is the "experimental IBD API" section:
https://github.com/tskit-dev/tskit/blob/c238d5707908f6c1af824a4c3150525fc4c6c3b3/c/tskit/tables.h#L2876
which I haven't bothered to read at all really.

I'm inclined to skip them, but mainly because I don't totally understand the implementation code.

castedo:
@jeromekelleher 
> I was worried about const poisoning, but it doesn't seem to have affected our test code at all

I've saved that fun for last :), specificly the last of three PRs. I've staged the most straight forward const-ifications in the two current open PRs. The third stage I have not PR'd yet but have most of it stached away locally in git. It changes way more files. We can see how scary it looks in a third draft PR.

jeromekelleher (approved):
LGTM, thanks @castedo! Looking forward to the next installation¬!

PR #920, opened 2020-10-22
constify tree C API, but short of changing tsk_tree_t type and stats functions

castedo:
This PR builds on top of #919 and is a candidate for fixing #844. This PR can be taken in addition to or instead of #919. It's ready to go.

However, to *really* make the C API "const-correct", three pointers inside `tsk_tree_t` should arguably be made const pointers. Due to this issue, I did not upgrade the following functions:
````
int tsk_tree_init(tsk_tree_t *self, tsk_treeseq_t *tree_sequence, tsk_flags_t options);
int tsk_tree_copy(tsk_tree_t *self, tsk_tree_t *dest, tsk_flags_t options);
const tsk_id_t *tsk_treeseq_get_samples(tsk_treeseq_t *self);
int tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_length);
````

I will describe this further constification in a separate draft PR with some implications I see in tweaking `tsk_tree_t`.

### DIFF
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -229,49 +229,50 @@ int tsk_treeseq_loadf(tsk_treeseq_t *self, FILE *file, tsk_flags_t options);
 int tsk_treeseq_dump(tsk_treeseq_t *self, const char *filename, tsk_flags_t options);
 int tsk_treeseq_dumpf(tsk_treeseq_t *self, FILE *file, tsk_flags_t options);
 int tsk_treeseq_copy_tables(
-    tsk_treeseq_t *self, tsk_table_collection_t *tables, tsk_flags_t options);
+    const tsk_treeseq_t *self, tsk_table_collection_t *tables, tsk_flags_t options);
 int tsk_treeseq_free(tsk_treeseq_t *self);
-void tsk_treeseq_print_state(tsk_treeseq_t *self, FILE *out);
+void tsk_treeseq_print_state(const tsk_treeseq_t *self, FILE *out);
 
 /** @} */
 
-tsk_size_t tsk_treeseq_get_num_nodes(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_edges(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_migrations(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_sites(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_mutations(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_provenances(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_populations(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_individuals(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_trees(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_num_samples(tsk_treeseq_t *self);
-char *tsk_treeseq_get_metadata(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_metadata_length(tsk_treeseq_t *self);
-char *tsk_treeseq_get_metadata_schema(tsk_treeseq_t *self);
-tsk_size_t tsk_treeseq_get_metadata_schema_length(tsk_treeseq_t *self);
-char *tsk_treeseq_get_file_uuid(tsk_treeseq_t *self);
-double tsk_treeseq_get_sequence_length(tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_nodes(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_edges(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_migrations(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_sites(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_mutations(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_provenances(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_populations(const tsk_treeseq_t *self);
+tsk_size_t tsk_treeseq_get_num_individuals(const tsk_treeseq_t *self);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AdminBot-tskit** [comment]
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/920/)

**Comment by jeromekelleher** [comment]
I'm on board with the constification - let's get #919 shipped away first and come back to this .

**Comment by castedo** [comment]
@jeromekelleher Here's the scoop now on this PR. It's ready to review. This PR constifies the non-stats tree functions. This PR does not include any of the stats functions. I'm going to constify those in a separate PR. I wasn't sure what's "public" vs not in the C API, so I was going with what shows up in the docs. With this PR effectively all methods appearing in the docs are constified.

This PR also does not constify the `tsk_tree_t` type. That change makes the biggest impact on random other parts of the codebase, so I've left that for last.

I stated earlier that there would be 3 PRs, but I'm currently planning 4 PRs. This PR is rather small so maybe in hind sight 3 larger PR would be preferable. Let me know what you think after all is done. I'm partly doing 4 PRs mainly to see how many git topic branches and PRs I can juggle and merge simultaneously without driving myself or you crazy. :laughing: Good testing for https://github.com/castedo/git-prepr

File: c/tskit/trees.h [hunk 1]
```diff
@@ -229,49 +229,50 @@ int tsk_treeseq_loadf(tsk_treeseq_t *self, FILE *file, tsk_flags_t options);
 int tsk_treeseq_dump(tsk_treeseq_t *self, const char *filename, tsk_flags_t options);
```
jeromekelleher, line 229: Both dump methods can have ``const self`` I think. In general, the tsk_treeseq_t class is immutable, so nearly every method should have have a ``const`` self. Other than, I think, ``init``, ``load``, and ``free``.

c/tskit/trees.h
```diff
@@ -362,42 +363,42 @@ int tsk_treeseq_f4(tsk_treeseq_t *self, tsk_size_t num_sample_sets,
...
+bool tsk_tree_has_sample_lists(const tsk_tree_t *self);
+bool tsk_tree_has_sample_counts(const tsk_tree_t *self);
+bool tsk_tree_equals(const tsk_tree_t *self, const tsk_tree_t *other);
 /* Returns true if u is a descendant of v; false otherwise */
-bool tsk_tree_is_descendant(tsk_tree_t *self, tsk_id_t u, tsk_id_t v);
-bool tsk_tree_is_sample(tsk_tree_t *self, tsk_id_t u);
+bool tsk_tree_is_descendant(const tsk_tree_t *self, tsk_id_t u, tsk_id_t v);
+bool tsk_tree_is_sample(const tsk_tree_t *self, tsk_id_t u);
 
 int tsk_tree_copy(tsk_tree_t *self, tsk_tree_t *dest, tsk_flags_t options);
```
jeromekelleher, line 388: Copy can have ``const self`` I think?

c/tskit/trees.h
```diff
@@ -362,42 +363,42 @@ int tsk_treeseq_f4(tsk_treeseq_t *self, tsk_size_t num_sample_sets,
...
-int tsk_tree_get_mrca(tsk_tree_t *self, tsk_id_t u, tsk_id_t v, tsk_id_t *mrca);
-int tsk_tree_get_num_samples(tsk_tree_t *self, tsk_id_t u, size_t *num_samples);
+int tsk_tree_get_parent(const tsk_tree_t *self, tsk_id_t u, tsk_id_t *parent);
+int tsk_tree_get_time(const tsk_tree_t *self, tsk_id_t u, double *t);
+int tsk_tree_get_mrca(const tsk_tree_t *self, tsk_id_t u, tsk_id_t v, tsk_id_t *mrca);
+int tsk_tree_get_num_samples(const tsk_tree_t *self, tsk_id_t u, size_t *num_samples);
 int tsk_tree_get_num_tracked_samples(
-    tsk_tree_t *self, tsk_id_t u, size_t *num_tracked_samples);
+    const tsk_tree_t *self, tsk_id_t u, size_t *num_tracked_samples);
 int tsk_tree_get_sites(tsk_tree_t *self, tsk_site_t **sites, tsk_size_t *sites_length);
```
jeromekelleher, line 400: const self here?

jeromekelleher (commented):
LGTM @castedo, some minor inconsistencies I think. Basically, I'm happy with ``const self`` for almost all tsk_treeseq_t methods and nearly all ``tsk_tree_t`` methods.

c/tskit/trees.h
castedo: I was worried about this TSK_NO_BUILD_INDEXES option and rebuilding of indexes within the implementation of dump. I think maybe when data inside is not sorted. It's hazy to me, I'm so new to it.

I've come across the terms 'physical state' and 'logical state': https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-const
So a question is whether this potential rebuilding of indexes is totally "invisible" and thus just change in "physical" state and not "logical" state. My inclination would be, in C, to avoid this kind of advanced subtly that probably makes more sense in C++, where data members can be private. But in C is harder to be sure somebody might not be "seeing" changes in "physical" state even if it's not change in "logical" state.

Is this rebuilding of indexes not possible in this TS dump?

c/tskit/trees.h
castedo: I've got `tsk_tree_get_sites` flipped to const in the final branch/PR `pr/const_c_api`. I've been organizing this change as part of the "changes in tsk_tree_t type" since they collectively trigger way more changes in other parts of the codebase.

c/tskit/trees.h
castedo: Ooo, yes, this one can get const early in this PR. I'll tweak and merge main and push an update.

c/tskit/trees.h
benjeffery: In my mind the indexes are part of the logical state of a `TableCollection` as they are exposed, and not an internal implementation detail.  In the current state then `dump` cannot be `const`.

Do we break much if we remove the option to build indexes on dump? It seems a bit un-intuative that dumping creates the index.

jeromekelleher: Great points, thanks both. I think dump can probably have the option to build the indexes removed, as this is left over from an earlier version of the code that didn't have such good support for indexes being present/not present. However, this is too much to do in this PR - let's open an issue to track tagging ``tsk_treeseq_dump`` as const and removing this option.

castedo: OK, resolved for this PR. I'll create an issue on this unless I hear otherwise.

jeromekelleher (approved):
LGTM!

castedo:
I think I forgot to mention that this is good to merge.

PR #916, opened 2020-10-20
Python access to Table Collection index

benjeffery:
Fixes #4 

There is a separate issue for LWT (#870) so I'll do a new PR for that.

PR #921, opened 2020-10-22
Add index to the LWT representation

### LINKED ISSUE #870: Add indexes to the dict encoding

benjeffery:
## Description
Adds indexes (optionally) to the LWT interface. Based on #916 as it needs python access to indexes.
Fixes #870 

# PR Checklist:

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -5226,31 +5226,31 @@ TableCollection_get_indexes(TableCollection *self, void *closure)
         goto out;
     }
 
-    if (!tsk_table_collection_has_index(self->tables, 0)) {
-        ret = Py_None;
-        goto out;
-    }
-
-    insertion = table_get_column_array(self->tables->indexes.num_edges,
-        self->tables->indexes.edge_insertion_order, NPY_INT32, sizeof(tsk_id_t));
-    if (insertion == NULL) {
-        goto out;
-    }
-    removal = table_get_column_array(self->tables->indexes.num_edges,
-        self->tables->indexes.edge_removal_order, NPY_INT32, sizeof(tsk_id_t));
-    if (removal == NULL) {
-        goto out;
-    }
     indexes_dict = PyDict_New();
     if (indexes_dict == NULL) {
         goto out;
     }
-    if (PyDict_SetItemString(indexes_dict, "edge_insertion_order", insertion) != 0) {
-        goto out;
+
+    if (tsk_table_collection_has_index(self->tables, 0)) {
+        insertion = table_get_column_array(self->tables->indexes.num_edges,
+            self->tables->indexes.edge_insertion_order, NPY_INT32, sizeof(tsk_id_t));
+        if (insertion == NULL) {
+            goto out;
+        }
+        removal = table_get_column_array(self->tables->indexes.num_edges,
+            self->tables->indexes.edge_removal_order, NPY_INT32, sizeof(tsk_id_t));
+        if (removal == NULL) {
+            goto out;
+        }
+
+        if (PyDict_SetItemString(indexes_dict, "edge_insertion_order", insertion) != 0) {
+            goto out;
+        }
+        if (PyDict_SetItemString(indexes_dict, "edge_removal_order", removal) != 0) {
+            goto out;
+        }
     }
-    if (PyDict_SetItemString(indexes_dict, "edge_removal_order", removal) != 0) {
-        goto out;
-    }
+
     ret = indexes_dict;
     indexes_dict = NULL;
 out:
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -5226,6 +5226,11 @@ TableCollection_get_indexes(TableCollection *self, void *closure)
         goto out;
     }
 
+    if (!tsk_table_collection_has_index(self->tables, 0)) {
+        ret = Py_None;
+        goto out;
+    }
+
     insertion = table_get_column_array(self->tables->indexes.num_edges,
         self->tables->indexes.edge_insertion_order, NPY_INT32, sizeof(tsk_id_t));
     if (insertion == NULL) {
--- a/python/lwt_interface/CHANGELOG.rst
+++ b/python/lwt_interface/CHANGELOG.rst
@@ -1,0 +1,6 @@
+--------------------
+[0.1.2] - 2020-10-22
+--------------------
+
+ - Added optional top-level key ``index`` which has contains ``edge_insertion_order`` and
+   ``edge_removal_order``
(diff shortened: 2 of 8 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/921/)

benjeffery:
I recommend merging #916 first then asking mergify to rebase before review.

c/tskit/tables.h
```diff
@@ -2721,6 +2721,26 @@ collection. Any existing index is first dropped using
...
+This method sets the edge insertion/removal index for this table collection
+The index arrays should have the same number of edges that are in the
+edge table. The index is not checked for validity.
+
+See the :ref:`sec_c_api_table_indexes` section for details on the index
+life-cycle.
+@endrst
+
+@param self A pointer to a tsk_table_collection_t object.
+@param edge_insertion_order Array of tsk_id_t edge ids
```
jeromekelleher, line 2737: Full stop missing here and next

python/lwt_interface/tskit_lwt_interface.h
```diff
@@ -1339,6 +1402,21 @@ parse_table_collection_dict(tsk_table_collection_t *tables, PyObject *tables_dic
         goto out;
     }
 
+    /* index */
+    value = get_table_dict_value(tables_dict, "index", false);
```
jeromekelleher, line 1406: Probably best to call this "indexes" just for consistency with the C struct?

jeromekelleher (commented):
LGTM, minor comments.

jeromekelleher:
Need to do some tests where we add bad indexes from the dict encoding. Things like:

- Negative/out of bounds edge IDs
- Nonsensical indexes (np.zeros for both, e.g.). What error occurs here when we try to run ``ts.trees()``?

python/lwt_interface/tskit_lwt_interface.h
benjeffery: For consistency we should also change the python property on `TableCollection` to `indexes` but then that is inconsistent with `build_index` and `has_index` which would be breaking changes to move to `indexes`... we may have to live with the python API being `index` and the c struct being `indexes`.

benjeffery:
> Need to do some tests where we add bad indexes from the dict encoding. Things like:
> 
> * Negative/out of bounds edge IDs
> * Nonsensical indexes (np.zeros for both, e.g.). What error occurs here when we try to run `ts.trees()`?

As it was nothing happened as indexes were always rebuilt on `load_tables` I'm adding a `build_indexes` flag to the python API in #916

benjeffery:
Almost there, needs a few more tests so setting to draft.

benjeffery:
@jeromekelleher I think this is ready to go - I've changed the sematics such that `TableCollection`s without indexes now give `None` instead of a dict with empty arrays. Previously roundtripping a `TableCollection` without indexes resulted in adding zero-length indexes, which didn't seem right!

python/_tskitmodule.c
```diff
@@ -5226,6 +5226,11 @@ TableCollection_get_indexes(TableCollection *self, void *closure)
         goto out;
     }
 
+    if (!tsk_table_collection_has_index(self->tables, 0)) {
```
jeromekelleher, line 5229: I don't know about this. I think it would be logical to return an empty dictionary. That way we're more robust to adding more indexes in the future, where the dictionary may contain the current edge_x indexes, but also maybe any future indexes. So, we might have to deal with the situation where the dictionary contains a variable number of keys - logically then, it makes sense to treat the zero-keys situation in the same way.

Think of ``tsk_table_collection_has_index(self, 0)`` as saying "do you have all possible indexes". So, we'd have (roughly)
```
dict = PyDict_New();
if (tsk_table_collection_has_index(self->tables, 0)) {
    // add key-value pairs for each edge_insertion and edge_removal
}
```
We can see how this is then extensible to having more indexes in the future - we can just test the existance of the each index on a case-by-case basis.

jeromekelleher (commented):
I'm not sure None is the right encoding for missing indexes here - see comment above.

python/_tskitmodule.c
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/921/commits/2855dec9485874d744de5d9a3a5da17c25c648b5

python/lwt_interface/CHANGELOG.rst
```diff
@@ -0,0 +1,6 @@
+--------------------
+[0.1.2] - 2020-10-22
+--------------------
+
+ - Added optional top-level key ``index`` which has contains ``edge_insertion_order`` and
```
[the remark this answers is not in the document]
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/921/commits/9618a3b3440cb4125f5fef0342e497cee9631681

python/lwt_interface/tskit_lwt_interface.h
```diff
@@ -1566,9 +1586,19 @@ write_table_arrays(tsk_table_collection_t *tables, PyObject *dict)
...
         { "provenances", provenance_cols, NULL, 0 },
+        /* We don't want to insert empty indexes, return an empty dict if there are none
+         */
+        { "indexes",
+            tsk_table_collection_has_index(tables, 0) ? indexes_cols : no_indexes_cols,
+            NULL, 0 },
     };
 
     for (j = 0; j < sizeof(table_descs) / sizeof(*table_descs); j++) {
+        if (strcmp(table_descs[j].name, "indexes") == 0
```
jeromekelleher, line 1597: Do we need to do this? I think it'd be better if the dict encoding had the empty dictionary as the output rather than excluding the key. This is what the TableCollection does, right?

jeromekelleher (commented):
Not sure the lwt encoding version is the same as the output of asdict() now? We should probably add a test that these are identical, somewhere.

benjeffery:
@jeromekelleher I've tidied this up and added the test for identical contents in the dicts.

python/lwt_interface/tskit_lwt_interface.h
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/921/commits/4d815295f13d97eaf2912d04cbdad09e6ef485d2

jeromekelleher (approved):

jeromekelleher:
LGTM, merge away whenever

PR #930, opened 2020-10-24
Allow span to be set when generating trees

hyanwong:
## Description

Extend the functionality of `Tree.unroot`, `tskit.all_trees` etc to allow trees with spans other than 1. Passing `**kwargs` means that we can also extends this if (for example) we want to be able to set node/root times on the generated trees. Also document that a new ts is created for each new tree, and change the API section titles "Metadata" and "Combinatorics" so they aren't confused with the pages of those names.

### DIFF
--- a/python/tskit/combinatorics.py
+++ b/python/tskit/combinatorics.py
@@ -255,7 +255,7 @@ class PartialTopologyCounter:
         return RankTree(children).rank()
 
 
-def all_trees(num_leaves):
+def all_trees(num_leaves, span=1):
     """
     Generates all unique leaf-labelled trees with ``num_leaves``
     leaves. See :ref:`sec_combinatorics` on the details of this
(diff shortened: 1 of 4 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/930/)

python/tskit/combinatorics.py
```diff
@@ -255,7 +255,7 @@ def join_topologies(child_topologies):
         return RankTree(children).rank()
 
 
-def all_trees(num_leaves):
+def all_trees(num_leaves, **kwargs):
```
jeromekelleher, line 258: I'm not sure about this - why not just have ``span=None`` here? If we don't expect to have lots of parameters, then it's just making it hard for users to know where parameters come from. It's the same amount of typing, for now.

jeromekelleher (commented):
Probably a useful thing to have. I'd be hesitant about invoking kwargs here though, we don't want to end up in matplotlib territory.

python/tskit/combinatorics.py
hyanwong: I um-ed and ah-ed about putting it in. The argument for doing so is that we might want to add other features such as specifying branch length or node labels, which we would simply add to the underlying `to_tsk_tree()` method, and then get all the rest for free. But I could explicitly specify `span` for the time being (and repeat the docstring description for all the methods that call  `to_tsk_tree()`)

hyanwong:
OK, I think I've addressed all the comments now @jeromekelleher - just took the basic option of setting `span=1` for all the functions - that way at least it's clearly documented (and I can't see any other default value would be sensible).

python/tskit/combinatorics.py
```diff
@@ -429,7 +432,7 @@ def label_rank(self):
         return self._label_rank
 
     @staticmethod
-    def unrank(rank, num_leaves, labels=None):
+    def unrank(rank, num_leaves, labels=None, **kwargs):
```
jeromekelleher, line 435: There's an extra ``**kwargs`` here

python/tskit/trees.py
```diff
@@ -859,11 +859,15 @@ def unrank(rank, num_leaves):
 
         :param tuple(int) rank: The rank of the tree to generate.
         :param int num_leaves: The number of leaves of the tree to generate.
+        :param float span: The genomic span of the returned tree. The tree will cover
+            the interval :math:`[0, L)` and the :attr:`~Tree.tree_sequence` from which
```
jeromekelleher, line 863: You don't define what L is here.

jeromekelleher (commented):
Few more issues.

python/tskit/combinatorics.py
hyanwong: Well spotted: I though I had found all those. Sorry.

python/tskit/trees.py
hyanwong: Good point. Copy/paste error

hyanwong:
Done, thanks @jeromekelleher

hyanwong:
Closing and reopening to force a travis rebuild

jeromekelleher (approved):

PR #931, opened 2020-10-24
Allow newicks to be output without branch lengths

hyanwong:
As discussed on slack. This only changes the Python newick generator. I also took the opportunity to make all the parameters keyword-only (which is a breaking change), but it's really not obvious to me what `tree.newick(8)` means - I think we should force `tree.newick(precision=8)`.

Incidentally, I was wondering if `__build_newick` should be `_build_newick` (single underscore) for consistency. But perhaps the double underscore was deliberate?

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,6 +4,9 @@
 
 **Features**
 
+- ``Tree.newick()`` now has extra option ``include_branch_lengths`` to allow branch
+  lengths to be omitted (:user:`hyanwong`, :pr:`931`).
+
 - Added ``Tree.generate_star`` static method to create star-topologies (:user:`hyanwong`,
   :pr:`934`).
 
@@ -28,6 +31,7 @@
 **Breaking changes**
 
 - The argument to ``ts.dump`` and ``tskit.load`` has been renamed `file` from `path`.
+- All arguments to ``Tree.newick()`` except precision are now keyword-only.
 
 --------------------
 [0.3.2] - 2020-09-29
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2092,6 +2092,28 @@ class TestTree(HighLevelTestCase):
                 newick_c = tree.newick(precision=precision)
                 assert newick_c == newick_py
 
+    def test_bifurcating_newick(self):
+        for n_tips in range(2, 6):
+            ts = msprime.simulate(n_tips, random_seed=1)  # msprime trees are binary
+            for tree in ts.trees():
+                base_newick = tree.newick(include_branch_lengths=False).strip(";")
+                for i in range(n_tips):
+                    # Each tip number (i+1) mentioned once
+                    assert base_newick.count(str(i + 1)) == 1
+                # Binary newick trees have 3 chars per extra tip: "(,)"
+                assert len(base_newick) == n_tips + 3 * (n_tips - 1)
+
+    def test_newick_topology_equiv(self):
+        replace_numeric = {ord(x): None for x in "1234567890:."}
+        for ts in get_example_tree_sequences():
+            for tree in ts.trees():
+                if tree.num_roots > 1:
+                    continue
+                plain_newick = tree.newick(node_labels={}, include_branch_lengths=False)
+                newick1 = tree.newick().translate(replace_numeric)
+                newick2 = tree.newick(node_labels={}).translate(replace_numeric)
+                assert newick1 == newick2 == plain_newick
+
     def test_newick_buffer_too_small_bug(self):
         nodes = io.StringIO(
             """\
--- a/python/tests/test_phylo_formats.py
+++ b/python/tests/test_phylo_formats.py
@@ -89,22 +89,73 @@ class TestNewick(TreeExamples):
     external Newick parser.
     """
 
-    def verify_newick_topology(self, tree, root=None, node_labels=None):
+    # 3 methods to return example tree sequences with internal samples:
+    # (copied from test_highlevel.py)
+    def all_nodes_samples_example(self):
+        n = 5
+        ts = msprime.simulate(n, random_seed=10, mutation_rate=5)
+        assert ts.num_mutations > 0
+        tables = ts.dump_tables()
+        nodes = tables.nodes
+        flags = nodes.flags
+        # Set all nodes to be samples.
+        flags[:] = tskit.NODE_IS_SAMPLE
+        nodes.flags = flags
+        return tables.tree_sequence()
+
+    def only_internal_samples_example(self):
+        n = 5
+        ts = msprime.simulate(n, random_seed=10, mutation_rate=5)
+        assert ts.num_mutations > 0
+        tables = ts.dump_tables()
+        nodes = tables.nodes
+        flags = nodes.flags
+        # Set just internal nodes to be samples.
+        flags[:] = 0
+        flags[n:] = tskit.NODE_IS_SAMPLE
+        nodes.flags = flags
+        return tables.tree_sequence()
+
+    def mixed_node_samples_example(self):
+        n = 5
+        ts = msprime.simulate(n, random_seed=10, mutation_rate=5)
+        assert ts.num_mutations > 0
+        tables = ts.dump_tables()
+        nodes = tables.nodes
+        flags = nodes.flags
+        # Set a mixture of internal and leaf samples.
+        flags[:] = 0
+        flags[n // 2 : n + n // 2] = tskit.NODE_IS_SAMPLE
+        nodes.flags = flags
+        return tables.tree_sequence()
+
+    def verify_newick_topology(
+        self, tree, root=None, node_labels=None, include_branch_lengths=True
+    ):
         if root is None:
             root = tree.root
-        ns = tree.newick(precision=16, root=root, node_labels=node_labels)
+        ns = tree.newick(
+            precision=16,
+            root=root,
+            node_labels=node_labels,
+            include_branch_lengths=include_branch_lengths,
+        )
         if node_labels is None:
             leaf_labels = {u: str(u + 1) for u in tree.leaves(root)}
         else:
             leaf_labels = {u: node_labels[u] for u in tree.leaves(root)}
-        newick_tree = newick.loads(ns)[0]
+        # default newick lib outputs 0.0 if length is None => replace the length_parser
+        newick_tree = newick.loads(
+            ns, length_parser=lambda x: None if x is None else float(x)
+        )[0]
         leaf_names = newick_tree.get_leaf_names()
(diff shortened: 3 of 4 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/931/)

python/CHANGELOG.rst
```diff
@@ -14,6 +17,7 @@
 **Breaking changes**
 
 - The argument to ``ts.dump`` and ``tskit.load`` has been renamed `file` from `path`.
+- Arguments to ``Tree.newick()`` are now all positional.
```
jeromekelleher, line 20: This is a bit gratuitous, I think. People will be using this code and it will probably break

python/tskit/trees.py
```diff
@@ -2148,10 +2148,10 @@ def nodes(self, root=None, order="preorder"):
                 yield from iterator(u)
 
     # TODO make this a bit less embarrassing by using an iterative method.
-    def __build_newick(self, node, precision, node_labels):
+    def __build_newick(self, node, precision, node_labels, branch_lengths):
```
jeromekelleher, line 2151: This is a good candidate for mandatory keyword only args.

python/tskit/trees.py
```diff
@@ -2160,12 +2160,16 @@ def __build_newick(self, node, precision, node_labels):
...
+                    child, precision, node_labels, branch_lengths
+                )
+                if branch_lengths:
+                    subtree += ":{0:.{1}f}".format(branch_length, precision)
+                s += subtree + ","
             s = s[:-1] + f"){label}"
         return s
 
-    def newick(self, precision=14, root=None, node_labels=None):
+    def newick(self, *, precision=14, root=None, node_labels=None, branch_lengths=True):
```
jeromekelleher, line 2172: This is a bit gratuitous I think and will probably break people's code (the newick method has been around a long time). I think making everything after ``precision`` keyword only is probably fine,though. (We can say in a comment that precision is positional for legacy only)

python/tskit/trees.py
```diff
@@ -2187,6 +2191,8 @@ def newick(self, precision=14, root=None, node_labels=None):
         :param dict node_labels: If specified, show custom labels for the nodes
             that are present in the map. Any nodes not specified in the map will
             not have a node label.
+        :param branch_lengths: If True (default), output branch lengths in the Newick
```
jeromekelleher, line 2194: Hmm, looking at this again in the presence of the other parameters I wonder if it's a bit misleading and it seems like we're providing a way to specify what hte branch lenghts *are*. Perhaps change to ``include_branch_lengths``? Probably better to be explicit.

jeromekelleher (changes requested):
Few changes needed. The double underscore was deliberate I think, as a very-private indeed embarrassing recursive function.

python/tskit/trees.py
hyanwong: > The double underscore was deliberate I think, as a very-private indeed embarrassing recursive function.

I thought that double underscore didn't mean "very private", but instead marked it for name mangling: https://dbader.org/blog/meaning-of-underscores-in-python

python/tests/test_highlevel.py
```diff
@@ -2054,6 +2054,17 @@ def test_newick_large_times(self):
                 newick_c = tree.newick(precision=precision)
                 assert newick_c == newick_py
 
+    def test_newick_nolengths(self):
```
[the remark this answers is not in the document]
hyanwong: It should do. It's perfectly valid newick format to omit branch lengths.

python/tskit/trees.py
jeromekelleher: Yes, that's what happens in practice.

hyanwong:
I've addressed all the change now.

python/tskit/trees.py
```diff
@@ -2198,6 +2214,9 @@ def newick(self, precision=14, root=None, node_labels=None):
                     "newick trees, one for each root."
                 )
             root = self.root
+        if not include_branch_lengths and node_labels is None:
+            # Force the python generator for simplicity, by specifying the default labels
```
jeromekelleher, line 2218:
```suggestion
            # The C engine always includes branch lengths so we force the Python engine by specifying the default labels
```

python/tskit/trees.py
```diff
@@ -2198,6 +2214,9 @@ def newick(self, precision=14, root=None, node_labels=None):
...
+            node_labels = {i: str(i + 1) for i in self.leaves()}
```
jeromekelleher, line 2219: Hmm, ``leaves()`` or ``samples()``? I think it should be ``samples()``. Leaves is tricky, as we can have non-sample leaves. 

What does the C engine do? We should add a test for this, since we are doing this workaround.

Maybe it would be better to not do this, and to change the condition on the line below instead to

```
if node_labels is None and include_branch_lengths:
```

Then it's completely explicit and we don't need any comments explaining what we're doing.

jeromekelleher (changes requested):
Probably better not to do the node_labels workaround, actually, so we don't have to test that the labels are the same in various corner cases.

python/tskit/trees.py
hyanwong: [We claim](https://tskit.readthedocs.io/en/latest/python-api.html?highlight=newick#tskit.Tree.newick) it does leaves only (regardless of whether they are samples), and that seems sensible to me. Some newick readers only bother with names for tips anyway. In particular, the docs currently say:

```
By default, leaf nodes are labelled with their numerical ID + 1, and internal nodes are not labelled.
```

hyanwong: > Then it's completely explicit and we don't need any comments explaining what we're doing.

We would still need to set the node labels to the leaves only when `node_labels == None` for the python Newick generating code. But happy to move the logic around.

jeromekelleher: Ah right, what you have sounds good then. We should add a test to verify that we get the same labels with and without ``include_branch_lengths`` in a corner case when we have non-sample leaves, though.

hyanwong:
This is ready for final review, @jeromekelleher. Note that to use the leaves != samples examples (specifically `get_internal_samples_examples()`) in test_highlevel.py, I've moved a load of tree-sequence-creating routines (e.g. `get_example_tree_sequences`) from `test_highlevel.py` to `tsutil.py`. I think that's probably a better place for them, as then they can all be used in different unit test files.

python/tests/test_phylo_formats.py
```diff
@@ -89,22 +89,37 @@ class TestNewick(TreeExamples):
     external Newick parser.
     """
 
-    def verify_newick_topology(self, tree, root=None, node_labels=None):
+    @staticmethod
+    def length_parser(x):
```
jeromekelleher, line 93: I don't think we need a static method for this, do we? A static method implies this function is useful outside the immediate test context, whereas this is really just for newick parsing.

 Seems like a good use for a lambda function:
```
 # By default the newick lib outputs length 0.0 if no branch lengths are present
newick_tree = newick.loads(ns, length_parser=lambda x: None if x is None else float(x))[0]
```

python/tests/test_highlevel.py
```diff
@@ -55,214 +54,6 @@
...
-    # Give an example with a gap at the end.
-    ts = msprime.simulate(10, random_seed=5, recombination_rate=1)
-    tables = get_table_collection_copy(ts.dump_tables(), 2)
-    tables.sites.clear()
-    tables.mutations.clear()
-    insert_uniform_mutations(tables, 100, list(ts.samples()))
-    yield tables.tree_sequence()
-
-
-def get_internal_samples_examples():
```
jeromekelleher, line 152: I don't think we want to promote this code up to utils.py. The pattern of yielding a bunch of different examples isn't a great one as it leads to long-running tests where it's hard to distinguish which test has failed. I would prefer to leave this module as-is, and continue to factor tests out of it and into more appropriate places, rather than have this deprecated testing pattern diffuse out into other modules.

Creating a ts with a non-sample leaf is pretty easy:
```python
tables = msprime.simulate(n, random_seed=1).dump_tables()
root = len(tables.nodes) - 1
leaf = tables.nodes.add_row(flags=0, time=0)
tables.edges.add_row(0, 1, root, leaf)
tables.sort()
ts = tables.tree_sequence()
```

jeromekelleher (commented):
Looks good, but let's not do the test refactoring.

python/tests/test_phylo_formats.py
hyanwong: Happy to use lambda

python/tests/test_highlevel.py
hyanwong: Ah, OK. I do think we want a nice range of trees & tree sequences that we call upon from different test functions, rather than writing them all out again (e.g. the example above doesn't test internal samples, which would have been another good check). But this particular refactoring seems not to be the right way to go about it.

jeromekelleher:
Can you bring this up to date please @hyanwong? A few small changes and we're good to merge I think.

hyanwong:
> Can you bring this up to date please @hyanwong? A few small changes and we're good to merge I think.

In the latest push, I've split the `get_internal_samples_examples()` code into 3 separate functions (`all_nodes_samples_example` `only_internal_samples_example` `mixed_node_samples_example`)and put them at the start of the class, but this does lead to quite a lot of code duplication. Suggestions for improving this are welcome - it feels a bit wrong to me.

benjeffery:
The pytest way to do this is a parameterised fixture:
```python
@pytest.fixture(scope="module", params=[(None, None), (5,None), (5//2, n+5//2)], ids=['All', 'Internal', 'Mixed'])
def ts_with_varying_sample_nodes(sample_node_slice):
    ts = msprime.simulate(n, random_seed=10, mutation_rate=5)
    assert ts.num_mutations > 0
    tables = ts.dump_tables()
    nodes = tables.nodes
    flags = nodes.flags
    # Set a mixture of internal and leaf samples.
    flags[:] = 0
    flags[slice(sample_node_slice)] = tskit.NODE_IS_SAMPLE
    nodes.flags = flags
    return tables.tree_sequence()
```

Then your tests use the fixture:
```python
def test_samples_differ_from_leaves(self, ts_with_varying_sample_nodes):
        for t in ts.trees():
            self.verify_newick_topology(t)
```

hyanwong:
Great, thanks @benjeffery - should we do this for the bevvy of tests in test_highlevel.py - or at least make the various test sequences available for all files in `tests`? Presumably that way we can stick them in `tsutils.py` and grab them from there in the various `test_XXX files`?

benjeffery:
Suite-wide fixtures are in `conftest.py` and where possible should be `scope="session"` to avoid running msprime a million times.

jeromekelleher (approved):

PR #934, opened 2020-10-25
Add a Tree.generate_star method

hyanwong:
## Description

Add `Tree.create_star`, `Tree.create_balanced`, and `Tree.create_empty` methods, as discussed on Slack.

The `Tree.create_empty` method is perhaps unlikely to be widely used, but I think is useful for completeness, and will also come in handy for testing (for example, I only just noticed that we currently bomb out with a non-obvious message when trying to plot an empty TS).

Happy to make these class methods if that is preferred.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,6 +4,9 @@
 
 **Features**
 
+- Added ``Tree.generate_star`` static method to create star-topologies (:user:`hyanwong`,
+  :pr:`934`).
+
 - Added ``equals`` method to TableCollection and each of the tables which
   provides more flexible equality comparisons, for example, allowing
   users to ignore metadata or provenance in the comparison.
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -7917,3 +7917,37 @@ class TestMissingData:
                 tree.is_isolated("abc")
             with pytest.raises(TypeError):
                 tree.is_isolated(1.1)
+
+
+class TestExampleTrees:
+    """
+    Test hard-coded example tree (sequence) generation
+    """
+
+    def test_star_equivalent(self):
+        for extra_params in [{}, {"span": 2.5}]:
+            for n in range(2, 6):
+                ts = tskit.Tree.generate_star(n, **extra_params).tree_sequence
+                equiv_ts = tskit.Tree.unrank((0, 0), n, **extra_params).tree_sequence
+                assert ts.tables.equals(equiv_ts.tables, ignore_provenance=True)
+
+    def test_star_bad_params(self):
+        for n in [-1, 0, 1, np.array([1, 2])]:
+            with pytest.raises(ValueError):
+                tskit.Tree.generate_star(n)
+        for n in [None, "", []]:
+            with pytest.raises(TypeError):
+                tskit.Tree.generate_star(n)
+        with pytest.raises(tskit.LibraryError):
+            tskit.Tree.generate_star(2, span=0)
+        with pytest.raises(tskit.LibraryError):
+            tskit.Tree.generate_star(2, branch_length=0)
+
+    def test_star_branch_length(self):
+        branch_length = 10
+        n = 7
+        ts = tskit.Tree.generate_star(n, branch_length=branch_length).tree_sequence
+        topological_equiv_ts = tskit.Tree.unrank((0, 0), n).tree_sequence
+
+        assert ts.node(ts.first().root).time == branch_length
+        assert ts.kc_distance(topological_equiv_ts) == 0
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -30,6 +30,7 @@ import concurrent.futures
 import copy
 import functools
 import itertools
+import json
 import math
 import os
 import textwrap
@@ -45,6 +46,7 @@ import tskit.drawing as drawing
 import tskit.exceptions as exceptions
 import tskit.formats as formats
 import tskit.metadata as metadata_module
+import tskit.provenance as provenance
 import tskit.tables as tables
 import tskit.util as util
 import tskit.vcf as vcf
@@ -2341,6 +2343,52 @@ class Tree:
         :rtype: float
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
+
+    @staticmethod
+    def generate_star(num_leaves, *, span=1, branch_length=1, record_provenance=True):
+        """
+        Generate a single :class:<Tree> whose leaf nodes all have the same parent (i.e.
+        a "star" tree). The leaf nodes are all at time 0 and are marked as sample nodes.
+
+        .. note::
+            This is similar to ``tskit.Tree.unrank((0,0), n, span=span)`` but is more
+            efficient for large n. However, the ``unrank`` method provides
+            a concise way of generating alternative (non-star) topologies.
+
+        :param int num_leaves: The number of leaf nodes in the returned tree (must be
+            be 2 or greater).
+        :param float span: The span of the tree, and therefore the
+            :attr:`~TreeSequence.sequence_length` of the :attr:`.tree_sequence`
+            property of the returned :class:<Tree>.
+        :param float branch_length: The length of every branch in the tree (equivalent
+            to the time of the root node).
+        :return: A star-shaped tree. Its corresponding :class:`TreeSequence` is available
+            via the :attr:`.tree_sequence` attribute.
+        :rtype: Tree
+        """
+        if num_leaves < 2:
+            raise ValueError("The number of leaves must be 2 or greater")
+        tc = tables.TableCollection(sequence_length=span)
+        tc.nodes.set_columns(
+            flags=np.full(num_leaves, NODE_IS_SAMPLE, dtype=np.uint32),
+            time=np.zeros(num_leaves),
+        )
+        root = tc.nodes.add_row(time=branch_length)
+        tc.edges.set_columns(
+            left=np.full(num_leaves, 0),
+            right=np.full(num_leaves, span),
+            parent=np.full(num_leaves, root, dtype=np.int32),
+            child=np.arange(num_leaves, dtype=np.int32),
+        )
+        if record_provenance:
+            # TODO replace with a version of https://github.com/tskit-dev/tskit/pull/243
+            # TODO also make sure we convert all the arguments so that they are
+            # definitely JSON encodable.
+            parameters = {"command": "generate_star", "TODO": "add parameters"}
+            tc.provenances.add_row(
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+        return tc.tree_sequence().first()
 
 
 def load(file):


hyanwong:
Leaving as a draft until #930 is merged, so I can uncomment a test for `span` in the unit tests.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/934/)

python/tests/test_topology.py
```diff
@@ -7917,3 +7917,69 @@ def test_is_isolated_bad(self):
...
                 tree.is_isolated(1.1)
+
+
+class TestExampleTrees:
+    """
+    Test the creation of hard-coded example tree (sequences)
+    """
+
+    def test_star_equivalent(self):
+        for extra_params in [{}]:  # , {'span':2.5}]: # Add after #930 is merged
```
jeromekelleher, line 7928: There's only one extra param, so loop seems unnecessary and makes the code less readable

python/tests/test_topology.py
```diff
@@ -7917,3 +7917,69 @@ def test_is_isolated_bad(self):
...
+            for n in range(0, 6):
+                ts = tskit.Tree.create_star(n, **extra_params).tree_sequence
+                if n == 0:
```
jeromekelleher, line 7931: I think it'd be clearer to have a separate ``test_empty`` method, rather than embedding this into a loop.

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            # TODO also make sure we convert all the arguments so that they are
+            # definitely JSON encodable.
+            parameters = {"command": "create_empty", "TODO": "add parameters"}
+            tc.provenances.add_row(
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+        return tc.tree_sequence().first()
+
+    @staticmethod
+    def create_star(n, *, span=1, branch_length=1, record_provenance=True):
```
jeromekelleher, line 2361: Maybe ``num_leaves`` rather than n - people might think n means the total number of nodes.

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+        The leaf nodes are all at time 0 and are marked as sample nodes.
+
+        :param int n: The number of leaf nodes in the tree
+        :param float span: The span of the tree, and therefore the
+            :attr:`~TreeSequence.sequence_length` of the :attr:`~Tree.tree_sequence`
+            property of the returned :class:<Tree>
+        :param float branch_length: The length of every branch in the tree (equivalent
+            to the time of the root node).
+
+        .. note::
```
jeromekelleher, line 2374: The note should go before the param blocks. Also we're missing a return description.

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            This is similar to ``tskit.Tree.unrank((0,0), n, span=span)`` but is more
+            efficient for large n. However, the ``unrank`` method provides
+            a concise way of generating alternative (non-star) topologies.
+        """
+        tc = tables.TableCollection(sequence_length=span)
+        tc.nodes.set_columns(
+            flags=np.full(n, NODE_IS_SAMPLE, dtype=np.uint32),
+            time=np.zeros(n),
+        )
+        if n > 1:
```
jeromekelleher, line 2384: I don't think we should make this exception. We should always return a tree with ``num_leaves`` + 1 nodes and num_leaves edges. (which holds true for all num_leaves >= 0, right?)

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+        return tc.tree_sequence().first()
+
+    @staticmethod
+    def create_star(n, *, span=1, branch_length=1, record_provenance=True):
+        """
+        Create a single :class:<Tree> and associated tree sequence where the ``n``
+        leaf nodes in the tree all have the same parent (i.e. a "star" tree).
+        The leaf nodes are all at time 0 and are marked as sample nodes.
```
jeromekelleher, line 2365: We should add a note here (and to the others) "The corresponding :class:`TreeSequence`` is available via the ``.tree_sequence`` attribute of the returned tree". 

The first line at the moment implies to me that we're returning a tuple (tree, ts), so I think this note would be more helpful and not distract.

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            parameters = {"command": "create_empty", "TODO": "add parameters"}
+            tc.provenances.add_row(
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+        return tc.tree_sequence().first()
+
+    @staticmethod
+    def create_star(n, *, span=1, branch_length=1, record_provenance=True):
+        """
+        Create a single :class:<Tree> and associated tree sequence where the ``n``
```
jeromekelleher, line 2363: Delete "and associated tree sequence" - see the note below.

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+        return tc.tree_sequence().first()
+
+    @staticmethod
+    def create_balanced(n, *, span=1, branch_length=1, record_provenance=True):
+        """
+        Create a single bifurcating :class:<Tree> and associated tree sequence where
+        the tree is a completely balanced (i.e. all parents have exactly 2 children,
+        and all leaves go through the the same number of nodes to get to the root).
```
jeromekelleher, line 2407: Same comments as above on the arrangement of the docstring

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            ``tskit.Tree.unrank((tskit.combinatorics.num_shapes(n)-1, 0), n, span=span)``
+            but is more efficient for large n. However, the ``unrank`` method provides
+            a concise way of generating alternative (non-balanced) topologies.
+        """
+        tc = tables.TableCollection(sequence_length=span)
+        tc.nodes.set_columns(
+            flags=np.full(n, NODE_IS_SAMPLE, dtype=np.uint32),
+            time=np.zeros(n),
+        )
+        if n & (n - 1) != 0:
```
jeromekelleher, line 2427: Seems a bit lame to insist ``num_leaves`` is a power of 2. Why not make the tree as balanced as possible? It'll be the same algorithm I'd imagine. It would be nice if this was also identical to ``tskit.Tree.unrank(N - 1, 0)``

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            raise ValueError("n must be an integer power of 2")
+        if n > 1:
```
jeromekelleher, line 2429: Again, I don't like the special case. We shouldn't need it here, right?

python/tskit/trees.py
```diff
@@ -2338,6 +2340,113 @@ def kc_distance(self, other, lambda_=0.0):
...
+            merge_nodes = list(range(n))
+            while len(merge_nodes):
```
jeromekelleher, line 2431: len(merge_nodes) > 0

jeromekelleher (changes requested):
Looks good, thanks @hyanwong. These will be really handy, I think. The main feedback is we shouldn't have special cases, and try to make simple rules about the number of nodes and edges returned hold for all n >= 0.

I agree it's nice to have all these have a common prefix. I'm not so sure about "create_" though; perhaps "make_"?

I'm not sure how useful the ``empty`` one is to be honest, it's already a one-liner, and I don't think the equivalances you're stating will hold once the special cases at 1 and 0 have been removed.

python/tests/test_topology.py
hyanwong: I've just uncommented the bit with "span" in, so it will test with with the default, and with a span now.

python/tests/test_topology.py
hyanwong: Sure

python/tskit/trees.py
hyanwong: Good point. That will match `unrank` too.

python/tskit/trees.py
hyanwong: I was trying to match `unrank`, @jeromekelleher . For num_leaves=1 that produces a TS with a single node and no edges. I think this is the right behaviour. We don't really want a unary node as the root, do we - it would be removed by `simplify()` if we were to ruin that on the TS anyway.

hyanwong:
> I agree it's nice to have all these have a common prefix. I'm not so sure about "create_" though; perhaps "make_"?

Or `generate`? Happy with any of those.

> I'm not sure how useful the `empty` one is to be honest, it's already a one-liner, and I don't think the equivalances you're stating will hold once the special cases at 1 and 0 have been removed.

I think it's useful so that we can test e.g. `tskit.Tree.generate_empty().draw_text()` and so on, which we have currently failed to do. It's also useful to be able to compare things to an empty TS, unless there is already some function to check if a TS is empty (didn't check, sorry!)

python/tskit/trees.py
hyanwong: Well, then the tree isn't actually balanced. So if anyone was to run it to test e.g. measures of tree balance, then they would fail. Given the name of the function, I think we should insist on creating a fully balanced tree, unless e.g. there is a flag "allow_partially_unbalanced" (default False) so that users have to say that they are completely sure that's what they want.

jeromekelleher:
> It's also useful to be able to compare things to an empty TS, unless there is already some function to check if a TS is empty (didn't check, sorry!)

Generating an empty ts is pretty trivial: ``tskit.TableCollection(1).tree_sequence()`` - but I don't feel particularly strongly about having an ``x_empty`` function or not. I'm not sure "empty" is the right word anyway (it's implying there's something *in* the tree), it's more like a "null" tree. I'm not sure it's worth the discussion of the semantics to include it.

python/tskit/trees.py
jeromekelleher: It's as balanced as possible for the given number of leaves though. This will basically be useless if we only allow powers of 2. What do we get for ``unrank(N - 1)`` for num_leaves = 5, say?

python/tskit/trees.py
jeromekelleher: Hmm. Maybe we should just raise a ValueError if n < 2 for now, and avoid this discussion. It's not particularly interesting what the semantics are in these degenerate cases and we can always fix them later if it does become important.

hyanwong: Yep, could do, I suppose.

python/tskit/trees.py
hyanwong: > > I agree it's nice to have all these have a common prefix. I'm not so sure about "create_" though; perhaps "make_"?
> 
> Or `generate`? Happy with any of those.

Just realised that "make_star" might imply that we are forcing an existing tree into a star shape. So if you don't like "create", I would prefer "generate". Perhaps @benjeffery might care to adjudicate?

hyanwong:
> Generating an empty ts is pretty trivial: `tskit.TableCollection(1).tree_sequence()` - but I don't feel particularly strongly about having an `x_empty` function or not.

That's not an obvious calling convention to people unfamiliar with tables. E.g. I might hope to get an empty ("null") TS by simply calling `tskit.TreeSequence()` with no params (and I did try that, when messing around). I only added it so that I could be explicit.

> I'm not sure "empty" is the right word anyway (it's implying there's something _in_ the tree), it's more like a "null" tree. I'm not sure it's worth the discussion of the semantics to include it.

It never struck me as lacking before. But I do think we have somewhat overlooked at the behaviour of various tskit functions for "null" tree sequences & trees, so this was a first step in doing something about that.

python/tskit/trees.py
benjeffery: > Perhaps @benjeffery might care to adjudicate?

Happy with `make` or `generate`, with a very slight preference for `generate`.

jeromekelleher:
> But maybe it should be possible to create null objects by doing tskit.TreeSequence() and tskit.Tree()

The semantics around object initialisation are extremely tricky because of the interaction with the C API and Python C API. I'd rather avoid this can of worms for now.

hyanwong:
So it's not obvious to me how to create semi-balanced trees that match `unrank(N - 1)` without doing some sort of recursion. Given that there's disagreement over these "semi-balanced" trees anyway, I've simply removed the `generate_balanced` code. I'll put it in a different PR so we can discuss it there. I think this PR, that now just has `generate_star`, is therefore ready to merge @jeromekelleher

python/CHANGELOG.rst
```diff
@@ -4,12 +4,19 @@
...
+- Added ``Tree.generate_star`` static method to create star-topologies (:user:`hyanwong`,
+  :pr:`934`).
+
 - Added ``equals`` method to TableCollection and each of the tables which
   provides more flexible equality comparisons, for example, allowing
   users to ignore metadata or provenance in the comparison.
   (:user:`mufernando`, :user:`jeromekelleher`, :issue:`896`, :pr:`897`,
   :issue:`913`, :pr:`917`).
 
+- Added ``TableCollection.equals``, a method that compares two table
```
jeromekelleher, line 16: This is already in the changelog

jeromekelleher (commented):
Looks great, thanks @hyanwong. Just need to fix up the changelog and we're good to merge.

python/CHANGELOG.rst
hyanwong: Good spot, sorry. I obviously didn't rebase properly. Done now.

jeromekelleher (approved):

PR #935, opened 2020-10-25
Ibd constant time pair lookup

molpopgen:
The current IBD algorithm spends most of its time looking up to which, if any, input pair a pair of nodes belongs.

This PR adds a matrix to `tsk_ibd_finder_t` that allows `tsk_ibd_finder_find_sample_pair_index2` to operate in constant time, at the cost of some extra memory.

To make this work, I add a new error code that gets raised when duplicate pairs are passed to the ibd finding code.

In addition to the existing tests, I ran the following stochastic test:

```py
import pickle
import time

import msprime

print(msprime.tskit.__version__)

ts = msprime.simulate(1000, recombination_rate=25.0, random_seed=420)


def make_pairs(n):
    rv = []
    for i in range(n):
        for j in range(i + 1, n):
            rv.append([i, j])

    return rv


for i in range(100, ts.num_samples + 1, 100):
    pairs = make_pairs(i)
    then = time.time()
    ibd = ts.tables.find_ibd(pairs)
    now = time.time()
    elapsed = now - then
    print(i, elapsed)
    with open(f"ibd_output_{i}.pickle", "wb") as f:
        pickle.dump(ibd, f)
    if elapsed > 5e3:
        break
```

The output for the current stable release is:

```
0.3.2
100 23.053314208984375
200 370.5223250389099
300 1860.3008773326874
400 5864.632394313812
```

The output for the code in this PR is:

```
0.3.3.dev1
100 0.11817097663879395
200 0.5299818515777588
300 1.2625401020050049
400 2.2465524673461914
500 3.626859188079834
600 5.582508087158203
700 7.96460223197937
800 11.132724046707153
900 14.832526206970215
1000 19.45262336730957
```

The speedup is over 2000x!

The following code compared the IBD outputs:

```py
import os
import pickle

import numpy as np

for i in range(100, 1000 + 1, 100):
    fn = f"ibd_output_{i}.pickle"
    fn2 = f"/home/kevin/src/tskit/python/ibd_output_{i}.pickle"
    if os.path.exists(fn) and os.path.exists(fn2):
        with open(fn, "rb") as f:
            stable_results = pickle.load(f)
        with open(fn2, "rb") as f:
            pr_results = pickle.load(f)
        assert len(stable_results) == len(pr_results)

        for k in stable_results.keys():
            assert k in pr_results
        for k in pr_results.keys():
            assert k in stable_results

        for key, value in stable_results.items():
            pr_value = pr_results[key]
            for x in ["left", "right", "node"]:
                assert np.array_equal(value[x], pr_value[x])
        print(f"{i} passed")
```

They all passed.

# PR Checklist:

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -5516,6 +5516,93 @@ out:
     return ret;
 }
 
+static int
+tsk_ibd_finder_index_samples(tsk_ibd_finder_t *self)
+{
+    int ret = 0;
+    size_t i;
+    tsk_id_t idx;
+
+    self->paired_nodes_index
+        = calloc(self->num_nodes, sizeof(*self->paired_nodes_index));
+
+    if (self->paired_nodes_index == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    for (i = 0; i < self->num_nodes; ++i) {
+        self->paired_nodes_index[i] = -1;
+    }
+
+    for (i = 0; i < 2 * self->num_pairs; ++i) {
+        self->paired_nodes_index[self->pairs[i]] = 1;
+    }
+
+    self->num_unique_nodes_in_pair = 0;
+    idx = 0;
+    for (i = 0; i < self->num_nodes; ++i) {
+        if (self->paired_nodes_index[i] == 1) {
+            ++self->num_unique_nodes_in_pair;
+            self->paired_nodes_index[i] = idx;
+            ++idx;
+        }
+    }
+
+out:
+    return ret;
+}
+
+// NOTE: future travellers may want to refactor
+// this to only allocate the upper triangle of the matrix.
+// Doing so would save a good bit of memory, but would
+// imply trickier indexing here and in
+// tsk_ibd_finder_find_sample_pair_index2.
+static int
+tsk_ibd_finder_build_pair_map(tsk_ibd_finder_t *self)
+{
+    int ret = 0;
+    size_t i;
+    tsk_id_t sample0, sample1;
+    tsk_id_t row, col;
+    size_t matrix_size = self->num_unique_nodes_in_pair * self->num_unique_nodes_in_pair;
+
+    self->pair_map = calloc(matrix_size, sizeof(*self->pair_map));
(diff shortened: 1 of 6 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/935/)

benjeffery (approved):
Awesome! Wish I had set up the benchmark tracking already. I'll let people more familiar with the IBD code weigh in before merging, but it looks good to me.

c/tskit/tables.c
```diff
@@ -5498,6 +5498,92 @@ tsk_ibd_finder_init_samples(tsk_ibd_finder_t *self)
...
+    for (i = 0; i < 2 * self->num_pairs; ++i) {
+        self->paired_nodes_index[self->pairs[i]] = 1;
+    }
+
+    self->num_unique_nodes_in_pair = 0;
+    idx = 0;
+    for (i = 0; i < self->num_nodes; ++i) {
+        if (self->paired_nodes_index[i] == 1) {
+            self->num_unique_nodes_in_pair++;
+            self->paired_nodes_index[i] = idx++;
```
jeromekelleher, line 5529: We avoid using the return value of ``++`` - can you split this into two statements please?

jeromekelleher (changes requested):
This is great, thanks @molpopgen! @gtsambos and I knew there was some low-hanging fruit here, but didn't realise the perf impact would be quite so profound! We're planning on moving to some sort of sparse representation of the output once Georgia has done some profiling so we might end up using a different way to hash the sample pairs. But, your change will make this profiling meaningful now, so thanks a million.

One minor style nit and we're good to merge I think.

c/tskit/tables.c
molpopgen: Just pushed a fix for this.  I can also squash this entire thing down to 1 commit?

benjeffery: Squash away!

molpopgen:
So, this is ready to go, unless we want to deal w/the code coverage?

benjeffery:
Coverage is just hard to test error paths, merging!

gtsambos (approved):

PR #936, opened 2020-10-26
move stats functions into stats.h/c

castedo:
In working on multiple PRs for #844 I saw the TODO in trees.h to move the stats functions in trees.h/c into stats.h/.c. There are also some typedefs that are duplicated and re-written (with diff names) so this reconciles them into just stats.h.

This PR will also make the final PRs for #844 slightly simpler and more organized.


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/936/)

castedo:
Here's some info on the line number changes with this move.

From main:
-    429 trees.h
-    53 stats.h

to:
-   355 trees.h
-   150 stats.h

From main:
- 5035 trees.c
- 475 stats.c

to:
-  2691 trees.c
-  2821 stats.c

jeromekelleher:
Thanks for this @castedo, you're doing great work here refactoring and improving the codebase.

I'm not sure about this one, though. The idea was to keep all the ``tsk_treeseq_t`` and ``tsk_tree_t`` methods in the ``trees.[c,h]`` files, so that it's easy to know what files you need to compile and to keep things arranged somewhat logically. I agree that there's a case to be made for moving this code to ``stats.[c,h]``, but it does violate this simple rule.

I agree trees.c is a bit of a behemoth and it would help to break it up a little bit, but fundamentally I would much rather fewer big files than lots of little ones. One of the key goals here is to make tskit easy to embed, and part of that is making it easy to know what files you need to compile in order to get the functionality you want.

castedo:
@jeromekelleher That sounds good too. So I'll remove the TODO comment in one of the other PRs I'm doing. There's also the duplicated typedefs, but that's such a small change I'm thinking it's better to slip that into one of these other PRs. Let me know if for some reasons having duplicated typedefs in the Python/C layer and test code was intentional.

jeromekelleher:
> Let me know if for some reasons having duplicated typedefs in the Python/C layer and test code was intentional.

No, this was surely a mistake. Thanks a lot.

PR #939, opened 2020-10-26
constify C stats api

castedo:
For #844. This builds on top of #920 which did not constify all the "stats" functions in tree.h.

This PR constifies all those stats functions and fixes duplicated typedefs by keeping them defined only once in trees.h.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -336,7 +336,7 @@ out:
  */
 int TSK_WARN_UNUSED
 tsk_treeseq_init(
-    tsk_treeseq_t *self, tsk_table_collection_t *tables, tsk_flags_t options)
+    tsk_treeseq_t *self, const tsk_table_collection_t *tables, tsk_flags_t options)
 {
     int ret = 0;
     tsk_id_t num_trees;
@@ -579,8 +579,8 @@ tsk_treeseq_get_num_trees(const tsk_treeseq_t *self)
     return self->num_trees;
 }
 
-double *
-tsk_treeseq_get_breakpoints(tsk_treeseq_t *self)
+const double *
+tsk_treeseq_get_breakpoints(const tsk_treeseq_t *self)
 {
     return self->breakpoints;
 }
@@ -641,9 +641,10 @@ increment_nd_array_value(double *array, tsk_size_t n, const tsk_size_t *shape,
 /* TODO flatten the reference sets input here and follow the same pattern used
  * in diversity, divergence, etc. */
 int TSK_WARN_UNUSED
-tsk_treeseq_genealogical_nearest_neighbours(tsk_treeseq_t *self, tsk_id_t *focal,
-    size_t num_focal, tsk_id_t **reference_sets, size_t *reference_set_size,
-    size_t num_reference_sets, tsk_flags_t TSK_UNUSED(options), double *ret_array)
+tsk_treeseq_genealogical_nearest_neighbours(const tsk_treeseq_t *self,
+    const tsk_id_t *focal, size_t num_focal, tsk_id_t *const *reference_sets,
+    const size_t *reference_set_size, size_t num_reference_sets,
+    tsk_flags_t TSK_UNUSED(options), double *ret_array)
 {
     int ret = 0;
     tsk_id_t u, v, p;
@@ -823,8 +824,8 @@ out:
 }
 
 int TSK_WARN_UNUSED
-tsk_treeseq_mean_descendants(tsk_treeseq_t *self, tsk_id_t **reference_sets,
-    size_t *reference_set_size, size_t num_reference_sets,
+tsk_treeseq_mean_descendants(const tsk_treeseq_t *self, tsk_id_t *const *reference_sets,
+    const size_t *reference_set_size, size_t num_reference_sets,
     tsk_flags_t TSK_UNUSED(options), double *ret_array)
 {
     int ret = 0;
@@ -989,7 +990,8 @@ out:
  ***********************************/
 
 static int
-tsk_treeseq_check_windows(tsk_treeseq_t *self, size_t num_windows, double *windows)
+tsk_treeseq_check_windows(
+    const tsk_treeseq_t *self, size_t num_windows, const double *windows)
 {
     int ret = TSK_ERR_BAD_WINDOWS;
     size_t j;
@@ -1053,9 +1055,10 @@ update_running_sum(tsk_id_t u, double sign, const double *restrict branch_length
 }
 
 static int
-tsk_treeseq_branch_general_stat(tsk_treeseq_t *self, size_t state_dim,
-    double *sample_weights, size_t result_dim, general_stat_func_t *f, void *f_params,
-    size_t num_windows, double *windows, double *result, tsk_flags_t TSK_UNUSED(options))
+tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, size_t state_dim,
+    const double *sample_weights, size_t result_dim, general_stat_func_t *f,
+    void *f_params, size_t num_windows, const double *windows, double *result,
+    tsk_flags_t TSK_UNUSED(options))
 {
     int ret = 0;
     tsk_id_t u, v;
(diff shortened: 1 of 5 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/939/)

jeromekelleher:
Thanks @castedo - would you mind marking these as "draft" PRs until they're ready for review/merging? It's tricky for me to keep track of what's what and know what GitHub updates I should pay attention to.

castedo:
Roughly speaking, this PR just constifies all the stats functions in trees.h/c.

Because this PR still does not change the types in trees.h, some stats functions can not yet be constified (e.g. tsk_treeseq_kc_distance). That happens in the next and final PR #940 .

Two functions have const params that are fun, e.g.
```
int tsk_treeseq_mean_descendants(..., tsk_id_t *const *reference_sets, ... )
```
This is a fun case where "Mediterranean/romance" noun-adjective order is the only way to express what stays constant. It might be intuitive reading it "in Spanish", e.g.  "tsk_id_t puntero constante puntero" where the const is an adjective to the noun to the left, not to the right. 

Heads up there's a non-stats tweak:
```
int tsk_treeseq_init(tsk_treeseq_t *self, const tsk_table_collection_t *tables, ... )
```
because the tables are copied.

c/tskit/trees.c
```diff
@@ -336,7 +336,7 @@ tsk_treeseq_init_nodes(tsk_treeseq_t *self)
  */
 int TSK_WARN_UNUSED
 tsk_treeseq_init(
-    tsk_treeseq_t *self, tsk_table_collection_t *tables, tsk_flags_t options)
+    tsk_treeseq_t *self, const tsk_table_collection_t *tables, tsk_flags_t options)
```
jeromekelleher, line 339: This one is slightly tricky. Currently we're taking a copy of the input tables so it's fine if it's const. But, we plan on making it possible to avoid this copy to make things a bit more efficient. Ultimately, we want to have zero-copy semantics over a table collection all the way through to the Python API (see #53), but this will be a fair bit of work to do with absolute safety.

So, the question is, *if* we're passing in a "frozen" table collection that we just keep a reference to, do we need to update it? I don't think we do, except if it was not indexed. But, it would almost certainly be better to raise an error in this case rather than building the indexes.

I think I've convinced myself that const is good here; @benjeffery, what do you think?

c/tskit/trees.c
```diff
@@ -1053,9 +1055,10 @@ update_running_sum(tsk_id_t u, double sign, const double *restrict branch_length
 }
 
 static int
-tsk_treeseq_branch_general_stat(tsk_treeseq_t *self, size_t state_dim,
-    double *sample_weights, size_t result_dim, general_stat_func_t *f, void *f_params,
-    size_t num_windows, double *windows, double *result, tsk_flags_t TSK_UNUSED(options))
+tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, size_t state_dim,
+    const double *sample_weights, size_t result_dim, general_stat_func_t *f,
+    void *f_params, size_t num_windows, const double *windows, double *result,
```
jeromekelleher, line 1060: I think ``f_params`` can be marked const too? The intention would certainly be for these to be read-only values that are passed through to the summary function.

c/tskit/trees.c
```diff
@@ -641,9 +641,10 @@ increment_nd_array_value(double *array, tsk_size_t n, const tsk_size_t *shape,
 /* TODO flatten the reference sets input here and follow the same pattern used
  * in diversity, divergence, etc. */
 int TSK_WARN_UNUSED
-tsk_treeseq_genealogical_nearest_neighbours(tsk_treeseq_t *self, tsk_id_t *focal,
-    size_t num_focal, tsk_id_t **reference_sets, size_t *reference_set_size,
-    size_t num_reference_sets, tsk_flags_t TSK_UNUSED(options), double *ret_array)
+tsk_treeseq_genealogical_nearest_neighbours(const tsk_treeseq_t *self,
+    const tsk_id_t *focal, size_t num_focal, tsk_id_t *const *reference_sets,
```
jeromekelleher, line 645: I see what you mean about the trickiness of this one @castedo. In this case, I think it's probably simplest to skip the const-ing of the ``reference_sets`` here and in ``mean_descendents``, because we want to change it anyway to use a flattened input of ``sample_sets`` like the ``diversity``, divergence`` etc. I've opened #948

c/tskit/trees.h
```diff
@@ -277,79 +277,101 @@ int tsk_treeseq_simplify(const tsk_treeseq_t *self, const tsk_id_t *samples,
...
+    const tsk_id_t *focal, size_t num_focal, tsk_id_t *const *reference_sets,
+    const size_t *reference_set_size, size_t num_reference_sets, tsk_flags_t options,
     double *ret_array);
+int tsk_treeseq_mean_descendants(const tsk_treeseq_t *self,
+    tsk_id_t *const *reference_sets, const size_t *reference_set_size,
+    size_t num_reference_sets, tsk_flags_t options, double *ret_array);
 
 /* TODO change all these size_t's to tsk_size_t */
 
 typedef int general_stat_func_t(size_t K, double *X, size_t M, double *Y, void *params);
```
jeromekelleher, line 290: Should this have ``const double *X`` and ``const void *params``?

jeromekelleher (commented):
This is excellent @castedo, a major improvement in documenting this rather tricky API. Some minor comments and questions.

c/tskit/trees.c
castedo: :+1: on #948, I bet that will be less confusing for most readers.

Until then, I see two sub-optimal options
1) use `*const` which might confuse some readers who are not familiar with `*const` or
2) don't use `*const` which will confuse readers who are familiar with it because the API is declaring that the function WILL change that parameter (all the other read-only params use const, it would stand out that this one isn't marked read-only)

I'm inclined to do 2) following the principle of using the "vocabulary/words" precisely and make those unfamiliar with that "vocabulary/word" look them up.

jeromekelleher: (2) sounds good. Hopefully these functions will be updated to use the correct API soon anyway.

c/tskit/trees.c
castedo: `f_params` can be const if `params` in `general_stat_func_t` is const:
https://github.com/tskit-dev/tskit/blob/5ac0bb430c59d38c32158c4dc798ca544ee20f3f/c/tskit/trees.h#L292

So I think the question comes down to which should be easier:
1) let callers pass immutable f_param or
2) let implementors of a general_stat_func_t function modify the general purpose pass-through param

For 2) I could imagine maybe an implementor wants to keep some parallel running statistic or something. This f_param/param is untyped and it's left to the implementor to do with it whatever they like. Unfortunately I have little sense whether that "whever they like" would always be read-it-only vs sometimes store-something.

castedo: I'm going to update and play around with those general_stat_func_t params. They could use better names and I'll see if it's easy and sensible to const some of them.

jeromekelleher: Yeah, that's not obvious I guess. The intention was that we're just passing in read-only values and I can't see a good reason why someone would want to update stuff within the  summary function. But then, maybe that's being overly prescriptive.

Is there a distinction between saying *we* don't update this input pointer vs *you* can't?

castedo: OK, I think I found a good reason to leave param not const:
https://github.com/tskit-dev/tskit/blob/d8a76acbf4529ae897fa731ddeee282cb32c6692/c/tests/test_stats.c#L306
It's test code, but still, it's an example of a summary function wanting to change values that f_param points to.

> Is there a distinction between saying _we_ don't update this input pointer vs _you_ can't?

Not with normal `const` usage, at least that I'm aware of. It's a distinction that makes sense, but I don't think the C language/compiler has something for this. The code could perform some const casting gymnastics and this distinction could be documented. But that doesn't seem worth it.

I've renamed the params to `general_stat_func_t` to readable names I found in `trees.c` and I'm making the `X`, to be renamed `state` parameter to be const, the only one. I'm thinking `f_param` should be left non-const. Custom code is simpler to read and it's more flexible. It's a little wonky. But in a way if something is `void *` it's kind of like saying the param is not typed and those all the fancy powerful C compiler typing is being skipped, including const.

c/tskit/trees.h
castedo: For reason I started replying about this in the other comment. The commit I just forced push now has this updated (and renamed the params to be more readable). But see the other comment for why I only consted one param and not params.

c/tskit/trees.c
benjeffery: Const all the way and deal with any issues that arise in my view. Besides, `tsk_treeseq_t` as a read-only view of `tsk_table_collection_t` is a key architectural decision of the library, we should enforce it.

castedo:
LGTM for merging.

jeromekelleher (approved):
Let's merge!

PR #976, opened 2020-11-06
Add documentation for genetic_relatedness

### LINKED ISSUE #971: Add documentation for ts.genetic_relatedness()

brieuclehmann:
I've added a docstring to the python call and a few lines to the stats documentation. @petrelharp please let me know if I've got the interpretations correct and if there's anything else I should add!

Fixes #971

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -43,6 +43,7 @@ Single site statistics
 - :meth:`TreeSequence.divergence`
 - :meth:`TreeSequence.segregating_sites`
 - :meth:`TreeSequence.allele_frequency_spectrum`
+- :meth:`TreeSequence.genetic_relatedness`
 
 ------------------------
 Patterson's f statistics
@@ -568,6 +569,12 @@ and boolean expressions (e.g., :math:`(x > 0)`) are interpreted as 0/1.
 
    For an unpolarized statistic with biallelic loci, this calculates
    :math:`p_1 (1-p_2) + (1 - p_1) p_2`.
+
+``genetic_relatedness``
+   :math:`f(x_i, x_j) = \frac{1}{2}(x_i - m)(x_j - m)`,
+
+   where :math:`m = \frac{1}{n}\sum_{k=1}^n x_k` with :math:`n` the total number
+   of samples.
 
 ``Y2``
    :math:`f(x_1, x_2) = \frac{x_1 (n_2 - x_2) (n_2 - x_2 - 1)}{n_1 n_2 (n_2 - 1)}`
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5509,6 +5509,48 @@ class TreeSequence:
         :ref:`polarised <sec_stats_polarisation>`,
         and :ref:`return value <sec_stats_output_format>`.
 
+        What is computed depends on ``mode``:
+
+        "site"
+            Number of pairwise allelic matches in the window between two
+            sample sets relative to the rest of the sample sets. To be precise,
+            let `m(u,v)` denote the total number of alleles shared between
+            nodes `u` and `v`, and let `m(I,J)` be the sum of `m(u,v)` over all
+            nodes `u` in sample set `I` and `v` in sample set `J`. Let `S` and
+            `T` be independently chosen sample sets. Then, for sample sets `I`
+            and `J`, this computes `E[m(I,J) - m(I,S) - m(J,T) + m(S,T)]`.
+            This can also be seen as the covariance of a quantitative trait
+            determined by additive contributions from the genomes in each
+            sample set. Let each allele be associated with an effect drawn from
+            a `N(0,1/2)` distribution, and let the trait value of a sample set
+            be the sum of its allele effects. Then, this computes the covariance
+            between the trait values of two sample sets. For example, to
+            compute covariance between the traits of diploid individuals, each
+            sample set would be the pair of genomes of each individual; if
+            ``proportion=True``, this then corresponds to :math:`K_{c0}` in
+            `Speed & Balding (2014) <https://www.nature.com/articles/nrg3821>`_.
+
+        "branch"
+            Total area of branches in the window ancestral to pairs of samples
+            in two sample sets relative to the rest of the sample sets. To be
+            precise, let `B(u,v)` denote the total area of all branches
+            ancestral to nodes `u` and `v`, and let `B(I,J)` be the sum of
+            `B(u,v)` over all nodes `u` in sample set `I` and `v` in sample set
+            `J`. Let `S` and `T` be two independently chosen sample sets. Then
+            for sample sets `I` and `J`, this computes
+            `E[B(I,J) - B(I,S) - B(J,T) + B(S,T)]`.
+
+        "node"
+            For each node, the proportion of the window over which pairs of
+            samples in two sample sets are descendants, relative to the rest of
+            the sample sets. To be precise, for each node `n`, let `N(u,v)`
+            denote the proportion of the window over which samples `u` and `v`
+            are descendants of `n`, and let and let `N(I,J)` be the sum of
+            `N(u,v)` over all nodes `u` in sample set `I` and `v` in sample set
+            `J`. Let `S` and `T` be two independently chosen sample sets. Then
+            for sample sets `I` and `J`, this computes
+            `E[N(I,J) - N(I,S) - N(J,T) + N(S,T)]`.
+
         :param list sample_sets: A list of lists of Node IDs, specifying the
             groups of nodes to compute the statistic with.
         :param list indexes: A list of 2-tuples, or None.
@@ -5518,8 +5560,12 @@ class TreeSequence:
             (defaults to "site").
         :param bool span_normalise: Whether to divide the result by the span of the
             window (defaults to True).
-        :param bool proportion: Whether to divide the result by the number of
-            segregating sites (defaults to True).
+        :param bool proportion: Whether to divide the result by
+            :meth:`.segregating_sites`, called with the same ``windows`` and
+            ``mode`` (defaults to True). Note that this counts sites
+            that are segregating between *any* of the samples of *any* of the
+            sample sets (rather than segregating between all of the samples of
+            the tree sequence).
         :return: A ndarray with shape equal to (num windows, num statistics).
         """
         if proportion:


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/976/)

jeromekelleher (commented):
LGTM

docs/stats.rst
```diff
@@ -569,6 +570,11 @@ and boolean expressions (e.g., :math:`(x > 0)`) are interpreted as 0/1.
    For an unpolarized statistic with biallelic loci, this calculates
    :math:`p_1 (1-p_2) + (1 - p_1) p_2`.
 
+``genetic_relatedness``
+   :math:`f(x_i, x_j) = \frac{1}{2}(x_i - m)(x_j - m)`,
+
+   where :math:`m = \frac{1}{n}\sum_{k=1}^n x_k` with :math:`n` the total number
+   of samples.
```
petrelharp, line 577: I think you'll need a blank line before `Y2`?

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
         :ref:`polarised <sec_stats_polarisation>`,
         and :ref:`return value <sec_stats_output_format>`.
 
+        What is computed depends on ``mode``:
+
+        "site"
+            Number of pairwise allelic matches between two samples relative to
```
petrelharp, line 5528: Number of pairwise allelic matches *at segregating sites*

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
...
+            the rest of the samples, per unit of chromosome length. To be
+            precise, let `m(a,b)` denote the total number of pairwise allelic
+            matches between samples `a` and `b`, and let `C` and `D` be
+            independently chosen individuals from the sample. Then, this
+            computes `E[m(a,b) - m(a,C) - m(b,D) + m(C,D)]`. If
+            ``proportion=True``, this corresponds to :math:`K_{c0}` in
+            `Doug & Balding (2014) <https://www.nature.com/articles/nrg3821>`_.
```
petrelharp, line 5535: It's Speed & Balding.

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
...
+
+        "branch"
+            Total area of branches ancestral to two samples relative to
+            the rest of the samples. To be precise, let `S(a,b)` denote the
+            shared branch area between samples `a` and `b`, and let `C` and `D`
```
petrelharp, line 5540: maybe "let `S(a,b)` denote the total area of all branches ancestral to both `a` and `b`, ..." since "shared" is ambiguous.

python/tskit/trees.py
petrelharp: All three of these should reference the window, I think - e.g. "pairwise allelic matches in the window"

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
...
+            be independently chosen individuals from the sample. Then, this
+            computes `E[S(a,b) - S(a,C) - S(b,D) + S(C,D)]`.
+
+        "node"
+            Total number of nodes ancestral to two samples relative to
+            the rest of the samples. To be precise, let `N(a,b)` denote the
+            number of nodes ancestral to samples `a` and `b`, and let `C` and
+            `D` be independently chosen individuals from the sample. Then, this
+            computes `E[N(a,b) - N(a,C) - N(b,D) + N(C,D)]`.
```
petrelharp, line 5549: A node statistic returns, for each node, the sum of the summary function applied to the split induced by the branch above that node, divided by window length. Here, the summary function gives the number of pairs that both inherit from the node, so this is something like "for each node n, let N(a,b) denote the proportion of the window over which a and b are both descendants from n..."

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
         :ref:`polarised <sec_stats_polarisation>`,
         and :ref:`return value <sec_stats_output_format>`.
 
+        What is computed depends on ``mode``:
```
petrelharp, line 5525: Below it's written sort of as if each sample set has only one chromosome (e.g., "between samples `a` and `b`") but also acknowledging that this is not true (e.g., "number of *pairwise* allelic matches"). (it's tricky to get this right!)

python/tskit/trees.py
```diff
@@ -5522,6 +5522,32 @@ def genetic_relatedness(
...
+
+        "site"
+            Number of pairwise allelic matches between two samples relative to
+            the rest of the samples, per unit of chromosome length. To be
+            precise, let `m(a,b)` denote the total number of pairwise allelic
+            matches between samples `a` and `b`, and let `C` and `D` be
+            independently chosen individuals from the sample. Then, this
```
petrelharp, line 5532: oh, whoops - this is talking about "individuals", but these are sample sets.

petrelharp:
Good start, thanks @brieuclehmann! But it's not quite correct yet - tricky to get precise! I've made comments, want to have another go at it?

brieuclehmann:
I also need to add documentation for the `proportion` argument, clarifying that segregating sites is computed across all samples in the sample sets.

python/tskit/trees.py
brieuclehmann: Ah right, I hadn't thought about the case where each sample set can have more than one sample...
If the x[j] can take values greater than 1, we don't have the identity 2(a−c)(b−d) = [a=b]−[b=c]−[a=d]+[c=d] (used in the overleaf document). So, do we necessarily have the same interpretation?

brieuclehmann: ah, I get it now! The number of pairwise matches is saying that we sum over the combinations of samples in different sample sets.

python/tskit/trees.py
brieuclehmann: OK wow I was completely off before! I think I get it now... Strictly speaking, do we want N(a,b) = \sum_(u,v) n(u,v) where n(u,v) is the proportion of the window over which a and b are both descendants from n? Where the sum is over all pairs of nodes (u,v) where u \in a and v \in b.
And similarly for the branch stat? (since branches ancestral to sample sets is ambiguous)

python/tskit/trees.py
petrelharp: Yeah - it's kinda wierd, but it's apparently the right thing for the covariance matrix. (hm: given this is the more interesting use of it, we should probably add the "covariance matrix of a quantitative trait" interpretation to the documentation?)

brieuclehmann: Yep, I agree it would be worth emphasising this. And maybe highlighting that to get the covariance between samples u and v, you need to use sample_sets = [[u], [v]]?
Would this be different to the interpretation of K_c0 in Speed & Balding?

petrelharp: Right, something like this. 

And, right - K_c0 is only written for diploids.

jeromekelleher:
How's this coming along? I lost the thread a while back - can you approve when it's ready to go please @petrelharp?

petrelharp:
lmk when you're ready for me to have another look, @brieuclehmann?

brieuclehmann:
Yep, ready now, apologies for not flagging it up!

python/tskit/trees.py
```diff
@@ -5522,6 +5522,44 @@ def genetic_relatedness(
...
+
+        "site"
+            Number of pairwise allelic matches in the window between two
+            sample sets relative to the rest of the sample sets. To be precise,
+            let `m(a,b)` denote the total number of pairwise allelic matches
+            between sample sets `a` and `b` and let `C` and `D` be
+            independently chosen sample sets. Then, this computes
+            `E[m(a,b) - m(a,C) - m(b,D) + m(C,D)]`.
+            This can also be seen as the covariance of a quantitative trait
+            corresponding to two sample sets (or "individuals"). Let each
```
petrelharp, line 5535: how about "the covariance of a quantitative trait determined by additive contributions from the genomes in each sample set."

python/tskit/trees.py
```diff
@@ -5522,6 +5522,44 @@ def genetic_relatedness(
...
+            allele be associated with an effect drawn from a `N(0,1/2)`
+            distribution, and let the trait value of a sample set be the sum of
+            its allele effects. Then, this computes covariance between the
+            trait values of two sample sets. For diploid sample sets, when
```
petrelharp, line 5539: And then "For example, to compute covariance between traits of diploid individuals, each sample set would be the pair of genomes of one individuals (and in this case, if ``proportion=True``, this corresponds to ...)."

python/tskit/trees.py
petrelharp: (since the "or individuals") might lead people to think you can pass in tskit-individuals here.

brieuclehmann:
Thanks @petrelharp I've added your suggestions!

brieuclehmann:
I can't figure out why these tests are failing, any advice @jeromekelleher ?

jeromekelleher:
Yes, this is weird. @benjeffery, would you mind taking a look at this please?

benjeffery:
Looks like we have a test that doesn't like not being on the main thread? I'll re-run for now, if this reoccurs I'll look into it.

python/tskit/trees.py
```diff
@@ -5500,6 +5500,48 @@ def genetic_relatedness(
...
+
+        "branch"
+            Total area of branches in the window ancestral to pairs of samples
+            in two sample sets relative to the rest of the sample sets. To be
+            precise, let `B(u,v)` denote the total area of all branches
+            ancestral to nodes `u` and `v`, and let `B(I,J)` be the sum of
+            `B(u,v)` over all nodes `u` in sample set `I` and `v` in sample set
+            `J`. Let `S` and `T` be two independently chosen sample sets. Then
+            for sample sets `I` and `J`, this computes
+            `E[B(u,v) - B(u,S) - B(v,T) + B(S,T)]`.
```
petrelharp, line 5532: whoops, this should be 
```
             `E[B(I,J) - B(I,S) - B(J,T) + B(S,T)]`.
```

python/tskit/trees.py
```diff
@@ -5500,6 +5500,48 @@ def genetic_relatedness(
...
+        "node"
+            For each node, the proportion of the window over which pairs of
+            samples in two sample sets are descendants, relative to the rest of
+            the sample sets. To be precise, for each node `n`, let `N(u,v)`
+            denote the proportion of the window over which samples `u` and `v`
+            are descendants of `n`, and let and let `N(I,J)` be the sum of
+            `N(u,v)` over all nodes `u` in sample set `I` and `v` in sample set
+            `J`. Let `S` and `T` be two independently chosen sample sets. Then
+            for sample sets `I` and `J`, this computes
+            `E[N(I,J) - N(I,S) - N(J,T) + N(S,T)]`.
```
petrelharp, line 5543: agree!

python/tskit/trees.py
```diff
@@ -5510,7 +5552,10 @@ def genetic_relatedness(
         :param bool span_normalise: Whether to divide the result by the span of the
             window (defaults to True).
         :param bool proportion: Whether to divide the result by the number of
-            segregating sites (defaults to True).
+            segregating sites (defaults to True). Note that this counts sites
+            that are segregating in *any* of the samples of *any* of the sample
+            sets (rather than segregating in all of the samples of the tree
+            sequence).
```
petrelharp, line 5558: This should say "segregating between" instead of "segregating in". Also, maybe it should say
```
Whether to divide the result by :meth:`.segregating_sites`, called with the same ``windows`` and ``mode`` (defaults to True).
```
... since if mode='branch' then it's not the *number* of segregating sites.

python/tskit/trees.py
brieuclehmann: ah yep, thanks for spotting!

petrelharp (approved):
Excellent!

benjeffery:
@mergifyio rebase

PR #977, opened 2020-11-06
Add division by segregating sites when proportion=True

### LINKED ISSUE #972: Add `proportion` argument in ts.genetic_relatedness()
Currently, genetic_relatedness() does _not_ divide by the number of segregating sites. We should add a `proportion` argument to the function to do exactly this, with the default being `proportion=True`.

brieuclehmann:
I've added the division by number of segregating sites, which wasn't there previously even though it was in the function arguments!

I've just noticed that the structure of Tajima's D and F_ST are quite different - is this what I should be aiming for?

Fixes #972


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/977/)

python/tskit/trees.py
```diff
@@ -5535,15 +5535,30 @@ def genetic_relatedness(
...
+            denominator = self.segregating_sites(
+                sample_sets=all_samples,
+                windows=windows,
+                mode=mode,
+                span_normalise=span_normalise,
+            )
+        else:
+            denominator = 1
+
+        return (
```
jeromekelleher, line 5550: Would be easier to read here if we assigned the numerator to a variable:
```python3
numerator = self.__ ...
return numerator / denominator
``

File: python/tskit/trees.py [hunk 2]
```diff
@@ -5535,15 +5535,30 @@ def genetic_relatedness(
...
-            2,
-            sample_sets,
-            indexes=indexes,
-            windows=windows,
-            mode=mode,
-            span_normalise=span_normalise,
-            polarised=polarised,
+        if proportion:
+            # TODO this should be done in C also
+            all_samples = list({u for s in sample_sets for u in s})
```
jeromekelleher, line 5540: Why do we need to do this ``all_samples`` step? I would have thought we could pass the ``sample_sets`` through without changes.

jeromekelleher (approved):
minor comments

python/tskit/trees.py
brieuclehmann: So, if `sample_sets=[[u] for u in ts.samples()]` then `ts.segregating_sites(sample_sets)` is applied to each sample independently (so returns a list of zeros of length num_samples).

Now... I can't remember why I was writing `sample_sets` in this way but must have copy-pasted it from somewhere else!

petrelharp: I agree: we want the denominator to be the number of sites that are segregating in *any* of the samples of any of the sample sets. (need to make sure to document that it is this, not segregating in all samples of the trees sequence)

brieuclehmann: Just to clarify - this means I should leave this bit as is?

jeromekelleher: Yes, that's my interpretation of @petrelharp's comment @brieuclehmann. Looks like we should update the documentation to clarify the point, though

brieuclehmann: Yep, I'll update that on the docs PR.

jeromekelleher (approved):
LGTM!

PR #981, opened 2020-11-09
Complete checks for uninitialised objects in C Python module

### LINKED ISSUE #933: Add tests for uninitialised Python objects
The Python C API can current segfault in quite a few places if a user deliberately creates an instance without calling ``__init__`` using ``__new__``. While this is a remote corner case, we should tie up this loose-end and avoid the segault. The way to do it is to make sure ``ClassName_check_state`` is called before accessing any of the memory of the embedded C object.

The test for doing this on the various objects can be adopted from [test_uninitialised](https://github.com/tskit-dev/msprime/blob/main/tests/test_lowlevel.py#L930) in msprime.

benjeffery:
## Description

It was possible to seg fault the C Python module. This PR adds checks to all methods.

Fixes #933 

# PR Checklist:

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -820,11 +820,11 @@ IndividualTable_add_row(IndividualTable *self, PyObject *args, PyObject *kwds)
     npy_intp *shape;
     static char *kwlist[] = { "flags", "location", "metadata", NULL };
 
-    if (!PyArg_ParseTupleAndKeywords(
-            args, kwds, "|iOO", kwlist, &flags, &py_location, &py_metadata)) {
+    if (IndividualTable_check_state(self) != 0) {
         goto out;
     }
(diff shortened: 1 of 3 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/981/)

python/_tskitmodule.c
```diff
@@ -5642,7 +5765,7 @@ TreeSequence_get_node(TreeSequence *self, PyObject *args)
     Py_ssize_t record_index, num_records;
     tsk_node_t record;
 
-    if (TreeSequence_check_tree_sequence(self) != 0) {
```
jeromekelleher, line 5645: Yes, I was a bit worried about inconsistencies - thanks for sorting that out.

jeromekelleher (approved):
Looks great, thanks!

PR #984, opened 2020-11-10
Add note on pytest output

### LINKED ISSUE #932: Document how to get pytest to print
We have ``-n4`` set as a default which is great for making the tests run quicker, but it does mean that ``pytest -vs`` does actually show prints any more.

It looks like you have to override this and set ``-n 0 -vs `` so that ``print()`` statements within your tests (which are vital for debugging) actually do something. I think we need to make this clear for new users, as test development could be very frustrating otherwise.

benjeffery:
Fixes #932

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -533,6 +533,13 @@ of the failing lines:
 
     $ python3 -m pytest --tb=line
 
+If you need to see the output of tests (e.g. ``print`` statements) then you need to use
+these flags to run a single thread and capture output:
+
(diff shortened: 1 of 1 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/984/)

jeromekelleher (approved):

benjeffery:
@Mergifyio refresh

PR #986, opened 2020-11-11
Add TableCollection dump/load

### LINKED ISSUE #14: Provide TableCollection.dump( ) method

benjeffery:
## Description

Adds `dump` and `load` to `TableCollection`.
Fixes #14 

# PR Checklist:

### DIFF
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -27,18 +27,14 @@ import collections
 import inspect
 import itertools
 import os
-import platform
 import random
 import tempfile
-import unittest
 
 import msprime
 import numpy as np
 import pytest
 
 import _tskit
-
-IS_WINDOWS = platform.system() == "Windows"
 
 
 def get_tracked_sample_counts(st, tracked_samples):
@@ -70,7 +66,7 @@ def get_sample_counts(tree_sequence, st):
     return nu
 
 
-class LowLevelTestCase(unittest.TestCase):
+class LowLevelTestCase:
     """
     Superclass of tests for the low-level interface.
     """
@@ -165,6 +161,32 @@ class TestTableCollection(LowLevelTestCase):
     """
     Tests for the low-level TableCollection class
     """
+
+    def test_file_errors(self):
+        tc1 = _tskit.TableCollection(1)
+        self.get_example_tree_sequence().dump_tables(tc1)
+
+        def loader(*args):
+            tc = _tskit.TableCollection(1)
+            tc.load(*args)
+
+        for func in [tc1.dump, loader]:
+            with pytest.raises(TypeError):
+                func()
+            for bad_type in [None, [], {}]:
+                with pytest.raises(TypeError):
+                    func(bad_type)
+
+    def test_dump_equality(self, tmp_path):
+        for ts in self.get_example_tree_sequences():
+            tc = _tskit.TableCollection(sequence_length=ts.get_sequence_length())
+            ts.dump_tables(tc)
+            with open(tmp_path / "tmp.trees", "wb") as f:
+                tc.dump(f)
+            with open(tmp_path / "tmp.trees", "rb") as f:
+                tc2 = _tskit.TableCollection()
+                tc2.load(f)
+            assert tc.equals(tc2)
 
     def test_reference_deletion(self):
         ts = msprime.simulate(10, mutation_rate=1, random_seed=1)
@@ -518,7 +540,6 @@ class TestTreeSequence(LowLevelTestCase, MetadataTestMixin):
     def tearDown(self):
         os.unlink(self.temp_file)
 
-    @pytest.mark.skipif(IS_WINDOWS, reason="File permissions on Windows")
     def test_file_errors(self):
         ts1 = self.get_example_tree_sequence()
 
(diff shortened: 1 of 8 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/986/)

benjeffery:
Codecov isn't happy due to lines in _tskitmodule.c which it says are not covered. When I set a breakpoint on those lines it gets hit though! Not sure what is happening there.

python/tests/test_lowlevel.py
```diff
@@ -166,6 +166,60 @@ class TestTableCollection(LowLevelTestCase):
     Tests for the low-level TableCollection class
     """
 
+    def setUp(self):
```
jeromekelleher, line 169: Hardly seems worth making a temp file for *every* test in this class, when we're only using  it in the verify_dump_equality function.

A useful pattern that works across the different oses is:

```python
with tempfile.TemporaryDirectory() as tempdir:
     tmp_file = pathlib.Path(tempdir.name) / "tmp.trees"
     # do your tests using tmp_file
```

python/tests/test_lowlevel.py
```diff
@@ -166,6 +166,60 @@ class TestTableCollection(LowLevelTestCase):
...
+        fd, self.temp_file = tempfile.mkstemp(prefix="msp_ll_tc_")
+        os.close(fd)
+
+    def tearDown(self):
+        os.unlink(self.temp_file)
+
+    @pytest.mark.skipif(IS_WINDOWS, reason="File permissions on Windows")
```
jeromekelleher, line 176: I don't see why this needs skipping on windows - we're not touching the file system are we?

python/tskit/tables.py
```diff
@@ -2258,6 +2258,32 @@ def __eq__(self, other):
...
+    @classmethod
+    def load(cls, file_or_path):
+        file, local_file = util.convert_file_like_to_open_file(file_or_path, "rb")
+        ll_tc = _tskit.TableCollection(1)
+        ll_tc.load(file)
+        tc = TableCollection(1)
+        tc._ll_tables = ll_tc
+        return tc
+
+    def dump(self, file_or_path, skip_checks=False):
```
jeromekelleher, line 2270: I'm not sure we want the ``skip_checks`` argument here as it might have some unwanted effects, and might be expensive. I think we should just write out the state as it is.

If we do have the checks we should probably implement at the C level using the check integrity function, rather than just calling ``self.tree_sequence()``.

jeromekelleher (commented):
LGTM, but a few comments above.

python/tskit/util.py
```diff
@@ -435,3 +436,33 @@ def tree_sequence_html(ts):
...
+        # must make sure that close is **not** called on the fd.
+        try:
+            _file = open(file_like, mode, closefd=False, buffering=0)
+        except TypeError:
+            pass
+    if _file is None:
+        # Assume that this is a file **but** we haven't opened it, so we must
+        # not close it.
+        if mode == "wb" and not hasattr(file_like, "write"):
+            raise TypeError("file object must have a write method")
```
jeromekelleher, line 465: Not hitting this with tests.

jeromekelleher:
Codecov has definitely gone a bit mad on the TableCollection_alloc coverage - that function must be being called. So, we can ignore codecov here I think.

python/tests/test_lowlevel.py
benjeffery: Copied from `test_highlevel` without thinking - fixed both to use pytests built-in tempdir in https://github.com/tskit-dev/tskit/pull/986/commits/82dcbe524773ea44f71c8dfd5dd4f5592891d94e

python/tests/test_lowlevel.py
benjeffery: Again copied without thinking, fixed both in https://github.com/tskit-dev/tskit/pull/986/commits/acda06f0ee1c5a02b13594991b768a0069d6323b

python/tskit/tables.py
benjeffery: Removed in https://github.com/tskit-dev/tskit/pull/986/commits/82edb1fa6a1c944a86ce731d14fecf13df0103d3

python/tskit/util.py
benjeffery: Extra test in https://github.com/tskit-dev/tskit/pull/986/commits/8cdebcf9abd4a4a854a68802ef9d98f9df8bd394

benjeffery:
@jeromekelleher Fixed up, still needs a squash as left several commits for easy change viewing.

python/tests/test_lowlevel.py
```diff
@@ -619,38 +599,35 @@ def test_num_nodes(self):
...
             tables.compute_mutation_times()
             ts = _tskit.TreeSequence()
             ts.load_tables(tables)
-            self.verify_dump_equality(ts)
+            with open(tmp_path / "temp.trees", "wb") as f:
+                ts.dump(f)
+            with open(tmp_path / "temp.trees", "rb") as f:
+                ts2 = _tskit.TreeSequence()
+                ts2.load(f)
+            assert ts.get_num_samples() == ts2.get_num_samples()
```
jeromekelleher, line 614: I did all this stuff long-hand before because we didn't have the table collection equality operator - we could just replace with the standard ``ts.tables == ts2.tables`` now.

jeromekelleher (approved):
Looks great, thanks @benjeffery! There's one potential cleanup, but merge away whenever.

python/tests/test_lowlevel.py
benjeffery: As this is the low-level ts I had to go with:
```python
            tc = _tskit.TableCollection(ts.get_sequence_length())
            ts.dump_tables(tc)
            tc2 = _tskit.TableCollection(ts2.get_sequence_length())
            ts2.dump_tables(tc2)
            assert tc.equals(tc2)
```

petrelharp:
yay! this will simplify some other code!

PR #1001, opened 2020-11-16
Add options to table collection clear

### LINKED ISSUE #929: Should tsk_table_collection_clear remove metadata?
We've probably discussed this before, but should tsk_table_collection_clear also clear out the metadata as well as all the table data? What about the schemas? I guess probably not, and the current behaviour is what we want. It's not totally obvious, though.

We should at the least update the documentation to make it clear what ``clear`` actually does, though.

benjeffery:
## Description

Adds options to `tsk_table_collection_clear`. Also changes the default behaviour by not clearing the provenance table. In the original issue I thought clear should wipe everything - but I've come round to the idea that the most common clearing operation is where you want to retain the provenance and metadata schemas. We should encourage retaining these by default in a table-modifying operation.

Fixes #929 

# PR Checklist:

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -11,6 +11,11 @@
   current behaviour. (:user:`mufernando`, :user:`jeromekelleher`,
   :issue:`896`, :pr:`897`, :issue:`913`, :pr:`917`).
 
+- Changed default behaviour of ``tsk_table_collection_clear`` to not clear
+  provenances and added ``options`` argument to optionally clear provenances
+  and schemas.
+  (:user:`benjeffery`, :issue:`929`, :pr:`1001`)
+
 - Exposed ``tsk_table_collection_set_indexes`` to the API.
   (:user:`benjeffery`, :issue:`870`, :pr:`921`)
 
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -5142,7 +5142,7 @@ test_table_collection_union(void)
         &tables_copy, &tables_empty, node_mapping, TSK_UNION_NO_CHECK_SHARED);
     CU_ASSERT_FATAL(tsk_table_collection_equals(&tables, &tables_copy, 0));
     // self is empty
-    ret = tsk_table_collection_clear(&tables_copy);
+    ret = tsk_table_collection_clear(&tables_copy, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     ret = tsk_table_collection_union(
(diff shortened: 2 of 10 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/1001/)

petrelharp:
LGTM! Although - we don't seem to have discussed this in #929 - I would have thought the default would be to clear provenance, since what's the point of keeping provenance in an empty table? Provenance says where the information came from, but if there's no information...

benjeffery:
My thinking here was based on the existing usages of `clear` - they both copied the provenance back in after clearing. You can see in this PR I have removed that copy. 

Basically you usually clear when you're modifying an existing table, which should retain provenance and metadata.

jeromekelleher:
> I would have thought the default would be to clear provenance, since what's the point of keeping provenance in an empty table? Provenance says where the information came from, but if there's no information...

This was my initial thought also.

jeromekelleher:
> My thinking here was based on the existing usages of clear - they both copied the provenance back in after clearing. You can see in this PR I have removed that copy.

I see, that's a good point. What if we changed the flags to KEEP_PROVENANCE and KEEP_METADATA? It seems reasonable to me that the default behaviour would be to nuke everything, and then you can optionally keep some stuff in special cases where you want them?

petrelharp:
Ok, that's a good point. I guess in practice if we want a totally brand-new table collection then we tend to just re-allocate a new one. So, I agree, what you've done here actually makes more sense. OK, never mind - I vote to keep it the way you've got it!

jeromekelleher:
How does this look at the Python level @benjeffery - did we not provide a TableCollection.clear() method? It might be worth doing this now, as the "direction" the flags need to be in usually becomes clearer after we think through the semantics in Python.

benjeffery:
> How does this look at the Python level @benjeffery - did we not provide a TableCollection.clear() method? It might be worth doing this now, as the "direction" the flags need to be in usually becomes clearer after we think through the semantics in Python.

There is no such method in the Python API - I assume the test suite would benefit from it. I'll have a look for potential usage ans add it to this PR - setting back to draft till that is in.

> Peter makes a good point too - if we want to totally crispy clean table collection, we can just make a new one, so I'm leaning towards agreeing with what you have.

I also agree with this - what's the point (especially in Python) of a complete clear?

benjeffery:
Only found one spot in the Python tests to use this - but I still think it is worth having. @jeromekelleher I think this is merge-ready.

jeromekelleher (approved):
LGTM!

PR #1004, opened 2020-11-17
Flesh out examples and add docs to LWT

### LINKED ISSUE #864: Document LightweightTableCollection interface and usage
See #862 for background.

Need to add documentation for how to use the LightweightTableCollection code in downstream projects. Ideally we want to add some minimal examples using the Python-C interface and Cython to our build examples.

benjeffery:
## Description

Adds fleshed out examples to the example module and some docs to the README. Not finished yet as I assume we want doc strings on everything?

Fixes #864 

# PR Checklist:


benjeffery:
@jeromekelleher Still draft as I assume we want docstrings - but thought it worth you having a look at the C Python code I added to the examples as I likely missed a detail or more standard way of doing something!

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/1004/)

python/lwt_interface/README.md
```diff
@@ -8,4 +8,55 @@ C API in your own compiled Python module (either via Cython
...
+In C you can access the tables in a `LightweightTableCollection` instance that is passed 
+to your function:
+```C
+LightweightTableCollection *lwt_tables = NULL;
+PyArg_ParseTuple(args, "O!", &LightweightTableCollectionType, &lwt_tables))
+```
+The variable `lwt_tables->tables` then contains a standard `tsk_table_collection_t`.
+
+To send tables back to python you can create an instance:
+```C
```
jeromekelleher, line 49: This is a bit tricky, as there's a lot of other infrastructure around this you need in order to make it memory safe etc. I think I'd rather keep the example C code chunks out of this file, and we just direct the reader to look at the example module for how to use it.

python/lwt_interface/example_c_module.c
```diff
@@ -38,7 +38,53 @@
...
+    }
+
+out:
+    return ret;
+}
+
+static PyObject * example_returning(PyObject *self, PyObject *args) {
+    PyObject * ret = NULL;
+    int err = -1;
+    LightweightTableCollection *tables = (LightweightTableCollection *)PyObject_CallFunction(
```
jeromekelleher, line 62: This is nice. I wonder if it's simpler to show an example where the C code modifies the tables? We don't strictly need to return a new TableCollection since we can do this easily create a new one in Python and pass it in.

But, if you're sure the refcounting is all solid here, then this is good too.

python/lwt_interface/tskit_lwt_interface.h
```diff
@@ -1813,7 +1813,7 @@ static PyMethodDef LightweightTableCollection_methods[] = {
     { NULL } /* Sentinel */
 };
 
-static PyTypeObject LightweightTableCollectionType = {
```
jeromekelleher, line 1816: Why remove the static here?

jeromekelleher (commented):
Looks great. What would be really super-useful here would be an example of how to use the lwt interface with Cython - but maybe we don't want to get into this.

python/lwt_interface/README.md
benjeffery: Yep, agreed and removed.

python/lwt_interface/example_c_module.c
benjeffery: I think it best if recommend always creating table collections in Python - I've added a note to the docs about this and changed to example to be a modifying example.

python/lwt_interface/tskit_lwt_interface.h
benjeffery: Opps!

benjeffery:
@jeromekelleher I've made some changes and fleshed out the example to actually build a tree sequence. I've also removed the creating of a table collection on the C side as we should discourage this.

python/lwt_interface/example_c_module.c
```diff
@@ -38,7 +38,99 @@
...
+    }
+    err = tsk_tree_first(&tree);
+    if (err < 0) {
+        handle_tskit_error(err);
+        goto out;
+    }
+    
+    /* Return true if the tree has more than one root */
+    if (tsk_tree_get_num_roots(&tree) > 1) {
+        ret = Py_True;
```
jeromekelleher, line 82: I think you need to Py_INCREF(Py_True) here. Safer to do something like
``ret = Py_BuildValue("i", tsk_tree_get_num_roots(&tree) > 1)``

(Is there a constructor code for bool?)

jeromekelleher (approved):
LGTM, some minor nits

python/lwt_interface/example_c_module.c
benjeffery: I went with `Py_BuildValue("O", condition ? Py_True: Py_False);` as there is no code for booleans. I've also fixed the bad `Py_None` just below. 
For some reason I had it in my head that these didn't need refcounting - but it is the exact opposite.

benjeffery:
@jeromekelleher fixed up.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1004/)

benjeffery:
>  Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1004/)

Ah-ha the bot thinks there are doc changes as it is comparing to `main` which has commits that this is PR not rebased on to. Hmm. Not worth worrying about, but would be possible for it to compare to the base commit of the PR.

benjeffery:
@terhorst We've done some work on the Python C extension interface examples and docs here - would you mind seeing if it makes sense? We'll be doing a cython example too eventually.

terhorst:
> @terhorst We've done some work on the Python C extension interface examples and docs here - would you mind seeing if it makes sense? We'll be doing a cython example too eventually.

Hi @benjeffery, sorry, I mentally filed this thread away to respond ... and then promptly threw the filing cabinet out the window ;) In short, I think this is great! Super helpful for what I wanted to accomplish. The only missing piece of the puzzle for me was Cython integration. Cython automatically generates `PyMODINIT_FUNC` and, as far as I can tell, there is no (supported) way to inject a call to `register_lwt_class(module)` in there. (I say supported because I was able to get it to work by `extern`ing the `module` variable and dropping the call directly into the top-level module code, but this felt very hacky and liable to break at some point in the future.) Also there are some issues with void pointers if you try to use `tskit_lwt_interface.h` in C++ code, but it's not a showstopper.

What I ended up doing was creating an extremely minimal C module, say `_lwtc.c`, that essentially just calls the register function. Then in Cython I wrapped it with an extension type declaration, i.e.
```cdef extern:
    ctypedef class xsmc._lwtc.LightweightTableCollection [object LightweightTableCollection]:
        cdef tsk_table_collection_t *tables
```
The key is being able to access the table collection pointer from Cython code. Once that's done, you can call all the rest of the `tsk_` API and you're off to the races.

jeromekelleher:
Awesome, thanks for the update @terhorst! Do you have a link to the code please? We'd like to make a minimal example so that others can benefit from your work here!

terhorst:
Sure, a minimal example is to define the LightweightTreeSequence module via

https://github.com/terhorst/xsmc/blob/arg/src/_lwtc.c

(essentially your example code with all the nonessentials stripped out), and then call it as so:

https://github.com/terhorst/xsmc/blob/arg/xsmc/_viterbi.pyx#L154-L157

Note that you'll need to have defined all the various tskit headers; that happens here:

https://github.com/terhorst/xsmc/blob/arg/xsmc/_tskit.pxd.

benjeffery:
@terhorst Great stuff, is there any chance you have the time to polish the code you've linked to into an example in the tskit codebase? No worries if not!

terhorst:
Here is a Cythonized version of one of the C API examples from the docs:

https://github.com/terhorst/tskit_cython_example

jeromekelleher:
This looks great, thanks @terhorst! @benjeffery, what's best way of linking this into our docs? Would it be best to get a copy into the tskit repo so we can make sure we don't break it (like the current C example?)

jeromekelleher:
Moving the discussion to #1124 so we don't lose track of it.

PR #1013, opened 2020-11-18
Basic implementation of tree polytomy resolution.

jeromekelleher:
@hyanwong this is my proposal for providing an interface for random polytomy resolution. It's still a work in progress and needs bunch of the nice interface feature you added, but I think it basically works. In particular, I think mutations should "just work", modulo epsilon being small enough.

I'm confident this is doing the job correctly because it's based on a well documented algorithm for generating binary trees uniformly. It's also reasonably efficient, and could easily be implemented in C if we so desired.

Thoughts?

### DIFF
--- a/python/tskit/combinatorics.py
+++ b/python/tskit/combinatorics.py
@@ -36,7 +36,7 @@ import numpy as np
 import tskit
 
 
-@attr.s
+@attr.s(eq=False)
 class TreeNode:
     """
     Simple linked tree node used for the random binary tree function below.
--- a/python/tskit/combinatorics.py
+++ b/python/tskit/combinatorics.py
@@ -28,10 +28,85 @@ import collections
 import functools
 import heapq
 import itertools
+import random
 
+import attr
 import numpy as np
 
 import tskit
+
+
+@attr.s
+class TreeNode:
+    """
+    Simple linked tree node used for the random binary tree function below.
+    """
+
+    parent = attr.ib(default=None)
+    children = attr.ib(factory=list)
+    label = attr.ib(default=None)
+
+
+def random_binary_tree(leaf_labels, rng):
+    """
+    Based on the description of Remy's method of generating "decorated"
+    random binary trees in TAOCP 7.2.1.6. This is not a direct
+    implementation of Algorithm R, because we are interested in
+    the leaf node labellings.
+    """
+    nodes = [TreeNode(label=leaf_labels[0])]
+    for label in leaf_labels[1:]:
+        n = TreeNode(label=label)
+        x = rng.choice(nodes)
+        internal = TreeNode(parent=x.parent, children=[x, n])
+        if x.parent is not None:
+            index = x.parent.children.index(x)
+            x.parent.children[index] = internal
+        rng.shuffle(internal.children)
+        x.parent = internal
+        n.parent = internal
+        nodes.extend([n, internal])
+
+    # TODO move this to testing code.
+    assert len(nodes) == 2 * len(leaf_labels) - 1
+    for node in nodes:
+        # print("check", node)
+        if node.label is not None:
+            assert len(node.children) == 0
+        for child in node.children:
+            assert child.parent == node
+
+    root = nodes[0]
+    while root.parent is not None:
+        root = root.parent
+    return root
+
+
+def resolve_polytomies(tree, epsilon, seed=None):
+    tables = tree.tree_sequence.dump_tables()
+    tables.keep_intervals([tree.interval], simplify=False)
+    tables.edges.clear()
+    rng = random.Random(seed)
+
+    for u in tree.nodes():
+        if tree.num_children(u) > 2:
+            root = random_binary_tree(tree.children(u), rng)
+            root.label = u
+            time = tree.time(u) - epsilon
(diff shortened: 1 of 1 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/1013/)

hyanwong:
I get a RecursionError here:

```
resolve_polytomies(tskit.Tree.generate_star(5), epsilon=1e-13, seed=68)
```

as follows:

```
RecursionError                            Traceback (most recent call last)
<ipython-input-16-fad009b190b8> in <module>
----> 1 resolve_polytomies(tskit.Tree.generate_star(5), epsilon=1e-13, seed=68)

<ipython-input-7-7018048d8755> in resolve_polytomies(tree, epsilon, seed)
     56      for u in tree.nodes():
     57          if tree.num_children(u) > 2:
---> 58              root = random_binary_tree(tree.children(u), rng)
     59              root.label = u
     60              time = tree.time(u) - epsilon

<ipython-input-7-7018048d8755> in random_binary_tree(leaf_labels, rng)
     27          internal = TreeNode(parent=x.parent, children=[x, n])
     28          if x.parent is not None:
---> 29              index = x.parent.children.index(x)
     30              x.parent.children[index] = internal
     31          rng.shuffle(internal.children)

<attrs generated eq __main__.TreeNode> in __eq__(self, other)
      2     if other.__class__ is not self.__class__:
      3         return NotImplemented
----> 4     return  (
      5         self.parent,
      6         self.children,

... last 1 frames repeated, from the frame below ...

<attrs generated eq __main__.TreeNode> in __eq__(self, other)
      2     if other.__class__ is not self.__class__:
      3         return NotImplemented
----> 4     return  (
      5         self.parent,
      6         self.children,

RecursionError: maximum recursion depth exceeded in comparison
```

jeromekelleher:
Thanks, yes, should be fixed now.

hyanwong:
Thanks @jeromekelleher - I did some testing and both algorithms produce equiprobable trees, and have the same scaling properties (mine is currently about 10% faster, but that's negligible and will be massively outweighed by any implementation in C). So I'm happy to use your implementation of Remy's method instead.

Testing on a real inferred TS from 1000G (about 5% of chr20) shows that iterating tree-by-tree is about 30 times slower than resolving each polytomy as it changes (i.e. doing it over the tree sequence). My code takes 1 min to resolve the TS into binary trees. A tree-by-tree approach looks like it will take about half an hour. That seems enough of a different to be worth thinking about implementing a TS method. I think my previous code can use your `random_binary_tree` function, so I guess that's relatively easy?

hyanwong:
P.s. can we standardise on whether we use `seed` or `random_seed`? I think msprime uses `random_seed`, whereas stdpopsim uses `seed`, which is already pretty confusing.

jeromekelleher:
OK, thanks for trying out the code @hyanwong.

After thinking about it more, I'm convinced that resolving the polytomies tree-by-tree is the right thing to do for the tree distance metric calculations. The distance metrics are inherently tree-by-tree, and what you're trying to do by resolving polytomies is to marginalise over all the possible resolutions *of that tree*. If we have a large polytomy that spans a large section of the genome then if we only resolve that once, we're sampling much less of the space (or at least, sampling it less evenly) that we would if we repeatedly did it tree-by-tree. Anyway, who's to say that we *should* have a single resolution of a polytomy across many trees? There's every chance that there could be different binary trees within the region spanned by the polytomy, we just don't know.

So, basically, statistically I think we should do the resolution tree-by-tree, even if it is slower. I think that by the time the other processing is done (outputting the resolved trees to newick, doing the RF distance comparison in R, etc) the actual resolution time probably won't dominate, so in practise it'll not be much slower.

Anyway - I guess I'll close this PR and add a commit to #809 with the updated functions.

hyanwong:
> After thinking about it more, I'm convinced that resolving the polytomies tree-by-tree is the right thing to do for the tree distance metric calculations. The distance metrics are inherently tree-by-tree, and what you're trying to do by resolving polytomies is to marginalise over all the possible resolutions _of that tree_. If we have a large polytomy that spans a large section of the genome then if we only resolve that once, we're sampling much less of the space (or at least, sampling it less evenly) that we would if we repeatedly did it tree-by-tree.

Yes, we are sampling topologies less (which is why it is faster!). But we never claim that the trees are independent anyway (i.e. we know that there is non-independence in the metric from tree to tree) so I don't think that matters. What *is* useful is to be able to output the resulting binary TS as a tree sequence. I can't do that with your code at the moment.

> Anyway, who's to say that we _should_ have a single resolution of a polytomy across many trees? There's every chance that there could be different binary trees within the region spanned by the polytomy, we just don't know.

Yes, but that also goes for the region within a tree. I genuinely don't think that we should re-resolve a polytomy, just because there's a minor change in a subtree that's miles away from the polytomy node, not even nested within it. Here's another way to put it: if we match, say, a Cornish sample onto the UKB+TGP tree sequence, we could well create another breakpoint in recent times, low down in the UK part of the tree. That shouldn't affect how a polytomy between 3 closely related San bushmen is resolved.

> Anyway - I guess I'll close this PR and add a commit to #809 with the updated functions.

Yes, thanks.

PR #1019, opened 2020-11-19
Make tc.equals kwargs only

### LINKED ISSUE #896: table collection equivalence
Right now, equality of table collections is defined in a very strict manner (see [here](https://github.com/tskit-dev/tskit/blob/2b7bd7d64923fe720d3fd3bd99d678373e9b63c6/c/tskit/tables.c#L7735)). All tables are compared, in addition to sequence length, metadata and metadata schema.

In discussions with @gtsambos, @petrelharp, @benjeffery and @hyanwong we came to the conclusion that it would be helpful to have a less strict equality (maybe _equivalence_) comparison, in which the metadata, metadata schema and provenance tables are ignored. I think this _equivalence_ operation is crucial, because in many cases we are interested in comparing the "biological" content of tree sequences, without worrying about how they were made to be.

Importantly, since the metadata attributes were added `tskit.TableCollection.union` stopped working, because there we subset tree sequences (produced by different runs of simulations) on their shared nodes and test for equality, which will fail because of metadata.

I propose we add two new flags to `tsk_table_collection_equals` to ignore the metadata/schema and provenance tables when comparing: `TSK_IGNORE_METADATA` and `TSK_IGNORE_PROVENANCE`.

Does this sound reasonable?

benjeffery:
## Description

Makes TableCollection equals options kwargs only

Fixes #896 

# PR Checklist:

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -2502,6 +2502,9 @@ class TestTableCollection:
         assert t1.equals(t2)
         assert t2.equals(t1)
 
+        with pytest.raises(TypeError):
+            t1.equals(t2, True)
+
     def test_sequence_length(self):
         for sequence_length in [0, 1, 100.1234]:
             tables = tskit.TableCollection(sequence_length=sequence_length)
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2247,6 +2247,7 @@ class TableCollection:
     def equals(
         self,
         other,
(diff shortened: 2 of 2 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1019/)

jeromekelleher (approved):

PR #1022, opened 2020-11-20
Add ts.coiterate method

### LINKED ISSUE #1021: Add TreeSequence.coiterate(other, **kwargs) method
As discussed in #815, this is a useful operation and should be part of the public API. 

The implementation is [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/tsutil.py#L1287), which I think is probably fine. We can leave the door open to coiterating along more than two tree sequences via ``*args``, but we don't need to worry about doing it for now.

We just need a handful of tests to make sure that (a) this does what we expect on a simple example; (b) the trees are returned in the correct order (i.e., ``tree1.tree_sequence is ts1``) and so on; (c) we test some cases like one ts has one tree and the other has 3, etc; (d) test error conditions are handled correctly.

I think we should get this in for 0.3.3

@benjeffery @hyanwong, would one of you mind picking this up please?

hyanwong:
Fixes #1021

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -4758,6 +4758,95 @@ class TestBadTrees:
             tskit.load_text(nodes=nodes, edges=edges, strict=False)
 
 
+class TestCoiteration:
+    """
+    Test ability to iterate over multiple (currently 2) tree sequences simultaneously
+    """
+
+    def test_identical_ts(self):
+        ts = msprime.simulate(4, recombination_rate=1, random_seed=123)
+        assert ts.num_trees > 1
+        total_iterations = 0
+        for tree, (_, t1, t2) in zip(ts.trees(), ts.coiterate(ts)):
+            total_iterations += 1
+            assert tree == t1 == t2
+        assert ts.num_trees == total_iterations
+
+    def test_intervals(self):
+        ts1 = msprime.simulate(4, recombination_rate=1, random_seed=1)
+        assert ts1.num_trees > 1
+        one_tree_ts = msprime.simulate(5, random_seed=2)
+        multi_tree_ts = msprime.simulate(5, recombination_rate=1, random_seed=2)
+        assert multi_tree_ts.num_trees > 1
+        for ts2 in (one_tree_ts, multi_tree_ts):
+            bp1 = set(ts1.breakpoints())
+            bp2 = set(ts2.breakpoints())
+            assert bp1 != bp2
+            breaks = set()
+            for interval, t1, t2 in ts1.coiterate(ts2):
+                assert set(interval) <= set(t1.interval) | set(t2.interval)
+                breaks.add(interval.left)
+                breaks.add(interval.right)
+                assert t1.tree_sequence == ts1
+                assert t2.tree_sequence == ts2
+            assert breaks == bp1 | bp2
+
+    def test_simple_ts(self):
+        nodes = """\
+        id      is_sample   time
+        0       1           0
+        1       1           0
+        2       1           0
+        3       0           1
+        4       0           2
+        """
+        edges1 = """\
+        left    right   parent  child
+        0       0.2       3       0,1
+        0       0.2       4       2,3
+        0.2     1         3       2,1
+        0.2     1         4       0,3
+        """
+        edges2 = """\
+        left    right   parent  child
+        0       0.8       3       2,1
+        0       0.8       4       0,3
+        0.8     1         3       0,1
+        0.8     1         4       2,3
+        """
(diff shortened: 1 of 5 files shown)


hyanwong:
@benjeffery - this is failing because sphinx can't find the `Interval` class, which is actually defined right at the top of the trees.py file. I can't figure out why it doesn't work, when sphinx finds the `Tree` class just fine. Might you have time to have a look?

benjeffery (commented):
A few nits and suggestions. The `kwargs` functionality is not being tested yet.

hyanwong:
> A few nits and suggestions. The `kwargs` functionality is not being tested yet. There are also some failing tests that refer to the old tsutil version of this. The doc build is failing as `Interval` is not documented, i.e. `python-api.rst` doesn't contain `.. autoclass:: Interval`. The class itself is also missing docstrings.

All great suggestions, thanks @benjeffery . I didn't realise that documentation was required to link (although I guess that's obvious, really), Should I document Interval, even though it's trivial?

benjeffery:
> Should I document Interval, even though it's trivial?

If we're returning iterators using it then I guess it goes in the "Simple Container Classes" section.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1022/)

hyanwong:
All done. Thanks @benjeffery

python/tests/test_topology.py
```diff
@@ -4785,6 +4785,94 @@ def test_partial_overlap_contradictory_children(self):
...
+        assert interval.left == 0.8
+        assert interval.right == 1
+        assert tree1 == ts1.at_index(1)
+        assert tree2 == ts2.at_index(1)
+
+    def test_nonequal_lengths(self):
+        ts1 = msprime.simulate(4, random_seed=1, length=2)
+        ts2 = msprime.simulate(4, random_seed=1)
+        with pytest.raises(ValueError, match="equal sequence length"):
+            next(ts1.coiterate(ts2))
```
jeromekelleher, line 4866: We shouldn't need the ``next`` here I think

python/tests/test_topology.py
```diff
@@ -4785,6 +4785,94 @@ def test_partial_overlap_contradictory_children(self):
...
+        assert ts1.num_trees > 1
+        one_tree_ts = msprime.simulate(5, random_seed=2)
+        multi_tree_ts = msprime.simulate(5, recombination_rate=1, random_seed=2)
+        assert multi_tree_ts.num_trees > 1
+        for ts2 in (one_tree_ts, multi_tree_ts):
+            bp1 = set(ts1.breakpoints())
+            bp2 = set(ts2.breakpoints())
+            assert bp1 != bp2
+            breaks = set()
+            for interval, t1, t2 in ts1.coiterate(ts2):
```
jeromekelleher, line 4813: Good to also verify that the intervals for t1 and t2 are both in ``interval``.

jeromekelleher (commented):
Very nice, thanks @hyanwong. A couple of small tweaks and we're good to merge.

python/tskit/trees.py
```diff
@@ -3850,6 +3863,43 @@ def trees(
         )
         return TreeIterator(tree)
 
+    def coiterate(self, other, **kwargs):
```
jeromekelleher, line 3866: Actually, might make more sense to put this next to the ``trees`` iterator in the file? It's nice to keep methods somewhat grouped by topic.

python/tests/test_topology.py
hyanwong: I think we do. If I remove it, we don't get the error raised, because we don't actually start the generator going?

python/tests/test_topology.py
hyanwong: I think it's `assert set(interval) <= set(t1.interval) | set(t2.interval)`. Will add.

python/tskit/trees.py
hyanwong: Erm. That's where it is, isn't it?

python/tests/test_topology.py
hyanwong: Done

python/tests/test_topology.py
benjeffery: Generator functions are only executed when after the first `next` so yeah, we do need it.

jeromekelleher: Yes, my bad.

python/tskit/trees.py
jeromekelleher: Yep, so it is. GitHub's way of showing context is a bit weird sometimes

jeromekelleher (approved):

PR #1023, opened 2020-11-20
Add tests for genetic_relatedness

### LINKED ISSUE #974: Add more tests for genetic_relatedness
From @petrelharp on slack:

- write some more tests (e.g., for the branch and node methods), and incorporate these into the testing framework for the other statistics.

brieuclehmann:
This is to add more extensive tests for genetic_relatedness, under the test_tree_stats framework.

I've run into a few issues so far:

- `ts.genetic_relatedness` has the additional argument `proportion=True` (to divide by segregating_sites) which requires some additional plumbing in the overall testing framework. I'm not sure how best to set this up!
- `ts.segregating_sites` returns a vector when we have multiple windows. This was causing some unexpected output when dividing by segregating_sites (n_window x n_window) instead of (n_window) when `proportion=True`.  I've fixed this by adding `[:, None]` (see line 5603 in trees.py). Is this OK?
- I've just been looking at the site tests for now, and just the `test_single_tree_infinite_sites` case. I'm getting the following error: `ValueError: Summary function does not return zero for both zero weight and total weight.` which I'm not sure how to fix.

Fixes #974 .


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1023/)

petrelharp:
> Summary function does not return zero for both zero weight and total weight. 

Well, should it? It seems like it should, since "zero weight" corresponds to bits of the tree sequence not inherited by anyone, and "total weight" corresponds to bits inherited by everyone in the sample sets. (So, for instance, a mutation inherited by none of the samples has weight zero, and shouldn't contribute to trait covariance.) So, if `x=[0,0,...,0]` then in fact `f(x) = 0`, which is good:
```
        def f(x):
            mx = np.mean(x)
            return np.array([(x[i] - mx) * (x[j] - mx) / 2 for i, j in indexes])
```
Now, "full sample weight" would be `x = n`, where `n[i] = len(sample_sets[i])`, I think. And if all the sample sets have the same size, then `f(n) = 0`, again. But I guess this would not be zero for sample sets of different sizes? For instance, if `sample_sets = [[0], [2,3]]` (the covariance between a haploid and a diploid), and `indexes=[(0,0), (0,1), (1,1)]`, then `f([1,2]) = [0.25, -0.25, 0.25]`, I think? Is this correct?

petrelharp:
I suspect this means we need to redefine it a bit to account for the not-all-sample-sets-the-same-size case. (But in a way that's equivalent if they are.)

python/tskit/trees.py
```diff
@@ -5600,7 +5600,7 @@ def genetic_relatedness(
                 windows=windows,
                 mode=mode,
                 span_normalise=span_normalise,
-            )
+            )[:, None]
```
petrelharp, line 5603: Good thought, but sadly this won't work [with `windows=None`](https://tskit.readthedocs.io/en/latest/stats.html#windows). I think it *will* work, without the `[:,None]`, if you pass `[all_samples]` as the sample sets instead of `all_samples`. (passing a list of sample IDs instead of a list of lists of sample ids drops a dimension in the output; that's what's causing the problem here).

petrelharp:
> ts.genetic_relatedness has the additional argument proportion=True (to divide by segregating_sites) which requires some additional plumbing in the overall testing framework. I'm not sure how best to set this up!

Ah, right. I think what you've done (redefine `verify_definition`) is a good way to go - good job figuring out the spaghetti in there. Now, you could add `proportion` as an argument to it, and then do:
```
    def verify_sample_sets_indexes(self, ts, sample_sets, indexes, windows):
        def f(x):
            mx = np.mean(x)
            return np.array([(x[i] - mx) * (x[j] - mx) / 2 for i, j in indexes])

        for proportion in [True, False]:
         self.verify_definition(
             ts,
             sample_sets,
             indexes,
             windows,
             f,
             ts.genetic_relatedness,
             genetic_relatedness,
             proportion=proportion,
         )

```
... ?

petrelharp:
I've carefully read the python 'test' versions, by the way, and they look great.

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,205 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+            sample_sets=all_samples, windows=windows, mode="site"
+        )
+    for j in range(len(windows) - 1):
+        begin = windows[j]
+        end = windows[j + 1]
+        site_positions = [x.position for x in ts.sites()]
+        for i, (ix, iy) in enumerate(indexes):
+            X = sample_sets[ix]
+            Y = sample_sets[iy]
+            # site_in_window = False
```
jeromekelleher, line 1797: I'm not sure where the ``site_in_window`` thing came from - was this copied from another similar function? If so, let's delete the commented out stuff here.

python/tskit/trees.py
jeromekelleher: What was the resolution here @brieuclehmann?

jeromekelleher (commented):
I've had a look through @brieuclehmann, and it all looks good to me (although I've not thought through the exact details). I'm happy to merge whenever @petrelharp is.

python/tests/test_tree_stats.py
brieuclehmann: Yep, I had copied over from divergence, will delete!

brieuclehmann:
There's still a fair bit amount of work to do as I haven't started looking at the branch or node statistics. I'm a bit swamped at the moment but will have a go over the next few days.

brieuclehmann:
> I suspect this means we need to redefine it a bit to account for the not-all-sample-sets-the-same-size case. (But in a way that's equivalent if they are.)

One way to do this would be:
```
n = np.array([len(x) for x in sample_sets])
mn = np.mean(n)
def f(x):
     mx = np.mean(x)
     return np.array([(mn * x[i] / n[i] - mx) * (mn * x[j] / n[j] - mx) / 2 for i, j in indexes])
```
This is compatible with the previous definition since `mn / n[i] = 1` for all `i` when all the sample sets have the same length. Does this seem reasonable @petrelharp ?

brieuclehmann:
I'm also getting lots of warnings:
```RuntimeWarning: invalid value encountered in true_divide
    return numerator / denominator```
which I'm pretty sure is to do with dividing by zero when number of segregating is zero.

**Comment by petrelharp** [comment]
For zero segregating sites, I guess we should be returning `nan`s. See (I think) Fst for how to do this?

**Comment by petrelharp** [comment]
And, I like both your suggestions - I was thinking along the same lines. Do you think one makes more sense than the other? (e.g., which one has a easier/more sensible verbal definition?)

**Comment by brieuclehmann** [comment]
I think the second one makes more sense (and requires less work :D); it corresponds to what you should get if you take the definition of (site) genetic_relatedness to be the centered genotype matrix covariance, and so has the same trait covariance interpretation from the overleaf document.

**Comment by petrelharp** [comment]
I agree! I guess we just should be more careful about what "centering" means?

File: python/tests/test_tree_stats.py (hunk 2 above)
Comment (petrelharp) [reply python/tests/test_tree_stats.py:1797]: heh, whoops - I should do a cleanup, i guess

Comment (brieuclehmann) [reply python/tests/test_tree_stats.py:1797]: Oh no, you actually use `site_in_window` in divergence! I just commented it out for genetic_relatedness.

File: python/tskit/trees.py (hunk 1 above)
Comment (brieuclehmann) [reply python/tskit/trees.py]: @petrelharp 's suggestion of using `[all_samples]` worked! Should I only 'resolve' something once I've pushed the change?

**Comment by brieuclehmann** [comment]
I've just pushed the latest changes for site_genetic_relatedness. This still fails on two tests (`test_ghost_allele` and `test_single_tree_jukes_cantor`) - I haven't looked into why this is the case yet.

I haven't yet looked at node_ or branch_genetic_relatedness; at the moment, these are just copied from divergence from now with the plumbing changed to include `proportion`. I should have time later this week to have a go at these. It's not immediately clear to me what the 'naive' definitions _should_ be but I haven't given it a great deal of thought yet!

File: python/tskit/trees.py (hunk 1 above)
Comment (petrelharp) [reply python/tskit/trees.py]: I think waiting for the push makes sense! (but, don't worry about pushing a lot... it's nice to know what happened with these things...)

**Comment by petrelharp** [comment]
> It's not immediately clear to me what the 'naive' definitions should be but I haven't given it a great deal of thought yet!

Taking our definition-in-words in the documentation should make this clear, hopefully? And if not, updating the definition? That's the idea - implement the description-in-words in as simple and obviously-correct a way as possible. The divergence code should be a good place to start.

**Comment by brieuclehmann** [comment]
I've added branch_genetic_relatedness, which is passing all the tests! It follows the same logic as the site_genetic_relatedness, but treating each branch as a site. 

Just node_genetic_relatedness and sorting the two failing tests in site_genetic_relatedness to go...

**Comment by brieuclehmann** [comment]
So I think the issue with test_ghost_allele and test_single_tree_jukes_cantor is that G = ts.genotype_matrix() contains values other than {0,1}. For example, in test_ghost_allele, we have G = [2 0 0 0]. We _should_ be treating this as G = [1 0 0 0] and computing the covariance based on this - this is what ts.genetic_relatedness() is currently doing, but not the naive implementation, so I would need to tweak that. Does this sound about right @petrelharp ?

**Comment by petrelharp** [comment]
> G = ts.genotype_matrix() contains values other than {0,1}.

Oh, right, that'll be it - good spotting! `site_genetic_relatedness` can be fixed up by rewriting using the more general `x != y` formulation from the theory paper.

**Comment by brieuclehmann** [comment]
Yep, I think I've fixed it! Everything passing without issues now.

**Comment by jeromekelleher** [comment]
Woot! Sounds like we're nearly there @brieuclehmann, we just need the node stat now?

**Comment by brieuclehmann** [comment]
The node stat is done too @jeromekelleher !

**Comment by jeromekelleher** [comment]
Excellent! I think we just need @petrelharp's approval on the PR and we can merge then.

File: python/tests/test_tree_stats.py [hunk 3]
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+                for k in range(ts.num_sites):
+                    if (site_positions[k] >= begin) and (site_positions[k] < end):
+                        for x in X:
+                            x_index = np.where(samples == x)[0][0]
+                            for y in Y:
+                                y_index = np.where(samples == y)[0][0]
+                                S += (
+                                    haps_centered[x_index][k]
+                                    * haps_centered[y_index][k]
+                                    / 2
```
petrelharp, line 1810: I can't seem to make this factor of 2 agree with the theory (from overleaf) - I suppose this means the overleaf is wrong? Do you know where the error is?
Screenshot from 2020-11-26 05-17-19

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+    ts, sample_sets, indexes, windows=None, span_normalise=True, proportion=True
+):
+    out = np.zeros((len(windows) - 1, len(indexes)))
+    samples = ts.samples()
+    all_samples = list({u for s in sample_sets for u in s})
+    haps = ts.genotype_matrix(isolated_as_missing=False).T
+    denom = np.ones(len(windows))
+    if proportion:
+        denom = ts.segregating_sites(
+            sample_sets=all_samples, windows=windows, mode="site"
```
petrelharp, line 1786: this should include `span_normalise` also

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+                haps_mean = haps.mean()
+                haps_centered = area * (haps - haps_mean)
+                for i, (ix, iy) in enumerate(indexes):
+                    X = sample_sets[ix]
+                    Y = sample_sets[iy]
+                    for x in X:
+                        x_index = np.where(samples == x)[0][0]
+                        for y in Y:
+                            y_index = np.where(samples == y)[0][0]
+                            out[j][i] += haps_centered[x_index] * haps_centered[y_index]
```
petrelharp, line 1855: I think it'd be more clear what's going on here if `area` was just `tr.branch_length(v) * span` and here we had
```
                             out[j][i] += area * haps_centered[x_index] * haps_centered[y_index]
```
instead of putting it in with a sqrt() above.

benjeffery:
Great - I think this has come in time for 0.3.3 which we're planning for tomorrow, now that #1030 is done.

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+                break
+            branches = [(c, tr.parent(c)) for c in tr.nodes()]
+            span = min(end, tr.interval[1]) - max(begin, tr.interval[0])
+            for B in branches:
+                v = B[0]
+                area = np.sqrt(tr.branch_length(v) * span)
+                haps = np.zeros(ts.num_samples)
+                for x, u in enumerate(samples):
+                    haps[x] = np.int(tr.is_descendant(u, v))
+                haps_mean = haps.mean()
```
petrelharp, line 1846: Hm, won't this make the statistic differ depending on what samples are in the tree sequence? Which should not happen? For instance, suppose we have a tree sequence with 10 samples; it should be the case that
```
ts.genetic_relatedness([[0,1], [2,3,4]], mode='branch')
```
is equal to the same thing, computed after removing samples 5-9:
```
sts = ts.simplify([0,1,2,3,4])
sts.genetic_relatedness([[0,1], [2,3,4]], mode='branch')
```
Here's a quick check:
```
ts = msprime.simulate(10, mutation_rate=3, recombination_rate=3, random_seed=23)
branch_genetic_relatedness(ts, [[0,1], [2,3,4]], indexes=[(0,1)], windows=[0, ts.sequence_length])
# array([[0.13483581]])
sts = ts.simplify([0,1,2,3,4])
branch_genetic_relatedness(sts, [[0,1], [2,3,4]], indexes=[(0,1)], windows=[0, ts.sequence_length])
# array([[-0.206612]])
## but all is good for the basic implementation:
ts.genetic_relatedness([[0,1], [2,3,4]], indexes=[(0,1)], windows=[0, ts.sequence_length], mode='branch')
# array([[-0.206612]])
sts.genetic_relatedness([[0,1], [2,3,4]], indexes=[(0,1)], windows=[0, ts.sequence_length], mode='branch')
# array([[-0.206612]])
```

petrelharp: I believe that the centering is equivalent to the formulation we have written down, but I haven't written it out to check?

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+def node_genetic_relatedness(
+    ts, sample_sets, indexes, windows=None, span_normalise=True, proportion=True
+):
+    out = np.zeros((len(windows) - 1, ts.num_nodes, len(indexes)))
+    samples = ts.samples()
+    denom = np.ones((len(windows), ts.num_nodes))
+    if proportion:
+        all_samples = list({u for s in sample_sets for u in s})
+        denom = ts.segregating_sites(
+            sample_sets=all_samples, windows=windows, mode="node"
```
petrelharp, line 1873: needs `span_normalise`

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+def branch_genetic_relatedness(
+    ts, sample_sets, indexes, windows=None, span_normalise=True, proportion=True
+):
+    out = np.zeros((len(windows) - 1, len(indexes)))
+    samples = ts.samples()
+    denom = np.ones(len(windows))
+    if proportion:
+        all_samples = list({u for s in sample_sets for u in s})
+        denom = ts.segregating_sites(
+            sample_sets=all_samples, windows=windows, mode="branch"
```
petrelharp, line 1828: `span_normalise`

python/tests/test_tree_stats.py
```diff
@@ -1768,6 +1768,273 @@ class TestSiteDivergence(TestDivergence, MutatedTopologyExamplesMixin):
...
+        end = windows[j + 1]
+        site_positions = [x.position for x in ts.sites()]
+        for i, (ix, iy) in enumerate(indexes):
+            X = sample_sets[ix]
+            Y = sample_sets[iy]
+            S = 0
+            for a in alleles:
+                this_haps = haps == a
+                haps_mean = this_haps.mean(axis=0)
+                haps_centered = this_haps - haps_mean
```
petrelharp, line 1800: oh darn, this has the same problem as with centering for the branch stat below.

petrelharp:
Looking good, and, nearly there! But, gee - this has turned up an oversight in the testing set-up! We weren't running tests were the union of all sample sets was not equal to all the samples. I've fixed this over in #1039; you could merge that in to get tests that fail for the issue related to centering I'm pointing out in the comments.

brieuclehmann:
> I've fixed this over in #1039; you could merge that in to get tests that fail for the issue related to centering I'm pointing out in the comments.

I fear my (lack of) git skills might cause some issues. How should I merge #1039 to my current branch?

benjeffery:
@brieuclehmann I can merge if you like - you want #1039 to be part of this PR, yes?

brieuclehmann:
Um, possibly! I'd like to apply the tests that @petrelharp has added so that I can check that my fixes has worked.

brieuclehmann:
Ah right, I just pushed through my proposed changes. How should I go about checking these on https://github.com/benjeffery/tskit/tree/1023-1039 ?

python/tests/test_tree_stats.py
brieuclehmann: hmmm, does it come down to the factor of 2 'trickery'? So now the naive implementation reflects equation (3) rather than (4)?

benjeffery:
I've just updated and am running the tests! The changes in 1039 are quite small - you could prob cut and paste them as a quick fix.

brieuclehmann:
> you could prob cut and paste them as a quick fix.

Good idea! I've done this and it seems to be working for the genetic_relatedness tests on my machine... Any suggestions?

benjeffery:
Add them to this PR and push to see if CI is happy?

petrelharp:
The factor of 2 is bothering me. Can you do a spot check that this definition agrees with the equation we cite, in the case of a binary alleles matrix? We could just tack one more test on to check the normalization, like:
```
class TestSiteGeneticRelatedness(TestGeneticRelatedness, MutatedTopologyExamplesMixin):
    mode = "site"

    def test_one_more_thing(self):
        ts = msprime.simulate(10, mutation_rate=0.01, sequence_length=100, recombination_rate=0.01, random_seed=23)
        A = ts.genetic_relatedness([[0,1], [2,3], [4,5]], mode="site", span_normalise=False, ...)
        B = < computed as in Speed and Balding >
        self.assertArrayAlmostEqual(A, B)
```

petrelharp:
> Good idea! I've done this and it seems to be working for the genetic_relatedness tests on my machine... Any suggestions?

That's great! The fixed you made to the tests look right after a quick glance to me, so I believe it...

brieuclehmann:
> We could just tack one more test on to check the normalization, like:

I've added a test @petrelharp - let me know if that's what you had in mind!

petrelharp:
> I've added a test @petrelharp - let me know if that's what you had in mind!

Yes, perfect! Just two things: (1) put in a comment at the top of the test what it's doing (comparing to Speed & Balding's K_{c0}), and (2) move the test to `TestSiteGeneticRelatedness` (since right now it's getting called three times, including under `TestBranchGeneticRelatedness` and `TestNodeGeneticRelatedness`, which inherit from the class it's defined in).

jeromekelleher:
How's this looking @brieuclehmann? It would be super-nice if we could get the last of @petrelharp's comments sorted out so we can ship 0.3.3 today.

brieuclehmann:
Just waiting for @petrelharp 's approval on the latest changes I believe @jeromekelleher .

jeromekelleher:
Ah, so you've implemented the Speed and Balding method? I think we're good to go then.

jeromekelleher (approved):

jeromekelleher:
Looks to me like the test has been implemented, so we should be pretty air-tight now. If there's any remaining issues we can follow them up and fix (with a bugfix release, if necessary).

This is great, congratulations and thanks @brieuclehmann!

petrelharp:
Ah good, yes, sorry, I was off in the woods.

PR #1026, opened 2020-11-21
More combinatorial functions

jeromekelleher:
Adds some more generators for common tree shapes, and makes some overall improvements to the combintorics code.

Supersedes #944 

# PR Checklist:


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1026/)

jeromekelleher:
This is ready for review now I think. The main thing that's added is the ``generate_balanced`` function, which can do arbitrary arities, .e.g:
```
       15             
  ┏━━━━━╋━━━━━━┓      
  ┃     ┃     14      
  ┃     ┃   ┏━┳┻━━┓   
 11    12   ┃ ┃  13   
┏━╋━┓ ┏━╋━┓ ┃ ┃ ┏━╋━━┓
0 1 2 3 4 5 6 7 8 9 10
```

I've made sure that the generated trees are identical to the trees we'd get from unrank, and documented how we allocate internal nodes there.

@daniel-goldstein, would you mind taking a quick look here to makre sure I haven't done anything silly?

jeromekelleher:
@benjeffery - any idea how we should do abstract test classes using pytest? It's annoying me that we need to duplicate the code for tests like ``test_provenance`` here. I'd have done a Mixin class here in the old days, so that all the test classes inherit the test, but the test isn't executed on the abstract class.

benjeffery:
Well you can still use a Mixin if you like! Often i think a plain function at module-level is simpler, but there are times Mixins feel right. 

Generally i think the tests here look good - using the parameterisation decorator is good as it isolates failing cases easily, note that if you want all combinations you can stack the decorator. Also I like that we're grouping tests by class.

jeromekelleher:
@benjeffery, any chance of another look at the tests? I'm quite pleased how this turned out, might be quite a nice pattern to adopt/adapt elsewhere?

python/tskit/combinatorics.py
```diff
@@ -82,6 +106,135 @@ def random_binary_tree(leaf_labels, rng):
...
+        parameters = {"command": "generate_star", "TODO": "add parameters"}
+        tables.provenances.add_row(
+            record=json.dumps(tskit.provenance.get_provenance_dict(parameters))
+        )
+    return tables.tree_sequence().first()
+
+
+def generate_comb(num_leaves, *, span, branch_length, record_provenance):
+    """
+    Generate a comb tree for the specified number of leaves.
```
hyanwong, line 163: Also called a "pectinate" tree - probably worth including in the docs for search purposes

python/tskit/trees.py
```diff
@@ -2451,29 +2451,79 @@ def generate_star(num_leaves, *, span=1, branch_length=1, record_provenance=True
...
+        :param int num_leaves: The number of leaf nodes in the returned tree (must be
+            be 2 or greater).
+        :param int arity: The maximum number of children a node can have in the returned
+            tree.
+        :param float span: The span of the tree, and therefore the
+            :attr:`~TreeSequence.sequence_length` of the :attr:`.tree_sequence`
+            property of the returned :class:<Tree>.
+        :param float branch_length: The minimum length of a branch in the tree (see
+            above for details on how internal node times are assigned).
+        :return: A star-shaped tree. Its corresponding :class:`TreeSequence` is available
```
hyanwong, line 2488: We aren't returning a star shaped tree here, but a balanced one

benjeffery:
> @benjeffery, any chance of another look at the tests? I'm quite pleased how this turned out, might be quite a nice pattern to adopt/adapt elsewhere?

I've just realised another way to do this as you can parameterise a whole class:
```python

@pytest.mark.parametrize('method', ["flip", "catch", "hurl"])
class TestCommonsStuff:
    def test_foobar(self, method):
        assert ....
```
and then put the specific tests in their own class.

It is a matter of taste as always - I find what you have already very clear.

jeromekelleher:
Good idea @benjeffery! I think I'll stick with the subclasses here as it's nice to be able to tack on extra specific tests as well as the common ones.

python/tskit/trees.py
```diff
@@ -2451,29 +2451,79 @@ def generate_star(num_leaves, *, span=1, branch_length=1, record_provenance=True
...
-            left=np.full(num_leaves, 0),
-            right=np.full(num_leaves, span),
-            parent=np.full(num_leaves, root, dtype=np.int32),
-            child=np.arange(num_leaves, dtype=np.int32),
+
+    @staticmethod
+    def generate_comb(num_leaves, *, span=1, branch_length=1, record_provenance=True):
+        """
+        Generate a :class:<Tree> in which all internal nodes have two children
+        and the left child is a leaf. This is a "comb" or "ladder" phylogeny.
```
hyanwong, line 2504: "comb", "ladder", or "pectinate"

grahamgower: Oh, you mean a https://en.wikipedia.org/wiki/Caterpillar_tree ? :P

hyanwong: Ah, yes, I've heard that as well, now you come to mention it. I asked a few phylogeneticists and they said the convention was "comb" or (more formally) "pectinate" when the tree is rooted. Caterpillar is nice, though.

jeromekelleher:
Thanks for the feedback, I've addressed your comments @hyanwong. I'm marked this as "merge-ready" so it just needs an approval and it'll go in.

daniel-goldstein:
Sorry spent all yesterday moving out for the holidays. I can still give it a look now if you'd like

jeromekelleher:
> Sorry spent all yesterday moving out for the holidays. I can still give it a look now if you'd like

That would be great, thanks @daniel-goldstein! No particular hurry.

python/tskit/combinatorics.py
```diff
@@ -37,6 +37,30 @@
...
+
+    Otherwise, we always return k chunks. The first k - 1 chunks will
+    contain exactly n // k items, and the last chunk the remainder.
+    """
+    n = len(lst)
+    if k <= 0 or int(k) != k:
+        raise ValueError("Number of chunks must be a positive integer")
+
+    if n > 0:
+        chunk_size = max(1, n // k)
```
daniel-goldstein, line 54: Can you do this instead with `yield from itertools.islice(lst, chunk_size)`?

daniel-goldstein (approved):
This looks really great! Loving the pytest decorators. Just one little comment

python/tskit/combinatorics.py
jeromekelleher: Hey, good idea. No, I don't think islice will do it, the semantics of what we want are really quite specific in order to line up with the unranked trees.  numpy has a [function](https://numpy.org/doc/stable/reference/generated/numpy.array_split.html) which almost does what we want, but not quite.

PR #1030, opened 2020-11-23
Fix parsimony on nonbinary trees

### LINKED ISSUE #987: Bug in map_mutations() for polytomy
@jeromekelleher 

I used map_mutations to remap genotypes onto my inferred tree sequence and got unexpected results. The total number of mutations was _higher_ than that inferred by `tsinfer`. I whittled it down to a reproducible, minimal example:

Screenshot 2020-11-11 at 12 57 56 PM

Here is a tree in the (simplified down) tree sequence. The genotype array is:
`[1, 0, 0, 0, 0, 1]` and alleles are `('C', 'T')`.

Now we run: 
`tree.map_mutations(np.array([1, 0, 0, 0, 0, 1]), ("C", "T"))`
And we get _three_ mutations:
`('C', [{'id': -1, 'site': -1, 'node': 6251, 'time': nan, 'derived_state': 'T', 'parent': -1, 'metadata': b''}, {'id': -1, 'site': -1, 'node': 2, 'time': nan, 'derived_state': 'C', 'parent': 0, 'metadata': b''}, {'id': -1, 'site': -1, 'node': 4, 'time': nan, 'derived_state': 'C', 'parent': 0, 'metadata': b''}])`

Am I correct in thinking this is a bug? Doc of `map_mutations()` states:

> The current implementation uses the Fitch parsimony algorithm to determine the minimum number of state transitions required to explain the data. In this model, transitions between any of the non-missing states is equally likely.

But map_mutations infers an extra mutation. Perhaps could be an issue with polytomies?

Here's code to reproduce:
```
import tskit
import io

def polytomy_tree_ts():
    nodes = io.StringIO("""\
    id      is_sample   time
    0       1           0
    1       1           0
    2       1           0
    3       1           0
    4       1           0
    5       1           0
    6       0           1
    7       0           2
    8       0           2
    9       0           3
    """)
    edges = io.StringIO("""\
    left    right   parent  child
    0       1       6       0,2,4
    0       1       7       6,5
    0       1       8       1,3
    0       1       9       7,8
    """)
    sites = io.StringIO("""\
    position    ancestral_state
    0.5         0
    """)
    mutations = io.StringIO("""\
    site    node    derived_state
    0       0       1
    0       5       1
    """)
    return tskit.load_text(nodes=nodes, edges=edges,sites=sites, mutations=mutations, strict=False)

ex = polytomy_tree_ts()
ex.first().map_mutations(var.genotypes, ("G", "C"))
```

jeromekelleher:
Closes #987 

# PR Checklist:

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -28,6 +28,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <math.h>
+#include <assert.h>
 
 #include <tskit/trees.h>
 
@@ -4268,6 +4269,7 @@ get_smallest_set_bit(uint64_t v)
     uint64_t t = 1;
     int8_t r = 0;
 
+    assert(v != 0);
     while ((v & t) == 0) {
         t <<= 1;
         r++;
@@ -4275,9 +4277,14 @@ get_smallest_set_bit(uint64_t v)
     return r;
 }
 
+#define HARTIGAN_MAX_ALLELES 64
+
 /* This interface is experimental. In the future, we should provide the option to
  * use a general cost matrix, in which case we'll use the Sankoff algorithm. For
  * now this is unused.
+ *
+ * The algorithm used here is Hartigan parsimony, "Minimum Mutation Fits to a
+ * Given Tree", Biometrics 1973.
  */
 int TSK_WARN_UNUSED
 tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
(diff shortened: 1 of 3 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1030/)

jeromekelleher:
This is ready to go I think. @hyanwong, can you take a look at this please? In particular, can you think of other test examples we should put in here, to make sure the result is parsimonious?

hyanwong:
Will do. Sorry for the slow reply - internet issues.

hyanwong:
> This is ready to go I think. @hyanwong, can you take a look at this please? In particular, can you think of other test examples we should put in here, to make sure the result is parsimonious?

This all looks great. Thanks for sorting this out @jeromekelleher . If you are looking for more trees to test, then perhaps

1. we could (trivially) test a star topology, which should give the ancestral state being the highest freq mutation, and the number of transitions being the count of lowest freq alleles (perhaps you have, but I can't find it by searching for "star" in the changes).
2. a few more tests with more than 2 states might be helpful. For instance,
    i. the star topology above with n tips and n-1 alleles - the ancestral state should be the only one present on 2 tips.
    ii. the least parsimonious possible version of your `tskit.Tree.generate_balanced(27, arity=3)` tree, where there are 3 alleles, distributed as `[0,1,2] * (27//3)` over the tips.

It's really helpful to have the "balanced tree" generation code now! Thanks for doing that. It also occurred to me that it's now trivial to produce a "random" binary tree for testing, by coupling `star_topology` with `split_polytomies` - do you think that would be useful, in which case should I open an issue?

hyanwong:
Oh, and another thing: should this deal with (a) internal sample nodes, especially when there is an internal sample node which is also a polytomy, and (b) tip nodes that aren't samples ("dangling nodes). I suspect (b) is already tested, but maybe not when some of the branches from a polytomy are dangling, and some aren't. 

(a) is more important to test than (b), I would think.

jeromekelleher:
> It's really helpful to have the "balanced tree" generation code now! Thanks for doing that. It also occurred to me that it's now trivial to produce a "random" binary tree for testing, by coupling star_topology with split_polytomies - do you think that would be useful, in which case should I open an issue?

I did think of this, but then wondered if it would really be of much use since we can already call msprime.simulate() to do much the same  thing. I was also slightly worried that people might use this as some sort of evolutionary model, when it's not. I could easily be convinced it's a good idea, though, if you think it's a useful thing to add.

jeromekelleher:
Thanks for the review, I'll add the tests you mention.

hyanwong:
Oh, and lastly (promise!), are we dealing OK with polytomies that have missing data - both where some or all of the children of a polytomy are marked as missing (e.g. a star topology with only 1 sample non missing), and where there is an internal sample but that sample is marked as missing (I guess this should behave just as if the internal sample node wasn't a sample at all.

hyanwong:
> I did think of this, but then wondered if it would really be of much use since we can already call msprime.simulate() to do much the same thing. I was also slightly worried that people might use this as some sort of evolutionary model, when it's not. I could easily be convinced it's a good idea, though, if you think it's a useful thing to add.

That's a very good point. The (relatively weak) arguments that I can see for it are:

1. It allows tskit to produce random trees without relying on an external dependency. That could be useful e.g. for SLiM testing, or what have you.
2. The distribution of coalescence tree topologies is not equiprobable - see below - the top plot is msprime topologies, the bottom is from split_polytomy.

```
import collections

import tqdm
import msprime
import tskit

msprime_counts = collections.defaultdict(int)
split_poly_counts = collections.defaultdict(int)
for s in tqdm.trange(1, 10000):
    msprime_counts[msprime.simulate(5, random_seed=s).first().rank()] += 1

for s in tqdm.trange(1, 10000):
    split_poly_counts[tskit.Tree.generate_star(5).split_polytomies(random_seed=s).rank()] += 1

fig, axes = plt.subplots(2)
axes[0].bar(list(range(len(msprime_counts))), msprime_counts.values())
axes[1].bar(list(range(len(split_poly_counts))), split_poly_counts.values())
```

image

jeromekelleher:
OK, let's break the ``generate_random`` discussion to a separate issue. Can you copy the comment above so we don't lose track of it please?

jeromekelleher:
I've added the extra tests @hyanwong - great ideas. The internal samples and missing data stuff was already very thoroughly tested, and I've added in a few extra cases where we're using the balanced trees of varying arity. I think we can be confident this is correct now!

hyanwong:
Great. This seems mergeable to me then.

python/tests/test_parsimony.py
```diff
@@ -635,56 +741,81 @@ class TestParsimonyMissingData(TestParsimonyBase):
     Tests that we correctly map_mutations when we have missing data.
     """
 
-    def test_all_missing(self):
-        for n in range(2, 10):
-            ts = msprime.simulate(n, random_seed=2)
-            tree = ts.first()
+    @pytest.mark.parametrize("n", range(2, 10))
+    def test_all_missing(self, n):
+        ts = msprime.simulate(n, random_seed=2)
```
benjeffery, line 746: The places you have `ts = msprime.simulate(n, random_seed=2)` could use a session-scoped parameterised fixture. https://docs.pytest.org/en/stable/fixture.html#fixture-parametrize

benjeffery (approved):
Loving the tests here. One suggestion about fixtures - but it is no biggie as happy either way.

hyanwong:
Great stuff. Does this automatically make it into the matching algorithm now?

jeromekelleher:
> Great stuff. Does this automatically make it into the likelihood compression part of the matching algorithm now?

No, that's got to be done separately.

c/tskit/trees.c
```diff
@@ -4268,16 +4269,22 @@ get_smallest_set_bit(uint64_t v)
     uint64_t t = 1;
     int8_t r = 0;
 
+    assert(v != 0);
```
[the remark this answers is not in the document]
jeromekelleher: This is pretty perf sensitive code, so I want to compile out the assert.

python/tests/test_parsimony.py
jeromekelleher: might pick that up later - it's hard to know where to draw the line when updating the test code!

PR #1031, opened 2020-11-23
Fix flag handling in Python C API

### LINKED ISSUE #1027: Flags are parsed as signed in the C Python extension.
`tables.nodes.flags` and `tables.individuals.flags` are listed as dtype `np.uint32`. But it looks like the C library is expecting them to be set as signed 32 bit ints instead:

```
ts = msprime.simulate(4, mutation_rate=10)
tables = ts.dump_tables()
tables.nodes.add_row(flags = np.iinfo(np.int32).max)  # works
tables.nodes.add_row(flags = np.iinfo(np.uint32).max)  # fails with "OverflowError: signed integer is greater than maximum"
tables.individuals.add_row(flags = np.iinfo(np.int32).max)  # works
tables.individuals.add_row(flags = np.iinfo(np.uint32).max)  # fails with "OverflowError: signed integer is greater than maximum"
```

benjeffery:
## Description

Flags were not handled right in the Python C API layer. The PR introduces a converter function to fix this.

Fixes #1027 

# PR Checklist:

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -736,6 +736,31 @@ out:
     return ret;
 }
 
+static int
+uint32_converter(PyObject *py_obj, uint32_t *uint_out)
+{
+    long long temp_long;
+    int ret = 0;
+
+    if (!PyArg_Parse(py_obj, "L", &temp_long)) {
+        goto out;
+    }
+    if (temp_long > UINT32_MAX) {
+        PyErr_SetString(PyExc_OverflowError, "unsigned int32 >= than 2^32");
+        goto out;
+    }
+    if (temp_long < 0) {
+        PyErr_SetString(
+            PyExc_ValueError, "Can't convert negative value to unsigned int");
+        goto out;
+    }
+
+    uint_out[0] = (uint32_t) temp_long;
+    ret = 1;
+out:
+    return ret;
+}
+
 /*===================================================================
  * IndividualTable
  *===================================================================
@@ -823,8 +848,8 @@ IndividualTable_add_row(IndividualTable *self, PyObject *args, PyObject *kwds)
     if (IndividualTable_check_state(self) != 0) {
         goto out;
     }
-    if (!PyArg_ParseTupleAndKeywords(
-            args, kwds, "|iOO", kwlist, &flags, &py_location, &py_metadata)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O&OO", kwlist, &uint32_converter,
+            &flags, &py_location, &py_metadata)) {
         goto out;
     }
     if (py_metadata != Py_None) {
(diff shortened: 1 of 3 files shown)


benjeffery:
Gah, this new code doesn't like `numpy.uint32` as a type to covert.

AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1031/)

benjeffery:
Thought of a better way to do this - resetting to draft.

benjeffery:
@jeromekelleher I've found a way which copes with `numpy.uint32`. Should be good to go.

benjeffery:
Ha spoke too soon, it failed on windows.

python/_tskitmodule.c
```diff
@@ -736,6 +736,31 @@ make_file(PyObject *fileobj, const char *mode)
     return ret;
 }
 
+static int
+uint32_converter(PyObject *py_obj, uint32_t *uint_out)
+{
+    int64_t temp_long;
+    int ret = 0;
+
+    if (!PyArg_Parse(py_obj, "l", &temp_long)) {
```
jeromekelleher, line 745: Probably best to use [this function](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLongAndOverflow) if we want to be absolutely 100% sure.

python/tskit/tables.py
```diff
@@ -721,6 +721,7 @@ def add_row(self, flags=0, time=0, population=-1, individual=-1, metadata=None):
         :rtype: int
         """
         metadata = self.metadata_schema.validate_and_encode_row(metadata)
+        print(flags)
```
jeromekelleher, line 724: print

python/_tskitmodule.c
jeromekelleher: Hmm, looks like we're already catching overflows properly anyway, so maybe no need to bother with that. We can't assume a long is 64 bit though, we need a ``long long`` for that.

jeromekelleher (commented):
LGTM, minor comments.

python/tskit/tables.py
benjeffery: Yeah I'm still debugging windows.

python/_tskitmodule.c
benjeffery: I tried `PyLong_AsLongLongAndOverflow` but it doesn't like the `np.uint32` objects that `add_row` currently recieves.

jeromekelleher: Dammit. PyArg_Parse is good anyway, just use a long long. That's probably the issue on windows, as a long is 32 bits there.

python/_tskitmodule.c
```diff
@@ -736,6 +736,32 @@ make_file(PyObject *fileobj, const char *mode)
     return ret;
 }
 
+static int
+uint32_converter(PyObject *py_obj, uint32_t *uint_out)
+{
+    int64_t temp_long;
```
jeromekelleher, line 742: This needs to be declared a ``long long`` I think, as that's what the python C code is assuming. It could happen that the long long is more than 64 bits or something - anyway, no point in mixing fixed widths and old style types. We can be sure a long long is at least 64 bits.

jeromekelleher (approved):
LGTM

PR #1037, opened 2020-11-25
Preliminary PR for random tree generation

### LINKED ISSUE #1033: Add Tree.generate_random() method
It's now trivial to add a `tskit.Tree.generate_random()` method for generating random bifurcating trees with equal probability for each topology. All we have to do is:

```
def generate_random(self, num_leaves, random_seed=None):
    return self.generate_star(num_leaves).split_polytomy(random_seed)
```

And we probably don't need another unit test for this, because we already test it [here](https://github.com/tskit-dev/tskit/blob/71beec57e14720346973a6a62be94d368c6ef03a/python/tests/test_combinatorics.py#L1100), so all we need to do (IMO) is change that function to `generate_random()` rather than `split_polytomies()`, and we should be covered.

The arguments for this are

1. It allows tskit to produce random trees without relying on an external dependency. That could be useful e.g. for SLiM testing, or what have you.
2. The distribution of coalescence tree topologies is not equiprobable - see below - the top plot is msprime topologies, the bottom is from split_polytomy.

```
import collections

import tqdm
import msprime
import tskit

msprime_counts = collections.defaultdict(int)
split_poly_counts = collections.defaultdict(int)
for s in tqdm.trange(1, 10000):
    msprime_counts[msprime.simulate(5, random_seed=s).first().rank()] += 1

for s in tqdm.trange(1, 10000):
    split_poly_counts[tskit.Tree.generate_star(5).split_polytomies(random_seed=s).rank()] += 1

fig, axes = plt.subplots(2)
axes[0].bar(list(range(len(msprime_counts))), msprime_counts.values())
axes[1].bar(list(range(len(split_poly_counts))), split_poly_counts.values())
```

image

I'm happy to submit this as a PR if the approach I describe is the right one.

hyanwong:
## Description

A draft PR to create random trees. No changelog yet, sorry!

Fixes #1033

# PR Checklist:

- [x ] Tests that fully cover new/changed functionality.

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -1092,19 +1092,6 @@ class TestPolytomySplitting:
             strict=False,
         )
 
-    @pytest.mark.slow
-    @pytest.mark.parametrize("n", [2, 3, 4, 5])
-    def test_all_topologies(self, n):
-        N = num_leaf_labelled_binary_trees(n)
-        ranks = collections.Counter()
-        tree = tskit.Tree.generate_star(n)
-        for seed in range(20 * N):
-            split_tree = tree.split_polytomies(random_seed=seed)
-            ranks[split_tree.rank()] += 1
-        # There are N possible binary trees here, we should have seen them
-        # all with high probability after 20 N attempts.
-        assert len(ranks) == N
-
     def verify_trees(self, source_tree, split_tree, epsilon=None):
         N = 0
         for u in split_tree.nodes():
@@ -1471,6 +1458,34 @@ class TestGenerateBalanced(TreeGeneratorTestBase):
             for v in tree.children(u):
                 # Special case cause n is a power of 2
                 assert tree.time(u) == tree.time(v) + branch_length
+
+
+class TestGenerateRandom(TreeGeneratorTestBase):
+    method_name = "generate_random"
+
+    @pytest.mark.slow
+    @pytest.mark.parametrize("n", [3, 4, 5])
+    def test_all_topologies(self, n):
+        N = num_leaf_labelled_binary_trees(n)
+        ranks = collections.Counter()
+        for seed in range(20 * N):
+            random_tree = tskit.Tree.generate_random(n, random_seed=seed)
+            ranks[random_tree.rank()] += 1
+        # There are N possible binary trees here, we should have seen them
+        # all with high probability after 20 N attempts.
+        assert len(ranks) == N
+
+    @pytest.mark.skip(
+        reason="random tree via random_split does not create leaves in a defined order"
+    )
+    def test_leaves(self, n):
+        pass
+
+    @pytest.mark.skip(
+        reason="random tree via random_split does not create nodes in a defined order"
+    )
+    def test_rank_unrank_round_trip(self, n):
+        pass
 
 
 class TestGenerateComb(TreeGeneratorTestBase):
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -1333,6 +1333,19 @@ class TestPolytomySplitting:
         split_tree = tree.split_polytomies(random_seed=14, tracked_samples=[0, 1])
         assert split_tree.num_tracked_samples() == 2
 
+    @pytest.mark.slow
+    @pytest.mark.parametrize("n", [3, 4, 5])
+    def test_all_topologies(self, n):
+        N = num_leaf_labelled_binary_trees(n)
+        ranks = collections.Counter()
+        for seed in range(20 * N):
+            star = tskit.Tree.generate_star(n)
+            random_tree = star.split_polytomies(random_seed=seed)
+            ranks[random_tree.rank()] += 1
+        # There are N possible binary trees here, we should have seen them
+        # all with high probability after 20 N attempts.
+        assert len(ranks) == N
+
 
 class TreeGeneratorTestBase:
     """
@@ -1460,8 +1473,11 @@ class TestGenerateBalanced(TreeGeneratorTestBase):
                 assert tree.time(u) == tree.time(v) + branch_length
 
 
-class TestGenerateRandom(TreeGeneratorTestBase):
-    method_name = "generate_random"
+class TestGenerateRandomBinary(TreeGeneratorTestBase):
+    method_name = "generate_random_binary"
+
+    def method(self, n, **kwargs):
+        return tskit.Tree.generate_random_binary(n, random_seed=53, **kwargs)
 
     @pytest.mark.slow
     @pytest.mark.parametrize("n", [3, 4, 5])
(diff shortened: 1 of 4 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1037/)

jeromekelleher:
I've made some updates here @hyanwong, mainly just renamed the method to ``generate_random_binary``. My original proposal was to force users to specify ``arity=2``, since i would expect ``arity=None`` to uniformly sample from *all* trees, not just the binary ones. But, this was fiddly, and it seems like the binary case is worth its own method.

It's mostly working, except for a test failure where the internal node labellings of the unranked tree disagrees with the generated one. I don't follow why this is, so I'd like to get to the bottom of it - this isn't super urgent anyway, it's just a nice-to-have at some point.

The disagreement is here: the first tree is the one we generate, and the second one is the result of unranking the rank of the first one:
```
tests/test_combinatorics.py::TestGenerateRandomBinary::test_rank_unrank_round_trip[8] rank =  (212, 3328)
5.00┊          14     ┊                                                                                                       
    ┊        ┏━━┻━━┓  ┊                                                                                                       
4.00┊       12     ┃  ┊  
    ┊     ┏━━┻━━┓  ┃  ┊                                                                                                       
3.00┊    11     ┃  ┃  ┊  
    ┊   ┏━┻━━┓  ┃  ┃  ┊                                        
2.00┊   9    ┃  ┃  ┃  ┊                                        
    ┊  ┏┻━┓  ┃  ┃  ┃  ┊       
1.00┊  8  ┃ 10  ┃ 13  ┊    
    ┊ ┏┻┓ ┃ ┏┻┓ ┃ ┏┻┓ ┊                                        
0.00┊ 0 5 1 6 7 2 3 4 ┊       
  0.00              1.00                                       
                                                               
5.00┊          14     ┊                                        
    ┊        ┏━━┻━━┓  ┊                                                                                                       
4.00┊       13     ┃  ┊                                                                                                       
    ┊     ┏━━┻━━┓  ┃  ┊        
3.00┊    12     ┃  ┃  ┊                                                                                                           ┊   ┏━┻━━┓  ┃  ┃  ┊                                                                                                       
2.00┊  11    ┃  ┃  ┃  ┊                                                                                                           ┊  ┏┻━┓  ┃  ┃  ┃  ┊                                        
1.00┊ 10  ┃  9  ┃  8  ┊       
    ┊ ┏┻┓ ┃ ┏┻┓ ┃ ┏┻┓ ┊                                        
0.00┊ 0 5 1 6 7 2 3 4 ┊                                                                                                       
  0.00              1.00              
```

Both of these are assigning internal node labels via postorder, but the unrank one has flipped the order of children among parents for some reason. The first tree is clearly "right", so I'd like to understand what the unranked tree is doing here. @daniel-goldstein, and chance you could take a look here please?

daniel-goldstein:
So I believe this has to do with a conflict in the canonical orientation between the ranking code and code to generate random binary trees. If you don't use the the minlex ordering when printing the unranked tree you should get the tree printed in a different layout where the assignment of the internal nodes looks "right". It is unfortunate that there isn't a single canonical orientation, but at least assigning only occurs when producing a `tskit.Tree` from a `RankTree`, so we could potentially impose a `minlex` ordering at that boundary.

hyanwong:
> I've made some updates here @hyanwong, mainly just renamed the method to `generate_random_binary`. My original proposal was to force users to specify `arity=2`, since i would expect `arity=None` to uniformly sample from _all_ trees, not just the binary ones. But, this was fiddly, and it seems like the binary case is worth its own method.

I liked your previous idea of specifying `arity=2. Why not create the `generate_random` function anyway, which just returns a NotImplementedError unless `arity=2`, in which case it uses your specific binary method? You could even implement a temporary hack which just uses unrank for the `arity=None` case, and raises an error if it is called with more than (say) 8 tips?

jeromekelleher:
> I liked your previous idea of specifying arity=2. Why not create the generate_random function anyway, which just returns a NotImplementedError unless arity=2, in which case it uses your specific binary method? You could even implement a temporary hack which just uses unrank for the arity=None case, and raises an error if it is called with more than (say) 8 tips?

Just doesn't seem worth the effort - there's a few hours work involved, and ``generate_random_binary`` is useful enough on its own. I've ended up sinking a fair bit of time into this stuff already...

jeromekelleher:
Thanks @daniel-goldstein, I'll take a look at this later.

jeromekelleher:
Aha, thanks @daniel-goldstein! I think I've fixed the issue now by canonicalising the order of the children within a node by sorting them by ``(num_leaves, min_label)``. This is *slightly* different to the canonical order used in the rank tree, which sorts by ``(num_leaves, rank, min_label)`` - do you think what I've done is correct? (Presumably we can get away with a simpler rule because these are binary trees?)

I've marked this as ready for review, but we should keep it back until after 0.3.3 is released; we can update the changelog and merge then.

python/tskit/trees.py
```diff
@@ -2541,6 +2541,50 @@ def generate_comb(num_leaves, *, span=1, branch_length=1, record_provenance=True
...
+            evolution. In order to simulate such trees, additional software
+            such as `msprime <https://github.com/tskit-dev/msprime>`` is required.
+
+        :param int num_leaves: The number of leaf nodes in the returned tree (must
+            be 2 or greater).
+        :param float span: The span of the tree, and therefore the
+            :attr:`~TreeSequence.sequence_length` of the :attr:`.tree_sequence`
+            property of the returned :class:<Tree>.
+        :param float branch_length: The time separating successive non-leaf nodes
+            from each other.
```
jeromekelleher, line 2575: Note: we need to document random_seed here.

jeromekelleher:
I think this is ready to go - @hyanwong, can you take a final look please?

jeromekelleher (approved):

python/tests/test_combinatorics.py
```diff
@@ -1346,6 +1333,19 @@ def test_kwargs(self):
...
+    def test_all_topologies(self, n):
+        N = num_leaf_labelled_binary_trees(n)
+        ranks = collections.Counter()
+        for seed in range(20 * N):
+            star = tskit.Tree.generate_star(n)
+            random_tree = star.split_polytomies(random_seed=seed)
+            ranks[random_tree.rank()] += 1
+        # There are N possible binary trees here, we should have seen them
+        # all with high probability after 20 N attempts.
+        assert len(ranks) == N
```
hyanwong, line 1347: This is a reasonable test, but I'm not sure it's very sensitive to the uniformity of the distribution of topologies. It may be that we don't care about testing that too much, but lack of testing this has caught out other phylogenetic libraries (e.g. R & Dendropy) so it would, I suppose, be good practice to have something (I think I has something like that in the previous tests, but it can turn out to be rather slow).

python/tskit/combinatorics.py
```diff
@@ -104,6 +131,24 @@ def random_binary_tree(leaf_labels, rng):
         root = nodes[0]
         while root.parent is not None:
             root = root.parent
+
+        # Canonicalise the order of the children within a node. This
+        # is given by (num_leaves, min_label). See also the
+        # RankTree.canonical_order function for the definition of
+        # how these are orderd during rank/unrank.
```
hyanwong, line 138: typo - "orderd"

python/tskit/combinatorics.py
```diff
@@ -71,6 +71,33 @@ class TreeNode:
     children = attr.ib(factory=list)
     label = attr.ib(default=None)
 
+    def as_tables(self, *, num_leaves, span, branch_length):
```
hyanwong, line 74: nice!

hyanwong (commented):
LGTM. One typo, and a possible addition to tests (but this may not be necessary, and could be done later anyway)

hyanwong:
NB: Name-wise, I sort-of prefer the conciseness of "generate_random(self, arity=2)`, and raising a NotImplementedError if anything other than 2 is specified, but we could make such a thing later if need be.

jeromekelleher:
> NB: Name-wise, I sort-of prefer the conciseness of generate_random(self, arity=2), and raising a NotImplementedError if anything other than 2 is specified, but we could make such a thing later if need be.

I started with this, and it turned out to be quite tedious in practise having to type ``arity=2`` every time. This is a big enough special case to warrant it's own method, I think. Anyway, I don't imagine anyone is going to implement an efficient general tree sampler any time soon, so there's no much point in making the interface more clumsy for it.

hyanwong:
> I started with this, and it turned out to be quite tedious in practise having to type `arity=2` every time. This is a big enough special case to warrant it's own method, I think. Anyway, I don't imagine anyone is going to implement an efficient general tree sampler any time soon, so there's no much point in making the interface more clumsy for it.

Sorry, I meant, with `arity=2` being the default, so you don't have to actually type it. That stick me as the value for arity that's most likely to be expected as a default, no?

python/tests/test_combinatorics.py
jeromekelleher: It's very difficult to do this sort of test rigorously, robustly and quickly enough to be a unit test. We'd need some sort of validation suite for this. Please open an issue to track if you think it's needed.

jeromekelleher:
No, I disagree with that. I wouldn't expect a general ``random_tree`` method to just choose from the binary trees by default. That would be weird.

python/tests/test_combinatorics.py
hyanwong: Yes, agreed. Probably only worth it if we require a validation suite for other reasons, I suppose. I can't think of any off hand, though.

hyanwong:
> No, I disagree with that. I wouldn't expect a general `random_tree` method to just choose from the binary trees by default. That would be weird.

Ah, I see, you mean if arity was none you would expect to pick from *all* trees, of any arity, with equal probability. Yes, I can see that as a more sensible default if arity is not specified. So don't mind me.

PR #1052, opened 2020-11-28
Change refs in map mutation docs from Fitch to Hartigan

hyanwong:
Missed a few of these. Worth doing a `grep -r Fitch *` once the code in haplotypes.c is changed, so we are sure we've caught them all (bar some references in the test suite, which still uses Fitch for testing binary trees)

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1211,7 +1211,7 @@ The output is::
 Missing data
 ++++++++++++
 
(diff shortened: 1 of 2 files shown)


AdminBot-tskit:
📖 Docs for this PR can be previewed [here](https://tskit-dev.github.io/tskit-docs/PR-1052/)

jeromekelleher (approved):
Good spot, thanks.

PR #1075, opened 2020-12-05
Update conftest.py

hyanwong:
The docstring description was confusing me, as it said all fixtures should end in _fixture, but then didn't show that in the example. I think the correction in the PR is correct @benjeffery ? Also since nearly all our code is in classes, I made the example into a class method. If this is right we should probably change the docstrings in a similar way in all the conftest.py files in other tskit repos?


benjeffery (approved):

benjeffery:
Thanks @hyanwong!

PR #1077, opened 2020-12-07
Change metadata schemas to use repr for the encoded schema, making str pretty print

### LINKED ISSUE #1074: MetadataSchema.__str__ should give pretty-printed schema
The output of ``print(schema)`` is hard to follow. We should give a more human-friendly version in ``__str__``. An easy way to do this would be to use ``pprint.pprint``

benjeffery:
Fixes #1074

This didn't quite seem to warrant a changelog, happy to add one though if anyone thinks otherwise.


jeromekelleher (approved):
LGTM!

PR #1081, opened 2020-12-10
allele remapper for write_vcf

petrelharp:
To be able to write valid VCF files, sometimes we will need to be able to change the ancestral/derived states into something else upon output. (The VCF standard only allows `ACGTN.` characters!) Here's some basic plumbing to make this sort of thing easy... and, I'm using it to do just that over in this pyslim PR: https://github.com/tskit-dev/pyslim/pull/115

In pyslim, I'm using it in two ways:
1. replace the ancestral state with a value pulled from the reference sequence, and derived states with nucleotides pulled from mutation metadata
2. replace arbitrary alleles with a unique string of ACGT.

There's one other small change here: previously, no ALT alleles would output an empty string; here I've changed this to output a `.`.

Also note that this allows outputting VCFs when there's more than 9 alleles per site using the existing (fast!) code, if you remap them to 9 or fewer alleles.

If it looks like a good approach, I can add some tests, and possibly move over the generic approach (2) as an option, if we want.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -4670,6 +4670,8 @@ class TreeSequence:
         individuals=None,
         individual_names=None,
         position_transform=None,
+        nucleotide_alleles=False,
+        allele_mapper=None,
     ):
         """
         Writes a VCF formatted file to the specified file-like object.
@@ -4726,6 +4728,8 @@ class TreeSequence:
         check that the alleles result in a valid VCF---for example, it is possible
         to use the tab character as an allele, leading to a broken VCF.
 
+        Empty alleles (i.e., alleles that are "") are written as ".".
+
         The ``position_transform`` argument provides a way to flexibly translate
         the genomic location of sites in tskit to the appropriate value in VCF.
         There are two fundamental differences in the way that tskit and VCF define
@@ -4739,6 +4743,19 @@ class TreeSequence:
         tskit uses 0-based. However, how coordinates are transformed depends
         on the VCF parser, and so we do **not** account for this change in
         coordinate system by default.
+
+        The ``allele_mapper`` argument allows the user to remap alleles in the
+        tree sequence (e.g., from integers to nucleotides). The value of
+        ``allele_mapper`` should be a function that takes a `.Variant` object
+        and returns a tuple of ``alleles, genotype_map``. The first should be
+        a list of unique alleles that will be output to the VCF file:
+        ``alleles[0]`` will be the REF allele and the remainder will be the ALT
+        alleles. The second, ``genotype_map``, is a numpy array of length equal
+        to the length of the number of alleles of the Variant object, and with
+        values that index ``alleles``.  This maps genotypes in the tree
+        sequence to genotypes in the output VCF. For Variant argument ``v``, if
+        ``v.alleles[j]`` should be mapped to ``alleles[k]`` in the output VCF,
+        then ``genotype_map[j]`` should be equal to ``k``.
 
         Example usage:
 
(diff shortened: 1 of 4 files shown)


petrelharp:
A question here is: should we keep the old behavior of outputting the raw alleles to the VCF as optional (or as default)? The "alleles" could include things like "23,5", which would lead to wrong VCF files (e.g., if the REF alleles are "23,5" and "2", then they'd get concatenated in the REF column to "23,5,2", and someone with genotype 2 *should* have allele "2" but would instead get shown as having allele "5"). What do you think, @attrna?

jeromekelleher:
Seems like a pretty sensible approach, I don't see what else we can do tbh. It would be nice to output valid VCF, but I think actually checking for that could end up being quite convoluted and expensive. I don't really know what's best.

petrelharp:
The motivation here is that I think generally improving our interoperability could help a lot with uptake, and lack of ACGT output is apparently a barrier for using a bunch of standard tools. (also, empty ALT fields)

And, how about this proposal: we add a `nucleotide_alleles` option to `write_vcf` that, if True, translates the alleles to a string of nucleotides, by mapping them sequentially to their base-4 encoding, replacing digits with ACGT. (we'd still only support up to 9 alleles per site, though) I guess that it should default to False, to preserve backwards compatibility. I've got this mostly worked out over in  tskit-dev/pyslim#115 - so, I'd just move that over here.

petrelharp:
I've added tests and implemented `nucleotide_alleles`. If things pass, it's ready for someone else to have a look at.

petrelharp:
Whoops - my mistake here. Since the library we're using for validation fails on empty alleles (e.g. state = `""`), I was writing these out as `"."` (in all cases, not just when using the allele remapper), but that means "missing". I'll fix this up, and let it keep writing out empty alleles (but ask the testing software to not produce any such alleles).  (But, note that we do want to write out a `.` allele in the case that there are *not* alternate alleles.)

petrelharp:
Any chance you could give this a look, @attrna? Ping me if you need to know how to install the code from this PR. Passing `nucleotide_alleles=True` to `ts.write_vcf( )` should now give you (more) valid VCF files, with all ACTG alleles (and no empty strings).

petrelharp:
@attrna reports no problems with these VCF files.

jeromekelleher:
Oops, sorry for the delay here @petrelharp, I'm a bad maintainer! Looking now.

petrelharp:
you've been working on more important things... =)

jeromekelleher:
Nothing is more important than VCF, surely!

python/tskit/trees.py
```diff
@@ -4740,6 +4744,19 @@ def write_vcf(
...
         coordinate system by default.
 
+        The ``allele_mapper`` argument allows the user to remap alleles in the
+        tree sequence (e.g., from integers to nucleotides). The value of
+        ``allele_mapper`` should be a function that takes a `.Variant` object
+        and returns a tuple of ``alleles, genotype_map``. The first should be
+        a list of unique alleles that will be output to the VCF file:
+        ``alleles[0]`` will be the REF allele and the remainder will be the ALT
+        alleles. The second, ``genotype_map``, is a numpy array of length equal
+        to the length of the number of alleles of the Variant object, and with
```
jeromekelleher, line 4754: delete "the length of", or maybe clearer as "is a numpy array of length equal to ``len(variant.alleles)``"

python/tskit/trees.py
```diff
@@ -4809,14 +4826,49 @@ def write_vcf(
             pre 0.2.0 legacy behaviour of rounding values to the nearest integer
             (starting from 1) and avoiding the output of identical positions
             by incrementing is used.
-        """
+        :param nucleotide_alleles: If True, remaps existing alleles to
```
jeromekelleher, line 4829: I'm not super keen on this ``nucleotide_alleles`` option. I can see it having unexpected consequences down the line...

jeromekelleher (commented):
I'm pretty luke-warm on this tbh. Really, this is just an artefact of the mismatch between SLiM's internal stacked mutation model and the tskit data model. They just fundamentally don't match, and while most things sort-of work in the current setup, there will be probably be other things down the line where we have to do some sort of nucleotide remapping to compensate. For example, haplotype matching won't work, I think, and would need to be mapped back to ACGT space.

So, wouldn't it be better to have a conversion utility in pyslim that takes a SLiM tree sequence and "nucleotidifies" it? That way the decisions about what's a sensible way of converting the SLiM mutations into nucleotides can be done withing the context of pyslim, which understands the complexities and can hide them from the user, as needed. The conversion need not be lossy, since you could stash the full context of the site/mutations in the metadata. Otherwise, we'll end up distributing this nucleotification logic all over tskit, the documentation, etc.

Msprime will be outputting nucleotides by default, soon, so I don't think that's a motivation for adding a feature like this here.

petrelharp:
Well, here's my motivation: I implemented this here, not just in pyslim, because msprime also produces non-nucleotide alleles, and so this is giving us a way of outputting legal VCF from an msprime simulation produced with InfiniteAlleles, or a codon model, for instance. It is by default doing a lossless, one-to-one translation, and is only affecting VCF output. It might feel fairly ad hoc, but it works, it's lossless, it's hard to know what else we could do by default, and it gives people a fairly simple route to doing something else if that makes more sense for their particular situation.

python/tskit/trees.py
petrelharp: I think we really want to make it possible for people to output legal VCFs, though. I'm not sure how else we'd do that.

jeromekelleher:
That's fair enough, but I think the operation would be better expressed as a transformation of the **tree sequence** rather than something we do just for VCF output. So, if we try to write out a tree sequence as VCF that doesn't use ACGT we can raise a warning and suggest that the user "nucleotidify" it. There will be other situations in which we'll want to transform non ACGT alleles into ACGT somehow, so let's make a method for it.

We'll still be able to get into situations where we generate invalid VCF, for example if we have adjacent sites that have lots of alleles and we end up mapping them to AAAAAA or whatever.

benjeffery:
+1 for separating this from VCF output.

petrelharp:
Update: this issue is not unique to VCF output, so we should deal with this higher-up, and arguably by remapping alleles directly. I'm worried this loses information, but need to think through it more. Anyhow, we're not going to open this can of worms here.

benjeffery:
Is there another issue open for tracking this functionality? Had a quick look and didn't see it.

petrelharp:
Nope. I don't have it clear enough to write an issue, currently.

petrelharp:
I've implemented allele remapping in pyslim here: https://github.com/tskit-dev/pyslim/pull/174

PR #1089, opened 2020-12-14
Report back a recommendation for epsilon

### LINKED ISSUE #1088: Provide feedback or adaptive epsilon for split polytomies?
We deliberately removed both the feedback about what epsilon value to used, and the ability to calculate a sensible epsilon value in split_polytomies, but that's now wasted a day of compute time because my simulations all aborted, firstly with the default, and again when I set epsilon to 1e-20, both with the somewhat unhelpful error message:

```
_tskit.LibraryError: ('time[parent] must be greater than time[child]', 'Epsilon=1e-20 not small enough to create new nodes below a polytomy, due to the time of one of child nodes being too close. ')
```
since I need to split polytomies before saving results. I'm now trying with 1e-40, but that might not work either (I guess this is because of trying to split polytomies in nodes produced via tsinfer path compression, which are placed by [some small epsilon](https://github.com/tskit-dev/tsinfer/blob/558172c617153f3057e99d2b03556fe28e4095c7/lib/tree_sequence_builder.c#L35) above their child nodes). So having an adaptive epsilon seems like a good plan here. Or if not that, at least suggesting what a possible working value of epsilon should be. I previously made some suggestions how to do this [here](https://github.com/tskit-dev/tskit/pull/815#issuecomment-732206285).

hyanwong:
## Description

Report back a recommended value if epsilon is too high when splitting polytomies. This might only be a sensible thing to do if the random_seed is fixed, otherwise the recommended value might need to be lowered again. Still a useful thing to report, though, I feel.

Fixes #1088 

# PR Checklist:

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -1085,9 +1085,6 @@ class TestPolytomySplitting:
     7       0.8      1.0      6       1,2,3,4
     """
 
-    def tree_polytomy_4(self):
-        return tskit.Tree.generate_star(4)
-
     def ts_polytomy_44344(self):
         return tskit.load_text(
             nodes=io.StringIO(self.nodes_polytomy_44344),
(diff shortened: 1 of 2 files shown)


jeromekelleher:
There's a few problems with this, unfortunately @hyanwong. I'll work up a counter proposal using ``nextafter`` as the default eps strategy.

PR #1090, opened 2020-12-14
Change default in split_polytomies to use nextafter.

jeromekelleher:
Generate the shortest possible branches when splitting polytomies, by default. This will be much more robust to different scales than the old fixed-size default, and, think, is about as good as you can do, ever.

I've also put some effort into making good error messages to allow people debug what happens when this goes wrong, which is extremely fiddly. Having worked with this a bit now and pushed it to it's limits, I don't really see a good reason for ever using the fixed epsilon, unless that's really what you want to do and you intend to put multiple mutations on the branches.

Supersedes #1089

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,4 +1,14 @@
 --------------------
+[0.3.5] - 2021-XX-XX
+--------------------
+
+**Features**
+
+- Change the default behaviour of Tree.split_polytomies to generate
+  the shortest possible branch lengths instead of a fixed epsilon of
+  1e-10. (:user:`jeromekelleher`, :issue:`1089`, :pr:`1090`)
+
+--------------------
 [0.3.4] - 2020-12-02
 --------------------
 
--- a/python/tskit/combinatorics.py
+++ b/python/tskit/combinatorics.py
@@ -256,8 +256,6 @@ def split_polytomies(
     method, which is the usual route through which this function is called.
     """
     allowed_methods = ["random"]
-    if epsilon is None:
-        epsilon = 1e-10
     if method is None:
         method = "random"
     if method not in allowed_methods:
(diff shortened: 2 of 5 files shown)


python/tskit/combinatorics.py
```diff
@@ -272,15 +270,55 @@ def split_polytomies(
...
-            time = tree.time(u) - epsilon
-            stack = [(child, time) for child in root.children]
+            root_time = tree.time(u)
+            stack = [(child, root_time) for child in root.children]
             while len(stack) > 0:
-                node, time = stack.pop()
+                node, parent_time = stack.pop()
                 if node.label is None:
-                    node.label = tables.nodes.add_row(time=time)
+                    if epsilon is None:
```
hyanwong, line 278: I do think you are correct to retain epsilon, even if it is not usually used, because we may wish later to insert nodes between the ones we've just created. I can't actually see why that would be needed for `split_polytomies`, but it is certainly needed when we create new nodes using path compression. In other words, having the only possibility for epsilon being the minimal representable distance might restrict how the TS could be used down the line, so allowing a tweakable parameter (even if it is not the default) seems justified.

python/tskit/combinatorics.py
```diff
@@ -272,15 +270,55 @@ def split_polytomies(
...
+                    node.label = tables.nodes.add_row(time=child_time)
+                else:
+                    assert len(node.children) == 0
+                    # This is a leaf node connecting back into the original tree
+                    child_time = tree.time(node.label)
+                if parent_time <= child_time:
+                    u = root.label
+                    min_child_time = min(tree.time(v) for v in tree.children(u))
+                    min_time = root_time - min_child_time
+                    message = (
```
hyanwong, line 291: Thanks for adding these messages in. This is really helpful. It would be even nicer for the user if we saved the "worst" message (ie. the example that would require the smallest epsilon to be set) until we have passed through all the nodes, and then output just that one after the loop (in the hope that correcting that one if possible would make the routine run properly next time). But I think this might be tricky given all the possible paths, so it's probably not worth it.

python/tskit/combinatorics.py
```diff
@@ -272,15 +270,55 @@ def split_polytomies(
...
+                        f"of {min_time} to the resolved leaves."
+                    )
+                    if epsilon is None:
+                        message += (
+                            " The time difference between nodes is so small that "
+                            "more nodes cannot be inserted between within the limits "
+                            "of floating point precision."
+                        )
+                    else:
+                        # We can also have parent_time == child_time if epsilon
```
hyanwong, line 303: I'm not sure I understand this comment. If epsilon === branch_length and both are >0, then parent_time cannot be equal to child_time? Or to put it another way, if epsilon is too large, isn't parent_time > child_time guaranteed?

python/tskit/trees.py
```diff
@@ -2398,8 +2398,25 @@ def split_polytomies(
...
+        Times are allocated to different levels of the tree, such that any newly
+        inserted sibling nodes will have the same time.
+
+        By default, the times of the newly generated children of a particular
+        node are the minimum representable distance in floating point arithmetic
+        from their parents (using the `nextafter
+        <https://numpy.org/doc/stable/reference/generated/numpy.nextafter.html>`_
+        function). Thus, the generated branches have the shortest possible nonzero
+        length. A fixed branch length between inserted nodes and their parents
+        can also be specified by using the ``epsilon`` parameter.
```
hyanwong, line 2411: Nicely described.

hyanwong (approved):
This all looks great @jeromekelleher - thanks. A few comments and something I don't understand, but I think it can be merged as-is.

python/tskit/combinatorics.py
jeromekelleher: That would be tricky to do and test - I don't think it's worth the effort tbh. Already we have twice as much code for making error messages as actually doing the work.

python/tskit/combinatorics.py
jeromekelleher: The issue happens when we *exactly* use up all the available branch length in the original tree, and so there's nothing left for the last branch in the resolved bit. I've updated the comment, thanks.

jeromekelleher (commented):
Thanks for the review @hyanwong.

jeromekelleher:
@benjeffery, I think this is ready to go - can you have a quick look over please?

python/CHANGELOG.rst
```diff
@@ -1,3 +1,13 @@
+--------------------
```
benjeffery, line 1: Sorry, seems I forgot the post-release for 0.3.4!

benjeffery (approved):
LGTM, lots of tricky stuff nicely dealt with!

python/CHANGELOG.rst
jeromekelleher: No worries - I had mucked up your workflow by setting the release number as part of a PR!

Issue #1073, opened 2020-12-03
Metadata decoding should use default value from schema


jeromekelleher:
The following schema defines a single property which has a default of 0:

```python
import tskit

schema = tskit.MetadataSchema({
    "codec": "json",
    "type": "object",
    "properties": {
        "something_with_default": {
            "type": "number",
            "default": 0,
        },
     },
    "required": [],
    "additionalProperties": True,
    }
)

tables = tskit.TableCollection(1)
tables.populations.metadata_schema = schema
tables.populations.add_row(metadata={"something_with_default": 1})
tables.populations.add_row(metadata={})

print(tables.populations)

ts = tables.tree_sequence()
print(list(ts.populations()))
```

The result is:

```
id      metadata
0       eyJzb21ldGhpbmdfd2l0aF9kZWZhdWx0IjoxfQ==
1       e30=
[{'id': 0, 'metadata': {'something_with_default': 1}}, {'id': 1, 'metadata': {}}]
```
So, we're not encoding the default value in the table (as expected) but we're also not returning it in the decoded dictionary. I would have expected the decoded result to be:
```
[{'id': 0, 'metadata': {'something_with_default': 1}}, {'id': 1, 'metadata': {'something_with_default': 0}}]
```
The ``default`` keyword is defined in [JSON schema](https://json-schema.org/understanding-json-schema/reference/generic.html?highlight=default#id2) but validators are not obliged to do anything with it. However, we could use the value during decoding easily enough in decoding, I would have thought?

This would be very useful for saving space in JSON metadata where we have a key that's usually some default value and in some exceptional case it's not.

benjeffery:
So I think this is best done by pre-indexing the keys with default values, then checking after normal decoding. Otherwise you have to do another JSON traversal.

jeromekelleher:
Yep, that's a nice way of doing it.

jeromekelleher:
I guess it makes no sense currently in struct, and it should raise an error if we try to set a default?

benjeffery:
Yep, can't have missing keys. I think an error is ok as you have to do quite a bit of work to go from JSON to `struct` so removing the defaults isn't much more.

jeromekelleher:
OK, makes sense. Let's set the struct behaviour (raise an error) at the same time as we implement the JSON behaviour then.

jeromekelleher:
[...]

PR #1091, opened 2020-12-15
Make default metadata schema dependant

### LINKED ISSUE #1084: Metadata should default to {} on Table.add_row if schema is set
Quite a bit of code breaks when we set a metadata schema because there isn't a useful default for ``metadata`` in ``add_row``, e.g.:
```python
schema = tskit.MetadataSchema({
    "codec": "json",
    "type": "object",
    "properties": {},
    "required": [],
    "additionalProperties": True,
    }
)

tables.populations.metadata_schema = schema
tables.populations.add_row(metadata={}) # Fine
tables.populations.add_row() # breaks
```

It seems reasonable to add the empty metadata as the default here, when a schema is defined.

benjeffery:
Fixes #1084

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v3.4.0
     hooks:
       - id: check-merge-conflict
       - id: debug-statements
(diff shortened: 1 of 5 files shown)


jeromekelleher:
LGTM @benjeffery. I'm slightly worried that this ``schema.default_value`` will be confusing when we start thinking about setting defaults, as discussed in #1073. Is it worth looking at this at the same time, since they are closely related?

benjeffery:
My thinking here is that it is not `add_row`'s concern to know what the default value for metadata should be. I think the schema's default metadata value will still be `{}` even after we add `default` functionality for keys. 

Maybe if I rename this attribute to `empty_value` instead?

jeromekelleher:
> Maybe if I rename this attribute to empty_value instead?

yes, good call.

benjeffery:
Ok, I've changed the attribute name. Should be good to go.

petrelharp:
LGTM - but perhaps there should be a note in the documentation about metadata schemas explaining that {} will be passed in if no metadata are provided, so this should be legal for the schema if it's desired to be able to do that.

jeromekelleher (approved):

jeromekelleher:
Yes, we should have a discussion of default handling. I guess we could add a section to the docs in this PR explaining this much, and then add to it when we have implemented #1073

benjeffery:
I've added a small note to the docs.

jeromekelleher:
I've had a look through the diff, but I'm not seeing any changes to the docs @benjeffery - maybe you forgot to push?

benjeffery:
Should be there now. I only added a note to the `add_row` docstring - we can add a full section about default handling after #1073.

benjeffery:
A pre-commit update is also along for the ride as its cache got in an inconsistent state.

PR #1108, opened 2021-01-02
updates to union, subset, and sort

### LINKED ISSUE #705: make sort stricter

### LINKED ISSUE #1076: Add ``assert_equal`` to TableCollection and TreeSequence
The '.equals`` method is very useful, but it doesn't tell us where the problem is when tables are not equal. We should add methods ``assert_equal`` which would initially look like 
```python
def assert_equal(self, other, ignore_provenance=...) # same signature as equals
    assert self.sequence_length == other.sequence_length
    assert self.nodes == other.nodes
    # etc
```
This would at least tell us which table the mismatch occurs in. We can imagine providing a more detailed picture of how tables differ at some point, but this could be added later as a refinement. This form ``assert_equal`` should integrate nicely with pytest, and clean up our testing code some more.

Seems like a straightforward win?

### LINKED ISSUE #1095: tidy up union
A few things:

- there's a note that [says we can remove the restriction](https://github.com/tskit-dev/tskit/blob/c9b2c920b6c4e1aaebcbc91dc4c6e13e919bdd08/c/tskit/tables.c#L9400) that the non-shared bits of other not be above the shared bit. I think we can do this now, since we're sorting on mutation time.
- I don't think that we're checking whether the same value appears more than once in `node_mapping`, and I suspect this could make bad things happen - but, maybe it's ok?

petrelharp:
There's a few things in this PR:

- implements canonical sort (closes #705 and incorporates #715)
- removes the restriction on shared history for `.union` (closes #1095)
- extends subset to work on completely unsorted tables
- added `shuffle_tables` and `assert_table_collection_equals` methods to `tests/tsutil.py`
- added more tests for subset, union, and sort generally

Notes:

*Canonical sort*:  This is an option, e.g. `tables.sort(canonical=True)` and `tsk_table_collection_sort(tables, TSK_SORT_CANONICAL)`. In addition to providing a method to canonicalize tables, this will ensure that mutation parents come before their children, something we didn't previously have.  This first runs `subset` to do the reordering of individuals and populations. It turned out our original proposal to sort mutations canonically (sort by mutation.parent) was just wrong, so instead we are sorting by number of "mutation descendants". (This is necessary when mutation times are not unique.) I was tempted to add this to the standard `.sort()`, but given that sort is a bottleneck in practice and we have no use case which produces tables with mutations out of parent-sorted order, I did not.

*keep_unused options*: To allow `subset` to be used to reorder tables and not discard information, I had to add `keep_unused_X` arguments to it, for `X` in `sites`, `individuals`, and `populations`. This seems fine, although I'm tempted to (a) either just have a single `keep_unused` argument that does all three, or (b) remove these from the python subset and write a separate python method called `reorder_ndoes( )` that's the python front-end.

*Unchanged populations:* Something I ran across in the process is that it's annoying to keep populations consistent. I added `unchanged_populations` arguments (and a corresponding `TSK_UNCHANGED_POPULATIONS` flag) to both `.subset( )` and `.sort( )` that don't reorder populations. However, we should probably write `reorder_populations` (and `reorder_individuals`) methods to make this sort of thing easier to deal with.

*assert table collections equal:* I thought that this would close #1076, but now I wonder if that was meant to be a full method (including in the C library), rather than just in `tsutil`?

*deduplicate_sites*: turns out this method, with the addition of mutation times, needs a sort before *and after* it, strictly speaking. 

This clearly needs a squash, but I think (hope?) everything works and is (possibly) good to go.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4961,13 +4961,13 @@ out:
 }
 
 int TSK_WARN_UNUSED
-tsk_table_collection_canonicalise(
-    tsk_table_collection_t *self, tsk_flags_t TSK_UNUSED(options))
+tsk_table_collection_canonicalise(tsk_table_collection_t *self, tsk_flags_t options)
 {
     int ret = 0;
     tsk_id_t k;
     tsk_id_t *nodes = NULL;
     tsk_table_sorter_t sorter;
+    tsk_flags_t subset_options = options & TSK_KEEP_UNREFERENCED;
 
     ret = tsk_table_sorter_init(&sorter, self, 0);
     if (ret != 0) {
@@ -4982,8 +4982,7 @@ tsk_table_collection_canonicalise(
     for (k = 0; k < (tsk_id_t) self->nodes.num_rows; k++) {
         nodes[k] = k;
     }
-    ret = tsk_table_collection_subset(
-        self, nodes, self->nodes.num_rows, TSK_KEEP_UNREFERENCED);
+    ret = tsk_table_collection_subset(self, nodes, self->nodes.num_rows, subset_options);
     if (ret != 0) {
         goto out;
     }
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -9511,8 +9511,8 @@ tsk_table_collection_subset(tsk_table_collection_t *self, const tsk_id_t *nodes,
             if (individual_map[k] == TSK_NULL) {
                 tsk_individual_table_get_row_unsafe(&tables.individuals, k, &ind);
                 ret = tsk_individual_table_add_row(&self->individuals, ind.flags,
-                    ind.location, ind.location_length, ind.metadata,
-                    ind.metadata_length);
+                    ind.location, ind.location_length, ind.parents, ind.parents_length,
+                    ind.metadata, ind.metadata_length);
                 if (ret < 0) {
                     goto out;
                 }
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4804,7 +4804,7 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
         if (ret != 0) {
             goto out;
         }
-        ret = tsk_table_sorter_sort_mutations(self);
+        ret = self->sort_mutations(self);
         if (ret != 0) {
             goto out;
         }
@@ -4838,8 +4838,9 @@ tsk_table_sorter_init(
         goto out;
     }
 
-    /* Set the sort_edges method to the default. */
+    /* Set the sort_edges and sort_mutations methods to the default. */
     self->sort_edges = tsk_table_sorter_sort_edges;
+    self->sort_mutations = tsk_table_sorter_sort_mutations;
 out:
     return ret;
 }
@@ -4849,6 +4850,153 @@ tsk_table_sorter_free(tsk_table_sorter_t *self)
 {
     tsk_safe_free(self->site_id_map);
     return 0;
+}
+
+/******************
+ * Canonical sort
+ ******************/
+
+typedef struct {
+    tsk_mutation_t mut;
+    int num_descendants;
+} mutation_canonical_sort_t;
+
+static int
+cmp_mutation_canonical(const void *a, const void *b)
+{
+    const mutation_canonical_sort_t *ia = (const mutation_canonical_sort_t *) a;
+    const mutation_canonical_sort_t *ib = (const mutation_canonical_sort_t *) b;
+    /* Compare mutations by site */
+    int ret = (ia->mut.site > ib->mut.site) - (ia->mut.site < ib->mut.site);
+    if (ret == 0 && !tsk_is_unknown_time(ia->mut.time)
+        && !tsk_is_unknown_time(ib->mut.time)) {
+        ret = (ia->mut.time < ib->mut.time) - (ia->mut.time > ib->mut.time);
+    }
+    if (ret == 0) {
+        ret = (ia->num_descendants < ib->num_descendants)
+              - (ia->num_descendants > ib->num_descendants);
+    }
+    if (ret == 0) {
+        ret = (ia->mut.node > ib->mut.node) - (ia->mut.node < ib->mut.node);
+    }
+    if (ret == 0) {
+        ret = (ia->mut.id > ib->mut.id) - (ia->mut.id < ib->mut.id);
+    }
+    return ret;
+}
+
+static int
+tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
+{
+    int ret = 0;
+    tsk_size_t j;
+    tsk_id_t parent, mapped_parent, p;
+    tsk_mutation_table_t *mutations = &self->tables->mutations;
+    tsk_size_t num_mutations = mutations->num_rows;
+    tsk_mutation_table_t copy;
+    mutation_canonical_sort_t *sorted_mutations
+        = malloc(num_mutations * sizeof(*sorted_mutations));
+    tsk_id_t *mutation_id_map = malloc(num_mutations * sizeof(*mutation_id_map));
+
+    ret = tsk_mutation_table_copy(mutations, &copy, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    if (mutation_id_map == NULL || sorted_mutations == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    /* compute numbers of descendants for each mutation */
+    for (j = 0; j < num_mutations; j++) {
+        sorted_mutations[j].num_descendants = 0;
+    }
+    for (j = 0; j < num_mutations; j++) {
+        p = mutations->parent[j];
+        while (p != TSK_NULL) {
+            sorted_mutations[p].num_descendants += 1;
+            p = mutations->parent[p];
+        }
+    }
+
+    for (j = 0; j < num_mutations; j++) {
+        tsk_mutation_table_get_row_unsafe(&copy, (tsk_id_t) j, &sorted_mutations[j].mut);
+        sorted_mutations[j].mut.site = self->site_id_map[sorted_mutations[j].mut.site];
+    }
+    ret = tsk_mutation_table_clear(mutations);
+    if (ret != 0) {
+        goto out;
+    }
+
+    qsort(sorted_mutations, num_mutations, sizeof(*sorted_mutations),
+        cmp_mutation_canonical);
+
+    /* Make a first pass through the sorted mutations to build the ID map. */
+    for (j = 0; j < num_mutations; j++) {
+        mutation_id_map[sorted_mutations[j].mut.id] = (tsk_id_t) j;
+    }
+
+    for (j = 0; j < num_mutations; j++) {
+        mapped_parent = TSK_NULL;
+        parent = sorted_mutations[j].mut.parent;
+        if (parent != TSK_NULL) {
+            mapped_parent = mutation_id_map[parent];
+        }
+        ret = tsk_mutation_table_add_row(mutations, sorted_mutations[j].mut.site,
+            sorted_mutations[j].mut.node, mapped_parent, sorted_mutations[j].mut.time,
+            sorted_mutations[j].mut.derived_state,
+            sorted_mutations[j].mut.derived_state_length,
+            sorted_mutations[j].mut.metadata, sorted_mutations[j].mut.metadata_length);
+        if (ret < 0) {
+            goto out;
+        }
+    }
+    ret = 0;
+
+out:
+    tsk_safe_free(mutation_id_map);
+    tsk_safe_free(sorted_mutations);
+    tsk_mutation_table_free(&copy);
+    return ret;
+}
+
+int TSK_WARN_UNUSED
+tsk_table_collection_canonicalise(
+    tsk_table_collection_t *self, tsk_flags_t TSK_UNUSED(options))
+{
+    int ret = 0;
+    tsk_id_t k;
+    tsk_id_t *nodes = NULL;
+    tsk_table_sorter_t sorter;
+
+    ret = tsk_table_sorter_init(&sorter, self, 0);
+    if (ret != 0) {
+        goto out;
+    }
+
+    nodes = malloc(self->nodes.num_rows * sizeof(*nodes));
+    if (nodes == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    for (k = 0; k < (tsk_id_t) self->nodes.num_rows; k++) {
+        nodes[k] = k;
+    }
+    ret = tsk_table_collection_subset(
+        self, nodes, self->nodes.num_rows, TSK_CANONICALISE);
+    if (ret != 0) {
+        goto out;
+    }
+    sorter.sort_mutations = tsk_table_sorter_sort_mutations_canonical;
+
+    ret = tsk_table_sorter_run(&sorter, NULL);
+    if (ret != 0) {
+        goto out;
+    }
+out:
+    tsk_safe_free(nodes);
+    tsk_table_sorter_free(&sorter);
+    return ret;
 }
 
 /*************************
@@ -8781,6 +8929,7 @@ tsk_table_collection_sort(
     if (ret != 0) {
         goto out;
     }
+
     ret = tsk_table_sorter_run(&sorter, start);
     if (ret != 0) {
         goto out;
@@ -9291,20 +9440,26 @@ out:
 }
 
 int TSK_WARN_UNUSED
-tsk_table_collection_subset(
-    tsk_table_collection_t *self, const tsk_id_t *nodes, tsk_size_t num_nodes)
+tsk_table_collection_subset(tsk_table_collection_t *self, const tsk_id_t *nodes,
+    tsk_size_t num_nodes, tsk_flags_t options)
 {
     int ret = 0;
-    tsk_id_t k, i, new_parent, new_child, new_node;
+    tsk_id_t j, k, new_parent, new_child, new_node, site_id;
     tsk_edge_t edge;
-    tsk_mutation_t mut;
-    tsk_site_t site;
     tsk_id_t *node_map = NULL;
     tsk_id_t *individual_map = NULL;
     tsk_id_t *population_map = NULL;
     tsk_id_t *site_map = NULL;
     tsk_id_t *mutation_map = NULL;
     tsk_table_collection_t tables;
+    tsk_individual_t ind;
+    tsk_population_t pop;
+    tsk_site_t site;
+    tsk_mutation_t mut;
+    bool canonicalise = !!(options & TSK_CANONICALISE);
+    bool no_filter_populations = !(options & TSK_FILTER_POPULATIONS);
+    bool no_filter_individuals = !(options & TSK_FILTER_INDIVIDUALS);
+    bool no_filter_sites = !(options & TSK_FILTER_SITES);
 
     ret = tsk_table_collection_copy(self, &tables, 0);
     if (ret != 0) {
@@ -9335,12 +9490,57 @@ tsk_table_collection_subset(
     memset(site_map, 0xff, tables.sites.num_rows * sizeof(*site_map));
     memset(mutation_map, 0xff, tables.mutations.num_rows * sizeof(*mutation_map));
 
+    if (no_filter_individuals && !canonicalise) {
+        ret = tsk_individual_table_copy(
+            &tables.individuals, &self->individuals, TSK_NO_INIT);
+        if (ret < 0) {
+            goto out;
+        }
+        for (k = 0; k < (tsk_id_t) tables.individuals.num_rows; k++) {
+            individual_map[k] = k;
+        }
+    }
+    if (no_filter_populations && !canonicalise) {
+        ret = tsk_population_table_copy(
+            &tables.populations, &self->populations, TSK_NO_INIT);
+        if (ret < 0) {
+            goto out;
+        }
+        for (k = 0; k < (tsk_id_t) tables.populations.num_rows; k++) {
+            population_map[k] = k;
+        }
+    }
+
     // nodes, individuals, populations
     for (k = 0; k < (tsk_id_t) num_nodes; k++) {
         ret = tsk_table_collection_add_and_remap_node(
             self, &tables, nodes[k], individual_map, population_map, node_map, true);
         if (ret < 0) {
             goto out;
+        }
+    }
+    if (canonicalise) {
+        // keep unused individuals and populations
+        for (k = 0; k < (tsk_id_t) tables.individuals.num_rows; k++) {
+            if (individual_map[k] == TSK_NULL) {
+                tsk_individual_table_get_row_unsafe(&tables.individuals, k, &ind);
+                ret = tsk_individual_table_add_row(&self->individuals, ind.flags,
+                    ind.location, ind.location_length, ind.metadata,
+                    ind.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+            }
+        }
+        for (k = 0; k < (tsk_id_t) tables.populations.num_rows; k++) {
+            if (population_map[k] == TSK_NULL) {
+                tsk_population_table_get_row_unsafe(&tables.populations, k, &pop);
+                ret = tsk_population_table_add_row(
+                    &self->populations, pop.metadata, pop.metadata_length);
+                if (ret < 0) {
+                    goto out;
+                }
+            }
         }
     }
 
@@ -9359,36 +9559,53 @@ tsk_table_collection_subset(
     }
 
     // mutations and sites
-    i = 0;
-    for (k = 0; k < (tsk_id_t) tables.sites.num_rows; k++) {
-        tsk_site_table_get_row_unsafe(&tables.sites, k, &site);
-        while ((i < (tsk_id_t) tables.mutations.num_rows)
-               && (tables.mutations.site[i] == site.id)) {
-            tsk_mutation_table_get_row_unsafe(&tables.mutations, i, &mut);
-            new_node = node_map[mut.node];
-            if (new_node != TSK_NULL) {
-                if (site_map[site.id] == TSK_NULL) {
-                    ret = tsk_site_table_add_row(&self->sites, site.position,
-                        site.ancestral_state, site.ancestral_state_length, site.metadata,
-                        site.metadata_length);
-                    if (ret < 0) {
-                        goto out;
-                    }
-                    site_map[site.id] = ret;
-                }
-                new_parent = TSK_NULL;
-                if (mut.parent != TSK_NULL) {
-                    new_parent = mutation_map[mut.parent];
-                }
-                ret = tsk_mutation_table_add_row(&self->mutations, site_map[site.id],
-                    new_node, new_parent, mut.time, mut.derived_state,
-                    mut.derived_state_length, mut.metadata, mut.metadata_length);
-                if (ret < 0) {
-                    goto out;
-                }
-                mutation_map[mut.id] = ret;
+    // make a first pass through to build the mutation_map so that
+    // mutation parent can be remapped even if the table is not in order
+    j = 0;
+    for (k = 0; k < (tsk_id_t) tables.mutations.num_rows; k++) {
+        if (node_map[tables.mutations.node[k]] != TSK_NULL) {
+            mutation_map[k] = j;
+            j++;
+            site_id = tables.mutations.site[k];
+            if (site_map[site_id] == TSK_NULL) {
+                // insert a sentinel non-NULL value:
+                site_map[site_id] = 1;
             }
-            i++;
+        }
+    }
+    // keep retained sites in their original order
+    j = 0;
+    for (k = 0; k < (tsk_id_t) tables.sites.num_rows; k++) {
+        if (canonicalise || no_filter_sites || site_map[k] != TSK_NULL) {
+            tsk_site_table_get_row_unsafe(&tables.sites, k, &site);
+            ret = tsk_site_table_add_row(&self->sites, site.position,
+                site.ancestral_state, site.ancestral_state_length, site.metadata,
+                site.metadata_length);
+            if (ret < 0) {
+                goto out;
+            }
+            site_map[k] = j;
+            j++;
+        }
+    }
+    for (k = 0; k < (tsk_id_t) tables.mutations.num_rows; k++) {
+        tsk_mutation_table_get_row_unsafe(&tables.mutations, k, &mut);
+        new_node = node_map[mut.node];
+        if (new_node != TSK_NULL) {
+            new_parent = TSK_NULL;
+            if (mut.parent != TSK_NULL) {
+                new_parent = mutation_map[mut.parent];
+            }
+            ret = tsk_mutation_table_add_row(&self->mutations, site_map[mut.site],
+                new_node, new_parent, mut.time, mut.derived_state,
+                mut.derived_state_length, mut.metadata, mut.metadata_length);
+            if (ret < 0) {
+                goto out;
+            }
+            tsk_bug_assert(mutation_map[mut.id] == ret);
+        }
+        if (ret < 0) {
+            goto out;
         }
     }
 
@@ -9441,7 +9658,6 @@ tsk_check_subset_equality(tsk_table_collection_t *self,
         }
     }
 
-    // TODO: strict sort before checking equality
     ret = tsk_table_collection_copy(self, &self_copy, 0);
     if (ret != 0) {
         goto out;
@@ -9450,11 +9666,21 @@ tsk_check_subset_equality(tsk_table_collection_t *self,
     if (ret != 0) {
         goto out;
     }
-    ret = tsk_table_collection_subset(&self_copy, self_nodes, num_shared_nodes);
+    ret = tsk_table_collection_subset(&self_copy, self_nodes, num_shared_nodes,
+        TSK_FILTER_POPULATIONS | TSK_FILTER_INDIVIDUALS | TSK_FILTER_SITES);
     if (ret != 0) {
         goto out;
     }
-    ret = tsk_table_collection_subset(&other_copy, other_nodes, num_shared_nodes);
+    ret = tsk_table_collection_subset(&other_copy, other_nodes, num_shared_nodes,
+        TSK_FILTER_POPULATIONS | TSK_FILTER_INDIVIDUALS | TSK_FILTER_SITES);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_table_collection_canonicalise(&self_copy, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_table_collection_canonicalise(&other_copy, 0);
     if (ret != 0) {
         goto out;
     }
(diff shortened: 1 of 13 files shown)


c/tskit/tables.c
```diff
@@ -4715,6 +4716,114 @@ tsk_table_sorter_free(tsk_table_sorter_t *self)
...
+        ret = (ia->mut.node > ib->mut.node) - (ia->mut.node < ib->mut.node);
+    }
+    if (ret == 0) {
+        ret = (ia->mut.id > ib->mut.id) - (ia->mut.id < ib->mut.id);
+    }
+    return ret;
+}
+
+static int
+tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
```
petrelharp, line 4753: This has a lot of code overlap with `tsk_table_sorter_sort_mutations`, and we could more cleanly just make `canonical` a flag that affects its behavior.

c/tskit/tables.c
```diff
@@ -8623,12 +8732,34 @@ tsk_table_collection_sort(
...
+        if (nodes == NULL) {
+            ret = TSK_ERR_NO_MEMORY;
+            goto out;
+        }
+        for (k = 0; k < (tsk_id_t) self->nodes.num_rows; k++) {
+            nodes[k] = k;
+        }
+        subset_options
+            |= (TSK_SUBSET_KEEP_UNUSED_POPULATIONS | TSK_SUBSET_KEEP_UNUSED_INDIVIDUALS
+                | TSK_SUBSET_KEEP_UNUSED_SITES);
```
petrelharp, line 8757: I am tempted to make a single macro that is `TSK_SUBSET_KEEP_UNUSED = TSK_SUBSET_KEEP_UNUSED_POPULATIONS | TSK_SUBSET_KEEP_UNUSED_INDIVIDUALS
                | TSK_SUBSET_KEEP_UNUSED_SITES`.

petrelharp: however, I think someone's going to want to use only some of these options at some point, so I vote to keep them separate.

c/tskit/tables.c
```diff
@@ -9206,36 +9385,56 @@ tsk_table_collection_subset(
     }
 
     // mutations and sites
-    i = 0;
+    // make a first pass through to build the mutation_map so that
+    // mutation parent can be remapped even if the table is not in order
```
petrelharp, line 9389: This next bit got more complicated, but `subset` is as a result a lot more robust, not requiring any sort of sorting whatsoever.

c/tskit/tables.h
```diff
@@ -684,6 +686,9 @@ typedef struct {
 /** @brief Do not run integrity checks before performing an operation. */
 #define TSK_NO_CHECK_INTEGRITY (1u << 29)
 
+/** @brief Do not reorder the population table. */
+#define TSK_UNCHANGED_POPULATIONS (1u << 28)
```
petrelharp, line 690: I think this might want to be a common option - and myabe even default, since changing population references could be a big gotcha.

python/tskit/tables.py
```diff
@@ -2648,6 +2668,9 @@ def deduplicate_sites(self):
         duplicate ``position`` (and keeping only the *first* entry for each
         site), and renumbering the ``site`` column of the mutation table
         appropriately.  This requires the site table to be sorted by position.
+
+        This method does not sort the tables afterwards, so mutations may no longer
+        be sorted by time.
```
petrelharp, line 2673: I think we may need to do something about this, but it's a job for a different PR.

petrelharp:
Ok - this is ready for someone to take a look at!

python/tskit/trees.py
```diff
@@ -5106,42 +5106,84 @@ def trim(self, record_provenance=True):
         tables.trim(record_provenance)
         return tables.tree_sequence()
 
-    def subset(self, nodes, record_provenance=True):
+    def subset(
+        self,
+        nodes,
+        record_provenance=True,
+        unchanged_populations=False,
+        keep_unused_populations=False,
```
jeromekelleher, line 5114: It would be good to mirror the options that we have in ``simplify``, which are ``filter_populations=True``, ``filter_individuals=True``, etc.

python/tskit/tables.py
```diff
@@ -2558,7 +2558,7 @@ def map_ancestors(self, *args, **kwargs):
         # A deprecated alias for link_ancestors()
         return self.link_ancestors(*args, **kwargs)
 
-    def sort(self, edge_start=0):
+    def sort(self, edge_start=0, canonical=False, unchanged_populations=False):
```
jeromekelleher, line 2561: I'm not sure this is the right way to do this. Wouldn't it be better to add a separate function ``canonicalise``, which does the sorting *and* optional filtering of unreferenced stuff?  Is there a good reason to not make a separate method? It would make the C implementation clearer I think, as well.

c/tskit/tables.c
```diff
@@ -9190,6 +9338,37 @@ tsk_table_collection_subset(
             goto out;
         }
     }
+    if (keep_individuals) {
+        for (k = 0; k < (tsk_id_t) tables.individuals.num_rows; k++) {
+            if (individual_map[k] == TSK_NULL) {
+                ret = tsk_individual_table_get_row(&tables.individuals, k, &ind);
```
jeromekelleher, line 9344: Can use the ``unsafe`` versions of the the get_row functions here since we're in charge of the ID values.

jeromekelleher (commented):
Overall looks good @petrelharp, although I haven't gone through everything in detail. A couple of high-level points, though.

jeromekelleher:
> assert table collections equal: I thought that this would close #1076, but now I wonder if that was meant to be a full method (including in the C library), rather than just in tsutil?

The idea of #1076 is to add this to the public Python API so that downstream projects like msprime etc can use it in their test cases, and get informative errors when the tables aren't equal. We don't want to do this as the C level specifically because we want to provide good errors. A good approach might be to first try to see if the tables are equal using the low-level C call, and then if they are not, do some table-by-table and maybe row-by-row comparisons in Python to figure out exactly where the tables differ.

jeromekelleher:
> TSK_UNCHANGED_POPULATIONS

Is a good idea and definitely useful, but I'm not sure about the name. It's in a different tense to the rest of the options (or something). The rest of the options like say TSK_FILTER_SITES describe *actions*, but this is describing a desired *state*. Something like ``TSK_NO_REORDER_POPULATIONS`` or something might be more consistent?

c/tskit/tables.h
```diff
@@ -717,6 +722,11 @@ typedef struct {
 /* Flags for table init. */
 #define TSK_NO_METADATA (1 << 0)
 
+/* Flags for subset() */
+#define TSK_SUBSET_KEEP_UNUSED_POPULATIONS (1 << 0)
```
jeromekelleher, line 726: We could reuse the TSK_FILTER_POPULATIONS options etc for simplify?

python/tskit/tables.py
petrelharp: Are you thinking that `canonicalize` would have a separate implementation, or would end up calling `sort`? I did it this way because there'd be a lot of code overlap.

jeromekelleher: I'm imagining that it would look something like
```c
int TSK_WARN_UNUSED
tsk_table_collection_canonicalise(tsk_table_collection_t *self,  tsk_flags_t options)
{
    int ret = 0;
    tsk_id_t k;
    tsk_flags_t subset_options = options & TSK_UNCHANGED_POPULATIONS;
    tsk_id_t *nodes;
    tsk_table_sorter_t sorter;

    ret = tsk_table_sorter_init(&sorter, self, options);
    if (ret != 0) {
        goto out;
    }

        nodes = malloc(self->nodes.num_rows * sizeof(*nodes));
        if (nodes == NULL) {
            ret = TSK_ERR_NO_MEMORY;
            goto out;
        }
        for (k = 0; k < (tsk_id_t) self->nodes.num_rows; k++) {
            nodes[k] = k;
        }
        subset_options
            |= (TSK_SUBSET_KEEP_UNUSED_POPULATIONS | TSK_SUBSET_KEEP_UNUSED_INDIVIDUALS
                | TSK_SUBSET_KEEP_UNUSED_SITES);
        ret = tsk_table_collection_subset(
            self, nodes, self->nodes.num_rows, subset_options);
        sorter.sort_mutations = tsk_table_sorter_sort_mutations_canonical;
        tsk_safe_free(nodes);
 
    ret = tsk_table_sorter_run(&sorter, start);
    if (ret != 0) {
        goto out;
    }
out:
    tsk_table_sorter_free(&sorter);
    return ret;
}
```

I think we can add a flag to the ``tsk_table_sorter_t`` struct to decide whether we're sorting mutations in canonical order or just the standard order (i.e., I agree with you that the current duplication in tsk_table_sorter_sort_mutations_canonical isn't very nice).

And there's no changes to the current ``tsk_table_collection_sort``.

How does that sound?

petrelharp: Sounds good, I had this in mind as an alternative also. I will make the change (and the other suggestions).

python/tskit/trees.py
petrelharp: Unfortunately, we can't exactly mirror `simplify`. Here we've got three options:

1. remove unused pops
2. reorder pops, with unused ones at the end
3. keep pops the same

This is because for `canonicalize` we need to reorder things, but not discard information - that's option (2). Similarly, the option for `individuals` toggles between options (1) and (2). But the `filter_X` flags for `simplify` toggle between options (1) and (3).

Maybe the best thing to do is to mirror `simplify` at the python level, but then leave an extra flag that does option (2) (e.g., `TSK_SUBSET_CANONICALIZE`) at the C level, perhaps not even provided in python?

petrelharp: I've gone ahead and done this change.

python/tskit/tables.py
petrelharp: > I think we can add a flag to the tsk_table_sorter_t struct to decide whether we're sorting mutations in canonical order or just the standard order (i.e., I agree with you that the current duplication in tsk_table_sorter_sort_mutations_canonical isn't very nice).

Ok, I've put back `sort` and made `canonicalise`, but I've kept `tsk_table_sorter_sort_mutations_canonical` separate from `tsk_table_sorter_sort_mutations` because the two functions use different types for `sorted_mutations`, and I don't know how to write the code in a way that removes the duplications.

petrelharp:
Ok - up-to-date now. I'm not sure how to de-duplicate the mutation sorting code without making the standard case less efficient. Have a look?

jeromekelleher:
Sorry I've been slow to get back to this @petrelharp, I've been tied up with PopGroup. Will take a good look tomorrow!

petrelharp:
No worries!

python/tests/tsutil.py
```diff
@@ -1282,3 +1567,47 @@ def genealogical_nearest_neighbours(ts, focal, reference_sets):
...
+    assert t1.metadata == t2.metadata
+    assert t1.metadata_bytes == t2.metadata_bytes
+    assert t1.sequence_length == t2.sequence_length
+    assert t1 == t2
+
+
+def assert_tables_equal(t1, t2, label=""):
+    if hasattr(t1, "metadata_schema"):
+        if t1.metadata_schema != t2.metadata_schema:
+            print(f"{label} :::::::::: t1 ::::::::::::")
```
jeromekelleher, line 1598: I don't think we should use prints here for a few reasons: (a) when we promote this to a real method (#1076) the message should be part of the exception that gets raised; (b) it's quite easy to miss prints in pytest anyway, and I think a lot of the people we're trying to help would miss the messages.

What about:
```python
if t1.metadata_schema != t2.metadata_schema:
    msg = (
         f"{label} :::::::::: t1 ::::::::::::\n{t1.metadata_schema}"
         f"{label} :::::::::: t2 ::::::::::::\n{t1.metadata_schema}"
   ) 
   raise AssertionError(msg)
```

python/tskit/tables.py
```diff
@@ -2601,6 +2601,24 @@ def sort(self, edge_start=0):
         self._ll_tables.sort(edge_start)
         # TODO add provenance
 
+    def canonicalise(self):
+        """
+        This puts the tables in *canonical* order - to do this, the individual
```
jeromekelleher, line 2606: I think it's probably worth a section in the data-model docs to describe what the canonical form is. We can then just refer to that here to describe what this is doing.

python/tskit/tables.py
```diff
@@ -2601,6 +2601,24 @@ def sort(self, edge_start=0):
         self._ll_tables.sort(edge_start)
         # TODO add provenance
 
+    def canonicalise(self):
```
jeromekelleher, line 2604: I think we should add ``filter_populations=True, filter_individuals=True`` here just like in simplify.  Shouldn't the default behaviour of canonicalise be to remove unreferenced entities? That way, if you call canonicalise on two things that slightly different versions of redundant information you'll get the same output, which, I would think, is what you want.

python/tskit/tables.py
jeromekelleher, line 2606: I would say canonical *form* (as opposed to *order*)

python/tskit/tables.py
```diff
@@ -2648,6 +2666,9 @@ def deduplicate_sites(self):
         duplicate ``position`` (and keeping only the *first* entry for each
         site), and renumbering the ``site`` column of the mutation table
         appropriately.  This requires the site table to be sorted by position.
+
+        This method does not sort the tables afterwards, so mutations may no longer
```
jeromekelleher, line 2670: Maybe a ``..warning:: `` is what we want for the moment then.

python/tskit/tables.py
```diff
@@ -2924,20 +2945,44 @@ def drop_index(self):
...
+        nodes,
+        record_provenance=True,
+        filter_populations=True,
+        filter_individuals=True,
+        filter_sites=True,
+        canonicalise=False,
+    ):
         """
         Modifies the tables in place to contain only the entries referring to
         the provided list of nodes, with nodes reordered according to the order
```
jeromekelleher, line 2959: More precisely "list of node IDs" - someone might thing these are Node objects.

python/tskit/tables.py
```diff
@@ -2924,20 +2945,44 @@ def drop_index(self):
...
         they appear in the list. See :meth:`TreeSequence.subset` for a more
         detailed description.
 
+        Note: the tables can be completely unsorted.
```
jeromekelleher, line 2963: Maybe "There are no sortedness requirements for the tables".

python/tskit/tables.py
```diff
@@ -2924,20 +2945,44 @@ def drop_index(self):
...
+
         :param list nodes: The list of nodes for which to retain information. This
             may be a numpy array (or array-like) object (dtype=np.int32).
         :param bool record_provenance: Whether to record a provenance entry
             in the provenance table for this operation.
+        :param bool filter_populations: Whether to remove populations not referenced by
+            retained nodes. If False, the population table will remain unchanged.
```
jeromekelleher, line 2970: They will be reordered, though, right?

python/tskit/tables.py
```diff
@@ -2924,20 +2945,44 @@ def drop_index(self):
...
+        :param bool filter_individuals: Whether to remove individuals not referenced by
+            retained nodes. If False, the individuals table will remain unchanged.
+        :param bool filter_sites: Whether to remove sites not referenced by
+            retained mutations. If False, the site table will remain unchanged.
+        :param bool canonicalise: If True, retains all unused entries, putting
```
jeromekelleher, line 2975: I don't think the canonical form should include any redundant information (see note above), so I don't think this is the right name for the parameter. Seems like we're getting into a bit of a mess here just to keep unreferenced objects in the data model. Is this something that we really need?

jeromekelleher (commented):
I've had a good look at the Python end of this and it looks good @petrelharp, but I think I'm confused about the entry point here. This started out as you wanting to do something specific, and has now turned into us defining a canonical form for tree sequences (which is great!). I'm not sure that the canonical form that I have in mind (i.e., no redundancy, at least by default) actually does what you want, though.

You want to keep population IDs stable through some union/subset etc operations, right?

python/tests/tsutil.py
```diff
@@ -135,6 +136,44 @@ def insert_branch_mutations(ts, mutations_per_branch=1):
...
+
+def insert_discrete_time_mutations(ts, num_times=4):
+    tables = ts.tables
+    tables.sites.clear()
+    tables.mutations.clear()
+    height = max([t.time(t.roots[0]) for t in ts.trees()])
+    for pos in range(int(tables.sequence_length)):
+        anc = "X" * pos
+        tables.sites.add_row(position=pos, ancestral_state=anc)
+        t = ts.at(pos)
```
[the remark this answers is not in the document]
petrelharp: I wanted the times to be the same at different sites, just to make more things that could go wrong in sorting.

python/tests/tsutil.py
```diff
@@ -753,6 +825,219 @@ def compute_mutation_times(ts):
...
+            metadata=n.metadata,
+        )
+    # edges
+    randomised_edges = list(orig.edges)
+    if shuffle_edges:
+        rng.shuffle(randomised_edges)
+    for e in randomised_edges:
+        tables.edges.add_row(e.left, e.right, e.parent, e.child, metadata=e.metadata)
+    # sites
+    randomised_sites = list(enumerate(orig.sites))
```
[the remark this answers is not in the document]
petrelharp: Yes!! This big was all @mufernando!

python/tests/tsutil.py
```diff
@@ -753,6 +825,219 @@ def compute_mutation_times(ts):
...
+            s.position, ancestral_state=s.ancestral_state, metadata=s.metadata
+        )
+    # mutations
+    randomised_mutations = list(enumerate(orig.mutations))
+    if shuffle_mutations:
+        if keep_mutation_parent_order:
+            # randomise *except* keeping parent mutations before children
+            mut_site_order = [mut.site for mut in orig.mutations]
+            rng.shuffle(mut_site_order)
+            mut_by_site = {s: [] for s in mut_site_order}
```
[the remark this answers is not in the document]
petrelharp: I guess, but this is a bit more clear what the keys of `mut_by_site` are going to be... I'd rather leave it?

python/tskit/tables.py
petrelharp: No! I changed the behavior, to match `simplify`.

petrelharp:
Ok, the question is: should `canonicalise` remove unreferenced individuals and populations? (see [comment](https://github.com/tskit-dev/tskit/pull/1108/files/46731d8d90e0c1f8d69a0a3103232dcab4245111#r553316204) and [comment](https://github.com/tskit-dev/tskit/pull/1108#discussion_r553320792). Gee, I'm not sure.

Here's some use cases:

1. Testing for shared overlap in `union`: we want to know that everything relevant to the nodes in question is equivalent. So, removing unreferenced nodes/individuals is OK.
2. Use in test suites that the C and python versions do exactly the same thing (e.g., for `subset` or `simplify`). In these cases, I could imagine it allowing bugs to slip through the cracks (eg if we accidentally leave some extra populations in but they're not referenced?) but it seems kinda unlikely.
3. Use in test suites for operations that should be "no change"; like `t2 = ts.copy(); no_op(t2); t1.canonicalise(); t2.canonicalse(); assert t1 == t2`. Dropping unreferenced inds/pops would mean we couldn't test `no_op` on tables with unreferenced things.
4. Putting totally out-of-order tables in order so mutation parents come before children. Here it seems bad to drop unreferenced things, although arguably this functionality should be provided by a different method.

I do like the idea of dropping unreferenced things, since we aren't actually canonically sorting these. But for (2-4) it does seem good to leave them in. So, maybe it should be optional, defaulting to `False`?

Right now, `subset` has an option to retain unreferenced nodes and individuals, I could rename this from `canonicalise` to `keep_unused_individuals` and `keep_unused_populations`, and add the same options to `TableCollection.canonicalise` that controls this behavior?

Or, maybe these test cases are unconvincing?

petrelharp:
I got the rest of those changes in, btw.

c/tskit/tables.h
```diff
@@ -2708,19 +2727,33 @@ they appear in the ``nodes`` argument. Specifically, this subsets and reorders
...
+4. Edges: if both parent and child are retained nodes.
+5. Mutations: if the mutation's node is a retained node.
+6. Sites, if `TSK_FILTER_SITES`: if any mutations remain at the site after
+    removing mutations.
 
 Retained edges, mutations, and sites appear in the same
-order as in the original tables.
+order as in the original tables. If any of `TSK_FILTER_INDIVIDUALS`,
+`TSK_FILTER_POPULATIONS`, or `TSK_FILTER_SITES` are *not* provided,
+then the respective tables will be *unchanged*.
```
jeromekelleher, line 2741: Oh, you want the other tables entirely unchanged? Ah, *that's* what I've been confused about. Right, you don't want to reorder the populations because you want stable population/individual IDs. 

Well, what if we add a flag ``TSK_STABLE_IDs`` or something to ``subset`` then which does the least amount of ID reshuffling as possible, which still doing the basic node-subset operation? This could be mutually exclusive to the FILTER_X operations, or we could drop them entirely if they're not useful.

Or maybe your original idea of a  ``reorder_nodes`` function is better? I'm finding it hard to understand the desired semantics of the ``subset`` function at this point, so maybe another function is a better idea?

jeromekelleher (commented):
Apologies it's taken so long to get to this @petrelharp. I'm still confused though, unfortunately. My takes:

1. The ``TableCollection.canonicalise()`` method is a useful operation which we should support. By default it should remove redundant information, but we should add a ``keep_unreferenced=False`` option to allow a user to keep unreferenced information at the end of the tables. It's probably not worth having ``keep_unreferenced_populations``, ``keep_unreferenced_individuals`` etc, but we can imagine doing this, if needed. Unreferenced stuff should definitely go at the **end** of the table, though, as otherwise we're not canonicalising in any meaninful way.
2. The ``subset(nodes)`` operation no longer requires sorted input - great!
3. The ``subset(nodes)`` operation now has a bunch of options and complicated semantics I don't understand :crying_cat_face: 
4. We might be able to replace the complicated semantics of ``subset(nodes)`` with ``reorder_nodes(nodes)`` operation which rejiggles the node table a bit but doesn't change any other tables?

I've made some comments above, but I think they probably reflect my confusion in that some are probably contradictory - sorry about that.

c/tskit/tables.h
```diff
@@ -686,13 +688,14 @@ typedef struct {
...
 
-/* Flags for simplify() */
+/* Flags for simplify() and/or subset() */
 #define TSK_FILTER_SITES (1 << 0)
 #define TSK_FILTER_POPULATIONS (1 << 1)
 #define TSK_FILTER_INDIVIDUALS (1 << 2)
 #define TSK_REDUCE_TO_SITE_TOPOLOGY (1 << 3)
 #define TSK_KEEP_UNARY (1 << 4)
 #define TSK_KEEP_INPUT_ROOTS (1 << 5)
+#define TSK_CANONICALISE (1 << 6)
```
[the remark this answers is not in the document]
petrelharp: That's not what this flag does - this flag is the thing we use within tsk_table_collection_canonicalise. (I will explain later.)

c/tskit/tables.h
```diff
@@ -2708,19 +2727,33 @@ they appear in the ``nodes`` argument. Specifically, this subsets and reorders
 each of the tables as follows:
 
 1. Nodes: if in the list of nodes, and in the order provided.
-2. Individuals and Populations: if referred to by a retained node,
-   and in the order first seen when traversing the list of retained nodes.
-3. Edges: if both parent and child are retained nodes.
-4. Mutations: if the mutation's node is a retained node.
-5. Sites: if any mutations remain at the site after removing mutations.
+2. Individuals, if `TSK_FILTER_INDIVIDUALS`: if referred to by a retained node,
```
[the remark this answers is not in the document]
petrelharp: sounds good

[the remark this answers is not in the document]
petrelharp: oh wait, that's not what this does at the moment...

petrelharp:
TODO:

- [ ] collapse the two `keep_unreferenced_X` options into one
- [ ] add `keep_unreferenced` as an option to `canonicalise`
- [ ] figure out what do do with subset's many options

And, hm, let's see. Dredging this back up from my memory... We have three behaviors for `subset`:

1. reorder and remove unreferenced individuals and populations - needed for `check_subset_equality`
2. reorder but don't remove unreferenced - needed for `canonicalise` (with `keep_unreferenced`)
3. don't touch the individual and population tables - useful to keep population references stable (the motivation for keeping individuals is less obvious); and to match the behavior of `simplify` with `filter_X=False`

I do think we need to do all of these things. You're suggesting separating out (2) as a different method, `tsk_table_collection_reorder_nodes`, maybe? I like this option? It would make the "reorder nodes" operation more discoverable. I may not be up for the refactoring required to get `reorder` to not have a bunch of code duplicated from `subset` right now, though.

Does that make things clear? What do you think?

jeromekelleher:
This is very helpful, thanks @petrelharp. Also very tricky!

It would be nice if we could separate out the operations of ``subset``ing from ordering, but I see how these are intrinsically tied up at the moment. Suppose we defined ``tables.subset(nodes)``so that it *only* touches the node table and keeps all the other tables as-is.

Then, we have a ``canonicalise(keep_unreferenced=False)`` function which puts the tables in a canonical form, which can optionally **not** remove unreferenced items.

So, you're operations above become:
1. tables.subset(nodes); tables.canonicalise()
2. (part of the implementation of canonicalise)
3. tables.canonicalise(keep_unreferenced=False)

I don't think it's that much less efficient to split up the subset() and ordering() concerns really. I guess one concern though is that we've released subset() under the current semantics, so we shouldn't change them too lightly.

petrelharp:
That also seems fine, except for changing the behavior of `subset` post-release. Probably no-one is relying on it? If someone is, it's @mufernando ... could you have a look at the last two comments and see what you think?

jeromekelleher:
I don't know, it's all very tricky. On the one hand it's nice that ``subset()`` sort of "radiates out" from the given set of nodes and,  optionally discards or not unreferenced entities. Nodes are the focal part of the data model, so it makes sense to make them the central entity we subset by. On the other hand, all the different keep/don't keep/reorder/don't reorder options are very hard to get your head around and keep track of.

What if there was just two options to subset then? By default we reorder and filter unreferenced; if KEEP_UNREFERENCED we keep unreferenced entitites at the end of the tables; and if MAINTAIN_REFERENCES we keep all other references besides nodes and edges stable. The ``canonicalise`` function can then call subset() and do some extra sorting.

Would this cover all the requirements? 

@mufernando, any input here?

jeromekelleher:
Hey, what if we made a method called ``rejigger_ids``, that would solve all our problems! (:zany_face: )

mufernando:
gee, this is way more complicated than I initially thought!

> I don't know, it's all very tricky. On the one hand it's nice that `subset()` sort of "radiates out" from the given set of nodes and, optionally discards or not unreferenced entities. Nodes are the focal part of the data model, so it makes sense to make them the central entity we subset by. On the other hand, all the different keep/don't keep/reorder/don't reorder options are very hard to get your head around and keep track of.
> 
> What if there was just two options to subset then? By default we reorder and filter unreferenced; if KEEP_UNREFERENCED we keep unreferenced entitites at the end of the tables; and if MAINTAIN_REFERENCES we keep all other references besides nodes and edges stable. The `canonicalise` function can then call subset() and do some extra sorting.
> 
> Would this cover all the requirements?
> 
> @mufernando, any input here?

I agree with this, we defined `subset` as an operation on a set of nodes, but that also affects the other tables and we should stick to that. Some of the other, more "simple" behavior that the `filter_X` flags are doing should really be relegated to a different operation such as `reorder_nodes`, which we can implement later on and refactor `subset` to use it. 

I'm not exactly clear on what the MAINTAIN_REFERENCES would do though.

petrelharp:
I'll have a look and try something out. Maybe if I just put some more effort into the documentation, it'll all be clear and natural as-is. =)

petrelharp:
Another data point on how we ended up with the current set of options:
Screenshot from 2021-01-26 15-55-11

petrelharp:
Looking above, @jeromekelleher, I see that you maybe weren't entirely happy with this being what "canonicalise" means. If you had something else in mind, we could reserve that keyword for some other yet-to-be-defined operation, and implement this as `sort(..., complete=True)` or something.

petrelharp:
Looking back, one option is to remove (3) entirely: we don't have a particular use case for "keep individual/population/site tables unchanged". Here's why I implemented it, though: I wanted to write a test where I (1) used `subset` to extract from a tree sequence both (a) all the relationships to a single individual; and (b) everything else, and then (2) use `union` to put (a) and (b) back together to get the original tree sequence. However, this runs into problems if the population table changes: if everything in (a) is from population 1, for instance, then what was population (1) becomes population (0) in (a), and so the pieces we union back together are no longer compatible (since we don't have a `population_map` argument to `union`). I thought that maintaining sites might be useful to people who, say, want to have genotypes comparable across different `subset`s.  I don't have any use case for keeping individual tables unchanged, and implemented it for symmetry with `simplify`.

So: I'd be willing to remove the `filter_sites`, `filter_individuals` and `filter_populations` flags entirely. It seems a bit of a shame, because `subset` is a fairly low-level tool, and having it be more flexible seems helpful (as it was for the task above).

I think the reason this is difficult is because of how things are named in `simplify`: since it uses `filter_X` to toggle between "keep unchanged" and "reorder and remove", it's hard to come up with another option that does the halfway operation of "just reorder, but don't remove".  Also, maybe we should not mirror `simplify`, as we want to avoid conflating the two in people's minds (they do pretty different things but easy to conflate!).

So, here's a proposal:
```
def subset(self, nodes, record_provenance=True,
                  reorder_populations=True,
                  remove_unreferenced=True):
    """
    Returns a tree sequence containing only information directly referencing the provided list of nodes to retain.
    The result will retain only the nodes whose IDs are listed in ``nodes``, only edges for which both parent
    and child are in ``nodes```, only mutations whose node is in ``nodes``, and only individuals that are referred to
    by one of the retained nodes.

    This has the side effect of reordering the nodes, individuals, and populations in the tree sequence: the nodes
    in the new tree sequence will be in the order provided in ``nodes``, and both individuals and populations will be
    ordered by the earliest retained node that refers to them. (However, ``reorder_populations`` may be set to False
    to keep the population table unchanged.)

    By default, the method removes all individuals and populations not referenced by any nodes, and all sites not
    referenced by any mutations. To retain these unreferencd individuals, populations, and sites, pass
    ``remove_unreferenced=False``. If this is done, the site table will remain unchanged, unreferenced individuals
    will appear at the end of the individuals table (and in their original order), and unreferenced populations will
    appear at the end of the population table (unless ``reorder_populations=False``).
    """
```

In this case, the three operations above are:
1.   reorder and remove unreferenced individuals and populations - `subset(  )`
2.   reorder but don't remove unreferenced - `subset(remove_unreferenced=False)`
3.   don't touch the population tables - `subset(reorder_populations=False)`
I've removed "don't touch individual tables" from (3) because I don't think there's really a use case for that. If there is, though, we could add `reorder_individuals` later.

petrelharp:
Update:
- my second proposal above sounds good for subset
- it remains to decide if `canonicalise` removes unreferenced things or not.
 
We *don't* need to retain unreferenced things for its use in `union` (and arguably it'd be better if it didn't, actually). So, here's my proposal:
- add a `remove_unreferenced=True` argument to `canonicalise`.

jeromekelleher:
I like the new signature for ``subset``, and canonicalise(remove_unreferenced=True) seems good to me. (Also keep_unreferenced=False is fine too, if this works better in terms of syncing with the C flags. I guess you'd have to do it as TSK_KEEP_UNREFERENCED, for options=0 to mean "remove unreferenced, right?)

jeromekelleher:
Apologies for sending things in the wrong direction with the misguided FILTER_X options, btw.

petrelharp:
(Note: subset [has not been released with flags](https://github.com/tskit-dev/tskit/blob/C_0.99.10/c/tskit/tables.h#L2738), so this will be an API change.)

c/tskit/tables.c
```diff
@@ -4715,6 +4716,152 @@ tsk_table_sorter_free(tsk_table_sorter_t *self)
...
+    }
+    if (ret == 0) {
+        ret = (ia->num_descendants < ib->num_descendants)
+              - (ia->num_descendants > ib->num_descendants);
+    }
+    if (ret == 0) {
+        ret = (ia->mut.node > ib->mut.node) - (ia->mut.node < ib->mut.node);
+    }
+    if (ret == 0) {
+        ret = (ia->mut.id > ib->mut.id) - (ia->mut.id < ib->mut.id);
```
petrelharp, line 4747: I don't know why Codecov says this is not covered - I've stuck in some printfs, and it IS covered, in `tests/test_tables.py::TestSortTables::test_no_mutation_parents`.

petrelharp:
I think this is ready to go, and passing all the tests, despite what github and codecov say (!?!?). Want to take a final (?) look, @jeromekelleher?

Note: I thought about calling the option `keep_unreferenced` rather than `remove_unreferenced`, but thought the double-negative default of `keep_unreferenced=False` was more confusing.

c/tskit/tables.c
jeromekelleher: Needs to be covered by C tests, Python tests aren't counted for coverage here.

jeromekelleher: I added an identical mutation to your test_canonicalise. Nice test!

jeromekelleher:
I've gone through this in detail @petrelharp, and I happy to merge. I've rebased to fixup recent breakages, and made a few small changes. Can you have a look at the last two commits please to make sure you're OK with the updates?

Some notes:

- For new additions to the API, it really does help if we don't hard-code defaults into the method signature unless we're absolutely **certain** they will never change (i.e., in the presence of all possible future parameters).
- Also good to make sure that new additions like this are keyword only.

petrelharp:
LGTM! Thanks for fixing those things up.

petrelharp:
I can squash & rebase later, but not right now - go ahead if you want.

petrelharp:
Whoops, sorry, almost had a git mishap there.

petrelharp:
Um, I guess I could click the "rebase and merge" button, but I'm nervous to since last time. (I guess we need to adjust our threshold on the codecov again, eh?)

jeromekelleher:
> Um, I guess I could click the "rebase and merge" button, but I'm nervous to since last time. (I guess we need to adjust our threshold on the codecov again, eh?)

I think it's fine having multiple commits on something big like this. The codecov levels aren't part of the mergify rules, so we don't have to worry about them. We should really use the AUTOMERGE-REQUESTED label here on tskit for merging stuff rather than the buttons (I've not been using it recently on msprime because there's too much stuff going on, and there isn't really anyone to do the reviewing; and someone can't approve their own PRs)

jeromekelleher (approved):

benjeffery:
> (I've not been using it recently on msprime because there's too much stuff going on, and there isn't really anyone to do the reviewing; and someone can't approve their own PRs)

We could add a label to merge without review if you are the PR author?

jeromekelleher:
> We could add a label to merge without review if you are the PR author?

It's not worth the bother - things will settle down soon enough in msprime, and it's bad practise anyway.

PR #1109, opened 2021-01-06
Raise error when additional_properties specified for struct codec

### LINKED ISSUE #1085: additionalProperties=True ignored in struct codec
Setting ``"additionalProperties": True`` in a struct metadata schema does not raise an error and is ignored.

We should probably raise an error if someone tries to set this.

benjeffery:
Closes #1085

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -455,6 +455,10 @@ class StructCodec(AbstractMetadataCodec):
             elif type(obj) == dict:
                 ret = {k: enforce_fixed_properties(v) for k, v in obj.items()}
                 if ret.get("type") == "object":
+                    if ret.get("additional_properties"):
+                        raise ValueError(
+                            "Struct codec does not support additional_properties"
+                        )
                     ret["required"] = list(ret.get("properties", {}).keys())
                     ret["additionalProperties"] = False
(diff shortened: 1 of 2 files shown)


python/tskit/metadata.py
```diff
@@ -455,6 +455,10 @@ def enforce_fixed_properties(obj):
             elif type(obj) == dict:
                 ret = {k: enforce_fixed_properties(v) for k, v in obj.items()}
                 if ret.get("type") == "object":
+                    if ret.get("additional_properties") is True:
```
jeromekelleher, line 458: Any reason for ``is True`` here? This is normally a no-no. Also, not clear why we're using ``ret.get()`` here rather than ``ret[]``.

benjeffery: Just realised that additional_properties can be an object so this needs to change. Will fix.

benjeffery: additional_properties is optional, hence the use of get. As it can be an object I've reverted to truthiness. Should be the right thing now.

jeromekelleher: Hah, I didn't realise that the default default value for ``get`` was None!

jeromekelleher (approved):

PR #1110, opened 2021-01-06
Add schema.asdict

### LINKED ISSUE #1070: Document MetadataSchema.schema or as asdict() method
There's currently no documented way to get the actual schema out of a MetadataSchema object. The ``.schema`` attribute feels a bit circular to me as a public API, so maybe we add an ``asdict()`` method?

benjeffery:
Fixes #1070 by adding asdict to MetadataSchema.

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -548,8 +548,16 @@ class MetadataSchema:
 
     @property
     def schema(self) -> Optional[Mapping[str, Any]]:
-        # Make schema read-only
-        return self._schema
+        # Return a copy to avoid unintentional mutation
+        return copy.deepcopy(self._schema)
+
+    def asdict(self) -> Optional[Mapping[str, Any]]:
+        """
+        Returns a dict representation of this schema. One possible use of this is to
+        modify this dict and then pass it to the ``MetadataSchema`` constructor to create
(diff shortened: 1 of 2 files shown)


python/tskit/metadata.py
```diff
@@ -544,8 +544,14 @@ def __eq__(self, other) -> bool:
...
     @property
     def schema(self) -> Optional[Mapping[str, Any]]:
-        # Make schema read-only
-        return self._schema
+        # Return a copy to avoid unintentional mutation
+        return copy.deepcopy(self._schema)
+
+    def asdict(self) -> Optional[Mapping[str, Any]]:
+        """
+        Return a deep copy of the schema.
```
jeromekelleher, line 552: This isn't a super descriptive comment - we should be distinguishing the schema as an object that we're modelling (which is ``self``) and the dictionary representation of the schema. This was one of the main things that confused me when getting my head around the API. The fact we're storing the dictionary representation of the schema in a public attribute called ``schema`` is a detail that people shouldn't need to know about.

benjeffery: Yes! I've hopefully improved it.

jeromekelleher (approved):
LGTM!

PR #1116, opened 2021-01-15
Make struct codec encode default values

benjeffery:
Partial work for #1073 on the struct side.

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -1006,6 +1006,23 @@ class TestStructCodec:
         assert ms.decode_row(ms.validate_and_encode_row(row_data)) == row_data
         assert ms.decode_row(ms.validate_and_encode_row(None)) is None
 
+    def test_default_values(self):
+        schema = {
+            "codec": "struct",
+            "type": "object",
+            "properties": {
+                "int": {"type": "number", "binaryFormat": "b", "default": 42},
+                "float": {"type": "number", "binaryFormat": "d"},
+            },
+        }
+        ms = metadata.MetadataSchema(schema)
+        row_data = {"float": 5.5}
+        assert ms.validate_and_encode_row(row_data) == b"\x00\x00\x00\x00\x00\x00\x16@*"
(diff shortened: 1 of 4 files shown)


python/tests/test_metadata.py
```diff
@@ -1494,6 +1511,37 @@ def test_additional_properties(self):
...
+                "int": {"type": "number", "binaryFormat": "i"},
+                "float": {"type": "number", "binaryFormat": "d"},
+            },
+            "required": ["float"],
+        }
+        with pytest.raises(
+            exceptions.MetadataSchemaValidationError,
+            match="Optional property 'int' must have a default value",
+        ):
+            self.encode(schema, {"float": 5.5})
```
jeromekelleher, line 1528: This should be raised when we create the schema rather than when encoding data I think? (probably is, but the test should reflect this)

python/tskit/metadata.py
```diff
@@ -454,12 +482,19 @@ def enforce_fixed_properties(obj):
...
                         )
-                    ret["required"] = list(ret.get("properties", {}).keys())
+                    # To prevent authors having to list required properties the default
+                    # is that all without a default are required.
+                    if "required" not in ret:
+                        ret["required"] = [
+                            prop
+                            for prop, sub_schema in ret.get("properties", {}).items()
+                            if "default" not in sub_schema
+                        ]
```
jeromekelleher, line 497: Can check here if any properties that *don't* have a default are in required too.

jeromekelleher (changes requested):
Looks good. I think we should validate the **schema** at construction time, though, so raise an error if MetadataSchema(schema) is called with conflicting required/default values.

python/tests/test_metadata.py
benjeffery: Modified test to just create schema, the schema was where the error was already being rasied.

python/tskit/metadata.py
benjeffery: That's checked on line 174

benjeffery:
@jeromekelleher Should be good to go now

jeromekelleher (approved):

PR #1117, opened 2021-01-15
Time join

jelliotthiggins:
@petrelharp @jeromekelleher - here's the new PR

Initial implementation of the time join functionality in tskit (instead of msprime or pyslim). Posting as a draft so you can hopefully take a look as I work on the tests! And copying over my (slightly edited) description for others' reference (discussion can be found on [msprime here](https://github.com/tskit-dev/msprime/pull/1347) and [pyslim here](https://github.com/tskit-dev/pyslim/pull/118)). 

In using pyslim, I found the process described in ["Vignette: Following up with more coalescent simulation" ](https://pyslim.readthedocs.io/en/latest/vignette_continuing.html)incredibly useful. To make it more accessible for all tree sequences instead of just those generated using SLiM or msprime, I added it to tskit as a `TableCollection` method called `time_join`. It is all contained in the high level python right now, although a C implementation could be useful long term.

It takes in a "recent" and an "ancient" `TreeSequence` and the length of time to separate them by before joining. It outputs a complete tree sequence, containing the entirety of both `TreeSequence` objects. It uses two helper methods for now, `set_time`, which performs the time adjustment on the tables, and `set_flags`, which removes the sample flags on the "ancient" `TreeSequence` (and could also be used to place sample flags in other cases)

The initial use described in the example taking a tree sequence produced by SLiM, recapitating, adding mutations, and continuing the simulation, but this could be done with any tree sequences.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3026,22 +3026,26 @@ class TableCollection:
                 record=json.dumps(provenance.get_provenance_dict(parameters))
             )
 
-    def time_join(recent_ts, ancient_ts, time, continue_nodes=None):
+    def time_join(self, recent_ts, time, continue_nodes=None):
         if continue_nodes is None:
-            continue_nodes = ancient_ts.samples()
+            continue_nodes = self.samples()
 
         new_tables = recent_ts.tables
         new_nodes = np.empty(0)
-        for root in set().union(*[set(t.roots) for t in recent_ts.trees()]):
-            # root is just an id, make sure I actually get the node
-            if root.time == time:
-                new_nodes = np.append(new_nodes, root)
-            elif root.time < time:
-                n = new_tables.nodes.add_row(time=time)
-                # need to get the tree to be able to get the interval
-                # new_tables.edges.add_row(
-                #    parent=n, child=t.root, left=t.interval.left, right=t.interval.right
-                # )
+        # NOTE - edges will be unique because of their interval
+        for t in recent_ts.trees():
+            for root in t.roots:
+                root = recent_ts.node(root)
+                if root.time == time:
+                    new_nodes = np.append(new_nodes, root)
+                elif root.time < time:
+                    n = new_tables.nodes.add_row(time=time)
+                    new_tables.edges.add_row(
+                        parent=n,
+                        child=root,
+                        left=t.interval.left,
+                        right=t.interval.right,
+                    )
                 new_nodes = np.append(new_nodes, n)
             else:
                 raise RuntimeError
@@ -3060,7 +3064,7 @@ class TableCollection:
             continue_nodes, len(new_nodes), replace=False
         )
 
-        tables = ancient_ts.tables
+        tables = self.tables
         tables.set_samples(continue_nodes, samples=False)
         tables.union(
             recent_ts.tables,
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2976,10 +2976,10 @@ class TableCollection:
         self,
         other,
         node_mapping,
-        offset_time=None,
         check_shared_equality=True,
         add_populations=True,
         record_provenance=True,
+        shift_time=0.0,
     ):
         """
         Modifies the table collection in place by adding the non-shared
@@ -3005,8 +3005,8 @@ class TableCollection:
         """
         node_mapping = util.safe_np_int_cast(node_mapping, np.int32)
 
-        # adjust table time
-        # adjust sample flags
+        if shift_time > 0.0:
+            self._ll_tables.set_time(shift_time)
 
         self._ll_tables.union(
             other._ll_tables,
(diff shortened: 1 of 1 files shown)


jelliotthiggins:
I haven't tried combining sites and mutations yet so that's definitely something to address. Do either of you know if it has been done elsewhere?

petrelharp:
Hm - in thinking this through, I realized a mistake, that I also made in the vignette. Suppose that a root `n` of a tree on the interval `[a, b)` has time `t`, which is more recent than the target `time`. Here, you're adding a new node at `time` and an edge from `n` to that new node on `[a,b)`. That isn't right, because there would have been recombination over that time period - for instance, if `[a,b) is very long, then probably `n` inherited from more than one distinct ancestor back at `time`. In the vignette, I neglected the possibliity of roots more recent than `time` entirely (and, happily, there weren't any), but [need to fix this](https://github.com/tskit-dev/pyslim/issues/127). (A way around this would be to add a single sample in a disconnected population that artifically keeps all lineages going until `end_time`.)

So: I think we should not deal with the case where some roots of trees are more recent than `time`, as just merging them on like this is not usually correct. (Good news! That makes this simpler!) Sorry about misleading you on this!

OK, other thoughts:

- both `set_samples` and `set_time` (I'd call it `shift_time`?) are very fundamental things that could be useful, but ought to be implemented in C if they're going to be standalone methods
- the `time_join` method itself also ought to be implemented in C, I think; this python is very useful in making things concrete and providing something to test against.

So: is doing the C something that you are interested in doing, @jelliotthiggins? Feel free to say that you'd rather stick wtih your python code...

petrelharp:
ps - I fixed up the vignette: https://github.com/tskit-dev/pyslim/pull/128

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+            recent_ts.tables,
+            node_map,
+            add_populations=False,
+            check_shared_equality=False,
+            shift_time=time,
+        )
+
+        return tables.tree_sequence()
+
+    def set_samples(self, nodes, samples=True):
```
jeromekelleher, line 3083: I don't think we want this method either - it's a very general name for doing something quite specific. If we add methods to the namespace of the TableCollection we need to (a) be very sure that the operation we're adding is generally useful, and not blocking us from doing something similar in the future under this name (i.e., is there something slightly different, but incompatible we might want to do in future under the same name?); (b) make sure it's very thoroughly tested and documented.

Since all we're doing is updating the flags, it would be much simpler to do something like
```python
tables.nodes.flags = new_flags
```

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
                 record=json.dumps(provenance.get_provenance_dict(parameters))
             )
 
+    def time_join(self, recent, time, continue_nodes=None):
```
jeromekelleher, line 3029: Probably more intuitive to call this parameter ``join_nodes``, or perhaps just ``nodes``?

Also, I think we can make ``time`` an optional parameter, making it default to the time of the oldest root in ``recent_ts``

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+        ancient_ts = self.tree_sequence()
+        recent_ts = recent.tree_sequence()
+        if continue_nodes is None:
+            continue_nodes = ancient_ts.samples()
+
+        new_tables = recent_ts.tables
+        new_nodes = np.empty(0, dtype=np.uint32)
```
jeromekelleher, line 3036: It's actually probably more efficient to use a Python list here, since we're appending one-by-one, i.e.
``new_nodes = []``.

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
                 record=json.dumps(provenance.get_provenance_dict(parameters))
             )
 
+    def time_join(self, recent, time, continue_nodes=None):
+        ancient_ts = self.tree_sequence()
```
jeromekelleher, line 3030: We're not actually using ``ancient_ts`` here for anything, and creating a ``tree_sequence`` is quite expensive. We can get the sample nodes directly from the tables.

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+            raise RuntimeError(
+                f"""More lineages present at the longest-ago end of `recent_ts` \
+                than there are nodes in `continue_nodes`. \
+                Setting sample_size={len(continue_nodes)} or leaving it to default \
+                will ensure compliance."""
+            )
+
+        node_map = np.repeat(tskit.NULL, recent_ts.num_nodes)
+
+        node_map[new_nodes] = np.random.choice(
```
jeromekelleher, line 3067: This needs a random seed to be set to make it reproducible. We should add a ``random_seed=None`` argument  to the ``time_join`` function, and allocate an instance of ``np.random.Random``.

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+        tables.set_samples(continue_nodes, samples=False)
+        tables.union(
+            recent_ts.tables,
+            node_map,
+            add_populations=False,
+            check_shared_equality=False,
+            shift_time=time,
+        )
+
+        return tables.tree_sequence()
```
jeromekelleher, line 3081: If this is a method of ``TableCollection`` it should update the current table collection *in place*, not return a tree sequence. So, the call to ``union`` should be ``self.union()``

python/tskit/tables.py
jeromekelleher, line 3029: Since we need the trees, I think we should require that the input argument is a tree sequence, so we call it ``recent_ts``.

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+                    n = new_tables.nodes.add_row(time=time)
+                    new_tables.edges.add_row(
+                        left=t.interval.left,
+                        right=t.interval.right,
+                        parent=n,
+                        child=root_id,
+                    )
+                    new_nodes = np.append(new_nodes, n)
+                else:
+                    raise RuntimeError(
```
jeromekelleher, line 3052: A ``ValueError`` is more appropriate here I think, since it's a problem with the input parameter values.

python/tskit/tables.py
```diff
@@ -3020,6 +3026,99 @@ def union(
...
+                        f"Root found older than join time (age = {root.time})"
+                    )
+        recent_ts = new_tables.tree_sequence()
+
+        if len(new_nodes) > len(continue_nodes):
+            raise RuntimeError(
```
jeromekelleher, line 3058: Likewise, this should be a ValueError

python/tskit/tables.py
```diff
@@ -2994,6 +2995,7 @@ def union(
             should be the index of the equivalent node in ``self``, or
             ``tskit.NULL`` if the node is not present in ``self`` (in which case it
             will be added to self).
+        :param float offset_time: Time to offset ``other`` by
```
jeromekelleher, line 2998: The parameter name doesn't agree with the code here.

jeromekelleher (commented):
Looks good @jelliotthiggins, I think the functionality is basically there but we need to reorganise things a little bit.

jeromekelleher:
ps. I don't think we need to do anything in C here, all the operations we need are easily available in the current Python API.

python/tskit/tables.py
jelliotthiggins: Should it also return the tables after the call to `self.union()` or simply update in place? (should both `ancient_tables.time_join(recent_ts)` and `joined_tables = ancient_tables.time_join(recent_ts)` work?)

jelliotthiggins:
Implemented almost all of those changes you pointed out @jeromekelleher (still have to deal with the random seed, wanted to dig a bit into how y'all usually do it for consistency). Still working on tests, it's been a busy week! 

> I haven't tried combining sites and mutations yet so that's definitely something to address. Do either of you know if it has been done elsewhere?

Any thoughts on this point in the meantime? 

> So: I think we should not deal with the case where some roots of trees are more recent than `time`, as just merging them on like this is not usually correct. (Good news! That makes this simpler!) Sorry about misleading you on this!

I haven't addressed this yet and I don't fully understand the single sample in a disconnected population hack @petrelharp - does that prevent coalescence somehow? In the hopes of us finding a solution I'm leaving the conditional in for the time being with the caveat that it's not a final solution!

python/tskit/tables.py
jeromekelleher: Update in place - this is the convention we've adopted for these table collection operations.

We can add a ``TreeSequence.time_join`` method that would return a new tree sequence (but we don't need to do that here).

jeromekelleher:
Looks good @jelliotthiggins! I think ``union`` should already handle the merging of sites and mutations, so it might work already. Best thing to do is write a few simple test cases to try to break it!

jelliotthiggins:
Sounds good, working on tests today. One thing of note - in the Vignette, unflagging samples is done using `~np.uint32(tskit.NODE_IS_SAMPLE)` (bitwise, so int value `4294967294`) but in generated TreeSequences non-samples are flagged with `0`. I'm going to make that change to using `np.uint32(0)` instead if that sounds ok, unless there's a reason to differentiate between original non-samples and former-sample non-samples.

jeromekelleher:
We need to be careful to preserve other bit values beside the NODE_IS_SAMPLE flag, so probably the vignette was unflagging samples by doing a bitwise mask with ~np.uint32(tskit.NODE_IS_SAMPLE). (That's a good test to add - do we preserve flags values properly?)

jelliotthiggins:
> We need to be careful to preserve other bit values beside the NODE_IS_SAMPLE flag, so probably the vignette was unflagging samples by doing a bitwise mask with ~np.uint32(tskit.NODE_IS_SAMPLE). (That's a good test to add - do we preserve flags values properly?)

Gotcha, my bit operations are rusty but that makes sense and I updated! Definitely worth adding a test on that.

benjeffery:
Closing due to inactivity - please re-open if needed.

PR #866, opened 2020-09-28
Individual parents

ivan-krukov:
Implement a ragged parent column, as per #852

PR #1119, opened 2021-01-18
Add KEEP_UNARY_IN_INDIVIDUALS option to simplify()

### LINKED ISSUE #1113: Keeping specific unary nodes though simplify
There are some cases where, during simplify, we might want to leave specific unary nodes in the TS, as long as they are on the lineage to a sample node. At the moment, we can only retain unary nodes by a. specifying `keep_unary=T` in which case *all* unary nodes are kept but risk being lost on the next simplify or b. marking the node as a "sample node" but then the node is always retained even if it is not on the lineage from another sample node. The specific example I have encountered with @mmosborn is where a "census" is performed at a certain time, and the nodes at that time, *and the individuals associated with them*, need to be retained as long as they are in the ancestry. This allows us to get information about individuals and nodes in the past (for example, the spatial location of the individual) even though the nodes aren't strictly "samples" because we don't think we have full genomic information about them.

@jeromekelleher suggests that a way to do this is via an option to simplify which says "keep unary nodes associated with an individual, unless they aren't on a path from a sample". I find that a little contorted, but my suggestion of flagging up the required unary nodes using a flag like NODE_IS_SAMPLE has other problems because it associates a permanent state with the node in question. Other ideas welcome below.

hyanwong:
## Description

Adds KEEP_UNARY_IN_INDIVIDUALS option to simplify()

Fixes #1113 

# PR Checklist:

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -6094,6 +6094,8 @@ simplifier_print_state(simplifier_t *self, FILE *out)
     fprintf(out, "\tkeep_unary              : %d\n", !!(self->options & TSK_KEEP_UNARY));
     fprintf(out, "\tkeep_input_roots        : %d\n",
         !!(self->options & TSK_KEEP_INPUT_ROOTS));
+    fprintf(out, "\tkeep_unary_in_individuals : %d\n",
+        !!(self->options & TSK_KEEP_UNARY_IN_INDIVIDUALS));
 
     fprintf(out, "===\nInput tables\n==\n");
     tsk_table_collection_print_state(&self->input_tables, out);
@@ -6635,8 +6637,16 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
     double left, right, prev_right;
     tsk_id_t ancestry_node;
     tsk_id_t output_id = self->node_id_map[input_id];
+
     bool is_sample = output_id != TSK_NULL;
-    bool keep_unary = !!(self->options & TSK_KEEP_UNARY);
+    bool keep_unary = false;
+    if (self->options & TSK_KEEP_UNARY) {
+        keep_unary = true;
+    }
+    if ((self->options & TSK_KEEP_UNARY_IN_INDIVIDUALS)
+        && (self->tables->nodes.individual[input_id] != TSK_NULL)) {
+        keep_unary = true;
+    }
 
     if (is_sample) {
         /* Free up the existing ancestry mapping. */
@@ -8545,6 +8555,11 @@ tsk_table_collection_simplify(tsk_table_collection_t *self, const tsk_id_t *samp
 
     /* Avoid calling to simplifier_free with uninit'd memory on error branches */
     memset(&simplifier, 0, sizeof(simplifier_t));
+
+    if ((options & TSK_KEEP_UNARY) && (options & TSK_KEEP_UNARY_IN_INDIVIDUALS)) {
+        ret = TSK_ERR_KEEP_UNARY_MUTUALLY_EXCLUSIVE;
+        goto out;
+    }
 
     /* For now we don't bother with edge metadata, but it can easily be
      * implemented. */
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -693,6 +693,7 @@ typedef struct {
 #define TSK_REDUCE_TO_SITE_TOPOLOGY (1 << 3)
 #define TSK_KEEP_UNARY (1 << 4)
 #define TSK_KEEP_INPUT_ROOTS (1 << 5)
+#define TSK_KEEP_UNARY_IN_INDIVIDUALS (1 << 6)
 
 /* Flags for check_integrity */
 #define TSK_CHECK_EDGE_ORDERING (1 << 0)
@@ -2671,6 +2672,9 @@ TSK_KEEP_INPUT_ROOTS
     By default simplify removes all topology ancestral the MRCAs of the samples.
     This option inserts edges from these MRCAs back to the roots of the input
     trees.
+TSK_KEEP_UNARY_IN_INDIVDUALS
+    This acts like TSK_KEEP_UNARY (and is mutually exclusive with that flag). It
+    keeps unary nodes, but only if the unary node is referenced from an individual.
 
 .. note:: Migrations are currently not supported by simplify, and an error will
     be raised if we attempt call simplify on a table collection with greater
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -6,6 +6,10 @@
 
 **Features**
 
+- Add ``TSK_KEEP_UNARY_IN_INDIVIDUALS`` flag to simplify, which allows the user to
+  keep unary nodes only if they belong to a tabled individual. This is useful for
+  simplification in forwards simulations (:user:`hyanwong`, :issue:`1113`, :pr:`1119`).
+
 **Bugfixes**
 
 
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -940,7 +940,18 @@ test_simplest_records(void)
     CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
     tsk_treeseq_free(&simplified);
 
+    ret = tsk_treeseq_simplify(&ts, sample_ids, 2,
+        TSK_KEEP_UNARY | TSK_KEEP_UNARY_IN_INDIVIDUALS, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_KEEP_UNARY_MUTUALLY_EXCLUSIVE);
+    tsk_treeseq_free(&simplified);
+
     ret = tsk_treeseq_simplify(&ts, sample_ids, 2, TSK_KEEP_UNARY, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
+    tsk_treeseq_free(&simplified);
+
+    ret = tsk_treeseq_simplify(
+        &ts, sample_ids, 2, TSK_KEEP_UNARY_IN_INDIVIDUALS, &simplified, NULL);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
     tsk_treeseq_free(&simplified);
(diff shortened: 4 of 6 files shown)


hyanwong:
If we want to get this into a C API release version so we can release SLiM 3.5.1, what do we need to do @benjeffery ?

benjeffery:
> If we want to get this into a C API release version so we can release SLiM 3.5.1, what do we need to do @benjeffery ?

I think we can push out a C API release soon after this is merged - we can hold #866 till after as it is still in progress anyway,

c/tskit/tables.h
```diff
@@ -2671,6 +2672,9 @@ TSK_KEEP_INPUT_ROOTS
     By default simplify removes all topology ancestral the MRCAs of the samples.
     This option inserts edges from these MRCAs back to the roots of the input
     trees.
+TSK_KEEP_UNARY_IN_INDIVDUALS
+    This acts like TSK_KEEP_UNARY (and is mutually exclusive with that flag). It
+    keeps unary nodes, but only the unary node is referenced from an individual.
```
jeromekelleher, line 2677: only if

jeromekelleher (approved):
LGTM! One nit.

c/tskit/tables.h
hyanwong: Corrected. Well spotted.

hyanwong:
> LGTM! One nit.

Thanks. I didn't know if we wanted to reorder the bits passed as flags to `simplify()` so that the two TSK_KEEP_UNARY bits were together? I don't know if the actual bit used is purely an internal thing or not, but I though better not to rearrange for the moment?

c/tskit/tables.c
```diff
@@ -6094,6 +6094,8 @@ simplifier_print_state(simplifier_t *self, FILE *out)
     fprintf(out, "\tkeep_unary              : %d\n", !!(self->options & TSK_KEEP_UNARY));
     fprintf(out, "\tkeep_input_roots        : %d\n",
         !!(self->options & TSK_KEEP_INPUT_ROOTS));
+    fprintf(out, "\tkeep_unary_in_indivifuals : %d\n",
```
petrelharp, line 6097: "indivifuals"

jeromekelleher:
> I don't know if the actual bit used is purely an internal thing or not, but I though better not to rearrange for the moment?

I don't think there's much point in changing the bit values - someone might be depending on them, and there isn't really much to be gained from having them contiguous (and when we tag 1.0 we'll have to get used to these compromises anyway).

petrelharp:
LGTM, and it's very straightforward, although how do we feel about it not really being tested for correctness? I'd say that this implementation is "obviously correct", but changes down the line to simplify's algorithm might break that.

c/tskit/tables.c
hyanwong: Oops 😊

hyanwong:
> LGTM, and it's very straightforward, although how do we feel about it not really being tested for correctness? I'd say that this implementation is "obviously correct", but changes down the line to simplify's algorithm might break that.

Correctness-wise don't think it's much different from KEEP_UNARY, is it? There are a few basic tests I added in the suite, but I imagine this might get tested more thoroughly (? for correctness) if it is implemented in the Python API too?

petrelharp:
> I imagine this might get tested more thoroughly (? for correctness) if it is implemented in the Python API too?

Right, I'm wondering if we need to actually do this.

hyanwong:
> > I imagine this might get tested more thoroughly (? for correctness) if it is implemented in the Python API too?
> 
> Right, I'm wondering if we need to actually do this.

I think not, because the new functionality is simply a slightly restricted version of keep_unary. I'm not sure how it could go seriously wrong?

petrelharp:
Right now, I agree - I don't see how it could go wrong. But, suppose that next month, someone else makes some other change to simplify that breaks this. They don't catch it becaue they add tests for the functionality that they've implemented, and assume that everything else (e.g., this) is well-tested elsewhere.

hyanwong:
But I have tested a (admittedly single tree) case where one unary node has an individual and another one doesn't, and I'd checked the number of nodes that are expected to remain after the KEEP_UNARY_IN_INDIVIDUALS flag. What other tests might you be thinking of?

jeromekelleher:
I think the ``test_simplest_unary_with_individuals`` does test this for correctness @petrelharp, doesn't it? I don't think we need to go down the implement-in-Python path, it's quite a simple addition, and the tests do fail if it's not implemented correctly (right @hyanwong?)

hyanwong:
> the tests do fail if it's not implemented correctly (right @hyanwong?)

Yes, they do fail if its not implemented, or not implemented correctly.

jeromekelleher:
As a follow up,  we should open an issue to track adding the functionality in Python, though, so we don't forget about it.

hyanwong:
> As a follow up, we should open an issue to track adding the functionality in Python, though, so we don't forget about it.

Now at https://github.com/tskit-dev/tskit/issues/1120

petrelharp:
> But I have tested a (admittedly single tree) case where one unary node has an individual and another one doesn't, and I've checked the number of nodes that are expected to remain after the KEEP_UNARY_IN_INDIVIDUALS flag. What other tests might you be thinking of?

Well, asyou say, there's tests for (a) equality of tables where the behavior is the same as KEEP_UNARY (b) equality of tables where it's the same as just plain simplify, and (c) for [equality of number of rows](https://github.com/tskit-dev/tskit/pull/1119/files#diff-ba4ed4f290b38255241a3267411ce136dd44b6477cbe0cbb3c39341948fd96f6R1054) where KEEP_UNARY_IN_INDIVIDUALS has new behavior. At minimum I'd compare to the tables that are expected, rather than just counting the number of rows.  It'd also be nice to put in a few more odd situations, like an individual on a unary node above the root; an individual on a unary node above a single sample, unary nodes without individuals in both situations, a case where a node was not unary but becomes unary because of simplification, and a case where an individual has two nodes, and one will be retained because of this but the other not. Like so:

ex

```
    const char *nodes = "1  0   0   -1\n"
                        "1  0   0   0\n"
                        "0  1   0   -1\n"
                        "0  1   0   1\n"
                        "0  2   0   -1\n"
                        "0  3   0   -1\n"
                        "0  3   0   2\n"
                        "0  1   0   -1\n"
                        "0  1   0   3\n"
                        "0  0   0   0\n"
                        "0  0   0   0\n"
                        "0  1   0   3\n";
    const char *edges = "0  2   2   0\n"
                        "0  2   3   1\n"
                        "0  2   4   2,3\n"
                        "0  1   5   4\n"
                        "0  1   6   4\n"
                        "2  3   7   0\n"
                        "2  3   8   1,9\n"
                        "2  3   11   10\n";
    const char *individuals = "0    2.5\n"
                              "0    1.5,3.1\n"
                              "0    1.1\n"
                              "0    1.9\n";
```

hyanwong:
> > But I have tested a (admittedly single tree) case where one unary node has an individual and another one doesn't, and I've checked the number of nodes that are expected to remain after the KEEP_UNARY_IN_INDIVIDUALS flag. What other tests might you be thinking of?
> At minimum I'd compare to the tables that are expected, rather than just counting the number of rows. 

Yes, fair point. Or maybe, simply test that the nodes are the expected ones (e.g. return the new node ids)

> It'd also be nice to put in a few more odd situations, like an individual on a unary node above the root; an individual on a unary node above a single sample, unary nodes without individuals in both situations, 

Shouldn't we then test the normal KEEP_UNARY in this way too. Or if we do (I haven't checked), why would it be any different with KEEP_UNARY_IN_INDIVIDUALS

> a case where a node was not unary but becomes unary because of simplification, and a case where an individual has two nodes, and one will be retained because of this but the other not. 

Yep, I can see that I guess.

petrelharp:
I think this is the expected result:
```
    const char *nodes = "1  0   0   -1\n"
                        "1  0   0   0\n"
                        "0  1   0   1\n"
                        "0  1   0   2\n"
                        "0  2   0   -1\n"
                        "0  3   0   3\n"
    const char *edges =
                        "0  2   2   1\n"
                        "2  3   3   1\n"
                        "0  2   4   0,2\n"
                        "1  2   5   4\n";
    const char *individuals = "0    2.5\n"
                              "0    1.5,3.1\n"
                              "0    1.9\n"
                              "0    1.1\n";
```

petrelharp:
That wasn't quite right, but I've got it sorted out now.

jeromekelleher:
Is this ready to go @petrelharp? It would be good to get this in and ship the next C API release so we can get this into SLiM and so we don't need to hold up #1125.

(Although, tracking individual parents properly would be very nice to be able to do in SLiM too - would it be worth pulling all this into a C API release and updating SLiM then?)

hyanwong:
I think we concluded in https://github.com/MesserLab/SLiM/issues/139#issuecomment-762900817 that on the principle of least harm we would default to setting KEEP_UNARY_IN_INDIVIDUALS to `false` by default in SLiM simplification. So SLiM 3.5.1 is not actually waiting up on this. Perhaps that argues for waiting until the individual parents stuff is merged too?

petrelharp:
I think this is ready to merge - I was just waiting for someone (@hyanwong?) to look at my additional test case. I think it's good, though - merge if you like. We should discuss the C/SLiM release on slack.

hyanwong:
>  I was just waiting for someone (@hyanwong?) to look at my additional test case. 

Oh sorry. I didn't realise you were waiting on me. Merge away.

benjeffery (approved):

benjeffery:
I am going to squash the commits and merge.

bhaller:
> I think we concluded in [MesserLab/SLiM#139 (comment)](https://github.com/MesserLab/SLiM/issues/139#issuecomment-762900817) that on the principle of least harm we would default to setting KEEP_UNARY_IN_INDIVIDUALS to `false` by default in SLiM simplification. So SLiM 3.5.1 is not actually waiting up on this. Perhaps that argues for waiting until the individual parents stuff is merged too?

As I understood the situation, SLiM 3.5.1 is waiting on this because it seemed better not to have to roll a 3.5.2 release with this feature, or to have the feature delayed until SLiM 3.6 (whatever that turns out to be) happens, which will probably be months down the road.  Anyway, I'm not sure where the right place is now to discuss this with you and @petrelharp – probably not on a merged PR! – but let's figure this out.  :->

hyanwong:
> As I understood the situation, SLiM 3.5.1 is waiting on this 

Actually, I think it's not. It's only waiting on my reworking of the pyslim docs to include "retaining" individuals. That doesn't need this PR in tskit.

> Anyway, I'm not sure where the right place is now to discuss this with you and @petrelharp – probably not on a merged PR! 

Indeed. I think the link above, https://github.com/MesserLab/SLiM/issues/139 is fine.

benjeffery:
Is that a definitive answer? Happy to tag a release if SLiM does need this. Next release after that will likely be a couple of weeks looking at the milestone.

hyanwong:
No hurry for a release from me any more (now that Peter and Ben argued me out of setting `KEEP_UNARY_IN_INDIVIDUALS` as the default). So punt it off for a couple of weeks and go and get yourself a G&T instead.

bhaller:
@hyanwong I don't understand your read on the situation.  Could you explain why what I wrote above is incorrect?

bhaller:
Ah, I see the discussion has moved to Slack...

Issue #117, opened 2019-02-07
Sort should support sorting migrations with empty suffix


jeromekelleher:
It should be possible to pass a set of tables with > 0 migrations to sort but where the bookmark says the chunk to sort is empty. This allows us to workaround issues where we have a sorted set of migrations that we want to pass to sort.

See also #22.

Issue #20, opened 2018-02-02
Support migrations in simplify


jeromekelleher:
Simplify current ignores migration records. We need to update simplify so that migrations are also processed. This should be simple to do; we just use the node ID map to translate the node IDs in the old table into new node IDs. If the old node ID maps to -1, we leave the migration out.

hyanwong:
Ping @Chris1221 who was interested in taking a look at this. I'm not sure how complex an issue this would be for someone unfamiliar with the codebase, but Jerome's comment at the top gives me hope that it's not too bad

jeromekelleher:
Not the ideal place to start, as simplify is pretty hairy and multi-layered. But,  would be a great way to learn the codebase and how the top-to-bottom development process works.

Chris1221:
Perfect, thank you @hyanwong! Yes, probably won't start here, but I'll keep it on my radar. This feature would be particularly useful for the `smcsmc` conversion to `tskit`, which is why I was interested in it. But you're right @jeromekelleher, it is probably pretty hairy.

jeromekelleher:
It's not so bad @Chris1221 --- I'd be happy to help you get up to speed. It'd be great to get smcsmc converted to tskit!

Chris1221:
Yea I think so too. I was talking to @hyanwong a little about it, but we've implemented a very hacky conversion to the basic tables in text format -- it needs quite a bit of refining but the major issue was that we treat migration events as nodes, which is different than `tskit`. I'm going to look at the `tsconvert` work done so far and see if I can do it in a better way. I think it's pretty crucial to get everything in comparable formats, then we can compare apples to apples for the inference methods.

jeromekelleher:
>  I think it's pretty crucial to get everything in comparable formats, then we can compare apples to apples for the inference methods.

Amen brother! I think it'd be well worth while spending a few days with us at the BDI then, I'm sure we can iron out the details between us.

Chris1221:
That sounds great! I'm going home for Christmas soon, so realistically I don't think I'll have enough time in the next week or so, but we should try to find some time early in the new year. It's pretty high on my to-do list for `smcsmc`. Having it tied into the `tskit` ecosystem will also make it more easy for users to approach the output (and better integrated into `stdpopsim`).

Chris1221:
I'm going to take a look at this in the next few days. If anyone else has been working on it, let me know!

jeromekelleher:
[...]

PR #1129, opened 2021-01-23
Support migrations in keep_intervals, delete_intervals, and trim

### LINKED ISSUE #22: sort_tables should sort migrations
Migrations should be sorted by time by ``sort_tables``. See also #389.

### LINKED ISSUE #1128: Support for migrations in keep_intervals, delete_intervals, and trim
I've been doing simulations with migration recently and would like to make `keep_intervals`, `delete_intervals`, and `trim` compatible with tree sequences containing migrations. I'd like to tackle support for migrations in `simplify` later, so with these changes `keep_intervals` and `delete_intervals` would still fail if `simplify=True` is passed.

awohns:
## Description
Adds support for running `keep_intervals`, `delete_intervals`, and `trim` on tree sequences that contain migrations. I've modified the python api for these functions to support migrations and written some tests. However, `keep_intervals` calls `sort`, and my implementation of sorting the migration table isn't worked out yet, I would appreciate any pointers, so to speak!

# Addresses the following issues
Closes #1128. If sorting the migration table is worked out, would close #22, and could be modified to address #117

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2886,16 +2886,17 @@ class TableCollection:
             in increasing order. The list of intervals must be interpretable as a
             2D numpy array with shape (N, 2), where N is the number of intervals.
         :param bool simplify: If True, run simplify on the tables so that nodes
-            no longer used are discarded. (Default: True).
+            no longer used are discarded. Must be ``False`` if input tree sequence
+            includes migrations. (Default: True).
         :param bool record_provenance: If ``True``, add details of this operation
             to the provenance table in this TableCollection. (Default: ``True``).
         """
         intervals = util.intervals_to_np_array(intervals, 0, self.sequence_length)
-        if len(self.migrations) > 0:
-            raise ValueError("Migrations not supported by keep_ and delete_ intervals")
 
         edges = self.edges.copy()
         self.edges.clear()
+        migrations = self.migrations.copy()
+        self.migrations.clear()
         keep_sites = np.repeat(False, self.sites.num_rows)
         for s, e in intervals:
             curr_keep_sites = np.logical_and(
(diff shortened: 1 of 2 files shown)


jeromekelleher:
@awohns, the sort stuff is a can of worms so I've opened a separate PR to do it. Can you have a look at #1131 please? Hopefully you can just drop the last commit here once that's done and then merging will be straightforward.

awohns:
Thanks @jeromekelleher. #1131 looks good to me, I added one small comment about documenting the sorting of the migration table, to be consistent with the other tables' documentation

jeromekelleher:
@awohns, this is ready to pick up again I think. We've merged support for sorting migrations.

awohns:
Ok, I've rolled back my C changes to sort and tests pass so I think this is ready for review!

jeromekelleher (approved):
LGTM!

jeromekelleher:
@hyanwong, can you have a quick look over this please to see if there's anything we've missed?

python/tskit/tables.py
```diff
@@ -2886,16 +2886,17 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
             in increasing order. The list of intervals must be interpretable as a
             2D numpy array with shape (N, 2), where N is the number of intervals.
         :param bool simplify: If True, run simplify on the tables so that nodes
-            no longer used are discarded. (Default: True).
+            no longer used are discarded. Must be ``False`` if input tree sequence
```
hyanwong, line 2889: I guess this is only because `simplify` itself does not support migrations? So should we note to change this when https://github.com/tskit-dev/tskit/issues/20 is fixed? Perhaps we should add a test of `trim()` with simplify=True on  a ts with migrations, that is currently expected to fail, and comment at the top of the test that once it starts passing, we need to change the docs?

python/tskit/tables.py
```diff
@@ -2925,7 +2938,13 @@ def keep_intervals(self, intervals, simplify=True, record_provenance=True):
 
     def _check_trim_conditions(self):
         if self.migrations.num_rows > 0:
-            raise ValueError("You cannot trim a tree sequence containing migrations")
+            if (np.min(self.migrations.left) < np.min(self.edges.left)) and (
+                np.max(self.migrations.right) > np.max(self.edges.right)
+            ):
+                raise ValueError(
```
hyanwong, line 2944: What can't we do this? I mean, we'd lose those migrations, right? But should we just issue a warning? Or should we say "this would lose information, so you have to deliberately remove it using delete_intervals([[trim_left, trim_right]]) first"?

hyanwong:
> @hyanwong, can you have a quick look over this please to see if there's anything we've missed?

I don't know much about migrations, but this looks like a simple change to me. 2 very minor comments above. I guess I don't need to approve or anything, though.

PR #1131, opened 2021-01-23
Support migrations in sort.

### LINKED ISSUE #22: sort_tables should sort migrations

### LINKED ISSUE #117: Sort should support sorting migrations with empty suffix

jeromekelleher:
Closes #22
Closes #117

(I've not done any docs because I can't see anywhere we've said this stuff wasn't supported)

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2019-2020 Tskit Developers
+ * Copyright (c) 2019-2021 Tskit Developers
  * Copyright (c) 2017-2018 University of Oxford
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -3507,9 +3507,8 @@ tsk_migration_table_dump_text(const tsk_migration_table_t *self, FILE *out)
     for (j = 0; j < self->num_rows; j++) {
         metadata_len = self->metadata_offset[j + 1] - self->metadata_offset[j];
         err = fprintf(out, "%.3f\t%.3f\t%d\t%d\t%d\t%f\t%.*s\n", self->left[j],
-            self->right[j], (int) self->node[j], (int) self->source[j],
-            (int) self->dest[j], self->time[j], metadata_len,
-            self->metadata + self->metadata_offset[j]);
+            self->right[j], self->node[j], self->source[j], self->dest[j], self->time[j],
+            metadata_len, self->metadata + self->metadata_offset[j]);
         if (err < 0) {
             goto out;
         }
@@ -3531,7 +3530,6 @@ tsk_migration_table_equals(const tsk_migration_table_t *self,
           && memcmp(self->source, other->source, self->num_rows * sizeof(tsk_id_t)) == 0
           && memcmp(self->dest, other->dest, self->num_rows * sizeof(tsk_id_t)) == 0
           && memcmp(self->time, other->time, self->num_rows * sizeof(double)) == 0;
-
     if (!(options & TSK_CMP_IGNORE_METADATA)) {
         ret = ret && self->metadata_length == other->metadata_length
               && self->metadata_schema_length == other->metadata_schema_length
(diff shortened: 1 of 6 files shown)


jeromekelleher:
@petrelharp I've added sort support for migrations here. What do you think of the sort key? Should we make the sort requirement for migrations stricter now? We'd need to update msprime to make sure things are output in this order, I think.

c/tskit/tables.c
```diff
@@ -4463,6 +4472,28 @@ cmp_edge(const void *a, const void *b)
...
 }
 
+static int
+cmp_migration(const void *a, const void *b)
+{
+    const migration_sort_t *ca = (const migration_sort_t *) a;
+    const migration_sort_t *cb = (const migration_sort_t *) b;
+
+    int ret = (ca->time > cb->time) - (ca->time < cb->time);
+    /* If time values are equal, sort by the source population */
```
awohns, line 4482: Perhaps worth noting the full sorting rules in `docs/data-model.rst`. Right now only the nondecreasing time value is listed.

awohns (approved):

c/tskit/tables.c
```diff
@@ -4463,6 +4472,28 @@ cmp_edge(const void *a, const void *b)
...
+    if (ret == 0) {
+        ret = (ca->source > cb->source) - (ca->source < cb->source);
```
benjeffery, line 4484: Would be nice to get these branches tested - I guess it means handcrafting some tests and that's a but painful?

benjeffery (approved):
LGTM, just one question on coverage.

petrelharp:
I vote to not add more sortedness requirements if we don't have a good algorithmic reason to. Plus, it's not obvious to me that `source` and `dest` are the natural next things to sort on - I would have thought `node` would come next? Or maybe `left`? So, I'd say we should *just* sort on `time` here, and add some more requirements if it turns out to be helpful.

jeromekelleher:
Well, we need to choose some keys to sort on to make things deterministic (which we definitely want, we had loads of problems in the past with different C implementations having slightly different sort orders). I chose the current ordering because that's what comes out of msprime. So, my vote would be to formalise the order that pops out naturally from msprime (starting [here](https://github.com/tskit-dev/msprime/blob/main/lib/msprime.c#L4276) - sort by (source, dest, left). Unless there's a good reason otherwise, we might as well go with an existing ordering.

Agreed on not adding any more sortedness requirements to the input, that seems like it's more trouble that it's worth.

petrelharp:
I'm a bit confused still: by "deterministic" you mean specifying a total order, right? Don't we need some *more* keys for that, since you could have more than one migration event with the same `time, source, dest`?

jeromekelleher:
> I'm a bit confused still: by "deterministic" you mean specifying a total order, right? Don't we need some more keys for that, since you could have more than one migration event with the same time, source, dest

Yeah, I mean a fully specified ordering, because the output of sort isn't deterministic otherwise. Ah, you're right ``(time, source, dest, left)`` isn't fully specified, we'd need ``(time, source, dest, left, node)`` because we could have multiple nodes migrating at once.

petrelharp:
Sure, sounds good. I guess that sorting of mutations can be skipped with the bookmark if it's a performance problem.  So, `sort` will make migrations sorted on all these things, but "valid tree sequence requirements" will still only require sortedness on `time`, right?

jeromekelleher:
Yep, agreed

jeromekelleher:
We should probably try to get #1108 in first, since @petrelharp has had to do a lot of messing around with it already. Unless you'd prefer to get this in first Peter? The updates I need to do are fairly small.

petrelharp:
I would love to not have to do more mucking around with #1108, yes.

benjeffery:
Is this ready for me to have a look over?

jeromekelleher:
No, not yet. Thanks for the ping though, will do ASAP. (Writing slides for a talk tomorrow atm!)

c/tskit/tables.c
jeromekelleher: We decided to not put any further constraints on the input requirements. The extra keys are to make sure that the sort is always fully specified.

jeromekelleher:
Ready for final review and merge here @benjeffery

benjeffery (approved):
Great stuff, annoyingly I can't find anything to comment on!

Issue #605, opened 2020-05-11
Site method to get genotypes


petrelharp:
Right now there doesn't seem to be an easy way to either (a) look up the genotype of a node at a particular site, or (b) get the array of genotypes, other than by iterating through the variants until you get there. Unless I'm just forgetting about the right way to do this?  This is an important thing to be able to do, it seems (and would be useful to me right now, in writing tests). How about
```
class Site:

def genotypes(nodes=<the samples>):
   '''
   Returns a numpy array of character strings giving the alleles carried by the listed nodes at this site.
   ''' 
```
Alternatively, this could be a method of the tree sequence, and also take a `position` argument.

jeromekelleher:
This is an expensive operation though, as we need to have the tree built. No reason we can't offer a TreeSequence method to do it though (making it a method of Site would imply a lot of cross-linkage - the site would need to know the TreeSequence).

Normally I imagine doing this sort of thing for every site, so sequentially going over them is a good pattern and efficient. What's your usage pattern here?

There is a ``samples`` argument to ``variants`` all right though, which *does* let you decode the state of non-sample nodes as well (IIRC).

petrelharp:
You're right, that my example was iterating over all the sites, in which case I could use `variants`. But, suppose that I want to look at the genotypes at a particular site (eg, I find the site with the largest allele freq difference between two pops); as far as I can tell we have no good way to do that.

jeromekelleher:
> You're right, that my example was iterating over all the sites, in which case I could use variants. But, suppose that I want to look at the genotypes at a particular site (eg, I find the site with the largest allele freq difference between two pops); as far as I can tell we have no good way to do that.

For this example, you'd need to be iterating over the trees to get the allele frequencies. Maybe we could make it a method of the Tree? That way it's explicit what the cost is. So, ``Tree.decode_genotypes(samples=None)``, where samples can be any node.

I don't like putting in methods that are unnecessarily inefficient. I've seen people do things like ``tree_list = [tree.at(site.position) for site in ts.sites()]``, which is horrible. I worry that we'd start seeing code like ``[site.genotypes() for site in ts.sites()]`` which would be hideously inefficient.

petrelharp:
> make it a method of the Tree

Sure, why not. I was a bit worried about getting *exactly* the right tree due to floating-point precision or something, but I guess you could do `ts.at(site.position).genotypes()`? It still feels kinda wierd, though.

[...]

PR #1132, opened 2021-01-23
add convenience data member samples for vargen

castedo:
This is a follow up to compile problem discussed in #slim Slack channel.

The `sample` data member of `tsk_vargen_t` was removed during the 'consitifcation' of the codebase. This breaks some code in SLiM and might break other code that expects this data member to be present. This PR should fix this breaking of backwards of compatibility.

HOWEVER, technically, the SLiM code that got broken probably doesn't need this data member and probably doesn't want to use using `tsk_vargen_t` in the way it's been using it which expects a `sample` data member. @bhaller is considering what change, *if any*, to make in SLiM based on this new info.

Regardless, this breaking of backwards compatibility begs the question whether this PR should get merged regardless of what SLiM does in this particular case.

This issue gets slightly more complicated regarding the renaming of the original data member to `alt_samples` but I think thsi is actually an orthogonal issue. The `alt_samples` data member is essentially a private implementation detail of vargen and it would be ideal if clients do not rely on it.

Additional tasks to likely add to this PR:

- [x] comments about what the heck is `alt_samples` etc...
- [x] some kind of indication that `alt_sample` should be private

### DIFF
--- a/c/tskit/genotypes.c
+++ b/c/tskit/genotypes.c
@@ -159,6 +159,7 @@ tsk_vargen_init(tsk_vargen_t *self, const tsk_treeseq_t *tree_sequence,
     memset(self, 0, sizeof(tsk_vargen_t));
 
     if (samples == NULL) {
+        self->samples = tsk_treeseq_get_samples(tree_sequence);
         self->num_samples = tsk_treeseq_get_num_samples(tree_sequence);
         self->sample_index_map = tsk_treeseq_get_sample_index_map(tree_sequence);
         num_samples_alloc = self->num_samples;
@@ -171,6 +172,7 @@ tsk_vargen_init(tsk_vargen_t *self, const tsk_treeseq_t *tree_sequence,
         if (ret != 0) {
             goto out;
         }
+        self->samples = samples;
         self->num_samples = num_samples;
         self->sample_index_map = self->alt_sample_index_map;
     }
--- a/c/tskit/genotypes.h
+++ b/c/tskit/genotypes.h
@@ -52,9 +52,8 @@ typedef struct {
     size_t num_samples;
     size_t num_sites;
     const tsk_treeseq_t *tree_sequence;
-    const tsk_id_t *sample_index_map;
-    tsk_id_t *alt_samples;
-    tsk_id_t *alt_sample_index_map;
+    const tsk_id_t *samples;          /* samples being used */
+    const tsk_id_t *sample_index_map; /* reverse index map being used */
     bool user_alleles;
     char *user_alleles_mem;
     size_t tree_site_index;
@@ -63,6 +62,9 @@ typedef struct {
     tsk_tree_t tree;
     tsk_flags_t options;
     tsk_variant_t variant;
+    // private: the following data members are not intended to be used externally
+    tsk_id_t *alt_samples; /* alternative subset of samples to use */
+    tsk_id_t *alt_sample_index_map;
 } tsk_vargen_t;
 
 int tsk_vargen_init(tsk_vargen_t *self, const tsk_treeseq_t *tree_sequence,


castedo:
The compile break was hit by upgrading SLiM to a newer tskit here:
https://github.com/MesserLab/SLiM/pull/147

bhaller:
I'll advocate in favor of adding a `samples` ivar to vargen.  It allows the code that uses a vargen to be agnostic about how the vargen was constructed, without jumping through hoops.  As I understand it, if I sent you a `tsk_vargen_t` in the mail (as a professor of mine used to say :->), and you opened the box and saw it there, how would you use it?  Without `samples`, you wouldn't know whether it was based on `ts->samples` or on alternative samples contained by `vg->alt_samples`, so you'd have to do a little dance: `if (vg->alt_samples) <use vg->alt_samples>; else <use ts->samples>;`.  This seems non-optimal to me; you should be able to just use whatever the vargen considers itself to be based upon, however it was constructed, rather than relying upon this internal implementation detail of `alt_samples` being NULL, etc.  You should be able to just do `<use vg->samples>;`.  Without this, uses of vargen are liable to hard-code whether they use `vg->alt_samples` or `ts->samples`, based upon knowledge of how the vargen was constructed; and then if the vargen gets constructed differently, or if the internal semantics of `alt_samples` change somehow, that assumption is suddenly exposed as a bug, perhaps in a completely different place in the code if the construction and the use of the vargen are separated.  (This is the relevance of the "somebody mailed you a vargen" thought experiment.)

That's my two cents.  :->

castedo:
> if I sent you a `tsk_vargen_t` in the mail (as a professor of mine used to say :->), and you opened the box and saw it there, how would you use it?

@bhaller Is the box carried by an African or European swallow?

bhaller:
Oh, I have one more thing to say.  I wonder whether the right fix might not be to rename `alt_samples` back to `samples`, guarantee that it always points to the samples (by setting it to `ts->samples` if alternative samples were not passed to the constructor), and keep a separate flag controlling whether the vargen owns `samples` (because they're alternative samples) or does not (because they're `ts->samples`)?  I have no particular opinion, but the "do I own this pointer?" flag seems like a more standard way of handling this problem in non-refcounted situations.

castedo:
That's how the old code worked, and should still work with `samples` HOWEVER it gets complicate with `sample_index_map` because vargen will actually build and write to the alternative index map and there are const and non-const pointers floating around. The tree sequence is always const. I suspect it will be cleaner to NOT take the older flag approach given the const and non-const pointer situation, but I'm going to re-look at the code and see whether I'm not seeing something cleaner.

bhaller:
> That's how the old code worked, and should still work with `samples`

This puzzles me, because when I changed SLiM to pass NULL instead of `ts->samples`, that caused `samples` to be NULL in the vargen, and SLiM crashed.  That's using tskit 0.3.0's code in SLiM, when `samples` still existed.  So it seems like `samples` was not, at that time, guaranteed to point to the samples whether "alternative" or not...?  Am I misunderstanding?

castedo:
You are correct that the old 'samples' was not guaranteed to point to samples. It has the same semantics as the current `alt_samples`.

I think my text "should still work with `samples`" is confusing wording.

I do see a clean way to update the code to use a flag and instead of pairs of const and non-const pointers. However I have to add const versions of the tsk_safe_free macro and many more lines of code so my thought is it's not a good quickfix for a 0.99 release. My inclination is to keep this PR a bare minimal 3 lines of new code and make a separate PR for the cleaner implementation, which can be done after 1.0. It's really more of an implementation detail.

bhaller:
Sure, whatever seems best.  Re: the current PR changes, does tskit nil out pointers when they're no longer used?  I.e., in the teardown for a vargen should `samples` be set to NULL?  Apart from that, seems good to me, and jumping through the extra hoops you describe sounds quite unnecessary (either now or after 1.0, really).  :->

jeromekelleher:
I haven't read through the detail of this, but I just wanted to drop in one data point quickly: the vargen API hasn't been documented and isn't "approximately 1.0" like the tables or trees APIs. I'm kind of hoping we won't actually need this class in the long run, and can instead add a ``get_variant(site_id)`` method or something similar to the tree class (see #605), which will be just as efficient, much more flexible, and do away with some annoying boilerplate.

So, given that we're hoping to drop this class before 1.0 is tagged, my vote would be to not worry too much about const correctness and to do whatever we need to avoid breaking people's code in the short term.

castedo:
@bhaller, regarding the nulling out of the const non-owned pointers during teardown (free) of tsk_varget_t, I'm torn. It could mitigate the difficulty of debugging invalid memory access and memory corruption bugs. However, non-pointer data members of tskit structures (e.g. count of number of elements, etc...) do not get zeroed out.
So nulling out const non-owned pointers might makes things even more confusing. So in this PR I've opted just to not add more lines. So the const non-owned pointers are treated like other data members in that they are all invalid and
should not be accessed.

As part of a hypothetical much larger change in tskit convention, I could imagine some sense in having a zero-it-all-out convention where in teardown/free a function is almost always called which zeros out the entire struct. But probably better discussed in a separate time and context than this PR.

castedo:
Ready for review. Overall, I've leaned towards only adding/changing a few lines.

There is one confusing detail worth pointing out: The semantics of this new `samples` data member is slightly different than the old `samples` data member. And the semantics of the old `samples` data member is exactly the same as the current `alt_samples` data member.

The new `samples` data member always points to a valid array of samples. It's whatever is getting used by `tsk_vargen_t`. That is, either the samples passed into `tsk_vargen_init` or the default which is all the samples of the tree sequence. In contrast, `alt_samples` and the old `samples` are NULL when no samples were explicitly passed to `tsk_vargen_init` (default behavior using all samples in the tree sequence).

@bhaller and I discussed in Slack and both of us think this makes sense, that `alt_samples` is essentially an implementation detail that should be internal to `tsk_vargen_t`. Why this change at all? It's a downstream side effect of making most tskit code const correct.                                                                Alternative approaches I've explored end up being more complicated.

bhaller:
> @bhaller, regarding the nulling out of the const non-owned pointers during teardown (free) of tsk_varget_t, I'm torn.

Whatever you think is best.  I wasn't proposing a change in tskit policy here; as you say, that is of course a separate, and bigger, issue, and I'm not advocating for it.  I was just doing a checkback on this PR, asking whether the `samples` ivar you're adding here is following current tskit policy regarding teardown of such state, since I do tend to nil out pointers on teardown in my code.  :->

benjeffery (approved):
Thanks for figuring this one out and coming up with a pragmatic fix. 
LGTM!

benjeffery:
@bhaller Do you need (another) C API release with this in for SLiM?

bhaller:
> @bhaller Do you need (another) C API release with this in for SLiM?

Yep, sorry, thanks.

bhaller:
Thanks @castedo, quick work.  :->

petrelharp:
> @bhaller Do you need (another) C API release with this in for SLiM?

Well, technically not, since we can just copy in whatever version of the C code we want. But it does seem more solid to say that it's a particular released version.

bhaller:
> > @bhaller Do you need (another) C API release with this in for SLiM?
> 
> Well, technically not, since we can just copy in whatever version of the C code we want. But it does seem more solid to say that it's a particular released version.

Yeah, it'd be best to stick to released tskit version to avoid confusion down the road, I think...

jeromekelleher (approved):

jeromekelleher:
Great, thanks for getting this fix turned around so quickly everyone! I think we can just go ahead and tag the next C API release now, there's no reason not to. Definitely agree with sticking to tagged releases for SLiM.

PR #1133, opened 2021-01-25
C_0.99.10 release prep

benjeffery:


c/CHANGELOG.rst
```diff
@@ -1,3 +1,12 @@
+----------------------
+[0.99.10] - 2021-01-25
+----------------------
+
+**Features**
+
+- Add ``samples`` to ``tsk_vargen_t``. (:user:`castedo`, :pr:`1132`).
```
jeromekelleher, line 7: I don't think we want to advertise this as a feature (since (a) vargen isn't documented and (b) slim shouldn't really be using these internal attrs anyway). I think we just call the release "Minor bugfix on internal APIs" and leave it at that.

jeromekelleher (approved):
LGTM, but I think we shouldn't document the actual change since it's an internal thing (which SLiM shouldn't be using, in an ideal world).

c/CHANGELOG.rst
benjeffery: Cool, I was thinking is wasn't normal changelog fodder - but didn't want an empty one. Changed to "Minor change to internal APIs".

jeromekelleher: But looks from my diff like the changes for 0.99.9 have been removed instead?

benjeffery: Blimey. Managed to sneak in a fix before mergify.

Issue #1111, opened 2021-01-07
Need a way to add entire rows to tables.


jeromekelleher:
When adding rows to tables we do a lot of things like
```python
tables.nodes.add_row(flags=node.flags, time=node.time, ...)
```

It would be simpler and more future proof if we had a keyword-only arg (that can't be supplied with any other argument), ``row`` which contains all the relevant values:

```python
tables.nodes.add_row(row=node)
```

Annoyingly, we can't do this because there's some ``add_row`` methods that have mandatory positional arguments, and the others have fixed numeric values for defaults in the signatures.

We could have a new method called ``add(row)`` which does this? I'm not that keen on the name though, it's really doing the same thing as ``add_row``, just being parameterised differently. 

I guess if we're going to take the Python list analogy further, we should probably call this method ``append``?

jeromekelleher:
Actually, ``append`` seems good. Any objections? This would save us a lot of annoying code, I think.

benjeffery:
Might as a well go the whole hog and also implement `extend` from the python list too?

jeromekelleher:
Yep, good idea.

petrelharp:
That *would* be nice. What kind of an object would `node` be?

jeromekelleher:
Anything that Duck-types to the required attributes, so we could either pass in a NodeTableRow or (I think) a Node.

jeromekelleher:
Similarly, we should add ``x_table_add_object(table, object)`` to the C API, i.e., ``tsk_individual_table_add_object(tsk_individual_table_t *table, individual_t *object)``. There's quite a few places we'd make the code cleaner with this.

PR #1143, opened 2021-01-26
Test suite updates for msprime 1.0

### LINKED ISSUE #1130: Fix test suite errors with msprime 1.0
Most of these seem to be around having times for mutations.

We should probably install the 1.0.0a3 release as part of CI now so that we don't get any surprises later.

jeromekelleher:
Closes #1130

This should be good to merge ASAP, as there's no real changes. Will need msprime 1.0.0a4 installed before it'll work, though, so don't be surprised if it breaks locally for devs for the moment.

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1772,7 +1772,8 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestCase):
         self.verify_basic_svg(svg, width=200 * ts.num_trees)
 
     def test_draw_integer_breaks_ts(self):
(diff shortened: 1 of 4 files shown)


benjeffery (approved):
LGTM! I see you are kwarg-ing the add row methods, as mentioned before we should consider making those kwarg only.

jeromekelleher:
> I see you are kwarg-ing the add row methods, as mentioned before we should consider making those kwarg only.

Possibly, but we don't want to break code either. I think trying to transition to using the ``append`` (#1111) methods is the best we can do - even if kwargs were set in this case, we'd still be missing any new attrs added.

PR #1146, opened 2021-01-27
Remove name from metadata example

### LINKED ISSUE #1071: Unused field in JSON schema example?
In the [metadata examples](https://tskit.readthedocs.io/en/latest/metadata.html#examples) docs we have the following example of the definition of a property:

```python
           collection_date": {
                "name": "Collection date",
                "description": "Date of sample collection in ISO format",
                 "type": "string",
                "binaryFormat": "10p",
                "pattern": "^([1-9][0-9]{3})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])?$",
            },
```

I can't find any reference to the ``name`` field here in JSON schema. Should this be the ``title`` from the [annotations](https://json-schema.org/understanding-json-schema/reference/generic.html#id2) section? In this case, I'm not sure it adds any explanatory power and we'd be better of leaving it out - the combination of the field name and description should be sufficient.

benjeffery:
Fixes #1071

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -273,7 +273,6 @@ to the individuals in a tree sequence:
             "properties": {
                 "accession_number": {"type": "integer", "binaryFormat": "i"},
                 "collection_date": {
-                    "name": "Collection date",
                     "description": "Date of sample collection in ISO format",
                     "type": "string",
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM

PR #1149, opened 2021-01-27
Use defaults when decoding JSON metadata

benjeffery:
A start on #1073 

So far this only does defaults on objects, and objects in objects, but does not do arrays. For example an array of objects could have a default key for all the instances in the array, or an array itself have a default value.

Note that where a deeper default exists it overrides a shallower one. This also means that if a shallower default of an object has many keys, and one of those keys has a deeper default, all of the shallow default will be ignored.

# PR Checklist:

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -515,6 +515,68 @@ class TestMetadataModule:
         assert ms.decode_row(ms.validate_and_encode_row(row_data)) == row_data
 
 
+class TestJSONCodec:
+    def test_simple_default(self):
+        schema = {
+            "codec": "json",
+            "type": "object",
+            "properties": {"number": {"type": "number", "default": 5}},
+        }
+        ms = tskit.MetadataSchema(schema)
+        assert ms.decode_row(ms.validate_and_encode_row({})) == {"number": 5}
+        assert ms.decode_row(ms.validate_and_encode_row({"number": 42})) == {
+            "number": 42
+        }
+
+    def test_nested_default(self):
+        schema = {
+            "codec": "json",
+            "type": "object",
+            "properties": {
+                "obj": {
+                    "type": "object",
+                    "properties": {
+                        "number": {"type": "number", "default": 5},
+                        "nested_obj": {
+                            "type": "object",
+                            "properties": {},
+                            "default": {"foo": "bar"},
+                        },
+                    },
+                }
+            },
+        }
+        ms = tskit.MetadataSchema(schema)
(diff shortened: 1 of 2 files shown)


python/tests/test_metadata.py
```diff
@@ -515,6 +515,68 @@ def decode(self, encoded):
...
+        schema = {
+            "codec": "json",
+            "type": "object",
+            "properties": {
+                "obj": {
+                    "type": "object",
+                    "properties": {
+                        "number": {"type": "number", "default": 5},
+                    },
+                    "default": {"number": 42},  # This will be ignored
```
jeromekelleher, line 572: Doesn't this default violate the schema anyway? 42 isn't an object, is it?

python/tskit/metadata.py
```diff
@@ -109,13 +110,33 @@ def register_metadata_codec(
 
 class JSONCodec(AbstractMetadataCodec):
     def __init__(self, schema: Mapping[str, Any]) -> None:
-        pass
+        # Recurse through the schema finding default values to fill in on decode
+        self.defaults = {}
+
+        def record_defaults(path, sub_schema):
```
jeromekelleher, line 116: Elegant!

python/tskit/metadata.py
```diff
@@ -109,13 +110,33 @@ def register_metadata_codec(
...
+
+        record_defaults((), schema)
 
     def encode(self, obj: Any) -> bytes:
         return tskit.canonical_json(obj).encode()
 
     def decode(self, encoded: bytes) -> Any:
-        return json.loads(encoded.decode())
+        result = json.loads(encoded.decode())
+        # Assign default values
```
jeromekelleher, line 130: I think it's logically simpler to start with the ``defaults`` dictionary, and write in anything that's in the decoded version, so
```python
result = copy.deepcopy(self.defaults)
recursive_update(result, decoded)
```

I agree this is made substantially trickier by the nestedness requirements.

I'm tempted to just implement this for top-level values initially (throwing an error if someone tries to set a default for a nested type), and see how far it gets us. It'll be a lot more work up front if we decide to support all possible combinations of nested defaults.

jeromekelleher (commented):
Looks good, and very elegant code. In the interest of not getting bogged down on details, though, I wonder if we should do something more limited in the first instance, though.

jeromekelleher:
Bumping my vote for just doing this on top-level stuff and raising an error if someone tries to do anything more complicated with defaults...

benjeffery:
Closed in favour of #1272

PR #1182, opened 2021-02-02
Svg improvements

### LINKED ISSUE #1155: SVG output: change mutation symbol to highlight sample nodes?
It's just struck me that in the SVG output we should really distinguish between sample nodes and other nodes. I think a neat way to do this would be to use a (black) square for sample nodes and change the mutations to perhaps a horizontal red line (or maybe a cross, if we want to be fancy) over the lineage. Thoughts?

hyanwong:
Make samples nodes square, use a translucent background for light/dark modes, and place line colours in the style file for ease of changing. Also use pytest tmp_path so we can look at saved output files.

Fixes #1155

# PR Checklist:

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -27,6 +27,7 @@ import collections
 import io
 import math
 import os
+import pathlib
 import re
 import xml.dom.minidom
 import xml.etree
@@ -204,6 +205,12 @@ class TestTreeDraw:
         return tskit.load_text(
             nodes, edges, sites=sites, mutations=mutations, strict=False
         )
+
+    def fail(self, *args, **kwargs):
+        """
+        Required for xmlunittest.XmlTestMixin to work with pytest not unittest
+        """
+        pytest.fail(*args, **kwargs)
 
 
 def closest_left_node(tree, u):
@@ -1434,11 +1441,11 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
             cls = group.attrib["class"]
             assert re.search(r"\broot\b", cls)
 
-    def test_draw_file(self, tmp_path):
-        # NB - deleting tmp_path dir unnecessary: this is done after 3 times by pytest
+    def test_draw_to_file(self, tmp_path):
+        # NB: to view output files for testing changes to drawing code, it is possible
+        # to save to a fixed directory using e.g. `pytest --basetemp=/tmp/svgtest ...`
         t = self.get_binary_tree()
         filename = tmp_path / "tree-draw.svg"
-        print(f"Saving tree.draw() to {filename}")  # `pytest -rP` shows path to SVG
         svg = t.draw(path=filename)
         assert os.path.getsize(filename) > 0
         with open(filename) as tmp:
@@ -1446,7 +1453,6 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         assert svg == other_svg
 
         filename = tmp_path / "tree-draw_svg.svg"
-        print(f"Saving tree.draw_svg() to {filename}")  # `pytest -rP` shows path to SVG
         svg = t.draw_svg(path=filename)
         assert os.path.getsize(filename) > 0
         with open(filename) as tmp:
@@ -1455,7 +1461,6 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         self.verify_basic_svg(other_svg)
 
         filename = tmp_path / "ts-draw_svg.svg"
-        print(f"Saving ts.draw_svg() to {filename}")  # `pytest -rP` shows path to SVG
         ts = self.get_simple_ts()
         svg = ts.draw_svg(path=filename)
         assert os.path.getsize(filename) > 0
@@ -1826,15 +1831,15 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         tables = ts.dump_tables()
         tables.edges.clear()
         ts_no_edges = tables.tree_sequence()
-        svg = ts_no_edges.draw_svg()  # This should just be a row of 10 circles
+        svg = ts_no_edges.draw_svg()  # This should just be a row of 10 sample nodes
         self.verify_basic_svg(svg)
-        assert svg.count("circle") == 10
-        assert svg.count("path") == 0
+        assert svg.count("rect") == 10  # Sample nodes are rectangles
+        assert svg.count('path class="edge"') == 0
 
         svg = ts_no_edges.draw_svg(force_root_branch=True)
         self.verify_basic_svg(svg)
-        assert svg.count("circle") == 10
-        assert svg.count("path") == 10
+        assert svg.count("rect") == 10
+        assert svg.count('path class="edge"') == 10
 
         # If there is a mutation, the root branches should be there too
         ts = msprime.mutate(ts, rate=1, random_seed=1)
@@ -1844,9 +1849,9 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         assert ts_no_edges.num_mutations > 0  # Should have some singletons
         svg = ts_no_edges.draw_svg()
         self.verify_basic_svg(svg)
-        assert svg.count("circle") == 10
-        assert svg.count("path") == 10
-        assert svg.count("rect") == ts_no_edges.num_mutations
+        assert svg.count("rect") == 10
+        assert svg.count('path class="edge"') == 10
+        assert svg.count('path class="sym"') == ts_no_edges.num_mutations
 
     def test_tree_root_branch(self):
         # in the simple_ts, there are root mutations in the first tree but not the second
@@ -1883,7 +1888,7 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         )
         # Prettify the SVG code for easy inspection
         svg = xml.dom.minidom.parseString(svg).toprettyxml()
-        svg_fn = os.path.join(os.path.dirname(__file__), "data", "svg", "tree.svg")
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "tree.svg"
         with open(svg_fn, "rb") as file:
             expected_svg = file.read()
         self.assertXmlEquivalentOutputs(svg, expected_svg)
@@ -1895,7 +1900,7 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         )
         # Prettify the SVG code for easy inspection
         svg = xml.dom.minidom.parseString(svg).toprettyxml()
-        svg_fn = os.path.join(os.path.dirname(__file__), "data", "svg", "mut_tree.svg")
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "mut_tree.svg"
         with open(svg_fn, "rb") as file:
             expected_svg = file.read()
         self.assertXmlEquivalentOutputs(svg, expected_svg)
@@ -1907,7 +1912,7 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         )
         # Prettify the SVG code for easy inspection
         svg = xml.dom.minidom.parseString(svg).toprettyxml()
-        svg_fn = os.path.join(os.path.dirname(__file__), "data", "svg", "ts.svg")
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "ts.svg"
         self.verify_basic_svg(svg, width=200 * ts.num_trees)
         with open(svg_fn, "rb") as file:
             expected_svg = file.read()
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -28,7 +28,6 @@ import io
 import math
 import os
 import re
-import tempfile
 import xml.dom.minidom
 import xml.etree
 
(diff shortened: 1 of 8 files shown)


python/tests/test_drawing.py
```diff
@@ -1913,6 +1919,32 @@ def test_known_svg_ts(self):
...
+        flags[n // 2 : n + n // 2] = tskit.NODE_IS_SAMPLE
+        nodes.flags = flags
+        ts = tables.tree_sequence()
+        svg = ts.draw_svg(
+            root_svg_attributes={"id": "123"},
+            style=".leaf .sym {fill: green} .sample .sym {fill: cyan}",
+        )
+        # Prettify the SVG code for easy inspection
+        svg = xml.dom.minidom.parseString(svg).toprettyxml()
+        svg_fn = os.path.join(
```
jeromekelleher, line 1940: pathlib is a bit nicer for this sort of thing: https://docs.python.org/3.7/library/pathlib.html

python/tskit/drawing.py
```diff
@@ -683,7 +693,26 @@ def draw(self):
 
             # Add node symbol + label next (visually above the edge subtending this node)
             # Symbols
-            curr_svg_group.add(dwg.circle(**self.node_attrs[u]))
+            if self.tree.is_sample(u):
+                curr_svg_group.add(
+                    dwg.rect(
+                        **{
```
jeromekelleher, line 699: This a lot of vspace. Why not create the args dict in a variable rather than doing everything in one call?

python/tskit/drawing.py
```diff
@@ -460,11 +461,21 @@ def __init__(
...
         self.node_attrs = {}
         self.node_label_attrs = {}
+        symbol_size = "{:g}".format(rnd(self.symbol_size))
+        half_symbol_size = "{:g}".format(rnd(self.symbol_size / 2))
         for u in tree.nodes():
             self.edge_attrs[u] = {}
             if edge_attrs is not None and u in edge_attrs:
                 self.edge_attrs[u].update(edge_attrs[u])
-            self.node_attrs[u] = {"r": "{:g}".format(rnd(self.symbol_size / 2))}
+            self.node_attrs[u] = {
```
jeromekelleher, line 470: This is a bit obscure - wouldn't it be easier to have a ``sample_node_attrs`` variable or something? Having to demultiplex the information and filter the non-relevant bits out is pretty messy and error prone (i.e, what happens when someone adds an extra attr to the circle? )

python/tskit/trees.py
```diff
@@ -1462,7 +1462,12 @@ def draw_svg(
         **kwargs,
     ):
         """
-        Return an SVG representation of a single tree.
+        Return an SVG representation of a single tree. Sample nodes are represented as
+        black squares, other nodes are black circles, and mutations are red crosses,
+        although these default styles can be altered (see below). By default, numeric
+        labels are draqwn beside nodes and mutations: these can be altered using the
```
jeromekelleher, line 1468: draqwn

python/tests/test_drawing.py
```diff
@@ -1435,34 +1440,35 @@ def verify_basic_svg(self, svg, width=200, height=200):
             cls = group.attrib["class"]
             assert re.search(r"\broot\b", cls)
 
-    def test_draw_file(self):
+    def test_draw_file(self, tmp_path):
+        # NB - deleting tmp_path dir unnecessary: this is done after 3 times by pytest
```
jeromekelleher, line 1444: Don't think we need this comment

python/tests/test_drawing.py
```diff
@@ -1435,34 +1440,35 @@ def verify_basic_svg(self, svg, width=200, height=200):
...
-            svg = ts.draw_svg(path=filename)
-            assert os.path.getsize(filename) > 0
-            with open(filename) as tmp:
-                other_svg = tmp.read()
-            self.verify_basic_svg(svg)
-            self.verify_basic_svg(other_svg)
-        finally:
-            os.unlink(filename)
+        filename = tmp_path / "tree-draw.svg"
+        print(f"Saving tree.draw() to {filename}")  # `pytest -rP` shows path to SVG
```
jeromekelleher, line 1447: I'm not keen on putting in prints. Won't this spam the output everytime someone runs ``pytest -vs``?

jeromekelleher (commented):
LGTM, minor comments.

python/tskit/drawing.py
hyanwong: Yeah, it was stuck in by black. I can use a variable.

python/tskit/drawing.py
hyanwong: Yes, it's a bit fiddly. But I assume that most people will *want* the same thing applied to both sample nodes and other nodes (e.g. `"stroke":"black", "fill":"white"`). I think all the other attributes are identical across rects and circles. Would we then have an extra param in the API (`sample_node_attrs=None`), and even a `sample_node_label_attrs=None` to match? Seems like it's growing a bit out of control.

Or do you mean we do this internally, but don't expose it to the user?

python/tests/test_drawing.py
hyanwong: >Won't this spam the output everytime someone runs `pytest -vs`?

It doesn't seem to, no (try it). I think stdout is normally captured by something else. The `-rP` is required to get anything out (it took me a while to find that). It's pretty useful being able to look at the files that have been created by the test suite, as they are in very obscure directories (e.g. `/private/var/folders/00/l36df9q14hv13hhw8wglncch0000gr/T/pytest-of-yan/pytest-541/popen-gw3/test_draw_file0/ts-draw_svg.svg`) so finding them is hard.

python/tskit/drawing.py
jeromekelleher: I mean internally, so the code is clear.

python/tests/test_drawing.py
jeromekelleher: It does. Run with ``-n0 -vs``, and it prints all outputs as they are generated. I use this all the time.

``-rP`` just means "show the output" from passed tests (which is only useful to you now because I've nagged about leaving stray prints in innumerable reviews).

The location is pretty logical for Linux users ( /tmp/pytest-of-jk/pytest-13/test_draw_file0/), and you can override the default location if you want with [--basetmpdir](https://docs.pytest.org/en/stable/tmpdir.html#the-default-base-temporary-directory).

So yes, please delete the prints.

hyanwong: Ah, it requires the `-n0` to print, I see.

I can delete. I would like to leave a comment about --basetmpdir somewhere though, as that'll be super useful for testing, and I'm bound to forget. Perhaps in a comment at the top of that test.

jeromekelleher: Sure - put in whatever comments you like (keeping the prints commented out, seems like something that'll handy in future, too).

hyanwong:
All suggestions addressed. I also took the opportunity to make the internal_samples_examples into fixtures so we can reuse them in different tests (and added some SVG tests to the examples for internal samples). That's a separate commit (the last one) in the PR, as it touches a few other places too.

python/tests/conftest.py
```diff
@@ -120,6 +120,62 @@ def ts_fixture():
...
+    nodes = tables.nodes
+    flags = nodes.flags
+    # Set a mixture of internal and leaf samples.
+    flags[:] = 0
+    flags[n // 2 : n + n // 2] = tskit.NODE_IS_SAMPLE
+    nodes.flags = flags
+    return tables.tree_sequence()
+
+
+@fixture(scope="session")
```
jeromekelleher, line 168: Silently iterating over examples within a test is an anti-pattern now, we shouldn't do it for new code. We don't want a fixture that contains multiple examples, we want to parametrise the test with multiple examples.

python/tests/test_drawing.py
```diff
@@ -1907,12 +1915,32 @@ def test_known_svg_ts(self):
...
         # Prettify the SVG code for easy inspection
         svg = xml.dom.minidom.parseString(svg).toprettyxml()
-        svg_fn = os.path.join(os.path.dirname(__file__), "data", "svg", "ts.svg")
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "ts.svg"
         self.verify_basic_svg(svg, width=200 * ts.num_trees)
         with open(svg_fn, "rb") as file:
             expected_svg = file.read()
         self.assertXmlEquivalentOutputs(svg, expected_svg)
 
+    def test_known_svg_internal_samples_ts(self, internal_samples_examples):
```
jeromekelleher, line 1924: Re the last comment, what if we did
```
@pytest.mark.parametrize("ts", internal_samples_examples)
def  def test_known_svg_internal_samples_ts(self, ts):
...
```

Parametrising is much better because it tells you *which* example has failed and lets you make things much more modular.

python/tests/test_highlevel.py
```diff
@@ -153,6 +153,7 @@ def get_gap_examples():
 def get_internal_samples_examples():
     """
     Returns example tree sequences with internal samples.
+    TODO - replace with internal_samples_examples fixtures in conftest.py
```
jeromekelleher, line 156: Delete

jeromekelleher (commented):
Looking good, but if we're going to refactor the test cases we're going to have to spend more time coming up with a good pattern that parametrises the tests with the examples. Half doing this is worse than not doing it all because we'll then have two different patterns to search and replace for when things are ultimately refactored to use pytest.mark.parametrize.

python/tests/conftest.py
hyanwong: Oh yes, you mean something like this (I've been porting something similar into SLiM / pyslim, but using dicts, rather than a class)

```
class TreeSequenceExamples:
    """A simple container for example tree seqs - seems a bit unnatural though?"""
    @classmethod
    def example1(cls):
        ...
        return ts

    @classmethod
    def example2(cls):
        ...
        return ts

    @classmethod
    def list_examples_1_and_2(cls):
        return ["example1", "example2"]

    @classmethod
    def list_all_examples(cls):
        return [meth for meth in dir(cls) if meth.startswith "example"]


@pytest.fixture(scope="session", params=TreeSequenceExamples.list_all_examples())
def example_ts(request):
     return getattr(TreeSequenceExamples, request.param)(TreeSequenceExamples)

# In test file

@pytest.mark.parametrize('example_ts', argvalues=['example1', 'example2'])
def test_my_examples(example_ts):
    # do something with example_ts

# or equivalantly

@pytest.mark.parametrize('example_ts', argvalues=TreeSequenceExamples.list_examples_1_and_2())
def test_my_examples(example_ts):
    # do something with example_ts
```

Just a rough idea. You or @benjeffery can probably come up with a more elegant scheme?

python/tests/test_drawing.py
hyanwong: Yes, you're absolutely right.

python/tests/conftest.py
hyanwong: Oh, I guess rather than using a class, which seems a bit hacky, we could put all the examples into a file , and `import` that into `conftest.py`?

benjeffery: In this particular case I would parameterise the fixture:
```python
n = 5
@fixture(scope="session",
         params=[('all', slice(None,None)), ('internal', slice(n,None)), ('mixed', slice(n // 2, n + n // 2))],
         ids=["all", "internal", "mixed"])
def node_samples_example(request):
    name, slice = request.param
    ts = msprime.simulate(n, random_seed=10, mutation_rate=5)
    assert ts.num_mutations > 0
    tables = ts.dump_tables()
    tables.metadata_schema = tskit.MetadataSchema({"codec":"json"})
    tables.metadata = {"name": name}
    nodes = tables.nodes
    flags = nodes.flags
    flags[:] = 0
    flags[slice] = tskit.NODE_IS_SAMPLE
    nodes.flags = flags
    return tables.tree_sequence()
```
The metadata is there so you can do `f"internal_sample_ts_{ts.metadata['name']}.svg"` in the test.

benjeffery: Are these fixtures used elsewhere? Only widley used ones should be in `conftest.py`.

python/tests/test_drawing.py
```diff
@@ -1435,34 +1441,36 @@ def verify_basic_svg(self, svg, width=200, height=200):
...
-            svg = ts.draw_svg(path=filename)
-            assert os.path.getsize(filename) > 0
-            with open(filename) as tmp:
-                other_svg = tmp.read()
-            self.verify_basic_svg(svg)
-            self.verify_basic_svg(other_svg)
-        finally:
-            os.unlink(filename)
+        filename = tmp_path / "tree-draw.svg"
+        # print(f"Saved tree.draw() to {filename}")  # `pytest -rP` shows path to SVG
```
benjeffery, line 1449: I think the comment above makes all these `print`s redundant.

python/tests/conftest.py
jeromekelleher: I don't know, I'd rather not spend time investigating it right now. Can we just keep this to what we need for the change in question?

benjeffery (commented):
Looks good @hyanwong, I've made a proposal for how a parameterised fixture could work.

python/tests/conftest.py
hyanwong: I'm just sticking stuff into another (non urgent) fixture-related PR. For the moment I'll just revert the last commit in this PR and can add the extra tests when the other (fixture) PR eventually gets merged. No point working something out only to have to rip it out again, and these tests are only niceties anyway.

python/tests/test_drawing.py
hyanwong: Re being used elsewhere, these functions are only used in 2 other files, one of which (test_highlevel.py) is a bit hard-coded so would take some work to convert to fixtures.

hyanwong:
I dropped the fixture stuff and simply don't test the look of SVGs with internal sample nodes at the moment. The tests can be reinserted once something like https://github.com/tskit-dev/tskit/pull/1185 eventually gets merged. No urgency for that though. This can now be merged as-is and additional tests added later.

python/tests/test_drawing.py
hyanwong: (print statements removed)

jeromekelleher (approved):

benjeffery:
@hyanwong CI problem was a transient github issue.

hyanwong:
Can this be merged @benjeffery - I'm doing some graphics descriptions today, so it might be helpful to refer to the new look.

benjeffery (approved):
Yep, LGTM! Can you file an issue for the missing tests?

PR #1183, opened 2021-02-02
Clarify keep_input_roots docs

### LINKED ISSUE #1151: Clarify behaviour of simplify with keep_input_roots=true and unary nodes
The documentation for keep_input_roots is misleading in the presence of unary nodes. We need to add a clarification like

> if keep_unary is True the path of unary nodes and edges from the MRCA to the root will also be retained.

hyanwong:
Fixes #1151, with @petrelharp 's rewording (I changed "root" to MRCA though, which I think was a simple typo in his suggestion)

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -2541,9 +2541,11 @@ class TableCollection:
         :param bool keep_unary: If True, any unary nodes (i.e. nodes with exactly
             one child) that exist on the path from samples to root will be preserved
             in the output. (Default: False)
-        :param bool keep_input_roots: If True, insert edges from the MRCAs of the
-            samples to the roots in the input trees. If False, no topology older
-            than the MRCAs of the samples will be included. (Default: False)
+        :param bool keep_input_roots: Whether to retain history ancestral to the
+            MRCA of the samples. If ``False``, no topology older than the MRCAs of the
+            samples will be included. If ``True`` the roots of all trees in the returned
+            tree sequence will be the same roots as in the original tree sequence.
+            (Default: False)
         :param bool record_provenance: If True, record details of this call to
             simplify in the returned tree sequence's provenance information
             (Default: True).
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

python/tskit/tables.py
```diff
@@ -2541,9 +2541,10 @@ def simplify(
         :param bool keep_unary: If True, any unary nodes (i.e. nodes with exactly
             one child) that exist on the path from samples to root will be preserved
             in the output. (Default: False)
-        :param bool keep_input_roots: If True, insert edges from the MRCAs of the
-            samples to the roots in the input trees. If False, no topology older
-            than the MRCAs of the samples will be included. (Default: False)
+        :param bool keep_input_roots: Whether to retain history ancestral to the
+            MRCA of the samples. If False, no topology older than the MRCAs of the
+            samples will be included; the roots of all trees in the returned tree
+            sequence will be MRCAs in the original tree sequence. (Default: False)
```
petrelharp, line 2547: this MRCA should be "root", I think.

hyanwong: Oh, did I miscorrect you, sorry. You said "the roots of all trees will be roots in the original tree sequence", which I find difficult to parse, and I think (?) is wrong isn't it? If `False`, the roots in the returned TS will be the MRCAs in the original TS. If `True`, the roots in the returned TS will be the same roots as in the original TS. I think, anyway?

hyanwong: Do you mean, perhaps

> Whether to retain history ancestral to the
> MRCA of the samples. If ``False``, no topology older than the MRCAs of the
> samples will be included. If ``True`` the roots of all trees in the returned tree
> sequence will be the same roots as in the original tree sequence. (Default: False)

petrelharp: Ah, yes, sorry - your suggestion is in fact exactly what I was thinking.  I suppose that the current statement is not wrong, but a bit confusing.

hyanwong: I'll change it to the quote above "Whether to retain history ancestral to the
MRCA of the samples. If False, no topology older than the MRCAs of the
samples will be included. If True the roots of all trees in the returned tree
sequence will be the same roots as in the original tree sequence. (Default: False)"

hyanwong: Done

petrelharp (approved):

PR #1185, opened 2021-02-03
Start putting example tree sequences into fixtures

hyanwong:
Here's my basic suggestion as to how we can make a variety of tree sequences available for testing to multiple modules, using fixtures. The idea is that any function that generates a tree sequence which could be used over multiple files just bungs the function in `ts_examples.py`, and it becomes accessible to any function using a generic `ts_fixture` parameter, via `@pytest.mark.parameterize`. I'm quite pleased with how it's turned out, but I don't know how easy it would be to add the possibility of tweaking parameters involved in simulating each tree sequence (e.g. `mutation_rate`, or `n`).

Ping @benjeffery who seems to be the fixture expert.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -42,7 +42,7 @@ import msprime
 import pytest
 from pytest import fixture
 
-import tskit
+from . import ts_examples
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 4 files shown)


benjeffery:
Thanks for having a look at this @hyanwong - if it is ok with you I'll park this until the after the next tskit release as I'd like to give it some proper thought.

hyanwong:
Yeah, no hurry here. I only did this because I'm repeatedly finding that I'm duplicating code all over the place when I write new tests, which seems wrong. It would, however, be useful to have a standard comment to use when I duplicate TS-producing code so we can find all these dups and rip out them out later, when (if) we "fixturize" it all.

jeromekelleher:
I'm starting to think fixtures are an antipattern actually, and we'd be better off using ``parametrize`` as the better way of modernising  the test suite. Since this PR is now a long way behind I think it's probably better to close it. Any objectsions?

Issue #1154, opened 2021-01-29
Bug in simplify with KEEP_UNARY_IN_INDIVIDUALS


hyanwong:
Yes, this really is a simplify bug, which is my own fault. In f2d906d21fee3b8fa8991f4666c76a016a623ffe I inserted a check to see if the node being simplified has an individual:

https://github.com/tskit-dev/tskit/blob/6c3fdc48ccbfb53e66898dbc1d1b689580d66464/c/tskit/tables.c#L6783

~~But it seems that `input_id` here does not correspond to the node number in the current `self->tables->nodes` - I'm assuming it's been changed already by some previous node jiggling in the simplification algorithm, but the tables haven't been updated.~~ (see below: I'm not sure this is the problem, actually)

This is hard to test in python, because `keep_unary_in_individuals` is not an option to the simplify command yet: it's only in the C interface. But hopefully someone who knows about simplify can verify that my triaging is correct. Otherwise I can post a tree sequence where this causes a problem, and link to my branch where I've implemented `keep_unary_in_individuals` in python.

hyanwong:
The TS file is at [keep_unary_in_individuals_test.trees.zip](https://github.com/tskit-dev/tskit/files/5895893/keep_unary_in_individuals_test.trees.zip), which is a ts which has been slimmed down to 1 tree. Node 3 is a unary node within individual 1, and it is on the path from node 25 to a root (node 15), so it should be kept in. When I try

```
ts.simplify([n.id for n in ts_small.nodes() if n.time==0], keep_unary_in_individuals=True).num_individuals
```

With my modified python tskit, that node disappears. In fact, when we get into `simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)` for `input_id` 3, `self->tables->nodes.individual[3]` equals `TSK_NULL`. Perhaps that means I'm wrong in my guess above. It looks like node 3 doesn't have an individual in `self->tables->nodes.individual` after all. So the node table itself has changed by this point.

hyanwong:
Aha!

I think it should be

```
self->input_tables.nodes[input_id]
```

(not `self->tables->nodes[input_id]`)

benjeffery:
@hyanwong Can we get a PR for the python implementation before the bug fix? Would be good to separate those.

hyanwong:
> @hyanwong Can we get a PR for the python implementation before the bug fix? Would be good to separate those.

[...]

PR #1186, opened 2021-02-03
Remap individual ids after simplification

benjeffery:
@jeromekelleher  Just trying to get my head round this by adding it to the python simplify first. Is it this simple if we aren't worrying about consistency with the genetic info?

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -31,6 +31,7 @@ import pathlib
 import pickle
 import platform
 import random
+import struct
 import time
 import unittest
 import warnings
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Er, yes, actually, I think that basically does it. There'll be some subtleties, but that should be the fundamental change we need, since we're already remapping the individual IDs.

What we really want here to test this stuff out is a way of simulating individuals in the simple forward simulator we have. I think we should be able to add an option ``record_individuals=False`` option and add an individual [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_wright_fisher.py#L154) and it should all work itself out, ish.

That way we can get some decent sized test data without having to hand-craft tree sequences.

benjeffery:
> What we really want here to test this stuff out is a way of simulating individuals in the simple forward simulator we have. I think we should be able to add an option `record_individuals=False` option and add an individual [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_wright_fisher.py#L154) and it should all work itself out, ish.

I've had a stab at that in https://github.com/tskit-dev/tskit/pull/1186/commits/0ad085bd0327859529ca70359c7284468ddfd98e adding the same level of testing that migrations have - which seems to just be counting the rows.

jeromekelleher:
> I've had a stab at that in 0ad085b adding the same level of testing that migrations have - which seems to just be counting the rows.

Nice! I'd need to take the code for a spin to be sure I know what's happening, but can't right now (should have started this presentation yesterday...)

benjeffery:
> > I've had a stab at that in [0ad085b](https://github.com/tskit-dev/tskit/commit/0ad085bd0327859529ca70359c7284468ddfd98e) adding the same level of testing that migrations have - which seems to just be counting the rows.
> 
> Nice! I'd need to take the code for a spin to be sure I know what's happening, but can't right now (should have started this presentation yesterday...)

No worries - I'm working on the c side now.

jeromekelleher:
Working on it now - will tack in a commit with some updates to the Python code.

jeromekelleher:
@benjeffery I pushed a commit with some updates. There was a slight logic bug in the individual ID remapping and I fixed up the tests a bit. The test refactoring could probably be done better, but it's a big improvement over what it was anyway.

jeromekelleher:
So the tests to be done are now to see if the identity relationships between nodes and individuals are retained through simplification. The easiest way to do that I think is to bung in some metadata into both, simplify, and then see if every node's individual has the same metadata as the node. Then, for those parents that happen to survive simplification (which probably won't be many, most will be associated with unary nodes), the parent relationships should be the same.

For ``filter_individuals``=False, we should just keep the whole pedigree in place.

jeromekelleher:
@hyanwong, this addition to the simulator will make your life easier in #1154.

petrelharp:
This is going to be so much easier to test now that we have metadata!

benjeffery:
I've added some tests using the forward sim.

benjeffery:
Ok, there is one last thing to decide here - in #1192 @jeromekelleher suggested simplify should cope with unordered individuals. The current re-referencing code in this PR only works with sorted individuals as it works in the same pass building the new table. It's no biggie to add another pass over the individual table though.

jeromekelleher:
@petrelharp, your call here. A second linear pass through the individuals is no biggie IMO, and probably worth not enforcing a full topological sort on the input data for.

petrelharp:
Sounds good - let's do the second pass. SLiM currently might have individuals before their parents (because the 'remembered' ones come up top of the table, so if a child is remembered and its parent is not...), and we could change that, but it might actually be a can of worms (because (a) if a child gets remembered before a parent, they'd be out of order; or (b) when we output, we stick the current generation on to the individual table, with the remembered indiviiduals up top; to enforce the order in these two situations we'd have to do the topological sort...).

benjeffery:
@jeromekelleher I've added tests with shuffled individuals (that triggered `tsk_bug_assert`), then modified the C to cope. Should be ready to go after a squash.

python/tests/test_tables.py
```diff
@@ -2361,6 +2362,134 @@ def test_samples_interface(self):
...
+        # Check we have a mixed up order
+        with pytest.raises(
+            tskit.LibraryError,
+            match="Individuals must be provided in an order where"
+            " children are after their parent individuals",
+        ):
+            tables.tree_sequence()
+
+        tables.simplify()
+        metadata = [
```
jeromekelleher, line 2453: Might be easier make a tree sequence out of it, and let machinery do the metadata processing for you?

jeromekelleher (commented):
LGTM, some minor suggestions to reduce the amount of code.

python/tests/test_tables.py
```diff
@@ -2361,6 +2362,134 @@ def test_samples_interface(self):
...
+                if parent != tskit.NULL:
+                    assert (
+                        ts.individual(parent).metadata["original_id"] == original_parent
+                    )
+        assert set(tables.individuals.parents) != {-1}
+
+    def test_shuffled_individual_parent_mapping(self, wf_sim_with_individual_metadata):
+        tables = wf_sim_with_individual_metadata.copy()
+        individuals = tables.individuals
+
```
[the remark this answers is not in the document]
benjeffery: Didn't know about that method! Doh!

python/tests/test_tables.py
benjeffery: We can't as the individuals being shuffled violates the requirements for a tree sequence. Hence the `raises` check above.

jeromekelleher: Doesn't ``simplify`` sort this out, or is that a follow on?

benjeffery: No, simplify doesn't sort the individuals. You'd have to do a `sort_tables` after, but then you're not just testing simplify.

jeromekelleher: Fair enough :+1:

benjeffery:
@jeromekelleher Comments addressed, thanks for pointing out the existing shuffle method, I should have had a look for one!

jeromekelleher (approved):
LGTM!

PR #1190, opened 2021-02-05
Implement keep_unary_in_individuals in Python

### LINKED ISSUE #1154: Bug in simplify with KEEP_UNARY_IN_INDIVIDUALS

hyanwong:
~~Adds python support for keep_unary="in_individuals", but does not fix #1154 . As requested by @benjeffery~~

Updated to include the fix + tests

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -6982,7 +6982,7 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
         keep_unary = true;
     }
     if ((self->options & TSK_KEEP_UNARY_IN_INDIVIDUALS)
-        && (self->tables->nodes.individual[input_id] != TSK_NULL)) {
+        && (self->input_tables.nodes.individual[input_id] != TSK_NULL)) {
         keep_unary = true;
     }
 
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -2368,12 +2368,13 @@ class TestSimplifyTables:
             9,
             10,
             seed=1,
-            deep_history=True,
+            deep_history=False,
             initial_generation_samples=False,
             num_loci=5,
             record_individuals=True,
         )
         assert tables.individuals.num_rows > 50
+        assert np.all(tables.nodes.individual >= 0)
         individuals_copy = tables.copy().individuals
         tables.individuals.clear()
         tables.individuals.metadata_schema = tskit.MetadataSchema({"codec": "json"})
@@ -2403,6 +2404,26 @@ class TestSimplifyTables:
                         ts.individual(parent).metadata["original_id"] == original_parent
                     )
         assert set(tables.individuals.parents) != {tskit.NULL}
+
+    def verify_complete_genetic_pedigree(self, tables):
+        ts = tables.tree_sequence()
+        for edge in ts.edges():
+            child = ts.individual(ts.node(edge.child).individual)
+            parent = ts.individual(ts.node(edge.parent).individual)
+            assert parent.id in child.parents
+            assert parent.metadata["original_id"] in child.metadata["original_parents"]
+
+    def test_no_complete_genetic_pedigree(self, wf_sim_with_individual_metadata):
+        tables = wf_sim_with_individual_metadata.copy()
+        tables.simplify()  # Will remove intermediate individuals
+        with pytest.raises(AssertionError):
+            self.verify_complete_genetic_pedigree(tables)
+
+    def test_complete_genetic_pedigree(self, wf_sim_with_individual_metadata):
+        for params in [{"keep_unary": True}, {"keep_unary_in_individuals": True}]:
+            tables = wf_sim_with_individual_metadata.copy()
+            tables.simplify(**params)  # Keep intermediate individuals
+            self.verify_complete_genetic_pedigree(tables)
 
     def test_shuffled_individual_parent_mapping(self, wf_sim_with_individual_metadata):
         tables = wf_sim_with_individual_metadata.copy()
--- a/python/tests/test_wright_fisher.py
+++ b/python/tests/test_wright_fisher.py
@@ -592,3 +592,58 @@ class TestSimplify:
         other_tables.provenances.clear()
         assert tables == other_tables
         self.verify_simplify(ts, small_ts, sub_samples, node_map)
+
+    @pytest.mark.parametrize("ts", wf_sims)
+    @pytest.mark.parametrize("nsamples", [2, 5])
+    def test_simplify_keep_unary(self, ts, nsamples):
+        np.random.seed(123)
+        ts = tsutil.mark_metadata(ts, "nodes")
+        sub_samples = random.sample(list(ts.samples()), min(nsamples, ts.num_samples))
+        random_nodes = np.random.choice(ts.num_nodes, ts.num_nodes // 2)
+        ts = tsutil.insert_individuals(ts, random_nodes)
+        ts = tsutil.mark_metadata(ts, "individuals")
+
+        for params in [{}, {"keep_unary": True}, {"keep_unary_in_individuals": True}]:
+            sts = ts.simplify(sub_samples, **params)
+            # check samples match
+            assert sts.num_samples == len(sub_samples)
+            for n, sn in zip(sub_samples, sts.samples()):
+                assert ts.node(n).metadata == sts.node(sn).metadata
+
+            # check that nodes are correctly retained: only nodes ancestral to
+            # retained samples, and: by default, only coalescent events; if
+            # keep_unary_in_individuals then also nodes in individuals; if
+            # keep_unary then all such nodes.
+            for t in ts.trees(tracked_samples=sub_samples):
+                st = sts.at(t.interval[0])
+                visited = [False for _ in sts.nodes()]
+                for n, sn in zip(sub_samples, sts.samples()):
+                    last_n = t.num_tracked_samples(n)
+                    while n != tskit.NULL:
+                        ind = ts.node(n).individual
+                        keep = False
+                        if t.num_tracked_samples(n) > last_n:
+                            # a coalescent node
+                            keep = True
+                        if "keep_unary_in_individuals" in params and ind != tskit.NULL:
+                            keep = True
+                        if "keep_unary" in params:
+                            keep = True
+                        if (n in sub_samples) or keep:
+                            visited[sn] = True
+                            assert sn != tskit.NULL
+                            assert ts.node(n).metadata == sts.node(sn).metadata
+                            assert t.num_tracked_samples(n) == st.num_samples(sn)
+                            if ind != tskit.NULL:
+                                sind = sts.node(sn).individual
+                                assert sind != tskit.NULL
+                                assert (
+                                    ts.individual(ind).metadata
+                                    == sts.individual(sind).metadata
+                                )
+                            sn = st.parent(sn)
+                        last_n = t.num_tracked_samples(n)
+                        n = t.parent(n)
+                st_nodes = list(st.nodes())
+                for k, v in enumerate(visited):
+                    assert v == (k in st_nodes)
--- a/python/tests/tsutil.py
+++ b/python/tests/tsutil.py
@@ -242,27 +242,41 @@ def insert_random_ploidy_individuals(ts, max_ploidy=5, max_dimension=3, seed=1):
     return tables.tree_sequence()
 
 
-def insert_individuals(ts, samples=None, ploidy=1):
+def insert_individuals(ts, nodes=None, ploidy=1):
     """
     Inserts individuals into the tree sequence using the specified list
-    of samples (or all samples if None) with the specified ploidy by combining
-    ploidy-sized chunks of the list.
+    of node (or use all sample nodes if None) with the specified ploidy by combining
+    ploidy-sized chunks of the list. Add metadata to the individuals so we can
+    track them
     """
-    if samples is None:
-        samples = ts.samples()
-    if len(samples) % ploidy != 0:
-        raise ValueError("number of samples must be divisible by ploidy")
+    if nodes is None:
+        nodes = ts.samples()
+    assert len(nodes) % ploidy == 0  # To allow mixed ploidies we could comment this out
     tables = ts.dump_tables()
     tables.individuals.clear()
     individual = tables.nodes.individual[:]
     individual[:] = tskit.NULL
     j = 0
-    while j < len(samples):
-        nodes = samples[j : j + ploidy]
-        ind_id = tables.individuals.add_row()
-        individual[nodes] = ind_id
+    while j < len(nodes):
+        nodes_in_individual = nodes[j : min(len(nodes), j + ploidy)]
+        # should we warn here if nodes[j : j + ploidy] are at different times?
+        # probably not, as although this is unusual, it is actually allowed
+        ind_id = tables.individuals.add_row(
+            metadata=f"orig_id {tables.individuals.num_rows}".encode()
+        )
+        individual[nodes_in_individual] = ind_id
         j += ploidy
     tables.nodes.individual = individual
+    return tables.tree_sequence()
+
+
+def mark_metadata(ts, table_name, prefix="orig_id:"):
+    """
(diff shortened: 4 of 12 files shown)


hyanwong:
(also no tests)

hyanwong:
Also fixes #1120 . Note that I've used the string value `"in_individuals"` to turn on the `KEEP_UNARY_IN_INDIVIDUALS` behaviour, rather than the originally suggested `"individuals"`, as I think `simplify(keep_unary="individuals")` sounds like we keep unary individuals (whatever that means) rather than keeping unary nodes if they are in an individual. Happy to change to whatever the consensus is, though.

python/tskit/tables.py
```diff
@@ -2538,9 +2538,13 @@ def simplify(
...
-        :param bool keep_unary: If True, any unary nodes (i.e. nodes with exactly
-            one child) that exist on the path from samples to root will be preserved
-            in the output. (Default: False)
+        :param str keep_unary: When to preserve unary nodes (i.e. nodes with
+            exactly one child) that exist on the path from samples to root. If
+            ``None`` all unary nodes are removed. If "all" all such nodes will
+            be preserved in the output. If "in_individuals", they are only preserved
+            if they are associated with an individual. (Note that the boolean values
+            ``True`` and ``False`` can be used as shortcuts for "all" and ``None``.)
+            (Default: None)
```
petrelharp, line 2547: This could lead to confusion; we should probably allow `"none"` and `"None"` as aliases for `None` here?

Oh, actually: how about we replace `None` with `"none"` here, and say the default is `"none"`, so that all the default values referred to in this text are strings?  (and then, of course, `None` will mean `"none"` below...)

petrelharp: and I think that if we don't talk about `None` in the text we won't get users accidentally upper-casing `"None"`...

hyanwong: Yes, that would be reasonable. At the moment it just errors out with strings other than "in_individuals". Happy to add "None" and "none" as aliases.

hyanwong: How do we feel about `""` and `"False"`. I assume we ignore them, but this is all getting a bit horrible!

hyanwong: Like this, @petrelharp?

```
      :param str keep_unary: When to preserve unary nodes (i.e. nodes with
            exactly one child) that exist on the path from samples to root. If
            "none" all unary nodes are removed. If "all" all such nodes will
            be preserved in the output. If "in_individuals", they are only preserved
            if they are associated with an individual. (Note that the boolean value
            ``True`` can be used as a shortcuts for "all" and ``False``,
            ``None``, or "None" as a shortcut for "none") (Default: "none")
```

Wow. 4 "none"s in a single line. 🤯

benjeffery: I'm not sure about the API here - is it not clearer to have two mutually exclusive flags like the C API?

benjeffery (commented):
I'm not a fan of the string/bool flags I'm afraid.
Are there also some tests we could add that don't hit #1154?

hyanwong:
> I'm not a fan of the string/bool flags I'm afraid.

Other suggestions welcome! I'm happy to sit on the fence on this one.

> Are there also some tests we could add that don't hit #1154?

Isn't that the only thing we are testing here?

python/tskit/tables.py
hyanwong: Depends on what future uses might be. I mooted some at https://github.com/tskit-dev/tskit/issues/1120#issuecomment-770179852 and another at https://github.com/tskit-dev/tskit/pull/1156#issuecomment-772081032. Whatever is most obvious to the user, I guess.

jeromekelleher:
Maybe we are trying to be too clever here with the API and it would be simpler to just add ``keep_unary_in_individuals`` to mirror the low-level API. We can always add a higher-level approach later if we need to.

@petrelharp, any objections?

hyanwong:
Now updated & swapped back to individual parameters `keep_unary` and `keep_unary_in_individuals`. Is this OK @benjeffery / @petrelharp ? Tests in https://github.com/tskit-dev/tskit/pull/1191 which is based off this PR. They are only separate because @benjeffery asked for that at the start. I can put them all in one PR if preferred.

python/tskit/tables.py
```diff
@@ -2493,6 +2493,7 @@ def simplify(
         filter_individuals=True,
         filter_sites=True,
         keep_unary=False,
+        keep_unary_in_individuals=False,
```
jeromekelleher, line 2496: Let's make the default in the signature ``None``, will give us some room to change the API later.

python/tskit/trees.py
```diff
@@ -4917,6 +4917,7 @@ def simplify(
         filter_individuals=True,
         filter_sites=True,
         keep_unary=False,
+        keep_unary_in_individuals=False,
```
jeromekelleher, line 4920: Ditto here.

c/tskit/tables.c
```diff
@@ -6982,7 +6982,7 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
         keep_unary = true;
     }
     if ((self->options & TSK_KEEP_UNARY_IN_INDIVIDUALS)
-        && (self->tables->nodes.individual[input_id] != TSK_NULL)) {
+        && (self->input_tables.nodes.individual[input_id] != TSK_NULL)) {
```
jeromekelleher, line 6985: This was a C bug, so as a minimum we need to add a C test that provoked the original bug

jeromekelleher (commented):
I'm a bit lost here as to what's going in what PR now. This one seems to include a fix for the unary_in_individuals bug, but doesn't include any tests. 

I think it would be simpler if we brought this and #1191 together into one and tried to get them merged ASAP.

We should include tests that use the pedigree information in the forward simulator that was introduced in recent commits.

hyanwong:
> I think it would be simpler if we brought this and #1191 together into one and tried to get them merged ASAP.

Yes, I agree. Since most of the discussion is in this PR & issue, I'll close the other one and add the tests here.

> We should include tests that use the pedigree information in the forward simulator that was introduced in recent commits.

I'm not sure what you mean. Why is the pedigree relevant to the unary nodes? Or do you mean that we keep *all* unary nodes and check that the pedigree info goes back from the samples to the roots correctly?

c/tskit/tables.c
hyanwong: Hmm, I think that might not be so easy. Part of the reason for exposing this to python was that it is much easier to test for this bug (the python tests I introduced fail unless the fix it there).

For the C tests, I suppose one way would be change the hard-coded example in https://github.com/tskit-dev/tskit/blob/028997baf34de806674309c9dc84297bc71b2d02/c/tests/test_trees.c#L1054 so it triggers this bug. It might take a bit of fiddling to find an example that does this, though (and the bug is dependent on the node visiting order within simplify(), I think)

python/tskit/tables.py
hyanwong: Do you want the default for `keep_unary` to be `None` too? This would be a change from the current API, but shouldn't actually change any behaviour.

jeromekelleher:
> I'm not sure what you mean. Why is the pedigree relevant to the unary nodes? Or do you mean that we keep all unary nodes and check that the pedigree info goes back from the samples to the roots correctly?

We're interested in unary nodes that refer to individuals, for which the application is forward simulations that store individuals. We have a simple forward simulator that stores invididauls now, so we should use it to test this feature.

hyanwong:
> > I'm not sure what you mean. Why is the pedigree relevant to the unary nodes? Or do you mean that we keep all unary nodes and check that the pedigree info goes back from the samples to the roots correctly?
> 
> We're interested in unary nodes that refer to individuals, for which the application is forward simulations that store individuals. We have a simple forward simulator that stores invididauls now, so we should use it to test this feature.

Yes, that's what the tests in #1191 do, ISTR. But they don't use the pedigree information, only the individuals. Perhaps it'll be clearer when I merge those tests into this PR?

hyanwong:
OK @jeromekelleher - this should be done now. I added an extra test that scans though all the edges and checks (after simplification either with keep_unary or keep_unary_with_individuals) that all the individuals along the genetic lineages are kept, along with their unary nodes.

I needed to flip the parameter in `wf_sim_with_individual_metadata` to `deep_history=False` so that all nodes had an individual. I hope that's OK @benjeffery ? From my reading of the test code that uses that fixture, the deep history is irrelevant, and isn't required for use in the tests.

hyanwong:
P.s. ping me if you need me to squash all these commits together after review?

c/tskit/tables.c
hyanwong: As discussed, the code without the C fix reliably fails the python tests, and the old C code didn't segfault, so I think we're covered.

python/tests/test_tables.py
```diff
@@ -2368,12 +2368,13 @@ def wf_sim_with_individual_metadata(self):
             9,
             10,
             seed=1,
-            deep_history=True,
+            deep_history=False,
```
petrelharp, line 2371: I haven't thought throught he implications of changing this for `test_shuffled_individual_parent_mapping`, have you?

hyanwong: Erm, why should this make a difference. We still shuffle the individuals and check that they are (a) shuffled and (b) the original individual ids correspond.

hyanwong: But perhaps I'm not understanding something?

jeromekelleher: Why change the default here? Seems gratuitous to me.

python/tests/tsutil.py
```diff
@@ -242,30 +242,45 @@ def insert_random_ploidy_individuals(ts, max_ploidy=5, max_dimension=3, seed=1):
     return tables.tree_sequence()
 
 
-def insert_individuals(ts, samples=None, ploidy=1):
+def insert_individuals(ts, nodes=None, ploidy=1, allow_mixed_ploidy=False):
```
jeromekelleher, line 245: Do we use this ``allow_mixed_ploidy`` argument? If not, why add it?

jeromekelleher (commented):
Looks good, but some gratuitous changes included AFAICT

python/tests/tsutil.py
hyanwong: Re deep_history - if we graft deep history onto the simulation we no longer have individuals for all the nodes (the new "deep history" nodes don't have individuals)

hyanwong: Re mixed ploidy, sure, can remove. It just seemed like a useful thing to flag up to our future selves. Perhaps all that's needed is to make a comment/warning in the docstring that this utility function could result in mixed ploidies?

python/tests/test_tables.py
hyanwong: See below. If we want to check that all nodes have an individual, and all the individuals point to their correct parents, then we can't include the deep history nodes, which have no individuals (and the nodes at the top of the WF generated simulation will not have the correct parents.

I.e. if we have a deep history, we (almost by definition) cannot fully map the trees onto the individuals pedigree. It seems worth testing that we were capturing *all* of the genetic pedigree in the individuals.

petrelharp: (yan has thought this over)

python/tests/tsutil.py
petrelharp: sounds good

python/tests/test_wright_fisher.py
```diff
@@ -598,27 +598,43 @@ def test_simplify_tables(self, ts, nsamples):
...
-                    assert node.individual >= 0
-                    assert node.metadata in md_map
-                    orig_metadata = md_map[node.metadata]
-                    assert orig_metadata == ts.individual(node.individual).metadata
-        assert num_nonsample_unary > 0
+        sts = ts.simplify(sub_samples, keep_unary_in_individuals=True)
+        # check samples match
+        assert sts.num_samples == len(sub_samples)
+        for n, sn in zip(sub_samples, sts.samples()):
+            print("::::", n, sn)
```
hyanwong, line 610: do you want this print in there?

petrelharp:
I've made the WF test stricter (now it's an if-and-only-if test).

python/tests/test_wright_fisher.py
petrelharp: whoops sorry

hyanwong: NP. I do this *all* the time (just ask Jerome)

petrelharp:
With the new test and pending the suggestions above I'm good with this. Do you want to take care of those and remove my print statement and squash, @hyanwong?

hyanwong:
OK, thanks @petrelharp and @jeromekelleher - I removed the print statement and the `allow_mixed_ploidy` param in the test suite (turns out we didn't have any tests with mixed ploidy, so I put the assert in, but we can remove later if we find we need it. I guess this is ready to go?

petrelharp:
Ah, wait - another thing to do. You've implemented the `keep_unary_in_individuals` in the python Simplifier, but are not verifying it does the same thing as `ts.simplify()`. (See the tests that do this for `keep_unary`, for instance.) Unless I'm missing something.

hyanwong:
> Ah, wait - another thing to do. You've implemented the `keep_unary_in_individuals` in the python Simplifier, but are not verifying it does the same thing as `ts.simplify()`. (See the tests that do this for `keep_unary`, for instance.) Unless I'm missing something.

Yeah, you're right. Is what you are meaning here?

https://github.com/tskit-dev/tskit/blob/a23b5b4fddd528be2ef94ab317632e35addf2c44/python/tests/test_topology.py#L2403

I had a look, and I think there might be a bug in that test. In particular, the ts-with-unary-nodes that is created is (I think) called `new_ts`, but the `tests.Simplifier` code is run on the *old* ts, that doesn't have any unary nodes in it. Or am I missing something?

hyanwong:
https://github.com/tskit-dev/tskit/pull/1190/commits/1296d99c95b1305322ba224dcbb9a6fecc2f993a adds a test for the python simplifier and also corrects what I think is a wrong variable name in the previous code (we weren't actually testing the ts with unary nodes before: oops).

If this is correct, I can squash the commit into the previous, if required

petrelharp:
Looks good to me. I made some minor adjustments to the WF test I wrote, so it's also testing correctness of `keep_unary` and the default as well (and removed the num_samples=10 case because the tests were starting to run slow). Have a look at what I did, then I think we can squash and merge?

hyanwong:
> Looks good to me. I made some minor adjustments to the WF test I wrote, so it's also testing correctness of `keep_unary` and the default as well (and removed the num_samples=10 case because the tests were starting to run slow). Have a look at what I did, then I think we can squash and merge?

LGTM. Squashed now.

Looking over it I realise that I *didn't* do the thing of slicing through the TS at a certain time and checking that all lineages were covered; it's done by picking nodes at random instead. But I guess the random thing is a harder test anyway, so all's fine. Also there's the slicing test in the [SLiM PR](https://github.com/MesserLab/SLiM/pull/158) that's just been merged, so the idea is saved somewhere, at least.

jeromekelleher (approved):
Looks like we're all happy, so good to go! Thanks @hyanwong and @petrelharp.

hyanwong:
> Looks like we're all happy, so good to go! Thanks @hyanwong and @petrelharp.

Yay. Thanks Jerome.

PR #1192, opened 2021-02-08
Add individual parent checks

### LINKED ISSUE #1137: `Individual.parents` should be added to integrity checks

benjeffery:
Fixes #1137 by adding checks for individual reference integrity and ordering.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7953,25 +7953,26 @@ tsk_table_collection_check_individual_integrity(
     const tsk_id_t num_individuals = (tsk_id_t) individuals.num_rows;
     const bool check_individual_ordering = options & TSK_CHECK_INDIVIDUAL_ORDERING;
 
-    /* Check parent references are valid */
-    for (j = 0; j < individuals.parents_length; j++) {
-        if (individuals.parents[j] != TSK_NULL
-            && (individuals.parents[j] < 0
-                   || individuals.parents[j] >= num_individuals)) {
-            ret = TSK_ERR_INDIVIDUAL_OUT_OF_BOUNDS;
-            goto out;
-        }
-    }
-
-    /* Check parents are ordered */
-    if (check_individual_ordering) {
-        for (j = 0; j < (tsk_size_t) num_individuals; j++) {
-            for (k = individuals.parents_offset[j];
-                 k < individuals.parents_offset[j + 1]; k++) {
-                if (individuals.parents[k] != TSK_NULL
-                    && individuals.parents[k] >= (tsk_id_t) j) {
-                    ret = TSK_ERR_UNSORTED_INDIVIDUALS;
-                }
+    for (j = 0; j < (tsk_size_t) num_individuals; j++) {
+        for (k = individuals.parents_offset[j]; k < individuals.parents_offset[j + 1];
+             k++) {
+            /* Check parent references are valid */
+            if (individuals.parents[k] != TSK_NULL
+                && (individuals.parents[k] < 0
+                       || individuals.parents[k] >= num_individuals)) {
+                ret = TSK_ERR_INDIVIDUAL_OUT_OF_BOUNDS;
+                goto out;
+            }
+            /* Check no-one is their own parent */
+            if (individuals.parents[k] == (tsk_id_t) j) {
+                ret = TSK_ERR_INDIVIDUAL_SELF_PARENT;
+                goto out;
+            }
+            /* Check parents are ordered */
+            if (check_individual_ordering && individuals.parents[k] != TSK_NULL
+                && individuals.parents[k] >= (tsk_id_t) j) {
+                ret = TSK_ERR_UNSORTED_INDIVIDUALS;
+                goto out;
             }
         }
     }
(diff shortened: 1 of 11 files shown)


c/tskit/tables.c
```diff
@@ -6831,8 +6831,9 @@ simplifier_init(simplifier_t *self, const tsk_id_t *samples, size_t num_samples,
      * debateable whether we need it. If we remove, we definitely need explicit
      * tests to ensure we're doing sensible things with duplicate sites.
      * (Particularly, re TSK_REDUCE_TO_SITE_TOPOLOGY.) */
-    ret = tsk_table_collection_check_integrity(tables,
-        TSK_CHECK_EDGE_ORDERING | TSK_CHECK_SITE_ORDERING | TSK_CHECK_SITE_DUPLICATES);
+    ret = tsk_table_collection_check_integrity(
+        tables, TSK_CHECK_EDGE_ORDERING | TSK_CHECK_SITE_ORDERING
+                    | TSK_CHECK_SITE_DUPLICATES | TSK_CHECK_INDIVIDUAL_ORDERING);
```
jeromekelleher, line 6836: Why do we need TSK_CHECK_INDIVIDUAL_ORDERING  this for simplify?

jeromekelleher (commented):
LGTM, but I don't think we need to check for ordered individuals in simplify. We should relax the requirement and put in a test (in Python) to verify that we can successfully simplify with unordered parents.

c/tskit/tables.c
benjeffery: Seems I got over zealous! Removed!

c/tskit/tables.c
```diff
@@ -7939,6 +7939,42 @@ tsk_table_collection_check_migration_integrity(
...
+    }
+
+    /* Check parents are ordered */
+    if (check_individual_ordering) {
+        for (j = 0; j < (tsk_size_t) num_individuals; j++) {
+            for (k = individuals.parents_offset[j];
+                 k < individuals.parents_offset[j + 1]; k++) {
+                if (individuals.parents[k] != TSK_NULL
+                    && individuals.parents[k] >= (tsk_id_t) j) {
+                    ret = TSK_ERR_UNSORTED_INDIVIDUALS;
```
petrelharp, line 7969: I suppose we might check that no individual is its own parent (even if we are not checking individual ordering).

python/tests/test_tables.py
```diff
@@ -2361,6 +2361,20 @@ def test_samples_interface(self):
...
+        tables.individuals.add_row(parents=[-2])
+        with pytest.raises(tskit.LibraryError, match="Individual out of bounds"):
+            tables.simplify()
+
+    def test_unsorted_individuals(self, simple_ts_fixture):
+        tables = simple_ts_fixture.dump_tables()
+        tables.individuals.clear()
+        tables.individuals.add_row(parents=[1])
+        tables.individuals.add_row(parents=[-1])
+        tables.simplify()
```
petrelharp, line 2376: does this test anything?

petrelharp:
I realized we can check for loops reasonably easily:
```
for ind in ts.individuals():
  for p in ind.parents:
    while p != tskit.NULL:
     if p == ind.id:
         raise ValueError("Inconsistent individual parents (time travel)")
```

petrelharp:
Two more things we might check?

c/tskit/tables.c
jeromekelleher: Could be a useful diagnostic. No good reason to be your own parent (unless people are being sloppy with the tables and want to fix things up at the end).

python/tests/test_tables.py
jeromekelleher: No, I don't think so.

We should check that the integrity of individual refs is maintained. A good test here could be to take a forward sim sim with parent references, shuffle them, simplify, and then use the metadata to confirm that the correct individuals are parents at the end of simplify.

jeromekelleher (commented):
Still needs some tweaks on testing simplify.

jeromekelleher:
> I realized we can check for loops reasonably easily:

Doesn't parent-before-child rule out loops?

c/tskit/tables.c
benjeffery: I'll add a test for this, as this code already forbids it.

petrelharp:
> Doesn't parent-before-child rule out loops?

Yes, sorry - never mind this.

python/tests/test_tables.py
benjeffery: This test belongs in #1186, I'll add it there.

benjeffery:
Ok, I've added the test for self-parentage and removed the simplify test that was doing nothing. Should be good to go.

petrelharp:
I was actually suggesting that we check for self parentage always (not just when checking for ordering). And, turns out that test I said wasn't testing anything actually was - it was testing whether simplify runs even if the individuals are unordered. I've just gone ahead and done those things; feel free to do whatever to this commit.

benjeffery:
@jeromekelleher I like @petrelharp 's addition here, more precise error messages are always a good thing. I think this PR is ready to go.

jeromekelleher (approved):

jeromekelleher:
OK, great!

Issue #1197, opened 2021-02-11
sort should ensure ordering of individual parents (also, mutation parents?)


petrelharp:
`sort( )` should produce valid tables; with #1192 one of the requirements not ensured by `sort` is that parent individuals come before children. We should do this! Also, I think we should do the same for mutations. (I'm only proposing sorting so that mutations who have their parent mutations specified get sorted correctly; otherwise we'd have to compute the mutation parents.)

Since sort is a bottleneck in forwards sims, and this sortedness is not required by simplify, perhaps this should be optional. (although it sounds annoying to make it optional)

jeromekelleher:
Agreed. But:

> Since sort is a bottleneck in forwards sims, and this sortedness is not required by simplify, perhaps this should be optional. (although it sounds annoying to make it optional)

I think edge sorting is the bottleneck, right? All the rest is pretty small fry compared to the large edge tables, so adding a few constant factors isn't going to make any difference. I would vote for doing it the way we think is right first, and then backing out an making it optional if it turns out to add to sort-time in forward sims.

benjeffery:
I'm going to have a stab at this, @jeromekelleher and I just talked it over and sorting individuals is equivalent to topologically sorting the directed graph of parent references. Kahn's algorithm looks promising for this, it also will detect cycles in the graph.

PR #1198, opened 2021-02-11
avoid possible infinite loop in canonicalize

petrelharp:
Looking at #1192 I realized there was potential for an infinite loop in the case of bad mutation parent input. Whoops! Here's a fix.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -4088,6 +4088,48 @@ test_sort_tables_mutation_times(void)
 }
 
 static void
+test_sort_tables_canonical_errors(void)
+{
+    int ret;
+    tsk_table_collection_t tables;
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1;
+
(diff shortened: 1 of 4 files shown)


petrelharp:
Whoops, this doesn't fix it. Consider 0->1, 1->2, 2->1. Huh, tricky.

jeromekelleher:
I guess we  have to sort first so that parents come before children, and then check for loops? (Like #1197?)

petrelharp:
The problem occurs while computing the number of descendant mutations, which is the key that allows us to sort so parents come before children if times are equal (or, unknown).

This is the problem of detecting cycles in a directed graph; I'll have a look at algorithms for that.

petrelharp:
Ok, this will correctly check things now. I think it's totally good enough.

petrelharp:
Wait, never mind - not right yet.

petrelharp:
Ok, for reals now! Doing the loop check in a boneheaded way.

jeromekelleher (approved):
LGTM!

jeromekelleher:
This is clever - detect a loop by letting it loop until we know it's definitely gone too far. Who cares if it's not the most efficient possible way of detecting the loop, it's an error anyway. Very elegant!

Issue #1138, opened 2021-01-25
Consider ordering requirements for individual table.


benjeffery:
#1125 adds parents to the individual table. Currently there are no sorting requirements on the individual table. 

@jeromekelleher and I discussed this and thought that enforcing that parents come before children in the table would presumably be useful and also ensure that there are no loops in the pedigree. Thoughts @petrelharp?

petrelharp:
Hm. It would presumably be useful. But it would also totally mess up how SLiM use the individual table. That's fine as long as we don't require this for `simplify_tables`, but it's a bit awkward to have these caveats about this algorithm not needing that sortedness property, etcetera. On the other hand, I see how this would make developing pedigree algorithms much easier and more efficient! So, I guess my suggestions are:
- we should at least have `canonicalise` put individual parents before children, and
- we could either add it to the list of standard sortedness requirements, but ensure simplify doesn't require it, *or*
- allow pedigree algorithms to require the property (which can be checked with two passes through the table).

Those two options aren't very different, as they're both introducing an extra property to keep track of; the question is just whether it's default or not. We could wait until someone starts developing pedigree algorithms to see how useful it is?

jeromekelleher:
That's good to know, thanks @petrelharp. I think we could add the ordering to the standard tree sequence requirements without breaking simplify cleanly enough by adding a ``CHECK_INDIVIDUAL_ORDERING`` to the list of flags for check_integrity [here](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.h#L698). Simplify won't need to provide this flag, as all it'll be doing is remapping IDs.

That would mean that individual IDs in SLiM would need to be rejiggered at the end before output so that the output file is a valid tree sequence, but that could be done efficiently with some version of canonicalise/subset or whatever. If the original IDs are important, then I guess these could be tracked via metadata or something.

I thought that parents-being-before-children would have mapped very naturally to forward simulations - SLiM must be using the individual table quite differently then?

> (which can be checked with two passes through the table).

It's just one pass, isn't it? All we're saying is basically there's no forward references: at row ``j``, the parent values must be < ``j``.

petrelharp:
> That would mean that individual IDs in SLiM would need to be rejiggered at the end before output so that the output file is a valid tree sequence, but that could be done efficiently with some version of canonicalise/subset or whatever. If the original IDs are important, then I guess these could be tracked via metadata or something.

[...]

PR #1199, opened 2021-02-15
Sort individuals

### LINKED ISSUE #1197: sort should ensure ordering of individual parents (also, mutation parents?)

benjeffery:
Sort individuals. Fixes #1197 #1138.

WIP!

Currently only a python implementation (deliberately written in a style to aid translation to C). Needs more tests etc, but thought it worth sharing progress.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -4959,6 +4959,121 @@ out:
     return ret;
 }
 
+static int
+tsk_table_sorter_sort_individuals(tsk_table_sorter_t *self)
+{
+    int ret = 0;
+    tsk_id_t i;
+    tsk_individual_table_t copy;
+    tsk_individual_t individual;
+    tsk_individual_table_t *individuals = &self->tables->individuals;
+    tsk_node_table_t *nodes = &self->tables->nodes;
+    tsk_size_t num_individuals = individuals->num_rows;
+    tsk_size_t current_todo = 0;
+    tsk_size_t todo_insertion_point = 0;
+    tsk_size_t *incoming_edge_count
+        = malloc(num_individuals * sizeof(*incoming_edge_count));
+    tsk_id_t *individuals_todo
+        = malloc((num_individuals + 1) * sizeof(*individuals_todo));
+    tsk_id_t *new_id_map = malloc(num_individuals * sizeof(*new_id_map));
+
+    ret = tsk_individual_table_copy(individuals, &copy, 0);
+    if (ret != 0) {
+        goto out;
+    }
+
+    if (incoming_edge_count == NULL || individuals_todo == NULL || new_id_map == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    for (i = 0; i < (tsk_id_t) num_individuals; i++) {
+        incoming_edge_count[i] = 0;
+        individuals_todo[i] = TSK_NULL;
+        new_id_map[i] = TSK_NULL;
+    }
+    individuals_todo[num_individuals] = TSK_NULL; /* Sentinel value */
+
+    /* First find the set of individuals that have no children by creating
+     * an array of incoming edge counts */
+    for (i = 0; i < (tsk_id_t) individuals->parents_length; i++) {
+        if (individuals->parents[i] != TSK_NULL) {
+            incoming_edge_count[individuals->parents[i]]++;
+        }
+    }
+    /* Use these as the starting points for checking all individuals,
+     * doing this in reverse makes the sort stable */
+    for (i = (tsk_id_t) num_individuals - 1; i >= 0; i--) {
+        if (incoming_edge_count[i] == 0) {
+            individuals_todo[todo_insertion_point] = i;
+            todo_insertion_point++;
+        }
+    }
+
+    /* Now emit individuals from the set that have no children, removing their edges
+     * as we go adding new individuals to the no children set. */
+    while (individuals_todo[current_todo] != TSK_NULL) {
+        tsk_individual_table_get_row_unsafe(
+            individuals, individuals_todo[current_todo], &individual);
+        for (i = 0; i < (tsk_id_t) individual.parents_length; i++) {
+            if (individual.parents[i] != TSK_NULL) {
+                incoming_edge_count[individual.parents[i]]--;
+                if (incoming_edge_count[individual.parents[i]] == 0) {
+                    individuals_todo[todo_insertion_point] = individual.parents[i];
+                    todo_insertion_point++;
+                }
+            }
+        }
+        current_todo++;
+    }
+
+    /* Any edges left are parts of cycles */
+    for (i = 0; i < (tsk_id_t) num_individuals; i++) {
+        if (incoming_edge_count[i] > 0) {
+            ret = TSK_ERR_INDIVIDUAL_PARENT_CYCLE;
+            goto out;
+        }
+    }
+
+    ret = tsk_individual_table_clear(individuals);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* The sorted individuals are in reverse order */
+    for (i = (tsk_id_t) num_individuals - 1; i >= 0; i--) {
+        tsk_individual_table_get_row_unsafe(&copy, individuals_todo[i], &individual);
+        ret = tsk_individual_table_add_row(individuals, individual.flags,
+            individual.location, individual.location_length, individual.parents,
+            individual.parents_length, individual.metadata, individual.metadata_length);
+        if (ret < 0) {
+            goto out;
+        }
+        new_id_map[individuals_todo[i]] = ret;
+    }
+
+    /* Rewrite the parent ids */
+    for (i = 0; i < (tsk_id_t) individuals->parents_length; i++) {
+        if (individuals->parents[i] != TSK_NULL) {
+            individuals->parents[i] = new_id_map[individuals->parents[i]];
+        }
+    }
+    /* Rewrite the node individual ids */
+    for (i = 0; i < (tsk_id_t) nodes->num_rows; i++) {
+        if (nodes->individual[i] != TSK_NULL) {
+            nodes->individual[i] = new_id_map[nodes->individual[i]];
+        }
+    }
+
+    ret = 0;
+out:
+    tsk_safe_free(incoming_edge_count);
+    tsk_safe_free(individuals_todo);
+    tsk_safe_free(new_id_map);
+    tsk_individual_table_free(&copy);
+    return ret;
+}
+
 int
 tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
 {
@@ -4966,6 +5081,7 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
     tsk_size_t edge_start = 0;
     tsk_size_t migration_start = 0;
     bool skip_sites = false;
+    bool skip_individuals = false;
 
     if (start != NULL) {
         if (start->edges > self->tables->edges.num_rows) {
@@ -4986,6 +5102,14 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
             && start->mutations == self->tables->mutations.num_rows) {
             skip_sites = true;
         } else if (start->sites != 0 || start->mutations != 0) {
+            ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
+            goto out;
+        }
+
+        /* Individuals also must be all sorted or skipped entirely */
+        if (start->individuals == self->tables->individuals.num_rows) {
+            skip_individuals = true;
+        } else if (start->individuals != 0) {
             ret = TSK_ERR_SORT_OFFSET_NOT_SUPPORTED;
             goto out;
         }
@@ -5015,6 +5139,12 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
             goto out;
         }
         ret = self->sort_mutations(self);
+        if (ret != 0) {
+            goto out;
+        }
+    }
+    if (!skip_individuals) {
+        ret = tsk_table_sorter_sort_individuals(self);
         if (ret != 0) {
             goto out;
         }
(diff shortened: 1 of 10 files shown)


jeromekelleher:
Looks very nice, eminently C-able!

benjeffery:
Was hoping to have this done yesterday, but taking longer than expected - the initial implementation's sort wasn't stable, that's fixed now. Currently fixing up tests in python side.

jeromekelleher:
> the initial implementation's sort wasn't stable, 

It's the gift that keeps on giving! Well done for spotting early.

petrelharp:
Nicely done. So the sort here is decreasing by "minimum number of hops to a childless descendant, and then smallest position in the table among the closest childless descendant".

benjeffery:
@petrelharp Yes! That's a great way of stating it. Should have this polished off today.

benjeffery:
While working on the python tests I realised we also need to tweak the canonicalisation as the sort there no longer meets the requirements of a tree sequence.

benjeffery:
We also have the classic msprime chicken and egg as all the sort testing code uses msprime that doesn't store the pedigree. Will add some `wf_sim` tree sequences to the sort tests for now.

petrelharp:
> While working on the python tests I realised we also need to tweak the canonicalisation as the sort there no longer meets the requirements of a tree sequence.

But `canonicalize` just uses the standard sort [with an enhanced mutation sort](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.c#L9058) so it'll get this change?

benjeffery:
> > While working on the python tests I realised we also need to tweak the canonicalisation as the sort there no longer meets the requirements of a tree sequence.
> 
> But `canonicalize` just uses the standard sort [with an enhanced mutation sort](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.c#L9058) so it'll get this change?

Ah, so I think the sort of individuals will be valid then. However, their sort order will depend on the ordering of the parents in each individual though - is that ok for canonicalisation or do we need to have a canonical ordering of the parent arrays?

petrelharp:
> Ah, so I think the sort of individuals will be valid then. However, their sort order will depend on the ordering of the parents in each individual though - is that ok for canonicalisation or do we need to have a canonical ordering of the parent arrays?

Oh, right - canonicalize uses `subset` to put the parents in order by node. That happens [before sort](https://github.com/tskit-dev/tskit/blob/main/c/tskit/tables.c#L9068), so I guess that with this change, canonicalise will sort individuals by "minimum number of hops to a childless descendant, and then smallest **node** among the closest childless descendant". It's still canonical, given the node ordering, I think? This will need some modifications to `canonicalise`'s docs, anyhow.

I guess that now `subset` can produce invalid tree sequences, though, since it reorders individuals by their nodes.

jeromekelleher:
> I guess that now subset can produce invalid tree sequences, though, since it reorders individuals by their nodes.

Ah, that's unfortunate...

benjeffery:
> I guess that now `subset` can produce invalid tree sequences, though, since it reorders individuals by their nodes.
Yes, this is true, I guess it can just sort the individuals as a last step? 

It doesn't seem like `subset` reorders the parent array here, so the final sort order is still dependant on the order the parents were specified in.

Also I can't see where `subset` remaps the ids in `individual.parent`?

petrelharp:
Ok, how about this: 
- have `subset` leave the individual table ordering alone (so, still subset individuals, but don't re-order them), and 
- come up with a canonical ordering for individuals, that either happens here, or if it's difficult, that `canonicalise` does.

petrelharp:
As for the order: modifying what you've got here it'd be easy to efficiently compute the total number of descendants of each individual (by recursively adding up the values of all children), so then we could have (num_descendants, smallest referring node) as a canonical sort key. I dunno whether to swap that in here or just do it in canonicalize - what you have here is nice because it doesn't use `qsort`, but then why would we bother implementing two different sorts in different places?

jeromekelleher:
> I dunno whether to swap that in here or just do it in canonicalize - what you have here is nice because it doesn't use qsort, but then why would we bother implementing two different sorts in different places?

Well, one argument would be one is O(n) and the other is O(n log n). I don't have a strong opinion about it though.

benjeffery:
As another option, I think if `subset` also sorts the parents in `individual.parents` then applying this sort will be canonical. 

I don't have enough information to know how to prioritise the speed of sort vs the complexity of having two sorts. Sorting is a very common operation though so I err to having that be O(n).

If it makes sense I'd like to finish this PR as-is and then fix the canonical issues in a separate one?

petrelharp:
>  I think if subset also sorts the parents in individual.parents then applying this sort will be canonical.

Um, I guess this is true? I'm a little unsure about this, given that the outcome of this sort depends on the order of parents. 

> If it makes sense I'd like to finish this PR as-is and then fix the canonical issues in a separate one?

That makes sense, but I think we should make sure we have good path forward: we'd like to at least make sure that if you canonicalise, a subsequent sort doesn't change anything, for instance.  And, that the order induced by canonicalise does not depend on the original order of the parent table, just on the order of the node table. And probably also not on the order that the different parents of a given individual are specified.

I'll think some more about this (although I'm ready to be done with sorting... =)

jeromekelleher:
Agree with all the above, but:

> : we'd like to at least make sure that if you canonicalise, a subsequent sort doesn't change anything, for instance. 

I wouldn't get too hung up on this; once the output of sort is guaranteed to be legal input for tables that's all that really matters. Canonical must be canonical though, there can be only one output from ``canonicalise``.

petrelharp:
> I wouldn't get too hung up on this; once the output of sort is guaranteed to be legal input for tree sequences that's all that really matters. Canonical must be canonical though, there can be only one output from canonicalise.

Ok, wise words there.  Here's my proposal, then:
1. make it so `subset` doesn't reorder the individual table (this is easy and makes it easier to understand anyhow)
2. let `sort` do whatever (so: this PR is fine)
3. change `canonicalize` so that it sorts individuals on `(num descendants, id of first referring node)` - this will be a bit annoying but straightforward

I've given it a fair bit of thought and come up with no better solution.

jeromekelleher:
Sounds like a plan, thanks @petrelharp.

So, I guess we merge the sort stuff here and file some issues to track the other two (which we resolve before release)?

petrelharp:
I've made a good start at `subset` and `canonicalise` (in #1226), and turned up another issue (`union`); I think I'll wait until this is merged to finish since I'm starting to step on the toes of this PR.

benjeffery:
Needs a squash and a changelog, but I finally think this is ready for review. I had to do the subset individual re-mapping (#1223 ) to get the tests to pass.

c/tskit/tables.c
```diff
@@ -4954,13 +4954,132 @@ tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
...
 }
 
+static int
+tsk_table_sorter_sort_individuals(tsk_table_sorter_t *self)
+{
+    int ret = 0;
+    tsk_id_t i;
+    tsk_individual_table_t copy;
+    tsk_individual_t individual;
+    tsk_individual_t copy_individual;
```
petrelharp, line 4964: you don't use `individual` and `copy_individual` at the same time, so maybe they should both be `individual`?

python/tests/tsutil.py
```diff
@@ -1077,6 +1093,18 @@ def py_sort(tables, use_num_descendants=False):
...
+    for mig_id in sorted_migs:
+        tables.migrations.add_row(
+            copy.migrations[mig_id].left,
+            copy.migrations[mig_id].right,
+            copy.migrations[mig_id].node,
+            copy.migrations[mig_id].source,
+            copy.migrations[mig_id].dest,
+            copy.migrations[mig_id].time,
+            copy.migrations[mig_id].metadata,
+        )
```
petrelharp, line 1105: did you mean to put the migration sort stuff in here?

petrelharp (approved):
Looks good! Very small comments, and: how about throwing in a test that inserts some `NULL` parents in addition to the actual parents? (Or, maybe we should just disallow them.)

python/tests/tsutil.py
benjeffery: Yes, otherwise a comparison of the python and c sorting fails.

c/tskit/tables.c
```diff
@@ -4954,13 +4954,132 @@ tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
...
+        individuals_todo[i] = TSK_NULL;
+        new_id_map[i] = TSK_NULL;
+    }
+    individuals_todo[num_individuals] = TSK_NULL; /* Sentinel value */
+
+    /* First find the set of individuals that have no children by creating
+     * an array of incoming edge counts */
+    for (i = 0; i < (tsk_id_t) individuals->parents_length; i++) {
+        if (individuals->parents[i] != TSK_NULL) {
+            incoming_edge_count[individuals->parents[i]] += 1;
```
jeromekelleher, line 4996: nit - might as well use ``++`` here for consistency

c/tskit/tables.c
```diff
@@ -4954,13 +4954,132 @@ tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
...
+
+    ret = tsk_individual_table_clear(individuals);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* The sorted individuals are in reverse order */
+    for (i = (tsk_id_t) num_individuals - 1; i >= 0; i--) {
+        tsk_individual_table_get_row_unsafe(
+            &copy, individuals_todo[i], &copy_individual);
```
jeromekelleher, line 5041: I agree with Peter - ``copy_individual`` doesn't really add any clarity over ``individual`` here.

c/tskit/tables.c
```diff
@@ -4954,13 +4954,132 @@ tsk_table_sorter_sort_mutations_canonical(tsk_table_sorter_t *self)
...
+    }
+
+    /* Rewrite the parent ids */
+    for (i = 0; i < (tsk_id_t) individuals->parents_length; i++) {
+        if (individuals->parents[i] != TSK_NULL) {
+            individuals->parents[i] = new_id_map[individuals->parents[i]];
+        }
+    }
+    /* Rewrite the node individual ids */
+    for (i = 0; i < (tsk_id_t) self->tables->nodes.num_rows; i++) {
```
jeromekelleher, line 5059: Might be worth taking a reference to the node table above (like we do for the individual table) to avoid a bit of pointer chasing in this loop.

c/tskit/tables.c
```diff
@@ -5014,6 +5141,9 @@ tsk_table_sorter_run(tsk_table_sorter_t *self, const tsk_bookmark_t *start)
             goto out;
         }
     }
+    if (!skip_individuals) {
+        ret = tsk_table_sorter_sort_individuals(self);
```
jeromekelleher, line 5145: I know it's redundant, but let's stick the ``if (ret != 0)`` stanza in here. It's easy to miss in diffs if someone updates the code subsequently, and forgets to put it in.

c/tskit/tables.c
```diff
@@ -9746,6 +9876,13 @@ tsk_table_collection_subset(tsk_table_collection_t *self, const tsk_id_t *nodes,
         }
     }
 
+    /* Rewrite the individual parent ids */
+    for (k = 0; k < (tsk_id_t) self->individuals.parents_length; k++) {
+        if (self->individuals.parents[k] != TSK_NULL) {
+            self->individuals.parents[k] = individual_map[self->individuals.parents[k]];
```
jeromekelleher, line 9882: We should try to cover this in tests? (or is that coming in one of the follow-ups?)

python/tests/tsutil.py
```diff
@@ -1002,6 +1002,19 @@ def cmp_edge(i, j, tables):
     return ret
 
 
+def cmp_migration(i, j, tables):
+    ret = tables.migrations.time[i] - tables.migrations.time[j]
```
jeromekelleher, line 1006: It doesn't really matter I'd imagine, but 
```
migration_i = tables.migrations[i]
migration_j = tables.migrations[j]
ret = migration_i.time - migration_j.time
# etc
```
would avoid repeatedly building and returning numpy arrays for each of the columns.

python/tests/tsutil.py
```diff
@@ -1651,3 +1679,55 @@ def assert_tables_equal(t1, t2, label=""):
             f"{label}: t1.num_rows {t1.num_rows} != {t2.num_rows} t2.num_rows"
         )
     assert t1 == t2
+
+
+def sort_individual_table(tables):
+    """
+    Sorts the indivial table by parents-before-children.
```
jeromekelleher, line 1686: typo, indivial

jeromekelleher (approved):
Very nice, a few minor comments. The algorithm is quite beautiful!

jeromekelleher:
> Looks good! Very small comments, and: how about throwing in a test that inserts some NULL parents in addition to the actual parents? (Or, maybe we should just disallow them.)

I think we should allow NULL parents - there will be times when having parents ``[-1, -1]`` will be handy and others when parents=``[]`` is more useful. We should allow both, I think.

c/tskit/tables.c
petrelharp: yes, this is in #1226.

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

python/tests/tsutil.py
benjeffery: Done

PR #1201, opened 2021-02-17
Refer to nodes, not samples in docstring

hyanwong:
The docs confused me. However, I wonder if we should call the parameter `nodes` rather than `samples` (and add `samples as a deprecated alias)? Since these are all keyword only, it shouldn't break any code.


jeromekelleher (approved):
LGTM

jeromekelleher:
> However, I wonder if we should call the parameter nodes rather than samples (and add `samples as a deprecated alias)? Since these are all keyword only, it shouldn't break any code.

Possibly - let's file an issue to track. It's not cut and dried though, as you saw, since samples and non-samples are treated differently.

PR #1202, opened 2021-02-17
Allow nodes to be specified in genotype_matrix()

### LINKED ISSUE #1200: Less memory-intensive method for getting a single haplotype, including ancestral haplotypes
A number of use-cases have come up recently where people want to get a single ancestral haplotype out of a tree sequence. The quick-and-dirty way to do this, as used by @saurabhbelsare , is to change the flag on the ancestral node to mark it as a sample, and then use the `haplotypes` iterator. This (presumably) will report the unknown sections as "missing data". I wonder if we want a `.haplotype()` method that gets a single haplotype (possibly an ancestral one) from a tree sequence.

Alternatively, and this is maybe better, we could wrap the method suggested above by modifying the `.haplotypes()` iterator to (optionally) take a list of nodes: `haplotypes(nodes=None, *, ...)`. If `nodes` is None, we default to the sample nodes, as currently, but if `nodes` is not None it can be a list of any nodes, including non-sample ones. In this case, we take a copy of the tables, mark the specified nodes as samples (and all other nodes as non-samples), simplify(), and output the haplotypes. This would also be a way to allow people to output a few sample haplotypes without slurping the whole variant matrix into memory (at the moment we advise that if only a few haplotypes are needed, the user should simplify first, but we don't actually do the work for them).

At the moment all the parameters to `ts.haplotypes()` are keyword only, so this won't break API compatibility.

hyanwong:
And also allow ancestral haplotypes to be output.

Fixes #1200

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -4248,20 +4248,11 @@ class TreeSequence:
         if nodes is None:
             nodes = self.samples()
 
-        ts = self
-        if isolated_as_missing and len(np.setdiff1d(nodes, ts.samples())) > 0:
-            # Can't use the variants method if isolated_as_missing is True and
-            # there are non-sample nodes, so simply mark the non-sample nodes as samples
-            tables = ts.dump_tables()
-            flags = tables.nodes.flags
-            flags &= np.invert(np.array([NODE_IS_SAMPLE], dtype=flags.dtype))
-            flags[nodes] |= NODE_IS_SAMPLE
-            tables.nodes.flags = flags
-            ts = tables.tree_sequence()
-
-        H = np.empty((ts.num_samples, ts.num_sites), dtype=np.int8)
+        H = np.empty((self.num_samples, self.num_sites), dtype=np.int8)
         missing_int8 = ord(missing_data_character.encode("ascii"))
-        for var in ts.variants(samples=nodes, isolated_as_missing=isolated_as_missing):
+        for var in self.variants(
+            samples=nodes, isolated_as_missing=isolated_as_missing
+        ):
             alleles = np.full(len(var.alleles), missing_int8, dtype=np.int8)
             for i, allele in enumerate(var.alleles):
                 if allele is not None:
@@ -4401,13 +4392,19 @@ class TreeSequence:
             yield Variant(site, alleles, genotypes)
 
     def genotype_matrix(
-        self, *, isolated_as_missing=None, alleles=None, impute_missing_data=None
+        self,
+        nodes=None,
+        *,
+        isolated_as_missing=None,
+        alleles=None,
+        impute_missing_data=None,
     ):
         """
         Returns an :math:`m \\times n` numpy array of the genotypes in this
-        tree sequence, where :math:`m` is the number of sites and :math:`n`
-        the number of samples. The genotypes are the indexes into the array
-        of ``alleles``, as described for the :class:`Variant` class.
+        tree sequence, where :math:`m` is the number of sites and :math:`n` is
+        the number of nodes (i.e. by default, the number of samples). The genotypes
+        are the indexes into the array of ``alleles``, as described for the
+        :class:`Variant` class.
 
         If isolated samples are present at a given site without mutations above them,
         they will be interpreted as :ref:`missing data<sec_data_model_missing_data>`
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Creating an intermediate copy isn't something we should do, IMO. I don't think this feature is worth spending much (any?) time on, it's a niche requirement and the ``genotype_matrix`` is nearly always more useful.

hyanwong:
> Creating an intermediate copy isn't something we should do, IMO. I don't think this feature is worth spending much (any?) time on, it's a niche requirement and the `genotype_matrix` is nearly always more useful.

Yeah, it wasn't very nice. But I think that the ability to extract ancestral haplotypes *is* useful. I'm nearly always thinking in terms of haplotypes, not sites.

jeromekelleher:
Yes, but the point is you're better off doing it with the genotype matrix (a numpy array) rather than a list of strings (.haplotypes()). You can do ``ts.genotype_matrix().T`` to get the haplotype matrix. This was the main reason I didn't add the extra options to .haplotypes, because working with strings is usually not what you want to do anyway.

hyanwong:
> Yes, but the point is you're better off doing it with the genotype matrix (a numpy array) rather than a list of strings (.haplotypes()). You can do `ts.genotype_matrix().T` to get the haplotype matrix. This was the main reason I didn't add the extra options to .haplotypes, because working with strings is usually not what you want to do anyway.

ISWYM. But `genotype_matrix` doesn't allow subsetting down to certain genomes. Perhaps it should, if a common use case is to find the genotypes for a subset of genomes/nodes? If so, it's unclear to me if we would want to allow non-sample nodes in there too? I'm happy to code up something like this for genotype_matrix(), but that would also involve taking a copy and marking only some nodes as samples, so the issue is the same.

If we don't provide this in tskit, it does make me think that it would be very useful to have some snippets of code somewhere semi-official that show how to do certain tasks (such as extracting ancestors), so that users don't end up recoding the methods from scratch each time. As I said, a few people now have wanted the ability to get the ancestral "haplotypes" (not necessarily the string version, though) out of a tree sequence. And I would find this useful too.

petrelharp:
I like the proposal of adding a `nodes` argument to `genotype_matrix`, that would allow non-sample nodes to be passed in. This is a nice small stop-gap recipe, though, in the meantime.

hyanwong:
> I like the proposal of adding a `nodes` argument to `genotype_matrix`, that would allow non-sample nodes to be passed in. This is a nice small stop-gap recipe, though, in the meantime.

If, in the case where we do `.genotype_matrix(nodes != None)`, we are happy to take a copy of the TS, flip the required flags, and output again, then this is a trivial moving of the code in this PR to the `genotype_matrix` function instead.

hyanwong:
Commit https://github.com/tskit-dev/tskit/pull/1202/commits/7f29b9b2a33de16a117d92db4a4603e4e1bb2dcf puts the code into the genotype_matrix method instead. As @jeromekelleher says, it's not terribly efficient to take a copy of the entire TS, but this is only done if we provide a `nodes` argument. Also, we only touch the flags and don't do a `simplify()`, so it shouldn't be CPU intensive, only wasteful of memory.

hyanwong:
It might save some fiddly code, and allow ancestral haplotypes to be output, if we simply used the genotype_matrix() function within `haplotypes()` instead of constructing the haplotypes by going through the variants iterator. The memory use should be the same, as we have to allocate an m x n matrix of int8 values either way (and we could loop over the genotype matrix rows, substituting in the allelic letters, rather than allocating an entire new array for the characters.

petrelharp:
I think this sort of useful hack would be good as a recipe somewhere (maybe as an answer to a discussion question?), but not actually as a method in tskit.

hyanwong:
> I think this sort of useful hack would be good as a recipe somewhere (maybe as an answer to a discussion question?), but not actually as a method in tskit.
> 
> _Edit:_ I mean the code in this PR, not the idea of adding arguments.

You mean, I should stick the code in a recipe in a discussion, but we should (in the longer term) think about adding a `nodes` argument to genotype_matrix() that does it in a less hacky way (presumably in the C code, as that's called almost immediately in the genotype_matrix() method)? If so, SGTM.

petrelharp:
Yes, exactly!

hyanwong:
https://github.com/tskit-dev/tskit/discussions/1206

PR #1208, opened 2021-02-17
Clarify that simplify() flags nodes as samples

hyanwong:
It wasn't clear to me that the actual flag values were changed on simplification(). I think this is the correct description (i.e. nodes that were internal samples are no longer flagged as samples after simplify(), if they were not passed in the the `samples` parameter.


petrelharp (approved):
LGTM! Thanks!

jeromekelleher (approved):
LGTM

PR #1211, opened 2021-02-19
Do not set samples in TreeSequence.simplify()

hyanwong:
I just noticed this: if `samples` is None, I think we should probably pass it straight through to Tables.simplify(), rather than have 2 bits of code, one in `TreeSequence.simplify()` and one in `Tables.simplify()` that do the same thing? That way we are also sure that the methods always do the same thing.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1106,10 +1106,19 @@ class TestTreeSequence(HighLevelTestCase):
                 a2 = [variant.alleles[g] for g in variant.genotypes]
                 assert a1 == a2
 
+    def verify_tables_api_equality(self, ts):
+        for samples in [None, list(ts.samples()), ts.samples()]:
+            tables = ts.dump_tables()
+            tables.simplify(samples=samples)
+            assert tables.equals(
+                ts.simplify(samples=samples).tables, ignore_timestamps=True
+            )
+
     @pytest.mark.slow
     def test_simplify(self):
         num_mutations = 0
         for ts in get_example_tree_sequences():
+            self.verify_tables_api_equality(ts)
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I guess, yeah. Do we have tests that confirm they do the same thing with ``samples=None`` already?

hyanwong:
> I guess, yeah. Do we have tests that confirm they do the same thing with `samples=None` already?

https://github.com/tskit-dev/tskit/blob/a23b5b4fddd528be2ef94ab317632e35addf2c44/python/tests/test_tables.py#L2205 tests that passing `None` is the same as passing `ts.samples()` so, I guess so, yeah.

jeromekelleher:
Yes, but that's not checking that ``ts.simplify().tables == tables.simplify()``, is it? To make sure that this change isn't subtly changing the behaviour we'd want to have a test case in doing that **before** you make the change and make sure the change doesn't break afterwards.

It's very rare that a PR is going to to through that has code changes and no tests.

hyanwong:
No - I misread your comment. I thought you meant that tables.simplify() (or ts.simplify()) does the same thing passing `None` or passing `ts.samples()`.

hyanwong:
(since, if `None` is passed to ts.simplify(), it simply subs ts.samples() into the parameter, and passes it to `tables.simplify()`)

benjeffery:
I've added a test for equality between the tables and ts API for simplify.

hyanwong:
Thanks Ben. Sorry to leave this one dangling.

benjeffery:
No problem, I'm spring cleaning.

jeromekelleher (approved):
LGTM!

PR #1213, opened 2021-02-22
Add `x_label` parameter for the X axis

hyanwong:
And correct some alignment bugs in SVG placement of trees in a tree sequence, where we accidentally padded the top of the trees in the tree sequence, rather than just the bottom.

This will result in nicer diagrams for the tutorial, and is an obvious addition. The only question is whether we want a label by default or not. It's less of a change, and less language-specific if we simply have no X label by default, so I just left it that way for the time being.

Visual representations to check this is what we want are in the `tests/data/svg` directory.

# PR Checklist:

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1918,6 +1918,36 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
             expected_svg = file.read()
         self.assertXmlEquivalentOutputs(svg, expected_svg)
 
+    def test_known_svg_ts_plain(self):
+        """
+        Plain style, with no background shading and a variable X axis
+        """
+        ts = self.get_simple_ts()
+        svg = ts.draw_svg(x_scale="treewise")
+        # Prettify the SVG code for easy inspection
+        svg = xml.dom.minidom.parseString(svg).toprettyxml()
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "ts_plain.svg"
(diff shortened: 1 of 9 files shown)


python/tests/test_drawing.py
```diff
@@ -1918,6 +1918,34 @@ def test_known_svg_ts(self):
...
+    def test_known_svg_ts_with_xlabel(self):
+        """
+        Style with X axis label
+        """
+        ts = self.get_simple_ts()
+        svg = ts.draw_svg(x_label="genomic position (bp)")
+        # Prettify the SVG code for easy inspection
+        svg = xml.dom.minidom.parseString(svg).toprettyxml()
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / "ts_xlabel.svg"
+        self.verify_basic_svg(svg, width=200 * ts.num_trees)
```
jeromekelleher, line 1944: Be nice to test that the specified x_label text is in the xml.

jeromekelleher (commented):
Looks good. I can't see any x-label in the example when I view it on github (and search this page for "genomic position" doesn't hit anything in the xml, so I'm not sure it's actually worked?

python/tests/test_drawing.py
hyanwong: Doh: I made a doc change then forgot to pass the parameter through. Makes the point that I should indeed check for the string in the SVG. Will update.

jeromekelleher (approved):
LGTM, very useful

PR #1215, opened 2021-02-23
Add chevrons for sites / mutations

### LINKED ISSUE #1194: Add mutation tickmarks on x-axis in SVG representation
@grahamgower suggested that in the SVG representation of a tree sequence with mutations, it would be good to add a red vertical line (or perhaps cross) on the X axis showing the placement of the mutation along the genome. I think this is a great idea. I don't know if we should also have some text near the tick so that we can map the mutations on the tree to the position on the x axis, but I suspect this would rapidly get annoying, so I think a simple line would be best.

hyanwong:
## Description

Adds indicators for site positions and number of mutations in svg representation, as long as x_scale="physical".

A new test against a known TS is added, at `tests/data/svg/ts_nonbinary.svg`. An example of a site with multiple mutations and a nonvariable site is seen in e.g. `tests/data/svg/ts.svg`

Fixes #1194 

# PR Checklist:

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -45,6 +45,7 @@ from tskit import drawing
 class TestTreeDraw:
     """
     Tests for the tree drawing functionality.
+    TODO - the get_XXX_tree() functions should probably be placed in fixtures
     """
 
     def get_binary_tree(self):
@@ -1851,7 +1852,9 @@ class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
         self.verify_basic_svg(svg)
         assert svg.count("rect") == 10
         assert svg.count('path class="edge"') == 10
-        assert svg.count('path class="sym"') == ts_no_edges.num_mutations
+        assert svg.count('path class="sym"') == (
+            ts_no_edges.num_mutations + ts_no_edges.num_sites
+        )
 
     def test_tree_root_branch(self):
         # in the simple_ts, there are root mutations in the first tree but not the second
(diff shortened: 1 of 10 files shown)


python/tskit/drawing.py
```diff
@@ -383,6 +383,28 @@ def __init__(
...
             )
+        if x_scale == "physical":
+            # Add sites as upper chevrons
+            for s in ts.sites():
+                x = axes_left + s.position * drawing_scale
+                muts = [None] if len(s.mutations) == 0 else s.mutations
+                site = axis.add(dwg.g(class_=f"site s{s.id}"))
+                for i, m in enumerate(muts):
+                    mut = dwg.g(
+                        class_="mut nonvariable" if m is None else f"mut m{m.id}"
```
jeromekelleher, line 394: I think we're going through quite a lot of logical contortions here because we're conflating sites and mutations.

jeromekelleher (commented):
I like it, but I think way we're conflating mutations and sites is confusing and unhelpful. How would a colour blind person tell the difference between a site with one or zero mutations?

How about we put in chevrons for the mutations like you've done, but put a light vertical tick for the site? So, we'd have something like
````
                          ^
#####|########|#######
```

The first site has no mutations and the second has one. Nice simple, unambiguous interpretation.

Edit: ugh, lack of monospace fonts. You get the idea

hyanwong:
> I like it, but I think way we're conflating mutations and sites is confusing and unhelpful. How would a colour blind person tell the difference between a site with one or zero mutations?

Yeah, good point. Putting a different symbol seems sensible, although if it is meant to represent a site in general, then there's an argument that *all* sites should have that symbol too. So I'd prefer something that explicitly says "empty site". I worry a little that a vertical line will seem like something different altogether, or even just a tick mark like the ones that appear below the line. So how about an empty downward pointing triangle, or an empty circle, or something?

> How about we put in chevrons for the mutations like you've done, but put a light vertical tick for the site? So, we'd have something like
> 
> ```
>                           ^
> #####|########|#######
> ```

Sorry - by "tick" do you mean a vertical line? Or a checkmark?

python/tskit/drawing.py
hyanwong: Are we? I thought this was quite logical. Each position has a site. The site contains (possibly multiple) mutations. The mutations contain a symbol.

It's true that an "empty" site still has a group labelled `<g class="mut nonvariable">...`, which I agree is wrong. We should probably just have a plain symbol without the "mut" class. Perhaps that's what you mean?

jeromekelleher:
> then there's an argument that all sites should have that symbol too.

Yes, we should represent all sites in the SiteTable. I don't think we should make any distinction between "empty" sites or not, the symbol appears at the given position if there is a site in the site table. I don't have strong opinions on what the symbol should be, just that it should exist for all sites (in the site table) and be visually distinct from mutations.

hyanwong:
Ah, I see. we could have a light grey check *underneath* the chevrons. Good idea. I'll see what that looks like.

hyanwong:
You mean something like this, @jeromekelleher ? Are the grey "site" lines too long?

Screenshot 2021-02-23 at 16 43 58

jeromekelleher:
Yes, that looks great! I would have left the tick go a little bit below the X-axis (say, half as far as the tree position ticks), so that we still see them even if there's a stack of chevrons. The also look a little tall, to my eye.

Not strong opinions though, happy to go with your aesthetics.

hyanwong:
> Yes, that looks great! I would have left the tick go a little bit below the X-axis (say, half as far as the tree position ticks), so that we still see them even if there's a stack of chevrons. 

I think I prefer keeping the different sides of the axis for different things, so they don't get mixed up. We might (in the future) want minor ticks on the underside e.g. for regularly spaced tick marks at 100, 200, 300 etc, so I think better not to use that bit of the axis.

> They also look a little tall, to my eye.

I moved the chevrons up a bit, so they don't impinge on the lower part of the axis. I tried a slightly shorter line and you lose the look of an arrow.

> Not strong opinions though, happy to go with your aesthetics.

I think it's OK now. I do wonder, however, why we are now using grey for the site line. Previously I suggested grey to mark an "empty" site, but since all sites have the line, I think should probably switch back to black? Here's that that looks like.

Screenshot 2021-02-23 at 19 29 32

jeromekelleher:
Black looks good. I'm happy to merge if you are?

hyanwong:
> Black looks good. I'm happy to merge if you are?

Great. Let me push those changes and then you can merge. I'll ping you here.

hyanwong:
OK, this is done, and passes on my machine. The current failure is something to do with python 3.9 on OS X, and I've seen it before. I'll install 3.9 here and see if I can replicated.

python/tests/test_drawing.py
```diff
@@ -1881,56 +1884,58 @@ def test_tree_root_branch(self):
...
         assert "H 0" in snippet2b
 
+    def verify_known_svg(self, svg, filename, **kwargs):
+        # expected SVG files can be inspected in tests/data/svg/*.svg
+        svg = xml.dom.minidom.parseString(
+            svg
+        ).toprettyxml()  # Prettify for easy viewing
+        self.verify_basic_svg(svg, **kwargs)
+        svg_fn = pathlib.Path(__file__).parent / "data" / "svg" / filename
+        # with open(svg_fn, "wt") as file:  # Uncomment to save new expected files
```
benjeffery, line 1894: This could be implemented as a a pytest flag. No urgent need, maybe next time we touch this code.

benjeffery (approved):
LGTM!

benjeffery:
@mergifyio rebase

python/tests/test_drawing.py
hyanwong: Good point. I've added this as a comment. Let's see if it sneaks in before automerge.

jeromekelleher (approved):

PR #1218, opened 2021-02-24
swap deprecated np.int and np.float for int & float

hyanwong:
Recent versions of numpy deprecate np.int and np.float for simply `int` and `float`.


jeromekelleher (approved):
Lgtm

PR #1230, opened 2021-03-03
Set mutation positions using mutation time if available

### LINKED ISSUE #840: draw_ methods ignore mutation_time
At the moment, `draw_svg()` spaces apart mutations evenly on a branch. Now that we have introduced mutation times we should really use the time value (if defined) to place mutations on branches when drawing trees. 

This would be easy, apart from the fact that there may be mutations above the root. In this case, we need to scale the tree so that, rather than a fixed root branch length (which is a hack anyway), the tree height is given by the oldest mutation above the root node. When plotting a whole tree sequence, we need to find the oldest of these values. Unfortunately, this is not quite as easy as finding `max(ts.tables.mutations.time)`, because I think we can have valid mutations that are not above any node on any tree. We previously hit this issue for node times, and solved it by defining a `ts.max_root_time` property. I wonder if we now need a `ts.max_mutation_time` property too?

hyanwong:
## Description

The first step to putting a y-axis on SVG plots.

Fixes #840

# PR Checklist:

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -24,12 +24,15 @@
 Module responsible for visualisations.
 """
 import collections
+import logging
 import math
 import numbers
+import operator
 
 import numpy as np
 import svgwrite
 
+import tskit.util as util
 from _tskit import NULL
 
 LEFT = "left"
@@ -126,6 +129,16 @@ def rnd(x):
         return x
     digits -= math.ceil(math.log10(abs(x)))
     return round(x, digits)
+
+
+def identity(x):
+    return x
+
+
+def log_transform(x):
+    # add 1 so that don't reach log(0) = -inf error.
+    # just shifts entire timeset by 1 unit so shouldn't affect anything
+    return np.log(x + 1)
 
 
 def add_text_in_group(dwg, elem, x, y, text, **kwargs):
@@ -265,7 +278,7 @@ class SvgTreeSequence:
         root_group = dwg.add(dwg.g(class_="tree-sequence"))
         if x_scale == "physical":
             background = root_group.add(dwg.g(class_="background"))
-            axis_top_pad = 20
+            axis_top_pad = 15
             tick_len = (0, 5)
         else:
             axis_top_pad = 5
@@ -347,6 +360,8 @@ class SvgTreeSequence:
             if x_scale == "treewise":
                 x = tree_x
             elif x_scale == "physical":
+                # Shift diagonal lines between tree & axis into the treebox a little
+                backgd_pad_y = axis_top_pad + svg_trees[0].treebox_y_offset
                 x = break_x
                 if i > 0 and i % 2 == 1:
                     # draw an alternating grey background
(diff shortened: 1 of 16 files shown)


hyanwong:
This took a bit of reworking to include mutations above the root. But worth sorting as this will impact implementing a Y-axis scale, which we can now do by looking at `node_y_coord_map` and `mut_y_coord_map`. We should first implement a Y-scale on trees, and then generalise to tree sequence. This PR also includes a few nice example plots in `tests/data/svg`

Screenshot 2021-03-03 at 12 27 09

Log scale:

Screenshot 2021-03-03 at 12 27 30

python/tskit/drawing.py
```diff
@@ -485,19 +490,21 @@ def __init__(
...
+        nodes = set(tree.nodes())
         for site in tree.sites():
             for mutation in site.mutations:
-                self.node_mutations[mutation.node].append(mutation)
-                if tree.parent(mutation.node) == NULL:
-                    self.mutations_over_root = True
+                if mutation.node in nodes:
+                    self.node_mutations[mutation.node].append(mutation)
+                    if tree.parent(mutation.node) == NULL:
+                        self.mutations_over_roots = True
```
jeromekelleher, line 500: We should probably raise a ValueError here if ``mutation.node not in nodes`` - better to fail noisily than to silently hide mutations

python/tskit/drawing.py
```diff
@@ -554,6 +561,16 @@ def __init__(
                 self.add_class(self.mutation_label_attrs[m], "lab")
         self.draw()
 
+    @staticmethod
```
jeromekelleher, line 564: These would be clearer as raw functions at the module level - ``self.identity()`` reads like "I want the identity of this object" and ``ClassName.identity()`` is similar.

jeromekelleher (commented):
Looks good, although haven't delved into all the details. Getting full coverage in tests will be a faff.

python/tskit/drawing.py
hyanwong: Good point. I initially defined them as lambdas, but black (or some other CI thing) didn't like that.

python/tskit/drawing.py
hyanwong: True. Note that we don't hide them in the "tree sequence" view: they are present on the X axis:

Screenshot 2021-03-03 at 12 39 30

But that's a bit obscure, so raising an error (or at least a warning) seems reasonable.

hyanwong: Thinking about this, I'm inclined to issue a warning instead of raising an error,. After all, we don't currently raise an error when e.g. there are nodes in the node table which are not part of any tree. So I don't think there currently an expectation that we *need* to show all the contents of the tables, is there?

hyanwong:
> Looks good, although haven't delved into all the details. Getting full coverage in tests will be a faff.

What extra coverage would you like? E.g. should we check that a mutation high above a root is still within the tree box, and that the root branch is long enough? You can see this in the example SVGs, but it's not explicitly checked, which would be sensible, I guess. I'm not sure what else would be useful. Perhaps an example from the WF simulator for some multi-root trees, especially if we can make the JukeCantor routine place mutations at defined times?

python/tskit/drawing.py
jeromekelleher: Sure, warning makes sense.

jeromekelleher:
My bad, looks like you have 100% coverage with the examples.

hyanwong:
> My bad, looks like you have 100% coverage with the examples.

OK, review comments done. Probably worth saving more extensive testing until we have y-scales, when it should be a bit more obvious where the mutations are, whether it's a log scale, etc. So maybe merge this if it's OK, and I'll get on with that.

hyanwong:
Rolled into #1236 too, so we could either close this PR and review all the changes together over there, or merge this first and rebase. Either is fine by me.

jeromekelleher:
Looks like a fair bit has changed in #1236, so will probably be reviewing twice if we go through it here too?

hyanwong:
> Looks like a fair bit has changed in #1236, so will probably be reviewing twice if we go through it here too?

Yeah, closing.

PR #1231, opened 2021-03-03
Make individual parents an optional column

### LINKED ISSUE #1212: make individual parent column optional and add previous flie format to tests
In testing pyslim, I noticed that current tskit `main` doesn't read older files, I think because of the missing individual parent column. I couldn't find fixing this backwards-compatibility thing on a list to do anywhere, so I'm posting it here, although I have probably just missed it.

More generally, I notice that in `tests/test_file_format.py` there's tests that load example files with the old hdf5 file format, but no saved examples of post-hdf5 trees files, although we've been through a few file versions. Perhaps we should add an example .trees file to `tests/data/old_formats` and a verification test each time we bump the file format?

benjeffery:
Fixes #1212


python/tests/test_file_format.py
```diff
@@ -240,6 +243,22 @@ def verify_tree_sequence(self, ts):
...
+            "edges",
+            "sites",
+            "migrations",
+            "mutations",
+        ]:
+            t = getattr(ts.tables, table)
+            assert t.num_rows > 0
+            assert t.metadata_schema == tskit.MetadataSchema({"codec": "json"})
+            assert t[2].metadata == f"n_{table}_2"
+        ts.tables.has_index()
```
jeromekelleher, line 260: Missing assert?

jeromekelleher (commented):
LGTM. We should add this pair of cols to the ``test_missing_optional_column_pairs`` test in test_file_formats.c as well though.

benjeffery:
Fixed the C tests and the assert, should be good to go.

jeromekelleher (approved):
LGTM

PR #1226, opened 2021-02-28
make subset and canonicalise deal with individual parents

petrelharp:
As discussed in  #1199, the plan is for 
1. subset to not reorder individuals
2. canonicalise to sort on (num descendants, first referring node)

WIP.

Issue #1225, opened 2021-02-28
union needs to remap individual parent IDs


petrelharp:
To do this, we need to first build the map of which individual IDs are shared between `self` and `other`, by looping over the shared nodes, [here](https://github.com/tskit-dev/tskit/blob/20c423f055f03b84c98980e6e99febba9a6217d9/c/tskit/tables.c#L9887).  (And, we don't have tests with individual parents to test this.)

(See also #1223.)

petrelharp:
Related: https://github.com/tskit-dev/tskit/pull/1226

benjeffery:
Just working through this, so individuals whose parents are no longer part of the tree sequence will have them set to `NULL`?

benjeffery:
Doh that was a bad question, I get it now.

benjeffery:
Ok, here's my shot at a better question. What happens when individuals whose nodes are marked as being equivalent in `other_node_mapping` have parents that are of nodes that are not? This smells like it should be an error?

jeromekelleher:
That does sound like it should be an error, but it might be tricky to actually detect? How much consistency checking are we doing now to make sure those nodes are actually equivalent?

jeromekelleher:
Well, I guess I'd do individual checks in there, to the extent that other aspects of node identity are checked?

petrelharp:
> Ok, here's my shot at a better question. What happens when individuals whose nodes are marked as being equivalent in other_node_mapping have parents that are of nodes that are not? This smells like it should be an error?

I don't quite understand the question: what's "have parents that are of nodes that are not"?

benjeffery:
I mean that the structure of node parents is not compatible with the structure of individual parents. So you could have parent individuals of node A in `other` whose nodes are not mapped to the nodes in `self` that are the parent nodes of the node that is mapped to A.

petrelharp:
Ah. I say we ignore that. There's very few - if any - situations we can be sure of an error. Suppose for instance that A is the parent of B, but only one of A's nodes is present, and B has inherited only from the other of A's nodes; this would look like 'inconsistent parents' unless we notice that we don't have all of A's nodes (and to do that, we'd have to assume a ploidy for A).

PR #1233, opened 2021-03-04
Remove inconsistant mutation error

### LINKED ISSUE #1180: Support silent mutations
Silent mutations are mutations which don't change the state i.e., two consecutive mutations with derived_state = "A".

Currently, support for these mutations is in a bit of grey area. We don't detect their existence at load time (see #1152) but we will throw an error when someone tries to decode the genotypes.

The original motivation for throwing this error was to help people find errors in their tree sequences - silent mutations seemed like something you would only do by mistake. However, it turns out that there's good mathematical reasons for wanting to have silent mutations from the perspective of simulating mutational processes (https://github.com/tskit-dev/msprime/issues/1440) and in particular trying to overlay multiple mutational processes is made hugely more complex if we **don't** have silent mutations. 

The changes we need to make should be minor:

- Remove the error raised during genotypes
- Update the test suite to make sure we have coverage of using silent mutations in various places (e.g., in the stats api)
- Update docs to clarify what silent mutations are and their support status.

See also: #1152, #325, https://github.com/tskit-dev/msprime/issues/1440

benjeffery:
Fixes #1180 
WIP, needs stats tests

### DIFF
--- a/c/tskit/core.h
+++ b/c/tskit/core.h
@@ -247,13 +247,12 @@ not found in the file.
 #define TSK_ERR_MUTATION_PARENT_EQUAL                               -501
 #define TSK_ERR_MUTATION_PARENT_AFTER_CHILD                         -502
 #define TSK_ERR_MUTATION_PARENT_INCONSISTENT                        -503
-#define TSK_ERR_INCONSISTENT_MUTATIONS                              -504
-#define TSK_ERR_UNSORTED_MUTATIONS                                  -505
-#define TSK_ERR_NON_SINGLE_CHAR_MUTATION                            -506
-#define TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE                     -507
-#define TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION            -508
-#define TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE                -509
-#define TSK_ERR_MUTATION_TIME_HAS_BOTH_KNOWN_AND_UNKNOWN            -510
+#define TSK_ERR_UNSORTED_MUTATIONS                                  -504
+#define TSK_ERR_NON_SINGLE_CHAR_MUTATION                            -505
+#define TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE                     -506
+#define TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_MUTATION            -507
+#define TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE                -508
+#define TSK_ERR_MUTATION_TIME_HAS_BOTH_KNOWN_AND_UNKNOWN            -509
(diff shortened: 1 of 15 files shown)


c/tskit/core.h
```diff
@@ -247,7 +247,6 @@ not found in the file.
 #define TSK_ERR_MUTATION_PARENT_EQUAL                               -501
 #define TSK_ERR_MUTATION_PARENT_AFTER_CHILD                         -502
 #define TSK_ERR_MUTATION_PARENT_INCONSISTENT                        -503
-#define TSK_ERR_INCONSISTENT_MUTATIONS                              -504
```
petrelharp, line 250: should probably decrement subsequent error codes?

jeromekelleher: I'm good either way - we definitely won't do that sort of thing after 1.0, but we're not there yet so :shrug:

c/tskit/genotypes.c
```diff
@@ -333,7 +330,7 @@ tsk_vargen_update_genotypes_i8_sample_list(
             index = list_next[index];
         }
     }
-out:
+
```
jeromekelleher, line 333: These functions can no long error, so we should make them void. Normally wouldn't bother, but this is perf sensitve.

jeromekelleher (commented):
LGTM, minor comments

c/tskit/genotypes.c
benjeffery: The return value appears to be a count of the number of missing genotypes, which is used to set the `has_missing_data` flag.

c/tskit/core.h
benjeffery: I've renumbered in https://github.com/tskit-dev/tskit/pull/1233/commits/614c001afe1d468b9c3698a2c4196e415a910202

c/tskit/genotypes.c
jeromekelleher: Ah - my bad.

jeromekelleher:
Good, that's a sign it's working! Let me know if you have any questions/would like me to take a look at anything.

benjeffery:
Ah, the issue is that a site with only silent mutations has a blank vcf `ALT` column. I guess it should be `.`.

benjeffery:
Last failure is due to this change exposing #1225 to the test suite. Will fix that in seperate PR and merge before this one.

petrelharp:
Oh dear, this means I should finish up https://github.com/tskit-dev/tskit/pull/1226

jeromekelleher:
What's the status here @benjeffery? Should we skip some tests here in this PR so we can merge it, then rebase #1226 and fix there?

benjeffery:
This only needs #1225, which I'm working on. Should be done in soon, so will just wait and rebase.

benjeffery:
#1225 is getting more complicated than I realised so have added a skip here.

jeromekelleher (approved):
LGTM

Issue #580, opened 2020-05-06
[Drawing] Add Y axis ticks to TreeSequence.draw_svg()


jeromekelleher:
TreeSequence.draw_text() has y axis ticks for each node time:

```
4.19┊   7     ┊         ┊
    ┊ ┏━┻━┓   ┊         ┊
0.21┊ ┃   ┃   ┊   6     ┊
    ┊ ┃   ┃   ┊ ┏━┻━┓   ┊
0.08┊ ┃   5   ┊ ┃   5   ┊
    ┊ ┃ ┏━┻┓  ┊ ┃ ┏━┻┓  ┊
0.05┊ ┃ ┃  4  ┊ ┃ ┃  4  ┊
    ┊ ┃ ┃ ┏┻┓ ┊ ┃ ┃ ┏┻┓ ┊
0.00┊ 1 3 0 2 ┊ 1 3 0 2 ┊
  0.00      2.00      10.00
```
which I find very handy.  I think we should do the same thing for TreeSequence.draw_svg  --- the only reason I didn't do it is because I ran out of time. This would be a good thing to get in as the default before we release, so we don't have to change default behaviour in the future.

Any objections?

hyanwong:
It would be really helpful in the future, once we decide on `time_units` (https://github.com/tskit-dev/tskit/issues/495) to include this in the label too (perhaps `time (years)`, `time (generations)` etc). It would reduce the scope for all kinds of interpretation error.

PR #1235, opened 2021-03-09
Require Python 3.6

benjeffery:
Fixes #1171

Issue #1171, opened 2021-02-01
Drop python 3.6


benjeffery:
Numpy has now dropped support for python3.6, also msprime. We should consider doing so too. Opinions? 
@jeromekelleher, @petrelharp

jeromekelleher:
I think we should drop it soon. If now is the easiest time to do so (since numpy just has), then let's do it.

PR #1236, opened 2021-03-09
Y axis svg

hyanwong:
## Description

Adds the possibility to put a Y axis & scale onto SVG plots. The axis-adding code is generalised, so it is available to both plots of single trees and whole tree sequences.

~Draft at the moment, to solicit feedback~

Fixes https://github.com/tskit-dev/tskit/issues/580 and #840
# PR Checklist:


hyanwong:
> Looks great. I've not gone through the logic in detail, ping me when you'd like some more feedback.

Thanks @jeromekelleher - I think I've covered most of this now, so some more detailed feedback would be good. If you flip through the SVG files in `tests/data/svg/` (e.g. using a web browser) you should be able to see a number of examples of how it works. I think the main thing to decide is the interface (which is why it's not yet documented).

Re interface, I've tried to retain current behaviour. So `tree.draw_svg()` does the same by default - i.e. is plotted with ought X or Y axes. But if you call `tree.draw_svg(x_label="Pos")` then it adds the entire x-axis, and similarly with `tree.draw_svg(y_label="Time")` (they can be combined).

For a tree *sequence*, the x axis is on by default (and can't be turned off), so that `ts.draw_svg(x_label="Pos")` simply adds the axis label. OTOH, `ts.draw_svg(y_label="Time")` (as with tree-specific version) actually turns the whole axis on.

The other two parameters I've added are `y_ticks` and `y_gridlines`, which have similarly "bespoke" behaviour. For both trees and tree sequences, the default, `y_ticks=None` sets the ticks as the node positions in the tree or the entire ts respectively. This is often a bit ugly, as trees usually have lots of nodes near the tips, although it's a bit nicer when using a log scale. You can also provide an iterable to `y_ticks`, to place ticks wherever you want, e.g. evenly spaced. But I haven't implemented a "tick locator" function that will find a nice set of (say) 8 ticks on a linear or log scale: If you call `y_ticks=8` it simply returns a NotImplementedError.

Gridlines can be force shown by `y_gridlines=True` or force hidden by `y_gridlines=False`. The default `y_gridlines=None` shows gridlines only if a user-specified set of y_ticks has been given. If `y_ticks` is the default (a tick for each unique node height), then gridlines are not shown by default, because the tick positions are likely to be a bit crappy.

python/tskit/drawing.py
```diff
@@ -221,7 +236,284 @@ def remap_style(original_map, new_key, none_value):
         return str(text_tree)
 
 
-class SvgTreeSequence:
+class Plotbox:
```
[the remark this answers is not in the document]
hyanwong: CI currently failing on Python 3.6 because dataclass is only available from Python >= 3.7. I guess by the time this PR is reviewed and merged we'll have [changed the min reqs to Python 3.7](https://github.com/tskit-dev/tskit/issues/1171) anyway.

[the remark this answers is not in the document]
benjeffery: Just merged! https://github.com/tskit-dev/tskit/pull/1235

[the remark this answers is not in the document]
hyanwong: Fast work, thanks @benjeffery !

benjeffery:
Looks great Yan, ping me when you want a full review.

hyanwong:
> Looks great Yan, ping me when you want a full review.

It's ready for review whenever you have time. I think I've covered all the code lines now.

python/tskit/trees.py
```diff
@@ -1608,6 +1613,21 @@ def draw_svg(
         :param bool force_root_branch: If ``True`` always plot a branch (edge) above the
             root(s) in the tree. If ``None`` (default) then only plot such root branches
             if there is a mutation above a root of the tree.
+        :param float symbol_size: Change the default size of the node and mutation
+            plotting symbols. If ``None`` (default) use a standard, sensible size.
+        :param str x_label: If given, add a labelled X axis to the plot, showing the
```
jeromekelleher, line 1618: I'm uncomfortable with using the ``label`` args as the way of turning on and off the axes. Reading code like 
```python
tree.draw_svg(x_label="", y_label="") 
```
it won't be obvious that this is to draw axes with no labels. I think it might be better to have independent ``x_axis=False``, ``y_axis=False`` arguments, even though it makes code a bit more verbose. That way we can give default axis labels ("Time" and "Genome", say), which would probably be what you want most of the time, and lead to more obvious code.

jeromekelleher (commented):
Looks great, but I'm not too keen on the current interface. I think we're trying to do too much with too few parameters. How about
```
x_axis: bool # Whether to show x axis; None for default
y_axis: bool # Whether to show y axis; None for default.
x_label: str # The x axis label; None for default. If x_axis is shown, default is "Genome", otherwise nothing
y_label: str # The y axis label; None for default. If y_axis is shown, default depends on value of time scale, otherwise nothing.
y_gridlines: bool. Whether to plot gridlines. Default is False. (Overly complex making default depend on y_ticks)
```

So, we have a set of relatively independent params. For example, you can, if you want, set a Y-axis label without a drawn y-axis (which you might want to do). By overloading too much functionality into ``label`` parameters I think we'd be (a) making for code that's hard to read and (b) painting ourselves into a corner in terms of flexibility.

What do you think? We don't have to support ``x_axis=False`` for TreeSequence, if that's awkward.

python/tskit/drawing.py
```diff
@@ -125,15 +149,13 @@ def rnd(x):
...
-
-
-def add_text_in_group(dwg, elem, x, y, text, **kwargs):
-    """
-    Add the text to the elem within a group. This allows text rotations to work smoothly
-    """
-    grp = elem.add(dwg.g(transform=f"translate({rnd(x)}, {rnd(y)})"))
-    grp.add(dwg.text(text, **kwargs))
+    x = round(x, digits)
+    try:
```
benjeffery, line 153: `if int(x) == x` would be a way to do this without the try except?

hyanwong:
> Looks great, but I'm not too keen on the current interface. I think we're trying to do too much with too few parameters. 

Yep, I'm completely agnostic about the API. Happy to go with your suggestion below. I don't think we should default to a standard axis label text though, as that opens us up to internationalisation issues  (and we haven't had it before anyway). 

> ```
> x_axis: bool # Whether to show x axis; None for default
> y_axis: bool # Whether to show y axis; None for default.
> x_label: str # The x axis label; None for default. If x_axis is shown, default is "Genome", otherwise nothing
> y_label: str # The y axis label; None for default. If y_axis is shown, default depends on value of time scale, otherwise nothing.
> y_gridlines: bool. Whether to plot gridlines. Default is False. (Overly complex making default depend on y_ticks)
> ```

> What do you think? We don't have to support `x_axis=False` for TreeSequence, if that's awkward.

I think it should be relatively easy to support  `x_axis=False` for TreeSequence, but I assume we would keep the defaults being that `x_axis=None` is equivalent to `x_axis=True` for a TreeSequence but `x_axis=False` for a Tree?

python/tskit/trees.py
```diff
@@ -1608,6 +1613,21 @@ def draw_svg(
...
+            start and end position of this tree along the genome. ``None`` (default)
+            hides the X axis
+        :param str y_label: If given, add a labelled Y axis to the plot, showing the
+            time (or rank, if ``tree_height_scale`` is ``"rank"``) of nodes and
+            mutations in the tree. ``None`` (default) hides the Y axis.
+            hides the X axis
+        :param list y_ticks: A list of y values at which to plot tickmarks (or
+            ``False`` for no tickmarks). If ``None``, plot one tickmark for each unique
```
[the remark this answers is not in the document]
hyanwong: True. Or `0` will, if we ever want to implement a tick locator.

jeromekelleher:
> I don't think we should default to a standard axis label text though, as that opens us up to internationalisation issues (and we haven't had it before anyway).

I think we can break with the old default here, it's generally handy to know that "genome position" (or something) is what the x axis refers to. Definitely for the Y axis having a good default of "Node rank" or "Time ago" would be really helpful, and is what you want most of the time. Unlabelled axes are one of the biggest mistakes made in data visualisation, so we should do the right thing by default here I think.

I wouldn't worry about i8n - we can deal with that if it becomes an issue.

> I think it should be relatively easy to support x_axis=False for TreeSequence, but I assume we would keep the defaults being that x_axis=None is equivalent to x_axis=True for a TreeSequence but x_axis=False for a Tree?

yes, I think that's the right default.

python/tskit/drawing.py
```diff
@@ -247,29 +565,30 @@ def __init__(
...
+        extra_background_shade_spacing = 15 if x_scale == "physical" else 0
+        plotbox_bottom_padding = 20 + extra_background_shade_spacing
+        x_axis_offset = 20  # placement of the x axis line within the padding
+        y_axis_offset = self.default_y_axis_offset
+        if y_label is not None:
+            y_axis_offset += self.line_height
+        if x_label is not None:
+            x_axis_offset += self.line_height
+            plotbox_bottom_padding += self.text_height
+        # treebox_bottom_pad = 0 if  else 14
```
benjeffery, line 591: Still needed?

python/tskit/drawing.py
```diff
@@ -552,99 +860,197 @@ def __init__(
...
+                label=y_label,
+                gridlines=y_gridlines,
+            )
+        self.draw_tree()
+
+    def process_mutations(self, mut_node, lower_bound, upper_bound, ignore_times=False):
+        """
+        Sort the self.node_mutations array for a given mut_node in reverse time order,
+        returning the oldest time.
+        The main complication is with UNKNOWN_TIME values: we replace those with
```
benjeffery, line 905: I'm guessing there is no way to tell that a time was unknown from the SVG? Should we add some indication?

hyanwong:
> I think we can break with the old default here, it's generally handy to know that "genome position" (or something) is what the x axis refers to. Definitely for the Y axis having a good default of "Node rank" or "Time ago" would be really helpful, and is what you want most of the time. Unlabelled axes are one of the biggest mistakes made in data visualisation, so we should do the right thing by default here I think.

If we are creating a default label, I would like it to reflect https://github.com/tskit-dev/tskit/issues/495. When that's solved, the default would change from "Time" to "Time (generations)" or "Time (years)", or "Time (days)" or whatever? That's another reason I was putting off this decision. But I guess we'd be happy for the defaults to change like this in the future?

Also, I assume we don't put "Log time" because the actual ticks will be labelled as (untransformed) values.

> I wouldn't worry about i8n - we can deal with that if it becomes an issue.

OK, your call. It just makes me wary.

python/tskit/trees.py
benjeffery: Agree here.

benjeffery (commented):
Wow, as usual viz code is a ton of work! Just a few comments.

python/tskit/drawing.py
hyanwong: Yep, if you prefer.

python/tskit/drawing.py
hyanwong: Oops, will cull.

python/tskit/drawing.py
hyanwong: Good point. It *should* normally be obvious, because the mutations will be spaced out evenly, but perhaps that's not enough? I could add an extra class to the mutation group, I suppose, to make it available for styling? What do you think @benjeffery ?

benjeffery: A class would be good, as only people with mixed times will care.

hyanwong: The CSS class is now (longwindedly) `unknown_time`, e.g. `<g class="mut m5 s2 unknown_time">`. Guess that's fine though, as it's only duplicated once per mutation.

hyanwong:
OK, it took a fair bit or re-ordering, but I think the new interface with y_axis=True, etc is working now. `y_axis` and `y_label` can be used independently, so you can add an axis label without the axis itself, although I can't imagine that being terribly popular:

Screenshot 2021-03-11 at 18 27 02

As before, have a browse through the files in `tests/data/svg` to see if the format looks right. We can use some of these files (or the general format) in a viz tutorial, I guess.

I think this is ready for a final review?

python/tskit/trees.py
```diff
@@ -1608,6 +1615,24 @@ def draw_svg(
...
+            do not plot a Y axis.
+        :param str x_label: Place a label under the plot. If ``None`` (default) and
+            there is an X axis, create and place an appropriate label.
+        :param str y_label: Place a label to the left of the plot. If ``None`` (default)
+            and there is a Y axis,  create and place an appropriate label.
+        :param list y_ticks: A list of Y values at which to plot tickmarks (``[]``
+            gives no tickmarks). If ``None``, plot one tickmark for each unique
+            node value.
+        :param bool y_gridlines: Whether to plot horizontal lines behind the tree
+            at each y tickmark. If ``None`` (default), only plot gridlines if a list
```
jeromekelleher, line 1634: I think y_gridlines should be independent of y_ticks. It would be annoying to have to turn off y_gridlines if you just wanted specific ticks.

python/tskit/drawing.py
```diff
@@ -117,6 +122,20 @@ def check_x_scale(x_scale):
     return x_scale
 
 
+def check_ticks(ticks, default_iterable):
+    """
```
jeromekelleher, line 126: I don't think this comment is correct? Also seems more complicated than necessary - when do we use the "Autocalculated tick mark locations not implemented." error?

python/tskit/drawing.py
```diff
@@ -221,7 +235,355 @@ def remap_style(original_map, new_key, none_value):
         return str(text_tree)
 
 
-class SvgTreeSequence:
+def add_class(attrs_dict, classes_str):
+    """Adds the classes_str to the 'class' key in attrs_dict, or creates it"""
+    try:
+        attrs_dict["class"] += " " + classes_str
```
jeromekelleher, line 241: Why not use a defaultdict here for these attrs_dicts?

python/tskit/drawing.py
```diff
@@ -221,7 +235,355 @@ def remap_style(original_map, new_key, none_value):
...
+        ".node .mut .sym {stroke-width: 1.5px}"
+        ".tree text, .tree-sequence text {dominant-baseline: central}"
+        ".plotbox .lab.lft {text-anchor: end}"
+        ".plotbox .lab.rgt {text-anchor: start}"
+    )
+
+    # TODO: we may want to make some of the constants below into parameters
+    text_height = 14  # May want to calculate this based on a font size
+    line_height = text_height * 1.2  # allowing padding above and below a line
+    root_branch_fraction = (
```
jeromekelleher, line 344: This looks odd - put the comment on the line above?

jeromekelleher (commented):
Looks great! Some minor comments on the code and interface, but basically ready to merge I think.

Looks like you need a few more permutations of x_axis=[True, False], y_axis=[True, False] to bump the test coverage up (looks like the tests are only covering one side of a branch in some bases)

python/tskit/drawing.py
hyanwong: Good point. I think we *do* want to (eventually) allow e.g. ticks=5 to find roughly 5 equally spaced ticks to use, but I haven't implemented that yet, hence the error to remind us to leave the possibility open. I could simply place that idea in the docstring, though? I like leaving NotImplementedErrors in the code to indicate definite things that need to be done, though.

python/tskit/trees.py
hyanwong: Oh yes, I simply forgot to change the documentation here. Thanks.

python/tskit/drawing.py
hyanwong: Doh, of course. I think this was how it was set up previously, so it's a historical thing. I'll change it.

python/tskit/drawing.py
hyanwong: Oh, this is Black being silly. I'll rephrase.

python/tskit/drawing.py
hyanwong: Ah, I've just had a look, and these dicts contain a mix of strings and floats/ints, as they are passed as kwargs to svgwrite. I've just tried changing them to defaultdict(str) and it makes svgwrite bomb out. So I think we should leave it as-is.

hyanwong:
> Looks like you need a few more permutations of x_axis=[True, False], y_axis=[True, False] to bump the test coverage up (looks like the tests are only covering one side of a branch in some bases)

Just committed this combinatorial monster:

```
@pytest.mark.parametrize("x_axis", (True, False))
@pytest.mark.parametrize("y_axis", (True, False))
@pytest.mark.parametrize("x_label", (True, False))
@pytest.mark.parametrize("y_label", (True, False))
@pytest.mark.parametrize("tree_height_scale", ("rank", "time", "log_time"))
@pytest.mark.parametrize("y_ticks", ([], [0, 1], None))
@pytest.mark.parametrize("y_gridlines", (True, False))
def test_draw_svg_parameter_combos(...)
```

Is that excessive @jeromekelleher ? It still runs in reasonable time on my machine.

python/tskit/drawing.py
```diff
@@ -117,6 +122,20 @@ def check_x_scale(x_scale):
...
+    If y_ticks is iterable, it is a list of tick positions. Otherwise return a default
+    or None if falsey
+    """
+    if ticks is None:
+        return default_iterable
+    try:
+        iter(ticks)
+        return ticks
+    except TypeError:
+        raise NotImplementedError("Autocalculated tick mark locations not implemented.")
```
jeromekelleher, line 136: This is still the wrong error class and message. Just return ``ticks`` and let the standard errors do the work.

jeromekelleher:
> Is that excessive @jeromekelleher ? It still runs in reasonable time on my machine.

... yes. How many tests is this? We don't need the cross product of all these things I think, maybe just a few specific combos where the parameters interact.

python/tskit/drawing.py
hyanwong: Weird. I did change that, and it's not what's now on my branch: https://github.com/hyanwong/tskit/blob/2382583b1f72de16646637aec9255c5bb0c01744/python/tskit/drawing.py#L125. Not sure what's up here @jeromekelleher

hyanwong:
> > Is that excessive @jeromekelleher ? It still runs in reasonable time on my machine.
> 
> ... yes. How many tests is this? We don't need the cross product of all these things I think, maybe just a few specific combos where the parameters interact.

2^5 * 3^2 = 288. Yes, I agree it seems excessive.

hyanwong:
I removed the x_axis/label specs and log_time, and now it's only 32 combos, which I deem acceptable. Also the re-push seems to have properly updated the PR, so now the NotImplementedError is gone. Note however, that I think the docs say that `NotImplementedError` need not *just* be used for base classes lacking a method, but can also be used "while the class is being developed to indicate that the real implementation still needs to be added." (which I think covers my use case).

I guess this can be merged - do you want me to squash the last commit into the previous one @jeromekelleher ?

jeromekelleher (approved):
LGTM!

PR #1238, opened 2021-03-10
Remap individuals after union

### LINKED ISSUE #1225: union needs to remap individual parent IDs

benjeffery:
Fixes #1225.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -9975,6 +9975,7 @@ tsk_table_collection_union(tsk_table_collection_t *self,
     int ret = 0;
     tsk_id_t k, i, new_parent, new_child;
     tsk_size_t num_shared_nodes = 0;
+    tsk_size_t num_individuals_self = self->individuals.num_rows;
     tsk_edge_t edge;
     tsk_mutation_t mut;
     tsk_site_t site;
@@ -10027,6 +10028,18 @@ tsk_table_collection_union(tsk_table_collection_t *self,
     memset(population_map, 0xff, other->populations.num_rows * sizeof(*population_map));
     memset(site_map, 0xff, other->sites.num_rows * sizeof(*site_map));
 
+    /* We have to map the individuals who are linked to nodes in the intersection first
+       as otherwise an individual linked to one node in the intersection and one in
+       `other` would be duplicated. We assume that the individual in `self` takes
+       priority.
+     */
+    for (k = 0; k < (tsk_id_t) other->nodes.num_rows; k++) {
+        if (other_node_mapping[k] != TSK_NULL
+            && other->nodes.individual[k] != TSK_NULL) {
+            individual_map[other->nodes.individual[k]]
+                = self->nodes.individual[other_node_mapping[k]];
+        }
+    }
     // nodes, individuals, populations
     for (k = 0; k < (tsk_id_t) other->nodes.num_rows; k++) {
         if (other_node_mapping[k] != TSK_NULL) {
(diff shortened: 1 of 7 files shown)


c/tskit/tables.c
```diff
@@ -10026,7 +10026,20 @@ tsk_table_collection_union(tsk_table_collection_t *self,
     memset(individual_map, 0xff, other->individuals.num_rows * sizeof(*individual_map));
     memset(population_map, 0xff, other->populations.num_rows * sizeof(*population_map));
     memset(site_map, 0xff, other->sites.num_rows * sizeof(*site_map));
-
+    for (k = 0; k < (tsk_id_t) other->individuals.num_rows; k++) {
+        individual_map[k] = TSK_NULL;
```
petrelharp, line 10030: isn't this taken care of by the `memset` above?

benjeffery: Ah of course!

benjeffery:
Just realised this isn't right as we should only remap the new individuals.

jeromekelleher:
Maybe a few test cases in Python would help clarify what is required here?

petrelharp:
Oh, right - I guess the individual parent remapping needs to happen during [add_and_remap_node](https://github.com/tskit-dev/tskit/pull/1238/files#diff-a7113e19292b94afe45f78ce230588105799e862a5f6f15d1af1ba81816f6e76R10048). Oh, wait, that won't work, because we don't know if we'll hit parent nodes before children.  Gah, this stuff is terrible.

Maybe the easiest thing is to separate out building the node table from building the individual table? First loop over nodes and build the individual map but don't actually add the rows to the individual table; then go through afterwards and add in the individuals?

benjeffery:
I think I have it now - remap after all the `add_and_remap_node` calls, but only loop over the new individuals. Should have a PR with tests tonight.

benjeffery:
Testing this is making me think `wf_sim` should always default to `record_individuals=True`. I'm going through the test failures from that now.

benjeffery:
I've just pushed a commit that does this, and it required changing tests to check the metadata of individuals. Seems a good change.

jeromekelleher (approved):

benjeffery:
@hyanwong getting an svg error here. Is it possible the svg isn't deterministic?

hyanwong:
> @hyanwong getting an svg error here. Is it possible the svg isn't deterministic?

What's the error? I can't imagine how SVG creation would fail to be deterministic, although I guess it could depend on something unexpected such as e.g. the node order. I'll double-check that the test suite uses random seeds for all the generated tree.I had some SVG changes when switching to the new msprime version, so of which seemed like differences in floating point rounding (although I have no idea why that might happen in this case)

benjeffery:
Ahhh, the nodes have individual classes! And as this PR adds individuals by default to wf sim the svg has changed. False alarm!

hyanwong:
Phew. If the sim has changed, you can overwrite the old SVG file like this:

```
pytest tests/test_drawing.py::TestDrawSvg::test_known_svg_ts_multiroot --overwrite-expected-visualizations
```

benjeffery:
Yeah, that flag is awesome!

PR #1246, opened 2021-03-16
"matrix multiplication" statistic

petrelharp:
Here's a draft of the C code to compute covariances of aribtrary weighted sums of genotypes. Borrowing from @brieuclehmann's code, this should always be true:
```
def genetic_relatedness_matrix(ts, sample_sets, mode):
    n = len(sample_sets)
    indexes = [(n1, n2) for n1, n2 in itertools.combinations_with_replacement(range(n), 2)]
    K = np.zeros((n,n))
    K[np.triu_indices(n)] = ts.genetic_relatedness(sample_sets, indexes, mode = mode, proportion=False, span_normalise=False)
    K = K + np.triu(K,1).transpose()
    return K

ts = msprime.simulate(10, length=1e6, recombination_rate=1e-8, Ne=1e4, mutation_rate=1e-8)
W = np.random.normal(size=2 * ts.num_samples).reshape((ts.num_samples, 2))
K = genetic_relatedness_matrix(ts, [[n] for n in ts.samples()], mode='site')
A = W.T @ K @ W
B = ts.genetic_relatedness_weighted(W, indexes=[(0,0), (0, 1), (1, 0), (1,1)], mode='site', span_normalise=False).reshape((2,2))
assert np.allclose(A, B)
```
(... and, it *is*, happy day!)

TODO: input checking, and tests. Tests should be straightforward, as we can just check for the property above, but hooking it in to the testing code will take some doing.

Also TODO:
- is this a good name? a good python API? I also thought about just making `weights` an argument to `genetic_relatedness`, which would call one method or the other depending on whether `sample_sets` or `weights` were present. This could be a general pattern, even.
- maybe there's some tidying to do at the C level, in terms of refactoring the various general stats computing helpers? but this could be put off.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2071,6 +2071,11 @@ typedef struct {
     const tsk_id_t *set_indexes;
 } sample_count_stat_params_t;
 
+typedef struct {
+    double *total_weights;
+    const tsk_id_t *index_tuples;
+} indexed_weight_stat_params_t;
+
 static int
 tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
     const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
@@ -3025,7 +3030,82 @@ out:
     return ret;
 }
 
-static int
+genetic_relatedness_weighted_summary_func(size_t state_dim, const double *state,
+    size_t result_dim, double *result, void *params)
+{
+    indexed_weight_stat_params_t args = *(indexed_weight_stat_params_t *) params;
+    const double *x = state;
+    tsk_id_t i, j;
+    size_t k;
+    double meanx, ni, nj;
+
+    meanx = state[state_dim - 1] / args.total_weights[state_dim - 1];
+    ;
+    for (k = 0; k < result_dim; k++) {
+        i = args.index_tuples[2 * k];
+        j = args.index_tuples[2 * k + 1];
+        ni = args.total_weights[i];
+        nj = args.total_weights[j];
+        result[k] = (x[i] - ni * meanx) * (x[j] - nj * meanx) / 2;
+    }
+    return 0;
+}
+
+int
+tsk_treeseq_genetic_relatedness_weighted(const tsk_treeseq_t *self,
+    tsk_size_t num_weights, const double *weights, tsk_size_t num_index_tuples,
+    const tsk_id_t *index_tuples, tsk_size_t num_windows, const double *windows,
+    double *result, tsk_flags_t options)
+{
+    int ret = 0;
+    tsk_size_t num_samples = self->num_samples;
+    size_t j, k;
+    indexed_weight_stat_params_t args;
+    const double *row;
+    double *new_row;
+    double *total_weights = malloc((num_weights + 1) * sizeof(*total_weights));
+    double *new_weights = malloc((num_weights + 1) * num_samples * sizeof(*new_weights));
+
+    if (total_weights == NULL || new_weights == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    // add a column of ones to W
+    for (k = 0; k < num_samples; k++) {
+        row = GET_2D_ROW(weights, num_weights, k);
+        new_row = GET_2D_ROW(new_weights, num_weights + 1, k);
+        for (j = 0; j < num_weights; j++) {
+            new_row[j] = row[j];
+        }
+        new_row[num_weights] = 1.0;
+    }
+
+    /* TODO: sanity check indexes */
+
+    for (j = 0; j < num_samples; j++) {
+        row = GET_2D_ROW(new_weights, num_weights + 1, j);
+        for (k = 0; k < num_weights + 1; k++) {
+            total_weights[k] += row[k];
+        }
+    }
+
+    args.total_weights = total_weights;
+    args.index_tuples = index_tuples;
+
+    ret = tsk_treeseq_general_stat(self, num_weights + 1, new_weights, num_index_tuples,
+        genetic_relatedness_weighted_summary_func, &args, num_windows, windows, result,
+        options);
+    if (ret != 0) {
+        goto out;
+    }
+
+out:
+    tsk_safe_free(total_weights);
+    tsk_safe_free(new_weights);
+    return ret;
+}
+
 Y2_summary_func(tsk_size_t TSK_UNUSED(state_dim), const double *state,
     tsk_size_t result_dim, double *result, void *params)
 {
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -3030,17 +3030,17 @@ out:
     return ret;
 }
 
-genetic_relatedness_weighted_summary_func(size_t state_dim, const double *state,
-    size_t result_dim, double *result, void *params)
+static int
+genetic_relatedness_weighted_summary_func(tsk_size_t state_dim, const double *state,
+    tsk_size_t result_dim, double *result, void *params)
 {
     indexed_weight_stat_params_t args = *(indexed_weight_stat_params_t *) params;
     const double *x = state;
     tsk_id_t i, j;
-    size_t k;
+    tsk_size_t k;
     double meanx, ni, nj;
 
     meanx = state[state_dim - 1] / args.total_weights[state_dim - 1];
-    ;
     for (k = 0; k < result_dim; k++) {
         i = args.index_tuples[2 * k];
         j = args.index_tuples[2 * k + 1];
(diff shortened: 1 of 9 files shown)


jeromekelleher:
Very cool!

> is this a good name? a good python API? I also thought about just making weights an argument to genetic_relatedness, which would call one method or the other depending on whether sample_sets or weights were present. This could be a general pattern, even.

I can't think of a better one, but I'd be +1 on adding a ``weights`` argument to genetic_relatedness. We'd need a different name though - is this the same "weight" as the W arg in the ``general_stat`` method? I can see this being confusing.

c/tskit/trees.c
```diff
@@ -2799,6 +2804,83 @@ tsk_treeseq_genetic_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample
...
+    size_t result_dim, double *result, void *params)
+{
+    indexed_weight_stat_params_t args = *(indexed_weight_stat_params_t *) params;
+    const double *x = state;
+    tsk_id_t i, j;
+    size_t k;
+    double meanx, ni, nj;
+
+    meanx = state[state_dim - 1] / args.total_weights[state_dim - 1];
+    ;
```
jeromekelleher, line 2818: stray ;

c/tskit/trees.c
```diff
@@ -2799,6 +2804,83 @@ tsk_treeseq_genetic_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample
...
+    double *total_weights = malloc((num_weights + 1) * sizeof(*total_weights));
+    double *new_weights = malloc((num_weights + 1) * num_samples * sizeof(*new_weights));
+
+    if (total_weights == NULL || new_weights == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    // add a column of ones to W
+    for (k = 0; k < num_samples; k++) {
```
jeromekelleher, line 2850: Nice to be consistent with j or k being the sample index. It's the opposite in the next code chunk.

jeromekelleher (commented):
Looks great!

I think we should add a ``genetic_relatedness_matrix`` method as well as a convenience, but you probably had that in mind too?

The code all looks basically right to me, although I have scrutinised the detail. My only question really is about the "weight" term and whether this will be confusing. If it's the same W as in general stat (which I think it is?) then we're all good.

petrelharp:
>  is this the same "weight" as the W arg in the general_stat method?

It is the same. So, maybe the argument should be called `W`.

> I think we should add a genetic_relatedness_matrix method as well as a convenience, but you probably had that in mind too?

Yeah, I guess? We still haven't gotten around to writing the `divergence_matrix` method either.

brieuclehmann:
Just been playing around with this and it seems like there might be some numerical stability issues. I tried @petrelharp 's code above with different seeds and occasionally get an AssertionError. Actually, the SUPER WEIRD thing is that sometimes it works with the same seed and sometimes it doesn't... i.e. I run exactly the same code twice and the first time it will be fine but the second time returns, e.g., B = [inf, inf ; inf, inf]. Has anyone seen anything like this before? This is on my MacBook btw.

```
def genetic_relatedness_matrix(ts, sample_sets, mode):
    n = len(sample_sets)
    indexes = [(n1, n2) for n1, n2 in itertools.combinations_with_replacement(range(n), 2)]
    K = np.zeros((n,n))
    K[np.triu_indices(n)] = ts.genetic_relatedness(sample_sets, indexes, mode = mode, proportion=False, span_normalise=False)
    K = K + np.triu(K,1).transpose()
    return K

seed = 42
ts = msprime.simulate(10, length=1e6, recombination_rate=1e-8, Ne=1e4, mutation_rate=1e-8, random_seed=seed)
np.random.seed(seed)
W = np.random.normal(size=2 * ts.num_samples).reshape((ts.num_samples, 2))
K = genetic_relatedness_matrix(ts, [[n] for n in ts.samples()], mode='site')
A = W.T @ K @ W
B = ts.genetic_relatedness_weighted(W, indexes=[(0,0), (0, 1), (1, 0), (1,1)], mode='site', span_normalise=False).reshape((2,2))
assert np.allclose(A, B)
```

brieuclehmann:
OK, still some bizarre behaviour on rescomp. See the following code to perform matrix-vector multiplication using the same tree sequence as above. I am using the old code for comparison as `mat_mul` then `mat_mul_stat` for the call to `genetic_relatedness_weighted`. On some occasions it returns the right output, on others not:

```
n_ind = int(ts.num_samples / 2)
sample_sets = [(2 * i, (2 * i) + 1) for i in range(n_ind)]
W_samples = np.array([[float(u in A) for A in sample_sets] for u in ts.samples()])
indexes = [(i, n_ind) for i in range(len(sample_sets))]
n = np.array([len(x) for x in sample_sets])
n_total = sum(n)

def mat_mul_stat(a):
    W = np.c_[W_samples, W_samples @ a]
    return ts.genetic_relatedness_weighted(
        W, indexes=indexes, mode="site", span_normalise=False
    )

def mat_mul(a):
    W = np.c_[W_samples, W_samples @ a]
    wts = np.sum(W, axis=0)
    # summary function
    def f(x):
        mx = np.sum(x[0:n_ind]) / n_total
        return np.array(
            [(x[i] - wts[i] * mx) * (x[j] - wts[j] * mx) / 2 for i, j in indexes]
        )
    return ts.general_stat(W, f, len(indexes), span_normalise=False)

np.random.seed(seed)
a = np.random.randn(n_ind)
K = genetic_relatedness_matrix(ts, sample_sets, "site")
print(mat_mul_stat(a))
print(K @ a)
print(mat_mul(a))
print(mat_mul_stat(a))
print(mat_mul_stat(a))
```

Returns:
```
[  73.45122395 -205.96299977   53.02061499  435.5723698  -352.63344965]
[  73.45122395 -205.96299977   53.02061499  432.12461049 -352.63344965]
[  73.45122395 -205.96299977   53.02061499  432.12461049 -352.63344965]
[  41.41281634 3375.49081761 3669.81766139 4035.61057064 3232.49239152]
[-842.64551716  690.48841404  967.14364331 1339.59209569  545.65397605]
```

petrelharp:
Yep, I'm seeing that too. Bizarrely, only in an interactive shell,not at the command line.  It must be that I've done something bad with memory management. I'm not sure how to track that down - can we run valgrind on the python tests?

brieuclehmann:
I'm also getting the problem when I run the above code in the command line via a script... Never used valgrind before but could look into it if that would help?

jeromekelleher:
> I'm not sure how to track that down - can we run valgrind on the python tests?

No, Python gets in the way. We'd have to reproduce the error as a C test case to be able to use valgrind.

I have a bit of time in the morning, I could take a look if someone sends me a script to reproduce?

petrelharp:
```
#!/usr/bin/python3
import msprime, tskit
import numpy as np
import itertools

def genetic_relatedness_matrix(ts, sample_sets, mode):
    n = len(sample_sets)
    indexes = [(n1, n2) for n1, n2 in itertools.combinations_with_replacement(range(n), 2)]
    K = np.zeros((n,n))
    K[np.triu_indices(n)] = ts.genetic_relatedness(sample_sets, indexes, mode = mode, proportion=False, span_normalise=False)
    K = K + np.triu(K,1).transpose()
    return K


ts = msprime.simulate(10, length=1e6, recombination_rate=1e-8, Ne=1e4, mutation_rate=1e-8, random_seed=42)

for seed in range(10):
    np.random.seed(seed)
    W = np.random.normal(size=2 * ts.num_samples).reshape((ts.num_samples, 2))
    K = genetic_relatedness_matrix(ts, [[n] for n in ts.samples()], mode='site')
    A = W.T @ K @ W
    B = ts.genetic_relatedness_weighted(W, indexes=[(0,0), (0, 1), (1, 0), (1,1)], mode='site', span_normalise=False).reshape((2,2))
    print('A:', A)
    print('B:', B)
    assert np.allclose(A, B)
```

jeromekelleher:
Thanks @petrelharp, this is a weird one. Reliably fails on the 10th seed, but not if we just run that seed on its own. I'll poke around, but it nearly has to be a memory management error.

jeromekelleher:
I think it's worthwhile putting in some tests to cover the remaining code paths now - this will have to be done later anyway, and it might unearth the problem we're having here. Stuff covering error paths in the Python C stuff is particularly important to exercise, subtle problems can manifest themselves in all sorts of unexpected ways and lots of time can be lost tracking them down.

petrelharp:
That sounds good. I'm pretty swamped right now, might you be able to have a go at the tests, @brieuclehmann?

brieuclehmann:
Yep, I'd be happy to! Please could you point me towards a relevant function/method that I could use as a template?

jeromekelleher:
Thanks @brieuclehmann but it's probably better if I put in a commit to test out the code paths I'm thinking about - the testing code is quite convoluted for this stuff, and it would be a good way for me to start getting up to speed on what the function is doing anyway. I'll have a go as soon as I can (hopefully tomorrow AM)

jeromekelleher:
I think that should do it - I think the issue was that the ``total_weights`` array was being malloced rather than calloced (and therefore never being initialised to 0). Most of the time the memory happened to be zero already so it worked.

c/tskit/trees.c
```diff
@@ -2799,6 +2804,77 @@ tsk_treeseq_genetic_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample
...
+        new_row = GET_2D_ROW(new_weights, num_weights + 1, j);
+        for (k = 0; k < num_weights; k++) {
+            new_row[k] = row[k];
+            total_weights[k] += row[k];
+        }
+        new_row[num_weights] = 1.0;
+    }
+    total_weights[num_weights] = num_samples;
+
+    /* TODO: sanity check indexes */
```
jeromekelleher, line 2860: Not sure what this TODO is for? Shouldn't tsk_treeseq_general_stat check the indexe_tuples?

petrelharp: I guess that's what it's for? Good point.

petrelharp:
> I think the issue was that the total_weights array was being malloced rather than calloced (and therefore never being initialised to 0)

Doh! Good catch.

jeromekelleher:
I think we should try to get this merged @petrelharp - do you know what's needed to get it in? We don't need to have it fully documented, but it would be good to get it in to main so we can experiment.

brieuclehmann:
Ping @petrelharp - I'd be happy to write some tests and more documentation in the coming days so that we can get this merged, but I think I'm not able to rebase the branch myself?

petrelharp:
Done!

jeromekelleher:
FYI I just pushed an update fixing the build problems.

brieuclehmann:
Would it be possible to rebase this onto main? I'd like to use `genetic_relatedness_weighted` and `decapitate` in the same script, i.e. to do PCA on a decapitated tree sequence.

(Sorry for dropping the ball on writing the tests & documentation for this - I will get round to it v soon!)

petrelharp:
Uh-oh, a segfault - ping @jeromekelleher?

jeromekelleher:
Looks like a simple compile error because of API changes we made for C 1.0. Should be fixed.

jeromekelleher:
Closing this in favour of #2785 (which includes these commits)

PR #1254, opened 2021-03-18
"Append row" for table classes

### LINKED ISSUE #1111: Need a way to add entire rows to tables.

benjeffery:
Closes #1111

### DIFF
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -158,13 +158,7 @@ class Simplifier:
         flags &= ~tskit.NODE_IS_SAMPLE
         if is_sample:
             flags |= tskit.NODE_IS_SAMPLE
-        output_id = self.tables.nodes.add_row(
-            flags=flags,
-            time=node.time,
-            population=node.population,
-            metadata=node.metadata,
-            individual=node.individual,
-        )
+        output_id = self.tables.nodes.append(node.replace(flags=flags))
         self.node_id_map[input_id] = output_id
         return output_id
 
@@ -186,9 +180,7 @@ class Simplifier:
         num_edges = 0
         for child in sorted(self.edge_buffer.keys()):
             for edge in self.edge_buffer[child]:
-                self.tables.edges.add_row(
-                    edge.left, edge.right, edge.parent, edge.child
-                )
+                self.tables.edges.append(edge)
                 num_edges += 1
         self.edge_buffer.clear()
         return num_edges
@@ -413,19 +405,14 @@ class Simplifier:
                         mapped_parent = -1
                         if mut.parent != -1:
                             mapped_parent = mutation_id_map[mut.parent]
-                        self.tables.mutations.add_row(
-                            site=len(self.tables.sites),
-                            node=self.mutation_node_map[mut.id],
-                            time=mut.time,
-                            parent=mapped_parent,
-                            derived_state=mut.derived_state,
-                            metadata=mut.metadata,
+                        self.tables.mutations.append(
+                            mut.replace(
+                                site=len(self.tables.sites),
+                                node=self.mutation_node_map[mut.id],
+                                parent=mapped_parent,
+                            )
                         )
-                self.tables.sites.add_row(
-                    position=site.position,
-                    ancestral_state=site.ancestral_state,
-                    metadata=site.metadata,
-                )
+                self.tables.sites.append(site)
 
     def finalise_references(self):
         input_populations = self.ts.tables.populations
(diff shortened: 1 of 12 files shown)


jeromekelleher:
> I'm not happy with this as-is, as it using attr.asdict won't support duck typed objects with the right properties.

There's a lot to be said for the way you're doing it now - it's less code and it's robust to adding new columns. If this works for both (.e.g) NodeTableRow and Node, then I'm happy. Are both of these attrs classes at the moment?

In the long run, we can say "any class that we can call ``dataclasses.asdict`` on which has at least the required fields" will work.

benjeffery:
I think I'll do the dataclasses migration now as a separate PR.

benjeffery:
I'm just dealing with the container classes. Seeing if they should be dataclasses. Will need tweaks as they have extra things like `individual.node` in.

benjeffery:
One other question is whether to take encoded or unencoded metadata when appending a container like `Node`. Taking encoded metadata would be more performant - but if the user was adding rows to a table with a different, incompatible schema then they are silently adding bad metadata. I lean towards taking the encoded metadata with a warning in the docs.

jeromekelleher:
I think we could be more flexible about this, and shouldn't worry too much about performance (this is row-by-row in Python, after all).

As a high-level requirement, something like
```python
for node in other_tables.nodes():
     tables.nodes.append(node)
```
```python
for node in ts.nodes():
     tables.nodes.append(node)
```

should always work, under the various permutations of schema present/not present in either the source or destination tables. We could perhaps emit warnings, if we're writing out without a schema, or something?

benjeffery:
Yes, def agree that those should work, current tests look like that. In the case where the destination table has no schema we can just take the bytes from the source object. The question is, when the receiving table has a schema, do we decode, validate against the schema and then encode?

jeromekelleher:
> I guess the answer is yes, as otherwise the user won't find out till that row fails to decode at what could be a much later date.

Yes, I think so. The guiding principle here is ease of use and developer friendliness; performance is very much secondary.

benjeffery:
So currently the table row classes do not decode metadata, so we couldn't check it. I can't remember why they don't, maybe as the table API is more low-level.

All this is making me wonder why we have different classes for trees and tables, might be simpler to merge the two. Lets talk about it later?

benjeffery:
This is now stacked on #1261 and is modified to take any class with the right attributes.

python/tests/test_highlevel.py
```diff
@@ -3170,6 +3164,30 @@ def get_instances(self, n):
...
+            "migrations",
+            "sites",
+            "mutations",
+            "populations",
+            "provenances",
+        ]:
+            table = getattr(tables, table_name)
+            for i in range(len(getattr(ts_fixture.tables, table_name))):
+                table.append(getattr(ts_fixture, table_name[:-1])(i))
+        print(ts_fixture.tables, tables)
```
jeromekelleher, line 3187: print

python/tests/simplify.py
```diff
@@ -158,13 +159,7 @@ def record_node(self, input_id, is_sample=False):
...
         if is_sample:
             flags |= tskit.NODE_IS_SAMPLE
-        output_id = self.tables.nodes.add_row(
-            flags=flags,
-            time=node.time,
-            population=node.population,
-            metadata=node.metadata,
-            individual=node.individual,
-        )
+        output_id = self.tables.nodes.append(replace(node, flags=flags))
```
jeromekelleher, line 162: I have to say I'm not a big fan of this pattern. It seems awkward to me to have to know that the rows are dataclasses, and then you need to call this function ``dataclasses.replace`` with the stuff you want to change. It's assuming that users are well versed in dataclasses.

What if we have a method
```python
def replace(self, **kwargs):
     return dataclasses.replace(self, **kwargs)
```
on the row classes so that this line would look like
```python
output_id = self.tables.nodes.append(node.replace(flags=flags))
```

at least then the recommended pattern is discoverable and we can document it locally.

jeromekelleher (commented):
Looks good - see comment about the common usage pattern for modified rows

python/tests/simplify.py
benjeffery: Yeah, I see what you mean. I'm not sure why those methods aren't on dataclasses in the first place. I've added a method by inheritance of a class in `util`

python/tests/test_highlevel.py
benjeffery: Fixed

jeromekelleher (approved):
LGTM!

PR #1257, opened 2021-03-18
Allow adding rows with null metadata

### LINKED ISSUE #1256: tables do not allow setting of metadata=None even if 'null' is allowed for metadata
We made it so that we can allow empty metadata by setting `type = ['object', 'null']` in the metadata schema. If so, then
```
schema.validate_and_encode_row(None)
```
returns `b''`. However, you *can't* actually do
```
table.add_row(metadata=None)
```
because you get the error
```
E           tskit.exceptions.MetadataValidationError:
... 
E           
E           On instance:
E               {}

```

The culprit is right here: https://github.com/tskit-dev/tskit/blob/main/python/tskit/metadata.py#L571
because `metadata=None` gets turned into `{}` here: https://github.com/tskit-dev/tskit/blob/main/python/tskit/tables.py#L549

A test and possible solution to follow.

petrelharp:
Closes #1256 although probably not in the best possible way.

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -439,11 +439,19 @@ class TestMetadataModule:
             with pytest.raises(exceptions.MetadataSchemaValidationError):
                 metadata.MetadataSchema(schema)
 
-    def test_null_union_top_level(self):
+    @pytest.mark.parametrize("codec", ["struct", "json"])
+    def test_null_union_top_level(self, codec):
         schema = {
-            "codec": "json",
+            "codec": f"{codec}",
             "type": ["object", "null"],
-            "properties": {"one": {"type": "string"}, "two": {"type": "number"}},
+            "properties": {
+                "one": {
+                    "type": "string",
+                    "binaryFormat": "1024s",
+                    "nullTerminated": True,
+                },
+                "two": {"type": "number", "binaryFormat": "i"},
+            },
         }
         ms = metadata.MetadataSchema(schema)
         row_data = {"one": "tree", "two": 5}
@@ -1025,6 +1033,25 @@ class TestStructCodec:
             "float": 5.5,
             "int": 42,
         }
+
+    def test_defaults_object_or_null(self):
+        schema = {
+            "codec": "struct",
+            "type": ["object", "null"],
+            "properties": {
+                "int": {"type": "number", "binaryFormat": "b", "default": 42},
+                "float": {"type": "number", "binaryFormat": "d"},
+            },
+        }
+        ms = metadata.MetadataSchema(schema)
+        row_data = {"float": 5.5}
+        assert ms.validate_and_encode_row(row_data) == b"\x00\x00\x00\x00\x00\x00\x16@*"
+        assert ms.decode_row(ms.validate_and_encode_row(row_data)) == {
+            "float": 5.5,
+            "int": 42,
+        }
+        assert ms.validate_and_encode_row(None) == b""
+        assert ms.decode_row(b"") is None
 
 
 class TestStructCodecRoundTrip:
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -851,6 +851,105 @@ class MetadataTestsMixin:
                 assert table[j].metadata == metadata_column[j]
                 assert table[j + num_rows].metadata == metadata_column[j]
 
+    @pytest.mark.parametrize(
+        "codec",
+        ["struct", "json"],
+    )
+    def test_set_null_metadata(self, codec):
+        table = self.table_class()
+        table.metadata_schema = metadata.MetadataSchema(
+            {
+                "codec": f"{codec}",
+                "title": "Example Metadata",
+                "type": ["object", "null"],
+                "properties": {
+                    "a": {"type": "number", "binaryFormat": "i"},
+                },
+                "required": ["a"],
+                "additionalProperties": False,
+            },
+        )
+        examples = [{"a": 4}, None]
+        for md in examples:
+            table.add_row(
+                **{
+                    **self.input_data_for_add_row(),
+                    "metadata": md,
+                }
+            )
+        assert table.num_rows == len(examples)
+        for md, row in zip(examples, table):
+            assert md == row.metadata
+
+    # only json allows leaving out of optional entries
+    def test_set_empty_metadata_json(self):
+        table = self.table_class()
+        table.metadata_schema = metadata.MetadataSchema(
+            {
+                "codec": "json",
+                "title": "Example Metadata",
+                "type": ["object", "null"],
+                "properties": {
+                    "a": {"type": "number", "binaryFormat": "i"},
+                },
+                "required": [],
+                "additionalProperties": False,
+            },
+        )
+        examples = [{"a": 4}, {}]
+        for md in examples:
+            table.add_row(
+                **{
+                    **self.input_data_for_add_row(),
+                    "metadata": md,
+                }
+            )
+        assert table.num_rows == len(examples)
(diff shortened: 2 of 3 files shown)


petrelharp:
*Edit:* This is *definitely* not the best way to do it.

jeromekelleher (commented):
LGTM, but I'll defer to @benjeffery

python/tests/test_metadata.py
```diff
@@ -440,15 +440,23 @@ def test_bad_top_level_type(self):
...
-        schema = {
-            "codec": "json",
-            "type": ["object", "null"],
-            "properties": {"one": {"type": "string"}, "two": {"type": "number"}},
-        }
-        ms = metadata.MetadataSchema(schema)
-        row_data = {"one": "tree", "two": 5}
-        assert ms.decode_row(ms.validate_and_encode_row(row_data)) == row_data
-        assert ms.decode_row(ms.validate_and_encode_row(None)) is None
+        for codec in ["json", "struct"]:
```
petrelharp, line 443: This was a test I added at first before I figured out where the problem was, and figured it's a good one, so I'd leave it in.

python/tests/test_tables.py
```diff
@@ -851,6 +851,32 @@ def test_round_trip_set_columns(self):
...
+                "title": "Example Metadata",
+                "type": ["object", "null"],
+                "properties": {
+                    "a": {"type": "number", "binaryFormat": "i"},
+                },
+                "required": ["a"],
+                "additionalProperties": False,
+            },
+        )
+        metadata_list = [{"a": 4}, None, {"a": 5}]
```
benjeffery, line 868: Might as well add `{}` here too to check it doesn't become `None` with future fiddling.

benjeffery (approved):
What misgivings did you have? This seems like the right way to fix this!
One test tweak suggestion.

python/tests/test_tables.py
petrelharp: That won't validate on this schema, but I'll add another test.

python/tests/test_tables.py
```diff
@@ -851,6 +851,61 @@ def test_round_trip_set_columns(self):
...
+                        **self.input_data_for_add_row(),
+                        "metadata": md,
+                    }
+                )
+            assert table.num_rows == len(metadata_list)
+            for md, row in zip(metadata_list, table):
+                assert md == row.metadata
+
+    @pytest.mark.skip("Fails, but shouldn't?")
+    def test_set_with_optional_properties(self):
```
petrelharp, line 882: > Might as well add {} here too to check it doesn't become None with future fiddling.

Good suggestion, @benjeffery! This fails for totally different reasons on both codecs.

petrelharp: With `struct` it fails with:
```
tskit/metadata.py:452: in <lambda>
    else b"".join(
>           sub_encoder(obj[key]) for key, sub_encoder in sub_encoders.items()
        )
    )
E   KeyError: 'a'
```
and with `json` it fails by turning the None into `{}`:
```
>               assert md == row.metadata
E               assert None == {}
E                +  where {} = IndividualTableRow(flags=3, location=array([], dtype=float64), parents=array([], dtype=int32), metadata={}).metadata
```

python/tests/test_tables.py
```diff
@@ -851,6 +851,67 @@ def test_round_trip_set_columns(self):
...
+                table.add_row(
+                    **{
+                        **self.input_data_for_add_row(),
+                        "metadata": md,
+                    }
+                )
+            assert table.num_rows == len(metadata_list)
+            for md, row in zip(metadata_list, table):
+                if isinstance(md, dict):
+                    # struct inserts defaults, json does not
```
petrelharp, line 908: This was surprising. Does the JSON codec do anything with defaults?

python/tskit/metadata.py
```diff
@@ -446,13 +446,23 @@ def make_object_or_null_encode(cls, sub_schema):
             key: StructCodec.make_encode(prop)
             for key, prop in sub_schema["properties"].items()
         }
-        return (
-            lambda obj: b""
-            if obj is None
-            else b"".join(
-                sub_encoder(obj[key]) for key, sub_encoder in sub_encoders.items()
-            )
-        )
```
petrelharp, line 455: This was a bug, I think - `make_object_encode` was using defaults, but `make_object_or_null_encode` was not, leading to a `KeyError` when trying to encode objects without all the required properties.

python/tskit/metadata.py
```diff
@@ -568,7 +578,10 @@ def __init__(self, schema: Optional[Mapping[str, Any]]) -> None:
             self._validate_row = TSKITMetadataSchemaValidator(schema).validate
             self.encode_row = codec_instance.encode
             self.decode_row = codec_instance.decode
-            self.empty_value = {}
+            if "type" in schema and "null" in schema["type"]:
```
petrelharp, line 581: Before I had this only putting the empty value as None if there were required attributes. I think this is wrong, and should be this way - None is "empty" if allowed.

petrelharp:
I think this is ready to go now, although maybe "json doesn't use defaults" is a bug for another PR?

python/tests/test_tables.py
benjeffery: It's in work here: https://github.com/tskit-dev/tskit/pull/1149

python/tests/test_metadata.py
jeromekelleher: Good to pytest.mark.parametrize this rather than looping

python/tests/test_tables.py
```diff
@@ -851,6 +851,66 @@ def test_round_trip_set_columns(self):
                 assert table[j].metadata == metadata_column[j]
                 assert table[j + num_rows].metadata == metadata_column[j]
 
+    def test_set_null_metadata(self):
+        for codec in ["struct", "json"]:
```
jeromekelleher, line 855: Likewise here:
```
@pytest.mark.parametrize("codec", ["struct", "json"])
def test_set_null_metadata(self, codec):
     table = self.table_class() 
     # ect
```

It's a great pattern once you get used to using it!

jeromekelleher (commented):
LGTM @petrelharp, minor comments.

benjeffery:
@petrelharp I'm adding some commits here, I'll also address the comments.

benjeffery:
Ok, I've added a skip to the test that needs JSON default handling. If you're happy @petrelharp I'll squash and merge.

python/tests/test_tables.py
```diff
@@ -851,65 +851,75 @@ def test_round_trip_set_columns(self):
...
-                {
-                    "codec": f"{codec}",
-                    "title": "Example Metadata",
-                    "type": ["object", "null"],
-                    "properties": {
-                        "a": {"type": "number", "binaryFormat": "i"},
-                    },
-                    "required": ["a"],
-                    "additionalProperties": False,
+    # Properties are not optional for struct
```
petrelharp, line 854: Is this because the default for struct is to require properties, and not for json?

benjeffery: Yes, struct has no way to encode missingness in the current design.

petrelharp: Ah, never mind, I got it.

petrelharp:
I tweaked your tweak to separate out some of the things we're testing and make it so we test with an explicit "required" entry; go ahead and squash & merge if you like it? Or, tweak more?

benjeffery:
Thanks for clearing this up @petrelharp! Accommodating legacy metadata has lead to some wrinkles in the design, but I think we're getting there.

Issue #186, opened 2019-05-22
Access edge information from a tree


hyanwong:
I find myself wanting to inspect edges in a tree - in particular, to find out how far a given edge in a tree spans over the rest of the tree sequence. This seems pretty hard to get at at the moment, other than by iterating over the edges and finding the one with the correct parent/child nodes and at the correct genomic position. It would be nice to have a function called e.g. `tree.edge_above(u)` that returned details of the edge above node `u` in a tree.  Or indeed, a function that returns the edge index (`i`) above a node in tree, so that I can at least do `ts.edge(i)`.

I'm willing to have a go at implementing this, but I've had a look at the low level and C code, and it looks like this is annoyingly more complicated to do than I had at first thought. Any pointers or recommendations would be thus rather appreciated.

jeromekelleher:
Yes, this is going to need some ground-up infrastructure to support. A workaround for now would be to co-iterate over the trees and edge diffs, maintaining the edge pointers yourself? Something like
```python
edge_map = {}
for tree, (_, edges_out, edges_in) in zip(ts.trees(), ts.edge_diffs):
     for edge in edges_out:
          del edge_map[edge.child]
      for edge in edges_in:
           edge_map[edge.child] = edge
      # Now edge map should contain what you need
```

Would this do the trick? 

Might not be hassle of all the C code?

hyanwong:
Yes, this is exactly what I am doing at the moment. Not so useful for large tree sets, but fine for me. 

Poss not worth doing it in the c code if it is a lot of work. I opened the issue as I thought it might be more generally useful. But happy to leave it up to you to divide whether to keep it open.

jeromekelleher:
OK, let's close this for now as it would require a fair bit of low-level engineering to enable (essentially adding an extra option to the tree iterator). If someone else is interested in this feature, please comment below and we can look again.

petrelharp:
another option is to use numpy, eg
```
edges = ts.tables.edges
np.nonzero(np.logical_and(edges.parent == 20,
                                            edges.child == 0,
                                            edges.left <= 0.5,
                                            edges.right > 1.0))
```
although this won't be very efficient to do it for lots of edges.

hyanwong:
Implemented as a high-level, non-optimised private function `ts._tree_node_edges()` in #1259. This issue kept in case we want to implement a low-level version in C.

hyanwong:
[...]

PR #1259, opened 2021-03-19
Add _node_edges private functions

hyanwong:
## Description

Addresses #186 with a non-optimised Python approach, so that we can easily obtain the edges above a node in a tree. The general approach was just discussed with @jeromekelleher. This should simplify #1258 

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2209,6 +2209,18 @@ class Tree:
             for u in roots:
                 yield from iterator(u)
 
+    def _node_edges(self):
+        """
+        Return a numpy array mapping the node IDs in this tree to the ID of the edge
+        above them. This is in lieu of a tree.edge(u) function, currently implemented
+        using the non-optimised Python TreeSequence._tree_node_edges() generator
+        """
+        for index, node_edge_map in enumerate(  # pragma: no branch
+            self.tree_sequence._tree_node_edges()
+        ):
+            if index == self.index:
+                return node_edge_map
+
     # TODO make this a bit less embarrassing by using an iterative method.
     def __build_newick(self, *, node, precision, node_labels, include_branch_lengths):
         """
@@ -4535,6 +4547,20 @@ class TreeSequence:
             encoded_metadata=metadata,
             metadata_decoder=self.table_metadata_schemas.edge.decode_row,
         )
+
+    def _tree_node_edges(self):
+        """
+        Return a generator over the trees in the tree sequence, yielding a numpy array
+        that maps the node IDs in the tree to the ID of the edge above them.
+        Currently this is a private, non-optimised Python implementation.
+        """
+        node_edges = np.full(self.num_nodes, NULL, dtype=np.int32)
+        for _, edges_out, edges_in in self.edge_diffs():
+            for e in edges_out:
+                node_edges[e.child] = NULL
+            for e in edges_in:
+                node_edges[e.child] = e.id
+            yield node_edges
 
     def migration(self, id_):
         """
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1586,6 +1586,23 @@ class TestTreeSequence(HighLevelTestCase):
         assert t1.equals(t2)
         assert t2.equals(t1)
 
+    def test_tree_node_edges(self):
+        for ts in get_example_tree_sequences():
+            edge_visited = np.zeros(ts.num_edges, dtype=bool)
+            for mapping, tree in zip(ts._tree_node_edges(), ts.trees()):
+                node_mapped = mapping >= 0
+                edge_visited[mapping[node_mapped]] = True
+                assert np.sum(node_mapped) == len(list(tree.nodes())) - tree.num_roots
+                for u in tree.nodes():
+                    if tree.parent(u) == tskit.NULL:
+                        assert mapping[u] == tskit.NULL
+                    else:
+                        edge = ts.edge(mapping[u])
+                        assert edge.child == u
+                        assert edge.left <= tree.interval.left
+                        assert edge.right >= tree.interval.right
+            assert np.all(edge_visited)
+
 
 class TestTreeSequenceMethodSignatures:
     ts = msprime.simulate(10, random_seed=1234)
@@ -2780,6 +2797,30 @@ class TestTree(HighLevelTestCase):
             tskit.Tree(ts, sample_counts=False)
             assert len(w) == 1
             assert issubclass(w[0].category, RuntimeWarning)
+
+    def test_node_edges(self):
+        ts = msprime.simulate(5, recombination_rate=1, random_seed=42)
+        assert ts.num_trees > 2
+        edge_table = ts.tables.edges
+        for tree in ts.trees():
+            nodes = set(tree.nodes())
+            midpoint = sum(tree.interval) / 2
+            mapping = tree._node_edges()
+            for node, edge in enumerate(mapping):
+                if node in nodes and tree.parent(node) != tskit.NULL:
+                    edge_above_node = np.where(
+                        np.logical_and.reduce(
+                            (
+                                edge_table.child == node,
+                                edge_table.left < midpoint,
+                                edge_table.right > midpoint,
+                            )
+                        )
+                    )[0]
+                    assert len(edge_above_node) == 1
+                    assert edge_above_node[0] == edge
+                else:
+                    assert edge == tskit.NULL
 
 
 class TestNodeOrdering(HighLevelTestCase):


hyanwong:
Currently showing an exciting CI failure which is nothing to do with this PR:

```
____________________ TestTables.test_simplify_access_sites _____________________
[gw1] linux -- Python 3.7.10 /usr/local/bin/python

self = <tests.test_threads.TestTables object at 0x7f8dfb6cea90>

    def test_simplify_access_sites(self):
>       self.run_simplify_access_table("sites", "position")

tests/test_threads.py:266: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_threads.py:242: in run_simplify_access_table
    self.run_failing_reader(writer, reader)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.test_threads.TestTables object at 0x7f8dfb6cea90>
writer = <function TestTables.run_simplify_access_table.<locals>.writer at 0x7f8e0c854290>
reader = <function TestTables.run_simplify_access_table.<locals>.reader at 0x7f8e0c854b90>
num_readers = 32

    def run_failing_reader(self, writer, reader, num_readers=32):
        """
        Runs a test in which a single writer acceses some tables
        and a bunch of other threads try to read the data.
        """
        barrier = threading.Barrier(num_readers + 1)
    
        def writer_proxy():
            barrier.wait()
            writer()
    
        def reader_proxy(thread_index, results):
            barrier.wait()
            # Attempts to operate on a table while locked should raise a RuntimeError
            try:
                reader(thread_index, results)
                results[thread_index] = 0
            except RuntimeError:
                results[thread_index] = 1
    
        writer_thread = threading.Thread(target=writer_proxy)
        writer_thread.start()
        results = run_threads(reader_proxy, num_readers)
        writer_thread.join()
    
>       failures = sum(results)
E       TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'

tests/test_threads.py:206: TypeError
----------------------------- Captured stderr call -----------------------------
Exception in thread Thread-242:
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/circleci/tskit/python/tests/test_threads.py", line 196, in reader_proxy
    reader(thread_index, results)
  File "/home/circleci/tskit/python/tests/test_threads.py", line 240, in reader
    assert x.shape[0] == len(table)
AssertionError: assert 152 == 38
 +  where 38 = len(<tskit.tables.SiteTable object at 0x7f8dfa6cd390>)
```

python/tskit/trees.py
```diff
@@ -2209,6 +2209,15 @@ def nodes(self, root=None, order="preorder"):
...
                 yield from iterator(u)
 
+    def _node_edges(self):
+        """
+        Return a numpy array mapping the node IDs in this tree to the ID of the edge
+        above them. This is in lieu of a tree.edge(u) function, currently implemented
+        using the non-optimised Python TreeSequence._tree_node_edges() generator
+        """
+        ts = self.tree_sequence
+        return next(x for i, x in enumerate(ts._tree_node_edges()) if i == self.index)
```
jeromekelleher, line 2219: This means we always iterate over all trees, right? How about
```python
for index, node_edge_map in enumerate(self.tree_sequence._tree_node_edges()):
    if index == self.index:
          return node_edge_map
```
It's a bit more obvious as well

jeromekelleher: Ah, maybe it doesn't, next will pop the first thing off the iterator. Anyway, I shouldn't have to think this hard to figure out what this is doing, it should be immediately obvious.

jeromekelleher (commented):
LGTM, minor comment.

jeromekelleher:
> Currently showing an exciting CI failure which is nothing to do with this PR. Have you seen this before @benjeffery ?

yes, this is getting increasingly frequent and annoying. Can we just skip this test please @benjeffery ?

python/tskit/trees.py
hyanwong: Yeah, it's a generator - method picked straight off Stack Overflow. But easy to switch to your version.

jeromekelleher:
Looks like we're missing code coverage

hyanwong:
So we are. But I can't see where:

Screenshot 2021-03-19 at 16 30 56

Is this because we also need to test when the edge_diffs() iterator returns immediately?

hyanwong:
> Is this because we also need to test when the edge_diffs() iterator returns immediately?

Oh no, it can't be that, because `get_example_tree_sequences()` yields an empty TS as the last iteration. So I have no idea what the coverage complaint is about here. What am I missing @jeromekelleher ?

python/tskit/trees.py
```diff
@@ -2209,6 +2209,16 @@ def nodes(self, root=None, order="preorder"):
             for u in roots:
                 yield from iterator(u)
 
+    def _node_edges(self):
+        """
+        Return a numpy array mapping the node IDs in this tree to the ID of the edge
+        above them. This is in lieu of a tree.edge(u) function, currently implemented
+        using the non-optimised Python TreeSequence._tree_node_edges() generator
+        """
+        for index, node_edge_map in enumerate(self.tree_sequence._tree_node_edges()):
```
jeromekelleher, line 2218: I think we need a ``# pragma: no branch`` here, because this loop never terminates

 https://coverage.readthedocs.io/en/coverage-5.5/branch.html#structurally-partial-branches

python/tskit/trees.py
```diff
@@ -4536,6 +4546,20 @@ def edge(self, id_):
...
         )
 
+    def _tree_node_edges(self):
+        """
+        Return a generator over the trees in the tree sequence, yielding a numpy array
+        that maps the node IDs in the tree to the ID of the edge above them.
+        Currently this is a private, non-optimised Python implementation.
+        """
+        node_edges = np.full(self.num_nodes, NULL, dtype=np.int32)
+        for _, edges_out, edges_in in self.edge_diffs():
```
jeromekelleher, line 4556: I have no idea why this one is marked as partial

jeromekelleher (commented):
Weird. Can you try the pragma for the first case? I don't understand the second one.

python/tskit/trees.py
benjeffery: This one also never gets to `StopIteration` as the caller stops early.

hyanwong: I don't quite get why. If I do `zip(ts.trees(), ts.edge_diffs())` does it not reach `StopIteration` simultaneously on both iterators?

benjeffery (approved):
LGTM - do we want an issue for the optimised versions of these?

hyanwong:
> LGTM - do we want an issue for the optimised versions of these?

There's a closed one at https://github.com/tskit-dev/tskit/issues/186

python/tskit/trees.py
jeromekelleher: Ah, no. Zip will always take the shortest iterator, so won't check the rest of the iters.

jeromekelleher: Try reversing the order.

benjeffery: No, just on one of them

benjeffery:
> yes, this is getting increasingly frequent and annoying. Can we just skip this test please @benjeffery ?

I'd like to pin down why this is happening



> > LGTM - do we want an issue for the optimised versions of these?
> 
> There's a closed one at #186

I've added the future label to that issue.

python/tskit/trees.py
jeromekelleher: Or itertools.zip_longest should exhaust bother iters.

jeromekelleher:
Doesn't look like the ``#pragma`` did anything. Let's just remove it and move on I'd say.

hyanwong:
> Doesn't look like the `#pragma` did anything. Let's just remove it and move on I'd say.

I think it's because it's long enough to make the line wrap, and the pragmas [don't like that](https://stackoverflow.com/questions/54111889/coverage-pragma-no-branch-with-multiline-statement). It's failing on that other thing now though.

jeromekelleher:
@mergifyIO rebase

hyanwong:
Yay. Covered now. I just moved the pragma comment to the previous line.

python/tskit/trees.py
grahamgower: Just FYI, Python 3.10 adds a boolean `strict` keyword arg to `zip()` which when `True` will raise a `ValueError` if the iterators don't all have the same length (consuming an equal number of elements from each iterator as a side effect).
https://www.python.org/dev/peps/pep-0618/

PR #1261, opened 2021-03-20
Make metadata lazy for table row classes

benjeffery:
We were decoding metadata on every table row access previously, this makes that lazy. This same method will be applied to the container classes that get returned from e.g. `ts.node(u)` when they are dataclasses.

Note this also gives table row objects a `.table` attribute (@hyanwong might be interested) This attribute does not get put into the result of `dataclasses.asdict` or used in the auto dataclasses methods such as `__eq__` or `__repr__`. These auto methods also only see decoded metadata. The dataclasses are still "frozen".

Needs a couple of extra tests and and one optimisation for the null schema case before merging. Also thinking the schema reference should be stored, as otherwise it could be modified before use by the row.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1241,17 +1241,7 @@ class TestTreeSequence(HighLevelTestCase):
         assert ts.num_mutations == len(other_mutations)
         assert ts.num_mutations == len(mutations)
         for mut, other_mut in zip(mutations, other_mutations):
-            # We cannot compare these directly as the mutations obtained
-            # from the mutations iterator will have extra deprecated
-            # attributes.
-            assert mut.id == other_mut.id
-            assert mut.site == other_mut.site
-            assert mut.parent == other_mut.parent
-            assert mut.node == other_mut.node
-            assert mut.metadata == other_mut.metadata
-            # Check the deprecated attrs.
-            assert mut.position == ts.site(mut.site).position
-            assert mut.index == mut.site
+            assert mut == other_mut
 
     def test_sites_mutations(self):
         # Check that the mutations iterator returns the correct values.
@@ -2103,17 +2093,7 @@ class TestTree(HighLevelTestCase):
         assert tree.num_mutations == len(other_mutations)
         assert tree.num_mutations == len(mutations)
         for mut, other_mut in zip(mutations, other_mutations):
-            # We cannot compare these directly as the mutations obtained
-            # from the mutations iterator will have extra deprecated
-            # attributes.
-            assert mut.id == other_mut.id
-            assert mut.site == other_mut.site
-            assert mut.parent == other_mut.parent
-            assert mut.node == other_mut.node
-            assert mut.metadata == other_mut.metadata
-            # Check the deprecated attrs.
-            assert mut.position == tree.tree_sequence.site(mut.site).position
-            assert mut.index == mut.site
+            assert mut == other_mut
 
     def test_simple_mutations(self):
         tree = self.get_tree()
@@ -2991,10 +2971,10 @@ class SimpleContainersWithMetadataMixin:
         (inst,) = self.get_instances(1)
         (inst2,) = self.get_instances(1)
         assert inst == inst2
-        inst._metadata_decoder = lambda m: "different decoder"
+        inst.metadata
         assert inst == inst2
-        inst._encoded_metadata = b"different"
-        assert not (inst == inst2)
+        inst._metadata = "different"
+        assert inst != inst2
 
     def test_decoder_run_once(self):
         # For a given instance, the decoded metadata should be cached, with the decoder
@@ -3002,6 +2982,7 @@ class SimpleContainersWithMetadataMixin:
         (inst,) = self.get_instances(1)
         times_run = 0
 
+        # Hack in a tracing decoder
         def decoder(m):
             nonlocal times_run
             times_run += 1
@@ -3019,12 +3000,12 @@ class TestIndividualContainer(SimpleContainersMixin, SimpleContainersWithMetadat
     def get_instances(self, n):
         return [
             tskit.Individual(
-                id_=j,
+                id=j,
                 flags=j,
                 location=[j],
                 parents=[j],
                 nodes=[j],
-                encoded_metadata=b"x" * j,
+                metadata=b"x" * j,
                 metadata_decoder=lambda m: m.decode() + "decoded",
             )
             for j in range(n)
@@ -3035,12 +3016,12 @@ class TestNodeContainer(SimpleContainersMixin, SimpleContainersWithMetadataMixin
     def get_instances(self, n):
         return [
             tskit.Node(
-                id_=j,
+                id=j,
                 flags=j,
                 time=j,
                 population=j,
                 individual=j,
-                encoded_metadata=b"x" * j,
+                metadata=b"x" * j,
                 metadata_decoder=lambda m: m.decode() + "decoded",
             )
             for j in range(n)
@@ -3055,9 +3036,9 @@ class TestEdgeContainer(SimpleContainersMixin, SimpleContainersWithMetadataMixin
                 right=j,
                 parent=j,
                 child=j,
-                encoded_metadata=b"x" * j,
+                metadata=b"x" * j,
                 metadata_decoder=lambda m: m.decode() + "decoded",
-                id_=j,
+                id=j,
             )
             for j in range(n)
         ]
@@ -3067,11 +3048,11 @@ class TestSiteContainer(SimpleContainersMixin, SimpleContainersWithMetadataMixin
     def get_instances(self, n):
         return [
             tskit.Site(
-                id_=j,
+                id=j,
                 position=j,
                 ancestral_state="A" * j,
                 mutations=TestMutationContainer().get_instances(j),
-                encoded_metadata=b"x" * j,
+                metadata=b"x" * j,
                 metadata_decoder=lambda m: m.decode() + "decoded",
             )
             for j in range(n)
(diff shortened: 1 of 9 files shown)


python/tskit/tables.py
```diff
@@ -43,16 +43,55 @@
...
+    def new_init(self, *args, table=None, **kwargs):
+        builtins.object.__setattr__(self, "table", table)
+        wrapped_init(self, *args, **kwargs)
+
+    cls.__init__ = new_init
+    cls.metadata = CachedMetadata()
+    return cls
+
+
+@lazy_metadata
```
jeromekelleher, line 87: Not obvious here why we need both a decorator and to inherit from LazyMetadata - one or other is easy to follow for a reader, but both is confusing. What's the rationale here?

jeromekelleher: I guess we should spell it out up here where these classes are defined anyway, need to encode this knowledge somewhere.

jeromekelleher (commented):
Looks good, see some comments.

python/tskit/tables.py
benjeffery: I couldn't figure out how to make an extra slot with a class decorator, annoying!

jeromekelleher: Should we bother with the decorator then? We should be able to implement this in a straighforward inheritancy type way, right?

benjeffery: Manged to do it all in a decorator in the end, can't do inheritance due to intercepting the constructor.

benjeffery:
@jeromekelleher I think this could have a review now. Docs are not building due to a warning I haven't figured out yet, but look good ignoring the warning.

Note that I've removed the deprecated attrs on mutation - these could be added back if you think it is too early.

python/tests/simplify.py
```diff
@@ -218,7 +219,7 @@ def record_edge(self, left, right, parent, child):
         else:
             last = self.edge_buffer[child][-1]
             if last.right == left:
-                last.right = right
+                builtins.object.__setattr__(last, "right", right)
```
jeromekelleher, line 222: This isn't a pattern I want to encourage people to use. This seems worse than not having Frozen instances.

python/tskit/tables.py
```diff
@@ -43,54 +42,55 @@
...
-            )
+        return (
+            isinstance(other, IndividualTableRow)
+            and self.flags == other.flags
+            and np.array_equal(self.location, other.location)
+            and np.array_equal(self.parents, other.parents)
+            and self.metadata == other.metadata
+        )
 
     def __neq__(self, other):
```
jeromekelleher, line 67: Two things here:

1. I think we can get rid of ``__neq__`` now, this is handled automatically in more recent versions of Python
2. Can we use the dataclass ``__eq__``? If not, let's add a comment saying why, as this will be a recurring question.

(Ah, answering my own question here: we have numpy arrays. Good to add a quick comment here to that effect)

python/tskit/trees.py
```diff
@@ -78,158 +79,96 @@ def span(self):
...
+    """
+    nodes: np.ndarray
+    """
+    The IDs of the nodes that are associated with this individual as
+    a numpy array (dtype=np.int32). If no nodes are associated with the
+    individual this array will be empty.
+    """
+    metadata: Any
+    """
+    The decoded :ref:`metadata <sec_metadata_definition>`
```
jeromekelleher, line 122: Well, only decoded if a schema is present, right?

python/tskit/trees.py
```diff
@@ -304,213 +238,207 @@ def span(self):
...
+    def __eq__(self, other):
+        return (
+            isinstance(other, Site)
+            and self.id == other.id
+            and self.position == other.position
+            and self.ancestral_state == other.ancestral_state
+            and np.array_equal(self.mutations, other.mutations)
+        )
+
+    def __neq__(self, other):
```
jeromekelleher, line 282: Don't need neq I think

python/tskit/trees.py
```diff
@@ -304,213 +238,207 @@ def span(self):
...
+    The decoded :ref:`metadata <sec_metadata_definition>` for this
+    mutation.
+    """
+    time: float
+    """
+    The occurrence time of this mutation.
     """
 
+    # To get default values on a frozen instance you have work around the freeze
     def __init__(
```
jeromekelleher, line 336: Would the post_init hook be better?

python/tskit/trees.py
```diff
@@ -560,71 +488,95 @@ class Variant(SimpleContainer):
     Modifying the attributes in this class will have **no effect** on the
     underlying tree sequence data.
 
-    :ivar site: The site object for this variant.
+    :ivar site:
```
jeromekelleher, line 491: We can get rid of the ivars entirely I think (hope?)

python/tskit/metadata.py
```diff
@@ -39,6 +41,8 @@
 import tskit
 import tskit.exceptions as exceptions
 
+__setattr__ = builtins.object.__setattr__
```
jeromekelleher, line 44: I wonder about the wisdom of this - don't we then export this into our namespace? If we're going to do something this ugly, the I'd prefer to feel the full brunt of the ugliness each time we use it, just to be reminded of how ugly it is.

jeromekelleher (commented):
Looks good, but I'm seriously questioning whether making things frozen is worth it. What problem are we trying to solve here?

Having to resort to this ``__setattr__`` stuff to do perfectly reasonable things seems much more likely to result in errors than anything the frozen is trying to prevent. Plus, I think we are likely to break code if we change the semantics to frozen now.

If we're trying to allow for things like:
```
tables.nodes[0].flags = 0
```

then we perhaps we could have table rows frozen for now (to allow this in future) but keep the stuff returned by the TreeSequence unfrozen?

jeromekelleher:
> Note that I've removed the deprecated attrs on mutation - these could be added back if you think it is too early.

When were they marked as deprecated? We should update the changelog with details.

benjeffery:
> Looks good, but I'm seriously questioning whether making things frozen is worth it. What problem are we trying to solve here?

The idea was to prevent users thinking they could mutate the classes - as it warned about in the old docs.  

> Having to resort to this `__setattr__` stuff to do perfectly reasonable things seems much more likely to result in errors than anything the frozen is trying to prevent. Plus, I think we are likely to break code if we change the semantics to frozen now.
> 
> If we're trying to allow for things like:
> 
> ```
> tables.nodes[0].flags = 0
> ```
> 
> then we perhaps we could have table rows frozen for now (to allow this in future) but keep the stuff returned by the TreeSequence unfrozen?

Do you mean the other way round? As in `TableRow` classes are unfrozen to allow for future mutation, and `Node` et al. are frozen? This seems to make sense to me.

python/tskit/metadata.py
benjeffery: I imported it to avoid the double lookup on each row access - I thinking importing it to something called `__builtin__object__setattr__` would be ugly enough?

python/tskit/trees.py
benjeffery: Possibly - would still need to use the ugly setattr.

python/tests/simplify.py
benjeffery: Yes agreed - I could rejig these instances to not use the ugly way. It only really needs to be used in the custom `__init__` and the metadata decorator.

jeromekelleher:
> Do you mean the other way round? As in TableRow classes are unfrozen to allow for future mutation, and Node et al. are frozen? This seems to make sense to me.

No - to me the only point in freezing the classes is so that we don't break people's code later when we change the semantics to "push the changes back into the tables". So, let people mutate away on the Node, Edge etc classes - we document that these are copies of the original data, and they can do whatever they like with them. Freezing the classes just for the sake of it, or to somehow prevent people from writing bad code seems ... unpythonic to me.

jeromekelleher:
How about we make the minimal changes to dataclasses first as cleanly as we can with the same semantics as we currently have (ish), and then look at whether we should freeze or not as a second pass?

benjeffery:
@jeromekelleher OK, semantics are now exactly as they were before. Think I have addressed the comments too. Just the docs 
to fix, for some reason it isn't finding the `Site` class.

python/tests/test_highlevel.py
```diff
@@ -2991,17 +2971,18 @@ def test_metadata(self):
         (inst,) = self.get_instances(1)
         (inst2,) = self.get_instances(1)
         assert inst == inst2
-        inst._metadata_decoder = lambda m: "different decoder"
+        inst.metadata
```
jeromekelleher, line 2974: What does this statement do?

python/tskit/tables.py
```diff
@@ -46,51 +45,49 @@
...
-@dataclass(eq=False, **dataclass_options)
+@metadata.lazy_decode
+@dataclass(**dataclass_options)
 class IndividualTableRow:
     __slots__ = ["flags", "location", "parents", "metadata"]
     flags: int
     location: np.ndarray
     parents: np.ndarray
-    metadata: bytes
+    metadata: Any
```
jeromekelleher, line 55: We could say ``metadata : Union[bytes, dict]`` couldn't we?

jeromekelleher (approved):
Looks great, seems like a major improvement. One comment on the type annotations.

python/tests/test_highlevel.py
benjeffery: I thought it was forcing decode of the metadata, but that is done by the equality operator above anyway, removed.

python/tskit/tables.py
benjeffery: Changed to `Optional[Union[bytes, dict]]`

PR #1264, opened 2021-03-23
Replace interval numbers with attr names

benjeffery:
I looked into replacing `Interval` with a dataclass, but we iterate on it a lot, so left it as is. I thought it worth keeping these changes from `[0]` to `.left`.


jeromekelleher (approved):
Nice improvement to readability, thanks. 

We definitely do need to keep the tuple interpretation though.

Issue #1263, opened 2021-03-22
tree_height_scale="rank"  Y axis should show time values, not rank


jeromekelleher:
The current behaviour of tree_height_scale="rank" isn't terribly useful, as we just show [0, 1, 2,...] in the Y axis.

We should show the actual time values, like we do in the text view:
```
37.00┊        13       ┊  
     ┊     ┏━━━┻━━━┓   ┊  
10.00┊     ┃      12   ┊  
     ┊     ┃     ┏━┻━┓ ┊  
8.00 ┊     ┃    11   ┃ ┊  
     ┊     ┃    ┏┻━┓ ┃ ┊  
5.00 ┊    10    ┃  ┃ ┃ ┊  
     ┊  ┏━━╋━┓  ┃  ┃ ┃ ┊  
1.00 ┊  9  ┃ ┃  8  ┃ ┃ ┊  
     ┊ ┏┻┓ ┃ ┃ ┏┻┓ ┃ ┃ ┊  
0.00 ┊ 0 3 1 7 2 6 5 4 ┊  
   0.00              1.00 
```
This might be slightly confusing, but it's very useful when you want to see a compact representation of the tree and want to know what the times are. There's basically no information in [0, 1, 2..]

PR #1266, opened 2021-03-24
Print actual time values for ranked times

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/1263

Note that the `y_ticks` param now allows either a list of positions (as before) *or* a dict of `{pos1: label1, pos2: label2}`. The dict use is not documented, and neither is the fact that the position used when placing bespoke labels with `scale="rank"` is the rank position, not the actual time.

# PR Checklist:

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -30,6 +30,9 @@ import math
 import numbers
 import operator
 from dataclasses import dataclass
+from typing import List
+from typing import Mapping
+from typing import Union
 
 import numpy as np
 import svgwrite
@@ -122,15 +125,16 @@ def check_x_scale(x_scale):
     return x_scale
 
 
-def check_ticks(ticks, default_iterable):
+def check_y_ticks(ticks: Union[List, Mapping, None]) -> Mapping:
     """
-    This is trivial, but implemented as a function so that later we can implement a tick
-    locator function, such that e.g. ticks=5 selects ~5 nicely spaced tick locations
-    (ideally with sensible behaviour for log scales)
+    Later we might want to implement a tick locator function, such that e.g. ticks=5
+    selects ~5 nicely spaced tick locations (with sensible behaviour for log scales)
     """
     if ticks is None:
-        return default_iterable
-    return ticks
+        return {}
+    if isinstance(ticks, Mapping):
+        return ticks
+    return {pos: f"{pos:.2f}" for pos in ticks}
 
 
 def rnd(x):
@@ -387,7 +391,7 @@ class SvgPlot:
             x_label = "Genome position"
         if y_label is None and y_axis:
             if tree_height_scale == "rank":
-                y_label = "Ranked node time"
+                y_label = "Node time"
             else:
                 y_label = "Time"
         self.x_label = x_label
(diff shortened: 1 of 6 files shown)


benjeffery (approved):
Seems fine to me as we're not changing the API, just adding an internal thing.

hyanwong:
> Seems fine to me as we're not changing the API, just adding an internal thing.

Well, the API now allows a dict to be passed to `y_ticks`. But we don't document that.

python/tskit/drawing.py
```diff
@@ -527,7 +527,7 @@ def draw_y_axis(
         self,
         upper=None,  # In plot coords
         lower=None,  # In plot coords
-        tick_positions=None,
+        ticks=(),  # A list of numerical positions or a dictionary of pos->label
```
jeromekelleher, line 530: It's not obvious why this default value is treated differently to the others and this syntax will be opaque to some.

python/tskit/drawing.py
```diff
@@ -551,28 +551,31 @@ def draw_y_axis(
...
         if self.y_axis:
             y_axis.add(dwg.line((x, rnd(lower)), (x, rnd(upper))))
-            if tick_positions is not None:
-                for pos in tick_positions:
-                    tick = y_axis.add(
-                        dwg.g(
-                            class_="tick",
-                            transform=f"translate({x} {rnd(self.y_transform(pos))})",
-                        )
+            try:
```
jeromekelleher, line 554: There's awful lot of logic in this function already. How about 
```python
def check_yticks(ticks: Union[List, Mapping, None]) -> Mapping: # Not sure what the actual types would be here, just e.g.
    if ticks is None:
        return {}
    if isinstance(ticks, Mapping):
       return ticks
   return {pos: "{pos:.2f}" for pos in ticks}
```

The check_yticks function can then be tested independently, and it's not bound up with loads of other, unrelated, logic.

Then later:

```
ticks = check_yticks(ticks)
```

python/tskit/drawing.py
```diff
@@ -551,28 +551,31 @@ def draw_y_axis(
...
+                tick_positions = ticks.keys()
+                tick_labels = ticks.values()
+            except AttributeError:
+                tick_positions = ticks
+                tick_labels = (f"{pos:.2f}" for pos in tick_positions)
+
+            for label, pos in zip(tick_labels, tick_positions):
```
jeromekelleher, line 561: I don't see a good reason for unzipping the ticks dictionary into two lists here, we just have to zip them back together again.

jeromekelleher (commented):
Looks OK, but things are getting convoluted. Let's try and pull some logic out of these big functions.

python/tskit/drawing.py
hyanwong: Yeah, I had `None` originally, but then I need to swap None for e.g. `[]` and test it, which takes more code lines, so 🤷

python/tskit/drawing.py
hyanwong: Yep, I guess that's possible. I'll work though the logic to see if it hangs together.

python/tskit/drawing.py
hyanwong: True - probably a hangover from when they were separate params. I'll change

hyanwong:
Good call on those changes. I've made them and it's looking much nicer. Thanks @jeromekelleher 

I also tacked on a (separate) commit to make the SVGs msprime version-independent. It's a bit ugly (rounds the node times & edge positions to 5dp in one of the examples), but is only in the testing code, so I thought I would include them to make @benjeffery 's life easier for switching to msprime 1.0 for CI.

jeromekelleher (approved):
LGTM

PR #1267, opened 2021-03-24
Use msprime alpha on CI

benjeffery:
This works - although we get different SVGs, so we won't be able to get both versions of msprime to pass with the same expected SVG files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           # It's sometimes necessary to nuke the cache, and the simplest
           # way to do it is to change the key. We can increment this
           # version number when we want to do this.
-          key: tskit-{{ .Branch }}-v4
+          key: tskit-{{ .Branch }}-v5
       - run:
           name: Checkout submodules
           command: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,7 +54,7 @@ jobs:
             /usr/share/miniconda/envs/anaconda-client-env
             ~/osx-conda
             ~/.profile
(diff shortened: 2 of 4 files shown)


jeromekelleher:
We could pull the alpha from conda-forge too, like we're doing in stdpopsim at the moment.

jeromekelleher:
https://github.com/popsim-consortium/stdpopsim/blob/6e51b822aa6885a850fa68654746e1426e0eaccc/.github/workflows/tests.yml#L81

benjeffery:
> We could pull the alpha from conda-forge too, like we're doing in stdpopsim at the moment.

As msprime 1.0 has wheels we don't actually need conda at all in the tskit CI anymore - is there any reason to keep it? We were only installing msprime with it and I don't think there is any meaningful testing that comes from using it.

benjeffery:
I've added a commit to do that, seems to work!

jeromekelleher:
That numpy ufunc failure thing looks nasty - any idea what's happening there?

hyanwong:
I have a fix for one of the SVG errors, so that it passes on both new and old. Do you want it @benjeffery ? The other one is discussed in Slack and I think is unavoidable (I could try making an example TS a different way, though).

hyanwong:
> I have a fix for one of the SVG errors, so that it passes on both new and old. Do you want it @benjeffery ? The other one is discussed in Slack and I think is unavoidable (I could try making an example TS a different way, though).

Actually there's a fix for both at https://github.com/tskit-dev/tskit/pull/1266/commits/9ec3ee0f0436fea6a19968ba24f3b8d60b4e2199

benjeffery:
> That numpy ufunc failure thing looks nasty - any idea what's happening there?

Cache issue, fixed.

jeromekelleher (approved):

PR #1272, opened 2021-03-26
Use defaults when decoding JSON metadata

### LINKED ISSUE #1073: Metadata decoding should use default value from schema

benjeffery:
Fixes #1073 

Only uses defaults for the first-level of objects. Raises error if the schema specifies deeper defaults.

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -65,8 +65,9 @@ JSON
 When ``json`` is specified as the ``codec`` in the schema the metadata is encoded in
 the human readable `JSON <https://www.json.org/json-en.html>`_ format. As this format
 is human readable and encodes numbers as text it uses more bytes than the ``struct``
-format. However it is simpler to configure as it doesn't have any specific constraints and
-doesn't require any format specifiers for each type in the schema.
+format. However it is simpler to configure as it doesn't require any format specifier
+for each type in the schema. Default values for properties can be specified for only
+the shallowest level of the metadata object.
 
 ------
 struct
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -60,17 +60,19 @@ def replace_root_refs(obj):
 TSKITMetadataSchemaValidator = jsonschema.validators.extend(
     jsonschema.validators.Draft7Validator
 )
-META_SCHEMA: Mapping[str, Any] = copy.deepcopy(TSKITMetadataSchemaValidator.META_SCHEMA)
+deref_meta_schema: Mapping[str, Any] = copy.deepcopy(
+    TSKITMetadataSchemaValidator.META_SCHEMA
+)
 # We need a top-level only required property so we need to rewrite any reference
 # to the top-level schema to a copy in a definition.
-META_SCHEMA = replace_root_refs(META_SCHEMA)
-META_SCHEMA["definitions"]["root"] = copy.deepcopy(META_SCHEMA)
-META_SCHEMA["codec"] = {"type": "string"}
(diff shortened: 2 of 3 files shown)


benjeffery:
Codecov is unhappy - but the diff looks all covered in the report?!

python/tskit/metadata.py
```diff
@@ -183,54 +218,59 @@ def required_validator(validator, required, instance, schema):
     TSKITMetadataSchemaValidator,
     {"type": binary_format_validator, "required": required_validator},
 )
-META_SCHEMA: Mapping[str, Any] = copy.deepcopy(StructCodecSchemaValidator.META_SCHEMA)
+stuct_meta_schema: Mapping[str, Any] = copy.deepcopy(
```
jeromekelleher, line 221: typo: stuct -> struct

docs/metadata.rst
```diff
@@ -65,8 +65,9 @@ JSON
 When ``json`` is specified as the ``codec`` in the schema the metadata is encoded in
 the human readable `JSON <https://www.json.org/json-en.html>`_ format. As this format
 is human readable and encodes numbers as text it uses more bytes than the ``struct``
-format. However it is simpler to configure as it doesn't have any specific constraints and
-doesn't require any format specifiers for each type in the schema.
+format. However it is simpler to configure as it doesn't require any format specifier
+for each type in the schema. Default values for properties can be specified for the only
```
jeromekelleher, line 69: extra "the only" here I think.

jeromekelleher (commented):
Looks like codecov is complaining about the partial hits. Since these are in critical code paths we should probably try to cover them all right.

jeromekelleher (approved):
LGTM

benjeffery:
Turns out "dark mode" in chrome hides the partial coverage lines! Fixed now.

jeromekelleher:
Hmm, I think we might have some stability issues with the C API, there's been a few different failures recently.

Can you try running stress_lowlevel for a while please @benjeffery, and see if that surfaces anything?

benjeffery:
Always fails on circle CI only though right?

jeromekelleher:
Yes, as far as I've seen.

PR #1274, opened 2021-03-27
Make returned values for edge diffs a namedtuple

hyanwong:
I always have to force myself to remember that the returned values from the `edge_diffs` iterator are `(intvl, out, in)` not `(intvl, in, out)` - I think it's because the Hokey Cokey goes in, out, in , out before shaking it all about.

This trivial addition should make it much clearer and stop mistakes down the line.

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -653,6 +653,13 @@ class TestTreeSequence(HighLevelTestCase):
         for ts in get_example_tree_sequences():
             self.verify_edge_diffs(ts)
 
+    def test_edge_diffs_names(self, simple_degree2_ts_fixture):
+        for val in simple_degree2_ts_fixture.edge_diffs():
+            assert len(val) == 3
+            assert val[0] == val.interval
+            assert val[1] == val.edges_out
+            assert val[2] == val.edges_in
+
     def test_edge_diffs_include_terminal(self):
         for ts in get_example_tree_sequences():
             edges = set()
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -59,6 +59,8 @@ CoalescenceRecord = collections.namedtuple(
 
 BaseInterval = collections.namedtuple("BaseInterval", ["left", "right"])
 
+EdgeDiff = collections.namedtuple("EdgeDiff", ["interval", "edges_out", "edges_in"])
+
 
 class Interval(BaseInterval):
     """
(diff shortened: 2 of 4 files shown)


python/tskit/trees.py
```diff
@@ -59,6 +59,8 @@
 
 BaseInterval = collections.namedtuple("BaseInterval", ["left", "right"])
 
+EdgeDiffs = collections.namedtuple("EdgeDiffs", ["interval", "edges_out", "edges_in"])
```
jeromekelleher, line 62: I think this is probably better as singular, ``EdgeDiff``. Each thing we return is a single diff, and together they are the diffs.

jeromekelleher (commented):
LGTM, and a good idea.

We should explicitly test the new and old behaviour, though.

hyanwong:
I took the opportunity to change the name of `simple_ts_fixture` to `simple_degree1_ts_fixture`, as there's only one tree in the ts, so it's useful to be clear about that in case we use it elsewhere in testing (I used `simple_degree2_ts_fixture` for the new edge_diffs test)

I called a tree sequence with one/two trees a "degree1/degree2 ts", but perhaps we want a different term. I know that @jeromekelleher wanted to avoid "unary/binary" because of confusion with node arity. I would prefer to avoid singleton / doubleton because of confusion with singleton & doubleton variants. We could, I suppose, use [monadic / dyadic](https://en.wikipedia.org/wiki/Arity#Terminology), but that sounds a bit maths-jargony to me. Other ideas welcome.

jeromekelleher:
The fixtures thing is a disimprovement if anything for me to be honest - making the variable name in the test depend on the global fixture name seems very fiddly. I'd prefer if we moved away from fixtures and towards a set of standard tree sequences that we ``parametrize`` by. So, this is fine for now, but let's not go any further in this direction please.

jeromekelleher (approved):

hyanwong:
> The fixtures thing is a disimprovement if anything for me to be honest - making the variable name in the test depend on the global fixture name seems very fiddly. I'd prefer if we moved away from fixtures and towards a set of standard tree sequences that we `parametrize` by. 

Yes, I agree. A standardised set is definitely the way forward. But that's not available yet, right?

jeromekelleher:
> Yes, I agree. A standardised set is definitely the way forward. But that's not available yet, right?

No.

hyanwong:
p.s. even in the parameterised form we presumably do want a name for the simple 1-tree or 2-tree ts so that we can pick just that one as the parameter? So the naming problem is still there.

jeromekelleher:
I just don't want to have to change loads of parameter names across the test suite later. We will need to think about the naming thing globally; I don't see any point in thinking about this unless we're looking at the whole picture (and we have other things to do right now). What's here is fine for now, let's just not go any further down this path.

python/tests/test_highlevel.py
```diff
@@ -653,6 +654,13 @@ def test_edge_diffs(self):
         for ts in get_example_tree_sequences():
             self.verify_edge_diffs(ts)
 
+    def test_edge_diffs_names(self, simple_degree2_ts_fixture):
```
benjeffery, line 657: This tests the item getter of the new named tuple, but we've lost the test of nested destructuring that was `for (left, right), edges_out, edges_in in ts.edge_diffs():`

benjeffery (approved):
Great, thanks @hyanwong this is an improvment. Just one comment on a test that has been lost in the conversion.

python/tests/test_highlevel.py
hyanwong: Oh yes. Good point. I've reinstated the old test scheme. I guess together with the new one that's covered now.

hyanwong:
Getting those "numpy.ufunc size changed, may indicate binary incompatibility" warning again, although the actual error is some multiprocessing thing here.

jeromekelleher:
I tried rerunning the circleCI job as it's unrelated to this PR. There's definitely something fishy going on here though @benjeffery, let's keep an eye on it.

benjeffery:
> I tried rerunning the circleCI job as it's unrelated to this PR. There's definitely something fishy going on here though @benjeffery, let's keep an eye on it.

Yep, don't like the look of it either, will try to recreate locally.

PR #1275, opened 2021-03-27
Clarify that Tree.samples() returns (numerical) node IDs

hyanwong:
It wasn't clear if the iterator was over node objects or IDs. Additionally make it clear that these are sample node IDs, not sample individual IDs.


petrelharp:
LGTM!

jeromekelleher (approved):

PR #1277, opened 2021-03-28
Clarify order or IDs returned by .samples()

### LINKED ISSUE #1276: Clarify if .samples() methods return ids in order
I *think* that the IDs returned by ts.samples() and ts.first.samples() are always in ascending order. If this is the case, could we note it in the documentation?

hyanwong:
Fixes #1276


benjeffery (approved):
LGTM!

PR #1279, opened 2021-03-29
Add has_single_root and has_multiple_roots

### LINKED ISSUE #958: Check if a tree (or ts) is multi-root?
Whenever I look at the [docs for num_roots](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.Tree.num_roots) the bit that says "Requires O(number of roots) time." puts me off using this route to check if I have a multiroot tree - I'm worried about how inefficient it might be. Yet sometimes I don't want the actual root, I just want to assert that I'm dealing with a single root tree. Might it be useful therefore to have an "is_multiroot()` method on a tree?

I also wonder if this is something we want to check for all the trees in a TS, and if so, it there's a more efficient way of doing this than just going through the trees one-by-one.

hyanwong:
Fixes #958. Also added the word "only" to "requires O(n) time", as discussed with @petrelharp on Slack.

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2148,6 +2148,26 @@ class TestTree(HighLevelTestCase):
         for u in tree.nodes():
             assert tree.num_children(u) == len(tree.children(u))
 
+    def test_root_properties(self):
+        for ts in get_example_tree_sequences():
+            for tree in ts.trees():
+                if tree.has_single_root:
+                    assert tree.num_roots == 1
+                    assert tree.root != tskit.NULL
+                elif tree.has_multiple_roots:
+                    assert tree.num_roots > 1
+                    with pytest.raises(ValueError, match="More than one root exists"):
+                        _ = tree.root
+                else:
+                    assert tree.num_roots == 0
+                    assert tree.root == tskit.NULL
+
+    def test_root_properties_empty_ts(self):
+        # NB - this can be removed once the example_tree_sequences contain an empty ts
+        tree = tskit.TableCollection(sequence_length=1).tree_sequence().first()
+        assert tree.num_roots == 0
+        assert tree.root == tskit.NULL
+
     def verify_newick(self, tree):
         """
         Verifies that we output the newick tree as expected.
@@ -2399,7 +2419,7 @@ class TestTree(HighLevelTestCase):
             "breadthfirst",
             "minlex_postorder",
         ]
-        if tree.num_roots == 1:
+        if tree.has_single_root:
             with pytest.raises(ValueError):
                 list(t1.nodes(order="bad order"))
             assert list(t1.nodes()) == list(t1.nodes(t1.get_root()))
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1269,11 +1269,37 @@ class Tree:
         The number of roots in this tree, as defined in the :attr:`~Tree.roots`
         attribute.
 
-        Requires O(number of roots) time.
+        Only requires O(number of roots) time.
 
         :rtype: int
         """
         return self._ll_tree.get_num_roots()
+
+    @property
+    def has_single_root(self):
+        """
+        ``True`` if this tree has a single root, ``False`` otherwise.
+        Equivalent to tree.num_roots == 1. This is a O(1) operation.
+
+        :rtype: bool
+        """
+        root = self.left_root
+        if root != NULL and self.right_sib(root) == NULL:
+            return True
+        return False
+
+    @property
+    def has_multiple_roots(self):
+        """
+        ``True`` if this tree has more than one root, ``False`` otherwise.
+        Equivalent to tree.num_roots > 1. This is a O(1) operation.
+
+        :rtype: bool
+        """
+        root = self.left_root
+        if root != NULL and self.right_sib(root) != NULL:
+            return True
+        return False
 
     @property
     def roots(self):
@@ -1293,7 +1319,7 @@ class Tree:
 
         The roots of the tree are returned in a list, in no particular order.
 
-        Requires O(number of roots) time.
+        Only requires O(number of roots) time.
 
         :return: The list of roots in this tree.
         :rtype: list
@@ -1319,10 +1345,9 @@ class Tree:
         :rtype: int
         :raises: :class:`ValueError` if this tree contains more than one root.
         """
-        root = self.left_root
-        if root != NULL and self.right_sib(root) != NULL:
+        if self.has_multiple_roots:
             raise ValueError("More than one root exists. Use tree.roots instead")
-        return root
+        return self.left_root
 
     def get_index(self):
         # Deprecated alias for self.index


python/tests/test_highlevel.py
```diff
@@ -2141,6 +2141,20 @@ def test_num_children(self):
...
+                if tree.has_single_root:
+                    assert tree.num_roots == 1
+                    assert tree.root != tskit.NULL
+                elif tree.has_multiple_roots:
+                    assert tree.num_roots > 1
+                    with pytest.raises(ValueError, match="More than one root exists"):
+                        _ = tree.root
+                else:
+                    assert tree.num_roots == 0
+                    assert tree.root == tskit.NULL
```
benjeffery, line 2156: Can we assert that all three cases are triggered?

benjeffery (approved):
LGTM! Just one suggestion to make the tests more paranoid.

python/tests/test_highlevel.py
hyanwong: Oh yes, I was actually wondering how to do that. I've done it in the latest commit with a `set()`

hyanwong: Oh, that was a good call. Turns out get_example_tree_sequences() doesn't return a completely empty ts. When it does, it reveals a lot of bugs. I'll try fixing some now, in a different PR.

jeromekelleher (approved):
LGTM. I'm not sure any of the three "Only"s in the documentation add clarity, but if that's what you and Peter decided then fine by me.

python/tskit/trees.py
```diff
@@ -1267,12 +1267,36 @@ def num_roots(self):
...
+        Only requires O(number of roots) time.
 
         :rtype: int
         """
         return self._ll_tree.get_num_roots()
 
+    @property
+    def has_single_root(self):
+        """
+        Only ``True`` if this tree has a single root.
```
jeromekelleher, line 1279: I would say something like "True if this tree has a single root and equivalent to ``tree.num_roots == 1``. This is a O(1) operation."

Good to be precise and also reassure people that it's an efficient operation.

hyanwong: I guess the "Only" makes it clear here that e.g. we don't return False for num_roots == 0. It's an "if and only if". Probably overly pedantic.

hyanwong: Maybe "False otherwise" reads better. I'll change it.

python/tskit/trees.py
```diff
@@ -3826,12 +3851,14 @@ def max_root_time(self):
...
-        with the definition of tree roots, if there are no edges in the tree
-        sequence we return the time of the oldest sample.
+        with the definition of tree roots, if there are no edges in the tree sequence
+        we return the time of the oldest sample, or ``None`` if there are no samples.
 
         :return: The maximum time of a root in this tree sequence.
         :rtype: float
         """
+        if self.num_samples == 0 and self.num_edges == 0:
+            return None
```
benjeffery, line 3861: In other places we raise value error for situations like this, so I'm not sure that `None` is the right return. If we're going with it then it should be in the docstring.

benjeffery (commented):
Thanks for fixing all those edge cases! Not sure about the `None` return though.

python/tests/test_highlevel.py
```diff
@@ -2213,7 +2248,7 @@ def test_newick_topology_equiv(self):
         replace_numeric = {ord(x): None for x in "1234567890:."}
         for ts in get_example_tree_sequences():
             for tree in ts.trees():
-                if tree.num_roots > 1:
+                if not tree.has_single_root:
                     continue
```
benjeffery, line 2252: Shouldn't this be expecting an error rather than continuing?

python/tskit/trees.py
hyanwong: It is in the docstring (above, I think), but perhaps you are right: if `max([])` returns an error then perhaps ts.max_root_time should too? If so we should document that.

python/tests/test_highlevel.py
hyanwong: I think we do that in the simple `test_newick` function. Since we are testing topological equivalence here I thought that was irrelevant. But perhaps we need to be testing that *both* fail in an equivalent way.

hyanwong: And, to follow-up, in pytest, how do I test between equivalence of 2 errors?

jeromekelleher:
Can we roll back on the changes for the empty ts please @hyanwong? This needs separate consideration. I'm "unapproving" in the mean time.

You can test the 0 roots behaviour with a separate small test, which is the preferred pattern anyway. Unit tests should be small and obvious.

hyanwong:
> Can we roll back on the changes for the empty ts please @hyanwong? 

Sure - I can put the commit in a separate PR.

> You can test the 0 roots behaviour with a separate small test, which is the preferred pattern anyway. Unit tests should be small and obvious.

Yep, this was only temporary until we have proper parameterisation for testing different tree seqs. But I think it was useful to change the default bevvy of tests to contain an empty ts, as it picked up buggy behaviour (e.g. for `newick()`), so we should make sure we do test this.

jeromekelleher:
> Yep, this was only temporary until we have proper parameterisation for testing different tree seqs. But I think it was useful to change the default bevvy of tests to contain an empty ts, as it picked up buggy behaviour (e.g. for newick()), so we should make sure we do test this.

Fine, but that's a significant change, unrelated to the title of this PR. Please open an issue to document and we can follow up with another PR.

hyanwong:
> Fine, but that's a significant change, unrelated to the title of this PR. Please open an issue to document and we can follow up with another PR.

Will do.

Annoyingly, I seem to hit this situation all the time: I try to address one issue, and end up having to fix 2 or 3 unrelated issues which arise as a result. But it takes a lot of time to unpick the interrelated changes into semi-independent PRs, because I've usually gone most of the way towards fixing the first issue before I find I need to change the other stuff. Not sure there's an easy solution, but I'm *always* hitting this.

hyanwong:
Just removed the extra commit, and added an extra test with a comment that it can be removed once we add the empty ts to the set we test by default.

benjeffery:
> Annoyingly, I seem to hit this situation all the time: I try to address one issue, and end up having to fix 2 or 3 unrelated issues which arise as a result. But it takes a lot of time to unpick the interrelated changes into semi-independent PRs, because I've usually gone most of the way towards fixing the first issue before I find I need to change the other stuff. Not sure there's an easy solution, but I'm _always_ hitting this.

I'm sure we have talked about this but what is your workflow here? Using `git add -p` and rebasing, along with occasional cherry-picking should make it mostly simple.

hyanwong:
> > Annoyingly, I seem to hit this situation all the time: I try to address one issue, and end up having to fix 2 or 3 unrelated issues which arise as a result. But it takes a lot of time to unpick the interrelated changes into semi-independent PRs, because I've usually gone most of the way towards fixing the first issue before I find I need to change the other stuff. Not sure there's an easy solution, but I'm _always_ hitting this.
> 
> I'm sure we have talked about this but what is your workflow here? Using `git add -p` and rebasing, along with occasional cherry-picking should make it mostly simple.

Actually, I didn't know about `git add -p`. That sounds pretty helpful, thanks.

hyanwong:
Failing on `[gw13] node down: Not properly terminated`-type errors again.

benjeffery:
> Failing on `[gw13] node down: Not properly terminated`-type errors again.

I'm trying to recreate those at #1280

python/tests/test_highlevel.py
hyanwong: Sorted this now. Hitting another issue :(

jeromekelleher:
> Annoyingly, I seem to hit this situation all the time: I try to address one issue, and end up having to fix 2 or 3 unrelated issues which arise as a result. But it takes a lot of time to unpick the interrelated changes into semi-independent PRs, because I've usually gone most of the way towards fixing the first issue before I find I need to change the other stuff. Not sure there's an easy solution, but I'm always hitting this.

You didn't *have* to make all these unrelated changes here - fix one thing at a time I think, and these sorts of problems will occur less frequently.

PR #1282, opened 2021-03-30
Test some wacky tree seqs

hyanwong:
## Description

Split off from #1279 as [requested](https://github.com/tskit-dev/tskit/pull/1279#issuecomment-810115211), and on which this is based (so merge that one first). As well as a completely empty tree sequence, I've also added in a tree sequence with no samples, which triggers some wacky behaviour because a tree with no roots does not list any nodes in `tree.nodes()`, even though the edge diffs can report topology in the tree.

At some point we want to take all these examples and pytest-parameterize them, but until then it's useful to have them all listed in the `get_example_tree_sequences()` function

Also, I think our description of max_node_time was wrong: it implied that we only return the time of the oldest sample if there are no edges in the tree, but we also (correctly, IMO) return the time of the oldest sample if there are edges in the ts but they attach to parents that are younger than the oldest sample.

Contains the fixes to max_root_time and newick testing requested by @benjeffery 

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -229,8 +229,12 @@ def get_example_tree_sequences(back_mutations=True, gaps=True, internal_samples=
     yield ts
     yield tsutil.add_random_metadata(ts)
     tables = ts.dump_tables()
+    tables.nodes.flags = np.zeros_like(tables.nodes.flags)
+    yield tables.tree_sequence()  # no samples
+    tables = ts.dump_tables()
     tables.edges.clear()
-    yield tables.tree_sequence()  # empty tree sequence
+    yield tables.tree_sequence()  # empty tree
+    yield tskit.TableCollection(sequence_length=1).tree_sequence()  # empty tree seq
 
 
 def get_bottleneck_examples():
@@ -470,11 +474,13 @@ class HighLevelTestCase:
             assert st1.left_root == st2.left_root
             assert sorted(list(roots)) == sorted(st1.roots)
             assert st1.roots == st2.roots
-            if len(roots) > 1:
+            if len(roots) == 0:
+                assert st1.root == tskit.NULL
+            elif len(roots) == 1:
+                assert st1.root == list(roots)[0]
+            else:
                 with pytest.raises(ValueError):
                     st1.root
-            else:
-                assert st1.root == list(roots)[0]
             assert st2 == st1
             assert not (st2 != st1)
             left, right = st1.get_interval()
@@ -597,6 +603,10 @@ class TestTreeSequence(HighLevelTestCase):
 
     def verify_pairwise_diversity(self, ts):
         haplotypes = ts.genotype_matrix(isolated_as_missing=False).T
+        if ts.num_samples == 0:
+            with pytest.raises(ValueError, match="at least one element"):
+                ts.get_pairwise_diversity()
+            return
         pi1 = ts.get_pairwise_diversity()
         pi2 = simple_get_pairwise_diversity(haplotypes)
         assert pi1 == pytest.approx(pi2)
@@ -882,7 +892,7 @@ class TestTreeSequence(HighLevelTestCase):
                 assert t1.parent_dict == t2.parent_dict
                 assert t1.index == 0
                 if "tracked_samples" in kwargs:
-                    assert t1.num_tracked_samples() != 0
+                    assert t1.num_tracked_samples() == ts.num_samples
                 else:
                     assert t1.num_tracked_samples() == 0
 
@@ -892,7 +902,7 @@ class TestTreeSequence(HighLevelTestCase):
                 assert t1.parent_dict == t2.parent_dict
                 assert t1.index == ts.num_trees - 1
                 if "tracked_samples" in kwargs:
-                    assert t1.num_tracked_samples() != 0
+                    assert t1.num_tracked_samples() == ts.num_samples
                 else:
                     assert t1.num_tracked_samples() == 0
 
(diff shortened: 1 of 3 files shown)


python/tests/test_highlevel.py
```diff
@@ -916,9 +926,11 @@ def test_trees_interface(self):
...
         for ts in get_example_tree_sequences():
-            with pytest.raises(ValueError):
+            with pytest.raises(ValueError, match="at least one element"):
                 ts.get_pairwise_diversity([])
             samples = list(ts.samples())
+            if len(samples) == 0:
+                continue
             assert ts.get_pairwise_diversity() == ts.get_pairwise_diversity(samples)
             assert ts.get_pairwise_diversity(samples[:2]) == ts.get_pairwise_diversity(
                 list(reversed(samples[:2]))
```
benjeffery, line 936:
```suggestion
            if len(samples) == 0:
                with pytest.raises(
                    ValueError, match="Sample sets must contain at least one element"
                ):
                    ts.get_pairwise_diversity()
            else:
                assert ts.get_pairwise_diversity() == ts.get_pairwise_diversity(samples)
                assert ts.get_pairwise_diversity(
                    samples[:2]
                ) == ts.get_pairwise_diversity(list(reversed(samples[:2])))

```

benjeffery (approved):
Thanks @hyanwong one small suggestion.

python/tests/test_highlevel.py
hyanwong: Good idea

python/tests/test_highlevel.py
```diff
@@ -1591,7 +1619,12 @@ def test_tree_node_edges(self):
             for mapping, tree in zip(ts._tree_node_edges(), ts.trees()):
                 node_mapped = mapping >= 0
                 edge_visited[mapping[node_mapped]] = True
-                assert np.sum(node_mapped) == len(list(tree.nodes())) - tree.num_roots
+                # Paradoxically, tree.nodes() does not necessarily list all the nodes
```
jeromekelleher, line 1622: I don't think it's a paradox, it naturally follows from the fact that it lists nodes in traversal order. The documentation isn't clear about this all right though; I've opened #1285

python/tests/test_highlevel.py
```diff
@@ -2213,12 +2266,21 @@ def test_newick_topology_equiv(self):
...
-                newick2 = tree.newick(node_labels={}).translate(replace_numeric)
-                assert newick1 == newick2 == plain_newick
+                if not tree.has_single_root:
+                    with pytest.raises(ValueError) as plain_newick_err:
+                        tree.newick(node_labels={}, include_branch_lengths=False)
+                    with pytest.raises(ValueError) as newick1_err:
+                        tree.newick().translate(replace_numeric)
+                    with pytest.raises(ValueError) as newick2_err:
+                        tree.newick(node_labels={}).translate(replace_numeric)
+                    assert str(newick1_err) == str(newick2_err) == str(plain_newick_err)
```
jeromekelleher, line 2276: Putting multiple asserts on one line is counter productive - you don't know which one has failed, if it fails (or at least it's not as easy to see)

python/tests/test_highlevel.py
```diff
@@ -2213,12 +2266,21 @@ def test_newick_topology_equiv(self):
...
-                    continue
-                plain_newick = tree.newick(node_labels={}, include_branch_lengths=False)
-                newick1 = tree.newick().translate(replace_numeric)
-                newick2 = tree.newick(node_labels={}).translate(replace_numeric)
-                assert newick1 == newick2 == plain_newick
+                if not tree.has_single_root:
+                    with pytest.raises(ValueError) as plain_newick_err:
+                        tree.newick(node_labels={}, include_branch_lengths=False)
+                    with pytest.raises(ValueError) as newick1_err:
+                        tree.newick().translate(replace_numeric)
```
jeromekelleher, line 2273: What's the point of the ``.translate`` here? This won't get called, right because newick() raises a ValueError?

python/tests/test_highlevel.py
```diff
@@ -2213,12 +2266,21 @@ def test_newick_topology_equiv(self):
...
+                    with pytest.raises(ValueError) as newick2_err:
+                        tree.newick(node_labels={}).translate(replace_numeric)
+                    assert str(newick1_err) == str(newick2_err) == str(plain_newick_err)
+                else:
+                    plain_newick = tree.newick(
+                        node_labels={}, include_branch_lengths=False
+                    )
+                    newick1 = tree.newick().translate(replace_numeric)
+                    newick2 = tree.newick(node_labels={}).translate(replace_numeric)
+                    assert newick1 == newick2 == plain_newick
```
jeromekelleher, line 2283: Likewise here with the multiple asserts

python/tskit/trees.py
```diff
@@ -3823,19 +3848,24 @@ def num_migrations(self):
...
+        Returns the time of the oldest root in any of the trees in this tree sequence.
         This is usually equal to ``np.max(ts.tables.nodes.time)`` but may not be
-        since there can be nodes that are not present in any tree. Consistent
-        with the definition of tree roots, if there are no edges in the tree
-        sequence we return the time of the oldest sample.
+        since there can be non-sample nodes that are not present in any tree. Note that
+        isolated samples are also defined as roots (so there can be a max_root_time
+        even in a tree sequence with no edges).
+
+        ..note::
```
jeromekelleher, line 3857: Missing a space here, ``.. note::``

python/tskit/trees.py
```diff
@@ -3823,19 +3848,24 @@ def num_migrations(self):
...
+            Since roots are defined by the ends of the upward paths from the set
+            of samples, a tree sequence with no samples has no roots; an error
+            will be raised in this case.
 
         :return: The maximum time of a root in this tree sequence.
         :rtype: float
         """
-        ret = max(self.node(u).time for u in self.samples())
+        ret = max(self.node(u).time for u in self.samples())  # Raises err if no samples
```
jeromekelleher, line 3865: Might as well just test this and raise the error explicitly:
```
if self.num_samples == 0:
    raise ValueError("max_root_time is not defined in a tree sequence with 0 samples")
```

python/tskit/trees.py
jeromekelleher: A ``raises ValueError`` would probably be the more systematic way of documenting this rather than a note block.

jeromekelleher (commented):
Looks good, thanks, minor comments. Needs a rebase.

python/tests/test_highlevel.py
hyanwong: Yeah, I commented there too. Happy to rephrase this comment

python/tests/test_highlevel.py
hyanwong: Well spotted. Copy-paste error. Thanks.

python/tskit/trees.py
hyanwong: Nice, good point

hyanwong:
All done

PR #1287, opened 2021-03-31
Add table extend method

### LINKED ISSUE #1271: tsk_X_table_extend method
This will be useful for the python fancy indexing of tables, but would likely be useful more widely:
Someting like:
`tsk_table_X_extend(self, other, num_rows, row_index)` where the rows `other[row_index]` are appended to `self` from `other`.

benjeffery:
Fixes #1271

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -644,6 +644,38 @@ out:
 }
 
 int
+tsk_individual_table_extend(tsk_individual_table_t *self,
+    const tsk_individual_table_t *other, tsk_size_t num_rows,
+    const tsk_id_t *row_indexes, tsk_flags_t TSK_UNUSED(options))
+{
+    int ret = 0;
+    tsk_size_t j;
+    tsk_individual_t individual;
+
+    /* We know how much to expand the non-ragged columns, so do it ahead of time */
+    ret = tsk_individual_table_expand_main_columns(self, num_rows);
+    if (ret != 0) {
+        goto out;
+    }
+    for (j = 0; j < num_rows; j++) {
+        ret = tsk_individual_table_get_row(
+            other, row_indexes == NULL ? (tsk_id_t) j : row_indexes[j], &individual);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_individual_table_add_row(self, individual.flags, individual.location,
+            individual.location_length, individual.parents, individual.parents_length,
+            individual.metadata, individual.metadata_length);
+        if (ret < 0) {
+            goto out;
+        }
+    }
+    ret = 0;
+out:
+    return ret;
+}
+
+int
 tsk_individual_table_free(tsk_individual_table_t *self)
 {
     tsk_safe_free(self->flags);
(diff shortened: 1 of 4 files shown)


c/tskit/tables.h
```diff
@@ -842,6 +842,30 @@ int tsk_individual_table_clear(tsk_individual_table_t *self);
 */
 int tsk_individual_table_truncate(tsk_individual_table_t *self, tsk_size_t num_rows);
 
+/**
+@brief Extends a table by appending rows copied from another table.
+
+@rst
+Rows from ``other``, specified by row index, are appended to the end of  ``self``.
```
jeromekelleher, line 849: How about 

> Appends the rows at the specified indexes from the table ``other`` to the end of this table. Row indexes can be repeated and in any order. If ``row_indexes`` is NULL,  append the first ``num_rows`` from ``other`` to this table.

c/tskit/tables.h
```diff
@@ -842,6 +842,30 @@ int tsk_individual_table_clear(tsk_individual_table_t *self);
...
+Row indexes can be repeated. If ``row_indexes`` is NULL all rows are copied.
+``other`` is not modified. Note that metadata is copied as-is and is not checked
+for compatibility with any existing schema on ``self``.
+@endrst
+
+@param self A pointer to a tsk_individual_table_t object where rows are to be added.
+@param other A pointer to a tsk_individual_table_t object where rows are copied from.
+@param num_row_indexes The number of entries to use from ``row_indexes``, ignored if
```
jeromekelleher, line 857: I was thinking more like: 
> @param num_rows The number of rows from other to append to this table.

This seems more flexible and useful: if we want to get the first 10 rows, say, we just pass in ``.., 10, NULL)``.

jeromekelleher (commented):
Looks good! I think we should tweak the semantics around ``num_rows`` though.

c/tskit/tables.h
benjeffery: Yes, great idea.

benjeffery:
OK, changes made, will crack on with the other tables.

benjeffery:
@jeromekelleher Should be good to go!

c/tskit/tables.c
```diff
@@ -643,6 +643,40 @@ tsk_individual_table_truncate(tsk_individual_table_t *self, tsk_size_t num_rows)
...
+
+    /* We know how much to expand the non-ragged columns, so do it ahead of time */
+    if (self->max_rows < self->num_rows + num_rows) {
+        ret = tsk_individual_table_expand_main_columns(self, num_rows);
+        if (ret != 0) {
+            goto out;
+        }
+    }
+    for (j = 0; j < num_rows; j++) {
+        ret = tsk_individual_table_get_row(
```
jeromekelleher, line 663: Ah, really nice how this automatically does all our range checking for us! So if row_indexes is NULL and ``num_rows`` is bigger than the table, we'll get an out-of-bounds, exactly as we should. Love it.

c/tskit/tables.c
```diff
@@ -643,6 +643,40 @@ tsk_individual_table_truncate(tsk_individual_table_t *self, tsk_size_t num_rows)
...
+    const tsk_individual_table_t *other, tsk_size_t num_rows,
+    const tsk_id_t *row_indexes, tsk_flags_t TSK_UNUSED(options))
+{
+    int ret = 0;
+    tsk_size_t j;
+    tsk_individual_t individual;
+
+    /* We know how much to expand the non-ragged columns, so do it ahead of time */
+    if (self->max_rows < self->num_rows + num_rows) {
+        ret = tsk_individual_table_expand_main_columns(self, num_rows);
```
jeromekelleher, line 657: Ah, sorry, should have spotted this the first time around. You can just call ``ret = tsk_individual_table_expand_main_columns(self, num_rows);`` without the first check. It does the check to see if there's space for the specified number of additional rows, and expands if necessary.

jeromekelleher (approved):
Really nice, thanks @benjeffery. One minor simplification we can make.

c/tskit/tables.c
benjeffery: There's a small perf cost, that could be avoided in the NULL case as we could check the range and then copy the columns as a `memcpy` (apart from the offsets). Lets do that if we see the need, otherwise just added complexity at this point.

jeromekelleher: yeah, we could special case this later if becomes an issue. For now, this is great.

c/tskit/tables.c
benjeffery: Of course! Fixed.

PR #1288, opened 2021-03-31
Don't draw empty flanks

hyanwong:
## Description

A draft for being able to specify only a certain genomic region to display on an SVG plot, using e.g. `x_lim=[1000, 2000]` to display only trees within the regions 1000-2000 base pairs. This is implemented by deleting the flanking intervals but keeping the coordinate system. As part of this, we want to make sure that we don't display the missing flanks in a tree sequence. This is of general benefit, as such sequences are likely to be generated e.g. by https://github.com/tskit-dev/msprime/pull/1599. Therefore this PR also changes the default behaviour of `ts.draw_svg()` so that if there are tree(s) to left or right that are completely empty (no edges and no mutations) they are not displayed. I think that's usually what will be wanted anyway.

As a nice addition, if a tree has been chopped off by specifying an x_lim, that tick won't be plotted, to imply that the tree continues to the left or right (and is simply not displayed). Here's the example in the tests:

Screenshot 2021-03-31 at 21 59 18

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1541,8 +1541,8 @@ class Tree:
         force_root_branch=None,
         symbol_size=None,
         x_axis=None,
-        y_axis=None,
         x_label=None,
+        y_axis=None,
         y_label=None,
         y_ticks=None,
         y_gridlines=None,
@@ -1698,11 +1698,11 @@ class Tree:
         :param bool x_axis: Should the plot have an X axis line, showing the start and
             end position of this tree along the genome. If ``None`` (default) do not
             plot an X axis.
+        :param str x_label: Place a label under the plot. If ``None`` (default) and
+            there is an X axis, create and place an appropriate label.
         :param bool y_axis: Should the plot have an Y axis line, showing time (or
             ranked node time if ``tree_height_scale="rank"``). If ``None`` (default)
             do not plot a Y axis.
-        :param str x_label: Place a label under the plot. If ``None`` (default) and
-            there is an X axis, create and place an appropriate label.
         :param str y_label: Place a label to the left of the plot. If ``None`` (default)
             and there is a Y axis,  create and place an appropriate label.
         :param list y_ticks: A list of Y values at which to plot tickmarks (``[]``
@@ -1735,8 +1735,8 @@ class Tree:
             force_root_branch=force_root_branch,
             symbol_size=symbol_size,
             x_axis=x_axis,
-            y_axis=y_axis,
             x_label=x_label,
+            y_axis=y_axis,
             y_label=y_label,
             y_ticks=y_ticks,
             y_gridlines=y_gridlines,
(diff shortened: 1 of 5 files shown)


jeromekelleher:
Looks good, thanks @hyanwong. I don't have time to review properly now, but I agree with the basic approach.

One thing to think about (which I'm sure you have) is how we can define an "empty" or "missing" tree, and what method we would put on the Tree class to easily identify this.

hyanwong:
Thanks @jeromekelleher - no hurry on this. I just wanted to have a concrete proposal down.

I've actually not thought about the general case of identifying an empty region. At the moment (and for this PR) I only do so for flanking regions, so I can simply look for `min(np.min(tables.edges.left), np.min(tables.sites.position))` on the LHS and `max(np.max(tables.edges.left), np.max(tables.sites.position))` on the RHS.

jeromekelleher:
Ah right, that's a good idea.

jeromekelleher:
I'm going to mute these threads @hyanwong - please ping me if you want input.

hyanwong:
Didn't realise these were pinging you, sorry @jeromekelleher .

If you think the approach is right, then it's probably ready for review: I'm done now. But there's no urgency at all for this - could happen many weeks away, as the logic is quite intricate.

hyanwong:
Now ready for review

python/tskit/trees.py
```diff
@@ -5445,6 +5455,17 @@ def draw_svg(
 
         :return: An SVG representation of a tree sequence.
         :rtype: str
+
```
benjeffery, line 5458: Do we need this note? I can't imagine this tripping anyone up and the docs for svg drawing are already a lot?

benjeffery (approved):
Looks good!
One question about the docs. 
I'm concerned that the drawing code is getting a lot of options and complexity, this one is a useful feature to add, but I wanted to flag that we need to consider before adding much more!

hyanwong:
> Looks good!
> One question about the docs.
> I'm concerned that the drawing code is getting a lot of options and complexity, this one is a useful feature to add, but I wanted to flag that we need to consider before adding much more!

I agree. Actually, I think some of the examples in the docs should go into the viz tutorial, and that should help slim stuff down a bit.

jeromekelleher (approved):
LGTM. I agree with Ben though, this code is complicated enough, let's not add any more to it for a while.

hyanwong (commented):
> LGTM. I agree with Ben though, this code is complicated enough, let's not add any more to it for a while.

Yep, I agree too. Time spent on viz would be better directed to the viz tutorial now, IMO.

python/tskit/trees.py
benjeffery: Yeah, when the viz tutorial is done you can do a reduction/interlinking pass through these docs.

benjeffery:
@hyanwong Need a changelog for this as it is a new documented API change.

hyanwong:
> @hyanwong Need a changelog for this as it is a new documented API change.

Thanks. Done.

PR #1289, opened 2021-03-31
Add the ability to set your own svg symbols

hyanwong:
This is a draft that allows symbols in an SVG plot to be swapped for any SVG drawing commands. It's perhaps a bit hacky, but @awohns might be able too use it to do some funky stuff, so I'm draft PRing it here.

His idea was to swap the symbols for pie charts showing the distribution of populations under a node, which thus allows us to plot trees with huge numbers of tips using summary pie charts at the tips. He can take this further for examples.

### DIFF
--- a/python/tests/data/svg/ts_alt_symbols.svg
+++ b/python/tests/data/svg/ts_alt_symbols.svg
@@ -1,0 +1,350 @@
+<?xml version="1.0" ?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xlink="http://www.w3.org/1999/xlink" baseProfile="full" height="200" version="1.1" width="1000">
+	<defs>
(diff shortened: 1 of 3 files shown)


jeromekelleher:
I'm all for this in principle, so long as it doesn't get too complicated. One high-level point though: we shouldn't have an API dependency on svgwrite (it's in maintenance mode, and we may need to drop it entirely at some point). Better to use raw SVG xml if necessary.

hyanwong:
> I'm all for this in principle, so long as it doesn't get too complicated. One high-level point though: we shouldn't have an API dependency on svgwrite (it's in maintenance mode, and we may need to drop it entirely at some point). Better to use raw SVG xml if necessary.

It's a good point. I tried to do this, but the problem is that (as far as I can see) svgwrite can't inject raw SVG xml into the SVGs it produces. So we are sort of stuck.

jeromekelleher:
Hmm, I see. Well, we'll need to look at the options then.

hyanwong:
https://github.com/cduck/drawSvg looks similar and more active, but less well documented and less mature I think :shrug:

benjeffery:
drawSVG would add `cairo` as a dependancy, which is pretty heavy.

hyanwong:
Isn't that only needed if you output PNG (i.e. is an optional dependency)? I didn't look too hard though.

benjeffery:
> Isn't that only needed if you output PNG (i.e. is an optional dependency)? I didn't look too hard though.

You're correct https://github.com/cduck/drawSvg/blob/master/drawSvg/raster.py#L10

jeromekelleher:
I've been wondering how much value we're actually getting out of svgwrite - given that it's just writing a bit of XML and we're only using a small fraction of its API, it may be simpler in the long run to hand-roll the SVG.

hyanwong:
FWIW there's a trivial example of a home-rolled svg creator class at https://www.codedrome.com/svg-library-in-python/ . Apart from adding elements to a nested set of dicts rather than building up a single string, so we can add them out of order, I suspect we wouldn't need much more than is implemented in that example. Or we could use some XML library.

jeromekelleher:
Yes, that does look quite straightforward. I'm happy to consider a PR where we replace svgwrite with our own implementation of the subset of its API that we use. We could evolve our SVG implementation from there then.

PR #1292, opened 2021-04-01
Add RankTree.leaves()

hyanwong:
## Description

Adds an iterator over the leaf labels of a RankTree. After discussion with @daniel-goldstein we simply called this `.leaves()` in analogy with the iterator `Tree.leaves()`. This also internally documents some properties of the order returned by both `.leaves()` and `.newick()`, and tests they are true up to topologies of size 5 or 7.

# PR Checklist:

### DIFF
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -481,6 +481,31 @@ class TestRankTree:
         six_leaf_sym = RankTree(children=[three_leaf_asym, three_leaf_asym])
         assert six_leaf_sym.is_symmetrical()
 
+    @pytest.mark.parametrize("n", range(7))
+    def test_leaves(self, n):
+        for tree in RankTree.all_unlabelled_trees(n):
+            first_labelled_tree = RankTree.unrank(n, (tree.shape_rank(), 0))
+            assert list(first_labelled_tree.leaves()) == list(range(n))
+
+    @pytest.mark.parametrize("n", range(5))
+    def test_no_leaves(self, n):
+        for tree in RankTree.all_unlabelled_trees(n):
+            assert list(tree.leaves()) == [None] * n
+
+    @pytest.mark.parametrize("n", range(6))
+    def test_newick_order(self, n):
+        del_digits = {ord(c): "" for c in "1234567890"}
+        del_other = {ord(c): "" for c in "(,);"}
+        for unlabelled_tree in RankTree.all_unlabelled_trees(n):
+            first_labelled_tree = RankTree.unrank(n, (unlabelled_tree.shape_rank(), 0))
+            newick_braces = first_labelled_tree.newick().translate(del_digits)
+            for tree in RankTree.all_labellings(unlabelled_tree):
+                nwk = tree.newick()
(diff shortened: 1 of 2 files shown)


python/tests/test_combinatorics.py
```diff
@@ -481,6 +481,31 @@ def test_is_symmetrical(self):
         six_leaf_sym = RankTree(children=[three_leaf_asym, three_leaf_asym])
         assert six_leaf_sym.is_symmetrical()
 
+    @pytest.mark.parametrize("n", range(7))
+    def test_leaves(self, n):
+        for tree in RankTree.all_unlabelled_trees(n):
+            first_labelled_tree = RankTree.unrank(n, (tree.shape_rank(), 0))
+            assert list(first_labelled_tree.leaves()) == list(range(n))
```
jeromekelleher, line 488: This test would pass if the leaves() iterator was implemented as ``return range(n)``. I don't think there's much value in iterating over the many, many trees here, it's just wasting electricity.

A much better test would be a few small hand rolled examples. Confirming we get the right result on a handful of chosen examples would be much better here.

python/tests/test_combinatorics.py
```diff
@@ -481,6 +481,31 @@ def test_is_symmetrical(self):
...
+
+    @pytest.mark.parametrize("n", range(5))
+    def test_no_leaves(self, n):
+        for tree in RankTree.all_unlabelled_trees(n):
```
jeromekelleher, line 492: This is excessive - testing this on one or two examples would be the same as iterating over all trees here.

python/tests/test_combinatorics.py
```diff
@@ -481,6 +481,31 @@ def test_is_symmetrical(self):
...
+            assert list(tree.leaves()) == [None] * n
+
+    @pytest.mark.parametrize("n", range(6))
+    def test_newick_order(self, n):
+        del_digits = {ord(c): "" for c in "1234567890"}
```
jeromekelleher, line 497: Also seems excessive to do this for so many trees, since it's a fairly simple property which if it's true for a handful of examples will be true for everything.

jeromekelleher (commented):
LGTM. The testing is very brute force though, and doesn't seem like a good use of resources.

benjeffery:
@hyanwong Do you want this PR in the next release? Seems it just needs some small tweaks?

hyanwong:
> @hyanwong Do you want this PR in the next release? Seems it just needs some small tweaks?

It's not urgent, as I found a different way to do what I needed in https://github.com/tskit-dev/tskit/pull/1296. But I guess it would be useful, so worth leaving open.

jeromekelleher:
I think we should either bring this up to date and merge or close it @hyanwong - there's no point in leaving old PRs gathering dust. It'll just be more work later.

benjeffery:
Yes, I'd rather have the PR queue be things that are on a path to being merged. We can always label with future and re-open if you don't have time to bring this upto date now.

hyanwong:
Fine - let's relabel with future and close. I think #1294 is more worthy of attention in this area at the moment.

PR #1296, opened 2021-04-02
Add dict_of_dicts to tree sequence

### LINKED ISSUE #1294: TreeSequence.as_dict_of_dicts() method
We already have a `Tree.as_dict_of_dicts()` method, used to read a tree into NetworkX.

It would be useful to read an entire tree sequence into NetworkX as a DAG, so I would like to implement `TreeSequence.as_dict_of_dicts()`. I imagine it  being just like the other method but as well as specifying the `'branch_length'` attribute it would also specify the `'left'` and `'right'` attributes for an edge.

Should be easy, I would have thought...

hyanwong:
## Description

Fixes #1294. Should probably add some tests for attributes of the graph like `branch_length`, `left`, and `right`.

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -9,6 +9,9 @@
 
 **Features**
 
+- New `TreeSequence.as_dict_of_dicts` method for reading an entire tree sequence into
+  a NetworkX graph (:user:`hyanwong`, :issue:`1294`, :pr:`1296`).
+
 - Add `Table.append` method for adding rows from classes such as `SiteTableRow` and
   `Site` (:user:`benjeffery`, :issue:`1111`, :pr:`1254`).
 
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1633,6 +1633,53 @@ class TestTreeSequence(HighLevelTestCase):
                         assert edge.right >= tree.interval.right
             assert np.all(edge_visited)
 
+    def test_as_dict_of_dicts(self):
+        for ts in get_example_tree_sequences():
+            ts = ts.simplify(keep_unary=True)  # remove unreferenced nodes
+            adj_dod = ts.as_dict_of_dicts()
+            g = nx.from_dict_of_dicts(
+                adj_dod, create_using=nx.MultiDiGraph, multigraph_input=True
+            )
+            self.verify_nx_graph_topology(ts, g)
+
+    def verify_nx_graph_topology(self, ts, g):
+        assert nx.is_directed_acyclic_graph(g)
+        assert g.number_of_edges() == ts.num_edges
+        # isolated nodes may not be in the graph, so add sample nodes to the graph set
+        assert set(g.nodes) | set(ts.samples()) == set(range(ts.num_nodes))
+        # all non root nodes
+        assert set(ts.tables.edges.child) == {n for n in g.nodes if g.in_degree(n) != 0}
+        # all non leaf nodes
+        assert set(ts.tables.edges.parent) == {
+            n for n in g.nodes if g.out_degree(n) != 0
+        }
+
+    def test_nx_tree_ts_equiv(self):
+        ts = msprime.simulate(10, recombination_rate=1, random_seed=123)
+        adj_dod = ts.as_dict_of_dicts()
+        g_first = nx.from_dict_of_dicts(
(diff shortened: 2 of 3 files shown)


jeromekelleher:
Very neat! Looks very nice, but I think we want to be sure that this is the "right" graph representation before we make it part of the public API. Some questions:

1. What do these graphs look like when we plot them? I.e., what does a single tree ts look like, and one with two trees?
2. What would be need to do to round trip the topology? Ideally we would want 

```python
d = ts.as_dict_of_dicts()
tsp = tskit.TreeSequence.from_dict_of_dicts(d)
assert tsp.equals(ts, ignore_provenance=True)
```

We would need to have some visibilty on whether this is possible before we can fix on the graph representation.

benjeffery:
Nice! In the round trip example @jeromekelleher, do you mean that just the topology would round trip? Or other things like populations and migrations? (I guess these could be included as an attribute at the root dict?)

jeromekelleher:
Just the topology initially, we don't have to worry about the other stuff like individuals, migrations etc. But yes, in the case of populations it'd probably be simplest to add them in as attrs in the root dict.

hyanwong:
AFAIK when NetworkX imports from dict of dicts you can only set attributes on edges: I don't know how to set them on nodes. So round tripping might be difficult. Even setting node times would have to be imperfectly guessed from edge lengths.

benjeffery:
> AFAIK when NetworkX imports from dict of dicts you can only set attributes on edges: I don't know how to set them on nodes. So round tripping might be difficult. Even setting node times would have to be imperfectly guessed from edge lengths.

We would round-trip the dicts, not the NetworkX representation.

jeromekelleher:
Yes - I'm afraid we have to do the due diligence here @hyanwong. We need to spend time exploring the properties of this representation if we're doing to make it part of the API. The first question we need to answer is "is this a bijection"?

hyanwong:
I assumed this would simply be an analogue to Tree.as_dict_of_dicts, so purely a (non complete) export format for NetworkX. If you want it to be something more, then it gets complex, I agree.

hyanwong:
We don't have a Tree.from_dod method, so I'm not sure we need (want?) a TS one?

hyanwong:
Apologies for the terse replies: I'm on a mobile. There's no reason to merge this any time soon. As it is, I need to use a slightly bespoke version anyway, which randomises the node IDs (otherwise locating isomorphic nodes is biased towards the input order). I just imagine that if *I* find a function like this useful for analysing the tree sequence as a graph, others might too. So it's probably worth having a similar function in the API. But as you say, worth doing the due diligence first. I'm in no hurry either way.

jeromekelleher:
OK - since we're making quite a big decision here (what's the graph representation of a tree sequence) @hyanwong, I think it would be better if you used the version you have for your own work, and we took our time with this. The Tree version is much simpler and we don't worry about round tripping it because there's no way of creating a Tree on it's own anyway. The TreeSequence is the core data structure.

So, I think we can mark this as a draft?

hyanwong:
> So, I think we can mark this as a draft?

Done.

FWIW I don't think `as_dict_of_dict` should be the "canonical" graph representation of a TS. I think this only needs to be a function used for reading a specific format into NetworkX. We shouldn't claim anything grander about it. If we want a canonical format we could just use `.__dict__()` (or something) instead.

jeromekelleher:
Could we update this method to return the representation I'm advocating for here @hyanwong? https://github.com/tskit-dev/tskit/discussions/2068#discussioncomment-1821721

I think it's better to annotate the nodes with their time, and the edges with intervals, rather than making a multigraph with branch lengths like we have here.

hyanwong:
> Could we update this method to return the representation I'm advocating for here @hyanwong? [#2068 (reply in thread)](https://github.com/tskit-dev/tskit/discussions/2068#discussioncomment-1821721)

I haven't figured out how to use the "dict_of_dicts" version to allocate attributes (like time) to nodes. So maybe this isn't the right way anyway?

> I think it's better to annotate the nodes with their time, and the edges with intervals, rather than making a multigraph with branch lengths like we have here. Making the multigraph version for viz should be simple enough as a post-processing step.

Hmm, is that true. Isn't it just as easy to go from multigraph -> graph as the other way round?

benjeffery:
Tidying up old PRs - please re-open if you wish.

PR #1300, opened 2021-04-07
Improve string and html output

### LINKED ISSUE #1270: __str__ for tables could have better formatting
Also limit num rows to 40ish.

benjeffery:
Fixes #1270 by making `__str__` for tables output unicode tables. 

Introduces `tskit.MAX_LINES` which controls the number of lines to truncate the table to, in this way a user can see the whole of a table by setting it to `None`

Also displays decoded metadata instead of base64 encoding it, which also applies to HTML output. Metadata is truncated, to try to keep tables ~80 chars width. This is not controllable. 

OLD:
Screenshot from 2021-04-07 23-56-32

NEW:
Screenshot from 2021-04-07 23-51-31

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -540,7 +540,15 @@ class CommonTestsMixin:
             table = self.table_class()
             table.set_columns(**input_data)
             s = str(table)
-            assert len(s.splitlines()) == num_rows + 1
+            assert len(s.splitlines()) == num_rows + 4
+        input_data = self.make_input_data(41)
+        table = self.table_class()
+        table.set_columns(**input_data)
+        assert "1 rows skipped" in str(table)
+        tskit.set_print_options(max_lines=None)
+        assert "1 rows skipped" not in str(table)
+        tskit.set_print_options(max_lines=40)
+        tskit.MAX_LINES = 40
 
     def test_repr_html(self):
         for num_rows in [0, 10, 40, 50]:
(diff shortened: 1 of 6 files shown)


python/tests/test_util.py
```diff
@@ -434,7 +445,25 @@ def test_unicode_table():
...
+"""
+    )
+
+    assert (
+        util.unicode_table(
+            [["1", "2", "3", "4"], ["5", "6", "7", "8"], ["90", "10", "11", "12"]],
+            title="TITLE",
+            row_separator=False,
+        )
+        == """╔═══════════╗
```
jeromekelleher, line 462: Nice. Might as well use the ``textwrap.dedent()`` trick here for new code so that we can keep the file indentation looking good.

jeromekelleher (commented):
LGTM, very nice!

I'm not sure about tskit.MAX_ROWs, though. ALL_CAPS to me says "constant". What if we followed the say numpy does it and add a ``set_print_options`` function or something?

python/tests/test_util.py
benjeffery: Ah yeah, forgot that trick. Done

benjeffery:
I've added `tskit.set_print_options`.

python/tests/test_tables.py
```diff
@@ -540,7 +540,16 @@ def test_str(self):
             table = self.table_class()
             table.set_columns(**input_data)
             s = str(table)
-            assert len(s.splitlines()) == num_rows + 1
+            assert len(s.splitlines()) == num_rows + 4
+        input_data = self.make_input_data(41)
+        table = self.table_class()
+        table.set_columns(**input_data)
+        print(table)
```
jeromekelleher, line 547: stray print

jeromekelleher (commented):
Do you think we could pick this one up and get it merged please @benjeffery? I'd like to use it in a presentation

jeromekelleher (approved):

python/tests/test_tables.py
benjeffery: Opps, fixed.

PR #1301, opened 2021-04-08
Remove references to generations

### LINKED ISSUE #1295: Remove references to "generations" in docs
Probably because of transfer from msprime, the tskit docs refer to time "in generations", but this doesn't need to be the case for a tree sequence in general. In fact, where there are overlapping generations, it is only really meaningful to specify time in some calendar-type unit. So we should probably remove mention of "in generations" from the tskit docs (but not the msprime docs). Here's an example:

```
    def branch_length(self, u):
        """
        Returns the length of the branch (in generations) joining the
        specified node to its parent. This is equivalent to

        >>> tree.time(tree.parent(u)) - tree.time(u)
```

benjeffery:
Fixes #1295

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -918,7 +918,7 @@ class Tree:
 
     def branch_length(self, u):
         """
-        Returns the length of the branch (in generations) joining the
+        Returns the length of the branch (in units of time) joining the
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM

PR #1304, opened 2021-04-08
Improve Tree str and html

### LINKED ISSUE #1269: HTML repr for Tree
Similar to tree sequence. __str__ should also give same info.

benjeffery:
Fixes #1269 

Examples:
Screenshot from 2021-04-08 17-07-04
Screenshot from 2021-04-08 17-06-45

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -37,9 +37,11 @@ import platform
 import random
 import re
 import tempfile
+import textwrap
 import unittest
 import uuid as _uuid
 import warnings
+from xml.etree import ElementTree
 
 import kastore
 import msprime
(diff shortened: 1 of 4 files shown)


benjeffery:
The original issue suggested average arity - but I couldn't see a quick way to get number of edges in the tree to do edges/nodes?

python/tests/test_highlevel.py
```diff
@@ -2145,10 +2146,38 @@ def test_complex_mutations(self):
...
+            ║Total Branch Length│10.427971║
+            ╚═══════════════════╧═════════╝
+        """
+            )[1:]
+        )
+
+    def test_html_repr(self, ts_fixture):
+        html = ts_fixture.first()._repr_html_()
+        assert len(html) > 1900
+        assert "<tr><td>Total Branch Length</td><td>10.427971</td></tr>" in html
```
jeromekelleher, line 2180: Be nice to test if this is valid XML  as well, just pushing it through one of the parsers will be fine.

jeromekelleher (approved):
LGTM, one small suggestion for improving tests.

We can add other stuff  to the table over time if we want, but this is a great start.

python/tests/test_highlevel.py
benjeffery: Done

PR #1312, opened 2021-04-15
Update CI to using msprime 1 release.

jeromekelleher:


jeromekelleher:
Hmmm - I'm starting to feel like these example SVGs are more trouble that they are worth. @benjeffery, what do you think?

hyanwong:
> Hmmm - I'm starting to feel like these example SVGs are more trouble that they are worth. @benjeffery, what do you think?

Strange. It was working before. So I think this has flagged up something weird here? From a brief inspection, the trees in this example are now actually different, as are the mutations. So something odd has changed. Here's the first two trees in the old version:

Screenshot 2021-04-19 at 11 34 48

And here's the new one:

Screenshot 2021-04-19 at 11 35 22

hyanwong:
> > Hmmm - I'm starting to feel like these example SVGs are more trouble that they are worth. @benjeffery, what do you think?
> 
> Strange. It was working before. So I think this has flagged up something weird here? 

Confirmed that this is a difference between msprime 1.0.0a6 and 1.0.0, which (AFAI can see) results in different trees & mutations in this test case

benjeffery:
> Hmmm - I'm starting to feel like these example SVGs are more trouble that they are worth. @benjeffery, what do you think?

I think it feel like that as msprime has been in flux. If they trigger on a lot of msprime releases then yes, we'll change to less specific checks.

benjeffery:
@hyanwong I can't get this to fail locally,  even with msprime `1.0.0` which makes no sense unless the conda version of msprime is generating different output ?!

benjeffery:
One more thought on the svg fixtures are that they also clutter up the diffs. If we think changing to more semantic checks isn't too much work we should do it.

hyanwong:
> @hyanwong I can't get this to fail locally,  even with msprime `1.0.0` which makes no sense unless the conda version of msprime is generating different output ?!

I get a different output on my Mac laptop when simply switching between 1.0.0a6 and 1.0.0 using pip install / uninstall. Did you make sure to force uninstall first? Sometimes that seems to make a difference. 

Do you want me to commit the new SVG? Or perhaps better, use a different non binary example, perhaps e.g. with the WF simulator?

hyanwong:
> One more thought on the svg fixtures are that they also clutter up the diffs. If we think changing to more semantic checks isn't too much work we should do it.

True, but it's really useful to see the visual effect of formatting. Perhaps these examples could be flagged somehow as not normally run? Or we could rely on the tskit viz tutorial? Not sure if that helps, though.

benjeffery:
> Do you want me to commit the new SVG? Or perhaps better, use a different non binary example, perhaps e.g. with the WF simulator?

Yes please, I've nuked my env and reinstalled and still can't get this to fail. wf sim examples would be great.

jeromekelleher:
Depending on specific msprime outputs is bad I think, we can have changes even with differences in the underlying GSL version which would be non-obvious to track down.

hyanwong:
> Depending on specific msprime outputs is bad I think, we can have changes even with differences in the underlying GSL version which would be non-obvious to track down.

Presumably it's OK if we use the built-in WF simulator instead, then? Or we could save a set of example tree sequences somewhere (generated by a fixed version of msprime), and use those in the tskit tree sequences (esp when we fixture-ize them)?

hyanwong:
Should be fixed by #1318 but note that this still uses msprime for generating the deep history of the trees. It might be useful to have deterministic code that simply joins all the roots of all trees in a ts together in a single grand MRCA. I don't think there's a function that does this already, is there? Would it be generally useful?

benjeffery:
@mergifyio rebase

benjeffery:
@hyanwong I've rebased to include the wf sim PR and this is still failing, and I still can't get it to fail locally. I'm going to mark the offending test(s) as skipped so we can move on for now, but I think we'll need to remove the byte-for-byte tests unless they are from trees that are made by the test suite alone.

hyanwong:
> @hyanwong I've rebased to include the wf sim PR and this is still failing, and I still can't get it to fail locally. I'm going to mark the offending test(s) as skipped so we can move on for now, but I think we'll need to remove the byte-for-byte tests unless they are from trees that are made by the test suite alone.

Sounds sensible. As above, for this purpose, do you think we should add a routine (e.g. to tsutil.py) to make all trees have a single root? Then we could make some slightly more conventional tree sequences with the built-in WF generator.

benjeffery:
> > @hyanwong I've rebased to include the wf sim PR and this is still failing, and I still can't get it to fail locally. I'm going to mark the offending test(s) as skipped so we can move on for now, but I think we'll need to remove the byte-for-byte tests unless they are from trees that are made by the test suite alone.
> 
> Sounds sensible. As above, for this purpose, do you think we should add a routine (e.g. to tsutil.py) to make all trees have a single root? Then we could make some slightly more conventional tree sequences with the built-in WF generator.

Only if you are very motivated and it's straightforward. Although it is nice to know if any accidental change is made to the SVG, I'm not sure the cost/benefit of these tests is that high.

benjeffery:
Now getting failing tests not related to drawing such as `tests/test_topology.py::TestKeepIntervals::test_wright_fisher_simplified` that are also passing locally. Something fishy going on, maybe conda related.

jeromekelleher:
I think these tests are well past the threshold of being more trouble than they are worth. I think we should remove them.

benjeffery:
The failures now are due to to no mutations being placed by `msprime.mutate` but only on OSX. I've tweaked the rate, lets see if that passes.

hyanwong:
> The failures now are due to to no mutations being placed by `msprime.mutate` but only on OSX. I've tweaked the rate, lets see if that passes.

There's a few built-in mutation generators aren't there? Shouldn't we switch to using them in tests anyway, to reduce reliance on msprime?

benjeffery:
Changing the seed to mutate has worked to fix OSX. Will back out the debug changes and merge.

benjeffery (approved):

jeromekelleher:
A quirk I'm seeing elsewhere @benjeffery is that msprime simulations run on OSX and linux using the msprime wheels are generating different results. The most likely explanation is that they are building in different versions of GSL - possibly something for us to look into, in the interests of sanity...

PR #1318, opened 2021-04-19
Use the internal WF simulator for nonbinary tree ex

hyanwong:
## Description

Remove dependency on exact msprime output - should solve problems in #1312


benjeffery (approved):
Thanks @hyanwong

Issue #1299, opened 2021-04-06
Add numpy arrays to Python Tree object


jeromekelleher:
It will be useful for some applications to have direct access to the underling ``parent``, ``left_child`` etc arrays in the quintuply linked tree representation. We can do this by adding some (read only) numpy arrays that use these pointers defined in the C structs, i.e.,

```c
typedef struct {
    PyObject_HEAD
    TreeSequence *tree_sequence;
    tsk_tree_t *tree;
    PyObject *left_array;
    PyObject *right_array; // etc
} Tree;
```

Then in ``Tree_init`` we can create the required numpy views, and make them available to Python following the steps [here](https://docs.python.org/3/extending/newtypes_tutorial.html#adding-data-and-methods-to-the-basic-example)

jeromekelleher:
See [tutorial suggestion](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_parsimony.py#L77) for an idea of how we might test out how useful this is.

jeromekelleher:
After a fun afternoon playing with numba, this has gone a long way up my priority list! I'd kinda like to get it in for 0.3.6 @benjeffery and I'm happy to do the coding (I'm confident it's straighforward, and it'd be a nice break from other stuff). What's the timeline for 0.3.6?

benjeffery:
Awesome stuff, numba is excellent. Looking at the 0.3.6 milestone I would hope to have it complete within 2 weeks, hopefully before. Happy to add this issue to that milestone.

PR #1320, opened 2021-04-19
Tree array access.

jeromekelleher:
WIP for #1299

Trickier than I thought because of the circular references between the array and the Tree object, which owns the memory.

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -1,3 +1,11 @@
+.. |tree_array_warning| replace:: This is a high-performance interface which
+    provides zero-copy access to memory used in the C library.
+    As a consequence, the values stored in this array will change as
+    the Tree state is modified as we move along the tree sequence. See the
+    :class:`.Tree` documentation for more details. Therefore, if you want to
+    compare arrays representing different trees along the sequence, you must
+    take **copies** of the arrays.
+
 .. currentmodule:: tskit
 .. _sec_python_api:
 
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2726,6 +2726,41 @@ class TestTree(HighLevelTestCase):
                 breakpoints[i + 1] - breakpoints[i]
             )
 
+    def verify_tree_arrays(self, tree):
+        ts = tree.tree_sequence
+        assert tree.parent_array.shape == (ts.num_nodes,)
+        assert tree.left_child_array.shape == (ts.num_nodes,)
+        assert tree.right_child_array.shape == (ts.num_nodes,)
+        assert tree.left_sib_array.shape == (ts.num_nodes,)
+        assert tree.right_sib_array.shape == (ts.num_nodes,)
+        for u in range(ts.num_nodes):
+            assert tree.parent(u) == tree.parent_array[u]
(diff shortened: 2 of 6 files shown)


jeromekelleher:
Looks like this is working well. I think it's probably a good idea to add the ``flags_array`` and ``time_array`` to the tree class also, since these are things we'll often be using in low-level algorithms, and it's a faff (and currently copy-ful, semantics-wise) to do this via the tree sequence tables arrays.

Should we go ahead and merge this much, and try it out for a bit before documenting, or go ahead and document it now and be done with it? I doubt we can change much here, it's pretty much a done-deal I think.

@benjeffery, any thoughts?

benjeffery (approved):
This is awesome stuff, couple of small comments.

I wonder about the `_array` naming though. Would `left_children`, `parents` etc. fit more with our existing names as we don't put type information in names normally?

benjeffery:
> Should we go ahead and merge this much, and try it out for a bit before documenting, or go ahead and document it now and be done with it? I doubt we can change much here, it's pretty much a done-deal I think.

> @benjeffery, any thoughts?

Lets merge this, then file issues under 0.3.6 for `time_array` and `flags_array` and the docs. I'm happy to pick those up if I get to them before you.

jeromekelleher:
> I wonder about the _array naming though. Would left_children, parents etc. fit more with our existing names as we don't put type information in names normally?

It's a bit awkward, I agree, but I'm not sure there's much choice. If we were starting again I'd get rid of the ``parent(u)`` etc functions and use these arrays instead. But, that ship has sailed, so ``parent_array`` seems like the least bad option. I don't want to use ``parents`` because this will be break with the convention that array names are singular. Also, I think it would be confusing if we had ``tree.left_children[u]`` and ``tree.left_child(u)`` giving the same thing.

jeromekelleher:
(Thanks for the review btw, will update later)

python/tests/test_highlevel.py
```diff
@@ -2692,6 +2692,39 @@ def test_interval(self):
...
+            assert tree.right_sib(u) == tree.right_sib_array[u]
+
+    def test_tree_arrays(self):
+        ts = msprime.simulate(10, recombination_rate=1, random_seed=1)
+        assert ts.num_trees > 1
+        for tree in ts.trees():
+            self.verify_tree_arrays(tree)
+
+    @pytest.mark.parametrize(
+        "array", ["parent", "left_child", "right_child", "left_sib", "right_sib"]
```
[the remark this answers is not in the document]
jeromekelleher: good call

jeromekelleher:
OK, updated and documented. @petrelharp, fancy taking a quick look?

(I finally figured out how to make repeated bits of text in RST: this would come in handy for the stats API!)

docs/python-api.rst
```diff
@@ -1,3 +1,11 @@
+.. |tree_array_warning| replace:: This is a high-performance interface which
```
petrelharp, line 1: woah nice

jeromekelleher: Yeah!

petrelharp:
I did look at this, btw - looks good.

PR #1328, opened 2021-04-23
Assert equal for tables and collections

### LINKED ISSUE #1076: Add ``assert_equal`` to TableCollection and TreeSequence

benjeffery:
WIP - Tables are done, Table Collection has a few options that I'm working on.

Fixes #1076

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -32,6 +32,7 @@ import pathlib
 import pickle
 import platform
 import random
+import re
 import struct
 import time
 import unittest
@@ -73,7 +74,7 @@ class UInt32Column(Column):
 class CharColumn(Column):
     def get_input(self, n):
         rng = np.random.RandomState(42)
-        return rng.randint(low=0, high=127, size=n, dtype=np.int8)
+        return rng.randint(low=65, high=122, size=n, dtype=np.int8)
 
 
 class DoubleColumn(Column):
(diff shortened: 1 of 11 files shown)


benjeffery:
@petrelharp as JK is away can I get a review here?

benjeffery:
@mergiyio rebase

benjeffery:
@Mergifyio rebase

benjeffery:
Coverage isn't updating for some reason.

python/tests/test_tables.py
```diff
@@ -1009,7 +1010,162 @@ def test_set_with_optional_properties(self, codec):
...
+    @pytest.fixture
+    def table1(self, test_rows):
+        table1 = self.table_class()
+        for row in test_rows[:5]:
+            table1.add_row(**row)
+        return table1
+
+    def test_equal(self, table1, test_rows):
+        table2 = self.table_class()
+        for row in test_rows[:5]:
```
petrelharp, line 1034: isn't this depending on the number of rows being 5? e.g., shouldn't you replace `5` with `table1.num_rows`?

python/tests/test_tables.py
```diff
@@ -1009,7 +1010,162 @@ def test_set_with_optional_properties(self, codec):
...
+                match=re.escape(
+                    f"self.{column_name2}={test_rows[4][column_name2]} "
+                    f"other.{column_name2}={test_rows[5][column_name2]}"
+                ),
+            ):
+                table1.assert_equals(table2)
+
+    def test_num_rows(self, table1, test_rows):
+        table2 = self.table_class()
+        for row in test_rows[:4]:
```
petrelharp, line 1117: `4` -> `table1.num_rows - 1`?

python/tests/test_tables.py
```diff
@@ -1009,7 +1010,162 @@ def test_set_with_optional_properties(self, codec):
...
+            table2.add_row(**row)
+        with pytest.raises(
+            AssertionError,
+            match=f"{type(table1).__name__} number of rows differ: self=5 other=4",
```
petrelharp, line 1121: oh I see, many things are depending on the number 5. perhaps it's OK.

petrelharp (approved):
Looks great! Thanks!!

python/tests/test_tables.py
benjeffery: The tables have a fix number of rows (10).

benjeffery:
> Looks great! Thanks!!
Thanks for the review!

Issue #1268, opened 2021-03-26
Define _repr_pretty_ for ts+tables &  _repr_svg_ for ts+trees


hyanwong:
It seems like Jupyter notebooks allow a whole set of `_repr_XXX_` methods, which, when `display()` is called, are tried in order depending on the abilities of the notebook. E.g.

```
# Quick hack to show _repr_XXX_ methods in a notebook
tskit.TreeSequence._repr_svg_ = tskit.TreeSequence.draw_svg
tskit.TreeSequence._repr_pretty_ = lambda self, p, cycle : p.text(str(self))
display(ts)  # Normal html view
display(ts, exclude=["text/html"])  # alternative SVG view
display(ts, exclude=["image/svg+xml", "text/html"])  # Will show the str() view
```

It seems sensible to provide `_repr_pretty_` for tree sequences, tables, and trees, and `_repr_svg_` for tree sequences and trees.

Actually, there's a question here about whether the "pretty" textual representation of a tree sequence should be  the `draw_text()` method (showing the trees) or the `__str__()` method (showing a tabular view).

jeromekelleher:
Totally, we should implement these.

One thing to think about is that these should work well on massive tree sequences as well as small example ones, so drawing isn't something that should be done unless it's asked for I think.

hyanwong:
> One thing to think about is that these should work well on massive tree sequences as well as small example ones, so drawing isn't something that should be done unless it's asked for I think.

Yeah, that's a really good point. I was half wondering if `display(ts)` should actually plot the svg, while `display(ts.tables)` should show what we currently have as the html tabular view, but that argues strongly against it.

I don't know if that would also apply to a tree: what would a user expect if they did `display(ts.first())` in a notebook? I think I would expect a graphical tree to be drawn. OTOH, that's inconsistent with `display(ts)` which shows the tabular form. So I don't know what's best here?

It would be trivial to define

```
class TreeSequence:
    def _repr_svg_(self, **kwargs):
        return self.draw_svg(path=None, **kwargs)
```

and similarly for `Tree`s, but would that conflict with your point above about explicitly asking for the SVG form rather than magically letting IPython/Jupyter do the work.

benjeffery:
Am I right in thinking `_repr_svg_` is only relevant for the Qt console? Most users will be using notebooks in browser.

hyanwong:
It works for me in the browser. Try the code below in a browser window:

```
import tskit
import msprime
[...]

PR #1329, opened 2021-04-26
Add basic _repr_svg_ method on a tree seq

hyanwong:
Partially addresses https://github.com/tskit-dev/tskit/issues/1268. Shame there's no way (I can find) of passing params to e.g. `IPython.display.display_svg`.

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1417,6 +1417,16 @@ class TestDrawTextExamples(TestTreeDraw):
                 t.draw_text(max_tree_height=bad_max_tree_height)
 
 
+class TestReprSvg(TestTreeDraw):
+    """
+    Tests the IPython _repr_svg_ shortcut
+    """
+
+    def test_repr_svg(self):
+        ts = self.get_simple_ts()
+        assert ts._repr_svg_() == ts.draw_svg()
+
+
 class TestDrawSvg(TestTreeDraw, xmlunittest.XmlTestMixin):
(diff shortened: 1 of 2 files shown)


benjeffery:
Your test failures here are unrelated - will push a fix shortly.

benjeffery:
Thanks @hyanwong - do you know if there are any jupyter backends that will try SVG before HTML? This may be an issue for large tree-sequences.

hyanwong:
I asked about priority order on SO. No reply yet: https://stackoverflow.com/questions/67253015/ipython-priority-order-of-repr-methods-using-display

benjeffery:
I've been throught the notebook source code, and it is still a mystery to me - lets see if anyone replies.

hyanwong:
`_repr_mimebundle_ ` defines a list of provided `_repr_` methods (https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject), and given that's an ordered list I suspect it might provide/override the priority order too.

hyanwong:
Someone knowledgeable has posted an answer, and from my reading, it looks like we might not want to provide `_repr_svg_` after all (see my question as a follow-up): https://stackoverflow.com/questions/67253015/ipython-priority-order-of-repr-methods-using-display/67276689#67276689

benjeffery:
Yes, I expected that to be the case. I think we hold off on this. We can do a version with appropriate safeguards if we want to later.

PR #1335, opened 2021-04-29
Fix bug where set_columns didn't update the metadata schema

### LINKED ISSUE #1334: Metadata schema cache is not updated after `set_columns`
This bug has been in the last couple of releases.

Nasty oversight (my fault) I just found is that when setting a `metadata_schema` using `set_columns` the metadata schema cache doesn't get updated. Found this as `copy` uses `set_columns`, and copies had differing schemas. Tests were passing as the copy passed equality as equality uses the low-level methods that use encoded schemas.

Fix is to update the cache in `set_columns`.

benjeffery:
Fixes #1334

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -579,7 +579,7 @@ class CommonTestsMixin:
                 copy = table.copy()
                 assert id(copy) != id(table)
                 assert isinstance(copy, self.table_class)
-                assert copy == table
+                copy.assert_equals(table)
                 table = copy
 
     def test_pickle(self):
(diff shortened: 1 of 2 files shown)


benjeffery:
@tskit-dev/admins Can I get a review here from someone? (Both to check my code and satisfy mergify!)

python/tests/test_tables.py
```diff
@@ -1009,6 +1009,23 @@ def test_set_with_optional_properties(self, codec):
...
                 assert md == row.metadata
 
+    def test_copy_metadata_schema(self):
+        table = self.table_class()
+        assert table.metadata_schema == tskit.MetadataSchema(None)
+        table.metadata_schema = tskit.MetadataSchema({"codec": "json"})
+        copy = table.copy()
+        table.assert_equals(copy)
+        # Check this independently to check the schema cache was invalidated
+        assert table.metadata_schema == copy.metadata_schema
```
petrelharp, line 1019: shouldn't you also have, say,
```
copy.metadata_schema = tskit.MetadataSchema(None)
assert table.metadata_schema != copy.metadata_schema
```

petrelharp: well I suppose this is getting tested in the metadata schema tests.

petrelharp (approved):
Ack, good catch. LGTM.

python/tests/test_tables.py
benjeffery: Added

PR #1337, opened 2021-05-03
Make dump tables operations const.

### LINKED ISSUE #1327: Semantics for table/treeseq "dump" functions.
(This issue is motivated by a conversation on Slack.)

The current functions to output tables and tree sequences to files have a few quirks:

* The functions expect a non-const pointer to "self".  The issue here is that output functions are usually expected to be non-modifying.
* The non-const-ness comes from the default behavior (options = 0), which will build the table indexes.  '
* The tree sequence version of the function shares this non-const pointer to self.  But the function is, in fact, de facto const because you cannot build a tree sequence from a table collection that is not indexed.

Proposal:

* Change the functions to take const pointers to self.
* In effect, this means TSK_NO_BUILD_INDEXES will be the default behavior.
* Keep the `options` as is for now, documenting that TSK_NO_BUILD_INDEXES is effectively deprecated (unless it is used elsewhere?).

Side effects:

* None for tree sequences (?)
* Tables will need to be indexed prior to calling `tsk_table_collection_dump` (if indexes are wanted in the output).  This change may break code for some cases.
* Presumably some tests will be affected?

molpopgen:
Fixes #1327 

# PR Checklist:

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -5080,11 +5080,11 @@ test_dump_unindexed_with_options(tsk_flags_t tc_options)
     CU_ASSERT_FALSE(tsk_table_collection_has_index(&tables, 0));
     ret = tsk_table_collection_dump(&tables, _tmp_file_name, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-    CU_ASSERT_TRUE(tsk_table_collection_has_index(&tables, 0));
+    CU_ASSERT_FALSE(tsk_table_collection_has_index(&tables, 0));
 
     ret = tsk_table_collection_load(&loaded, _tmp_file_name, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-    CU_ASSERT_TRUE(tsk_table_collection_has_index(&loaded, 0));
+    CU_ASSERT_FALSE(tsk_table_collection_has_index(&loaded, 0));
     CU_ASSERT_TRUE(tsk_node_table_equals(&tables.nodes, &loaded.nodes, 0));
     CU_ASSERT_TRUE(tsk_edge_table_equals(&tables.edges, &loaded.edges, 0));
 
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Looks great, thanks for this @molpopgen. I'd be inclined to do a "hard break" here and by removing the ``TSK_NO_BUILD_INDEXES`` flag and not adding the new ``TSK_BUILD_INDEXES_IF_ABSENT``. The old behaviour is just plain weird I think, and the only code we're likely to break is the test suite (which we can fix by calling ``build_indexes`` after load, if we need to). I'd see it as a pre-1.0 opportunity to clean things up a bit.

What do you think?

benjeffery:
> What do you think?

I know I originally suggested the flag, as a way to avoid any breakage, but now thinking about it I think it will be such a rare thing that it isn't worth the complexity.

molpopgen:
Sounds good.  I'll undo the new flag, remove the old, etc..

c/tests/test_tables.c
```diff
@@ -5078,6 +5078,8 @@ test_dump_unindexed_with_options(tsk_flags_t tc_options)
     parse_edges(single_tree_ex_edges, &tables.edges);
     CU_ASSERT_EQUAL_FATAL(tables.edges.num_rows, 6);
     CU_ASSERT_FALSE(tsk_table_collection_has_index(&tables, 0));
+    ret = tsk_table_collection_build_index(&tables, 0);
+    CU_ASSERT_EQUAL(ret, 0);
     ret = tsk_table_collection_dump(&tables, _tmp_file_name, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
```
jeromekelleher, line 5084: I think what we'd want to test is that we *don't* have an index after loading when we write a table collection without an index. So, delete the new lines and change the assert to FALSE below.

c/tests/test_tables.c
```diff
@@ -5273,8 +5271,8 @@ test_dump_fail_no_file(void)
     errno = 0;
 
     /* Trying to dump without first sorting fails */
```
jeromekelleher, line 5273: We should be able to dump and load the unsorted tables without problems. Can just change the test to asserting ret is 0.

c/tskit/tables.c
```diff
@@ -50,7 +50,7 @@ typedef struct {
 
 typedef struct {
     const char *name;
-    void *array;
+    const void *array;
```
jeromekelleher, line 53: This change is good, I'm just wondering: was it necessary to keep the compiler happy or did you spot the inconsistency and fix it?

jeromekelleher (approved):
Looks great, thanks @molpopgen. Some minor tweaks to the tests and we're good to merge.

We should also update the CHANGELOG, since this is an API and semantics breaking change.

c/tskit/tables.c
molpopgen: Compiler required it.  I am a bit concerned that the other arrays did not require similar casts, but decided to do the minimal necessary changes.

c/tests/test_tables.c
molpopgen: i

c/tests/test_tables.c
molpopgen: I think this is taken care of.

c/tests/test_tables.c
```diff
@@ -5273,8 +5264,8 @@ test_dump_fail_no_file(void)
...
-    ret = tsk_table_collection_dump(&t1, _tmp_file_name, 0);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_EDGES_NOT_SORTED_PARENT_TIME);
+    /* ret = tsk_table_collection_dump(&t1, _tmp_file_name, 0); */
```
jeromekelleher, line 5267: stray comments here

jeromekelleher:
Looks great @molpopgen - just need to delete a few stray comments and squash

molpopgen:
@jeromekelleher -- I think this is ready.  Go ahead and merge!

benjeffery:
Sorry, meant to check this over earlier, but LGTM too! Thanks @molpopgen

PR #1344, opened 2021-05-10
Prevent extending from self

benjeffery:
When extending, it was possible to extend using rows from `self` however this is unsafe as `add_row` can re-allocate any ragged arrays, making the data form `get_row` invalid. Seems simplest to prevent doing this operation as I can't see much use for it.


jeromekelleher (approved):
Yep, good catch and call.

PR #1348, opened 2021-05-12
Add fancy indexing to tables

### LINKED ISSUE #1221: Allow tables to be subset using numpy indexing
We have `tables.edges.truncate(up_to_row_n)`, but for https://github.com/tskit-dev/tskit/discussions/1220 it would be useful to be able to subset the edges to a selection to keep. At least in the case of edges, I think in so doing we would always create a valid tree sequence, right?

I'm guessing it would look something like this (untested) method of an edge table, but done using the `[]` operator (see @jeromekelleher's comment below)

```
def subset(self, keep):
    metadata, metadata_offset = keep_with_offset(keep, self.metadata, self.metadata_offset)
    self.set_columns(
        left = self.left[keep],
        right=self.right[keep],
        parent=self.parent[keep],
        child=self.child[keep],
        metadata=metadata
        metadata_offset=metadata_offset)
```

The fiddling with metadata is probably tricky enough to want to incorporate this as a tskit method, rather than expect the user to write it from scratch, IMO. Does this seem worth incorporating into tskit? I suggest it as I imagine it might be a generally useful method, and I can't see it causing any tricky edge cases.

benjeffery:
Fixes #1221

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -115,6 +115,31 @@ class CommonTestsMixin:
             for j in range(num_rows)
         ]
 
+    @pytest.fixture
+    def test_rows(self, scope="session"):
+        test_rows = self.make_transposed_input_data(10)
+        # Annoyingly we have to tweak some types as once added to a row and then put in
+        # an error message things come out differently
+        for n in range(10):
+            for col in test_rows[n].keys():
+                if col in ["timestamp", "record", "ancestral_state", "derived_state"]:
+                    test_rows[n][col] = bytes(test_rows[n][col]).decode("ascii")
+        return test_rows
+
+    @pytest.fixture
+    def table(self, test_rows):
+        table = self.table_class()
+        for row in test_rows:
+            table.add_row(**row)
+        return table
+
+    @pytest.fixture
+    def table_5row(self, test_rows):
+        table_5row = self.table_class()
+        for row in test_rows[:5]:
+            table_5row.add_row(**row)
+        return table_5row
+
     def test_max_rows_increment(self):
         for bad_value in [-1, -(2 ** 10)]:
             with pytest.raises(ValueError):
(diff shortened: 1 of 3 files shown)


python/tests/test_tables.py
```diff
@@ -1031,52 +1056,34 @@ def test_set_columns_metadata_schema(self):
...
+            match=f"Types differ: self={type(table_5row)} other=<class 'int'>",
         ):
-            table1.assert_equals(42)
+            table_5row.assert_equals(42)
 
-    def test_metadata_schema(self, table1):
-        if hasattr(table1, "metadata_schema"):
-            assert table1.metadata_schema == tskit.MetadataSchema(None)
-            table2 = table1.copy()
+    def test_metadata_schema(self, table_5row):
```
jeromekelleher, line 1072: I like the abstract idea of fixtures (i.e., not making up test data over and over again), but I have to say I find this idiom of changing all the variables in a test to (effectively) a global variable really nasty. These tests are now objectively less readable that they were before. Wouldn't a parametrize with a set of locally-defined example tables be better?

jeromekelleher: Hmm, I see, you've made it a 5row table so that you know in later tests that you can do things like ``table[:4]`` or whatever. Maybe disregard above comment then.

python/tskit/tables.py
```diff
@@ -335,16 +336,41 @@ def __setattr__(self, name, value):
 
     def __getitem__(self, index):
         """
-        Return the specified row of this table, decoding metadata if it is present.
-        Supports negative indexing, e.g. ``table[-5]``.
+        If passed an integer, return the specified row of this table, decoding metadata
+        if it is present. Supports negative indexing, e.g. ``table[-5]``.
+        If passed a slice, iterable or array return a new table containing the specified
+        rows. Similar to numpy fancy indexing, if the array or iterables contains
+        booleans then in acts as a mask, returning those rows for which are True.
```
jeromekelleher, line 343: s/in/it

jeromekelleher: The sentence isn't quite grammatical, something like "returning those rows for which the mask is True."

python/tskit/tables.py
```diff
@@ -335,16 +336,41 @@ def __setattr__(self, name, value):
...
+
+        :param index: the zero-index of a desired row, a slice of the desired rows, an
+            iterable or array of the desired row numbers, or a boolean array to use as
+            a mask.
+        """
+
+        if isinstance(index, numbers.Integral):
+            # Single row by integer
```
jeromekelleher, line 351: Out of curiosity, could you do ``index = slice(index).indices(len(self))[0]``?

More obscure, I guess, but saves a bit of fiddly code?

jeromekelleher (approved):
Awesome! Minor comments.

python/tests/test_tables.py
benjeffery: The main change is moving the fixture to where it can be used by `AssertEqualsMixin` and `FancyIndexingMixin`. I agree the non-locality of fixtures can be a readability issue, but generally think it is worth the trade off.

python/tskit/tables.py
benjeffery: Unfortunately not:
```python
>>> slice(5).indices(10)
(0, 5, 1)
```

python/tskit/tables.py

PR #1458, opened 2021-05-17
Update tutorial.rst

minesrebollo:
typo: .) to ).

## Description

Fixes #(issue) <- Putting the issue number here will auto-close the issue when this PR is merged 

# PR Checklist:

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -958,7 +958,7 @@ the excellent `scikit-allel
 <http://scikit-allel.readthedocs.io/en/latest/index.html>`_
 for the basic `plotting code
 <http://scikit-allel.readthedocs.io/en/latest/_modules/allel/stats/ld.html#plot_pairwise_ld>`_
-used here.)
+used here).
 
 .. code-block:: python
 


petrelharp:
Thanks for checking! But I agree - it's right the way it is.

minesrebollo:
Yes, I was wrong! Sorry.

I will use tskit and tsinfer and I'm reading the documentation and
tutorials, they are very detailed and helpful, thank you!

I wanted to contribute as I spotted some things (in Python API) but
couldn't find how to suggest changes and got excited when I could edit this
on and rushed before double-checking. I realized I was wrong and tried to
delete it later but apparently I didn't do it correctly. Sorry again.

What I saw in Python API and couldn't change I guess it is part of the auto
summary. For instance, the overview of ltrim is truncated (it ends in
"i.e.", purple rectangle in the screenshot I attached). Similar to this
happens in generate_star and num_children that also have i.e. Interval
overview sentence seems to be missing something.

Maybe it is my screen but I couldn't make use of the whole screen to read
the overviews (blue rectangle in the screenshot).

There is also a parenthesis as superscript that shouldn't in the formula of
generate_random_binary.

Last, is it possible that there is an n missing in the word give in
tskit.tree sample_lists parameter description? (Maybe it is not, I am not a
native so sorry if it is ok like that.)

Best,
Inés

[image: Screen Shot 2021-05-17 at 16.54.56.png]

petrelharp:
> What I saw in Python API and couldn't change I guess it is part of the auto
summary. For instance, the overview of ltrim is truncated (it ends in
"i.e.", purple rectangle in the screenshot I attached). Similar to this
happens in generate_star and num_children that also have i.e. Interval
overview sentence seems to be missing something.

I don't see your screenshot (I'm not sure if email attachments come through - you might need to attach them directly in github?), but I see what you mean. I think that the code is autogenerating a summary based on the first "sentence", which it thinks is up until the first period-followed-by-a-space.  Maybe the right way to fix this would be to put a comma after i.e., like `i.e.,` (which is more correct anyhow).

> There is also a parenthesis as superscript that shouldn't in the formula of
generate_random_binary.

Hah, why so there is. Good catch. I think it's supposed to be 2-to-the-power-of-(n-2), so `2^{n-2}` instead of `2^(n-2)`.

These things are autogenerated, so to fix them you'd need to go to where they're defined in the python code. For instance,
[here's](https://github.com/tskit-dev/tskit/blob/main/python/tskit/trees.py#L2788) where the parenthesis problem occurs. Want to have a go at fixing them?

> I wanted to contribute as I spotted some things (in Python API) but
couldn't find how to suggest changes and got excited when I could edit this
on and rushed before double-checking. 

No worries, we've all been there. =)

minesrebollo:
> These things are autogenerated, so to fix them you'd need to go to where they're defined in the python code. For instance,
> [here's](https://github.com/tskit-dev/tskit/blob/main/python/tskit/trees.py#L2788) where the parenthesis problem occurs. Want to have a go at fixing them?

Thank you for pointing it out!!

PR #1460, opened 2021-05-18
Update trees.py

minesrebollo:
## Description

The autogenerated summary of methods and functions is truncated at every i.e. because, as pointed by @petrelharp, it is detected as the end of the sentence. I added commas after i.e. to solve that and then to every other occurrence of i.e. for consistency.

Random_binary formula had () instead of {} so the superscript was not right.

Missing "n" in sample_lists description ... ?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -70,7 +70,7 @@ class Interval(BaseInterval):
         in the interval.
     :vartype left: float
     :ivar right: The right hand end of the iterval. By convention this value is *not*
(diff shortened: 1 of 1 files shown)


minesrebollo:
I am having some pre-commit issues with Lint. At first I thought the problem was the length of the new line and changed it but it still doesn't pass...

benjeffery:
I've fixed the linting (was just a code format fix) and squashed down to one commit. Many thanks @minesrebollo!

benjeffery (approved):

PR #1468, opened 2021-05-25
Use integer format for text trees if positions are integers

hyanwong:
As we do for the SVG format. This is useful now that msprime.sim_ancestry defaults to integer breakpoints.

I'm not sure if we need additional tests, as I've adjusted some cases in the existing tests which have 0,1 breakpoints. I also feel it's probably too trivial to stick in the changelog?


benjeffery (approved):
@hyanwong I'm guessing that there are some tests that do and don't meet the integer criteria already - so were hitting both paths?

hyanwong:
Yes, but only by comparing against known (saved) formats, rather than explicitly testing the integer-ness of the labels. But extracting the labels to test against, out of all that SVG (or even out of the text representation) is a pain, so I reckon that's the easiest way to test it all.

Issue #1345, opened 2021-05-10
Method to create brick table


awohns:
One helpful enhancement to `tskit` is to create a "brick table" from a tree sequence. A brick table is an augmented edge 
table where edges are bifurcated when they have different descendants across adjacent trees. These bifurcated edges can be used to trace tracts of genome which are inherited as units. This is useful in applications where one is interested in patterns of descent where _position_ matters, for instance if looking at which bits of an ancestral haplotype are inherited by its descendants. The term "brick" refers to how the result can be visualised as stacked bricks.

For example, consider the following tree sequence:
![image](https://user-images.githubusercontent.com/10664424/117722571-387c5e80-b1af-11eb-9de0-15029b13958f.png)

Note that the edge from node 6 to node 7 exists in both trees, but has different descendants. The brick version of this tree sequence would bifurcate this edge, resulting in two edges with 6 as the child and 7 as the parent, the first from 0-0.62 and the second from 0.62-1.

I've implemented this in a straightforward manner using  `ts.trees()` and `ts.edge_diffs()` where we identify nodes with different parents across adjacent trees. We then climb the adjacent trees from the node with different parents, bifurcating edges along this path which straddle both trees.

I'm not sure what the best name for this method is... one idea is to simply call it `split_edges`. It would also make sense add an option `map_edges`, along the lines of `map_nodes` in `simplify`. If this parameter is specified, the method would returns a tuple containing the resulting tree sequence and a numpy array mapping edge IDs in the current tree sequence to their corresponding edge IDs in the returned tree sequence.

If this sounds interesting to the community, I can create a PR with my implementation.

petrelharp:
Yes! And, I like `split_edges`. Question: is it different descendant *nodes* or different descendant *samples*? I think the latter?

petrelharp:
Thinking ahead, perhaps the next thing to implement is the analogue of `genetic_relatedness` but for edges, and so if we've first split up the edge table then we'll get it for bricks. (need to write down the details though)

hyanwong:
I like the description of a "brick" table. It makes it much clearer what the `split_edges` function is doing. It would probably be even clearer if we could visualise the edges as horizontal bars (i.e. really like Tetris bricks). There's an argument that there should be some viz method that shows this somehow (perhaps ts.tables.draw_svg(), but then I'm not sure how we represent the mutations table).

[...]

PR #1469, opened 2021-05-25
Implement split_edges

awohns:
Implements a naive, python version of `ts.split_edges` as described in #1345 with a few tests.

I've made it a draft PR for now. If the approach and output look reasonable I'll work up the C version. 

My implementation uses `edge_diffs` to bifurcate edges either by `leaf`, `node` or `tree`. In `mode=leaf`, edges are bifurcated when they different descendant leaves in different trees. In `mode=node`, edges are bifurcated when they have different descendant nodes in different trees. When `mode=tree`, edges that exist in multiple trees are bifurcated to create a "JBOT" (just a bunch of trees). 

I originally had `mode=sample` instead of `mode=leaf`, but that mean keeping track of the samples beneath each edge as we go along if there were internal samples or non-sample leaves. This seemed less elegant and algorithmically problematic than using `edge_diffs`... but perhaps not. Perhaps there's a better way to implement this?

It's also worth mentioning that we might want an `edge` (or `brick`) mode which analogously splits edges based on the edges beneath it, but I haven't implemented that here.

# PR Checklist:

- [ ] Functionality matches what we want
- [ ] C version of function
- [ ] Tests that fully cover new functionality.
- [ ] Documentation of `split_edges()`
- [ ] Update changelog

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -6423,7 +6423,7 @@ class TestMutationTime:
         )
         mutations = io.StringIO(
             """\
-        site	node	time	derived_state	parent
+        site    node    time    derived_state   parent
         0       1       1.5     1               -1
         0       2       1.5     1               -1
         0       3       0.5     2               1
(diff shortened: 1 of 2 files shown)


petrelharp:
I'll read through it in detail - but, it would really help if first you wrote down what the proposed API is? And, more importantly, what exactly it returns? (i.e., when exactly edges get split into what)

benjeffery:
Tidying up old PRs - please re-open if you want to get this in.

PR #1473, opened 2021-05-27
Typed namedtuples

hyanwong:
Should be a like-for-like switch, but helps with typing and documentation

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -34,6 +34,7 @@ import textwrap
 import warnings
 from dataclasses import dataclass
 from typing import Any
+from typing import NamedTuple
 from typing import Optional
 from typing import Union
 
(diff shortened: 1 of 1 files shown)


benjeffery (approved):
Looks fine to me, although I wonder why there was a `BaseInterval` class? 
Removing this is, strictly speaking, a breaking change. I'd like to check with @jeromekelleher on this as he maybe aware of possible usage of this base class.

hyanwong:
> Looks fine to me, although I wonder why there was a `BaseInterval` class?
> Removing this is, strictly speaking, a breaking change. I'd like to check with @jeromekelleher on this as he maybe aware of possible usage of this base class.

I stuck it in in a previous version (before knowing about typed NamedTuple) so that I could derive a class from it to add the `span` property. I don't think it's referenced anywhere else in the codebase (and it's was never documented, so I think we needn't treat it as a breaking change)

benjeffery:
Ah yes, thanks for jogging my memory. This should be fine then.

PR #1476, opened 2021-05-28
Add explicit links to definition terms

hyanwong:
This is useful if we want to explain e.g. what a "sample" is. I haven't looked to see if it messes up formatting, but I hope not.

### DIFF
--- a/docs/data-model.rst
+++ b/docs/data-model.rst
@@ -28,11 +28,15 @@ Definitions
 
 To begin, here are definitions of some key ideas encountered later.
 
+.. _sec_data_model_definitions_tree:
+
 tree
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I'm not sure this works, you need sections to refer to.

Can you try using these labels at least once each to see if they actually work?

hyanwong:
> I'm not sure this works, you need sections to refer to.
> 
> Can you try using these labels at least once each to see if they actually work?

I just tried it and it worked fine: I added
```
:ref:`test me<sec_data_model_definitions_mutation>`
```

to an .rst page and it created the link as expected (the docs say it should be OK too).

benjeffery (approved):

PR #1478, opened 2021-06-02
Use offsets to get labels right in draw_svg(x_lim)

### LINKED ISSUE #1477: Bug in labelling during svg_draw with x_lim
The `x_lim` parameter to `draw_svg` uses `keep_intervals(..., simplify=False)` under the hood to reduce the TS to only the trees that are needed to plot. But even with simplify=False, this deletes sites (and mutations) outside the selected regions, and changes the number of trees. As a result, the plotted mutation labels are wrong, and the classes for trees (e.g. `t0`), sites (e.g. `s3`) and mutations(e.g. `m7`) are wrong.

After some to-and-fro-ing, it seems like the simplest option is to take account of the offsets from new to old sites, mutations, and trees.

hyanwong:
## Description

Fixes #1477

# PR Checklist:

### DIFF
--- a/python/tests/data/svg/ts_x_lim.svg
+++ b/python/tests/data/svg/ts_x_lim.svg
@@ -27,31 +27,31 @@
 						<text>0.79</text>
 					</g>
 				</g>
-				<g class="site s0" transform="translate(25.9364 163.2)">
+				<g class="site s1" transform="translate(25.9364 163.2)">
 					<line class="sym" x1="0" x2="0" y1="0" y2="-10"/>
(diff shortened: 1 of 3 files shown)


benjeffery (approved):
Thanks @hyanwong 
Seems to be the easiest way to fix this, my concern about the complexity of the drawing code is only rising though!

jeromekelleher (approved):
LGTM

benjeffery:
No idea why all the CI has decided to ignore this. Trying a re-push

PR #1487, opened 2021-06-07
Added fam file parser

nikbaya:
## Description

Added new text_formats.py module with the function `parse_fam`, which converts PLINK fam files into tskit format.

### DIFF
--- a/python/tests/test_text_formats.py
+++ b/python/tests/test_text_formats.py
@@ -1,0 +1,249 @@
+# MIT License
+#
+# Copyright (c) 2021 Tskit Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for converting fam file to tskit
+"""
+import dataclasses
+import tempfile
+from dataclasses import asdict
+
+import numpy as np
+import pytest
+
+import tskit
+
+
+@dataclasses.dataclass
+class FamEntry:
+    fid: str = "0"
+    iid: str = "0"
+    pat: str = "0"
+    mat: str = "0"
+    sex: str = "0"
+    phen: str = None
+
+    def get_row(self, delimiter="\t"):
+        return delimiter.join([x for x in asdict(self).values() if x is not None])
+
+
+class TestParseFam:
+    """
+    Tests for the parse_fam function.
+    """
+
+    def get_parsed_fam(self, entries, delimiter="\t"):
+        content = "\n".join([entry.get_row(delimiter=delimiter) for entry in entries])
+        with tempfile.TemporaryFile() as f:
+            f.write(bytes(content, "utf-8"))
+            f.seek(0)
+            return tskit.parse_fam(f)
+
+    def test_empty_file(self):
+        entries = []
+        with pytest.warns(UserWarning):
+            tb = self.get_parsed_fam(entries=entries)
+        assert len(tb) == 0
+
+    @pytest.mark.parametrize("iid", ["1", "a", "100", "abc"])
+    def test_single_line(self, iid):
+        entries = [FamEntry(iid=iid)]
+        tb = self.get_parsed_fam(entries=entries)
+        assert len(tb) == 1
+        assert np.array_equal(tb[0].parents, [-1, -1])
+        assert tb[0].metadata["plink_fid"] == "0"
+        assert tb[0].metadata["plink_iid"] == str(iid)
+        assert tb[0].metadata["sex"] == 0
+
+    @pytest.mark.parametrize("iids", [("1", "2"), ("a", "b")])
+    def test_multiple_line_file(self, iids):
+        # test both integer and string IIDs
+        iid1, iid2 = iids
+        entries = [FamEntry(iid=iid1), FamEntry(iid=iid2)]
+        tb = self.get_parsed_fam(entries=entries)
+        assert len(tb) == 2
+        for idx in range(2):
+            assert np.array_equal(tb[idx].parents, [-1, -1])
+            assert tb[idx].metadata["plink_fid"] == "0"
+            assert tb[idx].metadata["plink_iid"] == str(entries[idx].iid)
+            assert tb[idx].metadata["sex"] == 0
+
+    @pytest.mark.parametrize("n_cols", range(1, 5))
+    def test_insufficient_cols(self, n_cols):
+        fields = list(asdict(FamEntry()))
+        entry = FamEntry(iid="1")
+        for field in fields[n_cols:]:
+            entry.__setattr__(field, None)
+        with pytest.raises(IndexError):
+            self.get_parsed_fam(entries=[entry])
+
+    def test_unrelated_duplicate_iids(self):
+        # Individuals have the same IID, but are in different families
+        entries = [FamEntry(iid="1"), FamEntry(fid="1", iid="1")]
+        tb = self.get_parsed_fam(entries=entries)
+        assert len(tb) == 2
+        assert tb[0].metadata["plink_fid"] == "0"
+        assert tb[0].metadata["plink_iid"] == "1"
+        assert tb[1].metadata["plink_fid"] == "1"
+        assert tb[1].metadata["plink_iid"] == "1"
+
+    def test_duplicate_rows(self):
+        entries = [FamEntry(iid="1"), FamEntry(iid="1")]
+        with pytest.raises(ValueError):
+            self.get_parsed_fam(entries=entries)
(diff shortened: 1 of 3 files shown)


hyanwong:
It's probably worth squashing all the commits here together into one, @nikbaya ? I think `git rebase -i` should do it, then flag all but the first to "squash".

benjeffery:
Great stuff @nikbaya - will get to reviewing this now. 

One quick pointer as I can see merge commits in this PR which can complicate things when we come to merge the PR. It is usually easier to rebase your branch as `main` is changing, rather than merge `main` into your branch. The other option is not to worry about rebasing at all until your work is done, the only time you really need to rebase onto `main` is when there have been changes on `main` that are relevant to your PR. A guide to the recommended workflow is at https://tskit.dev/tskit/docs/stable/development.html#git-workflow

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+        }
+    )
+    for plink_fid, plink_iid, pat, mat, sex in individuals:
+        sex = int(sex)
+        if not (sex in range(3)):
+            raise ValueError(
+                "Sex must be one of the following: 0 (unknown), 1 (male), 2 (female)"
+            )
+        metadata_dict = {"plink_fid": plink_fid, "plink_iid": plink_iid, "sex": sex}
+        pat = plink_fid + " " + pat if pat != "0" else pat
```
jeromekelleher, line 82: If you do setdefault, then you can do ``pat_id = f"{plink_fid} {pat}"

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+        mat = plink_fid + " " + mat if mat != "0" else mat
+        tb.add_row(parents=[id_map[pat], id_map[mat]], metadata=metadata_dict)
```
jeromekelleher, line 84: I think we need to do a topological sort here on the parents table to make sure that it can be used as a valid tree sequence. Doing ``tables.sort()`` should do it.

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+    return tc
```
jeromekelleher, line 85: I'm not sure there's much point in returning ``tc`` here, since we're passing it in as a parameter.

python/tskit/text_formats.py
jeromekelleher: We should test this by making an example where some children come before their parents, and then calling ``ts = tables.tree_sequence()`` on it.

jeromekelleher (commented):
This is great, thanks @nikbaya! Some comments above.

Echoing other comments above, your life would be simpler if you stopped merging and used rebase instread. The link Ben put in a above is quite helpful, it's worth following the guide fully once.

benjeffery:
I've added a discussion at https://github.com/tskit-dev/tskit/discussions/1488 about these kinds of routines

python/tests/test_text_formats.py
```diff
@@ -0,0 +1,198 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for converting fam file to tskit
+"""
+import os
+import tempfile
+import unittest
```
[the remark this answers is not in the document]
benjeffery: You can save a lot of boring busy work with `unittest2pytest`

jeromekelleher:
Thanks for that @benjeffery. I think for now we can keep this PR going and aim to merge it, but avoid making parse_fam part of the documented API. That way we can use it for ongoing work, but it quite self contained and will be easy to move into tsconvert if we choose to do so.

python/tests/test_text_formats.py
```diff
@@ -0,0 +1,198 @@
...
+
+class TestParseFam(unittest.TestCase):
+    """
+    Tests for the parse_fam function.
+    """
+
+    def get_parsed_fam(self, entries, delimiter="\t"):
+        tc = tskit.TableCollection(0)
+        content = "\n".join([delimiter.join(list(map(str, row))) for row in entries])
+        with tempfile.TemporaryDirectory() as temp_dir:
```
benjeffery, line 43: tmp_path from https://docs.pytest.org/en/6.2.x/fixture.html is the `pytest` way here.

python/tests/test_text_formats.py
```diff
@@ -0,0 +1,198 @@
...
+            fam_path = os.path.join(temp_dir, "test.fam")
+            with open(fam_path, "w") as f:
+                f.write(content)
+            tc = tskit.parse_fam(fam_path=f.name, tc=tc)
+        tb = tc.individuals
+        return tb
+
+    def test_empty_file(self):
+        entries = [[]]
+        self.assertWarns(UserWarning, self.get_parsed_fam, entries=entries)
```
benjeffery, line 53:
```python
with pytest.warns(UserWarning):
    self.get_parsed_fam(entries=entries)
```

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+            raise ValueError("Duplicate PLINK ID")
+        id_map[plink_id] = tskit_id
+    id_map["0"] = "-1"
+
+    tb = tc.individuals
+    tb.metadata_schema = tskit.MetadataSchema(
+        {
+            "codec": "json",
+            "type": "object",
+            "properties": {
```
benjeffery, line 66: I'd be tempted to put properties `fid`,`iid` and `sex` under a `plink` object.

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+# SOFTWARE.
+"""
+Module responsible for working with text format data.
+"""
+import numpy as np
+
+import tskit
+
+
+def parse_fam(fam_path: str, tc: tskit.TableCollection):
```
benjeffery, line 30: Most of our conversion routines return a new `TableCollection`, not modify an existing one. Is the thinking here that the dominant use case is adding individuals to an existing tree sequence?

hyanwong: The TC manipulation routines, OTOH, modify a TC in-place. For those, we decided that we should impose the rule that table collection routines modify in-place, whereas tree sequence routines alway return a new TS instance (which they have to, because of immutability). So I wonder if Nik is right here, and any TC routines that take a TC as input should do an in-place modification?

benjeffery: Yes, but an input TC isn't needed here. So maybe the right thing is to return the TS.

hyanwong: Oh, I guess I didn't understand what the `tc: tskit.TableCollection` was doing in the function sig then, sorry - didn't look at the code itself.

jeromekelleher: The original thinking was that this is only affecting the individual table, so it's not going to return a very interesting (or useful) TableCollection on its own. The only way you could use it is to either update the returned tables to set sequence length, add edges/nodes etc, or to copy the individual table that we create here.

There is an argument for just returning the IndividualTable I guess, which would clarify this?

benjeffery: Ah I see. Returning an `IndividualTable` is conceptually clearer - but it is currently a bit clunky to use. You can't assign tables  to a `TableCollection` although I see no reason why you shouldn't be able to if we implemented it: https://github.com/tskit-dev/tskit/issues/1489

python/tskit/text_formats.py
```diff
@@ -0,0 +1,85 @@
...
+    id_map = {}  # dict for translating PLINK ID to tskit IndividualTable ID
+    for tskit_id, (plink_fid, plink_iid, _pat, _mat, _sex) in enumerate(individuals):
+        # need this to guarantee uniqueness of individual ID
+        plink_id = (
+            str(plink_fid) + " " + str(plink_iid)
+        )  # include space between strings to ensure uniqueness
+        if plink_id in id_map:
+            raise ValueError("Duplicate PLINK ID")
+        id_map[plink_id] = tskit_id
+    id_map["0"] = "-1"
```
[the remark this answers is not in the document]
nikbaya: Shouldn't I be using `setdefault` later when I'm trying to key into `id_map` with the parent IDs? For instance:
```
id_map.setdefault(pat, "-1")
```
would return the corresponding value for the key `pat`, otherwise it returns "-1"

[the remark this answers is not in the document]
nikbaya: just a note that using setdefault will not raise a KeyError when trying to map a nonzero parent ID of a parent not present in the fam file.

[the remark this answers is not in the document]
jeromekelleher: Yes, I think the right option here is to raise a KeyError if there are unknown IDs in the file. We can add options to do something different later if we need to , but the default should be to fail noisily I think.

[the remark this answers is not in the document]
jeromekelleher: Sorry, my bad on the setdefault thing, I misunderstood what it did. Really the correct thing to do is just map "0" to "-1" and raise a KeyError if anything else is specified.

python/tests/test_text_formats.py
nikbaya: I’m having some trouble figuring out the pytest tmp_path fixture. I don’t want to pass tmp_path as an argument to each test function. How can I access tmp_path within get_parsed_fam?

python/tskit/text_formats.py
nikbaya: how would I do this?

python/tskit/text_formats.py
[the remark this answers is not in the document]
jeromekelleher: We should resolve this I think, even if it means not bothering with type annotations. We want to support reading from a File object, not just a path. 

Also, let's change to returning a tskit.InividivualTable rather than taking the TableCollection as a parameter (which is a bit ugly)

python/tskit/text_formats.py
jeromekelleher: Let's keep it flat for now - I think we'll need to think hard about the metadata we want to support for individuals, and what we have now almost certainly isn't it. This will get us going, though.

jeromekelleher (commented):
I think we can wrap this up soon @nikbaya, just a few changes to make.

python/tests/test_text_formats.py.bak
```diff
@@ -0,0 +1,215 @@
+# MIT License
+#
```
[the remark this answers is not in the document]
nikbaya: could you clarify what you mean by an extra file?

nikbaya:
@jeromekelleher you suggested not taking a TableCollection as a parameter (https://github.com/tskit-dev/tskit/pull/1487#discussion_r651935295). Should the only argument be a file object? If so, should I be creating an empty TableCollection in which to put the fam file contents? What sequence length should I use?

jeromekelleher (approved):
LGTM!

PR #1496, opened 2021-06-15
Cache decoded schemas, to speed up replicates

benjeffery:
Fixes #1945

`MetadataSchema` classes are immutable so this is an easy win.

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -26,6 +26,7 @@ import abc
 import builtins
 import collections
 import copy
+import functools
 import json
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
Awesome! There's zero overhead right, we're just using the string's hash as a dictionary key I'd imagine?

benjeffery:
> Awesome! There's zero overhead right, we're just using the string's hash as a dictionary key I'd imagine?

Yep, underlying implementation is a `dict` so that's exactly what will happen.

Issue #1498, opened 2021-06-17
Restructure statistics.html


hyanwong:
After a brief chat with @petrelharp we thought it might be time to revisit the general structure of https://tskit.dev/tskit/docs/stable/stats.html, because (for instance) it doesn't mention the LDcalc stuff, or have obvious room for @gtsambos 's IBD stuff, both of which can be thought of as "multi-site stats". We should also add GNN, which is a single site stat, and probably leave room somewhere for talking about time windows.

The current structure can be roughly summarised as follows:

```
# Statistics
intro
----------------
## Available statistics

### Single site stats (just lists them out below)
Some general stats (e.g. diversity, AFS)
#### Patterson’s f statistics
f4, f3, f2
#### Y statistics
Y3, Y2
#### Trait correlations
#### Derived statistics
#### General methods

### AFS
complex stuff about gotchas when calculating AFS for multi-allelic sites

----------------
## Interface
intro
### Mode
### Sample sets and indexes
#### One-way methods
#### Multi-way methods
### Windows
### Span normalize
### Output format
### Output dimensions

----------------
## General API
stuff here about general API for single site stats, e.g. polarization

```

I'll follow up below with a suggestion for an updated structure.

hyanwong:
I've worked up one rearrangement into a rough PR at https://github.com/tskit-dev/tskit/pull/1499, which puts the main list of "Available stats" into a table rather than a set of lists with headers, as I found the long-list style was difficult to see at-a-glance:

<img width="708" alt="Screenshot 2021-06-17 at 12 26 36" src="https://user-images.githubusercontent.com/4699014/122387913-44ee9680-cf67-11eb-879f-11978f7a3b36.png">


Links from the titles in the table take you to a "notes" section below, so the AFS details fit there nicely. The doc structure then allows for a place for the multi-site stats docs (assuming they don't fit into the "General API":

```
# Statistics
intro - introduce single-site vs multi-site stats

## Available stats
*new table*
### Notes
#### Patterson’s f statistics
#### Y statistics
#### Trait correlations
#### Derived statistics
#### AFS

## Single site stats
### Interface
#### Mode
...
### General API
#### Polarization
...

## Multi site stats
todo
```

gtsambos:
Hey @hyanwong, it's about these time statistics got a top-down restructure -- thanks for taking this on. I have two comments:

[...]

Issue #1237, opened 2021-03-10
windowed GNN C implementation


awohns:
#683 provides a python implementation of the windowed GNN statistic. Next steps are to modify the C implementation of `genealogical_nearest_neighbours` to support `windows` and `time_windows` and to update the documentation.

I've reproduced the documentation I wrote for a `windowed_genealogical_nearest_neighbours` function below. Bits of this can just be spliced into the original `genealogical_nearest_neighbours` documentation.

A related issue is to expand the stats API to support time windows, probably as they're defined in GNN, (which will negate the need for some of the documentation below), but that's a bigger undertaking.

```
def windowed_genealogical_nearest_neighbours(
        self,
        focal,
        reference_sets,
        windows=None,
        time_windows=None,
        span_normalise=True,
        time_normalise=True,
    ):
        """
        Returns genealogical nearest neighbour proportions partitioned
        into span-based `windows` and time-based `time_windows`.
        If neither `windows` nor `time_windows` are specified, the output is
        identical to that of :meth:`TreeSequence.genealogical_nearest_neighbours`.
        Passing arguments to `windows` and `time_windows` each increase the dimensions
        of the output array by 1. If `windows` is not None, GNN proportions are computed
        in windows defined by user-provided coordinates, as defined in the
        :ref:`stats API <sec_stats_windows>`. If `time_windows` is not none,
        GNN proportions are computed in user-provided time windows.
        For each focal node in each tree, the relevant time window is found based on the
        age of the most recent common ancestral node :math:`a` between the focal node
        and any other node present in the reference sets, as defined in
        :meth:`TreeSequence.genealogical_nearest_neighbours`.
        See the :ref:`statistics interface <sec_stats_interface>` section for details on
        :ref:`windows <sec_stats_windows>` and
        :ref:`span normalise <sec_stats_span_normalise>`. While `time_windows` are
        analagous to `windows`, note that `time_windows` do not need to extend to the
        age of the oldest root node. The `time_normalise` parameter is analagous to
        `span_normalise` with normalisation constants determined by the span of
        sequence assigned to each `time_window`.
        .. warning:: The interface for this method is preliminary and may be subject to
            backwards incompatible changes in the near future. The long-term stable
            API for this method will be consistent with other :ref:`sec_stats`.
        :param list focal: A list of :math:`n` nodes whose GNNs should be calculated.
        :param list sample_sets: A list of :math:`m` lists of node IDs.
[...]

PR #1499, opened 2021-06-17
Stats page: add a table, rework sections

hyanwong:
## Description

Addresses some of #1498 - but this is only a draft for comments


petrelharp:
Thanks for having a go at this. Here's how this renders, for reference:
Screenshot from 2021-06-17 05-52-10

Let's see - LD should be under "one way" not "multi way" since it's not a comparison between >=2 groups of samples... but, "multi way" is referring to [this stuff](https://tskit.dev/tskit/docs/stable/stats.html#multi-way-methods) which doesn't even apply to LD (it's not using the same API yet). So, I'm not sure if we can have the column headers?

But more generally - I'm not sure that having things classified in this summary table up top is helpful - why not just a summary table, like for instance of the methods of a TreeSequence [here](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence)?

Other changes look good.

hyanwong:
> Let's see - LD should be under "one way" not "multi way" since it's not a comparison between >=2 groups of samples...

Yep, sorry - I didn't pay much attention!
> but, "multi way" is referring to [this stuff](https://tskit.dev/tskit/docs/stable/stats.html#multi-way-methods) which doesn't even apply to LD (it's not using the same API yet). So, I'm not sure if we can have the column headers?

I'm not sure it's quite about the API (although I guess we eventually might want to have similar or identical APIs for multi-site and single site stats). I think it's more about the logic. You can have a stat which is defined on a single sample set, or compares between sample sets somehow, and that's true both for multi and single site measures, right?

We might want some way to indicate that the single site stuff has a general API and the multi-site stuff doesn't. But I think having both the terms multi-way and multi-site on the page means that we need some very clear way to make the distinction, and the table does it for me (and, in addition, squeezes it all onto a single screen, which I find helpful as an overview (I can get lost in a long list like the TS methods)

> But more generally - I'm not sure that having things classified in this summary table up top is helpful - why not just a summary table, like for instance of the methods of a TreeSequence [here](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence)?

See above. But I'm very welcome to changes, especially if they mean there's a one-page summary of what's available.

petrelharp:
It's true that this is a distinction we *could* make up top, but I don't think we *need* to (and it's feeling kinda cluttered) - I don't think we need to remind users here that LD works on more than one site, for instance.

hyanwong:
> It's true that this is a distinction we _could_ make up top, but I don't think we _need_ to (and it's feeling kinda cluttered) - I don't think we need to remind users here that LD works on more than one site, for instance.

Fair enough. But then I think we need some wording to make clear the distinction between the terms multi-site and multi-way (both of which we will use on the page). Otherwise I think people will think they are the same thing. Or maybe think of another terminology for one of them?

So up the top, you would prefer something like 

- Single site
  - One-way
    - allele_frequency_spectrum
    - diversity
    - segregating_sites
    - Trait correlations: trait_covariance, trait_correlation, trait_linear_mode
    - Fst
  - Multi-way
    - divergence
    - genetic_relatedness
    - Patterson's f stats: f4, f3, f2
    - Y stats: Y3, Y2
    - Tajima's D
    - GNN
- Multi-site
  - LD
  - IBD

petrelharp:
I think the role of the table up top is mostly as a quick reference, so people can quickly find the stat they are looking for. And I'm a bit worried that saying 'single site' up here will make people think that we're only calculating these things one SNP at a time, so I'm leaning towards a flat table. Do you think that breaking things out in this table will introduce the distinctions?

hyanwong:
Maybe we just need a different terminology for "single/multi site"? Here's a mock-up of the list style. Does that work better for you? At least it fits on one page, so fine by me.

Screenshot 2021-06-17 at 16 39 47

petrelharp:
Heck, that looks good. Maybe add an explanatory sentence like "Single-site statistics are ones that are averages across sites in windows of the genome."

hyanwong:
Updated the PR to the list for and added @petrelharp 's suggestion.

Note that I include GNN here although it isn't in the stats framework yet. It definitely is a single site, multi-way statistic, but it explicitly ignored branch lengths. I guess we could simply explain that in a section, a bit like the one labelled "Y statistics".

It also has a time-windowing feature, which we don't discuss in this documentation (although is that fully pushed now @awohns ?)

petrelharp:
> Note that I include GNN here although it isn't in the stats framework yet. 

I think we were previously not documenting GNN because we were aiming to change the interface to match the general framework? This would be nice to get out of the way.

awohns:
A python version of windowed GNN is in the test suite (with tests),  but not the main API yet. This is something I had on the back burner but I will now make a priority since it looks like it would be a good time to incorporate it into the stats framework as @hyanwong and @petrelharp say.

petrelharp:
that'd be great!

hyanwong:
Yes, it would be a good time @awohns - thanks if you can do it!

awohns:
Hi @hyanwong and @petrelharp, I've become swamped with other commitments for the next month so won't be able to address the time windows (#1237) for a few weeks. Please feel free to go ahead on this without time-windows for now! (or if you want to take a crack it, I can show you what I've done with the time windows so far!)

jeromekelleher:
Looks like there's some good updates here but we got a bit sidetracked - could we finish this up and log issues for anything that can't be done immediately please @hyanwong?

PR #1511, opened 2021-06-22
Fix check offsets

### LINKED ISSUE #1509: metadata overflow causes crash
Over in https://github.com/MesserLab/SLiM/issues/198 @bhaller has tracked down a bug in how `max_metadata_length` is resized - see [this comment](https://github.com/MesserLab/SLiM/issues/198#issuecomment-864493497) for the details.

benjeffery:
Fixes #1509

Turns out this was wrong in every call to `check_offset_overflow`. 

Tested by twiddling the internal state of the table struct. We could probably make large columns, but this seemed simpler.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -406,6 +406,38 @@ out:
 }
 
 static int
+expand_ragged_column(tsk_size_t current_length, tsk_size_t additional_length,
+    tsk_size_t max_length_increment, tsk_size_t *max_length, void **column,
+    size_t element_size)
+{
+    int ret = 0;
+    tsk_size_t increment = TSK_MAX(additional_length, max_length_increment);
+
+    if (check_offset_overflow(current_length, additional_length)) {
+        ret = TSK_ERR_COLUMN_OVERFLOW;
+        goto out;
+    }
+    if ((current_length + additional_length) > *max_length) {
+        if (check_offset_overflow(*max_length, increment)) {
+            /* Here we could allocate to the maximum size.
+             * Instead we are erroring out as this is much easier to test.
+             * The cost is that (at most) the last "max_length_increment"-1
+             * bytes of the possible array space can't be used. */
+            ret = TSK_ERR_COLUMN_OVERFLOW;
+            goto out;
+        }
+
+        ret = expand_column(column, *max_length + increment, element_size);
+        if (ret != 0) {
+            goto out;
+        }
+        *max_length += increment;
+    }
+out:
+    return ret;
+}
+
+static int
 replace_string(
     char **str, tsk_size_t *len, const char *new_str, const tsk_size_t new_len)
 {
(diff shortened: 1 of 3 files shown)


c/tskit/tables.c
```diff
@@ -267,7 +267,7 @@ tsk_individual_table_expand_location(
         = TSK_MAX(additional_length, self->max_location_length_increment);
     tsk_size_t new_size = self->max_location_length + increment;
 
-    if (check_offset_overflow(self->location_length, increment)) {
+    if (check_offset_overflow(self->max_location_length, increment)) {
```
petrelharp, line 270: Since we are checking `max_location_length` instead of `location_length` then the check should go inside of the next `if` statement, where we actually expand `max_location_length`, shouldn't it?

c/tskit/tables.c
```diff
@@ -267,7 +267,7 @@ tsk_individual_table_expand_location(
         = TSK_MAX(additional_length, self->max_location_length_increment);
```
petrelharp, line 267: unrelated, but this looks wrong? I would think the increment should be `MAX(length, min_length)`? Or is `max_location_length_increment` poorly named?

petrelharp: Oh, never mind - I see it's (max_location_length) increment, not max (location_length_increment).

c/tskit/tables.c
benjeffery: Yes! You're right. As it stands now it will error out on the extension one before the overflow.

petrelharp: Note that this is assuming that `location_length` is always less than `max_location_length`, so we don't need to check for overflow on the former. (which is true)

benjeffery: You have to be careful though, as if `self->metadata_length` overflows you don't expand the column at all and crash.

jeromekelleher:
I think it's probably worth abstracting out an ``expand_ragged_column`` here which ``tsk_individual_table_expand_parents`` etc could call, as there's a lot of duplicated code which we could get rid of by using a few pointers. Otherwise we'll be making the same changes a dozen or more times here.

I guess the call would look something like
```c
static int
tsk_individual_table_expand_parents(
    tsk_individual_table_t *self, tsk_size_t additional_length)
{
    return expand_ragged_column(self->max_parents_length_increment, &self->max_parents_length,
        &self->parents);
}
```

benjeffery:
Yep, will do that. Also reduces test duplication which was getting boring!

c/tskit/tables.c
petrelharp: So do we need another check on `location_length`? This is only checking `max_location_length`.

jeromekelleher:
What's the status of this @benjeffery? I think it's worth adding the ``expand_ragged_column`` infrastructure, as it'll be handy in anycase for doing things like setting individual rows.

benjeffery:
It's waiting for me to get back around to it - finishing up an `tsconvert` PR then will get to it today.

benjeffery:
Ok, after some head-scratching this is done. A few queries:

- See the comment in `expand_ragged_columns` instead of erroring when `max_length+increment` overflow we could/should just allocate to the max size - however this can't practically be tested.
- All the `tsk_TABLE_table_expand_RAGGEDCOLUMN` functions are now one-liners that could be removed - I kept them in as `expand_ragged_column` has 6 arguments and the existing calls look cleaner in the higher level methods.
- Shouldn't `tsk_TABLE_table_set_max_COLUMN_length_increment` be `tsk_TABLE_table_set_min_COLUMN_length_increment`??!!

c/tskit/tables.c
```diff
@@ -2990,7 +2901,7 @@ tsk_mutation_table_expand_metadata(tsk_mutation_table_t *self, size_t additional
         = (tsk_size_t) TSK_MAX(additional_length, self->max_metadata_length_increment);
     tsk_size_t new_size = self->max_metadata_length + increment;
 
-    if (check_offset_overflow(self->metadata_length, increment)) {
```
jeromekelleher, line 2993: Shouldn't this be expand_ragged_column also?

jeromekelleher (commented):
LGTM, just need to double-check we're getting all the ragged columns.

jeromekelleher:
> See the comment in expand_ragged_columns. Instead of erroring when max_length+increment overflow we could/should just allocate to the max size - however this can't practically be tested.

Yeah - let's not worry about it though. I guess it'll be become more pressing at some point when we implement the doubling behaviour (#5), when increment will become large quickly. I think it's fine for now, though.

> All the tsk_TABLE_table_expand_RAGGEDCOLUMN functions are now one-liners that could be removed - I kept them in as expand_ragged_column has 6 arguments and the existing calls look cleaner in the higher level methods.

Yeah, let's keep then. More readable this way as you say.

> Shouldn't tsk_TABLE_table_set_max_COLUMN_length_increment be tsk_TABLE_table_set_min_COLUMN_length_increment??!!

I guess ``tsk_TABLE_table_set_COLUMN_length_increment`` would be better again?

c/tskit/tables.c
benjeffery: Whoops, bot sure how I missed that one. Have searched through again to check the others.

benjeffery:
> > Shouldn't tsk_TABLE_table_set_max_COLUMN_length_increment be tsk_TABLE_table_set_min_COLUMN_length_increment??!!
> 
> I guess `tsk_TABLE_table_set_COLUMN_length_increment` would be better again?

Yes, it is a public method, but undocumented so we could change it. However if #5 is planned then not worth it as it will be obsolete then.

jeromekelleher:
>  However if #5 is planned then not worth it as it will be obsolete then.

I think the idea for #5 is to support both the current behaviour, and also doubling. If ``size_increment == 0``, then ``increment = 2 * len(table)``; otherwise, ``increment = size_increment``. So, we can do it simply enough in the current framework, while keeping the flexibility.

benjeffery:
@jeromekelleher I think this is ready to go

jeromekelleher (approved):

jeromekelleher:
This needs some attention @benjeffery - some compiler warnings popping up

Issue #343, opened 2019-08-27
Convert tsk_id_t/tsk_size_t to 64 bit. 


jeromekelleher:
It seems that requiring metadata columns to being < 4G is an unwelcome limitation in forward simulation applications, so we should consider upgrading the metadata_offset columns (and other offset columns) to 64 bit integers. Here is what it would entail:

1. Create a ``tsk_offset_t`` typedef and go through the C tables API making sure this is used for all offset columns (metadata_offset, ancestral_state_offset, etc). Typedef this to uint64_t.
2. Increment the file-format minor version. In the file writing code, look at the last value in each offset array. If it's < UINT32_MAX, store it as a 32 bit value; if not, store as a 64 bit. In the reading code, check which type is being used to store and update accordingly. Note that this means we don't be able to use the current zero-copy behaviour where we use the memory in the kastore to back the arrays. Figure out the best approach here (note, using the memory in the kastore was motivated by using mmap for io, which we've dropped now as it's inherently dangerous and hard to do properly cross-platform).
3. Clean up _tskitmodule.c to use the correct new sizes for numpy arrays (possibly also needing to backport this over to msprime where we use the LightweightTableCollection for interchange).

From a user perspective, this will mean that old versions of tskit won't be able to read newer files. New versions of tskit will continue to read older files without issues, as we're making the code more flexible in terms of expected types.

pinging @petrelharp and @molpopgen for opinions.

jeromekelleher:
Note: arguably we should go the whole hog here and typedef ``tsk_size_t`` to ``uint64_t`` and ``tsk_id_t`` to int64_t. It's less clear that we really do need more than 2*31 rows in a table though, and this will definitely affect memory usage and (possibly) performance. Perhaps not as much as we might think though, so as we're disrupting things it might be as well to try this out and see what the perf implication really are.

petrelharp:
It seems like we need to allow bigger tables (see #402), but on the other hand, the implications for memory usage are pretty unfortunate. The least bad other solution I can think of is to have tables switch over from 32-bit to 64-bit once they get big, although that'd be quite annoying.

jeromekelleher:
> It seems like we need to allow bigger tables (see #402), but on the other hand, the implications for memory usage are pretty unfortunate. The least bad other solution I can think of is to have tables switch over from 32-bit to 64-bit once they get big, although that'd be quite annoying.

[...]

PR #1512, opened 2021-06-22
Change printfs to cast to long-long.

jeromekelleher:
A step towards 64 bit tables (#343), getting rid of some of the compiler warning noise.

I think this is harmless, because we were already casting to (int) in most places already, which was technically wrong. Casting up to long long won't do any harm. I've also added in a couple of other no controversy changes here, as I was looking at it.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -701,11 +701,12 @@ tsk_individual_table_print_state(const tsk_individual_table_t *self, FILE *out)
 
     fprintf(out, TABLE_SEP);
     fprintf(out, "tsk_individual_tbl: %p:\n", (const void *) self);
-    fprintf(out, "num_rows          = %d\tmax= %d\tincrement = %d)\n",
-        (int) self->num_rows, (int) self->max_rows, (int) self->max_rows_increment);
-    fprintf(out, "metadata_length = %d\tmax= %d\tincrement = %d)\n",
-        (int) self->metadata_length, (int) self->max_metadata_length,
-        (int) self->max_metadata_length_increment);
+    fprintf(out, "num_rows          = %lld\tmax= %lld\tincrement = %lld)\n",
+        (long long) self->num_rows, (long long) self->max_rows,
+        (long long) self->max_rows_increment);
+    fprintf(out, "metadata_length = %lld\tmax= %lld\tincrement = %lld)\n",
+        (long long) self->metadata_length, (long long) self->max_metadata_length,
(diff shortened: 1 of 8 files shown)


petrelharp:
Does the airspeed velocity thing still work? It would be great to verify that this sort of thing is indeed harmless...

jeromekelleher:
In this case I have no doubts - the printfs aren't every called in production, so make no difference. The numerical casts are just to silence warnings - these things were being upcast to double anyway.

I'm not sure we did have an ASV setup for tskit?

petrelharp:
> I'm not sure we did have an ASV setup for tskit?

Right, good point. I think it'd be a good thing to prioritize, though.

benjeffery:
> > I'm not sure we did have an ASV setup for tskit?
> 
> Right, good point. I think it'd be a good thing to prioritize, though.

Yep, we never set this up and the one for msprime is offline as the laptop I was running it on had some issues. I'm sure recently someone (maybe @jeromekelleher?) mentioned an alternative benchmarking tool. I can picture it in my head but have failed to remember the name.

c/tskit/tables.c
```diff
@@ -2600,9 +2607,9 @@ tsk_site_table_dump_text(const tsk_site_table_t *self, FILE *out)
         ancestral_state_len
             = self->ancestral_state_offset[j + 1] - self->ancestral_state_offset[j];
         metadata_len = self->metadata_offset[j + 1] - self->metadata_offset[j];
-        err = fprintf(out, "%d\t%f\t%.*s\t%.*s\n", (int) j, self->position[j],
+        err = fprintf(out, "%lld\t%f\t%.*s\t%.*s\n", (long long) j, self->position[j],
             ancestral_state_len, self->ancestral_state + self->ancestral_state_offset[j],
-            metadata_len, self->metadata + self->metadata_offset[j]);
+            (int) metadata_len, self->metadata + self->metadata_offset[j]);
```
benjeffery, line 2612: I'm not getting why `metadata_len` gets a cast but `ancestral_state_len` doesn't.

benjeffery (approved):
LGTM - one query.

jeromekelleher:
Overall, I'm not sure the ASV thing is actually worth the (substantial!) time and effort it takes to make it meaningful.

c/tskit/tables.c
jeromekelleher: Good spot - updated.

PR #1513, opened 2021-06-22
Add row rvs

jeromekelleher:
Stacked on #1512

Adopts the new convention that we use ``ret_id = x_table_add_row`` and then check if this is negative rather than using the ``ret`` value directly like we do now. I think this is a readability improvement anyway, so good to do regardless.


benjeffery (approved):
LGTM!

Issue #1514, opened 2021-06-22
Update dev-docs since conda-minimal.txt is gone


awohns:
Looks like the [requirements section in the development documentation](https://tskit.dev/tskit/docs/stable/development.html#requirements) refers to `python/requirements/conda-minimal.txt` which is now gone. I guess this should be updated to `python/requirements/requirements.txt`?

awohns:
Additionally, using conda compilers (as suggested in the line: `conda install clang_osx-64  clangxx_osx-64` in the docs) doesn't work, but using the built-in mac compiler does (same is true for @hyanwong). Should we also take this out?

hyanwong:
Note that weirdly, the condo compilers *are* required for compiling tsinfer. Go figure...

awohns:
Yep, this is now resolved!

PR #1515, opened 2021-06-22
fixes to macos installation docs

awohns:
Fixes issues mentioned in #1514.
1. The file `python/requirements/conda-minimal.txt` doesn't exist anymore so I delete the line where packages are installed from it.
2. The conda compilers do not work for either @hyanwong or I, while the built-in mac ones do. I delete any mention of the conda compiler option here, but perhaps others think that we should still mention that you can try that option, as it seems that it worked before? I also deleted the bit about supporting conda compilers with Mac versions older than "Mojave", but open to keeping this too. Worth noting for posterity that you **need** the conda compilers for `tsinfer`, as the built in ones don't work there.
3. (I also fix a minor typo in the docs where the word "interface" was duplicated)

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -31,7 +31,7 @@ structure:
 - The ``python`` directory contains the Python library and command line interface,
   which is what most contributors are likely to be interested in. Please
   see the :ref:`sec_development_python` section for details. The
-  low-level :ref:`sec_development_python_c` interface is also defined here.
+  low-level :ref:`sec_development_python_c` is also defined here.
 
 - The ``c`` directory contains the high-performance C library code. Please
   see the :ref:`sec_development_c` for details on how to contribute.
@@ -80,13 +80,8 @@ For OSX and Windows users we recommending using
 `conda <https://docs.conda.io/projects/conda/en/latest/>`__,
 and isolating development in a dedicated environment as follows::
 
-    $ conda create -q -n tskit-dev
-    $ source activate tskit-dev
-    $ conda install -c conda-forge --yes --file=python/requirements/conda-minimal.txt
-    $ conda install -c conda-forge doxygen
-    $ conda install -c bioconda --yes pysam
-    $ pip install -r python/requirements/development.txt
-
+    $ conda env create -f python/requirements/development.yml
+    $ conda activate tskit-dev
 
 On macOS, conda builds are generally done using ``clang`` packages that are kept up to date:
 
@@ -100,14 +95,20 @@ on versions of macOS older than "Mojave":
 
 .. code-block:: bash
 
+    $ cd python
     $ CONDA_BUILD_SYSROOT=/ python3 setup.py build_ext -i
 
 On more recent macOS releases, you may omit the ``CONDA_BUILD_SYSROOT`` prefix.
 
+If you run into issues with the conda compiler, be sure that your command line tools are installed
+and up to date (you should also reboot your system after installing CLI tools). Note that you may
+also have to install a `specific version of the Xcode command line tools
+<https://stackoverflow.com/a/64416852/2752221>`_.
+
 .. note::
 
    The use of the C toolchain on macOS is a moving target.  The above advice
-   was written on 23 January, 2020 and was validated by a few ``tskit`` contributors.
+   was updated on 22 June, 2021 and was validated by a few ``tskit`` contributors.
    Caveat emptor, etc..
 
 
@@ -651,9 +652,7 @@ On Debian/Ubuntu, these can be installed using
     if so, you can install it instead with ``pip`` by running
     ``pip3 install clang-format==6.0.1 && ln -s clang-format $(which clang-format)-6.0``.
 
-Conda users can install the basic requirements as follows::
-
-    $ conda install -c conda-forge ninja meson cunit
+Conda users can install the basic requirements from `python/requirements/development.txt`.
 
 Unfortunately clang-format is not available on conda, but it is not essential.
 
--- a/python/requirements/development.txt
+++ b/python/requirements/development.txt
@@ -8,10 +8,13 @@ flake8
 h5py>=2.6.0
 jsonschema>=3.0.0
 kastore
+# More recent meson is too strict. See #1515.
+meson<=0.56
 msgpack>=1.0.0
 msprime>=1.0.0
 networkx
 newick
+ninja
 numpy
 packaging
 portion
--- a/python/requirements/development.yml
+++ b/python/requirements/development.yml
@@ -1,0 +1,42 @@
+name: tskit-dev
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - autodocsumm==0.1.13 #Pinned as errors on 0.2.0
+  - biopython
+  - black
+  - breathe
+  - codecov
+  - coverage
+  - cunit
+  - doxygen
+  - flake8
+  - h5py>=2.6.0
+  - jsonschema>=3.0.0
+  - kastore
+  - meson<=0.56 # More recent meson is too strict. See #1515.
+  - msprime>=1.0.0
+  - networkx
+  - ninja
+  - numpy
+  - packaging
+  - pip
+  - portion
+  - pre-commit
+  - pyparsing
+  - pysam
+  - pytest
+  - pytest-cov
+  - pytest-xdist
+  - PyVCF
+  - sphinx<=3.3.1
+  - sphinx-argparse
+  - sphinx-issues
+  - sphinx_rtd_theme
+  - svgwrite>=1.1.10
+  - pip:
+    - newick
+    - xmlunittest
+    - msgpack>=1.0.0
+    - python_jsonschema_objects


jeromekelleher:
LGTM, thanks @awohns  - any mac users want to validate this?

benjeffery:
Pinging @tskit-dev/all here to see if anyone can check this on their mac.

Long term I probably need to get an OSX machine myself, but having never used one I'm probably not best placed to figure out the issues.

hyanwong:
I assume it passes CI without the Conda compilers? Or do we not test the meson build stuff on Mac CI?

molpopgen:
I can confirm, but not till after the 5th.

molpopgen:
> The conda compilers do not work for either @hyanwong or I, while the built-in mac ones do.

This is very surprising.

hyanwong:
I think it might be the linker. We get this:

```
ld: warning: -pie being ignored. It is only used when linking a main executable
ld: unsupported tapi file type '!tapi-tbd' in YAML file '/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib/libSystem.tbd' for architecture x86_64
```

molpopgen:
That's a warning, not an error. So things should be working if that's the only output.

molpopgen:
Any chance that one or both of you updated macOS but did not re run the cli tools installer?

molpopgen:
Also, if either of you are using brew, then that greatly complicates things, I'm afraid.  Any macOS  instructions should start from a brand new user account uncontaminated by other package managers.  Yes, it is a pain, but it is the only way to really know what's up

hyanwong:
I think the second line is counting as an error. But actually, I'm not getting it now, only the 

```
ld: warning: -pie being ignored. It is only used when linking a main executable
```

Which doesn't stop compilation. And you are right - the tests are passing now. I did update my CLI tools when helping @awohns so perhaps that was the problem. Perhaps he can verify?

molpopgen:
> I think the second line is counting as an error. But actually, I'm not getting it now, only the 
> 
> ```
> ld: warning: -pie being ignored. It is only used when linking a main executable
> ```
> 
> Which doesn't stop compilation. And you are right - the tests are passing now. I did update my CLI tools when helping @awohns so perhaps that was the problem. Perhaps he can verify?

This is a constant pain point for macOS.  I thought I'd added something about it to the docs?

The first line is common, and can be safely ignored.

hyanwong:
By the way, @molpopgen - do you have any idea why the conda compilers are required on OS X to stop tsinfer from segfaulting? https://github.com/tskit-dev/tsinfer/issues/376#issuecomment-747406463 (fix to avoid mixed compilers in https://github.com/tskit-dev/tsinfer/commit/558f4f767b5b46a76476435b73696a7ee74eb909#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR87). It's been annoying and we haven't quite got to the bottom of it.

Not worth spending much time over, since we have a fix. I'd just like to know.

molpopgen:
> By the way, @molpopgen - do you have any idea why the conda compilers are required on OS X to stop tsinfer from segfaulting? https://github.com/tskit-dev/tsinfer/issues/376#issuecomment-747406463 (fix to avoid mixed compilers in https://github.com/tskit-dev/tsinfer/commit/558f4f767b5b46a76476435b73696a7ee74eb909#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR87). It's been annoying and we haven't quite got to the bottom of it.
> 
> Not worth spending much time over, since we have a fix. I'd just like to know.

I have a guess.  Conda requires that you use conda compilers to compile stuff--the system takes a lot of care to set up loads of shell variables that us old folks had to memorize and do by hand in the past (uphill, in snow, both ways). 

If you use the system compilers, you can have really bad stuff happen like where your tool and it's dependency disagree on what a type looks like, due to differences in compiler and/or linker.  This happens because system compilers have no idea where conda is, and conda is set to ignore the system.  So compiling with the system compilers risks using dependencies external to conda, which could be the wrong version or they "look different" due to differences between compilers. 

All this is true on Linux, too.  If you are compiling stuff that has to work in your conda env, install the compilers.  The conda docs skip this, as it is only developers who have to fight with this.  The majority of conda use is installing stuff that's already compiled.

molpopgen:
> Thanks @molpopgen. Re-running the cli tools installer (`xcode-select --install`) fixed the clang error for me so the conda compilers work now.
> 
> I'll update the PR to retain the conda compiler installation

Seems like the docs should also say (in 96 point bold font) to run that updater.

bhaller:
Here's a stackoverflow that suggests that you might need to install a specific version of the Xcode command-line tools, not the default version, depending upon the macOS version this is happening on: https://stackoverflow.com/a/64416852/2752221

molpopgen:
> @molpopgen sorry for all the noise but, the conda compilers do not in fact seem to be working for me despite upgrading command line tools. I thought they worked but I wasn't in the conda environment where i had installed the compiler. I get the following error when trying to run `make` in the python directory (using the conda environment):
> ```
> ld: warning: -pie being ignored. It is only used when linking a main executable
> ld: unsupported tapi file type '!tapi-tbd' in YAML file '/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/lib/libSystem.tbd' for architecture x86_64
> clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
> error: command '/Users/awohns/miniconda3/envs/tskit-dev/bin/x86_64-apple-darwin13.4.0-clang' failed with exit code 1
> ```

Did you do a "meson clean" and "build" before trying to make? If you didn't clean out the previous build system, then was still trying to look for all of the wrong compilers, I think...

awohns:
> Here's a stackoverflow that suggests that you might need to install a specific version of the Xcode command-line tools, not the default version, depending upon the macOS version this is happening on: https://stackoverflow.com/a/64416852/2752221

Thanks @bhaller, I think this was the issue! I tried installing xcode command line tools version 11.5 and was able to run `make` in the `python` directory without issue. I can add a note to the documentation about this.

@molpopgen, I wasn't able to run meson at all before this so I don't think that was the issue. With the downgraded command line tools, and after installing meson version 0.57 (as noted in [this msprime issue](https://github.com/tskit-dev/msprime/issues/1485)), I can now successfully run `meson build` in the `c` directory. @jeromekelleher, meson is not in the requirements right now, but I guess we could add it, with the pinned version?

jeromekelleher:
Yes, I think we should pin meson in the requirements.txt, this is done now in msprime AFAIK

awohns:
Ok, I've added meson, ninja, and cunit to the requirements, along with noting the command line tools issue. I think this should be good to go if it looks alright to all?

python/requirements/development.txt
```diff
@@ -4,14 +4,18 @@ black
 breathe
 codecov
 coverage
+cunit
```
benjeffery, line 7: `cunit` isn't in PyPI so this fails for `pip` users.

benjeffery (commented):
Looks all good for OSX - but there is an issue with `pip` based toolchain. We'll need a seperate `conda-development.txt` file.

benjeffery:
@awohns do you want me to wrap this one up?

awohns:
@benjeffery, so there are also a number of packages that aren't on conda-forge or bioconda (see below). So for pip, I guess we should have a `development-pip.txt` file that has everything except cunit (which can be installed with conda) and then for conda we have a `development-conda.txt` for everything that can be installed on conda (and then instructions to pip install the packages listed below?).

```
PackagesNotFoundError: The following packages are not available from current channels:

  - newick
  - xmlunittest
  - msgpack[version='>=1.0.0']
  - python_jsonschema_objects
```

molpopgen:
I may be wrong, but isn't there a way to specify a pip install in the conda
yaml file?

benjeffery:
> I may be wrong, but isn't there a way to specify a pip install in the conda yaml file?

Yep, example here: https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually

docs/development.rst
```diff
@@ -82,7 +82,6 @@ and isolating development in a dedicated environment as follows::
 
     $ conda create -q -n tskit-dev
     $ source activate tskit-dev
```
mufernando, line 84: should replace `source` with `conda` here.

docs/development.rst
```diff
@@ -104,10 +103,14 @@ on versions of macOS older than "Mojave":
 
 On more recent macOS releases, you may omit the ``CONDA_BUILD_SYSROOT`` prefix.
 
+If you run into issues with the conda compiler, be sure that your command line tools are installed
```
mufernando, line 106: also might be worth it to reboot your system after reinstalling CLI tools

docs/development.rst
```diff
@@ -82,7 +82,6 @@ and isolating development in a dedicated environment as follows::
 
     $ conda create -q -n tskit-dev
     $ source activate tskit-dev
-    $ conda install -c conda-forge --yes --file=python/requirements/conda-minimal.txt
     $ conda install -c conda-forge doxygen
     $ conda install -c bioconda --yes pysam
```
mufernando, line 86: can't we install doxygen and pysam all at once with:
`conda -c conda-forge -c bioconda doxygen pysam`

mufernando:
I just tested the contents of the PR and it worked perfectly for me. I made some comments to clarify/simplify some things, but LGTM overall.

what else do we need to merge this?

docs/development.rst
```diff
@@ -82,7 +82,6 @@ and isolating development in a dedicated environment as follows::
 
     $ conda create -q -n tskit-dev
```
mufernando, line 83: also, can we remove all the $ so that it is easier to paste it all at once?

benjeffery:
I think this was waiting for someone with OSX to confirm it all, which you've just done! @awohns or @mufernando do you have time to make the edits to the PR suggest above or shall I jump in?

awohns:
I can finish it up!

awohns:
Ok, I've changed the mac pipeline to use a yml file called `development.yml`. This works for me and is simpler than what we had before.
I think it's ready to merge if this looks good to others!

benjeffery (approved):
Wonderful, thanks for seeing this through @awohns. :star:
I'll merge now as it is a real improvement and any possible issues can be fixed in subsequent PRs.

PR #1516, opened 2021-06-23
Use ret_num_trees for check integrity return

benjeffery:
Similarly to #1513

### DIFF
--- a/c/tests/test_minimal_cpp.cpp
+++ b/c/tests/test_minimal_cpp.cpp
@@ -150,7 +150,7 @@ test_edge_sorting()
     tsk_table_collection_t tables;
     tsk_id_t n = 10;
     tsk_id_t j;
-    tsk_id_t ret = tsk_table_collection_init(&tables, 0);
+    int ret = tsk_table_collection_init(&tables, 0);
     assert(ret == 0);
 
     tables.sequence_length = 1.0;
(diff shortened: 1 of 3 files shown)


jeromekelleher:
I'm not sure about this, because check_integrity only returns the number of trees **if we ask it to**. So, in most of these cases, the variable name will be actively misleading. I think we should use ``ret_num_trees`` in the cases where it is computing the number of trees, and cast to int in all the other cases, with a comment to the effect of "Casting to int is safe here because we do not compute the number of trees".

jeromekelleher:
It was an unfortunate API choice, to be honest, in retrospect maybe we should have added another function or something.

benjeffery:
> I'm not sure about this, because check_integrity only returns the number of trees **if we ask it to**. So, in most of these cases, the variable name will be actively misleading. I think we should use `ret_num_trees` in the cases where it is computing the number of trees, and cast to int in all the other cases, with a comment to the effect of "Casting to int is safe here because we do not compute the number of trees".

Ah yes, I had forgotten that the number of trees is not always returned.

benjeffery:
@jeromekelleher now casting in all calls, except where TSK_CHECK_TREES is specified, where we use `ret_num_trees`

c/tests/test_minimal_cpp.cpp
```diff
@@ -168,7 +168,9 @@ test_edge_sorting()
     assert(tables.edges.num_rows == (tsk_size_t) n - 1);
 
     /* Make sure the edges are unsorted */
-    ret = tsk_table_collection_check_integrity(&tables, TSK_CHECK_EDGE_ORDERING);
+    /* Not calling TSK_CHECK_TREES so casting is safe */
+    ret = (tsk_id_t) tsk_table_collection_check_integrity(
```
jeromekelleher, line 172: We don't need to cast here, ret is declared as a tsk_id_t and the return value is already tsk_id_t. I think we should change the type of ``ret`` to int here (I probably changed this, before I realised it was a mistake to change ``ret`` values to tsk_id_t

jeromekelleher (commented):
LGTM, but the C++ version is wrong

c/tests/test_minimal_cpp.cpp
benjeffery: Good catch, I've made `ret` an `int` here.

jeromekelleher (approved):

PR #1517, opened 2021-06-23
Replace direct references to sizes

benjeffery:
Almost all of these direct references to 32 should have been `tsk_X` anyway. 

I've also modified the overflow checkers to not need a wider type - please double check the logic here!

I've also `#defined` the kastore types - I know we'll have to do something more clever than just changing them but it still seemed a good thing to do.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -410,14 +410,14 @@ test_node_table(void)
     tsk_id_t ret_id;
     tsk_node_table_t table, table2;
     tsk_node_t node, node2;
-    uint32_t num_rows = 100;
+    tsk_size_t num_rows = 100;
     tsk_id_t j;
-    uint32_t *flags;
+    tsk_flags_t *flags;
     tsk_id_t *population;
     double *time;
     tsk_id_t *individual;
     char *metadata;
-    uint32_t *metadata_offset;
+    tsk_size_t *metadata_offset;
     const char *test_metadata = "test";
     tsk_size_t test_metadata_length = 4;
     char metadata_copy[test_metadata_length + 1];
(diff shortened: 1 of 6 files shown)


c/tests/test_tables.c
```diff
@@ -491,18 +491,18 @@ test_node_table(void)
     CU_ASSERT_EQUAL(table.metadata_length, 0);
 
     num_rows *= 2;
-    flags = malloc(num_rows * sizeof(uint32_t));
+    flags = malloc(num_rows * sizeof(tsk_flags_t));
     CU_ASSERT_FATAL(flags != NULL);
-    memset(flags, 1, num_rows * sizeof(uint32_t));
-    population = malloc(num_rows * sizeof(uint32_t));
+    memset(flags, 1, num_rows * sizeof(tsk_flags_t));
+    population = malloc(num_rows * sizeof(tsk_size_t));
```
jeromekelleher, line 497: This and all the other id references below should be ``tsk_id_t`` I think.  Better again to use ``sizeof(*population)``, but that'd be more fiddly to do and not worth the effort I think.

c/tskit/tables.c
```diff
@@ -71,8 +70,7 @@ check_table_overflow(tsk_size_t current_size, tsk_size_t additional_rows)
 static bool
 check_offset_overflow(tsk_size_t current_size, tsk_size_t additional_elements)
 {
-    uint64_t new_size = (uint64_t) current_size + (uint64_t) additional_elements;
```
jeromekelleher, line 74: Let's not change the logic here yet, we know this works. Your version should work, but I can see some corner cases where it might fall over, so let's not change too many things at once.

c/tskit/tables.c
```diff
@@ -919,19 +918,20 @@ tsk_individual_table_load(tsk_individual_table_t *self, kastore_t *store)
         metadata_schema_length;
 
     read_table_col_t read_cols[] = {
-        { "individuals/flags", (void **) &flags, &num_rows, 0, KAS_UINT32, 0 },
+        { "individuals/flags", (void **) &flags, &num_rows, 0, TSK_FLAG_STORAGE_TYPE,
```
jeromekelleher, line 921: Nit - a single "flag" is a bit, I think we should call it ``TSK_FLAGS_STORAGE_TYPE`` to be consistent

c/tskit/tables.h
```diff
@@ -62,6 +64,8 @@ Sizes in tskit are defined by the ``tsk_size_t`` type.
 @endrst
 */
 typedef uint32_t tsk_size_t;
+#define TSK_MAX_SIZE (tsk_size_t) UINT32_MAX
```
jeromekelleher, line 67: Maybe put in some extra brackets, to be save, ``((tsk_size_t) UINT32_MAX)``

jeromekelleher (commented):
This is excellent, thanks @benjeffery! A couple of minor points.

c/tskit/tables.c
benjeffery: Ok, I'll save this patch for the 64bit branch. I don't think we can guarantee that a 128bit type is available so that branch will need to use something like this.

c/tests/test_tables.c
benjeffery: I've switched them to the right types - perils of search and replace!

c/tskit/tables.c
benjeffery: Yep, good one. Fixed.

c/tskit/tables.h
benjeffery: I've removed the cast - not sure why it was there.

benjeffery:
@jeromekelleher All fixed up I think.

c/tskit/tables.c
jeromekelleher: It's pretty academic at that point though, no conceivable computer could overflow a 64 bit number.

jeromekelleher (approved):

PR #1518, opened 2021-06-23
Final type width cleanups

benjeffery:
Some (hopefully) final typing clean ups that let us switch 64/32 in minimal places.

PR #1519, opened 2021-06-23
64 bit test branch

benjeffery:
Not for merging.

Stacked on #1516, #1517 and #1518.

@bhaller this branch passes all the C tests and has 64bit wide ids and sizes. Will be interesting to see SLiM performance with this code!

To change from 32-64 only the following changes are needed: 

It is also possible to keep 32bit ids but have 64bit sizes, which would fix the metadata issue, but avoid table bloat.

### DIFF
--- a/c/examples/error_handling.c
+++ b/c/examples/error_handling.c
@@ -19,9 +19,9 @@ main(int argc, char **argv)
         fprintf(stderr, "%s", tsk_strerror(ret));
         exit(EXIT_FAILURE);
     }
-    printf("Loaded tree sequence with %d nodes and %d edges from %s\n",
-        tsk_treeseq_get_num_nodes(&ts),
-        tsk_treeseq_get_num_edges(&ts),
+    printf("Loaded tree sequence with %lld nodes and %lld edges from %s\n",
+        (long long)tsk_treeseq_get_num_nodes(&ts),
+        (long long)tsk_treeseq_get_num_edges(&ts),
         argv[1]);
     tsk_treeseq_free(&ts);
 
--- a/c/examples/haploid_wright_fisher.c
+++ b/c/examples/haploid_wright_fisher.c
@@ -58,19 +58,19 @@ simulate(
             children[j] = child;
         }
         if (t % simplify_interval == 0) {
-            printf("Simplify at generation %d: (%d nodes %d edges)",
-                t,
-                tables->nodes.num_rows,
-                tables->edges.num_rows);
+            printf("Simplify at generation %lld: (%lld nodes %lld edges)",
+                (long long)t,
+                (long long)tables->nodes.num_rows,
+                (long long)tables->edges.num_rows);
             /* Note: Edges must be sorted for simplify to work, and we use a brute force
              * approach of sorting each time here for simplicity. This is inefficient. */
             ret = tsk_table_collection_sort(tables, NULL, 0);
             check_tsk_error(ret);
             ret = tsk_table_collection_simplify(tables, children, N, 0, NULL);
             check_tsk_error(ret);
-            printf(" -> (%d nodes %d edges)\n",
-                tables->nodes.num_rows,
-                tables->edges.num_rows);
+            printf(" -> (%lld nodes %lld edges)\n",
+                (long long)tables->nodes.num_rows,
+                (long long)tables->edges.num_rows);
             for (j = 0; j < N; j++) {
                 children[j] = j;
             }
--- a/c/examples/streaming.c
+++ b/c/examples/streaming.c
@@ -24,8 +24,8 @@ main(int argc, char **argv)
             break;
         }
         check_tsk_error(ret);
-        fprintf(stderr, "Tree sequence %d had %d mutations\n", j,
-            (int) tables.mutations.num_rows);
+        fprintf(stderr, "Tree sequence %d had %lld mutations\n", j,
+            (long long) tables.mutations.num_rows);
         ret = tsk_mutation_table_truncate(&tables.mutations, 0);
         check_tsk_error(ret);
         ret = tsk_table_collection_dumpf(&tables, stdout, 0);
--- a/c/examples/tree_iteration.c
+++ b/c/examples/tree_iteration.c
@@ -26,17 +26,17 @@ main(int argc, char **argv)
 
     printf("Iterate forwards\n");
     for (iter = tsk_tree_first(&tree); iter == 1; iter = tsk_tree_next(&tree)) {
-        printf("\ttree %d has %d roots\n",
-            tsk_tree_get_index(&tree),
-            tsk_tree_get_num_roots(&tree));
+        printf("\ttree %lld has %lld roots\n",
+            (long long)tsk_tree_get_index(&tree),
+            (long long)tsk_tree_get_num_roots(&tree));
     }
     check_tsk_error(iter);
 
     printf("Iterate backwards\n");
     for (iter = tsk_tree_last(&tree); iter == 1; iter = tsk_tree_prev(&tree)) {
-        printf("\ttree %d has %d roots\n",
-            tsk_tree_get_index(&tree),
-            tsk_tree_get_num_roots(&tree));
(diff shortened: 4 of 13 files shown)


bhaller:
I'll have a look at this soon, thanks.  @petrelharp, I should be able to simply copy these files into SLiM's code, is that correct?  Do we have any patches relative to tskit right now?

jeromekelleher:
I can confirm the C library is compiling without warnings, all our tests are passing and valgrind is happy. I haven't tried pushing the limits, though I guess the simplest way to do this would be to make a small C program that wrote lot of rows to the node table (definitely with a suitably large row chunk size!)

petrelharp:
Nope, no patches - just paste it in.

bhaller:
`tsk_strerror()` should perhaps take a 64-bit int, so that when one calls a function that returns tsk_id_t, one does not have to reduce the result to a 32-bit int to pass it to `tsk_strerror()`?

I am not at all a fan of unsigned types in C/C++; I think they are a source of bugs with little positive to recommend them.  If you folks agree with that opinion, would this PR offer an opportunity to switch to int64_t rather than uint64_t?  Certainly I don't imagine that we actually need the extra bit of headroom.  @jeromekelleher @petrelharp @molpopgen 

Running tests now, seems to work, will post a performance summary on issue #343.

jeromekelleher:
Changing from unsigned to signed for size types is an orthogonal change @bhaller, which would be very noisy from our perspective, with not a lot of benefit. Please do open an issue to discuss if you think this is something we should seriously consider for the C API version 1.0.

> tsk_strerror() should perhaps take a 64-bit int, so that when one calls a function that returns tsk_id_t, one does not have to reduce the result to a 32-bit int to pass it to tsk_strerror()?

I don't think we should do this, because it's only the ``x_table_add_row`` functions that return tsk_id_t and all other functions return an int (with one exception, I think). So, we'd be optimising in the wrong direction there, IMO.

bhaller:
> Changing from unsigned to signed for size types is an orthogonal change @bhaller, which would be very noisy from our perspective, with not a lot of benefit. Please do open an issue to discuss if you think this is something we should seriously consider for the C API version 1.0.

Nah, just thought I'd test the waters.  Doesn't sound like a good idea.

> > tsk_strerror() should perhaps take a 64-bit int, so that when one calls a function that returns tsk_id_t, one does not have to reduce the result to a 32-bit int to pass it to tsk_strerror()?
> 
> I don't think we should do this, because it's only the `x_table_add_row` functions that return tsk_id_t and all other functions return an int (with one exception, I think). So, we'd be optimising in the wrong direction there, IMO.

Hmm.  So when one calls and `x_table_add_row` function and gets a `tsk_id_t`, what is the protocol, then?  Test whether the return value is `< 0`, and if it is that indicates an error; and then what?  Is it guaranteed to be safe to cast down to `int` in order to call `tsk_strerror()`?  Is doing so considered to be good/standard practice?

Mostly unrelated: In SLiM we often have methods that make a series of calls to tskit.  In the past, we have just declared `int ret;` at the top, and then done each call to tskit with `ret = ...`, testing `ret` for an error at each step.  Now, since the return value from the `x_table_add_row` functions no longer fits in `int`, that paradigm needs to shift in an uglier direction.  I'm not sure that this is a *problem*, and I have no proposed fix for it; I just thought I'd note that I encountered it as an issue when porting SLiM to the 64-bit tskit.

Other than these two issues (`tsk_strerror()` and `int ret;`) the port was smooth.

molpopgen:
> Mostly unrelated: In SLiM we often have methods that make a series of calls to tskit.  In the past, we have just declared `int ret;` at the top, and then done each call to tskit with `ret = ...`, testing `ret` for an error at each step.  Now, since the return value from the `x_table_add_row` functions no longer fits in `int`, that paradigm needs to shift in an uglier direction.  I'm not sure that this is a *problem*, and I have no proposed fix for it; I just thought I'd note that I encountered it as an issue when porting SLiM to the 64-bit tskit.
> 

This is a strong argument for the AAA rule: "almost always auto" (coined by Herb Sutter--his blog goes into many deep dives on this).  Just change int to auto and be done.

bhaller:
> > Mostly unrelated: In SLiM we often have methods that make a series of calls to tskit.  In the past, we have just declared `int ret;` at the top, and then done each call to tskit with `ret = ...`, testing `ret` for an error at each step.  Now, since the return value from the `x_table_add_row` functions no longer fits in `int`, that paradigm needs to shift in an uglier direction.  I'm not sure that this is a _problem_, and I have no proposed fix for it; I just thought I'd note that I encountered it as an issue when porting SLiM to the 64-bit tskit.
> 
> This is a strong argument for the AAA rule: "almost always auto" (coined by Herb Sutter--his blog goes into many deep dives on this). Just change int to auto and be done.

Well, it's not really quite that simple, because of `tsk_strerror()`.  If `auto` results in `ret` being `int64_t`, then *every* call to `tsk_strerror()` will need to include an explicit down-cast from `auto` (i.e. `int64_t`) to `int`, and all the return values that are `int` will be getting cast upward to `int64_t` and then back down to `int` (eww).  That's the "uglier direction" I referred to.  :->  For that uglification to go away, either `tsk_strerror()` needs to accept `tsk_id_t`, or the return values from the `x_table_add_row` functions need to be handled using a different local variable, not `ret` (which is still a little uglier than the 32-bit paradigm).

bhaller:
I'm also not particularly a fan of `auto`, as it leads to code that is harder to read and might behave in unexpected ways.  Compile-time type-checking is a wonderful thing, IMHO, and using `auto` can circumvent that type-checking and produce incorrect behavior as a result, by "magically" selecting the wrong/unintended type for you.  I haven't read Sutter's blog, but my gut reaction to "almost always auto" is "wow, I'm glad I don't have to maintain *his* code" :->.  Anyhow, that's a side topic that we needn't/shouldn't discuss in this issue.  :->

jeromekelleher:
> Hmm. So when one calls and x_table_add_row function and gets a tsk_id_t, what is the protocol, then? Test whether the return value is < 0, and if it is that indicates an error; and then what? Is it guaranteed to be safe to cast down to int in order to call tsk_strerror()? Is doing so considered to be good/standard practice?

All good points @bhaller, and something I've been scratching my head about over the last few days. 
The convention we've adopted is this:
```
ret_id = x_table_add_row(...)
if (ret_id < 0) {
    ret = (int) ret_id; // guaranteed safe; negative values are only ever error codes
    goto out;
}
```
I agree it's a change from the normal conventions and is a bit ugly. It does nicely emphasise that the value you get back is an ID, unless an error occurs.  It's a balance between the conveniece of getting the ID of the newly added row back (which is often handy and does avoid annoying bugs), and keeping with the convention that the return value is an error marker only. I guess we could add a ``row_id`` pointer to the function call, but this seems more ugly to me.

bhaller:
> > Hmm. So when one calls and x_table_add_row function and gets a tsk_id_t, what is the protocol, then? Test whether the return value is < 0, and if it is that indicates an error; and then what? Is it guaranteed to be safe to cast down to int in order to call tsk_strerror()? Is doing so considered to be good/standard practice?
> 
> All good points @bhaller, and something I've been scratching my head about over the last few days.
> The convention we've adopted is this:
> 
> ```
> ret_id = x_table_add_row(...)
> if (ret_id < 0) {
>     ret = (int) ret_id; // guaranteed safe; negative values are only ever error codes
>     goto out;
> }
> ```
> 
> I agree it's a change from the normal conventions and is a bit ugly. It does nicely emphasise that the value you get back is an ID, unless an error occurs. It's a balance between the conveniece of getting the ID of the newly added row back (which is often handy and does avoid annoying bugs), and keeping with the convention that the return value is an error marker only. I guess we could add a `row_id` pointer to the function call, but this seems more ugly to me.

OK, if that's the standard paradigm SLiM will adopt it (assuming we go to 64-bit at all).  Thanks for the clarification.

benjeffery:
Closing this PR now that we have decided a path forward.

PR #1520, opened 2021-06-23
Y lab rotation & tick label tweaks

hyanwong:
## Description

This fixes a minor "bug" so that viewing a default SVG in Inkscape will properly rotate the Y axis label

It also tidies up the grouping of tick marks so they are more consistent with the other labels, and can be rotated via css without annoying pixel tweaking. Apologies about all the updated SVG files @benjeffery - now that some of these examples are in the viz tutorial, we might be able to offload the visual checking to that, I suppose.


hyanwong:
Probably not worth adding anything to the changeling for this TBH.

jeromekelleher (approved):

benjeffery (approved):

PR #1524, opened 2021-06-23
Switch to typed hints for draw_svg and document y_ticks

hyanwong:
Simply updates the documentation for the draw_svg parameter `y_ticks`, but also switches to type hints for those 2 functions, and adds the sphinx module needed to use these.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -34,8 +34,11 @@ import textwrap
 import warnings
 from dataclasses import dataclass
 from typing import Any
+from typing import List
+from typing import Mapping
 from typing import NamedTuple
 from typing import Optional
+from typing import Tuple
 from typing import Union
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


python/tskit/trees.py
```diff
@@ -1589,101 +1592,104 @@ def draw_text(
...
-        symbol_size=None,
-        x_axis=None,
-        x_label=None,
-        y_axis=None,
-        y_label=None,
-        y_ticks=None,
-        y_gridlines=None,
-        all_edge_mutations=None,
+        size: Optional[Tuple[int, int]] = None,
+        time_scale: "TreeSequence.SvgTimeScale" = None,
```
jeromekelleher, line 1605: I think we want to use ``from __future__ import typing`` to avoid some of this awkwardness with string types.

jeromekelleher (commented):
I'm basically on board with this. Two things:

1. What can we do if we do the ``from __future__ import typing`` thing?
2. Can you paste in a screen show of how the rendered docs look before and after, so we can make an informed choice about the changes?

python/tskit/trees.py
hyanwong: Is that `from __future__ import annotations`? I think it's python 3.7, so do we want that as a minimum version?

hyanwong:
Here's a screenshot

Screenshot 2021-06-24 at 10 10 09


I must say, I do find the specification for functions a bit clunky now: essentially all params will need to be on a separate line for any method with more than 3 or so params, and I find it makes the concept of what the method does a little less easy to grok immediately. But I guess I'll get used to it. I assume the pros outweigh the cons.

python/tskit/trees.py
jeromekelleher: It's already in msprime

hyanwong:
> 1. What can we do if we do the `from __future__ import typing` thing?

It seems not to help in the doc building phase, AFAICS. When I do that and remove the quotes I get e.g.

```
Cannot resolve forward reference in type annotations of "tskit.TreeSequence.draw_svg": name 'SvgOrder' is not defined
```

and also things like

```
tskit/trees.py:docstring of tskit.Tree.draw_svg:: WARNING: py:class reference target not found: SvgMaxTime
```

I guess this is something to do with Sphinx, but maybe @benjeffery knows more?

jeromekelleher:
Maybe we should decouple this discussion from the actual API changes you wanted to make @hyanwong? Seems like this is going to be a significant tooling change.

hyanwong:
> Maybe we should decouple this discussion from the actual API changes you wanted to make @hyanwong? Seems like this is going to be a significant tooling change.

Sure. It's not an API change anyway, it's just documenting that a param can be a dict as well as a list (it could be before, but was undocumented). Hence this all being about documentation.

jeromekelleher:
Right - I don't feel like taking on a tooling change right now, so I think it's probably best to drop the typehints stuff for the moment.

hyanwong:
Sure. I think @benjeffery has plans for automating all this stuff, so I thought this would be a useful way to dip a toe in the waters. I guess leaving this issue open, but deferring it 'til later is the best option?

benjeffery:
My preference would be to close this PR, don't delete the branch though, and open an issue with the proposal.

PR #1531, opened 2021-06-28
Add _TSK_BIG_TABLES typedef.

### LINKED ISSUE #1528: Add compile flag for supporting 64 bit IDs
The discussion in #343 makes it clear that it is likely that we will need to support very large tables at some point in the future, and so it would be wise to plan ahead for this. My suggestion for this is the following:

- Add some defines that allow us to build the tskit library with 64 bit IDs (default is definitely 32 bit). Do not change any of the storage types in the file format, this is purely just about the size of the tsk_id_t and tsk_size_t types. Do not document this, and state clearly in the source code that this is not a supported configuration.
- Add a step to the CircleCI config that builds the C library like this and runs the tests.

This way we'll be able to keep the library in 64bit readiness and should be able to make the jump  fairly easily if/when we need to. Running the tests in 64bit mode in CI should prevent us from going down any 32 bit only rabbit holes.

Any objections?

jeromekelleher:
Closes #1528

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,12 @@ jobs:
             ninja -C build-clang
 
       - run:
+          name: Compile C with gcc in 64 bit mode
+          command: |
+            CFLAGS=-D_TSK_BIG_TABLES CPPFLAGS=-D_TSK_BIG_TABLES meson build-gcc-64 c
+            ninja -C build-gcc-64
+
+      - run:
           name: Run C tests
           command: |
             ninja -C build-gcc test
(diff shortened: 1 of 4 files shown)


benjeffery (approved):
Awesome! Looks good to me - you mentioned you had some concerns about this way of overflow checking, are you happy with this now?

jeromekelleher:
> Looks good to me - you mentioned you had some concerns about this way of overflow checking, are you happy with this now?

I thought some more about it and you're right, it is much better! I got rid of the checks for > 0 since these are unsigned types, but basically it's what you proposed.

I'm slightly worried about the fact we don't have any tests for the boundary behaviour for num_rows though, I tried leaving out the "+1" and all the tests still passed.

benjeffery:
Yep, added an issue for that https://github.com/tskit-dev/tskit/issues/1532

jeromekelleher:
@mergifyIO rebase

jeromekelleher:
Yep, this is working on CircleCI, just verified it's definitely running in 64 bit mode by checking the total memory used reported in valgrind (*much* more in 64 bit mode).

PR #1540, opened 2021-06-29
Use Hartigan parsimony for the standard test

hyanwong:
Just spotted this @jeromekelleher - I think in all the "normal" parsimony tests we should be testing the Hartigan algorithm, rather than the fitch one, no?


jeromekelleher:
I don't think it really matters here as it's just testing the test-code, and they are binary alleles. Probably it should verify that both the local hartigan and fitch algorithms raise errors in this case?

benjeffery:
Yeah, looks like both should be tested.

hyanwong:
Done

benjeffery (approved):

PR #1541, opened 2021-06-29
Clarify+test parsimony placement with unary nodes

### LINKED ISSUE #1078: map_mutations does not conform to documented behaviour with unary nodes
Spotted in https://github.com/tskit-dev/tsinfer/pull/401. The `map_mutations` documentation says "For each mutation, `node` is the tree node at the bottom of the branch on which the transition occurred,", but actually the mutation is placed on the highest tree node on the branch on which the transition occurred. Here's a test case which I guess might be useful for a unit test: I was expecting the mutation to be associated with the leaf node 0, but it's actually been placed on the unary node above that (node 3)

```
import tskit
import io

nodes = io.StringIO(
    """\
id  is_sample   time    population
0   1   0.000000    0   -1
1   1   0.000000    0   -1
2   1   0.000000    0   -1
3   0   1.000000    0   -1
4   0   1.000000    0   -1
5   0   2.000000    0   -1
"""
)
edges = io.StringIO(
    """\
left    right   parent  child
0.000000    1.000000    3  0
0.000000    1.000000    4  1
0.000000    1.000000    5  2
0.000000    1.000000    5  3
0.000000    1.000000    5  4
"""
)

ts = tskit.load_text(nodes, edges, sequence_length=1, strict=False)
print(ts.first().draw_text())
print(ts.first().map_mutations(genotypes = [1, 0, 0], alleles=["0", "1"]))
```

gives:

```
  5  
┏━╋━┓
3 4 ┃
┃ ┃ ┃
0 1 2
('0', [{'id': -1, 'site': -1, 'node': 3, 'time': nan, 'derived_state': '1', 'parent': -1, 'metadata': b''}])
```

I guess we either need to change the documentation or the `map_mutations()` behaviour. I wonder what the most logical thing to do here is?

hyanwong:
Fixes #1078

### DIFF
--- a/python/tests/test_parsimony.py
+++ b/python/tests/test_parsimony.py
@@ -1172,6 +1172,84 @@ class TestParsimonyExamplesBalancedTernary(TestParsimonyBase):
         assert len(transitions) == 18
 
 
+class TestParsimonyExamplesUnary(TestParsimonyBase):
+    """
+    Some examples on a tree with unary nodes. The mutation should be placed
+    on the highest node along the lineage compatible with the parsimonious placement
+    """
+
+    #        9
+    #      ┏━┻━┓
+    #      8   ┃
+    #    ┏━┻━┓ ┃
+    #    6   7 ┃
+    #    ┃   ┃ ┃
+    #    5   ┃ ┃
+    #  ┏━╋━┓ ┃ ┃
+    #  0 2 3 1 4
+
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
I think this is guaranteed because the mutations are allocated on the downward pass, so the first node we encounter will be chosen.

Issue #1542, opened 2021-06-29
Allow fixed ancestral states / internal samples in map_mutations


hyanwong:
A few times now I have wanted to be able to map mutations parsimoniously onto a tree *while specifying an ancestral state X*. This is identical to setting the root node as a sample with state X.

It's unclear to me from the documentation or the tests whether the Hartigan algorithm is guaranteed to give the most parsimonious reconstruction when internal nodes are allowed to be samples. The Sankoff-score-based implementation (which I think is simply used to test the Hartigan method) has a comment here:

https://github.com/tskit-dev/tskit/blob/7d3e47a55e3f6fec5e2602ec1e700d82b0a309b8/python/tests/test_parsimony.py#L285

And there's a test of a tree with internal samples here:

https://github.com/tskit-dev/tskit/blob/7d3e47a55e3f6fec5e2602ec1e700d82b0a309b8/python/tests/test_parsimony.py#L589

which checks that the method does at least reconstruct the correct internal states. But I can't see any assurance that it's the most parsimonious reconstruction. Do you know @jeromekelleher ? If so, then setting an ancestral state is simply the same as changing the root to be a sample with a given state, so it would just be a question of whether we should create an interface to allow this to be specified (e.g. `Tree.map_mutations(genotypes, alleles, ancestral_state=None)` where `ancestral_state` could optionally be set to a specific allelic value from among the alleles (or a genotype value, I guess).

jeromekelleher:
It's a bug if it's not a parsimonious reconstruction, and I'm fairly sure that it's correct when you have internal samples. You could specify the ancestral state in the way you say.

The Sankoff stuff is just used for testing a part of the API, I didn't bother making it more complicated since it's not part of the API.

hyanwong:
Great, thanks Jerome. It would be nice to document that this works with internal samples (and I might add a test case just to check).

Do you think it is worth adding an `ancestral_state` argument to `map_mutations`, or is that over complicating the interface (and hence we would expect users to do the internal-sample-setting thing if they also need to do what I need to)?

jeromekelleher:
I think it's probably a good idea to add an ``ancestral_state`` argument, but I haven't looked at the implementation so not sure how tricky it'll be.

hyanwong:
OK, so I'll leave this issue open and have a look at the implementation myself.

jeromekelleher:
[...]

PR #1550, opened 2021-07-01
Add ancestral_state to map_mutations

hyanwong:
Starts to address #1542. Currently this is a draft version of a change to the python-only Hartigan `map_mutations` code. I *think* this is right, but it could do with checking and also we should add a few round-trip tests with ancestral state specified (I'm not sure how comprehensive to make this, e.g. whether to patch it in to the round trip tests higher up).

The doesn't code up the C version, but I guess that should be easy. It also doesn't add the ancestral state function to the fitch map_mutations code, but we don't use that except for testing anyway.

I guess it's OK that the algorithm is allowed to put a mutation above the root, so that the ancestral state for the whole tree switches immediately?

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -2936,6 +2936,23 @@ test_simplest_map_mutations(void)
     CU_ASSERT_EQUAL_FATAL(transitions[0].state, 1);
     free(transitions);
 
+    /* Assign the ancestral_state */
+    genotypes[0] = 1;
+    genotypes[1] = 1;
+    ancestral_state = 0;
+    ret = tsk_tree_map_mutations(&t, genotypes, NULL, TSK_MM_FIXED_ANCESTRAL_STATE,
+        &ancestral_state, &num_transitions, &transitions);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(ancestral_state, 0);
+    CU_ASSERT_EQUAL_FATAL(num_transitions, 2);
+    CU_ASSERT_EQUAL_FATAL(transitions[0].node, 0);
+    CU_ASSERT_EQUAL_FATAL(transitions[0].parent, TSK_NULL);
+    CU_ASSERT_EQUAL_FATAL(transitions[0].state, 1);
+    CU_ASSERT_EQUAL_FATAL(transitions[1].node, 1);
+    CU_ASSERT_EQUAL_FATAL(transitions[1].parent, TSK_NULL);
+    CU_ASSERT_EQUAL_FATAL(transitions[1].state, 1);
+    free(transitions);
+
     tsk_tree_free(&t);
     tsk_treeseq_free(&ts);
 }
@@ -4387,6 +4404,24 @@ test_single_tree_map_mutations(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     CU_ASSERT_EQUAL_FATAL(num_transitions, 3);
     free(transitions);
+
+    ancestral_state = 5;
+    ret = tsk_tree_map_mutations(&t, genotypes, NULL, TSK_MM_FIXED_ANCESTRAL_STATE,
+        &ancestral_state, &num_transitions, &transitions);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(num_transitions, 4);
+    CU_ASSERT_EQUAL_FATAL(ancestral_state, 5);
+    free(transitions);
+
+    ancestral_state = -1;
+    ret = tsk_tree_map_mutations(&t, genotypes, NULL, TSK_MM_FIXED_ANCESTRAL_STATE,
+        &ancestral_state, &num_transitions, &transitions);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_ANCESTRAL_STATE);
+
+    ancestral_state = 64;
+    ret = tsk_tree_map_mutations(&t, genotypes, NULL, TSK_MM_FIXED_ANCESTRAL_STATE,
+        &ancestral_state, &num_transitions, &transitions);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_ANCESTRAL_STATE);
 
     genotypes[0] = 64;
     ret = tsk_tree_map_mutations(
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -4276,9 +4276,8 @@ get_smallest_set_bit(uint64_t v)
  */
 int TSK_WARN_UNUSED
 tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
-    double *TSK_UNUSED(cost_matrix), tsk_flags_t TSK_UNUSED(options),
-    int8_t *r_ancestral_state, tsk_size_t *r_num_transitions,
-    tsk_state_transition_t **r_transitions)
+    double *TSK_UNUSED(cost_matrix), tsk_flags_t options, int8_t *r_ancestral_state,
+    tsk_size_t *r_num_transitions, tsk_state_transition_t **r_transitions)
 {
     int ret = 0;
     struct stack_elem {
@@ -4335,6 +4334,17 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
     }
     num_alleles++;
 
+    ancestral_state = 0; /* keep compiler happy */
+    if (options & TSK_MM_FIXED_ANCESTRAL_STATE) {
+        ancestral_state = *r_ancestral_state;
+        if ((ancestral_state < 0) || (ancestral_state >= HARTIGAN_MAX_ALLELES)) {
+            ret = TSK_ERR_BAD_ANCESTRAL_STATE;
+            goto out;
+        } else if (ancestral_state >= num_alleles) {
+            num_alleles = (int8_t)(ancestral_state + 1);
+        }
+    }
+
     for (root = self->left_root; root != TSK_NULL; root = self->right_sib[root]) {
         /* Do a post order traversal */
         postorder_stack[0] = root;
@@ -4374,27 +4384,29 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
         }
     }
 
-    optimal_root_set = 0;
-    /* TODO it's annoying that this is essentially the same as the
-     * visit function above. It would be nice if we had an extra
-     * node that was the parent of all roots, then the algorithm
-     * would work as-is */
-    memset(allele_count, 0, ((size_t) num_alleles) * sizeof(*allele_count));
-    for (root = self->left_root; root != TSK_NULL; root = right_sib[root]) {
+    if (!(options & TSK_MM_FIXED_ANCESTRAL_STATE)) {
+        optimal_root_set = 0;
+        /* TODO it's annoying that this is essentially the same as the
+         * visit function above. It would be nice if we had an extra
+         * node that was the parent of all roots, then the algorithm
+         * would work as-is */
+        memset(allele_count, 0, ((size_t) num_alleles) * sizeof(*allele_count));
+        for (root = self->left_root; root != TSK_NULL; root = right_sib[root]) {
+            for (allele = 0; allele < num_alleles; allele++) {
+                allele_count[allele] += bit_is_set(optimal_set[root], allele);
+            }
+        }
+        max_allele_count = 0;
         for (allele = 0; allele < num_alleles; allele++) {
-            allele_count[allele] += bit_is_set(optimal_set[root], allele);
+            max_allele_count = TSK_MAX(max_allele_count, allele_count[allele]);
         }
-    }
-    max_allele_count = 0;
-    for (allele = 0; allele < num_alleles; allele++) {
-        max_allele_count = TSK_MAX(max_allele_count, allele_count[allele]);
-    }
-    for (allele = 0; allele < num_alleles; allele++) {
-        if (allele_count[allele] == max_allele_count) {
-            optimal_root_set = set_bit(optimal_root_set, allele);
+        for (allele = 0; allele < num_alleles; allele++) {
+            if (allele_count[allele] == max_allele_count) {
+                optimal_root_set = set_bit(optimal_root_set, allele);
+            }
         }
+        ancestral_state = get_smallest_set_bit(optimal_root_set);
     }
-    ancestral_state = get_smallest_set_bit(optimal_root_set);
 
     num_transitions = 0;
     for (root = self->left_root; root != TSK_NULL; root = self->right_sib[root]) {
--- a/python/tests/test_parsimony.py
+++ b/python/tests/test_parsimony.py
@@ -181,7 +181,7 @@ def fitch_map_mutations(tree, genotypes, alleles):
     return alleles[ancestral_state], mutations
 
 
-def hartigan_map_mutations(tree, genotypes, alleles):
+def hartigan_map_mutations(tree, genotypes, alleles, ancestral_state=None):
     """
     Returns a Hartigan parsimony reconstruction for the specified set of genotypes.
     The reconstruction is specified by returning the ancestral state and a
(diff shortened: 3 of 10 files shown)


python/tests/test_parsimony.py
```diff
@@ -217,14 +217,18 @@ def hartigan_map_mutations(tree, genotypes, alleles):
...
 
-    allele_count[:] = 0
-    for v in tree.roots:
-        for j in range(num_alleles):
-            allele_count[j] += optimal_set[v, j]
-    max_allele_count = np.max(allele_count)
-    optimal_root_set = np.zeros(num_alleles, dtype=int)
-    optimal_root_set[allele_count == max_allele_count] = 1
-    ancestral_state = np.argmax(optimal_root_set)
+    if ancestral_state is None or ancestral_state == tskit.MISSING_DATA:
```
jeromekelleher, line 220: I see where you're coming from, but is there an advantage to allowing MISSING_DATA as well as None? Isn't it likely to mask mistakes? I think we should leave it out until we hit a clear case where we need it.

python/tests/test_parsimony.py
```diff
@@ -1250,6 +1258,69 @@ def test_reversed_mutation_over_7(self):
...
+
+    def test_mutation_over_3(self):
+        genotypes = [0, 0, 0, 1, 0, 0]
+        ancestral_state, transitions = self.do_map_mutations(
+            self.tree, genotypes, ancestral_state=tskit.MISSING_DATA, compare_lib=False
+        )
+        assert ancestral_state == "0"
+        assert len(transitions) == 1
+        assert transitions[0] == tskit.Mutation(node=3, parent=-1, derived_state="1")
+        genotypes = [0, 0, 0, 1, 0, 0]
```
jeromekelleher, line 1303: Helpful to put in a blank like to separate these different tests visually

jeromekelleher (commented):
This looks basically right to me.

jeromekelleher:
> I guess it's OK that the algorithm is allowed to put a mutation above the root, so that the ancestral state for the whole tree switches immediately?

I don't understand what you mean here - the actual site.ancestral_state is guaranteed to be the value you asked for, right? Once this holds, then it's up to the algorithm to decide a parsimonious assignment, and putting a mutation over the root will often be the only way to do this (.e.g, genotypes = [0,...,0] and specifying ancestral state = 1).

hyanwong:
> I don't understand what you mean here - the actual site.ancestral_state is guaranteed to be the value you asked for, right? Once this holds, then it's up to the algorithm to decide a parsimonious assignment, and putting a mutation over the root will often be the only way to do this (.e.g, genotypes = [0,...,0] and specifying ancestral state = 1).

Yep, it just depends if by ancestral state you mean the state of the MRCA. If that were the case, then the most parsimonious explanation would be to have all branches descending from the MRCA to have a separate switch. I guess it's just a question of making sure the documentation is clear that the ancestral state is not necessarily that shown by the grand MRCA, if mutations above the root are allowed (perhaps that's obvious, but it bears saying somewhere).

hyanwong:
> This looks basically right to me.

Thanks @jeromekelleher . It would be useful to get this in for the tsdate paper revisions, so I'll have a go at the C side. Are there any tips on how in C you are supposed to check for a variable that can get passed in as either an integer or `None` in python?

jeromekelleher:
In the Python C API you'd make it an optional value. It's not obvious what the right approach to getting the plumbing set up, might be simplest if I put in a commit here doing the C side and then perhaps you could write more tests?

hyanwong:
> might be simplest if I put in a commit here doing the C side and then perhaps you could write more tests?

If you have time, that would be great, thanks. Then I can see how it's done too.

python/tests/test_parsimony.py
hyanwong: Good point

hyanwong:
Changes pushed - this now has compare_lib=True, so tests should fail until the C version is working.

jeromekelleher:
Done - C stuff was slightly tricky because we already had an ``ancestral_state`` parameter used to send it back. Using a flag to say that this is actually an input seems OK, though.

I think we should probably support specifying the ancestral state as a string as well, since it's easy enough to do. What do you think?

jeromekelleher:
Also, it's not clear to me when we'll have mutations over roots and mutations over the children of the root - it would be good to test this on a few different oddball scenarios to make sure it's doing what we think it's doing.

hyanwong:
> Done - C stuff was slightly tricky because we already had an `ancestral_state` parameter used to send it back. Using a flag to say that this is actually an input seems OK, though.

Thanks - quick work.

> I think we should probably support specifying the ancestral state as a string as well, since it's easy enough to do. What do you think?

If you don't mind variable type params, that's fine, yeah. I guess this would just be in Python though?

jeromekelleher:
> If you don't mind variable type params, that's fine, yeah. I guess this would just be in Python though?

Yes, just at the top-level Python function. See the comment.

hyanwong:
I think that there's a slight wrinkle: at the moment we calculate the number of alleles as the maximum number in the genotype array. But actually we should be calculating it as the length of the provided alleles list (or one less than that if the last item in the alleles list is `None`). Think of the case where we specify an alleles list of `["A", "B", "C"], have a genotype array of 0s and 1s, and specify the `ancestral_state` as 2 (or "C").

For the python code, we'd just have

```
num_alleles = len(alleles)
if alleles[-1] is None:
    num_alleles -= 1
```

(instead of `num_alleles = np.max(genotypes[not_missing]) + 1`)

I guess something similar could happen in the C code? I assume it was done like that previously simply for efficiency reasons, e.g. if the user specified a long list of alleles but only a few were actually present in the genotype array?

jeromekelleher:
This is a quirk all right, and we could work around it by letting anything < 64 be the ancestral state (from the C code's perspective, which doesn't know anything about alleles). My thought was though, that specifying a value 2 (say) as the ancestrel state when the genotypes array is full of 0s and 1s is probably a mistake.

We can change it though, if you think there's a reasonable use case.

hyanwong:
> This is a quirk all right, and we could work around it by letting anything < 64 be the ancestral state (from the C code's perspective, which doesn't know anything about alleles). My thought was though, that specifying a value 2 (say) as the ancestral state when the genotypes array is full of 0s and 1s is probably a mistake.
> 
> We can change it though, if you think there's a reasonable use case.

I think it needn't be a mistake to specify an acetal state that isn't in the haplotype array. Indeed, it is quite likely to happen in user code. We have had a fair number of examples brought up by users where the ancestral state in the VCF is, say, "C", but the samples that are actually used in the analysis have only "A" and "T". It would be a mistake, IMO, not to be able to lay down variation on these triallelic sites via parsimony in tsinfer, which is one of the use-cases of this new functionality.

I've just pushed a commit which changes both the C and python versions as appropriate, along with a few more tests. Note that I haven't added to the C tests, so the check in the C code that ancestral_state < HARTIGAN_MAX_ALLELES is not tested.

jeromekelleher:
Yep, looks good.

hyanwong:
Squashed and ready for review

hyanwong:
This is failing CI because the C checker is failing to notice that `ancestral_state` is set either in line 4338 (`if (options & TSK_MM_FIXED_ANCESTRAL_STATE)`) or 4407 (`if (!(options & TSK_MM_FIXED_ANCESTRAL_STATE))`). How do I get the checker to realise that this covers all the branches, given that there's a bit of code that needs to be run between the two if statements (because it depends on the first `if`, and is then used in the second `if`)?

jeromekelleher:
Just set ancestral_state to zero before the ``if``, like
```
ancestral_state = 0; /* keep compiler happy */
``

**Comment by hyanwong** [comment]
> Just set ancestral_state to zero before the `if`, like
> 
> ```
> ancestral_state = 0; /* keep compiler happy */
> ```

Perfect, thanks. Now it's complaining about

```
num_alleles = ancestral_state + 1;
```

presumably because although both `num_alleles` and `ancestral_state` are int8_t, the "1" isn't. Do I just cast the result, or can I convert the 1 to an int8_t value?

**Comment by jeromekelleher** [comment]
Probably cast the 1 - either is fine though

File: c/tests/test_trees.c [hunk 3]
```diff
@@ -4388,6 +4405,13 @@ test_single_tree_map_mutations(void)
     CU_ASSERT_EQUAL_FATAL(num_transitions, 3);
     free(transitions);
 
+    ancestral_state = 5;
+    ret = tsk_tree_map_mutations(&t, genotypes, NULL, TSK_MM_FIXED_ANCESTRAL_STATE,
+        &ancestral_state, &num_transitions, &transitions);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(num_transitions, 4);
+    free(transitions);
```
jeromekelleher, line 4413: Missing the test for bad_ancestral state now - need to add another check. Also, good to check that ancestral_state is 5 in this one.

c/tskit/trees.c
```diff
@@ -4335,6 +4334,19 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
     }
     num_alleles++;
 
+    if (options & TSK_MM_FIXED_ANCESTRAL_STATE) {
+        ancestral_state = *r_ancestral_state;
+        if ((ancestral_state < 0) || (ancestral_state >= HARTIGAN_MAX_ALLELES)) {
+            ret = TSK_ERR_BAD_ANCESTRAL_STATE;
+            goto out;
```
jeromekelleher, line 4341: Missing test coverage - see previous comment.

c/tskit/trees.c
```diff
@@ -4335,6 +4334,19 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
...
+        } else if (ancestral_state >= num_alleles) {
+            num_alleles = ancestral_state;
```
jeromekelleher, line 4343: Why not ``num_alleles = ancestral_state + 1``?

c/tskit/trees.c
```diff
@@ -4335,6 +4334,19 @@ tsk_tree_map_mutations(tsk_tree_t *self, int8_t *genotypes,
...
+            num_alleles++;
+        }
+    } else {
```
jeromekelleher, line 4346: I wouldn't bother with the else here, just assign ancestral_state = 0 immediately before the if statement. Uses less vertical space and looks less like deliberate logic.

python/tests/test_parsimony.py
```diff
@@ -217,14 +226,18 @@ def hartigan_map_mutations(tree, genotypes, alleles):
...
+        allele_count[:] = 0
+        for v in tree.roots:
+            for j in range(num_alleles):
+                allele_count[j] += optimal_set[v, j]
+        max_allele_count = np.max(allele_count)
+        optimal_root_set = np.zeros(num_alleles, dtype=int)
+        optimal_root_set[allele_count == max_allele_count] = 1
+        ancestral_state = np.argmax(optimal_root_set)
+    else:
+        if ancestral_state < 0 or ancestral_state >= len(alleles):
```
jeromekelleher, line 239: The second clause doesn't count here now. Better to put the check further with the check around line 205 so that this logic is all in one place.

python/tests/test_parsimony.py
```diff
@@ -520,16 +533,45 @@ class TestParsimonyBase:
...
+                genotypes, alleles, ancestral_state
+            )
+            ancestral_state2, transitions2 = hartigan_map_mutations(
+                tree, genotypes, alleles, ancestral_state
+            )
+            assert ancestral_state1 == ancestral_state2
+            assert len(transitions1) == len(transitions2)
+            assert transitions1 == transitions2
+        else:
+            # Test the implementation in this file first
```
jeromekelleher, line 553: What's this change for? It's equivalent to what was there before isn't it, just with code duplication?

python/tskit/trees.py
```diff
@@ -2498,14 +2498,35 @@ def map_mutations(self, genotypes, alleles):
         :param tuple(str) alleles: The alleles for the specified ``genotypes``. Each
             positive value in the ``genotypes`` array is treated as an index into this
             list of alleles.
+        :param ancestral_state: A fixed ancestral state, specified either as a
+            non-negative integer less than the number of alleles, or a string which
+            must be one of the ``alleles`` provided above. If ``None`` (default) then
+            the most parsimonious ancestral state is chosen. Note that if the ancestral
```
jeromekelleher, line 2504: "one of the most parsimonious ancestral states"

python/tskit/trees.py
```diff
@@ -2498,14 +2498,35 @@ def map_mutations(self, genotypes, alleles):
...
+        :type ancestral_state: Union[int, str]
         :return: The inferred ancestral state and list of mutations on this tree
             that encode the specified observations.
         :rtype: (str, list(tskit.Mutation))
         """
         genotypes = util.safe_np_int_cast(genotypes, np.int8)
-        if np.max(genotypes) >= 64:
+        max_alleles = np.max(genotypes)
+        if ancestral_state is not None:
+            try:
```
jeromekelleher, line 2516: I don't see the advantage of this try except over ``isinstance(ancestral_state, str)``, which is clearer in intent

jeromekelleher (commented):
Some final things to clear up

c/tskit/trees.c
hyanwong: That fails with `error: conversion from ‘int’ to ‘int8_t’ {aka ‘signed char’} may change value`, as does the following (which I don't quite understand)

```
num_alleles = ancestral_state + (int8_t) 1
```

I don't mind whatever syntax as long as it compiles. Should I try `num_alleles = (int8_t) (ancestral_state + 1)`

python/tests/test_parsimony.py
hyanwong: I don't think so. In the previous version, whether or not we specified `compare_lib`, we ran `hartigan_map_mutations` first. Therefore if that function raised an error, we never went on to test whether `tskit.map_mutations` raised the same error. The way I've structured it, if you specify `compare_lib` you test `tskit.map_mutations` first, so can check for the correct error raised there.

python/tskit/trees.py
hyanwong: Good point. I wonder if people will misread "one of the most" as "almost but not quite the best parsimony", because colloquially we say "one of the most" to mean "among the top XXX of their class". How about "an ancestral state is chosen which provides the most parsimonious placement of mutations"?

python/tskit/trees.py
hyanwong: Happy to use `isinstance`. ISTR you thought use of is instance was indicative of poor code structure, that's all. I can't see a way around this sort of thing here, though.

jeromekelleher: Can't really avoid isinstance when doing type polymorphism, it's better than opaque try/excepts

python/tskit/trees.py
jeromekelleher: It's not getting across the arbitrariness though, you're making it sound like there's a unique placement of mutations that gives the best parsimony, when there's really a single best parsimony *score* and often lots of ways of achieving that score, of which we choose one arbitrarily.

Also, I notice that we haven't emphasised that if you choose the ancestral state you will get back a parsimonious assignment of mutations conditioned on that ancestral state, and this may not be fully parsimonious.

python/tests/test_parsimony.py
jeromekelleher: If we're checking for errors then call the library version directly.

c/tskit/trees.c
jeromekelleher: I don't know - there must be a way to do it. If you can't figure it out then let me know and I'll fix it.

python/tests/test_parsimony.py
hyanwong: Yep, good point.

python/tskit/trees.py
hyanwong: > It's not getting across the arbitrariness though, 

I hoped that saying "an ancestral state" rather than "the ancestral state" made that point. But perhaps that's too subtle?

> Also, I notice that we haven't emphasised that if you choose the ancestral state you will get back a parsimonious assignment of mutations conditioned on that ancestral state, and this may not be fully parsimonious.

Yes, good point. Will add.

hyanwong: I now say 

            If ``None`` (default) then
            an ancestral state is chosen arbitrarily from among those that provide
            the most parsimonious placement of mutations. Note that if the ancestral
            state is specified, the placement of mutations may not be as parsimonious
            as that which could be achieved by leaving the ancestral state unspecified;
            additionally it may lead to mutations being placed above the root node(s) of
            the tree (for example if all the samples have a genotype of 1 but the
            ancestral state is fixed to be 0).

Which I think covers it all?

hyanwong:
Seems to all be passing now, and all comments addressed. Thanks for the review @jeromekelleher

jeromekelleher (approved):

PR #1553, opened 2021-07-05
Add ts.copy

benjeffery:
This seems to be a very common pattern:

```python
    tables = ts.dump_tables()
    #do something with tables e.g.
    tables.provenances.add_row(record=record)
    ts = tables.tree_sequence()
```
This PR adds `TreeSequence.copy` which takes an optional `modifier` argument. With this the above becomes:
```python
ts = ts.copy(modifier=lambda tables: tables.provenances.add_row(record=record))
```

Having the words "copy" and "modify" in the resulting code clarify what is going on to the reader. To me this is clearer, but I'd like to hear some opinions as lambdas(or a locally defined function) aren't generally the most popular thing in python.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1637,6 +1637,21 @@ class TestTreeSequence(HighLevelTestCase):
                         assert edge.right >= tree.interval.right
             assert np.all(edge_visited)
 
+    def test_copy(self, ts_fixture):
+        # No modifier results in a pure copy
+        assert ts_fixture.copy() == ts_fixture
+        # No-op modifier results in a pure copy
+        assert ts_fixture.copy(modifier=lambda tables: tables) == ts_fixture
+        # Changing a table works
+        diff_ts = ts_fixture.copy(
+            modifier=lambda tables: tables.provenances.add_row(record="TEST")
+        )
+        assert diff_ts != ts_fixture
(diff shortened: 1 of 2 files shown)


petrelharp:
Hm - on the one hand, I have seen the error a few times of forgetting to do `ts = tables.tree_sequence()` at the end, and this would help with that. On the other hand, I don't think it's that common to want to do a short one-liner to the tables; so usually people would use this like
```
def modify_tables(tables):
  # do stuff

ts = ts.copy(modifier=modify_tables)
```
At first I wasn't excited: it's harder to incrementally build up the operations in `# do stuff`if they're packed away in a function, and isn't any less code than the original method. **But** it encourages writing modular code! So, I'm enthusiastic. (I wouldn't say it's a big deal, though.)

jeromekelleher:
I'm pretty "meh" on this one tbh - lambdas with side-effects are pretty frowned-upon I think, and callbacks tend to confuse people in general.

benjeffery:
> callbacks tend to confuse people in general

I think this is true and for most this will be confusing. Having done JavaScript where everything is a lambda for a few years has made me lambda crazy, so I have to avoid the urge in Python!

jeromekelleher:
Working with [twisted](https://twistedmatrix.com/trac/) gave me a strong aversion to callbacks!

PR #1556, opened 2021-07-06
Fix pickling of TreeSequence subclasses

### LINKED ISSUE #1555: Unpickling TreeSequence from pyslim doesn't initialize metadata, and raises exception
I've run into an unexpected exception with tskit 0.3.6 — a description and MRE are below. The zip file [tskit_mre.zip](https://github.com/tskit-dev/tskit/files/6767413/tskit_mre.zip) contains a SLiM simulation routine based on recipe 17.1 and an example tree sequence with metadata. Loading the tree using pyslim and trying to pickle and unpickle it as below, 

```python
import pyslim as psl
import tskit as tsk
import pickle

[x.__version__ for x in (psl, tsk)]
# ['0.600', '0.3.6']

tr = psl.load('./test.tree')
with open('test.pkl', 'wb') as f:
  pickle.dump(tr, file=f)

with open('test.pkl', 'rb') as f:
   tr2 = pickle.load(file=f)
```
leads to the following error,

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-11-dc9156f6d230> in <module>
      1 with open('../data/slim_sims/test.pkl', 'rb') as f:
----> 2     t = pickle.load(f)

~/miniconda3/envs/bioinfo/lib/python3.7/site-packages/tskit/trees.py in __setstate__(self, tc)
   3435 
   3436     def __setstate__(self, tc):
-> 3437         self.__init__(tc.tree_sequence().ll_tree_sequence)
   3438 
   3439     def __eq__(self, other):

~/miniconda3/envs/bioinfo/lib/python3.7/site-packages/pyslim/slim_tree_sequence.py in __init__(self, ts, reference_sequence, legacy_metadata)
    155     def __init__(self, ts, reference_sequence=None, legacy_metadata=False):
    156         self.legacy_metadata = legacy_metadata
--> 157         if not (isinstance(ts.metadata, dict) and 'SLiM' in ts.metadata
    158                 and ts.metadata['SLiM']['file_version'] in compatible_slim_file_versions):
    159             tables = ts.dump_tables()

AttributeError: '_tskit.TreeSequence' object has no attribute 'metadata'
```

This is with tskit 0.3.6 — on a different machine with 0.3.5, this works fine, so I suspect this is a regression from this change in how pickled tree sequences are initialized: https://github.com/tskit-dev/tskit/pull/1298/files#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8L3318. I would submit a PR, but I'm not sure the best way to initialize the object in a way that preserves the metadata.

benjeffery:
Fixes #1555

@petrelharp pyslim should probably add a pickle-unpickle test?

We could add sub-classing tests to tskit, but as it's not a recommended path I think the onus is on libraries doing so to test?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -3423,7 +3423,7 @@ class TreeSequence:
         return self.dump_tables()
 
     def __setstate__(self, tc):
-        self.__init__(tc.tree_sequence().ll_tree_sequence)
+        TreeSequence.__init__(self, tc.tree_sequence().ll_tree_sequence)
 
     def __eq__(self, other):
         return self.tables == other.tables


python/tskit/trees.py
```diff
@@ -3423,7 +3423,7 @@ def __getstate__(self):
         return self.dump_tables()
 
     def __setstate__(self, tc):
-        self.__init__(tc.tree_sequence().ll_tree_sequence)
+        TreeSequence.__init__(self, tc.tree_sequence().ll_tree_sequence)
```
jeromekelleher, line 3426: Hmm, so this explicitly skips the subclass init. What are subclasses supposed to do then, I guess define their own ``__setstate__``

Is there any official guidance on what we're suppose to do here?

jeromekelleher:
Not obvious to me that tskit is doing anything wrong here - isn't the responsibility on subclasses to make sure they are initialising the object correctly? Did we change the signature of TreeSequence at some point?

python/tskit/trees.py
benjeffery: Yes, the proper fix is for pyslim to override `__setstate__` in it's subclass, calling `super().__init__` before doing it's own work.

benjeffery:
> Not obvious to me that tskit is doing anything wrong here - isn't the responsibility on subclasses to make sure they are initialising the object correctly? Did we change the signature of TreeSequence at some point?

Not since the copy from msprime code. You're right that this better fixed in pyslim.

jeromekelleher:
The fix here seems like worse behaviour to me - I think this should be fixed in pyslim.

PR #1559, opened 2021-07-08
Unknown time doc

### LINKED ISSUE #1504: is_unknown_time is not documented on the Python API page
I think this is just an oversight, as it has a docstring.

hyanwong:
## Description

This creates a new documentation section for utility functions (there's only 1 documented there at the moment)

The second commit moves the "Constants" section of the API docs to a top-level section, rather than nesting it within "Trees and tree sequences", which I think is much more logical, as the constants are used in e.g. the tables API too.

Fixes #1504

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -54,31 +54,6 @@ about particular nodes in the tree.
     :members:
     :autosummary:
 
-+++++++++
-Constants
-+++++++++
-
-.. autodata:: NULL
-    :annotation: = -1
-
-.. autodata:: NODE_IS_SAMPLE
-    :annotation: = 1
-
-.. autodata:: MISSING_DATA
-    :annotation: = -1
-
-.. autodata:: FORWARD
-    :annotation: = 1
-
-.. autodata:: REVERSE
-    :annotation: = -1
-
-.. autodata:: ALLELES_ACGT
-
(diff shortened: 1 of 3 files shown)


benjeffery (approved):

jeromekelleher (approved):

hyanwong:
You don't know how to remove the annoying "`(*args, **kwargs)`" auto-doc-ed string do you @benjeffery?
Screenshot 2021-07-08 at 12 49 41

It doesn't happen with the documentation for `tskit.ALLELES_ACGT`, presumably because that is not defined as a `_tskit.XXX` variable: https://github.com/tskit-dev/tskit/blob/70a370ade325237435654adb4e1030ac7108d9a6/python/tskit/__init__.py#L47

benjeffery:
Hmm, does `autoclass::` work for that one?

hyanwong:
No, sadly not, You need `autodata`. I think it's something to do with MagicMock, as when I remove the ` :annotation: = -1` line, I get this:
Screenshot 2021-07-08 at 14 03 07

But I have no idea what MagicMock actually *does*, so I'm completely lost after that.

hyanwong:
Aha - looks like we don't actually need the MagicMock stuff any more, I think? I'll PR to remove it and that fixes everything AFAICS.

Issue #1503, opened 2021-06-18
Document TableRow.replace() ?


hyanwong:
The most common thing to want to do when using e.g. `NodeTable.append()` is to want to change one variable, e.g. `time`. At the moment this requires you to 

```
import dataclasses
node_table.append(dataclasses.replace(row, time=new_val))
```

Is it worth having the syntactic sugar

```
node_table.append(row.replace(time=new_val))
```

to avoid having to import dataclasses and remember the syntax? This could be a lightweight method on a TableRow class, but is it worth duplicating the data class functionality here? I think on balance it might be, especially for newbies to the wonderful world of tree sequences (it might also encourage using `.append()` instead of `.add_row(x, y, z)`).

benjeffery:
As a data point I have typed the second (currently incorrect) version almost every time I've written this piece of code!

jeromekelleher:
I thought we had done this? I wasn't keen on the long-term API dependency on dataclasses implied by the ``dataclasses.replace(row, time=new_val))`` - if we encourage users to do this, our rows must be dataclasses forever.

hyanwong:
That's a good point too. Looks like we should just implement it then? Simply this?

```
def replace(self, **kwargs)
    detaclasses.replace(self, **kwargs)
```

I don't know what the "/" in the replace docs parameter list is though: https://docs.python.org/3/library/dataclasses.html#dataclasses.replace

grahamgower:
> I don't know what the "/" in the replace docs parameter list is though: https://docs.python.org/3/library/dataclasses.html#dataclasses.replace

That means the first parameter is positional only. I.e., the keyword *cannot* be used to specify that param.
https://www.python.org/dev/peps/pep-0570/

benjeffery:
I'm clearly losing my mind as I implemented this already: https://github.com/tskit-dev/tskit/blob/main/python/tskit/util.py#L37

But then I'm sure @hyanwong got an error trying to use `.replace` on slack the other day?

hyanwong:
I thought I did too. But it works now. Hmmm.

hyanwong:
Reopening because I think this should be documented somewhere?

PR #1560, opened 2021-07-08
Document the .replace method

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/1503

### DIFF
--- a/docs/python-api.rst
+++ b/docs/python-api.rst
@@ -151,16 +151,19 @@ Tables
 ++++++
 
 The :ref:`tables API <sec_binary_interchange>` provides an efficient way of working
-with and interchanging :ref:`tree sequence data <sec_data_model>`. Each table
-class (e.g, :class:`NodeTable`, :class:`EdgeTable`) has a specific set of
-columns with fixed types, and a set of methods for setting and getting the data
+with and interchanging :ref:`tree sequence data <sec_data_model>`. Each table class
+(e.g, :class:`NodeTable`, :class:`EdgeTable`, :class:`SiteTable`) has a specific set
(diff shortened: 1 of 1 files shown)


docs/python-api.rst
```diff
@@ -173,16 +173,19 @@ Tables
...
-of the data. For example::
+Each table supports accessing the data either by row or column.
+
+To access the data in a *column*, we can use standard attribute access which will
+return a copy of the column data as a numpy array. To access the data in a *row*, say
+row number ``j`` in table ``t``, simply use ``t[j]``. The returned row has attributes
+allowing contents to be accessed by name, e.g. ``site_table[0].position``,
+``sites_table[0].ancestral_state``, ``sites_table[0].metadata`` etc. Note that row
+contents cannot be changed. To create a new row with changed content,
+row objects have a ``replace`` method. For example::
```
jeromekelleher, line 188: Nice to link to a method, like "(e.g., :meth:`Site.replace`)" or something?

jeromekelleher (approved):
Looks good, minor suggestion

docs/python-api.rst
hyanwong: Yeah, We don't list the actual method anywhere. I think either we'd need to list all the TableRow definitions, or perhaps just do:

```
.. autoclass:: Tskit.util.Dataclass
```

I did that, but the removed it, as then we'd be actually documenting this data class, and presumably therefore be committed to retaining it (and not e.g. implementing `replace` in a different way. So I thought it easiest not to commit to a particular implementation for the time being.

jeromekelleher: OK, let's leave it then.

PR #1561, opened 2021-07-08
Remove _tskit module mocking

hyanwong:
AFAICS we don't need to do this "mocking" think any more - I can compile the docs without it, and I seem to remember someone saying that this way of doing things was deprecated now.

If we remove this, it fixes https://github.com/tskit-dev/tskit/pull/1559#issuecomment-876374053

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,26 +12,10 @@
 import os
 import subprocess
 import sys
-from unittest.mock import MagicMock
 
 from docutils import nodes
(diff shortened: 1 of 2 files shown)


benjeffery (approved):
Nice, I meant to remove that a while ago.

jeromekelleher (approved):
Good call - this was all needed because we couldn't build the low-level module on RTD. But we're not on RTD anymore, so it's not needed.

PR #1562, opened 2021-07-08
Correct docs to point out that individuals are sorted

hyanwong:
Fixed in https://github.com/tskit-dev/tskit/pull/1199 but not documented


jeromekelleher (approved):
LGTM

PR #1571, opened 2021-07-17
Support storing 64 bit offsets

jeromekelleher:
Stacked on #1552

This adds support for either 32 or 64 bit offsets in the file format and a new flag for dump that forces offsets to be 64 bit. This is an intermediate step towards having 64 bit offsets in memory. The TSK_DUMP_FORCE_OFFSET64 will be useful because it will allow us to test that we are loading correctly from 64 bit offset arrays in the kastore without needed to create huge metadata arrays.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,11 +50,14 @@ jobs:
             CC=clang CXX=clang++ meson build-clang c
             ninja -C build-clang
 
-      - run:
-          name: Compile C with gcc in 64 bit mode
-          command: |
-            CFLAGS=-D_TSK_BIG_TABLES CPPFLAGS=-D_TSK_BIG_TABLES meson build-gcc-64 c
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I think this is ready for review @benjeffery - it doesn't affect any existing behaviour, so should be OK to merge I think.

jeromekelleher:
Ah, crap, the 64 bit BIG_TABLE tests aren't going to work. I think it's probably simplest to disable these for a few days until the follow-ups can be done @benjeffery? Otherwise, we'll end up having to implement full change in this PR I think.

jeromekelleher:
I've added a commit to disable these tests - if you agree with this course of action @benjeffery we should open an issue so we don't forget to turn them back on again.

benjeffery (approved):
LGTM, Seems a good idea to do this incrementally.

PR #1579, opened 2021-07-21
Convert most size_t values to tsk_size_t.

### LINKED ISSUE #1573: Remove size_t types from public API
Following on from the decision in #1527 to typedef tsk_size__t to 64 bit, and to deal with the 32 bit issues that arise, it would be much simpler if there were no size_t types in the public API. Most of these are in the stats API, and there's no good reason for them to be size_t rather than tsk_size_t.

jeromekelleher:
Clarify the conventions around using size_t vs tsk_size_t

Closes #1573

Removes (nearly) all size_t values from the public API, and clarifies how we should use them within the library. This is a good step towards 1.0 I think, and should clear up some of the mess on 32 bit machines when we have 64 bit tsk_size_t.

### DIFF
--- a/c/tests/test_genotypes.c
+++ b/c/tests/test_genotypes.c
@@ -880,8 +880,8 @@ test_single_tree_silent_mutations_i16(void)
     tsk_flags_t options[] = { 0, TSK_16_BIT_GENOTYPES };
     tsk_id_t all_samples[] = { 0, 1, 2, 3 };
     tsk_id_t *samples[] = { NULL, all_samples };
-    size_t num_samples = 4;
-    size_t s, f;
+    tsk_size_t num_samples = 4;
+    tsk_size_t s, f;
     int ret;
 
     tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges, NULL,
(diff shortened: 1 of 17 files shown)


benjeffery (approved):
Great!

Searching through for other `size_t` that should be `tsk_size_t` these stand out, unless I've misunderstood:

`tables.c:7953`
`tables.c:8230` <-- Seems overflow unsafe, but very unlikely to be an issue?
`tables.c:9864`

jeromekelleher:
Thanks @benjeffery!

> tables.c:7953

num_sites is being used as the argument to malloc and memcpy, so size_t is correct. I guess there's a potential overflow with the ``num_sites + 2``,  but then the size_t would overflow anyway when doing the multiply by sizeof(...), so I don't think there's any problem there.

> tables.c:8230 <-- Seems overflow unsafe, but very unlikely to be an issue?

Similarly, ``num_nodes_alloc`` is being used as the argument to calloc. Good point about the overflow, let's see. I don't think it's possible, because tables.num_rows is at most 2**31 and size_t will be a uint32_t at least.

There's probably a bunch of tricky overflows around the code to be honest if people are working with large tables on a 32 bit system, because of all the ``malloc(x * sizeof(...))``. But, I doubt one of these large tables would fit into 2G of RAM in the first place anyway, though, so it's probably not worth worrying about.

> tables.c:9864
This is correct because we're talking about the size of buffers for the kastore API.

So, I think we're good?

benjeffery:
Ok, seems my misunderstanding is was parsing "computing memory block sizes" as "counting bytes".

PR #1584, opened 2021-07-22
Test 32bit on CI

benjeffery:
Adds a 32bit build and test to circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,17 +1,8 @@
 version: 2.1
-jobs:
-  build:
-    docker:
-      - image: circleci/python:3.7-buster
-    working_directory: /home/circleci/tskit
+
+commands:
+  setup:
     steps:
-      - checkout
-      - run: sudo chown -R circleci:circleci *
-      - restore_cache:
-          # It's sometimes necessary to nuke the cache, and the simplest
-          # way to do it is to change the key. We can increment this
-          # version number when we want to do this.
-          key: tskit-{{ .Branch }}-v5
       - run:
           name: Checkout submodules
           command: |
@@ -19,7 +10,7 @@ jobs:
       - run:
           name: Install dependencies and set path
           command: |
-            sudo apt-get install ninja-build libcunit1-dev valgrind clang doxygen
+            sudo apt-get install -y ninja-build libcunit1-dev valgrind clang doxygen
             # Install meson to the system packages so we can run it as root
             sudo pip install meson==0.55
             pip install numpy==1.18.5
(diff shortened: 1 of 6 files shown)


benjeffery:
@jeromekelleher Here's the CI for 32bit.

jeromekelleher:
Awesome! I worry that we're duplicating some stuff a bit pointlessly, though, and particularly whether CodeCov will code with having two uploads from the different runs? I think we can pull stuff out of the compile_and_test target though - I'll add some comments.

jeromekelleher:
I think we could probably pull out:

- Run gcov & upload coverage
- Test building with meson subproject
- Install shared library and hand-compile program.
- Run example make file
- Upload LWT coverage

There's an argument for uploading the coverage stuff twice, since CodeCov is supposed to be able to handle it, and it'll give us visibility on testing in any #defs we need to do. I'm not sure I trust CodeCov though, tbh.

I don't think there's any point in including the makefile or the example programs though. We'd probably end up having to put a bunch of pointless-looking casts in to them to compile cleanly on 32 bit (but maybe not?)

benjeffery:
For simplicity's sake I'd rather keep the jobs the same unless as you say we'd have to put in work to make the 32 version pass. The  steps we could remove only take a few seconds.

benjeffery:
The tests don't pass as root so I've had to make a custom docker image using the one we had before as base. It fails on an error that isn't immediately obvious to me now. 
```
  Test: test_single_tree_compressed_matrix ...FAILED
    1. ../c/tests/test_haplotype_matching.c:351  - CU_ASSERT_EQUAL(decoded[j][2],0.0)
```

jeromekelleher:
Hmm, that's maybe something to do with IEEE maths not being fully present on i386. I'll try it out on my laptop and see if I get the same error.

jeromekelleher:
I've updated this @benjeffery with a few fixes. Turns out one of them was a real bug, so I guess 32 bit CI is showing some value straight away.

Note I rebased and force pushed to bring in the size_t stuff, in case that was affecting things.

jeromekelleher:
Looks like we're nearly there, one more error in the Python tests at tests/test_stats.py:198. I'll take a look.

jeromekelleher:
Hopefully this is ready to go now @benjeffery ...

jeromekelleher:
Looks like codecov is OK with this, so I say we merge. Thanks for sorting this out @benjeffery, it's a great bit of infrastructure.

jeromekelleher (approved):

benjeffery:
> Looks like codecov is OK with this, so I say we merge. 

I've got one tweak to move from my personal dockerhub to the tskit one now I have access to it

PR #1585, opened 2021-07-23
64 bit tsk_size_t

### LINKED ISSUE #1527: Change offset columns to 64 bit.
Following the discussion in #343, we've decided that metadata and other ragged columns should be 64 bit. This should be done for C API 1.0.

I think we can either change the tsk_size_t typedef, or create a new type ``tsk_offset_t`` (or something) which is used for just the offset columns. Let's discuss the pros and cons of that here, as well as the details involved in this transition.

### LINKED ISSUE #1530: Support 32 or 64 bit sizes in offset columns
#1529 added some infrastructure abstract out reading and writing offset columns. Here's how I think we can build on this to allow us to support either 32 bit or 64 bit sizes for the offset column (see #1527) in the file format:

- add a new field ``tsk_size_t * offset_array_mem`` to the ``read_table_ragged_col_t`` struct
- Then, in ``read_table_ragged_cols`` if the type of the offset_col is ``KAS_UINT32``, then we allocate the required memory for this column into the offset_array_mem, and update ``col->offset_array_dest`` to point to this memory. We then loop through the array of 32 bit values from the kastore, and cast them into the newly allocated array.
- As part of the ``out`` block in (.e.g) ``tsk_individual_table_load`` we call a new function ``free_read_table_memory(cols, ragged_cols, properties);``, which will loop through the ragged cols and tsk_safe_free the ``offset_array_mem``. (I think we should include the ``cols`` and ``properties`` here while we're at it, because we may do something similar for them soon.)
- We'll probably have to update the struct initialisers read_table_ragged_col_t to set ``offset_array_mem`` to NULL; either that or we change to using named initialsers (like ``.name="individuals/location"``, etc) which I *think* guarantee that anything that's not explicity assigned is set to zero.

### LINKED ISSUE #1554: 64bit offsets - LWT compatibility
We're moving to 64bit offset columns in #1527. This will create incompatibility in the `LightweightTableCollection` `dict` representation which is used to transfer a tree-sequence between python modules that have compiled different versions of the tskit C API.

There are two situations, a 64bit tskit reading a dict from a 32bit tskit, and vice-versa.

The first of these is relatively straight forward - we already use [`PyArray_FromAny`](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_FromAny) which will silently apply any cast which is safe, so will read the 32bit offset array just fine.

The second case will fail as the cast is unsafe. My suggestion for dealing with this is that the 64bit code writes a 32bit array to any `X_offset` column if safe to do so and all the offsets are below the maximum value. If they aren't it writes 64bit and the 32bit code will error out. Unfortunately the error message will not be friendly, but there is no way to retrospectively fix that.

We should add some tests to CI with old msprime versions to test this.

jeromekelleher:
Stacked on #1579.

Closes #1530
Closes #1527
Closes #1554

### DIFF
--- a/c/tests/test_file_format.c
+++ b/c/tests/test_file_format.c
@@ -359,10 +359,10 @@ verify_bad_offset_columns(tsk_treeseq_t *ts, const char *offset_col)
     int ret = 0;
     kastore_t store;
     tsk_table_collection_t tables;
-    tsk_size_t *offset_array, *offset_copy;
+    uint32_t *offset_array, *offset_copy;
     size_t offset_len;
     int type;
-    tsk_size_t data_len;
+    uint32_t data_len;
 
     ret = tsk_treeseq_dump(ts, _tmp_file_name, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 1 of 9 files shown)


jeromekelleher:
Pausing this for now, going to tackle  #1589 first, get some infrastructure in place for the Python C interface before making the transition.

jeromekelleher:
Closing in favour of #1618

Issue #1545, opened 2021-06-30
Update a single row


benjeffery:
### Discussed in https://github.com/tskit-dev/tskit/discussions/1543

<div type='discussions-op-text'>

<sup>Originally posted by **hyanwong** June 30, 2021</sup>
At the moment I think adding to the metadata arrays (e.g. on populations, individuals, or nodes) is quite intricate, and I have to look it up every time. I think it's something like this:

```
ts = msprime.sim_ancestry(10)
tables = ts.dump_tables()
pop_metadatas = [p.metadata for p in ts.populations()]  # Why is there no tables.population.get_metadatas() method?
pop_metadatas[0]['name'] += "(the only population)"
tables.populations.packset_metadata(
    [tables.populations.metadata_schema.validate_and_encode_row(r) for r in pop_metadatas])
new_ts = tables.tree_sequence()
print([p for p in new_ts.populations()])
```

(it looks me about 10 minutes to look up how to do that, and the information is scattered all over the docs - I don't think a beginner could do it TBH).

Should we have some convenience methods to help with this? I quite often want to edit metadata on just a single entry in a table: e.g. label a particular individual, flag up a single node, etc.

We should also give an example of doing this in the [metadata tutorial](https://tskit.dev/tutorials/metadata.html)</div>

benjeffery:
I've added this to the C 1.0 milestone as we'll need something like: `tsk_update_char_ragged_array(tsk_size_t *offset_to_modify, char *array_to_modify, tsk_size_t *row_indexes, tsk_size_t *new_offset, char *new_array)`

jeromekelleher:
I'm not sure this is something we want in the public API - wouldn't it be better to support updating a single row? If people are going to be updating multiple rows, then they should work column wise anyway.

benjeffery:
Gah, of course, forgetting the C milestone is for the public API. Was thinking this would just be used by the CPython code.

jeromekelleher:
Mind if we generalise this to "update single row" @benjeffery? No point in just focusing on the ragged columns here, we want an ``x_table_set_row`` as discussed [here](https://github.com/tskit-dev/tskit/discussions/1543#discussioncomment-942289)

jeromekelleher:
Removing C API tag as it's done there.

benjeffery:
[...]

PR #1597, opened 2021-07-29
Add update row to low level python

benjeffery:
Initial low level work for #1545 
WIP - only individuals done

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -780,6 +780,31 @@ out:
 }
 
 static int
+tsk_id_converter(PyObject *py_obj, tsk_id_t *id_out)
+{
+    long long temp_long;
+    int ret = 0;
+
+    if (!PyArg_Parse(py_obj, "L", &temp_long)) {
+        goto out;
+    }
+    if (temp_long > TSK_MAX_ID) {
+        PyErr_SetString(PyExc_OverflowError, "Value too large for tskit id type");
+        goto out;
+    }
+    if (temp_long < TSK_NULL) {
+        PyErr_SetString(
+            PyExc_ValueError, "tskit ids must be NULL(-1), 0 or a positive number");
+        goto out;
+    }
+
+    id_out[0] = (tsk_id_t) temp_long;
+    ret = 1;
+out:
+    return ret;
+}
+
+static int
 int32_array_converter(PyObject *py_obj, PyArrayObject **array_out)
 {
     int ret = 0;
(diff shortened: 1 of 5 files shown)


benjeffery:
@jeromekelleher Can I get a quick review here before the copy-paste-athon?

python/_tskitmodule.c
```diff
@@ -779,6 +779,31 @@ uint32_converter(PyObject *py_obj, uint32_t *uint_out)
...
+    int ret = 0;
+
+    if (!PyArg_Parse(py_obj, "L", &temp_long)) {
+        goto out;
+    }
+    if (temp_long > TSK_MAX_ID) {
+        PyErr_SetString(PyExc_OverflowError, "Value too large for tskit id type");
+        goto out;
+    }
+    if (temp_long < TSK_NULL) {
```
jeromekelleher, line 795: Not sure this is worth checking at this level, but I don't feel strongly about it.

python/_tskitmodule.c
```diff
@@ -865,14 +890,17 @@ IndividualTable_init(IndividualTable *self, PyObject *args, PyObject *kwds)
 }
 
 static PyObject *
-IndividualTable_add_row(IndividualTable *self, PyObject *args, PyObject *kwds)
+IndividualTable_add_or_update_row(IndividualTable *self, PyObject *args, PyObject *kwds)
```
jeromekelleher, line 893: I don't like this pattern - trying to back-infer how a method was called always seems like a bad idea, and leads to convoluted logic.

jeromekelleher (commented):
Hmm, not sure about this pattern. I see what you're trying to do though, and avoiding duplicating all the arg parsing logic is definition worth doing.

Let me have a think about it and see if I can come up with anything better. So, hold off on the copy-paste-athon for a bit if you don't mind, please.

jeromekelleher:
I've had a look at this, and actually the IndividualTable is the wrong place to start as it's by far and away the most complicated example. For the other tables, it'll be much simpler to just copy and paste and tweak in the obvious way than to try and factor out the common code. Try seeing what it looks like for EdgeTable, and then consider which option (i.e., trying to factor out the argument parsing code, or just copy-paste-altering) would be easier for a new contributor to understand.

python/_tskitmodule.c
benjeffery: I've done this this way as it should cope with both 32 and 64 bit ids should we wish to support that in future.

python/_tskitmodule.c
```diff
@@ -1981,6 +2125,45 @@ EdgeTable_add_row(EdgeTable *self, PyObject *args, PyObject *kwds)
...
 
+static PyObject *
+EdgeTable_update_row(EdgeTable *self, PyObject *args, PyObject *kwds)
+{
+    PyObject *ret = NULL;
+    int err;
+    tsk_id_t row_index = -1;
+    double left = 0.0;
+    double right = 1.0;
+    int parent;
```
jeromekelleher, line 2136: might as well save a big of vspace here, ``int parent, child;``. Same for left and right, we're not using the values we're setting, so no point in setting them.

python/_tskitmodule.c
```diff
@@ -1981,6 +2125,45 @@ EdgeTable_add_row(EdgeTable *self, PyObject *args, PyObject *kwds)
...
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O&ddii|O", kwlist, &tsk_id_converter,
+            &row_index, &left, &right, &parent, &child, &py_metadata)) {
+        goto out;
+    }
+    if (py_metadata != Py_None) {
+        if (PyBytes_AsStringAndSize(py_metadata, &metadata, &metadata_length) < 0) {
+            goto out;
+        }
+    }
+    err = tsk_edge_table_update_row(
```
jeromekelleher, line 2156: Should probably cast the ints to ``tsk_id_t``, and metadata_length to ``tsk_size_t``. (No need to go over old code, but for new code good to do it more correctly)

jeromekelleher (commented):
LGTM, minor points on type safety

python/_tskitmodule.c
jeromekelleher: Or, maybe we should just use your nice ``tsk_id_converter`` on all the tsk_id_t values, and save this annoying little dance?

python/_tskitmodule.c
benjeffery: Yep, done in a couple of the methods

python/_tskitmodule.c
benjeffery: I've used `tsk_id_converter` and cast the string lengths where needed (including ancestral_state etc.)

benjeffery: I've used `tsk_id_converter` and cast the string lengths where needed (including ancestral_state etc.)

jeromekelleher (commented):
Looks good, thanks @benjeffery. Unfortunately I think we do need to get a bit more test coverage on the argument parsing error and default handling - perhaps we could add a handful of update_row invocations to where we're doing this stuff for add_row? Exhastive testing of all the options would be really tedious and a bit pointless, but we want some coverage on the error cases for the new methods (I see we do have good tests for the row_id case, all right though, which does give a lot of assurance that the error handling code is working properly).

I'm paranoid about this error handling stuff as it has turned into a real time-sink in the past trying to track down segfaults caused by weird behaviour in small mistakes with ArgParseTuple

benjeffery:
>  perhaps we could add a handful of update_row invocations to where we're doing this stuff for add_row?

In that case I'm tempted to update the old `add_row` code so that the tests can be identical.

jeromekelleher:
Changes look good, great to cut down on our technical dept a bit like this.

benjeffery:
Ok, I think this is good to go. I didn't add the update tests where we are doing add row tests as they are in the higher-level tests. I also spotted some missing `add_row` tests.

jeromekelleher (approved):
Looks great, thanks @benjeffery!

PR #1600, opened 2021-07-30
Add __setitem__ to tables.

benjeffery:
Stacked on #1597

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -344,7 +344,7 @@ class BaseTable:
         is True. Note that as the result is a new table, the row ids will change as tskit
         row ids are row indexes.
 
-        :param index: the zero-index of a desired row, a slice of the desired rows, an
+        :param index: the index of a desired row, a slice of the desired rows, an
             iterable or array of the desired row numbers, or a boolean array to use as
             a mask.
         """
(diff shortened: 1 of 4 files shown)


benjeffery:
Missing some coverage here, will fix.

python/tskit/tables.py
```diff
@@ -374,6 +374,43 @@ def __getitem__(self, index):
 
         return ret
 
+    def __setitem__(self, index, new_row):
+        """
+        Replaces a row of this table at the specified index with information from a
+        row-like object. Metadata, will be validated and encoded according to the table's
+        :attr:`metadata_schema<tskit.IndividualTable.metadata_schema>`.
+
+        :param index: the zero-index of the row to change
```
jeromekelleher, line 383: Maybe delete the zero here? All indexes are zero-based, might just be confusing.

python/tskit/tables.py
```diff
@@ -374,6 +374,43 @@ def __getitem__(self, index):
...
+            e.g. ``ts.individual(i)`` work for this purpose, along with any other
+            object with the correct attributes.
+        """
+        if isinstance(index, numbers.Integral):
+            # Single row by integer
+            if index < 0:
+                index += len(self)
+            if index < 0 or index >= len(self):
+                raise IndexError("Index out of bounds")
+        else:
```
jeromekelleher, line 395: Let's just think ahead here to make sure we're not painting ourselves into a corner. What would the semantics be for other types of updates? I guess we could support other types of assignment here.

If the index argument is a slice/boolean array, then the ``new_row`` argument should be a Sequence of row_like values (of the same length as the index argument), and then we sequentially call ``update_row`` on these? Wouldn't be particularly efficient, but we could push the implementation down into C if we wanted to, and it would allow us to do things like

```python
tables.edges[:] = tables.edges[::-1]
```

which I'm sure would be handy. 

I'm not suggesting we do this now, just making sure we leave the door open for it in the future. I think TypeError on anything that's not an Integral for the index value does this?

python/tskit/tables.py
```diff
@@ -374,6 +374,43 @@ def __getitem__(self, index):
...
+            raise TypeError("Index must be integer")
+
+        row_data = {
+            column: getattr(new_row, column)
+            for column in self.column_names
+            if "_offset" not in column
+        }
+
+        # Encode the meatdata - note that if this becomes a perf bottleneck it is
```
jeromekelleher, line 404: "meatdata"

jeromekelleher (approved):
LGTM! Minor comments.

python/tskit/tables.py
benjeffery: This was cut and pasted from `__getitem__` so I've removed it there also.

python/tskit/tables.py
benjeffery: Yep, that all makes sense. The code as-is raises `TypeError` on anything non-`Integral`.

python/tskit/tables.py
benjeffery: Must have been hungry when writing this, fixed.

PR #1601, opened 2021-07-30
Add note to 'at' methods

benjeffery:
Seeing https://github.com/tskit-dev/tskit/discussions/1595 I thought a note should be added to the random access methods to warn about the non-constant runtime.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -808,6 +808,14 @@ class Tree:
         index in the parent tree sequence. Negative indexes following the
         standard Python conventions are allowed, i.e., ``index=-1`` will
         seek to the last tree in the sequence.
+        |LinearTraversalWarning|
+
+        .. |LinearTraversalWarning| replace:: warning::
+           The current implementation of this operation is linear in the number of
+           trees, so may be inefficient for large tree sequences. See
+           <this issue <https://github.com/tskit-dev/tskit/issues/684>_ for more
+           information.
+
 
         :param int index: The tree index to seek to.
         :raises IndexError: If an index outside the acceptable range is provided.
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -4186,6 +4186,9 @@ def at(self, position, **kwargs):
         Returns the tree covering the specified genomic location. The returned tree
         will have ``tree.interval.left`` <= ``position`` < ``tree.interval.right``.
         See also :meth:`Tree.seek`.
+        Note that as tree sequences are stored tree-by-tree this method has to pass
```
jeromekelleher, line 4189: Maybe a warning block would stand out a bit more? Something like "The current implementation of this operation is linear in the number of trees, so may be inefficient for large tree sequences. See `<this issue <https://github.com/tskit-dev/tskit/issues/684>`_ for more information."

It's nothing fundamental, we just haven't implemented the necessary indexing to do it.

jeromekelleher (commented):
Good idea - I think a warning block will be more effective at getting people to notice. We should probably also put this warning on Tree.seek and seek_index. (Could use a sphinx replacement to avoid the duplication, I think I did one for the stats API somewhere.)

python/tskit/trees.py
```diff
@@ -4202,6 +4212,10 @@ def at(self, position, **kwargs):
     def at_index(self, index, **kwargs):
         """
         Returns the tree at the specified index. See also :meth:`Tree.seek_index`.
+        Note that as tree sequences are stored tree-by-tree this method has to pass
```
jeromekelleher, line 4215: Did you mean to keep this sentence as well as LinearTraversalWarning?

benjeffery: I'm still arguing with sphinx syntax...

benjeffery:
@jeromekelleher Annoyingly my local sphinx won't build, so I can't check that there isn't a double warning where the repeated section is defined.

jeromekelleher:
Might be easier to follow the pattern for ``tree_array_warning`` at the top of python-api.rst?

benjeffery:
I've managed to get this to build and it works as intended.

jeromekelleher (approved):

PR #1602, opened 2021-08-01
TableCollection force_offset_64

### LINKED ISSUE #1598: Add force_offset64 option to TableCollection.asdict
To keep compatibility with existing code, the TableCollection.asdict method will need to return 32 bit offset values, like the lightweight TableCollection. So, as another step on the 64 bit road, we should add the option to force returning 64 bit offset values. This is actually trickier then for the LightweightTableCollection, because we're getting the table columns via attribute accesses rather than method calls, and so will have to change some of the low-level infrastructure.

Another tricky question arises then: when someone does a direct access to (say) ``NodeTable.metadata_offset`` should they get a 64 bit value, or a 32/64 bit value depending on the size? I think probably 64 bit, but it may turn out that this breaks too much code. We'll see, I guess.

jeromekelleher:
Closes #1598

### DIFF
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -408,6 +408,37 @@ class TestTableCollection(LowLevelTestCase):
         with pytest.raises(TypeError):
             tc.equals(tc, ignore_timestamps=bad_bool)
 
+    def test_asdict(self):
+        for ts in self.get_example_tree_sequences():
+            tc = _tskit.TableCollection(sequence_length=ts.get_sequence_length())
+            ts.dump_tables(tc)
+            d = tc.asdict()
+            # Method is tested extensively elsewhere, just basic sanity check here
+            assert isinstance(d, dict)
+            assert len(d) > 0
+
+    def test_fromdict(self):
+        for ts in self.get_example_tree_sequences():
+            tc1 = _tskit.TableCollection(sequence_length=ts.get_sequence_length())
+            ts.dump_tables(tc1)
+            d = tc1.asdict()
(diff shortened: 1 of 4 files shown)


jeromekelleher:
This should be ready to go @benjeffery. The last commit is a bit of a drive-by, but I spotted a case where we were rerunning a bunch of tests for no benefit, and it was an easy fix.

python/tests/test_lowlevel.py
```diff
@@ -408,6 +408,37 @@ def test_equals_bad_args(self):
...
+        for ts in self.get_example_tree_sequences():
+            tc1 = _tskit.TableCollection(sequence_length=ts.get_sequence_length())
+            ts.dump_tables(tc1)
+            d = tc1.asdict()
+            tc2 = _tskit.TableCollection(sequence_length=0)
+            tc2.fromdict(d)
+            assert tc1.equals(tc2)
+
+    def test_asdict_bad_args(self):
+        ts = msprime.simulate(10, random_seed=1242)
```
benjeffery, line 430: You could use `simple_degree1_ts_fixture` here, but I don't feel strongly about it.

benjeffery (approved):
Excellent, we're almost there with 64bit offsets!

benjeffery:
I've also run `stress_lowlevel` here. Seems all good at 200 iterations, will leave it running a bit more.

python/tests/test_lowlevel.py
jeromekelleher: I was following the conventions in the rest of the class, so I don't think it's worth changing to a fixture here.

PR #1611, opened 2021-08-04
Use new ts __str__ for cli info command

benjeffery:
The current CLI `info` could be improved by using the recently added `ts.__str__`.

This will break any code that was using the CLI to get any info - but I don't think that is likely.


jeromekelleher (approved):
LGTM

PR #1615, opened 2021-08-05
Added docstring for find_ibd API.

gtsambos:
I (finally) added some API documentation for `find-ibd`.

Some eyes over it would be useful if you have some spare brainspace: @petrelharp, @molpopgen, @jeromekelleher perhaps?

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3451,6 +3451,61 @@ class TableCollection:
             )
 
     def find_ibd(self, samples, max_time=None, min_length=None):
+        """
+        Returns a dictionary containing all IBD (identical-by-descent) segments for the
+        supplied list of sample pairs.
+
+        A pair of nodes ``(u, v)`` has an IBD segment with a left and right coordinate
+        ``[left, right)`` and ancestral node ``a`` iff the most recent common ancestor
+        of the segment ``[left, right)`` in nodes ``u`` and ``v`` is ``a``, and the
+        segment has been inherited along the same genealogical path (ie. it has not
+        been broken by recombination). The segments returned are the longest possible
+        ones.
+
+        Note that this definition is purely genealogical -- allelic states
+        *are not* considered here. If used without time or length thresholds, the
+        segments returned for a given pair will partition the span of the contig
+        represented by the tree sequence.
+
+        Each item in the outputted dictionary object contains the IBD segments for a
+        particular sample pair. Suppose the output is
+
+        .. code-block:: python
+
+            {
+                (0, 1): {
+                    "left": array([0.0]),
+                    "right": array([1.0]),
+                    "node": array([4]),
+                },
+                (0, 2): {
+                    "left": array([0.5, 0.0]),
+                    "right": array([1.0, 0.5]),
+                    "node": array([3, 5]),
+                },
+            }
+
+        This indicates that samples 0 and 1 share a single IBD segment on the interval
+        `[0.0, 1.0)` inherited from the ancestor with node ID 4. Samples 0 and 2 share
+        one IBD segment on the interval `[0.5, 1.0)` inherited from node 3, and another
+        segment on the interval `[0.0, 0.5)` inherited from node 5.
+
+        :param list samples: A list of pairs of integers, with each pair specified as
+            a tuple. In each pair, the integers correspond to the node IDs of a pair of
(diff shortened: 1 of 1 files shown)


python/tskit/tables.py
```diff
@@ -3451,6 +3451,44 @@ def union(
...
+
+        Note that this definition is purely genealogical -- allelic states
+        *are not* considered here. If used without time or length thresholds, the
+        segments returned for a given pair will partition the span of the contig
+        represented by the tree sequence.
+
+        :param list samples: A list of pairs of integers, with each pair specified as
+            a tuple. In each pair, the integers correspond to the node IDs of a pair of
+            chromosomes whose IBD segments will be returned by
+            :meth:`.TableCollection.find_ibd`.
```
jeromekelleher, line 3473: Might as well say "this method" - not much point in linking back to the same place.

python/tskit/tables.py
```diff
@@ -3451,6 +3451,44 @@ def union(
...
+        :param float max_time: Only segments inherited from common
+            ancestors whose node times are more recent than the specified time
+            will be returned. Specifying a maximum time is strongly recommended when
+            working with large tree sequences.
+        :param float min_length: Only segments longer than the specified length
+            will be returned.
+        :return: A dictionary object indexed by the specified sample pairs. The value
+            of each item is itself a dictionary indexed by the keys ``left``, ``right``
+            and ``node``. If the outputted object is saved to a variable called ``out``
+            and ``(u, v)`` is a specified sample pair, then
```
jeromekelleher, line 3483: Rather than putting this detail here in the return block, I wonder if it would be better to have another paragraph above in the description where we describe the output (maybe giving an example of the output?), and reserve this for describing the types. We should probably be precise about the dtypes as well (float64, and int32).

jeromekelleher (commented):
Looks great. I think an example would help, where we show the dictionary output for some example and explain how to interpret it.

python/tskit/tables.py
```diff
@@ -3467,26 +3467,31 @@ def find_ibd(self, samples, max_time=None, min_length=None):
         segments returned for a given pair will partition the span of the contig
         represented by the tree sequence.
 
+        Each item in the outputted dictionary object contains the IBD segments for a
+        particular sample pair. Suppose the output is
```
jeromekelleher, line 3471: I don't think this will render well in rst (the backticks is Markdown syntax afaik) - have you looked at the output?

python/tskit/tables.py
```diff
@@ -3451,6 +3451,49 @@ def union(
...
+        *are not* considered here. If used without time or length thresholds, the
+        segments returned for a given pair will partition the span of the contig
+        represented by the tree sequence.
+
+        Each item in the outputted dictionary object contains the IBD segments for a
+        particular sample pair. Suppose the output is
+        ```{(0, 1): {'left': array([0.]), 'right': array([1.]), 'node': array([4])},
+        (0, 2): {'left':array([0.5, 0.0]), 'right':array([1.0, 0.5]),
+        'node':array([3, 5])}}
+        ```
```
benjeffery, line 3475: In the output this line is included without a linebreak, so is very long - using a code block seems to work better:
```suggestion

        .. code-block:: python

            {
                (0, 1): {
                    "left": array([0.0]),
                    "right": array([1.0]),
                    "node": array([4]),
                },
                (0, 2): {
                    "left": array([0.5, 0.0]),
                    "right": array([1.0, 0.5]),
                    "node": array([3, 5]),
                },
            }

```

benjeffery (commented):
Great! One suggestion about the example.

I also wondered why this method is on `TableCollection` and not `TreeSequence` where it more naturally belongs in my mind?

python/tskit/tables.py
gtsambos: thanks @benjeffery, this solves all of the above problems in one!

gtsambos:
> I also wondered why this method is on TableCollection and not TreeSequence where it more naturally belongs in my mind?

Yep, this has been brought up by others too @benjeffery -- I will make a wrapper for find_ibd that works on the TreeSequence objects too.

jeromekelleher (approved):
LGTM! Ready for merge after a squash I think.

benjeffery (approved):

petrelharp:
Late here, but this looks great.

PR #1620, opened 2021-08-10
c library shims

jeromekelleher:
Add tsk_X shims for functions like malloc, memcpy as part of the transition to 64 bit tsk_size_t.


benjeffery (approved):
Looks good!

Is there a reason `tskit/tables.c:154` and `371` aren't switched over?

`10344` in the same file still has the zero malloc guard logic that can be removed.

jeromekelleher:
thanks for the close review @benjeffery - great spot on both counts! Should be good to go now.

PR #1622, opened 2021-08-11
Added a TreeSequence wrapper for find_ibd.

### LINKED ISSUE #1617: Make find_ibd a method for TreeSequence objects.
See #1615 -- this would be convenient.

gtsambos:
Closes #1617.

I only made very basic tests as the find_ibd method on tables has already been tested thoroughly.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3452,43 +3452,17 @@ class TableCollection:
 
     def find_ibd(self, samples, max_time=None, min_length=None):
         """
-        Returns a dictionary containing all IBD (identical-by-descent) segments for the
-        supplied list of sample pairs.
+        Equivalent to the :meth:`TreeSequence.find_ibd` method; please see its
+        documentation for more details, and use this method only if you specifically need
+        to work with a :class:`TableCollection` object.
 
-        A pair of nodes ``(u, v)`` has an IBD segment with a left and right coordinate
-        ``[left, right)`` and ancestral node ``a`` iff the most recent common ancestor
-        of the segment ``[left, right)`` in nodes ``u`` and ``v`` is ``a``, and the
-        segment has been inherited along the same genealogical path (ie. it has not
-        been broken by recombination). The segments returned are the longest possible
-        ones.
-
-        Note that this definition is purely genealogical -- allelic states
-        *are not* considered here. If used without time or length thresholds, the
-        segments returned for a given pair will partition the span of the contig
-        represented by the tree sequence.
-
-        Each item in the outputted dictionary object contains the IBD segments for a
-        particular sample pair. Suppose the output is
-
-        .. code-block:: python
-
-            {
-                (0, 1): {
-                    "left": array([0.0]),
-                    "right": array([1.0]),
-                    "node": array([4]),
-                },
-                (0, 2): {
-                    "left": array([0.5, 0.0]),
-                    "right": array([1.0, 0.5]),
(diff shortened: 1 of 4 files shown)


python/tskit/trees.py
```diff
@@ -7160,6 +7160,30 @@ def count_topologies(self, sample_sets=None):
 
         yield from combinatorics.treeseq_count_topologies(self, sample_sets)
 
+    def find_ibd(self, samples, max_time=None, min_length=None):
+        """
+        A wrapper for :meth:`TableCollection.find_ibd`. See this method for more info.
```
petrelharp, line 7165: Most methods have the TreeSequence method the "primary" one, I think - perhaps the main docs should live here?

gtsambos: Good point! I don't have strong feelings either way, I'm happy to shift the main docs over there. Will update the PR in the afternoon.

python/tskit/tables.py
```diff
@@ -3452,43 +3452,7 @@ def union(
...
-                    "right": array([1.0, 0.5]),
-                    "node": array([3, 5]),
-                },
-            }
-
-        This indicates that samples 0 and 1 share a single IBD segment on the interval
-        `[0.0, 1.0)` inherited from the ancestor with node ID 4. Samples 0 and 2 share
-        one IBD segment on the interval `[0.5, 1.0)` inherited from node 3, and another
-        segment on the interval `[0.0, 0.5)` inherited from node 5.
+        Same as :meth:`TreeSequence.find_ibd`. See this method for more info.
```
jeromekelleher, line 3456: I guess we should write down what the input requirements are - tables must be sorted I guess?

python/tskit/tables.py
```diff
@@ -3452,43 +3452,11 @@ def union(
...
+        The table collection must correspond to a valid :class:`TreeSequence` and must
```
jeromekelleher, line 3456: That's not true I think - the tables don't have to be indexed.

How about
```
Equivalent to the :meth:`TreeSequence.find_ibd` method; please see its documentation for more details.

This method has the same input data requirements as :meth:`TableCollection.simplify`.

File: python/tskit/tables.py [hunk 4]
```diff
@@ -3452,43 +3452,11 @@ def union(
...
+        be sorted. Use :meth:`TableCollection.sort` beforehand to ensure this. To ensure
+        that any adjacent IBD segments are concatenated into a single segment in the
```
jeromekelleher, line 3458: Some will read this as telling them to *always* run ``squash``, but this is only needed in special cases right? (Which are quite remote, I think?)

benjeffery (approved):
LGTM modulo JK's doc comments.

python/tskit/tables.py
gtsambos: I think, now that we've made the 'default' method the `TreeSequence` one, people are only going to use the TableCollection version of this function directly when they've manually constructed the tables themselves. In this case, I think I'd be encouraging them to squash and sort the tables as best practice anyway, because they have probably been deliberately constructed to be special in some way.  Perhaps I could add a comment recommending that people use the TreeSequence version of this method unless they specifically need this one?

jeromekelleher: That's a good idea. Sorting by default wouldn't be a good practise to encourage though, as this is very expensive for large tables.

jeromekelleher (approved):
Awesome, thanks @gtsambos

PR #1623, opened 2021-08-12
Adding a centre=False option to genetic_relatedness statistic

### LINKED ISSUE #1619: new (?) statistic to efficiently calculate shared times between every pair of samples at every tree
Hi all, I've been messing around with different ways to calculate the shared time between every pair of samples in many trees, which is a useful metric because it describes the covariance of characteristics of the samples (e.g., traits, locations) under Brownian motion. At first I couldn't figure out how to use an existing statistic to do this, so I used the ```general_stat``` function to make one of my own:

```python
import numpy as np
def shared_times(ts):
    """Use general_stat function to calculate shared branch lengths between all samples at all trees.
    
    Parameters
    ts: tskit tree sequence
    
    Returns
    A (n x k x k) numpy array, with n the number of trees in the sequence and k the number of sample nodes. 
    At each tree the (k x k) matrix gives the shared evolutionary times between each pair of sample nodes.
    """
    
    k = ts.num_samples #number of samples
    W = np.identity(k) #each node i in [0,1,...,k] given row vector of weights with 1 in column i and 0's elsewhere 
    def f(x): return (x.reshape(-1,1) * x).flatten() #determine which pairs of samples share the branch above a node, flattened

    return ts.general_stat(
        W, f, k**2, mode='branch', windows='trees', polarised=True, strict=False
    ).reshape(ts.num_trees, k, k)
```

For example:

```python
import msprime
ts = msprime.sim_ancestry(samples=5, population_size=1e4, sequence_length=1e4, recombination_rate=1e-8, random_seed=1)
sts = shared_times(ts)
```

The result can also be achieved by simply looping through trees and samples to calculate pairwise TMRCAs:

```python
def shared_times(ts):
    k = ts.num_samples
    sts = np.zeros((ts.num_trees,k,k))
    for t,tree in enumerate(ts.trees()):
        T = tree.time(tree.root)
        for i in range(k):
            for j in range(i):
                st = T - tree.tmrca(i,j)
                sts[t,i,j] = st
                sts[t,j,i] = st
    return sts
```

but this is considerably slower when there are many trees (although it seems to be faster when there are few trees with many samples, e.g., ```ts = msprime.sim_ancestry(samples=2e2, population_size=1e4, random_seed=1)``` -- not sure why).

Coming back to this much later (today), I realized you can use the divergence statistic, with a few tweaks, to do the same:

```python
def shared_times(ts):
    
    k = ts.num_samples
    
    # get 2*tmrcas
    sample_sets = [[i] for i in ts.samples()]
    indexes = [(i,j) for i in ts.samples() for j in range(i,k)] #compare each sample with each other only once (entries of upper triangular)
    divs = ts.divergence(sample_sets=sample_sets, indexes=indexes, mode='branch', windows='trees')
    divs[np.isnan(divs)] = 0 #tmrcas with self are said to be nan, so convert to 0

    # convert to matrices
    divs_mat = np.zeros((ts.num_trees,k,k)) 
    for ix,(i,j) in enumerate(indexes):
        divs_mat[:,i,j] = divs[:,ix] #convert list to upper triangle
        divs_mat[:,j,i] = divs[:,ix] #fill in lower triangle symmetrically
    
    # convert to shared times
    sts = np.zeros(divs_mat.shape)
    for i,div in enumerate(divs_mat):
        sts[i] = (np.max(div) - div)/2 #convert from 2*tmrcas to shared times
    
    return sts
```

This latter method seems to be the fastest for small tree sequences (eg, ```ts = msprime.sim_ancestry(samples=5, population_size=1e4, sequence_length=1e6, recombination_rate=1e-8, random_seed=1)```) but loses this advantage as tree sequences get larger (eg, ```msprime.sim_ancestry(samples=1e2, population_size=1e4, sequence_length=1e6, recombination_rate=1e-8, random_seed=1)```). It also uses >2x the RAM of the first method, which quickly becomes important with larger tree sequences.

I'm wondering if my ```general_stat``` matrix formulation would be useful as a new statistic of its own: related to the ```divergence``` statistic but calculated differently (I think), sometimes faster, sometimes less memory intensive, and semantically simpler (for creating matrices). If so, I guess I'd have to also think about what happens as you vary the options in ```general_stat```, e.g., ```mode='node'```. Anyway, just wanted to put this out there in case it is helpful/useful. Thanks for building all this stuff!

(This is my first time writing an issue so sorry if I've messed up somehow -- I'd be keen to hear how to do this better!)

mmosmond:
## Description

Adding ```centre=False``` option to ```genetic_relatedness``` statistic, so that it can calculate, for example, the non-mean-centred shared times between samples.

So far I've just added the option to the Python test function and a test (which passes). Next step is to develop the C code, but I haven't figured out how to do it. I think a similar change (an option to set mean to 0) needs to be added to [```genetic_relatedness_summary_func```](https://github.com/tskit-dev/tskit/blob/35e7d5f466857b00337d4a0596eb1bbbe69b57a0/c/tskit/trees.c#L2782).  

Fixes #1619 

Thanks for any help you can give @petrelharp or others!

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1306,7 +1306,7 @@ tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
     }
 
     if (parent == NULL || branch_length == NULL || state == NULL || running_sum == NULL
-        || summary == NULL) {
+        || summary == NULL || zero_state == NULL || zero_summary == NULL) {
         ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1297,6 +1297,8 @@ tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
     double *state = tsk_calloc(num_nodes * state_dim, sizeof(*state));
     double *summary = tsk_calloc(num_nodes * result_dim, sizeof(*summary));
     double *running_sum = tsk_calloc(result_dim, sizeof(*running_sum));
+    double *zero_state = tsk_calloc(state_dim, sizeof(*zero_state));
+    double *zero_summary = tsk_calloc(result_dim, sizeof(*zero_state));
 
     if (self->time_uncalibrated && !(options & TSK_STAT_ALLOW_TIME_UNCALIBRATED)) {
         ret = TSK_ERR_TIME_UNCALIBRATED;
@@ -1310,6 +1312,15 @@ tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
     }
     tsk_memset(parent, 0xff, num_nodes * sizeof(*parent));
 
+    /* If f is not strict, we may need to set conditions for non-sample nodes as well. */
+    ret = f(state_dim, zero_state, result_dim, zero_summary, f_params);
+    if (ret != 0) {
+        goto out;
+    }
+    for (j = 0; j < num_nodes; j++) { // we could skip this if zero_summary is zero
+        summary_u = GET_2D_ROW(summary, result_dim, j);
+        tsk_memcpy(summary_u, zero_summary, result_dim * sizeof(*zero_summary));
+    }
     /* Set the initial conditions */
     for (j = 0; j < self->num_samples; j++) {
         u = self->samples[j];
@@ -1322,6 +1333,7 @@ tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
             goto out;
         }
     }
+
     tsk_memset(result, 0, num_windows * result_dim * sizeof(*result));
 
     /* Iterate over the trees */
@@ -1425,6 +1437,8 @@ out:
     tsk_safe_free(state);
     tsk_safe_free(summary);
     tsk_safe_free(running_sum);
+    tsk_safe_free(zero_state);
+    tsk_safe_free(zero_summary);
     return ret;
 }
 
@@ -4542,21 +4556,39 @@ genetic_relatedness_summary_func(tsk_size_t state_dim, const double *state,
     tsk_id_t i, j;
     tsk_size_t k;
     double sumx = 0;
-    double sumn = 0;
     double meanx, ni, nj;
 
     for (k = 0; k < state_dim; k++) {
-        sumx += x[k];
-        sumn += (double) args.sample_set_sizes[k];
+        sumx += x[k] / (double) args.sample_set_sizes[k];
     }
 
-    meanx = sumx / sumn;
+    meanx = sumx / (double) state_dim;
     for (k = 0; k < result_dim; k++) {
         i = args.set_indexes[2 * k];
         j = args.set_indexes[2 * k + 1];
         ni = (double) args.sample_set_sizes[i];
         nj = (double) args.sample_set_sizes[j];
-        result[k] = (x[i] - ni * meanx) * (x[j] - nj * meanx) / 2;
+        result[k] = (x[i] / ni - meanx) * (x[j] / nj - meanx);
+    }
+    return 0;
+}
+
+static int
+genetic_relatedness_noncentred_summary_func(tsk_size_t TSK_UNUSED(state_dim),
+    const double *state, tsk_size_t result_dim, double *result, void *params)
+{
+    sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
+    const double *x = state;
+    tsk_id_t i, j;
+    tsk_size_t k;
+    double ni, nj;
+
+    for (k = 0; k < result_dim; k++) {
+        i = args.set_indexes[2 * k];
+        j = args.set_indexes[2 * k + 1];
+        ni = (double) args.sample_set_sizes[i];
+        nj = (double) args.sample_set_sizes[j];
+        result[k] = x[i] * x[j] / (ni * nj);
     }
     return 0;
 }
@@ -4572,9 +4604,16 @@ tsk_treeseq_genetic_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample
     if (ret != 0) {
         goto out;
     }
-    ret = tsk_treeseq_sample_count_stat(self, num_sample_sets, sample_set_sizes,
-        sample_sets, num_index_tuples, index_tuples, genetic_relatedness_summary_func,
-        num_windows, windows, options, result);
+    if (!(options & TSK_STAT_NONCENTRED)) {
+        ret = tsk_treeseq_sample_count_stat(self, num_sample_sets, sample_set_sizes,
+            sample_sets, num_index_tuples, index_tuples,
+            genetic_relatedness_summary_func, num_windows, windows, options, result);
+    } else {
+        ret = tsk_treeseq_sample_count_stat(self, num_sample_sets, sample_set_sizes,
+            sample_sets, num_index_tuples, index_tuples,
+            genetic_relatedness_noncentred_summary_func, num_windows, windows, options,
+            result);
+    }
 out:
     return ret;
 }
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2086,6 +2086,7 @@ typedef struct {
 } sample_count_stat_params_t;
 
 typedef struct {
+    tsk_size_t num_samples;
     double *total_weights;
     const tsk_id_t *index_tuples;
 } indexed_weight_stat_params_t;
@@ -4626,15 +4627,32 @@ genetic_relatedness_weighted_summary_func(tsk_size_t state_dim, const double *st
     const double *x = state;
     tsk_id_t i, j;
     tsk_size_t k;
-    double meanx, ni, nj;
+    double pn, ni, nj;
 
-    meanx = state[state_dim - 1] / args.total_weights[state_dim - 1];
+    pn = state[state_dim - 1];
     for (k = 0; k < result_dim; k++) {
         i = args.index_tuples[2 * k];
         j = args.index_tuples[2 * k + 1];
         ni = args.total_weights[i];
         nj = args.total_weights[j];
-        result[k] = (x[i] - ni * meanx) * (x[j] - nj * meanx) / 2;
+        result[k] = (x[i] - ni * pn) * (x[j] - nj * pn);
+    }
+    return 0;
+}
+
+static int
+genetic_relatedness_weighted_noncentred_summary_func(tsk_size_t TSK_UNUSED(state_dim),
+    const double *state, tsk_size_t result_dim, double *result, void *params)
+{
+    indexed_weight_stat_params_t args = *(indexed_weight_stat_params_t *) params;
+    const double *x = state;
+    tsk_id_t i, j;
+    tsk_size_t k;
+
+    for (k = 0; k < result_dim; k++) {
+        i = args.index_tuples[2 * k];
+        j = args.index_tuples[2 * k + 1];
+        result[k] = x[i] * x[j];
     }
     return 0;
 }
@@ -4672,17 +4690,26 @@ tsk_treeseq_genetic_relatedness_weighted(const tsk_treeseq_t *self,
             new_row[k] = row[k];
             total_weights[k] += row[k];
         }
-        new_row[num_weights] = 1.0;
+        new_row[num_weights] = 1.0 / (double) num_samples;
     }
-    total_weights[num_weights] = (double) num_samples;
+    total_weights[num_weights] = 1.0;
 
     args.total_weights = total_weights;
     args.index_tuples = index_tuples;
-    ret = tsk_treeseq_general_stat(self, num_weights + 1, new_weights, num_index_tuples,
-        genetic_relatedness_weighted_summary_func, &args, num_windows, windows, options,
-        result);
-    if (ret != 0) {
-        goto out;
+    if (!(options & TSK_STAT_NONCENTRED)) {
+        ret = tsk_treeseq_general_stat(self, num_weights + 1, new_weights,
+            num_index_tuples, genetic_relatedness_weighted_summary_func, &args,
+            num_windows, windows, options, result);
+        if (ret != 0) {
+            goto out;
+        }
+    } else {
+        ret = tsk_treeseq_general_stat(self, num_weights + 1, new_weights,
+            num_index_tuples, genetic_relatedness_weighted_noncentred_summary_func,
+            &args, num_windows, windows, options, result);
+        if (ret != 0) {
+            goto out;
+        }
     }
 
 out:
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -54,6 +54,9 @@ extern "C" {
 #define TSK_STAT_ALLOW_TIME_UNCALIBRATED (1 << 12)
 #define TSK_STAT_PAIR_NORMALISE          (1 << 13)
 
+/* Stat-specific options */
+#define TSK_STAT_NONCENTRED         (1 << 12)
+
 /* Options for map_mutations */
 #define TSK_MM_FIXED_ANCESTRAL_STATE (1 << 0)
 
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -53,9 +53,7 @@ extern "C" {
 #define TSK_STAT_SPAN_NORMALISE          (1 << 11)
 #define TSK_STAT_ALLOW_TIME_UNCALIBRATED (1 << 12)
 #define TSK_STAT_PAIR_NORMALISE          (1 << 13)
-
-/* Stat-specific options */
-#define TSK_STAT_NONCENTRED         (1 << 12)
+#define TSK_STAT_NONCENTRED              (1 << 14)
 
 /* Options for map_mutations */
 #define TSK_MM_FIXED_ANCESTRAL_STATE (1 << 0)
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -560,11 +560,12 @@ below, and above each branch or node; if polarised, then only the weight below i
 
 ### Strictness, and which branches count?
 
-Most statistics are not be affected by invariant sites,
+Most statistics are not affected by invariant sites,
 and hence do not depend on any part of the tree that is not ancestral to any of the sample sets.
-However, some statistics are different: for instance, {meth}`TreeSequence.genetic_relatedness`
+However, some statistics are different: for instance, 
+given a pair of samples, {meth}`TreeSequence.genetic_relatedness`
 with `centre=False` (and `polarised=True`, the default for that method)
-adds up for a pair of samples the total number of alleles (or total area of branches) that is
+adds up the total number of alleles (or total area of branches) that is
 either ancestral to both samples *or ancestral to neither*.
 So, it depends on what else is in the tree sequence.
 (For this reason, we don't recommend actually *using* this combination of options for genetic
@@ -604,7 +605,7 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
 
   unless the two indices are the same, when the diversity function is used.
 
-  For an unpolarized statistic with biallelic loci, this calculates
+  For an unpolarised statistic with biallelic loci, this calculates
   {math}`p_1 (1-p_2) + (1 - p_1) p_2`.
 
 `genetic_relatedness, centre=True`
@@ -612,8 +613,9 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
 
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
   of sample sets.
-  For an polarised statistic (the default) with biallelic loci, this calculates
-  {math}`(p_1 - \bar{p}) (p_2 - \bar{p})`.
+  For a polarised statistic (the default) with biallelic loci, this calculates
+  {math}`(p_1 - \bar{p}) (p_2 - \bar{p})`, where {math}`\bar{p}` is the average
+  derived allele frequency across sample sets.
 
 `genetic_relatedness, centre=False`
 : {math}`f(x_i, x_j) = (x_i / n_i) (x_j / n_j)`.
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -607,18 +607,30 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
   For an unpolarized statistic with biallelic loci, this calculates
   {math}`p_1 (1-p_2) + (1 - p_1) p_2`.
 
-`genetic_relatedness`
-: {math}`f(x_i, x_j) = \frac{1}{2}(x_i - m)(x_j - m)`,
+`genetic_relatedness, centre=False`
+: {math}`f(x_i, x_j) = (x_i / n_i) (x_j / n_j)`,
 
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
   of samples.
 
-`genetic_relatedness_weighted`
-: {math}`f(w_i, w_j, x_i, x_j) = \frac{1}{2}(x_i - w_i m) (x_j - w_j m)`,
+`genetic_relatedness, centre=True`
+: {math}`f(x_i, x_j) = (x_i / n_i - m)(x_j / n_j - m)`,
 
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
-  of samples, and {math}`w_j = \sum_{k=1}^n W_kj` is the sum of the weights in the {math}`j`th column of the weight matrix.
-  
+  of sample sets.
+
+`genetic_relatedness_weighted, centre=True`
+: {math}`f(w_i, w_j, x_i, x_j) = (x_i - w_i p) (x_j - w_j p)`,
+
+  where {math}`p` is the proportion of all samples below the focal node,
+  and {math}`w_j = \sum_{k=1}^n W_{kj}` is the sum of the weights in the {math}`j`th column of the weight matrix.
+
+`genetic_relatedness_weighted, centre=False`
+: {math}`f(w_i, w_j, x_i, x_j) = x_i x_j`,
+
+  where {math}`p` is the proportion of all samples below the focal node,
+  and {math}`w_j = \sum_{k=1}^n W_{kj}` is the sum of the weights in the {math}`j`th column of the weight matrix.
+
 `Y2`
 : {math}`f(x_1, x_2) = \frac{x_1 (n_2 - x_2) (n_2 - x_2 - 1)}{n_1 n_2 (n_2 - 1)}`
 
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -556,6 +556,25 @@ associated with each allele; but if polarised, then the ancestral allele is left
 For branch or node statistics, summary functions are applied to the total weight or number of samples
 below, and above each branch or node; if polarised, then only the weight below is used.
 
+(sec_stats_strictness)=
+
+### Strictness, and which branches count?
+
+Most statistics are not be affected by invariant sites,
+and hence do not depend on any part of the tree that is not ancestral to any of the sample sets.
+However, some statistics are different: for instance, {meth}`TreeSequence.genetic_relatedness`
+with `centre=False` (and `polarised=True`, the default for that method)
+adds up for a pair of samples the total number of alleles (or total area of branches) that is
+either ancestral to both samples *or ancestral to neither*.
+So, it depends on what else is in the tree sequence.
+(For this reason, we don't recommend actually *using* this combination of options for genetic
+relatedness.)
+
+In terms of the summary function {math}`f(x)`, "not affected by invariant sites" translates to
+{math}`f(0) = f(n) = 0`, where {math}`n` is the vector of sample set sizes.
+By default, {meth}`TreeSequence.general_stat` checks if the summary function satisfies this condition,
+and throws an error if not; this check can be disabled by setting `strict=False`.
+
 
 (sec_stats_summary_functions)=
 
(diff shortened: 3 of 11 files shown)


petrelharp:
Ok, this should do it, I think? The way the option gets passed through feels a little inelegant, but I think is ok.

We should think through the noncentred version a bit - what's the use case? Note that it will depend on, for instance, the number of fixed, derived mutations, which is a property we like statistics to not have.

mmosmond:
This looks like the right thing to me. One semantic note though, should the new method you made be called ```genetic_relatedness_noncentred_summary_func``` rather than ```genetic_relatedness_centred_summary_func```?

I was trying to get this to run in python and there seems to be a problem there. I'm not sure if this is correct but I added the ```centred=T/F``` option to ```__k_way_sample_set_stat``` in ```trees.py```, but then there seems to be an error in parsing the keyword arguments, and I'm not sure what to do to fix it. More specifically, 

```python
import tskit, msprime

ts = msprime.sim_ancestry(samples=10, sequence_length=10, recombination_rate=1, random_seed=1)

n = ts.num_samples
sample_sets = [[i] for i in ts.samples()]
indexes = [(i,j) for i in ts.samples() for j in ts.samples()]

ts.genetic_relatedness(sample_sets, indexes, mode='branch', windows='trees', span_normalise=True, proportion=False, centred=False)
```

gives ```SystemError: More keyword list entries (8) than format specifiers (7)```, where the error comes from line 5883 in ```trees.py```.

On the use case, getting the shared times between sample lineages is what I am interested in, eg, to get the MLE of the rate of evolution under Brownian motion. And even if you want to eventually mean centre things, getting the noncentred version is essential if you were interested in getting the shared times in multiple/particular epochs, eg, to let the rate of evolution be piecewise constant.

petrelharp:
> getting the shared times between sample lineages is what I am interested in, eg, to get the MLE of the rate of evolution under Brownian motion. 

Hm - I'm not sure what you mean (but, I believe you). Ah, I think I see - you must be doing this separately on each tree? (ie with `windows='trees'`)?

> if you were interested in getting the shared times in multiple/particular epochs, eg, to let the rate of evolution be piecewise constant.

It's on our list to add time windows to all the stats - it sounds like you might need this?

mmosmond:
Thanks @petrelharp -- it would have taken me a long time to figure out that format specifier. 

Yes, I'm interested in doing this tree by tree.

And yeah, a time-window sounds very convenient for what I'd like to do (and then I should be able to just work with the centred version).

I did a test to make sure this gives me what I think it should:
```python
import tskit, msprime
import numpy as np

# run a small sim
ts = msprime.sim_ancestry(samples=3, sequence_length=1e3, recombination_rate=1e-3, random_seed=1)

# calculate two different measures of genetic_relatedness
n = ts.num_samples
sample_sets = [[i] for i in ts.samples()]
indexes = [(i,j) for i in ts.samples() for j in ts.samples()]
#centred version, NONPOLARISED
sts_grel_mc = ts.genetic_relatedness(sample_sets, indexes, mode='branch', windows='trees', span_normalise=True, polarised=False, proportion=False, centred=True).reshape(ts.num_trees, n, n) 
#noncentred version, POLARISED
sts_grel = ts.genetic_relatedness(sample_sets, indexes, mode='branch', windows='trees', span_normalise=True, polarised=True, proportion=False, centred=False).reshape(ts.num_trees, n, n) 

# my functions to get shared times with general_stat and mean centre them, for comparison with above
def shared_times(ts):
    k = ts.num_samples #number of samples
    W = np.identity(k) #each node i in [0,1,...,k] given row vector of weights with 1 in column i and 0's elsewhere 
    def f(x): return (x.reshape(-1,1) * x).flatten() #determine which pairs of samples share the branch above a node, flattened
    return ts.general_stat(
        W, f, k**2, mode='branch', windows='trees', polarised=True, strict=False
    ).reshape(ts.num_trees, k, k)

def mean_centre(square_matrix):
    k,k = square_matrix.shape
    Tmat = np.identity(k) - [[1/k for _ in range(k)] for _ in range(k)] 
    return np.matmul(Tmat, np.matmul(square_matrix, np.transpose(Tmat)))

# now calculate the shared times and mean-centred shared times with this method
sts_gstat = shared_times(ts)
sts_gstat_mc = np.array(list(map(mean_centre, sts_gstat)))

# and finally compare the two approaches
np.all(abs(sts_gstat_mc - sts_grel_mc) < 1e-8)
np.all(abs(sts_grel * 2 - sts_gstat) < 1e-8) #NOTE THE FACTOR OF 2
```
It turns out that to get the mean-centred shared times from ```genetic_relatedness``` we need ```polarised=False``` but to get the non-mean-centred shared times from ```genetic_relatedness``` we need ```polarised=True``` and to multiply the result by 2. 

The latter bit makes sense to me: we want to simply multiply a branch length by 1 if both lineages share it (eqn 8 in Ralph et al 2020 Genetics), and since here f(x,y) = x * y / 2, we also need to multiply by 2. I'm a bit surprised by the former bit (that we need ```polarised=False``` to get the mean-centred shared times), but glad everything works!! 

Do we need more tests? Or should I go ahead with the workflow and 1) add to docstring/docs, 2) add to the change log, and 3)  send this to review?

petrelharp:
> And yeah, a time-window sounds very convenient for what I'd like to do (and then I should be able to just work with the centred version).

Hm - I'm confused on two things: how will you do this *without* a time window? and, I don't think that the centred version will work for this, since you want the centring to be done once, across all time windows, rather than separately by time window? (I *think* that's how the centering would work with time windows, since the summary stat would be applied separately to each chunk of time x genome.)

mmosmond:
>  how will you do this without a time window?

I'd calculate the non-centred shared times with ```genetic_relatedness```, write a little script to convert the output to the shared times (in each tree) in each epoch, multiply each epoch by a factor, add them all back together, and mean centre.

>  I don't think that the centred version will work for this, since you want the centring to be done once, across all time windows, rather than separately by time window?

Right, good point. But in my experience the splitting into epochs and then mean centring isn't a big deal, so what we've done here already seems to do the bulk of the work.

petrelharp:
> It turns out that to get the mean-centred shared times from genetic_relatedness we need `polarised=False` but to get the non-mean-centred shared times from genetic_relatedness we need `polarised=True` and to multiply the result by 2.

The first thing is good, since that's how we designed the statistic (since this is something estimable without knowing polarization, it should be computed with `polarised=False`; and this is the default, since polarization is not observable from the data).

As for requring polarisation to get the shared times: this should make sense once we understand what removing the mean centring is doing. Let's see, from the docs, the `site` version does:
```
Number of pairwise allelic matches in the window between two sample sets relative to the rest of the sample sets. To be precise, let m(u,v) denote the total number of alleles shared between nodes u and v, and let m(I,J) be the sum of m(u,v) over all nodes u in sample set I and v in sample set J. Let S and T be independently chosen sample sets. Then, for sample sets I and J, this computes E[m(I,J) - m(I,S) - m(J,T) + m(S,T)]. 
```
Ok, and this is just removing the "relative" and we get `E[m(I,J)]`.  With `polarised=False` this includes the ancestral allele as well.
Hm, and the `branch` version:
```
Total area of branches in the window ancestral to pairs of samples in two sample sets relative to the rest of the sample sets. To be precise, let B(u,v) denote the total area of all branches ancestral to nodes u and v, and let B(I,J) be the sum of B(u,v) over all nodes u in sample set I and v in sample set J. Let S and T be two independently chosen sample sets. Then for sample sets I and J, this computes E[B(I,J) - B(I,S) - B(J,T) + B(S,T)].
```
Hm, this is puzzling. Let's see - going back to the summary function, which without mean-centering is
```
 f(x1, x2) = x1 * x2 / 2
```
Ah-ha: so if each of these are single samples, then x1 and x2 are each 0/1, indicating whether the branch/allele is above the sample. With polarised=False, this is applied to each branch for *both* of the allele frequencies resulting from a mutation on that branch; i.e., `f(x1, x2) + f(1-x1, 1-x2)` (since I'm doing the example of single samples). So if a branch is ancestral to *neither* sample, then it counts also. This makes good sense, because as we noted above, a mutation that is inherited by neither sample contributes to the shared number of alleles, since both get the ancestral allele at that site. 

Somehow this does the right thing for the centred version. I forget why the factor of 2 is there, but I'm sure it's supposed to be there?

So - I think that everything is as expected, although confusing - do you agree? Might you be able to draft something for the docs explaining this (eg that you need `polarised=True` to get non-mean-centred shared branches, and maybe why)?

petrelharp:
> I'd calculate the non-centred shared times with genetic_relatedness, write a little script to convert the output to the shared times (in each tree) in each epoch, multiply each epoch by a factor, add them all back together, and mean centre.

How's this work? For each tree you get a single number; how's it get split into epochs? (sorry, I'm being slow!)

petrelharp:
Also our addition seems to have broken some other things - see the failing tests (and let me know if you want me to investigate).

mmosmond:
Thanks @petrelharp!

> So if a branch is ancestral to neither sample, then it counts also.

Ah, now I get it, thanks.

> Somehow this does the right thing for the centred version.

Yeah, and I just realized that an alternative way to get the centred shared times is to use ```polarised=True``` and multiply by 2 (as we do for the non-centred version). This is the most intuitive way, for me, to think about these branch statistics, as it means you just add up the lengths of shared branches (in reality you add half the length of each shared branch, but then multiply by everything by 2).

>  I forget why the factor of 2 is there, but I'm sure it's supposed to be there?

This makes more sense for me when ```mode='site'``` and ```polarised=False``` , because then if two sample nodes share an allele at a site they get a 1/2 for sharing the allele and 1/2 for sharing the absence of the other allele, giving 1. But it is harder for me to justify the 1/2 when used as a branch statistic. More thought needed...

> So - I think that everything is as expected, although confusing - do you agree? Might you be able to draft something for the docs explaining this (eg that you need polarised=True to get non-mean-centred shared branches, and maybe why)?

Yep and yep.

> How's this work? For each tree you get a single number; how's it get split into epochs? (sorry, I'm being slow!)

With ```n_e``` epochs and ```k``` samples, for each tree you get ```n_e``` ```(k x k)``` matrices. I've been doing this like (I'm sure this can be improved!):

```python
def epochs(shared_times, split_times):

    T = shared_times[0,0] #TMRCA in this tree (any diagonal entry)
    chopped_split_times = [t for t in split_times if t < T] #chop off epochs at TMRCA
    tsplits = [0] + chopped_split_times + [T] #append 0 to front of split times, append TMRCA to end
    Ts = [tsplits[i + 1] - tsplits[i] for i in range(len(tsplits)-1)] #amount of time in each epoch
    
    stimes = [] #place to store shared time matrices for all epochs

    # 0s for epochs beyond TMRCA
    n = len(split_times) - len(chopped_split_times)
    if n  > 0:
        k = len(shared_times)
        for _ in range(n):
            stimes.append(np.zeros((k,k)))
    
    # epochs within TMRCA
    for i in range(len(tsplits)-1):
        # shared times in remaining epochs (subtract elapsed time, set to zero if negative, find the min vs time in epoch)
        stimes.append(np.minimum(np.maximum(shared_times - (tsplits[-1] - tsplits[-(i+1)]), 0), Ts[-(i+1)])) 
    
    return np.array(stimes) #most distant epoch listed first

# little example using sts_gstat from thread above
split_times = [2, 8, 10] #this will define 4 epochs based on these three split times
sts_epochs = []
# apply to each tree
for i,sts in enumerate(sts_gstat):
    sts_epochs.append(epochs(sts, split_times))
```

> The current tests are great! But, I think we still need tests of centreing for site and node statistics.

OK, I'll work on this.

> Also our addition seems to have broken some other things - see the failing tests (and let me know if you want me to investigate).

And I'll try to fix these too.

Thanks again.

petrelharp:
> Yeah, and I just realized that an alternative way to get the centred shared times is to use polarised=True and multiply by 2 (as we do for the non-centred version). This is the most intuitive way, for me, to think about these branch statistics, as it means you just add up the lengths of shared branches (in reality you add half the length of each shared branch, but then multiply by everything by 2).

Hm, I'm not sure this works out - each branch only counts in one direction, since for the case of single nodes, if x1 * x2 is nonzero then (1 - x1) * (1 - x2) is zero. More generally, the factor of 2 doesn't come from adding both alleles and dividing by 2, since it applies to multiallelic sites, if I recall correctly.

petrelharp:
We have the proof written up in an overleaf, I'll send it to you.

petrelharp:
And, oh! I see about the epochs. It only works if you're working with only single samples on a single tree and all samples are at the same time, right?

mmosmond:
> Hm, I'm not sure this works out - each branch only counts in one direction, since for the case of single nodes, if x1 * x2 is nonzero then (1 - x1) * (1 - x2) is zero. More generally, the factor of 2 doesn't come from adding both alleles and dividing by 2, since it applies to multiallelic sites, if I recall correctly.

Oh, I guess my reasoning is wrong here then (and I'd be curious to see the proof you mention). But the approach does work: 
```python
sts_grel_mc_pol = ts.genetic_relatedness(sample_sets, indexes, mode='branch', windows='trees', span_normalise=True, polarised=True, proportion=False, centred=True).reshape(ts.num_trees, n, n)
np.all(abs(sts_grel_mc_pol * 2 - sts_grel_mc) < 1e-8)
```

mmosmond:
> And, oh! I see about the epochs. It only works if you're working with only single samples on a single tree and all samples are at the same time, right?

Yeah, that's the only case I've thought about.

mmosmond:
Yeah, a hand would help. I took a stab at fixing the failing tests but can't quite piece it all together. It looks like the problem is passing the extra ```centre=False``` keyword argument to the lower level diversity stat, but I don't get exactly where/how that is happening in the tests, and so I'm hesitant to try any fixes.

petrelharp:
I'll have a look. Separately, I see that we've called the option `centred`, while in the test code it's `centre`. The latter matches other options better, being a verb rather than an adjective (like `span_normalise`). Shall we change it to `centre`?

petrelharp:
Ah-ha: the problem is that [Fst calls `__k_way_sample_set_stat`](https://github.com/tskit-dev/tskit/blob/main/python/tskit/trees.py#L6760) with an anonymous python function that itself [calls the lower-level `diversity` method](https://github.com/tskit-dev/tskit/blob/main/python/tskit/trees.py#L6732); we've made it so that [`centred` is always an argument to the lower-level k-way method](https://github.com/tskit-dev/tskit/pull/1623/files#diff-76eaf4cb701579960d1e1f888222aa140f64f1b4622c28d6907d9b78c6c63871R8135), but not to [the one-way method](https://github.com/tskit-dev/tskit/blob/main/python/_tskitmodule.c#L7942). One fix would be to add `centred` to `TreeSequence_one_way_stat_method` as we have for `TreeSequence_k_way_stat_method`, although it's never used, or to do something in the anonymous python Fst function - but a much better solution would be to figure out a way to add an argument to just one method in the C code without adding it to all the other stats methods (without a lot of boilerplate).

Any suggestions here, @benjeffery, @jeromekelleher?

petrelharp:
I've done (two versions of) a way to fix Fst specific to the Fst method. That's enough for you to proceed here anyhow, @mmosmond .

mmosmond:
> Shall we change it to centre?
 
I was thinking the same and just committed the changes to ```_tskitmodule.c``` and ```trees.py``` (the ```trees.c/h``` files still refer to a ```noncentred``` option/summary function, but in that context it seems appropriate to leave as is

jeromekelleher:
I'm not following the issue here @petrelharp, but maybe the simplest thing is to hack what needs to be hacked here and to create an issue to track the necessary "unhacking"?

petrelharp:
Ok, @mmosmond - the conclusion in #1648 is that you can leave things the way they are here, and we'll follow up with the unhacking separately.

petrelharp:
Is that enough to get you going again?

petrelharp:
Need any assistance to get this wrapped up, @mmosmond?

mmosmond:
Hey @petrelharp, sorry, having a hard time finding the time right now (and it takes me a while to navigate the code). If you can find the time go for it! Sorry to drag this out...

petrelharp:
All good! I'll see if I can get to it...

petrelharp:
I could do something on it, if it'd be useful to you - shall we get it in there, @mmosmond?

mmosmond:
I could try and help next week some, maybe with some guidance on what exactly needs to be done? (I attempted to continue this a month ago but wasn't sure what files needed editing to get the checks to work out.) 

I would like to use the non-centered option in some updates I have planned for ```sparg```, but I likely won't get around to those until the new year, so don't rush this on account of me.

jeromekelleher:
Are we blocking on ##1648 ? I can pick that up ASAP if so

petrelharp:
No, that's not blocking - I've got a temporary way to do this; the idea was to shift things over to #1648 afterwards. (however, we might want to do that before C 1.0!).

All we need to do is get the tests working. I'll try to get it a good ways along next week.

mmosmond:
Hey @petrelharp, I took another look at this to see if I could push it along. I think I'm just not understanding how the tests work (may have something to do with my almost nonexistant C ability). Codecov says line 2821 in trees.c is not being tested, but I've put in a test on line 1356 of test_stats.c (following the centred example). (It looks like the rest of the missing tests would follow if line 2821 was tested.) So the test isn't running? 

I've only now found the tree sequences in testlib.c, but I think the resulting stat doesn't matter here, we just need the function to not run an error (ret=0), no?

benjeffery:
> Codecov says line 2821 in trees.c is not being tested, but I've put in a test on line 1356 of test_stats.c (following the centred example). (It looks like the rest of the missing tests would follow if line 2821 was tested.) So the test isn't running?

I can confirm that line 1356 of test_stats.c is being run and that 2821 in trees.c is not. You need an `||` not an `&&` on 1357 of test_stats.c

benjeffery:
@mmosmond

> Codecov says line 2821 in trees.c is not being tested, but I've put in a test on line 1356 of test_stats.c (following the centred example). (It looks like the rest of the missing tests would follow if line 2821 was tested.) So the test isn't running?

I can confirm that line 1356 of test_stats.c is being run and that 2821 in trees.c is not. You need an `|` not an `&` on 1357 of test_stats.c

mmosmond:
Thanks @benjeffery! Glad to see I had it mostly figured out, just have to watch out when negating an and/or!

mmosmond:
ah, sorry, not yet. i'll see if i can finish it up before the new year. think it's mostly just documentation left...

petrelharp:
Hi, @mmosmond - this has come up as something we need for another project, so I'm going to make this happen, very soon.

mmosmond:
amazing! sorry to have left this unfinished for so long. i got a bit lost in all the code and let it fall down my todo list.

petrelharp:
Setting up the tests here I got very confused for a bit, because I wasn't properly considering `polarised=False` (the default), of course. Here's the deal: as currently defined,
- if `polarised=True, centred=False`, then the relatedness between `i` and `j` is the total branch area ancestral to both `i` and `j`, divided by two.
- if `polarised=False, centred=False`, then it is the total area that is ancestral to either both *or neither* `i` and `j` (since with unpolarised data, we can't tell the difference between those), divided by two.

This is why we write the "naive definition" tests to compare to, of course.

However, this does bring up another issue: the annoying factor of two that appeared above. As currently implemented, with `mode="branch"` or `mode="site"` and biallelic sites only,
```
ts.genetic_relatedness(...) == 2 * ts.genetic_relatedness(..., polarised=True)
```
(because default is `centred=True, polarised=False`).

I'm working on sorting things out - I *think* that everything is done right, there's just counterintuitive aspects of this, but will verify.

python/tests/test_tree_stats.py
```diff
@@ -67,22 +67,22 @@ def subset_combos(*args, p=0.5, min_tests=3):
     # of them, using this function, below. If we don't set a seed, a different
     # random set is run each time. Ensures that at least min_tests are run.
     # Uncomment this line to run all tests (takes about an hour):
-    # p = 1.0
+    p = 1.0
```
petrelharp, line 70: TODO: revisit

petrelharp:
Okay, I think this is ready!!

It's a big one because there were a bunch of intertangled issues. Here's the outline:
1. Switched the definition of genetic_relatedness to be the *average* over samples in samples sets, rather than the sum (see discussion in #2888).
2. Changed the default of `genetic_relatedness` to `polarised=True`.
3. Also switched the implementation of centring in genetic_relatedness to be the average over sample sets, rather than over all samples across those sample sets (the implementation differed from the documentation; see discussion in #2888). 
4. Added the `centre=False` argument to the testing framework and to `genetic_relatedness` and `genetic_relatedness_weighted`.
5. Better and more complete testing for various genetic relatedness things.
6. Along the way, found and fixed #2983 (which was exposed since the uncentred stat is not strict) and #2984 (which was a corner case we missed).
7. Added docs to explain a few things I found surprising along the way (and had got me before enough times I thought I should write them down) - see `docs/data_model.md` on non-ancestral material, and `docs/stats.md` on polarisation and strictness.
8. Documented `genetic_relatedness_matrix`.
9. Did some hacking around to get `centre` in as an additional option passed all the way from python down into the C library.

petrelharp:
@mmosmond if you'd like to have a look - particularly at the documentation? - now's the time, and it would be very useful!

python/tests/test_tree_stats.py
petrelharp: This appears to not be having the stated effect (of making tests take an hour); but I'd like to not address this here. I can confirm that tests are actually running.

c/tskit/trees.c
```diff
@@ -1297,6 +1297,8 @@ tsk_treeseq_branch_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
     double *state = tsk_calloc(num_nodes * state_dim, sizeof(*state));
     double *summary = tsk_calloc(num_nodes * result_dim, sizeof(*summary));
     double *running_sum = tsk_calloc(result_dim, sizeof(*running_sum));
+    double *zero_state = tsk_calloc(state_dim, sizeof(*zero_state));
+    double *zero_summary = tsk_calloc(result_dim, sizeof(*zero_state));
```
jeromekelleher, line 1301: missing malloc checks

c/tskit/trees.c
```diff
@@ -3928,17 +3985,26 @@ tsk_treeseq_genetic_relatedness_weighted(const tsk_treeseq_t *self,
...
+            num_windows, windows, options, result);
+        if (ret != 0) {
+            goto out;
+        }
+    } else {
+        ret = tsk_treeseq_general_stat(self, num_weights + 1, new_weights,
+            num_index_tuples, genetic_relatedness_weighted_noncentred_summary_func,
+            &args, num_windows, windows, options, result);
+        if (ret != 0) {
+            goto out;
```
jeromekelleher, line 4006: Should probably try and hit this with tests?

c/tskit/trees.h
```diff
@@ -54,6 +54,9 @@ extern "C" {
 #define TSK_STAT_ALLOW_TIME_UNCALIBRATED (1 << 12)
 #define TSK_STAT_PAIR_NORMALISE          (1 << 13)
 
+/* Stat-specific options */
+#define TSK_STAT_NONCENTRED         (1 << 12)
```
jeromekelleher, line 58: Why not 1 << 14? We could pass both TSK_STAT_NONCENTRED  and TSK_STAT_ALLOW_TIME_UNCALIBRATED  to the stats API, couldn't we?

python/_tskitmodule.c
```diff
@@ -9594,6 +9596,10 @@ TreeSequence_k_way_stat_method(TreeSequence *self, PyObject *args, PyObject *kwd
     if (polarised) {
         options |= TSK_STAT_POLARISED;
     }
+    if (!centre) {
+        // only currently used by genetic_relatedness
+        options |= TSK_STAT_NONCENTRED;
```
jeromekelleher, line 9601: Needs  to be covered in low-level tests

python/_tskitmodule.c
```diff
@@ -9678,6 +9685,10 @@ TreeSequence_k_way_weighted_stat_method(TreeSequence *self, PyObject *args,
     if (polarised) {
         options |= TSK_STAT_POLARISED;
     }
+    if (!centre) {
+        // only currently used by genetic_relatedness_weighted
+        options |= TSK_STAT_NONCENTRED;
```
jeromekelleher, line 9690: Likewise here

mmosmond:
will take a look asap but likely not until tomorrow

python/tskit/trees.py
```diff
@@ -8777,12 +8871,20 @@ def Fst(
         # around with indexes and samples sets twice.
 
         def fst_func(sample_set_sizes, flattened, indexes, **kwargs):
```
jeromekelleher, line 8873: Would it be simpler to just duplicate the signature here rather than use the kwargs shortcut? Then you shouldn't need to do any messing around with inspect

jeromekelleher (commented):
Spotted a few minor things, otherwise LGTM

c/tskit/trees.h
petrelharp: yes, not sure what I was thinking here

python/tskit/trees.py
petrelharp: lol, good idea!

petrelharp:
Whoops - I guess I forgot to check codecov for `_tskitmodule.c`? I'm a bit puzzled why it wasn't more obvious when I looked at the codecov report - I thought I'd looked through everything! I think I've got all those bits; we'll see if codecov agrees.

petrelharp:
Okay - looks good to go pending @mmosmond's review of the docs and concepts!

docs/stats.md
```diff
@@ -556,6 +556,25 @@ associated with each allele; but if polarised, then the ancestral allele is left
 For branch or node statistics, summary functions are applied to the total weight or number of samples
 below, and above each branch or node; if polarised, then only the weight below is used.
 
+(sec_stats_strictness)=
+
+### Strictness, and which branches count?
+
+Most statistics are not be affected by invariant sites,
```
brieuclehmann, line 563: 'are not affected'

docs/stats.md
```diff
@@ -588,18 +607,30 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
-`genetic_relatedness`
-: {math}`f(x_i, x_j) = \frac{1}{2}(x_i - m)(x_j - m)`,
+`genetic_relatedness, centre=True`
+: {math}`f(x_i, x_j) = (x_i / n_i - m)(x_j / n_j - m)`,
 
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
-  of samples.
+  of sample sets.
+  For an polarised statistic (the default) with biallelic loci, this calculates
+  {math}`(p_1 - \bar{p}) (p_2 - \bar{p})`.
```
brieuclehmann, line 616: Worth clarifying what is meant by $\bar{p}$ precisely

docs/stats.md
```diff
@@ -588,18 +607,30 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
 
-`genetic_relatedness`
-: {math}`f(x_i, x_j) = \frac{1}{2}(x_i - m)(x_j - m)`,
+`genetic_relatedness, centre=True`
+: {math}`f(x_i, x_j) = (x_i / n_i - m)(x_j / n_j - m)`,
 
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
-  of samples.
+  of sample sets.
+  For an polarised statistic (the default) with biallelic loci, this calculates
```
brieuclehmann, line 615: 'a polarised'

docs/stats.md
```diff
@@ -588,18 +607,30 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
   For an unpolarized statistic with biallelic loci, this calculates
```
brieuclehmann, line 607: 'unpolarised'

docs/stats.md
petrelharp:
```suggestion
Most statistics are not affected by invariant sites,
```
lol good call

docs/stats.md
petrelharp, line 607:
```suggestion
  For an unpolarised statistic with biallelic loci, this calculates
```
fine I guess ;)

docs/stats.md
petrelharp:
```suggestion
  For a polarised statistic (the default) with biallelic loci, this calculates
```

docs/stats.md
petrelharp, line 616:
```suggestion
  {math}`(p_1 - \bar{p}) (p_2 - \bar{p})`, where {math}`\bar{p}` is the average
  derived allele frequency across sample sets.
```

jeromekelleher (approved):

jeromekelleher:
Shall we merge?

petrelharp:
Yes! @mmosmond still happy for any suggestions / changes - we can put them in separately.

mmosmond:
Wow, amazing to see how you guys work! No wonder this software can do all the things it can do.

This looks good to me. I can get exactly what I want (shared branch length in each tree for all pairs of samples) with
```python
import tskit, msprime
import numpy as np
ts = msprime.sim_ancestry(samples=3, sequence_length=1e3, recombination_rate=1e-3, random_seed=1)
n = ts.num_samples
sample_sets = [[i] for i in ts.samples()] #each sample in its own set
indexes = [(i,j) for i in ts.samples() for j in ts.samples()] #every pairwise comparison
ts.genetic_relatedness(sample_sets, indexes, 
                                  mode='branch', windows='trees', 
                                  span_normalise=True, polarised=True, 
                                  proportion=False, centre=False).reshape(ts.num_trees, n, n) 
```
No more pesky factor of 2!

petrelharp:
Yep, that's right! And you probably already noticed this, but if you don't want the actual matrix but just want to do operations with it, you can do vector-matrix-vector products with `genetic_relatedness_weights`, and matrix-vector products when #2980 goes in.

petrelharp:
Your attention to detail on the documentation would be particularly helpful, though, @mmosmond - so if you haven't read it carefully then that would be helpful.

mmosmond:
Yeah I took a look yesterday, and again now, and (with a focus on branch statistics) all looks ok to me. The only thing I wonder about is -- as a relative novice with a very specific aim -- if it should be said explicitly under "branch" that you can get the shared branch lengths between two samples in each tree with `mode='branch', windows='trees', proportion=False, centre=False`. But that is probably putting too much emphasis on an overly specific case. The upside would be to provide a concrete example for the branch statistic, like $K_{c0}$ does for the site statistic.

petrelharp:
Perhaps a short example in the tutorials would be nice for that?

PR #1626, opened 2021-08-14
Consistently use space to separate SVG transforms

### LINKED ISSUE #1625: Drawing: bug in x-axis labels rsvg and cairosvg.
Using SVG 2 operations severely limits the options that we have when exporting the SVGs. We basically have to use Chromium for things to render correctly, but that makes controlling things like fonts very hard. It's also a massive dependency and limits portability.

How much SVG 2 are we using @hyanwong? I think unless there's a major loss in functionality, it would be better to do things in a slightly uglier way, using the older standard.

Using svg 1.1, we should be able to use cairosvg to output to PDF, which gives a really nice way to get to publication quality graphics directly within a script.

hyanwong:
May fix #1625

### DIFF
--- a/python/tests/data/svg/internal_sample_ts.svg
+++ b/python/tests/data/svg/internal_sample_ts.svg
@@ -13,44 +13,44 @@
 		</g>
 		<g class="axes">
 			<g class="x-axis">
-				<g transform="translate(500,200)">
+				<g transform="translate(500 200)">
 					<text class="lab" text-anchor="middle" transform="translate(0 -11)">Genome position</text>
 				</g>
 				<line x1="20" x2="980" y1="163.2" y2="163.2"/>
 				<g class="ticks">
 					<g class="tick" transform="translate(20 163.2)">
 						<line x1="0" x2="0" y1="0" y2="5"/>
-						<g transform="translate(0,6)">
+						<g transform="translate(0 6)">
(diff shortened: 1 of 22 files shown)


jeromekelleher:
LGTM. I wonder if it's worth adding some tests based on [svglib](https://pypi.org/project/svglib/), just to see if it throws any warnings etc when run through the examples?

hyanwong:
I suspect that it might not warn in that particular case, but I guess there's no harm in adding extra stuff. OTOH, because we are using svg library, chances are we shouldn't have many badly formatted SVGs. Either way, probably best to merge this first and add extra tests in a future PR?

hyanwong:
I just tried `svglib`. It errors out on the (correct) style `.background path:nth-child(odd) {}` but doesn't detect a problem with the comma in `<g transform="translate(0,6)">` (and actually, I suspect that the comma is allowed, and it's just that Cairo is being too picky/buggy here - at least, they use a comma in the `translate` specification in the [MDN svg dev docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Basic_Transformations), and hunting around the internet suggests legacy SVG should allow either commas or spaces).

jeromekelleher (approved):

jeromekelleher:
OK, thanks. LGTM then.

PR #1627, opened 2021-08-16
Added an extra min_length threshold check in find_ibd.

gtsambos:
As per a private comment to @jeromekelleher , this small alteration doesn't change any of the output (all the tests still pass), but it should greatly reduce the internal memory needs of `find_ibd` when a minimum length is specified. (Basically, this stops lots of tiny segments from being appended to the master lists of ancestral segments)

I think we originally were considering various more flexible definitions of IBD where this condition is not wanted, but we've now settled on a definition of IBD that requires the IBD segments to maintain distinct genealogical paths along their length, so this condition is no longer needed I think.

Unless there are any objections, I'll change the relevant part of the C codebase too

@petrelharp @dvukcevic

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7415,17 +7415,19 @@ tsk_ibd_finder_enqueue_segment(
     void *p;
 
     tsk_bug_assert(left < right);
-    /* Make sure we always have room for one more segment in the queue so we
-     * can put a tail sentinel on it */
-    if (self->segment_queue_size == self->max_segment_queue_size - 1) {
-        self->max_segment_queue_size *= 2;
-        p = tsk_realloc(self->segment_queue,
-            self->max_segment_queue_size * sizeof(*self->segment_queue));
-        if (p == NULL) {
-            ret = TSK_ERR_NO_MEMORY;
-            goto out;
+    if (right - left > self->min_length) {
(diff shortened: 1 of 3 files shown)


jeromekelleher:
Sounds like a sensible option to me @gtsambos, let's do it.

gtsambos:
a (quite minor) issue: do you think the > should be >=? (Perhaps it doesn't matter as long as the documentation gets it right...)

jeromekelleher:
I think ``>`` is the most convenient, this let's use keep an easy default behaviour at zero?

gtsambos:
Sounds good to me.
Btw, I don't think the Codecov failing here is a big deal. The untested lines were part of the original codebase too, and they are basically testing for a memory overflow problem (so kindof hard to make a test for in the first place).

jeromekelleher:
Sure - we should add *some* tests though, just to make sure it's doing what it's supposed to. For example, making it longer than sequence length should result in zero IBD segs, right?

gtsambos:
Yup, I'll do that tomorrow. A test of the > versus >= behaviour might be good too.

jeromekelleher (approved):
LGTM - just needs a squash

gtsambos:
LMK if anything else should be done on this @jeromekelleher

PR #1652, opened 2021-09-01
Fix nbytes tests

### LINKED ISSUE #1638: Fix nbytes tests
The nbytes tests in test_tables.py were broken by #1618. Fix.

benjeffery:
Fixes #1638 

Note that this means that 'nbytes' will always count offsets as 64 bit.


jeromekelleher (approved):
LGTM

PR #1655, opened 2021-09-02
Avl tree

### LINKED ISSUE #1640: Dictionary data structures library
There's a number of places in which we're missing having efficient dictionary data structures in tskit (mainly hash tables and AVL trees). See #1639 as a current example.

I wanted to start this thread to explore some possibilities for libraries that we could use, before we start considering rolling our own.

Some requirements:

1. Has to be fully cross platform
2. Embeddable - we can include it in our build, an external library dependency is not an option
3. License compatible with our MIT licence
4. Multi-threaded - anything that uses global library-level memory would be treated with deepest suspicion.
5. "Good". Other people should be using it in production, preferably for a long time.
6. Full error checking compatible with our model 
6. Not have a terrible API 

Here's some options:

## [libdict](https://github.com/fmela/libdict)

Pros:
- Extensive suite of dictionary types
- Looks well written, clean OO type API
- 68 stars/42 forks on GitHub, so presumably reasonably well used

Cons:

- Not much documentation; [this](https://ftp.tw.freebsd.org/distfiles/libdict.html) is all I could find.
- No guidance on how to embed (but Makefile looks pretty simple, and someone has make a PR showing how to build with [meson](https://github.com/fmela/libdict/pull/37/files)
- No releases

## [klib](https://attractivechaos.github.io/klib/#About)

Pros
- Specifically designed for embedding
- Has lots of other things we might want at some point
- Extensively used (embedded in htslib, AFAIK) - 3.4K stars
- Pretty good documentation linked from the README
- Very efficient (probably - haven't looked at any actual benchmarks)

Cons

- Container types like hash tables use macros (efficient, but very hard to read)
- Coding style is extremely dense, very difficult to read IMO (especially matters when implemented via macros)

jeromekelleher:
Closes #1640

We don't need much from our AVL tree, so I've implemented Knuths version.

### DIFF
--- a/c/tests/test_core.c
+++ b/c/tests/test_core.c
@@ -371,6 +371,134 @@ test_malloc_overflow(void)
 #endif
 }
 
+static int
+validate_avl_node(tsk_avl_node_int_t *node)
+{
+    int height, lheight, rheight;
+
+    if (node == NULL) {
+        return 0;
+    }
+    lheight = validate_avl_node(node->llink);
+    rheight = validate_avl_node(node->rlink);
+    height = 1 + TSK_MAX(lheight, rheight);
+
+    if (lheight != 0 && rheight != 0) {
+        CU_ASSERT_FATAL(node->balance == rheight - lheight);
+    } else if (lheight == 0 && rheight == 0) {
+        CU_ASSERT_FATAL(height == 1);
+        CU_ASSERT_FATAL(node->balance == 0);
+    } else {
+        CU_ASSERT_FATAL(height == 2);
+        if (lheight == 0) {
+            CU_ASSERT_FATAL(node->balance == 1);
+        } else {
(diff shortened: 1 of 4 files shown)


jeromekelleher:
This is ready to go in I think. The implementation is bare-bones, but I think it does what we need for the IBD result work. It's not part of the public API so we can change things to make it more general later if we want (i.e., have more general keys).

It's a direct implementation of the version in TAOCP, so we can be confident it's correct  and will work as well as any other implementation. The tests are pretty basic, but they're validating the fundamental properties of the data structure and checking all code paths, so there's a high degree of confidence it's in good shape.

jeromekelleher:
@mergifyio rebase

benjeffery (approved):
LGTM!

PR #1637, opened 2021-08-19
IBD result struct

jeromekelleher:
This is stacked on #1618 so will be very noisy until that's merged

This is some initial work on refactoring the IBD code as we discussed @gtsambos. The idea is that we encapsulate all the IBD segment result information in this ``tsk_ibd_result_t`` struct, which we make available to the caller. All the details about how we store the segments for a pair of samples are hidden away, which gives us flexibility to do various things in the future.

Still some work to be done to really make use of this, though.

PR #1659, opened 2021-09-03
Ibd avl tree

jeromekelleher:
This converts the IBD result object to use an AVL tree internally for storage instead of a lookup matrix. We're not changing the semantics of the object yet so that we can store segments for arbitrary sample pairs, this is just the minimal changes required to change the backend. 

I'll follow up with more PRs for how we can change the semantics later, probably good to get this squared away first so that we can keep the diffs smaller.

Stacked on #1655 and #1637

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7118,105 +7118,105 @@ out:
  * IBD Result
  *************************/
 
+typedef struct {
+    tsk_segment_t *head;
+    tsk_segment_t *tail;
+} tsk_segment_list_t;
+
+/* This maps two positive integers 0 <= a < b < N into the set
+ * {0, ..., N^2}. For us to overflow an int64, N would need to
+ * be > sqrt(2^63), ~3 * 10^9. The maximum value for a 32bit int
+ * is ~2 * 10^9, so this can't happen here, however it is
+ * theoretically possible with 64 bit IDs. It would require
+ * a *very* large node table --- assuming 24 bytes per row
+ * it would be at least 67GiB. To make sure this eventuality
+ * doesn't happen, we have a tsk_bug_assert in the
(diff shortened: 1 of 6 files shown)


molpopgen:
This is awesome!

jeromekelleher:
This is ready to go - @gtsambos, basically this just swaps out the array-based indexing code you were using with its direct equivalent using the AVL trees. There's no semantic updates here, that's happening in #1660.

gtsambos (approved):
This is fantastic, Jerome!

c/tskit/tables.c
```diff
@@ -7118,105 +7118,95 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
+static int64_t
+pair_to_integer(tsk_id_t a, tsk_id_t b, tsk_size_t N)
+{
+    tsk_id_t tmp;
+    if (a > b) {
+        tmp = a;
+        a = b;
+        b = tmp;
+    }
+    return ((int64_t) a) * (int64_t) N + (int64_t) b;
```
benjeffery, line 7135: If I understand what is going on here correctly and the result needs to be unique for each pair of nodes then this won't work for 64bit IDs? Just checking I understand that this isn't future proof and it is worth documenting.

benjeffery (approved):
LGTM, one question about future proofing.

c/tskit/tables.c
jeromekelleher: Hah, good point. I've updated with a discussion and a bug_assert to make sure we can't overflow. If we ever need to deal with tables this big, then I'm sure we'll be hitting a bunch of other problems too :wink:

benjeffery: Great! I know it seems ridiculous as we're far away from node tables that big.

PR #1660, opened 2021-09-04
Ibd result all pairs

### LINKED ISSUE #1639: Remove mandatory sample_pairs argument in find_ibd
With the updates to find_ibd  in #1637 it's now possible to reason about the IBD segments without having to explicitly convert them all to Python objects, removing a significant memory block. However, we still have to provide the list of sample pairs explicitly as input, which basically rules out using find_ibd as a way to discover recent IBD segments in a large tree sequence. It should be possible to call ``ts.find_ibd(max_time=100)`` on a tree sequence with a large number of samples (10^6, say), and to get back a sensible result. At the moment this isn't possible.

Technically, what we want to achieve this is a sparse way of storing the IBD segments for a given pair of nodes ``(a, b)`` in the IBD result struct. A hash table is the obvious choice, but pulling in a dependency for a C hash table is tedious and would inevitably have some drawbacks. However, given the key-space is so well defined here, I think we can probably roll our own simple hash table (with linear probing) implementation easily enough. All we need is a hash function h, that approximately uniformly distributes a pair of integers (a, b) into the hash table range. So, concretely, suppose 0 <= a < b < 10**6 and H = 10000 (say), what's a good function to do this?

Any hash function aficionados out there?

jeromekelleher:
Change the IBD result object to allow us find IBD among all pairs.

Stacked on #1659, #1655, #1637

Closes #1639

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -4760,47 +4760,154 @@ test_link_ancestors_multiple_to_single_tree(void)
 }
 
 static void
-test_ibd_finder(void)
+verify_ibd_segment_list(tsk_segment_list_t *list, tsk_size_t num_nodes)
+{
+    tsk_segment_t *seg;
+    double total_span = 0;
+    tsk_size_t num_segments = 0;
+    /* double last_right = 0; */
+
+    for (seg = list->head; seg != NULL; seg = seg->next) {
+        CU_ASSERT_FATAL(seg->left < seg->right);
+        CU_ASSERT_FATAL(seg->node >= 0);
+        CU_ASSERT_FATAL(seg->node < (tsk_id_t) num_nodes);
+        total_span += seg->right - seg->left;
+        num_segments++;
+
+        /* TODO the segments are not necessarily in order - raise an issue
+         * to discuss this */
+        /* CU_ASSERT_FATAL(seg->left >= last_right); */
+        /* last_right = seg->right; */
+    }
+    CU_ASSERT_EQUAL_FATAL(total_span, list->total_span);
+    CU_ASSERT_EQUAL_FATAL(num_segments, list->num_segments);
+}
+
+static void
+verify_ibd_result(tsk_ibd_result_t *result)
 {
     int ret;
-    int k;
+    tsk_size_t j;
+    tsk_id_t a, b;
+    int64_t index;
+    tsk_size_t total_segments = 0;
+    tsk_size_t num_pairs = tsk_ibd_result_get_num_pairs(result);
+    tsk_id_t *pairs
+        = tsk_malloc(2 * tsk_ibd_result_get_num_pairs(result) * sizeof(*pairs));
+    tsk_id_t *pairs2
+        = tsk_malloc(2 * tsk_ibd_result_get_num_pairs(result) * sizeof(*pairs));
+    tsk_segment_list_t **lists
+        = tsk_malloc(tsk_ibd_result_get_num_pairs(result) * sizeof(*lists));
+    tsk_avl_node_int_t **avl_nodes
+        = tsk_malloc(result->pair_map.size * sizeof(*avl_nodes));
+
+    CU_ASSERT_FATAL(pairs != NULL);
+    CU_ASSERT_FATAL(pairs2 != NULL);
+    CU_ASSERT_FATAL(avl_nodes != NULL);
+    CU_ASSERT_FATAL(lists != NULL);
+    CU_ASSERT_EQUAL_FATAL(num_pairs, result->pair_map.size);
+    tsk_ibd_result_print_state(result, _devnull);
+
+    ret = tsk_ibd_result_get_keys(result, pairs);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_avl_tree_int_ordered_nodes(&result->pair_map, avl_nodes);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    for (j = 0; j < num_pairs; j++) {
+        a = pairs[2 * j];
+        b = pairs[2 * j + 1];
+        index = a * (int64_t) result->num_nodes + b;
+        CU_ASSERT(a < b);
+        CU_ASSERT_EQUAL(tsk_avl_tree_int_search(&result->pair_map, index), avl_nodes[j]);
+        index = b * (int64_t) result->num_nodes + a;
+        CU_ASSERT_EQUAL(tsk_avl_tree_int_search(&result->pair_map, index), NULL);
+    }
+
+    ret = tsk_ibd_result_get_items(result, pairs2, lists);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    for (j = 0; j < num_pairs; j++) {
+        CU_ASSERT_EQUAL_FATAL(pairs[2 * j], pairs2[2 * j]);
+        CU_ASSERT_EQUAL_FATAL(pairs[2 * j + 1], pairs2[2 * j + 1]);
+        verify_ibd_segment_list(lists[j], result->num_nodes);
+        total_segments += lists[j]->num_segments;
+    }
+    CU_ASSERT_EQUAL_FATAL(result->total_segments, total_segments);
+
+    free(pairs);
+    free(pairs2);
+    free(lists);
+    free(avl_nodes);
+}
+
+static void
+test_ibd_finder_single_tree(void)
+{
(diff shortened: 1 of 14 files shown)


jeromekelleher:
This is working pretty well:

```python
import resource
import tskit
import time

# ts = tskit.load("benchmark-n1M-m100M.trees")
ts = tskit.load("benchmark-n100k-m1M.trees")
print(ts)

before = time.perf_counter()
result = ts.find_ibd(max_time=100)

info = resource.getrusage(resource.RUSAGE_SELF)
# Result is in KiB
gb = info.ru_maxrss / (1024**2)

print(result)

print("time = ", duration)
print("mem  = ", gb)
```
Gives
```
╔═════════════════════════╗
║TreeSequence             ║
╠═══════════════╤═════════╣
║Trees          │     4518║
╟───────────────┼─────────╢
║Sequence Length│1000000.0║
╟───────────────┼─────────╢
║Sample Nodes   │   100000║
╟───────────────┼─────────╢
║Total Size     │ 14.0 MiB║
╚═══════════════╧═════════╝
╔═══════════╤══════╤═════════╤════════════╗
║Table      │Rows  │Size     │Has Metadata║
╠═══════════╪══════╪═════════╪════════════╣
║Edges      │217239│  6.6 MiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Individuals│     0│ 24 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Migrations │     0│  8 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Mutations  │  4943│178.6 KiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Nodes      │203674│  5.4 MiB│          No║
╟───────────┼──────┼─────────┼────────────╢
║Populations│     1│ 16 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Provenances│     1│597 Bytes│          No║
╟───────────┼──────┼─────────┼────────────╢
║Sites      │  4943│120.7 KiB│          No║
╚═══════════╧══════╧═════════╧════════════╝

IBD Result:
max_time       = 100
min_length     = 0
total segments = 86966219
num node pairs = 46006545

time =  26.705644345027395
mem  =  5.849822998046875
```

So, we 're finding the IBD segments for a tree sequence with 100K samples in 26 seconds and ~6G of RAM. There's a lot more work to be done, but I think we're definitely on the right track. There's 46 million pairs that are IBD by 100 generations ago.

jeromekelleher:
Here's the perf report:
```  
14.45%  python3  _tskit.cpython-38-x86_64-linux-gnu.so         [.] tsk_avl_tree_int_insert                                                                                                 ▒
   4.89%  python3  _tskit.cpython-38-x86_64-linux-gnu.so         [.] tsk_table_collection_find_ibd                                                                                           ▒
   1.21%  python3  _tskit.cpython-38-x86_64-linux-gnu.so         [.] tsk_blkalloc_get                                                                                                        ▒
   1.03%  python3  _tskit.cpython-38-x86_64-linux-gnu.so         [.] tsk_ibd_result_alloc_new_pair                                                                                           ▒
   0.71%  python3  [unknown]                                     [k] 0xffffffffa74727e7                                                                                                      ▒
   0.44%  python3  [unknown]                                     [k] 0xffffffffa7600ae7                                                                                                      ▒
   0.40%  python3  [unknown]                                     [k] 0xffffffffa6a308be                                                                                                      ▒
   0.24%  python3  [unknown]                                     [k] 0xffffffffa6c6d1ae                           
```
So - basically all the time is spent in the AVL tree (which is fine)

gtsambos:
Awesome. This looks like a huge step up. Pair-lookup was also taking most of the time beforehand, so it's not surprising that the AVL tree is where most of the time is spent here.

c/tests/test_tables.c
```diff
@@ -4812,38 +4942,47 @@ test_ibd_finder_multiple_trees(void)
...
-    tsk_id_t samples[] = { 0, 1, 0, 2 };
-    tsk_id_t *pair;
+    tsk_id_t samples[] = { 0, 1, 2 };
+    tsk_id_t pairs[][2] = { { 0, 1 }, { 0, 2 } };
+    tsk_size_t num_samples = 3;
     tsk_size_t num_pairs = 2;
     tsk_ibd_result_t result;
-    double true_left[2][2] = { { 0.0, 0.7 }, { 0.7, 0.0 } };
-    double true_right[2][2] = { { 0.7, 1.0 }, { 1.0, 0.7 } };
+    double true_left[2][2] = { { 0.0, 0.75 }, { 0.75, 0.0 } };
```
jeromekelleher, line 4950: In case anyone is wondering, this is for 32 bit tests. 0.7 isn't a round number, and it's troublesome for comparing equality on some hardware.

jeromekelleher:
This is ready for review now I think. I've filed a bunch of issues to follow up on, so there's still stuff to be cleaned up. It'd be good to get this much in though.

benjeffery (approved):
Been through, most closely checked the python C API code but all seems good.

gtsambos:
Just getting on to this now; hope to be done by tonight!

python/tskit/tables.py
```diff
@@ -2351,19 +2350,46 @@ class IbdResult(collections.abc.Mapping):
...
+
+    @property
+    def total_segments(self):
+        return self._ll_result.total_segments
+
+    @property
+    def pairs(self):
+        return self._ll_result.get_keys()
+
+    def __str__(self):
```
gtsambos, line 2366: Cool! This will be super useful.

c/tests/test_tables.c
```diff
@@ -4760,47 +4760,176 @@ test_link_ancestors_multiple_to_single_tree(void)
...
+    /* double last_right = 0; */
+
+    for (seg = list->head; seg != NULL; seg = seg->next) {
+        CU_ASSERT_FATAL(seg->left < seg->right);
+        CU_ASSERT_FATAL(seg->node >= 0);
+        CU_ASSERT_FATAL(seg->node < (tsk_id_t) num_nodes);
+        total_span += seg->right - seg->left;
+        num_segments++;
+
+        /* TODO the segments are not necessarily in order - issue #1682 */
```
gtsambos, line 4777: I think this is expected, though I know it's annoying. I'll comment more in the issue about it.

c/tests/test_tables.c
```diff
@@ -4760,47 +4760,176 @@ test_link_ancestors_multiple_to_single_tree(void)
...
+
+    ret = tsk_ibd_result_get_items(result, pairs2, lists);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    for (j = 0; j < num_pairs; j++) {
+        CU_ASSERT_EQUAL_FATAL(pairs[2 * j], pairs2[2 * j]);
+        CU_ASSERT_EQUAL_FATAL(pairs[2 * j + 1], pairs2[2 * j + 1]);
+        verify_ibd_segment_list(lists[j], result->num_nodes);
+        total_segments += lists[j]->num_segments;
+    }
+    CU_ASSERT_EQUAL_FATAL(result->total_segments, total_segments);
```
gtsambos, line 4833: I'm a bit unclear on what this is testing for. Is this some way of checking that there are only keys for pairs with non-empty sets of IBD segments or something?

c/tskit/tables.h
```diff
@@ -625,6 +625,13 @@ typedef struct _tsk_segment_t {
     tsk_id_t node;
 } tsk_segment_t;
 
+typedef struct {
+    tsk_size_t num_segments;
+    double total_span;
```
gtsambos, line 630: What's this for?

gtsambos (approved):
Nice one @jeromekelleher --you've been so busy 🤯 👍  only really minor questions on this. I didn't comment in great detail on the AVL tree stuff as I only have a general idea of what they are needed for  here.

c/tskit/tables.h
gtsambos: (the `total_span` attribute)

c/tests/test_tables.c
jeromekelleher: It's mainly check that the "keys" we return are correct. Just making sure that the two different ways we have of getting the same thing (they pairs) are the same.

c/tskit/tables.h
jeromekelleher: The sum of the ``right - left`` values in the segments. I figured this was a handy summary which would be useful when we optionally skip storing the segments.

PR #1690, opened 2021-09-10
add vectorised metadata, closes #1676

### LINKED ISSUE #1676: vectorised metadata extraction
A common thing people - at least with SLiM tree sequences - want to do is "get a vector of values from metadata where every object has the thing in metadata", e.g.
```
np.array([ind.metadata['slim_time'] for ind in ts.individuals()])
```
We do in pyslim, and @bodkan does this also (somewhat painfully?) in slendr using reticulate. The list comprehension is pretty easy to do, but to make it easier, perhaps there's an API we could make up? like
```
ts.metadata_vector('individuals', 'slim_time')
```
for instance?  Perhaps with an optional `default` argument that provides the value to be used when the key is not present in a metadata?

petrelharp:
Here is a start at adding `metadata_vector`. I haven't figured out the right way to get keys that are a few nested levels down, though - do you know an elegant way to do this, @benjeffery?

Fixes #1676

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -31,7 +31,9 @@ import itertools
 import json
 import numbers
 import warnings
+from collections.abc import Mapping
 from dataclasses import dataclass
+from functools import reduce
 from typing import Any
 from typing import Optional
 from typing import Union
@@ -46,6 +48,17 @@ import tskit.util as util
 from tskit import UNKNOWN_TIME
 
 dataclass_options = {"frozen": True}
+
+
+# Needed for cases where `None` can be an appropriate kwarg value,
+# we override the meta so that it looks good in the docs.
+class NotSetMeta(type):
+    def __repr__(cls):
+        return "Not set"
+
+
+class NOTSET(metaclass=NotSetMeta):
+    pass
 
 
 @metadata.lazy_decode
@@ -584,6 +597,53 @@ class MetadataMixin:
         self._metadata_schema_cache = metadata.parse_metadata_schema(
             self.ll_table.metadata_schema
         )
+
+    def metadata_vector(self, key, *, dtype=None, default_value=NOTSET):
+        """
+        Returns a numpy array of metadata values obtained by extracting ``key``
+        from each metadata entry, and using ``default_value`` if the key is
+        not present. ``key`` may be a list, in which case nested values are returned.
+        For instance, ``key = ["a", "x"]`` will return an array of
+        ``row.metadata["a"]["x"]`` values, iterated over rows in this table.
+
+        :param str key: The name, or a list of names, of metadata entries.
+        :param str dtype: The dtype of the result (can usually be omitted).
+        :param object default_value: The value to be inserted if the metadata key
+            is not present. Note that for numeric columns, a default value of None
+            will result in a non-numeric array. The default behaviour is to raise
+            ``KeyError`` on missing entries.
+        """
+
+        if default_value == NOTSET:
+
+            def getter(d, k):
+                return d[k]
+
+        else:
+
+            def getter(d, k):
+                return (
+                    d.get(k, default_value) if isinstance(d, Mapping) else default_value
+                )
+
+        if isinstance(key, list):
+            out = np.array(
+                [
+                    reduce(
+                        getter,
+                        key,
+                        row.metadata,
+                    )
+                    for row in self
+                ],
+                dtype=dtype,
+            )
+        else:
+            out = np.array(
+                [getter(row.metadata, key) for row in self],
+                dtype=dtype,
+            )
+        return out
 
 
 class IndividualTable(BaseTable, MetadataMixin):
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,7 +4,7 @@
 
 **Breaking changes**
 
-- The CLI `info` command now gives more detailed information on the tree sequence
+- The CLI ``info`` command now gives more detailed information on the tree sequence
   (:user:`benjeffery`, :pr:`1611`)
 
 - 64 bits are now used to store the sizes of ragged table columns such as metadata,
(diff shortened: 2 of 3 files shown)


benjeffery:
For getting the nested keys how about something like:
`reduce(lambda d, k: d.get(k, default) if isinstance(d, Mapping) else default, key, row.metadata)`

You'd need to check if key was a list first.

petrelharp:
Wow that is much more elegant than I would have done.

petrelharp:
Needs some more tests for weird metadata, but this seems to work.

petrelharp:
Ok, I think this is good to go?

python/tskit/tables.py
```diff
@@ -585,6 +586,42 @@ def _update_metadata_schema_cache_from_ll(self) -> None:
             self.ll_table.metadata_schema
         )
 
+    def metadata_vector(self, key, *, dtype=None, default_value=None):
+        """
+        Returns a numpy array of metadata values obtained by extracting ``key``
+        from each metadata entry, and inserting ``default_value`` if the key is
```
benjeffery, line 592: "inserting" suggests that we might be modifying the metadata
```suggestion
        from each metadata entry, and using ``default_value`` if the key is
```

python/tskit/tables.py
```diff
@@ -585,6 +586,42 @@ def _update_metadata_schema_cache_from_ll(self) -> None:
...
+        :param dtype dtype: The dtype of the result (can usually be omitted).
+        :param object default_value: The value to be inserted if the metadata key
+            is not present. Note that for numeric columns, a default value of None
+            will result in a non-numeric array.
+        """
+
+        if isinstance(key, list):
+            out = np.array(
+                [
+                    functools.reduce(
```
benjeffery, line 607: If you use `from functools import reduce` you save an attribute look up in the tight loop. Same with `Mapping` below.

python/tskit/tables.py
```diff
@@ -585,6 +586,42 @@ def _update_metadata_schema_cache_from_ll(self) -> None:
...
+                        lambda d, k: d.get(k, default_value)
```
benjeffery, line 608: We only want to use `get` if a default value is supplied so that `KeyError` is raised for a missing entry. As `None` could be a valid desired default, you'll need a singleton class `NO_DEFAULT` to check for and use as the default kwarg.

python/tskit/tables.py
```diff
@@ -585,6 +586,42 @@ def _update_metadata_schema_cache_from_ll(self) -> None:
             self.ll_table.metadata_schema
         )
 
+    def metadata_vector(self, key, *, dtype=None, default_value=None):
```
benjeffery, line 589: Just an idea but you could make `key` optional and return the whole metadata object if it is not specified?

python/tskit/tables.py
```diff
@@ -585,6 +586,42 @@ def _update_metadata_schema_cache_from_ll(self) -> None:
...
+        ``row.metadata["a"]["x"]`` values, iterated over rows in this table.
+
+        :param str key: The name, or a list of names, of metadata entries.
+        :param dtype dtype: The dtype of the result (can usually be omitted).
+        :param object default_value: The value to be inserted if the metadata key
+            is not present. Note that for numeric columns, a default value of None
+            will result in a non-numeric array.
+        """
+
+        if isinstance(key, list):
```
benjeffery, line 604:
```suggestion
        if isinstance(key, collections.abc.Sequence):
```

benjeffery (commented):
Looking good, I think we need to handle the default differently though.

python/tskit/tables.py
petrelharp: Hm - nice idea, but that might encourage bad habits? Currently, metadata has to be a dictionary (not eg a single number), so I'm not sure about this. We could always extend it later to cover this case without breaking this API.

python/tskit/tables.py
petrelharp: woah! ok!

python/tskit/tables.py
petrelharp: oh dear

petrelharp:
Thanks! I think I took care of those things. Too bad python doesn't have the [missing() function](https://stat.ethz.ch/R-manual/R-devel/library/base/html/missing.html).

molpopgen:
I haven't scanned the diffs, but curious if this works/is tested on both codec flavors? (JSON and struct)

petrelharp:
Whoops - I was too clever with the tests, and hadn't caught that it was doing totally the wrong thing. Since strings are a `Sequence`, when I asked for `key="abc"` it was treating that the same as `key=["a", "b", "c"]`.  Thanks for catching that, codecov!

petrelharp:
> I haven't scanned the diffs, but curious if this works/is tested on both codec flavors? (JSON and struct)

It certainly *should* work - this is only using things in the layer that the codec is invisible!

python/tskit/tables.py
jeromekelleher: I agree with Peter here

benjeffery:
@petrelharp I've added a commit here with an extra tests and a tweak to the default value so that it looks better in the docs. Let me know what you think.

python/tskit/tables.py
```diff
@@ -49,8 +49,16 @@
...
 dataclass_options = {"frozen": True}
 
-# we'll need this because python can't tell if kwargs are passed in not
-DEFAULT = object()
+
+# Needed for cases where `None` can be an appropriate kwarg value,
+# we override the meta so that it looks good in the docs.
+class NotSetMeta(type):
+    def __repr__(cls):
+        return "Not set"
```
petrelharp, line 57: ah-ha! nice magic!

petrelharp:
Rebased; should be ready to merge.

benjeffery (approved):

benjeffery:
Added changelog - merging.

PR #1696, opened 2021-09-13
Fix typo in trees.py

awohns:
Small typo in docstring of simplify() prevented inline code block from ending.


petrelharp (approved):
Ack, thanks for fixing that!

PR #1700, opened 2021-09-14
adding time slice to ts.samples()

### LINKED ISSUE #1692: time slice in samples
I always seem to want to get samples from a tree sequence within a time slice. Although this is not hard to do, I thought more people might be running into this situation and we could add a `start_time` and `end_time` argument to `TreeSequence.samples()`. This could be done just on the Python API end with a simple boolean mask applied on top of the array (the same way we do it for populations), see [here](https://tskit.dev/tskit/docs/stable/_modules/tskit/trees.html#TreeSequence.samples).

Let me know if you think this is worth implementing and I can work on a PR.

mufernando:
## Description

WIP

Fixes #1692 

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -346,6 +346,23 @@ def get_mrca(pi, x, y):
     return mrca
 
 
+def get_samples(ts, time=None, population=None):
+    samples = []
+    for node in ts.nodes():
+        keep = bool(node.is_sample())
+        if time is not None:
+            if isinstance(time, (int, float)):
+                keep &= np.isclose(node.time, time)
+            if isinstance(time, (tuple, list, np.ndarray)):
+                keep &= node.time >= time[0]
+                keep &= node.time < time[1]
+        if population is not None:
+            keep &= node.population == population
+        if keep:
+            samples.append(node.id)
+    return np.array(samples)
+
+
 class TestMRCACalculator:
     """
     Class to test the Schieber-Vishkin algorithm.
@@ -509,11 +526,14 @@ class TestNumpySamples:
     various methods.
     """
 
-    def get_tree_sequence(self, num_demes=4):
-        n = 40
+    def get_tree_sequence(self, num_demes=4, times=None, n=40):
+        if times is None:
+            times = [0]
         return msprime.simulate(
             samples=[
-                msprime.Sample(time=0, population=j % num_demes) for j in range(n)
+                msprime.Sample(time=t, population=j % num_demes)
+                for j in range(n)
+                for t in times
             ],
             population_configurations=[
                 msprime.PopulationConfiguration() for _ in range(num_demes)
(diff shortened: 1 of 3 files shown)


mufernando:
A few questions:

- Should both min and max be inclusive? 
- Does this approach look reasonable (or do we expect too much of a performance hit?)?
- Ideas for more tests?

jeromekelleher:
LGTM, thanks @mufernando!

- I'd vote for making max exclusive following our usual interval semantics so that for any sample with time t, min_time <= t < max_time.
- In terms of testing, I think checking explicitly with a few hand coded examples on a small tree sequence would be good. (as in, you explicitly assert for the known set of samples)

petrelharp:
> I'd vote for making max exclusive following our usual interval semantics so that for any sample with time t, min_time <= t < max_time.

I agree... except, how do we deal with the probably most common use for this, which is to get everyone alive at a given time (eg 0.0)? For that most people would probably try `ts.samples(max_time = 0.0)`, which would *not* work. 

One option would be to make the argument `time_interval` instead of two separate `min` and `max` times, and if it's just a single number instead of a tuple, we return all the samples whose times are equal to that number?  Or, add two arguments: `ts.samples(time_interval=None, time=None)`, where `time_interval` is a closed-on-the-left-open-on-the-right interval, and `time` is just a single time, and you can't do both of them?

I agree about the testing, and be sure to include some times like `1/3` where we might run into floating point error.

mufernando:
I agree with @petrelharp that one of the most used cases would be to look at a single time point (t=0), so I went ahead and added two new parameters `time` and `time_interval`. 

I never seem to know exactly how to test on a single known answer, so I did the next best thing (that I could think of) which was to re-implement the behavior independently using a simple for loop instead of numpy.

mufernando:
ah, also unsure whether theses tests should be where they are within `test_highlevel.py:TestNumpySamples`.

jeromekelleher:
Looks good @mufernando. I'm not sure there's much point in having ``time`` and ``time_interval`` though - how about we have a single parameter ``time``, and 

- If ``time`` is a numeric value, return all samples whose node time is approximately equal (as determined by numpy.isclose) to the specified value
- Otherwise, if ``time`` is a pair of values ``(min_time, max_time)`` and we return all samples whose node time t is in this interval such that ``min_time <= t < max_time``.

That seems straightforward enough?

mufernando:
> Looks good @mufernando. I'm not sure there's much point in having `time` and `time_interval` though - how about we have a single parameter `time`, and
> 
> * If `time` is a numeric value, return all samples whose node time is approximately equal (as determined by numpy.isclose) to the specified value
> * Otherwise, if `time` is a pair of values `(min_time, max_time)` and we return all samples whose node time t is in this interval such that `min_time <= t < max_time`.
> 
> That seems straightforward enough?

Done!

waiting for a review and then I can squash and rebase.

petrelharp:
LGTM! I added a few more tests, including one where we build a simple explicit tree sequence.

petrelharp (approved):

python/tskit/trees.py
```diff
@@ -4800,15 +4800,21 @@ def get_samples(self, population_id=None):
...
+        """
+        Returns an array of the sample node IDs in this tree sequence. If
+        `population` is specified, only return sample IDs from that population.
+        It is also possible to restrict samples by time using the parameter
+        `time`. If `time` is a numeric value, only return sample IDs whose node
+        time is approximately equal to the specified time. If `time` is a pair
+        of values of the form `(min_time, max_time)`, only return sample IDs
+        whose node time `t` is in this interval such that `min_time <= t < max_time`.
+
+        :param int population: The population of interest. If None, return all
```
jeromekelleher, line 4813: This isn't quite true - should be "If None, do not filter samples by population". Same for the ``time`` arg below.

python/tskit/trees.py
```diff
@@ -4820,10 +4826,25 @@ def samples(self, population=None, population_id=None):
...
             population = population_id
         samples = self._ll_tree_sequence.get_samples()
+        keep = np.full(shape=samples.shape, fill_value=True)
         if population is not None:
             sample_population = self.tables.nodes.population[samples]
-            samples = samples[sample_population == population]
-        return samples
+            keep = np.logical_and(keep, sample_population == population)
+        if time is not None:
+            if isinstance(time, (int, float)):
```
jeromekelleher, line 4834: ``if isinstance(time, numbers.Number)``

is more flexible here (current version will fail if you use a numpy 0D array as input)

python/tskit/trees.py
```diff
@@ -4820,10 +4826,25 @@ def samples(self, population=None, population_id=None):
...
+                sample_times = self.tables.nodes.time[samples]
+                keep = np.logical_and(keep, np.isclose(sample_times, time))
+            elif isinstance(time, (tuple, list)) and len(time) == 2:
```
jeromekelleher, line 4837: I think it'd be better to coerce the input to a numpy array, something like
```python
else:
     time = np.ndarray(time, dtype=float) # This should raise TypeErrors/ValueErrors as needed
     if time.shape != (2,):
         raise ValueError("time interval must be a pair of values (min_time, max_time)")
     # as before
```

We want to make sure that people can pass in numpy arrays as parameters, which the current approach doesn't allow.

An alternative is to do something like
```python
else:
     min_time, max_time = time 
     if not (isinstance(min_time, numbers.Number) and isinstance(max_time, numbers.Number)):
          raise TypeError(...)
```
which is perhaps clearer and more in line with future changes in Python via [PEP 622](https://www.python.org/dev/peps/pep-0622/)

Any thoughts here anyone? @benjeffery?

jeromekelleher (commented):
LGTM, some minor changes in how we handle the type polymorphism suggested.

We should add some simple tests to make sure that numpy arrays are supported as input. I.e, something like
```python
x = np.array([1, 2])
ts.samples(time=x[0]) # time == 1
ts.samples(time=x) # time between 1 and 2
```

python/tskit/trees.py
mufernando: aha, thanks for letting me know!

mufernando: it doesn't look like this would work with a 0D array.

image

python/tskit/trees.py
mufernando: I agree it is better to coerce the input into a numpy array. So now I'm doing that from the get-go (note I'm using `np.array` and not `np.ndarray` because the latter is not used for array creation).

I opted to check for the shape explicitly as (1,) and (2,) so that we can give a more informative error in the case the input shape is wonky.

mufernando:
Updated to work with numpy arrays (including 0d and (1,) arrays). Added some tests with these types of array also.

python/tests/test_highlevel.py
```diff
@@ -541,6 +561,149 @@ def test_samples(self):
...
+        times = [0, 0.1, 1 / 3, 1 / 4, 5 / 7]
+        ts = self.get_tree_sequence(num_demes=2, n=20, times=times)
+        for time in times:
+            assert np.array_equal(get_samples(ts, time=time), ts.samples(time=time))
+            for population in (None, 0):
+                assert np.array_equal(
+                    get_samples(ts, time=time, population=population),
+                    ts.samples(time=time, population=population),
+                )
+
```
jeromekelleher, line 574: These are good candidates for parametrize:
```python
@pytest.mark.parametrize("time_interval", [
     [0, 0.1],
￼            (0, 1 / 3),
￼            np.array([1 / 4, 2 / 3]),
￼            (0.345, 5 / 7),
￼            (-1, 1),
￼        ])
def test_samples_time_interval(self, time_interval):
     # body of the existing loop
```
can do the same thing with ``test_samples_time`` above.

This approach is better because it tells you *which* of the cases failed when it fails.

jeromekelleher (approved):
LGTM, thanks @mufernando. One final comment about using parametrize in tests rather than loops to get better visibility on test examples.

python/tests/test_highlevel.py
mufernando: oh, cool! didn't know about that. will update.

mufernando:
I think we are ready to merge!

Thanks you all!

python/tests/test_highlevel.py
```diff
@@ -541,6 +561,150 @@ def test_samples(self):
...
+            [0, 0.1],
+            (0, 1 / 3),
+            np.array([1 / 4, 2 / 3]),
+            (0.345, 5 / 7),
+            (-1, 1),
+        ],
+    )
+    def test_samples_time_interval(self, time_interval):
+        rng = np.random.default_rng(seed=931)
+        times = rng.uniform(low=time_interval[0], high=time_interval[1], size=20)
```
petrelharp, line 586: hm, actually your test here is always returning all the samples, no? shouldn't maybe the times be uniform on `[0, 2*time_interval[1]]` or something?

python/tests/test_highlevel.py
```diff
@@ -541,6 +561,150 @@ def test_samples(self):
             ]
         assert total == ts.num_samples
 
+    @pytest.mark.parametrize("time", [0, 0.1, 1 / 3, 1 / 4, 5 / 7])
+    def test_samples_time(self, time):
+        ts = self.get_tree_sequence(num_demes=2, n=20, times=[time])
```
petrelharp, line 566: same thing here

petrelharp:
Sorry, one more thing! (see comments)

python/tests/test_highlevel.py
mufernando: oops!

mufernando:
Fixed it!

Issue #1691, opened 2021-09-10
Add a "fake root" to the tsk_tree struct


jeromekelleher:
When writing algorithms with the quintuply linked tree structure for trees that contain multiple roots, it's quite tedious that we need to do things like
```python
for root in tree.roots():
     do_algorithm(root)
```
It would be nicer if we could do something like 
```python
do_algorithm(tree.fake_root)
```
(I don't like the term ``fake_root`` but can't think of anything better).

The way this would be implemented in terms of the quintuply linked tree would be to add one more element to the pointer arrays, and to let this value (num_nodes) correspond to the ID of the fake root. Then, ``tree.left_child_array[-1]`` would correspond to the left root, etc.

This is a breaking change in terms of the low-level details of how we do tree traversals etc in C, so I think it would be good to get it done before 1.0.

molpopgen:
In one sense, this is adding a new head to the root list, so `tree.root_head` or `tree.head_root` may be a decent name?

benjeffery:
`virtual_root`?

jeromekelleher:
Nice, all good ideas. Lemme hack something up and we can see how it looks and take a vote.

jeromekelleher:
Aha, forgot how tricky this root tracking code is! I'm hoping this will make it less complicated, but I'll have to get my head back into it. Staring with [Algorithm T](https://github.com/tskit-dev/tskit/blob/9cd44736be9fbc0092ad26628531bccf69e1255c/python/tests/tsutil.py#L1139) is probably the way to go. The [RootTrackingTree](https://github.com/tskit-dev/tskit/blob/main/python/tests/tsutil.py#L1351) has a full implementation of the current root tracking code. It's always bothered me that there's more code for tracking the roots than there is for all the other cases (which are quite elegant I think).

One property we (almost) definitely want to preserve is that ``parent[u] == -1`` for all the real roots. It would break all sorts of everything if they pointed to the virtual root instead.

molpopgen:
Yeah, roots and the sample list updating stuff.  I've banged my head on the latter a few times, and either made it worse/slower or wrong.

molpopgen:
> One property we (almost) definitely want to preserve is that `parent[u] == -1` for all the real roots. It would break all sorts of everything if they pointed to the virtual root instead.

Yep, that would be a BIG break.

petrelharp:
This is a nice idea. It also sounds very tricky.

jeromekelleher:
[...]

PR #1704, opened 2021-09-15
Virtual root in trees

jeromekelleher:
See #1691

Still a WIP, but it definitely works so I think we should do it. The code is *so* much cleaner this way.


python/tests/tsutil.py
```diff
@@ -1175,6 +1175,143 @@ def algorithm_T(ts):
...
+            self.left_sib[c] = -1
+            self.right_sib[c] = -1
+        else:
+            self.right_sib[u] = c
+            self.left_sib[c] = u
+            self.right_sib[c] = -1
+        self.right_child[p] = c
+
+    def is_root(self, u):
+        return self.num_samples[u] >= self.root_threshold
```
petrelharp, line 1237: I don't understand what's going on here?

petrelharp: Oh, I see - you only use this method on path ends. So like this method is actually "would this be a root at this threshold if it didn't have a parent" not "is this a root"?

jeromekelleher: Yes, good call. This should really be ``is_potential_root``.

This is based on the current code by the way - the root threshold logic has been in there for a while. See the [C code](https://github.com/tskit-dev/tskit/blob/e1d64e30eeee291214020354faa787485efeabb4/c/tskit/trees.c#L3982) for example. This version is much simpler because we're not making special cases for the branch setting on roots.

benjeffery:
> Some benchmarks:

Oh wow. Awesome!

I'm in the weeds with vargen, will take a look over this later today.

jeromekelleher:
Bumping this one - @benjeffery any chance you could take a look through please? Good to get this diff pushed through.

PR #1707, opened 2021-09-16
Create Python IbdSegmentList and IbdSegment

### LINKED ISSUE #1681: Add SegmentList object in Python, and return this rather than dict encoding
Currently the IbdResult object returns a dictionary {"left": array, "right": array, "node": array} for each IBD pair to encode the segments. This isn't a great design, as it doesn't give any way of putting functionality on these segment lists.

- Add a SegmentList object in the Python C API which we return from the lowlevel _get function. This is a front-end for the C segment_list_t struct.  Keep a reference to the IbdResult object to ensure the memory accesses are safe.
- This can have accessors ``left``, ``right``, ``array`` which return the corresponding arrays
- Have some other accessors that give stats of the segment list.
- The high-level Python interface can also support iteration (I guess implementing the Sequence protocol, or something) which would return a list of (left, right, node) namedtuples.

So, you'd be able to do stuff like
```python
for key, value in ibd_result.items():
    print(key, value.left) # Array of left coordinates
    for seg in value:
       print(seg.left, seg.right, seg.node)
```
Document this in the high-level Python API as the result. (See #1657)

Do this before #1680

jeromekelleher:
This updates the types that we return from the IBD calculations to make them more flexible and renames ``find_ibd`` to ``ibd_segments``

### DIFF
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -393,7 +393,9 @@ class TestIbd:
         with pytest.raises(SystemError):
             result.print_state()
         with pytest.raises(SystemError):
-            result.total_segments
+            result.num_segments
+        with pytest.raises(SystemError):
+            result.total_span
         with pytest.raises(SystemError):
             result.get_keys()
 
@@ -485,7 +487,8 @@ class TestIbd:
     def test_direct_instantiation(self):
         # Nobody should do this, but just in case
         result = _tskit.IbdResult()
-        assert result.total_segments == 0
+        assert result.num_segments == 0
+        assert result.total_span == 0
         with tempfile.TemporaryFile("w+") as f:
             result.print_state(f)
             f.seek(0)
@@ -502,6 +505,20 @@ class TestIbdSegmentList:
         for attr in attrs:
             with pytest.raises(SystemError, match="not initialised"):
                 getattr(seglist, attr)
+
+    def test_memory_management(self):
+        ts = msprime.simulate(10, random_seed=1)
+        tc = ts.tables._ll_tables
+        result = tc.find_ibd()
+        del ts, tc
+        lst = result.get(0, 1)
+        assert lst.num_segments == 1
+        del result
+        assert lst.num_segments == 1
+        # Do some allocs to see if we're still working properly
+        x = sum(list(range(1000)))
+        assert x > 0
+        assert lst.num_segments == 1
 
 
 class TestTableMethods:
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -387,7 +387,7 @@ class TestTableCollection(LowLevelTestCase):
 
 class TestIbd:
     def test_uninitialised(self):
-        result = _tskit.IbdResult.__new__(_tskit.IbdResult)
+        result = _tskit.IbdSegments.__new__(_tskit.IbdSegments)
         with pytest.raises(SystemError):
             result.get(0, 1)
         with pytest.raises(SystemError):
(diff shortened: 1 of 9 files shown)


jeromekelleher:
This diff is probably big enough that we should merge it before going ahead with further changes. I'll open a few follow up issues (mostly around documentation). This is nearly the end for the big IBD changes and I think it'll settle down pretty quickly now.

jeromekelleher:
Nice to get this merged fairly soon - any chance you could take a look through please @gtsambos?

python/tests/test_lowlevel.py
```diff
@@ -387,28 +387,30 @@ def test_fromdict_bad_args(self):
 
 class TestIbd:
     def test_uninitialised(self):
-        result = _tskit.IbdResult.__new__(_tskit.IbdResult)
+        result = _tskit.IbdSegments.__new__(_tskit.IbdSegments)
         with pytest.raises(SystemError):
             result.get(0, 1)
         with pytest.raises(SystemError):
             result.print_state()
         with pytest.raises(SystemError):
```
gtsambos, line 395: Is there a reason why these new attributes are uninitialised (as opposed to 0)?

python/tests/test_lowlevel.py
```diff
@@ -489,6 +497,30 @@ def test_direct_instantiation(self):
...
+class TestIbdSegmentList:
+    def test_direct_instantiation(self):
+        # Nobody should do this, but just in case
+        seglist = _tskit.IbdSegmentList()
+        attrs = ["num_segments", "total_span", "left", "right", "node"]
+        for attr in attrs:
+            with pytest.raises(SystemError, match="not initialised"):
+                getattr(seglist, attr)
+
+    def test_memory_management(self):
```
gtsambos, line 509: I don't understand what this test is doing -- can you explain a bit?

gtsambos (approved):
Looks good Jerome! These new classes (esp. the IbdSegment one) look like they'll make this output a lot nicer to work with.

python/tests/test_lowlevel.py
jeromekelleher: This is reflecting a tricky detail about how we're managing memory. The low-level IbdSegments class is a thin wrapper around the ``tsk_ibd_segments_t`` class in C. We only ever create a new IbdSegments class as a result of calling ``tables.find_ibd()``, which ensures that the memory management is handled correctly. What this test is doing is ensuring that if someone deliberately sidesteps all this, and tries to create an IbdSegments instance manually they don't get a segfault (which they would, if we weren't carefully checking the state of things).

Basically this will never happen, it's just low-level stuff to make sure the memory management is all water tight.

python/tests/test_lowlevel.py
jeromekelleher: It's basically the same as the previous comment - we're checking some tricky details about how memory is managed, making sure we don't leak or segfault in odd situations. In this case, the ownership of the underlying memory is being managed by the object that we ``del``, and we're making sure that this object doesn't actually disappear until after all the things that reference it also disappear.

Some background on [reference counting](https://towardsdatascience.com/understanding-reference-counting-in-python-3894b71b5611) in Python might clear things up a bit, if you're interested.

jeromekelleher:
@benjeffery, any chance you could have a quick look over this please?

python/tests/test_lowlevel.py
```diff
@@ -489,6 +497,30 @@ def test_direct_instantiation(self):
...
+        ts = msprime.simulate(10, random_seed=1)
+        tc = ts.tables._ll_tables
+        result = tc.ibd_segments()
+        del ts, tc
+        lst = result.get(0, 1)
+        assert lst.num_segments == 1
+        del result
+        assert lst.num_segments == 1
```
benjeffery, line 517: Is it worth adding some `gc.collect()` after the `del` to be ultra paranoid here?

python/tskit/tables.py
```diff
@@ -2395,6 +2395,56 @@ def packset_timestamp(self, timestamps):
...
+    left: float
+    right: float
+    node: int
+
+    @property
+    def span(self):
+        return self.right - self.left
+
+
+class IbdSegmentList(collections.abc.Iterable, collections.abc.Sized):
```
benjeffery, line 2410: Nice.

benjeffery (approved):
LGTM!
One paranoia suggestion.

I am running the stress test on this branch, looks good so far but will leave running.

python/tests/test_lowlevel.py
jeromekelleher: Sure! Good call.

jeromekelleher:
Great, thanks @benjeffery. Updated with nice extra paranoia - please merge whenever you're happy.

PR #1709, opened 2021-09-16
Use doubling behaviour on ragged columns

### LINKED ISSUE #1703: Use doubling strategy for allocating memory in ragged columns?
In #1683 we updated the default strategy for allocating memory for the fixed size table columns to use doubling.

Should we do something similar for the ragged columns like ``metadata``? It's probably what an API user would expect I think. We should also document the ``tsk_individual_table_set_max_metadata_length_increment`` etc methods so that power users can keep control of memory, if they want to.

benjeffery:
Fixes #1703 

Initial try for feedback before tests and copypasteathon.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -35,7 +35,6 @@
 #define _TSK_WORKAROUND_FALSE_CLANG_WARNING
 #include <tskit/tables.h>
 
-#define DEFAULT_SIZE_INCREMENT 1024
 #define TABLE_SEP "-----------------------------------------\n"
 
 #define TSK_COL_OPTIONAL (1 << 0)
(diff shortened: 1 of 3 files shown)


c/tskit/tables.c
```diff
@@ -511,6 +511,53 @@ calculate_max_rows(tsk_size_t num_rows, tsk_size_t max_rows,
...
+            new_max_length = TSK_MIN(max_length * 2, TSK_MAX_SIZE);
+            /* Add some constraints to prevent very small allocations */
+            if (new_max_length < 65536) {
+                new_max_length = 65536;
+            }
+            /* Prevent allocating more than 100MB additional unless needed*/
+            if (new_max_length - max_length > 104857600) {
+                new_max_length = max_length + 104857600;
+            }
+            new_max_length = TSK_MAX(new_max_length, current_length + additional_length);
```
jeromekelleher, line 541: Maybe factor this TSK_MAX out to after the two branches? It's doing the same thing, right?

jeromekelleher (commented):
LGTM, one suggestion

benjeffery:
Tests done, should be good to merge.

jeromekelleher (approved):
LGTM!

PR #1710, opened 2021-09-16
Document append/set columns in C API.

### LINKED ISSUE #1672: Document set/append columns in the C API
These methods are currently undocumented, they are advanced use methods that need care, but have their place and should be documented.

benjeffery:
WIP for feedback before copypastathon
Fixes #1672

### DIFF
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -859,7 +859,7 @@ and is not checked for compatibility with any existing schema on this table.
 @param other A pointer to a tsk_individual_table_t object where rows are copied from.
 @param num_rows The number of rows from ``other`` to append to this table.
 @param row_indexes Array of row indexes in ``other``. If ``NULL`` is passed then the
-first ``num_rows`` of ``other`` are used.
+    first ``num_rows`` of ``other`` are used.
 @param options Bitwise option flags. Currently unused; should be
     set to zero to ensure compatibility with later versions of tskit.
 @return Return 0 on success or a negative value on failure.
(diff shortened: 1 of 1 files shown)


c/tskit/tables.h
```diff
@@ -961,19 +961,61 @@ on and may change arbitrarily between versions.
...
-/* Undocumented methods */
+@rst
+Clears the data columns of this table and then copies column data from the specified
+set of arrays. The supplied arrays should all contain data on the same number of rows.
+The metadata schema is not affected.
+@endrst
 
+@param self A pointer to a tsk_individual_table_t object.
+@param num_rows The number of rows to copy from the specifed arrays
+@param flags A pointer to the tsk_flags_t array of flags to be copied
```
jeromekelleher, line 975: Nitpicking here, but this isn't a pointer to an array, it's a pointer or an array (a pointer to an array might be interpreted as tsk_flags_t **). I see we're using this language in the ``add_row`` methods for things like location though. Hmm.

I would say something like ``The array of location values to be copied`` maybe? The type information is redundant and leaving out the pointer language makes it a bit clearer maybe? But, we should probably be consistent, so maybe ``The array of double location values to be copied.``?

jeromekelleher (commented):
Looks good, nitpick about pointer vs array

c/tskit/tables.h
```diff
@@ -961,19 +961,61 @@ on and may change arbitrarily between versions.
...
+@param location A pointer to the double array of locations to be copied
+@param location_offset A pointer to the tsk_size_t array of location offsets to be copied
+@param parents A pointer to the tsk_id_t array of parents to be copied
+@param parents_offset A pointer to the tsk_size_t array of parent offsets to be copied
+@param metadata A pointer to the char array of metadata to be copied
+@param metadata_offset A pointer to the tsk_size_t array of metadata offsets to be copied
+@return Return 0 on success or a negative value on failure.
```
petrelharp, line 982: Nitpicking *even more*, but some of these have periods at the end, and others not. I vote to add periods. =)

c/tskit/tables.h
```diff
@@ -961,19 +961,61 @@ on and may change arbitrarily between versions.
...
 
-/** @} */
+/**
+@brief Replace this table's data by copying from a set of column arrays
 
-/* Undocumented methods */
+@rst
+Clears the data columns of this table and then copies column data from the specified
+set of arrays. The supplied arrays should all contain data on the same number of rows.
+The metadata schema is not affected.
```
petrelharp, line 970: Should we say here somewhere that if there's no locations (for instance) to pass `NULL, 0`? Or maybe that's general-purpose advice about ragged columns we don't need to repeat.

petrelharp:
LGTM!

benjeffery:
Turns out some of the argument names were wrong! :scream: 
Fixedup up and copied to all the tables.

c/tskit/tables.h
```diff
@@ -1930,6 +2095,54 @@ on and may change arbitrarily between versions.
...
+Copies column data from the specified set of arrays to create new rows at the end of the
+table. The supplied arrays should all contain data on the same number of rows. The
+metadata schema is not affected.
+@endrst
+
+@param self A pointer to a tsk_site_table_t object.
+@param num_rows The number of rows to copy from the specifed arrays.
+@param position The array of double position values to be copied.
+@param ancestral_state The array of char ancestral state values to be copied.
+@param ancestral_state_offset The array of tsk_size_t ancestral state offset values to be
```
jeromekelleher, line 2135: maybe indent the "copied" line here? Looks like doxygen doesn't care, so entirely optional

jeromekelleher (approved):
LGTM!

c/tskit/tables.h
benjeffery: Fixed it there and a few other places.

PR #1711, opened 2021-09-16
Document unknown time

### LINKED ISSUE #1685: Document tsk_is_unknown_time
I think this method would be essential for some uses of tskit and should be in the public API.

benjeffery:
Fixes #1685

### DIFF
--- a/c/tskit/core.h
+++ b/c/tskit/core.h
@@ -492,6 +492,20 @@ tsk_size_t tsk_search_sorted(const double *array, tsk_size_t size, double value)
 
 double tsk_round(double x, unsigned int ndigits);
 
+/**
+@brief Check if a number is TSK_UNKNOWN_TIME
+
+@rst
+Unknown time values in tskit are represented by a particular NaN value. Since NaN values
+are not equal to each other by definition, a simple comparison like
+``mutation.time == TSK_UNKNOWN_TIME`` will fail even if the mutation's time is
+TSK_UNKNOWN_TIME. This function compares the underlying bit representation of a double
+value and returns true iff it is equal to the specific NaN value TSK_UNKNOWN_TIME.
+@endrst
+
+@param val The number to check
+@return true if the number is TSK_UNKNOWN_TIME else false
+*/
 bool tsk_is_unknown_time(double val);
 
 #define TSK_UUID_SIZE 36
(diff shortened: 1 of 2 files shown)


c/tskit/core.h
```diff
@@ -492,10 +492,31 @@ tsk_size_t tsk_search_sorted(const double *array, tsk_size_t size, double value)
 
 double tsk_round(double x, unsigned int ndigits);
 
-bool tsk_is_unknown_time(double val);
+/**
+@brief Check if a number is TSK_UNKNOWN_TIME
+
+@rst
```
jeromekelleher, line 498: This is launching in to the middle a bit I think. How about

> Unknown time values in tskit are represented by a particular NaN value. Since NaN values are not equal to each other by definition, a simple comparison like ``mutation.time == TSK_UNKNOWN_TIME`` will fail even if the mutation's time is TSK_UNKNOWN_TIME.  This function compares the underlying bit representation of a double value and returns true iff it is equal to the specific NaN value TSK_UNKNOWN_TIME.

I think that basically covers it.

c/tskit/core.h
```diff
@@ -492,10 +492,31 @@ tsk_size_t tsk_search_sorted(const double *array, tsk_size_t size, double value)
...
+value in this set. The usual NAN is a differing value, but both are NAN.
+See https://en.wikipedia.org/wiki/NaN
 
+This function only returns true for ``tskit.is_unknown_time(tskit.UNKNOWN_TIME)``
+and will return false for ``tskit.is_unknown_time(NAN)`` or any other NAN or
+non-NAN value.
+@endrst
+
+@param val The number to check
+@return True if the number is TSK_UNKNOWN_TIME else False
```
jeromekelleher, line 513: This is pythonish - bools are lowercase in C, like ``true`` and ``false``

c/tskit/core.h
```diff
@@ -492,10 +492,31 @@ tsk_size_t tsk_search_sorted(const double *array, tsk_size_t size, double value)
...
+*/
+bool tsk_is_unknown_time(double val);
+NAN
```
jeromekelleher, line 516: Something weird here?

jeromekelleher (commented):
Looks good, suggested rewording

benjeffery:
@jeromekelleher Fixed up!

jeromekelleher (approved):
LGTM!

Issue #1713, opened 2021-09-16
add TSK_NO_CHECK_ORDERING to flags for `table_collection_check_consistency`


petrelharp:
Over in SLiM we've come across an issue that doing consistency checks helped us find, and we'd like to do them more often. However, we can't [check for mutations that are younger than their node](https://github.com/tskit-dev/tskit/blob/73ece3a93d3d52f03bc1bf348218df2d3a09786f/c/tskit/tables.c#L9377) without also [checking for mutations whose sites are out of order](https://github.com/tskit-dev/tskit/blob/73ece3a93d3d52f03bc1bf348218df2d3a09786f/c/tskit/tables.c#L9343) because there's only one flag, [`TSK_CHECK_MUTATION_ORDERING`](TSK_CHECK_MUTATION_ORDERING) that controls them both (and other things). As SLiM goes along we'd really like to (in DEBUG mode only) check for *consistency* of the tables without checking for the ordering properties that don't affect the validity of the tables. So, similar to how we currently [have `TSK_NO_CHECK_POPULATION_REFS`](https://tskit.dev/tskit/docs/stable/c-api.html#c.tsk_table_collection_check_integrity) it'd be nice to also have `TSK_NO_CHECK_ORDERING` that would disable those checks that we list as "for algorithmic efficiency" over in [the data model documentation](https://tskit.dev/tskit/docs/stable/data-model.html#valid-tree-sequence-requirements), namely, the flag would *not* check:
- sortedness of edges wrt parent time, child, and left, and contgiuousness of edges
- sortedness of sites wrt position
- sortedness of mutations wrt site and parent mutation
- sortedness of migrations wrt time
- sortedness of individuals wrt parent individual

jeromekelleher:
SGTM - I'd favour breaking up the current checks to being more modular rather than making a "super flag" which tries to do all these things at once though.

We can still change the flags before 1.0.

benjeffery:
Great to get these types of things in for 1.0.

petrelharp:
How about `TSK_NO_CHECK_{INDIVIDUAL,EDGE,SITE,MUTATION,MIGRATION}_SORTEDNESS`, and `TSK_NO_CHECK_SORTEDNESS` that implies all of them?

jeromekelleher:
SGTM!

bhaller:
[...]

PR #1714, opened 2021-09-16
remove out of date note about silent mutations

petrelharp:
I noticed while filing #1713 an out of date note about not allowing silent mutations.


jeromekelleher (approved):
Good spot, thanks @petrelharp!

PR #1722, opened 2021-09-17
Rearrange checks

### LINKED ISSUE #1713: add TSK_NO_CHECK_ORDERING to flags for `table_collection_check_consistency`

petrelharp:
This will close #1713. Since we already have all the `TSK_CHECK_X_ORDERING` flags we needed (except X=MIGRATIONS), I just went through to make sure that the things checked under CHECK_ORDERING were all ordering-related. They were, except for mutations. So, I've moved a bunch of consistency checks out of `TSK_CHECK_MUTATION_ORDERING`, i.e., checks that weren't previously done by default that now are: This turned up a few nonsensical tables in the tests. But we should make sure it's actually what we want to do? These checks are:
- mutation time younger than node
- mixed known/unknown times at a site
- site of parent mutation agrees with site

Other notes:
- I don't think we need a `TSK_CHECK_ORDERING` flag, since you get that from `TSK_CHECK_TREES`, and I don't think anyone's going to want to check ordering without checking the other optional properties.
- Hm, *however*, `TSK_CHECK_TREES` also implies running `tsk_table_collection_check_tree_integrity`, which is more expensive (it builds trees; everything else is a single pass through the tables). So, maybe we do want a generic `CHECK_ORDERING`?
- There was no previous check for migration ordering; I've implemented it (although we don't have a `sort` for it yet...)
- I've also made the printing out of tables a tad nicer (it was hard for me to find the table I wanted)

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -623,7 +623,7 @@ write_metadata_schema_header(
 {
     const char *fmt = "#metadata_schema#\n"
                       "%.*s\n"
-                      "#end#metadata_schema\n";
+                      "#end#metadata_schema\n" TABLE_SEP;
     return fprintf(out, fmt, (int) metadata_schema_length, metadata_schema);
 }
 
@@ -1134,7 +1134,7 @@ tsk_individual_table_print_state(const tsk_individual_table_t *self, FILE *out)
 {
     tsk_size_t j, k;
 
-    fprintf(out, TABLE_SEP);
+    fprintf(out, "\n" TABLE_SEP);
     fprintf(out, "tsk_individual_tbl: %p:\n", (const void *) self);
     fprintf(out, "num_rows          = %lld\tmax= %lld\tincrement = %lld)\n",
         (long long) self->num_rows, (long long) self->max_rows,
@@ -1782,7 +1782,7 @@ tsk_node_table_print_state(const tsk_node_table_t *self, FILE *out)
 {
     tsk_size_t j, k;
 
-    fprintf(out, TABLE_SEP);
+    fprintf(out, "\n" TABLE_SEP);
     fprintf(out, "tsk_node_tbl: %p:\n", (const void *) self);
     fprintf(out, "num_rows          = %lld\tmax= %lld\tincrement = %lld)\n",
         (long long) self->num_rows, (long long) self->max_rows,
@@ -2431,7 +2431,7 @@ tsk_edge_table_print_state(const tsk_edge_table_t *self, FILE *out)
 {
     int ret;
 
-    fprintf(out, TABLE_SEP);
+    fprintf(out, "\n" TABLE_SEP);
     fprintf(out, "edge_table: %p:\n", (const void *) self);
     fprintf(out, "options         = 0x%X\n", self->options);
     fprintf(out, "num_rows        = %lld\tmax= %lld\tincrement = %lld)\n",
@@ -3112,7 +3112,7 @@ tsk_site_table_print_state(const tsk_site_table_t *self, FILE *out)
 {
     int ret;
 
-    fprintf(out, TABLE_SEP);
+    fprintf(out, "\n" TABLE_SEP);
     fprintf(out, "site_table: %p:\n", (const void *) self);
     fprintf(out, "num_rows = %lld\t(max= %lld\tincrement = %lld)\n",
         (long long) self->num_rows, (long long) self->max_rows,
@@ -3780,7 +3780,7 @@ tsk_mutation_table_print_state(const tsk_mutation_table_t *self, FILE *out)
 {
     int ret;
 
-    fprintf(out, TABLE_SEP);
+    fprintf(out, "\n" TABLE_SEP);
     fprintf(out, "mutation_table: %p:\n", (const void *) self);
     fprintf(out, "num_rows = %lld\tmax= %lld\tincrement = %lld)\n",
         (long long) self->num_rows, (long long) self->max_rows,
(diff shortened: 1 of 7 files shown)


petrelharp:
I've tested this increased checking with SLiM (the treerec tests and pyslim), and nothing bad happened - do you see anything we wouldn't want to do each time we simplify, @bhaller?

bhaller:
> I've tested this increased checking with SLiM (the treerec tests and pyslim), and nothing bad happened - do you see anything we wouldn't want to do each time we simplify, @bhaller?

I'm quite out of my depth on this, I would say.  :->  But how's this: if you make a PR for SLiM that pulls these changes in, with the appropriate changes to the places in SLiM where we do integrity checks, then I'll test it on my side with my big test suite, with crosschecks enabled, etc.?  And I'll also make sure that it would catch Elissa's bug even with coalescence checking disabled, in DEBUG mode at least.  Sound good?

c/tskit/tables.c
```diff
@@ -9327,35 +9329,63 @@ tsk_table_collection_check_mutation_integrity(
             ret = TSK_ERR_MUTATION_PARENT_EQUAL;
             goto out;
         }
-        /* Check if time is nonfinite */
+        /* Check if time is nonfinite and less recent than node time */
```
jeromekelleher, line 9332: I had to read this a few times - should be *or* less recent, right?

c/tskit/tables.c
```diff
@@ -9327,35 +9329,63 @@ tsk_table_collection_check_mutation_integrity(
...
+            if (mutations.site[j - 1] != mutations.site[j]) {
+                last_known_time = INFINITY;
+                num_known_times = 0;
+                num_unknown_times = 0;
+            }
+        }
+
+        /* Check known/unknown times are not both present on a site */
+        if (unknown_time) {
+            num_unknown_times += 1;
```
jeromekelleher, line 9357: num_unknown_times++; is a bit more C idiomatic

c/tskit/tables.c
```diff
@@ -9327,35 +9329,63 @@ tsk_table_collection_check_mutation_integrity(
...
                 ret = TSK_ERR_TIME_NONFINITE;
                 goto out;
             }
+            if (mutation_time < node_time[mutations.node[j]]) {
+                ret = TSK_ERR_MUTATION_TIME_YOUNGER_THAN_NODE;
+                goto out;
+            }
+        }
+
+        /* reset checks if needed */
```
jeromekelleher, line 9346: Wasn't immediately obvious what the "if needed" meant here. Maybe "Reset counters for the next site". Also, can make it a bit more concise by
```c
if (j > 0 && mutations.site[j - 1] != mutations.site[j]) {
    ...
}
```
this is guaranteed to be safe as short circuiting booleans is part of the C standard.

c/tskit/tables.c
```diff
@@ -9366,29 +9396,11 @@ tsk_table_collection_check_mutation_integrity(
...
                     goto out;
                 }
-                if (mutations.site[parent_mut] != mutations.site[j]) {
-                    ret = TSK_ERR_MUTATION_PARENT_DIFFERENT_SITE;
-                    goto out;
-                }
             }
 
-            /* Check time value ordering */
+            /* Check time ordering, we do this after the time checks above, so
```
jeromekelleher, line 9401: Doesn't quite parse for me. Maybe "Check time ordering. We do..."

jeromekelleher (approved):
LGTM - minor nitpicks

bhaller:
Passes all checks on my side.  Thumbs up.

petrelharp:
Thanks - fixed up those things (and the docs). After I fix whatever github's version of clang-format tells me to do (since it differs from my local one) and push again this should be ready to go.  I noticed that the checks for mixing up known and unknown mutation times within sites don't actually work if the tables aren't ordered, but I also decided we don't care.

benjeffery (approved):
LGTM!
Just needs a changelog @petrelharp

benjeffery:
Woop! Merging.

Issue #1644, opened 2021-08-24
Add top-level "time_units" field


jeromekelleher:
As @hyanwong suggests in https://github.com/tskit-dev/tskit/issues/495#issuecomment-904022584 there are very good reasons for having a top-level ``time_units`` field in the tree sequence/table collection/file format. One use could be to issue a warning if people are using branch length statistics and using uncalibrated time units, where different branch lengths are not comparable.

I guess the type should be a string. We should probably have a number of "known" values. Two I can think of are

- generations_ago
- years_ago
- uncalibrated

I think the "ago" suffix is necessary, because in principle these could be actual datetime values (we could change the time-travel requirement depending then, if so). Tsinfer (and any other method producing arbitrary-ish time values) should use "uncalibrated". The branch length stats would then issue a warning (in Python) if computed on "uncalibrated" tree sequences.

In terms of how we handle compatibility, I think we have to assume ``generations_ago`` if the ``time_units`` field isn't present. Otherwise, we would suddenly start issuing confusing warnings to people using existing tree sequences.

How do we deal with ``time_units`` that are unknown to the library? Do we

1) Hard error in C and refuse to create a tree sequence
2) Raise a warning at load time in Python and treat as "uncalibrated"?
3) Something else?

petrelharp:
For a related issue, see https://github.com/bodkan/slendr/issues/61

benjeffery:
I've added this to the C1.0 milestone as we should consider this before we release the supported API.

I think assuming legacy tree-sequences to be `gererations-ago` isn't quite right as for many that won't be right and could cause confusion/errors. Might it be better to have an `unset` or `legacy` value that doesn't raise errors?

petrelharp:
Just to be clear here, it's only `tsinfer` that'll (currently) be making these `uncalibrated` tree sequences, right? And for `unknown`, we won't error? If so, this sounds good.

> I prefer ALLOW_TIME_UNCALIBRATED to FORCE but that's only a minor thing.

Me, too.

jeromekelleher:
Yeah, AFAIK tsinfer is the only thing that makes tree sequences with uncalibrated time, and yes, the default will be to assume "unknown", which will maintain the current behviour. So, we can't break anyone's code.

> ALLOW_TIME_UNCALIBRATED

Sure, SGTM

hyanwong:
[...]

PR #1733, opened 2021-09-21
time_units in the C API

benjeffery:
A start on #1644 

For now it is set to the default on `clear` and is non-optional for `equals` but we may want to rethink that.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -147,9 +147,11 @@ test_table_collection_equals_options(void)
     tsk_id_t ret_id;
     tsk_table_collection_t tc1, tc2;
 
+    char example_time_units[100] = "An example of time units with unicode ⏰";
     char example_metadata[100] = "An example of metadata with unicode 🎄🌳🌴🌲🎋";
     char example_metadata_schema[100]
         = "An example of metadata schema with unicode 🎄🌳🌴🌲🎋";
+    tsk_size_t example_time_units_length = (tsk_size_t) strlen(example_time_units);
     tsk_size_t example_metadata_length = (tsk_size_t) strlen(example_metadata);
     tsk_size_t example_metadata_schema_length
         = (tsk_size_t) strlen(example_metadata_schema);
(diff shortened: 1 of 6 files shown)


c/tests/test_tables.c
```diff
@@ -403,6 +410,73 @@ test_table_collection_metadata(void)
...
 }
 
+static void
+test_table_collection_time_units(void)
+{
+    int ret;
+    tsk_table_collection_t tc1, tc2;
+
+    char example_time_units[100] = "An example of time units with unicode ⏰";
+    tsk_size_t example_time_units_length = (tsk_size_t) strlen(example_time_units);
```
jeromekelleher, line 420: I'm not sure strlen works when you have unicode in general, it'll stop at the first embedded zero byte, right? Not sure putting the unicode in there does much anyway?

jeromekelleher (commented):
It's not clear how much validation we should do here. Do we have a set of accepted values (like an enum) which we strictly validate against, or do we accept anything and only take special action when certain known values used? I think the conclusion we came to was the only action we'd take is raise an error if we try to do things like compute branch lengths when ``time_units=="uncalibrated"``.

So, I guess we accept anything then, and just check for uncalibrated during init and set the appropriate flag?

c/tests/test_tables.c
benjeffery: It's fine in this test code, we don't do `strlen` anywhere with foreign string data though for that reason as you say.

benjeffery:
> check for uncalibrated during init and set the appropriate flag

I assume you're talking about the init of the (e.g.) branch length code?

benjeffery:
Tests fixed, will follow up with python/lwt changes.

jeromekelleher:
> > check for uncalibrated during init and set the appropriate flag
> I assume you're talking about the init of the (e.g.) branch length code?

Somewhere in ``tsk_treeseq_init`` we'd do something like
```
self->time_uncalibrated = strncmp(self->time_units, self->time_units_length, "uncalibrated") == 0; // or whatever it is.
```

Then we put in checks for ``self->time_uncalibrated`` in various methods that depend on branch lengths.

jeromekelleher (approved):
LGTM - I guess we should merge this much and deal with the semantics of ``uncalibrated" as a follow-up?

We should add this to the various print_state() and __str__ implementations.

benjeffery:
> LGTM - I guess we should merge this much and deal with the semantics of ``uncalibrated" as a follow-up?

Yes, was thinking to get the ground work in first.
 
> We should add this to the various print_state() and **str** implementations.

Good spot! I'll add the C ones of those to this PR.

PR #1734, opened 2021-09-23
Add time_units to python API

benjeffery:
Part of #1644 

Needs docs and change log. 
Next PR is to add this to the tree sequence in both C and Python.

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -1396,7 +1396,7 @@ IndividualTable_set_metadata_schema(IndividualTable *self, PyObject *arg, void *
     if (IndividualTable_check_state(self) != 0) {
         goto out;
     }
-    metadata_schema = parse_metadata_schema_arg(arg, &metadata_schema_length);
+    metadata_schema = parse_unicode_arg(arg, &metadata_schema_length);
     if (metadata_schema == NULL) {
         goto out;
     }
@@ -1963,7 +1963,7 @@ NodeTable_set_metadata_schema(NodeTable *self, PyObject *arg, void *closure)
     if (NodeTable_check_state(self) != 0) {
         goto out;
     }
-    metadata_schema = parse_metadata_schema_arg(arg, &metadata_schema_length);
+    metadata_schema = parse_unicode_arg(arg, &metadata_schema_length);
     if (metadata_schema == NULL) {
         goto out;
     }
@@ -2534,7 +2534,7 @@ EdgeTable_set_metadata_schema(EdgeTable *self, PyObject *arg, void *closure)
     if (EdgeTable_check_state(self) != 0) {
         goto out;
     }
-    metadata_schema = parse_metadata_schema_arg(arg, &metadata_schema_length);
+    metadata_schema = parse_unicode_arg(arg, &metadata_schema_length);
     if (metadata_schema == NULL) {
         goto out;
     }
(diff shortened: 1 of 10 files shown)


python/lwt_interface/tskit_lwt_interface.h
```diff
@@ -1334,6 +1335,27 @@ parse_table_collection_dict(tsk_table_collection_t *tables, PyObject *tables_dic
...
+    value = get_table_dict_value(tables_dict, "time_units", false);
+    if (value == NULL) {
+        goto out;
+    }
+    if (value != Py_None) {
+        if (!PyBytes_Check(value)) {
+            PyErr_Format(PyExc_TypeError, "'time_units' is not bytes");
+            goto out;
+        }
+        err = PyBytes_AsStringAndSize(value, &time_units, &time_units_length);
```
jeromekelleher, line 1348: Likewise here, I think unicode will be a lot simpler for interchange.

python/tskit/tables.py
```diff
@@ -2637,6 +2638,17 @@ def metadata_bytes(self) -> Any:
         """
         return self._ll_tables.metadata
 
+    @property
+    def time_units(self) -> Any:
```
jeromekelleher, line 2642: annotation should by ``str``?

jeromekelleher (commented):
LGTM, but see point about using unicode in Python-C layer. Can follow the example of the metadata schema, I think.

python/_tskitmodule.c
```diff
@@ -6028,6 +6028,51 @@ TableCollection_get_file_uuid(TableCollection *self, void *closure)
...
+
+    if (TableCollection_check_state(self) != 0) {
+        goto out;
+    }
+    if (arg == NULL) {
+        PyErr_Format(PyExc_AttributeError,
+            "Cannot del time_units, set to empty string (b\"\") to clear.");
+        goto out;
+    }
+    err = PyBytes_AsStringAndSize(arg, &time_units, &time_units_length);
```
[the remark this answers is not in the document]
benjeffery: Yes - good call.

benjeffery:
All fixed up, should be good to go.

python/lwt_interface/dict_encoding_testlib.py
```diff
@@ -202,6 +202,15 @@ def test_missing_sequence_length(self, tables):
         with pytest.raises(TypeError):
             lwt.fromdict(d)
 
+    def test_missing_time_units(self, tables):
+        assert tables.time_units != b""
```
jeromekelleher, line 206: Probably don't want to check bytes now?

python/lwt_interface/tskit_lwt_interface.h
```diff
@@ -1334,6 +1335,27 @@ parse_table_collection_dict(tsk_table_collection_t *tables, PyObject *tables_dic
...
+    if (value == NULL) {
+        goto out;
+    }
+    if (value != Py_None) {
+        if (!PyUnicode_Check(value)) {
+            PyErr_Format(PyExc_TypeError, "'time_units' is not a string");
+            goto out;
+        }
+        time_units = parse_unicode_arg(value, &time_units_length);
+        if (time_units == NULL) {
```
jeromekelleher, line 1349: We should try to cover this failure if we can - maybe pass in a non-utf8 encodeable string? Might not be that easy to do though, I guess.

jeromekelleher (approved):
LGTM - nice to push the coverage up a tiny bit if we could, but no biggie.

python/lwt_interface/tskit_lwt_interface.h
benjeffery: Yep extra test with `"\ud861\udd37"` added.

PR #1740, opened 2021-09-29
Fix logic bug in find ibd.

### LINKED ISSUE #1739: ibd_segments crashes sometimes
I got this error message
```
Python(86357,0x1171a15c0) malloc: Incorrect checksum for freed object 0x7fa5a028a800: probably modified after being freed.
Corrupt value: 0x41117b6000000000
Python(86357,0x1171a15c0) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```
when I ran the following code:
```python
ts = msprime.sim_mutations(ts, rate=1.65e-8)
ibd = ts.find_ibd(min_length=min_length, max_time=max_time)
```
This code used to work fine, so I assume it's happened because of the recent changes. I'm using the development version of tskit at [this commit](https://github.com/tskit-dev/tskit/tree/9cd44736be9fbc0092ad26628531bccf69e1255c).

jeromekelleher:
Closes #1739

I think this was a straightforward logic bug in how we were handling min_length. We were only checking if the segment queue needed expanding if the span was greater than min_length, but always adding to it. So basically any time we had a min_length set for a largish example we were going to hit this.

I can confirm the Python tests were provoking the issue. I tried reproducing in C, but it was too awkward and didn't seem worth the effort.

I refactored things a little while I was in there.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7618,8 +7618,7 @@ tsk_ibd_finder_enqueue_segment(
     tsk_segment_t *seg;
     void *p;
 
-    tsk_bug_assert(left < right);
-    if (right - left > self->min_length) {
+    if ((right - left) > self->min_length) {
         /* Make sure we always have room for one more segment in the queue so we
          * can put a tail sentinel on it */
         if (self->segment_queue_size == self->max_segment_queue_size - 1) {
@@ -7632,12 +7631,12 @@ tsk_ibd_finder_enqueue_segment(
             }
             self->segment_queue = p;
         }
+        seg = self->segment_queue + self->segment_queue_size;
+        seg->left = left;
+        seg->right = right;
+        seg->node = node;
+        self->segment_queue_size++;
     }
-    seg = self->segment_queue + self->segment_queue_size;
-    seg->left = left;
-    seg->right = right;
-    seg->node = node;
-    self->segment_queue_size++;
 out:
     return ret;
 }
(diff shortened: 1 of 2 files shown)


benjeffery (approved):
LGTM!

c/tskit/tables.c
```diff
@@ -7779,18 +7777,9 @@ tsk_ibd_finder_run(tsk_ibd_finder_t *self)
             }
```
gtsambos, line 7777: Do we need to enclose this call to `enqueue_segment` inside `if (seg->right - seg->left > self->min_length)`?

gtsambos (approved):
Thanks for looking at this so quickly Jerome -- just the one comment from me.

c/tskit/tables.c
jeromekelleher: I guess it doesn't actually matter in terms of the output, because we filter the segments we record separately. This is reducing the amount of ancestral material track, so should help with performance. Hmm, maybe we should put the check back into ``enqueue_segment``  - I'll update.

jeromekelleher: OK, updated - what do you think @gtsambos?

gtsambos: Yeah, this latest version looks good to me. Anything that reduces unnecessary ancestral material is good -- some users may only be interested in the relatively short list of "large" segments

gtsambos:
Not sure if I need to 'approve' again, but if so, I do!

PR #1760, opened 2021-10-06
Time units error

### LINKED ISSUE #1644: Add top-level "time_units" field

benjeffery:
Fixes #1644 

WIP for feedback.

Is this how you saw this working @jeromekelleher ?

I've added errors to `tsk_treeseq_allele_frequency_spectrum` when `TSK_STAT_BRANCH` and `tsk_treeseq_general_stat` when `TSK_STAT_BRANCH`.

I have not errored on 'tsk_treeseq_kc_distance' as although it uses branch length, it seems like the dimension isn't important there unless you are comparing kc between different tree sequences and we can't detect that.


benjeffery:
@mergifyio rebase

jeromekelleher:
Yeah, this is pretty much it @benjeffery I think. It would be good to add a private ``time_uncalibrated`` boolean to the ``tsk_treeseq_t`` struct though, which we set at init time, so that we don't have to do the tedious strcmping each time.

benjeffery:
Not sure why coverage is unhappy here - there are no lines marked as uncovered.

benjeffery:
Should be good to go now.

PR #1761, opened 2021-10-06
Remove warning stream and TSK_SAMPLE_COUNTS

### LINKED ISSUE #1744: Remove warning stream and TSK_SAMPLE_COUNTS option
#1742 added the concept of a "warn" stream for where the C library could issue warnings. However, this is inherently thread-unsafe, and could lead to nasty problems. We can easily imagine multiple threads doing something that issues warnings, and this messing without our mechanism for getting the stream.

So, I think we should get rid of this idea and remove the only place where we're currently using it.

The debug stream is different I think, as this is definitely just an internal developer option.

benjeffery:
Fixes #1744


jeromekelleher (approved):
LGTM!

Issue #1774, opened 2021-10-08
Remove ordering requirement for the individuals table


bhaller:
In #1138 it was decided to require an ordering for the individuals table: children must come after their individual parents.  This has turned out to be a problem for SLiM, because it has its own ordering requirements for the individuals table which conflict with this policy.  This was not noticed until now because SLiM wasn't using the parents column; but now that it is putting pedigree-based info in the parents column, now we have a conflict between the two ordering policies.  Further discussion of that can be found at https://github.com/MesserLab/SLiM/pull/231.

Discussion on Slack indicates that nobody is actually using this ordering requirement for anything, and that the best solution might be to simply remove tskit's ordering requirement.  Quoting @jeromekelleher throughout:

> Short answer, "no" we're not doing anything with it, and actually, I was wondering if we had got it wrong too. It actually feels quite unnatural when you start working with it, especially if you were to simulate a pedigree in the correct time direction (i.e., backwards :wink:). You'd end up having to reverse the individual table, or else buffer all your individuals and write them in afterwards.
> 
> Are you talking about removing the sortedness requirement entirely, or reversing to children-before-parents?
> 
> I'm open to either.
> 
> ...
> 
> To be honest I haven't found any advantage to having the sort order so far, it's basically just gotten in the way. Any time you try to rely on it, you end up just sorting by time in any case because weird corner cases crop up with ordering of edges otherwise.
> 
> Given that the number of individuals will typically be relatively small, that sort of thing isn't a big deal, so I'd not be particularly upset if we ditched the ordering requirement.

So.  Any objections to simply removing this sorting requirement?  @petrelharp @benjeffery not sure who else to tag...

petrelharp:
> When building a pedigree-based relatedness matrix, outwith tskit, we want parents before progeny so that we can process pedigree records forward in time and accumulate relatedness coefficients appropriately. But one can do topological sort of the pedigree graph (DAG) to achieve this if needed.

Well, it'd be a shame to put @benjeffery's nice code that does the sorting to waste, so we could still provide the functionality (e.g. `tables.sort(individual_parent_sort=False)`).

petrelharp:
But I think if it's getting in the way we should definitely remove it.

bhaller:
Hey campers.  FYI, it would be good to know the decision on this one way or the other; some work on the SLiM side is waiting to know which way this will go.  Implementation is no rush, but a thumbs-up/thumbs-down on the proposal would be helpful.  :->  Thanks!

[...]

PR #1789, opened 2021-10-13
Don't sort individuals by default

benjeffery:
Part of #1774 

I started ripping loads of code out, then realised that maybe this is a less destructive way to do this. (Commented out code will return when the individual sort method uses it.)

WIP for feedback, next is removing the requirement from checks

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -16,7 +16,14 @@
 
 - FIXME breaking changes for tree API and virtual root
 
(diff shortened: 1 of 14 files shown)


benjeffery:
@jeromekelleher I think this is ready, although do we need to check integrity before we sort?

benjeffery:
@jerome all fixed up and changlog added, assuming CI is now happy this is good to go.

PR #1802, opened 2021-10-16
Start conversion of docs to jupyter-book

### LINKED ISSUE #1588: Migrate docs to jupyterbook

hyanwong:
## Description

Initial attempt at converting the docs to JupyterBook. ~~For the moment, I have only attempted to convert the `python-api` file to markdown from `.rst` format. The other pages are all still in `.rst`~~ The `c-api` and `stats` pages are still in .rst format though.

Fixes #1588


hyanwong:
Note that we are getting the following errors:

```
tskit/python/tskit/trees.py:docstring of tskit.TreeSequence.at:1: WARNING: Undefined substitution referenced: "LinearTraversalWarning".
...
tskit/python/tskit/trees.py:docstring of tskit.Tree.parent_array:8: WARNING: Undefined substitution referenced: "tree_array_warning".
```

which are something to do with JupyterBook not liking directives like this: https://github.com/tskit-dev/tskit/blob/ca657b4ac8f99c2394df67222c2ea80519057454/python/tskit/trees.py#L813

And also warnings like

```
tskit/docs/c-api.rst:198: WARNING: c:identifier reference target not found: int32_t
...
```

which suggests that the `nitpick_ignore` lines in the new `_config.yml` are being ignored. I'm not sure what the correct format is here: any ideas @benjeffery ?

hyanwong:
> That's annoyingly nitpicky from sphinx if it insists on lists of tuples! 

Logged as a git issue in Sphinx: https://github.com/sphinx-doc/sphinx/issues/9748

hyanwong:
> Let's forget about nitpicks for now - we can follow that up separately. Just comment the block out.

Agree. But I guess there's no need to comment them out. They are just ignored at the moment because I've removed the "-n" flag in build.sh.

I'm still stuck on the `replace` issue though.

hyanwong:
> Hmm, this is tricky. I was able to get substitutions working within rst files by defining them `_config.yml` within the `rst_prolog`, but it doesn't seem to work in any markdown files.
> 
> I would suggest keeping the Python API file as rst for the first pass, and hopefully this will keep the substitutions working as before.

Another option, which I've just tried out, and which works, is to save each of the warnings in a file in the doc directory, say `tree_array_warning.rst`, containing:

```
.. warning:: This is a high-performance interface which
    provides zero-copy access to memory used in the C library.
    As a consequence, the values stored in this array will change as
    the Tree state is modified as we move along the tree sequence. See the
    :class:`.Tree` documentation for more details. Therefore, if you want to
    compare arrays representing different trees along the sequence, you must
    take **copies** of the arrays.
```

Then we can use `include` in the doctoring instead:

```
    @property
    def parent_array(self):
        """
        A numpy array (dtype=np.int32) encoding the parent of each node
        in this tree, such that ``tree.parent_array[u] == tree.parent(u)``
        for all ``0 <= u < ts.num_nodes``. See the :meth:`~.parent`
        method for details on the semantics of tree parents and the
        :ref:`sec_data_model_tree_structure` section for information on the
        quintuply linked tree encoding.

        .. include:: tree_array_warning.rst
        """
        return self._parent_array
```

I quite like this solution, as we encapsulate the type of box (i.e the `.. warning::` part) in the file itself, and it's obvious where the replacement is stored. What do you think @jeromekelleher ?

hyanwong:
I'm not really sure why we are getting lots of "No module named 'tskit'" errors on doc CI here. It works fine on my machine, although I guess that has tskit properly installed. Do I need to do something strange to get tskit imported itself in the tskit repo?

jeromekelleher:
It's probably something to do with paths - a straight copy from msprime won't work because the directory structure is different. You'll need to put "../python" into sys.path somewhere instead of ".."

grahamgower:
> In fact, the Makefile is a bit weird, with a `dist` target (to set the TSKIT version) as well as `all`. I'm not sure if it's quite right. Perhaps someone else could take a peek?

The `all` target is what you run offline as a developer. The `dist` target gets run when deploying the docs to their online home. The latter just sets the version string appropriately, which isn't important for development.

hyanwong:
> The `all` target is what you run offline as a developer. The `dist` target gets run when deploying the docs to their online home. The latter just sets the version string appropriately, which isn't important for development.

Thanks. I presume `dist` should have the Doxygen step in it, then.

I'm also unclear what 

```
dev:
	PYTHONPATH=${PYPATH} ./build.sh
```

does. I would have thought it should be 

```
dev:
	export PYTHONPATH=${PYPATH}
	./build.sh
```

but the first one compiles fine, and the second one doesn't find the appropriate tskit version. I'm trying to sort this for the issue noted above.

hyanwong:
Should we just merge this, and deal with improvements in later PRs? tagging @benjeffery as I think it'll need 2 approvals to get merged?

astrojuanlu:
FYI, in case you ever want to go back to have versioned docs on RTD while keeping Jupyter Book, have a look at https://jupyterbook.org/publish/readthedocs.html (still to be released)

jeromekelleher:
Thanks very much @astrojuanlu, that's great to know.

Issue #1808, opened 2021-10-18
Missing IBD segments


jeromekelleher:
Adding some new tests based on a simplified IBD definition in #1809 has raised some issues. There are definitely some problems with the IBD segments implementation. Here's one example:

```
tests/test_ibd.py::TestIbdDefinition::test_all_pairs[ts8]                                     
3.19┊     8     ┊                              
    ┊   ┏━┻━━┓  ┊                              
1.44┊   7    ┃  ┊                              
    ┊ ┏━┻┓   ┃  ┊
0.52┊ ┃  ┃   6  ┊      
    ┊ ┃  ┃  ┏┻┓ ┊                                                                             
0.30┊ ┃  5  ┃ ┃ ┊                                                                             
    ┊ ┃ ┏┻┓ ┃ ┃ ┊                                                                             
0.00┊ 0 3 4 1 2 ┊                                                                             
    0           1                                                                             
                                                                                              
LIB: defaultdict(<class 'list'>, {(5, 7): [IbdSegment(left=0, right=1.0, node=7)], (6, 8): [IbdSegment(left=0.0, right=1.0, node=8)], (6, 7): [IbdSegment(left=0.0, right=1.0, node=8)], (7, 8
): [IbdSegment(left=0, right=1.0, node=8)]})   
DEF: {(5, 6): [IbdSegment(left=0, right=1.0, node=8)], (5, 7): [IbdSegment(left=0, right=1.0, node=7)], (5, 8): [IbdSegment(left=0, right=1.0, node=8)], (6, 7): [IbdSegment(left=0, right=1.0
, node=8)], (6, 8): [IbdSegment(left=0, right=1.0, node=8)], (7, 8): [IbdSegment(left=0, right=1.0, node=8)]}
FAILED                                                                       
```

Here we have no IBD segment reported for ``(5, 6)`` in the lib version (this is an example with internal samples). I'm not sure if all problems are to do with internal samples, though.

Probably related, #1677 and #1678

PR #1809, opened 2021-10-18
Finish up IBD

jeromekelleher:
Stacked on #1806 
                 
Try to clear up the remaining stuff to do for IBD. First up though, we seem to have some bugs in the implementation, which we definitely need to fix (#1808).

I can try and get to the bottom of this @gtsambos, but we need to agree on whether my definition of IBD is correct. This is what I have:

```python
def naive_ibd(ts, a, b):
    """
    Returns the IBD segments along the genome for a and b.
    """

    def path(tree, u, v):
        ret = [u]
        while u != v:
            u = tree.parent(u)
            ret.append(u)
        return ret

    tree = ts.first()
    mrca = tree.mrca(a, b)
    last_paths = [path(tree, a, mrca), path(tree, b, mrca)]
    last_mrca = mrca
    left = 0.0
    segs = []
    while tree.next():
        mrca = tree.mrca(a, b)
        paths = [path(tree, a, mrca), path(tree, b, mrca)]
        if paths != last_paths:
            segs.append(tskit.IbdSegment(left, tree.interval.left, last_mrca))
            last_paths = paths
            left = tree.interval.left
            last_mrca = mrca

    segs.append(tskit.IbdSegment(left, ts.sequence_length, last_mrca))
    # Filter out segments with no mrca
    return [seg for seg in segs if seg.node != -1]

def naive_ibd_all_pairs(ts, samples=None):
    samples = ts.samples() if samples is None else samples
    all_pairs_map = {
        (a, b): naive_ibd(ts, a, b) for a, b in itertools.combinations(samples, 2)
    }
    # Filter out pairs with empty segment lists
    return {key: value for key, value in all_pairs_map.items() if len(value) > 0}
```

@petrelharp, do you agree with this as the right starting point for IBD definitions?

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -5455,6 +5455,22 @@ test_ibd_segments_debug(void)
 }
 
 static void
+test_ibd_segments_caterpillar_tree(void)
+{
+    int ret;
+    tsk_ibd_segments_t result;
+    tsk_treeseq_t *ts = caterpillar_tree(100, 1, 5);
+
+    /* We're just testing out the memory expansion in ibd_finder */
+    ret = tsk_table_collection_ibd_within(ts->tables, &result, NULL, 0, 0.0, DBL_MAX, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_ibd_segments_free(&result);
+
+    tsk_treeseq_free(ts);
+    free(ts);
+}
+
+static void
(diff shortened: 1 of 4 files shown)


gtsambos:
Just scanning this now @jeromekelleher -- the simplified definition looks good to me! Thanks for all this work.

jeromekelleher:
I went ahead and implemented the updates in C, and I think it's all working now and should be ready to go!

PR #1819, opened 2021-10-19
Implement TreeSequence.discrete_genome

### LINKED ISSUE #1144: Add discrete_genome property to TreeSequence
It's probably worth defining a cached property on the TreeSequence object called ``discrete_genome`` that checks if all genome coordinates are integers. Or, I guess, to do it properly, we'd add a ``discrete_genome`` bool attribute to the ``tsk_treeseq_t`` struct which checks if all genome coordinates are integers during initialisation and sets the flag accordingly. I don't think we should add it to the table collection, because then we'd have to worry about the flag going out of date, as the tables are updated.

This would help settle some other long-standing issues like #353.

jeromekelleher:
Closes #1144

This should unplug a few outstanding issues. We could make things a bit more granular and just focus on discrete sites/breakpoints etc, but I don't think it's worth the bother. Tree sequences mixing discrete and continuous coordinates are going to be rare, and probably a result of a mistake somewhere along the way.

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -926,6 +926,104 @@ verify_empty_tree_sequence(tsk_treeseq_t *ts, double sequence_length)
  *======================================================*/
 
 static void
+test_simplest_discrete_genome(void)
+{
+    const char *nodes = "1  0   0\n"
+                        "1  0   0\n"
+                        "0  1   0";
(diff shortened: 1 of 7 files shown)


jeromekelleher:
Thanks @benjeffery - I didn't see the conflict, sorry.

PR #1825, opened 2021-10-20
Output alignments

jeromekelleher:
Laying the groundwork for #353, this adds the ability to output a full sequence for each sample optionally using some reference genome. Since we'll often want to do this some some basic random reference, I've added a simple utility function to generate random reference sequences (which we could make more complicated, if we wanted).

It works like this:
```python

import msprime

ts = msprime.sim_ancestry(4, sequence_length=45, random_seed=234)
ts = msprime.sim_mutations(ts, rate=1e-1, random_seed=2345)

print("L = ", ts.sequence_length, ts.num_sites)
reference = tskit.random_nucleotides(ts.sequence_length, seed=42)
print(reference)
print()
for a in ts.alignments(reference_sequence=reference):
    print(a)
```
giving
```
L =  45.0 37
TGATTGAATCTTTTGAGGGTCACGGCCCGGAAGCCAGAATTTCGG

TCATTCTCCGACCCTGCGATCAGTTATTACTAGTCAAATGTTCTC
TAATTTTCAACCGCGGAAGTCATCTATGTCTAGTCTAAGGTCCCC
TATTTCTCCACCGCTGCGGTTCTCTACTAGTAGTCAAATGTCCTA
TCATTCTCCGACCCTGCGATCAGTTATTACTAGTCAAATGTTCTC
AGATTTACTGAACCTTCGGTCGTCGTTTACGAGTTAGAGATACTC
TATTTTTGAATCGGTGCAGTCGTCCATGACCAGGCAAAGGTTCGC
AGATTTACTGAACCTTCGGTCGTCGTTTACGAGTTAGAGATACTC
TATTTTTGAAACGGTGCAGTCGTCCACGACTAGGCAAAGGTTCGC
```

It's built on existing technology, so the semantics of tricky stuff like missing data should be clear. Hopefully the limited context of a discrete genome and simple mutations should make this semantically straightforward also. I've called it ``alignments`` as we can imagine actually dealing with indels and stuff at some point and outputtting them like this. Also, some time in the future we can imagine automatically using the tree sequence's own reference sequence by default (#146), if it's present.

I haven't written any tests, as I wanted to see what the basic reaction to the idea was.

Any thoughts @benjeffery @hyanwong @petrelharp ?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -4500,6 +4500,39 @@ class TreeSequence:
             if tree2.interval.right == right:
                 tree2 = next(trees2, None)
 
+    def _haplotypes_array(self, *, isolated_as_missing, missing_data_character):
+        H = np.empty((self.num_samples, self.num_sites), dtype=np.int8)
+        missing_int8 = ord(missing_data_character.encode("ascii"))
+        for var in self.variants(isolated_as_missing=isolated_as_missing):
+            alleles = np.full(len(var.alleles), missing_int8, dtype=np.int8)
+            for i, allele in enumerate(var.alleles):
+                if allele is not None:
+                    if len(allele) != 1:
+                        raise TypeError(
+                            "Multi-letter allele or deletion detected at site {}".format(
+                                var.site.id
+                            )
+                        )
+                    try:
+                        ascii_allele = allele.encode("ascii")
+                    except UnicodeEncodeError:
+                        raise TypeError(
+                            "Non-ascii character in allele at site {}".format(
+                                var.site.id
+                            )
+                        )
+                    allele_int8 = ord(ascii_allele)
+                    if allele_int8 == missing_int8:
+                        raise ValueError(
+                            "The missing data character '{}' clashes with an "
+                            "existing allele at site {}".format(
+                                missing_data_character, var.site.id
+                            )
+                        )
+                    alleles[i] = allele_int8
+            H[:, var.site.id] = alleles[var.genotypes]
+        return H
+
     def haplotypes(
         self,
         *,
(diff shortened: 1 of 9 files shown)


python/tskit/trees.py
```diff
@@ -4677,6 +4683,107 @@ def genotype_matrix(
...
+        .. note:: This is inherently a **zero-based** representation of the sequence
+            coordinate space. Care will be needed when interacting with other
+            libraries and upstream coordinate spaces.
+
+        If a ``reference_sequence`` is supplied this will be used to fill
+        in the nucleotide positions between the sites in the tree sequence; otherwise,
+        nucleotides between sites are treated as missing data and therefore filled
+        in with the ``missing_data_character`` (by default "-").
+
+        .. warning:: The future behaviour of this function will change when the
```
benjeffery, line 4708: I'm wondering if it might be better here to error out when reference_sequence=None with a message saying that this will be supported in future?

benjeffery (approved):
LGTM, just one thought about the ref seq.

jeromekelleher:
Good question, probably one to get wider input on. @petrelharp @bhaller @molpopgen this adds a new function ``ts.alignments()`` which takes an optional ``reference_sequence`` argument (see the diff for the full description/docs).

The question is, when a ``reference_sequence`` is no specified, should we:

.1 Fill in gaps between sites with missing data; or 
.2 Raise an error saying "we haven't implemented this yet" (i.e., storing a reference sequence in the tree sequence, #146)

Currently we're doing 1, which is more convenient, but will lead to behaviour changes in the future which could result in bugs. Maybe 2 is better? Any thoughts?

bhaller:
I would lean towards (2) just because it preserves flexibility for later.  I gather you folks don't want to settle on a final design for references sequences now; so it seems wise to make sure that you don't paint yourselves into a corner.  If you decide you want the behavior of (1) later you can always add it, whereas if you go with (1) now and later realize that something different would be better you'll have a backward-compatibility problem, right?  So, I don't understand all the issues here (e.g., I don't really understand what "missing data" really does/implies in tskit), but on general principles I'd vote for (2).

molpopgen:
I think 2 is the way to go at the moment.

jeromekelleher:
Yeah, makes sense. I'll update to do (2) - thanks for the input @bhaller and @molpopgen!

jeromekelleher:
Updated - think this is ready to go now.

PR #1826, opened 2021-10-20
Allow skipping of sites and mutations in sort

### LINKED ISSUE #1475: Sort only the edge table
From Python, I would like to be able to sort the edge table on its own, without sorting any of the other tables. After discussion with @benjeffery it seems like this should be relatively easy functionality to add. Since the TableCollection call signature is currently:

```
def sort(self, edge_start=0)
```

he suggests (I think) that we could add `sites_start`, `mutations_start` and `individuals_start` as params, and if we set all those to the max, the corresponding table wouldn't be sorted. Correct me if I'm wrong, Ben.

This is to address #1474, because `TableCollection.keep_intervals()` does a sort on all the tables, whereas all we really want it to do is just to sort the edges.

benjeffery:
Fixes #1475


benjeffery:
All fixed up, merging.

PR #1831, opened 2021-10-21
Make TreeSequence.samples args kw only

### LINKED ISSUE #1715: Make arguments after ``population`` in TreeSequence.samples keyword only
We add a new ``time`` argument to ts.samples in #1692. We should make sure that this is used only as a keyword, especially with the deprecated ``population_id`` argument still lying around.

I suggest we change the signature to 
```python
def samples(self, population=None, *, population_id=None, time=None):
    ...
```

benjeffery:
Fixes #1715


benjeffery:
Test added and set to merge.

PR #1838, opened 2021-10-22
Change node labelling and add rooting to Nexus.

### LINKED ISSUE #1803: Node labels in to_nexus are ambiguous
As @jeetsukumaran points out our default node labels in ``to_nexus`` are ambiguous as they are in the form ``"tsk_{node.id}_{node.flags}"``, and these would be interpreted as ``"tsk {node.id} {node.flags}"`` by conformant nexus parsers.

Since it's unlikely many people are using the to_nexus functionality at the moment (given that it's broken! #1785 ) I think we're free to change things around a bit. 

The idea of including the ID and flags in the label is so that samples can be identified, but maybe the flags are extraneous for most applications (the samples are nearly always be the leaf nodes), and we would be better off keeping things simple and just including the ID. The idea of the ``tsk`` prefix is that some things get confused by integer names, but maybe that's extraneous as well. So, perhaps the best default behaviour would be to simply use the bare ID values?

As we're updating the method, I think it's worthwhile adding (or at least planning for) custom node labels as well, via a mapping like we use in the current "newick" method. (Which we're planning on deprecating, #1671 )

Any thoughts here @jeetsukumaran? Great to get your input here so that we can at least lay the foundations for getting good interoperability with the classical phylo formats.

### LINKED ISSUE #1815: Nexus trees should be marked with "rooted" indicator
As @jeetsukumaran notes in https://github.com/tskit-dev/tskit/issues/1803#issuecomment-945215915

> The other big thing is to whether or not the trees should be seen as rooted vs unrooted. Typically, a tree string is prepended with "[&R]" to indicate a rooted tree vs "[&U]" an unrooted tree. This is a fairly standard convention and widely -supported. The problem is when such a token is not present, should the trees be considered unrooted (even if there is a basal bifurcation) or rooted? Here there is no consensus. Many and maybe most bioinformatics programs assume rooted trees, but PAUP* (which for me is the reference standard for the Nexus foramt) assumes unrooted trees (and so does DendroPy).

I responded:

> Trees are always rooted in tskit as far as I know, since there's a time associated with each node and therefore always an unambiguously "deepest node". So I think adding the [&R] prefix sounds like a good idea.

So I think we should add the rooting prefix.

jeromekelleher:
Batch of improvements for nexus handling.

Closes #1803
Closes #1815


jeromekelleher:
@mergifyio rebase

benjeffery:
@mergifyio rebase

python/tskit/trees.py
```diff
@@ -2327,56 +2331,83 @@ def _node_edges(self):
...
+            5 + max_label_size + math.ceil(math.log10(root_time)) + precision
+        )
+        buffer_size = 1 + single_node_size * self.num_nodes
+        return self._ll_tree.get_newick(
+            precision=precision,
+            root=root,
+            buffer_size=buffer_size,
+            legacy_ms_labels=legacy_ms_labels,
+        )
+        return
```
benjeffery, line 2350: Looks like an extra return slipped in.

benjeffery (approved):
Very nice, just one fix. Agree with the naming you've gone for here.

python/tskit/trees.py
jeromekelleher: Aha, thanks! Surprising flake8 didn't spot this.

PR #1861, opened 2021-10-27
Fixup num_tracked_samples to work with virtual_root

jeromekelleher:
Closes #1724

I haven't tested the set_from_sample_list as I'm kinda hoping to delete that code path.

PR #1862, opened 2021-10-27
Remove tree.left_root

### LINKED ISSUE #1796: Remove tree.left_root from C API.
#1704 added thevirtual root, but didn't remove the existing ``left_root``.

jeromekelleher:
Stacked on #1861

Closes #1796

The cases where we actually do want to explicitly iterate over the roots are rare; it's much better to push the virtual root onto a stack and let the general tree algorithms take care of things. So, there will be a bit of breakage in user code, but I think it's worth it to move them onto better ways of doing things.

TODO - the examples need updating here and we need some changelogging


jeromekelleher:
This is ready to go in I think - I've created #1886 to track updating the documentation, which needs some better examples.

benjeffery (approved):
LGTM!

PR #1864, opened 2021-10-28
Refactor LD calculation

jeromekelleher:
Refactor the LD calculator to use standard tree operations. Most of the old complexity was tied up in trying to maintain state between calls, but this is fragile and most likely pointless.

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2019 Tskit Developers
+ * Copyright (c) 2019-2021 Tskit Developers
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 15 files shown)


jeromekelleher:
This is ready for review I think. I've ended up making quite a few changes in order to make the LD calculation code simpler and to remove our dependency on awkward parts of the tree API like ``marked`` and ``sample_lists``. There are some nice new features to the C API though:

- tsk_tree_seek(x) moves the tree to position x in a reasonably intelligent way. We can imagine making this better in future when we have the seek index implemented.
- the ``preorder_samples`` method should make implementing ``tree_get_variant`` straightforward
- I'm not sure what we'll do about ``track_descendant_samples`` for version 1.0 API. I think it's probably best to keep all the ``tracked_samples`` stuff as undocumented in the C API in case we want to change the details later. So, I think it's fine like this, for now.

There's also a good bit of paying down on technical dept - I got rid of the old buffer API stuff used from the pre-numpy days. Even though we're removing the LDCalculator, this will still be handy when implementing the ultimate LD APIs (#1900 )

The downside of this change is that there is a perf regression when calculating large numbers of LD values in a structured way. Much of the complexity of the old implementation was in trying to maintain state between calls, and this did help when computing large matrices.

Here's some benchmarking:
```python
print("version:", tskit.__version__)

ts = tskit.load("../benchmark-n100k-m1M.trees")
ld_calc = tskit.LdCalculator(ts)
before = time.perf_counter()
A = ld_calc.r2_matrix()
print(f"Matrix shape{A.shape} = {duration:.2f}")


ts = tskit.load("../unified_genealogy_sim_mutations.trees")
print("loaded num_sites =", ts.num_sites)

before = time.perf_counter()
ld_calc = tskit.LdCalculator(ts)
a = ld_calc.get_r2_array(0)
print(f"All with 0 = {duration:.2f}")

before = time.perf_counter()
d = 1_000_000
for j in range(min(ts.num_sites, 1000)):
    a = ld_calc.get_r2_array(j, max_distance=d)
print(f"1000 sites, within 1Mb upstream = {duration:.2f}")
```
We get:
```
version: 0.3.7
Matrix shape(4943, 4943) = 8.88
loaded num_sites = 9626354
All with 0 = 4.13
1000 sites, within 1Mb upstream = 20.09
```
and
```
version: 0.3.8.dev1
Matrix shape(4943, 4943) = 28.09
loaded num_sites = 9626354
All with 0 = 2.14
1000 sites, within 1Mb upstream = 1.07
```

So, a mixed bag: version 0.3.7 is ~3x faster in computing the small LD matrix in a nice clean simulation, but the new code is 20x faster when computing localised LD in a real data tree sequence. (I think most of the perf gain here is from not using "sample lists", as maintaining these is quite expensive on trees with lots of polytomies.)

I'm not overly bothered about the LD matrix being a bit slower than before and am happy to log this as an issue for us to track to see if we can replicate the old performance with a new and better API (#1900). 

Any thoughts @benjeffery @petrelharp @apragsdale?

benjeffery (approved):
Wow, some great stuff in here. I've gone through pretty closely. I'm happy with the perf regression as it isn't a total loss and this is such an improvement.

petrelharp:
The performance hit seems not a big deal. I say go for it!

apragsdale:
This looks great, and to me the performance hit for the LD matrix seems not so bad.

jeromekelleher:
OK, let's merge!

PR #1865, opened 2021-10-29
Convert stats docs to markdown format

hyanwong:
Useful because we may want to expand these to include code examples.


jeromekelleher (approved):
LGTM, thanks.

PR #1866, opened 2021-10-31
Reduce number of tests in treestats

jeromekelleher:
This is an attempt to make the test suite a bit less painful to run. I've had to do something because we're regularly hitting timeouts on the CircleCI runs, which means we're not getting any updates on CodeCov.

We're not losing anything by removing these tests, and this makes things at least practical to run for now. We still need to do something about the TraitLinearModel tests, they are just too slow.

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -557,10 +557,6 @@ class TopologyExamplesMixin:
     actual tests.
     """
 
-    def test_single_tree(self):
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This seems to reduce runtime on CircleCI from ~17 minutes to 12, which is hopefully short enough to be reliable again. We really need to do something about the test suite though, it's making it much too hard to make quick changes.

PR #1868, opened 2021-10-31
Try to make circleCI run more reliably

jeromekelleher:
We're currently using 4 threads but only have 2 VCPUs. Let's see if this makes things any less wobbly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ commands:
           name: Run Python tests
           command: |
             cd python
-            python -m pytest --cov=tskit  --cov-report=xml --cov-branch -n2 tests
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This moves the code coverage upload for the main Python tests to GitHub actions, so that we don't have to run all of the tests on CircleCI any more. We only run a couple of test modules which are interacting most closely with the low-level C module. There's some complexity now, in that we're uploading to codecov from a few different places, but it seems worth it since CI is *much* faster.

There's a little drop off in our test coverage in the low-level module, but we can make an issue to track this.

jeromekelleher:
Looks like this is working so I'm going to merge to see if it makes the dev environment a bit more stable.

benjeffery:
Great, makes sense now that the actions tests are mature.

PR #1876, opened 2021-11-02
Fix type for size array

### LINKED ISSUE #1805: sample_set_sizes using np.uint32 rather thank np.uint64
In e.g. ``__one_way_sample_set_stat`` we create a numpy array for tsk_size_t which is 32 bit instead of 64 bit. Totally harmless, since the underlying code will cast up, but might as well fix.

benjeffery:
Fixes #1805

I've had a good look for other instances.


jeromekelleher (approved):
Lgtm

PR #1877, opened 2021-11-02
Cache some test functions to speed up tests

jeromekelleher:
This should knock another 30 seconds or so off the test suite runtime. I haven't looked to see if there are any more of these we can cache, will update later.

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -43,6 +43,25 @@ import tskit.exceptions as exceptions
 np.random.seed(5)
 
 
+def cached_np(func):
+    """
+    Decorator to speed up functions that take numpy arrays as positional
+    arguments that get called a lot with the same arguments.
+
+    # See https://github.com/tskit-dev/tskit/issues/1856 for more info
+    """
+    cache = {}
+
+    def f(*args):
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I can't find any other low-hanging fruit here, barring a significant update to the test_tree_stats file I don't think we'll do much better. This is a major improvement though.

benjeffery (approved):
LGTM, but I'm getting very little benefit, maybe 1-2s. I guess this works better singly threaded?

jeromekelleher:
Ah - well, it's no harm anyway.

molpopgen:
Is it possible for the caching to be a race condition when running tests in parallel?

benjeffery:
> Is it possible for the caching to be a race condition when running tests in parallel?

Don't think so, as the parallel tests run in separate processes, likely why there is less benefit as the cache is not shared.

molpopgen:
> > Is it possible for the caching to be a race condition when running tests in parallel?
> 
> Don't think so, as the parallel tests run in separate processes, likely why there is less benefit as the cache is not shared.

Right forgot about that.  Although they are ugly, that's exactly why session wide fixtures exist in pytest.

PR #1881, opened 2021-11-02
Use pattern rules in C examples Makefile

clwgg:
## Description

While playing with the C-API for the first time I found the example code in `c/examples/` extremely helpful. To avoid having to touch the build system for the API I used the `Makefile` in this folder to compile my custom examples. To make this easier I changed the Makefile to use pattern rules. This way, a new compile target simply has to be added to the `targets`, without any other manual changes. This PR contains these changes (without the addition of any other custom examples) in case it is helpful or of interest for the project. If not, feel free to dismiss the PR.


benjeffery (approved):
LGTM! Makes sense and looks like `clean` was out of date as it was.

Issue #1854, opened 2021-10-25
a way to read .trees metadata without loading the whole file would be useful


bhaller:
This came up in SLiM (https://github.com/MesserLab/SLiM/discussions/233) but I think it has general utility, as I'll describe.  What the user wants is a way to get the metadata from a .trees file without the big overhead of loading the whole .trees into memory (whether as tables or a tree sequence); just the metadata.

In SLiM they want this so they can read in parameters that they previously put into the metadata, before the point in their script where they would actually load in the .trees data.

This seems generally useful to me because someone might wish to, e.g., loop in Python over the .trees files in a directory that contains many of them, and do "something" with each .trees file that has metadata with a certain property.  Like: process all the .trees files that come from msprime but not those from SLiM, or copy all those where their parameter XYZZY had a value of 15.5 into a different folder, or whatever.

So for SLiM it'd be great to have C API for this; for other uses Python API seems called for.  Would be nice to have it in C API 1.0 – we would use it immediately in SLiM – but it's not a big deal if it isn't.

benjeffery:
@clwgg @bhaller this is a great opportunity to debug the dev docs, so please don't hold back on corrections or suggestions.

clwgg:
@benjeffery thanks, the dev docs are pretty fantastic! they made it much much easier to get into the project from the outside, especially when touching the backend for the first time. I just opened a PR that fixes two broken links I found (https://github.com/tskit-dev/tskit/pull/1880) -- otherwise I don't think they missed anything I felt I needed.

clwgg:
alright, I just opened a (very-much-a-)draft PR for this feature. looking forward to your feedback (:

bhaller:
Hi @clwgg, @benjeffery.  Just wondering whether this is likely to make it into C API 1.0.  @clwgg, do you care?  (If it makes it into C API 1.0 then it will make it into SLiM 3.7; if not, probably not.)

jeromekelleher:
It's basically ready, we'll get it in for 0.4.0/1.0

bhaller:
[...]

Issue #1720, opened 2021-09-17
Review C API flags


jeromekelleher:
We can't change any flags (numbers or names) after 1.0 is tagged, so we should go through them and make sure we're happy with them all.

cf. #1713

jeromekelleher:
Note: think about the new TSK_LOAD_SKIP_TABLES flag added in #1882

benjeffery:
Most flags, for example all the simplify flags need namespacing.

jeromekelleher:
To close this, we want to document all the flags we're using as constants in sphinx, and make sure we have these referred to in the HTML documentation. We probably want to formalise the groupings of the args.

PR #1882, opened 2021-11-02
read metadata without loading tables

clwgg:
## Description

This is a first stab at https://github.com/tskit-dev/tskit/issues/1854. I think there are a couple of decisions to make here that I don't feel qualified to make, so I would greatly appreciate any feedback. I've tried to document the process in the commit log as best I could, but to avoid everyone having to stumble through those I thought I'd put a little write-up here as well.

The main decision point I hit early on is whether `tsk_table_collection_load_indexes` should be part of the conditional within `tsk_table_collection_loadf_inited` or not. Without any further changes, either decision leads to errors, but ONLY if a table_collection loaded with `TSK_LOAD_SKIP_TABLES` is ever attempted to be put into a tree sequence (which I implemented but ended up reverting for this PR since I don't know if its necessary for the use cases we envision -- more on this below).

Let me walk you through each of these cases, assuming that `TSK_LOAD_SKIP_TABLES` is set:

If `tsk_table_collection_load_indexes` is **not** called, the table collection won't have it's indexes set, so if we attempt to load this table collection into a tree sequence we encounter an error in `tsk_table_collection_check_integrity`.

If `tsk_table_collection_load_indexes` is called and we attempt to add indexes to a table collection without having loaded its tables, we encounter an error within the `..._load_indexes` function when checking whether `(num_rows != self->edges.num_rows)` since we never actually loaded the edge table.

What I ended up doing is to call the `..._load_indexes` function, but pass the `TSK_LOAD_SKIP_TABLES` option to it to skip the `(num_rows != self->edges.num_rows)` check in case the option is set. My motivation was that it feels cleaner to have an empty but indexed table collection, than an empty, un-indexed table collection that throws an error whenever it tries to touch a tree sequence. Please note though that I'm not totally sure if there's any unwanted consequences of skipping that check (it seems to just check if the edge table was loaded correctly or was malformed in some way), so I rely on your advice here.

I then exposed this flag to the python `TableCollection` class through its C interface. I ended up reverting a change that allowed to also pass this option all the way through a TreeSequence since I don't really see the use case of that, but of course I'd be happy to re-revert this if needed. What I did instead is add a top-level `tskit.load_metadata` function which just calls `TableCollection.load(file, skip_tables=True)` and returns the metadata. I added this mostly as a proof-of-principle of what I am envisioning the use-case of this flag to be.

# PR Checklist:

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -9971,22 +9971,30 @@ bool
 tsk_table_collection_equals(const tsk_table_collection_t *self,
     const tsk_table_collection_t *other, tsk_flags_t options)
 {
-    bool ret
-        = self->sequence_length == other->sequence_length
-          && tsk_individual_table_equals(
-                 &self->individuals, &other->individuals, options)
-          && tsk_node_table_equals(&self->nodes, &other->nodes, options)
-          && tsk_edge_table_equals(&self->edges, &other->edges, options)
-          && tsk_migration_table_equals(&self->migrations, &other->migrations, options)
-          && tsk_site_table_equals(&self->sites, &other->sites, options)
-          && tsk_mutation_table_equals(&self->mutations, &other->mutations, options)
-          && tsk_population_table_equals(
-                 &self->populations, &other->populations, options)
-          && (self->time_units_length == other->time_units_length
-                 && tsk_memcmp(self->time_units, other->time_units,
-                        self->time_units_length * sizeof(char))
-                        == 0);
-
+    bool ret = self->sequence_length == other->sequence_length
+               && self->time_units_length == other->time_units_length
+               && tsk_memcmp(self->time_units, other->time_units,
+                      self->time_units_length * sizeof(char))
+                      == 0;
+    if (!(options & TSK_CMP_IGNORE_TABLES)) {
+        ret = ret
+              && tsk_individual_table_equals(
+                     &self->individuals, &other->individuals, options)
+              && tsk_node_table_equals(&self->nodes, &other->nodes, options)
+              && tsk_edge_table_equals(&self->edges, &other->edges, options)
+              && tsk_migration_table_equals(
+                     &self->migrations, &other->migrations, options)
+              && tsk_site_table_equals(&self->sites, &other->sites, options)
+              && tsk_mutation_table_equals(&self->mutations, &other->mutations, options)
+              && tsk_population_table_equals(
+                     &self->populations, &other->populations, options);
+        /* TSK_CMP_IGNORE_TABLES implies TSK_CMP_IGNORE_PROVENANCE */
+        if (!(options & TSK_CMP_IGNORE_PROVENANCE)) {
+            ret = ret
+                  && tsk_provenance_table_equals(
+                         &self->provenances, &other->provenances, options);
+        }
+    }
     /* TSK_CMP_IGNORE_TS_METADATA is implied by TSK_CMP_IGNORE_METADATA */
     if (options & TSK_CMP_IGNORE_METADATA) {
         options |= TSK_CMP_IGNORE_TS_METADATA;
@@ -10002,12 +10010,6 @@ tsk_table_collection_equals(const tsk_table_collection_t *self,
                             self->metadata_schema_length * sizeof(char))
                             == 0);
     }
-    if (!(options & TSK_CMP_IGNORE_PROVENANCE)) {
-        ret = ret
-              && tsk_provenance_table_equals(
-                     &self->provenances, &other->provenances, options);
-    }
-
     ret = ret
           && tsk_reference_sequence_equals(
                  self->reference_sequence, other->reference_sequence, options);
@@ -10512,12 +10514,15 @@ out:
 }
 
 static int TSK_WARN_UNUSED
-tsk_table_collection_loadf_inited(tsk_table_collection_t *self, FILE *file)
+tsk_table_collection_loadf_inited(
+    tsk_table_collection_t *self, FILE *file, tsk_flags_t options)
 {
     int ret = 0;
     kastore_t store;
 
-    ret = kastore_openf(&store, file, "r", KAS_READ_ALL);
+    int kas_flags = options & TSK_LOAD_SKIP_TABLES ? 0 : KAS_READ_ALL;
+    ret = kastore_openf(&store, file, "r", kas_flags);
+
     if (ret != 0) {
         if (ret == KAS_ERR_EOF) {
             /* KAS_ERR_EOF means that we tried to read a store from the stream
@@ -10534,41 +10539,48 @@ tsk_table_collection_loadf_inited(tsk_table_collection_t *self, FILE *file)
     if (ret != 0) {
         goto out;
     }
-    ret = tsk_node_table_load(&self->nodes, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_edge_table_load(&self->edges, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_site_table_load(&self->sites, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_mutation_table_load(&self->mutations, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_migration_table_load(&self->migrations, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_individual_table_load(&self->individuals, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_population_table_load(&self->populations, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_provenance_table_load(&self->provenances, &store);
-    if (ret != 0) {
-        goto out;
-    }
-    ret = tsk_table_collection_load_indexes(self, &store);
-    if (ret != 0) {
-        goto out;
+    if (!(options & TSK_LOAD_SKIP_TABLES)) {
+        ret = tsk_node_table_load(&self->nodes, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_edge_table_load(&self->edges, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_site_table_load(&self->sites, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_mutation_table_load(&self->mutations, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_migration_table_load(&self->migrations, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_individual_table_load(&self->individuals, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_population_table_load(&self->populations, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_provenance_table_load(&self->provenances, &store);
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_table_collection_load_indexes(self, &store);
+        if (ret != 0) {
+            goto out;
+        }
+    } else {
+        ret = tsk_table_collection_build_index(self, 0);
+        if (ret != 0) {
+            goto out;
+        }
     }
     ret = tsk_table_collection_load_reference_sequence(self, &store);
     if (ret != 0) {
@@ -10597,7 +10609,7 @@ tsk_table_collection_loadf(tsk_table_collection_t *self, FILE *file, tsk_flags_t
             goto out;
         }
     }
-    ret = tsk_table_collection_loadf_inited(self, file);
+    ret = tsk_table_collection_loadf_inited(self, file, options);
     if (ret != 0) {
         goto out;
     }
@@ -10623,7 +10635,7 @@ tsk_table_collection_load(
         ret = TSK_ERR_IO;
         goto out;
     }
-    ret = tsk_table_collection_loadf_inited(self, file);
+    ret = tsk_table_collection_loadf_inited(self, file, options);
     if (ret != 0) {
         goto out;
     }
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -724,6 +724,11 @@ typedef struct {
 /* Flags for table collection init */
 #define TSK_NO_EDGE_METADATA (1 << 0)
 
+/* Flags for table collection load */
+/* This shares an interface with table collection init.
+   TODO: review as part of #1720 */
+#define TSK_LOAD_SKIP_TABLES (1 << 1)
+
 /* Flags for table init. */
 #define TSK_NO_METADATA (1 << 0)
 
@@ -736,6 +741,7 @@ typedef struct {
 #define TSK_CMP_IGNORE_PROVENANCE (1 << 1)
 #define TSK_CMP_IGNORE_METADATA (1 << 2)
 #define TSK_CMP_IGNORE_TIMESTAMPS (1 << 3)
+#define TSK_CMP_IGNORE_TABLES (1 << 4)
 
 /* Flags for table collection clear */
 #define TSK_CLEAR_METADATA_SCHEMAS (1 << 0)
@@ -3412,6 +3418,9 @@ TSK_CMP_IGNORE_TS_METADATA
 TSK_CMP_IGNORE_TIMESTAMPS
     Do not include the timestamp information when comparing the provenance
     tables. This has no effect if TSK_CMP_IGNORE_PROVENANCE is specified.
+TSK_CMP_IGNORE_TABLES
+    Do not include any tables in the comparison, thus comparing only the
+    top-level information of the table collections being compared.
 @endrst
 
 @param self A pointer to a tsk_table_collection_t object.
@@ -3470,6 +3479,9 @@ If the file contains multiple table collections, this function will load
 the first. Please see the :c:func:`tsk_table_collection_loadf` for details
 on how to sequentially load table collections from a stream.
 
+If the TSK_LOAD_SKIP_TABLES option is set, only the top-level
+information of the table collection will be read, leaving all tables empty.
+
 **Options**
 
 Options can be specified by providing one or more of the following bitwise
@@ -3477,6 +3489,8 @@ flags:
 
 TSK_NO_INIT
     Do not initialise this :c:type:`tsk_table_collection_t` before loading.
+TSK_LOAD_SKIP_TABLES
+    Skip reading tables, and only load top-level information.
 
 **Examples**
 
@@ -3524,6 +3538,14 @@ different error conditions. Please see the
 :ref:`sec_c_api_examples_file_streaming` section for an example of how to
 sequentially load tree sequences from a stream.
 
+Please note that this streaming behaviour is not supported if the
+TSK_LOAD_SKIP_TABLES option is set. With this option, only the top-level
+information of the table collection will be read, leaving all tables empty. When
+attempting to read from a stream with multiple table collection definitions and
+the TSK_LOAD_SKIP_TABLES option set, only the top-level information of the first
+table collection will be read on the first call to
+:c:func:`tsk_table_collection_loadf`, with subsequent calls leading to errors.
+
 **Options**
 
 Options can be specified by providing one or more of the following bitwise
@@ -3531,6 +3553,8 @@ flags:
 
 TSK_NO_INIT
     Do not initialise this :c:type:`tsk_table_collection_t` before loading.
+TSK_LOAD_SKIP_TABLES
+    Skip reading tables, and only load top-level information.
 
 @endrst
 
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -6887,15 +6887,16 @@ TableCollection_equals(TableCollection *self, PyObject *args, PyObject *kwds)
     int ignore_ts_metadata = false;
     int ignore_provenance = false;
     int ignore_timestamps = true;
+    int ignore_tables = false;
     static char *kwlist[] = { "other", "ignore_metadata", "ignore_ts_metadata",
-        "ignore_provenance", "ignore_timestamps", NULL };
+        "ignore_provenance", "ignore_timestamps", "ignore_tables", NULL };
 
     if (TableCollection_check_state(self)) {
         goto out;
     }
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!|iiii", kwlist, &TableCollectionType,
-            &other, &ignore_metadata, &ignore_ts_metadata, &ignore_provenance,
-            &ignore_timestamps)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!|iiiii", kwlist,
+            &TableCollectionType, &other, &ignore_metadata, &ignore_ts_metadata,
+            &ignore_provenance, &ignore_timestamps, &ignore_tables)) {
         goto out;
     }
     if (ignore_metadata) {
@@ -6909,6 +6910,9 @@ TableCollection_equals(TableCollection *self, PyObject *args, PyObject *kwds)
     }
     if (ignore_timestamps) {
         options |= TSK_CMP_IGNORE_TIMESTAMPS;
+    }
+    if (ignore_tables) {
+        options |= TSK_CMP_IGNORE_TABLES;
     }
     if (TableCollection_check_state(other) != 0) {
         goto out;
@@ -6999,10 +7003,16 @@ TableCollection_load(TableCollection *self, PyObject *args, PyObject *kwds)
     PyObject *ret = NULL;
     PyObject *py_file;
     FILE *file = NULL;
-    static char *kwlist[] = { "file", NULL };
+    tsk_flags_t options = 0;
+    int skip_tables = false;
+    static char *kwlist[] = { "file", "skip_tables", NULL };
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &py_file)) {
+    if (!PyArg_ParseTupleAndKeywords(
+            args, kwds, "O|i", kwlist, &py_file, &skip_tables)) {
         goto out;
+    }
+    if (skip_tables) {
+        options |= TSK_LOAD_SKIP_TABLES;
     }
     file = make_file(py_file, "rb");
     if (file == NULL) {
@@ -7021,7 +7031,7 @@ TableCollection_load(TableCollection *self, PyObject *args, PyObject *kwds)
     if (err != 0) {
         goto out;
     }
-    err = tsk_table_collection_loadf(self->tables, file, 0);
+    err = tsk_table_collection_loadf(self->tables, file, options);
     if (err != 0) {
         handle_library_error(err);
         goto out;
@@ -7383,10 +7393,16 @@ TreeSequence_load(TreeSequence *self, PyObject *args, PyObject *kwds)
     PyObject *ret = NULL;
     PyObject *py_file;
     FILE *file = NULL;
-    static char *kwlist[] = { "file", NULL };
+    tsk_flags_t options = 0;
+    int skip_tables = false;
+    static char *kwlist[] = { "file", "skip_tables", NULL };
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &py_file)) {
+    if (!PyArg_ParseTupleAndKeywords(
+            args, kwds, "O|i", kwlist, &py_file, &skip_tables)) {
         goto out;
+    }
+    if (skip_tables) {
+        options |= TSK_LOAD_SKIP_TABLES;
     }
     file = make_file(py_file, "rb");
     if (file == NULL) {
@@ -7405,7 +7421,7 @@ TreeSequence_load(TreeSequence *self, PyObject *args, PyObject *kwds)
     if (err != 0) {
         goto out;
     }
-    err = tsk_treeseq_loadf(self->tree_sequence, file, 0);
+    err = tsk_treeseq_loadf(self->tree_sequence, file, options);
     if (err != 0) {
         handle_library_error(err);
         goto out;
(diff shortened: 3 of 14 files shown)


clwgg:
Please note that this is very much a draft -- I am of course happy to clean this up and add tests and everything once some of the decisions I outline above have been made.

c/tskit/tables.c
```diff
@@ -10324,39 +10327,41 @@ tsk_table_collection_loadf_inited(tsk_table_collection_t *self, FILE *file)
...
+        if (ret != 0) {
+            goto out;
+        }
+        ret = tsk_provenance_table_load(&self->provenances, &store);
+        if (ret != 0) {
+            goto out;
+        }
     }
-    ret = tsk_table_collection_load_indexes(self, &store);
+    ret = tsk_table_collection_load_indexes(self, &store, options);
```
jeromekelleher, line 10364: Might as well skip skip loading the indexes as well as the tables, since the indexes don't make any sense without the edge table. I would just put this into the same "if" block as the table loads.

python/tskit/tables.py
```diff
@@ -3754,3 +3754,8 @@ def ibd_segments(
             store_pairs=store_pairs,
             store_segments=store_segments,
         )
+
+
+def load_metadata(file):
```
jeromekelleher, line 3759: I don't think there's much point in providing the ``load_metadata`` function to be honest - it's just as easy to call

```python
metadata = tskit.load(path, skip_tables=True).metadata
```

It just doesn't seem worth having the extra function to document since it's already a one-liner and people might want to access other top-level info from the tree sequence too.

jeromekelleher (commented):
This looks great, thanks @clwgg! I think you've done a great job and have basically got it right.

I think it would be better to build an empty index in the ``skip_tables`` case rather than complicating the logic for load_indexes. It'll basically be a no-op, and I think the logic would be simpler.

Does this work for you?

benjeffery:
Great stuff, glad you could get going so quickly. FWIW I agree with @jeromekelleher on both having an empty index and that `load_metadata` isn't needed.

clwgg:
Thanks a lot for the quick replies! I just have two clarifying questions for how to go forward.

**Question 1:**
My motivation for `load_metadata` was that currently I expose the `skip_tables` option only to `TableCollection.load`. This is done completely through the TableCollection backend, so the `TSK_LOAD_SKIP_TABLES` flag never actually touches a tree sequence and exists only in "TableCollection land".

My understanding is that `tskit.load` calls `TreeSequence.load` however, so to make `skip_tables` available here, I would need to propagate it all the way from `TreeSequence.load` -> `_tskit.TreeSequence_load` -> `tsk_treeseq_load` until it hits `tsk_table_collection_load`. This was in fact my original plan (and I had implemented the backend part before reverting it here: https://github.com/tskit-dev/tskit/pull/1882/commits/20fe73f2924c6c30c14411f7040809cfa7adee16), until I realized that I could just expose it through the TableCollection only. I am happy to implement the TreeSequence part if you prefer to expose the option in this way -- I just wanted to double check here.

**Question 2:**
If we do actually make a TreeSequence out of the TableCollection loaded with `TSK_LOAD_SKIP_TABLES` we can't completely skip loading the indexes, since this leads `tsk_table_collection_check_integrity` to complain (which is called from `tsk_treeseq_init` within `tsk_treeseq_load`). Are you suggesting to add a function that adds a "dummy" empty index if `TSK_LOAD_SKIP_TABLES` is set, instead of having `tsk_table_collection_load_indexes` do it?

benjeffery:
Thanks, good questions.

This is an advanced use case, I don't think we need to add it to `TreeSequence` or `tskit.load`.

It is fine for `tsk_table_collection_check_integrity` to complain, there is no use case where you need a `TreeSequence` with just metadata that isn't covered by the `TableCollection` API.

jeromekelleher:
I've had a look at this @clwgg and I think we can make it work all right. The load function looks a bit like this:
```c
        }
        ret = tsk_table_collection_load_indexes(self, &store);
        if (ret != 0) {
            goto out;
        }
    } else {
        ret = tsk_table_collection_build_index(self, 0);
        if (ret != 0) {
            goto out;
        }
    }
```
Then, for testing, we can add something like this function to the ``tests/test_file_format.c`` file:
```c
static void
test_skip_tables(void)
{
    int ret;
    tsk_treeseq_t *ts1 = caterpillar_tree(5, 3, 3);
    tsk_treeseq_t ts2;

    ret = tsk_treeseq_dump(ts1, _tmp_file_name, 0);
    ret = tsk_table_collection_load(&tables, _tmp_file_name, TSK_LOAD_SKIP_TABLES);

    CU_ASSERT_TRUE(
        tsk_table_collection_equals(&tables, ts1->tables, TSK_CMP_IGNORE_TABLES));
    CU_ASSERT_EQUAL(tables.individuals.num_rows, 0);
    CU_ASSERT_EQUAL(tables.nodes.num_rows, 0);
    CU_ASSERT_EQUAL(tables.edges.num_rows, 0);
    CU_ASSERT_EQUAL(tables.migrations.num_rows, 0);
    CU_ASSERT_EQUAL(tables.sites.num_rows, 0);
    CU_ASSERT_EQUAL(tables.mutations.num_rows, 0);
    CU_ASSERT_EQUAL(tables.provenances.num_rows, 0);

    /* We should be able to make a tree sequence */
    ret = tsk_treeseq_init(&ts2, &tables, 0);
    tsk_treeseq_free(&ts2);
    
    /* Do the same thing with treeseq API */
    ret = tsk_treeseq_load(&ts2, _tmp_file_name, TSK_LOAD_SKIP_TABLES);
    CU_ASSERT_TRUE(tsk_table_collection_equals(&tables, ts2.tables, 0));
    tsk_treeseq_free(&ts2);

    tsk_treeseq_free(ts1);
    free(ts1); 
}
```

This uses a new flag ``TSK_CMP_IGNORE_TABLES`` to table_collection_equals, which I think is worth adding, and the compare function would now look like:
```c
bool
tsk_table_collection_equals(const tsk_table_collection_t *self,
    const tsk_table_collection_t *other, tsk_flags_t options)
{
    bool ret = self->sequence_length == other->sequence_length
               && self->time_units_length == other->time_units_length
               && tsk_memcmp(self->time_units, other->time_units,
                      self->time_units_length * sizeof(char))
                      == 0;
    if (!(options & TSK_CMP_IGNORE_TABLES)) {
        ret = ret
              && tsk_individual_table_equals(
                     &self->individuals, &other->individuals, options)
              && tsk_node_table_equals(&self->nodes, &other->nodes, options)
              && tsk_edge_table_equals(&self->edges, &other->edges, options)
              && tsk_migration_table_equals(
                     &self->migrations, &other->migrations, options)
              && tsk_site_table_equals(&self->sites, &other->sites, options)
              && tsk_mutation_table_equals(&self->mutations, &other->mutations, options)
              && tsk_population_table_equals(
                     &self->populations, &other->populations, options);
        /* TSK_CMP_IGNORE_TABLES impliest TSK_CMP_IGNORE_PROVENANCE */
        if (!(options & TSK_CMP_IGNORE_PROVENANCE)) {
            ret = ret
                  && tsk_provenance_table_equals(
                         &self->provenances, &other->provenances, options);
        }
    }
    /* TSK_CMP_IGNORE_TS_METADATA is implied by TSK_CMP_IGNORE_METADATA */
    if (options & TSK_CMP_IGNORE_METADATA) {
        options |= TSK_CMP_IGNORE_TS_METADATA;
    }
    if (!(options & TSK_CMP_IGNORE_TS_METADATA)) {
        ret = ret
              && (self->metadata_length == other->metadata_length
                     && self->metadata_schema_length == other->metadata_schema_length
                     && tsk_memcmp(self->metadata, other->metadata,
                            self->metadata_length * sizeof(char))
                            == 0
                     && tsk_memcmp(self->metadata_schema, other->metadata_schema,
                            self->metadata_schema_length * sizeof(char))
                            == 0);
    }
    return ret;
}
```

Note the tests don't *quite* pass with these changes - there's one more update that needs to be made (which I'll leave you to figure, if you like!)

clwgg:
Thanks for the great pointers @jeromekelleher! I had missed the `tsk_table_collection_build_index` function on my first pass, but of course it makes a lot more sense to use that one!

I just reverted a bunch of my changes and implemented your suggestions. The C tests you suggested are passing with this version. Should I had python tests for this as well, at least to test if the flag to `TableCollection.load` works?

Also, are you ok with the `TSK_LOAD_SKIP_TABLES` flag implementation? Or should we leave more space for other `table collection init` flags?

clwgg:
I just implemented your suggestions and added the tests you wrote. Given that you included `tsk_treeseq_load(&ts2, _tmp_file_name, TSK_LOAD_SKIP_TABLES);` in the test suite, I also re-added exposing the flag for tree sequence loading here: https://github.com/tskit-dev/tskit/pull/1882/commits/65823c6080b8f3ca4db06146c8ffdb6580b17a77. Does this mean I should also expose it to python `TreeSequence.load`? Sorry that I keep asking this, just trying to make sure since I'm getting some "mixed signals ;)".

clwgg:
This is a random side note, but I can't figure out how to make clang-format use the indentation style that the test suite is complaining about above when mixing multi-line parentheses and lines starting with logical operators. The `.clang-format` file in the repo doesn't reproduce it for me, and neither do a bunch of other config options I've tried. I'm on clang version 12.0.1 -- please let me know if anyone has an idea (:

clwgg:
> Does this mean I should also expose it to python `TreeSequence.load`? 

Ok, I just added this for completeness' sake -- please let me know if you would like to have this kept in here, and I can start adding python tests for whatever aspects of this we'd like to expose to python.

c/tskit/tables.h
```diff
@@ -712,6 +712,13 @@ typedef struct {
 /* Flags for table collection init */
 #define TSK_NO_EDGE_METADATA (1 << 0)
 
+/* Flags for table collection load */
+/* This shares an interface with table collection init, so we have to
+ * increment offset by one.
+ * NOTE: Should these two be under the same header?
```
jeromekelleher, line 718: Good questions - let's do this as part of #1720

Can you leave a comment here please, "TODO review as part of https://github.com/tskit-dev/tskit/issues/1720"

python/_tskitmodule.c
```diff
@@ -6982,11 +6982,17 @@ TableCollection_load(TableCollection *self, PyObject *args, PyObject *kwds)
...
+    int skip_tables = false;
+    static char *kwlist[] = { "file", "skip_tables", NULL };
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &py_file)) {
+    if (!PyArg_ParseTupleAndKeywords(
+            args, kwds, "O|i", kwlist, &py_file, &skip_tables)) {
         goto out;
     }
+    if (skip_tables) {
+        options |= TSK_LOAD_SKIP_TABLES;
```
jeromekelleher, line 6994: We need to add a test to the ``tests/test_lowlevel.py`` to cover this. You could add a test like ``load_skip_tables`` [here](https://github.com/tskit-dev/tskit/blob/ab99547d024c737698ec8ce6f06a2d8d56e52302/python/tests/test_lowlevel.py#L166)  that does some very simple tests on the ``skip_tables`` argument (do we get a TypeError on bad types) and just checks that it runs ok with ``skip_tables=True``. (No need to test any properties of the ts, we'll do that elsewhere)

Will need to do the same thing for the TreeSequence.load method. This can be a copy-paste basically into the appropriate test class.

jeromekelleher (commented):
Looks just right to me, thanks @clwgg! Some comments about testing above.

To test the overall functionality in Python, I think the best thing to do is add a new test class ``TestSkipTables`` to the ``tests/test_file_format.py`` file. We don't have to do anything exhaustive, just check the basic properties of the tables/tree sequences we're loading on a few examples.

This would be simplest if we added the ``ignore_tables`` argument to assert_equals, but maybe this PR is getting big enough as it is and this should be done in a follow up. It's up to you.

jeromekelleher:
> This is a random side note, but I can't figure out how to make clang-format use the indentation style that the test suite is complaining about when mixing multi-line parentheses and lines starting with logical operators. The .clang-format file in the repo doesn't reproduce it for me, and neither do a bunch of other config options I've tried. I'm on clang version 12.0.1 -- please let me know if anyone has an idea (:

This is super annoying @clwgg but you have to use the exact version of clang-format :frowning_face:  The [docs](https://tskit.dev/tskit/docs/stable/development.html#sec-development-c-requirements) should help.

jeromekelleher:
So yes, just to be clear on your question above, I think we should support this on both the TableCollection.load and TreeSequence.load code paths.

jeromekelleher:
Ah, I realised there's a wrinkle here that I forgot about. We need to change the load flags to kastore to tell it READ_ALL only if we're not in the "skip tables" case. Otherwise, we just read the whole file anyway. This is easily done:

```c
 int kas_flags = options & TSK_LOAD_SKIP_TABLES? 0: KAS_READ_ALL;

 ret = kastore_openf(&store, file, "r", kas_flags);
```

I started writing a few tests to check if this was working:
```python
def assert_tables_empty(tables):
    for table in tables.name_map.values():
        assert len(table) == 0


class TestSkipTables:
    def test_ts_read_path_interface(self, tmp_path, ts_fixture):
        # Check the fixture has metadata and a schema
        assert ts_fixture.metadata_schema is not None
        assert len(ts_fixture.metadata) > 0
        save_path = tmp_path / "tmp.trees"
        ts_fixture.dump(save_path)
        ts_no_tables = tskit.load(save_path, skip_tables=True)
        assert ts_no_tables.metadata == ts_fixture.metadata
        assert ts_no_tables.sequence_length == ts_fixture.sequence_length
        assert_tables_empty(ts_no_tables.tables)

    def test_ts_read_one_stream(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        ts_fixture.dump(save_path)
        with open(save_path, "rb") as f:
            ts_no_tables = tskit.load(save_path, skip_tables=True)
            assert ts_no_tables.metadata == ts_fixture.metadata
            assert ts_no_tables.sequence_length == ts_fixture.sequence_length
            assert_tables_empty(ts_no_tables.tables)

    def test_ts_read_multi_stream_fails(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        ts_fixture.dump(save_path)
        with open(save_path, "rb") as f:
            ts_no_tables = tskit.load(f, skip_tables=True)
            # The second load should fail because we've not read to the
            # end of the file.
            with pytest.raises(tskit.FileFormatError):
                tskit.load(f)
```

We should probably document this corner case (you can't read multiple tree sequences from a stream in ``skip_tables`` mode).

clwgg:
Great catch on the `KAS_READ_ALL` flag! I added that and am looking into the python tests now. Quick note on the ones you drafted (thanks for that!): the last one `test_ts_read_multi_stream_fails()` actually throws an `EOFError` for me, instead of a `tskit.FileFormatError` (I assume from [here](https://github.com/tskit-dev/tskit/blob/ab99547d024c737698ec8ce6f06a2d8d56e52302/python/_tskitmodule.c#L269)?). Is that the right one to include for the test? Or does the error code need changing?

clwgg:
ok, while working on the tests it did indeed seem like a good idea to expose the `TSK_CMP_IGNORE_TABLES` flag we introduced as `ignore_tables` to python. I tried to hunt down all the places where this would need adding and hope I caught them all.

clwgg:
Sorry for the mess with that last commit, but I was looking for places where the `ignore_*` flags to the `.equals()` methods get tested and just added a couple of in-line comments to not loose track of them -- any input on the notes in that commit (https://github.com/tskit-dev/tskit/pull/1882/commits/87974d84ab1c42978cde7c8b146948607de1c7c4) would of course be very much appreciated!

clwgg:
> some very simple tests on the skip_tables argument (do we get a TypeError on bad types) and just checks that it runs ok with skip_tables=True. (No need to test any properties of the ts, we'll do that elsewhere)

Sorry @jeromekelleher I just re-read this after I had implemented some of those tests already. The last commit (https://github.com/tskit-dev/tskit/pull/1882/commits/cd6775b34196e7ffe1511b95924a32c65e43f9f9) now also added a `bad_bool` test that I think is closer to what you were envisioning. I'm happy to pull the `tables_are_skipped` stuff out of the lowlevel tests if you think this would be better tested in `test_file_format.py`.

python/tests/test_lowlevel.py
```diff
@@ -164,6 +164,36 @@ class TestTableCollection(LowLevelTestCase):
     Tests for the low-level TableCollection class
     """
 
+    def test_skip_tables(self, tmp_path):
+        def tables_are_skipped(tc, tc_skip):
+            return (
```
jeromekelleher, line 169: It would be a little better to structure this like
```def assert_tables_are_skipped(tc, tc_skip):
          assert not tc.equals(tc_skip)
          assert tc.equals(tc_skip, ignore_tables=True) 
           ... etc
```
The reasons is that if the tests every fail, the specific problem will be immediately apparent. This way, you'd have to put in some prints to try and figure out which one of the assertions was failing.

python/tests/test_tables.py
```diff
@@ -4003,6 +4003,8 @@ def test_ignore_timestamps(self, t1, t2):
         t1.assert_equals(t2, ignore_provenance=True)
         t1.assert_equals(t2, ignore_timestamps=True)
 
+    # NOTE: test `ignore_tables` here too?
```
jeromekelleher, line 4006: Yes please!

python/tskit/tables.py
```diff
@@ -2850,6 +2856,10 @@ def assert_equals(
                 f" differs: self={self.sequence_length} other={other.sequence_length}"
             )
 
+        # NOTE: Do we need to anything "extra" to account for `ignore_tables`?
```
jeromekelleher, line 2859: I think this is fine - we don't keep any particular state to say that the tables were created by "ignore_tables" so failing on the number of edges being different is as good as any.

jeromekelleher (commented):
I think this looks great, thanks @clwgg ! I think you just need to fill out the few remaining testing corner cases, add some documentation and we're done.

Thanks for following up with the ``ignore_tables`` functionality here, you're going above and beyond!

jeromekelleher:
> I'm happy to pull the tables_are_skipped stuff out of the lowlevel tests if you think this would be better tested in test_file_format.py.

I think that's probably better all right. The low-level tests are really just to make sure we're not making any basic mistakes at the Python-C level and we're exercising all the code/error paths. Let's to the logic testing in ``test_file_formats`` like I suggested above. We can test ``ignore_tables`` in test_tables.py where you suggested.

clwgg:
Hi @jeromekelleher! I've been working on adding more tests, and I might be getting a little too much into the weeds of this... One thing I was wondering about the `multi_stream` test you proposed above, is what the "base case" we're testing against is (i.e. without the flag we're testing for), but I'm not sure if that's the right way of thinking about it. Here are a few examples:

**A** - The test case you proposed
``` python
    def test_ts_read_multi_stream_fails(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        ts_fixture.dump(save_path)
        with open(save_path, "rb") as f:
            ts_no_tables = tskit.load(f, skip_tables=True)
            # The second load should fail because we've not read to the
            # end of the file.
            with pytest.raises(exceptions.FileFormatError):
                tskit.load(f)

```
My understanding is that this throws a `FileFormatError` because we're trying to read a new tree sequence from a file handle that is now positioned somewhere inside of the file, so in this case doing something like `f.seek(0)` between the calls to `tskit.load()` would work to resolve the error.
If we don't set `skip_tables` on the first load, the equivalent test throws an `EOFError` instead, since we're now at the end of the file after the first read:

**B**
``` python
    def test_if_this_works(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        ts_fixture.dump(save_path)
        with open(save_path, "rb") as f:
            ts1 = tskit.load(f)
            with pytest.raises(EOFError):
                ts2 = tskit.load(f)
```

So really what **A** is testing is whether we throw a `FileFormatError` instead of an `EOFError` by setting the `skip_tables` flag, correct?
With that, I was wondering if you think it would be worth it to test the "multistream" behavior more explicitly? One way to do that would be something like:

**C**
``` python
    def test_double_file(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        with open(save_path, "wb") as f:
            ts_fixture.dump(f)
            ts_fixture.dump(f)
        with open(save_path, "rb") as f:
            ts1 = tskit.load(f)
            ts2 = tskit.load(f)
        assert ts_fixture.equals(ts1)
        assert ts_fixture.equals(ts2)

    def test_double_file_fails(self, tmp_path, ts_fixture):
        save_path = tmp_path / "tmp.trees"
        with open(save_path, "wb") as f:
            ts_fixture.dump(f)
            ts_fixture.dump(f)
        with open(save_path, "rb") as f:
            ts1 = tskit.load(f, skip_tables=True)
            with pytest.raises(exceptions.FileFormatError):
                ts2 = tskit.load(f)
```
Here, we use the concatenation of two tree sequence dumps in a single file to simulate reading from a non-seekable stream, which works fine without `skip_tables`, but throws the appropriate error when `skip_tables` is true. I think the concatenation into a single file looks a little ugly, but I thought it was easier to simulate it this way rather than having to set up sockets or something.

Anyway, I'm curious to hear what you think about this, and I'll try to get less distracted while adding the remaining tests...

jeromekelleher:
Yes, you've got it right @clwgg. I think your proposal of C is good. We are testing read-multiple-from-stream behaviour pretty thoroughly in [test_fileobj.py](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_fileobj.py) but having this test nearby would help to make the behaviour on ``skip_tables`` clearer.

The base-case here is that the ``test_double_file_fails`` test will fail if we don't provide the correct flags to kastore.

jeromekelleher:
We're hoping to get 0.4.0 shipped in the coming week @clwgg - do you think this can be finished up by then?

clwgg:
great! Yes, I think so.

clwgg:
Alright, I added more tests and API documentation, so I think this is pretty much ready for final(ish) review! There is one new issue that CI is complaining about:

- The python test case `test_ts_twofile_stream_fails()`, which I outlined in my "Option C" above as `test_double_file_fails`, is failing. Unfortunately I haven't been able to reproduce that locally, and I'm not totally sure what is going wrong.

Apart from that I am of course happy to make any other changes based on your review, especially of these new additions!

c/tskit/tables.h
```diff
@@ -3512,13 +3526,23 @@ different error conditions. Please see the
 :ref:`sec_c_api_examples_file_streaming` section for an example of how to
 sequentially load tree sequences from a stream.
 
+Please note that this streaming behaviour is disabled if the
+TSK_LOAD_SKIP_TABLES option is set. With this option, only the top-level
+metadata of the table collection will be read, leaving all tables empty. When
+attempting to read from a stream with multiple table collection definitions and
+the TSK_LOAD_SKIP_TABLES option set, only the top-level metadata of the first
```
jeromekelleher, line 3533:
```suggestion
the TSK_LOAD_SKIP_TABLES option set, only the top-level information of the first
```

c/tskit/tables.h
```diff
@@ -3512,13 +3526,23 @@ different error conditions. Please see the
...
+
 **Options**
 
 Options can be specified by providing one or more of the following bitwise
 flags:
 
 TSK_NO_INIT
     Do not initialise this :c:type:`tsk_table_collection_t` before loading.
+TSK_LOAD_SKIP_TABLES
+    Skip reading tables, and only load top-level metadata.
```
jeromekelleher, line 3545:
```suggestion
    Skip reading tables, and only load top-level information.
```

python/tskit/tables.py
```diff
@@ -2772,6 +2773,8 @@ def equals(
         :param bool ignore_timestamps: If True the provenance timestamp column
             is ignored in the comparison. If ``ignore_provenance`` is True, this
             parameter has no effect.
+        :param bool ignore_tables: If True no tables are included in the
+            comparison, thus comparing only the top-level metadata.
```
jeromekelleher, line 2777:
```suggestion
            comparison, thus comparing only the top-level information.
```

python/tskit/tables.py
```diff
@@ -2813,6 +2818,8 @@ def assert_equals(
...
```
jeromekelleher, line 2822:
```suggestion
            comparison, thus comparing only the top-level information.
```

python/tskit/tables.py
```diff
@@ -2873,10 +2881,10 @@ def __getstate__(self):
         return self.asdict()
 
     @classmethod
-    def load(cls, file_or_path):
+    def load(cls, file_or_path, skip_tables=False):
         file, local_file = util.convert_file_like_to_open_file(file_or_path, "rb")
         ll_tc = _tskit.TableCollection(1)
-        ll_tc.load(file)
+        ll_tc.load(file, skip_tables=bool(skip_tables))
```
jeromekelleher, line 2887:
```suggestion
        ll_tc.load(file, skip_tables=skip_tables)
```

python/tskit/tables.py
```diff
@@ -2873,10 +2881,10 @@ def __getstate__(self):
         return self.asdict()
 
     @classmethod
-    def load(cls, file_or_path):
+    def load(cls, file_or_path, skip_tables=False):
```
jeromekelleher, line 2884:
```suggestion
    def load(cls, file_or_path, *, skip_tables=False):
```
Safer make this a keyword only arg

python/tskit/tables.py
jeromekelleher: Casting to bool here could have unintended consequences, as we might want to make ``skip_tables=None`` the default later (and infer a sensible default based on other param values). If users want to use Falsey behaviour they can cast to bool easily enough themselves.

python/tskit/trees.py
```diff
@@ -2845,19 +2845,24 @@ def generate_random_binary(
         )
 
 
-def load(file):
+def load(file, skip_tables=False):
```
jeromekelleher, line 2848:
```suggestion
def load(file, *, skip_tables=False):
```

python/tskit/trees.py
```diff
@@ -2845,19 +2845,24 @@ def generate_random_binary(
...
     """
     Loads a tree sequence from the specified file object or path. The file must be in the
     :ref:`tree sequence file format <sec_tree_sequence_file_format>` produced by the
     :meth:`TreeSequence.dump` method.
 
     :param str file: The file object or path of the ``.trees`` file containing the
         tree sequence we wish to load.
+    :param bool skip_tables: If True, no tables are read from the ``.trees``
+        file and only the top-level metadata is loaded as a tree sequence object.
```
jeromekelleher, line 2857:
```suggestion
        file and only the top-level information is populated in the tree sequence object.
```

python/tskit/trees.py
```diff
@@ -3503,11 +3510,11 @@ def aslist(self, **kwargs):
         return [tree.copy() for tree in self.trees(**kwargs)]
 
     @classmethod
-    def load(cls, file_or_path):
+    def load(cls, file_or_path, skip_tables=False):
```
jeromekelleher, line 3513:
```suggestion
    def load(cls, file_or_path,  *, skip_tables=False):
```

python/tskit/trees.py
```diff
@@ -3503,11 +3510,11 @@ def aslist(self, **kwargs):
...
         file, local_file = util.convert_file_like_to_open_file(file_or_path, "rb")
         try:
             ts = _tskit.TreeSequence()
-            ts.load(file)
+            ts.load(file, skip_tables=bool(skip_tables))
```
jeromekelleher, line 3517:
```suggestion
            ts.load(file, skip_tables=skip_tables)
```

python/tests/test_file_format.py
```diff
@@ -1075,3 +1075,75 @@ def test_load_bad_formats(self):
...
+
+    def test_ts_twofile_stream_fails(self, tmp_path, ts_fixture):
+        # We can't skip_tables while reading from a stream
+        save_path = tmp_path / "tmp.trees"
+        with open(save_path, "wb") as f:
+            ts_fixture.dump(f)
+            ts_fixture.dump(f)
+        with open(save_path, "rb") as f:
+            tskit.load(f, skip_tables=True)
+            with pytest.raises(exceptions.FileFormatError):
```
jeromekelleher, line 1129: I'm not sure why this isn't working - it should (and does on my machine). I think it might be something to do with the complicated exception chaining coming out of  #1938 and Python 3.7. This code is correct anyway, we'll make whatever changes needed to fix it.

jeromekelleher (commented):
Looks good to me @clwgg! I've made a batch of suggestions.

Once these are in, can you [rebase and squash](https://popsim-consortium.github.io/stdpopsim-docs/latest/development.html#rebasing) the commits please? I think it's ready to merge then.

clwgg:
done!

clwgg:
oh, I just noticed that I can reproduce the failing test locally now that I rebased onto current main!

jeromekelleher:
> oh, I just noticed that I can reproduce the failing test locally now that I rebased onto the current upstream main!

Ah, that's good. What version of Python are you using?

clwgg:
To me it does look like the issue you first reported in https://github.com/tskit-dev/tskit/issues/1938 -- I get a `TypeError: expected str, bytes or os.PathLike object, not BufferedReader` from the `open()` call in `raise_hdf5_format_error`, which is triggered by the `try: except: FileFormatError` in `TreeSequence.load()`

clwgg:
I'm on `python 3.9.7`.

clwgg:
it looks like exception chaining is only triggered on an `OSError` in `raise_hdf5_format_error`, while the `TypeError` actually does throw an exception, right? I don't know, I might be totally off here.

jeromekelleher:
Aha, that's exactly it @clwgg ``TypeError`` is not an ``OSError``. A quick fix would be to add ``TypeError`` to the expected exceptions.

jeromekelleher:
This needs another rerbase @clwgg, but should be good to merge then.

clwgg:
done!

jeromekelleher (approved):
Fantastic, thanks @clwgg! We got there :smile:

I updated the commit to include some changelog information; merging.

PR #1883, opened 2021-11-03
Don't import HDF5 in error state

### LINKED ISSUE #1858: File load error path imports h5py
The error path for loading tskit files ends up loading h5py [here](https://github.com/tskit-dev/tskit/blob/c9f3dbe08b25196daacb718eb60290523543bf2c/python/tskit/trees.py#L3520) which is leads to obscure downstream problems. It's definitely a bug that our error path involves importing such a complex and problematic third party dependency that isn't one of our registered dependencies, so we should do something different here.

benjeffery:
Fixes #1858


jeromekelleher (approved):
Great!

Issue #1817, opened 2021-10-18
Handling of node labels should be consistent across Phylo export formats


jeromekelleher:
As discussed in https://github.com/tskit-dev/tskit/issues/1803#issuecomment-945964208 it's probably a good idea if we establish the convention that nodes are labelled with a string like ``tsk{node.id}``, and this should be the consistent default across phylo export, i.e., in 

- TreeSequence.to_nexus (see #1803)
- Tree.to_newick (see #1671)
- TreeSequence.to_fasta/write_fasta (#353)

jeromekelleher:
Closing this, resolved.

PR #1889, opened 2021-11-05
Update fasta code

jeromekelleher:
Update the FASTA output to use the new alignments method.

Stacked on #1825

### DIFF
--- a/python/tests/test_fasta.py
+++ b/python/tests/test_fasta.py
@@ -23,26 +23,89 @@
 """
 Test cases for fasta output in tskit.
 """
+import functools
 import io
-import itertools
-import os
-import tempfile
 
+import dendropy
 import msprime
+import numpy as np
 import pytest
 from Bio import SeqIO
 
-from tests import tsutil
+import tests
+import tskit
 
 
 # setting up some basic haplotype data for tests
-def create_data(length):
-    ts = msprime.simulate(
-        sample_size=10, length=length, mutation_rate=1e-2, random_seed=123
+@functools.lru_cache(maxsize=100)
+def create_data(sequence_length):
+    ts = msprime.sim_ancestry(
+        samples=5, sequence_length=sequence_length, random_seed=123
     )
-    ts = tsutil.jukes_cantor(ts, length, 1.0, seed=123)
-    assert ts.num_sites == length
+    ts = msprime.sim_mutations(ts, rate=0.1, random_seed=1234)
+    assert ts.num_sites > 5
     return ts
+
+
+@tests.cached_example
+def missing_data_example():
+    # 2.00┊   4     ┊
+    #     ┊ ┏━┻┓    ┊
+    # 1.00┊ ┃  3    ┊
+    #     ┊ ┃ ┏┻┓   ┊
+    # 0.00┊ 0 1 2 5 ┊
+    #     0        10
+    #      |      |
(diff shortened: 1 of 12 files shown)


jeromekelleher:
Ready for review - can you have a look please @hyanwong, most of this is your code originally I think.

jeromekelleher:
Hmmm, this is complicated by #1893.

python/tests/test_fasta.py
```diff
@@ -23,28 +23,91 @@
...
-    ts = msprime.simulate(
-        sample_size=10, length=length, mutation_rate=1e-2, random_seed=123
+@functools.lru_cache(maxsize=100)
+def create_data(sequence_length):
+    ts = msprime.sim_ancestry(
+        samples=5, sequence_length=sequence_length, random_seed=123
     )
-    ts = tsutil.jukes_cantor(ts, length, 1.0, seed=123)
-    assert ts.num_sites == length
+    ts = msprime.sim_mutations(ts, rate=0.1, random_seed=1234)
```
hyanwong, line 45: Nice to switch to sim_mutations now: 👍

hyanwong:
> Ready for review - can you have a look please @hyanwong, most of this is your code originally I think.

Pending #1893 this LGTM. Two things that I think people will want to be able to do are:

1. Only output a few of the haplotypes, or a short section of the sequence. This can be done, of course, using `.simplify()` in the first instance and `.keep_intervals` in the second, but it would be good to show people how to do this, either in a tutorial and/or mentioning it in the docstring. There is a (weak) argument for a parameter that says which samples to output, as `simplify()` will change the node numbers, meaning the labels might not be what is expected. I imagine people might do e.g. `my_ts.simplify([4, 5, 6]).as_fasta()` and expect the labels to be [`n4`, `n5`, `n6`] rather than [`n0`, `n1`, `n2`]. Something like `my_ts.simplify().as_fasta(samples=[4, 5, 6]) would allow us to output a string with [`n4`, `n5`, `n6`] as labels. But it may not be worth adding to the complexity of the  as_fasta/write_fasta functions, if we can just warn people to be careful
2. Have a human-readable label for the FASTA strings (was previously provided by the `sequence_ids` parameter). This would reduce the problems in 1., as I assume most people would label using metadata.

One solution to 2., which I don't think this should be part of this PR, but may be worth considering, is to have a general labelling class that could also be used in other output routines, such as `draw_svg`. E.g. one or more of the following:

```
# as a function
ts.write_fasta(node_labels=NodeLabels(lambda n: n.metadata['name'])
# or as a mapping
ts.write_fasta(node_labels=NodeLabels({n.id: n.metadata['name'] for n in ts.nodes()})
# or as a simple array
ts.write_fasta(node_labels=NodeLabels([n.metadata['name'] for n in ts.nodes()])
```

jeromekelleher:
Thanks @hyanwong.

1. We should definitely do this at some point, but the functionality needs to be propagated back through ``.alignments`` etc first, and that's some work.
2. Adding custom labels like we have for ``as_newick`` would definitely be useful, but again the ability to pick specific samples for alignments would be necessary to make this work in general. For now, it's more important that the default node labelling is simple and consistent across various export functions (#1817).

jeromekelleher:
I'm going to close this in favour of #1894, as it'll be easier to review the fasta code as "new" when fixing #1893

PR #1894, opened 2021-11-07
Finalise alignment export

### LINKED ISSUE #1893: Default missing data character should not be "-"
In #436 we update the ``haplotypes`` method to include support for missing data, and added the ``missing_data_char`` parameter. This defaults to a '"-"'. In retrospect I think this is an error and we should change it for 0.4.0.

The "-" character in an alignment is nearly always interpreted as a **gap** (see these [FASTA format page](https://en.wikipedia.org/wiki/FASTA_format) or these [nexus examples](http://wiki.christophchamp.com/index.php?title=NEXUS_file_format), i.e., indicating that there are some indels. This is not what we are indicating in the haplotypes. We are indicating the presence of missing data, i.e., that we don't know what the state of a particular site for a particular sample is. The most common convention for this seems to be using the "N" character.

The DNA alphabet in [dendropy](https://github.com/jeetsukumaran/DendroPy/blob/29fd294bf05d890ebf6a8d576c501e471db27ca1/src/dendropy/datamodel/charstatemodel.py) follows this convention, so unless there's a compelling reason not to I think we should change the default missing data character to "N".

@hyanwong, do you remember the original thinking for using "-" here rather than "N"? Any objections to us changing this?

jeromekelleher:
Add support for outputting alignments in FASTA and nexus formats, and fixes #1893.


jeromekelleher:
This is ready for a look, can you have a look please @hyanwong @benjeffery?

jeromekelleher:
Bumping this - this is a major new piece of functionality so it would be good to get some eyes on it.

python/tskit/trees.py
```diff
@@ -4901,10 +4883,16 @@ def alignments(
...
             a[:] = np.frombuffer(ref_bytes, dtype=np.int8)
-        H = self._haplotypes_array(
-            isolated_as_missing=isolated_as_missing,
-            missing_data_character=missing_data_character,
-        )
+
+        # By rights we should detect this in the variants iterator in C,
+        # but in reality we'll not run this on any large tree sequences
+        # anyway, so lets not worry about the perf.
+        if any(tree.num_roots > 1 for tree in self.trees()):
```
hyanwong, line 4890: Is this the right way to test for missing data? This will fail e.g. on forward-simulated tree sequences, even though we might know the sequence. It will also fail if a sample is isolated but has a mutation above it: in this case the data is not usually thought of as missing.

python/tskit/trees.py
```diff
@@ -4842,43 +4837,30 @@ def alignments(
...
-            be used to represent missing data.
-            If any normal allele contains this character, an error is raised.
-            Default: '-'.
+        :param str reference_sequence: The reference sequence to fill in
+            gaps between sites in the alignments.
         :rtype: collections.abc.Iterable
         :raises: ValueError
-            if all coordinates in this tree sequence are not discrete, or
-            if the ``missing_data_character`` exists in one of the alleles,
+            if all coordinates in this tree sequence are not discrete,
```
hyanwong, line 4856: perhaps "if any site in this tree sequence is not at an integer position"?

python/tskit/trees.py
```diff
@@ -4764,7 +4765,7 @@ def genotype_matrix(
 
         :param bool isolated_as_missing: If True, the allele assigned to
             missing samples (i.e., isolated samples without mutations) is
-            the ``missing_data_character``. If False, missing samples will be
+            :data:`.MISSING_DATA` (-1). If False, missing samples will be
```
hyanwong, line 4768: Is this right? Do we assign the *allele* (which is a string) to be "-1"? Surely not: it's the *genotype* that is -1, and the allele is some other thing, perhaps "N"?

python/tskit/trees.py
```diff
@@ -4842,43 +4837,30 @@ def alignments(
...
-            if the ``missing_data_character`` exists in one of the alleles,
+            if all coordinates in this tree sequence are not discrete,
             or if the ``reference_sequence`` is not of the correct length.
-        :raises: TypeError if the ``missing_data_character`` or any of the alleles
-            at a site are not a single ascii character.
+        :raises: TypeError if any of the alleles at a site are not a
+            single ascii character.
         """
         if not self.discrete_genome:
             raise ValueError("sequence alignments only defined for discrete genomes")
```
hyanwong, line 4862: NB: in the future, it would be reasonable, I think, to output alignment for non-discrete genomes by filling the spaces between each sit with a "gap" character.

hyanwong (commented):
This all looks fine to me, modulo some minor comments (especially about how we detect missing data).

One more general point is that we should perhaps think about how in the future we would deal with indels, so that we don't paint ourselves into a corner now. However , it's hard to imagine how to align sequences with indels, unless the reference allele is the longest one (e.g. ATC) and any mutations are the same length or shorter. There would also be a weird interaction between e.g. a site at position 10 with ancestral_state "ATC" and a site at position 11 with ancestral state "G". Nevertheless, will *will* have to deal with indels at some point, so it might be worth thinking about how other libraries do it.

hyanwong:
Note that I think this also closes #353 doesn't it? Unless we want to leave an issue open to deal with indels?

jeromekelleher:
> Note that I think this also closes #353 doesn't it? Unless we want to leave an issue open to deal with indels?

Yep - it's in the commit messages

python/tskit/trees.py
benjeffery: This is for `genotype_matrix(` though right?

benjeffery (approved):
LGTM, % @hyanwong's questions.

python/tskit/trees.py
hyanwong: Ah, good point. Then the word "allele" is a bit confusing, I think? Perhaps "If True, the genotype value assigned to missing samples (i.e., isolated samples without mutations) is :data:`.MISSING_DATA` (-1). If False, missing samples will be assigned the genotype value corresponding to the ancestral state." (I don't think we require the ancestral state to be always index 0 any more, do we?)

python/tskit/trees.py
jeromekelleher: It's more general than that - ``discrete_genome`` means that *all* coordinates are discrete, including the underlying trees, migrations, everything. There's a good reason for making it this strict - missing data is well defined only when we have integer coordinates for the edges, so we know exactly that from base x to y the sample is isolated

python/tskit/trees.py
jeromekelleher: I disagree - see point about missing data and discrete coordinates above. There's no point in complicating this any further by bringing in continuous coordinates, it's already going to be very complicated once we've dealt with indels etc. Continuous coordinates are a pretty niche requirement in reality anyway, people who want them are unlikely to also want alignments (and if they do, well, they can figure out what the semantics should be).

python/tskit/trees.py
jeromekelleher: Yes, you're right, the forward simulation case would be incorrectly identified. I've pushed an update with as good a workaround as I can come up with, modulo actually sitting down and implementing the full thing in C.

python/tskit/trees.py
jeromekelleher: Yes, I spotted this error as I was going through. I've updated the text a bit (and also copied to variants, where the same text was used)

jeromekelleher:
Thanks for the comments @hyanwong - I've made another pass. We're not going to get this perfect for missing data on this pass, but it'll be good enough to cover 99% of use cases. I'm fairly sure it's currently conservative, so that we may throw an error where not strictly necessary, but we'll never output something that's actually wrong.

jeromekelleher:
Re indels, I agree we'll have to tackle this at some point. The only thing we'll need to worry about here I think in terms of forward compatibility is whether we insist that the reference sequence is equal to the sequence length when storing the reference data (#146)

python/tskit/trees.py
hyanwong: OK - all I really meant was that it should be "any" not "all". So perhaps just "if any coordinate in this tree sequence is not discrete"?

hyanwong:
> Re indels, I agree we'll have to tackle this at some point. The only thing we'll need to worry about here I think in terms of forward compatibility is whether we insist that the reference sequence is equal to the sequence length when storing the reference data (#146). I don't think it affects the code here.

Right. I was wondering how people store multiple sequences in a FASTA file if there are indels? Do they just allow each sequence to be a different length, and to hell with the alignment & reference sequence? Or do they pad out all the deletions with "-", and potentially have more characters in each sequence than in the reference? It might be helpful to talk to someone who knows about this sort of thing, although I agree it doesn't seem like it should affect this PR.

hyanwong (approved):

python/tskit/trees.py
jeromekelleher: Ah yes, good point!

jeromekelleher:
OK, merging this! Now the big question is whether we go ahead and release 0.4.0 like this or if we implement reference sequences first...

PR #1904, opened 2021-11-10
Rename ibd_segment to identity_segment

### LINKED ISSUE #1857: Rename IBDSegments etc to IdentitySegments
The structures we're using to return IBD segments could be used in future to return IBS (identity-by-state) segments or combined IBD+IBS segments. To avoid future confusion, rename the structures to IdentitySegments, tsk_identity_segments_t etc.

In any case it's a better name, since "IBD" may not be immediately obvious to code readers.

jeromekelleher:
Closes #1857

Straightforward renaming, should be uncontroversial

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -5335,9 +5335,9 @@ test_link_ancestors_multiple_to_single_tree(void)
 }
 
 static void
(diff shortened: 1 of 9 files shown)


jeromekelleher:
This is a straightforward rename that people have already agreed to so I'm going to merge.

PR #1910, opened 2021-11-12
load_text only creates populations if none given

### LINKED ISSUE #1909: tskit.load_text creates an extra population
When I provide metadata for a single population in text format to `tskit.load_text`, it creates an additional population id 0 with no metadata:

```
import io
import tskit

nodes = """\
is_sample   time    population
1           0       0
1           0       0
0           1.0     -1
"""
edges = """\
left    right   parent  child
0       10      2       0
0       10      2       1
"""
populations = """\
metadata
some_metadata
"""
ts = tskit.load_text(
    nodes=io.StringIO(nodes),
    edges=io.StringIO(edges),
    populations=io.StringIO(populations),
    strict=False,
    base64_metadata=False,
)

print(ts.num_populations)
print(ts.tables.populations)
```

gives

```
2
╔══╤════════════════╗
║id│metadata        ║
╠══╪════════════════╣
║0 │             b''║
║1 │b'some_metadata'║
╚══╧════════════════╝
```

hyanwong:
Fixes #1909

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -112,6 +112,9 @@
   (:user:`jeetsukumaran`, :user:`jeromekelleher`, :issue:`1785`, :pr:`1835`,
   :pr:`1836`, :pr:`1838`)
 
+- `load_text` created additional populations even if the population table was specified,
+  and didn't strip newlines from input text (:user:`hyanwong`, :issue:`1909`, :pr:`1910`)
+
 
 --------------------
 [0.3.7] - 2021-07-08
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2736,6 +2736,7 @@ class TestTreeSequenceTextIO(HighLevelTestCase):
         assert ts1.num_edges == ts2.num_edges
         assert ts1.num_sites == ts2.num_sites
         assert ts1.num_mutations == ts2.num_mutations
+        assert ts1.num_populations == ts2.num_populations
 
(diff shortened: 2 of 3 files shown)


jeromekelleher:
Looks good. It would be nice to add a couple of minimal tests like (paraphrased)
```python
def test_parse_text_no_populations(self):
    nodes = """\
       is_sample  time  population
        1                0      2
     """
     tables = parse_text(nodes)
     assert len(tables.nodes) == 1
     assert len(tables.populations) == 3

def test_parse_text_populations(self):
       populations = """\
       metadata
       metadata1
       metadata2
       """
       tables = parse_text(nodes)     
       assert len(tables.populations) == 2
       assert tables.popualtions[0].metadata == "metadata1"
       assert tables.popualtions[0].metadata == "metadata2"
```

This will exercise both paths. I'm becoming a big fan of small obvious tests, they're so much more effective than the big elaborate code-based approaches that I used to favour.

hyanwong:
> Looks good. It would be nice to add a couple of minimal tests...

Good call. This exposed another bug: we were forgetting to strip the "\n" from each line when doing strict text reading. I've added this in, but apart from the test you suggested above, which only applied to the population table, this change doesn't have an associated unit test.

(I've also changed the `strip("\n")` to `rstrip("\n")` which is a bit more specific.

jeromekelleher (approved):
LGTM

benjeffery (approved):

PR #1911, opened 2021-11-12
Reference sequence - C API

benjeffery:
WIP - Looking into this I think we just do it.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -325,6 +325,117 @@ test_table_collection_simplify_errors(void)
 }
 
 static void
+test_reference_sequence(void)
+{
+    int ret;
+    tsk_reference_sequence_t *r1 = NULL;
+    tsk_reference_sequence_t *r2 = NULL;
+
+    char example_data[100] = "An example string with unicode 🎄🌳🌴🌲🎋";
+    tsk_size_t example_data_length = (tsk_size_t) strlen(example_data);
+    char example_url[100] = "An example url with unicode 🎄🌳🌴🌲🎋";
+    tsk_size_t example_url_length = (tsk_size_t) strlen(example_url);
+    char example_metadata[100] = "An example metadata with unicode 🎄🌳🌴🌲🎋";
+    tsk_size_t example_metadata_length = (tsk_size_t) strlen(example_metadata);
+    char example_schema[100] = "An example schema with unicode 🎄🌳🌴🌲🎋";
+    tsk_size_t example_schema_length = (tsk_size_t) strlen(example_schema);
+
+    // Test equality
+    CU_ASSERT_TRUE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    r1 = tsk_malloc(sizeof(tsk_reference_sequence_t));
+    CU_ASSERT_NOT_EQUAL_FATAL(r1, NULL);
+    tsk_reference_sequence_init(r1);
+
+    ret = tsk_reference_sequence_set_data(r1, example_data, example_data_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    r2 = tsk_malloc(sizeof(tsk_reference_sequence_t));
+    CU_ASSERT_NOT_EQUAL_FATAL(r2, NULL);
+    tsk_reference_sequence_init(r2);
+
+    ret = tsk_reference_sequence_set_data(r1, "", 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    ret = tsk_reference_sequence_set_data(r1, example_data, example_data_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    ret = tsk_reference_sequence_set_data(r2, example_data, example_data_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    ret = tsk_reference_sequence_set_url(r1, example_url, example_url_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_equals(r1, r2, 0));
+    ret = tsk_reference_sequence_set_url(r2, example_url, example_url_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_equals(r1, r2, 0));
+
+    ret = tsk_reference_sequence_set_metadata(
+        r1, example_metadata, example_metadata_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_equals(r1, r2, 0));
+    ret = tsk_reference_sequence_set_metadata(
+        r2, example_metadata, example_metadata_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_equals(r1, r2, 0));
+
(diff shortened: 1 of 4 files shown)


benjeffery:
WIP, but probably useful to check I am heading in the right direction.

benjeffery:
I think this is worth you looking at now @jeromekelleher

c/tskit/tables.c
```diff
@@ -627,6 +627,20 @@ write_metadata_schema_header(
     return fprintf(out, fmt, (int) metadata_schema_length, metadata_schema);
 }
 
+int
+tsk_reference_sequence_free(tsk_reference_sequence_t *self)
+{
+    if (self != NULL) {
```
jeromekelleher, line 633: We're checking this already in table_collection_free. Best to stick with the standard idiom of assuming self is not NULL.

c/tskit/tables.c
```diff
@@ -9833,6 +9847,9 @@ tsk_table_collection_free(tsk_table_collection_t *self)
     tsk_mutation_table_free(&self->mutations);
     tsk_population_table_free(&self->populations);
     tsk_provenance_table_free(&self->provenances);
+    if (self->reference_sequence) {
```
jeromekelleher, line 9850:
```suggestion
    if (self->reference_sequence != NULL) {
```

c/tskit/tables.c
```diff
@@ -9882,6 +9899,47 @@ tsk_table_collection_equals(const tsk_table_collection_t *self,
               && tsk_provenance_table_equals(
                      &self->provenances, &other->provenances, options);
     }
+
+    ret = ret
+          && ((self->reference_sequence == NULL && other->reference_sequence == NULL)
```
jeromekelleher, line 9904: might as well abstract most of this out to ``tsk_reference_sequence_equals``?

c/tests/test_tables.c
```diff
@@ -324,6 +324,143 @@ test_table_collection_simplify_errors(void)
...
+    ret = tsk_reference_sequence_set_metadata_schema(
+        tc1.reference_sequence, example_schema, example_schema_length);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_dump(&tc1, _tmp_file_name, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_load(&tc2, _tmp_file_name, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(&tc1, &tc2, 0));
+    tsk_table_collection_free(&tc1);
+    tsk_table_collection_free(&tc2);
```
jeromekelleher, line 461: Looks good, but not clear to me what bits are testing the table_collection and what bits are testing ``reference_sequence``. Perhaps better to abstract out the operations we need to ``tsk_reference_sequence_t`` and test that outside of the context of the table collection, *then* test the table collection operations that depend on the reference sequence?

c/tskit/tables.c
```diff
@@ -9909,6 +9967,40 @@ tsk_table_collection_set_metadata_schema(tsk_table_collection_t *self,
         metadata_schema, metadata_schema_length);
 }
```
jeromekelleher, line 9969: Let's group these together with ``tsk_referenece_sequence_free``, so all the methods of the object are contiguous in the file (I guess with a standard banner)

c/tskit/tables.c
```diff
@@ -10087,6 +10179,41 @@ tsk_table_collection_copy(const tsk_table_collection_t *self,
         goto out;
     }
 
+    tsk_reference_sequence_free(dest->reference_sequence);
+    dest->reference_sequence = NULL;
+
+    if (self->reference_sequence != NULL) {
+        dest->reference_sequence = tsk_malloc(sizeof(tsk_reference_sequence_t));
```
jeromekelleher, line 10186: ``tsk_reference_sequence_copy``

c/tskit/tables.c
```diff
@@ -10087,6 +10179,41 @@ tsk_table_collection_copy(const tsk_table_collection_t *self,
         goto out;
     }
 
+    tsk_reference_sequence_free(dest->reference_sequence);
```
jeromekelleher, line 10182: I see, this is where free testing for NULL is handy. Still, easy to do 
```c
if (dest->reference_sequence != NULL) {
    tsk_reference_sequence_free(dest->reference_sequence);
}
```

c/tskit/tables.c
```diff
@@ -10301,6 +10428,74 @@ tsk_table_collection_load_indexes(tsk_table_collection_t *self, kastore_t *store
     return ret;
 }
 
+static int
+tsk_reference_sequence_load(tsk_reference_sequence_t **self, kastore_t *store)
```
jeromekelleher, line 10432: I'm finding this a bit convoluted. How about we make this a method of the table collection, like 
```
tsk_table_collection_load_reference_sequence...
```
then later
```c
reference_sequence_loaded = data != NULL || url != NULL || metadata != NULL || metadata_schema != NULL;
if (self->reference_sequence != NULL) {
     free
}
if (reference_sequence_loaded) {
    self->reference_sequence = malloc
     if (data != NULL) {
         ret = tsk_reference_sequence_set_data(*self, data, (tsk_size_t) data_length);  
         if (ret != 0) {
             goto out;
         }
     }
     // etc
    }
```
I think this captures the various permutations of reference/no reference before/after ok, right?

c/tskit/tables.c
```diff
@@ -10460,6 +10659,22 @@ tsk_table_collection_write_format_data(const tsk_table_collection_t *self,
     return ret;
 }
 
+static int TSK_WARN_UNUSED
+tsk_reference_sequence_dump(const tsk_reference_sequence_t *self, kastore_t *store,
```
jeromekelleher, line 10663: I'd make this a method of the table collection as well, keep consistency with the nearby methods. Doesn't make sense to have a load method and no dump method, I think?

c/tskit/tables.c
```diff
@@ -10548,6 +10763,12 @@ tsk_table_collection_dumpf(
     if (ret != 0) {
         goto out;
     }
+    if (self->reference_sequence) {
```
jeromekelleher, line 10766:
```suggestion
    if (self->reference_sequence != NULL) {
```

jeromekelleher (commented):
LGTM - some suggestions for moving things around so that most of the code is in the ``tsk_reference_sequence_t`` class, except for the table dump/load methods. These need to be in the table_collection class because we need only create a reference_sequence object if there is data to be loaded.

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tskit/tables.c
benjeffery: Done

c/tests/test_tables.c
benjeffery: Done

benjeffery:
Ok, after a very annoying double free I think this is ready to be looked at again.

jeromekelleher (approved):
Keeping track of whether the reference sequence is NULL or not is tricky and tedious, so I wonder if it would be simpler if we kept a ``state`` variable instead and embedded the reference_sequence_t struct in the table collection.

But let's merge this first - there's a few minor changes to make which hopefully make things follow the current idioms a bit better.

benjeffery:
> embedded the reference_sequence_t struct in the table collection.

That's how I originally had it!

I've addressed the comments, working on the python side now.

jeromekelleher:
> I've addressed the comments, working on the python side now.

How about get this merged in first? There's some errors in circleic

benjeffery:
> There's some errors in circleic

Fixed!

PR #1912, opened 2021-11-12
Tree variant hack

jeromekelleher:
Experimental code to explore API options for #605 

Next step is to put together some Python plumbing around the Variant2 type and see what perf is like for this simple implementation.

cc @benjeffery

### DIFF
--- a/c/tests/test_genotypes.c
+++ b/c/tests/test_genotypes.c
@@ -452,8 +452,9 @@ test_single_tree_char_alphabet2(void)
     ret = tsk_variant2_init(&var, &ts, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
-    ret = tsk_tree_decode_variant(&t, &t.sites[0], &var);
+    ret = tsk_variant2_decode(&var, &t, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(var.site.position, 0);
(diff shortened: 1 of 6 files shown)


jeromekelleher:
I've done up the plumbing and I think the basic pattern of making all the action happen in ``tsk_variant_decode(self, tree, site_index)`` works well. The Python infrastructure works well too, giving us a natural way to get zero-copy access to the genotype data, when we want it.

However, there is a substantial performance penalty for not using SAMPLE_LISTS when doing the decoding (I'm seeing about a 2x perf drop in a few different situations). Given that this is a fundamental operation that people do *a lot*, I don't think we can justify getting rid of sample lists just for the sake of slightly simpler code. Maybe someday someone will come up with a better algorithm for maintaining them :smile:

So: I think the plan is straightforward. We follow through on the path you're on in https://github.com/benjeffery/tskit/tree/remove-vargen @benjeffery, changing as little as possible. Porting the state into the variant_t object should be straightforward enough and we can make the changes in stages (first pass, keep the vargen_t object).

How does that sound @benjeffery? I'll keep this PR open for discussion, and maybe we can reuse bits of it later on when we're removing the vargen/VariantGenerator infrostructure.

benjeffery:
Sounds good, I agree that 2x for this operation is worth the additional complexity. Will finish the ref seq work and then get this done.

jeromekelleher:
This is out of date now, closing.

PR #1913, opened 2021-11-14
Rename tsk_tree_t.left/right to interval.left

### LINKED ISSUE #1686: Change `tsk_tree_t.left/right` to an interval.
Marked as a TODO in current code.

jeromekelleher:
Closes #1686

Doesn't look too bad noise-wise, and seems worth the update?

I wondered if there was something fancy we could do with unions that would allow us to use either ``tree.interval.left`` or tree.interval[0]``, depending on the need, but I don't think it's possible in C99 (and would be of very marginal benefit anyway)


benjeffery (approved):
LGTM!

PR #1919, opened 2021-11-17
Remove note about change of mutation state

hyanwong:
I think we now allow this anyway. See https://github.com/tskit-dev/tskit/issues/1180


jeromekelleher (approved):
Yep good spot

PR #1928, opened 2021-11-20
Reinstate nitpick

jeromekelleher:
We need to get nitpick mode working on the docs again, so here's some work towards that.

It looks like we can use ``nitpick_ignore_regex`` rather than ``nitpick_ignore`` and this will work. This is because it looks like ``nitpick_ignore`` searches for a literal tuple, where as the regex version iterates (more sensibly): https://github.com/sphinx-doc/sphinx/blob/aea2af418ea8d10dead88537ad560643325a583c/sphinx/transforms/post_transforms/__init__.py#L178

This does require a newer version of sphinx, so  there'll be some mucking about with version pinning.

### DIFF
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -160,7 +160,8 @@ typedef struct {
      the tree's genomic interval.
      */
     tsk_size_t num_edges;
-    /**
+    /* FIXME Undocumenting this for now to resolve some sphinx/doxygen issues */
+    /*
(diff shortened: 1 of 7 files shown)


jeromekelleher:
I've fiddled around with the autdocsumm thing a bit and I can't see any obvious way to fix it, so I think it's best to merge this. Not having nitpick mode means the docs will very quickly get broken.

benjeffery (approved):
Nice find with the regex fix!

PR #1931, opened 2021-11-22
Analyzer fixes

### LINKED ISSUE #1930: static analyzer issues for tskit
Ran the static analyzer on SLiM for the first time in a while, and it has flagged a variety of issues, most of which are nits, but at least one of which seems like a legit bug, albeit a very minor one.  Even the ones that are nits would be nice to get fixed, so that the static analyzer results are more clean in future; but unlike building with zero warnings, I'm content for that to be more of an aspirational goal.  :->  Details follow.  Note this is all for the version of tskit presently in SLiM, which is a little behind you guys; apologies if anything here is out of date.

---------------------

in read_table_ragged_cols(), tables.c:183, values stored to `data_col_present` and `offset_col_present` are never read (nit):

```
    for (col = cols; col->name != NULL; col++) {
        data_col_present = false;
        offset_col_present = false;
```

---------------------

in tsk_ibd_result_add_segment(), tables.c:7374, value stored to `ret` never read (actual bug, but marginal):

```
        if (avl_node == NULL) {
            ret = TSK_ERR_NO_MEMORY;
        }
```

should do `goto out;`.

---------------------

in simplifier_output_sites(), tables.c:8709, _both_ stores to `mapped_parent` never read (not clear to me if this is a bug or not):

```
                mapped_parent = TSK_NULL;
                if (input_parent != TSK_NULL) {
                    mapped_parent = self->mutation_id_map[input_parent];
                }
```

---------------------

in tsk_table_collection_check_tree_integrity(), tables.c:9558, store to `tree_right` never read (nit):

```
    tree_left = 0;
    tree_right = sequence_length;
    num_trees = 0;
```

---------------------

in tsk_individual_table_load(), tables.c:1360, third argument (`metadata_schema_length`) is an uninitialized value (nit, but gets flagged a bunch of times):

```
    if (metadata_schema != NULL) {
        ret = tsk_individual_table_set_metadata_schema(
            self, metadata_schema, metadata_schema_length);
        if (ret != 0) {
            goto out;
        }
    }
```

This traces down to the logic in read_table_properties(), which calls `ret = kastore_gets()`.  If that returns `ret != 0` it transforms the error code with tsk_set_kas_error() and returns, without ever writing into `*property->len_dest`.  This is fine, I imagine, since the error code is still guaranteed to be non-zero.  The static analyzer doesn't understand what `tsk_set_kas_error()` is doing, though, so it says "ah, but what if ret is now 0?  Then `metadata_schema_length` is uninitialized!"  I think the correct fix is perhaps to add an `assert(ret != 0);` right after the call to `tsk_set_kas_error()` in `read_table_properties()`, which will reassure the static analyzer on that point.  I'm 95% sure I followed the static analyzer's reasoning here correctly; and adding the `assert()` in SLiM's copy of the code does fix the warning from the static analyzer, so it seems like the right fix.  :->

---------------------

I hope this is useful.  The static analyzer kind of straddles the line between "useful" and "spam".  :->

jeromekelleher:
Closes #1930

I've mostly made the changes suggested @bhaller. The one in ibd_segments had already been caught, and that was the only actual bug.

The only change I've made is to move the assert into ``set_kas_error`` rather than doing it in the location you suggested. This seemed like it would be a better approach, if this function was upsetting the static analyser.

What do you think?

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -279,6 +279,7 @@ read_table_properties(
             ret = kastore_gets(store, property->name, property->array_dest, &len, &type);
             if (ret != 0) {
                 ret = tsk_set_kas_error(ret);
+                assert(ret != 0); /* Tell static analysers that we're handling errors */
                 goto out;
             }
             if (type != property->type) {
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -179,8 +179,6 @@ read_table_ragged_cols(kastore_t *store, tsk_size_t *num_rows,
     tsk_size_t *offset_array;
 
     for (col = cols; col->name != NULL; col++) {
-        data_col_present = false;
-        offset_col_present = false;
         ret = kastore_containss(store, col->name);
         if (ret < 0) {
             ret = tsk_set_kas_error(ret);
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -9588,7 +9588,6 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
     tsk_memset(parent, 0xff, self->nodes.num_rows * sizeof(*parent));
 
     tree_left = 0;
-    tree_right = sequence_length;
     num_trees = 0;
     j = 0;
     k = 0;
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -8737,7 +8737,7 @@ simplifier_output_sites(simplifier_t *self)
     tsk_id_t input_mutation, mapped_parent, site_start, site_end;
     tsk_id_t num_input_sites = (tsk_id_t) self->input_tables.sites.num_rows;
     tsk_id_t num_input_mutations = (tsk_id_t) self->input_tables.mutations.num_rows;
-    tsk_id_t input_parent, num_output_mutations, num_output_site_mutations;
+    tsk_id_t num_output_mutations, num_output_site_mutations;
     tsk_id_t mapped_node;
     bool keep_site;
     bool filter_sites = !!(self->options & TSK_FILTER_SITES);
@@ -8755,11 +8755,6 @@ simplifier_output_sites(simplifier_t *self)
                && self->input_tables.mutations.site[input_mutation] == site.id) {
             mapped_node = self->mutation_node_map[input_mutation];
             if (mapped_node != TSK_NULL) {
-                input_parent = self->input_tables.mutations.parent[input_mutation];
-                mapped_parent = TSK_NULL;
-                if (input_parent != TSK_NULL) {
-                    mapped_parent = self->mutation_id_map[input_parent];
-                }
                 self->mutation_id_map[input_mutation] = num_output_mutations;
                 num_output_mutations++;
                 num_output_site_mutations++;


benjeffery (approved):
LGTM

bhaller:
> Closes #1930
> 
> I've mostly made the changes suggested @bhaller. The one in ibd_segments had already been caught, and that was the only actual bug.
> 
> The only change I've made is to move the assert into `set_kas_error` rather than doing it in the location you suggested. This seemed like it would be a better approach, if this function was upsetting the static analyser.
> 
> What do you think?

I agree that doing the fix in `set_kas_error()` is cleaner.  I actually tried putting it there, though, and that did not seem to be sufficient to convince the static analyzer that there was not a problem, oddly, which is why I suggested putting it outside.  Strangely, the static analyzer did not show steps walking into `set_kas_error` and back out again; it seemed to just regard it as a black box that altered `ret` in an unknowable way, even when an `assert()` was inside it.  Not sure why that would be; perhaps it exceeded a maximum stack depth that it's willing to go down into for analysis, or something.

jeromekelleher:
Ah, thanks @bhaller - looks like we need a different approach so. I'll update with your version in a minute.

(I took the AUTOMERGE label off @benjeffery as we need to make some tweaks)

benjeffery:
> (I took the AUTOMERGE label off @benjeffery as we need to make some tweaks)

Whoops, sorry thought this was done.

jeromekelleher:
> I actually tried putting it there, though, and that did not seem to be sufficient to convince the static analyzer that there was not a problem, oddly, which is why I suggested putting it outside. Strangely, the static analyzer did not show steps walking into set_kas_error and back out again; it seemed to just regard it as a black box that altered ret in an unknowable way, even when an assert() was inside it. 

Hmm, perhaps because it's in a different compilation unit? Would you mind trying making a copy of set_kas_error (say set_kas_error2 or someting) and putting it in the same file to see if it'll peer into it's innards then? It'd be good to know.

bhaller:
> > I actually tried putting it there, though, and that did not seem to be sufficient to convince the static analyzer that there was not a problem, oddly, which is why I suggested putting it outside. Strangely, the static analyzer did not show steps walking into set_kas_error and back out again; it seemed to just regard it as a black box that altered ret in an unknowable way, even when an assert() was inside it.
> 
> Hmm, perhaps because it's in a different compilation unit? Would you mind trying making a copy of set_kas_error (say set_kas_error2 or someting) and putting it in the same file to see if it'll peer into its innards then? It'd be good to know.

Tried that just now; didn't seem to make a difference.  Here's a screenshot:

Constrictor 1

Notice that the "Assuming `ret` is not equal to zero" line is selected in the log on the left; that is assuming that the return value from `kastore_gets()` is not equal to zero, and so it enters the `if (ret != 0)` branch at line 298.  Then it calls `tsk_set_kas_error_2()`, which you can see declared immediately above, as a `static` in the same compilation unit.  Nevertheless, there are no "Calling..." and "Entered call" entries in the log on the left, as there were for the other nested function calls in the log; the static analyzer does not seem to have actually gone into this call for some reason.  That can also be seen graphically; if it had gone into `tsk_set_kas_error_2()`, the GUI would show light blue arrows inside it in the code view, since it would be part of the trace, but there are no light blue arrows: not part of the trace, for some reason.  So the assert there, at line 277, is not seen by the analyzer, and a couple of lines lower in the log on the left, you can see "Assuming `ret` is equal to zero"; that assumption is contrary to the fact asserted at line 277, since that `ret` is the same on returned by `tsk_set_kas_error_2()`.  So, it does not appear to be a compilation unit thing.  My guess is still that it has to do with a maximum depth, beyond which it just assumes that functions act as black boxes and could return anything, or something like that.  There does seem to be a build setting governing "deep" versus "shallow" analysis (https://stackoverflow.com/questions/29635938/whats-the-difference-between-deep-and-shallow-static-analysis) but the above is with "deep".  It looks like that still has a limited "maximum number of nodes", if https://codechecker.readthedocs.io/en/latest/analyzer/checker_and_analyzer_configuration/ is relevant.  I don't see a way to choose an even higher maximum than "deep" in Xcode's UI, but from that last link it looks like maybe that can be configured if you run the analysis at the command line instead of in Xcode.  Adventures for another day.  :->

So, for practical purposes, if the goal is to get rid of the static analyzer warning in Xcode (that's certainly _my_ goal :->), the `assert()` can't go in `tsk_set_kas_error()`.  :-<

jeromekelleher:
Thanks for trying this out @bhaller - I've put in the ``assert`` where you originally suggested. Let's hope it placates the static analysis gods :smile:

jeromekelleher:
I can confirm that ``clang-tidy -p build/compile_commands.json tskit/tables.c`` gets rid of this warning, so this is good to merge.

PR #1932, opened 2021-11-22
Address comments about metadata docs

hyanwong:
As per initial "minor style suggestion" at https://github.com/tskit-dev/tutorials/pull/143#issuecomment-974149092n - does this read better @LukeAndersonTrocme ?

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -18,43 +18,46 @@ kernelspec:
 
 # Metadata
 
-The tree-sequence and every entity within it (nodes, mutations, edges,  etc.) can have
-metadata associated with them. This is intended for storing and passing on information
-that tskit itself does not use or interpret. For example information derived from a VCF
-INFO field, or administrative information (such as unique identifiers) relating to
-samples and populations. Note that provenance information about how a tree sequence
-was created should not be stored in metadata, instead the provenance mechanisms in
-tskit should be used (see {ref}`sec_provenance`).
+The tree-sequence and all the entities within it (nodes, mutations, edges,  etc.) can
+have metadata associated with them. This is intended for storing and passing on
+information that tskit itself does not use or interpret, for example information derived
+from a VCF INFO field, or administrative information (such as unique identifiers)
+relating to samples and populations. Note that provenance information about how a tree
+sequence was created should not be stored in metadata, instead the provenance mechanisms
+in tskit should be used (see {ref}`sec_provenance`).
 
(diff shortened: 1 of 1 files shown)


docs/metadata.md
```diff
@@ -18,43 +18,45 @@ kernelspec:
...
+metadata automatically and, most importantly, tells downstream users and tools how to
+decode and interpret the metadata. For example, the `msprime` schema for populations
+requires both a `name` and a `description` for each defined population: these names and
+descriptions can assist downstream users in understanding and using `msprime` tree
+sequences. It is best practice to populate such metadata fields if your files will be
+used by any third party, or if you wish to remember what the rows refer to some time
+after making the file!
+
+Technically, schemas describe what information is stored in each metadata record, and
+how it is to be encoded. Every node's metadata follows the node schema, every mutation's
```
jeromekelleher, line 40: maybe also,

> plus some optional rules about the types and ranges of data that can be stored

jeromekelleher (approved):
LGTM, minor suggestion

LukeAndersonTrocme:
Great update! I like the way Schemas are introduced right from the start, and all those hyperlinks are useful as well. 

Just to nitpick, when you say "Typically these schemas are defined by the software which created the tree-sequence file" I wonder if you could mention or point to an example?

hyanwong:
> Great update! I like the way Schemas are introduced right from the start, and all those hyperlinks are useful as well.
> 
> Just to nitpick, when you say "Typically these schemas are defined by the software which created the tree-sequence file" I wonder if you could mention or point to an example?

I'm not sure many packages do this yet (although there is current discussion on how SLiM is doing it). However, we have mentioned the `msprime` example in the paragraph above, so we could say "typically, as in the example of `msprime` above, schemas are defined by the software which created the tree-sequence file"?

LukeAndersonTrocme:
"typically, as in the example of msprime above, schemas are defined by the software which created the tree-sequence file" is perfect. I asked mainly because I wasn't sure how many other packages supported this already!

benjeffery:
> "typically, as in the example of msprime above, schemas are defined by the software which created the tree-sequence file" is perfect. I asked mainly because I wasn't sure how many other packages supported this already!

We're in very early days of tree sequence metadata, so tools are still integrating it. Thankyou so much for all your feedback, it's invaluable.

hyanwong:
At the moment, "ideally" might be a better adjective than "typically" :)

jeromekelleher:
Are we good to merge here?

PR #1939, opened 2021-11-24
catch TypeError in raise_hdf5_format_error

### LINKED ISSUE #1938: raise_hdf5_format_error doesn't work for streams
The recently introduced  [raise_hdf5_format_error](https://github.com/tskit-dev/tskit/blob/315e47a6c063098c72533c10d575464edcabc747/python/tskit/formats.py#L266) can be called on either a path or a file object, but assumes the input is a path.

Since we can't sniff the first few bytes of a file object (it might be an unseekable stream, like a socket) it's pretty questionable whether this functionality is worth it

clwgg:
## Description

The `filename` argument to `raise_hdf5_format_error` is expected to be a path and throws a `TypeError` if `TreeSequence.load` is called on a stream that triggers a `tskit.FileFormatError`. This PR catches that TypeError and throws the original exception. As mentioned by @jeromekelleher in https://github.com/tskit-dev/tskit/issues/1938 it doesn't make a ton of sense to try to do the `hdf_magic` comparison on a stream anyway, since we could be anywhere in a non-seekable stream.

Fixes #1938.
Makes newly introduced tests in https://github.com/tskit-dev/tskit/pull/1882 pass properly.

### DIFF
--- a/python/tskit/formats.py
+++ b/python/tskit/formats.py
@@ -268,6 +268,13 @@ def raise_hdf5_format_error(filename, original_exception):
     Tries to open the specified file as a legacy HDF5 file. If it looks like
     an msprime format HDF5 file, raise an error advising to run tskit upgrade.
     """
+
+    def raise_original_exception():
+        # We want to keep a useful error message here as well as the chaining history.
+        raise exceptions.FileFormatError(
+            str(original_exception)
+        ) from original_exception
+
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thanks for this @clwgg - it seems easier to just remove this special case which I've done over in #1942. It would be great if you could take a quick look.

clwgg:
sounds good!

PR #1942, opened 2021-11-25
Remove special case handling for HDF5 files.

### LINKED ISSUE #1938: raise_hdf5_format_error doesn't work for streams

jeromekelleher:
Closes #1938

This seems simpler than trying to untangle what happens when we have stream arguments. The new error message hopefully provides as much help anyway.


benjeffery (approved):
LGTM, much simpler

PR #1944, opened 2021-11-25
Refseq update

jeromekelleher:
WIP, still working out the details.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -360,11 +360,75 @@ test_table_collection_simplify_errors(void)
 }
 
 static void
+test_reference_sequence_state_machine(void)
+{
+
+    tsk_reference_sequence_t r1;
+
+    tsk_reference_sequence_init(&r1, 0);
+    CU_ASSERT_EQUAL(r1.data, NULL);
+    CU_ASSERT_EQUAL(r1.url, NULL);
+    CU_ASSERT_EQUAL(r1.metadata, NULL);
+    CU_ASSERT_EQUAL(r1.metadata_schema, NULL);
+    CU_ASSERT_TRUE(tsk_reference_sequence_is_null(&r1));
+
+    CU_ASSERT_EQUAL(tsk_reference_sequence_set_data(&r1, "x", 1), 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_is_null(&r1));
+    /* Setting the value back to NULL makes the reference whole object NULL */
+    CU_ASSERT_EQUAL(tsk_reference_sequence_set_data(&r1, NULL, 0), 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_is_null(&r1));
+    tsk_reference_sequence_free(&r1);
+    CU_ASSERT_TRUE(tsk_reference_sequence_is_null(&r1));
+
+    /* Any empty string is the same thing. */
+    tsk_reference_sequence_init(&r1, 0);
+    CU_ASSERT_EQUAL(tsk_reference_sequence_set_data(&r1, "", 0), 0);
+    CU_ASSERT_TRUE(tsk_reference_sequence_is_null(&r1));
+    tsk_reference_sequence_free(&r1);
+
+    tsk_reference_sequence_init(&r1, 0);
+    CU_ASSERT_EQUAL(tsk_reference_sequence_set_url(&r1, "x", 1), 0);
+    CU_ASSERT_FALSE(tsk_reference_sequence_is_null(&r1));
+    tsk_reference_sequence_free(&r1);
+
+    tsk_reference_sequence_init(&r1, 0);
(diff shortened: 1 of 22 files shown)


jeromekelleher:
I think this is ready for a look @benjeffery, if you wouldn't mind. I think most of the stuff has been covered, other than some issues that have popped up on the TableCollection tests, some extra file format tests, and some minimal docs.

Do we bump the file format and dict encoding minor numbers for this?

jeromekelleher:
Good to get your eyes on this if possible please @benjeffery - hopefully I can get it finished in one more pass, it would be really helpful to see what you think.

benjeffery:
> Good to get your eyes on this if possible please @benjeffery - hopefully I can get it finished in one more pass, it would be really helpful to see what you think.

Sorry, having a proper look now.

jeromekelleher:
OK, this is ready for review I think! I think I've covered all the high-level things and opened issues for the stuff we haven't done. Do you see anything missing @benjeffery?

c/tests/test_tables.c
```diff
@@ -359,12 +359,77 @@ test_table_collection_simplify_errors(void)
...
+{
+
+    tsk_reference_sequence_t r1;
+
+    tsk_reference_sequence_init(&r1, 0);
+    CU_ASSERT_EQUAL(r1.data, NULL);
+    CU_ASSERT_EQUAL(r1.url, NULL);
+    CU_ASSERT_EQUAL(r1.metadata, NULL);
+    CU_ASSERT_EQUAL(r1.metadata_schema, NULL);
+    CU_ASSERT_EQUAL(r1.data, NULL);
```
benjeffery, line 373: Duplicate line.

c/tskit/tables.c
```diff
@@ -10700,7 +10700,10 @@ tsk_table_collection_reference_sequence_dump(const tsk_table_collection_t *self,
             ref->metadata_schema_length, KAS_UINT8 },
         { .name = NULL },
     };
-    return write_table_cols(store, write_cols, 0);
+    if (!tsk_reference_sequence_is_null(&self->reference_sequence)) {
```
benjeffery, line 10703:
```suggestion
    if (tsk_table_collection_has_reference_sequence(self)) {
```

c/tskit/tables.h
```diff
@@ -4004,6 +4004,10 @@ life-cycle.
 bool tsk_table_collection_has_index(
     const tsk_table_collection_t *self, tsk_flags_t options);
 
+/* TODO not clear this should have the "options" flags. */
+bool tsk_table_collection_has_reference_sequence(
+    const tsk_table_collection_t *self, tsk_flags_t options);
```
benjeffery, line 4009: I can only imagine this having options if we use it to check fetch status, but I think that would be clearer as an additional function.

docs/python-api.md
```diff
@@ -1130,6 +1130,7 @@ The following constants are used throughout the `tskit` API.
 
 ```{eval-rst}
 .. autoclass:: TableCollection
+    :inherited-members:
```
benjeffery, line 1155: TableCollection doesn't inherit I thought?

python/tests/test_reference_sequence.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+        tables = tskit.TableCollection.fromdict(d)
+        assert not tables.has_reference_sequence()
+
+
+class TestSummaries:
+    def test_repr(self):
+        tables = tskit.TableCollection(1)
+        refseq = tables.reference_sequence
+        # TODO add better tests when summaries are updates
```
benjeffery, line 147:
```suggestion
        # TODO add better tests when summaries are updated
```

benjeffery (approved):
Couple of small things, overall I think this is great.
I have some concern about `.data` being a string at this point when we know we want to do zero-copy via arrays. It would likely be a breaking change to do so. Maybe we use `as_string` or similar for now at the public API? Then `data` can be the array later?

docs/python-api.md
benjeffery: Ah, when I got to the metadata stuff this made sense.

jeromekelleher:
> have some concern about .data being a string at this point when we know we want to do zero-copy via arrays. It would likely be a breaking change to do so. Maybe we use as_string or similar for now at the public API? Then data can be the array later?

I've opened issues #1989 and #1988 to address this - what do you think? Basically, the ``.data`` attribute is there, but high-level users won't really be seeing it.

jeromekelleher:
Thanks for the review @benjeffery, I think I've addressed those points. Ready to go I think!

benjeffery:
> > have some concern about .data being a string at this point when we know we want to do zero-copy via arrays. It would likely be a breaking change to do so. Maybe we use as_string or similar for now at the public API? Then data can be the array later?
> 
> I've opened issues #1989 and #1988 to address this - what do you think? Basically, the `.data` attribute is there, but high-level users won't really be seeing it.

Great, if `data` isn't the way that high-level users get at the sequence then it's all good.

PR #1945, opened 2021-11-25
Show an eg of a tree sequence with multiple roots

hyanwong:
It's confusing at first to think of multi-root tree as a single tree: most phylogenetic people would think off this as 2 trees. Hopefully showing an example for a whole tree sequence clarifies the distinction.

I've also shown the virtual root in red italics not bold, which I think ties in better with the concept of "virtual" (bold means more "real" to me)

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -1048,33 +1048,38 @@ position 100 to 200, has a single root. Finally the third tree, from position
 
 #### The virtual root
 
-To aid keeping track of multiple roots within a single tree, tskit uses a
-special additional node called the **virtual root**. All roots in all the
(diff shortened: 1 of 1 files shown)


hyanwong:
While I was looking at this section, I also took the opportunity to restructure the text explaining about the virtual root, as I found the current text goes into quite a lot of detail, before explaining that the user doesn't usually need to think about it at all! So I've put that bit up top, and described the virtual root as a bookkeeping device, which I think is a reasonable way to think of it. I've put this as a separate commit in case what I've written is worse or makes it less clear.

benjeffery (approved):
LGTM!

jeromekelleher (approved):
This is great, thanks @hyanwong

PR #1950, opened 2021-11-26
Skip validation when schema is trivial

### LINKED ISSUE #1923: Bypass metadata validation when codec=JSON and zero properties
In this case there's nothing to validate anyway, right? This would probably save good chunk of time on a common case.

benjeffery:
Fixes #1923

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -32,6 +32,7 @@ import pprint
 import struct
 import tempfile
 import unittest
+from unittest.mock import patch
 
 import msgpack
 import msprime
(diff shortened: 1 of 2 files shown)


python/tests/test_metadata.py
```diff
@@ -542,6 +542,26 @@ def test_bad_type_error(self):
         ):
             ms.validate_and_encode_row(tskit.TableCollection(1))
 
+    def test_skip_validation(self):
+        ms = tskit.MetadataSchema({"codec": "json"})
```
jeromekelleher, line 546: I guess mock.patch would be a bit shorter/more idiomatic here?
```python
with mock.patch(ms._validate_row) as mocked_validate:
     ms.validate_and_encode_row({})]
     assert mocked_validate.call_count == 0
```
something along these lines should do it.

python/tests/test_metadata.py
```diff
@@ -542,6 +542,26 @@ def test_bad_type_error(self):
...
+        triggered = False
+
+        def tracer(data):
+            nonlocal triggered
+            triggered = True
+
+        ms._validate_row = tracer
+        ms.validate_and_encode_row({})
+        assert not triggered
+
```
jeromekelleher, line 557: Might as well make this two tests, since they're indepdent if you use the ``mock`` thing

python/tests/test_metadata.py
```diff
@@ -542,6 +542,26 @@ def test_bad_type_error(self):
...
+        ms = tskit.MetadataSchema({"codec": "json", "properties": {"foo": {}}})
+        assert not ms._bypass_validation
+        triggered = False
+        ms._validate_row = tracer
+        ms.validate_and_encode_row({})
+        assert triggered
+
```
jeromekelleher, line 564: We should also check that validate is always called on non-json codecs right?

python/tskit/metadata.py
```diff
@@ -623,8 +632,10 @@ def __init__(self, schema: Optional[Mapping[str, Any]]) -> None:
             codec_instance = codec_cls(schema)
             self._string = tskit.canonical_json(schema)
             self._validate_row = TSKITMetadataSchemaValidator(schema).validate
+            self._bypass_validation = codec_cls.is_schema_trivial(schema)
```
jeromekelleher, line 635: Not clear why this is a class method - why not ``codec_instance.is_schema_trivial()``?

jeromekelleher (approved):
LGTM, some suggestions above.

python/tests/test_metadata.py
benjeffery: Done

python/tests/test_metadata.py
benjeffery: Done

python/tests/test_metadata.py
benjeffery: Done

python/tskit/metadata.py
benjeffery: Just matching the other methods that deal with schemas such as `modify schema`

PR #1958, opened 2021-11-28
Rejiggle c deps

jeromekelleher:
This changes around the repo layout a little to reduce the amount of symlinking we're doing. It's mainly motivated by the need to move away from the current approach so that we can work with versions of meson > 0.55, which we've been pinning on for a while an has been cropping up as a problem with devs on a regular basis.

The upside is that meson works much better using it's native subproject mechanism, and I think the overall project structure is a bit less complicated.

The downside is that we have two different ways of specifying the kastore dependency for C library development and the Python module, which will both have to be updated when we do a version bump. We'll need to document this clearly. (For example, I've forgotten to update the submodule to kastore C_2.0.1). I guess the best thing to do is put in a few tests that check the version number in various places to make it harder to forget this.

Any thoughts? @molpopgen, how does this affect you for the downstream Rust lib?


molpopgen:
The short answer is that I'd need to dig in a bit in order to know.  Right now, I manually copy files over, but I'd prefer using a submodule.  The submodule approach currently fails because the build system sees more than one symlink to the same file.

jeromekelleher:
> The short answer is that I'd need to dig in a bit in order to know. Right now, I manually copy files over, but I'd prefer using a submodule. The submodule approach currently fails because the build system sees more than one symlink to the same file.

This should help with that hopefully as there's now there's a lot less symlinking, and none under the ``c`` directory. You'd probably need to do your own submodule for kastore rather than pulling it in recursively from tskit, but that's probably OK?

molpopgen:
> > The short answer is that I'd need to dig in a bit in order to know. Right now, I manually copy files over, but I'd prefer using a submodule. The submodule approach currently fails because the build system sees more than one symlink to the same file.
> 
> This should help with that hopefully as there's now there's a lot less symlinking, and none under the ``c`` directory. You'd probably need to do your own submodule for kastore rather than pulling it in recursively from tskit, but that's probably OK?
> 

That should work. I'll just have to have a way to remember to update both of them as necessary.

benjeffery:
I assume this would be superseded by #1967 ?

jeromekelleher:
Yes - I'll update when I get a chance

jeromekelleher:
Actually easier to start again, closing.

PR #1964, opened 2021-11-29
Clearer docs distinguishing .tables from .dump_tables

hyanwong:
Even though under the hood they currently do exactly the same!


benjeffery:
If they do the same thing isn't this a little confusing without explaining why you should want to use dump_tables?

jeromekelleher (approved):
LGTM

jeromekelleher:
Hopefully we'll implement dump_tables like ``ts.tables.copy()`` someday.

PR #1968, opened 2021-11-29
Deprecate Tree.num_nodes

### LINKED ISSUE #1966: Deprecate Tree.num_nodes?
Even though it's documented, I think it's super confusing that `Tree.num_nodes` returns the number of nodes in the entire tree sequence, not the number of nodes in the tree itself. I suggest we might want to deprecate it and say that the user should use `Tree.tree_sequence.num_nodes` instead, which is much clearer.

hyanwong:
Fixes #1966

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -880,11 +880,11 @@ class HighLevelTestCase:
 
     def verify_tree_mrcas(self, st):
         # Check the mrcas
-        oriented_forest = [st.get_parent(j) for j in range(st.num_nodes)]
+        oriented_forest = [st.get_parent(j) for j in range(st.tree_sequence.num_nodes)]
         mrca_calc = tests.MRCACalculator(oriented_forest)
         # We've done exhaustive tests elsewhere, no need to go
         # through the combinations.
-        for j in range(st.num_nodes):
+        for j in range(st.tree_sequence.num_nodes):
             mrca = st.get_mrca(0, j)
             assert mrca == mrca_calc.get_mrca(0, j)
             if mrca != tskit.NULL:
@@ -3050,7 +3050,7 @@ class TestTree(HighLevelTestCase):
             ) | {root}
 
             # test MRCA
-            if tree.num_nodes < 20:
+            if tree.tree_sequence.num_nodes < 20:
                 for u, v in itertools.combinations(tree.nodes(), 2):
                     mrca = nx.lowest_common_ancestor(g, u, v)
                     if mrca is None:
@@ -3165,12 +3165,12 @@ class TestTree(HighLevelTestCase):
             return v in path
 
         tree = self.get_tree()
-        for u, v in itertools.product(range(tree.num_nodes), repeat=2):
+        for u, v in itertools.product(range(tree.tree_sequence.num_nodes), repeat=2):
             assert is_descendant(tree, u, v) == tree.is_descendant(u, v)
         # All nodes are descendents of themselves
-        for u in range(tree.num_nodes + 1):
+        for u in range(tree.tree_sequence.num_nodes + 1):
             assert tree.is_descendant(u, u)
-        for bad_node in [-1, -2, tree.num_nodes + 1]:
+        for bad_node in [-1, -2, tree.tree_sequence.num_nodes + 1]:
             with pytest.raises(ValueError):
                 tree.is_descendant(0, bad_node)
             with pytest.raises(ValueError):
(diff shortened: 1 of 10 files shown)


jeromekelleher:
LGTM.

We might as well remove the C python bits as well and replace with ``self.tree_sequence.num_nodes``, since it's not really used.

If you're comfortable doing that here @hyanwong go head, otherwise we can make a quick issue to track.

hyanwong:
> We might as well remove the C python bits as well and replace with `self.tree_sequence.num_nodes`, since it's not really used.
> 
> If you're comfortable doing that here @hyanwong go head, otherwise we can make a quick issue to track.

I can remove the C interface too. In another commit, is it worth changing the deprecation warnings elsewhere to FutureWarning rather than RuntimeWarning (as they are now)? Happy to do that too since my head is in it.

hyanwong:
> > We might as well remove the C python bits as well and replace with `self.tree_sequence.num_nodes`, since it's not really used.
> 
> I can remove the C interface too.

Removed in latest commit.

I notice that the C interface also has `tree->num_nodes`, which could, I presume, be removed and replaced by `tree->tree_sequence->tables->nodes.num_rows`. That's a breaking change though. Do you want that done too?

benjeffery:
Theres a broken test here, also we need to have a test for the warning behaviour.

hyanwong:
> Theres a broken test here, also we need to have a test for the warning behaviour.

Yep, on it, thanks @benjeffery  but sorting out my PCR test ATM!

hyanwong:
> > Theres a broken test here, also we need to have a test for the warning behaviour.
> 
> Yep, on it, thanks @benjeffery but sorting out my PCR test ATM!

Done now, I hope (had to update a fair few tests, but was a trivial change). Also I changed the number of nodes reported by `str(tree)` and `tree._repr_html_` to be the actual number in the tree (i.e. reachable from the roots), which I think is probably what people would want to know?

I think the codecov report can be ignored, but say if you think an extra test is needed @benjeffery ?

python/tskit/trees.py
```diff
@@ -1403,12 +1403,23 @@ def is_descendant(self, u, v):
...
         """
         Returns the number of nodes in the :class:`TreeSequence` this tree is in.
-        Equivalent to ``tree.tree_sequence.num_nodes``. To find the number of
-        nodes that are reachable from all roots use ``len(list(tree.nodes()))``.
+        Equivalent to ``tree.tree_sequence.num_nodes``.
+
+        .. deprecated:: 0.4
+            Use :attr:`Tree.tree_sequence.num_nodes<TreeSequence.num_nodes>` if you want
+            the number of nodes in the entire tree sequence, or
+            ``len(list(tree.nodes()))`` to find the number of nodes that are
```
jeromekelleher, line 1411: I guess we should make this as efficient as we can, so ``len(tree.preorder())`` would be better.

python/tskit/util.py
```diff
@@ -524,7 +524,7 @@ def tree_html(tree):
                       <tr><td>Index</td><td>{tree.index}</td></tr>
                       <tr><td>Interval</td><td>{tree.interval.left:.8g}-{tree.interval.right:.8g} ({tree.span:.8g})</td></tr>
                       <tr><td>Roots</td><td>{tree.num_roots}</td></tr>
-                      <tr><td>Nodes</td><td>{tree.num_nodes}</td></tr>
+                      <tr><td>Nodes</td><td>{len(list(tree.nodes()))}</td></tr>
```
jeromekelleher, line 527: Definitely here using ``len(tree.preorder())`` is better.

jeromekelleher (approved):
LGTM, thanks for following up on this messy one. Not sure why some lines are coming up as not covered by tests?

jeromekelleher:
The changes should definitely be covered by tests, but maybe codecov is acting up?

python/tskit/trees.py
hyanwong: Yes, nice one

jeromekelleher:
I don't understand what codecov is complaining about, I'm happy to merge.

python/tests/test_highlevel.py
```diff
@@ -3206,10 +3206,19 @@ def test_apis(self):
             assert t1.get_mrca(*pair) == t1.mrca(*pair)
             assert t1.get_tmrca(*pair) == t1.tmrca(*pair)
 
+    @pytest.mark.filterwarnings("ignore::FutureWarning")
```
benjeffery, line 3209: Whats this doing? Shouldn't the warnings be caught by `pytest.warns`?

python/tests/test_highlevel.py
```diff
@@ -3206,10 +3206,19 @@ def test_apis(self):
...
         t1 = self.get_tree()
         assert t1.get_length() == t1.span
         assert t1.length == t1.span
+        assert t1.num_nodes == t1.tree_sequence.num_nodes
+
+    @pytest.mark.filterwarnings("error::FutureWarning")
+    def test_deprecated_api_warnings(self):
+        # Deprecated and will be removed
+        t1 = self.get_tree()
+        with pytest.raises(FutureWarning, match="Tree.tree_sequence.num_nodes"):
```
benjeffery, line 3220: Should also be `with pytest.warns`?

benjeffery (approved):
Small question about the tests.

python/tests/test_highlevel.py
hyanwong: Changed

hyanwong: I think we want to ignore the warnings when testing the functionality (which is what this does), and then check for the warnings in the next function using `pytest.warns`?

hyanwong:
I think this is right now @benjeffery. Unless you prefer testing for both the warning and the functionality in one go, i.e.

```
with pytest.warns(FutureWarning, match="Tree.tree_sequence.num_nodes"):
            assert t1.num_nodes == t1.tree_sequence.num_nodes
```

But that's mixing up 2 separate things, I think?

hyanwong:
I think this can be merged now?

PR #1973, opened 2021-11-30
Vendor kastore

### LINKED ISSUE #1687: Remove tree diff iterator removal?
There's a note to remove the tree diff iterator in `trees.h`. However I can see this is being used in the KC distance (the note to suggest removing pre-dates the KC implementation) and haplotype matching code.

I suggest we just remove the note, and if we're happy with the iterator document it.

jeromekelleher:
Second attempt at simplifying C dependencies after  #1958.

Closes #1687


jeromekelleher:
I think this is ready to go - there will be some short-term pain for devs though as git gets confused about having/not having submodules.

benjeffery (approved):
Works well here and seems that `meson dist` also works. I added a changelog.

jeromekelleher:
Great! Let's merge, no point in putting this one off.

PR #1980, opened 2021-12-02
Swap boilerplate table warning

hyanwong:
Removes duplication, and we might want to change this text anyway, as we can now substitute a new numpy array for each column using the setters


jeromekelleher (approved):
LGTM

hyanwong:
Do we want to change the text? E.g.

```
The numpy arrays returned by table attribute accesses are copies of the underlying data. In particular, this means that editing individual values in the arrays will not change the table data. Instead, you should set entire columns or rows at once (see :ref:`sec_tables_api_accessing_table_data`)
```

jeromekelleher:
Yeah, that looks good.

hyanwong:
Changed

PR #1987, opened 2021-12-02
Add header structure to ref docs

hyanwong:
I've found it's really helpful to have header links on the RHS navbar in the docs for each class. This adds them. I wasn't sure what title to use. E.g. we already use the title "The Tree Class" in the summaries higher up the page. I wondered about "**Tree** class documentation", but that's very long and clutters up the navbar. So I ended up using "**Tree** docs", which is short if a bit jargon-y.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -19,8 +19,8 @@ kernelspec:
 # Python API
 
 This page documents the full tskit Python API. Brief thematic summaries of common
-classes and methods are presented first. The {ref}`sec_reference_api` at the end
-then contains full details which aim to be concise, precise and exhaustive.
+classes and methods are presented first. The {ref}`sec_python_api_reference` section
+at the end then contains full details which aim to be concise, precise and exhaustive.
 Note that this may not therefore be the best place to start if you are new
 to a particular piece of functionality.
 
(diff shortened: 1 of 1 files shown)


docs/python-api.md
```diff
@@ -1079,12 +1079,22 @@ The following constants are used throughout the `tskit` API.
 
 ### Tree and tree sequence classes
 
+#### {class}`Tree` docs
```
jeromekelleher, line 1082: How about API instead of docs?

jeromekelleher: Does the {class}`Tree` thing in the title do something different to just "Tree docs"?

jeromekelleher (approved):
LGTM - some suggestions.

docs/python-api.md
hyanwong: > Does the {class}Tree thing in the title do something different to just "Tree docs"?

It marks it in bold in the navbar, which I think is visually helpful. But not much otherwise, given it links to the documentation just underneath it.

jeromekelleher: OK, just checking it didn't do something mad with the links.

hyanwong: > How about API instead of docs?

Yes, that would work. If so, I wonder if we want to call the whole section "API reference" too, instead of "Reference documentation"

jeromekelleher: Sure, whatever you think

hyanwong: Done. Should be OK to merge

jeromekelleher (approved):

Issue #1977, opened 2021-12-02
Zero copy loading from kastore


jeromekelleher:
As discussed in #1970 the current default behaviour on calling ``tsk_table_collection_load`` is:

- Call kastore_load() with flags=LOAD_ALL
- For each table, call tsk_x_table_set_columns (or similar for reference sequence)

This means that we have two copies of the table/reference sequence data present in memory, and this can be an issue for downstream users, if there are very large files.

We could, in principle, avoid this in the following way:

- Add a method to kastore to return the pointer to the memory buffer returned by malloc that is used to store the file contents when we specify LOAD_ALL (we read in the whole file at once, or nearly all anyway, we have to read the header first to find out how big it is).
- Add an option to kastore_free/init to tell it to **not** free this buffer when the kastore is freed.
- Add an option to tsk_x_table_set_columns to tell it to **not** copy the input data, but to just use the specified pointers directly. This would need to set some internal flag also to tell it to not free these pointers.
- Similarly, the various set_string methods would need an option ``TSK_BORROW_REFERENCE``. This would then handle the reference sequence problem.
- We'd need to do something similar for the top level metadata in the table collection.

This is all straightforward enough, but the difficulty occurs when someone tries to update this table collection. The simplest thing would be to say that it's strictly read-only. However, this does mean that *if* someone wanted to update these large models they'd have to take a copy, and then we'd be back to square one in terms of memory high-water mark.

So, to do this, we'd need to implement the C-level state-machine enforcing read-only behaviour on tables and table collections (and reference sequences). Cf. #1906, #993, #1917

@bhaller - what do you think? Would this work for your use-case, or would you need to be able to subsequently write to the tables that you read in?

bhaller:
I don't think SLiM would use this facility, no; we need to be able to modify the tables.  As far as I can tell, the copy from kastore is unavoidable in general.  In #1970 I'm specifically concerned with avoiding the copy of the reference sequence, because in that specific case (a) the reference sequence can be extremely large (the tables could too, I suppose, but a 1 GB – 10 GB reference sequence is commonplace), (b) the copy *can* be avoided because SLiM doesn't want/need tskit to keep the reference sequence itself, and (c) SLiM's own copy of the reference sequence can be made directly from kastore, and is 2-bits-per-byte, so that avoids having double the memory overhead.

[...]

PR #1990, opened 2021-12-02
Add options flags to string setters

### LINKED ISSUE #1979: Add `options` flag to all string setting methods
As discussed in #1977 we may want to avoid always taking a copy of input strings when doing things like ``tsk_table_collection_set_metadata_schema``, ``tsk_reference_sequence_set_data`` etc.

It seems sensible to add an ``options`` flag to each of these methods to keep the door open for this in the future.

jeromekelleher:
Closes #1979

The idea of this was to open the ground for #1977, but I'm having second thoughts now because the ``x_table_set_columns`` doesn't have an options value (which I forgot).

I guess we could add an option to ``tsk_x_table_init(table, options)`` which is like ``TSK_EXTERNAL_MEMORY`` or something, which flicks a switch were we store pointers in ``set_columns`` and ``set_metadata_schema``, ``set_reference_sequence_data`` etc. So then, we don't need these options values.

I guess the alternative is to add ``options`` to all the ``x_tables_set_columns``.

Any thoughts @benjeffery?

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -33,6 +33,10 @@
 - ``kastore`` is now vendored into this repo instead of being a git submodule. Developers need to run
   ``git submodule update``. (:user:`jeromekelleher`, :issue:`1687`, :pr:`1973`)
 
+- All string setting methods (e.g. `tsk_table_collection_set_metadata`) now take an extra
+  "options" argument (:user:`jeromekelleher`, :pr:`1990`, :issue:`1979`). To
+  retain the current behaviour set options=0 for all calls.
+
 
 **Features**
 
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -190,12 +190,12 @@ test_table_collection_equals_options(void)
 
     // Equivalent except for time_units
     ret = tsk_table_collection_set_metadata(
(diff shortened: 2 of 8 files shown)


jeromekelleher:
Ah, I'd forgotten about the update case where we want to pass ownership of a large string to the table. This is easy to do, and definitely requires the options flag, so it's a done-deal that we need this. I'll update later.

benjeffery:
Adding options to `x_tables_set_columns` seems exactly the kind of thing we should be doing before 1.0, so makes sense to me.

jeromekelleher:
I was thinking the same thing...

jeromekelleher:
OK, I think this is ready to go - might was well get the messy stuff out of the way before starting on #2006.

jeromekelleher:
Oops, needs a CHANGELOG entry...

jeromekelleher:
Ugh, maybe we don't want to do this. As noted in https://github.com/tskit-dev/tskit/issues/2006#issuecomment-985776440 we'll have to drop all the const qualifiers in the signatures if we do it this way, which is particularly annoying when dealing with literal strings.

Maybe we should have a separate set of methods instead?

jeromekelleher:
Closed in favour of "takeset" methods, #2016

PR #2010, opened 2021-12-03
Rework data-models isolated samples 

### LINKED ISSUE #2002: Embed "missing data" as a subsection of an "encoding genotypes" section in data-models
The "data-models" doc page introduces tables and data encoding in tables, then (I think correctly) talks about how we encode a tree in tskit. The final section is about missing data. But I think that's not quite right. I think the last section should be about how genotypes arise from trees + sites + mutations (i.e. the encoding of genetic data), which can be very short, and which can then incorporate the missing data section. I'll PR this separately.

hyanwong:
Fixes #2002

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -733,7 +733,7 @@ Schema section (TODO).
 
 (sec_table_indexes)=
 
-## Table indexes
+### Table indexes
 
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM, thanks.

Issue #1971, opened 2021-11-30
Add the skip_reference_sequence and ignore_reference_sequence options


jeromekelleher:
We recently added the concept of "table data" in 82a56e7 with the addition of the ``skip_tables`` flag to tskit.load() and the ``ignore_tables`` flag to ``TableCollection.equals()`` (and the corresponding flags to the C API). Since that change was made we also in parallel added basic support for reference sequence data. As @bhaller points out (https://github.com/tskit-dev/tskit/issues/1854#issuecomment-982263133) the skip_tables option loads the reference sequence data.

The skip_tables option was initially motivated by the desire to get access to the top-level metadata only (#1854). Providing access *only* to the metadata is a non-starter I think, because it's much simpler to skip loading stuff into the table collection that it is to provide separate APIs for accessing the metadata. So, there will always be some extra info that comes with the metadata, and this is correct I think: what if I was going through a bunch of files just to read their ``uuid`` values? This isn't metadata, and I wouldn't want to read the whole file just to get them either.

The question then is what we do from this point. Since we want the option of not loading reference sequence data, the options as I see it are: 

1. Add similar flags like ``skip_reference_sequence`` and ``ignore_reference_sequence`` to ``load`` and ``equals``
2. Regard ``reference_sequence`` as ``table`` data, and document as such
3. Rename the ``skip_tables`` and ``ignore_tables`` flags to something like ``top_level_only`` and be clear that we don't consider ``reference_sequence`` as top level data.

Any thoughts @bhaller @clwgg?

benjeffery:
To me, the reference sequence is in the same category as the table data, and we use the existing flags. Happy to hear of any use cases that need just the reference though.

bhaller:
> To me, the reference sequence is in the same category as the table data, and we use the existing flags.

This seems fine to me.  Of course, who knows what different functionality end users might request next :->.  It might be that simply providing multiple flags – load this, don't load that – would provide flexibility for the future that might prove useful.  But perhaps it makes sense to wait and see, rather than trying to put all that flexibility in at the outset (perhaps complicating the API unnecessarily).

If the policy choice is that the reference sequence is "table data", then perhaps rather than renaming the flag away from `skip_tables`, the reference sequence ought to be renamed to reflect that policy?  It could be called, e.g., `refseq_table` or some such?

[...]

PR #2011, opened 2021-12-04
Take string methods for reference_sequence

jeromekelleher:
A first pass to see what "take" methods will look like for strings. This is done on the reference sequence so that @bhaller can incorporate it into SLiM before we've gone through the full update.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -596,6 +596,7 @@ out:
     return ret;
 }
 
+/* TODO rename to copy_string or replace_and_copy_string */
 static int
 replace_string(
     char **str, tsk_size_t *len, const char *new_str, const tsk_size_t new_len)
@@ -614,6 +615,15 @@ replace_string(
     }
 out:
     return ret;
+}
+
+static int
+takeset_string(char **str, tsk_size_t *len, char *new_str, const tsk_size_t new_len)
+{
+    tsk_safe_free(*str);
+    *str = new_str;
+    *len = new_len;
+    return 0;
 }
 
 static int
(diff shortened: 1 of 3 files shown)


c/tskit/tables.c
```diff
@@ -659,18 +669,8 @@ bool
 tsk_reference_sequence_equals(const tsk_reference_sequence_t *self,
     const tsk_reference_sequence_t *other, tsk_flags_t options)
 {
-    int ret;
```
jeromekelleher, line 662: This is an incidental change - I realised these checks were unnecessary now.

bhaller:
LGTM @jeromekelleher.  So this will get merged, #1971 will get merged, and then I'll pull the tskit sources into SLiM again and give it a whirl; IIRC at that point all the reference sequence API changes should be settled out, so we can make sure that it dovetails with SLiM nicely.  Sound good?

jeromekelleher:
Yep, that's the plan @bhaller. I think we'll push out a 0.99 release of the C API once this and #1971 have been merged, a beta for Python 0.4.0. I'd really like to do this in a small number of days...

bhaller:
> Yep, that's the plan @bhaller. I think we'll push out a 0.99 release of the C API once this and #1971 have been merged, and a beta for Python 0.4.0. I'd really like to do this in a small number of days...

Yes, me too – I leave for a vacation in two weeks, and very much want to release SLiM 3.7 with all of this stuff before then.  As soon as 0.99 is released I'll get going on the next steps.  Great to have an official release with this before 1.0, BTW; it will allow all the broken tests in SLiM's test suite, due to the individuals table sorting change, to go green prior to 1.0's release, which will be quite helpful.  Thanks!

benjeffery (approved):
LGTM, I think `takeset` is quite clever as it looks odd enough that I would make no assumptions and go read the docs :laughing:

bhaller:
SLiM is happy, and tskit 0.99.15 is now merged into SLiM's master branch.  Onward!

PR #2019, opened 2021-12-06
Add the skip_reference_sequence and ignore_reference_sequence options

### LINKED ISSUE #1971: Add the skip_reference_sequence and ignore_reference_sequence options

clwgg:
## Description

Fixes #1971

Adds two new C flags: `TSK_LOAD_SKIP_REFERENCE_SEQUENCE` and `TSK_CMP_IGNORE_REFERENCE_SEQUENCE`.
These are exposed to python as the `skip_reference_sequence` flag to `TableCollection.load` and `TreeSequence.load`, as well as the `ignore_reference_sequence` flag to `TableCollection.equals` and `TreeSequence.equals`.
Since the flags were written this way in the issue I kept them "written out", though I think an argument could be made to shorten `reference_sequence` to `refseq` in all four cases (your call @jeromekelleher, @benjeffery).

# PR Checklist:

### DIFF
--- a/c/tests/test_file_format.c
+++ b/c/tests/test_file_format.c
@@ -1338,6 +1338,67 @@ test_skip_tables(void)
     free(ts1);
 }
 
+static void
+test_skip_reference_sequence(void)
+{
+    int ret;
+    tsk_treeseq_t *ts1 = caterpillar_tree(5, 3, 3);
+    tsk_treeseq_t ts2;
+    tsk_table_collection_t t1, t2;
+    FILE *f;
+
+    CU_ASSERT_TRUE(tsk_treeseq_has_reference_sequence(ts1));
+
+    ret = tsk_treeseq_dump(ts1, _tmp_file_name, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_load(
+        &t1, _tmp_file_name, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    CU_ASSERT_FALSE(tsk_table_collection_equals(&t1, ts1->tables, 0));
+    CU_ASSERT_TRUE(tsk_table_collection_equals(
+        &t1, ts1->tables, TSK_CMP_IGNORE_REFERENCE_SEQUENCE));
+    CU_ASSERT_FALSE(tsk_table_collection_has_reference_sequence(&t1));
+
+    /* Test _loadf code path as well */
+    f = fopen(_tmp_file_name, "r+");
+    ret = tsk_table_collection_loadf(&t2, f, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(&t1, &t2, 0));
+    fclose(f);
+    tsk_table_collection_free(&t2);
+
+    /* Setting TSK_LOAD_SKIP_REFERENCE_SEQUENCE only reads part of the file */
+    f = fopen(_tmp_file_name, "r+");
+    ret = tsk_table_collection_loadf(&t2, f, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_NOT_EQUAL(fgetc(f), EOF);
+    fclose(f);
+    tsk_table_collection_free(&t2);
+
+    /* We should be able to make a tree sequence */
+    ret = tsk_treeseq_init(&ts2, &t1, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_treeseq_free(&ts2);
+
+    /* Do the same thing with treeseq API */
+    ret = tsk_treeseq_load(&ts2, _tmp_file_name, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(&t1, ts2.tables, 0));
+    tsk_treeseq_free(&ts2);
+
+    f = fopen(_tmp_file_name, "r+");
+    ret = tsk_treeseq_loadf(&ts2, f, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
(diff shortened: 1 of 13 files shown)


python/tskit/trees.py
```diff
@@ -2997,11 +2997,18 @@ def load(file, *, skip_tables=False):
         Please note that with this option set, it is not possible to load data from
         a stream of multiple tree sequences using consecutive calls to
         :meth:`tskit.load`.
+    :param bool skip_reference_sequence: If True, the tree sequence is read
```
jeromekelleher, line 3000: I wonder if we should say that the ``skip_x`` options require a seekable stream, and they'll fail if you call them on a socket or stdin?

c/tskit/tables.c
```diff
@@ -10518,7 +10521,11 @@ tsk_table_collection_loadf_inited(
     int ret = 0;
     kastore_t store;
 
-    int kas_flags = options & TSK_LOAD_SKIP_TABLES ? 0 : KAS_READ_ALL;
+    int kas_flags = KAS_READ_ALL;
+    if ((options & TSK_LOAD_SKIP_TABLES)
+        | (options & TSK_LOAD_SKIP_REFERENCE_SEQUENCE)) {
```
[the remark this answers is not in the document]
jeromekelleher: I guess if you wanted to be fancy you could say ``options & (TSK_LOAD_SKIP_TABLES|TSK_LOAD_SKIP_REFERENCE_SEQUENCE)``, but I think I'd prefer the more obvious version.

jeromekelleher (commented):
This is fantastic, thanks @clwgg. I spotted a few minor issues.

Can you update the changelogs also please? (Should be a copy/paste from the ignore_tables entries)

jeromekelleher:
> Since the flags were written this way in the issue I kept them "written out", though I think an argument could be made to shorten reference_sequence to refseq in all four cases 

I agree there's an argument, but we've erred on the side of verbosity elsewhere so we might as well stick with what we have now, IMO. I'm happy to change if someone else feels strongly about it, though, I'm not bothered either way.

c/tskit/tables.h
```diff
@@ -3539,10 +3547,13 @@ different error conditions. Please see the
 sequentially load tree sequences from a stream.
 
 Please note that this streaming behaviour is not supported if the
-TSK_LOAD_SKIP_TABLES option is set. With this option, only the top-level
-information of the table collection will be read, leaving all tables empty. When
-attempting to read from a stream with multiple table collection definitions and
-the TSK_LOAD_SKIP_TABLES option set, only the top-level information of the first
+TSK_LOAD_SKIP_TABLES or TSK_LOAD_SKIP_REFERENCE_SEQUENCE option is set.
+If the TSK_LOAD_SKIP_TABLES option is set, only the top-level
+information of the table collection will be read, leaving all tables empty.
```
benjeffery, line 3552:
```suggestion
If the TSK_LOAD_SKIP_TABLES option is set, only the non-table
information from the table collection will be read, leaving all tables with zero rows and no metadata or schema.
```

benjeffery (approved):
LGTM, though one doc string needed a little tweak.

c/tests/test_file_format.c
```diff
@@ -1338,6 +1338,67 @@ test_skip_tables(void)
...
+        &t1, _tmp_file_name, TSK_LOAD_SKIP_REFERENCE_SEQUENCE);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    CU_ASSERT_FALSE(tsk_table_collection_equals(&t1, ts1->tables, 0));
+    CU_ASSERT_TRUE(tsk_table_collection_equals(
+        &t1, ts1->tables, TSK_CMP_IGNORE_REFERENCE_SEQUENCE));
+    CU_ASSERT_FALSE(tsk_table_collection_has_reference_sequence(&t1));
+
+    /* Test _loadf code path as well */
+    f = fopen(_tmp_file_name, "r+");
```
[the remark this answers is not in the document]
clwgg: it does pass -- there's the `tsk_treeseq_dump` on line 1352. do you think it would be better to dump from the table_collection?

c/tskit/tables.c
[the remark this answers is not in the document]
clwgg: ooops :shame:

python/tskit/trees.py
clwgg: yeah, I was wondering about that but wasn't sure how to fit it into the :param structure. should I just throw it in at the top-level before the :param blocks?

c/tskit/tables.h
clwgg: thanks @benjeffery! since that's the `skip_tables` docstring, should I open another PR for that? or are y'all ok with throwing that in with the `skip_refseq` PR?

benjeffery: It should be in this PR as that text only needs changing because of the ref seq.

python/tskit/trees.py
jeromekelleher: Throw it in before the :param: blocks yeah, maybe as a ``.. warning::`` or something?

c/tests/test_file_format.c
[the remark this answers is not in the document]
jeromekelleher: Doh! :facepalm:

clwgg:
- added documentation review suggestions
- added code review suggestions
- added changelogs
- rebased & squashed

jeromekelleher:
Looks great, thanks @clwgg, let's merge!

Issue #1921, opened 2021-11-18
Document TableRow classes


hyanwong:
E.g. in https://tskit.dev/tskit/docs/latest/python-api.html#tables - other "general" functions like `add_row` etc are documented for each table, so I guess `replace()` should be too.

benjeffery:
The table row classes that `replace` is a method on are undocumented as they used to be very simple. They should probably be documented now. Changing the title of the issue.

hyanwong:
Oh yes, this is on a *row*, sorry.

hyanwong:
Fixed in #2020

PR #2020, opened 2021-12-06
Document table row classes & provenance classes

hyanwong:
A proposal for fixing #1921. Not point in repeating documentation here, I reckon, so I'm just pointing to the equivalent attributes in the tree sequence container class.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -1234,7 +1234,8 @@ Also see the {ref}`sec_tables_api_table_collection` summary.
 % Overriding the default signatures for the tables here as they will be
 % confusing to most users.
 
-#### The {class}`IndividualTable` class
+
+#### {class}`IndividualTable` classes
 
 ```{eval-rst}
 .. autoclass:: IndividualTable()
@@ -1243,7 +1244,20 @@ Also see the {ref}`sec_tables_api_table_collection` summary.
     :special-members: __getitem__
 ```
 
-#### The {class}`NodeTable` class
+##### Associated row class
+
+A row returned from an {class}`IndividualTable` is an instance of the following
+basic class, where each attribute matches an identically named attribute in the
+{class}`Individual` class.
+
+```{eval-rst}
+.. autoclass:: IndividualTableRow()
+    :members:
+    :inherited-members:
+```
+
+
+#### {class}`NodeTable` classes
 
 ```{eval-rst}
 .. autoclass:: NodeTable()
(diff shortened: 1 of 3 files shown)


docs/python-api.md
```diff
@@ -1243,6 +1243,18 @@ Also see the {ref}`sec_tables_api_table_collection` summary.
     :special-members: __getitem__
 ```
 
+##### Associated row class
+
+A row returned from an {class}`IndividualTable` is an instance of the following
+simple {mod}`dataclass<python:dataclasses>`. Each attribute in this class matches
```
jeromekelleher, line 1249: I don't think we want to document them as dataclasses - we might want to change this implementation detail in the future.

python/tskit/tables.py
```diff
@@ -64,11 +64,15 @@ class NOTSET(metaclass=NotSetMeta):
...
+    """
+    A row in an :class:`IndividualTable`.
+    """
+
     __slots__ = ["flags", "location", "parents", "metadata"]
-    flags: int
-    location: np.ndarray
-    parents: np.ndarray
-    metadata: Optional[Union[bytes, dict]]
+    flags: int  #: See :attr:`Individual.flags`
```
jeromekelleher, line 72: Agree there's no point in duplicating here.

Might was well use the standard docstring form though, like
```
""" See :attr:`Individual.flags` """
flags: int
```
as there will be cases in which the form you have will be too long for a single line, and it's good to keep things uniform, even if just for greppability.

jeromekelleher (commented):
LGTM, minor comments

docs/python-api.md
hyanwong: Yeah, that's a good point.

hyanwong:
Now fixes #1921. Note that this PR also adds documentation to the Provenance class.

python/tskit/trees.py
```diff
@@ -603,10 +603,22 @@ def __eq__(self, other):
 
 @dataclass
 class Provenance(util.Dataclass):
+    """
+    A provenance entry in a tree sequence, detailing what how this tree
+    sequence was generated, or what has been done to it (see :ref:`sec_provenance`).
+    """
+
```
benjeffery, line 610:
```suggestion
    """
    A provenance entry in a tree sequence, detailing how this tree
    sequence was generated, or subsequent operations on it (see :ref:`sec_provenance`).
    """

```

benjeffery (approved):
LGTM, one sentence didn't make sense.

python/tskit/trees.py
hyanwong: Oops. Editing error. Nice spot, thanks.

jeromekelleher (approved):

hyanwong:
I forgot to take this off draft mode. Done now, so can be flagged for merge.

PR #2023, opened 2021-12-06
Use ignore_provenance in tests

### LINKED ISSUE #905: Update tests to use the new equals with options
> There are a [few](https://github.com/tskit- dev/tskit/blob/40ce01eb47006a38f79f70dd2f733e7ecf3c26eb/python/tests/test_topology.py#L2430) [places](https://github.com/tskit-dev/tskit/blob/40ce01eb47006a38f79f70dd2f733e7ecf3c26eb/python/tests/test_tables.py#L1995) where we clear provenances before comparing equality. We could use this function there, or (probably better) call things like `simplify()` with `record_provenance=False` in the tests.

_Originally posted by @hyanwong in https://github.com/tskit-dev/tskit/issues/896#issuecomment-704528632_

Now with #807 we can stop clearing the provenance table before checking for equality. Instead of using `t1 == t2` you may use `t1.equals(t2, ignore_provenance=True)`

hyanwong:
Fixes #905 and also removes a load of asserts in test_topology.py::TestGeneralSamples::verify_permuted_nodes() which AFAICS did nothing apart from test that the tree sequences (or underlying tables) were the same.


jeromekelleher (approved):
LGTM, nice cleanup

PR #2024, opened 2021-12-06
Zero-copy loading for tree-sequences

### LINKED ISSUE #2132: Avoid copy of tables when creating treeseq_t
We want avoid copying the tables when we create a new tree sequence [here](https://github.com/tskit-dev/tskit/blob/36313962161f24d5493904cc6011f1aff7b90c5a/c/tskit/trees.c#L495)
 
See #2024 for prototype.

jeromekelleher:
Fixes #2132 

Here's the basic idea for what we're thinking about regarding ``tsk_treeseq_init`` and ``tsk_treeseq_load`` @molpopgen 

This should avoid one of the copies we get, passing ownership of the tables to the tree sequence.

There's some details here around the exact semantics of exactly who owns what when (which is why the tests are failing, I thing), but I'm sure it can be worked out.

Will this work in Rust-land?

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -2292,11 +2292,6 @@ test_simplest_bad_edges(void)
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     tsk_treeseq_free(&ts);
 
-    /* NULL for tables should be an error */
-    ret = tsk_treeseq_init(&ts, NULL, load_flags);
-    CU_ASSERT_EQUAL(ret, TSK_ERR_BAD_PARAM_VALUE);
-    tsk_treeseq_free(&ts);
-
     /* Bad population ID */
     tables.nodes.population[0] = -2;
     ret = tsk_treeseq_init(&ts, &tables, load_flags);
@@ -5831,11 +5826,11 @@ test_empty_tree_kc(void)
 
     ret = tsk_table_collection_init(&tables, 0);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
-    ret = tsk_treeseq_init(&ts, &tables, TSK_BUILD_INDEXES | TSK_SAMPLE_LISTS);
+    ret = tsk_treeseq_init(&ts, &tables, TSK_BUILD_INDEXES);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_SEQUENCE_LENGTH);
     tsk_treeseq_free(&ts);
     tables.sequence_length = 1.0;
-    ret = tsk_treeseq_init(&ts, &tables, TSK_BUILD_INDEXES | TSK_SAMPLE_LISTS);
+    ret = tsk_treeseq_init(&ts, &tables, TSK_BUILD_INDEXES);
     CU_ASSERT_EQUAL_FATAL(ret, 0);
 
     verify_empty_tree_sequence(&ts, 1.0);
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -10747,6 +10747,28 @@ out:
     return ret;
 }
 
+static int TSK_WARN_UNUSED
+tsk_table_collection_set_file_uuid(tsk_table_collection_t *self, const char *uuid)
+{
+    int ret = 0;
+
+    tsk_safe_free(self->file_uuid);
+    self->file_uuid = NULL;
+
+    if (uuid != NULL) {
+        /* Allow space for \0 so we can print it as a string */
+        self->file_uuid = tsk_malloc(TSK_UUID_SIZE + 1);
+        if (self->file_uuid == NULL) {
+            ret = TSK_ERR_NO_MEMORY;
+            goto out;
+        }
+        tsk_memcpy(self->file_uuid, uuid, TSK_UUID_SIZE);
+        self->file_uuid[TSK_UUID_SIZE] = '\0';
+    }
+out:
+    return ret;
+}
+
 int TSK_WARN_UNUSED
 tsk_table_collection_copy(const tsk_table_collection_t *self,
     tsk_table_collection_t *dest, tsk_flags_t options)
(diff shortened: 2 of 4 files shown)


molpopgen:
I suspect that this will work.  I may need to mess around with how I store a table collection, and I'd need to make myself an example to try this out.

molpopgen:
@jeromekelleher -- I have a mock-up [here](https://github.com/molpopgen/tskit_ownwership_mockup).  If you look in the C files, I have a comment about the use of valgrind to show that this scheme works. I think this lines up well with what's here?

jeromekelleher:
Some details to be worked out here around edge metadata, but this basically works I think. Not sure about the flag name, either.

I think we can probably let go of the idea of having read-only access in C (document things as "undefined behaviour"), which should let us get rid of a bunch of the TODOs around init-time here.

jeromekelleher:
> @jeromekelleher -- I have a mock-up here. If you look in the C files, I have a comment about the use of valgrind to show that this scheme works. I think this lines up well with what's here?

Yes, this looks exactly right @molpopgen. The ``tables`` pointer must be allocated by malloc/calloc/etc and must point to a valid table collection. We free the pointer in ``tsk_table_collection_free``.

molpopgen:
> > @jeromekelleher -- I have a mock-up here. If you look in the C files, I have a comment about the use of valgrind to show that this scheme works. I think this lines up well with what's here?
> 
> Yes, this looks exactly right @molpopgen. The ``tables`` pointer must be allocated by malloc/calloc/etc and must point to a valid table collection. We free the pointer in ``tsk_table_collection_free``.

After reading a lot more last night, my example may work by luck rather than design. I'm using one of rust's smart pointers, which uses the rust allocator. That allocator does not guarantee to be malloc! That makes sense of course, but I had blissfully suppressed that possibility in my mind. It's not a big deal, and it just means that I need to store raw pointers instead of smart pointers.

benjeffery:
Here's some insight into what is happening with memory usage with `tskit.load`.
Here's the code:
```
import time
import tskit
time.sleep(0.1)
tc = tskit.TableCollection.load("msp.ts")
time.sleep(0.1)
del tc
time.sleep(0.1)
ts = tskit.load("msp.ts")
time.sleep(0.1)
```
We load a table collection (for comparison) clear it, and then load as tree sequence.
Three code versions:
 -  `main` (as of yesterday with no zero-copy functionality) 
 - `takeset_columns` (zero-copy TableCollection loading)
 - `TSK_TAKE_TABLES` (this branch)
main
take-cols
take-tables

As you can see `TSK_TAKE_TABLES` removes one copy making the loading completely zero-copy. The remaining memory increase at about 1.8s to 2s is from the following "convenience" arrays. Some of which I assume could be made lazy?
(these are all the entries bigger than 10000bytes for the msprime ts):

- #nodes - `tsk_table_collection_check_tree_integrity`  `parent` array. Temporary!
- #samples - `tsk_treeseq_init_nodes`:`ts->samples`
- #nodes - `tsk_treeseq_init_nodes`:`ts->sample_index_map`
- #mutations - `tsk_treeseq_init_sites`:`ts->site_mutations_mem`
- #sites - `tsk_treeseq_init_sites`:`ts->site_mutations_length`
- #sites - `tsk_treeseq_init_sites`:`ts->site_mutations`
- #sites - `tsk_treeseq_init_sites`:`ts->tree_sites_mem`
- #node_referenced_individuals - `tsk_treeseq_init_individuals`:`ts->individual_nodes_mem`
- #individuals - `tsk_treeseq_init_individuals`:`ts->individual_nodes`
- #trees - `tsk_treeseq_init_trees`:`ts->tree_sites_length`
- #trees - `tsk_treeseq_init_trees`:`ts->tree_sites`
- #trees - `tsk_treeseq_init_trees`:`ts->breakpoints`

jeromekelleher:
This is great, thanks @benjeffery!

I'm relaxed about the extra memory used in the additional arrays used in the tree sequence - these were all conscious space-time tradeoffs, which I'd need significant convincing to give up on. If we really need to cut down memory at some point we can look into it, but I don't think there's any point right now.

I'm curious about what's happening between t=1.3 and 1.8 - I guess this is all ``check_integrity``?

jeromekelleher:
So we've chopped off ~200M off the peak usage here, right, ~25%?

benjeffery:
> I'm curious about what's happening between t=1.3 and 1.8 - I guess this is all check_integrity 

Yes - 31.7% of load time is `tsk_table_collection_check_integrity`. I guess #2067 would fix that.

> So we've chopped off ~200M off the peak usage here, right, ~25%?

Yes, that's about right. I think it is worth noting that `ts->site_mutations_mem` which grows with the number of mutations appears to only be used in the LD calculator and `tsk_viterbi_matrix_traceback`.

jeromekelleher:
> . I think it is worth noting that ts->site_mutations_mem which grows with the number of mutations appears to only be used in the LD calculator and tsk_viterbi_matrix_traceback. 

That is good to know. It is used in ``tsk_treeseq_get_site`` though which is a fundamental operation, the memory management for site and mutation objects is quite tricky. We'd have to change the API quite a bit to make them allocated on demand.

benjeffery:
> > . I think it is worth noting that ts->site_mutations_mem which grows with the number of mutations appears to only be used in the LD calculator and tsk_viterbi_matrix_traceback.
> 
> That is good to know. It is used in `tsk_treeseq_get_site` though which is a fundamental operation, the memory management for site and mutation objects is quite tricky. We'd have to change the API quite a bit to make them allocated on demand.

Yeah, those are the methods that use `tsk_treeseq_get_site`. I agree that it would be a lot of work to make them generate on demand! Lets wait until we have an actual problem in the massive tree sequence glorious future.

benjeffery:
Do we need any extra testing here? Looks like everything but the memory error paths are covered (except copy errors which are false hard to trigger).

benjeffery (approved):

hyanwong:
> I think it is worth noting that ts->site_mutations_mem which grows with the number of mutations appears to only be used in the LD calculator and tsk_viterbi_matrix_traceback.

Do we have any idea of the amounts that each of these convenience arrays contribute to the memory footprint. If there is one that makes up the lion's share, it might be worth thinking about, otherwise it probably isn't worth spending time on anyway.

benjeffery:
> otherwise it probably isn't worth spending time on anyway.

If we or others hit some constraint we at least have things we can do. I think we leave it till then.

jeromekelleher:
These really aren't convenience arrays, and it would require serious re-engineering of the memory layout to get rid of them. Even if they're not used much at the C layer, they're used extensively by the Python C API. They're a conscious space/(time/code complexity) tradeoff, and there's no reason to start messing with them now.

PR #2041, opened 2021-12-09
Add note about using divergence/diversity for mrcas

hyanwong:
I think this is useful, and I only just realised it. Ping @awohns too.


petrelharp:
Ah, very good idea.

molpopgen:
Likewise, the branch flavor of the segregating sites statistic is the sum of branch lengths for a set of nodes.

petrelharp:
I'm not worried about that failing test; I'll merge this.

PR #2042, opened 2021-12-09
Clarify "dead leaf" docs & leaves iterator

### LINKED ISSUE #339: tree.leaves(u) returns valid output even for invalid nodes u
See the screenshot below. Node 6 is not in the first tree anywhere, but `tree.leaves(6)` returns output that is totally valid.

Screen Shot 2019-08-23 at 4 47 08 pm

Perhaps this behaviour is needed for something, but it led to some difficult-to-trace errors when I was working on [this](https://github.com/tskit-dev/msprime/pull/799).

hyanwong:
Fixes #339

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1002,10 +1002,11 @@ class Tree:
 
         As this is defined by a traversal of the tree, technically we
         return the sum of all branch lengths that are reachable from
-        roots. Thus, this is the sum of all branches that are ancestral
+        roots. Thus, this is the total length of all branches that are connected
         to at least one sample. This distinction is only important
         in tree sequences that contain 'dead branches', i.e., those
-        that define topology not ancestral to any samples.
+        that define topology that is not connected to a tree root
+        (see :ref:`sec_data_model_tree_dead_leaves_and_branches`)
 
         :return: The sum of lengths of branches in this tree.
         :rtype: float
@@ -1374,6 +1375,16 @@ class Tree:
         Returns True if the specified node is a leaf. A node :math:`u` is a
         leaf if it has zero children.
 
+        .. note::
+            :math:`u` can be any node in the entire tree sequence, including ones
+            which are not connected via branches to a root node of the tree (and which
+            are therefore not conventionally considered part of the tree). Indeed, if
+            there are many trees in the tree sequence, it is common for the majority of
+            non-sample nodes to be :meth:`isolated<is_isolated>` in any one
+            tree. By the definition above, this method will return ``True`` for such
+            a tree when a node of this sort is specified. Such nodes can be thought of
+            as "dead leaves", see :ref:`sec_data_model_tree_dead_leaves_and_branches`.
+
         :param int u: The node of interest.
         :return: True if u is a leaf node.
         :rtype: bool
@@ -1383,7 +1394,8 @@ class Tree:
     def is_isolated(self, u):
         """
         Returns True if the specified node is isolated in this tree: that is
-        it has no parents and no children. Sample nodes that are isolated
+        it has no parents and no children (note that all isolated nodes in the tree
+        are therefore also :meth:`leaves<Tree.is_leaf>`). Sample nodes that are isolated
         and have no mutations above them are used to represent
         :ref:`missing data<sec_data_model_missing_data>`.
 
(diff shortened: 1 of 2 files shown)


hyanwong:
Also clarifies that `total_branch_length` calculates all branches connected to a root, not just the ones ancestral to a sample.

python/tskit/trees.py
```diff
@@ -2004,9 +2016,17 @@ def get_leaves(self, u):
 
     def leaves(self, u=None):
         """
-        Returns an iterator over all the leaves in this tree that are
-        underneath the specified node. If u is not specified, return all leaves
-        in the tree.
+        Returns an iterator over all the leaves in this tree that are underneath
+        the specified node. If :math:`u`  is not specified, return all leaves on
```
jeromekelleher, line 2020:
```suggestion
        Returns an iterator over all the leaves in this tree that descend from
        the specified node. If :math:`u`  is not specified, return all leaves on
```

python/tskit/trees.py
```diff
@@ -2004,9 +2016,17 @@ def get_leaves(self, u):
...
+        the tree (i.e. all leaves reachable from the tree root(s), see note below).
+
+        .. note::
+            :math:`u` can be any node in the entire tree sequence, including ones
+            which are not connected via branches to a root node of the tree. If
+            called on such a node, the iterator will happily return "dead" leaves
+            (see :meth:`is_leaf`) which cannot be reached from a root of this tree.
```
jeromekelleher, line 2027:
```suggestion
            called on such a node, the iterator will return the node itself, since it 
            is considered a "dead" leaf (see :meth:`is_leaf`) which cannot be reached from a root of this tree.
```

jeromekelleher (approved):
LGTM, couple of suggestions

python/tskit/trees.py
hyanwong: That's not strictly true. You could also be calling `leaves()` on a node which is on a dead section of topology. In this case it will return all the dead leaves descending from that node on the (dead) topology.

jeromekelleher: Ah, yes, you're right. OK, maybe just delete the "happily" then? This doesn't impart much information.

I  guess what you're doing in this case is manually specifying that "u" is a root for the purpose of this operation.

hyanwong:
I suggest we might want to add the following to the "data-model" section on tree encoding and roots:

```
### Dead leaves and branches

In a `tskit` tree, a *leaf node* is defined as a node without any children. The
implications of this turn out to be slighly unintuitive, and so are worth briefly
documenting here. Firstly, the same node can be a leaf in one tree, and not a leaf
in the next tree along the tree sequence. Secondly all isolated nodes must be leaves
(as by definition they have no children). Thirdly sample nodes need not be leaves
(they could be "internal samples"); likewise leaf nodes need not be samples.

Node 7 in the example above provides a good case study. Note that it is a root node
with at least one child in trees 0 and 2, but it is isolated in tree 1. Strictly,
because it is isolated in tree 1 it is also a leaf node in that tree, although it is
not attached to a root, not a sample, and is therefore not plotted. In this case, we
can think of it as a "dead leaf" (we don't normally plot dead leaves). In fact, in a
large tree sequence of many trees, most ancestral nodes will be isolated in any given
tree, and therefore most nodes will be of this sort. However, these dead leaves
are excluded from most calculations on trees, because such calculations traverse the
tree by either starting at a root and working down, or starting at a sample and working
up. Normally when we refer to the leaves of a tree, it is shorthand for the leaves
**on** the tree (that is, attached via branches, to one of the the tree roots).
Dead leaves are excluded from this definition.

Note that it is also possible to have trees in which there are "dead branches": that is
sections of topology which are not accessible from a root, and whose tips are all
dead leaves. Although valid, this is a relatively unusual state of affairs, and such
branches are not plotted by the standard :ref:`sec_tskit_viz` methods. The
:meth:`Tree.nodes` method will not, be default, traverse through dead branches, although
it can be made to do so by specifying the ID of a dead nodes as the root for traversal.
```

I can add this to the PR is you agree @jeromekelleher ?

hyanwong:
> I can add this to the PR is you agree @jeromekelleher ?

Added anyway. Executive decision. :)

jeromekelleher:
I like it, I think it's an intuitive way to explain what's going on. I say we merge.

Issue #2033, opened 2021-12-08
Add examples to Statistics page 


jeromekelleher:
It would be good to add an example of using at least some of the statistics on the Statistics page. It's a bit of a wall of text at the moment, and it's hard to get an intuitive feel for what, e.g., `allele_frequency_spectrum`` returns.

hyanwong:
To avoid duplication we could point, for example, to https://tskit.dev/tutorials/analysing_tree_sequences.html#allele-frequency-spectra in that particular case. I'm not sure on the exact balance between examples in tutorials and in the docs. ISWYM about having a few examples there though.

jeromekelleher:
It's tricky, I agree. We could try to draw the line between "explanatory" (tutorials) and "illustrative examples" (docs)?

hyanwong:
> It's tricky, I agree. We could try to draw the line between "explanatory" (tutorials) and "illustrative examples" (docs)?

Yes, a good distinction.I even wonder if we have an "examples" section up top (just after the "Available Statistic" section - and I want to move the "notes" to the end: I have a branch somewhere to PR with this) :

```
## Examples

:::{code-cell}
:"tags": ["hide-input"]
import msprime
ts = msprime.sim_ancestry(4, sequence_length=1000, recombination_rate=0.001, random_seed=1234)
ts = msprime.sim_mutations(ts, rate=0.001, random_seed=2)
print(
    "Example tree sequence has sequence length",
    int(ts.sequence_length),
    "with",
    ts.num_samples,
    "samples:",
    ts.num_trees,
    "trees &",
    ts.num_sites,
    "variable sites",
)
:::

:::{code-cell}
afs = ts.allele_frequency_spectrum(polarised=True, span_normalise=False)
print(afs) # afs[1] gives number of singletons, afs[2] doubletons, etc:::
```

PR #2055, opened 2021-12-10
Add some examples to stats

hyanwong:
And move notes to the end. Addresses some of #2033

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -124,22 +124,22 @@ Markdown(
 #### Basic calling convention
 
 ```{code-cell} ipython3
-afs = ts.allele_frequency_spectrum()
-print(afs) # afs[1] gives number of singletons, afs[2] doubletons, etc
+pi = ts.diversity()
+print(pi) # Genetic diversity within the sample set
 ```
 
 #### Restrict to {ref}`sample sets<sec_stats_sample_sets>`
 
 ```{code-cell} ipython3
-pi = ts.diversity(sample_sets=ts.samples(population=0))
-print(pi)  # Genetic diversity within population 0
+pi_0 = ts.diversity(sample_sets=ts.samples(population=0))
+print(pi_0)  # Genetic diversity within population 0
 ```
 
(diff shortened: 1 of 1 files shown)


docs/stats.md
```diff
@@ -80,8 +83,8 @@ whereas {ref}`sec_stats_sample_sets_multi_way` compare 2 or more sets of samples
 * Multi site
     * {meth}`~LdCalculator` (note this is soon to be deprecated)
 
-:::{todo}
-Add information about the new IBD stats.
+:::{seealso}
+The identity by descent methods **link**.
```
jeromekelleher, line 87: Can link to "{ref}`sec_ibd`" now can't we?

docs/stats.md
```diff
@@ -92,121 +95,73 @@ them using the general framework directly, so the versions above should be prefe
...
 
-#### Y statistics
-
-{meth}`~TreeSequence.Y3` and {meth}`~TreeSequence.Y2` are the `Y` statistics introduced
-by [Ashander et al (2018)](https://www.biorxiv.org/content/10.1101/354530v1)
-as a three-sample intermediate between diversity/divergence (which are
-pairwise) and Patterson's f statistics (which are four-way).
+#### Basic calling convention
 
+```{code-cell} ipython3
```
jeromekelleher, line 130: Is the AFS what we want as the first example? It'd be better to use something simple and more representative - the AFS doesn't look like any of the other stats.

jeromekelleher (approved):
LGTM

docs/stats.md
hyanwong: I've taken it out, as on consideration I don't think it's really "stats", but we can refer to it later I guess.

docs/stats.md
hyanwong: True. I put it first because the AFS is the basis for most single site stats. But I think you're right. Maybe segregating sites as a simple one.

hyanwong: I switched to diversity, to keep the examples similar.

PR #2057, opened 2021-12-10
Initial sketch of read-only tables implementatation

jeromekelleher:
Initial proof of concept for making the TableCollection returned by ``ts.tables`` read-only. I've only done the plumbing for the NodeTable here, but the extension is pretty clear. The implementation is pretty confusing at the moment, I think it could be simplified quite a bit.

I also had a look at what we could do in terms of returning direct read-only views of the column data. I think we could do this when the tables are in read-only model, so that ``ts.tables.nodes.time`` is an efficient operation. It doesn't look too bad, actually.

While we're breaking stuff, we should also set all column arrays that we return to being read-only, even when they are copies returned when the tables aren't in read-only mode. This will hopefully let us support the (much more difficult) direct setting of table column memory from numpy arrays, at some point.

Any thoughts?

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -66,9 +66,13 @@ static PyObject *TskitNoSampleListsError;
  * Because C code executed here represents atomic Python operations
  * (while the GIL is held), this should be safe */
 
+/* NOTE:2021-12-10: the "lock" attribute doesn't seem to do anything -
+ * where do we set it? */
+
 typedef struct _TableCollection {
     PyObject_HEAD
     tsk_table_collection_t *tables;
+    PyObject *ts;
 } TableCollection;
 
  /* The table pointer in each of the Table classes either points to locally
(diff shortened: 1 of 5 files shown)


benjeffery:
I'm trying to wrap my head round this. So when tables get made from a tree sequence the tables are marked as read only by having a reference to the ts, makes sense. 

What about the other way around when `TableCollection.tree_sequence` is called? Do we keep that as a copy operation?

jeromekelleher:
> What about the other way around when TableCollection.tree_sequence is called? Do we keep that as a copy operation?

For the first pass I think yes. The main thing we're trying to streamline here is:

```python
ts = tskit.load("bigfile.trees")
# analyse ts in lots of ways including using ts.tables.nodes.time and other column references
```

The creating-a-tree sequence-from-tables operation is much more niche I think. 

I guess we *could* mark the input TableCollection as read-only when we call ``.tree_sequence()`` on it, though.

benjeffery:
Great, was thinking that would be the first workflow to optimise, and doing just that one first makes sense.

> I guess we could mark the input TableCollection as read-only when we call .tree_sequence() on it, though.

Pretty sure this would break a lot of things and is an unexpected side-effect to make the TC read-only, we'd likely need to make copy be the default and add an argument, or have a separate method?

jeromekelleher:
We should make sure this change fixes #2080 while we're in here doing this type of stuff.

jeromekelleher:
Revisiting this in the light of #2441 and recent updates (and rebasing on them), and it looks like it basically works so we should probably try and push it through. I'll tip away at it as I have time.

This *is* a significant breaking change so we'd need a major version number bump for it.

PR #2069, opened 2021-12-14
Shorter note about span normalise in docs

hyanwong:
Because msprime now tends to default to base pair units, I think we can shorten this doc comment

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -429,15 +429,13 @@ this case, computes a statistic with the pattern of genotypes at each site,
 and normalising would divide these statistics by the distance to the previous variant site
 (probably not what you want to do).
 
-And, a final note about "span": in tree sequences produced by `msprime`
-coordinates along the sequence are **continuous**,
-so the "spans" used here may not correspond to distance along the genome in (say) base pairs.
-For instance, pairwise sequence divergence is usually a number between 0 and 1
-because it is the proportion of bases that differ;
-this will only be true if the `sequence_length`, and hence the "spans" are measured in base pairs
-(which you ensure in `msprime` by setting recombination and mutation rates equal to the values
-in units of crossovers and mutations per base pair, respectively).
-
+:::{note}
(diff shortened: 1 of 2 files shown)


docs/stats.md
```diff
@@ -429,15 +429,12 @@ this case, computes a statistic with the pattern of genotypes at each site,
...
-this will only be true if the `sequence_length`, and hence the "spans" are measured in base pairs
-(which you ensure in `msprime` by setting recombination and mutation rates equal to the values
-in units of crossovers and mutations per base pair, respectively).
-
+:::{note}
+The scaling of the genomic coordinates can affect this normalization. Take, for
+example, pairwise sequence divergence. This is usually a number between 0 and 1
+because it is the proportion of bases that differ; this will only be true if the
+`sequence_length`, and hence the "spans" are measured in base pairs.
+:::
```
petrelharp, line 437: Good spot. This is kinda confusing, since if the mutations are added with `discrete_genome` then it'll be between 0 and 1 regardless of what you think the units mean. Let's see how about:
```suggestion
:::{note}
The resulting values are in units of "per unit of sequence length" - for instance, pairwise
sequence divergence is in units of "differenes per unit of sequence length". By default,
{func}`.sim_mutations` adds mutations in discrete coordinates, and so coordinates are usually
interpreted as being measured in base pairs, in which case span normallized statistics are in
units of "per base pair".
:::
```

hyanwong:
I've reworded @petrelharp 's suggesting a little to avoid repetition:

> The resulting values are scaled "per unit of sequence length" - for instance, pairwise
> sequence divergence is measured in "differences per unit of sequence length". Functions
> such as {func}`msprime:msprime.sim_mutations` will by default add mutations in discrete
> coordinates, which are usually interpreted as a scale measured in base pairs, in which
> case span normalised statistics are in units of "per base pair".

jeromekelleher (approved):
LGTM

petrelharp:
Heh, right, thanks - got my wires crossed there.

PR #2073, opened 2021-12-16
Fix dump text

### LINKED ISSUE #1860: TreeSequence.dump_text fails if metadata schema is present
```python
import sys
ts = msprime.sim_ancestry(4, sequence_length=45, random_seed=234)
ts.dump_text(populations=sys.stdout)
```

gives
```
id      metadata
  File "tmp.py", line 14, in <module>
    ts.dump_text(populations=sys.stdout)
  File "/home/jk/work/github/tskit/python/tskit/trees.py", line 3759, in dump_text
    metadata = base64.b64encode(metadata).decode(encoding)
  File "/usr/lib/python3.8/base64.py", line 58, in b64encode
    encoded = binascii.b2a_base64(s, newline=False)
TypeError: a bytes-like object is required, not 'dict'
```

I guess the right behaviour here is to make the ``base64_metadata`` option default to None, and to output decoded metadata as JSON if a schema is present (by default) and base64 encode it otherwise. If base64_metadata is True, then we output the raw bytes always.

We should move the implementation of this function to ``text_formats.py`` while we're fixing it up too, to try and downsize trees.py a bit and keep all the text processing stuff in one place.

benjeffery:
Fixes #1860

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -307,9 +307,6 @@ def get_example_tree_sequences(back_mutations=True, gaps=True, internal_samples=
     ts = msprime.sim_ancestry(
         8, sequence_length=40, recombination_rate=0.1, random_seed=seed
     )
-    # FIXME various tests in this file assume bytes as metadata. Also
-    # see this bug in the text file format:
-    # https://github.com/tskit-dev/tskit/issues/1860
     tables = ts.dump_tables()
     tables.populations.metadata_schema = tskit.MetadataSchema(None)
     ts = tables.tree_sequence()
@@ -326,6 +323,7 @@ def get_example_tree_sequences(back_mutations=True, gaps=True, internal_samples=
     tables.edges.clear()
     yield tables.tree_sequence()  # empty tree
     yield tskit.TableCollection(sequence_length=1).tree_sequence()  # empty tree seq
+    yield tsutil.all_fields_ts()
 
 
 def get_bottleneck_examples():
@@ -1429,8 +1427,13 @@ class TestTreeSequence(HighLevelTestCase):
                 assert edge.span == edge.right - edge.left
 
     def test_edgesets(self):
+        tested = False
         for ts in get_example_tree_sequences():
-            self.verify_edgesets(ts)
+            # Can't get edgesets with metadata
+            if ts.tables.edges.metadata_schema == tskit.MetadataSchema(None):
+                self.verify_edgesets(ts)
+                tested = True
+        assert tested
 
     def test_breakpoints(self):
         for ts in get_example_tree_sequences():
@@ -1480,7 +1483,7 @@ class TestTreeSequence(HighLevelTestCase):
             python_time = tsutil.compute_mutation_times(ts)
             tables.compute_mutation_times()
             assert np.allclose(
-                python_time, tables.mutations.time, rtol=1e-15, atol=1e-15
+                python_time, tables.mutations.time, rtol=1e-10, atol=1e-10
             )
             # Check we have valid times
             tables.tree_sequence()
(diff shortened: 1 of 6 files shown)


python/tests/test_highlevel.py
```diff
@@ -2719,34 +2743,76 @@ def convert(v):
...
+    ):
+        """
+        Verifies that the populations we output have the correct form.
+        """
+
+        def convert(v):
+            return "{:.{}f}".format(v, precision)
+
+        output_populations = populations_file.read().splitlines()
+        print(output_populations)
```
jeromekelleher, line 2762: stray print

python/tests/test_highlevel.py
```diff
@@ -2719,34 +2743,76 @@ def convert(v):
...
-                edges_file.seek(0)
-                sites_file.seek(0)
-                mutations_file.seek(0)
-                individuals_file.seek(0)
-                self.verify_nodes_format(ts, nodes_file, precision)
-                self.verify_edges_format(ts, edges_file, precision)
-                self.verify_sites_format(ts, sites_file, precision)
-                self.verify_mutations_format(ts, mutations_file, precision)
-                self.verify_individuals_format(ts, individuals_file, precision)
+                for base64_metadata in [True, False]:
```
jeromekelleher, line 2779: I don't think there's any point in taking the cross product of precision and base64_metadata, it's not really testing anything useful. I think we're going to need to move away from the old "sledgehammer" approach to testing, and be more selective about how we do things (and more obvious, using examples).

Feels like a little bit of light refactoring here and pytest.mark.parametrize would help? I've been trying to modernize test methods that I encounter as I go, otherwise it'll never get done.

python/tskit/text_formats.py
```diff
@@ -248,3 +250,181 @@ def build_newick(tree, *, root, precision, node_labels, include_branch_lengths):
...
+            "time",
+            "population",
+            "individual",
+            "metadata",
+            sep="\t",
+            file=nodes,
+        )
+        for node in ts.nodes():
+            metadata = node.metadata
+            if isinstance(metadata, bytes) and base64_metadata:
```
jeromekelleher, line 281: Should probably make a function for this? Not point in duplicating this logic n times.
```
def get_text_metadata(metadata, base64metadata):
     if isinstance ...
         metadata = base64.b64encode(metadata).decode(encoding)
     return metadata
```

jeromekelleher (commented):
Looks good. The testing for this stuff is a mess though, I think it's probably worth taking the opportunity to fix it up.

benjeffery:
@jeromekelleher All fixed up!

python/tests/test_highlevel.py
```diff
@@ -1834,20 +1835,22 @@ def verify_tables_api_equality(self, ts):
...
-                sample_sizes |= {1}
-            if n > 2:
-                sample_sizes |= {2, max(2, n // 2), n - 1}
-            for k in sample_sizes:
-                subset = random.sample(list(ts.samples()), k)
-                self.verify_simplify_topology(ts, subset)
-                self.verify_simplify_equality(ts, subset)
-                self.verify_simplify_variants(ts, subset)
+            # Can't simplify edges with metadata
+            if ts.tables.edges.metadata_schema == tskit.MetadataSchema(schema=None):
```
jeromekelleher, line 1839: This isn't a great pattern I think - I've opened #2079 to track. We shouldn't have to create a new metadata schema object to determine whether the current one is set or not.

python/tskit/text_formats.py
```diff
@@ -248,3 +250,166 @@ def build_newick(tree, *, root, precision, node_labels, include_branch_lengths):
         include_branch_lengths=include_branch_lengths,
     )
     return s + ";"
+
+
+def dump_text(
+    ts,
```
jeromekelleher, line 256: Most of these should be keyword only args I think

jeromekelleher (approved):
LGTM, one comment.

benjeffery:
Hmm, got some unrelated failing tests here. Looking into it.

PR #2099, opened 2022-01-17
Refactor vargen to store state (except tree) in variant struct

### LINKED ISSUE #2083: Refactor variant decoding state into variant_t struct
#605 is a top level issue saying that we want to have a way to get genotypes from a variant without using the vargen_t struct.

A step along the way to this goal is to refactor the vargen_t struct so that all the state (except for the tree used for iteration) is maintained in the ``variant_t`` struct. We should be able to do this without breaking any client code.

See #1912 for reference also, which does some quick hacks along this line.

benjeffery:
Fixes #2083 

This still has a small change to the tests, but I think that can be fixed, hence this is a draft.

### DIFF
--- a/c/tskit/genotypes.c
+++ b/c/tskit/genotypes.c
@@ -35,25 +35,24 @@
  * ======================================================== */
 
 void
-tsk_vargen_print_state(const tsk_vargen_t *self, FILE *out)
+tsk_variant_print_state(const tsk_variant_t *self, FILE *out)
 {
     tsk_size_t j;
 
-    fprintf(out, "tsk_vargen state\n");
-    fprintf(out, "tree_index = %lld\n", (long long) self->tree.index);
-    fprintf(out, "tree_site_index = %lld\n", (long long) self->tree_site_index);
+    fprintf(out, "tsk_variant state\n");
     fprintf(out, "user_alleles = %lld\n", (long long) self->user_alleles);
-    fprintf(out, "num_alleles = %lld\n", (long long) self->variant.num_alleles);
-    for (j = 0; j < self->variant.num_alleles; j++) {
-        fprintf(out, "\tlen = %lld, '%.*s'\n",
-            (long long) self->variant.allele_lengths[j],
-            (int) self->variant.allele_lengths[j], self->variant.alleles[j]);
+    fprintf(out, "num_alleles = %lld\n", (long long) self->num_alleles);
+    for (j = 0; j < self->num_alleles; j++) {
+        fprintf(out, "\tlen = %lld, '%.*s'\n", (long long) self->allele_lengths[j],
+            (int) self->allele_lengths[j], self->alleles[j]);
     }
     fprintf(out, "num_samples = %lld\n", (long long) self->num_samples);
-    for (j = 0; j < tsk_treeseq_get_num_nodes(self->tree_sequence); j++) {
-        fprintf(out, "\t%lld -> %lld\n", (long long) j,
-            (long long) self->sample_index_map[j]);
-    }
+}
+
+void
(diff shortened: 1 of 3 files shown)


c/tskit/genotypes.c
```diff
@@ -35,25 +35,28 @@
...
 void
-tsk_vargen_print_state(const tsk_vargen_t *self, FILE *out)
+tsk_variant_print_state(const tsk_variant_t *self, FILE *out)
 {
     tsk_size_t j;
 
-    fprintf(out, "tsk_vargen state\n");
-    fprintf(out, "tree_index = %lld\n", (long long) self->tree.index);
-    fprintf(out, "tree_site_index = %lld\n", (long long) self->tree_site_index);
+    fprintf(out, "tsk_vatiant state\n");
```
jeromekelleher, line 42: s/vatiant/variant

c/tskit/genotypes.c
```diff
@@ -149,45 +152,30 @@ vargen_init_samples_and_index_map(tsk_vargen_t *self, const tsk_treeseq_t *tree_
...
         self->num_samples = num_samples;
-        self->sample_index_map = self->alt_sample_index_map;
     }
-    self->num_sites = tsk_treeseq_get_num_sites(tree_sequence);
-    self->tree_sequence = tree_sequence;
+
     self->options = options;
+
     if (self->options & TSK_16_BIT_GENOTYPES) {
-        self->variant.genotypes.i16
```
jeromekelleher, line 185: Not clear why we're not still allocating genotypes here?

c/tskit/genotypes.c
```diff
@@ -623,7 +617,75 @@ tsk_vargen_update_site(tsk_vargen_t *self)
...
+}
+
+int
+tsk_tree_get_variant(tsk_tree_t *self, const tsk_site_t *site, tsk_variant_t *variant,
+    tsk_flags_t options)
+{
+    int ret = 0;
+    tsk_size_t num_samples_alloc;
+
+    variant->tree = self;
```
jeromekelleher, line 632: There's too much internal state manipulation of the variant/vargen going on here - why isn't the allocation being done in ``variant_init``? We're doing O(n) operations and allocs here which we should be avoiding if possible.

c/tskit/genotypes.c
```diff
@@ -623,7 +617,75 @@ tsk_vargen_update_site(tsk_vargen_t *self)
...
         num_missing -= (tsk_size_t) no_longer_missing;
     }
-    var->has_missing_data = num_missing > 0;
+    self->has_missing_data = num_missing > 0;
+out:
+    return ret;
+}
+
+int
+tsk_tree_get_variant(tsk_tree_t *self, const tsk_site_t *site, tsk_variant_t *variant,
```
jeromekelleher, line 626: Do we need this method for the first pass? We just want to refactor the vargen for now, keeping the same semantics and performance as before, right?

jeromekelleher (commented):
Excellent progress, this is working out quite nicely. I don't see why we need the tree_get_variant method though, and we should definitely keep the existing pattern of mallocs when running through the variants using ``vargen``.

The changes to the tests are totally fine, I see no problem with that.

benjeffery:
@jeromekelleher You're right, I was trying to do two steps at once and got a lot of the init stuff in the wrong place thinking I needed a tree. I've moved it around so that a lot more is done in `variant_init`.

benjeffery:
Getting a weird codecov error here. Investigating.

benjeffery:
> Dang, need a little more coverage.

Actually, looks like error states, so all good and ready for review.

benjeffery:
@jeromekelleher good spots! Fixed up.

PR #2104, opened 2022-01-21
Make null strings decode to empty object in JSON codec

### LINKED ISSUE #2064: empty metadata should be allowed by the JSON codec
As we (@petrelharp) discovered in https://github.com/MesserLab/SLiM/issues/258, the `struct` codec allows zero-length metadata to indicate "no metadata", but the `JSON` codec does not; it insists upon a 4-byte "null" being present.  Not the end of the world, but seems a bit silly.  If you have a big table with many rows, and its schema says `JSON`, but you don't want to attach any metadata to it, you have to put "nullnullnull......." into the metadata.  I don't feel _strongly_ that this is wrong, but maybe it's wrong?

This came up for the population table in SLiM, with this schema:

`$schema":"http://json-schema.org/schema#","additionalProperties":true,"codec":"json","description":"SLiM schema for population metadata.","examples":[{"bounds_x0":0.0,"bounds_x1":100.0,"bounds_y0":0.0,"bounds_y1":100.0,"female_cloning_fraction":0.25,"male_cloning_fraction":0.0,"migration_records":[{"migration_rate":0.9,"source_subpop":1},{"migration_rate":0.1,"source_subpop":2}],"selfing_fraction":0.5,"sex_ratio":0.5,"slim_id":2,"name":"p2"}],"properties":{"bounds_x0":{"description":"The minimum x-coordinate in this subpopulation.","type":"number"},"bounds_x1":{"description":"The maximum x-coordinate in this subpopulation.","type":"number"},"bounds_y0":{"description":"The minimum y-coordinate in this subpopulation.","type":"number"},"bounds_y1":{"description":"The maximum y-coordinate in this subpopulation.","type":"number"},"bounds_z0":{"description":"The minimum z-coordinate in this subpopulation.","type":"number"},"bounds_z1":{"description":"The maximum z-coordinate in this subpopulation.","type":"number"},"description":{"description":"A description of this subpopulation.","type":"string"},"female_cloning_fraction":{"description":"The frequency with which females in this subpopulation reproduce clonally (for WF models).","type":"number"},"male_cloning_fraction":{"description":"The frequency with which males in this subpopulation reproduce clonally (for WF models).","type":"number"},"migration_records":{"items":{"properties":{"migration_rate":{"description":"The fraction of children in this subpopulation that are composed of 'migrants' from the source subpopulation (in WF models).","type":"number"},"source_subpop":{"description":"The ID of the subpopulation migrants come from (in WF models).","type":"integer"}},"required":["source_subpop","migration_rate"],"type":"object"},"type":"array"},"name":{"description":"A human-readable name for this subpopulation.","type":"string"},"selfing_fraction":{"description":"The frequency with which individuals in this subpopulation self (for WF models).","type":"number"},"sex_ratio":{"description":"This subpopulation's sex ratio (for WF models).","type":"number"},"slim_id":{"description":"The ID of this population in SLiM. Note that this is called a 'subpopulation' in SLiM.","type":"integer"}},"required":["slim_id"],"type":["object","null"]}`

The relevant part of which is the `"type":["object","null"]` at the end.  We allow "null", with the intended meaning of "empty metadata", but are required to actually put in "null" rather than just being allowed to leave it empty.  Is that really necessary?

benjeffery:
Fixes #2064

### DIFF
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -503,6 +503,7 @@ class TestJSONCodec:
             "properties": {"number": {"type": "number", "default": 5}},
         }
         ms = tskit.MetadataSchema(schema)
+        assert ms.decode_row(b"") == {"number": 5}
         assert ms.decode_row(ms.validate_and_encode_row({})) == {"number": 5}
         assert ms.decode_row(ms.validate_and_encode_row({"number": 42})) == {
             "number": 42
@@ -571,6 +572,10 @@ class TestJSONCodec:
         with patch.object(ms, "_validate_row", return_value=True) as mocked_validate:
             ms.validate_and_encode_row({"int": 1})
             assert mocked_validate.call_count == 1
+
+    def test_zero_length(self):
+        ms = tskit.MetadataSchema({"codec": "json"})
+        assert ms.decode_row(b"") == {}
 
 
 class TestStructCodec:
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -122,7 +122,9 @@ the human readable [JSON](https://www.json.org/json-en.html) format. As this for
 is human readable and encodes numbers as text it uses more bytes than the `struct`
 format. However it is simpler to configure as it doesn't require any format specifier
 for each type in the schema. Default values for properties can be specified for only
-the shallowest level of the metadata object.
+the shallowest level of the metadata object. Tskit deviates from standard JSON in that
+empty metadata is interpreted as an empty object. This is to allow setting of a schema
+to a table with out the need to modify all existing empty rows.
 
 
 ### struct
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -7,6 +7,12 @@
 - ``VcfWriter.write`` now prints the site ID of variants in the ID field of the output VCF files.
   (:user:`roohy`, :issue:`2103`, :pr:`2107`)
 
+**Breaking Changes**
+
+- The JSON metadata codec now interprets the empty string as an empty object. This means
+  that applying a schema to an existing table will no longer necessitate modifying the
+  existing rows. (:user:`benjeffery`, :issue:`2064`, :pr:`2104`)
+
 
 ----------------------
 [0.4.1] - 2022-01-11
--- a/python/lwt_interface/dict_encoding_testlib.py
+++ b/python/lwt_interface/dict_encoding_testlib.py
@@ -27,7 +27,6 @@ compiled module exporting the LightweightTableCollection class.
 See the test_example_c_module file for an example.
 """
 import copy
-import json
 
 import kastore
 import msprime
@@ -222,8 +221,7 @@ class TestMissingData:
         lwt.fromdict(d)
         tables = tskit.TableCollection.fromdict(lwt.asdict())
         # Empty byte field still gets interpreted by schema
-        with pytest.raises(json.decoder.JSONDecodeError):
-            tables.metadata
+        assert tables.metadata == {}
 
     def test_missing_metadata_schema(self, tables):
         assert repr(tables.metadata_schema) != ""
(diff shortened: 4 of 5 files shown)


python/tests/test_metadata.py
```diff
@@ -572,6 +572,10 @@ def test_dont_skip_validation_other_codecs(self):
             ms.validate_and_encode_row({"int": 1})
             assert mocked_validate.call_count == 1
 
+    def test_zero_length(self):
+        ms = tskit.MetadataSchema({"codec": "json"})
+        assert ms.decode_row(b"") == {}
+
```
jeromekelleher, line 579: We're not currently testing the interaction between the empty metadata fields and the defaults. Can you add a few tests with examples to make this explicit please? (This could easily be implemented in a way that bypasses adding the default values into the result)

jeromekelleher (approved):
LGTM, I think this is a win.

We should get more votes on this though, as it's a potentially breaking change. @petrelharp @molpopgen would you mind taking a quick look please?

benjeffery:
@jeromekelleher Test added

molpopgen:
> LGTM, I think this is a win.
> 
> We should get more votes on this though, as it's a potentially breaking change. @petrelharp @molpopgen would you mind taking a quick look please?

I'm afraid that I haven't been following this one.  But in fwdpy11 I either use the codec struct for metadata and the one table where I use JSON has no empty metadata allowed (I think).  In tskit-rust, empty metadata is already handled by returning a None type.  My only question would be regarding top-level metadata, but returning an empty object seems fine there (for my purposes).  I do not allow top-level metadata to be empty upon export from fwdpy11.

jeromekelleher:
Thanks @molpopgen, once question:

>  In tskit-rust, empty metadata is already handled by returning a None type.

Does this conflict with our proposed behaviour of returning the empty object? I guess it must be inconsistent with the idea of defining defaults on top of empty metadata strings like we do here? For example, if your schema has an attribute "a" with default = "abc", then running ``schema.decode(b"")`` or ``schema.decode(b"{}")``will both return ``{"a": "abc"}``.

molpopgen:
One question did come to mind about these changes. For the struct codec, I am assuming that one receives None when the metadata are empty. That is not changing? If that does change, then it is a breaking change for me.

On the rust side, I see no conflict. Default parameters don't always make sense for compiled languages. We could accommodate such a thing by requiring that metadata compatible types implement the Default trait, but I see little reason to do so. To me, a None is much more clear than to have to check if an object instance is equal to the default value.

benjeffery:
> . For the struct codec, I am assuming that one receives None when the metadata are empty. That is not changing? If that does change, then it is a breaking change for me.

This is correct, not changing and also needed for compatibility with SLIm.

jeromekelleher:
Having a quick look through the Rust metadata with @benjeffery, it seems like the Rust metadata mechanisms are concerned with serialising/deserialising data structures **within** rust rather than being concerted with decoding arbitrary metadata.

What would happen if we tried to read, say, the population metadata from an msprime tree sequence in Rust @molpopgen? Would you just get back a bunch of bytes, or would it try to give something like a Python dictionary of the contents?

molpopgen:
> Having a quick look through the Rust metadata with @benjeffery, it seems like the Rust metadata mechanisms are concerned with serialising/deserialising data structures **within** rust rather than being concerted with decoding arbitrary metadata.
> 
> What would happen if we tried to read, say, the population metadata from an msprime tree sequence in Rust @molpopgen? Would you just get back a bunch of bytes, or would it try to give something like a Python dictionary of the contents?

You would need to do the work of defining a struct on the rust side to represent the metadata and then use the procedural macros to register that type as being serialized using JSON.  Then that should all work, but I've not tested anything.  The number of lines of code should be something like the number of metadata fields plus two, where the extra two are the struct name and the metadata registration macro call.

You are not able to call the functions to retrieve metadata unless you provide the name of such a struct.

molpopgen:
In principle one could add a function to get the metadata as bytes. But such an interface immediately implies using unsafe code blocks and all that that entails. Raw bytes from C are not compatible with rust strings, and the types used to handle them all required code blocks marked as unsafe. these concerns are why I have not added such interface.

jeromekelleher:
OK, sounds good. So, it sounds like these differences between rust and Python and really differences, so there's no point in worrying about it. 

I say we merge!

molpopgen:
Yes, these are real language differences that force API differences.  Strong static typing and trying to prevent insecure code makes rust different.

benjeffery:
Merging!

PR #2108, opened 2022-01-27
32bit genotypes

### LINKED ISSUE #463: [C API] Change vargen genotypes (and elsewhere) to int32?
The vargen struct is complicated by having a genotypes.i8 and genotypes.i16 as well as the complexity we get in the code from changing between the two.

If we're only looking at a single site, then there's probably no point in this: it's unlikely we save any real time or space by having the int8 type. It's only when we're storing a genotype **matrix** that this will really matter, but that's not something that we should be doing really.

Probably we could percolate this change right through the Python API also without affecting users.

benjeffery:
Fixes #463

WIP, getting some odd, unrelated test failures. Trying on CI.

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -2,7 +2,14 @@
 [0.99.16] - 2022-0X-XX
 ----------------------
 
-- Make dumping of tables and tree seqences to disk a zero-copy operation.
+**Breaking changes**
+
+- Change the type of genotypes to ``int32_t``, removing the TSK_16_BIT_GENOTYPES flag option.
+  (:user:`benjeffery`, :issue:`463`, :pr:`2108`)
+
+**Features**
+
+- Make dumping of tables and tree sequences to disk a zero-copy operation.
   (:user:`benjeffery`, :issue:`2111`, :pr:`2124`)
 
 ----------------------
--- a/c/tests/test_genotypes.c
+++ b/c/tests/test_genotypes.c
@@ -49,20 +49,8 @@ test_simplest_missing_data(void)
     CU_ASSERT_EQUAL_FATAL(ret, 1);
     CU_ASSERT_EQUAL(var->site->position, 0.0);
     CU_ASSERT_TRUE(var->has_missing_data);
-    CU_ASSERT_EQUAL(var->genotypes.i8[0], TSK_MISSING_DATA);
-    CU_ASSERT_EQUAL(var->genotypes.i8[1], TSK_MISSING_DATA);
-    ret = tsk_vargen_next(&vargen, &var);
-    CU_ASSERT_EQUAL_FATAL(ret, 0);
-    tsk_vargen_free(&vargen);
-
-    ret = tsk_vargen_init(&vargen, &ts, NULL, 0, NULL, TSK_16_BIT_GENOTYPES);
-    CU_ASSERT_EQUAL_FATAL(ret, 0);
-    ret = tsk_vargen_next(&vargen, &var);
-    CU_ASSERT_EQUAL_FATAL(ret, 1);
-    CU_ASSERT_EQUAL(var->site->position, 0.0);
-    CU_ASSERT_TRUE(var->has_missing_data);
-    CU_ASSERT_EQUAL(var->genotypes.i16[0], TSK_MISSING_DATA);
-    CU_ASSERT_EQUAL(var->genotypes.i16[1], TSK_MISSING_DATA);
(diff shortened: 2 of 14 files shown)


jeromekelleher:
Might be worth doing a check perf check to see what promoting up to 32 bit before going through the painful stuff? Where should be a quick s/i8/i32 change you could do globally that would be easy.

benjeffery:
> Might be worth doing a check perf check to see what promoting up to 32 bit before going through the painful stuff? Where should be a quick s/i8/i32 change you could do globally that would be easy.

In tracing all the types we also have to change to 32 in map_mutations so doing the lot and then testing that and genotypes for perf.

bhaller:
If I understand what's being proposed, this should be fine.  We only use vargen_t for crosschecking the tskit data structures against SLiM (i.e., in debug mode) and when loading a .trees file from disk, I believe.  In all places where we currently call tsk_vargen_init() we pass TSK_16_BIT_GENOTYPES; we don't use 8-bit.  Is 16-bit still going to be supported, or is it also going away in favor of 32-bit?  Is the TSK_16_BIT_GENOTYPES flag going to be deprecated, or honored, or just ignored?  Anyhow, I could dig into the diffs here to answer my own questions, but this really ought to be no problem for us, if I'm not totally misunderstanding what is at issue.

benjeffery:
Thanks for the questions @bhaller - under the current proposal the 16bit option will be removed and the flag will be ignored, hence why we want to get this sorted for C API 1.0.

benjeffery:
Some testing:
Screenshot from 2022-02-04 10-59-47


Python:
```
for v in ts.variants():
    v.genotypes
```
This branch:  129s
Main: 134s

C:
```
    ret = tsk_treeseq_load(&ts, "perf.ts", 0);
    ret = tsk_vargen_init(&vargen, &ts, samples, 2, NULL, 0);
    while (tsk_vargen_next(&vargen, &var) == 1){
    };
```
This branch: 312s
Main: 312s

What I don't understand is why the python is so much faster. Must be a mistake right? Looking into it.

benjeffery:
Ok think I have sorted my issues out.
Looks like the 32bit genotypes branch is faster by 2-3%.

benjeffery:
No noticeable impact on ram, as you would expect for iterating. Just needs a changelog.

bhaller:
> Thanks for the questions @bhaller - under the current proposal the 16bit option will be removed and the flag will be ignored, hence why we want to get this sorted for C API 1.0.

OK, good to know.  Full steam ahead; will be a trivial fix when we decide to merge the next tskit version, I think.  Thanks.

benjeffery:
@jeromekelleher What sort of info did you want from `perf`?

Here's the summary:
`main`:
```
         43,598.48 msec task-clock                #    1.000 CPUs utilized          
               594      context-switches          #    0.014 K/sec                  
                 0      cpu-migrations            #    0.000 K/sec                  
            32,826      page-faults               #    0.753 K/sec                  
   198,588,349,911      cycles                    #    4.555 GHz                    
   215,228,429,541      instructions              #    1.08  insn per cycle         
    19,162,813,346      branches                  #  439.529 M/sec                  
           778,793      branch-misses             #    0.00% of all branches        

      43.606464362 seconds time elapsed

      43.554717000 seconds user
       0.044002000 seconds sys
```
`branch`:
```
         45,034.81 msec task-clock                #    1.000 CPUs utilized          
               130      context-switches          #    0.003 K/sec                  
                 0      cpu-migrations            #    0.000 K/sec                  
            33,219      page-faults               #    0.738 K/sec                  
   205,385,182,502      cycles                    #    4.561 GHz                    
   234,245,790,040      instructions              #    1.14  insn per cycle         
    19,162,553,042      branches                  #  425.505 M/sec                  
           795,300      branch-misses             #    0.00% of all branches        

      45.036363412 seconds time elapsed

      44.991126000 seconds user
       0.043999000 seconds sys
```
The flame graph from both is almost 100% `tsk_variant_update_site`

Screenshot from 2022-02-07 12-46-59

jeromekelleher:
Looks great! The first ~10 lines of the ``perf report`` screen for both is the most informative I think.

jeromekelleher:
OK, so there's a very slight increase in the time spent in ``tsk_variant_update_genotypes_sample_list``, but it's within the range we'd expect for statistical noise. 

I'm sold then, let's make the change!

benjeffery:
@jeromekelleher Need an approving review here for the merge.

jeromekelleher (approved):

Issue #1315, opened 2021-04-16
add statistic to compute density of coalescences through time


petrelharp:
This is also used to look for selection.  See also #1314 - maybe this is related, as each coalescence is the end of a linage.

nspope:
I have been doing this with topological constraints -- e.g. calculating density of coalescence events between two tips, conditional on neither first coalescing with an outgroup -- and this gives rise to a set of statistics that are informative about time-varying migration rates. If this twist is of interest and is not already in progress, would be happy to put together a PR.

petrelharp:
That'd be *excellent*.

nspope:
Great, then I'll take a crack at it & will try to model after the time-varying GNN API.

This is tangential but provides some context: the application of these statistics is in the spirit of RELATE's way of calculating effective population size (e.g. simply inverting pairwise coalescence density to get Ne), but extended to the multi-deme case. Say we have two demes, A and B, then there are six possible topologies of three (labelled) tips: ((A,A),B), ((A,B),B), etc. Under a piecewise-constant demography where Ne/migration rates vary over epochs, the predicted density of first coalesence events for each possible topology is solvable via standard methods for continuous-time Markov processes. These can be fit to the observed density (via a multinomial pmf) to get estimates of Ne/migration matrix per epoch. It turns out that this is a sort of nonlinear state-space model, so is well-suited for particle filters, etc.

petrelharp:
We discussed this in the meeting today and got enthusiastic about it, especially @janaobsteter, @gregorgorjanc, @awohns.

Also, Chuck Langley suggested providing an easy way of telling if the data are consistent with a single-Ne model (for instance; plot within- and between-coalescence rates).

petrelharp:
We might get moving on this soon, @nspope? What's your status here?

nspope:
Hi @petrelharp, I have a python mock-up that I can put into a PR when I have a minute later this week. The implementation is straightforward for arbitrary pairs of tips -- e.g. no topological constraints, such as conditioning on neither tip first coalescing with an outgroup -- but gets complicated when trying to add arbitrary topological constraints.

Anyway it should serve to get the conversation going.

nspope:
@gregorgorjanc I agree, I think we can start with a simple, general implementation and go from there. I'm happy to put that together based on what I already have, which was for the (more specific) application I described.

gregorgorjanc:
[...]

PR #2119, opened 2022-02-04
Density of coalescences / effective population size estimates

nspope:
WIP related to #1315. Edited 14 Feb 22 into something resembling a usable API. @petrelharp @gregorgorjanc

This mock-up consists of a class in `test_stats.py` that: (a) walks over nodes in marginal trees; (b) assigns nodes to "blocks" of adjacent trees that can be easily resampled; (c) calculates weights for node times according to a user-provided function -- for example, calculating the number of pairs/trios of a particular population labelling that coalesce in a particular node -- then sums these across trees for each unique node/block combination; (d) provides member functions for the ecdf of the resulting distribution of event times, and a simple estimator of average event rate in time intervals; (e) provides a generator for  bootstrapping that reweights nodes by resampling blocks of trees. Conceptually, this is a concise representation of a distribution of event times; the distribution of what exactly depends on the choice of how nodes are weighted. Special cases include pairwise coalescence times, pairwise cross-coalescence times, and first coalescence times of trios/quartets. Applications for these include the fitting of single- or multi-population demographic models with time-varying Ne/migration, but they are likely interesting in their own right as summary statistics (e.g. in sliding windows across the genome).

This could easily be refactored as a standalone function (still allowing arbitrary weighting) or several functions (with defined weights, e.g. for pairwise coalescence). The reason it's written as a class is to allow quick recomputation of coalescence rates for different discritizations of time, or fast querying of specific event times w/ the accompanying weight. This allows some interactivity in selecting the resolution of the time discritization, without having to pass over the marginal trees each time. The idea is that the reduction to a single record per unique node id keeps the storage requirements low. But this design might not translate very easily to local estimates on genomic windows.

### usage example
``` python
sample_sets = [population1, population2, population3] # distinct sets of samples
time_distr = CoalescenceTimeDistribution(ts, sample_sets, 
                weight_func = count_leaf_pairs, num_blocks = 100)

time_grid = numpy.linspace(0., max_time, num_time_points)

# returns array with shape ( length_of_weighting_function_output , num_time_points )
time_distr.ecdf(time_grid)
time_distr.num_coalesced(time_grid)
time_distr.num_uncoalesced(time_grid)

# returns array with shape ( length_of_weighting_function_output , num_time_points - 1)
# e.g. "time_grid" is used as breaks for windows wherein time-average rates are calculated
time_distr.interval_coalescence_rate(time_grid)
# for this choice of weight_func, the inverse of the coalescence rate is an estimate of haploid Ne
1./time_distr.interval_coalescence_rate(time_grid)
 
# generates copies with node weights adjusted according to a block resampling scheme
time_distr_boot = time_distr.block_bootstrap(num_replicates, random_seed)
next(time_distr_boot).interval_coalescence_rate(time_grid) # resampled summary statistics
```
where
``` python
def count_leaf_pairs(nd, tr, sample_sets):
    ss = sample_sets
    wts = np.array([0]*len(ss), dtype="int32")
    if tr.is_internal(nd):
       # number of descendants per child across sample sets
       ch = tr.children(nd)
       nd = [[len([j for j in tr.leaves(c) if j in s]) for s in ss] for c in ch]
       # weights are the number of pairs with a TMRCA at node
       for s in range(len(ss)):
           for i in range(len(ch)-1):
               for j in range(i+1, len(ch)):
                   wts[s] += nd[i][s] * nd[j][s]
    return wts
```
is one possible weighting function, that returns the number of pairs of samples that coalesce at the node across different sets of samples (e.g. populations). More complex choices could include counting of topologies matching a particular population labeling and position of focal node, for example counting embedded topologies of the form `((A,B):node,C)` where `A,B,C` are population labels and `node` is the node ID.

### Ne estimation and confidence intervals
This only really applies to pairwise coalescence rates, but could serve as the starting point for a tutorial. Assume a single population with piecewise-constant effective population size through time. The `i`th epoch has duration `d_i` and effective population size `n_i`. The probability that a pair of lineages will coalesce by the end of epoch `i` is `p_i = 1 - exp(-d_i / n_i)` (conditional on the pair having not already coalesced by the beginning of the epoch). Let there be `k_i` independent pairs of lineages that haven't coalesced by the start of epoch `i` -- for example, genealogies for independently segregating tracts of DNA that have been split by non-overlapping pairs of samples. Let the number of pairs that coalesce during epoch `i` be `x_i`. Because sample pairs/genealogies are independent, `x_i` follows a binomial distribution with size `k_i` and probability `p_i`. Then, the MLE for `n_i` is `-d_i/log(1 - x_i/k_i)`. Its standard error can be calculated from by twice-differentiating the binomial pmf with regard to `n_i`, evaluating at the MLE, etc. Confidence intervals could then be calculated using the normal approximation, or by inverting the endpoints of a more accurate binomial confidence interval (because there's a 1-to-1 relationship between `n_i` and `p_i`).

### Validation
Here are single-population Ne trajectories estimated using the method described above, for 100 simulations from a "sin wave" demography. These are calculated using `x_i, k_i` summed over 10 pairs of samples. Each simulation uses 10000 independent genealogies (so `k_0` is 100000 for each simulation). The black line is the truth, colors are estimates across simulations.

Here is bias / RMSE / confidence interval coverage from the same set of simulations, summing `x_i, k_i` across
 different numbers of sample pairs. The estimates are unbiased regardless of the number of sample pairs used, but are more accurate with more sample pairs. For estimates calculated with a single sample pair, confidence intervals have the desired 95% coverage. However coverage declines rapidly when more sample pairs are used, especially for older windows. I think this is because older coalescence events get counted multiple times like described my comment below, so the binomial model no longer applies. Block bootstrapping produces CIs with much better coverage (not shown).

### DIFF
--- a/python/tskit/stats.py
+++ b/python/tskit/stats.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2021 Tskit Developers
+# Copyright (c) 2018-2022 Tskit Developers
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -22,12 +22,16 @@
 """
 Module responsible for computing various statistics on tree sequences.
 """
+import copy
+import itertools
 import sys
 import threading
 
 import numpy as np
 
 import _tskit
+import tskit
+import tskit.trees as trees
 
 
 class LdCalculator:
@@ -163,3 +167,511 @@ class LdCalculator:
             A[j, j + 1 :] = a
             A[j + 1 :, j] = a
         return A
+
+
+class CoalescenceTimeTable:
+    """
+    Container for sorted coalescence times, weights, and block assignments.
+    """
+
+    def __init__(self, time, block, weights):
+        assert time.shape[0] == weights.shape[0] == block.shape[0]
+        assert time.ndim == block.ndim == 1
+        assert weights.ndim == 2
+        self.num_weights = weights.shape[1]
+        # remove empty records
+        not_empty = np.sum(weights, 1) > 0
+        self.num_records = sum(not_empty)
+        self.time = time[not_empty]
+        self.block = block[not_empty]
+        self.weights = weights[not_empty, :]
+        # add left boundary at time 0
+        self.num_records += 1
+        self.time = np.pad(self.time, (0, 1))
+        self.block = np.pad(self.block, (0, 1))
+        self.weights = np.pad(self.weights, ((0, 1), (0, 0)))
+        # sort by node time
+        time_order = np.argsort(self.time)
+        self.time = self.time[time_order]
+        self.block = self.block[time_order]
+        self.weights = self.weights[time_order, :]
+        # calculate quantiles
+        self.num_blocks = 1 + np.max(self.block) if self.num_records > 0 else 0
+        self.block_multiplier = np.ones(self.num_blocks)
+        self.cum_weights = np.cumsum(self.weights, 0)
+        self.quantile = np.empty((self.num_records, self.num_weights))
+        self.quantile[:] = np.nan
+        for i in range(self.num_weights):
+            if self.cum_weights[-1, i] > 0:
+                self.quantile[:, i] = self.cum_weights[:, i] / self.cum_weights[-1, i]
+
+    def resample_blocks(self, block_multiplier):
+        assert block_multiplier.shape[0] == self.num_blocks
+        assert np.sum(block_multiplier) == self.num_blocks
+        self.block_multiplier = block_multiplier
+        for i in range(self.num_weights):
+            self.cum_weights[:, i] = np.cumsum(
+                self.weights[:, i] * self.block_multiplier[self.block], 0
+            )
+            if self.cum_weights[-1, i] > 0:
+                self.quantile[:, i] = self.cum_weights[:, i] / self.cum_weights[-1, i]
+
+
+class CoalescenceTimeDistribution:
+    """
+    Class to precompute a table of sorted/weighted node times, from which to calculate
+    the empirical distribution function and estimate coalescence rates in time windows.
+    """
+
+    @staticmethod
+    def _count_coalescence_events(node, tree, sample_sets):
+        # TODO this will count unary nodes: should it count nodes
+        # with >1 child instead?
+        return np.array([1], dtype=np.int32)
+
+    @staticmethod
+    def _count_pair_coalescence_events(node, tree, sample_sets):
+        """
+        Count the number of pairs that coalesce in node, within and between the
+        sets of samples in ``sample_sets``. The count of pairs with members that
+        belong to sets :math:`a` and :math:`b` is:
+
+        .. math:
+
+            \\sum_{i \\neq j} (C_i(a) C_j(b) + C_i(b) C_j(a))/(1 - \\mathbb{I}[a = b])
+
+        where :math:`C_i(a)` is the number of samples from set :math:`a`
+        descended from child :math:`i`.  The values in the output are ordered
+        canonically; e.g. if ``len(sample_sets) == 2`` then the values would
+        correspond to counts of pairs with set labels ``[(0,0), (0,1), (1,1)]``.
+        """
+
+        # TODO needs to be optimized, use np.intersect1d
+        children = tree.children(node)
+        samples_per_child = [set(list(tree.samples(c))) for c in children]
+        sample_counts = np.zeros((len(sample_sets), len(children)), dtype=np.int32)
+        for i, s1 in enumerate(samples_per_child):
+            for a, s2 in enumerate([set(s) for s in sample_sets]):
+                sample_counts[a, i] = len(s1 & s2)
+
+        pair_counts = []
+        for a, b in itertools.combinations_with_replacement(
+            range(sample_counts.shape[0]), 2
+        ):
+            count = 0
+            for i, j in itertools.combinations(range(sample_counts.shape[1]), 2):
+                count += (
+                    sample_counts[a, i] * sample_counts[b, j]
+                    + sample_counts[a, j] * sample_counts[b, i]
+                ) / (1 + int(a == b))
+            pair_counts.append(count)
+
+        return np.array(pair_counts, dtype=np.int32)
+
+    @staticmethod
+    def _count_trio_first_coalescence_events(node, tree, sample_sets):
+        """
+        Count the number of pairs that coalesce in node with an outgroup,
+        within and between the sets of samples in ``sample_sets``. In other
+        words, count topologies of the form ``((A,B):node,C)`` where ``A,B,C``
+        are labels and `node` is the node ID.  The count of pairs with members
+        that belong to sets :math:`a` and :math:`b` with outgroup :math:`c` is:
+
+        .. math:
+
+            \\sum_{i \\neq j} (C_i(a) C_j(b) + C_i(b) C_j(a)) \\times
+            O(c) / (1 - \\mathbb{I}[a = b])
+
+        where :math:`C_i(a)` is the number of samples from set :math:`a`
+        descended from child :math:`i` of the node, and :math:`O(c)` is the
+        number of samples from set :math:`c` that are *not* descended from the
+        node.  The values in the output are ordered canonically by pair then
+        outgroup; e.g. if ``len(sample_sets) == 2`` then the values would
+        correspond to counts of pairs with set labels,
+        ``[((0,0),0), ((0,0),1), ..., ((0,1),0), ((0,1),1), ...]``.
+        """
+        samples = list(tree.samples(node))
+        outg_counts = [len(s) - len(np.intersect1d(samples, s)) for s in sample_sets]
+        pair_counts = CoalescenceTimeDistribution._count_pair_coalescence_events(
+            node, tree, sample_sets
+        )
+        trio_counts = []
+        for i in pair_counts:
+            for j in outg_counts:
+                trio_counts.append(i * j)
+        return np.array(trio_counts, dtype=np.int32)
+
+    def _update_weights_by_edge_diff(self, tree, edge_diff, running_weights):
+        """
+        Update ``running_weights`` to reflect ``tree`` using edge differences
+        ``edge_diff`` with the previous tree.
+        """
+
+        assert edge_diff.interval == tree.interval
+
+        # nodes that have been removed from tree
+        removed = {i.child for i in edge_diff.edges_out if tree.is_isolated(i.child)}
+        # TODO: What if sample is removed from tree? In that case should all
+        # nodes be updated for trio first coalescences?
+
+        # nodes where descendant subtree has been altered
+        modified = {i.parent for i in edge_diff.edges_in}
+        for i in copy.deepcopy(modified):
+            while tree.parent(i) != tskit.NULL and not tree.parent(i) in modified:
+                i = tree.parent(i)
+                modified.add(i)
+
+        # recalculate weights for current tree
+        for i in removed:
+            running_weights[i, :] = 0
+        for i in modified:
+            running_weights[i, :] = self.weight_func(i, tree, self.sample_sets)
+        self.weight_func_evals += len(modified)
+
+    def _build_ecdf_table_for_window(
+        self, left, right, tree, edge_diffs, running_weights
+    ):
+        """
+        Construct ECDF table for genomic interval [left, right]. Update ``tree``,
+        ``edge_diffs``, and ``running_weights`` for input for next window. Trees are
+        counted as belonging to any interval with which they overlap, and thus
+        can be used in several intervals. Thus, the concatenation of ECDF
+        tables across multiple intervals is not the same as the ECDF table
+        for the union of those intervals. Trees within intervals are chunked
+        into roughly equal-sized blocks for bootstrapping.
+        """
+
+        assert tree.interval.left <= left and right > left
+
+        # assign trees in window to equal-sized blocks with unique id
+        other_tree = tree.copy()
+        # TODO: is a full copy of the tree needed, given that the original is
+        # mutated below?
+        if right >= other_tree.tree_sequence.sequence_length:
+            other_tree.last()
+        else:
+            # other_tree.seek(right) won't work if `right` is recomb breakpoint
+            while other_tree.interval.right < right:
+                other_tree.next()
+        tree_idx = np.arange(tree.index, other_tree.index + 1) - tree.index
+        tree_offset = tree.index
+        num_blocks = min(self.num_blocks, len(tree_idx))
+        tree_blocks = np.floor_divide(num_blocks * tree_idx, len(tree_idx))
+
+        # calculate span weights
+        # TODO: if bootstrapping, does block span need to be tracked
+        # and used to renormalise each replicate?
+        other_tree.seek(tree.interval.left)
+        tree_span = [
+            min(other_tree.interval.right, right) - max(other_tree.interval.left, left)
+        ]
+        while other_tree.index < tree_offset + tree_idx[-1]:
+            other_tree.next()
+            tree_span.append(
+                min(other_tree.interval.right, right)
+                - max(other_tree.interval.left, left)
+            )
+        tree_span = np.array(tree_span) / sum(tree_span)
+
+        # storage if using single window, block for entire tree sequence
+        buffer_size = self.buffer_size
+        table_size = buffer_size
+        time = np.zeros(table_size)
+        block = np.zeros(table_size, dtype=np.int32)
+        weights = np.zeros((table_size, self.num_weights))
+
+        # assemble table of coalescence times in window
+        indices = np.zeros(tree.tree_sequence.num_nodes, dtype=np.int32) - 1
+        last_block = np.zeros(tree.tree_sequence.num_nodes, dtype=np.int32) - 1
+        num_record = 0
+        while tree.index != tskit.NULL:
+            if tree.interval.right > left:
+                current_block = tree_blocks[tree.index - tree_offset]
+                if self.span_normalise:
+                    span_weight = tree_span[tree.index - tree_offset]
+                else:
+                    span_weight = 1.0
+                nodes_in_tree = np.array(
+                    [i for i in tree.nodes() if tree.is_internal(i)]
+                )
+                # TODO this will fail if all nodes are isolated (masked tree)
+                nodes_to_add = nodes_in_tree[
+                    np.where(last_block[nodes_in_tree] != current_block)
+                ]
+                if len(nodes_to_add) > 0:
+                    idx = np.arange(num_record, num_record + len(nodes_to_add))
+                    last_block[nodes_to_add] = current_block
+                    indices[nodes_to_add] = idx
+                    if table_size < num_record + len(nodes_to_add):
+                        table_size += buffer_size
+                        time = np.pad(time, (0, buffer_size))
+                        block = np.pad(block, (0, buffer_size))
+                        weights = np.pad(weights, ((0, buffer_size), (0, 0)))
+                    time[idx] = [tree.time(i) for i in nodes_to_add]
+                    block[idx] = current_block
+                    num_record += len(nodes_to_add)
+                weights[indices[nodes_in_tree], :] += (
+                    span_weight * running_weights[nodes_in_tree, :]
+                )
+
+            if tree.interval.right < right:
+                # if current tree does not cross window boundary, move to next
+                tree.next()
+                self._update_weights_by_edge_diff(
+                    tree, next(edge_diffs), running_weights
+                )
+            else:
+                # use current tree as initial tree for next window
+                break
+
+        return CoalescenceTimeTable(time, block, weights)
+
+    def _generate_ecdf_tables(self, ts, window_breaks):
+        """
+        Return generator for ECDF tables across genomic windows defined by
+        ``window_breaks``.
+
+        ..note:: This could be used in methods in place of loops over
+            pre-assembled tables.
+        """
+
+        tree = ts.first()
+        edge_diffs = ts.edge_diffs()
+        running_weights = np.zeros((ts.num_nodes, self.num_weights))
+        self._update_weights_by_edge_diff(tree, next(edge_diffs), running_weights)
+        for left, right in zip(window_breaks[:-1], window_breaks[1:]):
+            yield self._build_ecdf_table_for_window(
+                left, right, tree, edge_diffs, running_weights
+            )
+
+    def __init__(
+        self,
+        ts,
+        sample_sets=None,
+        weight_func=None,
+        window_breaks=None,
+        blocks_per_window=None,
+        span_normalise=True,
+    ):
+
+        assert isinstance(ts, trees.TreeSequence)
+
+        if sample_sets is None:
+            sample_sets = [list(ts.samples())]
+        assert all([isinstance(i, list) for i in sample_sets])
+        assert all([i in ts.samples() for j in sample_sets for i in j])
+        self.sample_sets = sample_sets
+
+        if weight_func is None or weight_func == "coalescence_events":
+            self.weight_func = self._count_coalescence_events
+        elif weight_func == "pair_coalescence_events":
+            self.weight_func = self._count_pair_coalescence_events
+        elif weight_func == "trio_first_coalescence_events":
+            self.weight_func = self._count_trio_first_coalescence_events
+        else:
+            assert callable(weight_func)
+            self.weight_func = weight_func
+        _weight_func_eval = self.weight_func(0, ts.first(), self.sample_sets)
+        assert isinstance(_weight_func_eval, np.ndarray)
+        assert _weight_func_eval.ndim == 1
+        self.num_weights = len(_weight_func_eval)
+
+        if window_breaks is None:
+            window_breaks = np.array([0.0, ts.sequence_length])
+        assert isinstance(window_breaks, np.ndarray)
+        assert window_breaks.ndim == 1
+        assert np.min(window_breaks) >= 0.0
+        assert np.max(window_breaks) <= ts.sequence_length
+        window_breaks = np.sort(np.unique(window_breaks))
+        self.windows = [
+            trees.Interval(left, right)
+            for left, right in zip(window_breaks[:-1], window_breaks[1:])
+        ]
+        self.num_windows = len(self.windows)
+
+        if blocks_per_window is None:
+            blocks_per_window = 1
+        assert isinstance(blocks_per_window, int)
+        assert blocks_per_window > 0
+        self.num_blocks = blocks_per_window
+
+        assert isinstance(span_normalise, bool)
+        self.span_normalise = span_normalise
+
+        self.buffer_size = ts.num_nodes
+        self.weight_func_evals = 0
+        self.tables = [table for table in self._generate_ecdf_tables(ts, window_breaks)]
+
+    # TODO
+    #
+    # def __str__(self):
+    #    return self.useful_text_summary()
+    #
+    # def __repr_html__(self):
+    #    return self.useful_html_summary()
+
+    def copy(self):
+        return copy.deepcopy(self)
+
+    def ecdf(self, times):
+        """
+        Returns the empirical distribution function evaluated at the time
+        points in ``times``.
+
+        The output array has shape ``(self.num_weights, len(times),
+        self.num_windows)``.
+        """
+
+        assert isinstance(times, np.ndarray)
+        assert times.ndim == 1
+
+        values = np.empty((self.num_weights, len(times), self.num_windows))
+        values[:] = np.nan
+        for k, table in enumerate(self.tables):
+            indices = np.searchsorted(table.time, times, side="right") - 1
+            assert all([0 <= i < table.num_records for i in indices])
+            values[:, :, k] = table.quantile[indices, :].T
+        return values
+
+    # TODO
+    #
+    # def quantile(self, times):
+    #   """
+    #   Return interpolated quantiles of coalescence times, using the same
+    #   approach as numpy.quantile(..., method="linear")
+    #   """
+
+    def num_coalesced(self, times):
+        """
+        Returns number of coalescence events that have occured by the time
+        points in ``times``.
+
+        The output array has shape ``(self.num_weights, len(times),
+        self.num_windows)``.
+        """
+
+        assert isinstance(times, np.ndarray)
+        assert times.ndim == 1
+
+        values = self.ecdf(times)
+        for k, table in enumerate(self.tables):
+            weight_totals = table.cum_weights[-1, :].reshape(values.shape[0], 1)
+            values[:, :, k] *= np.tile(weight_totals, (1, values.shape[1]))
+        return values
+
+    def num_uncoalesced(self, times):
+        """
+        Returns the number of coalescence events remaining by the time points
+        in ``times``.
+
+        The output array has shape ``(self.num_weights, len(times),
+        self.num_windows)``.
+        """
+
+        values = 1.0 - self.ecdf(times)
+        for k, table in enumerate(self.tables):
+            weight_totals = table.cum_weights[-1, :].reshape(values.shape[0], 1)
+            values[:, :, k] *= np.tile(weight_totals, (1, values.shape[1]))
+        return values
+
+    def mean(self, since=0.0):
+        """
+        Returns the average time between ``since`` and the coalescence events
+        that occurred after ``since``.
+
+        Note that ``1/self.mean(left)`` is an estimate of the coalescence rate
+        over the interval (left, infinity).
+
+        The output array has shape ``(self.num_weights, self.num_windows)``.
+
+        ..note:: Check for overflow in ``np.average``.
+        """
+
+        assert isinstance(since, float) and since >= 0.0
+
+        values = np.empty((self.num_weights, self.num_windows))
+        values[:] = np.nan
+        for k, table in enumerate(self.tables):
+            index = np.searchsorted(table.time, since, side="right")
+            if index == table.num_records:
+                values[:, k] = np.nan
+            else:
+                for i in range(self.num_weights):
+                    if table.cum_weights[-1, i] > 0:
+                        multiplier = table.block_multiplier[table.block[index:]]
+                        values[i, k] = np.average(
+                            table.time[index:] - since,
+                            weights=table.weights[index:, i] * multiplier,
+                        )
+        return values
+
+    def coalescence_probability_in_intervals(self, time_breaks):
+        """
+        Returns the proportion of coalescence events occurring in the time
+        intervals defined by ``time_breaks``, out of events that have not
+        yet occurred by the intervals' left boundaries.
+
+        The output array has shape ``(self.num_weights, len(time_breaks)-1,
+        self.num_windows)``.
+        """
+
+        assert isinstance(time_breaks, np.ndarray)
+
+        time_breaks = np.sort(np.unique(time_breaks))
+        num_coalesced = self.num_coalesced(time_breaks)
+        num_uncoalesced = self.num_uncoalesced(time_breaks)
+        numer = num_coalesced[:, 1:, :] - num_coalesced[:, :-1, :]
+        denom = num_uncoalesced[:, :-1, :]
+        return numer / np.where(np.isclose(denom, 0.0), np.nan, denom)
+
+    def coalescence_rate_in_intervals(self, time_breaks):
+        """
+        Returns the interval-censored Kaplan-Meier estimate of the hazard rate for
+        coalesence events within the time intervals defined by ``time_breaks``. The
+        estimator is,
+
+        .. math::
+            \\hat{c}_{l,r} = \\begin{cases}
+              \\log(1 - x_{l,r}/k_{l})/(l - r) & \\mathrm{if~} x_{l,r} < k_{l} \\\\
+              \\hat{c}_{l,r} = k_{l} / t_{l,r} & \\mathrm{if~} x_{l,r} = k_{l}
+            \\end{cases}
+
+        and is undefined where :math:`k_{l} = 0`. Here, :math:`x_{l,r}` is the
+        number of events occuring in time interval :math:`(l, r]`,
+        :math:`k_{l}` is the number of events remaining at time :math:`l`, and
+        :math:`t_{l,r}` is the sum of event times occurring in the interval
+        :math:`(l, r]`.
+
+        The output array has shape ``(self.num_weights, len(time_breaks)-1,
+        self.num_windows)``.
+        """
+
+        assert isinstance(time_breaks, np.ndarray)
+
+        time_breaks = np.sort(np.unique(time_breaks))
+        phi = self.coalescence_probability_in_intervals(time_breaks)
+        duration = np.reshape(time_breaks[1:] - time_breaks[:-1], (1, phi.shape[1], 1))
+        numer = -np.log(1.0 - np.where(np.isclose(phi, 1.0), np.nan, phi))
+        denom = np.tile(duration, (self.num_weights, 1, self.num_windows))
+        for i, j, k in np.argwhere(np.isclose(phi, 1.0)):
+            numer[i, j, k] = 1.0
+            denom[i, j, k] = self.mean(time_breaks[j])[i, k]
+        return numer / denom
+
+    def block_bootstrap(self, num_replicates=1, random_seed=None):
+        """
+        Return a generator that produces ``num_replicates`` copies of the
+        object where blocks within genomic windows are randomly resampled.
+
+        ..note:: Copying could be expensive.
+        """
+
+        rng = np.random.default_rng(random_seed)
+        for _i in range(num_replicates):
+            replicate = self.copy()
+            for table in replicate.tables:
+                block_multiplier = rng.multinomial(
+                    table.num_blocks, [1.0 / table.num_blocks] * table.num_blocks
+                )
+                table.resample_blocks(block_multiplier)
+            yield replicate
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -46,6 +46,7 @@ import tskit
 import tskit.combinatorics as combinatorics
 import tskit.drawing as drawing
 import tskit.metadata as metadata_module
+import tskit.stats as stats
 import tskit.tables as tables
 import tskit.text_formats as text_formats
 import tskit.util as util
@@ -8718,6 +8719,26 @@ class TreeSequence:
             min_span=min_span,
             store_segments=store_segments,
             store_pairs=store_pairs,
+        )
+
+    def coalescence_time_distribution(
+        self,
+        *,
+        sample_sets=None,
+        weight_func=None,
+        window_breaks=None,
+        blocks_per_window=None,
+        span_normalise=False,
+    ):
+        # TODO docstring, not yet in API
+
+        return stats.CoalescenceTimeDistribution(
+            self,
+            sample_sets=sample_sets,
+            weight_func=weight_func,
+            window_breaks=window_breaks,
+            blocks_per_window=blocks_per_window,
+            span_normalise=span_normalise,
         )
 
     ############################################
--- a/python/tests/test_coaltime_distribution.py
+++ b/python/tests/test_coaltime_distribution.py
@@ -1,0 +1,1093 @@
+# MIT License
+#
+# Copyright (c) 2018-2022 Tskit Developers
+# Copyright (C) 2016 University of Oxford
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Test cases for coalescence time distribution objects in tskit.
+"""
+import msprime
+import numpy as np
+
+import tests
+import tskit
+
+
+class TestCoalescenceTimeDistribution:
+    """
+    Tree sequences used in tests of classes `CoalescenceTimeTable` and
+    `CoalescenceTimeDistribution`
+    """
+
+    @tests.cached_example
+    def ts_multimerger_six_leaves(self):
+        """
+        29.00┊    9        ┊
+             ┊ ┏━━┻━━┓     ┊
+        8.00 ┊ ┃     8     ┊
+             ┊ ┃   ┏━┻━━┓  ┊
+        5.00 ┊ ┃   7    ┃  ┊
+             ┊ ┃ ┏━╋━┓  ┃  ┊
+        1.00 ┊ ┃ ┃ ┃ ┃  6  ┊
+             ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊
+        0.00 ┊ 0 1 2 4 3 5 ┊
+             0            100
+        """
+        tables = tskit.TableCollection(sequence_length=100)
+        tables.nodes.set_columns(
+            time=np.array([0, 0, 0, 0, 0, 0, 1, 5, 8, 29]),
+            flags=np.array([1, 1, 1, 1, 1, 1, 0, 0, 0, 0], dtype="uint32"),
+        )
+        tables.edges.set_columns(
+            left=np.repeat([0.0], 9),
+            right=np.repeat([100.0], 9),
+            parent=np.array([6, 6, 7, 7, 7, 8, 8, 9, 9], dtype="int32"),
+            child=np.array([3, 5, 1, 2, 4, 6, 7, 0, 8], dtype="int32"),
+        )
+        return tables.tree_sequence()
+
+    @tests.cached_example
+    def ts_multimerger_eight_leaves(self):
+        """
+        10.00┊         13      ┊
+             ┊       ┏━━┻━━┓   ┊
+        8.00 ┊      12     ┃   ┊
+             ┊     ┏━┻━┓   ┃   ┊
+        6.00 ┊    11   ┃   ┃   ┊
+             ┊  ┏━━╋━┓ ┃   ┃   ┊
+        2.00 ┊ 10  ┃ ┃ ┃   9   ┊
+             ┊ ┏┻┓ ┃ ┃ ┃  ┏┻━┓ ┊
+        1.00 ┊ ┃ ┃ ┃ ┃ ┃  8  ┃ ┊
+             ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┊
+        0.00 ┊ 0 7 4 5 6 1 2 3 ┊
+             0                100
+        """
+        tables = tskit.TableCollection(sequence_length=100)
+        tables.nodes.set_columns(
+            time=np.array([0] * 8 + [1, 2, 2, 6, 8, 10]),
+            flags=np.repeat([1, 0], [8, 6]).astype("uint32"),
+        )
+        tables.edges.set_columns(
+            left=np.repeat([0], 13),
+            right=np.repeat([100], 13),
+            parent=np.array(
+                [8, 8, 9, 9, 10, 10, 11, 11, 11, 12, 12, 13, 13], dtype="int32"
+            ),
+            child=np.array([1, 2, 3, 8, 0, 7, 4, 5, 10, 6, 11, 9, 12], dtype="int32"),
+        )
+        return tables.tree_sequence()
+
+    @tests.cached_example
+    def ts_two_trees_four_leaves(self):
+        """
+        1.74┊   7     ┊   7     ┊
+            ┊ ┏━┻━┓   ┊ ┏━┻━┓   ┊
+        0.73┊ ┃   6   ┊ ┃   ┃   ┊
+            ┊ ┃ ┏━┻┓  ┊ ┃   ┃   ┊
+        0.59┊ ┃ ┃  5  ┊ ┃   5   ┊
+            ┊ ┃ ┃ ┏┻┓ ┊ ┃  ┏┻━┓ ┊
+        0.54┊ ┃ ┃ ┃ ┃ ┊ ┃  4  ┃ ┊
+            ┊ ┃ ┃ ┃ ┃ ┊ ┃ ┏┻┓ ┃ ┊
+        0.00┊ 0 1 2 3 ┊ 0 1 2 3 ┊
+          0.00      0.88      1.00
+        """
+        tables = tskit.TableCollection(sequence_length=1)
+        tables.nodes.set_columns(
+            time=np.array([0, 0, 0, 0, 0.54, 0.59, 0.73, 1.74]),
+            flags=np.array([1, 1, 1, 1, 0, 0, 0, 0], dtype="uint32"),
+        )
+        tables.edges.set_columns(
+            left=np.array([0.88, 0.88, 0, 0, 0.88, 0, 0, 0, 0.88, 0]),
+            right=np.array([1, 1, 0.88, 1, 1, 0.88, 0.88, 1, 1, 0.88]),
+            parent=np.array([4, 4, 5, 5, 5, 6, 6, 7, 7, 7], dtype="int32"),
+            child=np.array([1, 2, 2, 3, 4, 1, 5, 0, 5, 6], dtype="int32"),
+        )
+        return tables.tree_sequence()
+
+    @tests.cached_example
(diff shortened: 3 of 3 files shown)


gregorgorjanc:
This looks great @nspope !!!

@janaobsteter you can try it on your bee tree sequence work

nspope:
There is something that I am not grasping about the dependence between the TMRCAs for two non-overlapping pairs of samples in the same genealogy. The TMRCAs for a single sample pair follow an exponential distribution, and counts of coalescence events in a time window follow a binomial distribution like I describe above. 

However, for the **sum** of counts over multiple non-overlapping sample pairs, the distribution is not binomial in general, because both pairs might have the same MRCA. For example, in the genealogy
```
5.01┊    6    ┊
    ┊  ┏━┻━┓  ┊
1.43┊  ┃   5  ┊
    ┊  ┃  ┏┻┓ ┊
0.66┊  4  ┃ ┃ ┊
    ┊ ┏┻┓ ┃ ┃ ┊
0.00┊ 0 3 1 2 ┊
    0         1
```
sample pairs (0, 1) and (3, 2) both have the same MRCA (6). This coalescent event is double-counted when tallying TMRCAs in time windows over both pairs. So, in this case the binomial model doesn't hold. The estimator `-d_i / log(1 - x_i/k_i)` still seems to be unbiased, but the std error is lower than it should be.

Something like this might work instead:
1. Iterate over internal nodes
2. For each internal node choose a pair of descendant tips that have not already been selected (for a previously visited node). If this is not possible, discard this node and move on.
3. Tally the times associated with the retained internal nodes across windows, and use these in the binomial model

I have to work through this procedure more carefully, but I think it would result in sums of independent counts per time window.

petrelharp:
Awesome, thanks @nspope. This is a great start, and there's quite a bit to think about here. Here's some thoughts so far (and some are probably silly due to not thinking enough!):

About the interface: we have a proposed way to specify time windows in #665, that talks through e.g. what to do if both `windows` and `time_windows` are passed in. What do you think?

About confidence intervals: Perhaps the issue of getting a good estimator and obtaining uncertainty estimates can be disentangled somewhat. You have great suggestions for how to get uncertainty for a single tree, but extending these ideas to a region of the genome with many trees seems difficult? But: maybe I'm missing something? Perhaps a more general-purpose way to get uncertainty would be a block bootstrap of some kind? (e.g., compute estimates in windows of the genome that are the original windows subdivided in 5, and use the variance across these?)

More generally, I'm wondering what is the interpretation of this estimator as an empirical description of the genealogies? For instance, is it related to the ECDF of the pairwise coalescence times? If so, why do we need the restriction to non-overlapping pairs of lineages? Having a nice empirical interpretation like this is nice for communication, and understanding what the statistic is telling us about the trees, independent of a particular model (ie a randomly mating pop of size Ne).

An alternative method might be to instead of taking pairwise TMRCAs of samples, to take the number of *distinct lineages*; so in your example:
```
5.01┊    6    ┊
    ┊  ┏━┻━┓  ┊
1.43┊  ┃   5  ┊
    ┊  ┃  ┏┻┓ ┊
0.66┊  4  ┃ ┃ ┊
    ┊ ┏┻┓ ┃ ┃ ┊
0.00┊ 0 3 1 2 ┊
    0         1
```
after 0 and 3 coalesce into 4 you've only got one lineage remaining, not two. Of course, doing things this way makes interpretation harder, especially if we're thinking about cross-coalescence rates.

jeromekelleher:
This does look great @nspope and it's a really important piece of functionality. I wonder if it'd be worth talking through at one of our community meetings? These happen every other Wednesday 11:30EST, next one is on 16th. Would you be willing to talk us through what you've done here, and maybe we could discuss how to take it forward?

nspope:
@jeromekelleher Sure, I can join you all on the 16th. If you need to reach me outside of Github with details on how to join, I'm at nspope at utexas dot edu.

@petrelharp Great, thanks for the feedback, I agree that summary statistics vs demographic inference should be distinct in the implementation. A block bootstrap is really easy to implement here, as it just requires "re-weighting" counts across genomic windows with multinomial weights. The restriction to non-overlapping sample pairs is totally unnecessary -- this was just my (misguided) attempt at getting counts of coalescences that better match the binomial model described in OP (to try to strike a balance between accurately quantifying uncertainty and using all available information). I'll respond in detail when I have a moment.

nspope:
Hey @petrelharp,

Here's my thinking ... Let `n(t)` be the haploid effective population size; then `c(t) = 1/n(t)` is the pairwise coalescence rate (or "hazard" in survival modelling jargon). The cumulative rate is `C(t) = \int_0^t c(x) dx`, and the cumulative distribution function of pair coalesence times is `F(t) = 1 - exp(-C(t))`. If `F'(t)` is the derivative with respect to time, then we have `n(t) = (1 - F(t)) / F'(t)`.

Lets say we have an empirical CDF `G(t)`, taken from the pairwise TMRCAs of a tree or tree sequence. Of course this is piecewise constant and not smooth, but its expectation (wrt possible genealogies) is `F(t)`. What we're after is an estimate of the effective population size / pair coalescence rate through time. One way to go about estimating this would be approximate `F(t)` by fitting some smooth function `F_hat(t)` to `G(t)`, then plug `F_hat` and its time derivative into the relationship above. This is an interesting option, but seems like a bit of a can of worms ... because we'd have to make decisions about how smooth to make `F_hat(t)` and that will have a huge influence on `F_hat'(t)` and thus the associated estimates of `n(t)` and `c(t)` (especially when we get far enough back in time that there are few uncoalesced lineages). There's also the issue that in any (real) dataset, there is inevitably going to be error in the inferred TMRCAs. So for now, maybe we can punt the "fit a smoother" idea down the road. 

A simpler alternative (that's more robust to error in the inferred TMRCAs) is to chunk up time into intervals and use `G(t)` to directly approximate `F(t)`. Then, we can reframe the problem as trying to get an average coalescence rate for the interval. When time is discretized like this, we really only care about whether a particular interval contains a given TMRCA, so it's analogous to an interval-censored survival model.

Consider a time interval `[a, b)`. The cumulative coalesence rate in this interval is `C(b) - C(a) = c_bar * (b - a)` where `c_bar = (C(b) - C(a)) / (b - a)` is the average rate of the interval. To relate this back to the estimator that I wrote about in the OP, let `k` be the number of uncoalesced lineages at time `a`, and let `x` be the number of lineages that coalescence in `[a, b)`. Using the ECDF, 

`x / k = (G(b) - G(a)) / (1 - G(a)) \approx (F(b) - F(a)) / (1 - F(a)) = 1 - exp(-(b - a) * c_bar)` 

and solving this gives `c_bar \approx -log(1 - x/k) / (b - a)` (the estimator). Finally, `1 / c_bar = (b - a) / \int_a^b 1/n(x) dx` is the harmonic average of effective population size over the interval, which I think is what interests @gregorgorjanc.

With regards to block bootstrapping -- I totally agree, this would be a great way to get getting confidence intervals for a global set estimates calculated for the entire tree sequence. But, I think it would be challenging to use block bootstrap to get CIs for local estimates calculated on small genomic intervals that may just contain a few trees. In this case, something like the binomial model I laid out in OP might work -- if the standard error could somehow be adjusted to account for the dependence between pair TMRCAs in a genealogy. Unlike everything written above, the non-independence of TMRCAs seems specific to the coalescent and not to survival modelling or non-homogeneous point processes in general.

Sorry for the text dump!

gregorgorjanc:
@nspope thanks for this great "text dump"!

Just a note that tsdate can estimate node ages, hence TMRCA, but there will inevitably be errors, so time-binning should work fine, at least for what we need.

petrelharp:
Great, @nspope  - this is perfect! Well-explained, and I think also a very good approach.  Here's some notes:

- The numbers `x` and `k` are the "raw" input here, and these are of independent interest (as a more fundamental descriptor of the lineages, and the ingredients for more sophisticated estimators), so I'd vote for writing the low-level methods that just compute and return the arrays of these, then the Ne(t) estimator as a layer on top of that. (And, maybe we just need `x`?)
- I think there's at least two options for what `x` (and `k`) is: one is "how many of the pairs of samples are not yet coalesced", and the other is (n choose 2), where n is "how many distinct lineages ancestral to the samples are there". This amounts to a different weighting of the lineages present at the start of the time bin, and I think the second option should get a smaller variance because the things we're putting into G(t) are more independent. (Since if i and j have already coalesced, then T(i,k) = T(j,k) by necessity.)  I think both could be useful, as the first one is closer to the data.
- This obviously extends to pairwise coalescence rates!
- I'm inclined to return coalescence rates instead of "effective population sizes"? If people want the latter, they can just invert them?
- Genomic windows can be pretty small and still contain a lot of trees, so the block bootstrap might still be useful. However, not always, you're right - and, how much independence we get depends on how long ago we're talking about (more recent times  would need bigger windows). I agree, though, the binomial confidence intervals seem promising - there ought to be some approximation we can make?

nspope:
Thanks @petrelharp,

>The numbers x and k are the "raw" input here, and these are of independent interest (as a more fundamental descriptor of the lineages, and the ingredients for more sophisticated estimators), so I'd vote for writing the low-level methods that just compute and return the arrays of these, then the Ne(t) estimator as a layer on top of that. (And, maybe we just need x?)

Here's an idea ... we organize these into a class. The member functions of this class are something like `ecdf(self, time)`, `num_uncoalesced(self, time)`, `num_coalesced(self, time)`, that return the values of `G(t), x(t), k(t)` along a supplied time grid. The member data of this class would be a table, where each row corresponds to a node in the tree sequence. Rows would be sorted by the time associated with each node. Columns of this table would include: the node times, the node positions, and the node weights. By node weights, I mean the number of pairs that coalesce in a given node, which (in a binary tree) is just the number of leaves descended from the left branch multiplied by the number of leaves descended from the right branch. Building this table would require just a single pass over the nodes of the tree sequence, when the class is instantiated. 

The rationale behind this is that once the TMRCAs (node times) are time-sorted and pair-weighted, it is very efficient to calculate any of the quantities above at an arbitrary time point. By contrast, it seems likely the user would want to play with "binning" of coalescence events into time bins to calculate the pair coalescence rates, by changing the number/size of the time windows so as to have some control on the tradeoff between variance and temporal resolution. In this case, we'd want to avoid iterating over nodes each time the estimators are calculated. So, there would be another member function along the lines of `coalescence_density(self, time_windows, num_bootstrap=0, block_distance=None)` that uses the sorted table to calculate the estimators and resamples nodes via a block scheme if `num_bootstrap > 0`. Of course there could be a local form of this that returns estimates across genomic windows.

But maybe I'm over-complicating things/reinventing the wheel? Or maybe there's a desire to limit the proliferation of classes in tskit?

>I think there's at least two options for what x (and k) is: one is "how many of the pairs of samples are not yet coalesced", and the other is (n choose 2), where n is "how many distinct lineages ancestral to the samples are there". This amounts to a different weighting of the lineages present at the start of the time bin, and I think the second option should get a smaller variance because the things we're putting into G(t) are more independent. (Since if i and j have already coalesced, then T(i,k) = T(j,k) by necessity.) I think both could be useful, as the first one is closer to the data.

Sure, I agree both are useful. However (unless I'm misunderstanding), their respective ECDFs don't have the same expectation? So only the "pairwise coalescence rate" has the nice duality with effective population size that makes interpretation easy.

nspope:
> I'm inclined to return coalescence rates instead of "effective population sizes"? If people want the latter, they can just invert them?

I agree with coalescence rates as a default, but maybe provide a flag that controls which is returned. The reason is because if returning std errors or quantities calculated from bootstraps, the conversion of these wouldn't necessarily be as simple as just inverting them.

nspope:
> how much independence we get depends on how long ago we're talking about (more recent times would need bigger windows). I agree, though, the binomial confidence intervals seem promising - there ought to be some approximation we can make?

Thinking about this further ... clustered outcomes are definitely a phenomenon that is modeled in survival analysis. For example, all the machines in the tootsie roll factory break down at the same time because someone added too much sugar to the caramel (or something). So there may well be a simple correction/approximation that could be relevant here.

vsbuffalo:
Hi All,

So I've been working on implementing a few coalescence rate estimators for my own work and ran across this nice count-based estimator by @nspope.

I have many samples though, and it looks like running this current implementation over n samples is O(n²). I think a faster alternative is to run through the marginal trees and just count the number of leave pairs subtended by each node. One needs to be careful though, as an m-way multiple merger creates m choose 2 lineages, and each lineage pair i, j from this focal node has n(i)n(j) pairwise samples where n(i) is the number of leaves of the descendant lineage of i. I bashed out a quick draft implementation:

```
def count_tree_coalscent_pairs(tree, epochs):
    nepochs = len(epochs)
    times = dict()
    for tr in tree.trees():
        for node in tr.nodes():
            if tr.is_leaf(node):
                continue
            nlineages = [len(list(tr.leaves(child))) for child in tr.children(node)]
            pairs = list(itertools.combinations(nlineages, 2))
            # get the number of subtended samples
            ns = np.array([x*y for x, y in pairs])
            t = tr.time(node)
            times[t] = np.sum(ns) + times.get(t, 0)
    counts = np.fromiter(times.values(), dtype=int)
    times = np.fromiter(times.keys(), dtype=float)
    indices = np.digitize(times, epochs)
    coalesced = np.zeros(shape=nepochs+1)
    for i, (idx, time, count) in enumerate(zip(indices, times, counts)):
        coalesced[idx] += count
    total = np.flip(np.cumsum(np.flip(coalesced[1:])))
    return coalesced[1:], total, epochs
```

Then, for some sample data:

```
tree = msprime.sim_ancestry(100, ploidy=2, recombination_rate=1e-8, sequence_length=1e6, population_size=1000, random_seed=5)
bins = np.linspace(0, 10000, 10)
```

Let's look at @nspope's total counts:

```
>>> t = np.stack([x['total'].tolist() for x in pairwise_coalescence_counts(tree, bins)['counts'].values()]).sum(axis=0)
>>> c = np.stack([x['coalesced'].tolist() for x in pairwise_coalescence_counts(tree, bins)['counts'].values()]).sum(axis=0)
>>> t, c
(array([3621800, 2202888,  731645,  409806,  117864,   60576,   56052,
          18684,       0]),
 array([1418912, 1471243,  321839,  291942,   57288,    4524,   37368,
          18684,       0]))
```

and my implementation (pardon the minor output differences, we can address this later):

```
(array([1418912., 1471243.,  321839.,  291942.,   57288.,    4524.,
          37368.,   18684.,       0.,       0.]),
 array([3621800., 2202888.,  731645.,  409806.,  117864.,   60576.,
          56052.,   18684.,       0.,       0.]),
 array([    0.        ,  1111.11111111,  2222.22222222,  3333.33333333,
         4444.44444444,  5555.55555556,  6666.66666667,  7777.77777778,
         8888.88888889, 10000.        ]))
```

Now, looking at the timing of @nspope's

```
>>> %timeit counts = pairwise_coalescence_counts(tree, bins)
6.16 s ± 317 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

and mine:
```
>>> %timeit count_tree_coalscent_pairs(tree, bins)
747 ms ± 6.01 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So I think this speeds things up a bit (~82x faster) and may be of interest. I will work to refine and document the code if this is of interest — pardon my quick implementation here.

nspope:
Thanks for the suggestion @vsbuffalo! Yup, I think we'll ultimately want to do a single pass over nodes. The tip-by-tip approach that I have currently implemented is really just a foolproof point of comparison/prototype, not in any way for speed or scalability. At this point, we're just hashing out exactly what we want to estimate, then I'll work on an efficient implementation.

vsbuffalo:
This is something I was wondering as well — I think there is a strong case that coalescent rate estimators should not be put in tskit, but perhaps a few of the functions that go into computing these rates are generalizable enough that they should make it in. 

Unlike with diversity, or Fst, there doesn't appear to be a sort of canonical best estimator of coalescent rates. If we knew tree generating processing itself, there are likely better, tailored rate estimators we can construct. One recent paper that sticks out in my head is [Edge and Coop (2019)](https://academic.oup.com/genetics/article/211/1/235/5931145?login=true), which developed three estimators to infer allele frequencies and trait PGSs from trees. I think there's a case that tskit would just have to contain all kinds of useful estimators if it contains one, since they all are useful or terrible under different circumstances. From my experiments, @nspope's estimator seems like it performs rather well compared to the waiting-time estimator of [Strimmer and Pybus (2001)](https://academic.oup.com/mbe/article/18/12/2298/1074372), but the latter is commonly used. 

I do think, however, there are some general utility functions we could include that would be useful for all kinds of estimators. For example, functions to bin pairwise coalescence events and totals as used here, count the number of lineages at a particular time slice, a general proper block bootstrap for an arbitrary estimation function, etc. seem both general and tricky to implement efficiently that they may be contenders for inclusion. Either way, I think there is a strong case that we should write up a vignette on how to estimate coalescent rates as I imagine this could be useful to many.

nspope:
@jeromekelleher @vsbuffalo In my opinion, it'd be useful to include a way to calculate counts of pair coalescence times in tskit, both across time/genomic windows and between defined groups (e.g. cross-coalescence between populations, with outgroup constraints, and so). My feeling is that this kind of summary statistic will show up in more and more inferential methods in the future: they're both informative and analytically tractable under a wide range of models. So, I think it is worth hashing out an efficient and extendable way to calculate/summarise/resample these.

The downstream use of these statistics for demographic inference (or inference about selection, etc.) definitely seems beyond the scope of tskit. However, the duality between pair coalescence rate and Ne in an unstructured population is so simple (and so interpret-able) that in my opinion it would be useful to include a simple estimator of time-averaged coalescence rate, as a summary statistic in its own right. And to write a brief tutorial on how this relates to Ne under specific conditions.

One idea is to organize the process of counting/resampling into a class that could serve as a base for spin-offs like cross-coalescences. So for example, `class PairCoalescenceTimes` that only passes over nodes in marginal trees once, on instantiation. This would build a table that includes: node id, node time, node interval (boundaries of the leftmost/rightmost trees containing the node); node weights (number of descendant pairs that coalescence at the node, summed over marginal trees). Then, sort the table by time and calculate cumulative weights. Querying the ECDF at an arbitrary time point is a binary search over the table, and the ECDF can be used to calculate counts of uncoalesced/coalesced lineages or average coalescence rate in windows. Block bootstrapping would involve assigning nodes to blocks, drawing block weights from a multinomial distribution, and reweighting nodes accordingly. So a use case might look like,
```
focal_sets = [population1, population2, population3] # distinct sets of samples
time_grid = numpy.linspace(0., max_time, num_time_points)
coal_times = PairCoalescenceTimes(ts, focal_sets)

# returns array with shape ( len(focal_sets) , len(time_grid) )
coal_times.ecdf(time_grid)
coal_times.num_coalesced(time_grid)
coal_times.num_uncoalesced(time_grid)

# returns array with shape ( len(focal_sets) , len(time_grid)-1 , bootstrap_replicates )
# e.g. "time_grid" is used as breaks for windows wherein time-average rates are calculated
coal_times.binned_coalescence_events(time_grid, bootstrap_replicates, block_size, random_seed) # raw counts
coal_times.binned_coalescence_rate(time_grid, bootstrap_replicates, block_size, random_seed) # simple estimator
```
I have a prototype of this idea that I'll push shortly, and if there's time perhaps we could discuss on Wed.

jeromekelleher:
This sounds great @nspope - looking forward to hearing more about this one Wednesday, and have a good discussion about how we get the infrastructure into tskit. I am sympathetic to @vsbuffalo's perspective here, but I totally see what you're saying too @nspope.

vsbuffalo:
Yea that looks like a very clean API @nspope! I do think if there were to be a coalescent rate implementation, the pairwise approach is the best (rather than say the Strimmer and Pybus estimator which bakes in the Kingman k choose 2 numerator). I think one thing we could keep in mind is whether these data could go into other estimators like [Speidel et al's](https://static-content.springer.com/esm/art%3A10.1038%2Fs41588-019-0484-x/MediaObjects/41588_2019_484_MOESM1_ESM.pdf) waiting time estimator (p. 15).

nspope:
I had one more thought regarding the discussion in #1315 (which prompted this PR) and the question of what fits into tskit (if anything). The way I've been viewing things has been a little myopic, I think. A more general perspective is that we are passing over nodes in marginal trees, and weighting node times so as to capture interesting relationships between particular topological features and particular points in time.

For example, if the node weights are uniform, then ECDF(t) approximates the distribution function of coalescence times of genealogies  of a given size. If the node weights are the number of sample pairs coalescing at the node, then ECDF(t) approximates the distribution function of pairwise coalescence times, that happens to have a nice connection to effective population size in an unstructured population. If node weights are the number of sample pairs from separate "groups" coalescing at the node, then ECDF(t) approximates the distribution function of cross-coalescence times. And so on -- the "features" counted by the weights could be any sort of topological pattern. The choice of what to count should be left to the user. For instance: I study adaptive introgression ... so I could use weights that count the possible topologies for 4 samples from different subspecies. Taken together these would give the ingredients for something like a time-aware ABBABABA statistic, that I could compare to simulations.

Whatever the choice of node weights, the distribution that is being approximated is a time-to-event (survival) process. So the question for implementation is, what summary statistics are useful ways of describing these processes? The ECDF, of course, but that's not a very concise description and will be sensitive to error in the inferred node times. Raw moments are maybe too concise. I think more useful statistics come from survival analysis, such as estimates of the average event rate (hazard function) in time windows.

I suppose what I'm suggesting is an interface to get summary statistics from any time-to-event distribution that arises from counting a particular topology in a tree sequence for each node. These statistics could include familiar estimators from survival analysis (like the interval estimator in the OP). Then, a block bootstrap scheme that generates bootstrap replicates wherein each node weight is multiplied by its corresponding block weight. For example, the API that I laid out [above](https://github.com/tskit-dev/tskit/pull/2119#issuecomment-1038326797) would look like this instead:
``` python
sample_sets = [population1, population2, population3] # distinct sets of samples
time_distr = CoalescenceTimeDistribution(ts, sample_sets, weighting_function = count_sample_pairs)

time_grid = numpy.linspace(0., max_time, num_time_points)

# returns array with shape ( length_of_weighting_function_output , num_time_points )
time_distr.ecdf(time_grid)
# etc. -- summary statistics at time points

# returns array with shape ( length_of_weighting_function_output , num_time_points - 1)
# e.g. "time_grid" is used as breaks for windows wherein time-average rates are calculated
time_distr.binned_hazard_rates(time_grid)
# etc. -- summary statistics in time intervals
 
# generates copies with node weights adjusted according to a block resampling scheme
time_distr_boot = time_distr.block_bootstrap(num_blocks, num_replicates, random_seed)
next(time_distr_boot).binned_hazard_rates(time_grid) # resampled summary statistics
```
where `weighting_function` would take a node id, a tree, and a partition of samples; and return a vector of weights for the node. For example, to count the number of coalescing pairs across each population:
``` python
def count_sample_pairs (node, tree, sample_sets):
    weights = [0]*len(sample_sets)
    if tree.is_internal(node):
        ch = tree.children(node)
        for j in range(len(sample_sets)):
            # count number of samples in sample set per child
            ss = sample_sets[j]
            nlineages = np.array([len([i for i in tree.leaves(j) if i in ss]) for j in ch])
            # get the number of subtended pairs
            weights[j] = functools.reduce(sum, (x*y for x, y in itertools.combinations(nlineages, 2)))
    return weights # counts of pairs from each sample set
```

Maybe this is too broad/tangential to discuss here. But (for me at least) would be a more useful functionality than the special case of estimating pair coalescence rate or Ne.

petrelharp:
Nice idea, @nspope - but, since nodes extend over many trees and the number of samples below them can change, the appropriate entry in this table wouldn't be the node, but instead each node could have many rows, one new one each time the samples below it change. @awohns has called this (or something close to it) a "brick".  If pre-computing this would make other things much quicker, then I'm for it; but, I'm not sure - I think computing summaries is pretty close to things we already do?

Here's some thoughts. Recall that right now we have the machinery all set up to keep track of, at each node, how many samples in various sample sets are below that node, and efficiently update this as we iterate over trees.  To keep things simpler, let's consider cross-coalescence between sample sets A and B. For instance, the node statististic corresponding to divergence between A and B counts, for a node n, what proportion of paths between (a, b) pairs pass through the branch above n. Let's call this `p(n)`. The fundamental thing we want to do here is to find out "what proportion of (a, b) pairs coalesce in this node"; if the children of node n are `c1`, ..., `ck` then this is `p(c1) + ... + p(ck) - p(n)`, since a path (a,b) contributes to p(c1) if a is below c1 and b is not (or vice-versa); so if a path contributes to  `p(c1)` but not `p(n)` then the MRCA must be `n`. (*note:* I think there's some double-counting that makes this not quite right, but it can be fixed.) So, this says that to compute what we'd like to, we can take the basic statistics machinery and modify it so that summaries can depend on the difference between the value at a node and its children (rather than just on the value at a node). Alternatively, we could write things summing across *edges*, where we get to know for each edge what is the value at the top and bottom of the edge, and the times of these.

*Edit:* I've just seen the suggestion to have a more general "weighting function" interface. I like it! I'm not yet sure it needs to be a separate class - that depends on the underlying computation, I think? - but, I strongly agree that if we work out what useful generalizations will be then we'll get the right API.

nspope:
@petrelharp ok, I'll work out a prototype and we can see if it's too redundant with the existing stats machinery. I still think it's possible to avoid multiple rows per node (because the total weight for a given node is the sum of weights for that node across marginal trees), but it'll quickly become obvious if I'm wrong!

petrelharp:
> I suppose what I'm suggesting is an interface to get summary statistics from any time-to-event distribution that arises from counting a particular topology in a tree sequence for each node.

Ok, what exactly is this class of statistics? Here's one suggestion (brainstorming): given some samples `a1, ..., am` and some others `b1, ..., bn`, then find the distribution of times of nodes that are MRCAs of `a1, ..., am` but not ancestors of `b1, ..., bn`? (Note that by adding time windows to our existing statistics we'd be able to get the time distribution of ancestors of `a1,... am` but not `b1, ... bn`; what's different here is that it's only the *MR*CAs.) This could be extended to averaging over draws from `m` and `n` sample sets. What did you have in mind?

>  I could use weights that count the possible topologies for 4 samples from different subspecies. Taken together these would give the ingredients for something like a time-aware ABBA/BABA statistic, that I could compare to simulations.

There's an important distinction to get straight here - ABBA/BABA statistics count *mutations*, and therefore behave like statistics of *branch lengths* (since mutations fall on branches). If we just added time windows to our current stats, we'd automatically be able to partition ABBA/BABA out by time. What you're referring to here is like ABBA/BABA but looking at *times* of *most recent* common ancestors. (Instead of, I guess, times of *all* ancestors satisfying some criterion; compare also to the "node" version of [f4](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.f4). This seems like a good idea, but I haven't thought through what the differences would be?

petrelharp:
> I still think it's possible to avoid multiple rows per node (because the total weight for a given node is the sum of weights for that node across marginal trees)

Ah, since each node has a unique time; good point. I'm liking this! Restating - the resulting object would be, essentially, a collection of weighted times; if weights are nonnegative you could think of this as a distribution of times.

One possible issue with the "return all unique times and associated weights" approach (rather than just having a method of the tree sequence that returns the already-binned objects) is that if you want to do this in many windows along the genome (say, 10,000 windows) the size of the output gets rather large.

nspope:
@petrelharp 

> Ok, what exactly is this class of statistics?

I'm thinking things like Kaplan-Meier product limit estimator, the interval hazard rate estimator in the OP (which has a name I'm forgetting) -- just estimators of the hazard function / survival function at particular points in time.

> There's an important distinction to get straight here - ABBA/BABA statistics count mutations, and therefore behave like statistics of branch lengths (since mutations fall on branches). 

Yes, I meant that as an loose analogy to a statistic that is sensitive to introgression, not as an equivalence. What I was getting at is that the distributions of TMRCAs for the different possible topologies of the 4 samples would give not only information about whether introgression was taking place, but when it was taking place. Similarly to how cross-coalescence carry information about both the level of gene flow between population and the timing of gene flow. If there's time on Wed I'll give an example of this from my own work

> Ah, since each node has a unique time; good point. I'm liking this! Restating - the resulting object would be, essentially, a collection of weighted times; if weights are nonnegative you could think of this as a distribution of times.

Yes, that's exactly it! In what I'm envisioning, weights are counts -- so it is a distribution of times (some of which are duplicated), and the weights are the multiplicities of each unique value.

>One possible issue with the "return all unique times and associated weights" approach (rather than just having a method of the tree sequence that returns the already-binned objects) is that if you want to do this in many windows along the genome (say, 10,000 windows) the size of the output gets rather large.

Right, I haven't thought this through for the "local" version. However, I *am* suggesting we keep track of the left-most/right-most extent of each node. If genomic windows are large, this could be used to get local estimates without needing to repeat the table building for each genomic window. Of course, it would be sloppy near the breakpoints of windows. My intent was to use this positional information for bootstrapping (where a little bit of sloppy is OK).

petrelharp:
I think if we get one or two more good examples besides "coalescence rate" then we'll see how to generalize.

> I am suggesting we keep track of the left-most/right-most extent of each node. If genomic windows are large, this could be used to get local estimates without needing to repeat the table building for each genomic window. Of course, it would be sloppy near the breakpoints of windows. My intent was to use this positional information for bootstrapping (where a little bit of sloppy is OK).

Good points in general - just a small note that this doesn't quite work since weights change along the span of a node. (Maybe this is what you mean by "sloppiness", though?)

nspope:
> I think if we get one or two more good examples besides "coalescence rate" then we'll see how to generalize.

I can provide an example of counting distinct topologies of sample trios (from different populations) to jointly estimate migration rates / Ne in time windows. Another thing that springs to mind is Figure 4C in [Spiedel et al](https://www.nature.com/articles/s41588-019-0484-x) -- there they are counting trees where exceptionally long branches are descended from particular nodes.

> Good points in general - just a small note that this doesn't quite work since weights change along the span of a node. (Maybe this is what you mean by "sloppiness", though?)

Yeah, that's what I meant by sloppy. Another option is to specify the genomic intervalswhen the class is instantiated. Then these "split-block" assignments could be recorded while the table is built. For example, when a node crosses a breakpoint (is found in two blocks), it would get two rows in the table, one for each block, that would have different weights.

jeromekelleher:
You mention counting tree topologies @nspope, so pointing out the [combinartorics stuff](https://tskit.dev/tskit/docs/stable/topological-analysis.html#sec-combinatorics) in case you're not aware of it.

petrelharp:
I'll be curious to see the details - there's a lot of possible ways to go with the details.

nspope:
@petrelharp OK I think this is to a point where we can tell if this idea will work or flop. A short example is in the (in progress) test class [here](https://github.com/nspope/tskit/blob/2d513e354267bb2fb13af014e5ad25f9967b4461/python/tests/test_stats.py#L1825-L1832), and the I updated the OP with a longer usage example.

  - For bootstrapping, I ended up dividing trees into blocks (rather than by genomic coordinates). I think this is reasonable -- the number of recombination events per block should be similar. The number of blocks is specified at the outset, and then each stored record is a unique node+block combination. A different strategy will definitely be needed for a local version for genomic windows.
  - With the recombination rate (1e-8) and block size (10 per Mb) I was playing around with, it seems like the storage requirements scale more or less with the number of trees. I don't think that's too bad? Definitely better than for every node in every marginal tree.
  - This would be easy enough to refactor into a standalone function, if you think that'd be better design.

I'm curious what you think. Also (I know it's a bit late to ask this) but how far has this discussion deviated from your original feature idea in #1315? If you think this is getting way over-complicated, I'm fine with pulling back to something more concise.

nspope:
> You mention counting tree topologies @nspope, so pointing out the [combinartorics stuff](https://tskit.dev/tskit/docs/stable/topological-analysis.html#sec-combinatorics) in case you're not aware of it.

Thanks @jeromekelleher! Yes, `tree.count_topologies` is awesome and very close to what I'm after, except it's not quite specific enough about the topology that's being counted. For example, count all embedded subtrees of the forms `((a,b):node,c)` where 'a,b,c' are population labels and 'node' is the node id. The idea would be to associate node times with the frequency of subtrees with a particular labeling.

python/tests/test_stats.py
```diff
@@ -1584,3 +1585,248 @@ def verify(self, ts, reference_sets, focal=None):
...
+            )
+
+        coalesced, _ = np.histogram(pairwise_tmrca, window_breaks)
+        total = len(pairwise_tmrca) - np.append([0], np.cumsum(coalesced))[:num_windows]
+        rate = -np.log(1.0 - coalesced / total) / duration
+
+        return {"coalesced": coalesced, "total": total, "rate": rate}
+
+    def simulate_equilibrium(self, num_chrom, num_samples, recomb_rate=0.0, Ne=10000):
+        # simulate panmictic population of stable size
```
petrelharp, line 1796: Minor note: It looks like maybe you're going for a statistical test here? That's not the sort of thing we have here - instead, we just have "correctness" tests; for that purpose, a small WF simulation is the way to go, since all the weird things happen (with short genome and high enough recomb and mutation rates to get multiple hits of each).

petrelharp:
This looks great! I haven't got into the guts of the computation, but thinking about the API: It seems like the advantage to the "separate class" approach is that it has the pre-computed, weighted list of times, easy for other downstream analysis we haven't thought of.  Right now, we only *use* this list by asking it to be binned in time windows.

I'll sketch out a possible functional alternative, for comparison. Some of the details below are probably wrong, but you get the idea:
```
def coalescence_time_counts(self, sample_sets, indexes, time_windows, windows=None, ...):
    """
    Computes the number of cross-coalescences that occur between each of the pairs of sample_sets indicated by pairs in
    `indexes`, binned by `time_windows` and (genomic) `windows` and weighted by the length of the genome over which they
    occur. The output will be a len(windows) x len(indexes) x len(time_windows) array.
    If indexes[j] is (a, b), then the [i, j, k]th element being the total length of all ancestral 
    chromosomes on which a pair of samples from sample_sets[a] and sample_sets[b] find the MRCA between
    genomic coordinates windows[i] and windows[i+1], and in the time period time_windows[k] to time_windows[k+1].
    """
```
This could be the workhorse, with perhaps `ts.cross_coalescence_time_ecdf(sample_set1, sample_set2, ...)` and `ts.cross_coalescence_rate(..., method="Kaplan-Meier")` providing summary methods?  And, just as we [allow `windows="trees"`](https://tskit.dev/tskit/docs/stable/stats.html#sec-stats-windows) to say "put a window at each distinct tree", we could allow `time_windows="nodes"` to get an output equivalent to what you're computing here?

I'm not sure what's the nicer API - I am leaning towards consistency with the statistics, but could easily be convinced otherwise - what you have sketched out is also very nice. The other consideration is efficiency: if we want to do this in a lot of windows along the genome (and, a heatmap of coalescence rates along the genome would make a very nice picture!) then you'd have to store a lot of things that are already in the tree sequence, arguing for just providing the methods to extract whatever we want.

jeromekelleher:
Thanks for an awesome talk on this today @nspope! Here's some thoughts on API.

I think we should arrange it like this:
```python

class CoalescenceTimeDistribution:
    def __init__(self, ts, sample_sets, ...):
        ...

    def __str__(self):
        return self.useful_text_summary()

    def __repr_html__(self):
        return self.useful_html_summary()

    ...

class TreeSequence:

    ...

    def coalescence_time_distribution(
            self, sample_sets=None, *, weight_func=None,...):
        return CoalescenceTimeDistribution(
            self, sample_sets, weight_func=weight_func, ...)
```

So, we'd use it like
```
dist = ts.coalescence_time_distribution()
print(dist) # summary
```
if you didn't want to faff around with assigning a variable, you can do
```
print(ts.coalescence_time_distribution().ecdf())
```

The clinching argument here in favour of making this an object we return is that we can do things like have a nice repr_html, which will be super useful when working in notebooks. We might even be able to automatically plot out the distribution (if we don't mind having a soft dependency on matplotlib) using some Jupyter magic.

We have precedent for this type of organisation as well: [TreeSequence.ibd_segments]( https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.ibd_segments) returns an [IbdSegments](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.IdentitySegments) instance. This is a flexible and powerful way of doing things, as it gives us full control of thinks like caching, and we can (if we want) go all the way down to C (as we do with IBD segments).

petrelharp:
I like this! But, how about when I want to make a heatmap of the coalescence time distribution in 500 windows along chr1 from 100K samples - how will we deal with this?

nspope:
@jeromekelleher Looks great!

@petrelharp What about making storing the ECDF table an option? So for example, if you set `distr = ts.coalescence_time_distribution(..., precalculate=False)`, then when `distr.interval_hazard_rate(time)` is called it'll walk over edge differences to calculate the needed values. I completely agree that storing the table would get cumbersome with genomic windows. On the other hand, I think it'd be pretty cool to be able to interactively choose the time scale for the heatmap -- for example, to first look at coarse patterns on a log scale and then zoom in to finer resolution over a particular era -- without having to wait.

If this is too complicated, then I'd be fine with doing computations on the fly. It could still be factored as a class like @jeromekelleher outlines above. Maybe the way forward, is for me to modify the table construction to use edge differences -- then we can see how fast it finishes on large-scale datasets. If it's super quick, then the table idea is probably unnecessary.

nspope:
I think the other source of overhead is the user-supplied weight function. This would be a non-issue if there was a built-in way to efficiently count topologies w/ a focal node (like I described today), but for an arbitrary weight function I could see an advantage to only passing over the data once.

petrelharp:
A lot of these questions are about size and speed, so maybe we'll just have to see how things shake out.

jeromekelleher:
Agreed - let's try things out with the above architecture and see how it goes. We should definitely keep the "windows along the genome" part as a key requirement.

nspope:
I just ran across this [totally awesome preprint](https://www.biorxiv.org/content/10.1101/2021.07.13.452277v1.full.pdf) from Osmond and Coop, that looks like it uses first coalescence times of trios in a way that is conceptually similar to what I described in talk (although the underlying model is different). It looks like Matthew Osmond has developed some accompanying software that is already up on Github ... we could ask for his input on this proposed API, with regard to extracting the tree statistics used by his method?

petrelharp:
Hey, good idea - any input, @mmosmond?

python/tests/test_stats.py
```diff
@@ -1603,87 +1603,120 @@ def __init__(self, ts, sample_sets, weight_func, num_blocks=1):
...
+        self.weights = np.zeros((rbuffer, self.num_weights), "int32")
+        self.time = np.zeros(rbuffer)
+
+        # node pointers/weights that are updated via edge differences
+        running_wts = np.zeros((ts.num_nodes, self.num_weights), "int32")
+        running_ptr = np.zeros(ts.num_nodes, "int32") - 1
+
+        self.num_record = 0
+        self.weight_func_evals = 0
+        for edge_diff in ts.edge_diffs():
```
jeromekelleher, line 1633: ts.at() is an expensive operation, you want do do something like this:
```
for edge_diff, tree in zip(ts.edge_diffs(), ts.trees()):
    assert edge_diff.interval == tree.interval
    ....
```

jeromekelleher (commented):
One algorithmic suggestions @nspope 

How do you want to proceed from here? I'd like to get something minimal merged in undocumented soon if we could, so that we can start trying it out and building on it. Once we're happy with the API we can then document it. Doing everything in one "big bang" PR is too hard I think.

What's the minimal functionality we can get in here do you think?

nspope:
Thanks @jeromekelleher that makes sense. 

I was thinking to merge after implementing local estimates across genomic windows. The ETA for this is sometime next week. Once genomic windows are in there, we'll be able to play around w/ realistic use cases. Then decide whether to traverse edge diffs once when the class is instantiated, or every time class methods are called. This choice will probably impact the API.

Does this sound good? Or would you prefer to merge in the next few days? In which case I think I'd just need to put together a minimal test for pytest & move the class into the library proper.

python/tests/test_stats.py
nspope: Yes, at the moment this class isn't actually testing anything -- it only generates some examples, so that I can more easily check for methodological correctness as I work out the algorithm. My plan was to make proper tests when we're about to merge. The intent is to run these through pytest, to compare the implementation in the library proper against the one defined here in `test_stats.py`, right?

nspope: Ultimately I was planning on using [the tests for GNN](https://github.com/tskit-dev/tskit/blob/2d513e354267bb2fb13af014e5ad25f9967b4461/python/tests/test_stats.py#L911) as a template.

jeromekelleher:
SGTM @nspope - keep us posted!

nspope:
Hi @jeromekelleher and @petrelharp,

This is ready for review ... The docstrings are rough, I'd prefer to finish these after the API is finalized if you don't mind. But should be sufficient to understand the design, I think. Here is a usage example:

``` python
import tskit
import msprime
import numpy as np

ts = msprime.sim_ancestry(
    samples=100, 
    ploidy=1,
    sequence_length=1e6, 
    recombination_rate=1e-6, 
    population_size=1000,
    random_seed=1,
)

sample_sets = [[0,1,2,3,4],[5,6,7,8,9]]
genomic_breaks = np.linspace(0.0, 1.0, 10) * ts.sequence_length
time_breaks = np.linspace(0.0, 1.0, 5) * ts.max_root_time

distr = ts.coalescence_time_distribution(
    sample_sets = sample_sets,
    weight_func = 'pair_coalescence_events',
    window_breaks = genomic_breaks,
    blocks_per_window = 10,
)

# output has shape (3, len(time_breaks)-1, len(genomic_breaks)-1)
# rows correspond to coalescences b/w pairs from sets (0,0), (0,1), (1,1)
distr.coalescence_rate_in_intervals(time_breaks)

distr_boot = distr.block_bootstrap(num_replicates=10, random_seed=1)
next(distr_boot).coalescence_rate_in_intervals(time_breaks)
```

python/tskit/trees.py
```diff
@@ -7712,6 +7713,73 @@ def ibd_segments(
...
 
+    def coalescence_time_distribution(
+        self,
+        *,
+        sample_sets=None,
+        weight_func=None,
+        window_breaks=None,
+        blocks_per_window=None,
+    ):
+        """
```
jeromekelleher, line 7724: As we're not putting this into the public API yet we should remove the docstring. I think you can just insert a blank line with a comment or something to do this?

python/tskit/trees.py
```diff
@@ -32,19 +32,20 @@
...
 import math
 import warnings
-from dataclasses import dataclass
 from typing import Any
 from typing import NamedTuple
 from typing import Optional
 from typing import Union
 
 import numpy as np
+from dataclasses import dataclass
```
jeromekelleher, line 41: This is in the wrong group, should be back where it was

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
         return A
+
+
+class CoalescenceTimeDistribution:
+    """
+    Class to precompute a table of sorted/weighted node times, from which to calculate
+    the empirical distribution function and estimate coalescence rates in time windows.
+    """
+
+    class CoalescenceTimeTable:
```
jeromekelleher, line 178: I don't think this needs to be an embedded class. Let's just pull it out into the module scope. It won't be part of the external API unless we say so, so it's fine to put it there.

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
+            assert self.num_records > 0
+            time_order = np.argsort(self.time)
+            self.time = self.time[time_order]
+            self.block = self.block[time_order]
+            self.weights = self.weights[time_order, :]
+            self.num_blocks = 1 + np.max(self.block)
+            self.cum_weights = np.cumsum(self.weights, 0)
+            self.quantile = self.cum_weights / self.cum_weights[-1, :]
+
+        def copy(self):
```
jeromekelleher, line 202: Looks like this copy method isn't used?

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
+        sample_sets=None,
+        weight_func=None,
+        window_breaks=None,
+        blocks_per_window=None,
+    ):
+
+        assert isinstance(ts, trees.TreeSequence)
+
+        if sample_sets is None:
+            self.sample_sets = [list(ts.samples())]
```
jeromekelleher, line 379: This needs testing

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
+        elif all([isinstance(i, (int, np.int32, np.int64)) for i in sample_sets]):
+            self.sample_sets = [sample_sets]
```
jeromekelleher, line 381: Likewise here

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
+        else:
```
jeromekelleher, line 382: There's quite a bit of complexity here around the input types - maybe we could just make it something simple for now, and create an issue to track adding more complex behaviour later?

python/tskit/stats.py
```diff
@@ -163,3 +167,457 @@ def r2_matrix(self):
...
+
+        assert isinstance(times, np.ndarray)
+        assert times.ndim == 1
+
+        values = np.zeros((self.num_weights, len(times), self.num_windows))
+        for k, table in enumerate(self.tables):
+            indices = np.searchsorted(table.time, times, side="right")
+            values[:, :, k] = np.array(
+                [
+                    [0.0] * self.num_weights
```
jeromekelleher, line 460: I think this could be an intermediate value assigned to a varianble? There's a lot of logic happening in here. Also probably more efficient to do things with numpy indexing rather than go through in a list comprehension.

jeromekelleher (commented):
Looks great @nspope!

The only issue for me in merging this as a work-in-progress is that it doesn't have enough testing. We'd definitely want full test coverage as a minimum, so I'd suggest simplifying interfaces for now so that  there aren't so many corner cases. It would also be great to see the methods tested out on a few hand crafted examples, so we can really *see* what the results are. It's a bit tedious, I know, but it's totally worth it.

python/tests/test_stats.py
```diff
@@ -1584,3 +1584,138 @@ def verify(self, ts, reference_sets, focal=None):
...
+        assert B4.shape == (n_weights, n_time - 1, n_interval)
+        assert B5.shape == (n_weights, n_time, n_interval)
+        assert B6.shape == (n_weights, n_time, n_interval)
+        assert np.allclose(B1, B7)
+        assert B8.shape == (n_weights, n_time, n_interval)
+        assert B9.shape == (n_weights, 3, n_interval)
+
+    def test_wright_fisher(self):
+        ts = self._simulate_dtwf(1)
+        self.verify(ts)
```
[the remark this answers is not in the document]
nspope: Thanks, all makes perfect sense! I'll put together some targeted tests on small worked examples.

python/tskit/trees.py
nspope: For some reason one of the precommit hooks is shuffling this import around on my end (`reorder-python-imports`). I guess I can just disable this next time I push.

petrelharp: nah, go with what the pre-commit hooks say!

nspope:
I'm a bit perplexed, these tests are all passing locally but failing the CI. Perhaps I should hard-code the simulated tree sequences used by the tests?

python/tests/test_stats.py
```diff
@@ -1584,3 +1584,885 @@ def verify(self, ts, reference_sets, focal=None):
...
+    1.00 ┊ ┃ ┃ ┃ ┃  6  ┊    0     1        1  0.25
+         ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊
+    0.00 ┊ 0 1 2 4 3 5 ┊    0     0        0  0.00 < to catch OOR
+         0            100
+    Uniform weights on nodes
+    """
+
+    @tests.cached_example
+    def coalescence_time_distribution(self):
+        ts = msprime.sim_ancestry(
```
jeromekelleher, line 1608: Unfortunately we can't rely on the exact outputs of msprime @nspope. They will differ abitrarily depending on platform and underlying GSL version. 

The simplest thing to do here is probably just to hand code up this example using the tables API. You could also output the text representation of the nodes and edges and store them in the file here, if you prefer (see test_topology.py for lots of examples of this)

jeromekelleher (commented):
Looks great @nspope! I think you just need to remove the dependency on msprime.

python/tests/test_stats.py
nspope: Working on that now, thanks! It does seem like the issue is the RNG across platforms.

nspope:
Hi @jeromekelleher looks like everything is getting hit by tests now. Let me know if you think anything needs more work before merging. Thanks!

mmosmond:
> I just ran across this [totally awesome preprint](https://www.biorxiv.org/content/10.1101/2021.07.13.452277v1.full.pdf) from Osmond and Coop, that looks like it uses first coalescence times of trios in a way that is conceptually similar to what I described in talk (although the underlying model is different). It looks like Matthew Osmond has developed some [accompanying software](https://github.com/mmosmond/sparg) that is already up on Github ... we could ask for his input on this proposed API, with regard to extracting the tree statistics used by his method?

Sorry for the slow response -- I'm on parental leave but eventually stitched together enough nap time to read the above and watch the tskit community meeting recording. Good stuff @nspope!

Our method doesn't use first coalescence times of trios. What we do do is 1) calculate the probability of the sample locations under a model of Brownian motion given the shared evolutionary times between each sample, 2) calculate the probability of the coalescence (node) times under a pure birth (Yule) model, and 3) calculate the probability of the coalescence times under the standard neutral coalescent, given a piece-wise constant Ne from Relate. We do this for each individual marginal tree we sample, and we resample branch lengths at those marginal trees for importance sampling. See eqn 9 in our preprint for more details.

So what we need to run our method on a tree sequence is, at each marginal tree we consider, 1) a matrix of the shared evolutionary times between each sample, 2) a vector of the coalescence times, and 3) the ability to resample branch lengths (ideally topologies too one day, but our method would need to be revised). We also need 4) an estimate of Ne. Quantity (1) can be calculated with current stats (see #1619 and linked issues). Quantity (2) is straightforward. I don't think it is easy to do (3) with tsinfer + tsdate at the moment, but @awohns has mentioned it wouldn't be too hard to give easy access to a posterior of branch lengths (if I'm remembering correctly). And finally -- and most relevantly to this PR -- we currently infer the tree sequence with Relate, and use their estimate of Ne.

So I think that if we could estimate Ne from a tree sequence we could run our method -- without importance sampling -- using just tsinfer + tsdate. I think this would be nice for at least two reasons, 1) streamlining our method by using less programs eventually (once importance sampling also possible) and 2) developing new related methods that take advantage of the shared info from one marginal tree to the next. It seems like it is pretty straightforward to get a piece-wise constant Ne from your API, so all looks good to me!

Looking forward to giving this a whirl.

nspope:
Thanks for clarifying @mmosmond, especially while on leave! I must have jumped to conclusions based on Fig 1 in your preprint. And thanks for the feedback on this API as well.

jeromekelleher (approved):
LGTM!

I'm happy to merge this as a stepping stone. 

@petrelharp have you any more thoughts here?

petrelharp:
I'll have a look in the next day or two.

jeromekelleher:
@petrelharp - I think we need your eyes on this so we can merge? 

@nspope - is there much in the new additions?

jeromekelleher:
@nspope @petrelharp - what's the status here? Should we try to get this merged in time for the upcoming 0.5.0 release?

nspope:
@jeromekelleher I think @petrelharp is going to take a look soon (before June), then this could be merged as a WIP.  The functionality people were interested in (Ne estimation) is good to go, just needs some documentation. Next, I'll work on making it faster/more scalable.

petrelharp:
I was hoping to sort this out soon, but the pyslim release has hijacked me. I'll have a look soon, though, if there's a deadline!

jeromekelleher:
I don't think there is a deadline as we're not documenting it, so can go either before/after upcoming release.

nspope:
@petrelharp and I just talked over this PR:
- The block bootstrap method used here would be great to incorporate into the general stats interface.
- To make that possible, the machinery here needs to be shifted so that rather than summing/storing weights per node, we're summing/storing weights per time window.
- Doing so will also make it possible to calculate these statistics across windows on very large tree sequences (storage costs are prohibitive for large tree sequences in the current implementation).
- If the user has the option of specifying a "general weight function" for nodes, then this function should have the form `def weight_func(parent, child, tree, sample_sets):` (e.g. operate on edges) as this will ensure the edge-diff-based update scheme works regardless of the function.

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+
+        # nodes that have been removed from tree
+        removed = {i.child for i in edge_diff.edges_out if tree.is_isolated(i.child)}
+        # TODO: What if sample is removed from tree? In that case should all
+        # nodes be updated for trio first coalescences?
+
+        # nodes where descendant subtree has been altered
+        modified = {i.parent for i in edge_diff.edges_in}
+        for i in copy.deepcopy(modified):
+            while not tree.parent(i) is tskit.NULL and not tree.parent(i) in modified:
```
jeromekelleher, line 320:
```suggestion
            while tree.parent(i) != tskit.NULL and not tree.parent(i) in modified:
```

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+        can be used in several intervals. Thus, the concatenation of ECDF
+        tables across multiple intervals is not the same as the ECDF table
+        for the union of those intervals. Trees within intervals are chunked
+        into roughly equal-sized blocks for bootstrapping.
+        """
+
+        assert tree.interval.left <= left and right > left
+
+        # assign trees in window to equal-sized blocks with unique id
+        _tree = tree.copy()
```
jeromekelleher, line 347: ``_tree`` isn't a great variable name I think. Maybe ``other_tree`` or ``tree2`` or something?

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+                _tree.next()
+        tree_idx = np.arange(tree.index, _tree.index + 1) - tree.index
+        tree_offset = tree.index
+        num_blocks = min(self.num_blocks, len(tree_idx))
+        tree_blocks = np.floor_divide(num_blocks * tree_idx, len(tree_idx))
+
+        # calculate span weights
+        # TODO: if bootstrapping, does block span need to be tracked
+        # and used to renormalise each replicate?
+        _tree = tree.copy()
```
jeromekelleher, line 362: If the interval isn't too big it would be more efficient to seek back to the original position, like ``
```
tree2.seek(tree.interveral.left)
```

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+
+        # storage if using single window, block for entire tree sequence
+        buffer_size = self.buffer_size
+        table_size = buffer_size
+        time = np.zeros(table_size)
+        block = np.zeros(table_size, "int32")
+        weights = np.zeros((table_size, self.num_weights))
+
+        # assemble table of coalescence times in window
+        indices = np.zeros(tree.tree_sequence.num_nodes, "int32") - 1
```
jeromekelleher, line 379: We usually use ``dtype=np.int32`` - saves a little bit of string looking up, presumably. Bit more explicit to use the named param for dtype also.

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+        last_block = np.zeros(tree.tree_sequence.num_nodes, "int32") - 1
+        num_record = 0
+        while tree != tskit.NULL:
```
jeromekelleher, line 382: This is never False - you probably meant ``while tree.index != tskit.NULL``? If so, probably missing some coverage in the tests.

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+                    block[idx] = current_block
+                    num_record += len(nodes_to_add)
+                weights[indices[nodes_in_tree], :] += (
+                    span_weight * running_weights[nodes_in_tree, :]
+                )
+
+            if tree.interval.right < right:
+                # if current tree does not cross window boundary, move to next
+                tree.next()
+                self._update_weights_by_edge_diff(
```
jeromekelleher, line 415: If we're mutating the input tree anyway here, do we really need to make a full copy above? If the window size isn't too big and we're not moving too many trees along the genome (and the sample size is large) then avoiding a full copy could be a significant perf win.

python/tskit/stats.py
```diff
@@ -163,3 +167,506 @@ def r2_matrix(self):
...
+                    tree, next(edge_diffs), running_weights
```
jeromekelleher, line 416: Is this ``next`` guaranteed to never raise a StopIteration? The I guess so, since we have a strict ``< right`` on the conditional on the way in?

python/tests/test_stats.py
```diff
@@ -1584,3 +1584,1066 @@ def verify(self, ts, reference_sets, focal=None):
                 mean = np.sum(T * L, axis=1) / np.sum(L)
                 assert np.allclose(mean, A[j])
         return A
+
```
jeromekelleher, line 1587: These tests are pretty self-contained, so I think it would be good to keep them in their own file. Say, ``test_coaltime_distribution.py``?

jeromekelleher (approved):
Very nice - some minor comments. I wonder if this could be computed from a single pass over the tree iterator, but that's something that could be tried out afterwards. I vote we merge and refine.

petrelharp:
A note about the final note - actually, all the stats we have in mind here can be expressed as additive, time-windowed summaries that follow the 'general stats' framework with the modification that it is additive across *edges*, the time window associated with an edge is the time of the parent, and the summary function of an edge gets to know both parent weights and child weights. (So, you wrote `weight_func(parent, child, tree, sample_sets)` but I think it's actually `summary_func(parent_weights, child_weights)`.)

One upside of this definition is that the implementation assumes that when a bit of the tree is changed, this only affects summaries further up the trees; so, as currently written it's easy to write summary functions that won't be calculated correctly.

The main question in my head about this is whether mirroring the general stats implementation is the way to go.

python/tskit/stats.py
nspope: Makes sense! I'll stick in a TODO for now. This "blocking scheme" is going to get revised anyway, because of generalizing the bootstrap machinery.

python/tskit/stats.py
nspope: Yes, as long as `right <= ts.sequence_length` -- which should be safe to assume given that this function is only used internally.

nspope:
sorry for letting this go stale @jeromekelleher -- I think it should be good to merge as a work-in-progress now.

jeromekelleher (approved):
Let's merge!

@nspope - would you mind going through and opening some issues to track what we need to do as follow-ups? I think it's probably worth making a "project" for this so we can follow progress, and be clear on when it's ready for release/documenting.

PR #2120, opened 2022-02-04
Provide direction argument to edge_diffs.

jeromekelleher:
Adds a "direction" argument to edge_diffs to allow us to sequentially generate trees in reverse.

Moves the implementation to Python resulting in a roughly 2X perf regression. I doubt this will be significant in practise, because people will usually be *doing* something with the diffs, which will probably be more expensive. Nonetheless, it is a regression so we should be careful.

If we decide we don't mind about the perf loss for this operation, we can delete some of the Python-C interface boilderplate.

I also modernised the tests for this a bit, while I was in there.

cc @astheeggeggs

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -63,6 +63,12 @@
 - Add Colless tree imbalance index.
   (:user:`jeremyguez`, :user:`jeromekelleher`, :issue:`2250`, :pr:`2266`, :pr:`2344`).
 
+- Add ``direction`` argument to ``TreeSequence.edge_diffs``, allowing iteration
+  over diffs in the reverse direction. NOTE: this comes with a ~10% performance
+  regression as the implementation was moved from C to Python for simplicity
+  and maintainability. Please open an issue if this affects your application.
+  (:user:`jeromekelleher`, :user:`benjeffery`, :pr:`2120`).
+
 **Breaking Changes**
 
 - The JSON metadata codec now interprets the empty string as an empty object. This means
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -146,12 +146,6 @@ typedef struct {
 typedef struct {
     PyObject_HEAD
     TreeSequence *tree_sequence;
-    tsk_diff_iter_t *tree_diff_iterator;
-} TreeDiffIterator;
-
-typedef struct {
-    PyObject_HEAD
-    TreeSequence *tree_sequence;
     tsk_variant_t *variant;
 } Variant;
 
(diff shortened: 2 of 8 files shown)


hyanwong:
> Moves the implementation to Python resulting in a roughly 2X perf regression. I doubt this will be significant in practise, because people will usually be _doing_ something with the diffs, which will probably be more expensive. Nonetheless, it is a regression so we should be careful.

I'm fairly sure that we use the edge diffs a lot in tsdate, don't we @awohns ? Perhaps it's worth testing if this causes any perf problems there (I suspect not though, as we are doing some maths on each diff)

awohns:
That's right @hyanwong, we use the edge diffs a lot. We iterate over them in both the prior and the inside/outside algorithm. Might be good to check if this has a significant effect on tsdate runtime?

jeromekelleher:
That's a good test I think - if this doesn't have a significant impact on a time of a run of tsdate, then I think getting rid of the C linkage is fine, and a small perf regression is something we can live with. If it makes tsdate 2x slower though, then I think we'll have to keep the current C-based implementation for the forwards direction anyway.

jeromekelleher:
@benjeffery - should be push this in for 0.5.0?

@awohns - what's the simplest way of testing out the perf implications for tsdate?

benjeffery:
If we are doing this (which seems the right thing to do I think) then yes it should be in 0.5.0.

petrelharp:
Oh, this would be nice! We were just wanting this!

awohns:
Results of some perf testing on this:
`ts.edge_diffs()` is used in two places in tsdate. The more expensive usage is constructing the prior. Here are the runtimes for that operation on an inferred tree sequence from 1000 genomes chromosome 22. _**Current**_ is the current implementation of `edge_diffs()` and _**Direction**_ is the implementation in this PR:

Current: 6 mins 44 seconds. Direction: 7mins 41 seconds.

jeromekelleher:
Right, that's too much of a difference. Will need to think this through again - thanks @awohns!

jeromekelleher:
Hmm, so looks like most of the time for this code is actually spent constructing edge objects and the difference is really about the different code paths we take for that now. Some stats:

- C code path, yield None (i.e., don't create any edges): 0.98 seconds
- Python code path, yield None: 1.91 seconds
- C code path, yield lists of tuples: 1.19 seconds
- Python code path, yield lists of tuples: 4.05 seconds
- C code path, yield list of locally constructed Edge: 7.16 seconds (This is the current implementation in main)
- Python code yield list of Edge using self.edge():  13.66 (implementation Wilder tried out)
- Python code yielding list of locally constructed Edge: 10.55 seconds

So, I think there's room for quite a bit of improvement here and we did have a pretty big perf regression.

I wonder if edge metadata is really worth it here, and if we should do something like this:
```
# fully locked down dataclass
class DiffEdge:
     left: float
     right: float
     parent: int
     child: int
     id: int
    
     @property
     def metadata(self):
          raise ValueError("Metadata access not supported on edge diffs. Please use ``ts.edge(e.id).metadata`` instead")
```

I think if we can get back to the performance of yielding lists of tuples, we can switch the the Python code path and get rid of some C complexity, as this will be a perf improvement over what's currently in the library.

@benjeffery, any thoughts?

benjeffery:
I've pushed an additional commit here which removes a function call in the `Edge` creation code path.

Here are the timings on my machine for 100k trees of 1.6M edges:

`main` 3s
This PR before commit (`ts.edge`): 7.12s
`Edge(*tuple)`: 4.8s
`Edge(*tuple) with `own_init`: 3.73s

So we're losing about 25% vs `main` with this python only code. I think that is acceptable and we should wrap this up by removing the C code and documenting the function if you agree?

benjeffery:
Using a simple DiffEdge as you described above gets you to 3.55 vs 3.73. It's not nothing, but I don't think that is a big enough win to justify adding a parallel class.

jeromekelleher:
OK then, I guess we should put this to the end-users. @awohns, would you mind trying this version out on your benchmark and letting us know if you're happy with the perf?

jeromekelleher:
> OK then, I guess we should put this to the end-users. @awohns, would you mind trying this version out on your benchmark and letting us know if you're happy with the perf?

Bumping this one @awohns - as a key user of edge_diffs we'd like to know if the perf drop here is acceptable for your applications. Would you mind checking your benchmark again? (It's also probably relevant for your "bricking" code). We'd like to release 0.5.0 in the next few days.

awohns:
Sorry, didn't see this until now @jeromekelleher.

So here's the timing for creating the tsdate prior of NYGC 1000 genomes chr22 (same file as before; 453,350 trees):
Current: 6 mins 29 seconds (389 seconds)
Direction: 7 mins 14 seconds (434 seconds)
So it's ~12% slower

And here's the perf for running `[naive_split_edges()](https://github.com/awohns/ldgm/blob/a152cb6db42658ab24d973757979af83a1c9c6d5/ldgm/bricks.py#L42)` on the first 30 MB of the same tree sequence (148,437 trees):
Current: 8 mins 57 seconds
Direction: 8 mins 2 seconds (482 seconds)
Which I guess means that `edge_diffs()` is not really a contributing factor to the overall runtime.

jeromekelleher:
Thanks @awohns. I think you should have the casting vote here on whether we keep the existing C code or not then - 10% is a significant perf drop, but not a major one and probably wouldn't be noticed. I honestly don't mind either way as we have the code in both cases (we can keep the current C-based approach as the default), I just want someone else to make the decision!

awohns:
Thanks for seeking my feedback @jeromekelleher and for input @benjeffery! 
I think that 10% is acceptable from my perspective, especially given that `edge_diffs` is not a dominant factor in the runtime of either `tsdate` or `ldgm`. So I'm happy for this to be merged!

jeromekelleher:
OK, great! @benjeffery - do you want to do the rest here or will I pick it up?

benjeffery:
> OK, great! @benjeffery - do you want to do the rest here or will I pick it up?

I can take this from here!

jeromekelleher:
whoops, took off the AUTOMERGE label there @benjeffery as I'd left some embarassing intermediate commits.Will squash those now.

jeromekelleher:
This also still leaves in the C infrastructure as dead code. Will pull out now.

jeromekelleher:
Documentation isn't finished either. Marking as draft.

benjeffery:
> whoops, took off the AUTOMERGE label there @benjeffery as I'd left some embarassing intermediate commits.Will squash those now.

Gah, sorry. I didn't think properly about the C and docs.

jeromekelleher:
OK, should be ready to go now.

PR #2121, opened 2022-02-04
issue 1340: changed mrca signature to take more than 2 args.

### LINKED ISSUE #1340: Change Tree.mrca signature to take > 2 args
Following the discussion in #1339, we should change the signature of Tree.mrca to accept more than two nodes. The initial implementation can be something like

```python
def mrca(self, *args):
   return functools.reduce(self._ll_tree.mrca, args)
```

We should check that ``Tree.mrca(u) == u``, which seems perfectly reasonable to me. We're not going to break anyone's code by doing this, and it's an obvious generalisation. We can make a more performant implementation in the future if we like, but this'll do fine for now.

savitakartik:
## Description

changed tree.mrca signature to take more than 2 args.
added tests for the changed functionality in a new class TestMRCA.

Fixes #1340  <- Putting the issue number here will auto-close the issue when this PR is merged 

# PR Checklist:

- [x] Changelogs, if there are API changes. (logged to python/CHANGELOG.rst, but might have mentioned the wrong PR ID in the changelog, sorry!)

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -12,6 +12,8 @@
 - The JSON metadata codec now interprets the empty string as an empty object. This means
   that applying a schema to an existing table will no longer necessitate modifying the
   existing rows. (:user:`benjeffery`, :issue:`2064`, :pr:`2104`)
+- ``tree.mrca`` now takes 2 or more arguments.
+  (:user:`savitakartik`, :issue:`1340`, :pr:`2121`)
 
 
 ----------------------
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -851,6 +851,40 @@ class TestTreeTraversals:
         assert tree_orders == expected_result
 
 
+class TestMRCA:
+    t = tskit.Tree.generate_balanced(3)
+    #  4
+    # ┏━┻┓
+    # ┃  3
+    # ┃ ┏┻┓
+    # 0 1 2
+
+    def test_two_or_more_args(self):
+        assert self.t.mrca(2, 1) == 3
+        assert self.t.mrca(0, 1, 2) == 4
+
+    def test_less_than_two_args(self):
+        with pytest.raises(ValueError):
+            self.t.mrca(1)
+
+    def test_no_args(self):
+        with pytest.raises(ValueError):
+            self.t.mrca()
+
+    def test_same_args(self):
+        assert self.t.mrca(0, 0, 0, 0) == 0
+
+    def test_different_tree_levels(self):
+        assert self.t.mrca(0, 3) == 4
+
+    def test_out_of_bounds_args(self):
+        with pytest.raises(ValueError):
+            self.t.mrca(0, 6)
+
+    def test_virtual_root_arg(self):
+        assert self.t.mrca(0, 5) == 5
+
+
 class TestMRCACalculator:
     """
     Class to test the Schieber-Vishkin algorithm.
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1017,16 +1017,17 @@ class Tree:
         # Deprecated alias for mrca
         return self.mrca(u, v)
 
-    def mrca(self, u, v):
+    def mrca(self, *args):
         """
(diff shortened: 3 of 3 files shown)


python/tests/test_highlevel.py
```diff
@@ -851,6 +851,52 @@ def test_minlex_postorder_multiple_roots(self):
...
 
 
+class TestMRCA:
+    t1 = tskit.Tree.generate_balanced(3)
+    #  4
+    # ┏━┻┓
+    # ┃  3
+    # ┃ ┏┻┓
+    # 0 1 2
+
```
jeromekelleher, line 861: When I'm reading the tests I need to keep in mind what t1 and t2 are. Unless there's a particular property of the 2 sample tree we're using, I think it'd be easier to just have a single tree that we're using as an example.

python/tests/test_highlevel.py
```diff
@@ -851,6 +851,52 @@ def test_minlex_postorder_multiple_roots(self):
...
+    t2 = tskit.Tree.generate_balanced(2)
+    # 2
+    # ┏┻┓
+    # 0 1
+
+    def test_two_or_more_args(self):
+        t2 = self.t2
```
jeromekelleher, line 868: There's probably not much point in assinging to a local variable here, can just do
```
self.tree.mrca(0, 1) == 2
# etc
```

benjeffery (approved):
LGTM! Thanks @savitakartik,not sure why the coverage is complaining. Fix the conflict and we'll see how it goes on a second run.

python/tskit/trees.py
```diff
@@ -1017,16 +1017,16 @@ def get_mrca(self, u, v):
...
 
-    def mrca(self, u, v):
+    def mrca(self, *args):
         """
         Returns the most recent common ancestor of the specified nodes.
-
-        :param int u: The first node.
-        :param int v: The second node.
-        :return: The most recent common ancestor of u and v.
+        :param int args: input node IDs.
```
jeromekelleher, line 1023: I'm not sure what the right way to make a docstring for variable args input like this - @benjeffery, any ideas?

python/CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@
 - ``VcfWriter.write`` now prints the site ID of variants in the ID field of the output VCF files.
   (:user:`roohy`, :issue:`2103`, :pr:`2107`)
 
+- ``tree.mrca`` now takes 2 or more arguments.
+  (:user:`savitakartik`, :issue:`1340`, :pr:`2120`)
```
jeromekelleher, line 11: Yes, this should be pr 2121 I think

jeromekelleher (commented):
Looks great @savitakartik! Some minor comments about the tests, but overall this is what we're looking for.

Looks like you need to resolve some conflicts and update your branch. See the [this guide](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#github-workflow) for help on rebasing your branch. (Don't worry if things go a bit wrong, this happens to everyone the first few times!)

python/tskit/trees.py
benjeffery: Good catch, this isn't rendering correctly for me at least. I think this is the best you can do:
```suggestion

        :param int `*args`: input node IDs, must be at least 2.
```

jeromekelleher:
Looks great, thanks @savitakartik. I think it's ready to go, after a squash and rebase. See the stdpopsim docs I linked to above for help on this.

@benjeffery or I would be happy to help you through the process, if you like - it is definitely confusing the first few times!

benjeffery:
> Looks great, thanks @savitakartik. I think it's ready to go, after a squash and rebase. See the stdpopsim docs I linked to above for help on this.
> 
> @benjeffery or I would be happy to help you through the process, if you like - it is definitely confusing the first few times!

Yep, happy to do it, or to talk you through it on slack/call. LMK!

jeromekelleher (approved):
Good to go :tada:

savitakartik:
thanks for your help with my first PR, Jerome and Ben!

PR #2124, opened 2022-02-08
Use KAS_BORROWS_ARRAY

### LINKED ISSUE #2111: Zero copy writing to kastore
We should now able able to write columns to the kastore without creating an intervening copy using KAS_PUT_BORROWS_ARRAY in kastore 2.0.1. 

Should be a simple enough change. It would be good to record the memory profile of writing out a large file here as well, to confirm that we're getting the peak memory usage benefits we expect.

benjeffery:
Fixes #2111 

Draft as still needs memory perf to check we have gains.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -350,6 +350,7 @@ write_offset_col(
     uint32_t *offset32 = NULL;
     tsk_size_t len = col->num_rows + 1;
     tsk_size_t j;
+    int32_t put_flags = 0;
     int type;
     const void *data;
     bool needs_64 = col->offset_array[col->num_rows] > UINT32_MAX;
@@ -361,6 +362,7 @@ write_offset_col(
     if (options & TSK_DUMP_FORCE_OFFSET_64 || needs_64) {
         type = KAS_UINT64;
         data = col->offset_array;
+        put_flags = KAS_BORROWS_ARRAY;
     } else {
         offset32 = tsk_malloc(len * sizeof(*offset32));
         if (offset32 == NULL) {
@@ -372,8 +374,9 @@ write_offset_col(
         }
         type = KAS_UINT32;
         data = offset32;
+        /* We've just allocated a temp buffer, so kas can't borrow so leave put_flags=0*/
     }
-    ret = kastore_puts(store, offset_col_name, data, (size_t) len, type, 0);
+    ret = kastore_puts(store, offset_col_name, data, (size_t) len, type, put_flags);
     if (ret != 0) {
         ret = tsk_set_kas_error(ret);
         goto out;
@@ -392,7 +395,7 @@ write_table_ragged_cols(
 
     for (col = write_cols; col->name != NULL; col++) {
         ret = kastore_puts(store, col->name, col->data_array, (size_t) col->data_len,
-            col->data_type, 0);
+            col->data_type, KAS_BORROWS_ARRAY);
         if (ret != 0) {
             ret = tsk_set_kas_error(ret);
             goto out;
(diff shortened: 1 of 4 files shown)


c/tskit/tables.c
```diff
@@ -10718,14 +10721,14 @@ tsk_table_collection_write_format_data(const tsk_table_collection_t *self,
     /* This stupid dance is to workaround the fact that compilers won't allow
      * casts to discard the 'const' qualifier. */
     tsk_memcpy(format_name, TSK_FILE_FORMAT_NAME, sizeof(format_name));
-    ret = write_table_cols(store, write_cols, 0);
+    ret = write_table_cols(store, write_cols, options, 0);
```
jeromekelleher, line 10724: This is the only place that we're **not** calling write_table_cols with the BORROW option, so I think it would be simpler to write a different function ``write_table_const_cols`` or something, which passes in the const values and avoids the stupid dance that the comment above here is complaining about.

Everywhere else is doing KAS_BORROWS_ARRAY right?

jeromekelleher (commented):
Looks good, but I think there's a simpler approach where we deal with the (small number) of const columns separately.

c/tskit/tables.c
benjeffery: I think so - so the idea would be to not pass kas_flags at all and use the const status to determine if we borrow?

benjeffery:
At the moment this is having no effect on max ram, digging in to check the changes are being used.

c/tskit/tables.c
jeromekelleher: Yeah - we borrow everything in the current code path (except the case you've identified where we're casting), and make a new code path to accept the const values where we don't borrow.

benjeffery:
> Plots from [memory-profiler](https://pypi.org/project/memory-profiler/) are super helpful for this, should let us see the different stages on a simple script like
> 
> ```
> ts = tskit.load("big-freakin-file.trees")
> ts.dump("big-freakin-file-copy.trees")
> ```
> 
> I guess we might have to manually turn up the sampling interval to see accurately what's happening though.

This is actually an interesting challenge as the write arrays are copied then freed without control going back to python. `/usr/bin/time -v` shows max ram - but this is the loading value as that still copies. I think there is a way by making the tables, then dumping.

jeromekelleher:
If you run it with ``mprof python script.py`` you can use the ``--interval`` option to set the sampling interval. This runs it as a forked subprocess, and doesn't know anything about Pytohn.

jeromekelleher:
Kablammo! Nicely done @benjeffery - it's rare we get such clear wins (you should post this to the slack, in case folks haven't seen it).

c/tskit/tables.c
benjeffery: Sorry, I'm confused here.
The arrays that are passed to kastore are all const as defined in `write_table_col_t`. In fact I just deleted the "stupid dance" (now in this PR) without changing anything else and nothing seems broken.

benjeffery:
@jeromekelleher As discussed I've bought the file format writing up a level so all the arrays to write have the correct lifetime. This simplifies things so we have one code path for everything (except the 32bit arrays).

c/tskit/tables.c
```diff
@@ -10790,12 +10761,36 @@ tsk_table_collection_dumpf(
...
-     * to modify the return value. */
-    ret = tsk_table_collection_write_format_data(self, &store, options);
+    /* Write format data */
+    ret = tsk_generate_uuid(uuid, 0);
     if (ret != 0) {
         goto out;
     }
+
+    ret = write_table_cols(&store,
+        (write_table_col_t[]){
```
jeromekelleher, line 10771: Can we assign this to a local variable please? Took me quite a bit to figure out what this was doing.

c/tskit/tables.c
```diff
@@ -372,8 +374,9 @@ write_offset_col(
         }
         type = KAS_UINT32;
         data = offset32;
+        /* We've just allocated a temp buffer, so kas can't borrow */
+        ret = kastore_puts(store, offset_col_name, data, (size_t) len, type, 0);
     }
-    ret = kastore_puts(store, offset_col_name, data, (size_t) len, type, 0);
```
jeromekelleher, line 376:
```
ret = kastore_puts(store, offset_col_name, data, (size_t) len, type, put_flags);
```

I think this is more robust and avoids the "don't check ret immediately" problem, which is always an error waiting to happen in my book.

benjeffery:
@jeromekelleher all fixed up!

jeromekelleher (approved):
:tada:

PR #1937, opened 2021-11-24
Remove general_stat cacheing

brieuclehmann:
Here's a first pass at removing cacheing to reduce memory consumption in the general_stat framework. I've only tweaked the branch C function, and the python tests are passing.

Without caching, however, `ts.genetic_relatedness` is still slower. For the script below, with caching takes about 10s, without caching: 20s. When I increased n_samples to 1000: with caching - 58s; without caching - 110s.

I may have introduced some unnecessary computation so would appreciate a sense-check!

```python
import itertools
import time
import msprime
import numpy as np
import tskit
from memory_profiler import profile

fp = open("memprof_cache.log", "a+")
@profile(stream=fp)
def genetic_relatedness_matrix(ts, sample_sets, indexes, mode):
    x = ts.genetic_relatedness(sample_sets, indexes, mode=mode, proportion=False, span_normalise=False)
    return x

seed = 42
n_samples = 500
ts = msprime.simulate(
    n_samples,
    length=1e6,
    recombination_rate=1e-8,
    Ne=1e4,
    mutation_rate=1e-7,
    random_seed=seed,
)

n_ind = int(ts.num_samples / 2)
sample_sets = [(2 * i, (2 * i) + 1) for i in range(n_ind)]

n = len(sample_sets)
indexes = [
    (n1, n2) for n1, n2 in itertools.combinations_with_replacement(range(n), 2)
]

start_time = time.time()
x = genetic_relatedness_matrix(ts, sample_sets, indexes, 'branch')
end_time = time.time()
print(end_time - start_time)
```

Issue #1547, opened 2021-06-30
add more linear algebra tools


petrelharp:
Let `G` be the "extended genotype matrix", whose rows are indexed by mutations (not sites) and samples, with `G[i,j] = 1` if the sample has inherited that mutation and `0` otherwise. This is something people want to work with, but for many applications we don't need to actually return the matrix, but instead just multiply it by things. For instance:

1. If `s` is a vector of effect sizes, then `G^T s` gives a vector of additive phenotypes computed by adding up the entries of `s` corresponding to all the mutations each sample has inherited
2. `G^T G` is related to the (site) genetic relatedness matrix; so #1246 is something in this direction.
3. `G G^T`` is similarly related to the LD matrix, and multiplying the LD matrix by things is important in eg [estimating heritabiliy](https://huwenboshi.github.io/hess/)
4. `G x` is similar to what we've called [`trait_covariance`](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.trait_covariance) with `mode="site", windows="sites"` - except that gives something per site, not per mutation.

Here "related to" means "up to some normalization that I'm not figuring out right now".

There's API and computational issues to work out here. For instance, we *could* just implement methods that do `G x` and `G^T y`, and use these to get `G^T G x`; however, that's not how we currently do things in #1246.

Preliminary brainstorming: besides #1246,

- `G^T s`: maybe `ts.additive_phenotypes(samples, effect_sizes)` would return `G^T effect_sizes`. Also, we probably want `ts.individual_phenotypes(individuals, effect_sizes, dominance_coefficients)` that does the analogous thing for diploids. (But, maybe it'd be nice to have a single function that you can pass either sample IDs or individual IDs somehow.)

- I need to better understand what computations are done in LD space to see if this is the right thing, but possibly we should implement `G G^T x` as `ts.ld_matrix_multiplication(x)`.

- And, maybe `G x` would be an option to `trait_covariance`? Or, `mutation_covariance(x)`?

gregorgorjanc:
@petrelharp sorry for being pedantic, but `G^T s` gives you additive genetic (or breeding) values, not phenotypes. To get phenotypes we would have to add to `G^T s` at least environmental values, as well as effects, appropriate for a given trait, such as sex, age, …

[...]

PR #2135, opened 2022-02-22
Makes stats cache optional

jeromekelleher:
Alternative approach for #1937

### DIFF
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -180,7 +180,13 @@ def naive_branch_general_stat(
 
 
 def branch_general_stat(
-    ts, sample_weights, summary_func, windows=None, polarised=False, span_normalise=True
+    ts,
+    sample_weights,
+    summary_func,
+    windows=None,
+    polarised=False,
+    span_normalise=True,
+    cache_summary=True,
 ):
     """
     Efficient implementation of the algorithm used as the basis for the
@@ -200,8 +206,6 @@ def branch_general_stat(
     time = ts.tables.nodes.time
     parent = np.zeros(ts.num_nodes, dtype=np.int32) - 1
     branch_length = np.zeros(ts.num_nodes)
-    # The value of summary_func(u) for every node.
-    summary = np.zeros((ts.num_nodes, result_dim))
     # The result for the current tree *not* weighted by span.
     running_sum = np.zeros(result_dim)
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This is tricky. We definitely want to keep a cache of some sort because we evaluate summary(u) many more times otherwise (not just twice I realise, we also evaluate summar(u) twice for each branch we traverse up the tree as we add and remove edges) and the summary_func is one of the major contributors to run time as it is.

What we could do is have a slightly more intelligent cache where we only store the values of summary(u) for nodes that are currently in the tree, which would reduce the storage a bit (at the cost of making things a bit more complicated). 

But then, for a coalescent simulation we only have O(n + t) nodes, so we're only really going to benefit from this for (say) n=1000 and t= 10^6. The reality is that we need O(n^3) storage to do this because we need to cache O(n^2) space for each node currently in the tree, unless we want to evaluate the O(n^2) summary function quite a few more times. Although, I guess the actual evaluation of branch_length * summary[u] is also O(n^2), so maybe here's not much difference really.

Any thoughts @petrelharp?

petrelharp:
Here's some thoughts:

- The cacheing is saving us only a constant multiple on the number of summary_func evaluations, since as you say, it's updating it twice for each added / removed edge; but we're going to have to evalute summary_func again for all such updated nodes anyways. However, this constant might be substantial.
- So, we could avoid cacheing entirely if we were instead clever about which nodes needed to be updated, and in which order. For instance, if we kept track of which nodes needed their contributions updated and then evaluated summary_func for those just once after edges had been added and removed, then we'd have nearly the minimal number of requried summary_func calls (we'd have one extra call per node - for instance, we'd call summary_func(u) when u is added and also when it is removed).
- I don't know how calling summary_func compares to looking up the pre-computed value? The later presumably costs something, especially when the table of such things gets big?
- An alternative *might* be something along the lines of [this algorithm](https://github.com/tskit-dev/tskit/issues/1547#issuecomment-882235129) - I *think* it'd be possible to compute the n^2 covariance matrix by cacheing at each node a vector of length n (something like the covariance of that node with all samples over some set of SNPs) and then propagating these length-n vectors downwards (as described in that link) until they get to the samples; the length-n vector at each sample would then be the rows of the matrix. I'm unclear right now how this would compare to what we're talking about, or if if would generalize beyond covariance.

PR #2032, opened 2021-12-08
Column takeset

jeromekelleher:
Provide x_table_takeset_columns

Part of #2016, #1977, #537, #538

This is my initial outline for discussion, and completely untested. What do you think @benjeffery?

PR #2143, opened 2022-03-03
Use `takeset` to make loading zero-copy

### LINKED ISSUE #2131: Use column takeset and kastore GET_TAKES_OWNSHIP for zero-copy loading
Should be a case of setting the appropriate flags to kastore_open and using takeset when reading the individual table columns.

benjeffery:
Fixes #2131 

Here is a comparison of this main and this branch's memory use (main first):
main
branch

Almost all of the remaining overhead is due to casting offsets up to 64bit, which uses 3x their size, with that off we get:
branch_nocast

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -52,7 +52,6 @@ typedef struct {
     int data_type;
     tsk_size_t **offset_array_dest;
     tsk_flags_t options;
-    tsk_size_t *offset_array_mem;
 } read_table_ragged_col_t;
 
 typedef struct {
@@ -155,7 +154,6 @@ cast_offset_array(read_table_ragged_col_t *col, uint32_t *source, tsk_size_t num
         ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
-    col->offset_array_mem = dest;
     *col->offset_array_dest = dest;
     for (j = 0; j < len; j++) {
         dest[j] = source[j];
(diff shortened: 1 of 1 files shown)


petrelharp:
Awesome to see this moving. What's memtestmsp.py doing?

benjeffery:
All the tests finally pass here. Some feedback on how I've solved the edge cases (e.g. `if` after `out:`) would be good.

jeromekelleher:
Let's get #2032 squared away first, will make it easier to see the new stuff here. I think we're pretty sold on the takeset idea basically working anyway, right?

benjeffery:
Here's a clean example that we're getting what we expect in a very simple case - this is a tree sequence with only one row in the populations table: `tc.populations.add_row(metadata=b"a"*10_000_000_000)` and the memory usage when we load it to a `TableCollection`:
main_big_ragged
main_big_ragged

jeromekelleher:
Brilliant! So all working as expected now then?

benjeffery:
> So all working as expected now then? 

Yes, I believe so! It has been useful to break all this down and check we understand it all. Basically,  the 0.05seconds after 0.2 are snipped out and don't happen, so as well as the memory usage falling the load is quicker!

benjeffery:
@mergifyio rebase

benjeffery:
Now that I've got a clean diff I can see a couple of errors, will fix today so don't review yet!

benjeffery:
@jeromekelleher Ok! Ready for review.

benjeffery:
@jeromekelleher I've made the changes we discussed - in doing so it revealed that `read_table_ragged_col_t` had two references to the offset array, one of which was direct and therefore non-NULLable by the caller of `read_tables`. I couldn't see why this was needed so have removed the second reference, it was added when we added support for 64bit.

jeromekelleher (approved):
Looks great, let's merge!

PR #2146, opened 2022-03-07
Flesh out metadata doc schema example

hyanwong:
As discussed with @benjeffery

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -70,43 +70,216 @@ API, see {ref}`sec_tutorial_metadata_binary`.
 
 In this section we give some examples of how to define metadata
 schemas and how to add metadata to various parts of a tree sequence
-using the Python API.
+using the Python API. For simplicity, these initial examples use the JSON codec
+(see {ref}`sec_metadata_codecs`).
 
 (sec_metadata_examples_top_level)=
 
 ### Top level
 
-```{eval-rst}
-.. todo:: Add examples of top-level metadata. One with the ``permissive_json``
-  schema first to to show the simplest possible way of doing it. Then
-  followed with an example where we describe the metadata also.
+Top level metadata is associated with the tree sequence as a whole, rather than
+any specific table. This is used, for example, by programs such as
+[SLiM](https://github.com/MesserLab/SLiM) to store information about the sort of
+model that was used to generate the tree sequence (but note that detailed information
+used to recreate the tree sequence is better stored in {ref}`sec_provenance`).
+
+Here's an example of adding your own top-level metadata to a tree sequence:
+
+```{code-cell}
+import tskit
+# Define some top-level metadata you might want to add to a tree sequence
+top_level_metadata = {
+    "taxonomy": {"species": "Arabidopsis lyrata", "subspecies": "petraea"},
+    "generation_time": 2,
+}
+
+# Generate a simple tree sequence of one random tree.
+ts = tskit.Tree.generate_random_binary(8, branch_length=10, random_seed=9).tree_sequence
+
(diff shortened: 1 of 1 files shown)


docs/metadata.md
```diff
@@ -392,22 +400,40 @@ schema = metadata.MetadataSchema(
                 "binaryFormat": "10p",
                 "pattern": "^([1-9][0-9]{3})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])?$",
             },
+            "location": {
+                "description": "latitude and longitude of the collected sample",
```
jeromekelleher, line 404: Not sure this is a great example - we want people to use Individual.location for things like this, right?

jeromekelleher (commented):
LGTM, but I'd choose a different example in case people think this is the recommended route for location info.

docs/metadata.md
hyanwong: That's a good point. Have you got a suggestion for a parameter that uses one (or a set of) floating point values that we could use?

jeromekelleher: Some kind of phenotype maybe. Height and bmi or something?

hyanwong: Perhaps height and sex (so that we can use an enum)?

jeromekelleher: Whatever - but lets not choose something divisive as an example.

hyanwong: Good point. I forget this is an arguable point.

benjeffery:
Thanks @hyanwong, looks good modulo JK's comments.
One more detail i should comment on - `additionalProperties` is optional here as it will be added to the schema for you if you use struct. The same thing is happening with `required` which is being set on the `location` object in this case.

docs/metadata.md
```diff
@@ -101,6 +101,15 @@ using the Python API.
 .. todo:: Add examples of adding table-level metadata schemas.
 ```
 
+### Defaults
+
+```{eval-rst}
+.. todo:: Add examples of setting default values.
+```
+
+Default values for properties can be specified for only
```
benjeffery, line 110: This is only the case for JSON codec.

docs/metadata.md
```diff
@@ -392,22 +400,41 @@ schema = metadata.MetadataSchema(
...
+                        "type": "number",
+                        "binaryFormat": "f",
+                    },
+                    "age": {
+                        "description": "Age in years at time of sampling, or -1 if unknown",
+                        "type": "number",
+                        "binaryFormat": "h",
+                    },
+                },
+                "default": {"height": float("NaN"), "age": -1},
```
benjeffery, line 418: I think these should be done as separate defaults for age and height, then with an empty object default for phenotype. This allows either both, or one to be missing.

docs/metadata.md
hyanwong: Oh, I thought you said it was also required by the struct codec. I misunderstood. But I bunged that in a separate commit, so I can remove that. Sorry for the misunderstanding.

docs/metadata.md
hyanwong: Yes, I can see that, but you said previously that it wasn't obvious that you could set a top-level dict of defaults like this, so I thought it was worth showing an example.

hyanwong:
This is now ready for merging, IMO, @benjeffery

benjeffery:
@mergifyio rebase

benjeffery:
Yeesh 32bit tests are failing hard. Will investigate.

benjeffery:
@mergifyio rebase

benjeffery (approved):

PR #2134, opened 2022-02-21
Fix collision of VERSION files with Clang++ <version> header files

gbinux:
## Description

Many of Apple clang++'s standard headers include a internal header file named version (without an extension name). With the `VERSION` files in the `c` folder, Apple clang++ mistakenly uses the VERSION file as its own header file `<version>`. This causes problems when compiling the cpp test codes on macOS.

It seems this is can be fixed by simply adding the `.txt` extension name to the `tskit/c/VERSION` and `subprojects/kastore/VERSION` files and update the 2nd line of`meson.build` to `version: files('VERSION.txt')`.

Fixes #2133 

# PR Checklist:

- [ ] Tests that fully cover new/changed functionality.
- [ ] Documentation including tutorial content if appropriate.
- [ ] Changelogs, if there are API changes.

PR #2149, opened 2022-03-08
Rename version file

### LINKED ISSUE #2133: VERSION files in tskit/kastore c folders lead to CPP compilation failure on macOS
When I used meson/ninjia/x86_64-apple-darwin13.4.0-clang++(in a Conda environment) to compile the sources files under `tskit/c`, I got many line of errors. 

One type of them are due to the failure to find system include paths. So I added the paths to meson.build for cpp files.  And it worked.
```
cpp_sys_dep = declare_dependency(include_directories: [
    '/Users/xxxx/mambaforge/envs/cxx/bin/../include/c++/v1',
    '/Users/xxx/mambaforge/envs/cxx/lib/clang/12.0.1/include',
    '/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/include',
    '/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks'
])

executable('cpp_sorting_example',
          sources: ['examples/cpp_sorting_example.cpp'], 
          link_with: [tskit_lib], dependencies: [cpp_sys_dep,lib_deps])
```

The second type of errors are like this:
```
../version:1:1: error: expected unqualified-id
0.99.15
```

If I change file content to `extern int xxxx;`, it will compile.

Many of Apple clang++'s standard headers include a internal header file named `version` (without an extension name).   I think the Apple clang++ mistakenly use the tskit/c VERSION file as its own header file (`version`).

It seems that this is can be fixed by simply adding the `.txt` extension name to the `tskit/c/VERSION` and `subprojects/kastore/VERSION` files and update the 2nd line of `meson.build` to `  version: files('VERSION.txt')`.

Have you seen this before? Am I missing something from meson part (like a meson way of setting the `-isystem`)?

benjeffery:
Fixes #2133


jeromekelleher:
It would be nice to update #2134 so that @gbinux gets git cred for this.

Should just be a case of rebasing once #2148 is in?

benjeffery:
Whoops, forgot we had a PR for this.

PR #2150, opened 2022-03-11
Refactor vargen to tree_get_variant

benjeffery:
Some work in the direction of #605.

Note that as it stands the variant and tree need to be init'd consistently with regard to sample lists. This seemed to be better than having `get_variant` check everytime as I expect the use case is mostly getting variants for a set of changing sites, not samples.


benjeffery:
@jeromekelleher Are we bringing this up to python before C1.0? Would be nice to do it all at once.

c/tskit/genotypes.c
```diff
@@ -474,20 +470,31 @@ tsk_variant_get_allele_index(tsk_variant_t *self, const char *allele, tsk_size_t
     return ret;
 }
 
-static int
-tsk_variant_update_site(tsk_variant_t *self)
```
jeromekelleher, line 478: Might be better to keep this as tsk_variant_update_site() and then just keep ``tsk_tree_get_variant`` as simple as possible (just a call to this method)?

Much of the functionality here really does feel like a "method" of the variant object, so it feels artificial/wrong putting it into the tree class.

jeromekelleher (commented):
See comment about keeping most code in the variant class.

Maybe this will help clarify the status of the flags.

I'm leaning towards making the vargen an undocumented/deprecated feature that we encourage existing users to move away from, and explicitly keep out of the 1.0 API.

jeromekelleher:
> Are we bringing this up to python before C1.0? 

I would remove the usage of the vargen_t from Python-C anyway. Probably as easy to add the Tree.get_variant() method and be done with it.

benjeffery:
Added some tests, will do the suggested refactor.

benjeffery:
Dammit, valgrind wants to play.

benjeffery:
Failure seems unrelated. Investigating!

benjeffery:
Seems some screwed up caching on msys2, works in a fresh PR.

jeromekelleher (approved):
LGTM. I'm not sure about the exact flags semantics here, but let's merge and iterate.

PR #2152, opened 2022-03-14
Remove vargen from genotype_matrix

benjeffery:
Part of #605 stacked on #2150

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -9280,19 +9280,22 @@ TreeSequence_get_genotype_matrix(TreeSequence *self, PyObject *args, PyObject *k
 {
     PyObject *ret = NULL;
     static char *kwlist[] = { "isolated_as_missing", "alleles", NULL };
-    int err;
+    int err, t_ret;
     tsk_size_t num_sites;
     tsk_size_t num_samples;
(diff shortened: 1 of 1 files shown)


python/_tskitmodule.c
```diff
@@ -9324,31 +9324,53 @@ TreeSequence_get_genotype_matrix(TreeSequence *self, PyObject *args, PyObject *k
         goto out;
     }
     V = (int32_t *) PyArray_DATA(genotype_matrix);
-    vg = PyMem_Malloc(sizeof(tsk_vargen_t));
-    if (vg == NULL) {
+    variant = PyMem_Malloc(sizeof(tsk_variant_t));
```
jeromekelleher, line 9327: Is there a reason for mallocing the tree and variant rather than using stack storage? It saves a little bit of messing around, and once you make sure to memset them to zero before doing anything it's safe.

jeromekelleher (approved):
LGTM - minor comment re memory allocs

python/_tskitmodule.c
benjeffery: I've moved it to the stack, couldn't see a reason not. The vargen was on the heap before this diff though.

PR #2154, opened 2022-03-14
Remove vargen from variant iterator

benjeffery:
WIP


jeromekelleher (commented):
I was thinking of just nuking the VariantGenerator object entirely. If we have the Tree.get_variant method then it's redundant isn't it? I can't see there being any real perf difference between calling this from Python in ``TreeSequence.variants`` and using this CPython infrastructure.

So, rather than changing this next, how about adding the Tree.get_variant method?

PR #2156, opened 2022-03-16
clarify genetic_relatedness

petrelharp:
It's not entirely clear that the `span_normalise` applies both to the numerator and denominator when `proportion=True`; this clarifies this.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -6673,7 +6673,7 @@ class TreeSequence:
         :param str mode: A string giving the "type" of the statistic to be computed
             (defaults to "site").
         :param bool span_normalise: Whether to divide the result by the span of the
-            window (defaults to True).
+            window (defaults to True). Has no effect if ``proportion`` is True.
         :param bool proportion: Whether to divide the result by
             :meth:`.segregating_sites`, called with the same ``windows`` and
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -6673,10 +6673,10 @@ def genetic_relatedness(
...
             (defaults to "site").
         :param bool span_normalise: Whether to divide the result by the span of the
-            window (defaults to True).
-        :param bool proportion: Whether to divide the result by
-            :meth:`.segregating_sites`, called with the same ``windows`` and
-            ``mode`` (defaults to True). Note that this counts sites
+            window (defaults to True). Has no effect if ``proportion`` is True.
+        :param bool proportion: Defaults to True.  Whether to divide the result by
+            :meth:`.segregating_sites`, called with the same ``windows``,
+            ``mode``, and ``span_normalise``. Note that this counts sites
```
jeromekelleher, line 6679: I'm confused with this: doesn't that contradict the assertion above "Has no effect if propoption is True"?

jeromekelleher: Oh, I see.

jeromekelleher (approved):
Happy to merge whenever you are

PR #2164, opened 2022-03-23
Fix doc f3->f2

benjeffery:
Should this f3 be an f2? I'm not completely sure as sometimes these are collectively known as "f3 statistics" but our f4 has f4 in this spot, so for consistency, this seems right.


petrelharp:
Lol, that definitely should be `f2`, not `f3`. Nice catch.

Issue #2139, opened 2022-03-02
Things to document in treeseq_t class for 1.0


jeromekelleher:
Document:

- ``tables`` member (noting it's read only and you're very much on your own if you mutate it)
- Basic accessors like ``get_num_nodes`` etc
- load, dump, init etc methods
- Methods for accessors get_breakpoints, get_samples etc
- row object getters, get_node() etc
- simplify

We probably *could* document the various stats methods, but we may want to change something like the exact order of parameters when we generalise things a bit (e.g., for general LD stats) so I think it's best to leave them for 1.0

jeromekelleher:
Closing this as under control in #2208

PR #2206, opened 2022-04-25
Document treeseq struct members

benjeffery:
Part of #2139

I've been very liberal here documenting most of the struct, but happy to remove any we don't want to commit to. I do see most of these as being useful to users of the class though and unlikely to change.

### DIFF
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -88,7 +88,9 @@ typedef struct {
     tsk_mutation_t *site_mutations_mem;
     tsk_mutation_t **site_mutations;
     tsk_size_t *site_mutations_length;
-    /* The underlying tables */
+    /** @brief  The table collection underlying this tree sequence, This table
+     *  collection must be treated as read-only, and any changes to it will
+     *  lead to undefined behaviour. */
     tsk_table_collection_t *tables;
 } tsk_treeseq_t;
 


c/tskit/trees.h
```diff
@@ -63,32 +63,42 @@ extern "C" {
...
-    /* Map individuals to the list of nodes that reference them */
     tsk_id_t *individual_nodes_mem;
+    /** @brief  A ragged array of node ids corresponding to each individual */
     tsk_id_t **individual_nodes;
+    /** @brief  The lengths of the arrays in individual_nodes */
     tsk_size_t *individual_nodes_length;
-    /* For each tree, a list of sites on that tree */
     tsk_site_t *tree_sites_mem;
+    /** @brief  A ragged array of site ids for each tree */
     tsk_site_t **tree_sites;
```
jeromekelleher, line 91: I don't think we should document these things like ``inidividual_nodes``, ``site_mutations`` etc because this ties us down to *always* keeping these arrays. We could imagine doing things in future like only mallocing memory for these if someone calls the method in question.

jeromekelleher (commented):
I'd have tended to go the other way and document nothing except ``tables``, since there's getter methods for them. However, we could get rid of the getters for things like ``num_trees``, which is probably simpler for users.

As a guiding principle, I'd say lets not document any struct member that requires extra memory to be malloced.

c/tskit/trees.h
```diff
@@ -88,7 +88,9 @@ typedef struct {
     tsk_mutation_t *site_mutations_mem;
     tsk_mutation_t **site_mutations;
     tsk_size_t *site_mutations_length;
-    /* The underlying tables */
+    /** @brief  The table collection underlying this tree sequence, note that this
```
jeromekelleher, line 91: Maybe strengthen that assertion up a bit:
``suggestion
    /** @brief  The table collection underlying this tree sequence. This table collection must be treated as read-only, and any changes to it will lead to undefined behaviour.
```

jeromekelleher (approved):
LGTM

PR #2209, opened 2022-04-25
Draft implementation of Variant.concordance()

### LINKED ISSUE #2199: Calculate concordance to assess imputation performance
We want to implement functions in the Variant class to calculate various metrics to evaluate genotype imputation performance (discussed #2193).

The simplest metric is the **_overall concordance_**, which is the proportion of genotypes imputed correctly. Some studies look at the **_non-reference concordance_**, which considers only the genotypes that contain a non-reference allele. I am thinking that we can implement a function that allows us to compute the concordance between the same sites in two tree sequences. ts_true contains the known genotypes (e.g., from msprime), and ts_imputed contains the imputed genotypes.

```
site_concordance = []
for variant1, variant2 in zip(ts_true.variants(), ts_imputed.variants()):
     concordance = variant1.concordance(variant2, non_reference_only = False).
     site_concordance.append(concordance)
```

If the option non_reference_only is set to True, then it is returning the non-reference concordance instead of the overall concordance.

szhan:
## Description

This adds a function to get the proportion of genotypes in this variant and another variant. It checks that the number of genotypes in both the variants is equal. Importantly, the result is only valid if the genotypes for both the variants are from the same samples and in the same order.

Fixes #2199

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -574,6 +574,35 @@ class Variant:
         """
         return len(self.alleles) - self.has_missing_data
 
+    def concordance(self, other):
+        """
+        Calculate the fraction of samples for which the genotypes
+        of this variant are equal to the genotypes of the other variant.
+        The result is valid only if the genotypes for both the variants
+        are in the same order with respect to their samples.
+        Also, genotypes can be missing. Two genotypes with the value
+        ``tskit.MISSING_DATA`` are treated as equal.
+        """
+        if other.genotypes.shape != self.genotypes.shape:
+            raise ValueError("Variants not comparable.")
+        n = self.genotypes.shape[0]
+        if self.alleles == other.alleles:
+            # Quick path.
+            return np.sum(self.genotypes == other.genotypes, dtype=float) / n
+        else:
+            # Slow path - remap the alleles so the genotypes are comparable.
+            allele_idx_self = (
+                self.alleles if self.has_missing_data else self.alleles[:-1] + [-1]
+            )
+            allele_idx_other = (
+                other.alleles if other.has_missing_data else other.alleles[:-1] + [-1]
+            )
+            genotypes_remap_self = [allele_idx_self[i] for i in self.genotypes]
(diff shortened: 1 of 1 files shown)


szhan:
Right now, it is allowing for missing data in the genotypes.

python/tskit/trees.py
```diff
@@ -574,6 +574,35 @@ def num_alleles(self) -> int:
         """
         return len(self.alleles) - self.has_missing_data
 
+    def concordance(self, other):
+        """
+        Calculate the fraction of samples for which the genotypes
+        of this variant are equal to the genotypes of the other variant.
+        The result is valid only if the genotypes for both the variants
+        are in the same order with respect to their samples.
+        Also, genotypes can be missing. Two genotypes with the value
```
jeromekelleher, line 583: We're not really dealing with missing data here I think? We're saying missing data is equal to itself, but not equal to any alleles.

I think that's all we can do, but we should document it a bit more clearly.

jeromekelleher (commented):
LGTM! 

Needs some tests. I would add a class TestConcordance to TestGenotypes, where you can test out various examples.

@benjeffery, what do you think the right way to generate test examples is here? It would be simplest if we just created some Variant objects by hand so we can compare them directly, rather than going roundabout through a tree sequence. Maybe we create a local builder function like:

```python
def make_variant(genotypes, alleles=("0", "1")):
      return tskit.Variant(genotypes=genotypes, alleles=alleles)
```

Will this be OK with the low-level coupling that we're using/planning on using?

benjeffery:
> @benjeffery, what do you think the right way to generate test examples is here?

I don't think your example will work as the `Variant` class is very tied to the low-level C-Python. I think the way to do this is to mock out `genotypes` and `alleles`:

```python
from unittest.mock import MagicMock

def test_something():     
    variant = Variant.__new__(Variant)
    variant.genotypes = MagicMock(return_value=[0,1,2,3,....])
    variant.alleles = MagicMock(return_value=['A','C',....])
    # do something

PR #2216, opened 2022-04-27
Implementation of allele_remap()

### LINKED ISSUE #2211: Create a function to map from one alleles list to another
I'm writing several functions to calculate metrics to assess imputation performance (see #2199 #2200 #2210). As pointed out in #2199 , when the alleles lists of two variants are not equal, the genotypes of the variants must be remapped so that the genotypes are comparable. Because this remapping needs to be done a lot to support #2199, #2200, and #2210, it may be good to have a separate function to do this remapping. @benjeffery and I chatted about adding such a function.

szhan:
Fixes #2211 

This is meant to be an internal function.

### DIFF
--- a/python/tests/test_util.py
+++ b/python/tests/test_util.py
@@ -540,3 +540,41 @@ class TestRandomNuceotides:
     def test_length_bad_type(self, length):
         with pytest.raises(TypeError, match="argument must be a string"):
             tskit.random_nucleotides(length)
+
+
+#
+# Tests for allele_remap
+#
+@pytest.mark.parametrize(
+    "alleles_from, alleles_to, allele_map",
+    [
+        # Case 1: alleles_to is longer than alleles_from.
+        (["A", "C", "G", "T"], ["G", "C"], np.array([2, 1, 0, 3], dtype="uint32")),
+        # Case 2: alleles_to is shorter than alleles_from.
+        (["G", "C"], ["A", "C", "G", "T"], np.array([2, 1], dtype="uint32")),
+        # Case 3: alleles_to is empty.
+        (["A", "C", "G", "T"], [], np.array([0, 1, 2, 3], dtype="uint32")),
+        # Case 4: alleles_from is empty.
+        ([], ["A", "C", "G", "T"], np.array([], dtype="uint32")),
+        # Case 5: Both lists are empty.
+        ([], [], np.array([], dtype="uint32")),
+        # Case 6: Both lists are tuples.
+        (("G", "C"), ("A", "C", "G", "T"), np.array([2, 1], dtype="uint32")),
+        # Case 7: Both lists are numpy arrays.
+        (
+            np.array(("G", "C")),
+            np.array(("A", "C", "G", "T")),
+            np.array([2, 1], dtype="uint32"),
+        ),
+        # Case 8: Lists are of two different types.
+        (np.array(("G", "C")), ["A", "C", "G", "T"], np.array([2, 1], dtype="uint32")),
+        # Case 9: Lists contain elements of arbitrary types.
+        (
+            ["ABBA", "CDCD"],
+            ["ABBA", "CDCD", "EFEF", "GG", 18],
+            np.array([0, 1], dtype="uint32"),
+        ),
+    ],
+)
+def test_allele_remap(alleles_from, alleles_to, allele_map):
+    assert np.array_equal(allele_map, tskit.allele_remap(alleles_from, alleles_to))
--- a/python/tests/test_util.py
+++ b/python/tests/test_util.py
@@ -549,17 +549,41 @@ class TestRandomNuceotides:
     "alleles_from, alleles_to, allele_map",
     [
         # Case 1: alleles_to is longer than alleles_from.
-        (["A", "C", "G", "T"], ["G", "C"], np.array([2, 1, 0, 3], dtype="uint32")),
+        (
+            ["A", "C", "G", "T"],
+            ["G", "C"],
+            np.array([2, 1, 0, 3], dtype="uint32"),
+        ),
         # Case 2: alleles_to is shorter than alleles_from.
-        (["G", "C"], ["A", "C", "G", "T"], np.array([2, 1], dtype="uint32")),
+        (
+            ["G", "C"],
+            ["A", "C", "G", "T"],
+            np.array([2, 1], dtype="uint32"),
+        ),
         # Case 3: alleles_to is empty.
-        (["A", "C", "G", "T"], [], np.array([0, 1, 2, 3], dtype="uint32")),
+        (
+            ["A", "C", "G", "T"],
+            [],
+            np.array([0, 1, 2, 3], dtype="uint32"),
+        ),
         # Case 4: alleles_from is empty.
-        ([], ["A", "C", "G", "T"], np.array([], dtype="uint32")),
+        (
+            [],
+            ["A", "C", "G", "T"],
+            np.array([], dtype="uint32"),
+        ),
         # Case 5: Both lists are empty.
-        ([], [], np.array([], dtype="uint32")),
+        (
+            [],
+            [],
+            np.array([], dtype="uint32"),
+        ),
         # Case 6: Both lists are tuples.
-        (("G", "C"), ("A", "C", "G", "T"), np.array([2, 1], dtype="uint32")),
+        (
+            ("G", "C"),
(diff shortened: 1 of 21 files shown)


szhan:
When committing, I got a message from ``flake8`` about having a bare ``except``. So, to work around this, I added ``Exception``. I was thinking to use ``IndexError`` instead, but that causes the function to terminate with an error. Is there a  better way to deal with a bare ``except`` here?

benjeffery:
The exception thrown is `ValueError` so you should catch that:
```python
>>> l = [1,2]
>>> l.index(2)
1
>>> l.index(3)
ValueError: 3 is not in list
```

szhan:
I'm trying to think of weird edge cases to test. But this function is quite simple, so I'm not sure if we need more tests. Should we add tests to check that ``alleles_to`` and ``alleles_from`` are Python ``list``?

python/tests/test_util.py
```diff
@@ -540,3 +540,27 @@ def test_length_bad_value(self, length):
...
         with pytest.raises(TypeError, match="argument must be a string"):
             tskit.random_nucleotides(length)
+
+
+class TestAlleleRemap:
+    a = ["A", "C", "G", "T"]
+    b = ["G", "C"]
+    c = []
+
+    def test_a_bigger_than_b(self):
```
benjeffery, line 550: Maybe rename these to `test_from_bigger_than_to` and `test_to_bigger_than_from` as currently `test_a_smaller_than_b` looks like an error as it isn't.

python/tests/test_util.py
```diff
@@ -540,3 +540,27 @@ def test_length_bad_value(self, length):
     def test_length_bad_type(self, length):
         with pytest.raises(TypeError, match="argument must be a string"):
             tskit.random_nucleotides(length)
+
+
+class TestAlleleRemap:
+    a = ["A", "C", "G", "T"]
+    b = ["G", "C"]
+    c = []
```
benjeffery, line 548: It would be good to add some more tests here you could use `@pytest.mark.parametrize` as in `test_reference_sequence:179`. Some ideas:
 - What happens if one/both the arguments are a single string?
 - Error states such as an argument being `None`, these tests help us understand what exceptions are thrown by a method and that they don't silently do something unwanted.

python/tskit/util.py
```diff
@@ -592,3 +592,33 @@ def random_nucleotides(length: numbers.Number, *, seed: Union[int, None] = None)
...
+
+def allele_remap(alleles_from, alleles_to):
+    """
+    Returns an index map from the elements in one list (alleles_from)
+    to the elements of another list (alleles_to).
+
+    If some elements in alleles_from are not in alleles_to,
+    then indices outside of alleles_to are used.
+
+    :param list alleles_from: A list of alleles to map from.
```
benjeffery, line 610: The current type requirement for this function isn't `list` but `Sequence` as things like strings, tuples and bytes support the `.index` method.
I wonder if we should be supporting numpy arrays as a possible input to this function? I know `Variant.alleles` returns a `tuple`, so maybe we don't need to.

python/tskit/util.py
```diff
@@ -592,3 +592,33 @@ def random_nucleotides(length: numbers.Number, *, seed: Union[int, None] = None)
...
     return a.tobytes().decode("ascii")
+
+
+#
+# Miscellaneous auxiliary methods.
+#
+
+
+def allele_remap(alleles_from, alleles_to):
+    """
```
benjeffery, line 603: I'm not 100% sure we want this method to be documented. Documenting a method declares it as something we think is useful externally and is a commitment to best-effort support indefinitely so it always has to be a considered action.

benjeffery (commented):
Looking good! Some questions and comments.

jeromekelleher:
Just a quick comment here on usability for this: we shouldn't be too strict about the input types, as we'd also like to be able to use this for comparing ``tsinfer.Variant`` objects with ``tskit.Variant``. I'm currently doing this and hitting the tedious allele shuffling problem.

python/tskit/util.py
szhan: Is it good to keep two sentences about what it does? Or is too obvious enough?

```
"""
Returns an index map from the elements in one list (alleles_from) to the elements of another list (alleles_to).

If some elements in alleles_from are not in alleles_to, then indices outside of alleles_to are used.
"""
```

benjeffery: Yes, fine to keep that, but format it with `#` as comments so it doesn't get picked up by sphinx.

python/tests/test_util.py
szhan: By "one/both the arguments are a single string", do you mean if at least one of the inputs is not a list, just a string?

szhan: Thanks, I get what you mean now!

python/tests/test_util.py
```diff
@@ -540,3 +540,29 @@ def test_length_bad_value(self, length):
     def test_length_bad_type(self, length):
         with pytest.raises(TypeError, match="argument must be a string"):
             tskit.random_nucleotides(length)
+
+
+#
+# Tests for allele_remap
+#
+@pytest.mark.parametrize(
```
benjeffery, line 548: Rather than have the same code repeated in three tests, combine all the `parametrize` lists, adding comments like "differs in length" and "both empty" before the line of those ones.

benjeffery (approved):
Sorry was waiting for @szhan to say this was ready.

One small niggle then we can merge.

python/tests/test_util.py
szhan: Maybe we can check that both the input lists are of the same type and are one of list, tuple or string.

benjeffery: I don't think we need to be overly restrictive - they don't need to be the same type. Also I almost forgot - can we add some numpy examples to the tests?

szhan: The function is using index(), which numpy arrays don't have. Maybe convert a numpy array into a Python list?

szhan: Ah! You mentioned below!

benjeffery: > The function is using `index()`, which numpy arrays don't have. Maybe convert a numpy array into a Python list?

Probably better to do it the other way around and convert to numpy array at the start, then we support all sequences that numpy does.

szhan: So, the inputs can be a Numpy array or any Python Sequence-type object. And the output is a Numpy array (or a Python list)?

szhan: Also, we are allowing for duplicate elements in either or both of the inputs?

szhan: I've updated the function. Now, the inputs can be a numpy array or any Python Sequence-type object, and the output is a numpy array.

szhan: When the input is a string, then it is treated as a list with one element, which is the string itself.

python/tests/test_util.py
```diff
@@ -548,15 +548,19 @@ def test_length_bad_type(self, length):
...
+        # Case 2: alleles_to is shorter than alleles_from.
+        (["G", "C"], ["A", "C", "G", "T"], np.array([2, 1], dtype="uint32")),
+        # Case 3: alleles_to is empty.
+        (["A", "C", "G", "T"], [], np.array([0, 1, 2, 3], dtype="uint32")),
+        # Case 4: alleles_from is empty.
+        ([], ["A", "C", "G", "T"], np.array([], dtype="uint32")),
+        # Case 5: Both lists are empty.
+        ([], [], np.array([], dtype="uint32")),
+        # Case 6: Both lists are tuples.
+        (("G", "C"), ("A", "C", "G", "T"), np.array([2, 1], dtype="uint32")),
```
jeromekelleher, line 562: What about multi-character alleles? We need to be able to handle arbitrary strings as alleles. Like ``["AAA", "AA", "A", ""]``, e.g.

szhan: Added a test for that.

benjeffery: Oh and one more - as we support arbitrary unicode in these fields it would be good to have a test with the classic poo emoji as one of the alleles.

benjeffery:
@szhan Is this ready for another look? I can also clean up the commits if you like? Seems to have a merge in at the moment.

szhan:
Cancelling and closing this PR. Merging these commits into my own local fork caused too many weird conflicts, so it is far easier to just redo it.

PR #2224, opened 2022-04-28
Rename int and load flags to common namespace

### LINKED ISSUE #2219: `TSK_NO_METADATA` and `TSK_NO_EDGE_METADATA` have the same value
We might as well undo this collision in case of future usage of `TSK_NO_METADATA` by `tsk_table_collection_init`.

benjeffery:
Fixes #2219

### DIFF
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -697,28 +697,21 @@ typedef struct {
 /* Leave room for more positive check flags */
 #define TSK_NO_CHECK_POPULATION_REFS (1 << 12)
 
-/* Flags for load tables */
-#define TSK_BUILD_INDEXES (1 << 0)
+/* These flags are for table collection load or init, or used as
+   flags on table collection or individual tables.
+ * As flags are passed though from load to init they share a namespace */
+#define TSK_LOAD_SKIP_TABLES (1 << 0)
+#define TSK_LOAD_SKIP_REFERENCE_SEQUENCE (1 << 1)
+#define TSK_TABLE_NO_METADATA (1 << 2)
+#define TSK_TC_NO_EDGE_METADATA (1 << 3)
 
 /* Flags for dump tables */
 /* We may not want to document this flag, but it's useful for testing
  * so we put it high up in the bit space, below the common options */
 #define TSK_DUMP_FORCE_OFFSET_64 (1 << 27)
 
-/* Flags for table collection init/copy */
-/* TODO: need to careful about what flags get passed to init, from other
- * functions. Review as part of #1720 */
-#define TSK_NO_EDGE_METADATA (1 << 0)
-#define TSK_COPY_FILE_UUID (1 << 1)
-
-/* Flags for table collection load */
-/* This shares an interface with table collection init.
-   TODO: review as part of #1720 */
-#define TSK_LOAD_SKIP_TABLES (1 << 1)
-#define TSK_LOAD_SKIP_REFERENCE_SEQUENCE (1 << 2)
-
-/* Flags for table init. */
-#define TSK_NO_METADATA (1 << 0)
+/* Flags for table collection copy */
+#define TSK_COPY_FILE_UUID (1 << 0)
 
 /* Flags for union() */
 #define TSK_UNION_NO_CHECK_SHARED (1 << 0)
@@ -1439,7 +1432,7 @@ are initialised and freed.
 Options can be specified by providing one or more of the following bitwise
 flags:
 
-TSK_NO_METADATA
+TSK_TABLE_NO_METADATA
     Do not allocate space to store metadata in this table. Operations
     attempting to add non-empty metadata to the table will fail
     with error TSK_ERR_METADATA_DISABLED.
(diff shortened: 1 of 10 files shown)


jeromekelleher (approved):
LGTM.

I think we probably need to think through the flags semantics for table collection init/table init a little more clearly, but this is a good step.

benjeffery:
Ha, I've actually created a collision here as all `load` flags are passed to `init`. I guess the only way to tidy this up nicely is to have a `TSK_LOAD_INIT_*` namespace, but it is a bit of a mouthful.

benjeffery:
@jeromekelleher After playing around I've added to the PR to try and sort out the load/init flags, giving them a shared name space. It actually revealed an error where the tests were only working due to a collision.

c/tskit/tables.h
```diff
@@ -697,28 +697,21 @@ typedef struct {
...
 
-/* Flags for load tables */
-#define TSK_BUILD_INDEXES (1 << 0)
+/* Flags for table collection load or init
+ * As flags are passed though from load to init they share a namespace */
+#define TSK_LOAD_INIT_NO_METADATA (1 << 0)
+#define TSK_LOAD_INIT_SKIP_TABLES (1 << 1)
+#define TSK_LOAD_INIT_SKIP_REFERENCE_SEQUENCE (1 << 2)
+#define TSK_LOAD_INIT_NO_EDGE_METADATA (1 << 3)
+#define TSK_LOAD_INIT_BUILD_INDEXES (1 << 4)
```
jeromekelleher, line 706: This one is different, in that it's only ever passed to tsk_treeseq_init(), right?

c/tskit/tables.h
```diff
@@ -697,28 +697,21 @@ typedef struct {
 /* Leave room for more positive check flags */
 #define TSK_NO_CHECK_POPULATION_REFS (1 << 12)
 
-/* Flags for load tables */
-#define TSK_BUILD_INDEXES (1 << 0)
+/* Flags for table collection load or init
+ * As flags are passed though from load to init they share a namespace */
+#define TSK_LOAD_INIT_NO_METADATA (1 << 0)
```
jeromekelleher, line 702: LOAD_INIT seems a bit weird. Aren't these properties of the table collection/tables, which we're specifying at init time? So we could say something like

```
// This table has no metadata
TSK_TABLE_NO_METADATA (1 << 0)

// This table collection will have no reference sequence/tables/edge metadata
TSK_TC_NO_REFERENCE_SEQUENCE(1 << 0)
TSK_TC_NO_TABLES(1 << 1)
TSK_TC_NO_EDGE_METADATA (1 << 2)
```

c/tskit/tables.h
jeromekelleher: Could call ``TSK_TS_INIT_BUILD_INDEXES``

jeromekelleher (commented):
I'm not sure this @benjeffery - what do you think of my suggestion of thinking of the flags to init/load as declaring *properties* of the thing (table,table collection) and namespace that way?

benjeffery:
> I'm not sure this @benjeffery - what do you think of my suggestion of thinking of the flags to init/load as declaring _properties_ of the thing (table,table collection) and namespace that way?

Yeah, will give that a try. Has to be better than this!

benjeffery:
@jeromekelleher I think this makes more sense, have the shared namespace as a concept, but not all the same name.

c/tskit/tables.h
```diff
@@ -697,28 +697,22 @@ typedef struct {
...
-/* Flags for load tables */
-#define TSK_BUILD_INDEXES (1 << 0)
+/* These flags are for table collection load or init, or used as
+   flags on table collection or individual tables.
+ * As flags are passed though from load to init they share a namespace */
+#define TSK_LOAD_SKIP_TABLES (1 << 0)
+#define TSK_TABLE_NO_METADATA (1 << 1)
+#define TSK_TC_SKIP_REFERENCE_SEQUENCE (1 << 2)
+#define TSK_TC_NO_EDGE_METADATA (1 << 3)
+#define TSK_TS_INIT_BUILD_INDEXES (1 << 4)
```
jeromekelleher, line 707: Doesn't TSK_TS_INIT_BUILD_INDEXES belong in trees.h though, and live in a separate namespace? I don't think it interacts with the TableCollection of Table inits at all.

Ah, but SKIP_TABLES and SKIP_REFSEQ can also be passed to ``tsk_treeseq_load``, right? So namespacing these as LOAD makes sense, maybe?

benjeffery: Yes, this is the issue - all flags that get passed to `tsk_treeseq_load` get passed to `tsk_table_collection_load` and then all passed on again to `tsk_table_collection_init` or `tsk_table_collection_loadf_inited`. So they all need to be in this block.

You're right though `TSK_TS_INIT_BUILD_INDEXES` is unrelated - got confused with tsk_tc_init. Will move that.

benjeffery:
I also need to add a detailed change log here.

c/tskit/tables.h
benjeffery: Will change the two "skips" to "loads" also as that makes more sense to me.

jeromekelleher:
OK great, so we have some flags that we specifiy for ``load``, and additionaly flags that can get passed through to ``init``. I think that's making sense now and we just have to be clear on the documentation for this.

PR #2233, opened 2022-05-03
Add error for taking table with no edge metadata

### LINKED ISSUE #2207: Check behaviour is tested for the combination of `TSK_NO_EDGE_METADATA` and `TSK_TAKE_OWNERSHIP`
We have a note about this at `trees.c:417` but I don't think a test was added when `TSK_TAKE_TABLES` was added.

benjeffery:
Fixes #2207 
Added an error for this case.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -407,6 +407,10 @@ tsk_treeseq_init(
     tsk_memset(self, 0, sizeof(*self));
     if (options & TSK_TAKE_OWNERSHIP) {
         self->tables = tables;
+        if (tables->edges.options & TSK_TABLE_NO_METADATA) {
+            ret = TSK_CANT_TAKE_OWNERSHIP_NO_EDGE_METADATA;
+            goto out;
+        }
     } else {
         self->tables = tsk_malloc(sizeof(*self->tables));
         if (self->tables == NULL) {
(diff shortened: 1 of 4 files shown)


c/tskit/trees.c
```diff
@@ -407,6 +407,10 @@ tsk_treeseq_init(
     tsk_memset(self, 0, sizeof(*self));
     if (options & TSK_TAKE_OWNERSHIP) {
         self->tables = tables;
+        if (tables->edges.options & TSK_NO_EDGE_METADATA) {
```
jeromekelleher, line 410: Shouldn't the flag be TSK_TABLE_NO_METADATA though? I thought that was the point of that flag - any table that has no metadata would use it. the TSK_TC_NO_EDGE_METADATA was a way to send that flag to edge_table_init.

Maybe we should wait until #2224 is in before tackling this.

jeromekelleher (commented):
Good catch, but I think the implementation is uncovering another bug?

c/tskit/trees.c
benjeffery: Ah, they have the same value as I haven't rebased, so if I had rebased my mistake would have been revealed. 

I've rebased and fixed this.

benjeffery:
@jeromekelleher I think this is ready to now.

jeromekelleher (approved):

PR #2235, opened 2022-05-03
Addition of position argument to TreeSequence.site()

### LINKED ISSUE #2234: Add position=None keyword only arg to TreeSequence.site() method
The ``TreeSequence.site()`` method currently takes a single argument, ``id``. It would be very handy for exploratory analysis to have a ``position=x`` argument.

We can use use np.binsearch on ``tables.sites.position`` to find it, and raise a KeyError if it doesn't exist.

@szhan Fancy picking this on up?

szhan:
Fixes #2234

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -25,6 +25,7 @@ Test cases for the high level interface to tskit.
 """
 import collections
 import dataclasses
+import decimal
 import inspect
 import io
 import itertools
@@ -2023,6 +2024,8 @@ class TestTreeSequence(HighLevelTestCase):
             for index, site in enumerate(ts.sites()):
                 s2 = ts.site(site.id)
                 assert s2 == site
+                s3 = ts.site(position=site.position)
+                assert s3 == site
                 assert site.position == sites.position[index]
                 assert site.position > previous_pos
                 previous_pos = site.position
@@ -4323,3 +4326,97 @@ class TestTskitConversionOutput(unittest.TestCase):
             ValueError, match="macs output only supports single letter alleles"
         ):
             ts.to_macs()
+
+
+class TestTreeSequenceGetSite:
+    """
+    Tests for getting Site objects from a TreeSequence object
+    by specifying the position.
+    """
+
+    def get_example_ts_discrete_coordinates(self):
+        tables = tskit.TableCollection(sequence_length=10)
+        tables.sites.add_row(position=3, ancestral_state="A")
+        tables.sites.add_row(position=5, ancestral_state="C")
+        tables.sites.add_row(position=7, ancestral_state="G")
+        return tables.tree_sequence()
+
+    def get_example_ts_continuous_coordinates(self):
+        tables = tskit.TableCollection(sequence_length=10)
+        tables.sites.add_row(position=0.5, ancestral_state="A")
+        tables.sites.add_row(position=6.2, ancestral_state="C")
+        tables.sites.add_row(position=8.3, ancestral_state="T")
+        return tables.tree_sequence()
+
+    def get_example_ts_without_sites(self):
+        tables = tskit.TableCollection(sequence_length=10)
+        return tables.tree_sequence()
+
+    @pytest.mark.parametrize("id_", [0, 1, 2])
+    def test_site_id(self, id_):
+        ts = self.get_example_ts_discrete_coordinates()
+        site = ts.site(id_)
+        assert site.id == id_
+
+    @pytest.mark.parametrize("position", [3, 5, 7])
+    def test_position_discrete_coordinates(self, position):
+        ts = self.get_example_ts_discrete_coordinates()
+        site = ts.site(position=position)
+        assert site.position == position
+
+    @pytest.mark.parametrize("position", [0.5, 6.2, 8.3])
+    def test_position_continuous_coordinates(self, position):
+        ts = self.get_example_ts_continuous_coordinates()
+        site = ts.site(position=position)
+        assert site.position == position
+
+    @pytest.mark.parametrize("position", [0, 2.999999999, 5.000000001, 9])
+    def test_position_not_found(self, position):
+        with pytest.raises(ValueError, match=r"There is no site at position"):
+            ts = self.get_example_ts_discrete_coordinates()
+            ts.site(position=position)
+
+    @pytest.mark.parametrize(
+        "position",
+        [
+            np.array([3], dtype=float)[0],
+            np.array([3], dtype=int)[0],
+            decimal.Decimal(3),
+        ],
+    )
+    def test_position_good_type(self, position):
+        ts = self.get_example_ts_discrete_coordinates()
+        ts.site(position=position)
+
+    def test_position_not_scalar(self):
+        with pytest.raises(
+            ValueError, match="Position must be provided as a scalar value."
+        ):
+            ts = self.get_example_ts_discrete_coordinates()
+            ts.site(position=[1, 4, 8])
+
+    @pytest.mark.parametrize("position", [-1, 10, 11])
+    def test_position_out_of_bounds(self, position):
+        with pytest.raises(
+            ValueError,
+            match="Position is beyond the coordinates defined by sequence length.",
+        ):
+            ts = self.get_example_ts_discrete_coordinates()
+            ts.site(position=position)
+
+    def test_query_position_siteless_ts(self):
+        with pytest.raises(ValueError, match=r"There is no site at position"):
+            ts = self.get_example_ts_without_sites()
+            ts.site(position=1)
+
+    def test_site_id_and_position_are_none(self):
+        with pytest.raises(TypeError, match="Site id or position must be provided."):
+            ts = self.get_example_ts_discrete_coordinates()
+            ts.site(None, position=None)
+
+    def test_site_id_and_position_are_specified(self):
+        with pytest.raises(
+            TypeError, match="Only one of site id or position needs to be provided."
+        ):
+            ts = self.get_example_ts_discrete_coordinates()
+            ts.site(0, position=3)
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5215,8 +5215,8 @@ class TreeSequence:
         with either the specified ID or position.
 
         When position is specified instead of site ID, a binary search is done
-        on the list of positions of the sites (pre-sorted in ascending order)
-        to try to find a site with the user-specified position.
+        on the list of positions of the sites to try to find a site
+        with the user-specified position.
 
         :rtype: :class:`Site`
         """
@@ -5232,7 +5232,7 @@ class TreeSequence:
                 raise ValueError(
                     "Position is beyond the coordinates defined by sequence length."
                 )
-            site_pos = self.tables.sites.position  # Sorted in ascending order
+            site_pos = self.tables.sites.position
             id_ = site_pos.searchsorted(position)
             if id_ >= len(site_pos) or site_pos[id_] != position:
                 raise ValueError(f"There is no site at position {position}.")
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5209,13 +5209,33 @@ class TreeSequence:
             metadata_decoder=self.table_metadata_schemas.mutation.decode_row,
         )
 
-    def site(self, id_):
+    def site(self, id_=None, *, position=None):
         """
         Returns the :ref:`site <sec_site_table_definition>` in this tree sequence
-        with the specified ID.
+        with either the specified ID or position.
+
+        When position is specified instead of site ID, a binary search is done
+        on the list of positions of the sites (pre-sorted in ascending order)
+        to try to find a site with the user-specified position.
 
         :rtype: :class:`Site`
         """
+        if id_ is None and position is None:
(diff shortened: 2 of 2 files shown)


benjeffery:
@szhan Tag me here when you want a 👀

jeromekelleher:
@szhan - it would be nice to get this one in for the next release. It mainly just needs some tests - do you want some pointers on how to do this?

szhan:
I'm trying to find where the tests are for ``TreeSequence.site()``. I'm not seeing them under ``class TestTreeSequence`` in ``test_highlevel.py``. Are they elsewhere?

jeromekelleher:
No, there's probably no specific testing for this method. I'd suggest adding a ``TestTreeSequenceGetSite`` method to tests/test_highlevel.py where you can add some simple tests.

benjeffery:
> I'm trying to find where the tests are for `TreeSequence.site()`. I'm not seeing them under `class TestTreeSequence` in `test_highlevel.py`. Are they elsewhere?

`test_highlevel:2006`

szhan:
I see a test for ``TreeSequence.sites()`` in which ``TreeSequence.site()`` get used, but not tests written specifically for ``TreeSequence.site()``.

python/tests/test_highlevel.py
```diff
@@ -4261,3 +4261,24 @@ def test_macs_error(self):
             ValueError, match="macs output only supports single letter alleles"
         ):
             ts.to_macs()
+
+
+class TestTreeSequenceGetSite:
```
benjeffery, line 4266: The tests at test_highlevel.py:2006 are specifically for the site method. These tests are a good start, but I think adding 
```python
s3 = ts.site(position=site.postion)
assert s3 = site
```
would be an good thing to add too.

python/tests/test_highlevel.py
```diff
@@ -4261,3 +4261,24 @@ def test_macs_error(self):
...
+    """
+    Tests for getting Site objects from a TreeSequence object
+    by specifying site id or position.
```
benjeffery, line 4269: These only do position though, so leave it at that?

python/tests/test_highlevel.py
```diff
@@ -4261,3 +4261,24 @@ def test_macs_error(self):
...
+
+    def get_example(self):
+        tables = tskit.TableCollection(sequence_length=10)
+        tables.sites.add_row(position=3, ancestral_state="A")
+        tables.sites.add_row(position=5, ancestral_state="C")
+        tables.sites.add_row(position=7, ancestral_state="G")
+        tables.sites.add_row(position=8.3, ancestral_state="T")
+        return tables.tree_sequence()
+
+    @pytest.mark.parametrize("position", [3.0, 5.0, 7.0, 8.3])
```
benjeffery, line 4280: Would be good add tests for the `0` and `10` edge cases here.

python/tskit/trees.py
```diff
@@ -5098,13 +5098,27 @@ def mutation(self, id_):
             metadata_decoder=self.table_metadata_schemas.mutation.decode_row,
         )
 
-    def site(self, id_):
+    def site(self, id_=None, position=None):
         """
         Returns the :ref:`site <sec_site_table_definition>` in this tree sequence
-        with the specified ID.
+        with either the specified ID or position.
```
benjeffery, line 5104: An explanation of the scaling characteristics would be good here as the caller should be informed that this method when called with `position=` does a binary search of the position array. We don't want people to iterate through positions to get sites!

python/tskit/trees.py
```diff
@@ -5098,13 +5098,27 @@ def mutation(self, id_):
...
 
         :rtype: :class:`Site`
         """
+        if id_ is None and position is None:
+            raise TypeError("Site id or position must be provided.")
```
benjeffery, line 5109: Missing coverage in all the error states.

python/tskit/trees.py
```diff
@@ -5098,13 +5098,27 @@ def mutation(self, id_):
...
+            if position < 0 or position >= self.sequence_length:
+                raise ValueError(
+                    "Position is beyond the coordinates defined by sequence length."
+                )
+            position = np.array(position)
+            if len(position.shape) > 0:
+                raise ValueError("Position must be provided as a scalar value.")
+            site_pos = self.tables.sites.position  # Sorted in ascending order
+            id_ = site_pos.searchsorted(position)
+            if site_pos[id_] != position:
```
benjeffery, line 5120: Don't you end up getting a `KeyError` here if the position is greater than the last site? `searchsorted` will return one-past-the-end.

benjeffery (commented):
Thanks @szhan!
Needs tests for the error states, some more detail on the docs and a fix for a possible KeyError.

python/tests/test_highlevel.py
```diff
@@ -4261,3 +4261,23 @@ def test_macs_error(self):
...
+        tables.sites.add_row(position=3, ancestral_state="A")
+        tables.sites.add_row(position=5, ancestral_state="C")
+        tables.sites.add_row(position=7, ancestral_state="G")
+        return tables.tree_sequence()
+
+    @pytest.mark.parametrize("position", [3.0, 5.0, 7.0])
+    def test_site_position(self, position):
+        ts = self.example()
+        site = ts.site(position=position)
+        assert site.position == position
```
jeromekelleher, line 4284: This is a good start. So we also need:
```
def test_site_id_and_position(self):
    pass

def test_no_args(self):
    pass

@pytest.mark.parametrize(...)
def test_position_bad_type(self, bad_type):
    pass

@pytest.mark.parametrize("position", [-1, 10, 11]):
def test_position_out_of_bounds(self, position):
    pass

@pytest.mark.parametrize("position", [0, 2.999999, 3.000001, 4.9999, 5.00001, 9.99999]):
def test_bad_position(self, position):
    pass

File: python/tests/test_highlevel.py [hunk 8]
```diff
@@ -4261,3 +4261,23 @@ def test_macs_error(self):
...
+    """
+
+    def get_example():
+        tables = tskit.TableCollection(sequence_length=10)
+        tables.sites.add_row(position=3, ancestral_state="A")
+        tables.sites.add_row(position=5, ancestral_state="C")
+        tables.sites.add_row(position=7, ancestral_state="G")
+        return tables.tree_sequence()
+
+    @pytest.mark.parametrize("position", [3.0, 5.0, 7.0])
```
jeromekelleher: Also want to test different number types here, so something like
```
[3.0, 5.0, 7.0, 3, np.array([3], dtype=float)[0], np.array([3], dtype=int)[0], decimal.Decimal(3)]
```

jeromekelleher (commented):
Good start, just needs more tests for all the tricky corner cases. There's probably more I haven't thought of here, like what happens when you pass a list of values, etc.

python/tskit/trees.py
```diff
@@ -5098,13 +5098,27 @@ def mutation(self, id_):
             metadata_decoder=self.table_metadata_schemas.mutation.decode_row,
         )
 
-    def site(self, id_):
+    def site(self, id_=None, position=None):
```
jeromekelleher, line 5101: Position should be keyword only, so need 
```
    def site(self, id_=None, *, position=None):
```

python/tskit/trees.py
szhan: I have added this assertion in ``site`` before ``searchsorted``. It only works for discrete coordinates though.

```
if position < 0 or position >= self.sequence_length:
    raise ValueError("Position is beyond the coordinates defined by sequence length.")
```

benjeffery: I don't think that works even for discrete coords - say for example if you had zero sites!

python/tests/test_highlevel.py
szhan: It should only work for discrete positions, not continuous?

szhan:
I keep running into this error when trying to ``git commit``:

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

python/tests/test_highlevel.py:44:1: F401 'curses.ACS_NEQUAL' imported but unused
python/tests/test_highlevel.py:52:1: F401 'python.tests.simplify.AncestorMap' imported but unused
```

So, I have been using ``SKIP=flake8`` for now.

python/tskit/trees.py
szhan: I have added a test to check that a KeyError is returned in such an instance.

python/tests/test_highlevel.py
szhan: I have added a check that position must be an integer or floating point. Is that too narrow?

python/tskit/trees.py
benjeffery: I don't think `KeyError` is the right exception to pass back to the user here. Can you add a test for getting a site from a table with no sites also to exercise this part and check we get the right exception? (`ValueError`)

szhan: I think I have corrected it now.

python/tests/test_highlevel.py
szhan: Ahhh, why there are so many types of numbers?

python/tests/test_highlevel.py
szhan: Could I do this to check whether ``position`` holds an integer?

```
a = 9.4
int(a) == a
a = 3.0
int(a) == a
```

python/tests/test_highlevel.py
benjeffery: Checking types like this is usually an anti-pattern in python 

> I keep running into this error when trying to `git commit`:
> 
> ```
> flake8...................................................................Failed
> - hook id: flake8
> - exit code: 1
> 
> python/tests/test_highlevel.py:44:1: F401 'curses.ACS_NEQUAL' imported but unused
> python/tests/test_highlevel.py:52:1: F401 'python.tests.simplify.AncestorMap' imported but unused
> ```
> 
> So, I have been using `SKIP=flake8` for now.

@szhan Found that these lines had been added 
to his patch without his knowledge!

benjeffery:
> I keep running into this error when trying to `git commit`:
> 
> ```
> flake8...................................................................Failed
> - hook id: flake8
> - exit code: 1
> 
> python/tests/test_highlevel.py:44:1: F401 'curses.ACS_NEQUAL' imported but unused
> python/tests/test_highlevel.py:52:1: F401 'python.tests.simplify.AncestorMap' imported but unused
> ```
> 
> So, I have been using `SKIP=flake8` for now.

Checking types like this is usually an anti-pattern in python 

> I keep running into this error when trying to `git commit`:
> 
> ```
> flake8...................................................................Failed
> - hook id: flake8
> - exit code: 1
> 
> python/tests/test_highlevel.py:44:1: F401 'curses.ACS_NEQUAL' imported but unused
> python/tests/test_highlevel.py:52:1: F401 'python.tests.simplify.AncestorMap' imported but unused
> ```
> 
> So, I have been using `SKIP=flake8` for now.

@szhan Found that these lines had been added to his patch without his knowledge!

python/tests/test_highlevel.py
szhan: NVM. I think I misunderstood ``test_bad_position()``.

python/tests/test_highlevel.py
```diff
@@ -4261,3 +4264,79 @@ def test_macs_error(self):
...
+
+    @pytest.mark.parametrize("position", [0.5, 6.2, 8.3])
+    def test_position_continuous_coordinates(self, position):
+        ts = self.get_example_ts_continuous_coordinates()
+        site = ts.site(position=position)
+        assert site.position == position
+
+    @pytest.mark.parametrize("position", [0, 2.999999999, 5.000000001, 9])
+    def test_position_not_found(self, position):
+        with pytest.raises(ValueError):
```
benjeffery, line 4307: For all these it is good to add a `match=` to check the error is the right one.

benjeffery (commented):
Getting there! One good idea to add `match=` to all the exception testers.

benjeffery:
Not sure what is going on with coverage here as those states must be getting hit. Locally I hit them if I set a breakpoint, but my coverage shows them as uncovered too.

szhan:
Jerome figured it out! I just had to rebase. It seems that codecov didn't like that my branch was based on an older version of tskit.

python/tskit/trees.py
```diff
@@ -5209,13 +5209,33 @@ def mutation(self, id_):
...
+            if len(position.shape) > 0:
+                raise ValueError("Position must be provided as a scalar value.")
+            if position < 0 or position >= self.sequence_length:
+                raise ValueError(
+                    "Position is beyond the coordinates defined by sequence length."
+                )
+            site_pos = self.tables.sites.position  # Sorted in ascending order
+            id_ = site_pos.searchsorted(position)
+            if id_ >= len(site_pos) or site_pos[id_] != position:
+                raise ValueError(f"There is no site at position {position}.")
```
jeromekelleher, line 5238: I wonder if this should be a KeyError @benjeffery?

jeromekelleher (approved):
LGTM!

python/tskit/trees.py
benjeffery: Yes, probably should be, after thinking about it.

jeromekelleher: Actually no, I take it back. list().index raises a ValueError when the value isn't found, and I guess this is equivalent.

PR #2242, opened 2022-05-05
Document tree traversals in C

jeromekelleher:
I think this the last piece of the puzzle for documenting the 1.0 tree API. I ended up adding two new functions, but I think traversing all nodes is a common enough use-case that we should give it its own function and not make people remember that -1 is the magic value you need to use.

### DIFF
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -1289,8 +1289,9 @@ int tsk_tree_get_sites(
 This function provides an upper bound on the number of nodes that
 can be reached in tree traversals, and is intended to be used
 for memory allocation purposes. If ``num_nodes`` is the number
-of nodes visited in a tree traversal from the virtual root
-(e.g., ``tsk_tree_preorder(tree, tree->virtual_root, nodes,
+of nodes visited in a tree traversal from the
+:ref:`virtual root<sec_data_model_tree_roots>`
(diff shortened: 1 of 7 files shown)


c/tskit/trees.h
```diff
@@ -977,17 +976,141 @@ bool tsk_tree_is_descendant(const tsk_tree_t *self, tsk_id_t u, tsk_id_t v);
...
+*/
+int tsk_tree_preorder(const tsk_tree_t *self, tsk_id_t *nodes, tsk_size_t *num_nodes);
+
+/**
+@brief Fill an array with the nodes of this tree starting from a particular node.
+
+@rst
+As for :c:func:`tsk_tree_preorder` but starting the traversal at a particular node.
+The value ``-1`` is treated as a special case, in which we visit all nodes reachable
+from the roots. The virtual root is also a valid input for this function.
```
benjeffery, line 1029: Maybe link to `sec-data-model-tree-roots` wherever we mention the virtual root?

benjeffery (approved):
LGTM, one possible suggestion to link to the roots explanation when we mention the virtual root.

c/tskit/trees.h
jeromekelleher: Good call, done. Ready to go now I think

Issue #1804, opened 2021-10-17
Refactor tests to use example tree sequences and parametrize


jeromekelleher:
We currently have a mishmash of different ways of running tests on a range of different topologies. I think a good pattern will be to use pytest.mark.parametrize with a selection of tree sequences that we create once. The ``get_example_tree_sequences`` list in tests/test_highlevel.py is a good place to start for this (and currently used in existing PRs like #1704).

We don't have to do it all at once, but it would be good to refactor a few of the places we use different topologies and set the standard pattern for how we'd to see things done in the future. To close this PR, we should:

1. Extract the ``get_example_tree_sequences()`` code from tests/test_highlevel.py into it's own module (maybe ``examples.py``?) and set it up so that the list of examples gets built the first time its called, and then cached afterwards.
2. Refactor the test_highlevel.py file to use this function instead of the local examples and fix any breakages (nice to do a bit more "parametrizing" there as well where possible instead of iterating over the examples within the function.
3. Port the existing "fixture" methods to use the same pattern.

I think pytest fixtures are an anti-pattern because you're defining a global variable that the reader is supposed to understand the meaning of, and know where to find the definition. This approach is much more transparent.

So, code like this should be canonical in test files going forward:

```python

from tests import examples

...

class TestSomeStuff:

     @pytest.mark.parametrize("ts", examples.standard_ts())
     def test_specific_thing_where_topology_doesnt_matter(self, ts):
            ...

     @pytest.mark.parametrize("ts", examples.sample_topologies())
     def test_specific_thing_where_topology_matters_and_we_want_a_variety(self, ts):
            ...
```

PR #2246, opened 2022-05-09
Initial framework for balance metrics and tests

jeromekelleher:
This is the first step for adding tree balance metrics as discussed in #2245.

If we agree with this approach, we can follow up with some specific issues to (a) add other metrics; (b) document them; and (c) implement them in C.


python/tests/test_balance_metrics.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class TestDefinitions:
+    @pytest.mark.parametrize("ts", get_example_tree_sequences())
+    def test_sackin(self, ts):
+        for tree in ts.trees():
+            assert tree.sackin_index() == sackin_index_definition(tree)
+
+
+class TestBalancedBinaryOdd:
```
benjeffery, line 46: There's quite a bit of repeated boilerplate here for what is looping over several different topologies. If we're doing #1804 then should these trees be defined in `examples.py` then this file is a test that checks a dict of tree_name->value pairs as a single parameterised test?

jeromekelleher:
@jeremyguez would you mind taking a quick look here to see what you think? Are there any other tree topology examples we should include here? The idea is that we'll add tests for each of the balance metrics to the classes as we add them, so we can see the what the values are for these simple cases, easily.

benjeffery (approved):
Looks good, I wonder if we should go for defining these topologies in an examples.py though.

python/tests/test_balance_metrics.py
jeromekelleher: There's a tension here though between avoiding repetition and being able to *see* what you're testing. I find it really helpful to be able to just look at the example and see what it is your testing, and see what the result should be. It helps a lot to be able to fine-tune the examples for specific things that you're doing, too, rather than trying to come up with a universal set of examples that cover all possibilities.

jeromekelleher: Also, might not be obvious here but the idea is to put test functions for all the other metrics into these classes, so the boilder-plate fraction will go down quite a bit.

jeremyguez:
@jeromekelleher 
The tested topologies look good enough to me. (We could also have the case of mixed arities in one tree, but I'm not sure if it matters much.)

jeromekelleher:
OK, great, thanks @jeremyguez. I don't think there's much need to have mixed arity in a tree, as these examples are really just to make it easy for us to see what the answer is in some representative situations. The systematic testing is done in the pytest.mark.parametrize where we compare the definition with the implementation over lots of examples.

OK, going to merge this then. Are you happy to pick this up from here @jeremyguez? (I'll do the C version of this now, so you have examples to follow there)

jeremyguez:
@jeromekelleher Yes, I started to look at #2249.

PR #2258, opened 2022-05-10
Sackin C implementation

jeromekelleher:
Also includes the C testing framework for the rest of the balance metrics. So, to implement other balance metrics in C we add the implementation, and then test the returned value in these three test function.

cc @jeremyguez

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -6392,6 +6392,90 @@ test_genealogical_nearest_neighbours_errors(void)
 }
 
 static void
+test_single_tree_balance(void)
+{
+    int ret;
+    tsk_treeseq_t ts;
+    tsk_tree_t t;
+    tsk_size_t sackin;
+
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Sigh - codecov has gone a bit haywire again. I think everything that can be covered is covered.

jeromekelleher:
Bumping this one @benjeffery

benjeffery (approved):
LGTM - One question, is a 32bit `depth` only safe when `_TSK_BIG_TABLES` is not used? I'm not sure how paranoid to be here, but maybe the type of `depth` should be conditional on that macro?

jeromekelleher:
> LGTM - One question, is a 32bit depth only safe when _TSK_BIG_TABLES is not used? I'm not sure how paranoid to be here, but maybe the type of depth should be conditional on that macro?

Hah, good spot. I changed the depth value to a ``tsk_id_t`` as that is always safe.

Issue #2249, opened 2022-05-10
Tree path length


jeromekelleher:
Define a function that computes the number of edges traversed from one node to another in a tree.

I think the simplest way to define this is in terms of the depth and MRCA like this:

```python
def path_length(tree, u, v):
    mrca = tree.mrca(u, v)
    if mrca == -1:
        return math.inf
    return tree.depth(u) + tree.depth(v) - 2 * tree.depth(mrca)
```

@jeremyguez, do you agree with this definition?

If we agree with this definition we can implement in C based on the definition of [get_mrca](https://github.com/tskit-dev/tskit/blob/main/c/tskit/trees.c#L3535). The first step would be implement in Python and get some tests in place within the ``test_tree_balance.py`` file added in #2246

jeromekelleher:
For motivation, see #2245

jeremyguez:
@jeromekelleher 
You mean `test_balance_metrics.py`?

I was just wondering: in that case the definition used in `test_balance_metrics.py` and the actual method in `trees.py` will be the same, unlike for #2246 where those were different (`sackin_index_definition(tree)` vs `sackin_index(self)`)?

jeromekelleher:
It may not be worth implementing this in C btw. I doubt there'll be much difference between the implementation above and an optimised C version in most cases, and we probably won't end up using the function explicitly in the B2 metric (as we'd hope to keep the required state on the stack, rather than recompute lengths over and over).

So, I think we can close this once we have the version above in the Python API, documented and tested with some well chosen examples.

jeremyguez:
@jeromekelleher I gave a first try at it at #2259.

jeromekelleher:
Closed in #2259

PR #2259, opened 2022-05-10
Tree path length

jeremyguez:
I give here a try at #2249. I commented out the last test (copied from `TestMRCA`) as the trees I generated with `tskit.Tree.generate_balanced` did not have a virtual node (don't know why). I added some assert comparing with `self.t.depth() `outputs, but not sure if it's a good idea.
@jeromekelleher

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -885,6 +885,68 @@ class TestMRCA:
         assert self.t.mrca(0, 5) == 5
 
 
+class TestPathLength:
+    t = tskit.Tree.generate_balanced(9)
+    #         16
+    #    ┏━━━━┻━━━┓
+    #    ┃       15
+    #    ┃     ┏━━┻━┓
+    #   11     ┃   14
+    #  ┏━┻━┓   ┃  ┏━┻┓
+    #  9  10  12  ┃ 13
+    # ┏┻┓ ┏┻┓ ┏┻┓ ┃ ┏┻┓
+    # 0 1 2 3 4 5 6 7 8
+
+    def test_tmrca_leaf(self):
+        assert self.t.path_length(0, 16) == 3
+        assert self.t.path_length(16, 0) == 3
+        assert self.t.path_length(7, 16) == 4
+
+    def test_equal_depth(self):
+        assert self.t.path_length(5, 16) == self.t.depth(5)
+
+    def test_two_leaves(self):
+        assert self.t.path_length(0, 8) == 7
+
+    def test_two_leaves_depth(self):
+        assert self.t.path_length(0, 8) == self.t.depth(0) + self.t.depth(8)
+
+    @pytest.mark.parametrize("args", [[], [1], [1, 2, 3]])
+    def test_bad_num_args(self, args):
+        with pytest.raises(TypeError):
+            self.t.path_length(*args)
+
+    @pytest.mark.parametrize("bad_arg", [[], "1"])
+    def test_bad_arg_type(self, bad_arg):
+        with pytest.raises(TypeError):
+            self.t.path_length(0, bad_arg)
+        with pytest.raises(TypeError):
+            self.t.path_length(bad_arg, 0)
+
+    def test_same_args(self):
+        assert self.t.path_length(10, 10) == 0
+
+    def test_different_tree_levels(self):
+        assert self.t.path_length(1, 10) == 3
+
+    def test_out_of_bounds_args(self):
+        with pytest.raises(ValueError):
+            self.t.path_length(0, 20)
+
+    @pytest.mark.parametrize("u", range(17))
+    def test_virtual_root_arg(self, u):
+        assert self.t.path_length(u, self.t.virtual_root) == self.t.depth(u) + 1
+        assert self.t.path_length(self.t.virtual_root, u) == self.t.depth(u) + 1
+
+    def test_both_args_virtual_root(self):
+        assert self.t.path_length(self.t.virtual_root, self.t.virtual_root) == 0
+
+    def test_no_mrca(self):
+        tree = self.t.copy()
+        tree.clear()
+        assert math.isinf(tree.path_length(0, 1))
+
+
 class TestMRCACalculator:
     """
     Class to test the Schieber-Vishkin algorithm.
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2793,6 +2793,27 @@ class Tree:
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def path_length(self, u, v):
+        """
+        Returns the path length between two nodes
+        (i.e., the number of edges between two nodes in this tree).
+        If the two nodes have a most recent common ancestor, then this is defined as
+        ``tree.depth(u) + tree.depth(v) - 2 * tree.depth(tree.mrca(u, v))``. If the nodes
+        do not have an MRCA (i.e., they are in disconnected subtrees) the path length
+        is infinity.
+
+        .. seealso:: See also the :meth:`.depth` method
+
+        :param int u: The first node for path length computation.
+        :param int v: The second node for path length computation.
+        :return: The number of edges between the two nodes.
+        :rtype: int
+        """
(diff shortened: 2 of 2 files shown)


python/tskit/trees.py
```diff
@@ -2793,6 +2793,25 @@ def kc_distance(self, other, lambda_=0.0):
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def path_length(self, *args):
```
jeromekelleher, line 2796: The ``mrca`` function is a special case here because we can sensibly ask what's the MRCA of more than two nodes. However, I think we should fix this function to just supporting exactly two nodes, so let's change the signature to:
```python
def path_length(self, u, v):
```

python/tskit/trees.py
```diff
@@ -2793,6 +2793,25 @@ def kc_distance(self, other, lambda_=0.0):
...
+
+        """
+        Returns the path length between two nodes
+        (i.e., the number of edges between two nodes in this tree).
+
+        :param int `*args`: input node IDs, must be at least 2.
```
jeromekelleher, line 2802: Can update the params here then to have the standard form.

python/tskit/trees.py
```diff
@@ -2793,6 +2793,25 @@ def kc_distance(self, other, lambda_=0.0):
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def path_length(self, *args):
+
+        """
+        Returns the path length between two nodes
+        (i.e., the number of edges between two nodes in this tree).
```
jeromekelleher, line 2800:
```suggestion
        (i.e., the number of edges between two nodes in this tree).
        If the two nodes have a most recent common ancestor, then this is defined as 
        ``tree.depth(u) + tree.depth(v) - 2 * tree.depth(tree.mrca(u, v))``. If the nodes 
        do not have an MRCA (i.e., they are in disconnected subtrees) the path length 
        is infinity.
        
        .. seealso:: See also the :meth:`.depth` method
      
```

jeromekelleher: Here's a bit more detail for the docstring. You'll need to compile the docs to see if the cross linking works, I'm never sure when I've got RST syntax right!

python/tests/test_highlevel.py
```diff
@@ -885,6 +885,54 @@ def test_virtual_root_arg(self):
...
+        assert self.t.path_length(10, 10) == 0
+
+    def test_different_tree_levels(self):
+        assert self.t.path_length(1, 10) == 3
+
+    def test_out_of_bounds_args(self):
+        with pytest.raises(ValueError):
+            self.t.path_length(0, 20)
+
+    # def test_virtual_root_arg(self):
```
jeromekelleher, line 932: So, we need to think about the semantics of the virtual root here. I guess the simplest definition would be that 
```
tree.path_length(u, tree.virtual_root) == tree.depth(u) + 1
tree.path_length(tree.virtual_root, u) == tree.depth(u) + 1
tree.path_length(tree.virtual_root, tree.virtual_root) == 0
```
which, it turns out happens automatically :hooray:

Here's some tests for this and checking the no-mrca case:
```python
    @pytest.mark.parametrize("u", range(17))
    def test_virtual_root_arg(self, u):
        self.t.path_length(u, self.t.virtual_root) == self.t.depth(u)
        self.t.path_length(self.t.virtual_root, u) == self.t.depth(u)

    def test_both_args_virtual_root(self):
        self.t.path_length(self.t.virtual_root, self.t.virtual_root) == 0

    def test_no_mrca(self):
        tree = self.t.copy()
        tree.clear()
        assert math.isinf(tree.path_length(0, 1))

**Review by jeromekelleher — COMMENTED** [review]
Looks great! Some minor comments, and we're good to go then.

File: python/tskit/trees.py (hunk 1 above)
Comment (jeremyguez) [reply python/tskit/trees.py:2796]: I was testing args length in the method to pass the tests `test_less_than_two_args` and `test_no_args` (copied from the `mrca` method). Do you think I keep those tests?

File: python/tests/test_highlevel.py (hunk 4 above)
Comment (jeremyguez) [reply python/tests/test_highlevel.py]: Nice!

I was just wondering why I can't access `virtual_root` in this simple code:

```python
t = tskit.Tree.generate_balanced(9)
print(t.virtual_root)

> AttributeError: 'Tree' object has no attribute 'virtual_root'
```

python/tskit/trees.py
jeromekelleher: No, can just drop them

python/tests/test_highlevel.py
```diff
@@ -885,6 +885,64 @@ def test_virtual_root_arg(self):
...
+        assert self.t.path_length(0, 8) == 7
+
+    def test_two_leaves_depth(self):
+        assert self.t.path_length(0, 8) == self.t.depth(0) + self.t.depth(8)
+
+    # def test_less_than_two_args(self):
+    #     with pytest.raises(ValueError):
+    #         self.t.path_length(1)
+    #
+    # def test_no_args(self):
```
jeromekelleher, line 918: I guess we should test the interface a little all right, so how about something like
```python
@pytest.mark.parametrize("args", [[], [1], [1, 2, 3]])
def test_bad_num_args(self, args):
     with pytest.raises(TypeError):
            self.t.path_length(*args)

@pytest.mark.parametrize("bad_arg", [[], "1"])
def test_bad_arg_type(self, bad_arg)
      with pytest.raises(ValueError):
             self.t.path_length(0, bad_arg)
       with pytest.raises(ValueError):
             self.t.path_length(bad_arg, 0) 
```
I'm not sure if they second one will always raise a ValueError or if some things are TypeErrors. See what works, we don't need to worry too much about testing all possibilities here.

jeromekelleher (commented):
Looks great, thanks @jeremyguez! I've one more suggestion above for testing the interface, and we're done then.

Can you squash your commits down to one then please? (See [this guide](https://popsim-consortium.github.io/stdpopsim-docs/stable/development.html#rebasing) if you're not familiar with squashing)

python/tskit/trees.py
jeromekelleher: See suggestion below.

python/tests/test_highlevel.py
jeremyguez: Indeed it worked with `TypeError`.
I followed the tutorial on squashing, very useful! I forgot to edit the text of the commit message afterwards, I don't know if it's a problem.

Just one little thing, in the tutorial on squashing it's `git rebase -i upstream/master` where it should be `git rebase -i upstream/main`. Took me a bit of time to get the origin of the error git was raising when I used the first command.

python/tskit/trees.py
```diff
@@ -2793,6 +2793,29 @@ def kc_distance(self, other, lambda_=0.0):
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def path_length(self, u, v):
+
```
jeromekelleher, line 2797: Ah, just spotted there's an extra blank line here. Can you delete this please?

python/tskit/trees.py
```diff
@@ -2793,6 +2793,29 @@ def kc_distance(self, other, lambda_=0.0):
...
+        is infinity.
+
+        .. seealso:: See also the :meth:`.depth` method
+
+        :param int u: The first node for path length computation.
+        :param int v: The second node for path length computation.
+        :return: The number of edges between the two nodes.
+        :rtype: int
+        """
+
```
jeromekelleher, line 2813: Also no need for this blank line

jeromekelleher (approved):
OK, looking good! Just one last formatting nitpick.

Just fyi, rather than adding another commit and squashing, this is a case where "amending" a commit is handy. ``git commit -a --amend`` should update the commit, and you can then force push to update the PR.

python/tests/test_highlevel.py
jeromekelleher: Ah, good to know, thanks.

jeromekelleher:
Fantastic, thanks @jeremyguez !

jeremyguez:
Thanks for your help @jeromekelleher !

PR #2260, opened 2022-05-11
Doc err macros

benjeffery:
WIP!

### DIFF
--- a/c/tskit/core.h
+++ b/c/tskit/core.h
@@ -283,7 +283,6 @@ library to read. To read the file you must upgrade the version
 of tskit.
 */
 #define TSK_ERR_FILE_VERSION_TOO_NEW                                -102
-/** @} */
 
 /**
 A column that is a required member of a table was not found in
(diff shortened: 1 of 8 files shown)


benjeffery:
I've removed some flags here, which were unused. I guess this is technically a breaking change.

c/tskit/core.h
```diff
@@ -301,96 +300,355 @@ not found in the file.
...
+@defgroup OOB_ERROR_GROUP Out of bounds errors.
+@{
+*/
+/**
+A bad value was provided for a ragged column offset, values should
+start at zero and be monotonically increasing.
+*/
 #define TSK_ERR_BAD_OFFSET                                          -200
-#define TSK_ERR_OUT_OF_BOUNDS                                       -201
+/* 201 was the now unused TSK_OUT_OF_BOUNDS */
```
jeromekelleher, line 314: We might as well use 201, maybe move TSK_ERR_SEEK_OUT_OF_BOUNDS up here and change its code to 201? That's the least noisy change, and it's vanishingly unlikely anyone is depending on its specific value.

python/requirements/development.txt
```diff
@@ -25,7 +25,7 @@ pysam
 pytest
 pytest-cov
 pytest-xdist
-PyVCF
+#PyVCF
```
jeromekelleher, line 28: This is accidental I assume?

jeromekelleher (approved):
Wow, thanks so much for doing this @benjeffery :star_struck: 

This is great.

jeromekelleher:
Agreed

benjeffery:
@jeromekelleher This is ready to go, did another pass through and found a couple of tweaks (e.g. max alleles is not 127 anymore!)

docs/c-api.rst
```diff
@@ -682,9 +682,121 @@ File format errors
...
+        :content-only:
+
+----------------------
+Distance metric errors
+----------------------
+
+.. doxygengroup:: DISTANCE_ERROR_GROUP
+        :content-only:
+
+-------------------------
```
jeromekelleher, line 765: Maybe we should leave HAPLOTYPE_ERROR_GROUP,  DISTANCE_ERROR_GROUP, IBD_ERROR_GROUP, MAPPING_ERROR_GROUP, and STATS_ERROR_GROUP out for now? Seems odd to put these in to the end-user documentation when we're not providing the corresponding functions in the API. You've done the hard work, to adding the groups back in to the RST here will be very easy later.

jeromekelleher (approved):
LGTM! I wonder if we should leave some of the groups out from the final docs, but it's up to you.

docs/c-api.rst
benjeffery: Yeah, got a bit carried away. Will fix in the in-progress polish PR.

Issue #2190, opened 2022-04-06
Add selection test p-values


a-ignatieva:
It would be nice to have a function to calculate the selection test described in the Relate paper (p.16 of supplement) for mutations in a given tree. Or perhaps this is something that could be added to the documentation as an example, it's easy to code up using existing functions. This is the code I have (and an example). 

```python
import msprime
import math
import numpy as np

# Get number of lineages at time just above node
def get_num_lineages(tree, node):
    if tree.is_sample(node):
        count = tree.num_samples()
    else:
        nodes_ordered = tree.timedesc()
        count = 1 + np.where(nodes_ordered == node)[0]
    return count

# Calculate Relate selection test p-value
def selection_pvalue(tree, mutation):
    node = mutation.node
    if tree.is_sample(node):
        # Can only compute this for mutations with frequency >= 2
        p = None
    else:
        N = tree.num_samples()
        fn = tree.num_samples(node)
        # k is the number of lineages just after the edge carrying the mutation splits into 2
        k = 1 + get_num_lineages(tree, node)[0]
        p = 0.0
        for f in range(fn, N - k + 3):
            p += (f - 1) * math.comb(N - f - 1, k - 3) / math.comb(N - 1, k - 1)
    return p
```

```python
ts = msprime.sim_ancestry(
    10,
    population_size = 1,
    sequence_length = 1,
    ploidy = 1,
    recombination_rate = 0,
    discrete_genome = False,
    random_seed = 1)

ts = msprime.sim_mutations(ts, 
                           rate=3, 
                           discrete_genome=False, 
                           random_seed=1, 
                           model=msprime.BinaryMutationModel())
```
![pic1](https://user-images.githubusercontent.com/36134434/162077209-c8909b9d-0f14-4a56-8be3-d9dd866a4b9d.svg)

```python
for m in ts.mutations():
    print(m.id, m.site, ts.site(m.site).position, selection_pvalue(ts.first(), m))
```
```
0 0 0.027387595968320966 None
1 1 0.09280081023462117 0.41666666666666663
2 2 0.2044522489886731 0.5952380952380951
3 3 0.23608897626399994 None
4 4 0.4306985700968653 1.0
5 5 0.5181525482330471 0.41666666666666663
6 6 0.7783892354927957 1.0
7 7 0.8463109182193875 0.7222222222222222
8 8 0.8650202453136444 0.41666666666666663
9 9 0.9325573612004519 None
10 10 0.939127794932574 1.0
```

[...]

PR #2266, opened 2022-05-11
Adding colless_index method and tests

jeremyguez:
Here is a first try at Colless index. @jeromekelleher

### DIFF
--- a/python/tests/test_balance_metrics.py
+++ b/python/tests/test_balance_metrics.py
@@ -36,11 +36,44 @@ def sackin_index_definition(tree):
     return sum(tree.depth(u) for u in tree.leaves())
 
 
+def colless_index_definition(tree):
+    is_binary = all(
+        tree.num_children(u) == 2 for u in tree.nodes() if tree.is_internal(u)
+    )
+    if tree.num_roots != 1:
+        raise ValueError("Colless index not defined for multiroot trees")
+    if not is_binary:
+        raise ValueError("Colless index not defined for nonbinary trees")
+
+    return sum(
+        abs(
+            len(list(tree.leaves(tree.left_child(u))))
+            - len(list(tree.leaves(tree.right_child(u))))
+        )
+        for u in tree.nodes()
+        if tree.is_internal(u)
+    )
+
+
 class TestDefinitions:
     @pytest.mark.parametrize("ts", get_example_tree_sequences())
     def test_sackin(self, ts):
         for tree in ts.trees():
             assert tree.sackin_index() == sackin_index_definition(tree)
+
+    @pytest.mark.parametrize("ts", get_example_tree_sequences())
+    def test_colless(self, ts):
+        for tree in ts.trees():
+            is_binary = all(
+                tree.num_children(u) == 2 for u in tree.nodes() if tree.is_internal(u)
+            )
+            if tree.num_roots != 1 or not is_binary:
+                with pytest.raises(ValueError):
+                    tree.colless_index()
+                with pytest.raises(ValueError):
+                    colless_index_definition(tree)
+            else:
+                assert tree.colless_index() == colless_index_definition(tree)
 
 
 class TestBalancedBinaryOdd:
@@ -50,13 +83,15 @@ class TestBalancedBinaryOdd:
     #     ┊ ┃ ┏┻┓ ┊
     # 0.00┊ 0 1 2 ┊
     #     0      1
-
     @tests.cached_example
     def tree(self):
         return tskit.Tree.generate_balanced(3)
 
     def test_sackin(self):
         assert self.tree().sackin_index() == 5
+
+    def test_colless(self):
+        assert self.tree().colless_index() == 1
 
 
 class TestBalancedBinaryEven:
@@ -66,13 +101,15 @@ class TestBalancedBinaryEven:
     #     ┊ ┏┻┓ ┏┻┓ ┊
     # 0.00┊ 0 1 2 3 ┊
     #     0         1
-
     @tests.cached_example
     def tree(self):
         return tskit.Tree.generate_balanced(4)
 
     def test_sackin(self):
         assert self.tree().sackin_index() == 8
+
+    def test_colless(self):
+        assert self.tree().colless_index() == 0
 
 
 class TestBalancedTernary:
@@ -89,6 +126,10 @@ class TestBalancedTernary:
     def test_sackin(self):
         assert self.tree().sackin_index() == 18
 
+    def test_colless(self):
+        with pytest.raises(ValueError):
+            self.tree().colless_index()
+
 
 class TestStarN10:
     # 1.00┊         10          ┊
@@ -101,6 +142,10 @@ class TestStarN10:
 
     def test_sackin(self):
         assert self.tree().sackin_index() == 10
+
+    def test_colless(self):
+        with pytest.raises(ValueError):
+            self.tree().colless_index()
 
 
 class TestCombN5:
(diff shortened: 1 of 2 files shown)


python/tests/test_balance_metrics.py
```diff
@@ -121,8 +135,37 @@ def tree(self):
...
+class TestMultiRootBinary:
+    # 3.00┊            15     ┊
+    #     ┊          ┏━━┻━┓   ┊
+    # 2.00┊   11     ┃   14   ┊
+    #     ┊  ┏━┻━┓   ┃  ┏━┻┓  ┊
+    # 1.00┊  9  10  12  ┃ 13  ┊
+    #     ┊ ┏┻┓ ┏┻┓ ┏┻┓ ┃ ┏┻┓ ┊
+    # 0.00┊ 0 1 2 3 4 5 6 7 8 ┊
+    #     0                   1
+    @tests.cached_example
```
jeremyguez, line 151: I added this test because the only multi roots test was ternary, thus expecting `math.inf` for Colless index. I wanted to add a multi root test that outputs a real value. @jeromekelleher

jeromekelleher:
I found a good resource on balance metrics [here](https://treebalance.wordpress.com/colless-index/) (it's really thorough, and there's *lots* of metrics). According to them, this metric is only defined for binary trees with a single root, but I think the simplest thing to do is raise a ValueError if we encounter a tree that isn't one of these. How about this?

```python
def colless_index(self):
            if self.num_roots != 1:
            raise ValueError("Colless index not defined for multiroot trees")
        num_leaves = np.zeros(self.tree_sequence.num_nodes, dtype=np.int32)
        total = 0
        for u in self.nodes(order="postorder"):
            num_children = 0
            for v in self.children(u):
                num_leaves[u] += num_leaves[v]
                num_children += 1
            if num_children == 0:
                num_leaves[u] = 1
            elif num_children != 2:
                raise ValueError("Colless index not defined for nonbinary trees")
            else:
                total += abs(
                    num_leaves[self.right_child(u)] - num_leaves[self.left_child(u)]
                )
        return total
```
and for a definition we have 
```python
def colless_index_definition(tree):
    is_binary = all(
        tree.num_children(u) == 2 for u in tree.nodes() if tree.is_internal(u)
    )
    if tree.num_roots != 1 or not is_binary:
        raise ValueError("Colless index only defined for binary trees with one root")

    # We have to use len(list(tree.leaves(u))) here because tree.num_leaves() actually
    # returns the number of *samples*.
    return sum(
        abs(
            len(list(tree.leaves(tree.left_child(u))))
            - len(list(tree.leaves(tree.right_child(u))))
        )
        for u in tree.nodes()
        if tree.is_internal(u)
    )
```

It gets a bit ugly here for a few reasons:
1. We don't have a short/efficient way of saying "is this a binary tree" (it was discussed in passing [here](https://github.com/tskit-dev/tskit/issues/2190#issuecomment-1091610584). In principle we should be able to track this efficiently when building the tree, but it's work to do it.)
2. The ``num_leaves`` function doesn't do what it sounds like it should do (and should be removed from the API really)

I do think it's better to just raise an error here if there's any doubt about how the metric is defined when we have polytomies and multiple roots.

jeremyguez:
Thanks @jeromekelleher , very interesting resource!

I agree to raise an error when the tree is not binary following this resource, unlike [Shao and Sokal (1990)](https://github.com/tskit-dev/tskit/files/8687111/shao1990.1.pdf) who defined Colless index in that case by using only the binary nodes for the computation. Because in the extreme case where many nodes are polytomies, the index ends up being not representative of the imbalance at all with the Shao and Sokal definition, as only few of its nodes are used for the computation.

I have two questions for the moment on your code:
1. Why isn't it possible to test for `number of leaves == number of nodes - number of leaves + 1` to check if the tree is binary?
2. I have some trouble understanding why do we need a method and a definition, and why are they different? Is this to test that  both ways of computing give the same results (the most efficient and the simplest)?

jeromekelleher:
> Why isn't it possible to test for number of leaves == number of nodes - number of leaves + 1 to check if the tree is binary?

Yes, I guess that'll work. We still have to do a full traversal to get those numbers though, so it's not any more efficient.

> I have some trouble understanding why do we need a method and a definition, and why are they different? Is this to test that both ways of computing give the same results (the most efficient and the simplest)?

Basically, yes. Think of it as efficiency of human understanding vs efficiency of computer implementation. Humans and computers think differently, so it's very helpful to write things down both ways. In particular, deciding what the right behaviour in corner cases should be is really helped by having the simplest mathematical definition that you can (as, ideally, the corner case behaviour falls naturally out of that). This happened in the path_length computation in #2249 where I was wondering what the semantics should be for the virtual root - by having a nice mathematical definition for path_length, this was really easy.

benjeffery:
Let me know if/when you want a review here, otherwise I assume you both have it covered.

jeremyguez:
@jeromekelleher 
Thank you for your detailed answer, very useful!

I compared those two functions in term of time of execution (`colless_index` and `colless_index2`):

```python

# First method
def colless_index(self):
        if self.num_roots != 1:
            raise ValueError("Colless index not defined for multiroot trees")
        num_leaves = np.zeros(self.tree_sequence.num_nodes, dtype=np.int32)
        total = 0
        for u in self.nodes(order="postorder"):
            num_children = 0
            for v in self.children(u):
                num_leaves[u] += num_leaves[v]
                num_children += 1
            if num_children == 0:
                num_leaves[u] = 1
            elif num_children != 2:
                raise ValueError("Colless index not defined for nonbinary trees")
            else:
                total += abs(
                    num_leaves[self.right_child(u)] - num_leaves[self.left_child(u)]
                )
        return total

# Second method
def colless_index2(tree):
    if tree.num_roots != 1:
        raise ValueError("Colless index not defined for multiroot trees")

    total = 0

    for u in tree.nodes():
        if tree.is_internal(u):
            if tree.num_children(u) == 2 :
                total+= abs(tree.num_samples(tree.right_child(u))-tree.num_samples(tree.left_child(u)))
            else:
                raise ValueError("Colless index not defined for nonbinary trees")
    return(total)


# Big random tree
random = tskit.Tree.generate_random_binary(100000)

start = time.perf_counter()
print(colless_index(random))
stop = time.perf_counter()
print(f"colless_index = {stop-start:.2g}")

start = time.perf_counter()
print(colless_index2(random))
stop = time.perf_counter()
print(f"colless_index2 = {stop-start:.2g}")

```

The answer was:

```
74278552
colless_index = 0.43
74278552
colless_index2 = 0.37
```

I tried this code multiple times and the answer were about the same. So apparently both methods give the same results and `colless_index2` is a bit faster. The results were the same with a comb and a balanced tree.

jeromekelleher:
The relative performance here isn't so important @jeremyguez because both are written in Python, and the Python overhead will dominate for most trees.

The main issue though is that we can't use ``tree.num_samples`` here because samples aren't necessarily leaves, and leaves aren't necessarily samples (although the usually will be, for many of the trees that we use). We have to manually count the leaves because we don't have an efficient way of doing that. If you subsitute the ``num_leaves`` implementation above, the "definition" version will be much slower.

The speed of the "definition" version is pretty unimportant though, unless it's catestrophically slow it won't matter for the trees we're using the in tests.

The ``colless_index`` method defined above is preferable as the basis of a C implementation because it's nice and simple. It would also work well under numba acceleration. Try this version on your big tree (but make sure you run it more than once to warm up the jit):

```python
import numba

@numba.njit
def _colless_index(postorder, left_child, right_sib):
    num_leaves = np.zeros_like(left_child)
    total = 0
    for u in postorder:
        v = left_child[u]
        while v != -1:
            num_leaves[u] += num_leaves[v]
            v = right_sib[v]
        v = left_child[u]
        if v == -1:
            num_leaves[u] = 1
        else:
            # NB assuming tree is binary! We'd probably check this before
            # invoking the method.
            total += abs(num_leaves[right_sib[v]] - num_leaves[v])
    return total


def colless_index_numba(tree):
    if tree.num_roots != 1:
        raise ValueError("Colless index not defined for multiroot trees")
    # TODO check tree.is_binary, somehow
    return _colless_index(tree.postorder(), tree.left_child_array, tree.right_child_array)

```

jeremyguez:
@jeromekelleher 
So I pushed the methods you proposed. There is just the `TestDefinitions` class about which I don't know how to compare the definition with the method, as `assert tree.colless_index() == colless_index_definition(tree)` wouldn't work since it sometimes raise a ValueError.

python/tests/test_balance_metrics.py
jeromekelleher: We can probably get rid of the multiroot ternary example, it's not checking anything we're not doing elsewhere.

jeremyguez: Yes, just removed it.

jeromekelleher:
Thanks @jeremyguez - all we can do for the TestDefinion is check if the trees are binary first, and then either assert they get the same value or raise a ValueError.

jeremyguez:
> Thanks @jeremyguez - all we can do for the TestDefinion is check if the trees are binary first, and then either assert they get the same value or raise a ValueError.

I pushed a version doing this.
Thanks for your answers @jeromekelleher.

jeromekelleher (approved):

jeromekelleher:
Great stuff, thanks @jeremyguez!

ps. We have a Slack workspace tskit-dev and it would be great if you could join up. Can you send me an email please if you'd like to join? jerome.kelleher@bdi.ox.ac.uk

PR #2278, opened 2022-05-17
Fix doc build

benjeffery:


benjeffery:
Hmm, this update is orthogonal to the issue in #2275 but maybe the issue there was transitory.

Issue #2251, opened 2022-05-10
B1 Balance index (Python)


jeromekelleher:
Add a definition of the B1 balance index following the pattern established for Sackin index in #2246. Requires #2249 

See #2245 for information.

cc @jeremyguez

jeremyguez:
I wrote two definitions of B1 (first one is using list comprehension). Any thoughts @jeromekelleher? 

```python

def b1_index_definition(tree):
    return (sum
        (1/max([path_length(tree, n, leaf)
        for leaf in tree.leaves(n)])
        for n in tree.nodes()
        if n != tree.root and tree.is_internal(n))
    )

def b1_index_definition2(tree):
    total = 0
    for n in tree.nodes():
        if n != tree.root and tree.is_internal(n):
            depth_list = []
            for leaf in tree.leaves(n):
                depth_list.append(tree.path_length(n, leaf))
            total += 1/max(depth_list)
    return(total)

```

Do you have ideas for the method implementation? Maybe by going postorder we can have something more efficient?

jeromekelleher:
I really like the list comprehension version @jeremyguez, very neat.

You're right, there is a very clean and efficient implementation of this using a postorder traversal. Do you want to have a go at figuring it out? It took me a while to see it and it is really nice, so I don't want to spoil it for you (unless you want me to!) :smile:

jeremyguez:
Thanks!

Yes, I'll give it a try! Here's what I got:

```python
def b1_index(self):
    depth_max = np.zeros(self.tree_sequence.num_nodes, dtype=np.int32)
    total = 0
    for u in self.nodes(order="postorder"):
        if self.is_internal(u) and u != self.root:
            depth_max[u] = max([depth_max[v] for v in self.children(u)])+1
            total += 1/depth_max[u]
        elif self.is_leaf(u):
            depth_max[u] = 0    
    return total
```

jeromekelleher:
Very nice! That's basically what I got:

```python
def b1_index(tree):
    max_path_length = np.zeros(tree.tree_sequence.num_nodes, dtype=int)
    total = 0
    for u in tree.postorder():
        if tree.parent(u) != tskit.NULL and tree.is_internal(u):
            max_path_length[u] = 1 + max(max_path_length[v] for v in tree.children(u))
            total += 1 / max_path_length[u]
    return total
```

The only real difference is that I'm testing to see if ``u`` is *a* root rather than *the* root, so that this will work for multirooted trees.

[...]

PR #2281, opened 2022-05-19
Adding B1 balance index and tests

jeremyguez:
Here is the PR for B1 balance index. Compared to https://github.com/tskit-dev/tskit/issues/2251 I just added a `round` function at the end. @jeromekelleher

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2814,6 +2814,29 @@ class Tree:
             return math.inf
         return self.depth(u) + self.depth(v) - 2 * self.depth(mrca)
 
+    def b1_index(self):
+        """
+        Returns the B1 balance index for this tree.
+        This is defined as the inverse of the sum of all longest paths
+        to leaves for each node besides roots.
+
+        .. seealso:: See `Shao and Sokal (1990)
+            <https://www.jstor.org/stable/2992186>`_ for details.
+
+        :return: The B1 balance index.
+        :rtype: float
+        """
+        # TODO implement in C
+        max_path_length = np.zeros(self.tree_sequence.num_nodes, dtype=int)
+        total = 0.0
+        for u in self.postorder():
+            if self.parent(u) != tskit.NULL and self.is_internal(u):
+                max_path_length[u] = 1 + max(
+                    max_path_length[v] for v in self.children(u)
+                )
+                total += 1 / max_path_length[u]
+        return total
+
     def colless_index(self):
         """
         Returns the Colless imbalance index for this tree.
--- a/python/tests/test_balance_metrics.py
+++ b/python/tests/test_balance_metrics.py
@@ -55,6 +55,14 @@ def colless_index_definition(tree):
     )
 
 
+def b1_index_definition(tree):
+    return sum(
+        1 / max(tree.path_length(n, leaf) for leaf in tree.leaves(n))
+        for n in tree.nodes()
+        if tree.parent(n) != tskit.NULL and tree.is_internal(n)
+    )
+
+
 class TestDefinitions:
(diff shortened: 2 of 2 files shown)


jeromekelleher:
Looks great! I don't understand why we're rounding though?

jeremyguez:
> Looks great! I don't understand why we're rounding though?

It's because of the tests, having something like this didn't look clean (and I felt it could lead to errors):

```python
def test_b1(self):
        assert self.tree().b1_index() == 1.8333333333333333
```

And I wasn't sure this was clean either, since it's not testing the output of the method itself:

```python
def test_b1(self):
        assert round(self.tree().b1_index(),3) == 1.833
```

But I guess there is a clean general way of working around tests in that case?

jeremyguez:
By the way: I wrote in the doc that the output is a `float`, but actually in some cases, like for the empty tree, the output is 0 which is an `int`.  Should we force the result to be a `float`, so it's always the same output type?

jeromekelleher:
Ah, I see. I think the usual way to handle this would be something like
```
def test_b1(self):
        assert self.tree().b1_index() == pytest.approx(1.8333)
```

> By the way: I wrote in the doc that the output is a float, but actually in some cases, like for the empty tree, the output is 0 which is an int. Should we force the result to be a float, so it's always the same output type?

Good point. The cleanest way to fix this would be to set ``total = 0.0`` at the top of the loop, which will force it to be a float.

jeremyguez:
By removing the rounding I get now this kind of errors:

```python
assert tree.b1_index() == b1_index_definition(tree)
E           assert 19.4718253968254 == 19.471825396825395
```
I guess I should use `python.approx()` in all tests?

jeremyguez:
I put `pytest.approx(1.833, rel=1e-3)`, as the default is using `rel=1e-6`.

Is it an issue if the definition method output is not always a float, or should I force it to float too? The tests do pass without it.

jeromekelleher:
Do the tests not pass at the default tolerance? I would have expected the values to be very close.

> Is it an issue if the definition method output is not always a float, or should I force it to float too? The tests do pass without it.

that's fine, it doesn't really matter.

jeremyguez:
> Do the tests not pass at the default tolerance? I would have expected the values to be very close.

For this test the default relative tolerance is indeed working, as the difference is very small:

```python
assert tree.b1_index() == pytest.approx(b1_index_definition(tree))
```

But for this one, the default 1e-6 relative tolerance is too small compared to the difference between 1.833333.. and 1.833:

```python
assert self.tree().b1_index() == pytest.approx(1.833)
```

So I had two solutions that pass the test:

```python
assert self.tree().b1_index() == pytest.approx(1.833, rel=1e-3)
```
or

```python
assert self.tree().b1_index() == pytest.approx(1.8333333)
```
The first one seemed cleaner to me.

jeromekelleher:
Ah, I see. When you're comparing with a fixed value either way is good. Whatever you prefer.

python/tskit/trees.py
```diff
@@ -2814,6 +2814,29 @@ def path_length(self, u, v):
...
+    def b1_index(self):
+        """
+        Returns the rounded B1 balance index for this tree.
+        This is defined as the inverse of the sum of all longest paths
+        to leaves for each node besides roots.
+
+        .. seealso:: See `Shao and Sokal (1990)
+            <https://www.jstor.org/stable/2992186>`_ for details.
+
+        :return: The B1 balance index (rounded).
```
jeromekelleher, line 2826: Delete (rounded)

jeromekelleher (approved):
LGTM, spotted one issue in docstring

PR #2283, opened 2022-05-20
Remove tsutil.decapitate

### LINKED ISSUE #2269: Replace tsutil.decapitate in tests
There is currently a function called ``decapitate`` which does something slightly different to the definition in #2240. It would be better to remove this test function and replace it's usages with the new method. Should be a straightforward job.

jeromekelleher:
Closes #2269

Waiting on #2240


benjeffery (approved):
LGTM!

PR #2291, opened 2022-05-22
Fixed note about plink, write_vcf, individual names and ploidy.

gtsambos:
The original code gives me the following error:
```
ValueError: Cannot specify ploidy when individuals present
```

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5440,7 +5440,7 @@ class TreeSequence:
                 n_dip_indv = int(ts.num_samples / 2)
                 indv_names = [f"tsk_{str(i)}indv" for i in range(n_dip_indv)]
                 with open("output.vcf", "w") as vcf_file:
-                    ts.write_vcf(vcf_file, ploidy=2, individual_names=indv_names)
+                    ts.write_vcf(vcf_file, individual_names=indv_names)
 
             Adding a second ``_`` (eg: ``tsk_0_indv``) is not recommended as
             ``plink`` uses ``_`` as the default separator for separating family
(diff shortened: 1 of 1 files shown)


gtsambos:
Hmm. so I originally thought this error was coming up because of the `individual_names` argument being used with `ploidy`, but now I think it's a more general issue to do with the now-default behaviour of msprime to simulate diploid individuals

python/tskit/trees.py
```diff
@@ -5352,11 +5352,13 @@ def write_vcf(
...
                 indv_names = [f"tsk_{str(i)}indv" for i in range(n_dip_indv)]
                 with open("output.vcf", "w") as vcf_file:
-                    ts.write_vcf(vcf_file, ploidy=2, individual_names=indv_names)
+                    ts.write_vcf(vcf_file, individual_names=indv_names)
 
             Adding a second ``_`` (eg: ``tsk_0_indv``) is not recommended as
             ``plink`` uses ``_`` as the default separator for separating family
-            id and individual id, and two ``_`` will throw an error.
+            id and individual id, and two ``_`` will throw an error. In this
+            case, ploidy does not need to be specified, as it is implied by the
```
jeromekelleher, line 5360: I think this last sentence is a bit confusing when it comes directly after the note about plink. I think we just need to say that ``ploidy`` cannot be provided when there is individual data in the tree sequence in the parameter documentation down further.

I think the issue here is that these examples were run using msprime 0.x and may be out of date now.

jeromekelleher (commented):
Good spot, thanks @gtsambos. I think the simplest thing to do is to update the examples to work with msprime 1.0, and put in a note somewhere saying that older code that used the ploidy argument may need to be updated.

gtsambos:
lmk if this suffices @jeromekelleher

python/tskit/trees.py
```diff
@@ -5521,7 +5521,8 @@ def write_vcf(
```
petrelharp, line 5521:
```suggestion

.. note::
   Older code often uses the ``ploidy=2`` argument, because previous versions of msprime did not
   output individual data. Specifying individuals in the tree sequence is more robust, and since
   tree sequences now  typically contain individuals (e.g., as produced by
   ``msprime.sim_ancestry( )``), this is not necessary, and the ``ploidy`` argument can safely be removed
   from old code in most cases.
```

petrelharp:
I think this is a pretty common thing, so I made a suggestion?

jeromekelleher:
LGTM, and Peter's suggestion is good.

PR #2292, opened 2022-05-23
Place result as the last arg in stats API

### LINKED ISSUE #2285: Update stats API to make options last non-output parameter
We're pretty consistent elsewhere on this pattern of having ``options`` be the last param non output param. When we're documenting the stats API, we need to update this, so we have eg

```c
typedef int one_way_weighted_method(const tsk_treeseq_t *self, tsk_size_t num_weights,
    const double *weights, tsk_size_t num_windows, const double *windows, tsk_flags_t options,
   double *result);
```

benjeffery:
Seemed to make sense to do this before 1.0 even though these are undocumented.
Fixes #2285

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -1,6 +1,6 @@
-----------------------
-[1.0.0b1] - 2022-05-11
-----------------------
+--------------------
+[1.0.0] - 2022-05-XX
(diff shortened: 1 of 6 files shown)


jeromekelleher (approved):
LGTM, thanks @benjeffery!

Issue #2241, opened 2022-05-05
Doc links for preorder and postorder broken


jeromekelleher:
Looks like formatting issue in the docs for the traversal functions.

benjeffery:
Fixed in #2298

Issue #1481, opened 2021-06-03
add methods to extract node information for individuals


petrelharp:
`time` and `population` is stored by *node* even though it is an individual property. (at least, in all our use cases so far...) So, it's helpful to be able to pull this information out for individuals from their nodes, and it's a bit annoying to do that efficiently. I've got pyslim pre-extracting this information [here](https://github.com/tskit-dev/pyslim/blob/ced273690ca930ef27f6efa205ea20ff419e7361/pyslim/slim_tree_sequence.py#L180) (and, checking that there's no conflicting information). I think this should probably be a tskit method (and, writing it in C would make efficient error checking much easier than what I did there =).

So, I'm proposing
```
class TreeSequence:
   def individual_times(self):
       """
       Returns an array of times of length equal to the number of individuals, so that all nodes of individual ``i``
       have ``time`` attribute equal to the ``i``-the entry of the returned array. Raises an error if the nodes of an individual
       do not all have the same time.
       """

   def individual_populations(self):
```

I suppose I should provide an example showing that doing this in python in the obvious way can take a lot of time, e.g.
```
individual_times = [ts.node(ind.nodes[0]).time if len(ind.nodes) > 0 else np.nan for ind in ts.individuals()]
```
... I swear, in applications it has taken a while.

I guess alternatively we could make these properties of the Individual, so if we wanted a numpy array we'd do
```
individual_times = np.array([ind.time for ind in ts.individuals()])
```

jeromekelleher:
SGTM. I think we should also make ``time`` and ``population`` properties of the Individual class, which will raise an error if the times aren't the same in all nodes. 

I think it makes sense to have them on the TreeSequence rather than the TableCollection.

benjeffery:
Yes, these can be analogous to `site.mutations`, in that they are conveniences in the `TreeSequence` entity classes. Having them in `TableCollection` is bug-prone as they can get out of sync.

They could be lazily created if expensive.

molpopgen:
This is useful.  I've kept the node info as individual metadata in order to do this kinda thing.

petrelharp:
I'm removing `SlimTreeSequence` from pyslim, which was providing this functionality there. I'd like to get this in to tskit so we won't have (much) breakage. I'll make up a PR.

petrelharp:
Proposal: also provide `individual_locations`, which is a numpy array with one row per individual, as long as everyone has the same length of location (and error otherwise).

[...]

PR #2298, opened 2022-05-24
individual convenience extractors

petrelharp:
Ok, here I'm adding `ts.individual_locations`, `ts.individual_times`, and `ts.individual_nodes`. These are all cached properties... which I *think* does what we want? To be clear, what (I think) we want is to (a) have them only be computed once, and (b) not be returning new copies of them every time they are accessed.  Hm - if this is working, I should make them read-only (how to do this?), so people don't think they can modify the tree sequence by modifying these.

The main decision I made that I'm not sure about is the case when some of an individuals' nodes have `population=-1` and some have a legit population. I made the decision - honestly, for convenience - to treat `-1` as "missing data", and so an individual with node populations `-1` and `1` would then be reported as having population `1`, with no contradiction. (i.e., the unique non-null population) This seems nice but I haven't thought through other implications of this sort of thing.

I also said I'd add `time` and `population` properties to the `Individual` class. However, I'm not sure what to put in for these properties in the case where these are not well-defined (ie the values are not consistent across nodes). Perhaps I should *not* do this and instruct people to use, e.g., `ts.individual_times[ind.id]` instead.

*Note:* I also made one of the generic "make up examples" methods in `tsutil` a bit more general (it was only adding individuals to samples); hopefully this does not break anything (if it does, that will likely be a bug).

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -571,6 +571,18 @@ tsk_strerror_internal(int err)
             ret = "Individuals cannot be their own ancestor. "
                   "(TSK_ERR_INDIVIDUAL_PARENT_CYCLE)";
             break;
+
+        case TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH:
+            ret = "Individual populations cannot be returned "
+                  "if an individual has nodes from more than one population. "
+                  "(TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH)";
+            break;
+
+        case TSK_ERR_INDIVIDUAL_TIME_MISMATCH:
+            ret = "Individual times cannot be returned "
+                  "if an individual has nodes from more than one time. "
+                  "(TSK_ERR_INDIVIDUAL_TIME_MISMATCH)";
+            break;
     }
     return ret;
 }
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -716,6 +716,68 @@ tsk_treeseq_has_reference_sequence(const tsk_treeseq_t *self)
     return tsk_table_collection_has_reference_sequence(self->tables);
 }
 
+int
+tsk_treeseq_get_individuals_population(const tsk_treeseq_t *self, tsk_id_t *output)
+{
+    int ret = 0;
+    tsk_size_t i, j;
+    tsk_individual_t ind;
+    tsk_id_t ind_pop;
+    const tsk_id_t *node_population = self->tables->nodes.population;
+    const tsk_size_t num_individuals = self->tables->individuals.num_rows;
+
+    tsk_memset(output, TSK_NULL, num_individuals * sizeof(*output));
+
+    for (i = 0; i < num_individuals; i++) {
+        ret = tsk_treeseq_get_individual(self, (tsk_id_t) i, &ind);
+        tsk_bug_assert(ret == 0);
+        if (ind.nodes_length > 0) {
+            ind_pop = -2;
+            for (j = 0; j < ind.nodes_length; j++) {
+                if (ind_pop == -2) {
+                    ind_pop = node_population[ind.nodes[j]];
+                } else if (ind_pop != node_population[ind.nodes[j]]) {
+                    ret = TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH;
+                    goto out;
+                }
+            }
+            output[ind.id] = ind_pop;
+        }
+    }
+out:
+    return ret;
+}
+
+int
+tsk_treeseq_get_individuals_time(const tsk_treeseq_t *self, double *output)
+{
+    int ret = 0;
+    tsk_size_t i, j;
+    tsk_individual_t ind;
+    double ind_time;
+    const double *node_time = self->tables->nodes.time;
+    const tsk_size_t num_individuals = self->tables->individuals.num_rows;
+
+    for (i = 0; i < num_individuals; i++) {
+        ret = tsk_treeseq_get_individual(self, (tsk_id_t) i, &ind);
+        tsk_bug_assert(ret == 0);
+        /* the default is UNKNOWN_TIME, but nodes cannot have
+         * UNKNOWN _TIME so this is safe. */
+        ind_time = TSK_UNKNOWN_TIME;
+        for (j = 0; j < ind.nodes_length; j++) {
+            if (j == 0) {
+                ind_time = node_time[ind.nodes[j]];
+            } else if (ind_time != node_time[ind.nodes[j]]) {
+                ret = TSK_ERR_INDIVIDUAL_TIME_MISMATCH;
+                goto out;
+            }
+        }
+        output[ind.id] = ind_time;
+    }
+out:
+    return ret;
+}
+
 /* Stats functions */
 
 #define GET_2D_ROW(array, row_len, row) (array + (((size_t)(row_len)) * (size_t) row))
--- a/c/tskit/trees.h
+++ b/c/tskit/trees.h
@@ -697,7 +697,7 @@ Returns the location of each node in the list of samples or
 @endrst
 
 @param self A pointer to a tsk_treeseq_t object.
-@return Returns the pointer to the breakpoint array.
+@return Returns the pointer to the array of sample indexes.
 */
 const tsk_id_t *tsk_treeseq_get_sample_index_map(const tsk_treeseq_t *self);
 
@@ -889,6 +889,9 @@ int tsk_treeseq_split_edges(const tsk_treeseq_t *self, double time, tsk_flags_t 
     tsk_flags_t options, tsk_treeseq_t *output);
 
 bool tsk_treeseq_has_reference_sequence(const tsk_treeseq_t *self);
+
+int tsk_treeseq_get_individuals_population(const tsk_treeseq_t *self, tsk_id_t *output);
+int tsk_treeseq_get_individuals_time(const tsk_treeseq_t *self, double *output);
 
 int tsk_treeseq_kc_distance(const tsk_treeseq_t *self, const tsk_treeseq_t *other,
     double lambda_, double *result);
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2448,6 +2448,122 @@ class TestTreeSequence(HighLevelTestCase):
                         assert edge.right >= tree.interval.right
             assert np.all(edge_visited)
 
+    def verify_individual_vectors(self, ts):
+        verify_times = np.repeat(np.nan, ts.num_individuals)
+        verify_populations = np.repeat(tskit.NULL, ts.num_individuals)
+        for ind in ts.individuals():
+            if len(ind.nodes) > 0:
+                t = {ts.node(n).time for n in ind.nodes}
+                p = {ts.node(n).population for n in ind.nodes}
+                assert len(t) <= 1
+                assert len(p) <= 1
+                verify_times[ind.id] = t.pop()
+                verify_populations[ind.id] = p.pop()
+
+        times = ts.individuals_time
+        populations = ts.individuals_population
+        assert np.array_equal(times, verify_times, equal_nan=True)
+        assert np.array_equal(populations, verify_populations, equal_nan=True)
+        times2 = ts.individuals_time
+        populations2 = ts.individuals_population
+        assert np.array_equal(times, times2, equal_nan=True)
+        assert np.array_equal(populations, populations2, equal_nan=True)
+        # check aliases also
+        times3 = ts.individual_times
+        populations3 = ts.individual_populations
+        assert np.array_equal(times, times3, equal_nan=True)
+        assert np.array_equal(populations, populations3, equal_nan=True)
+
+    def test_individuals_population_errors(self):
+        t = tskit.TableCollection(sequence_length=1)
+        t.individuals.add_row()
+        t.individuals.add_row()
+        for j in range(2):
+            t.populations.add_row()
+            t.nodes.add_row(time=0, population=j, individual=0)
+        ts = t.tree_sequence()
+        with pytest.raises(
+            _tskit.LibraryError, match="TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH"
+        ):
+            _ = ts.individuals_population
+        # inconsistent but NULL populations are also an error
+        t.nodes.clear()
+        t.nodes.add_row(time=0, population=1, individual=0)
+        t.nodes.add_row(time=0, population=tskit.NULL, individual=0)
+        ts = t.tree_sequence()
+        with pytest.raises(
+            _tskit.LibraryError, match="TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH"
+        ):
+            _ = ts.individuals_population
+        t.nodes.clear()
+        t.nodes.add_row(time=0, population=tskit.NULL, individual=1)
+        t.nodes.add_row(time=0, population=0, individual=1)
+        ts = t.tree_sequence()
+        with pytest.raises(
+            _tskit.LibraryError, match="TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH"
+        ):
+            _ = ts.individuals_population
+
+    def test_individuals_time_errors(self):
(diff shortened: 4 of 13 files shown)


petrelharp:
Well, allowing non-sample individuals *did* break some things, albeit in predictable ways - mostly around VCF output. (I'm a bit nervous we're not adequately testing VCF output in the case where there's non-sample individuals, though.)

python/tests/test_tables.py
```diff
@@ -4621,7 +4621,9 @@ def get_msprime_example(self, sample_size, T, seed):
             random_seed=seed,
         )
         ts = tsutil.add_random_metadata(ts, seed)
-        ts = tsutil.insert_random_ploidy_individuals(ts, max_ploidy=1)
+        ts = tsutil.insert_random_ploidy_individuals(
+            ts, max_ploidy=1, samples_only=True
```
petrelharp, line 4625: this is in `TestUnion` - without `samples_only=True` it caused an error, which I worried about, but looked at the situation in detail, and everything is fine (the problem comes from wanting to split up a tree sequence into disjoint bits to then union back together, which is easier if all the individuals are from the same time, and we're not losing generality in the testing here, really)

petrelharp:
> I also said I'd add time and population properties to the Individual class. However, I'm not sure what to put in for these properties in the case where these are not well-defined (ie the values are not consistent across nodes). Perhaps I should not do this and instruct people to use, e.g., ts.individual_times[ind.id] instead.

Thinking about this a bit more - `ind.time` could, for instance, return all the distinct `time`s when the individual's nodes have more than one `time`. But, that's a bit icky because the return type would then need to be always list-like even in the common case when it's just really a single float. (Maybe not a big deal?) 

Alternatively we could put our "missing time" in for the value when `time` is not unique, and `tskit.NULL` when the population is not unique? (We couldn't distinguish the latter case from the case where nodes all agree, but have `tskit.NULL` for their population, but perhaps that's ok?)

Out of these options I like the second one.

python/tskit/trees.py
```diff
@@ -5073,6 +5076,71 @@ def alignments(self, *, reference_sequence=None, missing_data_character=None):
...
+    @property
+    def individual_populations(self):
+        """
+        Returns the length-``num_individuals`` array containing, for each
+        individual, the ``population`` attribute of their nodes, or
+        ``tskit.NULL`` for individuals with no nodes. Errors if any individual
+        has nodes with inconsistent non-NULL populations.
+        """
+        if self._individual_populations is None:
+            pops = np.repeat(tskit.NULL, self.num_individuals)
```
jeromekelleher, line 5088: Better to be explicity about the ``dtype=np.int32`` here.

python/tskit/trees.py
```diff
@@ -5073,6 +5076,71 @@ def alignments(self, *, reference_sequence=None, missing_data_character=None):
...
+            for n in self.nodes():
```
jeromekelleher, line 5089: What about doing it this way?
```python
for ind in self.individuals():
     if len(ind.nodes) > 0: 
          ind_pops = set(self.node(u).population for u in ind.nodes)
          if len(ind_pops) > 1:
               raise ValueError("...")
           pops[ind.id] = ind_pops.pop() # or whatever the set operation is, can't remember
```            
This avoids iteration over all nodes, and makes the "populations must be exactly equal" semantics easier to implement. 

I wouldn't worry too much about the efficiency of doing set operations here, as we'll probably want to drop this functionality down to the C api at some point anyway. I'm pretty sure I've written the equivalent C code in msprime, so we could just copy that.

python/tests/test_highlevel.py
```diff
@@ -2445,6 +2445,103 @@ def test_tree_node_edges(self):
...
+        t.clear()
+        t.individuals.add_row(location=[1.0, 2.0])
+        t.individuals.add_row(location=[])
+        t.individuals.add_row(location=[1.0, 2.0])
+        t.individuals.add_row(location=[])
+        ts = t.tree_sequence()
+        with pytest.raises(ValueError, match="locations"):
+            _ = ts.individual_locations
+
+    def test_individual_locations(self):
```
jeromekelleher, line 2528: Better to do this sort of thing with pytest.mark.parametrize. Here you'd do
```
@pytest.mark.parametrize("nlocs", [0, 1, 4])
@pytest.mark.parametrize("num_indivs", [0, 3])
def test_individual_locations(self, nlocs, num_indivis):
```

python/tests/test_highlevel.py
```diff
@@ -2445,6 +2445,103 @@ def test_tree_node_edges(self):
...
+    def test_individual_times_errors(self):
+        t = tskit.TableCollection(sequence_length=1)
+        t.individuals.add_row()
+        for j in range(2):
+            t.nodes.add_row(time=j, individual=0)
+        ts = t.tree_sequence()
+        with pytest.raises(ValueError, match="time"):
+            _ = ts.individual_times
+
+    def test_individual_vectors(self):
```
jeromekelleher, line 2498: This is a another place to use pytest.mark.parametrize

python/tests/test_highlevel.py
```diff
@@ -2445,6 +2445,103 @@ def test_tree_node_edges(self):
...
+                t = {ts.node(n).time for n in ind.nodes}
+                p = {ts.node(n).population for n in ind.nodes}
+                assert len(t) <= 1
+                assert len(p) <= 1
+                verify_times[ind.id] = t.pop()
+                verify_populations[ind.id] = p.pop()
+
+        times = ts.individual_times
+        populations = ts.individual_populations
+        assert np.array_equal(times, verify_times, equal_nan=True)
```
jeromekelleher, line 2462: This makes me wonder if individuals with no nodes should have an infinite time rather than nan? np.inf == np.inf is True.

python/tskit/trees.py
```diff
@@ -2195,8 +2195,8 @@ def num_tracked_samples(self, u=None):
 
     def preorder(self, u=NULL):
         """
-        Returns a numpy array of node ids in preorder
-        <https://en.wikipedia.org/wiki/Tree_traversal#Pre-order_(NLR)>. If the node u
+        Returns a numpy array of node ids in `preorder
```
jeromekelleher, line 2198: Closes #2241

jeromekelleher (commented):
LGTM!

Two high-level things:

1. I don't like the semantics around -1, I think it would be better to require strict equality of population (etc) ids.
2. Bike-sheddingly, I wonder about the names. We ended up using things like ``individuals_time`` in tsinfer ([eg](https://github.com/tskit-dev/tsinfer/blob/10d7274fceeb0c15933037e62ced59771aa63630/tsinfer/formats.py#L1166)). The rationale was we usually use arrays in the singlar in the tables API (nodes.time, say), and it's then inconsistent to flip this around. We could imagine doing this for all the table arrays (plus some extras, like say, ``ts.mutations_edge``), where we're just essentially replacing a "." with a "_".

We probably would want to push the implementation down to C for speed, but that wouldn't have to happen straight away.

petrelharp:
>     1. I don't like the semantics around -1, I think it would be better to require strict equality of population (etc) ids.

Hmph. That will make the implementation more annoying. =) But, agreed.

>     2. Bike-sheddingly, I wonder about the names. We ended up using things like `individuals_time` in tsinfer ([eg](https://github.com/tskit-dev/tsinfer/blob/10d7274fceeb0c15933037e62ced59771aa63630/tsinfer/formats.py#L1166)). The rationale was we usually use arrays in the singlar in the tables API (nodes.time, say), and it's then inconsistent to flip this around. We could imagine doing this for all the table arrays (plus some extras, like say, `ts.mutations_edge`), where we're just essentially replacing a "." with a "_".

This makes sense!

python/tskit/trees.py
petrelharp: That'd work, but I was writing this to be translatable to C more easily (which I was planning to do in this PR; if these operations are slow they'll be a bottleneck for things I do!)

python/tskit/trees.py
petrelharp: hah how about that

petrelharp:
However, changing it from `ts.individual_times` to `ts.individuals_time` is a breaking change from pyslim and will totally break code which would otherwise not be broken. I guess I"m voting to keep it the way it is, for this reason.

python/tests/test_highlevel.py
jeromekelleher: Yeah, fair enough.

python/tskit/trees.py
jeromekelleher: This is totally translatable to efficient C - I've done it all already [here](https://github.com/tskit-dev/msprime/blob/c11a2f12c72dd054a0ade0767474184ceec8281c/lib/msprime.c#L7207) in msprime.

jeromekelleher:
> However, changing it from ts.individual_times to ts.individuals_time is a breaking change from pyslim and will totally break code which would otherwise not be broken. I guess I"m voting to keep it the way it is, for this reason.

Not breaking code is good... How about we do it both ways, like this?

```python
@property
def individuals_time(self):
    """
    Documentation
    """
    # implementation

@property
def individual_times(self):
    # Undocumented alias for individuals_time to avoid breaking pre (VERSION) pyslim code
    return individuals_time
```

python/tskit/trees.py
```diff
@@ -5085,17 +5085,18 @@ def individual_populations(self):
         has nodes with inconsistent non-NULL populations.
         """
         if self._individual_populations is None:
-            pops = np.repeat(tskit.NULL, self.num_individuals)
+            pops = np.full(self.num_individuals, tskit.NULL, dtype=np.int32)
+            seen = np.full(self.num_individuals, False, dtype=bool)
```
jeromekelleher, line 5089: You don't need the full ``seen`` array - you can look at the local values for a given individual. In C-like code you'd have:
```
for ind in self.individuals():
     if len(ind.nodes) > 0: 
          ind_pop = -2
          for u in ind.nodes:
               if ind_pop == -2:
                  ind_pop = tables.nodes.population[u]
               elif ind_pop != tables.nodes.population[u]:
                   raise ValueError(..)
           population[ind.id] = ind_pop
```

petrelharp: Sorry, I'm confused - in C we won't have ready access to the individual's nodes when iterating over individuals, right?  Oh! I see, we *do*! Sorry, I was looking at the `stable` documentation, which I see hasn't updated to C 1.0 yet. Yes, that is a lot easier.

jeromekelleher:
LGTM - shall we merge this much and follow up with a C implementation?

petrelharp:
Er, I think I've got the C implementation now?

petrelharp:
I *think* this is good to go, although I haven't seen the codecov report for the C code.

Except for the darn clang-format versioning issue; I guess I'll fix those things up by hand.

petrelharp:
Ok, I've got the compile-with-big-tables issues fixed, and now CircleCI is showing a failure in the C tests that I do not see locally:
```
  Test: test_treeseq_get_individuals_time ...FAILED
    1. ../c/tests/test_trees.c:6846  - CU_ASSERT_EQUAL_FATAL(output[0],3.2)
```
Any idea what might be causing that?

jeromekelleher (commented):
LGTM!

The failure is because you're using decimal fractions which are recurrent in base 2, and older x86 math processors don't handle them in the way you might expect.

c/tskit/trees.c
```diff
@@ -716,6 +716,72 @@ tsk_treeseq_has_reference_sequence(const tsk_treeseq_t *self)
...
 }
 
+int
+tsk_treeseq_get_individuals_population(const tsk_treeseq_t *self, tsk_id_t *output)
+{
+    int ret = 0;
+    tsk_size_t i, j;
+    tsk_individual_t ind;
+    tsk_node_t node;
+    tsk_id_t ind_pop;
```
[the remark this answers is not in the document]
petrelharp: oooo, clever!

c/tskit/core.c
```diff
@@ -571,6 +571,18 @@ tsk_strerror_internal(int err)
             ret = "Individuals cannot be their own ancestor. "
                   "(TSK_ERR_INDIVIDUAL_PARENT_CYCLE)";
             break;
+
+        case TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH:
+            ret = "Individual populations cannot be returned "
+                  "if an individual has nodes from more than one population. "
+                  "(TSK_ERR_MISMATCHING_INDIVIDUAL_POPULATIONS)";
```
jeromekelleher, line 578: string out of date

c/tskit/core.c
```diff
@@ -571,6 +571,18 @@ tsk_strerror_internal(int err)
...
+            break;
+
+        case TSK_ERR_INDIVIDUAL_TIME_MISMATCH:
+            ret = "Individual populations cannot be returned "
+                  "if an individual has nodes from more than one time. "
+                  "(TSK_ERR_MISMATCHING_INDIVIDUAL_TIMES)";
```
jeromekelleher, line 584: Ditto here

petrelharp:
Thanks! Great suggestions there. I'll squash this, just on the off chance that it's actually really ready to go now. <crossing fingers>

jeromekelleher (approved):
LGTM, spotted one small detail.

c/tskit/core.c
petrelharp: lol whoops

petrelharp: fixed!

jeromekelleher:
Oh yeah, can you update the Python changelog with this also please? Good to merge whenever then I think.

petrelharp:
Did the changelog. Note that this does *not* (necessarily) close #1481 because that one also discusses adding `population` and `time` to the `Individual` class. (I'm going to hop over there to record my thoughts on that topic.)

c/tskit/trees.h
```diff
@@ -697,7 +697,7 @@ Returns the location of each node in the list of samples or
 @endrst
 
 @param self A pointer to a tsk_treeseq_t object.
-@return Returns the pointer to the breakpoint array.
+@return Returns the pointer to the array of sample indexes.
```
benjeffery, line 700: Nice spot - my bad I think.

python/tests/test_highlevel.py
```diff
@@ -2448,6 +2448,114 @@ def test_tree_node_edges(self):
...
+
+    def test_individuals_population_errors(self):
+        t = tskit.TableCollection(sequence_length=1)
+        t.individuals.add_row()
+        t.individuals.add_row()
+        for j in range(2):
+            t.populations.add_row()
+            t.nodes.add_row(time=0, population=j, individual=0)
+        ts = t.tree_sequence()
+        with pytest.raises(_tskit.LibraryError, match="population"):
```
benjeffery, line 2485: Nitpicky - but I'm trying to move to a standard where we use the macro for the string match. e.g: 
```suggestion
        with pytest.raises(_tskit.LibraryError, match="TSK_ERR_INDIVIDUAL_POPULATION_MISMATCH"):
```

python/tskit/trees.py
```diff
@@ -5073,6 +5076,71 @@ def alignments(self, *, reference_sequence=None, missing_data_character=None):
...
+        """
+        if self._individuals_time is None:
+            self._individuals_time = self._ll_tree_sequence.get_individuals_time()
+        return self._individuals_time
+
+    @property
+    def individual_times(self):
+        # Undocumented alias for individuals_time to avoid breaking
+        # pre-1.0 pyslim code
+        return self.individuals_time
```
benjeffery, line 5115: Can we cover this line (and the populations one) if pyslim is depending on it?

benjeffery (commented):
LGTM, just a few nits.

python/tskit/trees.py
petrelharp: Uh-oh, I did put in something that covers that but looks like I was too quick with the search-and-replace. =)

petrelharp:
OK - this is ready to go! I'll someone else hit 'merge' though.

benjeffery:
> OK - this is ready to go! I'll someone else hit 'merge' though.

I'll get the bot to hit the button! Thanks @petrelharp!

Issue #2236, opened 2022-05-04
"decapitate" function to delete edges older than a particular time.


jeromekelleher:
As discussed in #281, we want a way to delete all the topology in a tree sequence that's older than a particular time. Various names have been suggested for this operation, but since we've already settled on "recapitation" for the inverse of this operation (putting older topology on top of an existing tree sequence) then it seems simplest to just call it ``decapitate``.

benjeffery:
Shame that `decapitate` has such unsavory connotations, but I'm not sure I can think of a more tasteful name that fits as well.

PR #2302, opened 2022-05-26
Delete_older operation

jeromekelleher:
Next step on the road to decapitate (#2240, #2236). The idea is to implement this fully as a TableCollection method (not bothering with ts equivalent) in C.

Then decapitate is just the composition of this function and ``split_edges``. The advantage of doing this function on its own, is that we then don't have to do a C version of decapitate (which requires faddling around with all the node default args again).

What do you think of the definition @petrelharp? Is the name ok? I haven't tested the mutation and migration stuff so far, could well be wrong.

### DIFF
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -4217,7 +4217,9 @@ int tsk_table_collection_deduplicate_sites(
 int tsk_table_collection_compute_mutation_parents(
     tsk_table_collection_t *self, tsk_flags_t options);
 int tsk_table_collection_compute_mutation_times(
-    tsk_table_collection_t *self, double *random, tsk_flags_t TSK_UNUSED(options));
+    tsk_table_collection_t *self, double *random, tsk_flags_t options);
+int tsk_table_collection_delete_older(
+    tsk_table_collection_t *self, double time, tsk_flags_t options);
 
 int tsk_table_collection_set_indexes(tsk_table_collection_t *self,
     tsk_id_t *edge_insertion_order, tsk_id_t *edge_removal_order);
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3862,6 +3862,30 @@ class TableCollection(metadata.MetadataProvider):
                 record=json.dumps(provenance.get_provenance_dict(parameters))
             )
 
+    def delete_older(self, time):
+        """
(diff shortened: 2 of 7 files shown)


petrelharp:
The name seems good! We also have `tables.delete_sites` and `tables.delete_intervals`, which do similar things.

jeromekelleher:
I think this is ready to go - I'll update the changelog and docstring (to say maybe you want decapitate?) for this when decapitate is done in #2240

c/tskit/tables.h
```diff
@@ -4217,7 +4217,9 @@ int tsk_table_collection_deduplicate_sites(
 int tsk_table_collection_compute_mutation_parents(
     tsk_table_collection_t *self, tsk_flags_t options);
 int tsk_table_collection_compute_mutation_times(
-    tsk_table_collection_t *self, double *random, tsk_flags_t TSK_UNUSED(options));
+    tsk_table_collection_t *self, double *random, tsk_flags_t options);
```
petrelharp, line 4220: whoops - you've un-marked tsk_table_collection_compute_mutation_times's options as unused (they are actually unused; I just checked)

petrelharp:
```suggestion
    tsk_table_collection_t *self, double *random, tsk_flags_t TSK_UNUSED(options));
```

python/tskit/tables.py
```diff
@@ -3862,6 +3862,30 @@ def trim(self, record_provenance=True):
...
             )
 
+    def delete_older(self, time):
+        """
+        Deletes edge, mutation and migration information at least as old as
+        the specified time.
+
+        For the purposes of this method, the time of an edge is the time of its
+        *parent* node, so that any any edge that intersects with the specified
+        time will be removed.
```
petrelharp, line 3872:
```suggestion
        For the purposes of this method, an edge covers the times from the child node
        up until the *parent* node, so that any any edge with parent node time > ``time``
        will be removed.
```
The non-stictness of this inequality was not obvious.

petrelharp (approved):
LGTM!

c/tskit/tables.h
jeromekelleher: You don't need to do this in the header file though, it's just in the .c.

benjeffery (approved):
LGTM.

PR #2305, opened 2022-05-27
time and population attributes on individual

petrelharp:
As discussed in #1481, we'd like to add `time` and `population` to the Individual object. I would like to make this happen so that we don't break old code using pyslim, which provided those attributes. (Of course, if adding them turns out to be a bad design decision, then we should not do it.)

At first I thought that these would be arguments to `__init__`, with default values so as to be backwards compatible. However, it turns out that [defaults + slots + dataclasses don't work well together](https://stackoverflow.com/questions/50180735/how-can-dataclasses-be-made-to-work-better-with-slots). Furthermore, if we actually add these to the dataclass then they'll get printed out, which in #1481 we said maybe we didn't want to do (at least, we wanted more control over the printing). So, here I'm just adding them on as properties post-init, which seems nice and simple. We can worry about printing later.

One issue is what to do in the case where the nodes have mismatching properties. In #1481 we said we wanted to throw an error in this case. To do that, we'd need to stick a special value in the `Individual._time` slot, and provide a `.time()` property that 
throws an error if we ask for `ind.time` and we see the special value. However, this seems silly - why not just go ahead and return the special value to the user and let them deal with it appropriately? In fact, `None` is a good choice for this special value, because (as opposed to, say, our "missing time" value) because it's not a number at all, so it'll quickly cause problems if people try to use it like a time.

The other question is what to return when an individual has *no* nodes. It would be nice to return something different in that case, although arguably it's not a different situation - in either case of mismatching or no nodes, the indiividual's time is "we don't know".  So, we could do something like this:
- `time`: None if no nodes, MISSING if inconsistent nodes
- `population`: None if no nodes, ??? if inconsistent nodes

(or, maybe vice-versa?)... but, what's `???`? It can't be `tskit.NULL` because that's a valid value for `Node.population`.

So, the way I've drafted this up, the properties are `None` in either case of no nodes or mismatching nodes. I currently think this is actually the right way to go, but other suggestions welcome. After all, if someone wants to disambiguate the two situataions they can go look at `len(ind.nodes) > 0`. And if we want to save people having to go look at `ind.nodes` in *all* situations, well then we ought to return actually the list of node times in the case where the times are mismatching (which still seems like the wrong way to go).

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -27,6 +27,7 @@ Module responsible for managing trees and tree sequences.
 from __future__ import annotations
 
 import base64
+import builtins
 import collections
 import concurrent.futures
 import functools
@@ -93,6 +94,19 @@ class EdgeDiff(NamedTuple):
     edges_in: list
 
 
+def store_tree_sequence(cls):
+    wrapped_init = cls.__init__
+
+    # Intercept the init to record the tree_sequence
+    def new_init(self, *args, tree_sequence=None, **kwargs):
+        builtins.object.__setattr__(self, "_tree_sequence", tree_sequence)
+        wrapped_init(self, *args, **kwargs)
+
+    cls.__init__ = new_init
+    return cls
+
+
+@store_tree_sequence
 @metadata_module.lazy_decode
 @dataclass
 class Individual(util.Dataclass):
@@ -106,7 +120,15 @@ class Individual(util.Dataclass):
     underlying tree sequence data.
     """
 
-    __slots__ = ["id", "flags", "location", "parents", "nodes", "metadata"]
+    __slots__ = [
+        "id",
+        "flags",
+        "location",
+        "parents",
+        "nodes",
+        "metadata",
+        "_tree_sequence",
+    ]
     id: int  # noqa A003
     """
     The integer ID of this individual. Varies from 0 to
@@ -136,6 +158,24 @@ class Individual(util.Dataclass):
     The :ref:`metadata <sec_metadata_definition>`
     for this individual, decoded if a schema applies.
     """
+
+    @property
+    def population(self) -> int:
+        populations = {self._tree_sequence.node(n).population for n in self.nodes}
+        if len(populations) > 1:
+            raise ValueError("Individual has nodes with mis-matched populations")
+        if len(populations) == 0:
+            return tskit.NULL
+        return populations.pop()
+
+    @property
+    def time(self) -> int:
+        times = {self._tree_sequence.node(n).time for n in self.nodes}
+        if len(times) > 1:
+            raise ValueError("Individual has nodes with mis-matched times")
+        if len(times) == 0:
+            return tskit.UNKNOWN_TIME
+        return times.pop()
 
     # Custom eq for the numpy arrays
     def __eq__(self, other):
@@ -5048,7 +5088,7 @@ class TreeSequence:
             metadata,
             nodes,
         ) = self._ll_tree_sequence.get_individual(id_)
-        return Individual(
+        ind = Individual(
             id=id_,
             flags=flags,
             location=location,
@@ -5056,7 +5096,9 @@ class TreeSequence:
             metadata=metadata,
             nodes=nodes,
             metadata_decoder=self.table_metadata_schemas.individual.decode_row,
+            tree_sequence=self,
         )
+        return ind
 
     def node(self, id_):
         """
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2564,6 +2564,97 @@ class TestTreeSequence(HighLevelTestCase):
         locs3 = ts.individual_locations
         assert np.array_equal(ts_locs, locs3)
 
+    def verify_individual_properties(self, ts):
+        for ind in ts.individuals():
+            times = [ts.node(n).time for n in ind.nodes]
+            if len(set(times)) > 1:
+                with pytest.raises(ValueError, match="mis-matched times"):
+                    _ = ind.time
+            elif len(times) == 0:
+                assert tskit.is_unknown_time(ind.time)
+            else:
+                assert len(set(times)) == 1
+                assert times[0] == ind.time
+                # test accessing more than once in case we mess up with {}.pop()
+                assert times[0] == ind.time
+            pops = [ts.node(n).population for n in ind.nodes]
+            if len(set(pops)) > 1:
+                with pytest.raises(ValueError, match="mis-matched populations"):
+                    _ = ind.population
+            elif len(pops) == 0:
+                assert ind.population is tskit.NULL
+            else:
+                assert len(set(pops)) == 1
+                assert ind.population == pops[0]
+                # test accessing more than once in case we mess up with {}.pop()
+                assert ind.population == pops[0]
+
+    def test_individual_getter_population(self):
+        tables = tskit.TableCollection(sequence_length=1)
+        for _ in range(2):
+            tables.populations.add_row()
+        pop_list = [
+            ((), tskit.NULL),
+            ((tskit.NULL,), tskit.NULL),
+            ((1,), 1),
+            ((1, 1, 1), 1),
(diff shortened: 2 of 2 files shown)


petrelharp:
Sorry, hold on - I'll put some more information here.

python/tskit/trees.py
```diff
@@ -148,6 +148,10 @@ def __eq__(self, other):
             and self.metadata == other.metadata
         )
 
+    def __post_init__(self):
+        self.time = None
+        self.population = None
+
```
jeromekelleher, line 154: Hmm, what about:
```
@property
def time(self):
    times = {self.node(u).time for u in self.nodes}
    if len(times) != 1:
         raise ValueError("This individuals does not have a well defined time")
    return times.pop()
```

jeromekelleher (commented):
I'm a bit wary of assigning particular values to the different cases when the time/population is undefined. I think it's better to just raise an error directly rather than depend on downstream code erroring out with e.g. None being used in a calculation.

The way about should avoid all the problems unless we specifically ask for the individual's ``time`` value.

jeromekelleher:
Another reason for just raising an error in all cases: this is what the array code does, so having different semantics seems inconsistent and confusing.

petrelharp:
> Another reason for just raising an error in all cases: this is what the array code does, so having different semantics seems inconsistent and confusing.

It does seem reasonable that `ts.individual_times` should act just like `[ind.time for ind in ts.individuals()]`. Since we can't error in `__post_init__` (I think??), this is arguing for passing a special value which means "error if this is asked for". That would be OK if a bit awkward (what value? missing for time and, um, -2 for population??).

python/tskit/trees.py
jeromekelleher: Only if you did ``ts.individual(k).time`` or ``[ind.time for ind in ts.individuals()]`` - which is all you can do I think?

This can't break any existing code because there's no place that we use these attributes in the library (which we should be *very* careful about doing, I agree)

jeromekelleher:
This seems like the perfect application then for doing it dynamically with a property?

python/tskit/trees.py
petrelharp: Sorry, I was being too hasty. I don't think we can do just this because (unless I'm wrong?) here `self` is an Individual object, not the TreeSequence, and so we can't go back and pull out the times from the nodes in a `@property`? Ah, but I could set `self.times = {self.node(u).time for u in nodes}` in `ts.individual(k)` and then do like so:
```
@property
def time(self):
  if len(self.times) == 1:
    return self.times[0]  # not right because it's a set
  else:
    raise ValueError(...)
```

petrelharp:
As above - I don't think it can be a property because at `@property`-time we don't have the tree sequence any more. But, I could just store the times and populations under the hood (e.g., as `Individual._times`) and then define a `@property` that does the appropriate checking. (Or, am I missing something?)

jeromekelleher:
> As above - I don't think it can be a property because at @property-time we don't have the tree sequence any more.
Ah dammit, forgot about that, my bad!

Hmm. I think we should have the same semantics for attribute and array access. Let's see what @benjeffery thinks about how to do that in the least-bad way  (he's back on Wednesday).

benjeffery:
I've pushed a commit here with my proposed solution:

- Now uses a `property` to calculate values on demand
- Stores the tree sequence in `Individual` but in a way that doesn't change the output of `asdict` or `repr` of the class
- Matches the semantics of the array methods, raising errors on mis-match and giving `tskit.NULL` or `UNKNOWN_TIME` when zero nodes.

Let me know what you think!

petrelharp:
Lol, I've just noticed that the commit I made this morning (and that I was referring to when I said "I think this does what we want it to") was rejected due to non-fast-forward-ness. (the hazards of hitting "git push" one minute before the kid wakes up)

Let's see - your approach seems good also? I don't really know the implications of keeping the tree sequence around rather than just some other information - that seems... different? I'm going to push my solution, so you can see it for comparison, but with your commit on top.

petrelharp:
Ok, here's the way I did it: 33d92fc6898b3766af59958e24a0bc2249ea1ea3

python/tskit/trees.py
```diff
@@ -137,6 +159,24 @@ class Individual(util.Dataclass):
...
     """
 
+    @property
+    def population(self) -> int:
+        populations = {self._tree_sequence.node(n).population for n in self.nodes}
+        if len(populations) > 1:
+            raise ValueError("Individual has nodes with mis-matched populations")
+        if len(populations) == 0:
+            return tskit.NULL
+        return next(iter(populations))
```
jeromekelleher, line 169: I think ``populations.pop()`` does the same thing, a bit more readably

benjeffery:
This is a trade-off between keeping a reference or pre-calculating something that might not be used. I'd prefer holding the reference, but don't have a good handle on how often `population` and `time` will be used in practise.

petrelharp:
Hopefully people use `ts.individuals_time` instead of `[ind.time for ind in ts.individuals()]`; but it's a good point - pre-calculating times and individuals would affect other things (eg `[ind.location for ind in ts.individuals()]`

python/tskit/trees.py
petrelharp: I've changed it to this.

jeromekelleher:
I think there was some collateral damage here with the variant class @petrelharp, and this needs a rebase?

It's a good point about the back references to the tree sequence potentially catching people out, I don't have a strong sense of which option is better. Probably doesn't make much difference in practise most of the time.

petrelharp:
OK, I've done some benchmarking. Using a 750Mb tree sequence I've got lying around with 2.6M individuals, doing
```
import tskit
ts = tskit.load("sim_0040/pme_1807386945562.trees")
n = 0
for ind in ts.individuals():
    n += (ind.flags & 65536 > 0)

print(f"Number alive: {n}")
```
Ben's code:
```
	User time (seconds): 14.13
	Maximum resident set size (kbytes): 1135388
```
My code:
```
	User time (seconds): 51.35
	Maximum resident set size (kbytes): 1139352
```
(we didn't expect the memory to be different in this case, I was just checking).

On the one hand this is a factor of three (nice job, Ben). But on the other hand this is a very big tree sequence and we're under a minute for both cases, so it's not a bottleneck; and we do see memory usage as a real bottleneck; keeping the reference around provides a gotcha for memory usage. But, again, maybe not very common?

I don't know; I'd go with your gut feeling? I keep waffling, but if the linkage back to the tree sequence doesn't make you worry, then I guess I'm leaning towards Ben's code?

jeromekelleher:
I don't really mind, I'll leave it to you and Ben!

WRT memory gotchas, these are going to be a thing. For example, if we get things working properly then something like this will catch people out:

```python
time = ts.tables.nodes.time
# We just wanted this array, so let's be clever and get rid of ts
del ts
# Aha, but the time array is backed by the memory of ts, so it's still alive!
```
this isn't going to happen at the moment, but invisible backrefs will become more of a thing as we optimise the memory model more.

petrelharp:
Sounds like it's up to you, @benjeffery! =)

benjeffery:
Thanks for the benchmarking! I'm assuming that your tree sequence was diploid? I guess the extra time taken is O(nodes-with-individuals)?

I think we go with the stored reference. My thinking here is that someone who is optimising memory usage is a rarer and more advanced user than the user who is iterating through individuals. My other consideration is that I can see this kind of stored reference being useful where we have other properties of tree-sequence entities that we might want to provide as properties that are not cheap to calculate. Hope that makes sense!

petrelharp:
OK; I squashed, and will merge when tests pass.

jeromekelleher:
The "AUTOMERGE REQUESTED" label is handy for this @petrelharp - just add the label when you're happy with the PR and let mergify do the rebasing and merging. It'll merge if there's one "approval" from a maintainer.

petrelharp:
Ah right, I keep forgetting about that.

PR #2311, opened 2022-06-01
Add extra detail to sample_lists error message

### LINKED ISSUE #2309: kc_distance() returns an error "sample_lists option must be enabled to perform this operation"
Applying `kc_distance()` in standard ways like this

```
tree = ts.first()
tree.kc_distance(other=tsi.first())
```
gives me a library error:
```
LibraryError: The sample_lists option must be enabled to perform this operation.
```

benjeffery:
Hard to be specific here, as we don't know which method, or even which language create the tree and should have set the option. A search of the docs for "sample_lists" does work to get you to the right place though.

Closes #2309


jeromekelleher (approved):
LGTM. 

Something we could also do is add a NoSampleListsException (or something) to the low-level Python C module, then catch this exception in ``kc_distance`` and give a clear instruction python-wise on what to do?

I think people are still going to struggle otherwise.

benjeffery:
Yes agreed - let's merge this but keep the issue open.

PR #2313, opened 2022-06-01
Remove PyVCF

### LINKED ISSUE #2312: Remove PyVCF
No longer supported and not worth the trouble.

benjeffery:
Fixes #2312 

We're losing a lot of tests if we don't replace PyVCF. Would `scikit-allele` be OK to add as a test dependency?


jeromekelleher:
Too slow @szhan !

jeromekelleher:
Scikit-allele is fine, not sure how authoritative it's implementation of VCF is though (or how well it'll be maintained going forward).

Could we rephrase those tests in terms of pysam APIs? Or we could move over to cyvcf2, which has a friendlier interface?

benjeffery:
> Too slow @szhan wink !

Whoops, sorry! I need to start checking the assignment status!

benjeffery:
Ok, have checked though, and I don't think we're losing anything critical here.

jeromekelleher (approved):
Nuke it!

PR #2323, opened 2022-06-08
test that all edges have been used

### LINKED ISSUE #2317: table indexes not actually getting dropped on clearing edges
I've run into a case where I'm getting errors on doing `tables.tree_sequence()` after clearing/rebuilding the edges because the indexes are stale. Here's a MWE, in which I (1) build some valid tables; (2) clear the edges and build a different set of valid tables; (3) try to get the resulting tree sequence.
```
import tskit

do_error = True

t = tskit.TableCollection(sequence_length=10)
t.nodes.add_row(time=0)
t.nodes.add_row(time=0)
t.nodes.add_row(time=1)
t.nodes.add_row(time=2)

t.edges.add_row(left=0, right=5, parent=2, child=0)
t.edges.add_row(left=0, right=5, parent=2, child=1)
t.edges.add_row(left=5, right=10, parent=3, child=0)
t.edges.add_row(left=5, right=10, parent=3, child=1)

# note it would suffice to just do t.build_index() for the next two lines
ts = t.tree_sequence()
t = ts.dump_tables()
print("Previous indexes:")
print(t.indexes)

t.edges.clear()
print("After clearing:")
print(t.indexes)

t.edges.add_row(left=0, right=5, parent=2, child=0)
t.edges.add_row(left=0, right=8, parent=2, child=1)
t.edges.add_row(left=5, right=10, parent=3, child=0)
t.edges.add_row(left=5, right=8, parent=3, child=2)

if do_error:
    print("Not building indexes.")
else:
    print("Rebuilding indexes.")
    t.build_index()

print("New indexes:")
print(t.indexes)

t.tree_sequence()
```
which produces
```
Previous indexes:
TableCollectionIndexes(edge_insertion_order=array([0, 1, 2, 3], dtype=int32), edge_removal_order=array([1, 0, 3, 2], dtype=int32))
After clearing:
TableCollectionIndexes(edge_insertion_order=None, edge_removal_order=None)
Not building indexes.
New indexes:
TableCollectionIndexes(edge_insertion_order=array([0, 1, 2, 3], dtype=int32), edge_removal_order=array([1, 0, 3, 2], dtype=int32))
::: (5.000000) 0 - 2 not 3
  File "/home/peter/debug/tskit/build_index/test.py", line 40, in <module>
    t.tree_sequence()
  File "/home/peter/.local/lib/python3.9/site-packages/tskit-0.4.1-py3.9-linux-x86_64.egg/tskit/tables.py", line 3267, in tree_sequence
    return tskit.TreeSequence.load_tables(self)
  File "/home/peter/.local/lib/python3.9/site-packages/tskit-0.4.1-py3.9-linux-x86_64.egg/tskit/trees.py", line 3730, in load_tables
    ts.load_tables(tables._ll_tables, build_indexes=build_indexes)
_tskit.LibraryError: Bad edges: contradictory children for a given parent over an interval. (TSK_ERR_BAD_EDGES_CONTRADICTORY_CHILDREN)
```

However, changing `do_error` to False doesn't throw an error and tells us what the indexes ought to have been for the second tree:
```
Previous indexes:
TableCollectionIndexes(edge_insertion_order=array([0, 1, 2, 3], dtype=int32), edge_removal_order=array([1, 0, 3, 2], dtype=int32))
After clearing:
TableCollectionIndexes(edge_insertion_order=None, edge_removal_order=None)
Rebuilding indexes.
New indexes:
TableCollectionIndexes(edge_insertion_order=array([0, 1, 2, 3], dtype=int32), edge_removal_order=array([0, 3, 1, 2], dtype=int32))
```

In short - even though printing out the indexes tells us they've been cleared, somehow they haven't been invalidated, and so doing `tables.tree_sequence()` tries to use them.

petrelharp:
This will at least ensure that the bad trees I could construct over at #2317 will get caught by check_integrity. I've also added a note about indexes to the "inconsitent children" error I was getting there.

So, I think this closes #2317.


c/tskit/core.h
```diff
@@ -528,21 +528,25 @@ The table collection was not indexed.
 */
 #define TSK_ERR_TABLES_NOT_INDEXED                                  -702
 /**
+There was an error with the table's indexes.
+*/
+#define TSK_ERR_TABLES_BAD_INDEXES                                  -703
+/**
 Tables cannot be larger than 2**31 rows.
 */
-#define TSK_ERR_TABLE_OVERFLOW                                      -703
```
jeromekelleher, line 533: We can't make these kinds of changes post 1.0. Just put the new error at the end of the table as 707

c/tskit/tables.c
```diff
@@ -10333,6 +10333,14 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
         }
         num_trees++;
     }
+    tsk_bug_assert(j == num_edges);
+    while (k < num_edges) {
```
jeromekelleher, line 10337: Nice!

jeromekelleher (approved):
Very nice!

Looks like you've got the wrong version of clang-format?

benjeffery:
@petrelharp If you're having trouble getting the right `clang-format` then just `git commit --no-verify` as it is _very_ easy for me to run the lint on the PR.

petrelharp:
I think the only way this could miss an edge is if the indexes had the same edge index more than once. If both `I` and `O` are of length `num_edges` and have the elements `0... num_edges-1`, then they each have to hit each edge more than once. And (I've just double-checked) we'll get an error if the indexes are the wrong length.

However, I suppose there could be nonsensical indexes? But more problematically, modifying my example above so the first edge in the new tree sequence has right endpoint 4 instead of 5 (i.e., less than sequence length) makes my checks here not work (and the example hits a bug_assert in tsk_tree_advance). So, good idea.

jeromekelleher:
Yep, LGTM. I'll probably be a head scratcher trying to hit all those error cases unfortunately, but I guess they can all happen given sufficiently weird index input.

petrelharp:
It was a head scratcher, all right. But I think I've hit all the places it's possible to go wrong.

petrelharp:
The codecov failure is just for an out-of-memory error. So I think this is good to go. (although I'm sure you'll find something =)

c/tskit/tables.c
```diff
@@ -10272,6 +10272,7 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
     const tsk_id_t *restrict edge_child = self->edges.child;
     const tsk_id_t *restrict edge_parent = self->edges.parent;
     tsk_id_t *restrict parent = NULL;
+    int *restrict used_edges = NULL;
```
jeromekelleher, line 10275: since this can only take values 0, 1 or 2 I think we should save a bit of memory and make it an ``int8_t``.

c/tskit/tables.c
```diff
@@ -10281,20 +10282,38 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
     }
     tsk_memset(parent, 0xff, self->nodes.num_rows * sizeof(*parent));
 
+    used_edges = tsk_malloc(num_edges * sizeof(*used_edges));
+    if (used_edges == NULL) {
```
jeromekelleher, line 10286: We can save a bit of vertical space (and coverage penalty) by putting this check in with the one for ``parent``, like `` if (parent == NULL || used_edges == NULL)``. This is a common pattern.

c/tskit/tables.c
```diff
@@ -10281,20 +10282,38 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
...
             parent[edge_child[O[k]]] = TSK_NULL;
+            used_edges[O[k]] += 1;
             k++;
         }
         while (j < num_edges && edge_left[I[j]] == tree_left) {
+            if (used_edges[I[j]] != 0) {
+                ret = TSK_ERR_TABLES_BAD_INDEXES;
+                goto out;
+            }
+            used_edges[I[j]] += 1;
```
jeromekelleher, line 10316: ``used_edges[e]++`` is a bit more C-idiomatic.

c/tskit/tables.c
```diff
@@ -10333,12 +10356,33 @@ tsk_table_collection_check_tree_integrity(const tsk_table_collection_t *self)
...
+         * since otherwise we would have added a different edge twice,
+         * and so hit the error above. */
+        if (edge_right[O[k]] != sequence_length) {
+            ret = TSK_ERR_TABLES_BAD_INDEXES;
+            goto out;
+        }
+        used_edges[O[k]] += 1;
+        k++;
+    }
+    for (j = 0; j < num_edges; j++) {
```
jeromekelleher, line 10371: Since this is super-perf critical I don't think we should do this check. I'm sure you've caught all the weird possibilities.

jeromekelleher (commented):
Looks great! A few more suggestions, as predicted :smile:

Sorry to nitpick, but this *is* totally perf critical code so it's important we get it as tight as we can.

PR #2327, opened 2022-06-08
Adding B2 index

jeremyguez:
Here is a PR for adding B2 index. 
I did not set 10 as a default for the base parameter (although Shao and Sokal (1990) used 10), in case people would assume another base without checking the default (e.g., Bienvenu et al. (2020) advises to use a base 2 for binary trees). I did all the tests with base=10, assuming tests that pass with one base would pass with any base.
What do you think @jeromekelleher?


python/tests/test_balance_metrics.py
```diff
@@ -32,6 +35,14 @@
 # we can remove this.
 
 
+def nodes_to_root(tree, u):
```
jeromekelleher, line 38: How about this:
```python
path = []
u = tree.parent(u)
while u != tskit.NULL:
      paty.append(u)
      u = tree.parent(u)
return path
```

(This might be a useful library function in fact. Something like,
```python
def path(self, u, v=None):
     """
     Returns the path between two nodes u and v, such that ``len(tree.path(u, v)) == tree.path_length(u, v)``.
     If the second node ``v`` is not specified, return the path from u to root.
     """
```

python/tests/test_balance_metrics.py
```diff
@@ -63,6 +74,16 @@ def b1_index_definition(tree):
...
 
 
+def b2_index_definition(tree, base):
+    if tree.num_roots != 1:
+        raise ValueError("B2 index not defined for multiroot trees")
+    proba = [
+        np.prod([1 / tree.num_children(u) for u in nodes_to_root(tree, leaf)])
+        for leaf in tree.leaves()
+    ]
+    return -sum(p * math.log(p, base) for p in proba)
```
jeromekelleher, line 84: Might as well stick with numpy ops here,
```
return -np.sum(proba * np.log(proba, base))
```

python/tests/test_balance_metrics.py
```diff
@@ -63,6 +74,16 @@ def b1_index_definition(tree):
     )
 
 
+def b2_index_definition(tree, base):
```
jeromekelleher, line 77: I think we can put in ``base=10`` by default here.

python/tskit/trees.py
```diff
@@ -2853,6 +2853,38 @@ def b1_index(self):
                 total += 1 / max_path_length[u]
         return total
 
+    def b2_index(self, base):
```
jeromekelleher, line 2856: Let's default base=10, and put in a paragraph justifying this decision in the docstring.

jeromekelleher (commented):
Looks good to me. I don't know what the best thing to do about the base is - I feel like we should put in a sensible default if the base doesn't matter all that much (which it probably doesn't?).

Is there some discussion somewhere about what the base means somewhere and what the consequences of choosing a given base are?

jeromekelleher:
@jeremyguez we're nearly ready to ship the 0.5.0 release - do you want to make a final push to get this in on time? I've done some final tidy-ups for the other metrics in #2346, and it would be nice to have all four in for the release.

I think the only decision we need to make here is whether we want a default base, and 10 seems OK to me?

jeremyguez:
@jeromekelleher yes, I am going to add the path function needed for B2 first (PR almost done) and then push B2 once the path function is there. Both should be done today (or by the end of the weekend). Does this timeline fit the schedule?

jeromekelleher:
> I'm finishing both today (or by the end of the weekend). Does this timeline fit the schedule?

Perfect! It'll be nice to get this all wrapped up for the release. Ping me when you'd like some input.

python/tests/test_balance_metrics.py
jeremyguez: Apparently, for numpy, `log` is always the natural logarithm (no base parameter, see [here](https://numpy.org/doc/stable/reference/generated/numpy.log.html)).

python/tskit/trees.py
jeremyguez: Good idea, I did that.

jeremyguez:
cc @jeromekelleher 
So I updated here the PR, and I will modify it for using the library `path` function when https://github.com/tskit-dev/tskit/pull/2350 is pushed.

python/tests/test_balance_metrics.py
jeromekelleher: Let's just update the version of ``nodes_to_root`` here to use the implementation I suggested and put #2350 on the back-burner for now. Searching ``tree.roots`` on every loop iteration is unnecessarily inefficient when we can just check what the parent is in O(1) time.

python/tests/test_balance_metrics.py
```diff
@@ -88,6 +109,19 @@ def test_b1(self, ts):
         for tree in ts.trees():
             assert tree.b1_index() == pytest.approx(b1_index_definition(tree))
 
+    @pytest.mark.parametrize("ts", get_example_tree_sequences())
```
jeromekelleher, line 112: Let's add another pytest.parametrize here like so:

```
@pytest.mark.parametrize("ts", get_example_tree_sequences())
@pytest.mark.parametrize("base", [2, 10, math.e, np.array([3])[0]])
def test_b2_base(self, ts, base):
    # as above

jeromekelleher (commented):
LGTM, just needs a couple of small changes and we're good to merge.

python/tests/test_balance_metrics.py
jeremyguez: Thanks for the input @jeromekelleher! I did the changes.

jeromekelleher:
Looks great, thanks @jeremyguez! Looks like you merged rather than rebased through - can you rebase and squash please?

jeremyguez:
> Looks great, thanks @jeremyguez! Looks like you merged rather than rebased through - can you rebase and squash please?

I tried to rebase, but too many conflicts to resolve manually, I don't know why...
So I made a new branch with the last version from this branch. The PR is in https://github.com/tskit-dev/tskit/pull/2353.
Sorry for the inconvenience!

jeromekelleher:
Closing in favour of #2353

PR #2331, opened 2022-06-09
Decapitate2

jeromekelleher:
Stacked on #2302.

Finally, decapitate fully implemented!

The testing is perhaps a bit overboard, but I had implemented all of this in #2240 and it seemed a waste to just throw it away.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3867,9 +3867,16 @@ class TableCollection(metadata.MetadataProvider):
         Deletes edge, mutation and migration information at least as old as
         the specified time.
 
-        For the purposes of this method, an edge covers the times from the child node
-        up until the *parent* node, so that any any edge with parent node time > ``time``
-        will be removed.
+        .. seealso:: This method is similar to the higher-level
+            :meth:`TreeSequence.decapitate` method, which also splits
+            edges that intersect with the given time.
+            :meth:`TreeSequence.decapitate`
+            is more useful for most purposes, and may be what
+            you need instead of this method!
+
+        For the purposes of this method, an edge covers the times from the
+        child node up until the *parent* node, so that any any edge with parent
+        node time > ``time`` will be removed.
 
         Any mutation whose time is >= ``time`` will be removed. A mutation's time
         is its associated ``time`` value, or the time of its node if the
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -219,6 +219,7 @@ which perform the same actions but modify the {class}`TableCollection` in place.
   TreeSequence.delete_sites
   TreeSequence.trim
   TreeSequence.split_edges
+  TreeSequence.decapitate
 ```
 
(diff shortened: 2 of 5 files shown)


jeromekelleher:
Will update once #2302 iis merged.

TODO: update changelog.

python/tskit/tables.py
```diff
@@ -3862,6 +3862,37 @@ def trim(self, record_provenance=True):
...
+        """
+        Deletes edge, mutation and migration information at least as old as
+        the specified time.
+
+        .. seealso:: This method is similar to the higher-level
+            :meth:`TreeSequence.decapitate` method, which also splits
+            edges that intersect with the given time.
+            :meth:`TreeSequence.decapitate`
+            is more useful for most purposes, and may be what
+            you need instead of this method!
```
petrelharp, line 3875: I don't think you need this note.

petrelharp: never mind, I think this is good. =)

petrelharp (approved):
Whew! Looks good!

jeromekelleher:
Ready for final review @benjeffery, if you'd like to take a look.

benjeffery (approved):
LGTM. Really nice testing here. I tried for a while to think of a case that wasn't covered and failed!

As a thought about the flags/population/metadata of new nodes, how come population is the only one that can inherit from the retained node? Sorry if you've explained elsewhere, but I don't see why that wouldn't be useful for flags and metadata. I also wonder why it is the child information that gets used, would it not be the deleted parent?

jeromekelleher:
> As a thought about the flags/population/metadata of new nodes, how come population is the only one that can inherit from the retained node? Sorry if you've explained elsewhere, but I don't see why that wouldn't be useful for flags and metadata. I also wonder why it is the child information that gets used, would it not be the deleted parent?

Good questions! Population is different here I think because, unless there is other information in the model, we'd assume that any genome (node) that exists along the edge will be in the same population as the child. Flags and metadata are different in that we're probably not making these kinds of assumptions about processes.

Why it inherits from the child rather than the parent is because of the generally half-openness of intervals.

The more I try to explain this, the weaker the reasoning is though...

Maybe we should also default the population to -1, and file an issue to track adding functionality later to allow copying from the child or parent, like maybe something like:

```python
def decapitate, time, *, flags=None, population=None, metadata=None, copy_parent=False, copy_child=False):
    """
    If flags, population or metadata are not specified and ``copy_parent`` is True, copy the value from the edge's 
    parent node (likewise for copy_child). If values are explicitly provided for any of the node fields, the field 
    in question will **not** be copied from the child/parent.
    """
```

@petrelharp, any thoughts?

petrelharp:
I think you're right - defaulting to `-1` is the right way to go. I don't see a strong use case for `copy_parent` now, either.

jeromekelleher:
OK, opened a new issue to track this before 0.5 (#2334) as I can't bear updating this PR any more!

Issue #2252, opened 2022-05-10
B2 Balance index (Python)


jeromekelleher:
Add a definition of the B2 balance index following the pattern established for Sackin index in #2246. Requires #2249

See #2245 for information.

cc @jeremyguez

Here's a starting point for an efficient version:
```python
def B2_by_traversal(tree, base):
    # Note that this will take into account the number of roots also, by considering
    # them as children of the virtual root.
    stack = [(tree.virtual_root, 1)]
    total_proba = 0
    while len(stack) > 0:
        u, path_product = stack.pop()
        if tree.is_leaf(u):
            total_proba -= path_product * math.log(path_product, base)
        else:
            path_product *= 1 / tree.num_children(u)
            for v in tree.children(u):
                stack.append((v, path_product))
    return total_proba
```

We should think about the semantics with multiple roots carefully, though.

jeromekelleher:
I like it!

You can easily update the nodes_to_root function to return the path to *a* root rather than *the* root.

jeremyguez:
This definition seems to be matching the efficient version you wrote (two changes in the nodes_to_root function):

```python
def nodes_to_root(tree, u):
    nodes = []
    while u not in tree.roots:
        u = tree.parent(u)
        nodes.append(u)
    nodes.append(tree.virtual_root)
    return(nodes)

def b2_index_definition (tree, base):
    proba = [np.prod([1/tree.num_children(u)
             for u in nodes_to_root(tree, l)])
             for l in tree.leaves()]
    return(-np.sum([p * math.log(p,base) for p in proba]))
```

However, defining B2 in such a way, seems like rooting the multiroot trees before computing the index (we thus have the same index before and after removing the root of a rooted tree for example). 

```python
tree = tskit.Tree.generate_comb(5)
tree.clear()
```

In that case, I was expecting an index of 0, but got 0.7. I think this is related to this virtual root taken into account.

By looking at the definition [here](https://treebalance.wordpress.com/b%e2%82%82-index/), B2 could be undefined for multiroot trees, but it is not explicit. I contacted the authors of the website for their opinion about this. What do you think @jeromekelleher?

jeromekelleher:
Sticking this into the 0.5 release for now, until @jeremyguez let's us know whether we should try to push it in or not.

PR #2350, opened 2022-06-19
Adding path function and tests

jeremyguez:
This PR is for adding the `path` function which outputs the path between nodes u and v. It is needed for the B2 index https://github.com/tskit-dev/tskit/pull/2327 (issue https://github.com/tskit-dev/tskit/pull/2252).

By the way, the `path_length` function (https://github.com/tskit-dev/tskit/issues/2249 https://github.com/tskit-dev/tskit/pull/2259) is maybe a little too constrained: if the tree has more than one root, the outputs is always `inf`. However, in non-rooted trees the paths between nodes that have the same root appear to be defined. For the moment, I defined the `path` function with this same constrain as in the `path_length` function, so both functions remain consistent.

cc @jeromekelleher

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -948,6 +948,73 @@ class TestPathLength:
         assert math.isinf(tree.path_length(0, 1))
 
 
+class TestPath:
+    t = tskit.Tree.generate_balanced(9)
+    #         16
+    #    ┏━━━━┻━━━┓
+    #    ┃       15
+    #    ┃     ┏━━┻━┓
+    #   11     ┃   14
+    #  ┏━┻━┓   ┃  ┏━┻┓
+    #  9  10  12  ┃ 13
+    # ┏┻┓ ┏┻┓ ┏┻┓ ┃ ┏┻┓
+    # 0 1 2 3 4 5 6 7 8
+
+    def test_tmrca_leaf(self):
+        assert self.t.path(0, 16) == [9, 11, 16]
+        assert self.t.path(16, 0) == [11, 9, 0]
+        assert self.t.path(7, 16) == [13, 14, 15, 16]
+
+    def test_equal_len_path(self):
+        assert len(self.t.path(5, 16)) == self.t.path_length(5, 16)
+
+    def test_two_leaves(self):
+        assert self.t.path(0, 8) == [9, 11, 16, 15, 14, 13, 8]
+
+    def test_one_node(self):
+        assert self.t.path(0) == self.t.path(0, 16)
+
+    def test_root(self):
+        assert self.t.path(16) == []
+
+    @pytest.mark.parametrize("args", [[], [1, 2, 3]])
+    def test_bad_num_args(self, args):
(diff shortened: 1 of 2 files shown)


python/tests/test_highlevel.py
```diff
@@ -948,6 +948,73 @@ def test_no_mrca(self):
...
+    def test_tmrca_leaf(self):
+        assert self.t.path(0, 16) == [9, 11, 16]
+        assert self.t.path(16, 0) == [11, 9, 0]
+        assert self.t.path(7, 16) == [13, 14, 15, 16]
+
+    def test_equal_len_path(self):
+        assert len(self.t.path(5, 16)) == self.t.path_length(5, 16)
+
+    def test_two_leaves(self):
+        assert self.t.path(0, 8) == [9, 11, 16, 15, 14, 13, 8]
```
jeromekelleher, line 972: It's not obvious why the second argument is included in the list here and not the first - why is ``8`` on the tail but 0 not on the front? It seems natural that ``t.path(u, v)``  should be equal to ``reversed(t.path(v, u))``.

I guess this is because of the requirement that ``len(t.path(u, v))`` == ``t.path_length(u, v)``? I guess this is because path_length counts the edges whereas this ``path`` function counts the nodes?

jeromekelleher (changes requested):
Hmm, the semantics of this are a bit tricky in retrospect and I wonder if we should reconsider getting it in for the upcoming release. It's not strictly necessary for the B2 function, I just thought it would be something useful to get in while we're thinking about this stuff.

jeromekelleher:
> By the way, the path_length function (#2249 #2259) is maybe a little too constrained: if the tree has more than one root, the outputs is always inf

I don't think this is true - if you ask for the length of a path that's within a given rooted subtree it'll work fine. We only check to see if the mrca is -1, and don't actually ask anything about multirooted trees in the implementation.

python/tests/test_highlevel.py
jeremyguez: Yes, I tried to fill the requirement, but I was not sure how. So I though about how we want to use it for B2, where we want the list to include the last node but not the first one, because the probability computation needs to use the ancestry nodes, like here:

```python
path = []
u = tree.parent(u)
while u != tskit.NULL:
      path.append(u)
      u = tree.parent(u)
return path
```

But indeed for a more general purpose the best would be to have none of the first and last nodes, or both of them.

jeremyguez:
> > By the way, the path_length function (#2249 #2259) is maybe a little too constrained: if the tree has more than one root, the outputs is always inf
> 
> I don't think this is true - if you ask for the length of a path that's within a given rooted subtree it'll work fine. We only check to see if the mrca is -1, and don't actually ask anything about multirooted trees in the implementation.

Oh yes right, sorry I got mixed up! Thanks for your answer @jeromekelleher.

benjeffery:
I'm assuming this isn't needed as https://github.com/tskit-dev/tskit/issues/2252 is complete.

PR #2351, opened 2022-06-19
Don't print empty byte strings for no metadata

### LINKED ISSUE #2349: Tabulated display for metadata could show nothing rather than b'' when a table has no metadata
As mentioned on slack, on a table with no metadata and a blank schema, currently something like the following 

```
print(ts.tables.nodes)
# or
ts._repr_html_()
```

will return a tabular view with the "metadata" column filled full of `b''`. I wonder if in this specific case (no metadata in the entire table, no metadata schema) we should simply not display anything at all in that column? This would be less confusing for newcomers, and not be mistaken for anything else, I think.

If there is a metadata schema set (e.g. JSON), then I guess we would do as we currently do and show `{}` for all metadata entries in a blank table.

hyanwong:
Fixes #2349 but no tests yet: I want to see if this is the right approach: i.e. should an empty metadata schema be treated as False when viewed as a bool? Thoughts @benjeffery ?

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -85,6 +85,9 @@
 - Arguments after ``ploidy`` in ``write_vcf`` marked as keyword only
   (:user:`jeromekelleher`, :pr:`2329`, :issue:`2315`).
 
+- When metadata equal to ``b''`` is printed to text or HTML tables it will render as
+  an empty string rather than ``"b''"``. (:user:`hyanwong`, :issue:`2349`, :pr:`2351`)
+
 ----------------------
 [0.4.1] - 2022-01-11
 ----------------------
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -635,9 +635,13 @@ class CommonTestsMixin:
         input_data = self.make_input_data(41)
         table = self.table_class()
         table.set_columns(**input_data)
+        blank_meta_row = 39
+        if "metadata" in input_data:
+            table[blank_meta_row] = table[blank_meta_row].replace(metadata=b"")
         assert "1 rows skipped" in str(table)
         tskit.set_print_options(max_lines=None)
         assert "1 rows skipped" not in str(table)
+        assert "b''" not in str(table)
         tskit.set_print_options(max_lines=40)
         tskit.MAX_LINES = 40
 
--- a/python/tests/test_util.py
+++ b/python/tests/test_util.py
@@ -410,9 +410,17 @@ def test_obj_to_collapsed_html(obj, expected):
 
 
 def test_truncate_string_end():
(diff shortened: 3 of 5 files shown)


jeromekelleher:
I'm not sure this is the best way to do things anyway - isn't the point that printing out the "b'...'" is unhelpful, and we should just strip the "b" and quotes out of the rendered metadata? I don't see any advantage to keeping them in.

So, we define a function
```def render_metadata(md):
       s = str(md)
       if isinstance(md, bytes):
            # Strip leading b' and trailing '
            s = s[2:-1]
       return s
```
and use this for all the table output functions?

hyanwong:
> I'm not sure this is the best way to do things anyway - isn't the point that printing out the "b'...'" is unhelpful, and we should just strip the "b" and quotes out of the rendered metadata? I don't see any advantage to keeping them in.

Happy to do this: I was taking a more conservative approach and assuming that we would want to distinguish e.g. metadata of `b'{"a": 1}'` from `json.dumps({"a": 1})`. I think removing the `b'` and `'` would make these two indistinguishable, right?

jeromekelleher:
:shrug: We probably don't expect there to be much metadata-with-no-schema out there, so maybe having the ugly ``b'stuiff'`` is useful.

But definitely lets not get into what the boolean interpretation of a schema should be.

benjeffery:
I'd like to keep this _really_ simple and only replace `b''` with `""` for display purposes and keep everything else as-is. This fixes the most common case of ugliness without causing confusion.

hyanwong:
> I'd like to keep this _really_ simple and only replace `b''` with `""` for display purposes and keep everything else as-is. This fixes the most common case of ugliness without causing confusion.

I.e. 

```
"" if isinstance(row.metadata, bytes) and len(row.metadata) == 0 else util.truncate_string_end(str(row.metadata))
```

?

jeromekelleher:
How about we use a function, like [this](https://github.com/tskit-dev/tskit/pull/2351#issuecomment-1160118434)? Then we avoid distributing this logic around many different places.

benjeffery:
> How about we use a function, like [this](https://github.com/tskit-dev/tskit/pull/2351#issuecomment-1160118434)? Then we avoid distributing this logic around many different places.

Happy with the function, but I think it should only check for `b''` as it needs to be clear when metadata is bytes.

hyanwong:
Done. I stuck a line in test_str to put a blank metadata value in there, but perhaps that's excessive as the function is tested in `test_utils.py`.

hyanwong:
Current test failure is a codecov fault, I think, nothing to do with this PR.

benjeffery (approved):
I added a changelog, as technically this is a breaking change.

Issue #1145, opened 2021-01-27
Display coordinates without decimal padding if discrete_genome


jeromekelleher:
If all the positions in a site table (or left/right coords in an edge table) are integers, then we should display them without the padding decimals, like here:

![Screenshot from 2021-01-27 10-16-24](https://user-images.githubusercontent.com/2664569/105977154-f197ac80-6088-11eb-91ae-e15847f7893f.png)

This was the original motivation for suggesting #1144, but I've just realised that we won't have access to this property when doing the display stuff, table by table.

So, I think the simplest thing to do here is to add a private helper method to the various tables called ``_has_discrete_genome()`` which does the check using numpy ops. For display purposes, this will be plenty fast enough.

jeromekelleher:
Also we should reduce the precision we're printing out at, it's currently excessive. Something equivalent to "%.4g" is usually good for humans to process IMO.

hyanwong:
In the drawing code we do this [here](https://github.com/tskit-dev/tskit/blob/eedbb849e6b2df4f80beca06dfe3c67abd86015e/python/tskit/drawing.py#L343). Just a note that it would be good to change this use a common method, if available.

PR #2352, opened 2022-06-19
Print to 0 dp if genome pos arrays are integers

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/1145 - where should the tests go?

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -681,6 +681,13 @@ class BaseTable:
             </div>
         """
 
+    def _columns_all_integer(self, *colnames):
+        """For displaying floating point values without loads of decimal places"""
+        return all(
+            np.all(getattr(self, col) == np.floor(getattr(self, col)))
+            for col in colnames
+        )
+
 
 class MetadataColumnMixin:
     """
@@ -1099,19 +1106,21 @@ class NodeTable(BaseTable, MetadataColumnMixin):
                 [-1],
                 range(self.num_rows - (limit - (limit // 2)), self.num_rows),
             )
+        decimal_places_times = 0 if self._columns_all_integer("time") else 8
         for j in indexes:
             row = self[j]
             if j == -1:
                 rows.append(f"__skipped__{self.num_rows-limit}")
             else:
                 rows.append(
-                    "{}\t{}\t{}\t{}\t{:.8f}\t{}".format(
+                    "{}\t{}\t{}\t{}\t{:.{dp}f}\t{}".format(
                         j,
                         row.flags,
                         row.population,
                         row.individual,
                         row.time,
                         util.render_metadata(row.metadata),
+                        dp=decimal_places_times,
                     ).split("\t")
                 )
         return headers, rows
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM - where are we testing this text display stuff @benjeffery?

benjeffery:
We don't test the exact content of the text output, there is only the very simple `test_str` and `test_repr_html`.
You could make `test_str` more meaningful if you like?

hyanwong:
I'll add something. For the DTWF I wonder if we should also print time to 0 dp if all times are integers?

hyanwong:
New test add, which is a bit fiddly, since (a) we need to extract the str representation from the table string, which I hack by splitting on "│", and (b) tables with `left` and `right` only show integers when both left and right columns are entirely integer. But it works!

python/tskit/tables.py
```diff
@@ -1088,6 +1088,9 @@ def __init__(self, max_rows_increment=0, ll_table=None):
             ll_table = _tskit.NodeTable(max_rows_increment=max_rows_increment)
         super().__init__(ll_table, NodeTableRow)
 
+    def _has_discrete_times(self):
```
benjeffery, line 1091: This is going to sound picky, but I'm not a fan of one-use functions like this. If we could see ourselves using this a lot then I think it's worth it, but I don't see us reusing these so it isn't worth breaking up the flow and adding to the large list of methods on a Table.

python/tskit/tables.py
```diff
@@ -1717,6 +1741,9 @@ def __init__(self, max_rows_increment=0, ll_table=None):
             ll_table = _tskit.SiteTable(max_rows_increment=max_rows_increment)
         super().__init__(ll_table, SiteTableRow)
 
+    def _has_discrete_coords(self):
+        return np.all(self.position) == np.floor(self.position)
```
benjeffery, line 1745: No coverage here as "position" needs to be in the tests.

benjeffery (approved):
Great, thanks @hyanwong. One nit about adding one-use functions.

python/tskit/tables.py
hyanwong: Oh, of course. Doh! sorry.

python/tskit/tables.py
hyanwong: I'm happy to go with whatever you suggest. I'm just following what @jeromekelleher suggested (that's me passing the buck)

benjeffery: Sure, that would be good.

hyanwong: Done. Much tidier now, good call.

python/tskit/tables.py
```diff
@@ -681,6 +681,13 @@ def _repr_html_(self):
             </div>
         """
 
+    def _columns_all_integer(self, *colnames):
+        """For displaying floating point values without loads of decimal places"""
```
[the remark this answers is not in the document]
hyanwong: Ah, nice catch. I didn't realise the distinction.

Might be worth putting this in the tskit dev style guide? If it's not there already.

hyanwong:
> One last nit!

At times I've wished we were at that stage with the kid's hair.

PR #2377, opened 2022-07-01
Add a class for svg strings

hyanwong:
As discussed in https://github.com/tskit-dev/tskit/discussions/2376


hyanwong:
The functionality is not actually documented in the `ts.draw_svg()` and `tree.draw_svg()` methods. Suggestions for how to document this would be good.

python/tskit/drawing.py
```diff
@@ -94,6 +94,12 @@ def linear_transform(self, y):
         return self.plot_min - (y - self.min_time) * y_scale
 
 
+class _svg_string(str):
```
benjeffery, line 97: Can we use SVGString? As a class name it has to be capitalised.

benjeffery (approved):
LGTM, one PEP8 nit.

python/tskit/drawing.py
hyanwong: Done, and a bit of documentation added.

hyanwong:
Docs failing because `SVGString` isn't a documented class (nothing in `drawing.py` is in the docs), but `draw_svg()` returns `:rtype: SVGString`. I can lie and set this back to `:rtype: str` or I can try to make the `SVGString` class documented somewhere, or I can somehow exclude `SVGString` from being checked by the docs generator. Which would you prefer @benjeffery ?

benjeffery:
As this is being returned by the public API it should probably be doc'd under https://tskit.dev/tskit/docs/stable/python-api.html#miscellaneous-classes

hyanwong:
> As this is being returned by the public API it should probably be doc'd under https://tskit.dev/tskit/docs/stable/python-api.html#miscellaneous-classes

Oh, very good call. Done.

Issue #2127, opened 2022-02-09
Add possibility to simplify but keep nodes that are partially unary + partially coalescent


hyanwong:
As discussed in https://github.com/tskit-dev/tskit/discussions/2089 and with some code providing a workaround in https://github.com/tskit-dev/what-is-an-arg-paper/issues/80.

Also note that issue https://github.com/tskit-dev/tskit/issues/1120 discusses how we might avoid a large increase in the number of parameters to simplify, when focussing on particular nodes. I'm not sure if that is relevant to the current case: the semantics of the parameters passed to simplify requires some thought, I reckon.

hyanwong:
Here is a plot to show the effect of running that versus the normal simplify. It looks like I was wrong when chatting to @petrelharp : on standard tree sequences we nearly always get compression when leaving the "unary-when-on-coalescent_nodes" on. Unless I've made some error here:

```
import msprime
import numpy as np
import tskit
import matplotlib.pyplot as plt


def simplify_keeping_unary_in_coal(ts, map_nodes=False):
    """
    Keep the unary regions of nodes that are coalescent at least someone in the tree seq
    Temporary hack until https://github.com/tskit-dev/tskit/issues/2127 is addressed
    """
    tables = ts.dump_tables()
    # remove existing individuals. We will reinstate them later
    tables.individuals.clear()
    tables.nodes.individual = np.full_like(tables.nodes.individual, tskit.NULL)

    _, node_map = ts.simplify(map_nodes=True)
    keep_nodes = np.where(node_map != tskit.NULL)[0]
    # Add an individual for each coalescent node, so we can run
    # simplify(keep_unary_in_individuals=True) to leave the unary portions in.
    for u in keep_nodes:
        i = tables.individuals.add_row()
        tables.nodes[u] = tables.nodes[u].replace(individual=i)
    node_map = tables.simplify(keep_unary_in_individuals=True)

    # Reinstate individuals
    tables.individuals.clear()
    for i in ts.individuals():
        tables.individuals.append(i)
    val, inverted_map = np.unique(node_map, return_index=True)
    inverted_map = inverted_map[val != tskit.NULL]
    tables.nodes.individual = ts.tables.nodes.individual[inverted_map]
    if map_nodes:
        return tables.tree_sequence(), node_map
    else:
        return tables.tree_sequence()
    
    
sample_size = [10, 100, 1000, 10000]

    
num_reps = 100

byte_diff = np.zeros((num_reps, len(sample_size)))
[...]

PR #2381, opened 2022-07-04
Keep unary for coalescent nodes in simplify

jeromekelleher:
See #2127. See also #1190 for when keep_unary_in_individuals was added.

This is a quick prototype to show how it can be done. Essentially we just need to make a second pass through the overlapping segments to see whether a node participates in any coalescences. The rest is refactoring.

Ideally we would like to use an enumeration like ``keep_unary="all" | "individuals" | "coalescent"``, but we need to be a bit careful about how this might interact with people using "truthy" inputs to ``keep_unary`` at the moment.

Here's an example:

```
When keep_unary = True:                                                                                               
3.86┊             24      ┊            24       ┊                     ┊ 
    ┊         ┏━━━━┻━━━━┓ ┊        ┏━━━━┻━━━━┓  ┊                     ┊ 
2.40┊        23         ┃ ┊       23         ┃  ┊            23       ┊ 
    ┊         ┃         ┃ ┊        ┃         ┃  ┊        ┏━━━━┻━━━━┓  ┊ 
1.32┊         ┃        21 ┊        ┃        21  ┊        ┃        22  ┊ 
    ┊         ┃         ┃ ┊        ┃         ┃  ┊        ┃         ┃  ┊ 
0.48┊        20         ┃ ┊       20         ┃  ┊       20         ┃  ┊ 
    ┊     ┏━━━┻━━━┓     ┃ ┊     ┏━━┻━━┓      ┃  ┊     ┏━━┻━━┓      ┃  ┊ 
0.38┊     ┃       ┃    19 ┊     ┃     ┃     19  ┊     ┃     ┃     19  ┊ 
    ┊     ┃       ┃     ┃ ┊     ┃     ┃     ┏┻┓ ┊     ┃     ┃     ┏┻┓ ┊ 
0.20┊    18       ┃     ┃ ┊    18     ┃     ┃ ┃ ┊    18     ┃     ┃ ┃ ┊ 
    ┊   ┏━┻━━┓    ┃     ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊ 
0.20┊  17    ┃    ┃     ┃ ┊  17   ┃   ┃     ┃ ┃ ┊  17   ┃   ┃     ┃ ┃ ┊ 
    ┊  ┏┻━┓  ┃    ┃     ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊ 
0.17┊  ┃  ┃ 16    ┃     ┃ ┊  ┃  ┃16   ┃     ┃ ┃ ┊  ┃  ┃16   ┃     ┃ ┃ ┊ 
    ┊  ┃  ┃ ┏┻┓   ┃     ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊ 
0.11┊ 15  ┃ ┃ ┃   ┃     ┃ ┊ 15  ┃ ┃   ┃     ┃ ┃ ┊ 15  ┃ ┃   ┃     ┃ ┃ ┊ 
    ┊ ┏┻┓ ┃ ┃ ┃   ┃     ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊ 
0.08┊ ┃ ┃ ┃ ┃ ┃  14     ┃ ┊ ┃ ┃ ┃ ┃  14     ┃ ┃ ┊ ┃ ┃ ┃ ┃  14     ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊ 
0.07┊ ┃ ┃ ┃ ┃ ┃ ┃  13   ┃ ┊ ┃ ┃ ┃ ┃ ┃  13   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  13   ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊ 
0.06┊ ┃ ┃ ┃ ┃ ┃ ┃ 12  ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 12  ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 12  ┃ ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊ 
0.06┊ ┃ ┃ ┃10 ┃ ┃ ┃ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃11 ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃11 ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┃ ┊ 
0.00┊ 0 4 1 2 7 5 6 9 8 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊ 
    0                     2                     4                    10 
                                                                                                                      
                                                                                                                      
When keep_unary_if_coalescent = True                                                                                  
3.86┊             20      ┊            20       ┊                     ┊ 
    ┊         ┏━━━━┻━━━━┓ ┊        ┏━━━━┻━━━━┓  ┊                     ┊ 
2.40┊        19         ┃ ┊       19         ┃  ┊            19       ┊ 
    ┊         ┃         ┃ ┊        ┃         ┃  ┊        ┏━━━━┻━━━━┓  ┊ 
0.48┊        18         ┃ ┊       18         ┃  ┊       18         ┃  ┊ 
    ┊     ┏━━━┻━━━┓     ┃ ┊     ┏━━┻━━┓      ┃  ┊     ┏━━┻━━┓      ┃  ┊ 
0.38┊     ┃       ┃    17 ┊     ┃     ┃     17  ┊     ┃     ┃     17  ┊ 
    ┊     ┃       ┃     ┃ ┊     ┃     ┃     ┏┻┓ ┊     ┃     ┃     ┏┻┓ ┊ 
0.20┊    16       ┃     ┃ ┊    16     ┃     ┃ ┃ ┊    16     ┃     ┃ ┃ ┊ 
    ┊   ┏━┻━━┓    ┃     ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊ 
0.20┊  15    ┃    ┃     ┃ ┊  15   ┃   ┃     ┃ ┃ ┊  15   ┃   ┃     ┃ ┃ ┊ 
    ┊  ┏┻━┓  ┃    ┃     ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊ 
0.17┊  ┃  ┃ 14    ┃     ┃ ┊  ┃  ┃14   ┃     ┃ ┃ ┊  ┃  ┃14   ┃     ┃ ┃ ┊ 
    ┊  ┃  ┃ ┏┻┓   ┃     ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊ 
0.11┊ 13  ┃ ┃ ┃   ┃     ┃ ┊ 13  ┃ ┃   ┃     ┃ ┃ ┊ 13  ┃ ┃   ┃     ┃ ┃ ┊ 
    ┊ ┏┻┓ ┃ ┃ ┃   ┃     ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊ 
0.08┊ ┃ ┃ ┃ ┃ ┃  12     ┃ ┊ ┃ ┃ ┃ ┃  12     ┃ ┃ ┊ ┃ ┃ ┃ ┃  12     ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊ 
0.07┊ ┃ ┃ ┃ ┃ ┃ ┃  11   ┃ ┊ ┃ ┃ ┃ ┃ ┃  11   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  11   ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊ 
0.06┊ ┃ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┃ ┊ 
    ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊ 
0.00┊ 0 4 1 2 7 5 6 9 8 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊ 
    0                     2                     4                    10 


When keep_unary = False                                    
3.86┊             20      ┊            20       ┊                     ┊                                                
    ┊         ┏━━━━┻━━━━┓ ┊        ┏━━━━┻━━━━┓  ┊                     ┊                                                
2.40┊         ┃         ┃ ┊        ┃         ┃  ┊            19       ┊                                                
    ┊         ┃         ┃ ┊        ┃         ┃  ┊        ┏━━━━┻━━━━┓  ┊                                                
0.48┊        18         ┃ ┊       18         ┃  ┊       18         ┃  ┊                                                
    ┊     ┏━━━┻━━━┓     ┃ ┊     ┏━━┻━━┓      ┃  ┊     ┏━━┻━━┓      ┃  ┊                                                
0.38┊     ┃       ┃     ┃ ┊     ┃     ┃     17  ┊     ┃     ┃     17  ┊                                                
    ┊     ┃       ┃     ┃ ┊     ┃     ┃     ┏┻┓ ┊     ┃     ┃     ┏┻┓ ┊                                                
0.20┊    16       ┃     ┃ ┊    16     ┃     ┃ ┃ ┊    16     ┃     ┃ ┃ ┊                                                
    ┊   ┏━┻━━┓    ┃     ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊   ┏━┻━┓   ┃     ┃ ┃ ┊                                                
0.20┊  15    ┃    ┃     ┃ ┊  15   ┃   ┃     ┃ ┃ ┊  15   ┃   ┃     ┃ ┃ ┊                                                
    ┊  ┏┻━┓  ┃    ┃     ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊  ┏┻━┓ ┃   ┃     ┃ ┃ ┊                                                
0.17┊  ┃  ┃ 14    ┃     ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊                                                
    ┊  ┃  ┃ ┏┻┓   ┃     ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊  ┃  ┃ ┃   ┃     ┃ ┃ ┊                                                
0.11┊ 13  ┃ ┃ ┃   ┃     ┃ ┊ 13  ┃ ┃   ┃     ┃ ┃ ┊ 13  ┃ ┃   ┃     ┃ ┃ ┊                                                
    ┊ ┏┻┓ ┃ ┃ ┃   ┃     ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊ ┏┻┓ ┃ ┃   ┃     ┃ ┃ ┊                                                
0.08┊ ┃ ┃ ┃ ┃ ┃  12     ┃ ┊ ┃ ┃ ┃ ┃  12     ┃ ┃ ┊ ┃ ┃ ┃ ┃  12     ┃ ┃ ┊                                                
    ┊ ┃ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┏━┻━┓   ┃ ┃ ┊                                                
0.07┊ ┃ ┃ ┃ ┃ ┃ ┃  11   ┃ ┊ ┃ ┃ ┃ ┃ ┃  11   ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  11   ┃ ┃ ┊                                                
    ┊ ┃ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃  ┏┻━┓ ┃ ┃ ┊                                                
0.06┊ ┃ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ 10  ┃ ┃ ┃ ┊                                                
    ┊ ┃ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊ ┃ ┃ ┃ ┃ ┃ ┏┻┓ ┃ ┃ ┃ ┊                                                
0.00┊ 0 4 1 2 7 5 6 9 8 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊ 0 4 1 7 5 6 9 8 2 3 ┊                                                
    0                     2                     4                    10                                                

```

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -1198,6 +1198,12 @@ test_simplest_records(void)
     CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
     tsk_treeseq_free(&simplified);
 
+    ret = tsk_treeseq_simplify(
+        &ts, sample_ids, 2, TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
+    tsk_treeseq_free(&simplified);
+
     tsk_treeseq_free(&ts);
 }
 
@@ -1376,6 +1382,48 @@ test_simplest_unary_with_individuals(void)
     tsk_treeseq_free(&simplified);
 
     tsk_treeseq_free(&expected);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_simplest_partially_unary(void)
+{
+    int ret;
+    const char *nodes = "1  0   0\n"
+                        "1  0   0\n"
+                        "0  1   0";
+    const char *edges = "0  2   2   0\n"
+                        "0  1   2   1\n";
+    tsk_treeseq_t ts, simplified;
+    tsk_id_t sample_ids[] = { 0, 1 };
+
+    tsk_treeseq_from_text(&ts, 2, nodes, edges, NULL, NULL, NULL, NULL, NULL, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_samples(&ts), 2);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_sequence_length(&ts), 2.0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_nodes(&ts), 3);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_trees(&ts), 2);
+
+    ret = tsk_treeseq_simplify(&ts, sample_ids, 2, 0, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_samples(&simplified), 2);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_sequence_length(&simplified), 2.0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_nodes(&simplified), 3);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_edges(&simplified), 2);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_num_trees(&simplified), 2);
+    tsk_treeseq_free(&simplified);
+
+    ret = tsk_treeseq_simplify(
+        &ts, sample_ids, 2, TSK_SIMPLIFY_KEEP_UNARY, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
+    tsk_treeseq_free(&simplified);
+
+    ret = tsk_treeseq_simplify(
+        &ts, sample_ids, 2, TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT, &simplified, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_table_collection_equals(ts.tables, simplified.tables, 0));
+    tsk_treeseq_free(&simplified);
+
     tsk_treeseq_free(&ts);
 }
 
@@ -7840,6 +7888,7 @@ main(int argc, char **argv)
         { "test_simplest_nonbinary_records", test_simplest_nonbinary_records },
         { "test_simplest_unary_records", test_simplest_unary_records },
         { "test_simplest_unary_with_individuals", test_simplest_unary_with_individuals },
+        { "test_simplest_partially_unary", test_simplest_partially_unary },
         { "test_simplest_non_sample_leaf_records",
             test_simplest_non_sample_leaf_records },
         { "test_simplest_degenerate_multiple_root_records",
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -8677,6 +8677,8 @@ simplifier_print_state(simplifier_t *self, FILE *out)
         !!(self->options & TSK_SIMPLIFY_KEEP_INPUT_ROOTS));
     fprintf(out, "\tkeep_unary_in_individuals : %d\n",
         !!(self->options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS));
+    fprintf(out, "\tkeep_unary_if_coalescent: %d\n",
+        !!(self->options & TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT));
 
     fprintf(out, "===\nInput tables\n==\n");
     tsk_table_collection_print_state(&self->input_tables, out);
@@ -9224,16 +9226,8 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
     double left, right, prev_right;
     tsk_id_t ancestry_node;
     tsk_id_t output_id = self->node_id_map[input_id];
-
     bool is_sample = output_id != TSK_NULL;
     bool keep_unary = false;
-    if (self->options & TSK_SIMPLIFY_KEEP_UNARY) {
-        keep_unary = true;
-    }
-    if ((self->options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS)
-        && (self->input_tables.nodes.individual[input_id] != TSK_NULL)) {
-        keep_unary = true;
-    }
 
     if (is_sample) {
         /* Free up the existing ancestry mapping. */
@@ -9241,6 +9235,32 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
         tsk_bug_assert(x->left == 0 && x->right == self->tables->sequence_length);
         self->ancestor_map_head[input_id] = NULL;
         self->ancestor_map_tail[input_id] = NULL;
+    }
+
+    if (self->options & TSK_SIMPLIFY_KEEP_UNARY) {
+        keep_unary = true;
+    }
+    if ((self->options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS)
+        && (self->input_tables.nodes.individual[input_id] != TSK_NULL)) {
+        keep_unary = true;
+    }
+    if (self->options & TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT) {
+        /* Make an initial pass through the overlapping segments to see
+         * if there's any coalescence.
+         */
+        ret = segment_overlapper_start(
+            &self->segment_overlapper, self->segment_queue, self->segment_queue_size);
+        if (ret != 0) {
+            goto out;
+        }
+        while ((ret = segment_overlapper_next(
+                    &self->segment_overlapper, &left, &right, &X, &num_overlapping))
+               == 1) {
+            if (num_overlapping > 1) {
+                keep_unary = true;
+                break;
+            }
+        }
     }
 
     ret = segment_overlapper_start(
@@ -11491,10 +11511,14 @@ tsk_table_collection_simplify(tsk_table_collection_t *self, const tsk_id_t *samp
     /* Avoid calling to simplifier_free with uninit'd memory on error branches */
     tsk_memset(&simplifier, 0, sizeof(simplifier_t));
 
-    if ((options & TSK_SIMPLIFY_KEEP_UNARY)
-        && (options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS)) {
-        ret = TSK_ERR_KEEP_UNARY_MUTUALLY_EXCLUSIVE;
-        goto out;
+    if ((options & TSK_SIMPLIFY_KEEP_UNARY)) {
+        // FIXME what about specifying keep_unary_in_individuals *and*
+        // unary_if_coalescent?
+        if ((options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS)
+            || (options & TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT)) {
+            ret = TSK_ERR_KEEP_UNARY_MUTUALLY_EXCLUSIVE;
+            goto out;
+        }
     }
 
     /* For now we don't bother with edge metadata, but it can easily be
--- a/c/tskit/tables.h
+++ b/c/tskit/tables.h
@@ -715,6 +715,13 @@ flag). It keeps unary nodes, but only if the unary node is referenced from an in
 @endrst
 */
 #define TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS (1 << 6)
+/**
+@rst
+DOCUMENT ME
+@endrst
+*/
+#define TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT (1 << 7)
+
 /** @} */
 
 /**
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -6590,19 +6590,21 @@ TableCollection_simplify(TableCollection *self, PyObject *args, PyObject *kwds)
     int filter_populations = false;
     int keep_unary = false;
     int keep_unary_in_individuals = false;
+    int keep_unary_if_coalescent = false;
     int keep_input_roots = false;
     int reduce_to_site_topology = false;
-    static char *kwlist[] = { "samples", "filter_sites", "filter_populations",
-        "filter_individuals", "reduce_to_site_topology", "keep_unary",
-        "keep_unary_in_individuals", "keep_input_roots", NULL };
+    static char *kwlist[]
+        = { "samples", "filter_sites", "filter_populations", "filter_individuals",
+              "reduce_to_site_topology", "keep_unary", "keep_unary_in_individuals",
+              "keep_unary_if_coalescent", "keep_input_roots", NULL };
 
     if (TableCollection_check_state(self) != 0) {
         goto out;
     }
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|iiiiiii", kwlist, &samples,
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|iiiiiiii", kwlist, &samples,
             &filter_sites, &filter_populations, &filter_individuals,
             &reduce_to_site_topology, &keep_unary, &keep_unary_in_individuals,
-            &keep_input_roots)) {
+            &keep_unary_if_coalescent, &keep_input_roots)) {
         goto out;
     }
     samples_array = (PyArrayObject *) PyArray_FROMANY(
@@ -6629,6 +6631,9 @@ TableCollection_simplify(TableCollection *self, PyObject *args, PyObject *kwds)
     }
     if (keep_unary_in_individuals) {
         options |= TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS;
+    }
+    if (keep_unary_if_coalescent) {
+        options |= TSK_SIMPLIFY_KEEP_UNARY_IF_COALESCENT;
     }
     if (keep_input_roots) {
         options |= TSK_SIMPLIFY_KEEP_INPUT_ROOTS;
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2021 Tskit Developers
+# Copyright (c) 2018-2022 Tskit Developers
 # Copyright (c) 2015-2018 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -110,6 +110,7 @@ class Simplifier:
         filter_individuals=True,
         keep_unary=False,
         keep_unary_in_individuals=False,
+        keep_unary_if_coalescent=False,
         keep_input_roots=False,
     ):
         self.ts = ts
@@ -121,6 +122,7 @@ class Simplifier:
         self.filter_individuals = filter_individuals
         self.keep_unary = keep_unary
         self.keep_unary_in_individuals = keep_unary_in_individuals
+        self.keep_unary_if_coalescent = keep_unary_if_coalescent
         self.keep_input_roots = keep_input_roots
         self.num_mutations = ts.num_mutations
         self.input_sites = list(ts.sites())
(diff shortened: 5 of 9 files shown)


petrelharp:
If I've got it right, this is retaining all ancestry for any nodes that are coalescent nodes anywhere in the sequence. This is slightly different to what `extend_edges` does, which is to output any input *edges* on which somewhere there is a coalescence. So, we could have a node that is a coalscent node on [0, 10], and then we output an edge for which it's a parent on [0, 15], but not a different edge for which it's a parent on [50, 60]. The argument for doing the second thing but not the first is that (a) it's not clear if we have any information to infer the existence of these totally-unary edges; and (b) these totally unary edges only increase the size of the tree sequence (as opposed to the partially-coalescent ones, which reduce it).

Thoughts?

hyanwong:
Just to get this right in my head, you mean that if the node has non-contiguous edges (i.e. "trapped material") where it exists in the topology from (say) 10-15 and 50-60, but is not on any lineage, even as a unary node, from 15-50, then the current implementation keeps the 15-50 edge too?

If so, I think that's right, and is not unreasonable behaviour, if you want to keep the number of nodes unaffected, but preserve as much of the known ancestry as possible. It would be interesting to see if they are inferred by `tsinfer`. I'll check now, but I suspect there are some cases where they will be. I can see the rationale for sometimes wanting to deleting them too, though.

hyanwong:
We do have "split" edges like this in tsinfer (we also have quite a lot of nodes that are unary *everywhere* too. Here's an example:

```python
import collections

import numpy as np
import msprime
import tskit
import tsinfer

def simplify_keeping_unary_in_coal(ts, map_nodes=False):
    """
    Keep the unary regions of nodes that are coalescent at least someone in the tree seq
    Temporary hack until https://github.com/tskit-dev/tskit/issues/2127 is addressed
    """
    tables = ts.dump_tables()
    # remove existing individuals. We will reinstate them later
    tables.individuals.clear()
    tables.nodes.individual = np.full_like(tables.nodes.individual, tskit.NULL)

    _, node_map = ts.simplify(map_nodes=True)
    keep_nodes = np.where(node_map != tskit.NULL)[0]
    # Add an individual for each coalescent node, so we can run
    # simplify(keep_unary_in_individuals=True) to leave the unary portions in.
    for u in keep_nodes:
        i = tables.individuals.add_row()
        tables.nodes[u] = tables.nodes[u].replace(individual=i)
    node_map = tables.simplify(keep_unary_in_individuals=True)

    # Reinstate individuals
    tables.individuals.clear()
    for i in ts.individuals():
        tables.individuals.append(i)
    val, inverted_map = np.unique(node_map, return_index=True)
    inverted_map = inverted_map[val != tskit.NULL]
    tables.nodes.individual = ts.tables.nodes.individual[inverted_map]
    if map_nodes:
        return tables.tree_sequence(), node_map
    else:
        return tables.tree_sequence()


num_samples = 8
ts  = msprime.sim_ancestry(
    num_samples,
    ploidy=1,
    sequence_length=10e8,
    recombination_rate=1e-8,
    record_full_arg=True,
    random_seed=222
)

ts = simplify_keeping_unary_in_coal(ts)
tsA = msprime.sim_mutations(ts, rate=1e-8, random_seed=123)
print(tsA.num_sites, "sites")

tsB = tsinfer.infer(tsinfer.SampleData.from_tree_sequence(tsA))
print(tsB.num_trees, "inferred trees")

def node_spans_max_children(ts):
    node_spans = collections.defaultdict(list)
    # node_id => [(left, right, [n_children1, n_children2,...]), ()]
    curr_parents = collections.defaultdict(set)

    for tree, diffs in zip(ts.trees(), ts.edge_diffs()):
        for e_in in diffs.edges_in:
            u = e_in.parent
            if len(curr_parents[u]) == 0:
                # node starts
                node_spans[u].append(
                    [diffs.interval.left, diffs.interval.right, tree.num_children(u)]
                )
            else:
                node_spans[u][-1][2] = max(node_spans[u][-1][2], tree.num_children(u))
            curr_parents[e_in.parent].add(e_in.id)
        for e_out in diffs.edges_out:
            u = e_out.parent
            curr_parents[u].remove(e_out.id)
            if len(curr_parents[u]) == 0:
                # node ends
                node_spans[u][-1][1] = diffs.interval.right


    for u, data in node_spans.items():
        max_children = max(contiguous[2] for contiguous in data)
        for contiguous in data:
            if max_children > 1 and contiguous[2] < 2:
                print("Node", u, "is contiguously unary from", contiguous)

node_spans_max_children(tsB)
```

```
44 sites
12 inferred trees
Node 20 is contiguously unary from [492084773.0, 533368834.0, 1]
```

```
tsB.draw_svg(time_scale="rank", style=".n20 > .sym {fill: red} .mut {display: none}")
```

unknown

petrelharp:
Ah, great, nice example. I do agree that sometimes it'd be conceptually nice to have those nodes in their entireity, but the two operations *are* in principle different, and we should figure out whether we want one or both of them. If the unconnected, unary-only segments of a coalescent nodes are also inferrable, then we should also include them! But, I am not sure that they are? Do you have a good reason that they might be?

So, one question is: are these unary-only spans real or artifactual? Like, when tsinfer infers a unary-only bit, how often is that unary-only bit "correct"? In the case you've got here, the extra unary bit does not seem to be correct, at least it isn't a parent to the same samples. On my machine (tsinfer 0.2.3) the node in question is 14, not 16, but:
```
t = tsA.first()
orig_node = t.mrca(1, 3, 7)
t = tsB.first()
new_node = t.mrca(1, 3, 7)
assert new_node == 14

left = 0
sA = set()
sB = set()
for interval, tA, tB in tsA.coiterate(tsB):
    new_sA = set(tA.samples(orig_node))
    new_sB = set(tB.samples(new_node))
    if sA != new_sA or sB != new_sB:
        print(f"{left} to {interval[1]} the node is a parent to:")
        print(f"         truth: {sA}")
        print(f"      inferred: {sB}")
    left = interval[1]
    sA = new_sA
    sB = new_sB
```
gives
```
0 to 18028285.0 the node is a parent to:
         truth: set()
      inferred: set()
160570091.0 to 169968287.0 the node is a parent to:
         truth: {1, 3, 7}
      inferred: {1, 3, 7}
173906518.0 to 178484736.0 the node is a parent to:
         truth: {3, 7}
      inferred: {1, 3, 7}
178484736.0 to 204789241.0 the node is a parent to:
         truth: {3, 7}
      inferred: {1, 3, 5, 6, 7}
204789241.0 to 254012804.0 the node is a parent to:
         truth: {3, 7}
      inferred: {3, 5, 7}
320784505.0 to 323579641.0 the node is a parent to:
         truth: {3, 7}
      inferred: set()
488755800.0 to 493484566.0 the node is a parent to:
         truth: {3, 5, 7}
      inferred: set()
548658128.0 to 564756120.0 the node is a parent to:
         truth: {0, 3, 5, 7}
      inferred: set()
576356796.0 to 577972638.0 the node is a parent to:
         truth: {0, 3, 5, 7}
      inferred: {3}
577972638.0 to 588165874.0 the node is a parent to:
         truth: {0, 5, 7}
      inferred: {3}
884623975.0 to 924227883.0 the node is a parent to:
         truth: {0, 5, 7}
      inferred: {3, 6}
983144527.0 to 993319578.0 the node is a parent to:
         truth: {0, 5, 7}
      inferred: set()
```

Have you got any situations where there's clearly information to infer those unary-only bits?

hyanwong:
Moving this to https://github.com/petrelharp/num_edges/issues/2 so the discussion can continue easily even when the PR is closed.

jeromekelleher:
I think what simplify is doing here is a reasonable operation, because it's not discarding any information. Given a full ARG, it keeps a full record of any topology for ancestors that are coalescent anywhere. This is a useful operation, and I think corresponds to what msprime may optionally output some day. 

Implementing the exact mirror of edge_extend in simplify would be tricky I think, and much more complicated than this because you'd have to do a lot of reasoning about what's adjacent to what, and exclude edges depending on what they are adjacent to. I'm not enthusiastic about trying to do this - partly because it would be hard, but mostly because I don't see a lot of point (other than to provide the mirror-image of ``edge_extend``). In practise the increase in size caused by extra unary-only edges will be negligible, I think, and not worth the substantially more complicated simplify algorithm (from say, the forward simulation perspective).

I guess we could do some experiments easily enough to see how significant the difference is by counting the number of unary-only edges output by this for a range of input parameters?

jeromekelleher:
I've always seen ``edge_extend`` as being a method to infer a subset of the true ancestry (i.e., what's returned by this method), which is why I said it would be exact in the SMC case and an approximation for things like full Hudson and gene conversion. I would then put it the other way around - we can infer a subset (~the vast majority?) of the true unary nodes using these properties you're exploiting.

I think it's a pretty strong assertion to say that the current ``edge_extend`` algorithm is the best possible and no extra information about unary nodes could ever be derived from the trees.

hyanwong:
This LGTM @jeromekelleher: it's nice it is relatively simple. I agree that a separate parameter is the way to go, in case people want to keep both unary in individuals and when coalescent (I don't think we test for combinations of the `keep_unary_XXX` params, do we?)

The only comment I would have is to think about the name of the parameter. To someone who hasn't thought about it a lot, `keep unary if coalescent` could possibly seem like a contradiction in terms. It's more like `keep_unary_if_coalescent_elsewhere`, but that's too long. We could have `keep_partially_unary`, although it's nice to have all the params starting with `keep_unary_XXXXX`, which I presume is why we hit upon the current naming.

hyanwong:
> The only other comment I would have is to think about the name of the parameter. 

Another possible name: `keep_unary_regions` (to emphasise that it's not that we are keeping unary *nodes* per se, but the unary regions of nodes that we would have kept anyway.

hyanwong:
Here's a possible docstring. It's a bit contorted, but the best I could do after a fair bit of rewording.

> If True, keep unary *regions* of retained nodes. After simplification, unary nodes may continue to exist in local trees, but only if those nodes would have been retained by the simplification process anyway (e.g. because they represent coalescences between two or more lineages from sample nodes to the root). If ``False``, these nodes will still be present in the tree sequence, but are subject to removal from lineages where they are not coalescent points.

petrelharp:
> I think it's a pretty strong assertion to say that the current edge_extend algorithm is the best possible and no extra information about unary nodes could ever be derived from the trees.

Totally agree - I wasn't trying to say that it was!

How about

> If True, keep all regions over which retained nodes are ancestral to samples. By default, nodes not in samples are only retained on regions of the genome where they are coalescent (i.e., a MRCA to at least two nodes in samples); if this option is True then nodes not in samples are still only retained if they are a MRCA to at least two nodes in samples, but all regions of the genome on which they are ancestral to a node in sample is also retained. Any extra retained information will show up as unary nodes on local trees. Default: False

petrelharp:
OK - I did some quick experiments. This `keep_unary` argument seems to usually increase the number of edges by about as much as `extend_edges` decreases it. It's very interesting how much of this stuff there is!
Screenshot from 2022-07-21 21-39-00

So - this operation might still be useful, but I was wanting to include it in SLiM, to produce smaller, faster files. It looks like this operation won't do that. I'm not sure about the bigger picture at the moment.

jeromekelleher:
Huh, isn't that interesting... I wonder if that's still true if you have a larger sample size? 6 is small enough that I can imagine odd things happening.

hyanwong:
I thought I had some code and plots that showed that the hacked version generally reduced the number of edges. It could have been with a bigger sample size: I can't remember.

petrelharp:
Hah, you're right about the number of samples:
Screenshot from 2022-07-22 08-25-25

... why would this happen? (i.e., why would more samples make the proportion of edges that are isolated unary edges on coalescent nodes smaller?)

jeromekelleher:
> ... why would this happen? (i.e., why would more samples make the proportion of edges that are isolated unary edges on coalescent nodes smaller?)

Not obvious to me...

petrelharp:
I wonder how this will change with a longer genome, though...

petrelharp:
Ack, here's results for a longer sequence (0.1M instead of 0.01M):
Screenshot from 2022-07-22 13-13-30

petrelharp:
Here's a bit more of an experiment changing both sampels and sequence length; recomb rate is 1e-8.
Screenshot from 2022-07-22 23-17-34

jeromekelleher:
Very interesting! I really didn't see this coming - the plot makes a clear practical case for edge_extend.

I guess we do need to implement it in simplify to be truly useful for forward sims. I think the present operation is useful also, but we'd need some new names I think.

Maybe we should move away from the phrase ``keep_unary`` entirely? I'm finding it less helpful every time I look at this stuff.

PR #2384, opened 2022-07-05
Add __str__ and _repr_html_ to Variant

### LINKED ISSUE #2367: Display methods for Variant
When we print out a Variant we get ``<tskit.genotypes.Variant object at 0x7f58b0483430>``

We should add a str and repr_html to fix this.

Did this change in 0.5.0? Was Variant a dataclass before, and therefore inherited display methods?

szhan:
Fixes #2367

### DIFF
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -25,7 +25,9 @@ Test cases for generating genotypes/haplotypes.
 import itertools
 import logging
 import random
+import re
 import textwrap
+from xml.etree import ElementTree
 
 import msprime
 import numpy as np
@@ -2100,3 +2102,104 @@ class TestVariant:
             freqs = variant.frequencies()
             assert caplog.text.count("frequencies undefined") == 1
         assert np.all(np.isnan(list(freqs.values())))
+
+    def test_variant_str(self):
+        """
+        Test using a simple dummy tree sequence for testing.
+        It has only one tree and one site, whose variant has the alleles
+        ('A', 'T', 'G', '💩', '', 'TAG', None).
+        """
+        tables = tskit.TableCollection(10)
+        for _ in np.arange(6):
+            tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        tables.sites.add_row(position=5, ancestral_state="A")
+        tables.mutations.add_row(site=0, node=0, derived_state="T")
+        tables.mutations.add_row(site=0, node=1, derived_state="G")
+        tables.mutations.add_row(site=0, node=2, derived_state="💩")
+        tables.mutations.add_row(site=0, node=3, derived_state="")
+        tables.mutations.add_row(site=0, node=4, derived_state="TAG")
+        ts = tables.tree_sequence()
+        v = next(ts.variants())
+        assert v.alleles == ("A", "T", "G", "💩", "", "TAG", None)
+        assert isinstance(str(v), str)
+        assert re.match(
+            textwrap.dedent(
+                r"""
+                ╔═+╗
+                ║Variant\s*║
+                ╠═+╤═+╣
+                ║Site id\s*│\s*0║
+                ╟─+┼─+╢
+                ║Site position\s*│\s*[0-9\.]+║
+                ╟─+┼─+╢
+                ║Number of samples\s*│\s*[0-9]+║
+                ╟─+┼─+╢
+                ║Number of alleles\s*│\s*[0-9]+║
+                ╟─+┼─+╢
+                ║Samples with allele \'A\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele \'T\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele \'G\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele \'💩\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele \'\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele \'TAG\'\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Samples with allele missing\s*│\s*[0-9]+\s*\([0-9\.]+\%\)║
+                ╟─+┼─+╢
+                ║Has missing data\s*│\s*True║
+                ╟─+┼─+╢
+                ║Isolated as missing\s*│\s*True║
+                ╚═+╧═+╝
+                """[
+                    1:
+                ]
+            ),
+            str(v),
+        )
+
+    def test_variant_str_no_samples(self):
+        tables = tskit.TableCollection(10)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        tables.sites.add_row(position=5, ancestral_state="A")
+        tables.mutations.add_row(site=0, node=0, derived_state="T")
+        ts = tables.tree_sequence()
+        v = next(ts.variants(samples=[]))
+        for allele in v.alleles:
+            if allele is not None:
+                assert (
+                    re.search(
+                        rf"║Samples with allele '{allele}'\s*│\s*0\s*\(nan\%\)║", str(v)
+                    )
+                    is not None
+                )
+
+    def test_variant_str_no_site(self):
+        tables = tskit.TableCollection(10)
+        ts = tables.tree_sequence()
+        v = tskit.Variant(ts)
+        s = str(v)
+        assert len(s.splitlines()) == 5
+        assert (
+            "This variant has not yet been decoded at a specific site, "
+            + "call Variant.decode to set the site"
+            in s
+        )
+
+    def test_variant_html_repr(self, ts_fixture):
+        v = next(ts_fixture.variants())
+        html = v._repr_html_()
+        # Parse to check valid
+        ElementTree.fromstring(html)
+        assert len(html) > 1900
+
+    def test_variant_html_repr_no_site(self):
+        tables = tskit.TableCollection(10)
+        ts = tables.tree_sequence()
+        v = tskit.Variant(ts)
+        html = v._repr_html_()
+        ElementTree.fromstring(html)
+        assert len(html) > 1600
--- a/python/tskit/genotypes.py
+++ b/python/tskit/genotypes.py
@@ -32,6 +32,7 @@ import numpy as np
 import _tskit
 import tskit
 import tskit.trees as trees
+import tskit.util as util
 
 
 class Variant:
@@ -125,7 +126,7 @@ class Variant:
     def _check_decoded(self):
         if self._ll_variant.site_id == tskit.NULL:
             raise ValueError(
-                "This variant has not yet been decoded at a specific site,"
+                "This variant has not yet been decoded at a specific site, "
                 "call Variant.decode to set the site."
             )
 
@@ -293,6 +294,46 @@ class Variant:
             for allele, count in self.counts().items()
             if not (allele is None and remove_missing)
         }
+
+    def __str__(self) -> str:
+        """
+        Return a plain text summary of the contents of a variant.
+        """
+        try:
+            site_id = self.site.id
+            site_position = self.site.position
+            counts = self.counts()
+            freqs = self.frequencies()
+            rows = (
+                [
+                    ["Site id", str(site_id)],
+                    ["Site position", str(site_position)],
+                    ["Number of samples", str(len(self.samples))],
+                    ["Number of alleles", str(self.num_alleles)],
+                ]
+                + [
+                    [
+                        f"""Samples with allele {'missing' if k is None
+                            else "'" + k + "'"}""",
+                        f"{counts[k]} ({freqs[k] * 100:.2g}%)",
+                    ]
+                    for k in self.alleles
+                ]
+                + [
+                    ["Has missing data", str(self.has_missing_data)],
+                    ["Isolated as missing", str(bool(self.isolated_as_missing))],
+                ]
+            )
+        except ValueError as err:
+            rows = [[str(err), ""]]
+        return util.unicode_table(rows, title="Variant")
+
+    def _repr_html_(self) -> str:
+        """
+        Return an html summary of a variant. Called by Jupyter notebooks
+        to render a Variant.
+        """
+        return util.variant_html(self)
 
 
 #
--- a/python/tskit/util.py
+++ b/python/tskit/util.py
@@ -542,6 +542,98 @@ def tree_html(tree):
             """  # noqa: B950
 
 
+def variant_html(variant):
+    class_type = "Variant"
+
+    url_tskit_logo = (
+        "https://raw.githubusercontent.com/tskit-dev/administrative/main/tskit_logo.svg"
+    )
+    url_variant_class_doc = (
+        "https://tskit.dev/tskit/docs/latest/python-api.html#the-variant-class"
+    )
+
+    html_body_head = f"""
+        <div>
+            <style>
+                .tskit-table thead tr th {{text-align: left;padding: 0.5em 0.5em;}}
+                .tskit-table tbody tr td {{padding: 0.5em 0.5em;}}
+                .tskit-table tbody tr td:first-of-type {{text-align: left;}}
+                .tskit-details-label {{vertical-align: top; padding-right:5px;}}
+                .tskit-table-set {{display: inline-flex;flex-wrap: wrap;margin: -12px 0 0 -12px;width: calc(100% + 12px);}}
+                .tskit-table-set-table {{margin: 12px 0 0 12px;}}
+                details {{display: inline-block;}}
+                summary {{cursor: pointer; outline: 0; display: list-item;}}
+            </style>
+            <div class="tskit-table-set">
+                <div class="tskit-table-set-table">
+                <table class="tskit-table">
+                    <thead>
+                    <tr>
+                        <th style="padding:0;line-height:21px;">
+                        <img style="height: 32px;display: inline-block;padding: 3px 5px 3px 0;" src="{url_tskit_logo}"/>
+                        <a target="_blank" href="{url_variant_class_doc}"> {class_type} </a>
+                        </th>
+                    </tr>
+                    </thead>
+                    <tbody>
+        """  # noqa: B950
+
+    html_body_tail = """
+                    </tbody>
+            </table>
+            </div>
+        </div>
+        </div>
+        """
+
+    try:
+        variant.site
+
+        site_id = variant.site.id
+        site_position = variant.site.position
+        num_samples = len(variant.samples)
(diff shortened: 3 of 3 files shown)


jeromekelleher:
Cool! Can you paste in what these look like please so we can have a look?

python/tskit/genotypes.py
```diff
@@ -222,6 +223,27 @@ def copy(self) -> Variant:
...
 
+    def __str__(self) -> str:
+        """
+        Return a plain text summary of the contents of a variant.
+        """
+        variant_rows = [
+            ["Site id", str(self.site.id)],
+            ["Samples", str(len(self.samples))],
+            ["Genotypes", str(len(self.genotypes))],
+            ["Alleles", str(self.num_alleles)],
```
benjeffery, line 234: Would be nice to give the allele counts here.

python/tskit/util.py
```diff
@@ -542,6 +542,45 @@ def tree_html(tree):
...
+                summary {{cursor: pointer; outline: 0; display: list-item;}}
+              </style>
+              <div class="tskit-table-set">
+                <div class="tskit-table-set-table">
+                  <table class="tskit-table">
+                    <thead>
+                      <tr>
+                        <th style="padding:0;line-height:21px;">
+                          <img style="height: 32px;display: inline-block;padding: 3px 5px 3px 0;" src="https://raw.githubusercontent.com/tskit-dev/administrative/main/tskit_logo.svg"/>
+                          <a target="_blank" href="https://tskit.dev/tskit/docs/latest/python-api.html#the-tree-class"> Tree </a>
```
benjeffery, line 565: Would be better to link to the documentation for `Variant`.

python/tskit/util.py
```diff
@@ -542,6 +542,45 @@ def tree_html(tree):
...
+                        </th>
+                      </tr>
+                    </thead>
+                    <tbody>
+                      <tr><td>Site Id</td><td>{variant.site.id}</td></tr>
+                      <tr><td>Sample Nodes</td><td>{len(variant.samples)}</td></tr>
+                      <tr><td>Genotypes</td><td>{len(variant.genotypes)}</td></tr>
+                      <tr><td>Alleles</td><td>{variant.num_alleles}</td></tr>
```
benjeffery, line 573: Again the allele counts would be good.

benjeffery (commented):
Good start, needs some more detail like the site position and the allele counts.

python/tskit/genotypes.py
hyanwong: Would it be nice to implement https://github.com/tskit-dev/tskit/issues/2390 to help with this?

benjeffery:
@szhan Let me know if you want some pointers here.

szhan:
Example output of [`__str__()`.
Screen Shot 2022-07-20 at 5 27 42 PM

szhan:
Example output of `_repr_html_()`.
Screen Shot 2022-07-20 at 5 59 01 PM

python/tskit/util.py
szhan: The link now points to `Variant` in the doc.

szhan:
The HTML text has really long lines, so it is failing the flake8 test. What should we do?

szhan:
I'm not sure why I can't get `re.match` to work... I'm expecting `re.match` to return a `Match` object not `None`, which is what I'm getting with the code snippet below...

```
import tests.tsutil as tsutil
import re

ts = tsutil.all_fields_ts()
v = next(ts.variants())
str(v)

m = re.match(
    r"allele",
    str(v),
)
print(m)
```

szhan:
> Cool! Can you paste in what these look like please so we can have a look?

I've attached a couple of images showing some examples.

szhan:
Should we also show the ancestral allele?

jeromekelleher:
> The HTML text has really long lines, so it is failing the flake8 test. What should we do?

Looks like you can split some of the longest easily enough (no reason to have lines > 100), but there's also a noqa you can put at the end to tell flake8 to ignore it (see further down in the same function).

jeromekelleher:
> Should we also show the ancestral allele?

Let's show them in the order they're returned by ``variant.alleles``. You can do something like
```python
counts = self.counts()
freqs = self.frequencies()
for allele in self.alleles:
       print(f"Allele {allele}: {count[allele]} ({freqs[allele] * 100: .2g}%)")
```

szhan:
Here is what the output looks like.

`__str__()`
Screen Shot 2022-07-21 at 1 04 50 PM

`_repr_html_()`
Screen Shot 2022-07-21 at 12 56 43 PM

hyanwong:
Looking good, thanks @szhan 

1. Might it be worth saying "Number of alleles" and "Number of samples" rather than just "allele" and "samples"?
2. Do we need the length of genotypes in there (isn't that always the number of samples)?
3. Should we put quotes around the allele string, in case any of the alleles are the empty string (e.g. for an indel). So `Allele 'A'` rather than `Allele A`?
4. Instead of Allele None should we say e.g. "Allele missing"?

hyanwong:
> 1. Might it be worth saying "Number of alleles" and "Number of samples" rather than just "allele" and "samples"?

Actually, I forgot that (a) it might not be all the sample nodes in the ts, if a subset of samples is specified and (b) it is possible to specify non samples to be counted too.

In;m therefore not sure of the best phrasing instead of "samples", but we could say `Number of samples used`?

`Number of

hyanwong:
> Looking good, thanks @szhan
> 
> 1. Might it be worth saying "Number of alleles" and "Number of samples" rather than just "allele" and "samples"? Or `# samples` and `# alleles` if you want a shorter abbreviation.
> 2. Do we need the length of genotypes in there (isn't that always the number of samples)?
> 3. Should we put quotes around the allele string, in case any of the alleles are the empty string (e.g. for an indel). So `Allele 'A'` rather than `Allele A`?
> 4. Instead of Allele None should we say e.g. "Allele missing"?
> 5. "Missing data" might be crueller as "Has missing data"?
> 6. Is isolated as missing meant to be a boolean rather than numeric?

szhan:
> > The HTML text has really long lines, so it is failing the flake8 test. What should we do?
> 
> Looks like you can split some of the longest easily enough (no reason to have lines > 100), but there's also a noqa you can put at the end to tell flake8 to ignore it (see further down in the same function).

There is already `# noqa: B950` at the end of 'variant_html()' in `util.py`.

szhan:
This?
Screen Shot 2022-07-21 at 2 11 16 PM

hyanwong:
> This?
> Screen Shot 2022-07-21 at 2 11 16 PM

Yeah, looks good. Maybe put "Allele missing" instead of `Allele 'None'`?

hyanwong:
Is it clear what the "3" in `3 (2.6%)` means? Should we say "Samples with allele 'A'", or is that too wordy?

Also there's a space between the open bracket and the percentage amount, I think?

szhan:
For the record, I used the following code snippet to produce the HTML representation shown above.

```
from random import random
import tskit
import msprime


ts = msprime.sim_mutations(
    msprime.sim_ancestry(
        5,
        sequence_length=100,
        random_seed=1,
    ),
    rate=0.01,
    random_seed=1,
)

v = next(ts.variants())
v
```

szhan:
The order of alleles in `Variant.alleles` of the first tree of `ts_fixture` is different when I run the code on my laptop versus when run on CircleCI?!

```
def test_variant_str(self, ts_fixture):
        v = next(ts_fixture.variants())
>       assert v.alleles == ("G", "C", "T", "A", None)
E       AssertionError: assert ('C', 'G', 'T', 'A', None) == ('G', 'C', 'T', 'A', None)
E         At index 0 diff: 'C' != 'G'
E         Full diff:
E         - ('G', 'C', 'T', 'A', None)
E         ?       -----
E         + ('C', 'G', 'T', 'A', None)
E         ?  +++++
```

jeromekelleher:
> Should I just change test_variant_str so that the expected allele list matches the list produced when run on CircleCI? But the test may not pass on other machines, e.g. mine.

Yuck. This is why we've been moving away from using simulations as the basis for tests, super annoying dealing with these inconsistencies.

How about creating a very simple tree sequence by hand so you can test? Like
```
tables = tskit.TableCollection(10)
tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
tables.sites.add_row(position=5, ancestral_state="A")
tables.mutations.add_row(site=0, node=0, derived_state="T")
tables.mutations.add_row(site=0, node=1, derived_state="G")
ts = tables.tree_sequence()
```

szhan:
> > Should I just change test_variant_str so that the expected allele list matches the list produced when run on CircleCI? But the test may not pass on other machines, e.g. mine.
> 
> Yuck. This is why we've been moving away from using simulations as the basis for tests, super annoying dealing with these inconsistencies.
> 
> How about creating a very simple tree sequence by hand so you can test? Like
> 
> ```
> tables = tskit.TableCollection(10)
> tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
> tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
> tables.sites.add_row(position=5, ancestral_state="A")
> tables.mutations.add_row(site=0, node=0, derived_state="T")
> tables.mutations.add_row(site=0, node=1, derived_state="G")
> ts = tables.tree_sequence()
> ```

Just introduced `simple_ts` to create a simple dummy ts for testing.

szhan:
All the tests have passed! Updating the change log.

hyanwong:
> Yuck. This is why we've been moving away from using simulations as the basis for tests, super annoying dealing with these inconsistencies.

It's sometime helpful to create a new TS using e.g. `tskit.Tree.generate_balanced(4).tree_sequence`, but that won't have mutations. It is worth having a deterministic routine in `tskit` (or maybe just in the test suite) that adds a site and some mutations to an existing tree sequence in one pass. E.g. 

```
def add_variation(ts, position, ancestral_state, derived_states, mutation_nodes=ts.samples(), mutation_times=None):
    """
    Return a new tree sequence with a new site added at ``position`` with the given ancestral state.
    and for each element $I$ in the list of ``derived states``, add a mutation above mutation_nodes[i] at time
    mutation_times[I] (or TIME_UNKNOWN if not specified)
    """
```

Then making a test case is as simple as

```
ts = tskit.Tree.generate_balanced(4).tree_sequence
ts = tskit.add_variation(ts, 0, "A", ["T", "C"])
```

useful?

szhan:
> > Yuck. This is why we've been moving away from using simulations as the basis for tests, super annoying dealing with these inconsistencies.
> 
> It's sometime helpful to create a new TS using e.g. `tskit.Tree.generate_balanced(4).tree_sequence`, but that won't have mutations. It is worth having a deterministic routine in `tskit` (or maybe just in the test suite) that adds a site and some mutations to an existing tree sequence in one pass? E.g.
> 
> ```
> def add_variation(ts, position, ancestral_state, derived_states, mutation_nodes=ts.samples(), mutation_times=None):
>     """
>     Return a new tree sequence with a new site added at ``position`` with the given ancestral state.
>     and for each element $i$ in the list of ``derived states``, add a mutation above mutation_nodes[i]
>     at time mutation_times[i] (or TIME_UNKNOWN if not specified)
>     """
> ```
> 
> Then making a test case is as simple as
> 
> ```
> ts = tskit.Tree.generate_balanced(4).tree_sequence
> ts = tskit.add_variation(ts, 0, "A", ["T", "C"])
> ```
> 
> useful?

Should we add this in separate PR?

jeromekelleher:
> Should we add this in separate PR?

We should split it off to a separate issue at least.

szhan:
> > Should we add this in separate PR?
> 
> We should split it off to a separate issue at least.

Should we do that before finishing this PR?

jeromekelleher:
Let's not bother with it, it's just a utility for testing. @hyanwong can create an to discuss  as it's his idea.

szhan:
When there are no sample nodes selected to get the genotypes, the outputs look like this.

`__str__()`
Screen Shot 2022-07-23 at 6 59 31 PM

`_repr_html_()`
Screen Shot 2022-07-23 at 6 57 03 PM

Also, it gives this warning:
`WARNING:root:No non-missing samples at this site, frequencies undefined`

szhan:
When trying to call `_str_()` and `_repr_html_()` on an un-decoded `Variant` object, we get this error:

```
ValueError: This variant has not yet been decoded at a specific site,call Variant.decode to set the site.
```

This should be the behaviour we want, right?

benjeffery:
> When trying to call `_str_()` and `_repr_html_()` on an un-decoded `Variant` object, we get this error:
> 
> ```
> ValueError: This variant has not yet been decoded at a specific site,call Variant.decode to set the site.
> ```
> 
> This should be the behaviour we want, right?

I don't think so, it should be a one-line table with "This variant has not yet been decoded at a specific site,call Variant.decode to set the site" as the content IMO.

szhan:
Something like this?

╔═══════════════════════════════════════════════════════════╗
║ This variant has not yet been decoded at a specific site, call Variant.decode to set the site.║
╚═══════════════════════════════════════════════════════════╝

jeromekelleher:
Yes, ``__str__``  etc should always succeed even if the variant hasn't been decoded yet.

jeromekelleher (approved):
LGTM!

python/tests/test_genotypes.py
```diff
@@ -2100,3 +2102,60 @@ def test_variant_frequencies_no_samples(self, ts_fixture, caplog):
...
+                ╟─+┼─+╢
+                ║Isolated as missing\s*│\s*True║
+                ╚═+╧═+╝
+                """[
+                    1:
+                ]
+            ),
+            str(v),
+        )
+
```
hyanwong, line 2155: Can you add in a check for no samples and also when the variant isn't set to a site?

```python
def test_variant_str_no_samples():
    v = next(ts_fixture.variants(samples=[]))
    for allele in v.alleles:
        if allele is not None:
            assert re.search(f"║Samples with allele '{allele}'\s*│\s*0\s*\(nan\%\)║", str(v)) is not None

def test_variant_str_no_site():
   v = tskit.Variant(ts_fixture)
   s = str(v)
   assert len(s.splitlines()) == 3
   assert "This variant has not yet been decoded at a specific site, call Variant.decode to set the site" in s


```

hyanwong (approved):

szhan:
When calling `__str__()` and `_repr_html_()` on an un-decoded `Variant` object, the outputs look like this.

Screen Shot 2022-07-27 at 3 12 11 PM
Screen Shot 2022-07-27 at 3 11 54 PM

python/tests/test_genotypes.py
```diff
@@ -2100,3 +2102,60 @@ def test_variant_frequencies_no_samples(self, ts_fixture, caplog):
...
+        It has only one tree and one site, whose variant has the alleles
+        ('A', 'T', 'G', None).
+        """
+        tables = tskit.TableCollection(10)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0)
+        tables.sites.add_row(position=5, ancestral_state="A")
+        tables.mutations.add_row(site=0, node=0, derived_state="T")
+        tables.mutations.add_row(site=0, node=1, derived_state="G")
```
hyanwong, line 2118: Can we put one of the derived states to a unicode character, e.g. the classic 💩 ?!

hyanwong: Oh, and set one of the others to '' (empty string) and other to multiple characters, e.g. `AAC`

szhan: The poop emoji messes up the alignment of the columns...
Screen Shot 2022-07-27 at 4 23 41 PM

szhan: It is because the poop emoji is of the same width as the other characters.

hyanwong: I don't think we care. This is an edge case. But it's worth keeping a weird character in the tests to check that we don't bug out when trying to print, etc.

hyanwong (approved):
LGTM

python/tskit/genotypes.py
```diff
@@ -294,6 +295,45 @@ def frequencies(self, remove_missing=None) -> dict[str, float]:
             if not (allele is None and remove_missing)
         }
 
+    def __str__(self) -> str:
+        """
+        Return a plain text summary of the contents of a variant.
+        """
+        try:
+            self.site
```
benjeffery, line 303: Is this line needed? I know the intent is to check the variant is decoded, but that happens a few lines lower anyway.

python/tskit/util.py
```diff
@@ -542,6 +542,98 @@ def tree_html(tree):
...
+        freqs = variant.frequencies()
+
+        return (
+            html_body_head
+            + f"""
+                <tr><td>Site Id</td><td>{site_id}</td></tr>
+                <tr><td>Site Position</td><td>{site_position}</td></tr>
+                <tr><td>Number of Samples</td><td>{num_samples}</td></tr>
+                <tr><td>Number of Alleles</td><td>{num_alleles}</td></tr>
+            """  # noqa: B950
```
benjeffery, line 609: Do you still need the `noqa` here?

python/tskit/util.py
```diff
@@ -542,6 +542,98 @@ def tree_html(tree):
...
+                <tr><td>Isolated as Missing</td><td>{isolated_as_missing}</td></tr>
+            """  # noqa: B590
+            + html_body_tail
+        )
+    except ValueError as err:
+        return (
+            html_body_head
+            + f"""
+                        <tr><td>Error</td><td>{str(err)}</td></tr>
+            """  # noqa: B950
```
benjeffery, line 632: This one aswell?

python/tskit/util.py
```diff
@@ -542,6 +542,98 @@ def tree_html(tree):
...
+            )
+            + f"""
+                <tr><td>Has Missing Data</td><td>{has_missing_data}</td></tr>
+                <tr><td>Isolated as Missing</td><td>{isolated_as_missing}</td></tr>
+            """  # noqa: B590
+            + html_body_tail
+        )
+    except ValueError as err:
+        return (
+            html_body_head
```
benjeffery, line 629: Might as well get this covered.

python/tskit/util.py
```diff
@@ -542,6 +542,98 @@ def tree_html(tree):
...
+                    + " "
+                    + f"({freqs[k] * 100:.2g}%)"
+                    + "</td></tr>"
+                    for k in variant.alleles
+                ]
+            )
+            + f"""
+                <tr><td>Has Missing Data</td><td>{has_missing_data}</td></tr>
+                <tr><td>Isolated as Missing</td><td>{isolated_as_missing}</td></tr>
+            """  # noqa: B590
```
benjeffery, line 624: And here, I don't think 590 is a valid bugbear code anyway.

benjeffery (approved):
Awesome! Just a couple of nits.

python/tskit/util.py
szhan: I've added this test.

```
def test_variant_html_repr_no_site(self):
        tables = tskit.TableCollection(10)
        ts = tables.tree_sequence()
        v = tskit.Variant(ts)
        html = v._repr_html_()
        ElementTree.fromstring(html)
        assert len(html) > 1600
```

But I'm not sure if 1600 is appropriate here. The length of the HTML string is 1633.

PR #2385, opened 2022-07-06
Add site.alleles method

### LINKED ISSUE #2380: site.alleles() method?
I just realised that I can't get the alleles defined at a site. I'd prefer not to use the variants operator to do this. Is there any reason why we can't define a helper function:

```python
class Site:
    def alleles(self):
        alleles = set((m.derived_state for m in site.mutations))
        alleles.add(site.ancestral_state)
        return alleles.
```

hyanwong:
## Description

Deliberately returns an (unordered) set

Fixes #2380

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2592,6 +2592,29 @@ class TestTreeSequence(HighLevelTestCase):
         self.verify_individual_properties(ts)
 
 
+class TestSiteAlleles:
+    def test_no_mutations(self):
+        tables = tskit.TableCollection(sequence_length=1)
+        tables.sites.add_row(0, ancestral_state="")
+        site = tables.tree_sequence().site(0)
+        assert site.alleles == {""}
+
+    @pytest.mark.parametrize("k", range(5))
+    def test_k_mutations(self, k):
+        tables = tskit.TableCollection(sequence_length=1)
+        tables.sites.add_row(0, ancestral_state="ABC")
+        tables.nodes.add_row(1, 0)
+        tables.nodes.add_row(1, 0)  # will not have any mutations => missing
+        for j in range(k):
+            tables.mutations.add_row(site=0, node=0, derived_state=str(j))
+        ts = tables.tree_sequence()
+        variant = next(ts.variants())
+        assert variant.has_missing_data
+        assert len(variant.site.alleles) == k + 1
+        assert "ABC" in variant.site.alleles
+        assert variant.site.alleles == set(variant.alleles[:-1])
+
+
 class TestEdgeDiffs:
     @pytest.mark.parametrize("ts", get_example_tree_sequences())
     def test_correct_trees_forward(self, ts):
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -357,6 +357,22 @@ class Site(util.Dataclass):
             and self.metadata == other.metadata
         )
 
+    @property
+    def alleles(self) -> set[str]:
+        """
+        Return the set of all the alleles defined at this site
+
+        .. note::
+            This deliberately returns an (unordered) *set* of the possible allelic
+            states (as defined by the site's ancestral allele and its associated
+            mutations). If you wish to obtain an (ordered) *list* of alleles, for
+            example to translate the numeric genotypes at a site into allelic states,
+            you should instead use ``.alleles`` attribute of the :class:`Variant` class,
+            which unlike this attribute includes ``None`` as a state when there is
+            missing data at a site.
+        """
+        return {self.ancestral_state} | {m.derived_state for m in self.mutations}
+
 
 @metadata_module.lazy_decode()
 @dataclass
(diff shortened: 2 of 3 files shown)


python/tests/test_highlevel.py
```diff
@@ -1985,6 +1985,30 @@ def test_sites(self):
             assert mutation_index == len(mutations)
         assert some_sites
 
+    def test_site_alleles(self):
```
jeromekelleher, line 1988: No real point in running a simulatino here and smaller tests would be better. Could do something like

```
class TestSiteAlleles:
     def test_no_mutations(self):
          tables = tskit.TableCollection()
          tables.sites.add_row(0, ancestral_state="A") 
          site = tables.tree_sequence().site(0)
          assert site.alleles == {"A"}

     @pytest.mark.parametrize(k, range(5))
     def test_k_mutations(self):
          tables = tskit.TableCollection()
          tables.sites.add_row(0, ancestral_state="-1") 
          for j in range(k):
                tables.mutations.add_row(site=0, node=0, derived_state=str(j))
           assert stuff
```

python/tskit/trees.py
```diff
@@ -357,6 +357,19 @@ def __eq__(self, other):
             and self.metadata == other.metadata
         )
 
+    def alleles(self):
```
jeromekelleher, line 360: Should this be a function or an attribute? The only reason I can think of for making it a function is if we can imagine adding some parameters later.

python/tskit/trees.py
```diff
@@ -357,6 +357,19 @@ def __eq__(self, other):
...
+        """
+        Return a set of all the alleles defined at this site (the ancestral state and
```
jeromekelleher, line 362:
```suggestion
        Return the set of all the alleles defined at this site (the ancestral state and
```

python/tskit/trees.py
```diff
@@ -357,6 +357,19 @@ def __eq__(self, other):
...
+        the derived states of any mutations).
+
+        .. note::
+            Unlike the .alleles attribute of the :class:`Variant` class, this does not
```
jeromekelleher, line 366: I'm a bit fuzzy on whether there can also be situations where there would be alleles in this set that are not in the alleles returned by the variant.

I think it's important to stress the "this is a set" thing here too.

jeromekelleher (commented):
Looks good, some suggeestions

python/tskit/trees.py
hyanwong: Happy with either. I then to think of attributes as "low cost", which I guess this is.

python/tskit/trees.py
hyanwong: > I'm a bit fuzzy on whether there can also be situations where there would be alleles in this set that are not in the alleles returned by the variant.

I think probably not: it looks like the alleles returned in the variants iterator include all the mutations in the TS, even if they are above nodes that are not in the tree topology:

```
import tskit
tables.nodes.add_row(0, 0)
tables.nodes.add_row(0, 0)
tables.sites.add_row(0, ancestral_state="A")
tables.mutations.add_row(site=0, derived_state="T", node=0)
tables.mutations.add_row(site=0, derived_state="G", node=1)
ts = tables.tree_sequence()
print("Num samples =", ts.num_samples, ". Alleles", next(ts.variants()).alleles)
```

> I think it's important to stress the "this is a set" thing here too.

Good point

python/tskit/trees.py
benjeffery, line 360: Adding a type hint here would be nice seeing we're returning a set - it gets copied through to the docs.

python/tskit/trees.py
```diff
@@ -357,6 +357,19 @@ def __eq__(self, other):
...
+        Return the set of all the alleles defined at this site (the ancestral state and
+        the derived states of any mutations).
+
+        .. note::
+            Unlike the .alleles attribute of the :class:`Variant` class, this does not
+            include ``None`` as an allele when there is missing data.
+        """
+        alleles = {m.derived_state for m in self.mutations}
+        alleles.add(self.ancestral_state)
+        return alleles
```
benjeffery, line 371: A matter of taste so I don't feel strongly, but this reads better to my eye.

`return {m.derived_state for m in self.mutations} | {self.ancestral_state}`

benjeffery (commented):
Can we add a test where the ancestral state is not a single character? Easy to trip up when python does things like this:
```python
>>> a = "foobar"
>>> set(a)
{'f', 'o', 'b', 'a', 'r'}
>>> {a}
{'foobar'}
```

python/tskit/trees.py
hyanwong: We now note: 

> This deliberately returns an (unordered) *set* of the possible allelic
            states (as defined by the site's ancestral allele and its associated
            mutations). If you wish to obtain an (ordered) *list* of alleles, for
            example to translate the numeric genotypes at a site into allelic states,
            you should instead use ``.alleles`` attribute of the :class:`Variant` class,
            which unlike this attribute includes ``None`` as a state when there is
            missing data at a site.

hyanwong:
Comments addressed. Thanks @jeromekelleher  and @benjeffery. Ready to merge I think.

benjeffery (approved):

PR #2389, opened 2022-07-07
Add table.replace_with

### LINKED ISSUE #1489: Assign a table to a TableCollection
Currently this results in an error:
```python
>>> tc = tskit.TableCollection(1)
>>> tc.individuals=tskit.IndividualTable()
AttributeError: can't set attribute
```
Is there any reason that it shouldn't be possible? The implementation of this would seem to be clearing the existing table and copying the table data across.

Came up in the context of https://github.com/tskit-dev/tskit/pull/1487#discussion_r648387416

hyanwong:
## Description

And output sensible error messages when trying to set tables directly.

Fixes #1489 

# PR Checklist:

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -823,6 +823,11 @@ class CommonTestsMixin:
                     t.append_columns(**input_data)
                 input_data[offset_col.name] = np.copy(original_offset)
 
+    def test_replace_with_wrong_class(self):
+        t = self.table_class()
+        with pytest.raises(TypeError, match="wrong type"):
+            t.replace_with(tskit.BaseTable(None, None))
+
 
 class MetadataTestsMixin:
     """
@@ -3834,6 +3839,12 @@ class TestTableCollection:
                 with pytest.raises(TypeError):
                     func(bad_filename)
 
+    def test_set_table(self):
+        tc = tskit.TableCollection()
+        for name, table in tc.table_name_map.items():
+            with pytest.raises(AttributeError, match="replace_with"):
+                setattr(tc, name, table)
+
 
 class TestEqualityOptions:
     def test_equals_provenance(self):
@@ -4382,6 +4393,20 @@ class TestBaseTable:
         t = tskit.BaseTable(None, None)
         with pytest.raises(NotImplementedError):
             t.set_columns()
+
+    def test_replace_with(self, ts_fixture):
+        # Although replace_with is a BaseTable method, it is simpler to test it
+        # on the subclasses directly, as some differ e.g. in having metadata schemas
+        original_tables = ts_fixture.dump_tables()
+        original_tables.nodes.metadata_schema = tskit.MetadataSchema.permissive_json()
+        new_tables = ts_fixture.dump_tables()
+        new_tables.clear(clear_provenance=True, clear_metadata_schemas=True)
+
+        # write all the data back in again
+        for name, table in new_tables.table_name_map.items():
+            new_table = getattr(original_tables, name)
+            table.replace_with(new_table)
+        new_tables.assert_equals(original_tables)
 
 
 class TestSubsetTables:
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -589,6 +589,23 @@ class BaseTable:
             }
         )
 
+    def replace_with(self, other):
+        # Overwrite the contents of this table with a copy of the other table
+        params = {}
+        for column in self.column_names:
+            try:
+                params[column] = getattr(other, column)
+            except AttributeError:
+                raise TypeError(
+                    "Replacement table has wrong type: it lacks a {column} column"
+                )
+        try:
+            # Not all tables have a metadata_schema: if they do, encode it with repr
+            params["metadata_schema"] = repr(other.metadata_schema)
+        except AttributeError:
+            pass
+        self.set_columns(**params)
+
     def clear(self):
         """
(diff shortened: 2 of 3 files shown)


python/tests/test_tables.py
```diff
@@ -4383,6 +4394,22 @@ def test_set_columns_not_implemented(self):
         with pytest.raises(NotImplementedError):
             t.set_columns()
 
+    def test_replace_with(self):
+        # Although replace_with is a BaseTable method, it is simpler to test it
+        # on the subclasses directly, as some differ e.g. in having metadata schemas
+        ts = msprime.sim_ancestry(2, sequence_length=1, random_seed=1)
```
benjeffery, line 4400: Can you use `ts_fixture` here as it contains data in all tables.

python/tskit/tables.py
```diff
@@ -582,6 +582,20 @@ def append(self, row):
             }
         )
 
+    def replace_with(self, other):
+        """
+        Entirely replace the contents of this table with another
+        """
+        if type(self) != type(other):
+            raise TypeError("replacement column is not of the same type")
```
benjeffery, line 590: Do you mean table here instead of column?

benjeffery (approved):
Nice, couple of comments.
I still keep coming back to the idea of assignment as using this, unless you `del table` you have two copies hanging around. Will have a think, but this is good for now.

python/tskit/tables.py
hyanwong: Good spot, thanks!

hyanwong:
> Nice, couple of comments.

Good comments, thanks Ben.

> I still keep coming back to the idea of assignment as using this, unless you `del table` you have two copies hanging around. Will have a think, but this is good for now.

Sure. If you can make it more efficient in a future iteration, that would be great! But as you say, merge for now and pick up later.

python/tskit/tables.py
```diff
@@ -582,6 +582,20 @@ def append(self, row):
             }
         )
 
+    def replace_with(self, other):
+        """
+        Entirely replace the contents of this table with another
+        """
+        if type(self) != type(other):
```
benjeffery, line 589: I think this is overly restrictive, and prevents accepting anything that might have the same columns but isn't the same type, for example a subclass.
I think we should instead try to get each column and give a nice error if `other` doesn't have the attribute.

python/tskit/tables.py
```diff
@@ -582,6 +582,20 @@ def append(self, row):
...
+            raise TypeError("replacement table is not of the same type")
+        params = {column: getattr(other, column) for column in self.column_names}
+        try:
```
benjeffery, line 592: If you go with my last comment then the logic is inverted here - we should be checking if `self` needs `metadata_schema` not if `other` has it.

python/tskit/tables.py
```diff
@@ -2852,55 +2866,103 @@ def individuals(self) -> IndividualTable:
         """
         return IndividualTable(ll_table=self._ll_tables.individuals)
 
+    @individuals.setter
+    def individuals(self, value):
+        raise AttributeError(
+            "Cannot set tables in a table collection: use table.replace_with() instead."
```
benjeffery, line 2872: Instead of repeating this string can you make it a class attribute?

python/tskit/tables.py
```diff
@@ -582,6 +582,20 @@ def append(self, row):
             }
         )
 
+    def replace_with(self, other):
+        """
+        Entirely replace the contents of this table with another
```
benjeffery, line 587: Can you make it clear here that the contents of this table are overwritten by copying the other table?
Then turn it into a comment as I'd like to keep this undocumented for now so that we can consider if proper assignment is the right solution.

benjeffery (commented):
Sorry for delaying this @hyanwong, I wanted some time to think about it as it isn't really `replace_with` but `overwrite_with_a_copy_of`. 
I think that eventually, we'll want to work out the (messy) details of actually replacing table in a collection, so this is kind of a stop-gap solution. As such I've asked in a comment to leave this undocumented for now. Please add an issue to suggest documenting it so we can discuss it.

python/tskit/tables.py
hyanwong: Good idea!

hyanwong:
Changes made, but if you have a better idea for how to do the replacement properly, then I'm happy to leave this for the time being anyway. I can always code up the functionality every time it is needed instead.

benjeffery (approved):
LGTM! Thanks @hyanwong!
There's a good chance that assignment is a bigger mouthful than we want to chew for the foreseeable so happy to have this in.

python/tskit/tables.py
```diff
@@ -589,6 +589,23 @@ def append(self, row):
...
+        for column in self.column_names:
+            try:
+                params[column] = getattr(other, column)
+            except AttributeError:
+                raise TypeError(
+                    "Replacement table has wrong type: it lacks a {column} column"
+                )
+        try:
+            # Not all tables have a metadata_schema: if they do, encode it with repr
+            params["metadata_schema"] = repr(other.metadata_schema)
```
jeromekelleher, line 604: Why use ``repr`` here and not a direct API? Seem indirect and brittle (we might change repr to something slightly different, perhaps)

jeromekelleher: That's fine internally maybe, but it's not a great convention. A well named method would be much clearer to the reader

PR #2391, opened 2022-07-07
Clarify the set_columns behaviour for schemas

hyanwong:
Make it clear that column data is overwritten but schemas may not be, and mention encoding.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -632,9 +632,8 @@ class BaseTable:
 
     def set_columns(self, **kwargs):
         """
-        Sets the values for each column in this :class:`Table` using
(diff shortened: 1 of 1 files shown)


benjeffery (approved):
Thanks for spotting this @hyanwong I've thought about this a bit as at first I thought it might be a bug. But I think this is the ideal behavior has often you want to replace the columns without wiping the schema.

hyanwong:
This is stuck for some GitHub reason. Just needs a virtual kick.

benjeffery:
> This is stuck for some GitHub reason. Just needs a virtual kick.

Lint has failed.

PR #2392, opened 2022-07-07
Make the returned rank a namedtuple

hyanwong:
This is a trivial improvement that seems worth doing for clarity. Not sure about how to document it. No tests yet until the idea is deemed sensible.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -1248,7 +1248,13 @@ Also see the {ref}`sec_python_api_tree_sequences` summary.
 ```{eval-rst}
 .. autoclass:: Interval()
     :members:
+```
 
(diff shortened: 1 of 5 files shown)


benjeffery:
Yes, would be a nice improvement. Normally we'd use a `dataclass` for such a thing. If we made it subscribable and iterable it would be backwards compatible.

hyanwong:
I think we should keep this a NamedTuple, so that it is of a tuple type, maintaining backwards compatibility. We already have a NamedTuple in the source, for EdgeDiffs, so adding another seems OK to me.

hyanwong:
Added documentation and a test. Ready to merge as far as I'm concerned.

Issue #504, opened 2020-03-30
allele frequency function


petrelharp:
We should implement the `TreeSequence.allele_frequencies(sample_sets)` function, which returns a numpy array of (non-ancestral allele frequencies) x (sample_sets).

Here's an implementation:
```
def allele_frequencies(ts, sample_sets=None):
    if sample_sets is None:
       sample_sets = [ts.samples()] 
    n = np.array([len(x) for x in sample_sets])
    def f(x):
       return x / n
    return ts.sample_count_stat(sample_sets, f, len(sample_sets), windows='sites', polarised=True, mode='site', strict=False, span_normalise=False)
```
**Edit:** originally this omitted `span_normalise=False`.

jeromekelleher:
Nice, good idea.

molpopgen:
It may be nice to optionally return the joint fs as a sparse matrix.

petrelharp:
> It may be nice to optionally return the joint fs as a sparse matrix.

I'm not sure what you mean here?

molpopgen:
I assume your return value is the marginal fs in each sample set, but sometimes you want the joint fs to instead, such that fs[2,3] is the number of mutations present twice in sample set 0 and three times in set 1.  Or maybe I'm not understanding what you are returning here.

petrelharp:
That sounds like the AFS? Which we already have? Here I'm wanting to return the list of allele frequencies for each SNP.

molpopgen:
Gotcha--my mistake.

jeromekelleher:
> That sounds like the AFS? Which we already have? Here I'm wanting to return the list of allele frequencies for each SNP.

Triggered my AFS implementation PTSD there - would *not* want to do that again!

molpopgen:
I shouldn't comment when I haven't been sleeping. 😵

petrelharp:
This computes the total frequency of all mutations that are segregating in the samples, by site. People will also want to know the frequencies of mutations, separately: we should have another method that returns something of length equal to the number of mutations. I don't think we can compute this easily with the general stat framework, but here is a method to pull out this information for the mutations at a particular site:
```
def mut_counts(site):
    counts = {}
    t = ts.at(site.position)
    for m in site.mutations:
        counts[m.id] = t.num_samples(m.node)
        if m.parent != tskit.NULL:
            assert(m.parent in counts)
            counts[m.id] -= counts[m.parent]
    return counts
```

petrelharp:
[...]

PR #2393, opened 2022-07-08
Implement allele counting methods

### LINKED ISSUE #2390: Allele counting methods for Variant?
Reading #2367, it strikes me that it would be useful to get counts and frequencies of alleles out of a variant class, rather than have to decode this yourself from the genotypes array. These counts or frequencies could be shown in the `_repr_html_` method:

```
for v in ts.variants()
    an_allele = v.alleles[0]
    print(v.frequency(an_allele), v.count(an_allele))
```

the `.frequency(allele)` method should probably have an option to discount the missing data sites in the numerator when calculating the freq?

hyanwong:
## Description

Add `variant.num_missing`, `variant.counts()`, and `variant.frequencies()`. Ping @szhan - is this useful for #2384 ?

Fixes #2390 

# PR Checklist:

### DIFF
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -23,6 +23,7 @@
 Test cases for generating genotypes/haplotypes.
 """
 import itertools
+import logging
 import random
 import textwrap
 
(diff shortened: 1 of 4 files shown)


hyanwong:
No tests yet. Is the interface right @benjeffery ?

szhan:
Ah, yes, they are useful for #2384.

szhan:
They are also useful for some information-theoretic imputation quality metrics that I want to implement soon.

benjeffery:
> No tests yet. Is the interface right @benjeffery ?

Interface looks good to me, although I'm not sure about raising an error on there being zero samples, what about an empty dict as the return in that case? That's what `counts` would do. 

It would also be nice to reuse `counts` in `frequencies`.

hyanwong:
I coded this to raise a warning if frequencies are calculated on variant with all missing samples or no samples. Is this right, and should I create a separate `logger` instance in this case @benjeffery ?

petrelharp:
This looks very nice. Note: this overlaps with #504.

python/tskit/genotypes.py
```diff
@@ -215,13 +225,55 @@ def copy(self) -> Variant:
...
+            be counted in the total number of samples used to calculate the frequency,
+            and no information on the frequency of missing data is returned. Otherwise
+            (default), samples with missing data are included when calculating
+            frequencies.
+        :return: A dictionary mapping allelic states to the frequency of each allele
+            among the samples
+        """
+        if remove_missing is None:
+            remove_missing = False
+        total = self.tree_sequence.num_samples
```
benjeffery, line 264: Pretty sure this should be `len(self.samples)`?

python/tskit/genotypes.py
```diff
@@ -215,13 +225,55 @@ def copy(self) -> Variant:
...
+        if remove_missing:
+            total -= self.num_missing
+        if total == 0:
+            logging.warning(
+                "No non-missing samples at this site, frequencies undefined"
```
benjeffery, line 269: This message will also be shown if there are zero samples and remove_missing is false, so I think the text needs a tweak.

benjeffery (changes requested):
There's a mistake in the denominator for frequencies, I think!

python/tskit/genotypes.py
hyanwong: Hmm, that's an interesting question. If you call `ts.variants(samples=[0,1])`, do you expect the allele frequencies to be calculated from the entire tree sequence, or just among the samples you have specified. I'm not sure?

hyanwong: No, I'm being stupid here. The counts are only among the specified samples, so you are right @benjeffery

benjeffery: If I've specified `samples` to `Variant` then I would expect that all its operations are relative to that set of samples. Instead of `report_local_frequencies` how about  `denominator` which defaults to `len(self.samples)`?

hyanwong:
Hmm, an exciting segfault here. WTF?

hyanwong:
Maybe release 5.1 first, as we might want to think about how it interacts with @petrelharp's suggestions in #504, and nothing is waiting on it.

If @szhan wants to use it for #2384, he can simply wait until we decide on the right API.

python/tskit/genotypes.py
```diff
@@ -224,6 +233,55 @@ def copy(self) -> Variant:
...
+        at this site to the number of samples possessing that allele among the
+        set of samples specified when creating this Variant (by default, this is all
+        the sample nodes in the tree sequence). Missing data is represented by an
+        allelic state of ``None``.
+
+        :return: A dictionary mapping each allelic state to the number of samples
+            with that allele.
+        """
+        counts = {a: 0 for a in self.alleles}
+        for g, count in zip(*np.unique(self.genotypes, return_counts=True)):
```
jeromekelleher, line 248: This seems like an inefficient way do this to me. unique requires a full sort, O(n log n), but we can also do this with num_alleles passes through the array, like
```
for j, allele in self.alleles:
      counts[allele] = np.sum(self.genotypes == j)
```

since num_alleles is typically small, this should be faster for large n.

python/tskit/genotypes.py
```diff
@@ -224,6 +233,55 @@ def copy(self) -> Variant:
...
+        Return a dictionary mapping each possible :attr:`allele <Variant.alleles>`
+        at this site to the number of samples possessing that allele among the
+        set of samples specified when creating this Variant (by default, this is all
+        the sample nodes in the tree sequence). Missing data is represented by an
+        allelic state of ``None``.
+
+        :return: A dictionary mapping each allelic state to the number of samples
+            with that allele.
+        """
+        counts = {a: 0 for a in self.alleles}
```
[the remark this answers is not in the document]
hyanwong: Yep, nice idea.

python/tskit/genotypes.py
hyanwong: Good point. Happy to go with this. Or we could use `np.bincount` since these are all likely to be integers from 0..n:

```python
bincounts = np.bincount
for j, allele in enumerate(self.alleles):
    counts[allele] = bin counts[j]
```

?

jeromekelleher: Will bincount work with ``-1``?

hyanwong: Ah, really good point! I forgot about -1.

hyanwong: We could hack it by adding 1, but then I think your suggestion is much cleaner.

hyanwong: I had to special-case the `-1` anyway, because you can't simply loop through the alleles and treat the indices as required: if the last element is `None` you need to check for `-1` in the array.

So I used bin count (which is cleaner and I guess a bit faster) when there is no missing data, and a modification of your code when there is missing data. It's not that clean, but the clearest I can come up with.

hyanwong:
>  Should also consider options in the stdlib for representating frequency mapping

I had a hunt around and couldn't see anything terribly obvious.

hyanwong:
Not urgent in any way, but I could do with some help in finding out why codecov claims that e.g. L185 of genotypes.py isn't covered by the tests. By my reckon it is covered in `tests/test_genotypes.py::TestVariant::test_variant_num_missing`, and I've checked by raising an error on L185 in the tests, so I don't know quite what's going on.

jeromekelleher (commented):
LGTM. Needs a few explicit tests, as it's quite hard to see what's really being tested when checking against the ts_fixture.

python/tskit/genotypes.py
```diff
@@ -224,6 +235,61 @@ def copy(self) -> Variant:
...
+        Returns a count for each possible :attr:`allele <Variant.alleles>`
+        at this site, giving the number of samples possessing that allele among the
+        set of samples specified when creating this Variant (by default, this is all
+        the sample nodes in the tree sequence). Missing data is represented by an
+        allelic state of ``None``.
+
+        :return: A counter of the number of samples associated with each allele.
+        """
+        counts = collections.Counter()
+        if self.alleles[-1] is None:
```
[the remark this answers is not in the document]
jeromekelleher: Fair enough

python/tests/test_genotypes.py
```diff
@@ -1931,3 +1932,71 @@ def test_variant_decode(self, ts_fixture):
...
+    def test_variant_frequencies_no_samples(self, ts_fixture, caplog):
+        tables = ts_fixture.dump_tables()
+        tables.nodes.flags = np.zeros_like(tables.nodes.flags)
+        ts = tables.tree_sequence()
+        variant = next(ts.variants())
+        assert ts.num_samples == 0
+        with caplog.at_level(logging.WARNING):
+            freqs = variant.frequencies()
+            assert caplog.text.count("frequencies undefined") == 1
+        assert np.all(np.isnan(list(freqs.values())))
```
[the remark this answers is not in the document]
hyanwong: Added `test_variant_nonsample_freqs` and `test_variant_nonsample_freqs`.

jeromekelleher (approved):
LGTM

python/tskit/genotypes.py
```diff
@@ -228,6 +239,61 @@ def copy(self) -> Variant:
...
+            # we have to treat the last element of the genotypes array as special
+            counts[None] = np.sum(self.genotypes == tskit.MISSING_DATA)
+            for i, allele in enumerate(self.alleles[:-1]):
+                counts[allele] = np.sum(self.genotypes == i)
+        else:
+            bincounts = np.bincount(self.genotypes)
+            for i, allele in enumerate(self.alleles):
+                counts[allele] = bincounts[i] if i < len(bincounts) else 0
```
petrelharp, line 261: Late here, but I think you could just do this?
```
            counts = np.bincount(self.genotypes, minlength=len(self.alleles))
```

petrelharp:
This is excellent!

PR #2395, opened 2022-07-08
Auto show return types in docs

hyanwong:
Requires us to document a few classes that should be public anyway, as they are returned from public functions.


benjeffery:
Cool, just looking at this now and getting a diff of the HTML.

benjeffery (approved):
LGTM

PR #2397, opened 2022-07-10
Allow haplotypes and alignments to have left, right, samples params

### LINKED ISSUE #2092: Feedback about reference sequences: reference_start and restrict alignments options
As part of the what-is-an-ARG paper, I thought it would be interesting to see if we could encode the Kreitman (1983) _D. melanogaster_ ADH data (https://www.nature.com/articles/304412a0) as a tree sequence, including the reference sequence from FlyBase. This has brought up a few issues. In particular, I'm only focussing on bases 14615494..14618223 of the genome, for which I am using the consensus sequence in Kreitman as the reference.

I think part of the genome like this is likely to be something people are likely to want to do. Moreover, it is extremely helpful if the site positions are in the reference genome coordinate space. So in an attempt to solicit feedback, here are 2 possible suggestions to help with this

1. It might be useful for the reference sequence to have a `start_position`, by default 0, but it could be e.g in this instance 14615494. When outputting alignments, all bases before the start position (and after the last character in the reference) could simply be filled with `N`
2. It would be useful to be able to output the alignments restricted to a particular region, presumably with an extra parameter or two to the `alignments` method, specifying that start and end positions.

I think 2. is not at all controversial, but maybe 1 needs some discussion?

hyanwong:
## Description

Adds "samples", "start", and "stop" parameters to `ts.haplotypes()` and `ts.alignments()`. This (partially) fixes #2092. I was worried that calling `var.decode(site_id)` by simply iterating over the site ids would be slow, but that's how the `sites()` method does it anyway.

Is this the right approach? I haven't written any new tests until it's deemed sensible.

# PR Checklist:

### DIFF
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -638,6 +638,33 @@ class TestVariantGenerator:
         assert non_missing_found
         assert missing_found
 
+    def test_limit_interval(self):
+        ts = self.get_tree_sequence()
+        test_variant = tskit.Variant(ts)
+        test_variant.decode(1)
+        for v in ts.variants(left=ts.site(1).position, right=ts.site(2).position):
+            # should only decode the first variant
+            assert v.site.id == 1
+            assert np.all(v.genotypes == test_variant.genotypes)
+            assert v.alleles == test_variant.alleles
+
+    def test_bad_left(self):
+        ts = tskit.TableCollection(10).tree_sequence()
+        for bad_left in [-1, 10, 100, np.nan, np.inf, -np.inf]:
+            with pytest.raises(ValueError, match="`left` not between"):
+                list(ts.variants(left=bad_left))
+
+    def test_bad_right(self):
+        ts = tskit.TableCollection(10).tree_sequence()
+        for bad_right in [-1, 0, 100, np.nan, np.inf, -np.inf]:
+            with pytest.raises(ValueError, match="`right` not between"):
+                list(ts.variants(right=bad_right))
+
+    def test_bad_left_right(self):
+        ts = tskit.TableCollection(10).tree_sequence()
+        with pytest.raises(ValueError, match="must be less than"):
+            list(ts.variants(left=1, right=1))
+
 
 class TestHaplotypeGenerator:
     """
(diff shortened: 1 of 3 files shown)


benjeffery:
(Note that as written your description will close #2092)

Should `variants` also get `start` and `stop`? Not sure why it shouldn't. And yes using `decode` is the way to do this. All the heavy lifting is in seeking and decoding so the Python overhead isn't bad.

hyanwong:
> (Note that as written your description will close #2092)

Yep, That was basically deliberate. I will open a new issue with the unaddressed bits there.

> Should `variants` also get `start` and `stop`? Not sure why it shouldn't. 

I guess it could. Do you want me to add it to this PR then?

hyanwong:
p.s. do we want "start" and "stop" or "left" and "right"?

benjeffery:
> p.s. do we want "start" and "stop" or "left" and "right"?

Everywhere else that we talk about genomic intervals we talk about left, right, so that seems to make the most sense.

hyanwong:
> > p.s. do we want "start" and "stop" or "left" and "right"?
> 
> Everywhere else that we talk about genomic intervals we talk about left, right, so that seems to make the most sense.

I agree. It's clearly a genomic position if we say `left` and `right`.

benjeffery:
@hyanwong Is this ready for review?

python/tskit/trees.py
```diff
@@ -5216,30 +5295,33 @@ def alignments(self, *, reference_sequence=None, missing_data_character=None):
...
         if reference_sequence is None:
             if self.has_reference_sequence():
                 # This may be inefficient - see #1989. However, since we're
                 # n copies of the reference sequence anyway, this is a relatively
-                # minor tweak.
-                reference_sequence = self.reference_sequence.data
+                # minor tweak. We may also want to recode the below not to use direct
+                # access to the .data attribute, e.g. if we allow reference sequences
+                # to start at non-zero positions
+                reference_sequence = self.reference_sequence.data[left:right]
```
benjeffery, line 5312: Ah, I was worried about negative indexing here, but that's checked in `_check_genomic_range`.

benjeffery (approved):
LGTM!
Thanks for picking this one up - great to use all the new `Variant` machinery.

python/tskit/trees.py
```diff
@@ -4980,14 +5034,16 @@ def variants(
...
         """
-        Returns an iterator over the variants (each site with its genotypes
-        and alleles) in this tree sequence. See the
-        :class:`Variant` class for details on the fields of each returned
-        object. The ``genotypes`` for the variants are numpy arrays,
-        corresponding to indexes into the ``alleles`` attribute in the
-        :class:`Variant` object. By default, the ``alleles`` for each
+        Returns an iterator over the variants between the ``left`` (inclusive)
+        and ``right`` (exclusive) genomic positions in this tree sequence. Each
+        returned :class:`Variant` object has a site, a short list of alleles and
```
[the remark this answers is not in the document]
hyanwong: Yep, I put that in there to contract with `genotypes, but will try to rephrase

hyanwong:
Changes made. Would be nice to merge before the new point release, but whatever.

python/tests/test_genotypes.py
```diff
@@ -638,6 +638,33 @@ def test_snipped_tree_sequence_mutations_over_isolated(self):
...
+        test_variant.decode(1)
+        for v in ts.variants(left=ts.site(1).position, right=ts.site(2).position):
+            # should only decode the first variant
+            assert v.site.id == 1
+            assert np.all(v.genotypes == test_variant.genotypes)
+            assert v.alleles == test_variant.alleles
+
+    def test_bad_left(self):
+        ts = tskit.TableCollection(10).tree_sequence()
+        for bad_left in [-1, 10, 100, np.nan, np.inf, -np.inf]:
```
jeromekelleher, line 653: We prefer pytest.mark.parametrize for this sort of thing going forward

python/tskit/trees.py
```diff
@@ -5090,13 +5149,13 @@ def variants(
             isolated_as_missing=isolated_as_missing,
             alleles=alleles,
         )
-        sites = range(self.num_sites)
+        start, stop = np.searchsorted(self.tables.sites.position, interval)
```
jeromekelleher, line 5152: We're incurring a full copy of the tables here even when we don't specify ``left`` or ``right`` - could have been easily avoided.

hyanwong: Good point. I'll code around it.

PR #2404, opened 2022-07-14
Rephrase the Variant docs a little

hyanwong:
For example, I think it is always true that ehe ``alleles`` tuple is guaranteed not to contain any duplicates, rather than being only true if the alleles list is generated.

### DIFF
--- a/python/tskit/genotypes.py
+++ b/python/tskit/genotypes.py
@@ -33,25 +33,26 @@ import tskit.trees as trees
 class Variant:
     """
     A variant in a tree sequence, describing the observed genetic variation
-    among samples for a given site. A variant consists of (a) the **alleles** that
-    may be observed at the sample nodes for this site; (b) the **genotypes**
-    mapping sample IDs to the observed alleles and (c) of a reference to
-    the :class:`Site` that the Variant has been decoded at.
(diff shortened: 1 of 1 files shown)


python/tskit/genotypes.py
```diff
@@ -33,25 +33,24 @@
...
     """
     A variant in a tree sequence, describing the observed genetic variation
-    among samples for a given site. A variant consists of (a) the **alleles** that
-    may be observed at the sample nodes for this site; (b) the **genotypes**
-    mapping sample IDs to the observed alleles and (c) of a reference to
-    the :class:`Site` that the Variant has been decoded at.
+    among samples for a given site. A variant consists of (a) a tuple of
+    **alleles** listing the potential allelic states which samples at this site
+    can possess; (b) the **genotypes** mapping sample IDs to the observed alleles
+    and (c) a reference to the :class:`Site` at which the Variant has been decoded.
```
benjeffery, line 39: This should also say that a list of samples is included.

benjeffery (approved):
Nice, just one thing I think we should add.

python/tskit/genotypes.py
hyanwong: Good point. Done

PR #2405, opened 2022-07-14
Correct some docstrings and return types

hyanwong:
I think we are allowed `def alleles(self) -> tuple[str | None, ...]:` for a return type? This is a python 3.10 thing though.


benjeffery (approved):
Seems to come out fine in the docs.

PR #2407, opened 2022-07-14
Hive off definitions into a glossary

### LINKED ISSUE #2325: Doc improvement: push definitions before the data model
As discussed on Slack @hyanwong , it might be useful for new users to show them the current sub-sub-section definitions before the sub-section Data model - because there can be quite a bit of confusion between nodes and individuals etc.

Maybe also call that new sub-section a glossary?

hyanwong:
Fixes #2325

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -6,6 +6,7 @@ parts:
   - file: installation
 - caption: Concepts
   chapters:
+  - file: glossary
   - file: data-model
   - file: metadata
   - file: provenance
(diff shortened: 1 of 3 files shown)


hyanwong:
We should probably change the internal section labels, but as these are referenced from other projects (e.g. the tutorials), we need to be rather careful, so I haven't done it here.

hyanwong:
We should also put the glossary in alphabetical order, and maybe include "ARG" as an entry?

PR #2408, opened 2022-07-14
Quickstart

hyanwong:
## Description

I realise that the "Getting Started" section is rather thin, and only has installation instructions. I wondered if it would be useful to give a flavour of how tskit is used, before we dive in, using a "Quickstart" section. What do people think?


jeromekelleher (approved):
LGTM

PR #2414, opened 2022-07-15
special case start and end when we want all the variants()

### LINKED ISSUE #2410: Perf regression in #2397
The implementation of  #2397 requires a full copy of the underlying tables to be created every time we call ``variants`` [here](https://github.com/tskit-dev/tskit/blob/ad6ecd0a6889fb336b6fb9061ce2d10010807569/python/tskit/trees.py#L5149).

This can be avoided by taking a special case for when ``left is None and right is None``.

hyanwong:
## Description

This happens a lot, as it is the default. We don't have access to the original values of left and right within the function, but I think the check here is very low cost, and as a bonus will work even if explicit small `left` and large `right` values are given.

Fixes #2410 

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1804,7 +1804,7 @@ class Tree:
             multiple trees. If ``False`` or ``None`` (default) mutations are only drawn
             on an edge if their site position exists within the genomic interval covered
             by this tree. If ``True``, all mutations on each edge of the tree are drawn,
-            even if the their genomic position is to the left or right of the tree
+            even if their genomic position is to the left or right of the tree
             itself. Note that this means that independent drawings of different trees
             from the same tree sequence may share some plotted mutations.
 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
More worried about the *variants* iterator, line 5149.

Probably doesn't matter much with haplotypes anyway, since we're already creating a massive matrix.

hyanwong:
Corrected. Apologies! We could check for `left <= site(0).position and right > site(-1).position`, but probably not worth the extra effort.

benjeffery (approved):
Thanks @hyanwong, sorry I missed this on the review.

python/tskit/trees.py
```diff
@@ -5146,7 +5146,13 @@ def variants(
             isolated_as_missing=isolated_as_missing,
             alleles=alleles,
         )
-        start, stop = np.searchsorted(self.tables.sites.position, interval)
+        if left == 0 and right == self.sequence_length:
+            # Special case: no need to search the whole of the the positions array
```
jeromekelleher, line 5150: typo. comment seems unnecessary, anyhow.

jeromekelleher (approved):

python/tskit/trees.py
hyanwong: Changed, and a few similar typos in the file corrected.

PR #2415, opened 2022-07-15
Simplify table.replace_with

### LINKED ISSUE #2411: Simplify ``replace_with`` implementation
Implementation in #2389 can be much simpler. I think this works:
```python
def replace_with(self, other):
     self.set_columns(**other.asdict())
```

A lot of work went into making the dictionary representation a solid and efficient way of moving table data around, so we should avoid duplicating its logic.

hyanwong:
Fixes #2411


benjeffery (approved):
LGTM!

PR #2418, opened 2022-07-18
Fill in testing gaps for interval and sample specs.

jeromekelleher:
Fills in some gaps in the testing for functionality introduced in #2397.

In particular, it wasn't obvious to me that the searchsorted calls were going to "come from the right direction" when the queries didn't exactly match a site position (and there were no tests at all for queries that didn't match the site positions).

I think I've covered all the cases here though.


hyanwong:
Thanks Jerome. Sorry to miss these. I think the non-site-posiiton values were previously being tested when using th 0..seq_len values, but that's been special-cased out, so great to test it with other values instead.

Issue #386, opened 2019-10-15
Count extant lineages at a given time


jeromekelleher:
An operation that came up as being a useful feature in the recent PopSim meeting is to be able to count the number of extant lineages in each tree at a given time. This is equivalent to (a) deleting all edges older than t and truncating all edges intersecting with it (this would require the planned ``time`` argument to ``delete_interval`` in #382); and (b) counting the number of roots in each tree. We could define the operation by just doing this, but it seems a bit inefficient, especially if we do this for lots of different times.

Here's a way we can do it more efficiently:
```python
def count_roots(ts, t):
    time = ts.tables.nodes.time    
    num_children = np.zeros(ts.num_nodes, dtype=int)
    num_roots = ts.num_samples
    ret = np.zeros(ts.num_trees, dtype=int)

    j = 0
    for _, edges_out, edges_in in ts.edge_diffs():
        for edge in edges_out:
            if time[edge.parent] <= t:
                num_children[edge.parent] -= 1
                if num_children[edge.parent] == 1:
                    num_roots += 1
        for edge in edges_in:
            if time[edge.parent] <= t:                
                num_children[edge.parent] += 1
                if num_children[edge.parent] == 2:
                    num_roots -= 1
        ret[j] = num_roots
        j += 1
    return ret
```

We can code this up pretty efficiently in C, and return the numpy array. Should we do it this way? Should we call the method ``TreeSequence.num_roots(time=None)``, which returns a numpy array like the above? (If time is None, it defaults to the oldest time in the ts)

petrelharp:
We definately want to do this, somehow! Note that another way to do it would be to remove everything *below* `t` and count the number of tips.

And gee, your algorithm is very nice. `num_roots` is a good name; but it might be nice to connote somehow the fact that these *aren't* actually  roots in the trees unless `time=None`, maybe `num_lineages_at( )` or `num_roots_at( )`?

p.s. Asking about number of lineages in the trees irregardless of whether they are above samples is even easier. Here is, for instance, a function that I think tells you how many lineages there are in the tree at position `pos` at time `t`:
```
def num_lineages(ts, pos, t):
   left = ts.tables.edges.left
   right = ts.tables.edges.right
   a = ts.tables.nodes.time[ts.tables.edges.parent]
   b = ts.tables.nodes.times[ts.tables.edges.child]
   return np.sum((left <= pos) * (right > pos) * (a < t) * (b >= t))
```

jeromekelleher:
[...]

PR #2422, opened 2022-07-19
Tree num lineages

jeromekelleher:
Adds the ``Tree.num_lineages(t)`` method as discussed in #386 and elsewhere. 

The definition is slightly tricky because we're not just counting the edges that intersect with a given time/space coordinate, but we're also counting the number of roots. It seems logical that the number of lineages in a tree for any time greater than the root time should be 1 (assuming a single root) rather than 0.

I didn't bother adding the vector form (where t can be a sorted list of values) because it's a lot more work, and we can add it later if we want. This will be efficient enough for most purposes I think.

@savitakartik, @a-ignatieva - we should now be able to define the Relate statistic p-value easily in terms of this, right?

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -5071,6 +5071,55 @@ out:
     return ret;
 }
 
+int
+tsk_tree_num_lineages(const tsk_tree_t *self, double t, tsk_size_t *result)
+{
+    int ret = 0;
+    const tsk_id_t *restrict right_child = self->right_child;
+    const tsk_id_t *restrict left_sib = self->left_sib;
+    const double *restrict time = self->tree_sequence->tables->nodes.time;
+    tsk_id_t *stack = tsk_tree_alloc_node_stack(self);
+    tsk_size_t num_lineages = 0;
+    int stack_top;
+    tsk_id_t u, v;
+    double child_time, parent_time;
+
+    if (stack == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    if (!tsk_isfinite(t)) {
+        ret = TSK_ERR_TIME_NONFINITE;
+        goto out;
+    }
+    /* Push the roots onto the stack */
+    stack_top = -1;
+    for (u = right_child[self->virtual_root]; u != TSK_NULL; u = left_sib[u]) {
+        stack_top++;
+        stack[stack_top] = u;
+    }
+
+    while (stack_top >= 0) {
+        u = stack[stack_top];
+        parent_time = time[u];
+        stack_top--;
+        for (v = right_child[u]; v != TSK_NULL; v = left_sib[v]) {
+            child_time = time[v];
+            /* Only traverse down the tree as far as we need to */
+            if (child_time > t) {
+                stack_top++;
+                stack[stack_top] = v;
+            } else if (t < parent_time) {
+                num_lineages++;
+            }
+        }
+    }
+    *result = num_lineages;
+out:
+    tsk_safe_free(stack);
+    return ret;
+}
+
 /* Parsimony methods */
 
 static inline uint64_t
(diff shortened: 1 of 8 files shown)


a-ignatieva:
Looks good @jeromekelleher! And yes then for the test we do k = 1 + t.num_lineages(t.time(m.node) + epsilon) or something (the number of lineages just below the node carrying the mutation, but we shouldn't count all of the lineages if there is a polytomy).

jeromekelleher:
@petrelharp - are you happy with this definition?

python/tskit/trees.py
```diff
@@ -2893,6 +2893,20 @@ def sackin_index(self):
         """
         return self._ll_tree.get_sackin_index()
 
+    def num_lineages(self, t):
+        """
+        Returns the number of lineages present in this tree at time ``t``. This
+        is defined as the number of branches in this tree (reachable from roots)
```
petrelharp, line 2899: Do you mean "reachable from roots" or "reachable from samples"? (hm, I guess this is the same because roots are defined to be ones reachable from samples, but someone has to remember this definition?) I guess I'd say
```
number of branches in this tree that are reachable from the samples
```

python/tskit/trees.py
```diff
@@ -2893,6 +2893,20 @@ def sackin_index(self):
...
+        that intersect with ``t``, plus the number of roots with time ``>= t``.
```
petrelharp, line 2900: shouldn't this be `<= t`?

petrelharp: Also, I'd say explicitly that if a node with three children has time `t` then it will count as one lineage, not three.

c/tskit/trees.c
```diff
@@ -5071,6 +5071,50 @@ tsk_tree_b2_index(const tsk_tree_t *self, double base, double *result)
...
+            /* Only traverse down the tree as far as we need to */
+            if (child_time > t) {
+                stack_top++;
+                stack[stack_top] = v;
+            } else {
+                num_lineages++;
+            }
+        }
+    }
+    *result = num_lineages;
```
petrelharp, line 5112: The algorithm is very nice!

petrelharp:
Looks good to me. I could imagine not wanting the roots included, but we have zero use cases for this in the many-disconnected-trees case, so no need to complexify this. (that could be an option added later if necessary)

One question: will this skip missing samples? (i.e., if a sample is missing it's not included in the roots, right?)

jeromekelleher:
> One question: will this skip missing samples? (i.e., if a sample is missing it's not included in the roots, right?)

Ah, hmm, great point. Actually no, it'll consider these are roots in the same way (see the test I just added in 8228392), so you'll count missing a node representing missing data as a lineage.

On reflection, that probably seems right? The lineage/missing sample does exist, we just don't know anything about it.

python/tskit/trees.py
jeromekelleher: done

petrelharp:
I don't think that's right - we don't know anything about it, so it might coalesce into the existing trees more recently than `t`. (So, computing the statistics correctly would require knowing how many non-missing samples there are here.)

jeromekelleher:
Hmm - so should we just count the number of edges intersecting with time ``t`` then? This gives the same answer most of the time, except defines ``num_lineages = 0`` when t >= time of the root. (So, in a binary singly rooted tree, we'd go from 2 lineages just before the root to 0 at and older than the root).

petrelharp:
That seems okay to me, yes?

jeromekelleher:
OK, I'll update. Better to have a simpler definition!

jeromekelleher:
@petrelharp - updated to simpler definition.

petrelharp (approved):
LGTM!

hyanwong (commented):
LGTM. I'm wondering about rephrasing the docstring, though, as if I were a newcomer I think I would find it a little hard to grok. If so, I'll make a minor PR.

Issue #1916, opened 2021-11-16
memory ram usage in ts.samples()


mufernando:
`ts.samples()` is using too much ram when the keyword arguments are specified.

The subsetting by population or time is done on the Python side and it involves copying the tables (twice if pop and time are specified!). See the code [here](https://github.com/tskit-dev/tskit/blob/main/python/tskit/trees.py#L5094).

For a 3Gb tree sequence this is the ram usage I was getting:

![image](https://user-images.githubusercontent.com/9950100/142057460-0525e809-1ccc-4848-90f3-4aa3cd1fac9c.png)

** updated the figure

jeromekelleher:
Thanks @mufernando this is very useful. Can you do this again with just a single population please? It's not clear how many populations there are, and therefore what size the ``sample_sets`` dictionary is (and how big a multiplier this will be).

The operation ``recap_tsu.samples()`` should take roughly 4 bytes * num_samples - if not, we have a leak.

mufernando:
I updated the figure!

jeromekelleher:
I've done some profiling here @mufernando and I don't think we're leaking memory in any major way, so I've removed the "bug" label. The main no parameters code path is definitely returning the right thing, and has the right size and the ``array.nbytes`` is the correct value. We can definitely streamline the other code paths, but there's nothing immediate we can do about them (but I've opened #1917 to give a way we can make them a lot better).

I think the "memit" (and memoryprofile stats in general) need to be taken with a pinch of salt. When you see the first example, it looks like calling ``.samples`` uses 17MiB which sounds like too much (we have 3.6M samples, so the samples array should use 13.7 MiB). But, this isn't measuring the size of the things in the cell directly, it's measuring the difference in the size of the process **overall** from the OS's perspective. Numpy asked to malloc 13 MiB, and the OS decided to increase the size of the processes memory space by 17 MiB instead - this is totally normal and exactly what you'd expect.

mufernando:
ah, great, thanks @jeromekelleher! also, I'm not sure if this is a Jupyter thing or not, but if you see the image I had before I edited (with for loops), `%%memit` said the peak was proportional to the number of loops. Do you think this could indicate Python's garbage collection is screwing us over or is this just a bug of `memit`? I know something similar was happening when I ran my Python script because of the amount of ram overall it needed to run in our cluster.

jeromekelleher:
[...]

Issue #760, opened 2020-08-10
Table collection returned by ts.tables not read-only.


jeromekelleher:
>  One thing that has been bugging me is the description of mutability.  If you have a tree sequence, calling the functions that change it do not raise an error (like the various `add_row` methods).  Rather, they just silently do nothing.

_Originally posted by @molpopgen in https://github.com/tskit-dev/tskit/pull/756#issuecomment-670972776_

It would be good to do this, and the plan *is* for the table collection to be read-only so that we can avoid taking copies when creating a new set of tables. However, I think there would be a fair bit of work required to do this at the Python level, and we'll still want to actually mark things as read-only at the C level (so we can be sure things are safe). So, I think we should just tackle this properly when we do it.

jeromekelleher:
Note: when we do this we should update the documentation for the arrays added in #2424 to take out the comment about ``ts.tables`` incurring a full table copy. There's also some other warnings scattered around the docs to find and remove.

PR #2424, opened 2022-07-20
Table array access in TreeSequence

jeromekelleher:
Add array access to TreeSequence and solve some performance issues/memory leaks. See below for more details.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -49,6 +49,7 @@ import msprime
 import networkx as nx
 import numpy as np
 import pytest
+from numpy.testing import assert_array_equal
 
 import _tskit
 import tests as tests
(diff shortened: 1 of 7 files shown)


jeromekelleher:
Before using the standard ``ts.trees()`` approach:

timedesc-leak-before-trees

Before using the sequential ``ts.at(j)`` approach:
timedesc-leak-before-at

This branch:

timedesc-leak-after

The difference is pretty stark - 60M vs around 650M. It's interesting that it doesn't seem like this is a real "leak" per-se, because the memory does hit a high-water mark. It's an awful lot more though, and maybe we're having to run garbage collection or something to get it back?

jeromekelleher:
Fixes #1916:

Benchmark:
```python
import msprime
import time


before = time.perf_counter()
ts = msprime.sim_ancestry(
    10000,
    sequence_length=1000000,
    population_size=10_000,
    recombination_rate=1e-8,
    random_seed=1234,
)
print(f"Simulation of {ts.num_trees} trees done after {duration:.2f} seconds")

for _ in range(1000):
    samples = ts.samples(time=0)
    # print(samples)
```
Before:
samples-mem-before

This branch:

samples-mem-after

jeromekelleher:
Performance improvement of ``ts.site(position=x)``. Benchmark:
```python
import msprime
import time


before = time.perf_counter()
ts = msprime.sim_ancestry(
    10000,
    sequence_length=1000000,
    population_size=10_000,
    recombination_rate=1e-8,
    random_seed=1234,
)
ts = msprime.sim_mutations(ts, rate=1e-7, random_seed=1)
print(f"Simulation of {ts.num_trees} trees done after {duration:.2f} seconds")

for site in ts.sites():
    s = ts.site(position=site.position)
    if site.id >= 1000:
        break
```

Before:
site-search-before

After:
site-search-after

jeromekelleher:
Performance for using ``ts.variants(left=x)``:

```python
import msprime
import time


before = time.perf_counter()
ts = msprime.sim_ancestry(
    10000,
    sequence_length=1000000,
    population_size=10_000,
    recombination_rate=1e-8,
    random_seed=1234,
)
ts = msprime.sim_mutations(ts, rate=1e-7, random_seed=1)
print(f"Simulation of {ts.num_trees} trees done after {duration:.2f} seconds")

for _ in range(10):
    for var in ts.variants(left=1):
        pass
```
Before:
varstart-before

After:
varstart-after

jeromekelleher:
Running edge_diffs:
```python
import msprime
import time


before = time.perf_counter()
ts = msprime.sim_ancestry(
    10000,
    sequence_length=1000000,
    population_size=10_000,
    recombination_rate=1e-8,
    random_seed=1234,
)
# ts = msprime.sim_mutations(ts, rate=1e-7, random_seed=1)
print(f"Simulation of {ts.num_trees} trees done after {duration:.2f} seconds")

before = time.perf_counter()
for _ in range(100):
    for _ in ts.edge_diffs():
        pass

print(f"Ran edge diffs in {duration:.2f} seconds")
```
Before
edge_diffs_before

After:

edge_diffs_before
:

jeromekelleher:
This is ready for review. The basic idea is to provide array access via the TreeSequence object in lieu of the "proper" solution of making TreeSequence.tables a read-only view (which is hard, engineering wise). On the way I've moved all internal usage of ``self.tables`` as they were all leading to significant performance problems (at best) and sometimes major memory leaks (#2428). I don't understand why some of these were leaking memory and others aren't, but we can follow that up later - at least library functions shouldn't leak and be efficient.

Since we've got some memory leaks in the released version we should probably get this out ASAP.

python/tskit/trees.py
```diff
@@ -3860,6 +3860,160 @@ class TreeSequence:
     the :meth:`.variants` method iterates over all sites and their genotypes.
     """
 
+    individuals_flags: np.array
+    """
+    Efficient access to the ``flags`` column in the
+    :ref:`sec_individual_table_definition` as a numpy array (dtype=np.uint32).
+    Equivalent to ``ts.tables.individuals.flags``.
```
petrelharp, line 3867: "Equivalent to" makes it sound like there's no difference (the "Efficient" above notwithstanding). Maybe "Obtains the same values as `xxx` but without copying the entire tables."? Or, maybe it's okay?

petrelharp (approved):
NICE!!! Looks good to me - just a docs suggestion/comment.

python/tskit/trees.py
jeromekelleher: I've added an extrac note about this

jeromekelleher:
Marking this one for merging as I need it to work with the 1.4M sample tree sequences on Zenodo (and there's another bug that needs resolving to get a VCF out of them). Please do follow up with issues for any review problems you spot @benjeffery @hyanwong

hyanwong:
Sorry for the slow review. LGTM - this is a great improvement. My only query would be if / how we intend to maintain this in the long term, once we address #760. Will we then deprecate these attributes, and simply have them as permanently maintained (and undocumented?) equivalents to `ts.tables.nodes.time` and friends?

jeromekelleher:
> Will we then deprecate these attributes, and simply have them as permanently maintained (and undocumented?) equivalents to ts.tables.nodes.time and friends?

I think we can keep them permanently as part of the public API. There is an advantage of fewer attribute lookups in ``ts.node_flags`` to ``ts.tables.nodes.flags`` as well as a bit less typing, so no harm in keeping them I think.

benjeffery (commented):
Been through this closely and looks good!

python/tests/test_highlevel.py
```diff
@@ -2591,6 +2592,81 @@ def test_individual_properties(self, n):
...
+        with pytest.raises(AttributeError, match="delete attribute"):
+            delattr(ts, array)
+        with pytest.raises(ValueError, match="read-only"):
+            a[:] = 1
+
+    def test_arrays_equal_to_tables(self, ts_fixture):
+        ts = ts_fixture
+        tables = ts.tables
+
+        assert_array_equal(ts.individuals_flags, tables.individuals.flags)
```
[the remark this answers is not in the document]
jeromekelleher: I thought this was simpler and easier to follow. There would have been some obscure string parsing and getattring in order to get the corresponding table array, and this seemed more ... obvious.

Issue #2428, opened 2022-07-21
Referring to tables in TableCollection leaks memory


jeromekelleher:
See analysis in #2424 for details, and e.g., #2427.

jeromekelleher:
TODO: add some reproducible example here using mprof

jeromekelleher:
Let's mimic the memory leak we see in ``ts.site(position=x)`` by accessing the relevant bits of the tables in a loop:

```python
import msprime
import time

before = time.perf_counter()
ts = msprime.sim_ancestry(
    10000,
    sequence_length=1000000,
    population_size=10_000,
    recombination_rate=1e-8,
    random_seed=1234,
)
ts = msprime.sim_mutations(ts, rate=1e-7, random_seed=1)
print(ts)
duration = time.perf_counter() - before
print(f"Simulation of {ts.num_trees} trees done after {duration:.2f} seconds")

for site in ts.sites():
    # pos = ts.tables.sites.position
    tables = ts.tables
    if site.id >= 1000:
        break
```

First, just accessing ``ts.tables`` like here:

![access-tables](https://user-images.githubusercontent.com/2664569/180258672-8799f4b0-e1e3-4641-b075-5249cf260674.png)

As it should be

jeromekelleher:
Now, change it around so we do ``pos=ts.tables.sites.position``:

![access-tables-sites-position](https://user-images.githubusercontent.com/2664569/180258993-4e6e60f3-862b-4fb5-8338-1007b905b94c.png)

We're using waaaaay more memory, and it looks like garbage collection isn't recovering it all.

jeromekelleher:
Accessing ``ts.tables.sites`` gives us:

![access-tables-sites](https://user-images.githubusercontent.com/2664569/180259503-e3825634-cf10-4f1f-9acc-a8c9596a64d0.png)

:thinking:

jeromekelleher:
It seems like there must be some circular references down in where we do the allocation of the Table instances. It's all quite complicated and twisty down there though, so I'm at a bit of a loss as to what's going on here.

molpopgen:
We've got a couple of these piling up.  Relevant to tskit-dev/msprime#1946?

jeromekelleher:
Note:  #2080 is relevant here, because we create a new instance of the high-level table class each time ``tables.nodes`` is accessed.

jeromekelleher:
[...]

PR #2434, opened 2022-07-22
Remove circular references in Table classes

benjeffery:
For discussion with respect to #2431 and #2428.
The high memory usage in the loop in #2428 is resolved if `gc.collect()` is called in the loop. This indicates that a circular reference is at fault. This reference is due to the attribute replacement in `MetadataColumnMixin.__init__` which results in the closure method being held in the dictionary of each table instance. I haven't traced the full cycle from there, but the closure holds lots of references. 
This PR does a small refactor to make `BaseTable` check for additional arguments to the row constructor, in this way the instance attribute doesn't need to be rewritten and the circular reference and the excessive memory usage are gone.

I've only done a quick proof here - this could probably be done in a cleaner way.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -344,8 +344,14 @@ class BaseTable:
 
     def __init__(self, ll_table, row_class, **kwargs):
         self.ll_table = ll_table
-        self.row_class = row_class
+        self._row_class = row_class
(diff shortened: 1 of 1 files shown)


benjeffery:
Closed in favour of #2439

Issue #2436, opened 2022-07-25
Implement state machine for tsk_variant_t and keep reference to ``tree_sequence`` in restricted_copy.


jeromekelleher:
We are currently using the ``tree_sequence`` attribute of a way of determining whether a variant is a frozen copy or not. We also use the ``variant->site.position`` attribute as a way of determining if the variant has been decoded. It would be simpler if we had a single state machine, which supported transitions:
```
VARIANT_STATE_NEW -> VARIANT_STATE_DECODED
VARIANT_STATE_DECODED -> VARIANT_STATE_DECODED
VARIANT_STATE_DECODED -> VARIANT_STATE_FROZEN_COPY
```

Thus, 

- If state == VARIANT_STATE_NEW then ``tsk_variant_restricted_copy`` should fail
- if state == VARIANT_STATE_FROZEN_COPY then ``tsk_variant_decode`` should fail

The current approach of using the ``tree_sequence`` is problematic because

- We're documenting this attribute and are **not** documenting that it is currently NULL for frozen copies 
- We're still referring to memory from the tree sequence from the initial variant through the ``site`` copy (e.g., the list of mutations is still pointing to memory from the original ts). Thus, we still have a dependency on the original ts. Note, we're currently getting away with this dependence in the Python C API layer because we don't refer to the pointers within the ``site`` reference, but we could easily forget this some day.  

I think we can also remove some complexity in ``tsk_variant_restricted_copy`` because we can then avoid taking copies of the alleles in the ``user_alleles`` memory.

jeromekelleher:
We could consider renaming this to ``tsk_variant_frozen_copy`` (keeping the current name as an alias)

benjeffery:
Closing for inactivity and labelling "future".

PR #2437, opened 2022-07-25
Fix segfault in tsk_variant_restricted_copy

### LINKED ISSUE #2429: crash inside tskit in SLiM's CI tests
The original issue is here:

https://github.com/MesserLab/SLiM/issues/334

So see that issue for details.  The important bit is the crash log on GitHub Actions here:

https://github.com/MesserLab/SLiM/runs/7205270523?check_suite_focus=true

After it dawned on me (with illumination provided by @petrelharp) that the log in GitHub Actions actually provided a full backtrace already (I'm not used to Python-style backtraces :->), it became clear that the crash appears to be in variant stuff inside `genotypes.py` in tskit.  It looks like it is triggered by `for var in ts.variants(isolated_as_missing=False):` at line 23 in SLiM's `test_consistency.py` test script.  Line numbers have shifted slightly, but the crash is somewhere vaguely around here: https://github.com/tskit-dev/tskit/blob/c12c1608aac8e2e826e53172c94c398d1972333b/python/tskit/genotypes.py#L237.

I believe @petrelharp is currently looking into this, but since it's probably an issue with the new variant stuff, probably @benjeffery  would be interested?

jeromekelleher:
Incorrect buffer size calculations caused segfault when long alleles are present.

Closes #2429

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -1,8 +1,12 @@
 --------------------
-[1.2.0] - 2022-XX-XX
+[1.1.1] - 2022-XX-XX
 --------------------
 
+**Bug fixes**
 
+- Fix segfault in tsk_variant_restricted_copy in tree sequences with large
+  numbers of alleles or very long alleles
+  (:user:`jeromekelleher`, :pr:`2437`, :issue:`2429`).
 
 --------------------
 [1.1.0] - 2022-07-14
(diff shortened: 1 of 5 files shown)


jeromekelleher:
This fixes the issue as simply as possible (and underlines the advantages of using the ``x= malloc(size * sizeof(*x));`` pattern), however I think we can do things more simply and robustly by using a proper state machine for the variant class (#2436). We can do this as a follow up once the immediate segfault has been fixed, and released.

benjeffery (approved):
LGTM!

Issue #2328, opened 2022-06-09
Export examples section needed


jeromekelleher:
When we're next thinking about how to rejig the docs, we badly need to get a section where we can give some explanation and text about how to export data to VCF etc. The docstring for write_vcf is huge at the moment, and it would be much better structured (and we could say more, giving example etc) if it was done in an example-driven way like the msprime docs.

I'm not sure where this fits in the overall structure, though.

Issue #1325, opened 2021-04-21
TestSubsetUnion::test_no_mutatation_times - Issue that test is skipped because of is closed, but test still fails.


benjeffery:
The test ``TestSubsetUnion::test_no_mutatation_times`` is skipped due to #1225, which is now closed. However the test still fails. 
https://github.com/tskit-dev/tskit/blob/main/python/tests/test_tables.py#L4195

@petrelharp maybe it is obvious to you what is happening here before I dig into it?

petrelharp:
It wasn't obvious, but I've worked out that it fails [at compute mutation parents](https://github.com/tskit-dev/tskit/blob/97cb1de1a2b04a4e19ae7d47442c2b3cd3ee00e3/c/tskit/tables.c#L10577) because of #1227/#1232.

benjeffery:
Ah, of course. I've added a note over there.

PR #2443, opened 2022-07-26
VCF handling of individuals 

jeromekelleher:
How we handle individuals in VCFs is currently quite muddled, as we're basically assuming that files either have a zero individuals or they have individuals that are all referred to by sample nodes.

This update tries to improve things:

- By default, we let the VCF samples correspond to the individuals that have sample nodes (in numerical order)
- We raise an error if individuals have mixed sample status nodes, or we try to specify an individual that does not have a sample node.

This seems like the most reasonable approach to me, but it's tricky so any thoughts much appreciated before I document and finalise this.


python/tskit/vcf.py
```diff
@@ -115,13 +111,34 @@ def __init__(
...
+        ts = self.tree_sequence
         self.samples = None
         self.individual_ploidies = []
+
+        if individuals is None:
+            # Find all sample nodes that reference individuals
+            self.individuals = np.unique(ts.nodes_individual[ts.samples()])
+            if len(self.individuals) == 1 and self.individuals[0] == tskit.NULL:
+                # No samples refer to individuals
+                self.individuals = []
```
petrelharp, line 128: I don't see anything wrong here, but it seems confusing that if someone passes `individuals = None` then later in the code we can end up with `individuals = []`, but the user passing in `individuals = []` is forbidden. Maybe just leave `individuals=None` in this case?

petrelharp:
This seems good! At first I wondered why there's the restriction to sample nodes, but I think it's actually a good one (to help avoid errors).

python/tskit/vcf.py
jeromekelleher: Good call - I'll try and make this cleaner.

jeromekelleher:
@petrelharp I've removed the TestSubsetUnion test here because it seemed quite brittle and seems to have depended somehow on some random number of individuals getting inserted between samples. Given that the other test was also disabled (see  #1325 for more discussion and it wasn't particularly obvious what it was doing I thought the simplest thing to do was remove it).

I can reinstate and add an option to ``insert_random_ploidy_individuals`` to put in required gaps if you prefer, though?

python/tskit/vcf.py
jeromekelleher: done

petrelharp:
Huh. I have no idea why the test would depend on that. Want to open an issue to check back in on `union`?

jeromekelleher:
> Huh. I have no idea why the test would depend on that. Want to open an issue to check back in on union?

I think it's more likely the test was wrong than union (I tried the test with a bunch of random seeds and it failed in different ways). Do you think union is well tested without this, or should we open an issue to track getting more test coverage on it?

benjeffery:
Just reviewing now - have tacked on a commit adding error text checks.

python/tests/test_vcf.py
```diff
@@ -703,3 +761,95 @@ def test_ok_with_pysam(self):
...
+        1\t2\t1\t0\t1\t.\tPASS\t.\tGT\t1
+        1\t4\t2\t0\t1\t.\tPASS\t.\tGT\t1
+        1\t6\t3\t0\t1\t.\tPASS\t.\tGT\t0"""
+        expected = textwrap.dedent(s)
+        assert drop_header(ts.as_vcf(individuals=[1])) == expected
+
+    def test_reversed(self):
+        ts = self.ts()
+        s = """\
+        #CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\ttsk_0\ttsk_1
```
benjeffery, line 849: Is the `tsk_X` numbering always sequential? Or should it be the individual ID?

benjeffery (approved):
LGTM, one query about the `tsk_X` ordering.

python/tests/test_vcf.py
jeromekelleher: Good call, I guess that's a bug in the existing code.

jeromekelleher: Opened #2445 to track as this would be a breaking change to switch it now, and we don't want anything other than bugfixes in the upcoming release.

jeromekelleher:
Working my way through trying to explain this I realise that I'll have to do #2328 as well, in order to actually explain this stuff in an anyway comprehensible way. Will hopefully get it done tomorrow morning.

benjeffery:
Ok, I'm ready to release when this is done.

jeromekelleher:
I think this is ready to go. The changes to the docs are pretty big, and I don't think they're still quite right so I think the best thing to do is to log an issue for following up, so we can get the bug fixes out while still moving in the right direction docs-wise. I'll file a few follow-ups now.

docs/export.md
```diff
@@ -0,0 +1,399 @@
...
+See the {ref}`sec_glossary` for more details on tskit's data model
+ and terminology.
+:::
+
+(sec_export_vcf_compression)=
+
+### Compressed output
+
+The simplest way to compress the VCF output is to use the
+is to to use the `tskit vcf` {ref}`command line interface<sec_cli>`
```
benjeffery, line 136:
```suggestion
`tskit vcf` {ref}`command line interface<sec_cli>`
```

python/tskit/trees.py
```diff
@@ -5972,17 +5972,41 @@ def write_vcf(
...
+        adjacent sample nodes using the ``ploidy`` argument. See the
+        :ref:`sec_export_vcf_constructing_gt` section for more details
+        and examples.
+
+        If individuals that are associated with sample nodes are defined in the
+        data model (see :ref:`sec_individual_table_definition`), the genotypes
+        for each of the individual's samples are combined into a phased
+        multiploid values at each site. By default, all individuals associated
+        with sample nodes are included in increasing order of individual ID are
+        included.
```
benjeffery, line 5993:
```suggestion
        with sample nodes are included in increasing order of individual ID.
```

python/tskit/trees.py
```diff
@@ -5972,17 +5972,41 @@ def write_vcf(
...
+        using the ``individuals`` argument. It is an error to specify any
+        individuals that are not associated with any nodes, or whose
+        nodes are not all samples.
+
+        Mixed-sample individuals (e.g., those associated with one node
+        that is a sample and another that is not) in the data model will
+        result in an error by default. However, such individuals can be
+        excluded using the ``individuals`` argument.
+
+        If there are no individuals in the tree sequence tree sequence,
```
benjeffery, line 6005:
```suggestion
        If there are no individuals in the tree sequence,
```

benjeffery (approved):
Wow, love the new `export.md`. Just a couple of typos.

jeromekelleher:
Great! Should be ready to go now.

Issue #2444, opened 2022-07-27
Performance statistics report


jeromekelleher:
It would be very useful to have some simple microbenchmarks we could run from time-to-time. We've tried using ASV, but it's very heavyweight and collects masses of data that's never used. 

Requirements:

- Make a script ``perf_benchmark.py`` which runs a standard set of microbenchmarks and outputs the result to a file. The idea is this file should be updated before every release, so that we can spot any perf regressions and also so that we have a per-release history of the benchmarks in the history.
- Update the developer docs to include how and when to run these benchmarks as part of the release process.

CPU time performance benchmarks:

Given a standard file

- [x] time to load
- [x] time to save
- [x] time to access ``ts.tables`` in a loop
- [x] time to access ``tables.nodes``, ``tables.individuals`` etc
- [x] time to access columns, ``nodes.flags``, etc
- [x] time to get first tree, ts.first()
- [x] time to seek to middle tree
- [x] time to iterate over all trees
- [x] time to access tree arrays, tree.parent etc.
- [x] time to decode all variants
- [x] time to iterate over all rows, with and without metadata
- [x] time to write to vcf (writing to devnull)

additional as I thought of them (BJ):
- [x] tree node arrays (`postorder` et al)
- [x] tree row accessors (`ts.node(42)`)
- [x] tree accessors (`tree.left_sib(42)`)
- [x] iterate over `tree.nodes`

It would be nice to track the memory usage here, but this may be much more difficult and not worth the effort.

benjeffery:
Done some research here, measuring the variance of different methods. The main aim is to avoid this phenomenon in micro benchmarks, where running code over a period of 20s reveals two performance regimes:
![no_running](https://user-images.githubusercontent.com/8552/181750485-32c40cc8-2097-4068-bffc-5fb2d808a781.png)

Using `timeit` and having a balance between "number of trials in the tight loop" and "number of repetitions of the loop" seems to avoid this phenomena and gives lowest variance. I've converged on:
```python
def autotime(setup: str, code: str)
    t = timeit.Timer(setup=setup, stmt=code)
    # We want each test to last around 1-2 seconds, and to have a good number of trials and
    # repetitions as this gives the lowest variance in testing.
    one_run = t.timeit(number=1)
    num_trials = int(max(1, 0.01 / one_run))
    num_repeats = math.ceil(2 / (one_run * num_trials))
    t1 = time.time()
    results = t.repeat(repeat=num_repeats, number=num_trials)
[...]

PR #2454, opened 2022-08-01
Benchmarking

benjeffery:
WIP #2444 

Needs some polish, but probably a good point to get early feedback on the direction. The results are appended to a JSON file that will be committed to the repo, and a very basic HTML report generated:

Screenshot from 2022-08-01 12-05-22

My thinking is that at under a minute we can add this to CI.


jeromekelleher:
Looks great!

Some unexpected differences here in the e.g. ``tree.parent_array``, which make me wonder if anything under 50ns should just be "green". Do you think these are significant?

Re running in CI, I guess we just run it to make sure it hasn't broken but don't pay any attention to the numbers? 

Probably worth spinning into its own Workflow, so we're not running it n-times in the Tests?

benjeffery:
> Looks great!
> 
> Some unexpected differences here in the e.g. `tree.parent_array`, which make me wonder if anything under 50ns should just be "green". Do you think these are significant?

Yes, the colouring was a quick hack. I think for those fast ones that is usual variance.
> 
> Re running in CI, I guess we just run it to make sure it hasn't broken but don't pay any attention to the numbers?
Exactly, mostly run it to check we didn't break it, occasionally look at the numbers if useful.

> Probably worth spinning into its own Workflow, so we're not running it n-times in the Tests?

Yeah, very simple action that will finish before all the other tests, so not making our CI longer in total.

benjeffery:
Ok, this needs some docs, but is ready for a code review.
Here is the full table - there are a couple of regressions that might be worth looking into.
screencapture-file-home-benj-projects-tskit-python-benchmark-bench-results-html-2022-08-02-12_20_56

jeromekelleher (approved):
Looks great!

The only major thing that's missing from the benchmarks is row-by-row access with/out metadata decoding, but we can add that in later.

petrelharp:
Wow!!!! This is great!

benjeffery:
I've added a test for decoding metadata. I've also added a script that runs the benchmarks across all released versions.
I've also added the script to CI (takes ~5m), the results are available as an artifact - they won't be directly comparable as it is a different CPU, but the relative changes should still be useful. I think this is ready to merge.

benjeffery:
BTW, this shows how great the backwards compatibility has been. Same 0.5.2 file for all these benchmarks!

python/benchmark/run-for-all-releases.py
```diff
@@ -0,0 +1,19 @@
...
+def versions(package_name):
+    url = f"https://pypi.org/pypi/{package_name}/json"
+    data = json.load(urlopen(url))
+    return sorted(data["releases"].keys(), key=StrictVersion)
+
+
+if __name__ == "__main__":
+    versions = [v for v in versions("tskit") if "a" not in v and "b" not in v]
+    for v in tqdm.tqdm(versions):
+        os.system(f"pip install tskit=={v}")
```
jeromekelleher, line 18: ``subprocess.run(check=True, shell=True)`` is a better option here

jeromekelleher, line 18: I think it's probably worth automating in a venv here, so that the devs environment doesn't get messed up by running this.

jeromekelleher, line 18: ``subprocess.run(check=True, shell=True)`` is a better option here

python/benchmark/run.py
```diff
@@ -0,0 +1,210 @@
...
+    ret["min_frequency"] = cpufreq.min
+    ret["current_frequency"] = cpufreq.current
+    ret["cpu_usage_per_core"] = [
+        percentage for percentage in psutil.cpu_percent(percpu=True, interval=1)
+    ]
+    ret["total_cpu_usage"] = psutil.cpu_percent()
+    return ret
+
+
+def make_file():
```
jeromekelleher, line 42: Slightly nicer:
```
benchfile = tskit_dir / "benchmark" / "bench.trees"
if not benchfile.exists():
     ...
     ts.dump(benchfile)
```
similar patterns below

jeromekelleher (approved):
Awesome!

Few minor suggestions, merge away whenever you're happy.

PR #2456, opened 2022-08-04
link_ancestors stops after all relevant edges have been processed

### LINKED ISSUE #2442: Improve efficiency of link_ancestors
I'm noticing that link_ancestors is scaling much more poorly than I intuitively expected (memory seems super-linear wrt sample size?! :( ) Will think a bit more about why this might be, but there is one reasonably small 'edit' to the algorithm we could make that should improve things somewhat.

Currently, we're iterating through all of the edges in the edge table (see [here](https://github.com/tskit-dev/tskit/blob/8c932c2290d1737c4c0d739abbee58f811b4b428/python/tests/simplify.py#L586)) but we should actually just stop as soon as we reach a point in the edge table where the parent node IDs exceed the largest of the supplied ancestral node IDs.

gtsambos:
Not finished yet, but shows one way of addressing the efficiency problems mentioned in #2442. Currently assumes that the nodes IDs are in the (canonical) backwards-time order, which is why I changed one of the tests. Otherwise nothing breaks!

Fixes #2442

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7372,6 +7372,7 @@ typedef struct {
     tsk_id_t *buffered_children;
     tsk_size_t num_buffered_children;
     double sequence_length;
+    double oldest_node_time;
 } ancestor_mapper_t;
 
 static tsk_segment_t *TSK_WARN_UNUSED
@@ -7512,6 +7513,23 @@ out:
     return ret;
 }
 
+static void
+ancestor_mapper_find_oldest_node(ancestor_mapper_t *self)
+{
+    const double *node_time = self->tables->nodes.time;
+    tsk_size_t j;
+    double max_time = -1;
+
+    for (j = 0; j < self->num_ancestors; j++) {
+        max_time = TSK_MAX(max_time, node_time[self->ancestors[j]]);
+    }
+    for (j = 0; j < self->num_samples; j++) {
+        max_time = TSK_MAX(max_time, node_time[self->samples[j]]);
+    }
+
+    self->oldest_node_time = max_time;
+}
+
 static int
 ancestor_mapper_init_samples(ancestor_mapper_t *self, tsk_id_t *samples)
 {
@@ -7626,6 +7644,7 @@ ancestor_mapper_init(ancestor_mapper_t *self, tsk_id_t *samples, tsk_size_t num_
     if (ret != 0) {
         goto out;
     }
+    ancestor_mapper_find_oldest_node(self);
     ret = tsk_edge_table_clear(self->result);
     if (ret != 0) {
         goto out;
@@ -7805,6 +7824,8 @@ ancestor_mapper_run(ancestor_mapper_t *self)
     tsk_id_t parent, current_parent;
     const tsk_edge_table_t *input_edges = &self->tables->edges;
     tsk_size_t num_edges = input_edges->num_rows;
+    const double *node_time = self->tables->nodes.time;
+    bool early_exit = false;
 
     if (num_edges > 0) {
         start = 0;
@@ -7817,14 +7838,21 @@ ancestor_mapper_run(ancestor_mapper_t *self)
                 if (ret != 0) {
                     goto out;
                 }
-                current_parent = parent;
                 start = j;
+                current_parent = parent;
+                if (node_time[current_parent] > self->oldest_node_time) {
+                    early_exit = true;
+                    break;
+                }
             }
         }
-        ret = ancestor_mapper_process_parent_edges(
-            self, current_parent, start, num_edges);
-        if (ret != 0) {
-            goto out;
+        if (!early_exit) {
+            /* If we didn't break out of the loop early, we need to still process
+             * the final parent */
+            ret = ancestor_mapper_process_parent_edges(self, current_parent, start, j);
+            if (ret != 0) {
+                goto out;
+            }
         }
     }
 out:
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2021 Tskit Developers
+# Copyright (c) 2022 Tskit Developers
 # Copyright (c) 2015-2018 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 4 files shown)


python/tests/simplify.py
```diff
@@ -582,12 +582,15 @@ def __init__(self, ts, sample, ancestors):
...
             self.add_ancestry(0, self.sequence_length, sample_id, sample_id)
         self.edge_buffer = {}
+        self.oldest_ancestor = max(max(ancestors), max(sample))
 
     def link_ancestors(self):
         if self.ts.num_edges > 0:
             all_edges = list(self.ts.edges())
             edges = all_edges[:1]
             for e in all_edges[1:]:
+                if e.parent > self.oldest_ancestor:
```
jeromekelleher, line 592: I don't think works in general, you're assuming that nodes have nondecreasing time. Better to check ``time[e.parent] > self.oldest_ancestor_time``.

jeromekelleher (commented):
Looks good, and quite a simple change to make for a big gain. Better to explicity compare times though.

python/tests/simplify.py
gtsambos: just checking, there isn't a max() function in any of the C libraries tskit relies on, right? if not I'll write that myself

benjeffery: There's python's native `max` or `numpy.ndarray.max`?

gtsambos:
Thanks Ben! It's the C version of this function that I'm less sure about --
if you could take a look at the function I've added in the latest commit
I'd really appreciate it!

gtsambos:
Let me know if this looks ready to go, @benjeffery or @jeromekelleher

python/tests/simplify.py
jeromekelleher: There's a TSK_MAX macro you can use in a loop.

c/tskit/tables.c
```diff
@@ -7512,6 +7513,32 @@ ancestor_mapper_add_ancestry(ancestor_mapper_t *self, tsk_id_t input_id, double
...
+static int
+ancestor_mapper_oldest_node(ancestor_mapper_t *self)
+{
+    const double *node_time = self->tables->nodes.time;
+    tsk_size_t a;
+    int ret = 0;
+    double max_time = -1;
+
+    for (a = 0; a < self->num_ancestors; a++) {
+        if (node_time[self->ancestors[a]] > max_time) {
```
jeromekelleher, line 7525: Could use ``max_time = TSK_MAX(max_time, node_time[self->ancestors[a]]);``

c/tskit/tables.c
```diff
@@ -7512,6 +7513,32 @@ ancestor_mapper_add_ancestry(ancestor_mapper_t *self, tsk_id_t input_id, double
...
+            max_time = node_time[self->ancestors[a]];
+        }
+    }
+    for (a = 0; a < self->num_samples; a++) {
+        if (node_time[self->samples[a]] > max_time) {
+            max_time = node_time[self->samples[a]];
+        }
+    }
+    if (max_time < 0) {
```
jeromekelleher, line 7534: I guess this can only happen if we have zero samples and zero ancestors, which we're already checking for?

c/tskit/tables.c
```diff
@@ -7818,6 +7850,10 @@ ancestor_mapper_run(ancestor_mapper_t *self)
                     goto out;
                 }
                 current_parent = parent;
+                if (node_time[current_parent] > self->oldest_node_time) {
+                    goto out;
```
jeromekelleher, line 7854: We only use ``goto out`` for error conditions. You can ``break`` here instead, right?

python/tests/simplify.py
```diff
@@ -582,12 +582,17 @@ def __init__(self, ts, sample, ancestors):
         for sample_id in sample:
             self.add_ancestry(0, self.sequence_length, sample_id, sample_id)
         self.edge_buffer = {}
+        self.oldest_ancestor_time = max(ts.tables.nodes.time[u] for u in ancestors)
```
jeromekelleher, line 585: ``ts.nodes_time[u]`` is the preferred approach now - it's a lot faster. Ditto below.

jeromekelleher (commented):
LGTM, a few minor implementation points.

jeromekelleher:
Can you update the CHANGELOG also please @gtsambos?

jeromekelleher:
The docs fail is probably transient, don't worry about it.

c/tskit/tables.c
gtsambos: I don't think we can use break, because we have more than one loop to break out of. We end up getting other errors:
```
stderr:
Bug detected in ../tskit/tables.c at line 7802. If you are using tskit directly please open an issue on GitHub, ideally with a reproducible example. (https://github.com/tskit-dev/tskit/issues) If you are using software that uses tskit, please report an issue to that software's issue tracker, at least initially.
```

benjeffery:
I think the docs might need a fix, I'll get on it.

c/tskit/tables.c
gtsambos: ok -- I put that final call inside another if statement

gtsambos:
the codecov gods are cruel and unforgiving! ⚡

c/tskit/tables.c
```diff
@@ -7818,13 +7838,19 @@ ancestor_mapper_run(ancestor_mapper_t *self)
                     goto out;
                 }
                 current_parent = parent;
+                if (node_time[current_parent] > self->oldest_node_time) {
+                    break;
+                }
+
                 start = j;
```
jeromekelleher, line 7845: I think we can make this a bit simpler - why do we have the last call to ancestor_mapper_process_parent_edges below? 

I think you should put the ``break`` after the ``start = j`` bit also, to keep the loop invariants correct.

gtsambos: I guess it depends on what makes things simpler -- the latest push changes it so that process_parent_edges is only called inside the loop, but it means everything in the loop has to be put inside conditional statements.

c/tskit/tables.c
```diff
@@ -7805,27 +7824,29 @@ ancestor_mapper_run(ancestor_mapper_t *self)
...
     tsk_size_t num_edges = input_edges->num_rows;
+    const double *node_time = self->tables->nodes.time;
 
     if (num_edges > 0) {
         start = 0;
         current_parent = input_edges->parent[0];
-        for (j = 0; j < num_edges; j++) {
-            parent = input_edges->parent[j];
-            if (parent != current_parent) {
+        parent = input_edges->parent[0];
```
gtsambos, line 7832: Note this line isn't strictly necessary, but without it gcc now throws a warning about using `parent` uninitialised (probably because the initialisation is happening inside an if statement now)

jeromekelleher:
I've added  a suggested refactoring here that avoids within-loop checks (which I find a bit confusing).

You seem to have made changes to the kastore submodule here which will need to be dropped.

gtsambos:
hmm, Jerome's commit credit was removed when I squashed and rebased the commit, somehow 😮‍💨 you can see from the commit message that his code is in there, though

python/tests/simplify.py
```diff
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2021 Tskit Developers
+# Copyright (c) 2022 Tskit Developers
```
jeromekelleher, line 3: Something went wrong here - should be 2019-2022

jeromekelleher (approved):
Looks great, thanks @gtsambos! 

One tiny nit on the copyright updates, but happy to merge whenever.

jeromekelleher:
> hmm, Jerome's commit credit was removed when I squashed and rebased the commit, somehow ￼ you can see from the commit message that his code is in there, though

That's fine, as I expected.

PR #2460, opened 2022-08-09
Added methods to sort and squash segments in the IbdFinder output.

### LINKED ISSUE #2459: Provide option to squash IBD output?
One of the most important comments I got in my thesis examination was regarding our definition of IBD. Our current implementation of `ibd_segments` treats segments as distinct if they have been inherited along distinct 'genealogical paths', which are essentially sequences of edges. So in this instance, 
```
    #
    #        4       |      4       |        4
    #       / \      |     / \      |       / \
    #      /   \     |    /   3     |      /   \
    #     /     \    |   2     \    |     /     \
    #    /       \   |  /       \   |    /       \
    #   0         1  | 0         1  |   0         1
    #                |              |
    #                0.2            0.7
```
there will be 3 separate IBD segments for the sample pair (0,1), because the paths in the middle tree explicitly contain some extra intermediary nodes. (This has caused confusion for other users before: see #1479).

I was pretty sure this behaviour was desirable until I read this comment of Graham's:

> if I’m understanding things correctly, I’m not sure the path definition actually lines up with most people’s definition of halotype IBD. The good news is that given segments called from the tree-sequence path definition you can always stitch them to form IBD segments in post-processing, but I caution against calling the original segments IBD. (*Indeed, there are many unsampled lineages in the population that will coalesce and recombine with the lineages [0] & [1] before [0] & [1] coalesce together, our definition of IBD should depend on the properties of just the history of the pair not of the coalescence pattern of other sampled or unsampled alleles.

This is a super valid point, and kind of major imo. In many of the tree sequences we come across, nodes like (2) and (3) will be coalescence nodes that relate our samples to some other sample that has recombined at the breakpoints (who we are perhaps uninterested in). It's bad if our definition of IBD between the pair (0, 1) depends on whether those other samples are included in the trees or not, and currently it does.

> The good news is that given segments called from the tree-sequence path definition you can always stitch them to form IBD segments in post-processing

We've certainly talked about doing this before, but I'm unsure of how much effort it will take. I'm about to push up a draft PR that implements these changes in our Python `IbdFinder`, but I'm not sure this would translate well to the C code given the fancy struct that Jerome made for the output. The necessary algorithms are themselves fairly simple -- a sort method and a squash method, both very similar to those used on edge tables.

pinging @jeromekelleher, @petrelharp

gtsambos:
Fixes #2459 -- see the issue for discussion about why this is needed. Draft only at the moment.

### PR Checklist:

### DIFF
--- a/python/tests/ibd.py
+++ b/python/tests/ibd.py
@@ -138,6 +138,24 @@ class IbdResult:
         key = (a, b) if a < b else (b, a)
         self.segments[key].append(tskit.IdentitySegment(seg.left, seg.right, seg.node))
 
+    def sort_and_squash_segments(self, a, b):
+        # Sort and squash in-place. This will work differently in the C implementation.
+        key = (a, b) if a < b else (b, a)
+        self.segments[key] = sorted(self.segments[key])
+        num_segs = len(self.segments[key])
+        new_list = []
+        if num_segs > 1:
+            for seg in self.segments[key]:
+                if len(new_list) == 0:
+                    new_list.append(seg)
+                else:
+                    n = new_list[-1]  # last segment
+                    if seg.node > n.node or seg.left > n.right:
+                        new_list.append(seg)
+                    else:
+                        new_list[-1] = tskit.IdentitySegment(n.left, seg.right, n.node)
+            self.segments[key] = new_list
+
 
 class IbdFinder:
     """
@@ -177,7 +195,7 @@ class IbdFinder:
         for u, a in enumerate(self.A):
             print(u, self.sample_set_id[u], a, sep="\t")
 
-    def run(self):
+    def run(self, squash=False):
         node_times = self.tables.nodes.time
         for e in self.ts.edges():
             time = node_times[e.parent]
@@ -195,11 +213,13 @@ class IbdFinder:
                 if intvl[1] - intvl[0] > self.min_span:
                     child_segs.append(Segment(intvl[0], intvl[1], s.node))
                 s = s.next
-            self.record_ibd(e.parent, child_segs)
+            self.record_ibd(e.parent, child_segs, squash=squash)
             self.A[e.parent].extend(child_segs)
+        # if squash:
+        #     self.result.squash()
         return self.result.segments
 
-    def record_ibd(self, current_parent, child_segs):
+    def record_ibd(self, current_parent, child_segs, squash):
         """
         Given the specified set of child segments for the current parent
         record the IBD segments that will occur as a result of adding these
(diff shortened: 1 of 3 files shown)


gtsambos:
Hey @hyanwong! Unfortunately, we both made some pretty involved changes to the test_highlevel.py file in the past few days. I think I've fixed up the merge conflict, but I can't figure out how to get my tests to run with the argument that I added into `get_example_tree_sequences()`. Currently I'm just skipping these tests because I don't fully understand the syntax of your additions -- do you know how to ensure these tests 'see' the argument I've added in?

hyanwong:
It was @benjeffery who fixed up `get_example_tree_sequences()` to run in parallel, cached, so that `parameterize` worked properly. He might be able to suggest a fix?

benjeffery:
I'll take a look now

benjeffery:
I think the simplest thing here is to remove `custom_max`.
I've done this in https://github.com/tskit-dev/tskit/commit/a3aa3da8f2ba991a48967011a230a32cf6fe62cd
Currently, I've set the `N` numbers to those generated when `custom_max` was `15`. Once all the IBD tests pass we should add in some even numbers again and skip them in the IBD tests? (I'm assuming that you need them to be odd, as all those generated by `custom_max` were?

gtsambos:
Thanks @benjeffery and sorry for the late reply -- I had to take a few days off to deal with some urgent moving-to-the-US things.

> Currently, I've set the N numbers to those generated when custom_max was 15. Once all the IBD tests pass we should add in some even numbers again and skip them in the IBD tests? (I'm assuming that you need them to be odd, as all those generated by custom_max were?

I added this argument because some of my changes in this PR make IbdFinder run much more slowly, to the point where they look like they're hanging. (This is likely a Python-specific problem -- as we discuss in #2459, the C code works differently because it uses an AVL tree to do the sorting). But I didn't want to omit the tests entirely, as they cover some useful edge cases, so I wanted to run them with smaller sample sizes than the ones hard-coded in.

benjeffery:
> I added this argument because some of my changes in this PR make IbdFinder run much more slowly, to the point where they look like they're hanging. (This is likely a Python-specific problem -- as we discuss in #2459, the C code works differently because it uses an AVL tree to do the sorting). But I didn't want to omit the tests entirely, as they cover some useful edge cases, so I wanted to run them with smaller sample sizes than the ones hard-coded in.

We can leave the slow examples in the examples - but skip them in these tests?

benjeffery:
I've rebased this to main - the tests here only add a few seconds on my machine. @jeromekelleher I think this is pretty complete if you want to have a quick look to check?

benjeffery (approved):

benjeffery:
@mergifyio update

PR #2462, opened 2022-08-10
Mock-up of ibd-segments trash collection method.

### LINKED ISSUE #2461: Garbage collection in ibd_segments, link_ancestors and simplify
All three of these methods rely on an object called `ancestor_map` (C codebase) or `A` (Python mockups) to store segments that descend from each ancestral node in the tree sequence. In the case of `ibd_segments` and `link_ancestors`, these can be a big memory hog (much bigger than the size of the tree sequences themselves). However, the segments corresponding to a given ancestral node are no longer needed once we've processed all of the edges with that node in the 'child' position of the edge table. By pruning away this garbage periodically, we could improve the memory usage of these methods (at the expense of some runtime).

The difficulty is in the freeing of memory -- since the `ancestor_map` objects have their memory allocated all at once, they also have to be freed all at once. It's difficult to imagine changing this without substantially complicating the existing code.

one alternative is to periodically create a pruned copy of the `ancestor_map`, and to then replace the original version with the pruned copy. I'm about to push up a PR that demonstrates this.

gtsambos:
## Description

This is a Python mockup of some functions that might help to keep the internal memory requirements of `ibd_segments` in check. Would work similarly for `link_ancestors` (and perhaps also for `simplify`, but with smaller gains). See #2461 for more details. I'm aware that I'm really pumping out these PRs and issues at the moment 😅 -- this one is more of a long-term project and not urgent for anyone to look at

Before taking this any further, there are a few things we'll need to figure out:
1. (under what conditions) does this actually improve memory usage? I need to do some profiling of these functions.
2. what are the costs to runtime? I suspect this will really depend on how frequently we run these methods -- I've included a `trash_interval` parameter here that controls this.
3. will this implementation translate well into C code? (I might need @jeromekelleher's insight here sometime in the future)

Fixes #2461 

### PR Checklist:

### DIFF
--- a/python/tests/ibd.py
+++ b/python/tests/ibd.py
@@ -144,7 +144,16 @@ class IbdFinder:
     Finds all IBD relationships between specified sample pairs in a tree sequence.
     """
 
-    def __init__(self, ts, *, within=None, between=None, min_span=0, max_time=None):
+    def __init__(
+        self,
+        ts,
+        *,
+        within=None,
+        between=None,
+        min_span=0,
+        max_time=None,
+        empty_trash=True,
+    ):
         self.ts = ts
         self.result = IbdResult()
         if within is not None and between is not None:
@@ -167,6 +176,10 @@ class IbdFinder:
             if self.sample_set_id[u] != -1:
                 self.A[u].append(Segment(0, ts.sequence_length, u))
         self.tables = self.ts.tables
+        self.empty_trash = empty_trash
+        # if self.empty_trash:
+        #     print("We're emptying the trash!!")
+        self.num_of_child_edges = [-1 for i in range(ts.num_nodes)]
 
     def print_state(self):
         print("IBD Finder")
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I don't think this is the way we'd go about it - better to either write an allocator that we can free object as we are done with them, or to just use malloc directly (but that would need some serious code scrutiny to make sure we do the right thing in all situations).

Maybe you could try doing the malloc/free way and as a quick prototype and see how it works?

jeromekelleher:
> Maybe you could try doing the malloc/free way and as a quick prototype and see how it works?

I think this would give a lot of insight - looking at this in Python won't really tell you much.

gtsambos:
Ah sorry I misunderstood, I thought you just  meant to reorganise the code so that it reflects how the malloc/free method would work in C. I'll get cracking on that.

benjeffery:
Hi @gtsambos, is this PR still something you're working on?

gtsambos:
Hi Ben! It would be fine to close this PR -- the changes that Jerome
mentions above are totally different to what I've done, so maybe better for
a different pr. I'm just working on this all rather slowly as I've been
moving around a lot since I last commented here!

benjeffery:
Thanks @gtsambos! I was just having a little spring-clean so no worries.

Issue #678, opened 2020-06-13
Add samples and sites options to ts.genotype_matrix?


hyanwong:
As noted in the docs, the `genotype_matrix` method can be very memory intensive. As we get to larger datasets, I suspect users might want to only extract portions of the matrix, e.g. certain samples, or certain regions. For example one extreme would be to export the haplotype for a single individual. The obvious way to do this is to run `simplify` to reduce the number of samples, and `keep_intervals` to reduce the number of sites, but this isn't likely to be obvious to the average user. Should we provide 2 parameters to the `genotype_matrix` method itself which, if not `None`,  run these two preprocessing steps to reduce the output before returning the genotype matrix?

jeromekelleher:
I think there's a ``samples`` argument to ``variants`` which we could also easily add to genotype matrix. That would be an easy PR, I think. Going through ``simplify`` etc seems excessive.

Sites is a bit less obvious, but I guess list of site IDs to output would be fine. For now, we could just skip any sites that are not in this list as we're iterating over the variants.

But, ah - this is done in C at the moment IIRC. That's probably pointless, to be honest, and we should probably just implement the ``genotype_matrix`` method in Python.

mufernando:
I'm interested in this, so might have a try.

jeromekelleher:
> I'm interested in this, so might have a go at it soon.

Great, thanks @mufernando! The first thing to do is to figure out if a Python version of the current ``genotype_matrix`` function performs much worse than what we have now. To do this, create a large tree sequence (mininum 100K samples) and do some timing checks.

If we don't lose any real perf by doing it in python, then we can get rid of the current Python-C function, and our lives will be much easier.

jeromekelleher:
To really do this efficiently for sites we'd need a method for efficiently seeking to particular trees, so I'm putting this in the random access project.

petrelharp:
In removing SlimTreeSequence over in pyslim, I've run in to the [`ts.mutations_at( )`](https://github.com/tskit-dev/pyslim/blob/main/pyslim/slim_tree_sequence.py#L459) method I provided there, that provides more or less this functionality. It'd be nice to point to a replacement in tskit. What's the proposal for this API? Is it a method of `Variant`? I'm not sure if anyone is using it, so perhaps I won't actually do it right away (maybe, though?), but it'd at least be nice to have a concrete path forward worked out, and perhaps I can re-use my testing code from pyslim at least.

petrelharp:
[...]

PR #2465, opened 2022-08-16
Add Python implementation of genotype_matrix

szhan:
Part of #678

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5256,9 +5256,17 @@ class TreeSequence:
         if isolated_as_missing is None:
             isolated_as_missing = not impute_missing_data
 
-        return self._ll_tree_sequence.get_genotype_matrix(
-            isolated_as_missing=isolated_as_missing, alleles=alleles
+        variant = tskit.Variant(
+            self, isolated_as_missing=isolated_as_missing, alleles=alleles
         )
+
+        ret = np.zeros(shape=(self.num_sites, self.num_samples), dtype=np.int32)
+
+        for site_id in range(self.num_sites):
+            variant.decode(site_id)
+            ret[site_id, :] = variant.genotypes
+
+        return ret
 
     def alignments(
         self,
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -9385,78 +9385,6 @@ out:
 }
 
 static PyObject *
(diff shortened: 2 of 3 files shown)


python/tskit/trees.py
```diff
@@ -5256,10 +5256,22 @@ def genotype_matrix(
...
 
-        return self._ll_tree_sequence.get_genotype_matrix(
-            isolated_as_missing=isolated_as_missing, alleles=alleles
+        variant = tskit.Variant(
+            self,
+            isolated_as_missing=isolated_as_missing,
+            alleles=alleles,
         )
 
+        ret = np.zeros(shape=(self.num_sites, self.num_samples))
```
benjeffery, line 5265:
```suggestion
        ret = np.zeros(shape=(self.num_sites, self.num_samples), dtype=np.int32)
```

benjeffery (commented):
Screenshot from 2022-08-16 21-35-55

3x slower than C!



I'll have a think about how to improve this - worst case we use the fast path for when the `samples`/`sites` are something like 60-80% full and use the python way when we are skipping a lot, or just add the arguments in C....

jeromekelleher:
I'm surprised this is so much slower - where is the time being spent?

benjeffery:
> I'm surprised this is so much slower - where is the time being spent?

@szhan Do you want to pair on some perf analysis?

szhan:
C implementation.
```
ts.genotype_matrix();'tiny.trees'  2.0 µ
ts.genotype_matrix();'bench.trees' 287.3 m
```

Python implementation.
```
ts.genotype_matrix();'tiny.trees'  6.2 µ
ts.genotype_matrix();'bench.trees' 455.9 m
```

It is taking 1.6 times longer to run `bench.trees`.

jeromekelleher:
Seems like an acceptable perf tradeoff to me.

We shouldn't forget to remove the Python C implementation, now it's unused.

benjeffery:
@szhan Do you know what to remove in `_tskitmodule.c`, or would you like some pointers? It should be part of this PR.

szhan:
I think `TreeSequence_get_genotype_matrix()` and the snippet `.ml_name = "get_genotype_matrix"`?

benjeffery:
> I think `TreeSequence_get_genotype_matrix()` and the snippet `.ml_name = "get_genotype_matrix"`?

Looks good, just the test that needs removing also.

benjeffery (approved):
Great, thanks @szhan

PR #2471, opened 2022-08-23
Add order argument to ts.nodes()

### LINKED ISSUE #2370: Add order argument to TreeSequence.nodes
I was just after getting the oldest node(s) in the tree sequence, which I did by `node_times = ts.tables.nodes.time; oldest = np.flatnonzero(node_times == np.max(node_times))` but beginners might not know how to do this sort of thing, which involves knowing about tables. Is it worth having a few wrapper functions for things like this (although I don't know where they would belong)?

hyanwong:
Fixes #2370

This uses a lambda function, which make it quite a neat and flexible change without affecting performance of the `SimpleContainerSequence.__getitem__` function. Is that OK?

It also means we can do things like:

```
ts = tskit.Tree.generate_balanced(10).tree_sequence
nodes_in_time_order = ts.nodes(order="timedesc")
nodes_in_time_order[0]  # returns node 18
```

But I think that's quite reasonable behaviour. If this seems like the right approach, I'll add some tests.

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -356,6 +356,8 @@ def make_example_tree_sequences():
     )
     tables = ts.dump_tables()
     tables.populations.metadata_schema = tskit.MetadataSchema(None)
+    ts = tskit.Tree.generate_balanced(8).tree_sequence
+    yield ("rev node order", ts.subset(np.arange(ts.num_nodes - 1, -1, -1)))
     ts = tables.tree_sequence()
     assert ts.num_trees > 1
     yield (
@@ -1440,6 +1442,27 @@ class TestTreeSequence(HighLevelTestCase):
     @pytest.mark.parametrize("ts", get_example_tree_sequences())
     def test_pairwise_diversity(self, ts):
         self.verify_pairwise_diversity(ts)
+
+    @pytest.mark.parametrize("order", ["abc", 0, 1, False])
+    def test_bad_node_iteration_order(self, order):
+        ts = tskit.TableCollection(1).tree_sequence()
+        with pytest.raises(ValueError, match="order"):
+            ts.nodes(order=order)
+
+    @pytest.mark.parametrize("ts", get_example_tree_sequences())
+    def test_node_iteration_order(self, ts):
+        order = [n.id for n in ts.nodes()]
+        assert order == list(range(ts.num_nodes))
+        order = [n.id for n in ts.nodes(order="id")]
+        assert order == list(range(ts.num_nodes))
+        order = np.array([n.id for n in ts.nodes(order="timeasc")], dtype=int)
+        assert np.all(ts.nodes_time[order] == np.sort(ts.nodes_time))
+        # Check it conforms to the order of parents in the edge table
+        parent_only_order = order[np.isin(order, ts.edges_parent)]
+        edge_parents = np.concatenate(
+            (ts.edges_parent[:-1][np.diff(ts.edges_parent) != 0], ts.edges_parent[-1:])
+        )
+        assert np.all(parent_only_order == edge_parents)
 
     def verify_edgesets(self, ts):
         """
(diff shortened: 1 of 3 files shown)


jeromekelleher:
LGTM. Not obvious to me why ``timedesc`` would have ascending node ID as the secondary key though (or at least I can't think of any practical usage of this). Wouldn't it be simpler if ``timedesc`` was equal to  reversed ``timeasc``?

hyanwong:
> Wouldn't it be simpler if timedesc was equal to reversed timeasc

There's no need to have a "timedesc" in that case, as the iterator is reversible: `reverse(ts.nodes(order="timeasc")`. I put "timedesc" in there in case the alternative was needed, for instance traversing down through an ARG while retaining the "standard" order for ties. But it's a very niche case. Maybe we remove "timedesc" until such time as someone else wants that particular functionality.

hyanwong:
@mergifyio rebase

python/tests/test_highlevel.py
```diff
@@ -1441,6 +1441,33 @@ def verify_pairwise_diversity(self, ts):
     def test_pairwise_diversity(self, ts):
         self.verify_pairwise_diversity(ts)
 
+    def test_bad_node_order(self):
+        ts = tskit.TableCollection(1).tree_sequence()
+        for order in ["abc", 0, 1, False]:
```
jeromekelleher, line 1446: This is a pytest.parametrize pattern - generally if you're using a loop in a test like this, change it to a parametrize.

python/tests/test_highlevel.py
```diff
@@ -1441,6 +1441,33 @@ def verify_pairwise_diversity(self, ts):
...
+            with pytest.raises(ValueError, match="order"):
+                ts.nodes(order=order)
+
+    def test_node_order(self):
```
jeromekelleher, line 1450: This test looks generic, so could parametrize by the example tree sequences instead of just this one example?

jeromekelleher (commented):
LGTM, some comments on testing.

python/tests/test_highlevel.py
hyanwong: Oh yes, of course.

python/tests/test_highlevel.py
hyanwong: Yes, but I'm slightly worried that the example tree sequences all have nodes in time order (they were mostly generated by msprime), so I rolled my own version where I was sure the nodes were out-of-time-order.

Perhaps that should be added to the example tree sequences. I thought it wasn't worth testing the other examples though, as they would be likely to return nodes in the same order regardless of the order parameter.

jeromekelleher: Right I missed that, sorry.

Perhaps you could check if the examples all have nodes in time order, and add this if not?

hyanwong: The `decapitate` and `decapitate recomb` examples have a few nodes out of time order. I guess that's by "accident" rather than design. We could either add another deliberately out-of-order ts to the examples (but that'll increase the testing load) or perhaps add a "meta" test to check that the `get_example_tree_sequences()` function has some examples with out-of-time-order nodes?

jeromekelleher: Just add in the example you've made. We need to review those examples anyway.

I don't like the idea of just testing this on a single example, we've hit errors before where corner cases get forgotten about. The examples contain some gnarly things like empty tree sequences which could affect, say, lexsort.

hyanwong: Will do. Waiting on #2479 to be merged to make the test case easier.

hyanwong: Example added in to the general set. Should be ready to merge.

benjeffery:
@hyanwong Shall I review now?

hyanwong:
> @hyanwong Shall I review now?

Yes please.

benjeffery (approved):

PR #2474, opened 2022-08-25
Pin sub-dependencies due to new PIP resolver

benjeffery:
Gah.
The new `pip` resolver is strict in the sense that it requires that all dependencies' sub-dependency version specifications are satisfied. To do this it tries out many versions, one of which (`jinga`) has a broken version that fails install. It also takes ages to do this. The proximal fix is to pin some sub-dependancy versions as I've done here. Long term it requires python packages to be far more careful about specifying versions of dependancies.


jeromekelleher:
Do we need to update our setup.cfgs accordingly?

PR #2478, opened 2022-08-26
Negative indexing

### LINKED ISSUE #1008: Negative table indexing
It's often useful to check the last provenance. We should support `ts.provenance(-1)`

hyanwong:
## Description

Fixes #1008

How should I do the tests? It looks like the containers are all tested using the `SimpleContainersMixin`, so I guess I should add something to that to text for negative indexes. But I'm not sure how the mixing stuff works, TBH.

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -16,13 +16,16 @@
    Default is ``None``, where all the sample nodes are selected.
    (:user:`szhan`, :pr:`2493`, :issue:`678`)
 
-**Changes**
+**Breaking Changes**
 
  - Single statistics computed with ``TreeSequence.general_stat`` are now
    returned as numpy scalars if windows=None, AND; samples is a single
    list or None (for a 1-way stat), OR indexes is None or a single list of 
    length k (instead of a list of length-k lists).
    (:user:`gtsambos`, :pr:`2417`, :issue:`2308`)
+
+ - Accessor methods such as ts.edge(n) and ts.node(n) now allow negative
+   indexes (:user:`hyanwong`, :pr:`2478`, :issue:`1008`)
 
 **Performance improvements**
 
--- a/python/tests/test_combinatorics.py
+++ b/python/tests/test_combinatorics.py
@@ -889,7 +889,7 @@ class TestCountTopologies:
         sample_sets = [[0], [1]]
         self.verify_value_error(tables.tree_sequence(), sample_sets)
 
-        sample_sets = [[0], [-1]]
+        sample_sets = [[0], [tables.nodes.num_rows]]
         self.verify_node_out_of_bounds_error(tables.tree_sequence(), sample_sets)
 
     def verify_value_error(self, ts, sample_sets=None):
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -351,13 +351,13 @@ def make_example_tree_sequences():
                 model=msprime.InfiniteSites(msprime.NUCLEOTIDES),
             ),
         )
+    ts = tskit.Tree.generate_balanced(8).tree_sequence
+    yield ("rev node order", ts.subset(np.arange(ts.num_nodes - 1, -1, -1)))
     ts = msprime.sim_ancestry(
         8, sequence_length=40, recombination_rate=0.1, random_seed=seed
     )
     tables = ts.dump_tables()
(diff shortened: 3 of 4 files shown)


jeromekelleher:
Hmm, tricky one. Let's take a chance to think this one through so we're sure there's no unintended consequences.

I think we should probably do it for actual tables also if we do it for the ts accessors, but we could log that as a follow up issue required for the next release.

benjeffery:
> What do you think @benjeffery?

Yes, better to do the tables as a separate issue. I've been trying to think of adverse consequences, the only thing I can think is that -ve indexing can hide bugs in index calculation - but it is a standard Python thing.

jeromekelleher:
Yeah - I don't think we can break anyone's code, unless they are depending on us throwing an error on a negative index right?

benjeffery:
> Yeah - I don't think we can break anyone's code, unless they are depending on us throwing an error on a negative index right?

Yes, that would be the only case - not impossible that someone has done that. Strictly speaking, this is a breaking change.

jeromekelleher:
Seems fine then once we highlight it as such in the release notes.

hyanwong:
> Seems fine then once we highlight it as such in the release notes.

it looks like the approach is OK then. Can someone (e.g. @benjeffery ) give me pointers about my question below:

> How should I do the tests? It looks like the containers are all tested using the SimpleContainersMixin, so I guess I should add something to that to text for negative indexes. But I'm not sure how the mixing stuff works, TBH.

benjeffery:
`SimpleContainersMixin` tests the containers returned by the row accessors.

Something like this added to `TestTreeSequence`:
```python
@pytest.mark.parametrize("ts", get_example_tree_sequences())
    def test_row_getter(self, ts):
        for table_name in ts.tables_dict.keys():
            sequence = getattr(ts, table_name)()
            length = getattr(ts, "num_" + table_name)
            for i, n in enumerate(sequence):
                assert i == n.id
                assert getattr(ts, table_name[:-1])(i) == n
                assert getattr(ts, table_name[:-1])(-(length-i)) == n
```
Haven't tested this though!

hyanwong:
Thanks for the testing code @benjeffery - ready for review.

benjeffery (commented):
Are there tests not visible in the diff for things like floating point, numpy ints, `None` and other oddities?

hyanwong:
> Are there tests not visible in the diff for things like floating point, numpy ints, `None` and other oddities?

Hmm, good point. I don't think those were ever tested, TBH. We just assumed that they would be dealt with by the default  python indexing stuff. I could add specific tests.

hyanwong:
> > Are there tests not visible in the diff for things like floating point, numpy ints, `None` and other oddities?
> 
> Hmm, good point. I don't think those were ever tested, TBH. We just assumed that they would be dealt with by the default python indexing stuff. I could add specific tests.

There are some tests in `test_lowlevel.py` for `get_individuals(id)` etc, but they only seem to test the integer bounds:

https://github.com/tskit-dev/tskit/blob/22d57fb080200cac60ae29970a4aadea2dcc86ec/python/tests/test_lowlevel.py#L2804

benjeffery:
Yeah, best to add a few. We've had weird things happen before!

hyanwong:
Best to add in low-level or highlevel? I guess in the same place as the existing tests (i.e. high-level, but in test_tables.py)

hyanwong:
> (i.e. high-level, but in test_tables.py)

Sorry, was getting my PRs in a twist. I meant highlevel.py

benjeffery:
Yep in `highlevel.py`.

hyanwong:
Do we have a simple ts for testing this sort of thing. No point doing it over all the `get_example_tree_sequences()`. Or I could do it with an empty TS.

hyanwong:
> Do we have a simple ts fixture for testing this sort of thing. No point doing it over all the `get_example_tree_sequences()`. Or I could do it with an empty TS.

Ah, got it: `simple_degree2_ts_fixture`

hyanwong:
Added, and raised an explicit error to match python indexing

benjeffery (approved):
Great, thanks @hyanwong!

hyanwong:
Thanks @benjeffery . You said:

> Yes, better to do the tables as a separate issue.

But I think we can already do `ts.tables.edges[-1]`, so I think we are all done here, no?

benjeffery:
Yep! Thanks!

PR #2479, opened 2022-08-26
Add a sort to ts.subset()

### LINKED ISSUE #2473: ts.subset() does not reorder edges as required for a valid tree sequence
E.g. if we use `subset` to reverse the node order, we get an error:

```
import msprime
ts = msprime.simulate(10)
ts.subset(np.arange(ts.num_nodes - 1, -1, -1))
```

We need to sort the edge table before returning the tree sequence. I guess we are happy to call `tables.sort()`, although it conceivably might have the side effect of changing the order of migrations.

Perhaps cleaner would be if we had a simple `tables.sort_edges()` function, but we don't.

hyanwong:
Fixes #2473. Docs adjusted to imply that migrations may be reordered (as a side effect of `sort()`). Simple test case used to check topological identity after reordering nodes (which would be too time consuming for large tree (sequences))

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1949,6 +1949,27 @@ class TestTreeSequence(HighLevelTestCase):
         with pytest.raises(_tskit.LibraryError):
             ts.simplify()
 
+    def test_subset_reverse_all_nodes(self):
+        ts = tskit.Tree.generate_comb(5).tree_sequence
+        assert np.all(ts.samples() == np.arange(ts.num_samples))
+        flipped_ids = np.flip(np.arange(ts.num_nodes))
+        new_ts = ts.subset(flipped_ids)
+        assert set(new_ts.samples()) == set(flipped_ids[np.arange(ts.num_samples)])
+        r1 = ts.first().rank()
+        r2 = new_ts.first().rank()
+        assert r1.shape == r2.shape
+        assert r1.label != r2.label
+
+    def test_subset_reverse_internal_nodes(self):
+        ts = tskit.Tree.generate_balanced(5).tree_sequence
+        internal_nodes = np.ones(ts.num_nodes, dtype=bool)
+        internal_nodes[ts.samples()] = False
+        node_ids = np.arange(ts.num_nodes)
+        node_ids[internal_nodes] = np.flip(node_ids[internal_nodes])
+        new_ts = ts.subset(node_ids)
+        assert np.any(new_ts.nodes_time != ts.nodes_time)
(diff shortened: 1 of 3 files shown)


python/tests/test_highlevel.py
```diff
@@ -1949,6 +1949,27 @@ def test_simplify_migrations_fails(self):
         with pytest.raises(_tskit.LibraryError):
             ts.simplify()
```
petrelharp, line 1951: perhaps these should go [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_tables.py#L4552)? And could call [`verify_subset`](https://github.com/tskit-dev/tskit/blob/main/python/tests/test_tables.py#L4614) there?

petrelharp (commented):
LGTM; one minor change to testing.

jeromekelleher (approved):
LGTM

python/tests/test_highlevel.py
hyanwong: Hmm, I could do, but I thought test_tables.py was specifically for testing TableCollection functions. The new functionality only applies to the ts.subset function (should it? Shouldn't the TableCollection function mirror the TreeSequence one exactly?). That was my rationale for dumping the tests in `test_highlevel.py`.

petrelharp: Hm, good point. Um, I guess I think we should call `sort` in TableCollection.subset, instead of just here?

hyanwong: Yep, that would probably be my preference.

There may be other functions which have a table and a tree sequence version, but where extra processing is required to make a valid tree sequence. It could be worth establishing a precedence for functions like this, if there is a use-case for not doing the extra processing in the tables version. We could have a parameter `make_valid_ts` that any such table function has, which is always set to `True` if called via the ts version.

It may not be that this isn't required for this particular function, so we could punt any decision down the line.

petrelharp: I thought about this, and I don't think we have a use case for doing `subset` followed by other processing steps that don't require sorted tables before `sort`ing (if we did, then it could be useful for efficiency to keep it out). So, let's not add the extra parameter.

benjeffery (approved):

jeromekelleher:
Were you ready to merge this @hyanwong and @petrelharp? Seems like the sort was intended to back into the tables version?

benjeffery:
Sorry, I misread the discussion as being about an additional change.

hyanwong:
It was meant to go into the tables version. I'll follow up with a minor PR to fix on Monday.

PR #2488, opened 2022-09-05
Remove description of individual table sort order

hyanwong:
We decided in https://github.com/tskit-dev/tskit/issues/1774 to remove the requirement, but the docs aren't all updated.


petrelharp (approved):
Good catch, thanks!

PR #2489, opened 2022-09-05
Add a sort to tables.subset()

### LINKED ISSUE #2487: Change sort in subset to tables implementation?
#2479 may have been merged prematurely.

As discussed:

_Originally posted by @petrelharp in https://github.com/tskit-dev/tskit/pull/2479#discussion_r956325113_

hyanwong:
Fixes #2487, see extra discussion in #2473

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,11 +2,6 @@
 [0.5.3] - 2022-XX-XX
 --------------------
 
-**Fixes**
-
- - ``ts.subset()`` produces valid tree sequences even if nodes are shuffled
-   out of time order (:user:`hyanwong`, :pr:`2479`, :issue:`2473`)
-
 **Features**
 
  - The ``ts.nodes`` method now takes an ``order`` parameter so that nodes
@@ -26,6 +21,11 @@
 
  - Accessor methods such as ts.edge(n) and ts.node(n) now allow negative
    indexes (:user:`hyanwong`, :pr:`2478`, :issue:`1008`)
+
+ - ``ts.subset()`` produces valid tree sequences even if nodes are shuffled
+   out of time order (:user:`hyanwong`, :pr:`2479`, :issue:`2473`), and the
+   same for ``tables.subset()`` (:user:`hyanwong`, :pr:`2489`). This involves
+   sorting the returned tables, potentially changing the returned edge order.
 
 **Performance improvements**
 
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -4510,19 +4510,15 @@ class TestSubsetTables:
         for k, p in zip(pops, subset.populations):
             pp = tables.populations[k]
             assert pp == p
-        edges = [
-            i
-            for i, e in enumerate(tables.edges)
+        # subset can reorder the edges: we need to check we have the same set
+        edges = {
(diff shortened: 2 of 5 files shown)


hyanwong:
@petrelharp: I seem to have opened up a can of worms. This is failing when comparing the tsutil python version of subset() ([here](https://github.com/tskit-dev/tskit/blob/70de29f1b7cb7e9df4ce9988d7ece3a6447c1dd7/python/tests/tsutil.py#L622)) to the new one, even though  I added a `tables.sort()` to both versions. Can you figure out why?

hyanwong:
> @petrelharp: I seem to have opened up a can of worms. This is failing on edge ordering when comparing the tsutil python version of subset() ([here](https://github.com/tskit-dev/tskit/blob/70de29f1b7cb7e9df4ce9988d7ece3a6447c1dd7/python/tests/tsutil.py#L622)) to the new one, even though I added a `tables.sort()` to both versions. Can you figure out why?

OK - fixed now. In `verify_subset` we now can't test that the edges are identical and in the same order as a subset of the full table, because sorting might change the edge order. So we have to make a set and test that instead. I've coded to test each element against the set, but it might be nicer to make another set from subset.edges and simply test set equality.

Ready for review.

python/tests/test_tables.py
```diff
@@ -4510,19 +4510,15 @@ def verify_subset(self, tables, nodes):
...
         assert subset.edges.num_rows == len(edges)
-        for k, e in zip(edges, subset.edges):
-            ee = tables.edges[k]
-            assert ee.left == e.left
-            assert ee.right == e.right
-            assert node_map[ee.parent] == e.parent
-            assert node_map[ee.child] == e.child
-            assert ee.metadata == e.metadata
+        for e in subset.edges:
+            assert e in edges
```
benjeffery, line 4521: This is a subtly looser check as now there can be edges in `edges` that aren't in `subset.edges`. I guess actually that could happen before as the zip would stop at the shortest. Adding a paranoid:
```
for e in edges:
    assert e in subset.edges
```
Would make it strict?

benjeffery (approved):
LGTM, one small paranoia point about the testing.

hyanwong:
Thanks @benjeffery - changed

python/CHANGELOG.rst
```diff
@@ -5,7 +5,8 @@
 **Fixes**
 
  - ``ts.subset()`` produces valid tree sequences even if nodes are shuffled
-   out of time order (:user:`hyanwong`, :pr:`2479`, :issue:`2473`)
+   out of time order (:user:`hyanwong`, :pr:`2479`, :issue:`2473`), and the
+   same for ``tables.subset()`` (:user:`hyanwong`, :pr:`2489`)
```
jeromekelleher, line 9: Is there something about the output of subset we need to note, since we're doing an extra sort?

jeromekelleher (approved):
LGTM, but I think we need to make it clear that there'll be some extra sorting in the tables.subset case, as this could potentially break some code.

python/CHANGELOG.rst
jeromekelleher: Yes, I think so.

hyanwong: OK, done.

PR #2494, opened 2022-09-06
Add sites argument to genotype matrix

szhan:
Part of #678

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5228,6 +5228,7 @@ class TreeSequence:
         self,
         *,
         samples=None,
+        sites=None,
         isolated_as_missing=None,
         alleles=None,
         impute_missing_data=None,
@@ -5256,6 +5257,8 @@ class TreeSequence:
 
         :param array_like samples: An array of node IDs for which to generate
             genotypes, or None for all sample nodes. Default: None.
+        :param array_like sites: An array of site IDs for which to generate
+            genotypes, or None for all sites. Default: None.
         :param bool isolated_as_missing: If True, the genotype value assigned to
             missing samples (i.e., isolated samples without mutations) is
             :data:`.MISSING_DATA` (-1). If False, missing samples will be
@@ -5293,9 +5296,11 @@ class TreeSequence:
         )
 
         num_samples = self.num_samples if samples is None else len(samples)
-        ret = np.zeros(shape=(self.num_sites, num_samples), dtype=np.int32)
+        num_sites = self.num_sites if sites is None else len(sites)
+        ret = np.zeros(shape=(num_sites, num_samples), dtype=np.int32)
 
-        for site_id in range(self.num_sites):
+        sites = range(self.num_sites) if sites is None else sites
+        for site_id in sites:
             variant.decode(site_id)
             ret[site_id, :] = variant.genotypes
 
--- a/python/tests/test_genotypes.py
+++ b/python/tests/test_genotypes.py
@@ -295,15 +295,17 @@ class TestVariantGenerator:
         assert len(variants) == 0
 
     @pytest.mark.parametrize("samples", [None, [1, 2], [2, 4], []])
-    @pytest.mark.parametrize("sites", [None, [2, 3], [1, 3], []])
+    @pytest.mark.parametrize("sites", [None, [1, 2], [1, 3], []])
     def test_genotype_matrix(self, samples, sites):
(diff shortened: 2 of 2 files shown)


szhan:
What should the output be when `sites=[]`?

szhan:
Also, what if there is an invalid site ID among a list of valid site IDs?

python/tskit/trees.py
```diff
@@ -5279,14 +5289,21 @@ def genotype_matrix(
...
+        num_samples = self.num_samples if samples is None else len(samples)
+        num_sites = self.num_sites if sites is None else len(sites)
+        ret = np.zeros(shape=(num_sites, num_samples), dtype=np.int32)
+
+        sites = range(self.num_sites) if sites is None else sites
 
-        for site_id in range(self.num_sites):
-            variant.decode(site_id)
-            ret[site_id, :] = variant.genotypes
+        for i in range(len(sites)):
```
benjeffery, line 5304: Why not just for `for site_id in sites`?

szhan:
> 

I think I'm going to add bad sample IDs in a follow-up PR. I should have done that in the other PR. Sorry.

jeromekelleher:
Is this what we want or should we use ``left`` and ``right`` arguments like we do for the ``variants`` method? There's some nasty performance gotchas involved in using arbitrary list of sites, which can be in any order.

Implementing ``left`` and ``right`` would be easy, since we can just delegate to ``variants``.

benjeffery:
Seems a shame to do `left`/`right` when a key use case is to output a subset of sites that meet some criteria. We could make it an error to specify non-monotonicly-increasing site positions to avoid the pathological performance cases?

benjeffery:
> Could do - but another key use-case is also likely to be getting a slice from `left` to `right`. So, we could do that first, and then add `sites` as a mutually exclusive alternative?

Yep, sounds like a good way to proceed.

jeromekelleher:
#678 is too broad I think - we should create smaller more focused issues that can be ticked off in a single PR. We can open an issue to track the ``sites`` argument, but lets implement the easy thing first (left and right). If someone really wants ``sites`` in particular they can comment on the new issue.

hyanwong:
> If someone really wants `sites` in particular they can comment on the new issue.

There's relevant discussion at https://github.com/tskit-dev/msprime/discussions/1685 I think.

hyanwong:
Re the API, I wonder if we want `left and `right` which makes sense to me in an iterator like `variants()`, or an interval (or set of intervals) like `keep_intervals()` instead? If the latter, to make it clear you could also accept an `Interval` object as well:

```
ts.genotype_matrix(interval=[0, 10])  # or (interval=tskit.Interval(0, 10)) for clarity
```

to roughly match

```
ts.keep_intervals([[0, 10], [20,30]])  # or ([tskit.Interval(0, 10), tskit.Interval(20, 30)]) for clarity
```

jeromekelleher:
I think it's simpler to follow the same form as ``variants`` - particularly since it's by far the easiest to implement. I don't see any reason why the ``intervals`` arguments wouldn't also apply to variants, and could/should be implemented there first.

hyanwong:
> I think it's simpler to follow the same form as `variants` - particularly since it's by far the easiest to implement. I don't see any reason why the `intervals` arguments wouldn't also apply to variants, and could/should be implemented there first.

Well, it seems reasonably natural to me to think of an iterator starting at a certain point into the sequence and stopping after a while. I find it a bit more of a jump to think of this in the context of a function that returns a whole matrix in one go. similarly, it seems a bit weird to me to have a *set* of multiple intervals passed to an iterator, but it makes perfect sense when passed to something that returns a single object, like `keep_intervals` or `genotype_matrix`. But I'm not really bothered either way: I just thought I should comment on what seemed natural to me.

hyanwong:
Note that the intervals thing is equivalent to `ts.keep_intervals([[a, b]]).genotype_matrix()`, and the sites thing is equivalent to `ts.delete_sites(sites_to_del)`, so perhaps all we need is `ts.keep_sites()` rather than adding more parameters?

benjeffery:
@szhan Shall we shelve this for now? Adding the samples arg was some great work, but there are open questions here I don't think we want to get bogged down in right now.

jeromekelleher:
I agree, let's close this and #678 (optionally opening another issue to track the adding the ``sites`` argument)

PR #2508, opened 2022-09-16
Pin jsonschema

### LINKED ISSUE #2481: tskit breaks with `jsonschema>=4.0.0`
We currently specify `jsonschema>=3.0.0` in `setup.cfg`. When run with `4.9.0` the schema validation test fail as they don't raise errors when schemas are invalid (for specific errors).

Two options to fix:
- Pin `jsonschema` in `setup.cfg`. Not ideal.
- Fix the code for version 4.10

I'm not sure how hard the later option is - will look into it.

benjeffery:
Fixes #2481


benjeffery:
Testing things on CI, will add pin to setup.cfg if things work. (Only as a temp measure, would rather fix for the new version)

benjeffery:
Failing on windows! Will try moving the pin earlier in the process.

jeromekelleher (approved):

PR #2523, opened 2022-09-28
Map_mutations example shows how to map parent IDs

hyanwong:
Using map_mutations to add stuff to a tree sequence requires quite a lot of tricks. I think it's worth giving the simplest of examples I can imagine here.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2745,8 +2745,27 @@ class Tree:
             will result in these missing observations being imputed to the
             most parsimonious state.
 
+        Because the ``parent`` in the returned list of mutations refers to the index
+        in that list, if you are adding mutations to an existing tree sequence, you
+        will need to maintain a map of list IDs to the newly added mutations, for
+        instance::
+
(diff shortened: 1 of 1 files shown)


benjeffery (approved):
Nice to have this in the docs as it isn't obvious.
I always do it this way to use a list rather than a dict and make the `NULL` exception more obvious:
```python
            ancestral, mutations = tree.map_mutations(genotype, ('A', 'C', 'G', 'T'))
            site_id = tables.sites.add_row(position=pos, ancestral_state=ancestral, metadata=None)
            mutation_ids = []
            for i, m in enumerate(mutations):
                mutation_id = tables.mutations.append(m.replace(
                    site=site_id,
                    parent= tskit.NULL if m.parent == tskit.NULL else mutation_ids[m.parent]
                ))
                mutation_ids.append(mutation_id)
```
But I really don't mind which you're more happy with.

jeromekelleher (approved):

benjeffery:
@mergifyio refresh

PR #2531, opened 2022-09-29
Access the left and right of an edge as an interval

hyanwong:
Is this useful at all, or just an extra maintenance burden?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -312,6 +312,16 @@ class Edge(util.Dataclass):
         """
         return self.right - self.left
 
+    @property
+    def interval(self):
+        """
+        Returns the left and right positions of this edge as an :class:`Interval`
+
+        :return: The interval covered by this edge.
+        :rtype: :class:`Interval`
+        """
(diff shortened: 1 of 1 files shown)


benjeffery:
Don't think this is a bad idea - what code do you have where this is good and what is the alternative?

hyanwong:
> Don't think this is a bad idea - what code do you have where this is good and what is the alternative?

I needed to quickly check whether a tree covered a whole edge. `tree.interval == edge.interval` seems more comprehensible than `tree.right == edge.right and tree.left == edge.left`, but it's not exactly a priority!

benjeffery:
Oh! That's a nice example. Seems like a good addition.

benjeffery:
Converting this to a draft as it needs tests.

hyanwong:
Yeah, meant to open as a draft, sorry.

benjeffery:
@hyanwong Is this still something you want to try and merge?

hyanwong:
I guess I might as well. The test will be trivial.

PR #2533, opened 2022-09-29
Fix mrca calculation for multiple roots

### LINKED ISSUE #2521: Clarifying MRCA for multiple roots
If a tree sequence has multiple roots, and we call `ts.mrca(ts.samples())`, we could either return a list of MRCAs or error out. We should specify in the docs which we do. At the moment it seems to return "Out of bounds":

```python
import tskit
ts = tskit.Tree.generate_balanced(10).tree_sequence
ts = ts.delete_intervals([ts.first().interval])
ts.first().mrca(*ts.samples())
```

```
-> 1007         return functools.reduce(self._ll_tree.get_mrca, args)
   1008 
   1009     def get_tmrca(self, u, v):

ValueError: Node index out of bounds
```

benjeffery:
FIxes #2521

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -7,6 +7,9 @@
  - The ``Variant`` object can now be initialized with 64 bit numpy ints as
    returned e.g. from np.where (:user:`hyanwong`, :pr:`2518`, :issue:`2514`)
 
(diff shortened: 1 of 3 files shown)


hyanwong:
Do we need to document what happens if there are multiple MRCAs?

Issue #2248, opened 2022-05-10
Support for python 3.11


benjeffery:
3.11 betas are coming out, with this version having some breakage in the C API.

benjeffery:
I've had a go at this but can't get numpy to build due to https://github.com/numpy/numpy/issues/21422 so will wait for things to settle down.

jeromekelleher:
Have you some pointers to what we can expect in terms of C API breakage @benjeffery? Nice to know what's coming up.

benjeffery:
The good news is "CPython 3.11 is on average [25% faster](https://github.com/faster-cpython/ideas/blob/main/main-vs-310.rst) than CPython 3.10 when measured with the [pyperformance](https://github.com/python/pyperformance) benchmark suite"!!

The C API changes are at https://docs.python.org/3.11/whatsnew/3.11.html#id6 None of it looks too bad.

jeromekelleher:
None of it obviously affects us from a quick glance, so hopefully we'll get through without changes :crossed_fingers:

benjeffery:
3.11 is released this month, so we should have another look at this.

benjeffery:
Bringing this forward to the next release milestone.

jeromekelleher:
Already have conda builds on conda-forge btw, e.g.: https://anaconda.org/conda-forge/tskit/files

benjeffery:
> Already have conda builds on conda-forge btw, e.g.: https://anaconda.org/conda-forge/tskit/files

Yeah, seeing that made me bump this up!

PR #2598, opened 2022-10-05
Change error messages for Python 3.11

benjeffery:
Part of #2248 
Our test suite passes with Python 3.11 if these small changes are applied.
However, we can't yet add it to CI as one of our test dependencies (`lshmm`) depends on `numba` which doesn't yet have 3.11 support. (see https://github.com/numba/numba/issues/8304)

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -2725,9 +2725,9 @@ class TestTreeSequence(HighLevelTestCase):
         ts = ts_fixture
         a = getattr(ts, array)
         assert isinstance(a, np.ndarray)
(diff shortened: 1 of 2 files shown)


benjeffery:
> Could just delete these strings? Probably sufficient to catch an AttributeError tbh.

Well one would hope that they aren't going to change again! I'm happy with the "or" clause.

jeromekelleher:
People will wonder in 10 years time what the first strings were for :wink:

benjeffery:
Cool, I removed the matches.

jeromekelleher:
Sorry, thought I'd approved already

Issue #2605, opened 2022-10-17
Add ``filter_nodes`` argument to simplify


jeromekelleher:
As discussed in various places (#2604 + elsewhere?) it would be good to have a ``filter_nodes=False`` argument to simplify.

This may be very simple to implement, and just be case of calling ``record_node`` for all nodes [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/simplify.py#L144) and not just the samples.

hyanwong:
Minor additional discussion at https://github.com/tskit-dev/what-is-an-arg-paper/discussions/92

PR #2606, opened 2022-10-17
filter_nodes for simplify

jeromekelleher:
Here's a quick first pass for #2605.

I think it basically works, but would need more testing and a bit of implementation polish (mainly just properly working out whether a node is a sample).

The main limitation is that the input samples must be the original tree sequence samples. We could probably relax that a bit, I guess (make it a subset?) but I didn't think it through.

@hyanwong - do you want to try this out on your applications and maybe come up with some tests?

### DIFF
--- a/python/tests/simplify.py
+++ b/python/tests/simplify.py
@@ -111,6 +111,7 @@ class Simplifier:
         keep_unary=False,
         keep_unary_in_individuals=False,
         keep_input_roots=False,
+        filter_nodes=True,
     ):
         self.ts = ts
         self.n = len(sample)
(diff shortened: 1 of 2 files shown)


hyanwong:
I think we decided that we should allow any set of nodes to be specified as samples, although I'd be happy simply to restrict it to `ts.samples()`, at least for the time being.

If we did allow different sample sets with `filter_nodes=False`, I think we said that the order of samples passed to the function should not be important (so that nodes never change their ID). Alternatively, we could simply error out if the node IDs provided as `samples` are not in ascending order.

hyanwong:
I wasn't sure how to add extra functionality to this PR (or whether I have permission to), so I have opened up another PR with the commit in this PR as the first commit and a few tests added as extra commits: #2619. Hence I'm closing this

Issue #2608, opened 2022-10-19
Switch `filter_X` arguments in simplify to default `None` treated as True.


hyanwong:
As discussed with Jerome, this will allow easier implementation of https://github.com/tskit-dev/tskit/issues/2607 and can be done before #2605

PR #2609, opened 2022-10-19
Simplify call sig uses `None` for filter_XX args

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/2608

I presume this doesn't need extra tests. Does it require an addition to the changelog?

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,4 +1,16 @@
 --------------------
+[0.5.4] - 2022-XX-XX
+--------------------
+
+**Breaking Changes**
+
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I think this is potentially breaking, if people were depending on the Falsey interpretation of None? Although maybe not,as the Python C layer is probably interpreting these as ints and so it wouldn't allow None as a parameter.

Good to double-check this and document the new behaviour in the CHANGELOG in any case.

hyanwong:
Previously passing ``None`` resulted in 

```
TypeError: an integer is required (got type NoneType)
```

Test added which now passes when ``None`` is provided

jeromekelleher (approved):
LGTM.

PR #2612, opened 2022-10-24
Get min time and max time of all nodes and mutations

### LINKED ISSUE #2271: Add TreeSequence.max_time, min_time attributes
There are many situations where it's handy to know what the max and min times are, and it would be nice if we could avoid having to do a full lookup of the ``np.max(ts.tables.nodes.time)`` to do it.

I think we can reasonably add these as attributes of the tsk_treeseq_t struct in C, filling them in during startup.

We can make them available to Python then easily, too.

szhan:
Fixes #2271 

# PR Checklist:

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -1158,6 +1158,108 @@ test_simplest_discrete_time(void)
 }
 
 static void
+test_simplest_min_time(void)
+{
+    int ret;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+
+    const char *nodes = "1  0.1 0  -1\n"
+                        "1  0.1 0  -1\n"
+                        "1  0.1 0  -1\n"
+                        "0  1 0  -1\n"
+                        "0  2 0  -1\n";
+    const char *edges = "0  2   3   0,1\n"
+                        "0  2   4   2,3\n";
+    const char *sites = "0  0\n"
+                        "1  0\n";
+    const char *mutations = "0  2   1   -1  0.5\n"
+                            "1  3   1   -1  1.5\n";
+
+    tsk_treeseq_from_text(&ts, 2, nodes, edges, NULL, sites, mutations, NULL, NULL, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_min_time(&ts), 0.1, 1E-6);
+
+    ret = tsk_table_collection_copy(ts.tables, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_treeseq_free(&ts);
+
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_min_time(&ts), 0.1, 1E-6);
+    tsk_treeseq_free(&ts);
+
+    /* Setting mutation times to unknown should have no effect on min time. */
+    tables.mutations.time[0] = TSK_UNKNOWN_TIME;
+    tables.mutations.time[1] = TSK_UNKNOWN_TIME;
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_min_time(&ts), 0.1, 1E-6);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[0] = 0.5;
+    tables.mutations.time[1] = 1.5;
+
+    /* An empty tree sequence has infinity min time. */
+    tsk_table_collection_clear(&tables, 0);
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_min_time(&ts), INFINITY);
+    tsk_treeseq_free(&ts);
+
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_simplest_max_time(void)
+{
+    int ret;
+    tsk_treeseq_t ts;
+    tsk_table_collection_t tables;
+
+    const char *nodes = "1  0.1 0  -1\n"
+                        "1  0.1 0  -1\n"
+                        "1  0.1 0  -1\n"
+                        "0  1 0  -1\n"
+                        "0  2 0  -1\n";
+    const char *edges = "0  2   3   0,1\n"
+                        "0  2   4   2,3\n";
+    const char *sites = "0  0\n"
+                        "1  0\n";
+    const char *mutations = "0  2   1   -1  0.5\n"
+                            "1  3   1   -1  1.5\n";
+
+    tsk_treeseq_from_text(&ts, 2, nodes, edges, NULL, sites, mutations, NULL, NULL, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_max_time(&ts), 2.0, 1E-6);
+
+    ret = tsk_table_collection_copy(ts.tables, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tsk_treeseq_free(&ts);
+
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_max_time(&ts), 2.0, 1E-6);
+    tsk_treeseq_free(&ts);
+
+    /* Setting mutation times to unknown should have no effect on max time. */
+    tables.mutations.time[0] = TSK_UNKNOWN_TIME;
+    tables.mutations.time[1] = TSK_UNKNOWN_TIME;
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_DOUBLE_EQUAL(tsk_treeseq_get_max_time(&ts), 2.0, 1E-6);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[0] = 0.5;
+    tables.mutations.time[1] = 1.5;
+
+    /* An empty tree sequence has negative infinity max time. */
+    tsk_table_collection_clear(&tables, 0);
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_max_time(&ts), -INFINITY);
+    tsk_treeseq_free(&ts);
+
+    tsk_table_collection_free(&tables);
+}
+
+static void
 test_simplest_records(void)
 {
     const char *nodes = "1  0   0\n"
@@ -7875,6 +7977,8 @@ main(int argc, char **argv)
         /* simplest example tests */
         { "test_simplest_discrete_genome", test_simplest_discrete_genome },
         { "test_simplest_discrete_time", test_simplest_discrete_time },
+        { "test_simplest_min_time", test_simplest_min_time },
+        { "test_simplest_max_time", test_simplest_max_time },
         { "test_simplest_records", test_simplest_records },
         { "test_simplest_nonbinary_records", test_simplest_nonbinary_records },
         { "test_simplest_unary_records", test_simplest_unary_records },
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -363,6 +363,13 @@ tsk_treeseq_init_mutations(tsk_treeseq_t *self)
             = discrete_times && (is_discrete(time[j]) || tsk_is_unknown_time(time[j]));
     }
     self->discrete_time = self->discrete_time && discrete_times;
+
+    for (j = 0; j < num_mutations; j++) {
+        if (!tsk_is_unknown_time(time[j])) {
+            self->min_time = TSK_MIN(self->min_time, time[j]);
+            self->max_time = TSK_MAX(self->max_time, time[j]);
+        }
+    }
 }
 
 static int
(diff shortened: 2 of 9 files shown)


szhan:
I just installed `clang-format`, but it is not version 6.0. The online docs `(https://tskit.dev/tskit/docs/stable/development.html#sec-development-c)` says that `"C code is formatted using [clang-format](https://clang.llvm.org/docs/ClangFormat.html) with a custom configuration and version 6.0`. Do I have to use version 6.0?

c/tskit/trees.c
```diff
@@ -363,6 +363,13 @@ tsk_treeseq_init_mutations(tsk_treeseq_t *self)
             = discrete_times && (is_discrete(time[j]) || tsk_is_unknown_time(time[j]));
     }
     self->discrete_time = self->discrete_time && discrete_times;
+
+    for (j = 0; j < num_mutations; j++) {
+        if (!tsk_is_unknown_time(time[j])) {
+            self->min_time = TSK_MIN(self->min_time, time[j]);
+            self->max_time = TSK_MIN(self->max_time, time[j]);
```
jeromekelleher, line 370: TSK_MAX

c/tskit/trees.c
```diff
@@ -405,6 +412,13 @@ tsk_treeseq_init_nodes(tsk_treeseq_t *self)
             = discrete_times && (is_discrete(time[j]) || tsk_is_unknown_time(time[j]));
     }
     self->discrete_time = self->discrete_time && discrete_times;
+
+    for (j = 0; j < num_nodes; j++) {
+        if (!tsk_is_unknown_time(time[j])) {
+            self->min_time = TSK_MIN(self->min_time, time[j]);
+            self->max_time = TSK_MIN(self->max_time, time[j]);
```
jeromekelleher, line 419: MAX

c/tskit/trees.c
```diff
@@ -452,6 +466,8 @@ tsk_treeseq_init(
     self->num_trees = (tsk_size_t) num_trees;
     self->discrete_genome = true;
     self->discrete_time = true;
+    self->min_time = -INFINITY;
```
jeromekelleher, line 469: I think you need to start with min = positive inf and max = negative inf for this to work.

jeromekelleher (commented):
The logic doesn't quite work, it needs some tests.

A golden rule is to always test C code from the C tests before doing any further up the API stack.

I would add a could of simple tests to ``test_trees.c`` first before doing anything else.

jeromekelleher:
Yeah, you gotta use clang-format 6 for complicated reasons unfortunately.

The [pip package](https://pypi.org/project/clang-format/) is very helpful for this.

szhan:
I'm trying to create some simple cases for testing. The columns in the strings for creating a table collection in `test_trees.c` correspond to the arguments in `add_row` and `set_columns` in the Python API, right? For example, in

```
    const char *nodes = "1  0   0\n"
                        "1  0   0\n"
                        "1  0   0\n"
                        "0  0   0\n"
                        "0  0   0";
```

the first row means `flags=1`, `time=0`, and `population=0` when setting the node table, right?

jeromekelleher:
Yes, that's right. The columns are almost the same as the text format, but with some extra little quirks. Probably easiest to talk you through this in person tbh.

szhan:
I want to add tests to show that the times in a mutation table are correctly handled. Having mutation times does not change min time and max time, because the youngest mutation cannot be younger than the youngest node and the oldest mutation cannot be older than the oldest node, right? Unless the leaves get pruned or the root gets shaved off.

szhan:
For the tests, I'm tempted to use the test cases that are already in `c/tests/testlib.c`. But I should try to come up with simpler cases, right?

jeromekelleher:
> Having mutation times does not change min time and max time, because the youngest mutation cannot be younger than the youngest node and the oldest mutation cannot be older than the oldest node, right? Unless the leaves get pruned or the root gets shaved off.

It can in pathological situations - you can have "dangling" mutations that are above nodes but not on an edge. We should probably have ruled them out of the data model in retrospect, but they are valid input at the moment.

jeromekelleher:
> For the tests, I'm tempted to use the test cases that are already in c/tests/testlib.c. But I should try to come up with simpler cases, right?

Can definitely reuse the examples like single_tree_ex and paper_ex. Might need to hand craft something for the mutation above a root thing, but do as little new stuff as you can get away with.

A good start would be to add an assert or two to the test_single_tree_good_records test.

szhan:
> > Having mutation times does not change min time and max time, because the youngest mutation cannot be younger than the youngest node and the oldest mutation cannot be older than the oldest node, right? Unless the leaves get pruned or the root gets shaved off.
> 
> It can in pathological situations - you can have "dangling" mutations that are above nodes but not on an edge. We should probably have ruled them out of the data model in retrospect, but they are valid input at the moment.

So, should we add tests to see that having mutation times do affect calculations of min and max times? But if pathological mutation times are considered as invalid in the future (?), then these tests will crap out.

jeromekelleher:
> So, should we add tests to see that having mutation times do affect calculations of min and max times? But if pathological mutation times are considered as invalid in the future (?), then these tests will crap out.

Definitely. Tests crapping out when you change things is a good thing!

jeromekelleher:
But, we don't have to get *too* detailed for the C tests. Really we just want to cover all code paths there - tricky stuff like this is probably best tested in Python when we're sure the basics works.

c/tests/test_trees.c
```diff
@@ -1157,6 +1157,134 @@ test_simplest_discrete_time(void)
...
+    tables.mutations.time[1] = TSK_UNKNOWN_TIME;
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_min_time(&ts), 0.1);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[0] = 1.5;
+    tables.mutations.time[1] = 0.5;
+
+    /* A bad mutation younger than the youngest node. */
+    tables.mutations.time[1] = 0.0;
```
jeromekelleher, line 1207: Don't need to bother with this - the values within the tree sequence are undefined if there's an error

c/tests/test_trees.c
```diff
@@ -1157,6 +1157,134 @@ test_simplest_discrete_time(void)
...
+    tables.mutations.time[1] = TSK_UNKNOWN_TIME;
+    ret = tsk_treeseq_init(&ts, &tables, TSK_TS_INIT_BUILD_INDEXES);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(tsk_treeseq_get_max_time(&ts), 2.0);
+    tsk_treeseq_free(&ts);
+    tables.mutations.time[0] = 1.5;
+    tables.mutations.time[1] = 0.5;
+
+    /* A bad mutation older than the root. */
+    tables.mutations.time[0] = 2.5;
```
jeromekelleher, line 1271: Likewise here, no need to test the values after an error condition.

jeromekelleher (commented):
Looks great!

jeromekelleher:
Doesn't look like the tests are actually passing - have you run these locally with e.g. ``./build/test_trees test_simplest_min_time``?

I think it would be best to separate out the Python C API stuff into another PR - let's get the C side working well first.

szhan:
> 

I'm getting this error `CU_ASSERT_EQUAL_FATAL(ret,0)`. Is it saying that a tree sequence is not valid? Am I getting the text format correct? I added the column names to help clarify my understanding.

c/tests/test_trees.c
jeromekelleher: Well I meant there's no need to test this case at all - if it's an error then the ``min_time`` is undefined.

jeromekelleher:
Text format looks OK, but I don't think you can add column headers. I haven't pored over the details though.

Are you running this OK locally?

szhan:
I took out the column names, and the test still fails. This assertion is failing, so it seems that the text format may not be correct.

```
ret = tsk_table_collection_copy(ts.tables, &tables, 0);
tsk_treeseq_free(&ts);
```

Here are the columns I think are in the tables:
```
node table
is_sample    individual  time

edge table
left right parent    child

site table
position ancestral_state

mutation table
site node  derived_state parent    time
```

benjeffery:
> I'm getting this error `CU_ASSERT_EQUAL_FATAL(ret,0)`. Is it saying that a tree sequence is not valid? Am I getting the text format correct? I added the column names to help clarify my understanding.

Probably helpful if I explain my process for fixing things like this (Sorry I haven't been in this week @szhan!)

First I set a conditional breakpoint on `testlib.c:788` (as reported by the output of `test_trees`) with the condition being `ret !=0`,
when the breakpoint fires checking the value of `ret` gives `-307`. In `core.h` this is the value for `TSK_ERR_BAD_EDGE_INTERVAL`. Fixing the edge interval to be 0-1 results in the same error. Removing the text column headers changes the error to `-306` which is `TSK_ERR_BAD_NODE_TIME_ORDERING` I'll leave that as an exercise to the reader? ;)

jeromekelleher:
(The text format parsing is super unhelpful @szhan as it's just a development tool and not part of the library. It's never been worth putting effort into making better.)

szhan:
Yay, made some progress! When `tsk_treeseq_init` does not return 0 (say, `TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE` instead), it seems that it is not updating `max_time`.

jeromekelleher:
> When tsk_treeseq_init does not return 0 (say, TSK_ERR_MUTATION_TIME_OLDER_THAN_PARENT_NODE instead), it seems that it is not updating max_time.

When tsk_treeseq_init return a non-zero value it's reporting an error and then **all bets are off** - the only thing that you can validly do after that is call ``tsk_treeseq_free(ts)``. No other operation is defined or supported. So, if the input is not a well formed tree sequence (one of these errors are detected) then the max/min time is undefined. 

Do you see what I mean? There's no point (worse, it's an actively bad idea) to write tests here that check the state of the tree sequence object after ``tsk_treeseq_init`` has returned nonzero status.

benjeffery:
Looks like you're just failing 32bit now right? Shall I check it?

c/tests/test_trees.c
szhan: Ah, I didn't see these comments!

benjeffery:
I've set up the 32bit docker image and can recreate the failure.
These are failing:
```
    1. ../c/tests/test_trees.c:1180  - CU_ASSERT_EQUAL(tsk_treeseq_get_min_time(&ts),0.1)
    2. ../c/tests/test_trees.c:1188  - CU_ASSERT_EQUAL(tsk_treeseq_get_min_time(&ts),0.1)
    3. ../c/tests/test_trees.c:1196  - CU_ASSERT_EQUAL(tsk_treeseq_get_min_time(&ts),0.1)
```
This is because `0.1` is not 0.1 (blame floating point!)
Use `CU_ASSERT_DOUBLE_EQUAL` to fix this.

jeromekelleher:
Another way to do this is to use 0.125 or something that's a round number in base 2. That'll compare exactly even on old FPUs.

szhan:
I hope that the C API is okay now. I'm adding `TreeSequence_get_min_time()` and `TreeSequence_get_max_time()` in `_tskitmodule.c`, as well as some tests in `test_lowlevel.py`. I was thinking to use `LowLevelClass.get_example_tree_sequence()` to generate an example ts, but it uses `msprime` so the ts max time may be different on different OS. Can I just manually create an example ts via tables? Like

```
tables = tskit.TableCollection(sequence_length=2)
tables.nodes.add_row(flags=1, time=0.1)
tables.nodes.add_row(flags=1, time=0.1)
tables.nodes.add_row(flags=1, time=0.1)
tables.nodes.add_row(flags=0, time=1)
tables.nodes.add_row(flags=0, time=2)
tables.edges.add_row(left=0, right=2, parent=3, child=0)
tables.edges.add_row(left=0, right=2, parent=3, child=1)
tables.edges.add_row(left=0, right=2, parent=4, child=2)
tables.edges.add_row(left=0, right=2, parent=4, child=3)
tables.sites.add_row(position=0, ancestral_state='0')
tables.sites.add_row(position=1, ancestral_state='0')
tables.mutations.add_row(site=0, node=2, derived_state='1', time=0.5)
tables.mutations.add_row(site=1, node=3, derived_state='1', time=1.5)
ts = tables.tree_sequence()
```

szhan:
Okay, checkpoint! The C API part has passed all the CI checks.

szhan:
I've added a function under `LowLevelTestCase` to create a predefined tree sequence without using `msprime` in order to test the low-level Python API. I'm going to add such a function to test the high-level Python API as well. I'm not sure if this is a good idea, partly because the snippet that creates the tables is repeated in both `test_lowlevel.py` and `test_highlevel.py`.

szhan:
I hope this is it, unless there should be more tests to add.

jeromekelleher (commented):
Tests still need some work I think. I think you can cut down the lowlevel to almost nothing, and I've made a suggestion for organising the high-level tests so that they are more understandable.

python/tests/test_highlevel.py
```diff
@@ -1441,6 +1496,30 @@ def is_discrete(a):
...
+
+    @pytest.mark.parametrize(
+        "ts,expected",
+        [
+            (ts, expected)
+            for ts, expected in zip(
+                get_preset_example_tree_sequences(), [2.0, 2.0, float("-Inf")]
+            )
+        ],
+    )
```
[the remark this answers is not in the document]
szhan: Ah, okay. I thought it'd be good to put the test after the test for `discrete_time`.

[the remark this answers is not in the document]
jeromekelleher: It's confusing all right - the tests in that file are a mishmash of things that are super old, when there was only a handful of things that needed to be tested. Now there's much more and we're using different test infrastructure and have changed the basic style of tests as well.

c/tskit/trees.h
```diff
@@ -736,6 +739,30 @@ then this flag will be true.
 */
 bool tsk_treeseq_get_discrete_genome(const tsk_treeseq_t *self);
 
+/**
```
jeromekelleher, line 742: Just realised we need to add this to the C CHANGELOG because we have added to the API.

szhan:
It is strange that sometimes the ubuntu tests fail while the mac/windows tests pass...

szhan:
I'm not sure how to deal with the missed test coverage in python/tskit/trees.py#L4307 and python/tskit/trees.py#L4321.

python/tests/test_lowlevel.py
```diff
@@ -138,6 +138,11 @@ def get_example_migration_tree_sequence(self):
         )
         return ts.ll_tree_sequence
 
+    def get_empty_tree_sequence(self):
+        tables = tskit.TableCollection(sequence_length=2)
+        ts = tables.tree_sequence()
+        return ts.ll_tree_sequence
+
```
benjeffery, line 145: This looks unused?

benjeffery (commented):
LGTM - you have unused code in the low level tests though.

python/tests/test_lowlevel.py
szhan: Ah, thanks for catching this! I neglected to remove this after reducing the low level tests.

benjeffery (approved):

Issue #2181, opened 2022-04-05
Variant method to get all alleles as a string


jeromekelleher:
There are often times when we'd like to get the genotypes for a given site as the actual alleles rather than the indexes into the alleles list. (This is what #2168 was about, I assume.)

We can add this as a method to the Variant class easily enough. We can just raise an error if there are any non-1 length alleles in there.

What do we call it? 

This is the replacement also for the old ``as_bytes`` option, which basically did the same thing.

It might look something like
```python
def genotypes_as_alleles(self): -> str
     """
     Returns a string s in which s[j] is the value of ``var.alleles[var.genotypes[j]]``. Raises an error if 
     all alleles are not of length 1.
     """
```

jeromekelleher:
Note that some of the changes made to the test suite for #2172 can be made simpler using this functionality.

benjeffery:
Might it be nice to have a general method:
`tskit.genotypes_as_alleles(alleles, genotypes)`, `tskit.genotypes_as_ragged_alleles(alleles, genotypes)` then
`Variant.genotypes_as_alleles` would use this?

jeromekelleher:
Good idea - I guess the distinction is whether you want to returned value to be a numpy array or a string. There's definitely value in getting a numpy array back too.

jeromekelleher:
@hyanwong - any thoughts on names here? I feel like this is quite a handy feature and we should support "encoding" the variant data in string or numpy format. We definitely shouldn't call the it "encode" though, as we already have the "decode" method which does something quite different.

hyanwong:
The reason I wanted this was to be able to compare sites from different encodings of the same tree sequence (which therefore could have alleles in a different order. I "hacked" around this in https://github.com/tskit-dev/tsinfer/issues/652#issuecomment-1084693017 by using `ts.genotype_matrix(alleles=tskit.ALLELES_ACGT)`, but it isn't an obvious thing to do, and I can imagine it catching out other people who want to compare the sites-by-samples matrix, or do sitewise comparisons. So I think this is handy, yes. Another possibility for my use case would be the ability to compare the variant genotypes with each other, so that you could do (say) the equivalent of `ts1.variant(4).identical_genotypes(ts2.variant(4))` or similar notation.

[...]

PR #2617, opened 2022-10-30
Add `states()` method to return actual genotype states as characters

hyanwong:
## Description

Adds `variant.genotype_values()`. See https://github.com/tskit-dev/tsinfer/issues/739

# PR Checklist:

### DIFF
--- a/python/tskit/genotypes.py
+++ b/python/tskit/genotypes.py
@@ -245,6 +245,35 @@ class Variant:
         variant_copy._ll_variant = self._ll_variant.restricted_copy()
         return variant_copy
 
+    def states(self, missing_data_string=None) -> np.ndarray:
+        """
+        Returns the allelic states at this site as an array of strings.
+
+        .. warning::
+            Using this method is inefficient compared to working with the
+            underlying integer representation of genotypes as returned by
+            the :attr:`~Variant.genotypes` property.
+
+        :param str missing_data_string: A string that will be used to represent missing
+            data. If any normal allele contains this character, an error is raised.
+            Default: `None`, treated as `'N'`.
+        :return: An numpy array of strings of length ``num_sites``.
+        """
+        if missing_data_string is None:
+            missing_data_string = "N"
+        elif not isinstance(missing_data_string, str):
+            # Must explicitly test here, otherwise we output a numpy object array
+            raise ValueError("Missing data string is not a string")
+        alleles = self.alleles
+        if alleles[-1] is None:
+            if missing_data_string in alleles:
+                raise ValueError(
+                    "An existing allele is equal to the "
+                    f"missing data string '{missing_data_string}'"
+                )
+            alleles = alleles[:-1] + (missing_data_string,)
+        return np.array(alleles)[self.genotypes]
+
     def counts(self) -> typing.Counter[str | None]:
         """
         Returns a :class:`python:collections.Counter` object providing counts for each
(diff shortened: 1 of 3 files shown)


python/tskit/genotypes.py
```diff
@@ -245,6 +245,19 @@ def copy(self) -> Variant:
         variant_copy._ll_variant = self._ll_variant.restricted_copy()
         return variant_copy
 
+    def genotype_values(self) -> np.ndarray:
+        """
+        Returns the genotypes at this site as an numpy array of strings (if
+        there is no missing data) or objects (if the genotypes contain missing data,
+        in which case some elements will be equal to ``None``),
+        rather than an array of integer indexes. Note that this is inefficient
```
benjeffery, line 253: Can we render this in a warning box?

hyanwong: Yes, good idea

hyanwong: Done. Is the method name OK, do you think?

hyanwong:
I'm not sure why this is failing CI: everything is covered AFAICS.

Ready to merge, I reckon, pending approval of the name of the function.

jeromekelleher (commented):
LGTM - not sure about the name though.

jeromekelleher:
Also not clear that an Object array is what people would want. Would it be better to have a "missing data symbol" or something like we do for ``haplotypes``?

hyanwong:
> Also not clear that an Object array is what people would want. Would it be better to have a "missing data symbol" or something like we do for `haplotypes`?

ISWYM, but using any particular character will mean that we can't use that for an allele type (which we might conceivably want to do). This is only going to bite for missing data, so I don't think it's too bad TBH.

python/tskit/genotypes.py
```diff
@@ -245,6 +245,22 @@ def copy(self) -> Variant:
         variant_copy._ll_variant = self._ll_variant.restricted_copy()
         return variant_copy
 
+    def genotype_values(self) -> np.ndarray:
```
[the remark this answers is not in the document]
hyanwong: SGTM

jeromekelleher:
Hmm - how about writing some client code for this with and without missing data? Something like the ``verify`` function in tsinfer would be a good one. We definitely don't want to have to have different code paths for dtypes with and without missing data.

We use "N" as the default missing data character elsewhere and its fine. We should be able to use this function to compare easily with e.g. ``alignments``. 

Is the ``np.array(list(ts.alignments())`` equal to ``np.array(var.states() for var in ts.variants()).T`` with and without missing data?

python/tskit/genotypes.py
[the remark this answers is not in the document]
hyanwong: you mean `v.genotypes_state()` or simply `v.states()`

benjeffery:
@hyanwong Would you like this to be in 0.5.4?

hyanwong:
Ah, good point. There's no hurry from my PoV. Is it just a question of my changing the name? I can do that easily enough.

jeromekelleher:
I don't think we should hurry this in  as we don't want to get stuck with Object arrays unless we have to. They have subtly different semantics to arrays of unicode.

We should have some example client code for this before committing ourselves to a particular representation.

hyanwong:
Sure. I have updated this so it should never return an object array anyway. But no hurry to get this in. It was to help @szhan and others who spent a fair bit of time comparing integer arrays between tree sequence before realising that they weren't actually comparable.

hyanwong:
I think this also fixes #2181

benjeffery (approved):

PR #2618, opened 2022-10-30
Add Tree.siblings()

### LINKED ISSUE #2616: Add Tree.siblings(u) method
It is sometimes useful to have the list of siblings of a given node, and it would be handy to have the method to do it directly:
```python
def siblings(self, u):
    sibs = []
    parent = self.parent(u)
    if parent != -1:
        sibs =  [v for v in self.children(parent) if u != v]
    return sibs
```

This should do the trick I think?

szhan:
Fixes #2616 

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -19,6 +19,12 @@
   extra room on the canvas e.g. for long labels or repositioned graphical
   elements (:user:`hyanwong`, :pr:`2646`, :issue:`2645`)
 
+- The ``Tree`` object now has the method ``siblings`` to get
+   the siblings of a node. It returns an empty tuple if the node
+   has no siblings, is not a node in the tree, is the virtual root,
+   or is an isolated non-sample node.
+   (:user:`szhan`, :pr:`2618`, :issue:`2616`)
+
 **Breaking Changes**
 
  - the ``filter_populations``, ``filter_individuals``, and ``filter_sites``
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -4317,6 +4317,69 @@ class TestTree(HighLevelTestCase):
                     assert edge == tskit.NULL
 
 
+class TestSiblings:
+    def test_balanced_binary_tree(self):
+        t = tskit.Tree.generate_balanced(num_leaves=3)
+        assert t.has_single_root
+        # Nodes 0 to 2 are leaves
+        for u in range(2):
+            assert t.is_leaf(u)
+        assert t.siblings(0) == (3,)
+        assert t.siblings(1) == (2,)
+        assert t.siblings(2) == (1,)
+        # Node 3 is the internal node
+        assert t.is_internal(3)
+        assert t.siblings(3) == (0,)
+        # Node 4 is the root
+        assert 4 == t.root
+        assert t.siblings(4) == tuple()
+        # Node 5 is the virtual root
+        assert 5 == t.virtual_root
+        assert t.siblings(5) == tuple()
+
+    def test_star(self):
+        t = tskit.Tree.generate_star(num_leaves=3)
+        assert t.has_single_root
+        # Nodes 0 to 2 are leaves
+        for u in range(2):
+            assert t.is_leaf(u)
+        assert t.siblings(0) == (1, 2)
+        assert t.siblings(1) == (0, 2)
+        assert t.siblings(2) == (0, 1)
+        # Node 3 is the root
+        assert 3 == t.root
+        assert t.siblings(3) == tuple()
+        # Node 4 is the virtual root
+        assert 4 == t.virtual_root
+        assert t.siblings(4) == tuple()
+
+    def test_multiroot_tree(self):
+        ts = tskit.Tree.generate_balanced(4, arity=2).tree_sequence
+        t = ts.decapitate(ts.node(5).time).first()
+        assert t.has_multiple_roots
+        # Nodes 0 to 3 are leaves
+        assert t.siblings(0) == (1,)
+        assert t.siblings(1) == (0,)
+        assert t.siblings(2) == (3,)
+        assert t.siblings(3) == (2,)
+        # Nodes 4 and 5 are both roots
+        assert 4 in t.roots
+        assert t.siblings(4) == (5,)
+        assert 5 in t.roots
+        assert t.siblings(5) == (4,)
+        # Node 7 is the virtual root
+        assert 7 == t.virtual_root
+        assert t.siblings(7) == tuple()
+
+    @pytest.mark.parametrize("flag,expected", [(0, ()), (1, (2,))])
+    def test_isolated_node(self, flag, expected):
+        tables = tskit.Tree.generate_balanced(2, arity=2).tree_sequence.dump_tables()
+        tables.nodes.add_row(flags=flag)  # Add node 3
+        t = tables.tree_sequence().first()
+        assert t.is_isolated(3)
+        assert t.siblings(3) == expected
+
+
 class TestNodeOrdering(HighLevelTestCase):
     """
     Verify that we can use any node ordering for internal nodes
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -1208,6 +1208,33 @@ class Tree:
         """
         return self._right_sib_array
 
+    def siblings(self, u):
+        """
+        Returns the sibling(s) of the specified node ``u`` as a tuple of integer
+        node IDs. If ``u`` has no siblings or is not a node in the current tree,
+        returns an empty tuple. If ``u`` is the root of a single-root tree,
+        returns an empty tuple; if ``u`` is the root of a multi-root tree,
+        returns the other roots (note all the roots are related by the virtual root).
+        If ``u`` is the virtual root (which has no siblings), returns an empty tuple.
+        If ``u`` is an isolated node, whether it has siblings or not depends on
+        whether it is a sample or non-sample node; if it is a sample node,
+        returns the root(s) of the tree, otherwise, returns an empty tuple.
+        The ordering of siblings  is arbitrary and should not be depended on;
+        see the :ref:`data model <sec_data_model_tree_structure>` section for details.
+
+        :param int u: The node of interest.
+        :return: The siblings of ``u``.
+        :rtype: tuple(int)
+        """
+        if u == self.virtual_root:
+            return tuple()
+        parent = self.parent(u)
+        if self.is_root(u):
+            parent = self.virtual_root
+        if parent != tskit.NULL:
+            return tuple(v for v in self.children(parent) if u != v)
+        return tuple()
+
     @property
     def num_children_array(self):
         """
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,7 +36,7 @@ jobs:
         id: cache
         with:
           path: venv
-          key: docs-venv-v4-${{ hashFiles(env.REQUIREMENTS) }}
+          key: docs-venv-v5-${{ hashFiles(env.REQUIREMENTS) }}
 
       - name: Build virtualenv
         if: steps.cache.outputs.cache-hit != 'true'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -157,7 +157,7 @@ jobs:
           python -m pytest -x --cov=tskit --cov-report=xml --cov-branch -n2 tests
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v2
+        uses: codecov/codecov-action@v3
         with:
           working-directory: python
           fail_ci_if_error: true


szhan:
I'm doing this to see how efficiently I can get a simple issue done.

szhan:
Should we also add `Tree.num_siblings` just like `Tree.num_children`?

szhan:
Is there a point to test this function on an empty tree? Since there are no nodes, `Tree.siblings()` should always return an empty tuple.

python/tests/test_highlevel.py
```diff
@@ -4209,6 +4209,22 @@ def test_node_edges(self):
                 else:
                     assert edge == tskit.NULL
 
+    @pytest.mark.parametrize(
+        "t,expected_num_siblings",
+        [
+            (tskit.Tree.generate_random_binary(num_leaves=3, random_seed=1), 1),
```
jeromekelleher, line 4215: I don't think there's much point in having a random  tree here, just use ``generate_balanced`` I think.

jeromekelleher (commented):
Very nice, thanks @szhan!

One thing we'd want to test is the sibs of a root are the other roots, which we're actually doing wrongly at the moment.

I think the simplest way to test this is to make a multiroot example like we do in e.g. test_balance_metrics and just test the values. So some thing like this would be fine:
```python
class TestSiblings:
    test_multiroot_example(self):
           tree = # example that we draw out so we can see the right values.
           assert tree.siblings(5) == (6,)  # e.g.
    test_balanced_ternary(self)
           # as above
```

This is a slightly different testing style to the rest of the file, but I think that's OK.

jeromekelleher:
> Should we also add Tree.num_siblings just like Tree.num_children?

No, there's no need I think.

jeromekelleher:
> Is there a point to test this function on an empty tree? Since there are no nodes, Tree.siblings() should always return an empty tuple.

No harm in testing it. Can add the ``test_empty_tree`` function to your new ``TestSiblings`` class.

benjeffery (commented):
Also don't forget to add a test that it behaves correctly with the virtual root, which you won't get from `nodes`

szhan:
The siblings of a root is the other roots (if they exist), because they are related by a virtual root, right?

benjeffery:
> The siblings of a root is the other roots (if they exist), because they are related by a virtual root, right?

That's what I would expect, yes.

python/tests/test_highlevel.py
szhan: Just checking that a tree is deterministically produced by `generate_balanced`. Is that right?

hyanwong: Yes, it's fully deterministic.

hyanwong:
> > The siblings of a root is the other roots (if they exist), because they are related by a virtual root, right?
> 
> That's what I would expect, yes.

I agree here that `tree.siblings(tree.roots[0])` should give the other roots (if any) in the tree, but I think that needs to be explicitly mentioned in the documentation, as it may not be obvious to users who haven't heard of the virtual root.

Also I agree with checking that the number of sibs of the virtual root (i.e.`tree.siblings(ts.num_nodes)` is allowed, and is the empty tuple).

python/tests/test_highlevel.py
```diff
@@ -4210,6 +4210,76 @@ def test_node_edges(self):
...
+        assert t.is_internal(3)
+        assert t.siblings(3) == (0,)
+        # Node 4 is the root
+        assert 4 == t.root
+        assert t.siblings(4) == tuple()
+        # Node 5 is the virtual root
+        assert 5 == t.virtual_root
+        assert t.siblings(5) == tuple()
+
+    def test_balanced_ternary_tree(self):
```
jeromekelleher, line 4233: rename to ``test_star``

szhan:
I'm not sure how to handle an empty tree, because no nodes exist and whenever I call `Tree.parent()` it raises `ValueError` "Node index out of bounds".

python/tskit/trees.py
```diff
@@ -1205,6 +1205,27 @@ def right_sib_array(self):
...
+        Also, the virtual root itself has no siblings, and so an empty tuple is
+        returned. The ordering of siblings is arbitrary and should not be depended on;
+        see the :ref:`data model <sec_data_model_tree_structure>` section for details.
+
+        :param int u: The node of interest.
+        :return: The siblings of ``u``.
+        :rtype: tuple(int)
+        """
+        parent = self.parent(u)
+        if self.has_multiple_roots and u in self.roots:
```
jeromekelleher, line 1224: This could be inefficient as the list of roots can be large (in a "missing" tree eg.)

How about:
```python
if u == self.virtual_root:
    return tuple()
parent = self.parent(u)
if parent == tskit.NULL:
     parent = self.virtual_root
return tuple(v for v in self.children(parent) if u != v)
```

jeromekelleher (commented):
LGTM! Couple of small additional suggestions.

szhan:
Or maybe try `Tree.parent(u)` and pass on its exception to `Tree.siblings(u)`?

jeromekelleher:
> I'm not sure how to handle an empty tree, because no nodes exist and whenever I call Tree.parent() it raises ValueError "Node index out of bounds".

That's fine I think - any other methods will do the same thing, so we can just forget about the empty tree case here

jeromekelleher:
> We don't have to worry about isolated nodes, right?

Yes, should probably test. Hmm, I guess my version above will the wrong in that case, won't it? Good catch!

szhan:
Okay, this is a bit odd to me...

```
tables = tskit.Tree.generate_balanced(2, arity=2).tree_sequence.dump_tables()
tables.nodes.add_row(flags=1)   # Add node 3
t = tables.tree_sequence().first()
display.SVG(t.draw_svg())
```

Node 3 is a root? `t.roots` returns `[3, 2]`

hyanwong:
> Okay, this is a bit odd to me...
> 
> ```
> tables = tskit.Tree.generate_balanced(2, arity=2).tree_sequence.dump_tables()
> tables.nodes.add_row(flags=1)   # Add node 3
> t = tables.tree_sequence().first()
> display.SVG(t.draw_svg())
> ```
> 
> Node 3 is a root? `t.roots` returns `[3, 2]`, but not when `flags=0`.

Isolated nodes count as roots: https://tskit.dev/tskit/docs/stable/data-model.html#roots

szhan:
>

szhan:
Okay, I hope this is it!

szhan:
All checks have passed! Updating changelog. Still too slow this time around.

python/CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 [0.5.4] - 2022-XX-XX
 --------------------
 
+**Features**
+ - The ``Tree`` object now has the method ``siblings`` to get
+   the siblings of a node. It returns an empty tuple if the node
+   has no siblings, is not a node in the tree, is a virtual root,
+   or is isolated. (:user:`szhan`, :pr:`2618`, :issue:`2616`)
```
hyanwong, line 9: I don't think it will return an empty tuple for isolated nodes, right? They are roots, so `siblings()` will return the other roots

szhan: An isolated sample node is a root, but an isolated non-sample node is not a root?

```
tables = tskit.Tree.generate_balanced(2, arity=2).tree_sequence.dump_tables()
tables.nodes.add_row(flags=0)   # Add node 3
t = tables.tree_sequence().first()
display.SVG(t.draw_svg())
```

`t.roots` returns `[2]`.

But when I set `flags=1`, `t.roots` returns `[3, 2]`.

So, it should really depend on whether the isolated node is a sample node or not?

szhan: Wait... when `flags=0`, `t.children(t.virtual_root)` returns `(2,)`, But when `flags=1`, that returns `(3, 2)` instead.

hyanwong: Yes.

"However, it is also possible for a [sample node](https://tskit.dev/tskit/docs/stable/glossary.html#sec-data-model-definitions-sample) to be isolated. Unlike other nodes, isolated sample nodes are still considered as being present on the tree (meaning they will still returned by the [Tree.nodes()](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.Tree.nodes) and [Tree.samples()](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.Tree.samples) methods): they are therefore plotted, but unconnected to any other nodes."

(https://tskit.dev/tskit/docs/stable/data-model.html#isolated-nodes)

hyanwong: So when an isolated node is a *sample* node, it is considered still present in the tree, as another root.

python/tests/test_highlevel.py
```diff
@@ -4210,6 +4210,84 @@ def test_node_edges(self):
...
+        # Node 3 is the root
+        assert 3 == t.root
+        assert t.siblings(3) == tuple()
+        # Node 4 is the virtual root
+        assert 4 == t.virtual_root
+        assert t.siblings(4) == tuple()
+
+    def test_multiroot_tree(self):
+        tables = tskit.Tree.generate_balanced(4, arity=2).tree_sequence.dump_tables()
+        edges = tables.edges.copy()
```
hyanwong, line 4260: Maybe easier to do

```
ts = tskit.Tree.generate_balanced(4, arity=2).tree_sequence
ts = ts.decapitate(ts.node(5).time)
```

python/tskit/trees.py
```diff
@@ -1205,6 +1205,33 @@ def right_sib_array(self):
...
+        see the :ref:`data model <sec_data_model_tree_structure>` section for details.
+
+        :param int u: The node of interest.
+        :return: The siblings of ``u``.
+        :rtype: tuple(int)
+        """
+        if u == self.virtual_root:
+            return tuple()
+        parent = self.parent(u)
+        if u in self.roots:
```
hyanwong, line 1229: I think this is more efficient as

```
if tree.num_samples(u) > 0 and tree.parent(u) == tskit.NULL
```

see #2620

hyanwong: Or only checking once:

```
parent = self.parent(u)
if parent == tskit.NULL:
    if self.num_samples(u) == 0:
       return tuple()  # no parent and not a root
    else:
        parent = self.virtual_root
return tuple(v for v in self.children(parent) if u != v)
```

jeromekelleher:
> All checks have passed! Updating changelog. Still too slow this time around.

These tree operations are tricky - I don't think anyone would have got this done in one pass!

hyanwong:
> > All checks have passed! Updating changelog. Still too slow this time around.
> 
> These tree operations are tricky - I don't think anyone would have got this done in one pass!

Yeah, all this stuff with isolated nodes is complicated, and then there's the virtual root as well. It's a steep learning curve @szhan - well done on getting there.

python/CHANGELOG.rst
```diff
@@ -2,6 +2,13 @@
 [0.5.4] - 2022-XX-XX
 --------------------
 
+**Features**
+ - The ``Tree`` object now has the method ``siblings`` to get
+   the siblings of a node. It returns an empty tuple if the node
+   has no siblings, is not a node in the tree, is a virtual root,
```
benjeffery, line 8: Seems nitpicky, but I think it is important to conceptualise the virtual root as singular.
```suggestion
   has no siblings, is not a node in the tree, is the virtual root,
```

python/tests/test_highlevel.py
```diff
@@ -4210,6 +4210,78 @@ def test_node_edges(self):
...
+
+    def test_star(self):
+        t = tskit.Tree.generate_star(num_leaves=3)
+        assert t.has_single_root
+        # Nodes 0 to 2 are leaves
+        for u in range(2):
+            assert t.is_leaf(u)
+        assert t.siblings(0) == (
+            1,
+            2,
```
benjeffery, line 4241: Removing the trailing comma here will stop the auto-formatting weirdness.

python/tests/test_highlevel.py
```diff
@@ -4210,6 +4210,78 @@ def test_node_edges(self):
...
+        )
+        assert t.siblings(1) == (
+            0,
```
benjeffery, line 4244: And here

benjeffery (commented):
Couple of nits!

benjeffery:
@szhan Shall I fix the conflict here and see if the tests pass? The failure looks transitory.

jeromekelleher:
I think this is basically done, right? Let's just get it in for the next releaset @szhan as it'll only get more difficult to update. Is it clear what needs to be done?

szhan:
The last time that I worked on it, there was something funny with CircleCI checks, I think. Otherwise, it should be done.

benjeffery:
@mergifyio rebase

benjeffery:
@Mergifyio rebase

python/tskit/trees.py
benjeffery: I went with `if self.is_root(u):` so that `root_threshold` is taken into account.

python/tests/test_highlevel.py
```diff
@@ -4210,6 +4210,78 @@ def test_node_edges(self):
...
+            2,
+        )
+        assert t.siblings(2) == (
+            0,
```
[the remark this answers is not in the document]
benjeffery: Fixed

python/tests/test_highlevel.py
benjeffery: Fixed

python/tests/test_highlevel.py
benjeffery: Fixed

python/CHANGELOG.rst
benjeffery: Fixed

benjeffery (approved):
Fixed up, looks good to go.

jeromekelleher (approved):

PR #2619, opened 2022-10-30
filter_nodes for simplify

### LINKED ISSUE #2605: Add ``filter_nodes`` argument to simplify

hyanwong:
## Description

A follow up to https://github.com/tskit-dev/tskit/pull/2606 with tests added. Currently intentionally failing until we decide whether to allow the passed-in `samples` to be anything other than `ts.samples()`. Will also need to implement this in the C version of simplify, and switch `compare_lib=True`

Fixes #2605.

# PR Checklist:

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -8697,6 +8697,8 @@ simplifier_print_state(simplifier_t *self, FILE *out)
     fprintf(out, "options:\n");
     fprintf(out, "\tfilter_unreferenced_sites   : %d\n",
         !!(self->options & TSK_SIMPLIFY_FILTER_SITES));
+    fprintf(out, "\tno_filter_nodes   : %d\n",
+        !!(self->options & TSK_SIMPLIFY_NO_FILTER_NODES));
     fprintf(out, "\treduce_to_site_topology : %d\n",
         !!(self->options & TSK_SIMPLIFY_REDUCE_TO_SITE_TOPOLOGY));
     fprintf(out, "\tkeep_unary              : %d\n",
@@ -9059,16 +9061,17 @@ out:
 }
 
 static int
-simplifier_init_samples(simplifier_t *self, const tsk_id_t *samples)
+simplifier_init_nodes(simplifier_t *self, const tsk_id_t *samples)
 {
     int ret = 0;
     tsk_id_t node_id;
     tsk_size_t j;
+    tsk_size_t num_nodes = self->input_tables.nodes.num_rows;
+    bool filter_nodes = !(self->options & TSK_SIMPLIFY_NO_FILTER_NODES);
+    bool is_sample;
 
-    /* Go through the samples to check for errors. */
     for (j = 0; j < self->num_samples; j++) {
-        if (samples[j] < 0
-            || samples[j] > (tsk_id_t) self->input_tables.nodes.num_rows) {
+        if (samples[j] < 0 || samples[j] > (tsk_id_t) num_nodes) {
             ret = TSK_ERR_NODE_OUT_OF_BOUNDS;
             goto out;
         }
@@ -9077,15 +9080,38 @@ simplifier_init_samples(simplifier_t *self, const tsk_id_t *samples)
             goto out;
         }
         self->is_sample[samples[j]] = true;
-        node_id = simplifier_record_node(self, samples[j], true);
-        if (node_id < 0) {
-            ret = (int) node_id;
-            goto out;
+    }
+
+    if (filter_nodes) {
+        /* Go through the samples to check for errors. */
+        for (j = 0; j < self->num_samples; j++) {
+            node_id = simplifier_record_node(self, samples[j], true);
+            if (node_id < 0) {
+                ret = (int) node_id;
+                goto out;
+            }
+            ret = simplifier_add_ancestry(
+                self, samples[j], 0, self->tables->sequence_length, node_id);
+            if (ret != 0) {
+                goto out;
+            }
         }
-        ret = simplifier_add_ancestry(
-            self, samples[j], 0, self->tables->sequence_length, node_id);
-        if (ret != 0) {
-            goto out;
+    } else {
+        /* record all the nodes, but only save ancestry for those in the sample */
+        for (j = 0; j < num_nodes; j++) {
+            is_sample = self->is_sample[j];
+            node_id = simplifier_record_node(self, (tsk_id_t) j, is_sample);
+            if (node_id < 0) {
+                ret = (int) node_id;
+                goto out;
+            }
+            if (is_sample) {
+                ret = simplifier_add_ancestry(
+                    self, node_id, 0, self->tables->sequence_length, node_id);
+                if (ret != 0) {
+                    goto out;
+                }
+            }
         }
     }
 out:
@@ -9174,7 +9200,7 @@ simplifier_init(simplifier_t *self, const tsk_id_t *samples, tsk_size_t num_samp
     if (ret != 0) {
         goto out;
     }
-    ret = simplifier_init_samples(self, samples);
+    ret = simplifier_init_nodes(self, samples);
     if (ret != 0) {
         goto out;
     }
@@ -9253,11 +9279,10 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
     tsk_id_t ancestry_node;
     tsk_id_t output_id = self->node_id_map[input_id];
 
-    bool is_sample = output_id != TSK_NULL;
-    bool keep_unary = false;
-    if (self->options & TSK_SIMPLIFY_KEEP_UNARY) {
-        keep_unary = true;
-    }
+    bool is_sample = self->is_sample[input_id];
+    /* bool is_sample = output_id != TSK_NULL; */
+    bool filter_nodes = !(self->options & TSK_SIMPLIFY_NO_FILTER_NODES);
+    bool keep_unary = !!(self->options & TSK_SIMPLIFY_KEEP_UNARY);
     if ((self->options & TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS)
         && (self->input_tables.nodes.individual[input_id] != TSK_NULL)) {
         keep_unary = true;
@@ -9348,7 +9373,7 @@ simplifier_merge_ancestors(simplifier_t *self, tsk_id_t input_id)
         if (ret != 0) {
             goto out;
         }
-        if (num_flushed_edges == 0 && !is_sample) {
+        if (filter_nodes && (num_flushed_edges == 0) && !is_sample) {
             ret = simplifier_rewind_node(self, input_id, output_id);
         }
     }
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -7159,7 +7159,6 @@ typedef struct {
 } segment_overlapper_t;
 
 typedef struct {
-    tsk_id_t *samples;
     tsk_size_t num_samples;
     tsk_flags_t options;
     tsk_table_collection_t *tables;
@@ -7168,6 +7167,7 @@ typedef struct {
     /* State for topology */
     tsk_segment_t **ancestor_map_head;
     tsk_segment_t **ancestor_map_tail;
+    /* Mapping of input node IDs to output node IDs. */
     tsk_id_t *node_id_map;
     bool *is_sample;
     /* Segments for a particular parent that are processed together */
@@ -7185,8 +7185,6 @@ typedef struct {
     tsk_size_t num_buffered_children;
     /* For each mutation, map its output node. */
     tsk_id_t *mutation_node_map;
-    /* Map of input mutation IDs to output mutation IDs. */
-    tsk_id_t *mutation_id_map;
     /* Map of input nodes to the list of input mutation IDs */
     mutation_id_list_t **node_mutation_list_map_head;
     mutation_id_list_t **node_mutation_list_map_tail;
@@ -8807,7 +8805,7 @@ out:
 /* Add a new node to the output node table corresponding to the specified input id.
  * Returns the new ID. */
 static tsk_id_t TSK_WARN_UNUSED
-simplifier_record_node(simplifier_t *self, tsk_id_t input_id, bool is_sample)
+simplifier_record_node(simplifier_t *self, tsk_id_t input_id)
 {
     tsk_node_t node;
     tsk_flags_t flags;
@@ -8815,7 +8813,7 @@ simplifier_record_node(simplifier_t *self, tsk_id_t input_id, bool is_sample)
     tsk_node_table_get_row_unsafe(&self->input_tables.nodes, (tsk_id_t) input_id, &node);
     /* Zero out the sample bit */
     flags = node.flags & (tsk_flags_t) ~TSK_NODE_IS_SAMPLE;
-    if (is_sample) {
+    if (self->is_sample[input_id]) {
         flags |= TSK_NODE_IS_SAMPLE;
     }
     self->node_id_map[input_id] = (tsk_id_t) self->tables->nodes.num_rows;
@@ -8878,7 +8876,7 @@ simplifier_init_position_lookup(simplifier_t *self)
         goto out;
     }
     self->position_lookup[0] = 0;
-    self->position_lookup[num_sites + 1] = self->tables->sequence_length;
+    self->position_lookup[num_sites + 1] = self->input_tables.sequence_length;
     tsk_memcpy(self->position_lookup + 1, self->input_tables.sites.position,
         num_sites * sizeof(double));
 out:
@@ -8922,7 +8920,7 @@ simplifier_record_edge(simplifier_t *self, double left, double right, tsk_id_t c
     interval_list_t *tail, *x;
     bool skip;
 
-    if (!!(self->options & TSK_SIMPLIFY_REDUCE_TO_SITE_TOPOLOGY)) {
+    if (self->options & TSK_SIMPLIFY_REDUCE_TO_SITE_TOPOLOGY) {
         skip = simplifier_map_reduced_coordinates(self, &left, &right);
         /* NOTE: we exit early here when reduce_coordindates has told us to
          * skip this edge, as it is not visible in the reduced tree sequence */
@@ -8968,8 +8966,6 @@ simplifier_init_sites(simplifier_t *self)
     mutation_id_list_t *list_node;
     tsk_size_t j;
 
-    self->mutation_id_map
-        = tsk_calloc(self->input_tables.mutations.num_rows, sizeof(tsk_id_t));
     self->mutation_node_map
         = tsk_calloc(self->input_tables.mutations.num_rows, sizeof(tsk_id_t));
     self->node_mutation_list_mem
@@ -8978,15 +8974,12 @@ simplifier_init_sites(simplifier_t *self)
         = tsk_calloc(self->input_tables.nodes.num_rows, sizeof(mutation_id_list_t *));
     self->node_mutation_list_map_tail
         = tsk_calloc(self->input_tables.nodes.num_rows, sizeof(mutation_id_list_t *));
-    if (self->mutation_id_map == NULL || self->mutation_node_map == NULL
-        || self->node_mutation_list_mem == NULL
+    if (self->mutation_node_map == NULL || self->node_mutation_list_mem == NULL
         || self->node_mutation_list_map_head == NULL
         || self->node_mutation_list_map_tail == NULL) {
         ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
-    tsk_memset(self->mutation_id_map, 0xff,
-        self->input_tables.mutations.num_rows * sizeof(tsk_id_t));
     tsk_memset(self->mutation_node_map, 0xff,
         self->input_tables.mutations.num_rows * sizeof(tsk_id_t));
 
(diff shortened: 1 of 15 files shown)


hyanwong:
I just realised that the easiest way to allow any nodes to be marked as samples is to switch the NODE_IS_SAMPLE flags before simplifying. Since we require the node order to remain the same, I think this is all we need to do.

hyanwong:
> I just realised that the easiest way to allow any nodes to be marked as samples is to switch the NODE_IS_SAMPLE flags before simplifying. Since we require the node order to remain the same, I think this is all we need to do.

Actually, it's even easier than that. Recent commit included: I think this works fine:

```
sample = set(sample)
for node in ts.nodes():
    is_sample = node.id in sample
    self.record_node(node.id, is_sample=is_sample)
    if is_sample:
        self.add_ancestry(node.id, 0, self.sequence_length, node.id)
```

hyanwong:
I think this now needs to be implemented in the C version of `simplify()`, right?

hyanwong:
I coded up an equivalent C version, but it is segfaulting. I've pushed the changes anyway, in case someone else can see why that might be happening.

hyanwong:
I have coded up the C equivalent, but there are 2 remaining issues:

1. In `merge_labeled_ancestors` a new line says `is_sample = input_id in self.samples  # HACK`, which worries me
2. The default C call will have `TSK_SIMPLIFY_FILTER_NODES` set to 0, so the defaults will change to *not* filter nodes, which I think is causing the C tests to fail. This seems like a rather large change in default behaviour. I'm not sure what the best options are here.

molpopgen:
I am trying to figure out what this does as I think it collides with something that one may need for (one approach to) parallel simplification.

The comments in #2605 are concerning to me re: marking all nodes as samples.  That would completely break the stuff that I am presenting in December by rendering it not possible. (Or not possible w/o another flag to simplify that suppresses the behavior described in #2605, but I'd consider that a serious design failure.)

Could someone provide a *simple* description of what this is doing -- I can't find one by by tracking the linked issues, etc..

hyanwong:
> The comments in #2605 are concerning to me re: marking all nodes as samples. 

I don't think we mark *all* nodes as samples. The current implementation of `simplify()` takes a list of nodes, and turns all those nodes into samples, keeping their ancestry WRT the other nodes, which are marked as non-samples. By default, the list of nodes is simply all the existing samples in the TS, but you could provide a different list of nodes, in which case these become the samples instead. As far as I know, `simplify()` has always worked like this.

> Could someone provide a _simple_ description of what this is doing -- I can't find one by by tracking the linked issues, etc..

Well, I think the key is here:

```
if filter_nodes:
    for sample_id in sample:
        output_id = self.record_node(sample_id, is_sample=True)
        self.add_ancestry(sample_id, 0, self.sequence_length, output_id)
else:
    assert list(sample) == list(ts.samples())
    for node in ts.nodes():
        self.record_node(node.id, node.is_sample())
        if node.is_sample():
           self.add_ancestry(node.id, 0, self.sequence_length, node.id)
```

So normally we record the passed-in node list, treating all of these nodes as samples, and add the ancestry of those nodes into the TS. If we *don't* filter nodes, we record *all* the nodes in the normal order (meaning that the node IDs don't change), but we only keep the ancestry of the samples. That means a number of nodes may end up not being referenced in the node table.

Does that make sense? Is my explanation too simplistic @molpopgen ? I feel you probably understand this much better than I do anyway.

molpopgen:
Yes, I think that this makes sense -- the work "all" in the comment elsewhere was ambiguous to me.

So, if filter nodes is False:

* all nodes get recorded
* only sample nodes have their ancestry preserved.

For this to make sense, all nodes that get recorded by have no referents after simplification should be marked as NULL in the output idmap?  Is that the case?

hyanwong:
> So, if filter nodes is False:
> * all nodes get recorded
> * only sample nodes have their ancestry preserved.

Yes, with the proviso that the sample nodes are defined as the ones passed in to simplify (which by default is all the nodes flagged with tskit.NODE_IS_SAMPLE in the original tree sequence)

> For this to make sense, all nodes that get recorded by have no referents after simplification should be marked as NULL in the output idmap? Is that the case?

The idmap doesn't tell you whether the node is used or not (I think), only if the node Id has changed. Even in the case of nodes that have no referents after simplification, the nodes are retained in the table (although not referenced anywhere), and the idmap is simply the identify map. In fact, by definition, the output idmap should be `0..num_nodes-1` and not contain any NULLs (because no nodes are actually removed).

molpopgen:
> The idmap doesn't tell you whether the node is used or not, only if the node Id has changed.

This isn't quite right -- a node that "simplifies out" of the tables (is not longer part of the ancestry anywhere) has -1/null in the idmap.  For what I envision, I require this to be true for this feature, too.

molpopgen:
> Yes, with the proviso that the sample nodes are defined as the ones passed in to simplify (which by default is all the nodes flagged with tskit.NODE_IS_SAMPLE in the original tree sequence)

This is not what I read in the code that you wrote above -- what is "passed in node list"?

molpopgen:
> In fact, by definition, the output idmap when `filter_nodes=False` should be `0..num_nodes-1` and not contain any NULLs (because no nodes are actually removed).

This is the sticking point, and breaks what I need this feature to do.  I require a way to: keep the input and output node tables the same length but also to know which nodes are no longer used.  The implementation of this that I have done takes this into account by entering NULL into idmap for the latter case.

molpopgen:
My implementtion is here: https://github.com/molpopgen/tskit-rust/tree/parallel_experiment.

I cannot make it easy to see diffs b/c that would trigger a PR against tskit-rust, but you can look at the commit history for the branch and see the changes to simplification.

hyanwong:
> This is not what I read in the code that you wrote above -- what is "passed in node list"?

The user calls `ts.simplify(samples=[10,11,12])`. The nodes 10, 11, and 12 become the samples in the output tree sequence. If not give, the `samples` parameter default to `ts.samples()`, so that the samples in the input tree sequence remain samples in the output tree sequence (albeit with changed node IDs).

> I require a way to: keep the input and output node tables the same length but also to know which nodes are no longer used. 

I'm sure we could find a way to report this. We could revise the meaning of the idmap in this case, or (I guess) we could search through the edge table (but that's not very performant). @jeromekelleher might have a sensible suggestion here.

My previous hack would work for this, as a matter of fact: do a normal `simplify` (removing unreferenced nodes) to create a new tree sequence, look at the nodes with NULL in the map, then edit the *original* tree sequence by porting the edges over from the new tree sequence back into the old one, with a suitable remapping of parent and child IDs. Maybe that's too hack for you, though?

hyanwong:
I think it could be reasonable to return an id_map with NULL for unused nodes in the `filter_nodes=False` case, but I don't know if it's actually very easy to implement. I will ask @jeromekelleher .

hyanwong:
I think this is the key part of @molpopgen 's alternative implementation (pasted as a screenshot as I can't work out how to get a link)

Screenshot 2022-11-01 at 11 39 09

molpopgen:
> The user calls `ts.simplify(samples=[10,11,12])`. The nodes 10, 11, and 12 become the samples in the output tree sequence. If not give, the `samples` parameter default to `ts.samples()`, so that the samples in the input tree sequence remain samples in the output tree sequence (albeit with changed node IDs).

Something doesn't sit right with this explanation.  If the user passes in `[10, 11, 12]`, and filter_nodes is False, then only nodes in `[10, 11, 12]` are marked as samples:

```python
    for node in ts.nodes():
        # True if node is in [10, 11, 12] and False otherwise
        is_sample = node.id in passed_in_node_list
       # All non-sample nodes not in [10, 11, 12] are not marked as 
       # samples
        self.record_node(node.id, is_sample=is_sample)
        if is_sample:
           # All non-sample nodes not in 10, 11, 12 do not have ancestry
           # added.  Unless any such nodes connect up to the ancestries
           # of [10, 11, 12], their output ID should be NULL?
           self.add_ancestry(node.id, 0, self.sequence_length, node.id)
```

molpopgen:
> I think it could be reasonable to return an id_map with NULL for unused nodes in the `filter_nodes=False` case, but I don't know if it's actually very easy to implement. I will ask @jeromekelleher .

This is what the existing code already does.  A node id is a pointer:  if that point is non-null, then its memory location (row) points to valid node data that exists in the tree sequence.  If the pointer is null, then that location (row) is not node data in the tree sequence.  Simplification causes what is known as "pointer invalidation" -- imagine an application tracking its own data using node id as a key.  After simplification, those data are cleaned.  NULL node ids mean that the data can be removed from the structure because the data correspond to a node no longer in the tree sequence.  This kind of design is an implied contract of the current API.  Breaking it by having a simplification flag that means non-null values are returned for ids not in the tree sequence is a backwards-incompatible change, IMO
.

hyanwong:
> Something doesn't sit right with this explanation. If the user passes in `[10, 11, 12]`, and filter_nodes is False, then only nodes in `[10, 11, 12]` are marked as samples:
> 
> ```python
>            ...
>            # All non-sample nodes not in 10, 11, 12 do not have ancestry
>            # added.  Unless any such nodes connect up to the ancestries
>            # of [10, 11, 12], their output ID should be NULL?
>            self.add_ancestry(node.id, 0, self.sequence_length, node.id)
> ```

I think that's right. When we call `record_node` (a few lines before) we add the node to the output IDs, and that gets stuck in the id_map.

hyanwong:
> 



> > I think it could be reasonable to return an id_map with NULL for unused nodes in the `filter_nodes=False` case, but I don't know if it's actually very easy to implement. I will ask @jeromekelleher .
> 
> This is what the existing code already does. A node id is a pointer: if that point is non-null, then its memory location (row) points to valid node data that exists in the tree sequence. If the pointer is null, then that location (row) is not node data in the tree sequence. Simplification causes what is known as "pointer invalidation" -- imagine an application tracking its own data using node id as a key. After simplification, those data are cleaned. NULL node ids mean that the data can be removed from the structure because the data correspond to a node no longer in the tree sequence. This kind of design is an implied contract of the current API. Breaking it by having a simplification flag that means non-null values are returned for ids not in the tree sequence is a backwards-incompatible change, IMO .

Well, the point is moot as after chatting to Jerome he agrees that it would be reasonable to return NULL for nodes that are not in the topology.

It is, however, debatable whether a node that is not referenced anywhere is "in the tree sequence". It and its metadata are still accessible using `ts.node(unreferenced_id)`, it's just not in the edge table (or the mutations table, I guess).

hyanwong:
> My implementtion is here: https://github.com/molpopgen/tskit-rust/tree/parallel_experiment.
> 
> I cannot make it easy to see diffs b/c that would trigger a PR against tskit-rust, but you can look at the commit history for the branch and see the changes to simplification (in the C code)

I implemented what I think is the equivalent of your code in the python version of simplify, to see if it worked there. Here are the changes, which I think mirror your C changes. But this modified Python simplify code now fails with e.g. `_tskit.LibraryError: Duplicate edges provided. (TSK_ERR_DUPLICATE_EDGES)`

https://github.com/tskit-dev/tskit/pull/2619/files#diff-bee19d2851ff4fa665bdf9e14403d339e7a501f8b08832b0816662b8105df20d

I know python is not your thing @molpopgen, but con you see what I've done wrong?

jeromekelleher:
I think it's probably simpler if I pick this up @hyanwong, I'll work through the C stuff.

hyanwong:
That would be great, thanks @jeromekelleher.

molpopgen:
> It is, however, debatable whether a node that is not referenced anywhere is "in the tree sequence". It and its metadata are still accessible using `ts.node(unreferenced_id)`, it's just not in the edge table (or the mutations table, I guess).

That's correct -- the idmap is doing us a very nice thing here.  It is telling us 1.) which pointers are still valid and what their current locations are and 2.) which pointers are no longer valid.

That second group is important!  You can over-write the data for those nodes with your next batches of births in a forward simulation, for example.

molpopgen:
> I know python is not your thing @molpopgen, but can you see what I've done wrong?

It may take me a day or 3.  I'm off of work for a bit.

jeromekelleher:
How's this look @hyanwong?

hyanwong:
Sorry, I can't seem to see any changes? Maybe we can talk it through at 2pm?

jeromekelleher:
Oops, mucked up my git pushing. Look ok now? Needs some squashing etc obvs

jeromekelleher:
The nodes are still in the data model fundamentally, so I think it's better if the node mapping is the identity array in the filter_nodes=False case. 

It is quite straightforward to find the unreferenced nodes afterwards if you want to get them. OK, it's a linear pass through the edge table, but this will be negligible extra cost after running simplify.

Does seem ok @molpopgen or do you think we really definitely should use the node mapping to return info about unreferenced nodes?
```
￼    def test_find_unreferenced_nodes(self):
￼        # Simple test to show we can find unreferenced nodes easily.
￼        # 2.00┊    6    ┊
￼        #     ┊  ┏━┻━┓  ┊
￼        # 1.00┊  4   5  ┊
￼        #     ┊ ┏┻┓ ┏┻┓ ┊
￼        # 0.00┊ 0 1 2 3 ┊
￼        #     0         1
￼        ts1 = tskit.Tree.generate_balanced(4).tree_sequence
￼        ts2, node_map = do_simplify(
￼            ts1, [0, 1, 2], filter_nodes=False, compare_lib=False
￼        )
￼        assert np.array_equal(node_map, np.arange(ts1.num_nodes))
￼        node_references = np.zeros(ts1.num_nodes, dtype=np.int32)
￼        node_references[ts2.edges_parent] += 1
￼        node_references[ts2.edges_child] += 1
￼        # Simplifying for [0, 1, 2] should remove references to node 3 and 5
￼        assert list(node_references) == [1, 1, 1, 0, 2, 0, 1]
```

molpopgen:
All I am saying is: not having "simplified out" nodes kills the parallelization method used in my pr to tskit-rust.  (NOTE: the approach has nothing to do w/tskit-rust.  It was just fastest to implement it there.)

I think that it also breaks existing contracts to not have them be NULL, as I stated earlier.  I don't think it is relevant if no existing programs break!

I think this is bringing up that I understand idmap differently.  I understand it as the fate of a set of pointers, and it is undefined behavior to try to do anything meaningful with ids that are no longer part of the ancestry of sample nodes.  So these funny nodes are part of the data structure (sitting in the middle of an allocation containing valid and invalid nodes), but not the data itself.  IMO, this is moot for the C API.  The problem seems to the that the current Python API will happily return node table rows for nodes that no longer exist?

In the end, what I have done elsewhere can be done via the edge table, but that is generally a larger data structure to traverse.  And it is redundant -- we will hit the same nodes over and over again.  In a performance-critical scenario it could start to matter?

jeromekelleher:
Hmm, looks like this has brought up a bug that was lurking in the initialisation code that we'll need to document in the CHANGELOG and consider more carefully. I think there may have been two separate issues in there, will have to look at again more carefully once CI is passing.

jeromekelleher:
> what mapping should we return if an individual or mutation still refers to a node that's not ancestral to any of the samples?

Another note: we need to test the behaviour here and document it in anycase. So, more tests needed also.

molpopgen:
> The edges are just one part of the data model - what mapping should we return if an individual or mutation still refers to a node that's not ancestral to any of the samples? I think we'd have to write some extra code to determine this, and we'd have to write up exactly what all these things mean in the context of all the different filter-x flags.

This is downstream of my concerns, but I agree that it is relevant to the Python API.
For a forward sim, there is no guarantee that these other tables are in use until the final data are exported.  I actually never intended this option to be public on the Python side.

jeromekelleher:
I've done a quick update on this to check that we're doing the right thing for mutations. I think it's basically good, but I'll go back over the implementation to make sure that we're not touching the input node table. There's only two things to decide I think:

1. Are we OK with the returned node mapping being the identity? I can see the attraction of using it to say "which of the nodes are referred to by edges" in the output, but I think it's slightly strained and would take a bit of explaining. I honestly can't see a linear pass through the output edges amounting to a measurable difference in performance for the parallel simplify case.
2. Should we also add a flag saying "don't set the sample flags" in the output node table? This is the only place that we'd have concurrent writes if we did the funky share-the-node-table thing. In practise it may not matter if multiple threads do write, as client code could just go ahead and let that happen and then go back in the end setting the flags properly. But maybe this is putting too much trust in the user?

hyanwong:
> 1. Are we OK with the returned node mapping being the identity? 

I've yo-yoed back and forth about this, and have no strong feelings either way. I agree with your assessment: it's attractive but a bit strained. Perhaps it depends on whether we can document it without it feeling odd?

> 2. Should we also add a flag saying "don't set the sample flags"

I like the idea of sharing the node table (e.g. between multiple edge tables), so anything that can be used to guarantee read-only access STGM.

molpopgen:
I am fine with having the output map be the simple identity map.

For number 2, I think this is a good idea if I understand it -- it means that the node table is totally untouched by simplification?  If so, I am on board.

jeromekelleher:
I think this does what we want, besides the "don't touch the flags" option? 

I guess we call it TSK_SIMPLIFY_NO_UPDATE_SAMPLE_FLAGS or something?

Does it make sense to allow this flag without NO_FILTER_NODES?

hyanwong:
> I guess we call it TSK_SIMPLIFY_NO_UPDATE_SAMPLE_FLAGS or something?

SGTM
 
> Does it make sense to allow this flag without NO_FILTER_NODES?

Hmm, good question. I think it might make sense, for example from the ARG point of view, we might want to remove the nodes that are irrelevant to (say) the last 3 generations, while only having the final generation as the samples. We could do this by specifying the last 3 generations as the "focal" nodes in simplify, with a subset of the nodes being those that are the 'real' samples. It seems a bit obscure, but not entirely useless.

jeromekelleher:
Aha, yes - we should make sure that we also don't touch the individual and population tables here when their not filtered.

petrelharp:
All this sounds great to me. Two suggestions:
- how about `TSK_SIMPLIFY_KEEP_NODES` or `TSK_SIMPLIFY_KEEP_ALL_NODES` to match [other flags](https://tskit.dev/tskit/docs/latest/c-api.html#c.tsk_table_collection_simplify)
- how about `TSK_SIMPLIFY_NO_MARK_SAMPLES` as more concise than NO_UPDATE_SAMPLE_FLAGS (and maybe more clear, because why is it "update"?)

jeromekelleher:
I think it's better to keep this as NO_FILTER following the population etc tables because the KEEP flags (in retrospect) are more about keeping *edges* than nodes. We have:
```
TSK_SIMPLIFY_FILTER_SITES
TSK_SIMPLIFY_FILTER_POPULATIONS
TSK_SIMPLIFY_FILTER_INDIVIDUALS
TSK_SIMPLIFY_REDUCE_TO_SITE_TOPOLOGY
TSK_SIMPLIFY_KEEP_UNARY
TSK_SIMPLIFY_KEEP_INPUT_ROOTS
TSK_SIMPLIFY_KEEP_UNARY_IN_INDIVIDUALS
```

We keep the edges, and therefore the corresponding nodes are not filtered.

I like ``TSK_SIMPLIFY_NO_MARK_SAMPLES`` though

jeromekelleher:
OK, this should implement the "don't touch the table pointers" semantics when we don't filter unreferenced nodes, sites, populations or individuals. I had to jiggle things around a bit to do it, but I think it was worth the refactoring. Hopefully it's a bit clearer how the copies are used now and it opens the door for future memory reductions as well.

I still need to document and add the NO_UPDATE_SAMPLE_FLAGS. I'm also hoping to add an example with multiple table collections using the same nodes and individuals, to see convince myself that this all works in practise but we'll see how that goes time-wise.

jeromekelleher:
I think this is ready for review and merge - I'll implement TSK_SIMPLIFY_NO_UPDATE_SAMPLE_FLAGS separately as this PR has ended up getting quite big.

Turns out making this memory safe required quite a bit of fiddling! I think the code it much better now though, so hopefully worth the effort.

c/tskit/tables.c
```diff
@@ -8920,7 +8920,7 @@ simplifier_record_edge(simplifier_t *self, double left, double right, tsk_id_t c
     interval_list_t *tail, *x;
     bool skip;
 
-    if (!!(self->options & TSK_SIMPLIFY_REDUCE_TO_SITE_TOPOLOGY)) {
```
jeromekelleher, line 8923: In case anyone is wondering, I got rid of a bunch of these "not not"s because they're not needed from the C perspective and they don't make the code any easier to read.

c/tskit/tables.h
```diff
@@ -686,6 +686,11 @@ reference them. */
 #define TSK_SIMPLIFY_FILTER_POPULATIONS (1 << 1)
 /** Remove individuals from the output if there are no nodes that reference them.*/
 #define TSK_SIMPLIFY_FILTER_INDIVIDUALS (1 << 2)
+/** Do not remove nodes from the output if there are no edges that reference
+them. Note that this flag is negated compare to other filtering options because
```
benjeffery, line 690:
```suggestion
them. Note that this flag is negated compared to other filtering options because
```

c/tskit/tables.h
```diff
@@ -3909,8 +3914,32 @@ A mapping from the node IDs in the table before simplification to their equivale
...
+little as possible, and all pointers are guaranteed to remain valid
+after simplification. The only changes made to an unfiltered table are
+to update any references to tables that may have changed (for example,
+remapping population IDs in the node table if
+`TSK_SIMPLIFY_FILTER_POPULATIONS` was specified) or altering the
+sample status flag of nodes.
+
+.. note:: It is possible for populations and individuals to be filtered
+   even if `TSK_SIMPLIFY_NO_FILTER_NODES` is specified because there
+   may be entirely unreferences entities in the input tables, which
```
benjeffery, line 3939:
```suggestion
   may be entirely unreferenced entities in the input tables, which
```

benjeffery (approved):
LGTM - I only found a couple of typos

jeromekelleher:
Added a few minor things from review from @petrelharp and @benjeffery - ready to go I think!

benjeffery:
`git push`ing here to trigger CI as for some reason circle CI won't let me click the button.

jeromekelleher:
@mergifyio rebase

PR #2621, opened 2022-10-31
Use recursive updates for weights in ``CoalescenceTimeDistribution``

### LINKED ISSUE #2614: Redesign weight functions for `CoalescenceTimeDistribution` to ensure correctness when propagating weights up trees
The `CoalescenceTimeDistribution` class uses a an arbitrary (and possibly user-supplied) function to compute combinatorial weights for each node in a marginal tree. The function signature is currently required to be,
```python
def weight_func(node, tree, sample_sets):
   node_weights = ...
   return node_weights
```
and this gets applied to a node every time its descendant subtree changes. The idea is that weights may be efficiently calculated by traversing edge diffs and maintaining a running array of weights.

Unfortunately, this update scheme isn't guaranteed to be correct for every possible weight function. A contrived example is if the weight function is intended to return random weights for every marginal node,
```python
def weight_func(node, tree, sample_sets):
   return np.random.uniform(...) #vector of random weights for node in tree
```
but this will produce incorrect results because only a small subset of nodes is updated for every marginal tree (e.g. there wouldn't be a distinct random weight for each internal node in each marginal tree). 

A better approach would be to restrict possible weight functions to have the signature,
```python
def weight_func(parent_weights, child_weights):
   updated_parent_weights = ...
   return updated_parent_weights
```
that are applied *per-edge*. This would ensure that the edge-diff update scheme always behaves as intended. This will also make it easier to implement efficient, recursive updates on a given tree.

nspope:
Fixes #2614.

- Use a recursive update rather than traversing trees independently for each node (details below).
- For `span_normalise=True`, divide by total span of nonmissing trees rather than total sequence length.
- Add a method to get interpolated quantiles
- Ignore unary nodes during tree traversals
- Test that blocking of nodes (e.g. for bootstrapping) doesn't influence methods

The core algorithm in `CoalescenceTimeDistribution` operates by counting the number of subtrees of a particular tip labeling, that are rooted at each node in a tree. Doing this by brute force involves a lot of redundant calculations. This PR implements a recursive approach that should be a lot more efficient. 

The idea is that a topology with $n$ tips can be constructed from two smaller topologies with $i$ and $n-i$ tips, respectively. So to count the number of topologies with $n$ tips rooted at a given node, we need to know the number of topologies of size $1, \dots, n-1$ rooted at each child of this node.

This is done by splitting the weight function into two parts; `initialize` (that gets called on tips) and `update` (that gets called on internal nodes). These look like:
```
def initialize_trio_count(node, sample_sets):
    # set dimensions based on ``sample_sets``
    n_lineages = np.array(...)
    n_pairs = np.array(...)

    return (n_lineages, n_pairs)

def update_trio_count(n_lineages_in_children, n_pairs_in_children)
    # get counts of lower-order topologies rooted at node, from counts in children
    n_lineages = np.sum(n_lineages_in_children, axis=...)
    n_pairs = combinatorial_function_of(n_lineages_in_children)

    # calculate desired weight
    n_trios = combinatorial_function_of(n_lineages_in_children, n_pairs_in_children)

    return n_trios, (n_lineages, n_pairs)
```
In other words, ``initialize`` sets the dimensions of a propagated "state" (counts of lower-order topologies); while ``update`` uses the states of children to: (A) calculate the state for the parent (in this case the tuple ``(n_lineages, n_pairs)``; (B) calculate the desired weight (in this case ``n_trios``).

### DIFF
--- a/python/tests/test_coaltime_distribution.py
+++ b/python/tests/test_coaltime_distribution.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2022 Tskit Developers
+# Copyright (c) 2018-2023 Tskit Developers
 # Copyright (C) 2016 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -25,6 +25,7 @@ Test cases for coalescence time distribution objects in tskit.
 """
 import msprime
 import numpy as np
+import pytest
 
 import tests
 import tskit
@@ -181,7 +182,7 @@ class TestCoalescenceTimeDistribution:
     @tests.cached_example
     def ts_many_edge_diffs(self):
         ts = msprime.sim_ancestry(
-            samples=75,
+            samples=80,
             ploidy=1,
             sequence_length=4,
             recombination_rate=10,
@@ -216,31 +217,31 @@ class TestUnweightedCoalescenceTimeTable(TestCoalescenceTimeDistribution):
         t = np.array([0, 1, 5, 8, 29])
         distr = self.coalescence_time_distribution()
         tt = distr.tables[0].time
-        assert np.allclose(t, tt)
+        np.testing.assert_allclose(t, tt)
 
     def test_block(self):
         b = np.array([0, 0, 0, 0, 0])
         distr = self.coalescence_time_distribution()
         tb = distr.tables[0].block
-        assert np.allclose(b, tb)
+        np.testing.assert_allclose(b, tb)
 
     def test_weights(self):
         w = np.array([[0, 1, 1, 1, 1]]).T
         distr = self.coalescence_time_distribution()
         tw = distr.tables[0].weights
-        assert np.allclose(w, tw)
+        np.testing.assert_allclose(w, tw)
 
     def test_cum_weights(self):
         c = np.array([[0, 1, 2, 3, 4]]).T
         distr = self.coalescence_time_distribution()
         tc = distr.tables[0].cum_weights
-        assert np.allclose(c, tc)
+        np.testing.assert_allclose(c, tc)
 
     def test_quantile(self):
         q = np.array([[0, 0.25, 0.50, 0.75, 1]]).T
         distr = self.coalescence_time_distribution()
         tq = distr.tables[0].quantile
-        assert np.allclose(q, tq)
+        np.testing.assert_allclose(q, tq)
 
 
 class TestPairWeightedCoalescenceTimeTable(TestCoalescenceTimeDistribution):
@@ -277,13 +278,13 @@ class TestPairWeightedCoalescenceTimeTable(TestCoalescenceTimeDistribution):
         t = np.array([0, 1, 5, 8, 29])
         distr = self.coalescence_time_distribution()
         tt = distr.tables[0].time
-        assert np.allclose(t, tt)
+        np.testing.assert_allclose(t, tt)
 
     def test_block(self):
         b = np.array([0, 0, 0, 0, 0])
         distr = self.coalescence_time_distribution()
         tb = distr.tables[0].block
-        assert np.allclose(b, tb)
+        np.testing.assert_allclose(b, tb)
 
     def test_weights(self):
         w = np.array(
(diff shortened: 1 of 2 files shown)


nspope:
Hi @jeromekelleher, sorry to sleep on this PR for so long. I think this is at a good point to merge, whenever you have the time to take a look. @petrelharp and I have talked about changing the algorithm to a simpler one that operates on edges, which I think is a good idea -- but the method is completely functional with these changes, and we'd like to try using it for https://github.com/tskit-dev/tsdate/issues/237.

The docs build is failing-- I haven't touched `docs/` at all, so I'm not sure why this would be? Do you have a suggestion on how to troubleshoot?

python/tskit/stats.py
```diff
@@ -215,24 +215,49 @@ def resample_blocks(self, block_multiplier):
             )
             if self.cum_weights[-1, i] > 0:
                 self.quantile[:, i] = self.cum_weights[:, i] / self.cum_weights[-1, i]
+            else:
+                self.quantile[:, i] = np.nan
```
jeromekelleher, line 219: This isn't covered by tests I think?

python/tests/test_coaltime_distribution.py
```diff
@@ -952,42 +989,117 @@ def test_trees_per_block(self):
...
+        distr_noblock = self.coalescence_time_distribution(num_blocks=1)
+        distr_block = self.coalescence_time_distribution(num_blocks=10)
+        t = np.linspace(0, distr_noblock.tables[0].time[-1] + 1, 5)
+        assert np.allclose(distr_noblock.ecdf(t), distr_block.ecdf(t))
+
+    def test_num_coalesced(self):
+        distr_noblock = self.coalescence_time_distribution(num_blocks=1)
+        distr_block = self.coalescence_time_distribution(num_blocks=10)
+        t = np.linspace(0, distr_noblock.tables[0].time[-1] + 1, 5)
+        assert np.allclose(distr_noblock.num_coalesced(t), distr_block.num_coalesced(t))
```
jeromekelleher, line 1020: Using ``np.testing.assert_allclose(..)`` is a bit more helpful as it'll give some info on what's different. should be an easy search and replace on this file (since it's new)

python/tests/test_coaltime_distribution.py
```diff
@@ -1016,24 +1167,39 @@ class TestSpanNormalisedCoalescenceTimeTable(TestCoalescenceTimeDistribution):
...
+                    for t in ts.trees()
+                ]
+            )
+            ts = ts.delete_intervals(mask)
         distr = ts.coalescence_time_distribution(
             span_normalise=True,
         )
         return distr
 
     def test_weights(self):
```
jeromekelleher, line 1190: pytest parametrize is better for this sort of thing:
```python
@pytest.mark.parametrize("with_missing_data", [True, False])
def test_weights(self, with_missing_data):
     ... as before
```

jeromekelleher (commented):
LGTM. I haven't gone through the details of the algorithm, but I'm very happy to merge this as a private experimental feature that we can iterate on before documenting.

I left a few minor comments re testing.

I'm not sure what's happened to the docs build - I suspect it may be easier to create a new PR from a different branch, as some package caches or something might have gotten out of date. Alternatively it could be transient and it'll be fixed on your next push.

jeromekelleher:
No point in making a new PR @nspope - the docs build is broken for me too over in #2700. Filing an issue.

nspope:
I think this is good to go @jeromekelleher.

python/tests/test_coaltime_distribution.py
jeromekelleher: Doesn't look like you made this change @nspope?

jeromekelleher:
It'd be nice to do the parametrize thing @nspope. Probably good to squash down the commits a bit, there's no point in saving things like "retrigger codecov" for posterity.

python/tests/test_coaltime_distribution.py
nspope: whoops, I made the change to the identical pattern here: https://github.com/tskit-dev/tskit/blob/604680259e9996a0d08d671c968d013999bd3381/python/tests/test_coaltime_distribution.py#L1305-L1307

I'll go through and fix

nspope:
Oh sorry, looks like I missed mark.parametrize in a few places on the last commit, thanks for catching. Should be good now.

jeromekelleher (approved):

jeromekelleher:
@mergifyio rebase

Issue #996, opened 2020-11-13
Windowed and non-span-normalised versions of kc_distance() and friends


hyanwong:
We would like to be able to report on the distance between two tree sequences (e.g. `kc_distance` and, [if implemented](https://github.com/tskit-dev/tskit/issues/995), `rf_distance`) both (a) calculated only at sites that exist in both tree sequences and (b) in windows along the genome.

We should make use of the conventions & implementation that currently exist for single tree statistics - for instance, (a) is similar to the normal stats calls with `span_normalize=False` with `mode="site"`, and the windowing functionality can be taken straight from the stats functions.

jeromekelleher:
Agreed, we should make these methods consistent wherever possible.

It would be good to see a concrete proposal for what this would look like @hyanwong - what are the arguments we'd need for the tree distance metrics and what would be the return types?

Issue #995, opened 2020-11-13
Add incremental tree-sequence-wide Robinson-Foulds distance (weighted & unweighted)


grahamgower:
Something like this:
https://gist.github.com/grahamgower/8ca50b9d889e4691d0faca7ace93dac2

hyanwong:
A +++ vote for this from me! This should complement the `kc_distance()` method that we currently have. Eventually we might hope to have versions that are windowed too, see #996

hyanwong:
Also note that although often (and rightly) criticised for lack of sensitivity and counterintuitive behaviour, in *practice* it actually performs quite well for ARG comparisons: https://doi.org/10.1093/sysbio/syu085 (esp fig 2): 
 > "The RF metric has been criticized for too rapidly reaching a plateau, but in this study it performed well, as did its branch-length variant RFL. Apparently neither recombinational rearrangement nor deteriorating tree inference easily drive RF to its maximum, ... This .. shows that theoretical reasoning about distance metrics is not a good substitute for testing their performance."

We also find that the RF metric comparing 2 tree sequences is quite a good discriminator. It would be useful to know the validity of the metric for trees with polytomies - I have seen conflicting information about this (e.g. the documentation in the R package 'phangorn` claims that polytomies invalidate the RF metric, but in practise they don't seem to, in my experience. Do you know anything about this @grahamgower ?

hyanwong:
A couple of extra observations. The elegant code posted by @grahamgower doesn't have an L0 (i.e. unweighted, topology only) distance. If I stick it in, I don't seem to get the right answer for the rooted unweighted RF. E.g. I think the following should give a distance of 10.0

```
>>> ts1, ts2 = msprime.simulate(10, random_seed=1, num_replicates=2)
>>> print(rf_distance(orig_ts, cmp_ts, lambda a, b : branches_pnorm(a, b, 0)))
23.0
```

Also, sadly, doing this for larger trees with recombinations is about 3 times slower than loading the trees into Dendropy and running `dendropy.calculate.treecompare.symmetric_difference` on each genome segment.

jeromekelleher:
> Also, sadly, doing this for larger trees with recombinations is about 3 times slower than loading the trees into Dendropy and running dendropy.calculate.treecompare.symmetric_difference on each genome segment.

@grahamgower did say the code wasn't efficient, so no surprises there.

grahamgower:
[...]

PR #2643, opened 2022-11-21
Add unweighted_rf_distance (symmetric_distance) function and tests

Billyzhang1229:
Add  Unweighted Robinson–Foulds Distance (symmetric difference of two trees) function with tests 

See also:
- #995
- #996

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2937,6 +2937,37 @@ class Tree:
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def _get_sample_sets(self):
+        ret = {}
+        for u in self.nodes(order="postorder"):
+            u_sample_set = set()
+            if self.is_sample(u):
+                u_sample_set.add(u)
+            for v in self.children(u):
+                u_sample_set |= ret[v]
+            ret[u] = frozenset(u_sample_set)
+        return ret
+
+    def rf_distance(self, other):
+        """
+        Returns the Robinson-Foulds distance between the specified pair of trees.
+
+        .. seealso::
+            See `Robinson & Foulds (1981)
+            <https://doi.org/10.1016/0025-5564(81)90043-2>`_ for more details.
+
+        :param Tree other: The other tree to compare to.
+        :return: The computed Robinson-Foulds distance between this tree and other.
+        :rtype: int
+        """
+        if self.num_roots != 1 or other.num_roots != 1:
+            raise ValueError("Trees must have a single root")
+
+        s1 = set(self._get_sample_sets().values())
+        s2 = set(other._get_sample_sets().values())
+
+        return len(s1.symmetric_difference(s2))
+
     def path_length(self, u, v):
         """
         Returns the path length between two nodes
--- a/python/tests/test_distance_metrics.py
+++ b/python/tests/test_distance_metrics.py
@@ -1,0 +1,215 @@
+# MIT License
+#
+# Copyright (c) 2023 Tskit Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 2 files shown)


benjeffery:
@Billyzhang1229 The docs error here is fixed in `main` you just need to rebase.

Billyzhang1229:
> @Billyzhang1229 The docs error here is fixed in `main` you just need to rebase. I can do that if needed - just ask.

@benjeffery Thank you so much! I think I could do the rebase part. But I have question about the coverage report, the report shows some decrease for the files that I didn't change, do I need to worry about it?

benjeffery:
> > @Billyzhang1229 The docs error here is fixed in `main` you just need to rebase. I can do that if needed - just ask.
> 
> @benjeffery Thank you so much! I think I could do the rebase part. But I have question about the coverage report, the report shows some decrease for the files that I didn't change, do I need to worry about it?

codecov has been pretty unreliable in reporting numbers. The comments on the PR are usually good. I'll double check this before merge, so don't worry about it.

jeromekelleher:
This looks great @Billyzhang1229, thanks! 

I think the main thing we need here is a bit more testing to reassure ourselves that what we're computing is the same as what'd get using existing methods. How about a small simulation based test like:
```
@pytest.mark.parametrize("n", [2, 3, 5, 10, 20])
def test_rf_distance_against_dendropy(n):
      trees = []
      for seed in [42, 43]:
           ts = msprime.sim_ancestry(n, ploidy=1, random_seed=seed)
           trees.append(ts.first())
       rf1 = trees[0].rf_distance(trees[1])
       rf2 = dendropy_rf_distance(*trees)
       assert rf1 == rf2
```

Other than that, we just need to think about the naming and be a bit more precise about the documentation - but I think myself and @hyanwong can help with that. @hyanwong - thoughts about function naming here?

Billyzhang1229:
@jeromekelleher Thanks for the feedback! I agree that additional testing would be beneficial to ensure the accuracy of our computations.

## Naming

I am currently using `symmetric_distance()` and `unweighted_rf_distance()` for function names, the proper name should be `unweighted_rf_distance`. The term `symmetric_distance` is a made-up term used to represent the symmetric difference between trees. 

I'm not sure which one is better for naming. So I just wrapped the `symmetric_distance()` with `unweighted_rf_distance()`.

```python
def unweighted_rf_distance(self, other):
    return self.symmetric_distance(other)
```

In the future, we may also want to consider adding a `weighted_rf_distance()` function or simply `rf_distance()`?
___
If there is anything else you would like me to add or change, please let me know.

jeromekelleher:
I think ``rf_distance`` is a good because it aligns with our existing ``Tree.kc_distance`` method. We can document that it's the unweighted distance, and if we like later can allow for the weighted version via a ``weighted`` argument.

I don't think  ``symmetric_distance`` as a synonym helps much and I would drop it, as people are familiar with RF distance.

We will need to be quite precise about what we're computing here in the documentation, but I can do that in a follow-on commit.

hyanwong:
I agree with Jerome. `rf_distance` allows us to implement `rf_distance(weighted=True)` later. Everyone calls it the RF distance. We can mention the term "symmetric_distance" in the docstring to make it searchable.

hyanwong:
In fact, if we want to, we could allow the `weighted` parameter and simply raise a NotImplementedError if it's `True`, simply to give us a reminder to implement it every time we use it! But that's probably an abuse of the API :)

python/tskit/trees.py
```diff
@@ -2903,6 +2903,52 @@ def kc_distance(self, other, lambda_=0.0):
         """
         return self._ll_tree.get_kc_distance(other._ll_tree, lambda_)
 
+    def _get_sample_sets(self):
+        ret = collections.defaultdict(set)
+        for u in self.samples():
```
jeromekelleher, line 2908: We should try to avoid making a separate copy of these as frozen sets. How about:
```
def _get_sample_sets(self):
     ret = {}
     for u in self.nodes(order="postorder"):
           u_sample_set = set()
           if self.is_sample(u):
                u_sample_set.add(u)
           for  v in self.children(u):
                 u_sample_set |= ret[v]
            ret[u] = frozenset(u_sample_set)
```
Then we should be able to work directly with the return value of this function?

jeromekelleher:
> In fact, if we want to, we could allow the weighted parameter and simply raise a NotImplementedError if it's True, simply to give us a reminder to implement it every time we use it! But that's probably an abuse of the API :)

I don't see much point in this - it's forward compatible by not having the argument and we don't need it for this PR.

hyanwong:
Sorry that we haven't got around to merging this yet @Billyzhang1229. It would be great to get this in, and I see you've added extra tests. Do you know how this performs in the case of trees with polytomies? There's an interesting publication at https://onlinelibrary.wiley.com/doi/full/10.1111/cla.12540 that sounds like it might give some useful pointers in this case.

benjeffery:
@mergify rebase

benjeffery (approved):

benjeffery:
@mergifyio rebase

PR #2646, opened 2022-11-24
Allow canvas_size to be specified in draw_svg

### LINKED ISSUE #2645: SVG canvas scaling and .replace method
A few times now, myself and others (e.g. @gtsambos) have wanted to make the SVG canvas a little bigger, e.g. to accommodate long or rotated labels, etc.

There's a quick hack to do this as follows:

```python
import tskit
from IPython.display import display, SVG
w = 300
h = 200
tree = tskit.Tree.generate_balanced(10)
svg = tree.draw_svg(
    size=(w, h),
    style=".leaf > .lab {text-anchor: start; transform: rotate(90deg) translate(6px)}",
    node_labels={u: "abcedfghij"[u] * 10 for u in ts.samples()},
)
display(svg) # labels cut off

# Hack to make room for labels
taller_svg = svg.replace(f'height="{h}"', f'height="{h+50}"', 1)
display(SVG(taller_svg))
```

So two things here:

1. should we implement the `.replace` method on the tskit.SVGstring class, so ti returns another SVGstring? That would save a small amount of hassle, but I don't want to be implementing all the other standard string methods that return a string. I guess this would be just
   ```
   def replace(self, *args, **kwargs):
       return tskit.SVGString(self.replace(*args, **kwargs))
   ```
2. should we be able to specify a different canvas size from the `size()` passed to `draw_svg` to avoid having to hack it altogether?

hyanwong:
## Description

Fixes #2645

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -8,6 +8,10 @@
   which are the minimum and maximum among the node times and mutation times,
   respectively. (:user:`szhan`, :pr:`2612`, :issue:`2271`)
 
+- The ``draw_svg`` methods now accept a ``canvas_size`` parameter to allow
+  extra room on the canvas e.g. for long labels or repositioned graphical
+  elements (:user:`hyanwong`, :pr:`2646`, :issue:`2645`)
+
 **Breaking Changes**
 
  - the ``filter_populations``, ``filter_individuals``, and ``filter_sites``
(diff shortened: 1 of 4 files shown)


benjeffery:
This was a transient CI fail.

benjeffery (approved):
LGTM!

jeromekelleher (approved):
LGTM

PR #2650, opened 2022-11-25
Add "title" class to axes titles

hyanwong:
## Description

A trivial but useful change that allows us to hide or move axis titles easily. Annoyingly it changes the example SVG files slightly, so there's a bit of data churn here.


jeromekelleher (approved):
LGTM - needs some documentation I guess?

hyanwong:
This sort of thing tends to get documented in the viz tutorial, as it's a bit involved for the docstring. I have a PR in the tutorials repo which includes a mention of it, but there's no point merging it until this comes out in the next tskit release.

jeromekelleher:
We should have some local documentation for the SVG structure though. Perhaps we should have a local "visualisation" page which documents the SVG structure and gives quick examples of how to use the various SVG elements?  So, the tutorial is about "how to do specific things" whereas the documentation is "document all the features"? Don't have to do it here, but we should track with an issue.

benjeffery (approved):

PR #2651, opened 2022-11-28
extend edges method

petrelharp:
Here we (me and @hfr1tz3 and @avabamf) have implemented the "extend_edges" and related things in tskit.

Here's the description of what this does (copied from the draft docstring):
```
        Returns a new tree sequence in which the span covered by ancestral nodes
        is "extended" to regions of the genome according to the following rule:
        If an ancestral segment corresponding to node `n` has parent `p` and
        child `c` on some portion of the genome, and on an adjacent segment of
        genome `p` is the immediate parent of `c`, then `n` is inserted into the
        edge from `p` to `c`. This involves extending the span of the edges
        from `p` to `n` and `n` to `c` and reducing the span of the edge from
        `p` to `c`. Since the latter edge may be removed entirely, this process
        reduces (or at least does not increase) the number of edges in the tree
        sequence.
        The method works by iterating over the genome to look for edges that can
        be extended in this way; the maximum number of such iterations is
        controlled by ``max_iter``.
        The rationale is that we know that `n` carries a portion of the segment
        of ancestral genome inherited by `c` from `p`, and so likely carries
        the *entire* inherited segment (since the implication otherwise would
        be that distinct recombined segments were passed down separately from
        `p` to `c`).
```

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -7644,3 +7644,400 @@ out:
     tsk_safe_free(sample_index_map);
     return ret;
 }
+
+/* ======================================================== *
+ * Extend edges
+ * ======================================================== */
+
+typedef struct _edge_list_t {
+    tsk_id_t edge;
+    // the `extended` flags records whether we have decided to extend
+    // this entry to the current tree?
+    bool extended;
+    struct _edge_list_t *next;
+} edge_list_t;
+
+static int
+extend_edges_append_entry(
+    edge_list_t **head, edge_list_t **tail, tsk_blkalloc_t *heap, tsk_id_t edge)
+{
+    int ret = 0;
+    edge_list_t *x = NULL;
+
+    x = tsk_blkalloc_get(heap, sizeof(*x));
+    if (x == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    x->edge = edge;
+    x->extended = false;
+    x->next = NULL;
+
+    if (*tail == NULL) {
+        *head = x;
+    } else {
+        (*tail)->next = x;
+    }
+    *tail = x;
+out:
+    return ret;
+}
+
+static void
+remove_unextended(edge_list_t **head, edge_list_t **tail)
+{
+    edge_list_t *px, *x;
+
+    px = *head;
+    while (px != NULL && !px->extended) {
+        px = px->next;
+    }
+    *head = px;
+    if (px != NULL) {
+        px->extended = false;
+        x = px->next;
+        while (x != NULL) {
+            if (x->extended) {
+                x->extended = false;
+                px->next = x;
+                px = x;
+            }
+            x = x->next;
+        }
+        px->next = NULL;
+    }
+    *tail = px;
+}
+
+static int
+tsk_treeseq_extend_edges_iter(
+    const tsk_treeseq_t *self, int direction, tsk_edge_table_t *edges)
+{
+    // Note: this modifies the edge table, but it does this by (a) removing
+    // some edges, and (b) extending left/right endpoints of others,
+    // while keeping order the same, and so this maintains sortedness
+    // (so, there is no need to sort afterwards).
+    int ret = 0;
+    tsk_id_t tj;
+    tsk_id_t e, e_out, e_in;
+    tsk_id_t c, p, p_in;
+    tsk_blkalloc_t edge_list_heap;
+    double *near_side, *far_side;
+    edge_list_t *edges_in_head, *edges_in_tail;
+    edge_list_t *edges_out_head, *edges_out_tail;
+    edge_list_t *ex_out, *ex_in;
+    double there, left, right;
+    bool forwards = (direction == TSK_DIR_FORWARD);
+    tsk_tree_position_t tree_pos;
+    bool valid;
+    const tsk_table_collection_t *tables = self->tables;
+    const tsk_size_t num_nodes = tables->nodes.num_rows;
+    const tsk_size_t num_edges = tables->edges.num_rows;
+    tsk_id_t *degree = tsk_calloc(num_nodes, sizeof(*degree));
+    tsk_id_t *out_parent = tsk_malloc(num_nodes * sizeof(*out_parent));
+    tsk_bool_t *keep = tsk_calloc(num_edges, sizeof(*keep));
+    bool *not_sample = tsk_malloc(num_nodes * sizeof(*not_sample));
+
+    tsk_memset(&edge_list_heap, 0, sizeof(edge_list_heap));
+    tsk_memset(&tree_pos, 0, sizeof(tree_pos));
+
+    if (keep == NULL || out_parent == NULL || degree == NULL || not_sample == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    tsk_memset(out_parent, 0xff, num_nodes * sizeof(*out_parent));
+
+    ret = tsk_blkalloc_init(&edge_list_heap, 8192);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_tree_position_init(&tree_pos, self, 0);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_edge_table_copy(&tables->edges, edges, TSK_NO_INIT);
+    if (ret != 0) {
+        goto out;
+    }
+
+    for (tj = 0; tj < (tsk_id_t) tables->nodes.num_rows; tj++) {
+        not_sample[tj] = ((tables->nodes.flags[tj] & TSK_NODE_IS_SAMPLE) == 0);
+    }
+
+    if (forwards) {
+        near_side = edges->left;
+        far_side = edges->right;
+    } else {
+        near_side = edges->right;
+        far_side = edges->left;
+    }
+    edges_in_head = NULL;
+    edges_in_tail = NULL;
+    edges_out_head = NULL;
+    edges_out_tail = NULL;
+    e_out = 0; // only to avoid an 'maybe uninitialized' compile warning
+
+    if (forwards) {
+        valid = tsk_tree_position_next(&tree_pos);
+    } else {
+        valid = tsk_tree_position_prev(&tree_pos);
+    }
+
+    while (valid) {
+        left = tree_pos.interval.left;
+        right = tree_pos.interval.right;
+        there = forwards ? right : left;
+
+        // remove entries that aren't being extended/postponed
+        // and update out_parent
+        for (ex_out = edges_out_head; ex_out != NULL; ex_out = ex_out->next) {
+            e = ex_out->edge;
+            out_parent[edges->child[e]] = TSK_NULL;
+        }
+        remove_unextended(&edges_in_head, &edges_in_tail);
+        remove_unextended(&edges_out_head, &edges_out_tail);
+        for (ex_out = edges_out_head; ex_out != NULL; ex_out = ex_out->next) {
+            e = ex_out->edge;
+            out_parent[edges->child[e]] = edges->parent[e];
+        }
+
+        for (tj = tree_pos.out.start; tj != tree_pos.out.stop; tj += direction) {
+            e = tree_pos.out.order[tj];
+            if (out_parent[edges->child[e]] == TSK_NULL) {
+                // add edge to pending_out
+                ret = extend_edges_append_entry(
+                    &edges_out_head, &edges_out_tail, &edge_list_heap, e);
+                if (ret != 0) {
+                    ret = TSK_ERR_NO_MEMORY;
+                    goto out;
+                }
+                out_parent[edges->child[e]] = edges->parent[e];
+            }
+        }
+        for (tj = tree_pos.in.start; tj != tree_pos.in.stop; tj += direction) {
+            e = tree_pos.in.order[tj];
+            // add edge to pending_in
+            ret = extend_edges_append_entry(
+                &edges_in_head, &edges_in_tail, &edge_list_heap, e);
+            if (ret != 0) {
+                ret = TSK_ERR_NO_MEMORY;
+                goto out;
+            }
+        }
+        for (ex_out = edges_out_head; ex_out != NULL; ex_out = ex_out->next) {
+            e_out = ex_out->edge;
+            degree[edges->parent[e_out]] -= 1;
+            degree[edges->child[e_out]] -= 1;
+            tsk_bug_assert(out_parent[edges->child[e_out]] == edges->parent[e_out]);
+        }
+        for (ex_in = edges_in_head; ex_in != NULL; ex_in = ex_in->next) {
+            e_in = ex_in->edge;
+            degree[edges->parent[e_in]] += 1;
+            degree[edges->child[e_in]] += 1;
+        }
+
+        for (ex_in = edges_in_head; ex_in != NULL; ex_in = ex_in->next) {
+            e_in = ex_in->edge;
+            // check whether the parent-child relationship exists in the
+            // sub-forest of edges to be removed:
+            // out_parent[p] != -1 only when it is the bottom of an edge to be
+            // removed, and degree[p] == 0 only if it is not in the new tree
+            c = edges->child[e_in];
+            p = out_parent[c];
+            p_in = edges->parent[e_in];
+            while ((p != TSK_NULL) && (degree[p] == 0) && (p != p_in) && not_sample[p]) {
+                p = out_parent[p];
+            }
+            if (p == p_in) {
+                // we can extend!
+                // But, we might have passed the interval that a
+                // postponed edge in covers, in which case
+                // we should skip postponing the edge in
+                if (far_side[e_in] != there) {
+                    ex_in->extended = true;
+                }
+                near_side[e_in] = there;
+                while (c != p) {
+                    for (ex_out = edges_out_head; ex_out != NULL;
+                         ex_out = ex_out->next) {
+                        e_out = ex_out->edge;
+                        if (edges->child[e_out] == c) {
+                            break;
+                        }
+                    }
+                    tsk_bug_assert(edges->child[e_out] == c);
+                    ex_out->extended = true;
+                    far_side[e_out] = there;
+                    // amend degree: the intermediate
+                    // nodes have 2 edges instead of 0
+                    tsk_bug_assert(degree[c] == 0 || c == edges->child[e_in]);
+                    if (degree[c] == 0) {
+                        degree[c] = 2;
+                    }
+                    c = out_parent[c];
+                }
+            }
+        }
+        if (forwards) {
+            valid = tsk_tree_position_next(&tree_pos);
+        } else {
+            valid = tsk_tree_position_prev(&tree_pos);
+        }
+    }
+
+    for (e = 0; e < (tsk_id_t) num_edges; e++) {
+        keep[e] = edges->left[e] < edges->right[e];
+    }
+    ret = tsk_edge_table_keep_rows(edges, keep, 0, NULL);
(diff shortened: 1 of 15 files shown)


python/tskit/trees.py
```diff
@@ -6829,9 +6829,9 @@ def _forward_extend(self, forwards=True):
         #     edgediffs = reversed(self.edge_diffs())
         # skip first tree
         num_edges = np.full(self.num_nodes, 0)
-        (_, edges_out, edges_in) = next(edgediffs)
-        assert len(edges_out) == 0
-        for e in edges_in:
+        (_, Edges_out, Edges_in) = next(edgediffs)
+        assert len(Edges_out) == 0
+        for e in Edges_in:
```
petrelharp, line 6834: I don't think this should change anything - does it?

python/tskit/trees.py
```diff
@@ -6857,24 +6857,27 @@ def _forward_extend(self, forwards=True):
...
+                
+            # Find smallest length right enpoint of all edges in edges_in and edges_out
+            # right should equal the endpoint of a T_k 
             right = self.sequence_length
-            if tj < M:
-                right = min(right, edges.left[I[tj]])
             if tk < M:
                 right = min(right, edges.right[O[tk]])
+            if tj < M:
+                right = min(right, edges.left[I[tj]])
```
petrelharp, line 6880: I also don't think this change in order should change anything?

petrelharp:
@hfr1tz3 I think you need to add `.ipynb_checkpoints` to the `.gitignore` (and un-do the adding of that last file); something like
```
cd python/tskit
git rm .ipynb_checkpoints/trees-checkpoint.py
echo ".ipynb_checkpoints" >> .gitignore
git add .gitignore
git commit -m 'remove checkpoint thing'
```
should do it.

python/tskit/.ipynb_checkpoints/trees-checkpoint.py
```diff
@@ -6863,12 +6865,14 @@ def _forward_extend(self, forwards=True):
                 edges_out.append(tk)
```
petrelharp, line 6865: Doh! Good (and, tricky) catch.

Would things be simpler before if we did
```
edges_out.append(O[tk])
```
here, instead of changing the various `tk` to `O[tk]` below? (And, same for `tj`.)

petrelharp: my vote is yes!

petrelharp:
TODO: implement this test (it needs both forwards and backwards):

petrelharp:
TODO:
1. test forwards + backwards python code
2. implement backwards in C also
3. move python code to testing
4. make TreeSequence method call the TableCollection method (which calls the C)
5. test again (which now calls C code)

petrelharp:
Reminder: once c code is edited, to test changes locally you've got to run `make` in the `python/` subdir. (And, currently C code isn't being run.)

python/tskit/trees.py
```diff
@@ -6836,8 +6836,8 @@ def _extend(self, forwards=True):
             I = self.indexes_edge_insertion_order
             O = self.indexes_edge_removal_order
         else:
-            I = np.flip(self.indexes_edge_insertion_order)
-            O = np.flip(self.indexes_edge_removal_order)
+            I = np.flip(self.indexes_edge_removal_order)
+            O = np.flip(self.indexes_edge_insertion_order)
```
petrelharp, line 6840: ah! good catch

hfr1tz3: We're glad it was an easy fix

petrelharp:
I've had a good think about how to use the trees to make the algorithm not O(k^3), and it's possible but not obvious (to me yet). Unless it becomes obvious I'd like to wait to see if it's an problem in practice.

And, I don't know what's up with codecov? The actions say they've uploaded the results, e.g. to https://app.codecov.io/github/tskit-dev/tskit/commit/8973c7b86ada7a19decd9f88031ea045dfd95e6a but at [that link](https://app.codecov.io/github/tskit-dev/tskit/commit/8973c7b86ada7a19decd9f88031ea045dfd95e6a) it says that
```
Commit YAML is invalid
Coverage data is unable to be displayed, as the commit YAML appears to be invalid.
```

petrelharp:
Note: to run the code on this branch locally (in, say, a different project):
- make sure you've got this branch checked out
- go to the `python/` subdirectory
- run `make`
- run `pip install .`

This should replace tskit with this one; to make sure you're using the right version you can first uninstall tskit, so if it didn't work then it'll be obvious.

petrelharp:
TODO:

- add this to the docs
- use the tree positioning stuff (see e.g. around l.60 in  `python/tests/test_tree_positioning.py` for how to use it) to do the iteration

hfr1tz3:
I am trying to run new empirical tests with our edge extend, but I am getting an error for some reason.
TypeError: cannot unpack non-iterable TreeSequence object
We should clear this up on Thursday.

petrelharp:
Can you post the code that produces that error, and/or the information above the error that says at whichi point the error occurs?

hfr1tz3:
> Can you post the code that produces that error, and/or the information above the error that says at whichi point the error occurs?

In last_truth on our num_edges repository cell 3, I installed our extend edges method and tried to run the code. However, I think I may be just calling the function wrong. I will look into it more tomorrow

petrelharp:
I have - after quite a bit of staring at things - modified the algorithm to use the new "tree position" method of iterating over trees. (The staring did result in better code and more robust tests, however.) However, I need some input here - probably from @jeromekelleher? So, right now (a) the python implementation in test_topology.py uses "tree position", but (b) the c version in `tables.c` uses `I`/`O`; and (c) in `trees.c` there is a C translation that I got most of the way through porting over before realizing I needed some guidance, namely:

The `tsk_tree_position_t` stuff needs a *tree sequence*, not just tables. This makes sense since it has to do with trees. However, tree sequence methods return a whole new tree sequence, while this algorithm pretty naturally modifies things more or less in place - it only makes a copy of the edge table. In fact, it rebuilds the edge table and the indexes a bunch of times. I feel like it is more natural for this reason to have extend edges be a table collection method (that operates in-place); but then to use the tree position thing we'll have to build a tree sequence within the function (and, `tables.c` doesn't even import `trees.h`). What's the right way to set this up?

c/tskit/trees.c
```diff
@@ -6803,3 +6803,268 @@ tsk_treeseq_divergence_matrix(const tsk_treeseq_t *self, tsk_size_t num_samples,
...
+        sign_int = -1;
+        near_side = edges.right;
+        far_side = edges.left;
+    }
+    edges_in_head = NULL;
+    edges_in_tail = NULL;
+    edges_out_head = NULL;
+    edges_out_tail = NULL;
+
+    if (forwards) {
```
jeromekelleher, line 6938: @duncanMR what do you think about adding something like ``tsk_tree_position_step(&tree_pos, direction)`` for this? That would make it quite neat for things like this.

c/tskit/trees.c
```diff
@@ -6803,3 +6803,268 @@ tsk_treeseq_divergence_matrix(const tsk_treeseq_t *self, tsk_size_t num_samples,
...
+    while (valid) {
+        left = tree_pos.interval.left;
+        right = tree_pos.interval.right;
+        there = forwards ? right : left;
+
+        // remove entries that aren't being extended/postponed
+        remove_unextended(&edges_in_head, &edges_in_tail);
+        remove_unextended(&edges_out_head, &edges_out_tail);
+
+        for (tj = tree_pos.out.start; tj != tree_pos.out.stop; tj += sign_int) {
```
jeromekelleher, line 6953: You could use 
```c
for (tj = tree_pos.out.start; tj != tree_pos.out.stop; tj += tree_pos.direction) {
}
```
which is partly the intention of having ``direction`` done like it is.

jeromekelleher (commented):
I haven't got my head around the algorithm fully, but it's much easier to understand now!

I vote for ditching the ``tsk_table_collection`` version - surely this is a premature optimisation?

c/tskit/trees.c
duncanMR: That would be a useful thing to have in the API and trivial to implement. I'd be happy to do it

petrelharp:
It seems like the way to do this (eg in `compute_mutation_parents`) would be to (a) make a tree sequence from the tables; (b) do the stuff to the tables; (c) forget about the tree sequence, leaving just the tables. I just don't have my head entirely around whether that's the right way to go.

jeromekelleher:
> I totally agree that the performance here isn't critical. But I do think that there's a conceptual hole here: how do we write a method that modifies tables in place and uses the trees to do so? (I mean, there's no technical barrier to that, but that's not how the API is set up.) A good example of something where that's exactly what we need to do is compute_mutation_parents, which uses the I/O method of constructing the trees.

I think the number of things we need to do with the tables in place that require iterating over the trees will be pretty small, going forward. So, I'd vote we just leave stuff like ``compute_mutation_parents`` etc as they are, and concentrate on using the tree_pos stuff for things that are logically operations on a tree sequence.

We don't *have* to depend on the tree sequence for the tree_position thing --- we're really only using the breakpoints array, which we only really need for long-range jumping. But, I think the abstraction is getting very leaky when we do things on both the tree sequence and table collection, just for the sake of avoiding a copy, or creating a ``treeseq_t`` object (which can be as cheap as validating a set of tables has the required tree properties).

petrelharp:
> We don't have to depend on the tree sequence for the tree_position thing --- we're really only using the breakpoints array, which we only really need for long-range jumping. But, I think the abstraction is getting very leaky when we do things on both the tree sequence and table collection, just for the sake of avoiding a copy, or creating a treeseq_t object (which can be as cheap as validating a set of tables has the required tree properties).

Seems like then if we were to write something like `compute_mutation_parents` with tree_positions, then we'd create a `treeseq_t` object - that seems fine? I am pretty tempted to do things that way here, because (a) it'd take less re-plumbing, and (b) one of the use cases for this method is that it compresses the tree sequence.

jeromekelleher:
I've gone over this and refactored things a bit. The memory management at the C level was really quite tricky, but hopefully it's settled now.

Some high-level stuff:

- There's some problem with mutations. I think you probably need to keep track of the edges that are extended, and then remap the mutations?
- Testing is pretty weak. I think we need a few more fully worked examples where we have the input and the exact expected output, after one or two iterations. I know this is tedious, but it really pays off in terms of helping the code reader understand what's going on 
- We're missing some checks like "the other tables shouldn't be affected" (but see point above about mutations)

c/tskit/trees.c
```diff
@@ -6803,3 +6803,280 @@ tsk_treeseq_divergence_matrix(const tsk_treeseq_t *self, tsk_size_t num_samples,
...
+    // Note: this modifies the edge table, but it does this by (a) removing
+    // some edges, and (b) extending left/right endpoints of others,
+    // while keeping order the same, and so this maintains sortedness
+    // (so, there is no need to sort afterwards).
+    int ret = 0;
+    tsk_id_t tj;
+    tsk_id_t e, e1, e2, e_in;
+    tsk_blkalloc_t edge_list_heap;
+    double *near_side, *far_side;
+    edge_list_t *edges_in_head, *edges_in_tail;
```
jeromekelleher, line 6884: If I was starting this from scratch, I would't bother with linked lists and do something like this instead:
```
tsk_bool_t extended = calloc(num_edges, sizeof(*extended));
tsk_id_t *edges_in = mallco(num_edges, sizeof(*edges_in));
tsk_id_t *edges_out = mallco(num_edges, sizeof(*edges_out));
tsk_size_t num_edge_in, num_edges_out;
```

and keep track of the lists like that. Repacking to get rid of unextended edges will be slightly trickier, but there's less faddling around with linked lists, memory allocation and so on, and it would almost certainly be more efficient.

But like I say - you might not want to get into this, since it's working.

and keep track of th

c/tskit/trees.h
```diff
@@ -891,6 +891,39 @@ int tsk_treeseq_simplify(const tsk_treeseq_t *self, const tsk_id_t *samples,
...
+If an ancestral segment corresponding to node `n` has parent `p` and
+child `c` on some portion of the genome, and on an adjacent segment of
+genome `p` is the immediate parent of `c`, then `n` is inserted into the
+edge from `p` to `c`. This involves extending the span of the edges
+from `p` to `n` and `n` to `c` and reducing the span of the edge from
+`p` to `c`. Since the latter edge may be removed entirely, this process
+reduces (or at least does not increase) the number of edges in the tree
+sequence.
+
+The method works by iterating over the genome to look for edges that can
```
jeromekelleher, line 908: Debatable whether we want to bother documenting the C API for this also?

jeromekelleher: Would save a bit of trouble worrying about , .e.g. whether ``max_iter`` should be an int, or some other type.

jeromekelleher (commented):
A few minor comments. I'm not sure what the right pattern will be in order to update the mutation table too, but I guess that'll take some figuring out first.

jeromekelleher:
One thing that might not be immediately obvious - I made a new file ``test_extend_edges.py``

petrelharp:
Ah, I see about the memory management. Makes sense.

petrelharp:
TODO: put mutations on the edges; proposed method:
- for each edge to be postponed,
- for each mutation on said edges,
- figure out what edge the mutation *should* be on

petrelharp:
Okay! I think this is ready to go:

0. I have re-written the algorithm so it is no longer O(n^3), where n is the number of edges differing between adjacent trees. Now it is O(n), in the usual case anyhow. This should make it more possible to apply it to Relate-produced tree sequences.
1. It does mutations now: it currently just errors if mutations do not have times, and directs the user to `impute_mutation_times`; we could always just keep mutations at the nodes they are assigned to, which would be equivalent to the (currently default) method of `impute_mutation_times` that gives mutations the time of their nodes; but it seems statistically much better to distribute mtuation times evenly over the edges, so we're making the user do that step explicitly.
2. More tests are possible, but I feel good about it the current set of them.
3. Migrations are a can of worms, so they are disallowed.

petrelharp:
Darn it, looks like we need more test coverage.

python/tests/test_lowlevel.py
```diff
@@ -1496,6 +1496,13 @@ def test_time_units(self):
             ts.load_tables(tables)
             assert ts.get_time_units() == value
 
+    def test_extend_edges_bad_args(self):
+        ts1 = self.get_example_tree_sequence(10)
+        with pytest.raises(TypeError):
+            ts1.extend_edges()
+        with pytest.raises(TypeError, match="an integer"):
+            ts1.extend_edges("sdf")
```
jeromekelleher, line 1504: Need to add a test for ``max_iter=0`` here (or something else that will trigger a C-level error). Otherwise error handling code won't come up as covered.

c/tskit/trees.c
```diff
@@ -7644,3 +7644,417 @@ tsk_treeseq_divergence_matrix(const tsk_treeseq_t *self, tsk_size_t num_samples,
...
+
+    ret = tsk_tree_position_init(&tree_pos, self, 0);
+    if (ret != 0) {
+        goto out;
+    }
+
+    valid = tsk_tree_position_next(&tree_pos);
+    next_mut = 0;
+
+    while (valid) {
```
jeromekelleher, line 7935: Some comment as Python code here - why not just use the standard tree API? Any perf difference will be neligible.

c/tskit/trees.h
jeromekelleher: So, we're adding this to the next C API release then? We should update the C changelog so.

c/tskit/trees.c
```diff
@@ -7644,3 +7644,417 @@ tsk_treeseq_divergence_matrix(const tsk_treeseq_t *self, tsk_size_t num_samples,
...
+    tsk_table_collection_t tables;
+    tsk_treeseq_t ts;
+    int iter, j;
+    tsk_size_t last_num_edges;
+    const int direction[] = { TSK_DIR_FORWARD, TSK_DIR_REVERSE };
+
+    tsk_memset(&tables, 0, sizeof(tables));
+    tsk_memset(&ts, 0, sizeof(ts));
+    tsk_memset(output, 0, sizeof(*output));
+
```
jeromekelleher, line 7987: Probably good to check that max_iter > 0. It's not helpful to silently do nothing here I think? Negative max_iter is surely an error anyway.

jeromekelleher (commented):
LGTM. I'm not seeing much you can do to update coverage.

Two minor comments:

- We should probably do some checks on max_iter in the C code. This will actually make it easier to bump up test coverage in, e.g., the Python C API
- Do we really want to error on unspecified mutation times? It would probably be simpler to "min" impute if unknown, and that's what people will end up doing most of the time anyway?


Edit - whoops, just saw your point about impute mutation times above. Ignore comment 2 then!

python/tests/test_extend_edges.py
```diff
@@ -0,0 +1,672 @@
...
+        for j in range(
+            tree_pos.out_range.start, tree_pos.out_range.stop, tskit.FORWARD
+        ):
+            e = tree_pos.out_range.order[j]
+            parent[edges.child[e]] = -1
+
+        for j in range(tree_pos.in_range.start, tree_pos.in_range.stop, tskit.FORWARD):
+            e = tree_pos.in_range.order[j]
+            parent[edges.child[e]] = edges.parent[e]
+
```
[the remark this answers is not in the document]
petrelharp: That was because we prototyped in python first, then moved it to C.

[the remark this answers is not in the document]
petrelharp: Oh I see - okay, really it was because I was copying what we did in the other function.

petrelharp:
Note: https://github.com/tskit-dev/tsdate/pull/317 depends on this.

python/tests/test_extend_edges.py
```diff
@@ -0,0 +1,672 @@
...
+    return ts
+
+
+def _slide_mutation_nodes_up(ts, mutations):
+    # adjusts mutations' nodes to place each mutation on the correct edge given
+    # their time; requires mutation times be nonmissing and the mutation times
+    # be >= their nodes' times.
+
+    assert np.all(~tskit.is_unknown_time(mutations.time)), "times must be known"
```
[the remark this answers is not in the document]
petrelharp: could, but as noted above let's leave this to the end user (we have a note about this in the docstring)

python/tests/test_extend_edges.py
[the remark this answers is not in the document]
petrelharp: Actually - it's not that simple because the mutations aren't actually on the trees - they're passed in via a separate table (because they wouldn't be legal on that tree sequence). But it can be simpler, you're right.

petrelharp:
I think this is ready to go, except for the doc build failure. Codecov says that I'm not testing the 
```
+                 ret = TSK_ERR_DISALLOWED_UNKNOWN_MUTATION_TIME;
```
line, but I [totally am](https://github.com/tskit-dev/tskit/pull/2651/files#diff-72d133a23d6fc9a6b8a292763ebf3fd326ee81cc0b771247a4c28c9546bbaa92R353) - not sure what's up with that?

petrelharp:
Note: I've also done like https://github.com/tskit-dev/msprime/pull/1395 in this PR to get it to pass the docs build.

petrelharp:
Okay! I think this is ready to merge!

jeromekelleher (commented):
LGTM, just spotted on mistake.

I think we can squash down to one commit (or 3 max, if you really want to keep my changes?)

petrelharp:
Okay! I fixed that problem, added another C test that codecov pointed out, rebased & squashed; I think we're ready to go!

jeromekelleher (approved):

PR #2652, opened 2022-11-29
Truncate SVG tree sequence plots by skipping trees

hyanwong:
For long tree sequences, adds a `skipped XXX trees` panel in the middle, rather than plotting all the trees:

Screenshot 2022-11-29 at 14 49 16

I had to extract some of the plotting code into a base class to plot a panel that didn't have a tree or axes etc, so there's a bit of code churn here. I also added the "ax-line" and "ax-skipped" classes to the line(s) drawn to represent the x axis.

Note that some values are being listed as e.g. `-2.84217e-14` rather than `0`, so the `rnd` function written by @benjeffery could probably do with tweaking for the case of zero?


jeromekelleher:
How big is this diff out of curiosity? I worry that these svg changes are adding up.

hyanwong:
> > How big is this diff out of curiosity? I worry that these svg changes are adding up.
> 
> You mean lines of code to maintain in `drawing.py`, or GitHub diffs (e.g. since the saved SVGs are changing)?

If the latter, with the new SVGs, the `tests/data/svg/` directory contains SVG files totalling 432K. Adjustments in the SVG model (e.g. adding an extra class) seem to require changing all (or nearly all) of these, as Git isn't clever enough to work out the diffs.

I think I've changed this twice recently, and this will make a third. I can't see it happening more than about 5 or 6 times a year though, so 3Mb churn per year.

benjeffery (approved):
LGTM - although looks like you're getting different SVG output on windows.

hyanwong:
> LGTM - although looks like you're getting different SVG output on windows.

I think this might be the rounding thing I mentioned above.

hyanwong:
> > LGTM - although looks like you're getting different SVG output on windows.
> 
> I think this might be the rounding thing I mentioned above.

FYI: it's the behaviour of https://github.com/tskit-dev/tskit/blob/ec992daab6ac880a3e6e2bb31e2dddfa463adbd5/python/tskit/drawing.py#L325 for values close to zero.

benjeffery:
Sorry @hyanwong, what is the issue here? The current code does the following with small numbers:
```python
        assert "0" == str(drawing.rnd(0))
        assert "1.23457e-05" == str(drawing.rnd(0.0000123456789))
        assert "1.23457e-07" == str(drawing.rnd(0.000000123456789))
```
Which seems fine?

hyanwong:
> Sorry @hyanwong, what is the issue here? The current code does the following with small numbers:
> 
> ```python
>         assert "0" == str(drawing.rnd(0))
>         assert "1.23457e-05" == str(drawing.rnd(0.0000123456789))
>         assert "1.23457e-07" == str(drawing.rnd(0.000000123456789))
> ```
> 
> Which seems fine?

I think it's what happens to numbers which are actually zero within rounding error, e.g. because some of the new code adds up intervals to get a total width, we end up with some values that used to be zero being expressed as e.g. 4.26326e-14. I wonder if anything below e.g. 1e-10 should be zero. Or perhaps I should just do the rounding at some other point for this particular case.

benjeffery:
Ah ok, so `rnd` is working as designed - it is truncating that value to 6 sig figs correctly. I'd rather coerce small values to zero outside the `rnd` code so that it retains its clear purpose. Is there one or two places you need to do this or is it every value in the SVG?

hyanwong:
It's only one or two places. I can probably round those to zero explicitly. I'll look.

hyanwong:
I fixed the rounding issue and skipped two tests on windows, which seem to fail because weirdly in this instance the msprime windows version gives a slightly different set of sample node IDs (presumably something to do with the GSL rng version)?

PR #2661, opened 2022-12-12
add tsk_tree_index_t

### LINKED ISSUE #2659: Add tsk_tree_seek_index
We never added a method in the C API to seek efficiently to the tree at the kth index. The implementation in the Python API is currently quite naive: https://github.com/tskit-dev/tskit/blob/ec992daab6ac880a3e6e2bb31e2dddfa463adbd5/python/tskit/trees.py#L811

molpopgen:
## Description

Adds tree indexing to tsk_treeseq_t.
The indexes allow efficient advancing to arbitrary trees.

Fixes #2659

# PR Checklist:


molpopgen:
Note to self: I may be not considering the direction of iteration during setup.

jeromekelleher:
Ahhh, now I get it! 

In big O terms, this PR seeks to a random tree in O(num_edges + num_samples), since we iterate over all edges in order to pull out those that intersect with the tree's interval, and then build that tree. We don't incur a ``height`` term here, because the tree gets built from the bottom up, so there's never any propagating of stuff from a newly attached subtree upwards to root.

But, using the standard code I think the complexity is ``O(num_edges + num_samples + num_trees * height)`` (The first tree costs ``num_samples`` to build, in worst case we iterate over all edges and in worst case at each tree we do O(height) work.) which isn't *that* much different. I think you could probably come up with examples where the current linear scan does quite a bit better, since it's rare we visit all edges in the table, this happens for every seek here.

The O(num_edges) cost is the problem, and is what we discussed in #684. Without something that's O(log(num_edges) + num_samples) for seeking to an arbitrary tree, I'm not all that enthusiastic about the extra complexity, as I think it'll be quite hard to figure out how to tune the current approach.

What sort of performance difference are you observing in your Rust experiments @molpopgen?

molpopgen:
I see a general 10x-ish improvement in tree access times unless the trees being sought are those at the edges of the tree sequence.

molpopgen:
@jeromekelleher  -- this is a benchmark result that I uploaded to the tskit Slack way back when.

[benchmark.pdf](https://github.com/tskit-dev/tskit/files/10220704/benchmark.pdf)

molpopgen:
@jeromekelleher -- here is another one from Slack, comparing to the current seek method for a larger tree sequence.  Circles are the current tskit, purple diamonds are my early experiments.

Also, the recent Python prototype is similar in being about an order of magnitude faster for an arbitrary seek (again, except for the trees at the ends).

with_seek

molpopgen:
I just revisited my rust repo, and here is tskit/rust relative performance on a large tree sequence with 10k diploid samples.  The time is the time to set up the i-th tree, measured every 5e3 trees.

relative_perf

molpopgen:
@jeromekelleher  -- I'll save on uploading more images, but I seem to be iterating towards something w/my rust prototype that is always faster than the current implementation, at least for the first half of the trees ;)

petrelharp:
Perhaps you know this, but I got curious, and the problem we want to solve here is the "interval stabbing problem"; here is an approach that is [`O(num_edges  * log(num_edges)`](https://link.springer.com/chapter/10.1007/978-3-642-10631-6_18). (it's O(n) but that's because they assume endpoints are O(n) integers so they can sort by bucket sort; we'd need to sort by left and right endpoint, which we already have in the edge insertion and removal orders?)

jeromekelleher:
Thanks for the plots @molpopgen, these are really helpful. I've no doubt we could do better at the "starting from NULL seek to the kth tree" problem (which is what we're timing here I think, since that's interesting from the parallel computation viewpoint). We probably don't need to store the additional three ``num_trees`` arrays to do this, we just need to skip "adding in" edges that aren't in the tree we're seeking to. If we had a specialised ``seek_from_null`` function that we called internally for this, I think it would work quite well.

jeromekelleher:
Something like this:
```python
    def seek(self, position):
        """
        Seek to the tree at the position.
        """
        sequence_length = self.sequence_length
        M = self.edges_left.shape[0]
        in_order = self.edge_insertion_order
        out_order = self.edge_removal_order
        edges_left = self.edges_left
        edges_right = self.edges_right
        edges_parent = self.edges_parent
        edges_child = self.edges_child

        j = 0
        k = 0
        left = 0
        right = -1

        while j < M or left <= sequence_length:
            while k < M and edges_right[out_order[k]] == left:
                k += 1
            while j < M and edges_left[in_order[j]] == left:
                if edges_left[in_order[j]] <= position < edges_right[in_order[j]]:
                    p = edges_parent[in_order[j]]
                    c = edges_child[in_order[j]]
                    self.insert_edge(p, c)
                j += 1
            right = sequence_length
            if j < M:
                right = min(right, edges_left[in_order[j]])
            if k < M:
                right = min(right, edges_right[out_order[k]])

            if left <= position < right:
                # We're in the target tree
                break
            left = right
       # In the C code we'd now make sure to set the various counters so that we can
       # advance in either direction based on j and k

```

A quick numba version I just knocked up on some arbitrary simulations was 2X faster than the ``ts.at(ts.sequence_length / 2)``.

molpopgen:
Here are some raw timings.  I went nuts and rewrote all this in rust, but I was smarter this time and just made `tsk_tree_insert_edge` public, which save a lot of hassle.

The two methods can be made very close.  The indexing approach has what looks like a 2x advantage.

benchmarks

Some details:

1. Each point is a seek to a tree index.
2. To do this, we are using the indexes generated by methods like in this PR to find the position for tskit to seek or for @jeromekelleher's method.

molpopgen:
I have managed to get rid of the positive slope in my last graphic for the indexed method.  Working on it for the "jk" approach but I may be out of smart pills/time for the day.

molpopgen:
Okay, I have a working prototype and will do a do-over on this PR.  I managed to speed up @jeromekelleher's method by quite a bit -- you can do the same thing w/less logic.  The scale of this graph makes it hard to see, but the `jk` method is at least 25% faster than what I posted yesterday.

benchmarks

molpopgen:
@jeromekelleher -- I pushed a completely new version.  It'd be helpful for you to have a look. I haven't dug into figure out why the failing tests fail.

molpopgen:
Note to self: the final test is failing b/c something is broken with respect to advancing trees. I can reproduce this in my other prototype.

c/tskit/trees.c
```diff
@@ -4551,47 +4551,95 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
+                break;
+            }
+            if (x >= edge_left[e] && x < edge_right[e]) {
+                p = edge_parent[e];
+                c = edge_child[e];
+                tsk_tree_insert_edge(self, p, c, e);
+            }
+        }
+    } else {
+        for (edge = 0; edge < num_edges; ++edge) {
```
jeromekelleher, line 4593: I'm not sure there's actually much point in this - we've documented that users shouldn't depend on the specific order of children within a node (which is what changes when you apply/remove edges in different orders) and anyway, it *still* won't be the same even if we seek in the same direction as the standard code. 

So, I think completely *identical* quintuply linked trees is a non-goal here, because it's essentially impossible.

c/tskit/trees.c
```diff
@@ -4551,47 +4551,95 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
+                tsk_tree_insert_edge(self, p, c, e);
+            }
+        }
+    }
+
+    /* NOTE: this is where I'm a bit nervous */
+
+    /* We need to get the index of the tree in order to update other
+     * data fields. */
+    for (tree_index = 0; tree_index < (tsk_id_t) num_trees - 1; ++tree_index) {
```
jeromekelleher, line 4610: Probably worth binary searching here: https://github.com/tskit-dev/tskit/blob/ec992daab6ac880a3e6e2bb31e2dddfa463adbd5/c/tskit/core.c#L761

c/tskit/trees.c
```diff
@@ -4551,47 +4551,95 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
+        if (breakpoints[tree_index + 1] > x) {
+            break;
+        }
+    }
+    self->index = tree_index;
+    self->interval.left = breakpoints[tree_index];
+    self->interval.right = breakpoints[tree_index + 1];
+    self->left_index = insertion[tree_index];
```
jeromekelleher, line 4618: This isn't right - you're still thinking of the per-tree arrays that you're storing. You need to store `e` here probably.

You need to keep track of the "removal" value too.

jeromekelleher (commented):
Looks great, but I don't think it's correct at the moment. Seeking from either end won't make any difference I think, so I'd just go in one direction to make things simpler.

You need to reason about the edge_removal index also, that's currently missing here.

c/tskit/trees.c
```diff
@@ -4551,47 +4551,95 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
     int ret = 0;
     const double L = tsk_treeseq_get_sequence_length(self->tree_sequence);
-    const double t_l = self->interval.left;
-    const double t_r = self->interval.right;
-    double distance_left, distance_right;
 
     if (x < 0 || x >= L) {
         ret = TSK_ERR_SEEK_OUT_OF_BOUNDS;
         goto out;
     }
```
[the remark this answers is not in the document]
molpopgen: > 

Seeking from either end gives the last plot that I posted and the best performance.  If you always seek from the left, you get the previous phenomenon of the time to initialize a tree being much higher for trees near the end (shown two plots ago).

c/tskit/trees.c
molpopgen: Sorry--not sure what is being discussed here?  The code that is highlighted appears unrelated to the comments?  If you are referring to the two seek directions, performance is the goal -- see my previous figures.  We don't want that positive slope in the index v. time plot, which results from always iteratiing from one direction.

c/tskit/trees.c
molpopgen: Yes, this is where things are going wrong.  I am unlikely to get back to this until next week.

c/tskit/trees.c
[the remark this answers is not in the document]
jeromekelleher: This bit is about detecting the null tree and only using this specialised seek in that case - we don't want to *always* null out the tree and do this seek. The idea is that by using this seek from new (null index) trees we catch the majority of the use cases.

c/tskit/trees.c
jeromekelleher: Ah, right, forgot about the ``break`` above. I get it now, ignore!

molpopgen:
Note to post-travel self: the tree direction is not being set, which is definitely causing additional problems.

molpopgen:
@jeromekelleher -- I'm quite puzzled now.  If you look at the last commits, I am comparing what I think the tree is to what tskit generates.  For some reason, sample tracking seems to not be working when seeking to a tree using the new method.

molpopgen:
> @jeromekelleher -- I'm quite puzzled now. If you look at the last commits, I am comparing what I think the tree is to what tskit generates. For some reason, sample tracking seems to not be working when seeking to a tree using the new method.

I think that this is due to the semantics of setting up tracked samples -- the flag is not processed during tree initialization.

c/tskit/trees.c
molpopgen: Done.

c/tskit/trees.c
molpopgen: Fixed.

molpopgen:
I think this one is close to ready.

molpopgen:
@jeromekelleher - This is ready for another look.

c/tskit/trees.c
```diff
@@ -4565,6 +4677,13 @@ tsk_tree_seek(tsk_tree_t *self, double x, tsk_flags_t TSK_UNUSED(options))
...
 
+    if (self->index == -1) {
+        ret = tsk_tree_seek_from_null(self, x, options);
+        if (ret < 0) {
+            goto out;
+        }
+    }
+
```
jeromekelleher, line 4686: The rest of the function should all be in an "else" (or maybe a another private utility function, like ``tsk_tree_seek_linear_search``), right? It's not clear that we're not doing the same thing twice (although probably with no effect).

molpopgen:
> 

It should be the same, but I'll take a look in the tskit-rust hack that I used to prototype.

c/tskit/trees.c
```diff
@@ -4551,8 +4551,120 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
+    tree_index = (tsk_id_t) tsk_search_sorted(breakpoints, num_trees + 1, x);
+    if (breakpoints[tree_index] > x) {
+        tree_index -= 1;
+    }
+    self->index = tree_index;
+    self->interval.left = breakpoints[tree_index];
+    self->interval.right = breakpoints[tree_index + 1];
+    self->left_index = j;
+    self->right_index = k;
+    self->direction = TSK_DIR_FORWARD;
```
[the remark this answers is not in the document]
molpopgen: j can be set directly when going forward.  But there are checks above for that, in which case j is seet "manually".  k is set manually no matter what.  I'd like to revisit this later.  I had to do these manual sets for the existing tests to pass.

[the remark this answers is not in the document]
jeromekelleher: I meant the ``direction`` - intuitively you'd think it would be FORWARD if we've gone though the edges forward and REVERSE otherwise, but maybe it doesn't matter?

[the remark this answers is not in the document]
molpopgen: oh, right -- sorry, github was highlighting a block of 4 lines. (I'm afraid I'll never be able to understand code review due to this behavior.)

I believe that it does not matter -- the direction stuff is updated later in the "advance to next tree logic".  I was also operating under the assumption that most people will want to initialize a tree and then move right along the tree sequence.

molpopgen:
@jeromekelleher  -- I think I addressed your comments.  The Python tests are failing due to codecov not connecting.

jeromekelleher:
Superb!

I'd like to do a bit more testing, but won't really know how to do it until I do some coding. Would you mind if I tack on a commit here to add these tests (and make a few style tweaks for consistency)?

molpopgen:
> 

No problem -- I'll squash and force push ASAP.

jeromekelleher:
Great, thanks @molpopgen. I think it would be good to live with this for a little bit in main before releasing, so will pick it up once the (imminent) 0.5.4 release drops.

jeromekelleher:
I've added a few more tests @molpopgen and I haven't been able to break it. I'm still a bit worried about the ``direction`` thing, because there's some tricky logic involved when we do a direction change (like tree.next(), tree.next() then tree.prev()), and I don't understand how this interacts with it. It feels like we should be doing the equivalent of seeking backwards for the right-hand side of the sequence here, but I can't figure out a way to verify this from Python. I think I'll have to write a C test to try and provoke/reassure myself it's OK.

jeromekelleher:
Yes, there's definitely tricky problems here unfortunately :disappointed: 

The tests for the LD calculator (which does extensive seeking around) are raising some errors and even segfaulting. I just tried always forcing the forward direction to see if setting ``direction`` was the problem, but that's not it.

Valgrind is showing a problem here which is probably the source of the segfault:
```
jk@tempest$ valgrind ./build/test_stats test_paper_ex_ld                
==8420== Memcheck, a memory error detector                                                    
==8420== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.              
==8420== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info                   
==8420== Command: ./build/test_stats test_paper_ex_ld                                         
==8420==               
                                                                                              
                                                                                              
     CUnit - A unit testing framework for C - Version 2.1-3                                   
     http://cunit.sourceforge.net/                                                            
                                                                                              
                                                                                              
Suite: tskit                                                                                  
  Test: test_paper_ex_ld ...==8420== Invalid read of size 4                   
==8420==    at 0x14BF39: tsk_tree_seek_from_null (trees.c:4637)                       
==8420==    by 0x14C2E2: tsk_tree_seek (trees.c:4732)          
==8420==    by 0x150A89: tsk_ld_calc_initialise (stats.c:120)
==8420==    by 0x151130: tsk_ld_calc_get_r2_array (stats.c:281)         
==8420==    by 0x10BEA1: verify_ld (test_stats.c:98)                                          
==8420==    by 0x1101AD: test_paper_ex_ld (test_stats.c:987)                          
==8420==    by 0x49C0CC6: ??? (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==8420==    by 0x49C1697: CU_run_test (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==8420==    by 0x15775A: test_main (testlib.c:1025)                     
==8420==    by 0x11391E: main (test_stats.c:1803)                                             
==8420==  Address 0x4fd012c is 4 bytes before a block of size 44 alloc'd                      
==8420==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==8420==    by 0x114D94: tsk_malloc (core.c:866)                                      
==8420==    by 0x1341CB: tsk_table_collection_build_index (tables.c:10890)
==8420==    by 0x13CD4E: tsk_treeseq_init (trees.c:456)       
==8420==    by 0x1563BB: tsk_treeseq_from_text (testlib.c:785)
==8420==    by 0x11019A: test_paper_ex_ld (test_stats.c:985)                  
==8420==    by 0x49C0CC6: ??? (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)        
==8420==    by 0x49C1697: CU_run_test (in /usr/lib/x86_64-linux-gnu/libcunit.so.1.0.1)
==8420==    by 0x15775A: test_main (testlib.c:1025)      
==8420==    by 0x11391E: main (test_stats.c:1803)                                             
==8420==                                                                           
```

This stuff is very subtle and it'll need quite a bit more testing I think. It would probably be best to implement it in Python first and get it fully tested and verified before jumping in to C.

molpopgen:
Bummer -- I'll take a look when I get some time.

jeromekelleher:
Fixing that segfault may be fairly straightforward, it just wasn't obvious to me when I looked at it last night how it could happen - the code in question looked watertight to me.

molpopgen:
@jeromekelleher -- the valgrind issue and segfaults are due to undefined behavior introduced [here](https://github.com/tskit-dev/tskit/pull/2661/commits/c2a4b478373a52411251c799cbac757ac2d44583#diff-1604fd5688ee5df1192527915ccc2769b428aa1bf4090f4cae5b24d05d3dea37R4636) (line 4636 in case it isn't clear from the link).

jeromekelleher:
Doh, was that my fault? Sorry!

molpopgen:
So we're back to a situation where all the tests are passing, etc..  Shall I squash tihs one?

jeromekelleher:
Yes please!

molpopgen:
AFAICT, this looks good to go?

python/tests/test_highlevel.py
```diff
@@ -4669,6 +4680,37 @@ def test_seek_0_from_3(self):
...
+            # reusing the seek from null under the hood.
+            t2 = tskit.Tree(ts)
+            if index <= ts.num_trees / 2:
+                while t2.index != index:
+                    t2.next()
+            else:
+                while t2.index != index:
+                    t2.prev()
+            assert t1.index == t2.index
+            assert np.all(t1.parent_array == t2.parent_array)
```
petrelharp, line 4700: IIRC there's more to a tree than the parent array; shouldn't this be testing a bit more? (do we have an equality operation for trees?)

python/tests/test_highlevel.py
```diff
@@ -4669,6 +4680,37 @@ def test_seek_0_from_3(self):
...
+    def test_seek_last_then_prev(self, ts):
+        t1 = tskit.Tree(ts)
+        t1.seek(ts.sequence_length - 0.00001)
+        assert t1.index == ts.num_trees - 1
+        t2 = tskit.Tree(ts)
+        t2.prev()
+        assert_trees_identical(t1, t2)
+        t1.prev()
+        t2.prev()
+        assert_trees_identical(t1, t2)
```
petrelharp, line 4712: oh yeah here

molpopgen:
"jupyter-book not found" :laughing:

jeromekelleher:
Opened #2740 to track updating the docs. This is ready to go now once we get CI working again.

jeromekelleher:
@mergifyio rebase

PR #2663, opened 2022-12-14
Don't update sample flags in simplify

jeromekelleher:
Closes #2662

~Stacked on #2619~

Issue #1034, opened 2020-11-25
Function to delete all mutations at a site but leave the site in the table


hyanwong:
With @awohns , we find that we need a function like `ts.delete sites` but which only deletes the mutations at the site, and leaves the site table untouched. This is basically just a slimmed down version of delete_sites, that looks something like this:

For the moment we will include the function below in tsdate, but it would be nice to get this functionality into tskit, I think. Perhaps it could be done with an extra parameter to the existing `delete_sites` function, or perhaps for didactic reasons we should deprecate `delete_sites` and introduce `delete_mutations(site_ids=XXX, delete_site=True)` which does the same as the current delete_sites function? If the latter, I guess in the longer term we could add functionality to delete a specific list of mutations, which need not be all of them at a site, but that's a bit complex and can be kicked down the road.

```
    def delete_site_mutations(self, site_ids, record_provenance=True):
        """
        Remove the mutations at the specified sites entirely from the mutations table in
        this collection.

        :param list[int] site_ids: A list of site IDs specifying the sites whose
            mutations will be removed.
        :param bool record_provenance: If ``True``, add details of this operation
            to the provenance table in this TableCollection. (Default: ``True``).
        """
        keep_sites = np.ones(len(self.sites), dtype=bool)
        site_ids = util.safe_np_int_cast(site_ids, np.int32)
        if np.any(site_ids < 0) or np.any(site_ids >= len(self.sites)):
            raise ValueError("Site ID out of bounds")
        keep_sites[site_ids] = 0
        keep_mutations = keep_sites[self.mutations.site]
        new_ds, new_ds_offset = keep_with_offset(
            keep_mutations,
            self.mutations.derived_state,
            self.mutations.derived_state_offset,
        )
        new_md, new_md_offset = keep_with_offset(
            keep_mutations, self.mutations.metadata, self.mutations.metadata_offset
        )
        # Mutation numbers will change, so the parent references need altering
        mutation_map = np.cumsum(keep_mutations, dtype=self.mutations.parent.dtype) - 1
        # Map parent == -1 to -1, and check this has worked (assumes tskit.NULL == -1)
        mutation_map = np.append(mutation_map, -1).astype(self.mutations.parent.dtype)
        assert mutation_map[tskit.NULL] == tskit.NULL
        self.mutations.set_columns(
            site=self.mutations.site[keep_mutations],
            node=self.mutations.node[keep_mutations],
            time=self.mutations.time[keep_mutations],
[...]

Issue #176, opened 2019-04-16
Multiple chromosomes


jeromekelleher:
@DomNelson and I were just discussing the possibility of adding support for multiple chromosomes to tskit. One possibility which seems like it might be a reasonably smooth path forward is the following:

1. Add a ``chromosome`` table with an ID, ``name``, ``length`` and ``metadata``.
2. Add a ``chromosome`` column to the edge table. Within a chromosome, coordinates must be from 0 to the chromosome's length.
3. Either deprecate the ``sequence_length`` property (probably best), or make it equal to the sum of the length of all chromosomes.

For things like ``trees()`` we could add an optional ``chromosome`` argument. If the tree sequence contains multiple chromosomes which would raise an error if it's not specified. For tree sequences with a single chromosome, things would continue to work as now.

Any thoughts @petrelharp, @hyanwong, @bhaller, @molpopgen?

molpopgen:
I have implemented discrete chromosomes in fwdpy11.  Length refers to the
total genome length, and the recombination functions work to ensure correct
recording of transmitted edges.  In practice, keeping the individual
chromosome lengths as a global variable in a python script has been
sufficient for any downstream analysis.    I'm in favor of not adding more
tables, although I do see the utility for considering "real" data.

molpopgen:
One question: if you add chromosome labels, and edges on chromosome i are
in [0,i-1), then do you plan to simplify separately by chromosome or change
how edges are sorted to include the chromosome label?

I've let edges cross "chromosome" boundaries, which occurs at 1 minus the
crossover rates between chromosomes.  The nice feature of this approach is
that none of the sorting and simplifying code needed touching.

bhaller:
SLiM allows simulation of multiple chromosomes, in practice, by letting the user define a recombination map with a recombination rate of 0.5 between specific pairs of bases that constitute the end of one chromosome and the start of the next.  With this method you can simulate as many chromosomes as you want and it works quite nicely, but it's not terribly aesthetically pleasing.  We may put a layer on top of this machinery that hides this implementation detail from the user and instead lets them formally define multiple chromosomes; but I imagine it will keep working the same way under the hood, more or less.  So the upshot is: like fwdpy, we're already doing this and haven't seen any particular need for a new chromosome table, etc.  But if those features were to appear in tskit, we could certainly adapt to using them.

[...]

PR #2665, opened 2022-12-15
Multichrom example

jeromekelleher:
~Stacked on #2663 (and #2619)~

@petrelharp @bhaller @molpopgen  You might be interested in this: the new file `c/examples/multichrom_wright_fisher.c` gives a simple example of multi chromosome simulations with parallel simplify (most of the other stuff is from other in-flight PRs)

This adds a simple example of running a multi chromosome forwards simulation and running the sort/simplify steps in parallel using pthreads. This was really just to make sure that the basic idea of share the node table across the different table collections would work, and that we were getting the details right with the new TSK_SIMPLIFY_NO_FILTER_NODES and TSK_SIMPLIFY_NO_UPDATE_SAMPLE_FLAGS.

Running this through helgrind gives no errors (I think it's pretty good at picking up data races):
```
$ valgrind --tool=helgrind ./build/multichrom_wright_fisher 1000 10 1 tmp/sdf 1234
```

So, hooray, @molpopgen's ingenious idea works and with the infrastructure in #2663 and #2619 we can do it pretty easily!

Another thing I ended up doing here in 31c47e402fd02a19e64bb90454be48cfb42b2cb8 is to add a basic implementation of "delete rows", which is an operation we badly need in the API I think (mentioned here: https://github.com/tskit-dev/tskit/issues/1034#issuecomment-1304885063). I think it's worth talking about that separately, so I'll open an issue for it.

We can probably let this sit for a while now anyway, and should try to clear up some other stuff first.

### DIFF
--- a/c/examples/multichrom_wright_fisher.c
+++ b/c/examples/multichrom_wright_fisher.c
@@ -1,0 +1,265 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <err.h>
+#include <string.h>
+
+#include <pthread.h>
+#include <tskit/tables.h>
+
+#define check_tsk_error(val)                                                            \
+    if (val < 0) {                                                                      \
+        errx(EXIT_FAILURE, "line %d: %s\n", __LINE__, tsk_strerror(val));               \
+    }
+
+static void
+init_tables(tsk_table_collection_t *tcs, int num_chroms)
+{
+    int j, ret;
+
+    for (j = 0; j < num_chroms; j++) {
+        ret = tsk_table_collection_init(&tcs[j], 0);
+        check_tsk_error(ret);
+        if (j > 0) {
+            tsk_node_table_free(&tcs[j].nodes);
+        }
+    }
+}
+
+static void
+free_tables(tsk_table_collection_t *tcs, int num_chroms)
+{
+    int j;
+
+    for (j = 0; j < num_chroms; j++) {
+        if (j > 0) {
+            /* Must not double free node table columns. */
+            memset(&tcs[j].nodes, 0, sizeof(tcs[j].nodes));
+        }
+        tsk_table_collection_free(&tcs[j]);
+    }
+}
+
+static void
+join_tables(tsk_table_collection_t *tcs, int num_chroms)
+{
+    int j, ret;
+
+    for (j = 1; j < num_chroms; j++) {
+        ret = tsk_edge_table_extend(
+            &tcs[0].edges, &tcs[j].edges, tcs[j].edges.num_rows, NULL, 0);
+        check_tsk_error(ret);
+    }
+    /* Get all the squashable edges next to each other */
+    ret = tsk_table_collection_sort(&tcs[0], NULL, 0);
+    check_tsk_error(ret);
+    ret = tsk_edge_table_squash(&tcs[0].edges);
+    check_tsk_error(ret);
+    /* We need to sort again after squash */
+    ret = tsk_table_collection_sort(&tcs[0], NULL, 0);
+    check_tsk_error(ret);
+    ret = tsk_table_collection_build_index(&tcs[0], 0);
+    check_tsk_error(ret);
+}
+
+struct chunk_work {
+    int chunk;
+    tsk_table_collection_t *tc;
+    int *samples;
+    int N;
+};
+
+void *
+simplify_chunk(void *arg)
+{
+    int ret;
+    struct chunk_work *work = (struct chunk_work *) arg;
+    tsk_size_t edges_before = work->tc->edges.num_rows;
+
+    ret = tsk_table_collection_sort(work->tc, NULL, 0);
+    check_tsk_error(ret);
+    ret = tsk_table_collection_simplify(work->tc, work->samples, work->N,
+        TSK_SIMPLIFY_NO_FILTER_NODES | TSK_SIMPLIFY_NO_UPDATE_SAMPLE_FLAGS, NULL);
+    check_tsk_error(ret);
+    /* NOTE: this printf makes helgrind complain */
+    printf("\tchunk %d: %lld -> %lld\n", work->chunk, (long long) edges_before,
+        (long long) work->tc->edges.num_rows);
+
+    return NULL;
+}
+
+void
+sort_and_simplify_all(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N)
+{
+    int j, ret;
+    struct chunk_work work[num_chroms];
+    pthread_t threads[num_chroms];
+
+    for (j = 1; j < num_chroms; j++) {
+        tcs[j].nodes = tcs[0].nodes;
+    }
+
+    for (j = 0; j < num_chroms; j++) {
+        work[j].chunk = j;
+        work[j].tc = &tcs[j];
+        work[j].samples = samples;
+        work[j].N = N;
+
+        ret = pthread_create(&threads[j], NULL, simplify_chunk, (void *) &work[j]);
+        if (ret != 0) {
+            errx(EXIT_FAILURE, "Pthread create failed");
+        }
+        /* simplify_chunk((void *) &work[j]); */
+    }
+    for (j = 0; j < num_chroms; j++) {
+        ret = pthread_join(threads[j], NULL);
+        if (ret != 0) {
+            errx(EXIT_FAILURE, "Pthread join failed");
+        }
+    }
+}
+
+void
+simplify_tables(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N)
+{
+    int j, k, ret;
+    const tsk_size_t num_nodes = tcs[0].nodes.num_rows;
+    tsk_bool_t *keep_nodes = malloc(num_nodes * sizeof(*keep_nodes));
+    tsk_id_t *node_id_map = malloc(num_nodes * sizeof(*node_id_map));
+    tsk_id_t *edge_child, *edge_parent;
+
+    if (keep_nodes == NULL || node_id_map == NULL) {
+        errx(EXIT_FAILURE, "Out of memory");
+    }
+
+    printf("Simplify %lld nodes\n", (long long) tcs[0].nodes.num_rows);
+    sort_and_simplify_all(tcs, num_chroms, samples, N);
+
+    for (j = 0; j < num_nodes; j++) {
+        keep_nodes[j] = false;
+    }
+    for (j = 0; j < N; j++) {
+        keep_nodes[samples[j]] = true;
+    }
+
+    for (j = 0; j < num_chroms; j++) {
+        edge_child = tcs[j].edges.child;
+        edge_parent = tcs[j].edges.parent;
+        for (k = 0; k < tcs[j].edges.num_rows; k++) {
+            keep_nodes[edge_child[k]] = true;
+            keep_nodes[edge_parent[k]] = true;
+        }
+    }
+    tsk_node_table_keep_rows(&tcs[0].nodes, keep_nodes, 0, node_id_map);
+    printf("\tdone: %lld nodes\n", (long long) tcs[0].nodes.num_rows);
+
+    /* Remap node references */
+    for (j = 0; j < num_chroms; j++) {
+        edge_child = tcs[j].edges.child;
+        edge_parent = tcs[j].edges.parent;
+        for (k = 0; k < tcs[j].edges.num_rows; k++) {
+            edge_child[k] = node_id_map[edge_child[k]];
+            edge_parent[k] = node_id_map[edge_parent[k]];
+        }
+        ret = tsk_table_collection_check_integrity(&tcs[j], 0);
+        check_tsk_error(ret);
+    }
+    for (j = 0; j < N; j++) {
+        samples[j] = node_id_map[samples[j]];
+    }
+    free(keep_nodes);
+    free(node_id_map);
+}
+
+void
+simulate(
+    tsk_table_collection_t *tcs, int num_chroms, int N, int T, int simplify_interval)
+{
+    tsk_id_t *buffer, *parents, *children, child, left_parent, right_parent;
+    bool left_is_first;
+    double chunk_left, chunk_right;
+    int ret, j, t, b, k;
+
+    assert(simplify_interval != 0); // leads to division by zero
+    buffer = malloc(2 * N * sizeof(tsk_id_t));
+    if (buffer == NULL) {
+        errx(EXIT_FAILURE, "Out of memory");
+    }
+    for (k = 0; k < num_chroms; k++) {
+        tcs[k].sequence_length = num_chroms;
+    }
+    parents = buffer;
+    for (j = 0; j < N; j++) {
+        parents[j]
+            = tsk_node_table_add_row(&tcs[0].nodes, 0, T, TSK_NULL, TSK_NULL, NULL, 0);
+        check_tsk_error(parents[j]);
+    }
+    b = 0;
+    for (t = T - 1; t >= 0; t--) {
+        /* Alternate between using the first and last N values in the buffer */
+        parents = buffer + (b * N);
+        b = (b + 1) % 2;
+        children = buffer + (b * N);
+        for (j = 0; j < N; j++) {
+            child = tsk_node_table_add_row(
+                &tcs[0].nodes, 0, t, TSK_NULL, TSK_NULL, NULL, 0);
+            check_tsk_error(child);
+            /* NOTE: the use of rand() is discouraged for
+             * research code and proper random number generator
+             * libraries should be preferred.
+             */
+            left_parent = parents[(size_t)((rand() / (1. + RAND_MAX)) * N)];
+            right_parent = parents[(size_t)((rand() / (1. + RAND_MAX)) * N)];
+            left_is_first = rand() < 0.5;
+            chunk_left = 0.0;
+            for (k = 0; k < num_chroms; k++) {
+                chunk_right = chunk_left + rand() / (1. + RAND_MAX);
+                /* a very tiny chance that right and left are equal */
+                if (chunk_right > chunk_left) {
+                    ret = tsk_edge_table_add_row(&tcs[k].edges, chunk_left, chunk_right,
+                        left_is_first ? left_parent : right_parent, child, NULL, 0);
+                    check_tsk_error(ret);
+                }
+                chunk_left += 1.0;
+                if (chunk_right < chunk_left) {
+                    ret = tsk_edge_table_add_row(&tcs[k].edges, chunk_right, chunk_left,
+                        left_is_first ? right_parent : left_parent, child, NULL, 0);
+                    check_tsk_error(ret);
+                }
+            }
+            children[j] = child;
+        }
+        if (t % simplify_interval == 0) {
+            simplify_tables(tcs, num_chroms, children, N);
+        }
+    }
+    /* Set the sample flags for final generation */
+    for (j = 0; j < N; j++) {
+        tcs[0].nodes.flags[children[j]] = TSK_NODE_IS_SAMPLE;
+    }
+    free(buffer);
+}
+
+int
+main(int argc, char **argv)
+{
+//    int ret;
+    int num_chroms;
+
+    if (argc != 7) {
+        errx(EXIT_FAILURE, "usage: N T simplify-interval output seed num-chroms");
+    }
+
+    num_chroms = atoi(argv[6]);
+    tsk_table_collection_t tcs[num_chroms];
+
+    srand((unsigned) atoi(argv[5]));
+    init_tables(tcs, num_chroms);
+    simulate(tcs, num_chroms, atoi(argv[1]), atoi(argv[2]), atoi(argv[3]));
+    join_tables(tcs, num_chroms);
+//    ret = tsk_table_collection_dump(&tcs[0], argv[4], 0);
+//    check_tsk_error(ret);
+    free_tables(tcs, num_chroms);
+
+    return 0;
+}
--- a/c/examples/multichrom_wright_fisher.c
+++ b/c/examples/multichrom_wright_fisher.c
@@ -122,7 +122,7 @@ sort_and_simplify_all(tsk_table_collection_t *tcs, int num_chroms, int *samples,
 void
 simplify_tables(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N)
 {
-    int j, k, ret;
+    int j, k, num_edges, ret;
     const tsk_size_t num_nodes = tcs[0].nodes.num_rows;
     tsk_bool_t *keep_nodes = malloc(num_nodes * sizeof(*keep_nodes));
     tsk_id_t *node_id_map = malloc(num_nodes * sizeof(*node_id_map));
@@ -137,15 +137,18 @@ simplify_tables(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N
 
     for (j = 0; j < num_nodes; j++) {
         keep_nodes[j] = false;
+        tcs[0].nodes.flags[j] &= (~TSK_NODE_IS_SAMPLE);
     }
     for (j = 0; j < N; j++) {
         keep_nodes[samples[j]] = true;
+        tcs[0].nodes.flags[samples[j]] |= TSK_NODE_IS_SAMPLE;
     }
 
     for (j = 0; j < num_chroms; j++) {
         edge_child = tcs[j].edges.child;
         edge_parent = tcs[j].edges.parent;
-        for (k = 0; k < tcs[j].edges.num_rows; k++) {
+        num_edges = tcs[j].edges.num_rows;
+        for (k = 0; k < num_edges; k++) {
             keep_nodes[edge_child[k]] = true;
             keep_nodes[edge_parent[k]] = true;
         }
@@ -157,7 +160,8 @@ simplify_tables(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N
     for (j = 0; j < num_chroms; j++) {
         edge_child = tcs[j].edges.child;
         edge_parent = tcs[j].edges.parent;
-        for (k = 0; k < tcs[j].edges.num_rows; k++) {
+        num_edges = tcs[j].edges.num_rows;
+        for (k = 0; k < num_edges; k++) {
             edge_child[k] = node_id_map[edge_child[k]];
             edge_parent[k] = node_id_map[edge_parent[k]];
         }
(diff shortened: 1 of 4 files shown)


molpopgen:
Great!  I'll take a closer look in a bit.

bhaller:
Super cool.  Good to know about helgrind, too; I will look into that too!  :->

hyanwong:
I guess this is relevant to https://github.com/tskit-dev/tskit/issues/176, and might inform how we deal with multiple chromosomes.

hyanwong:
>  I would tentatively lean towards not connecting these two things.

I think it's not exclusive. If you *want* to represent multiple chromosomes, then this gives a nice way to do it. But there's not reason why you couldn't also do it on a single chromosome, or indeed on part of a multi-chromosome set.

But I do take you point that there's an elegancy about keeping the things completely independent.

jeromekelleher:
This is really about low-level C API stuff here for now - how we might (or not) surface this to higher level things like representing multiple chromosomes is something we should wait and see (which is I think what we're all saying!).

jeromekelleher:
@petrelharp @bhaller - I realised there was an easy way to join the tables by appending all the edges to the first table, sorting and then squashing adjacent edges. This *should* be equivalent to the haploid_wright_fisher example now (but that'll need to be verified).

jeromekelleher:
@petrelharp @bhaller - I realised there was an easy way to join the tables by appending all the edges to the first table, sorting and then squashing adjacent edges. This *should* be equivalent to the haploid_wright_fisher example now (but that'll need to be verified).

jeromekelleher:
Ah, probably need to set the sample flags for the final generation.

jeromekelleher:
One more update - I've marked the sample flags and updated the original WF example so that it also indexes the output, and I think the outputs are now identical up to node reordering:
```
tree 0:
  num_sites: 0
  interval:  (0.000000, 0.006008)
                                                     322                                                         321                                                                         >
                                    ┏━━━━━┳━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━┓                                     ┏━━━┻━━━━┓                                                                     >
                  312               ┃     ┃                             ┃                                     ┃        ┃                                                                     >
          ┏━━━━━━━━┻━━━━━━━━┓       ┃     ┃                             ┃                                     ┃        ┃                                                                     >
         299                ┃       ┃     ┃                             ┃                                     ┃        ┃                                                                     >
    ┏━━━━━┻━━━━━┓           ┃       ┃     ┃                             ┃                                     ┃        ┃                                                                     >
    ┃           ┃           ┃       ┃     ┃                             ┃                                    278       ┃                                                                     >
    ┃           ┃           ┃       ┃     ┃                             ┃                                   ┏━┻┓       ┃                                                                     >
    ┃           ┃           ┃       ┃     ┃                            258                   264            ┃  ┃       ┃                                                                     >
    ┃           ┃           ┃       ┃     ┃                     ┏━━━━━━━┻━━━━━━━┓     ┏━━━━━━━┻━━━━━━━┓     ┃  ┃       ┃                                                                     >
   238          ┃           ┃       ┃     ┃                     ┃               ┃     ┃               ┃     ┃  ┃       ┃                                                                     >
  ┏━┻━┓         ┃           ┃       ┃     ┃                     ┃               ┃     ┃               ┃     ┃  ┃       ┃                                                                     >
  ┃   ┃        228          ┃       ┃     ┃                    229              ┃     ┃               ┃     ┃  ┃       ┃                                               219                   >
  ┃   ┃     ┏━━━┻━━━┓       ┃       ┃     ┃           ┏━━━━━━━━━┻━┳━━━━━━━┓     ┃     ┃               ┃     ┃  ┃       ┃                                ┏━━━━━━━━━━━━━━━┻━┳━━━━━━━┳━━━━━┓    >
  ┃   ┃     ┃       ┃      195     199    ┃           ┃           ┃       ┃     ┃     ┃              184    ┃  ┃      192            188                ┃                 ┃       ┃     ┃    >
  ┃   ┃     ┃       ┃   ┏━━━╋━━━━┓ ┏┻━┓   ┃           ┃           ┃       ┃     ┃     ┃           ┏━━━┻━━━┓ ┃  ┃    ┏━━┻━━┓     ┏━━━━━┻━┳━━━┓           ┃                 ┃       ┃     ┃    >
 176  ┃     ┃       ┃   ┃  167   ┃ ┃  ┃  170          ┃          172     162    ┃     ┃          165      ┃ ┃  ┃    ┃    181    ┃       ┃   ┃          169                ┃       ┃    180   >
┏━┻┓  ┃     ┃       ┃   ┃  ┏┻━┓  ┃ ┃  ┃ ┏━┻━┓         ┃      ┏━━━━╋━━━┓  ┏┻━┓   ┃     ┃     ┏━━━━━╋━━━━┓  ┃ ┃  ┃    ┃   ┏━┻┓    ┃       ┃   ┃   ┏━━━━━┳━┻━━━━━━┓          ┃       ┃    ┏┻━┓  >
┃  ┃  ┃    102     138  ┃  ┃  ┃  ┃ ┃  ┃ ┃  121       146     ┃   101  ┃  ┃  ┃  124   129   105   116   ┃  ┃ ┃  ┃   107  ┃  ┃   134     144  ┃   ┃    120      126        119     123   ┃  ┃  >
┃  ┃  ┃  ┏━━╋━━┓  ┏━┻┓  ┃  ┃  ┃  ┃ ┃  ┃ ┃  ┏┻━┓  ┏━━┳━┻┳━━┓  ┃  ┏━┻┓  ┃  ┃  ┃  ┏┻━┓ ┏━╋━━┓ ┏┻━┓  ┏┻━┓  ┃  ┃ ┃  ┃  ┏━┻┓  ┃  ┃ ┏━━╋━━┓  ┏━┻┓  ┃   ┃  ┏━━╋━━┓  ┏━━╋━━┓  ┏━━┳━┻┳━━┓  ┏┻━┓  ┃  ┃  >
0 96 57 14 69 90 58 87 11 20 93 47 1 34 8 66 85 17 26 89 91 21 40 54 95 75 81 18 31 2 9 12 3 36 16 49 22 72 4 32 45 55 67 88 5 42 92 62 98 73 6 7 19 60 84 38 70 76 24 28 43 74 63 77 68 79 1>

tree 1:
  num_sites: 0
  interval:  (0.006008, 0.008218)
:
```
and the second
```
tree 0:
  num_sites: 0
  interval:  (0.000000, 0.006008)
                                                                            1                                                                                    0                           >
                                                   ┏━━━━━━━━┳━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━┓                                                      ┏━━━━━┻━━━━━┓                     >
                         18                        ┃        ┃                                       ┃                                                      ┃           ┃                     >
              ┏━━━━━━━━━━━┻━━━━━━━━━━━━┓           ┃        ┃                                       ┃                                                      ┃           ┃                     >
             41                        ┃           ┃        ┃                                       ┃                                                      ┃           ┃                     >
      ┏━━━━━━━┻━━━━━━━┓                ┃           ┃        ┃                                       ┃                                                      ┃           ┃                     >
      ┃               ┃                ┃           ┃        ┃                                       ┃                                                     53           ┃                     >
      ┃               ┃                ┃           ┃        ┃                                       ┃                                                    ┏━┻━┓         ┃                     >
      ┃               ┃                ┃           ┃        ┃                                      60                              66                    ┃   ┃         ┃                     >
      ┃               ┃                ┃           ┃        ┃                            ┏━━━━━━━━━━┻━━━━━━━━━━┓         ┏━━━━━━━━━━┻━━━━━━━━━━━┓        ┃   ┃         ┃                     >
     73               ┃                ┃           ┃        ┃                            ┃                     ┃         ┃                      ┃        ┃   ┃         ┃                     >
   ┏━━┻━━┓            ┃                ┃           ┃        ┃                            ┃                     ┃         ┃                      ┃        ┃   ┃         ┃                     >
   ┃     ┃           108               ┃           ┃        ┃                           109                    ┃         ┃                      ┃        ┃   ┃         ┃                     >
   ┃     ┃       ┏━━━━┻━━━━┓           ┃           ┃        ┃              ┏━━━━━━━━━━━━━┻━┳━━━━━━━━━━━┓       ┃         ┃                      ┃        ┃   ┃         ┃                     >
   ┃     ┃       ┃         ┃          129         133       ┃              ┃               ┃           ┃       ┃         ┃                     118       ┃   ┃        126                   1>
   ┃     ┃       ┃         ┃     ┏━━━━━╋━━━━━┓   ┏━┻━┓      ┃              ┃               ┃           ┃       ┃         ┃                ┏━━━━━┻━━━━┓   ┃   ┃     ┏━━━┻━━━┓         ┏━━━━━━━>
  170    ┃       ┃         ┃     ┃    161    ┃   ┃   ┃     164             ┃              166         156      ┃         ┃               159         ┃   ┃   ┃     ┃      175        ┃       >
 ┏━┻━┓   ┃       ┃         ┃     ┃   ┏━┻━┓   ┃   ┃   ┃   ┏━━┻━━┓           ┃         ┏━━━━━╋━━━━━┓   ┏━┻━┓     ┃         ┃         ┏━━━━━━┻┳━━━━━┓   ┃   ┃   ┃     ┃     ┏━┻━┓       ┃       >
 ┃   ┃   ┃      180       216    ┃   ┃   ┃   ┃   ┃   ┃   ┃    199         224        ┃    179    ┃   ┃   ┃    202       207       183     194    ┃   ┃   ┃   ┃    185    ┃   ┃      212      >
 ┃   ┃   ┃   ┏━━━╋━━━┓   ┏━┻━┓   ┃   ┃   ┃   ┃   ┃   ┃   ┃   ┏━┻━┓   ┏━━━┳━┻━┳━━━┓   ┃   ┏━┻━┓   ┃   ┃   ┃   ┏━┻━┓   ┏━━━╋━━━┓   ┏━┻━┓   ┏━┻━┓   ┃   ┃   ┃   ┃   ┏━┻━┓   ┃   ┃   ┏━━━╋━━━┓   >
231 327 288 245 300 321 289 318 242 251 324 278 232 265 239 297 316 248 257 320 322 252 271 285 326 306 312 249 262 233 240 243 234 267 247 280 253 303 235 263 276 286 298 319 236 273 323 2
```

We don't reorder the samples from 0 - n in the new version. If you do want this you can run simplify on it again and it'll renumber the output. (You could do this more cheaply too if you wanted to for compatibility purposes.)

jeromekelleher:
I think we can meaningfully compare the timings of this with haploid_wright_fisher now @petrelharp and @bhaller

petrelharp:
I think I know what's going on with the dissappointing scaling here. Here's an example from running the code:
```
$ N=10000;  T=5000; S=1000
$ ./haploid_wright_fisher $N $T $S n10k.trees 42
Simplify at generation 4000: (10010000 nodes 20000000 edges) -> (84542 nodes 439275 edges)
Simplify at generation 3000: (10084542 nodes 20439275 edges) -> (92436 nodes 494357 edges)
Simplify at generation 2000: (10092436 nodes 20494357 edges) -> (97097 nodes 527279 edges)
Simplify at generation 1000: (10097097 nodes 20527279 edges) -> (100073 nodes 547226 edges)
Simplify at generation 0: (10100073 nodes 20547226 edges) -> (102133 nodes 560584 edges)

$ ./multichrom_wright_fisher $N $T $S n10k-multichrom.trees 42
Simplify 10010000 nodes
	chunk 1: 11250775 -> 70861
	chunk 5: 11249822 -> 68845
	chunk 0: 11250882 -> 69585
	chunk 6: 11248979 -> 71402
	chunk 3: 11250703 -> 71395
	chunk 7: 11247502 -> 71564
	chunk 4: 11251640 -> 71959
	chunk 2: 11249697 -> 71370
```
The numbers for the multichromosome example are numbers of edges before and after simplification. Notice that *each* worker has about one-half the total number of edges: there were 20M edges for the whole simulation (2 edges * 10K individuals * 1K generations); and each of the 8 workers have 10M edges to deal with.

This is because: suppose that a chromosome has been inherited in `k` pieces, i.e., there are `k` edges that partition that chromosome. If we divide the chromosome among `n` workers, then we will get `n + k - 1` edges, since we're breaking the chromosome at (a) `n` places, and also at (b) `k` places. Unless we get lucky and some of the breakpoints fall exactly on a division between adjacent chunks. (In this simulation, everyone is divided into exactly `k=2` pieces.)

Now, what does simplify have to do under the hood? It's got a data structure that records the collection of ancestral segments that it's tracking; and then it goes through edges one at a time and deals with them. This means that if the cost of dealing with each edge is O(1), then our maximal speedup would be `(n+k-1)/(nk) ~ 1/k`. (This is because we've increased the number of edges by a factor of `(n+k-1)/k` but then divided that among `n` workers.)

However, the work required by simplify's data structure does scale with the number of segments it is tracking, to some degree, since although simplify has to *look* at all those edges, it only has to actually *do* something when it encounters an edge that some of its segments move along. If this was a substantial contribution to runtime then we might expect better speedups. The things we'd need to know are (a) what portion of simplify's work is "do something to the segments" versus "look at the next edge"; and (b) once we divide up a chromosome into chunks, how much does that reduce the proportion of its edges through which a segment of ancestry passes?  I'm not very clear on the point, but I suspect the answer to (b) is "not much"; if so, then there's no hope for a speedup.

However, this approach *would* work just fine if each worker got literally one chromosome, since then edge breakpoints would usually (well, half the time) fall exactly on the divisions between workers. I may do a modification to the code here to have a look at that case, and see if we have wins there.

ping @jeromekelleher @molpopgen

petrelharp:
I did some very simple benchmarking - comparing runtimes for `k` chromosomes, done (a) in parallel, and (b) not. (To do this I added a new file, `multichrom_wright_fisher_singlethreaded.c`, which if we merge this example in probably don't want to keep.)
Here's the script:
```
#!/bin/bash

N=1000
T=5000
S=200

for k in $(seq 12)
do
    /usr/bin/time ./multichrom_wright_fisher_singlethreaded $N $T $S n10k.trees 42 $k &> single${k}.log
    /usr/bin/time ./multichrom_wright_fisher $N $T $S n10k.trees 42 $k &> multi${k}.log
done

echo "type num_chroms user system elapsed percent_cpu N T S" > results.txt
for t in single multi
do
    (for x in ${t}*.log; do echo ${x%.log} $(tail -n 2 $x | head -n 1 | cut -f 1-4 -d ' ') $N $T $S; done | sed -e 's/user//' | sed -e 's/chroms//' | sed -e 's/system//' | sed -e 's/elapsed//' | sed -e 's/.CPU//' | sort -n ) | tr ' ' '\t' >> results.txt
done
```

With that, I get this: black points are singlethreaded, red are multithreaded. Note that the simulations are still doing a fair bit of non-parallel work that is included in these timings:
Screenshot from 2023-09-11 22-27-31

This is looking very good! The max CPU utilization percent reported is 675%. Here's the rest of the results:
```
     type num_chroms  user system elapsed percent_cpu
1   multi          1  4.22   0.48    4.70          99
2   multi          2  8.50   1.02    5.36         177
3   multi          3 13.21   1.47    5.92         247
4   multi          4 18.17   2.01    6.52         309
5   multi          5 24.05   2.86    8.27         325
6   multi          6 30.71   3.70    9.02         381
7   multi          7 38.30   4.45    9.79         436
8   multi          8 46.23   5.26   10.56         487
9   multi          9 54.43   6.11   11.41         530
10  multi         10 63.40   7.06   12.16         579
11  multi         11 73.02   8.35   12.95         628
12  multi         12 83.72   9.45   13.79         675
13 single          1  3.97   0.05    4.02         100
14 single          2  7.91   0.16    8.08         100
15 single          3 12.31   0.61   12.93          99
16 single          4 16.69   0.82   17.52          99
17 single          5 21.33   1.14   22.47          99
18 single          6 25.93   1.39   27.33          99
19 single          7 30.78   1.79   32.57          99
20 single          8 35.85   2.18   38.04          99
21 single          9 40.84   1.96   42.81          99
22 single         10 45.17   2.69   47.87          99
23 single         11 50.77   3.11   53.89          99
24 single         12 55.90   3.41   59.32          99
```

jeromekelleher:
Aha! That's great @petrelharp, it really clarifies things :+1:

petrelharp:
I'm thinking it's worth fixing up the multichrom example to merge in? @bhaller is enthusiastic about getting this into SLiM at some point. I think this just will require switching over `delete_rows` to `keep_rows`, and removing the extra file I put in?

jeromekelleher:
OK, I'll fix it up when I get a chance.

jeromekelleher:
I've updated this so that it's a clean build against upstream now @petrelharp. I think it needs a bit of documentation though - I don't really follow what the algorithm is doing. Could you update with some comments to help clarify that the model is, etc?

jeromekelleher:
I think the docs build is because of a stale cache. The simplest thing is probably to create a new PR. I've squashed my original commits down to one.

molpopgen:
I'm away on vacation for a couple of weeks, but:

1. The speedup should be 75-80% at most if one doesn't do one "chromosome" per thread. That's what I showed w/my prototype when I presented this method.
2. @petrelharp -- if I understand your figure, 12 chromosomes means 12 threads and a 6x speedup?  Is that correct?  If you run `perf`, you should be able to quantify the % of time that is purely single-threaded, which will give a better sense of how efficient the threading is.

jeromekelleher:
It's fairly esoteric knowledge @bhaller which I'm not sure is worth keeping a PR open in perpetuity for. We've learned the key lessons from it, I think, and moved on?

bhaller:
But anybody who wants to actually use tskit in a multithreaded manner (which would be me, and very probably others, sooner or later) will continue to find it useful.  Which is the point of an example, isn't it?  Why purge it from the knowledge pool?

jeromekelleher:
If it's useful we should merge it I guess - I didn't think too hard about it tbh. Keeping it here in limbo isn't an answer.

Does anyone want to have a look over and make suggestions for what would be needed to make this a useful documentation example?

bhaller:
I nominate @petrelharp :->.  In all seriousness, he knows the tskit ecosystem way better than I do, knows your doc system, and has been involved in this all along.  Can I volunteer you, Peter?  :->

petrelharp:
I think the proposal is to add something to the docs, [here](https://tskit.dev/tskit/docs/latest/c-api.html#examples). I could do that, sure. I'm imagining just talking through, like "we can have the tables for each chromosome share the same node and individual tables; here's how we do that".

petrelharp:
Go for it!

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,265 @@
...
+
+static void
+free_tables(tsk_table_collection_t *tcs, int num_chroms)
+{
+    int j;
+
+    for (j = 0; j < num_chroms; j++) {
+        if (j > 0) {
+            /* Must not double free node table columns. */
+            memset(&tcs[j].nodes, 0, sizeof(tcs[j].nodes));
```
bhaller, line 37: As I mentioned on Slack, it seems safer to keep the shared tables zeroed out all the time except when they need to be there, to avoid accidental modification of one of the shared tables, use of stale data, stale pointers, etc.; there are lots of risks here.  For SLiM I wrote a method that copies the main table collection's shared tables (we're sharing node, individual, and population) into the other table collections, and another method that zeros out those copies.  Then I bracket operations like simplification and writing to disk with those methods.  All the rest of the time the copies of the shared tables are zeroed for safety.  Just a suggestion; might point people in a safer direction, even though it doesn't matter for this model as it is here.

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,265 @@
...
+    tsk_id_t *edge_child, *edge_parent;
+
+    if (keep_nodes == NULL || node_id_map == NULL) {
+        errx(EXIT_FAILURE, "Out of memory");
+    }
+
+    printf("Simplify %lld nodes\n", (long long) tcs[0].nodes.num_rows);
+    sort_and_simplify_all(tcs, num_chroms, samples, N);
+
+    for (j = 0; j < num_nodes; j++) {
```
bhaller, line 138: rather than this loop, I used calloc() for keep_nodes; the kernel can provide zeroed pages faster than any loop can, even if the compiler turns it into memset()...

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,265 @@
...
+        keep_nodes[j] = false;
+    }
+    for (j = 0; j < N; j++) {
+        keep_nodes[samples[j]] = true;
+    }
+
+    for (j = 0; j < num_chroms; j++) {
+        edge_child = tcs[j].edges.child;
+        edge_parent = tcs[j].edges.parent;
+        for (k = 0; k < tcs[j].edges.num_rows; k++) {
```
bhaller, line 148: here and at line 160, I put `tcs[j].edges.num_rows` into a local temporary; I think the compiler might have to refetch the value from memory every time through, otherwise, since it might not be smart enough to realize that there is not a `restrict` issue here (i.e., that the loop's work is guaranteed not to change the value of `num_rows` through a pointer write)...

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,265 @@
...
+            edge_child[k] = node_id_map[edge_child[k]];
+            edge_parent[k] = node_id_map[edge_parent[k]];
+        }
+        ret = tsk_table_collection_check_integrity(&tcs[j], 0);
+        check_tsk_error(ret);
+    }
+    for (j = 0; j < N; j++) {
```
bhaller, line 167: Here I actually have a question, and maybe a comment would be useful.  The question is: is the result of filtering the node table ourselves, with `tsk_node_table_keep_rows()` as done here, different from letting simplify do the filtering for us?  Is the order of the nodes that results different?  I ask because SLiM's current code assumes that the N samples are in positions 0:(N-1) in the node table after simplification.  The code here kind of lumps the samples together with all of the nodes referenced by edges, and tells tskit "keep all of these", without distinguishing the two; so I'm guessing that SLiM can no longer make that 0:(N-1) assumption, and needs to remap the N samples as you do here, using `node_id_map`.  But I'd like to understand why.  Does simplification filter the node table in a different and more sophisticated way?  Or is there some other reason why SLiM's assumption happened to be valid?  @petrelharp wrote that code for SLiM, I think, so maybe he wants to hear this.

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,265 @@
...
+    for (j = 0; j < num_chroms; j++) {
+        ret = pthread_join(threads[j], NULL);
+        if (ret != 0) {
+            errx(EXIT_FAILURE, "Pthread join failed");
+        }
+    }
+}
+
+void
+simplify_tables(tsk_table_collection_t *tcs, int num_chroms, int *samples, int N)
```
bhaller, line 123: I'd suggest a name like `samples_count` instead of `N`, but maybe that's just me :->

c/examples/multichrom_wright_fisher.c
jeromekelleher: The reason for doing the node filtering ourselves here is that the different per-chromosome simplify operations will come up with different node mappings. That's the only real reason for using tsk_node_table_keep_rows.  You can't assume that the samples to 0,...n - 1 in this case, no (and would need to do the remapping).

c/examples/multichrom_wright_fisher.c
petrelharp: good idea - I went to go do this, but noticed that this is the `N` in the WF model; so I'm voting to keep it this way (it's actually the total population size as well as the number of samples)

c/examples/multichrom_wright_fisher.c
petrelharp: thoughts, @jeromekelleher ?

petrelharp:
This code doesn't reset the `sample` flags; I've added a bit that does that, and also done one of @bhaller's suggestion. There's one more suggestion that I don't know where the balance between 'code complexity' and 'safe coding' is; let's get that sorted or not and merge this?

c/examples/multichrom_wright_fisher.c
jeromekelleher: Agreed

c/examples/multichrom_wright_fisher.c
jeromekelleher: Yes, I think it's clearer this way for the purposes of example code, and the perf optimisation will be obvious to anyone who understands it well enough to use it.

c/examples/multichrom_wright_fisher.c
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    num_chroms = atoi(argv[6]);
+    tsk_table_collection_t tcs[num_chroms];
+
+    srand((unsigned) atoi(argv[5]));
+    init_tables(tcs, num_chroms);
+    simulate(tcs, num_chroms, atoi(argv[1]), atoi(argv[2]), atoi(argv[3]));
+    join_tables(tcs, num_chroms);
+//    ret = tsk_table_collection_dump(&tcs[0], argv[4], 0);
```
jeromekelleher, line 264: Why have we commented this out?

c/examples/multichrom_wright_fisher_singlethreaded.c
```diff
@@ -0,0 +1,120 @@
...
+
+    if (argc != 7) {
+        errx(EXIT_FAILURE, "usage: N T simplify-interval output seed num-chroms");
+    }
+    ret = tsk_table_collection_init(&tables, 0);
+    check_tsk_error(ret);
+    srand((unsigned)atoi(argv[5]));
+    simulate(&tables, atoi(argv[6]), atoi(argv[1]), atoi(argv[2]), atoi(argv[3]));
+
+//    /* Sort and index so that the result can be opened as a tree sequence */
```
jeromekelleher, line 110: Again, not clear why we're not outputting the results any more?

c/examples/multichrom_wright_fisher.c
jeromekelleher: I think there's a lot of foot guns lying around here, and I'd rather keep the code as simple as possible without giving the reader the (likely false) impression that the code is perfectly safe. Unless there's a specific thing to protect against in the code, I don't think there's any point in complicating the code any further. So I'd vote to keep as-is.

jeromekelleher (commented):
I think we should basically keep it as-is. I don't think we'd make the code any safer in reality by making it more complicated - this is very much an advanced topic, which requires some understanding of how all this stuff works to use.

Before merging, we should fix-up the examples so they write output and run them in CI (which I'm happy to do)

c/examples/multichrom_wright_fisher.c
petrelharp: not sure, will uncomment

petrelharp:
Okay - fixed those comments up. I'm not sure why the builds are failing?

jeromekelleher:
I'll pick it up, thanks @petrelharp

PR #2668, opened 2022-12-23
Implement is_empty method

### LINKED ISSUE #2640: Tree.is_empty method
As discussed in https://github.com/tskit-dev/tskit/issues/2600#issuecomment-1319750880, it would be good to have an standard way of checking if a tree is "empty". However, there are 2 definitions of empty:

1. No topology, anywhere. This is what happens when we `delete_intervals` e.g. when we simulate a certain region, or when we remove the flanking regions in tsinfer.
2. No topology reachable from the roots (where a root can also be subject to the `root_threshold` parameter, such that isolated samples may not always be considered a root).

I think that an `is_empty()` method would be a good place to document these two different meanings. So something like:

```python
def is_empty(self, check_roots=None) -> bool:
    """
    Check if this tree is "empty" (i.e. has no topology). A tree is empty if it has no edges. However, it
    may also be considered empty if it contains edges which represent "dead branches" (i.e. not
    reachable from the :meth:`~Tree.roots` of the tree). To consider a tree like this as empty, specify
    ``check_roots=True``.

    :param bool check_roots: Should we also consider a tree empty if it has topology which is unconnected
        to any of the roots of the tree. Default: ``None`` treated as ``False``.
    :return: ``True`` if this tree is empty, ``False`` otherwise.
    """
    if check_roots:
        if tree.num_roots == tree.num_samples():  # A common case, when root_threshold is 1
            return True
        else:
            # More inefficient if this is an empty tree with many samples, e.g. a missing region of genome
            return set(tree.roots) <= set(tree.tree_sequence.samples())
    else:
         return self.num_edges == 0
```

hyanwong:
## Description

Fixes #2640

# PR Checklist:

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,6 +4,9 @@
 
 **Features**
 
+- A new ``Tree.is_empty`` method can be used to determine if a tree is empty
+  for 2 definitions of emptiness (:user:`hyanwong`, :pr:`2668`, :issue:`2640`)
+
 - The ``TreeSequence`` object now has the attributes ``min_time`` and ``max_time``,
   which are the minimum and maximum among the node times and mutation times,
   respectively. (:user:`szhan`, :pr:`2612`, :issue:`2271`)
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -1764,6 +1764,85 @@ class TestZeroRoots:
         self.verify(ts, no_root_ts)
 
 
+class TestEmptyTrees(TopologyTestCase):
+    """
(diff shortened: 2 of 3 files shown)


benjeffery (approved):
Thanks @hyanwong! LGTM!

jeromekelleher:
Looks good, but I don't really see the value in the more complicated definition. This should have another name (if we need it at all).

hyanwong:
We couldn't decide on a sensible name ("branchless" was best), but came to the conclusion that `tree.num_edges == 0` is pretty clear, so killing this idea unless there's greater demand. Best would be to document this somewhere though

PR #2669, opened 2022-12-23
Implement is_root

### LINKED ISSUE #2620: `tree.is_root()` method
It's tempting to check whether a node is a root by looking for `u in tree.roots` without realising that `tree.roots` might be huge. I think a more efficient way to check is

```
def is_root(self, u)
    return self.num_samples(u) > 0 and self.parent(u) == tskit.NULL
```

Is that right @jeromekelleher ? This might be useful for @szhan in #2618

hyanwong:
## Description

Fixes #2620

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -3899,6 +3899,21 @@ class TestTree(HighLevelTestCase):
         tree = tskit.Tree.generate_balanced(3, root_threshold=4)
         assert tree.num_roots == 0
 
+    @pytest.mark.parametrize("root_threshold", [1, 2, 3])
+    def test_is_root(self, root_threshold):
+        # Make a tree with multiple roots with different numbers of samples under each
+        ts = tskit.Tree.generate_balanced(5).tree_sequence
+        ts = ts.decapitate(ts.max_root_time - 0.1)
+        tables = ts.dump_tables()
+        tables.nodes.add_row(flags=0)  # Isolated non-sample
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE)  # Isolated sample
+        ts = tables.tree_sequence()
+        assert {ts.first().num_samples(u) for u in ts.first().roots} == {1, 2, 3}
+        tree = ts.first(root_threshold=root_threshold)
+        roots = set(tree.roots)
+        for u in range(ts.num_nodes):  # Will also test isolated nodes
+            assert tree.is_root(u) == (u in roots)
+
     def test_is_descendant(self):
         def is_descendant(tree, u, v):
             path = []
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,6 +4,9 @@
 
 **Features**
 
+- A new ``Tree.is_root`` method avoids the need to to search the potentially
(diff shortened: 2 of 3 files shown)


python/tests/test_highlevel.py
```diff
@@ -3867,6 +3867,21 @@ def test_branch_length_empty_tree(self):
         assert tree.branch_length(1) == 0
         assert tree.total_branch_length == 0
 
+    @pytest.mark.parametrize("root_threshold", [1, 2, 3])
```
benjeffery, line 3870: What happens with pathological values of `root_theshold`? (-ve, fractional, infinities) Always good to test.

python/tests/test_highlevel.py
```diff
@@ -3867,6 +3867,21 @@ def test_branch_length_empty_tree(self):
...
+        ts = ts.decapitate(ts.max_root_time - 0.1)
+        tables = ts.dump_tables()
+        tables.nodes.add_row(flags=0)  # Isolated non-sample
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE)  # Isolated sample
+        ts = tables.tree_sequence()
+        assert {ts.first().num_samples(u) for u in ts.first().roots} == {1, 2, 3}
+        tree = ts.first(root_threshold=root_threshold)
+        roots = set(tree.roots)
+        for u in range(ts.num_nodes):  # Will also test isolated nodes
+            assert tree.is_root(u) == (u in roots)
```
benjeffery, line 3883: This tests that the Python `is_root` definition is the same as the (C API's tree iterator + Python `roots`) definition for the `generate_balanced` case, but are there any other topologies that we should check that this holds?

It's not clear to me that visiting the siblings of the left root (as `Tree.roots` does) checks that the sibling meets the `root_threshold`? Maybe I'm missing something here.

benjeffery (commented):
Thanks @hyanwong! I've made some suggestions about testing.
I'm confused about how `Tree.roots` takes account of `root_threshold`, but I assume I'm missing something.

python/tests/test_highlevel.py
hyanwong: Good point. This is actually tested elsewhere, but not thoroughly enough, it appears. See #2674.

I'll add an extra test in when that PR is merged.

python/tests/test_highlevel.py
jeromekelleher: That's taken care of in the lower-level code @benjeffery - stuff only gets into the list of roots if it passes this test.

jeromekelleher (approved):
LGTM

python/tests/test_highlevel.py
jeromekelleher: As in, something can only become a child of the virtual root if it passes the "rootiness" test

benjeffery: Ah, great. I assume this was taken care of, but thought it worth checking.

python/tests/test_highlevel.py
hyanwong: The root_threshold stuff is now tested elsewhere in https://github.com/tskit-dev/tskit/pull/2674 so I think we're covered now

PR #2670, opened 2022-12-24
Skip empty

### LINKED ISSUE #2600: How to skip empty regions at the start and end of a tree seq
We are about to make a change to tsinfer so that the regions before the last site and after the last site in a tree sequence have no topology. I imagine that this could be quite confusing for people looping through the trees, as sometimes (if the first site is not at position 0), the first tree will have num_roots == num_samples, and will not have any topology.

I suspect it would be useful to be able to skip these empty regions at the start and end of a tree sequence when using the `trees()` iterator. Have we any suggestions for a nice interface to do this?

hyanwong:
## Description

Based off #2668, and fixes #2600. I'm not sure if storing the `skip_empty` value as a python attribute in the `Tree` class is the right approach (it's not what's done for root_threshold) but from the tests it appears as if the call signature of `Tree()` and `ts.trees()` should be the same. Otherwise I would just have this as a parameter sent to the `TreeIterator` iterator class from the `ts.trees()` method, and not bother storing it in the `Tree` class itself.

Suggestions for other approaches welcome

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -652,6 +652,7 @@ class Tree:
         tracked_samples=None,
         *,
         sample_lists=False,
+        skip_empty=None,
         root_threshold=1,
         sample_counts=None,
     ):
@@ -671,6 +672,7 @@ class Tree:
 
         self._tree_sequence = tree_sequence
         self._ll_tree = _tskit.Tree(tree_sequence.ll_tree_sequence, **kwargs)
+        self.skip_empty = skip_empty  # This is a python-only facility, not in _ll
         self._ll_tree.set_root_threshold(root_threshold)
         self._make_arrays()
 
@@ -3851,6 +3853,7 @@ class TreeIterator:
         self.tree = tree
         self.more_trees = True
         self.forward = True
+        self.no_skip_empty = True if tree.skip_empty is None else not tree.skip_empty
 
     def __iter__(self):
         return self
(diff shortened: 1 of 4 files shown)


python/tskit/trees.py
```diff
@@ -3822,6 +3853,7 @@ def __init__(self, tree):
         self.tree = tree
         self.more_trees = True
         self.forward = True
+        self.no_skip_empty = True if tree.skip_empty is None else not tree.skip_empty
```
jeromekelleher, line 3856: More natural to make ``skip_empty`` a parameter of the TreeIterator class, so we call it like:
```
return TreeIterator(tree, skip_empty=skip_empty)
```

python/tskit/trees.py
```diff
@@ -3831,15 +3863,19 @@ def __reversed__(self):
         return self
 
     def __next__(self):
-        if self.forward:
-            self.more_trees = self.more_trees and self.tree.next()
-        else:
-            self.more_trees = self.more_trees and self.tree.prev()
-        if not self.more_trees:
-            raise StopIteration()
+        while True:
```
jeromekelleher, line 3866: This looks correct, but I think we'd need a good bit of testing to check what happens when we have runs of empty trees in the middle. Although, come to think of it I don't think we can have two adjacent empty trees, so maybe this while look could be replace with something simpler?

jeromekelleher (commented):
Looks good, but I'm certain now that having a single simple definition of ``is_empty`` is the right approach.  There's no need to add attributes to the tree object then, since it's purely a property of the iterator.

hyanwong:
Closed due to https://github.com/tskit-dev/tskit/issues/2600#issuecomment-1377301009

PR #2674, opened 2023-01-06
Ban negative root_thresholds

hyanwong:
I spotted this as a result of https://github.com/tskit-dev/tskit/pull/2669

Is it worth a changelog entry?

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2022 Tskit Developers
+# Copyright (c) 2018-2023 Tskit Developers
 # Copyright (c) 2015-2018 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


hyanwong:
Hmm, not sure why this is failing now. Seems a trivial change to me.

hyanwong:
Oh, I see, the `root_threshold` value in C is of type `tsk_size_t` so can't be negative.

We should either change the type, or check for negative values before passing the root_threshold to the C function in `_tskitmodule.c`. I'm not sure which is preferred.

hyanwong:
@mergifyio rebase

jeromekelleher:
> Is it worth a changelog entry?

Depends on what happened before - what is  the current behaviour if we set a root threshold of 0 of -1?

Should probably drop the copyright notice updates on files that aren't actually changed here like trees.c

hyanwong:
> > Is it worth a changelog entry?
> 
> Depends on what happened before - what is the current behaviour if we set a root threshold of 0 of -1?

It is treated as an unsigned in by `PyArg_ParseTuple(args, "I", &threshold)` I think, so presumably treated as a huge unsigned integer.

> Should probably drop the copyright notice updates on files that aren't actually changed here like trees.c

Sure. That was a left-over from a previous attempt, but as soon as you touch a file, pre-commit wants you to change the date. I can simply force pre-commit not to run in that case.

jeromekelleher:
> It is treated as an unsigned in by PyArg_ParseTuple(args, "I", &threshold) I think, so presumably treated as a huge unsigned integer.

OK - but we should check this to make sure we're not introducing bugs, particularly for 0.

hyanwong:
I think the code for checking 0 is left untouched (and the tests for that are still in there)

jeromekelleher:
OK, if all we're changing is the negative case then this is fine. It wasn't clear from the diff.

hyanwong:
> OK, if all we're changing is the negative case then this is fine. It wasn't clear from the diff.

The new python code checks for `<=0` and raises an error. The older C code checks simply for 0 (since it is an unsigned int by the time it gets to C), and that is still present, and tested in the lowlevel routine. So the new python check changes the error type when called from within python, as it is intercepted before it hits C. I could change the Python code to check for only strictly less than zero, but it seemed more useful to the user to raise the same error for both.

jeromekelleher (approved):

jeromekelleher:
LGTM, just good to clear up the copyright changes on untouched files.

hyanwong:
> LGTM, just good to clear up the copyright changes on untouched files.

Done

jeromekelleher (approved):

PR #2685, opened 2023-01-12
Implement edge differences from table collections.

molpopgen:
## Description

This PR allows for edge differences to be iterated over from a table collection.
The current API only has this feature for tree sequences, but all of the functionality
happens at the level of the tables.
There are several possible use cases for this, too.

# PR Checklist:

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2019-2022 Tskit Developers
+ * Copyright (c) 2019-2023 Tskit Developers
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -23,6 +23,7 @@
  */
 
 #include "testlib.h"
+#include "tskit/core.h"
 #include <tskit/tables.h>
 
 #include <float.h>
@@ -10420,6 +10421,35 @@ test_table_collection_delete_older(void)
     tsk_treeseq_free(&ts);
 }
 
+static void
+test_table_collection_edge_diffs_errors(void)
+{
+    int ret;
+    tsk_id_t ret_id;
+    tsk_table_collection_t tables;
+    tsk_diff_iter_t iter;
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+    tables.sequence_length = 1;
+    ret_id
+        = tsk_node_table_add_row(&tables.nodes, TSK_NODE_IS_SAMPLE, 0, -1, -1, NULL, 0);
+    CU_ASSERT_EQUAL(ret_id, 0);
+    ret_id = tsk_node_table_add_row(&tables.nodes, 0, 1, -1, -1, NULL, 0);
+    CU_ASSERT_EQUAL(ret_id, 1);
+    ret = (int) tsk_edge_table_add_row(&tables.edges, 0., 1., 1, 0, NULL, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+
+    ret = tsk_diff_iter_init(&iter, &tables, -1, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_TABLES_NOT_INDEXED);
+
+    tables.nodes.time[0] = 1;
+    ret = tsk_diff_iter_init(&iter, &tables, -1, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_BAD_NODE_TIME_ORDERING);
+
+    tsk_table_collection_free(&tables);
+}
+
 int
 main(int argc, char **argv)
 {
@@ -10542,6 +10572,8 @@ main(int argc, char **argv)
         { "test_table_collection_takeset_indexes",
             test_table_collection_takeset_indexes },
         { "test_table_collection_delete_older", test_table_collection_delete_older },
+        { "test_table_collection_edge_diffs_errors",
+            test_table_collection_edge_diffs_errors },
         { NULL, NULL },
     };
 
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2019-2022 Tskit Developers
+ * Copyright (c) 2019-2023 Tskit Developers
  * Copyright (c) 2017-2018 University of Oxford
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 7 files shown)


c/tests/test_tables.c
```diff
@@ -10420,6 +10421,20 @@ test_table_collection_delete_older(void)
...
+{
+    int ret;
+    tsk_table_collection_t tables;
+    tsk_diff_iter_t iter;
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = tsk_edge_table_add_row(&tables.edges, 0., 1., 0, 1, NULL, 0);
+    CU_ASSERT_EQUAL(ret, 0);
+    ret = tsk_diff_iter_init_tables(&iter, &tables, 0);
+    CU_ASSERT_EQUAL(ret, TSK_ERR_TABLES_NOT_INDEXED);
```
jeromekelleher, line 10435: Need to free the tables here or the valgrind tests will fail.

jeromekelleher: Best free the diff_iter too I guess

c/tskit/tables.c
```diff
@@ -33,6 +34,7 @@
 #include <math.h>
 
 #include <tskit/tables.h>
+#include <tskit/trees.h>
```
jeromekelleher, line 37: We don't want to include trees.h here - the idea is that you don't have to compile the trees code if you're only interested in the tables API.

c/tskit/tables.c
```diff
@@ -23,6 +23,7 @@
  * SOFTWARE.
  */
 
+#include "tskit/core.h"
```
jeromekelleher, line 26: This should be down in the other group? Unless there's something funky going on with shadowing the other imports?

jeromekelleher (commented):
LGTM

Good call on moving this into the tables code. We haven't documented it, so I'd prefer to just change the API wholesale by moving the code into the tables.c/h files, changing the signature to take a tsk_table_collection_t and fixing up the handful of places that we use this internally. It's not documented, so it's just an internal refactor.

As a side note, I'm not a big fan of this code and I would like to replace it with something cleaner at some point. I guess you don't mind too much once the same functionality is available from Rust's perspective?

c/tskit/tables.c
molpopgen: Yeah, I got stuck on this once I saw how the imports work out.  We don't need trees.h beyond a couple of constants.

molpopgen:
> As a side note, I'm not a big fan of this code and I would like to replace it with something cleaner at some point. I guess you don't mind too much once the same functionality is available from Rust's perspective?

We can change liberally here given that the public API remains the same, IMO.  We don't want to force tskit 2.0 just yet!

molpopgen:
@jeromekelleher -- this one will remain a draft for a little bit.  I need to ponder the imports and put out some other fires (upcoming mid term exams, etc..)

molpopgen:
> > As a side note, I'm not a big fan of this code and I would like to replace it with something cleaner at some point. I guess you don't mind too much once the same functionality is available from Rust's perspective?
> 
> We can change liberally here given that the public API remains the same, IMO. We don't want to force tskit 2.0 just yet!

Ah -- just saw your comment re: whole API change.  I can get on board with that!

jeromekelleher:
Great, no hurry from my perspective.

c/tskit/tables.c
molpopgen: I "fixed" this by moving the relevant defines to tables.h.  I think I got the doxygroup stuff right?

c/tskit/tables.c
molpopgen: I was able to delete this w/no harm.  Not sure what I was thinking.

c/tskit/tables.c
```diff
@@ -13003,3 +13003,164 @@ tsk_squash_edges(tsk_edge_t *edges, tsk_size_t num_edges, tsk_size_t *num_output
...
+    tsk_bug_assert(tables != NULL);
+    tsk_memset(self, 0, sizeof(tsk_diff_iter_t));
+    self->num_nodes = tables->nodes.num_rows;
+    self->num_edges = tables->edges.num_rows;
+    self->tables = tables;
+    self->insertion_index = 0;
+    self->removal_index = 0;
+    self->tree_left = 0;
+    self->tree_index = -1;
+    self->last_index
```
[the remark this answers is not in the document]
molpopgen: This PR gives us two ways to initialize this iterator:

* from a tree sequence 
* from a table collection

Prior to this PR, `self->last_index` got pulled from the tree sequence's `num_trees` field.  But, since a table collection doesn't know how many trees it has, this PR always calls the table validator to get that value. The table validation is not necessary when the input is a table collection owned/referred to by a tree sequence -- those tables have already been validated.

c/tests/test_tables.c
molpopgen: Done.

c/tskit/tables.c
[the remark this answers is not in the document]
jeromekelleher: Right, gotcha. that is annoying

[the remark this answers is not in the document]
molpopgen: So it seems we have a couple of options: use flags.  Currently, the flag is assumed to be zero, so we'd perhaps want a non-zero to say "count trees" and require that for tables.  However, that flag setup is actually the opposite of what is needed by default behavior and is impossible (?) to verify at run time (we cannot know where the table collection pointer comes from).  Need to ponder it a bit more.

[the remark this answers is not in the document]
molpopgen: Another option is to somehow use an internal API to always "do the right thing".  That's where I need to think harder.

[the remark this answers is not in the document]
jeromekelleher: It's a bit fiddly, but I guess we could change the signature of the init function to take ``num_trees`` as a parameter, which you pass ``-1`` if you don't know it. It's an internal only API not used that much, so not too terrible?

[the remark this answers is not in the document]
molpopgen: That was one of my first thoughts.  I agree that it isn't too bad.  My concern is that it is public via C, even though it is undocumented.  The only way to tell the user that the value needs to be correct is via documentaton.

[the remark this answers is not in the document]
jeromekelleher: It's an internal API and I don't think we should document it, so :shrug:

We can coordinate on not breaking tskit-rust I guess, maybe just with some comments to remind us (rather than actual docs)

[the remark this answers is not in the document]
jeromekelleher: We have to have the freedom to change undocumented things, otherwise you go mad trying to write anything new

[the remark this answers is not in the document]
molpopgen: Yeah, that works for me.  tskit-rust will just figure itself out -- I'll manage the constructor functions to do the right thing and any breakage will only happen internally, which is no big deal.

molpopgen:
@jeromekelleher -- I think this one is ready for a closer look.

jeromekelleher:
I just tweaked the API a bit to make it consistent, and covered a couple of loopholes. Is this ok, or just gratuitous breakage for Rust? I can revert the ``tsk_edge_diff_init_from_ts`` change if you like.

molpopgen:
I think that this is fine re: new function names.  For the rust side of things, this stuff is all just called internally, so the encapsulation means we have an easy fix when the time comes. (That's also what we get for using undocumented API!)

jeromekelleher:
OK, great - I think we just need to fix up the build breakage and it's ready to go. I'll fix this up and merge when I get a chance.

molpopgen:
@jeromekelleher -- I think I got this sorted out.  The Python test failures look like codecov (again/still) having issues.

jeromekelleher:
OK, great, let's squash and merge!

c/tskit/tables.c
```diff
@@ -13003,3 +13003,170 @@ tsk_squash_edges(tsk_edge_t *edges, tsk_size_t num_edges, tsk_size_t *num_output
...
+    if (num_trees < 0) {
+        num_trees = tsk_table_collection_check_integrity(self->tables, TSK_CHECK_TREES);
+        if (num_trees < 0) {
+            ret = (int) num_trees;
+            goto out;
+        }
+    }
+    self->last_index = num_trees;
+
+    if (!tsk_table_collection_has_index(self->tables, 0)) {
```
jeromekelleher, line 13035: Ah, just spotted that this is redundant now, as TSK_CHECK_TREES includes it

molpopgen: Okay, I'll give it a spin after removing that block.

molpopgen:
@jerome -- I removed the redundancy, rebased onto  main, and squashed.  I expect that Codecov still hates me, though, as far as the Python tests are concerned.

jeromekelleher (approved):

Issue #2666, opened 2022-12-15
Method to in-place subset a table


jeromekelleher:
In #2665 I added a draft function that looked like this:

```c
int
tsk_node_table_delete_rows(tsk_node_table_t *self, bool *delete_rows,
    tsk_flags_t TSK_UNUSED(options), tsk_id_t *id_map)
{
    int ret;
    tsk_node_table_t copy;
    tsk_node_t node;
    tsk_size_t j;
    tsk_id_t ret_id;

    ret = tsk_node_table_copy(self, &copy, 0);
    if (ret != 0) {
        goto out;
    }
    ret = tsk_node_table_clear(self);
    if (ret != 0) {
        goto out;
    }
    for (j = 0; j < copy.num_rows; j++) {
        if (id_map != NULL) {
            id_map[j] = TSK_NULL;
        }
        if (!delete_rows[j]) {
            tsk_node_table_get_row_unsafe(&copy, (tsk_id_t) j, &node);
            ret_id = tsk_node_table_add_row(self, node.flags, node.time, node.population,
                node.individual, node.metadata, node.metadata_length);
            if (ret_id < 0) {
                ret = (int) ret_id;
                goto out;
            }
            if (id_map != NULL) {
                id_map[j] = ret_id;
            }
        }
    }
out:
    tsk_node_table_free(&copy);
    return ret;
}
```

The idea is that we can keep/delete a subset of the rows in a table, in place, and optionally return the mapping of new-to-old IDs. I implemented this as ``delete_rows`` here, but it felt a bit unnatural and maybe it would be better to do it as ``keep_rows`` instead? Maybe ``subset(bool *keep_rows)``? Any better ideas?

I juggled with the idea of providing a list of row indexes instead, but I think the boolean mask is more generally useful, and is easy to set up if you do have a list of rows you want to get rid of/keep.

The idea is to do this for all tables, and for tables that self-reference (e.g. mutation table) to automatically remap those references and to (by default) return an error if you try to delete a row that has references to it.

See also #1034

benjeffery:
Sounds like a good addition to the API - a boolean mask seems to best way to me. I assume that we're not too concerned with perf here? If we are an easy win would be to expand the table once (as the number of rows needed is known) and call `tsk_node_table_add_row_internal` instead.

benjeffery:
One more thought, just realised - the code above is almost equal to:
```
ret = tsk_node_table_copy(self, &copy, 0);
...
[...]

PR #2700, opened 2023-01-25
Table subset

jeromekelleher:
A start on #2666

The nice thing about the boolean mask is that it does let us do the subsetting efficiently in-place (because the output order is fixed) so I think we should probably do this more efficiently, avoiding the intermediate copy. I'll tip along at this as I get a chance.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -1492,6 +1492,106 @@ test_node_table_update_row(void)
 }
 
 static void
+test_node_table_keep_rows(void)
+{
+    int ret;
+    tsk_id_t ret_id;
+    tsk_size_t j;
+    tsk_node_table_t source, t1, t2;
+    tsk_node_t row;
+    bool keep[3] = { 1, 1, 1 };
+    tsk_id_t id_map[3];
+    const char *metadata = "ABC";
+    tsk_id_t indexes[] = { 0, 1, 2 };
+
+    ret = tsk_node_table_init(&source, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret_id = tsk_node_table_add_row(&source, 0, 1.0, 2, 3, metadata, 1);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(&source, 1, 2.0, 3, 4, metadata, 2);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(&source, 2, 3.0, 4, 5, metadata, 3);
+    CU_ASSERT_FATAL(ret_id >= 0);
+
+    ret = tsk_node_table_copy(&source, &t1, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_node_table_keep_rows(&t1, keep, 0, id_map);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_node_table_equals(&t1, &source, 0));
+
+    ret = tsk_node_table_keep_rows(&t1, keep, 0, NULL);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_TRUE(tsk_node_table_equals(&t1, &source, 0));
+    CU_ASSERT_EQUAL_FATAL(id_map[0], 0);
+    CU_ASSERT_EQUAL_FATAL(id_map[1], 1);
+    CU_ASSERT_EQUAL_FATAL(id_map[2], 2);
+
+    keep[0] = 0;
+    keep[1] = 0;
+    keep[2] = 0;
+    ret = tsk_node_table_keep_rows(&t1, keep, 0, id_map);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(t1.num_rows, 0);
+    CU_ASSERT_EQUAL_FATAL(id_map[0], -1);
+    CU_ASSERT_EQUAL_FATAL(id_map[1], -1);
+    CU_ASSERT_EQUAL_FATAL(id_map[2], -1);
+
+    ret = tsk_node_table_copy(&source, &t1, TSK_NO_INIT);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    keep[0] = 0;
+    keep[1] = 1;
+    keep[2] = 0;
+    ret = tsk_node_table_keep_rows(&t1, keep, 0, id_map);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(t1.num_rows, 1);
+    CU_ASSERT_EQUAL_FATAL(id_map[0], -1);
+    CU_ASSERT_EQUAL_FATAL(id_map[1], 0);
+    CU_ASSERT_EQUAL_FATAL(id_map[2], -1);
+
+    ret = tsk_node_table_get_row(&t1, 0, &row);
+    CU_ASSERT_EQUAL_FATAL(row.flags, 1);
+    CU_ASSERT_EQUAL_FATAL(row.time, 2.0);
+    CU_ASSERT_EQUAL_FATAL(row.population, 3);
+    CU_ASSERT_EQUAL_FATAL(row.individual, 4);
+    CU_ASSERT_EQUAL_FATAL(row.metadata_length, 2);
+    CU_ASSERT_EQUAL_FATAL(row.metadata[0], 'A');
+    CU_ASSERT_EQUAL_FATAL(row.metadata[1], 'B');
+
+    tsk_node_table_free(&t1);
+
+    keep[0] = 0;
+    keep[1] = 0;
+    keep[2] = 0;
+    /* Keeping first n rows equivalent to truncate */
+    for (j = 0; j < source.num_rows; j++) {
+        ret = tsk_node_table_copy(&source, &t2, 0);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = tsk_node_table_copy(&source, &t1, 0);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        ret = tsk_node_table_truncate(&t1, j + 1);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        keep[j] = 1;
+        ret = tsk_node_table_keep_rows(&t2, keep, 0, NULL);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        CU_ASSERT_TRUE(tsk_node_table_equals(&t1, &t2, 0));
+
+        /* Adding the remaining rows back on to the table gives the original
+         * table */
+        ret = tsk_node_table_extend(
+            &t2, &source, source.num_rows - j - 1, indexes + j + 1, 0);
+        CU_ASSERT_EQUAL_FATAL(ret, 0);
+        CU_ASSERT_TRUE(tsk_node_table_equals(&source, &t2, 0));
+
+        tsk_node_table_free(&t1);
+        tsk_node_table_free(&t2);
+    }
+
+    tsk_node_table_free(&source);
+}
+
+static void
 test_edge_table_with_options(tsk_flags_t options)
 {
     int ret;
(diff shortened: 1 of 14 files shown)


jeromekelleher:
This is ready for a review now @benjeffery - I went with ``x_table_keep_rows`` rather than ``subset`` because the ``table_collection_subset`` does something a bit different, and it would have been confusing to use the same word.

Ping @petrelharp and @bhaller also, since this is a basic utility needed for breaking up the simplify operation into smaller bits for parallelisation

jeromekelleher:
Dammit, went copy-paste mad too soon, and didn't test it properly. Back to draft!

jeromekelleher:
I think the core of this is more-or-less in place now, and it would be good to get some feedback.

The main thing that might raise some controversy is the semantics around mutation parent ID remapping (look at mutation_table_keep_rows). Basically, we want to make it easy to get rid of mutations, which is currently a major faff because of the parents column. This does involve making sure that reference integrity is maintained, though, and the approach I've taken is to do some basic checks and to error-out if someone tries to do something that doesn't appear to make sense. We could easily add some options to change this behaviour, but I think this is a good default.

If we agree I'll write up the details in the documentation, and apply the same logic to the individual table (which is the only other place we have self-referential tables.

@molpopgen - this is also adjacent to the "more modular simplify" work you're interested in, so good to see what you think of the proposed API also.

molpopgen:
If I'm following along correctly: after simplifying "chunks", the client aggregates a list of nodes to delete, and can get back a new node "global" node mapping for the start of the next recording epoch?

jeromekelleher:
> If I'm following along correctly: after simplifying "chunks", the client aggregates a list of nodes to delete, and can get back a new node "global" node mapping for the start of the next recording epoch?

Yeah, that's one of the things that's made easier - you can efficiently repack the node table at the end of a simplification epoch to get rid of unused nodes. Also, I'm planning to use this internally within the simplifier to make updating the site, mutation and individual tables easier. We should be able to get to a point where we don't copy any of the tables during simplification (although will need to think about how to handle the input edges - but that's close to what you're thinking about too).

molpopgen:
Sounds good.  I'll try to take a closer look at this early this week.  My "modular simplification" idea has the nice property of not interacting with the input edge table.  Hopefully that'll mean things "just work" in the face of any refactoring.

c/tests/test_tables.c
```diff
@@ -1490,6 +1490,98 @@ test_node_table_update_row(void)
...
+    CU_ASSERT_EQUAL_FATAL(id_map[1], 1);
+    CU_ASSERT_EQUAL_FATAL(id_map[2], 2);
+
+    keep[0] = 0;
+    keep[1] = 0;
+    keep[2] = 0;
+    ret = tsk_node_table_keep_rows(&t1, keep, 0, id_map);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL_FATAL(t1.num_rows, 0);
+    CU_ASSERT_EQUAL_FATAL(id_map[0], -1);
```
benjeffery, line 1535: I know it is a small thing, but I'm never quite sure where the line is between using `-1` and `TSK_NULL`. In the implementation it is `TSK_NULL`

c/tests/test_tables.c
```diff
@@ -1490,6 +1490,98 @@ test_node_table_update_row(void)
...
+    CU_ASSERT_EQUAL_FATAL(row.metadata_length, 2);
+    CU_ASSERT_EQUAL_FATAL(row.metadata[0], 'A');
+    CU_ASSERT_EQUAL_FATAL(row.metadata[1], 'B');
+
+    tsk_node_table_free(&t1);
+
+    keep[0] = 0;
+    keep[1] = 0;
+    keep[2] = 0;
+    /* Keeping first n rows equivalent to truncate */
```
benjeffery, line 1565: Nice

python/tests/test_lowlevel.py
```diff
@@ -763,6 +763,40 @@ def test_table_extend_types(
...
+    @pytest.mark.parametrize("table_name", ["nodes"])
+    def test_table_keep_rows_errors(self, table_name, ts_fixture):
+        table = getattr(ts_fixture.tables, table_name)
+        n = len(table)
+        ll_table = table.ll_table
+        with pytest.raises(ValueError, match="must be of length"):
+            ll_table.keep_rows(np.ones(n - 1, dtype=bool))
+        with pytest.raises(ValueError, match="must be of length"):
+            ll_table.keep_rows(np.ones(n + 1, dtype=bool))
+        with pytest.raises(TypeError, match="Cannot cast"):
```
benjeffery, line 775: This probably will be in the high-level tests, but given recent issues, probably worth adding tests for things like `False`.

benjeffery (commented):
Very nice, seems all the hairy details are under control.
One extra test that might be worth doing is adding rows to the table after this operation - seeing that it resizes by changing `num_rows` and the old memory getting reused. As it looks this should be all fine, as `max_rows` keeps track of that, but maybe worth testing? Fine if not, but it's the only extra thing I could think of!

c/tests/test_tables.c
jeromekelleher: It's tricky all right, they're sort of interchangable in my head. You do need to know that TSK_NULL is -1 as a developer so I guess I'm just lazy with typing. Happy to change if you think it improves readability

jeromekelleher:
Good call @benjeffery - I'll add some tests for adding more rows

c/tests/test_tables.c
bhaller: > It's tricky all right, they're sort of interchangable in my head. You do need to know that TSK_NULL is -1 as a developer so I guess I'm just lazy with typing. Happy to change if you think it improves readability

I'm curious about this too.  Why does one need to know that it is -1, as a developer?  Couldn't there be a strict rule that TSK_NULL should *always* be used, in cases where it applies?  Might it be helpful to actually change TSK_NULL to -2, for a test build, and see what breaks, and fix all of those places?  Etc.  I have always found this to be one of the more confusing parts of tskit's APIs; maybe it could be made more rigorous?  Anyhow, just random thoughts as I'm looking at these GitHub emails.  :->

jeromekelleher: In this case you need to know that -1 is valid input and -2 isn't because we're testing the validation code. Given this knowledge, -1 is sometimes more readable than TSK_NULL (say, if we're declaring an array like ``{-1, 0, -1, 2}``). But it's a judgement call, and opinions will differ.

bhaller: Hmm.  To me, if -1 means `TSK_NULL` in a given context, then declaring an array as `{-1, 0, -1, 2}` would be very confusing, because it would imply that -1 is "just another value" and mask its special meaning, whereas `{TSK_NULL, 0, TSK_NULL, 2}` would be much more clear.  But as you say, I guess opinions vary!  :->  Coming from the C++ world, I would be itching to find a way to make using -1 be literally illegal and cause a compile error, in such contexts.  (I don't think such a mechanism exists, though, even in C++; things like "enum class" are gestures in that direction, but don't go far enough to cover the usage case that `TSK_NULL` inhabits.)  This is parallel to my opinion that using `0` to mean `NULL`/`nullptr` (in C/C++) is confusing and wrong; in C you don't get a compile error, but in C++ you actually do (or perhaps it's a warning, depending on settings?); `nullptr` was specifically invented to allow using 0 to become a compile error, as I understand it.  Anyhow, interesting discussion.  Personally, I'd vote for doing a pass over tskit to enforce API purity concerning `TSK_NULL`, but I happily admit that that is just my opinion!  :->

bhaller: (And I'd gently suggest that if this point is confusing even for @benjeffery, then it is probably a point of confusion worth clearing up for pretty much every user of tskit...)

jeromekelleher: Well, I'm not sure I made this clear at the top, but I'm talking about tskit C library core developers here, not people writing client code. They are certainly not expected to know that TSK_NULL is -1. These tests are for core developers to read and work with. I don't think Ben is confused about -1 being TSK_NULL, he just wasn't sure when to use one or the other stylistically.

I really don't think we should enforce always using TSK_NULL within the library code, as this would result in code like this:
```c
if (id < TSK_NULL || id >= self->num_rows) {
    ..  error
}
```
You have to know that TSK_NULL is -1 for this to be safe, and given that, it's better to just write -1 and be explicit about it.

Or I guess you could do 
```c
if ((id != TSK_NULL && id < 0) || id >= self->num_rows) {
     ... error
}
```
then you could have TSK_NULL as -2, but what's the point? Writing 

```c
if (id < -1|| id >= self->num_rows) {
     ... error
}
```
is simple, clear and obvious that you're not allowing any loopholes.

TSK_NULL **will always be** -1. There are good reasons for it to be this value, from multiple different angles and I can foresee no circumstances in which it would be a good idea to change it to a different value.

So, gently disagreeing with suggestion :stuck_out_tongue_winking_eye:

bhaller: To me,

`if ((id < 0 || id >= self->num_rows) && (id != TSK_NULL))`

is clearer in its meaning.  The goal is not, of course, to redefine `TSK_NULL` to be -2 instead.  It's just to make it clear in the code that `TSK_NULL` is a special value.  With

`if (id < -1 || id >= self->num_rows)`

it looks, to my eye, as if either (a) there is a bug in the code and it allows -1 incorrectly, or (b) an index of -1 is for some reason permitted.  When written that way, the code does not express that the reason that -1 is permitted is that -1 is `TSK_NULL`; the reader has to _guess_ that that might be the reason, and is left wondering whether that guess is correct.

But I guess this comes down to personal coding style, and there's no need to beat the dead horse further.  :->  It's just interesting to explore these philosophical differences, I think.  I certainly understand better now why the tskit code treats -1 vs. `TSK_NULL` in the way it does.

jeromekelleher: I think part of it is that -1 predates TSK_NULL usage by quite a long way, so (a) I'm used to thinking about it this way, and (b) lots of the code was written before TSK_NULL existed. I can see how it looks sloppy when coming from the other perspective of (correctly) thinking about TSK_NULL as an anonymous special value.

jeromekelleher:
OK, I think this is ready for final review now. Jeez that was a lot of code for a simple operation, but I'm hoping this (and a few similar ops like #2707) will help us trim down the code as well as making things more efficient over time.

molpopgen:
Great -- I was feeling guilty for not looking yet. We have a holiday weekend coming up, but I'll definitely look next week.

benjeffery (approved):
LGTM!

c/tskit/tables.h
```diff
@@ -1425,6 +1468,36 @@ and is not checked for compatibility with any existing schema on this table.
...
+@param keep Array of boolean flags describing whether a particular
+    row should be kept or not. Must be at least ``num_rows`` long.
+@param options Bitwise option flags. Currently unused; should be
+    set to zero to ensure compatibility with later versions of tskit.
+@param id_map An array in which to store the mapping between new
+    and old IDs. If NULL, this will be ignored.
+@return Return 0 on success or a negative value on failure.
+*/
+int tsk_node_table_keep_rows(
+    tsk_node_table_t *self, const bool *keep, tsk_flags_t options, tsk_id_t *id_map);
```
molpopgen, line 1499: Overall, the API makes sense.  There's one technical issue that should be documented:

* When interacting with C functions, C++ code will tend to store things in a vector and obtain the pointer via the data member function.  For example, `x.data()` yields a pointer to the first element, which is then passed to a C API.
* C++ `vector<bool>` totally breaks all of what I just said.  Due to a grave lapse of judgement, whether bool is stored or individual elements are bit fields is implementation defined.

There's nothing we can do except to note that "the pointer from a c++ vector<bool> is not the right thing to pass in here".

molpopgen (approved):
This LGTM.  There's one technical issue that can only be handled via documentation.

c/tskit/tables.h
bhaller: Or tell people to use `std::vector<uint8_t>` instead of `std::vector<bool>`, right?  As I understand it, that avoids the problem.

molpopgen: Yes, so long as uint8_t  and bool have the same sizeof everywhere.  That's true for all platforms that I expect to encounter, but I'd still use a `static_assert` to check it at compile time.

molpopgen: Yet another reason why I'm drifting away from C++.  From cppreference:

```
bool - type, capable of holding one of the two values: [true](https://en.cppreference.com/w/cpp/language/bool_literal) or [false](https://en.cppreference.com/w/cpp/language/bool_literal). The value of sizeof(bool) is implementation defined and might differ from 1.
```

And the first passes but the second fails:

```cpp
#include <cstdint>
#include <type_traits>

int
main(int, char**)
{
    static_assert(sizeof(std::uint8_t) == sizeof(bool), "sizeof failure");
    static_assert(std::is_same<std::uint8_t, bool>::value,
                  "uint8_t and bool aren't the same");
}

```

I think we have to say "be careful here" when using C++.  Too much is platform-dependent.

bhaller: Technically, true that they could be different – but as you said, on all platforms that we expect to encounter, they are in fact the same.  A matter of taste how much one wants to worry about that, I guess.  :->

jeromekelleher:
That's a really good point @molpopgen  that I struggled a bit with at the Python C API level as well - the C99 bool type is frustratingly vague and I think it could end up being a problem in the future.

How about we follow [numpy](https://github.com/numpy/numpy/blob/main/numpy/core/include/numpy/npy_common.h#L352) and do this:
```c
typedef uint8_t tsk_bool_t;
```

That should make things a bit easier to do portably? 

For C++ we put in documentation warning saying "make sure you use ``std::vector<tsk_bool_t>`` and not  ``std::vector<bool>``?

bhaller:
> typedef uint8_t tsk_bool_t;

That's what I do in SLiM/Eidos:

`typedef uint8_t eidos_logical_t;`

molpopgen:
> That's a really good point @molpopgen that I struggled a bit with at the Python C API level as well - the C99 bool type is frustratingly vague and I think it could end up being a problem in the future.
> 
> How about we follow [numpy](https://github.com/numpy/numpy/blob/main/numpy/core/include/numpy/npy_common.h#L352) and do this:
> 
> ```c
> typedef uint8_t tsk_bool_t;
> ```
> 
> and then change the signature of these functions to use `tsk_bool_t` rather than `bool`.
> 
> That should make things a bit easier to do portably?
> 
> For C++ we put in documentation warning saying "make sure you use `std::vector<tsk_bool_t>` and not `std::vector<bool>`?

Yes, this is the way to go.

jeromekelleher:
OK, great, I'll make that change.

jeromekelleher:
OK, I think this is ready to go now!

molpopgen:
LGTM. (This codecov thing is annoying!)

jeromekelleher:
I'm going to merge this manually, as the only thing that's failing is one codecov upload, which has occurred successfully on other builds.

PR #2710, opened 2023-02-17
Divergence matrix

jeromekelleher:
Initial divergence matrix code adapted from https://github.com/petrelharp/divmat

@petrelharp - I've just copied in the divmat code and tweaked things a bit to cover the annoying non-sample topology cases. This could probably be handled more efficiently, but I don't think it's worth worrying about.

The C is just a framework at the moment, I'm going to work on that for a bit and will ping you when it's ready for a look.

### DIFF
--- a/c/avl.c
+++ b/c/avl.c
@@ -1,0 +1,604 @@
+// clang-format off
+/*****************************************************************************
+
+    avl.c - Source code for the AVL-tree library.
+
+    Copyright (C) 1998  Michael H. Buselli <cosine@cosine.org>
+    Copyright (C) 2000-2002  Wessel Dankers <wsl@nl.linux.org>
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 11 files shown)


jeromekelleher:
As a way of getting some indicative perf numbers I implemented this temporarily using the avl tree implementation we use in msprime. All the tests are passing, so I think it's correct. This is now deleting 0 values as you'd expect.

Unfortunately the performance is still worse than the tree-by-tree numba version 😢

I may have some errors in there causing this? I don't really have time to dig in unfortunately

Any thoughts @petrelharp ?

petrelharp:
Hmph! I'll have a look, but perhaps we can pull in Sriram?

petrelharp:
My other thoughts are:
- we could do some basic counting of operations (in the python code?) for both methods, to verify the new algorithm is actually doing less operations
- your benchmarking was on a big enough example, I suppose?

jeromekelleher:
Superseded by #2736 - closing

Issue #2126, opened 2022-02-08
Accessing metadata schema values not obvious


hyanwong:
```
>>> print(ts.tables.nodes.metadata_schema)
OrderedDict([('codec', 'json')])
>>> ts.tables.nodes.metadata_schema["codec"]
TypeError: 'MetadataSchema' object is not subscriptable
>>> ts.tables.nodes.metadata_schema.asdict()["codec"]
'json'
```

Having to use `.asdict()` to get the schema values, when the printed version says it's an OrderedDict is pretty unintuitive, I would say.

jeromekelleher:
In general we need higher level APIs for accessing and modifying metadata schemas - do we have some issues open to track this @benjeffery?

At the moment you have to work directly with the ``schema`` attribute on the MetadataSchema, which you treat as a plain dictionary and is interpreted directly as JSON schema.

hyanwong:
> At the moment you have to work directly with the schema attribute on the MetadataSchema, which you treat as a plain dictionary and is interpreted directly as JSON schema.

So I think it's just a bit confusing that `print(ts.tables.nodes.metadata_schema)` and `print(ts.tables.nodes.metadata_schema.schema)` give the same thing. Perhaps the string version of metadata_schema should imply that you need to use .schema to get at the good stuff.

benjeffery:
Thanks for spotting this @hyanwong - I've added this to the metadata project. We'll soon take all the metadata issues and plan out a coherent high-level API for the schemas.

PR #2720, opened 2023-03-02
Md null schema

hyanwong:
While thinking with metadata for the `tskit` poster, I remembered this on the queue. As discussed with @benjeffery - this is mostly a documentation PR, but it also changes behaviour so that `str(tskit.MetadataSchema(schema=None))` returns `"Null_schema"` rather than `None`, which was pretty confusing, because `tskit.MetadataSchema(schema=None) is None` always returns `False`.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -24,6 +24,9 @@
   in the presence of non-ancestral material (very rare).
   (:user:`petrelharp`, :issue:`2983`, :pr:`1623`)
 
+- Printing ``tskit.MetadataSchema(schema=None)`` now shows ``"Null_schema"`` rather
+  than ``None``, to avoid confusion (:user:`hyanwong`, :pr:`2720`)
+
 **Features**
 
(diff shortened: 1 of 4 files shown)


hyanwong:
Having looked at https://github.com/tskit-dev/tskit/issues/2126, I wonder if, instead, we should report the `str()` as `MetadataSchema(None)`, and indeed report *all* schemas like that? This would fully address #2126

hyanwong:
I now think we should use this for the `str` representation of a schema:

```python
    def __str__(self) -> str:
        if isinstance(self._schema, collections.OrderedDict):
            s = pprint.pformat(dict(self._schema))
        else:
            s = pprint.pformat(self._schema)
        if "\n" in s:
            return f"MetadataSchema(\n{s}\n)"
        else:
            return f"MetadataSchema({s})"
```

What do you think, @benjeffery ? Will this break anything?

benjeffery:
I've rebased this and make the string `tskit.MetadataSchema(None)`

benjeffery:
@mergifyio rebase

Issue #2729, opened 2023-04-03
Possible bug in the AFS calculation


lukashuebner:
I get strange results when trying to calculate the AFS for the following dataset:

```cpp
// Same as the example from the PLOS paper but with back and recurring mutations
/*
0.25┊     8   ┊         ┊         ┊
    ┊   ┏━┻━┓ ┊         ┊         ┊
0.20┊   ┃   ┃ ┊         ┊   7     ┊
    ┊   ┃   ┃ ┊         ┊ ┏━┻━┓   ┊
0.17┊   6   ┃ ┊   6     ┊ ┃   ┃   ┊
    ┊ ┏━┻┓  ┃ ┊ ┏━┻━┓   ┊ ┃   ┃   ┊
0.09┊ ┃  5  ┃ ┊ ┃   5   ┊ ┃   5   ┊
    ┊ ┃ ┏┻┓ ┃ ┊ ┃ ┏━┻┓  ┊ ┃ ┏━┻┓  ┊
0.07┊ ┃ ┃ ┃ ┃ ┊ ┃ ┃  4  ┊ ┃ ┃  4  ┊
    ┊ ┃ ┃ ┃ ┃ ┊ ┃ ┃ ┏┻┓ ┊ ┃ ┃ ┏┻┓ ┊
0.00┊ 0 1 3 2 ┊ 0 1 2 3 ┊ 0 1 2 3 ┊
  0.00      2.00      7.00      10.00
*/
char const* multi_tree_back_recurrent_nodes = "1  0       -1   0\n"
                                              "1  0       -1   0\n"
                                              "1  0       -1   1\n"
                                              "1  0       -1   1\n"
                                              "0  0.071   -1   -1\n"
                                              "0  0.090   -1   -1\n"
                                              "0  0.170   -1   -1\n"
                                              "0  0.202   -1   -1\n"
                                              "0  0.253   -1   -1\n";

char const* multi_tree_back_recurrent_edges = "2 10 4 2\n"
                                              "2 10 4 3\n"
                                              "0 10 5 1\n"
                                              "0 2  5 3\n"
                                              "2 10 5 4\n"
                                              "0 7  6 0,5\n"
                                              "7 10 7 0,5\n"
                                              "0 2  8 2,6\n";

char const* multi_tree_back_recurrent_sites = "1      0\n"
                                              "4.5    0\n"
                                              "8.5    0\n";

/* site, node, derived_state, [parent, time] */
[...]

PR #2730, opened 2023-04-03
compute mutation parents on text input

petrelharp:
Over in #2729 we saw a bug due to not computing mutation parents on text input. This isn't really a bug, as the text input is not part of the public API, but I could easily imagine us making the same error in test code.

(More generally we ought to think about doing this as part of tree sequence instantiation; I seem to recall us having a reason not to do it, but I don't remember specifically what it was?)

### DIFF
--- a/c/tests/test_trees.c
+++ b/c/tests/test_trees.c
@@ -3739,13 +3739,13 @@ test_simplest_mutation_edges(void)
     const char *mutations = "0    1     1\n"
                             "0    0     1\n"
                             "0    2     1\n"
+                            "1    2     1\n"
                             "1    0     1\n"
-                            "1    1     1\n"
-                            "1    2     1\n";
+                            "1    1     1\n";
     tsk_treeseq_t ts;
     tsk_tree_t tree;
     /* We have mutations over roots, samples and just isolated nodes */
-    tsk_id_t mutation_edges[] = { -1, 0, -1, 1, -1, -1 };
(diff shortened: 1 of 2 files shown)


petrelharp:
Closes #2729.

benjeffery:
@mergifyio rebase

petrelharp:
Whoops; looks like I've made some silly mistake.

petrelharp:
Okay - I got this to work; I had to rearrange the tables in one of the tests.

HOWEVER, I'm worried that putting in `compute_mutation_parents` might *not* be the right way to go here. This is because there may be some operations we want to do without having indexes or mutation parents, so I am nervous putting these into all our examples.

jeromekelleher:
> HOWEVER, I'm worried that putting in compute_mutation_parents might not be the right way to go here. This is because there may be some operations we want to do without having indexes or mutation parents, so I am nervous putting these into all our examples.

I'm tending to agree actually. These functions are just for creating test examples, and it'll be easy to forget that any mutation parents you specify will be nuked by this call.

Maybe we should either leave it as it is, or only compute mutation parents when that column is missing from the input?

petrelharp:
Maybe we add a flag to the function, that if true does indexes and computes mutation parents, and if false then does not? That'd make it real hard to miss the fact that the function does or doesn't do it.

benjeffery:
@petrelharp I'm having a tidy up of PRs - I think this one can be closed?

PR #2736, opened 2023-04-12
divmat tree by tree

jeromekelleher:
This is a simple implementation of the divergence matrix, based on the infrastructure developed in
#2710.

The idea is that we do the very straighforward and naive thing of going pairwise through the samples,
but compute the MRCAs using a constant time algorithm. It's slightly faster than the current AVL
tree based approach in #2710, while using far less memory. Another advantage is that this implementation
is simple to parallelise by adding left and right coordinate arguments
and taking advantage of the new quick seeking in #2661. I think this is  probably
more important in practise now than good algorithms, as we'll need to throw lots of threads
at these calculations whatever happens.

I think we'll probably want to keep this implementation around as an option in the long run anyway,
since it's likely to be faster than more complex algorithms for small n, and it uses very little
memory (which may also be important in practise).

So, I vote we push ahead with fleshing out the API for this?

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -1191,9 +1191,11 @@ out:
  * General stats framework
  ***********************************/
 
+#define TSK_REQUIRE_FULL_SPAN 1
+
 static int
-tsk_treeseq_check_windows(
-    const tsk_treeseq_t *self, tsk_size_t num_windows, const double *windows)
+tsk_treeseq_check_windows(const tsk_treeseq_t *self, tsk_size_t num_windows,
+    const double *windows, tsk_flags_t options)
 {
     int ret = TSK_ERR_BAD_WINDOWS;
     tsk_size_t j;
@@ -1202,12 +1204,23 @@ tsk_treeseq_check_windows(
         ret = TSK_ERR_BAD_NUM_WINDOWS;
         goto out;
     }
-    /* TODO these restrictions can be lifted later if we want a specific interval. */
-    if (windows[0] != 0) {
-        goto out;
-    }
-    if (windows[num_windows] != self->tables->sequence_length) {
-        goto out;
+    if (options & TSK_REQUIRE_FULL_SPAN) {
+        /* TODO the general stat code currently requires that we include the
+         * entire tree sequence span. This should be relaxed, so hopefully
+         * this branch (and the option) can be removed at some point */
+        if (windows[0] != 0) {
+            goto out;
+        }
+        if (windows[num_windows] != self->tables->sequence_length) {
+            goto out;
+        }
+    } else {
+        if (windows[0] < 0) {
+            goto out;
+        }
+        if (windows[num_windows] > self->tables->sequence_length) {
+            goto out;
+        }
     }
     for (j = 0; j < num_windows; j++) {
         if (windows[j] >= windows[j + 1]) {
@@ -1960,7 +1973,8 @@ tsk_treeseq_general_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
         num_windows = 1;
         windows = default_windows;
     } else {
-        ret = tsk_treeseq_check_windows(self, num_windows, windows);
+        ret = tsk_treeseq_check_windows(
+            self, num_windows, windows, TSK_REQUIRE_FULL_SPAN);
         if (ret != 0) {
             goto out;
         }
@@ -2468,7 +2482,7 @@ tsk_treeseq_allele_frequency_spectrum(const tsk_treeseq_t *self,
     bool stat_site = !!(options & TSK_STAT_SITE);
     bool stat_branch = !!(options & TSK_STAT_BRANCH);
     bool stat_node = !!(options & TSK_STAT_NODE);
-    double default_windows[] = { 0, self->tables->sequence_length };
+    const double default_windows[] = { 0, self->tables->sequence_length };
     const tsk_size_t num_nodes = self->tables->nodes.num_rows;
     const tsk_size_t K = num_sample_sets + 1;
     tsk_size_t j, k, l, afs_size;
@@ -2496,7 +2510,8 @@ tsk_treeseq_allele_frequency_spectrum(const tsk_treeseq_t *self,
         num_windows = 1;
         windows = default_windows;
     } else {
-        ret = tsk_treeseq_check_windows(self, num_windows, windows);
+        ret = tsk_treeseq_check_windows(
+            self, num_windows, windows, TSK_REQUIRE_FULL_SPAN);
         if (ret != 0) {
             goto out;
         }
@@ -3331,7 +3346,7 @@ tsk_treeseq_simplify(const tsk_treeseq_t *self, const tsk_id_t *samples,
     }
     ret = tsk_treeseq_init(
         output, tables, TSK_TS_INIT_BUILD_INDEXES | TSK_TAKE_OWNERSHIP);
-    /* Once tsk_tree_init has returned ownership of tables is transferred */
+    /* Once tsk_treeseq_init has returned ownership of tables is transferred */
     tables = NULL;
 out:
     if (tables != NULL) {
@@ -3459,6 +3474,20 @@ out:
 /* ======================================================== *
  * Tree
  * ======================================================== */
+
+/* Return the root for the specified node.
+ * NOTE: no bounds checking is done here.
+ */
+static tsk_id_t
+tsk_tree_get_node_root(const tsk_tree_t *self, tsk_id_t u)
+{
+    const tsk_id_t *restrict parent = self->parent;
+
+    while (parent[u] != TSK_NULL) {
+        u = parent[u];
+    }
+    return u;
+}
 
 int TSK_WARN_UNUSED
 tsk_tree_init(tsk_tree_t *self, const tsk_treeseq_t *tree_sequence, tsk_flags_t options)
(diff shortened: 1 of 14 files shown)


jeromekelleher:
This is ready for a look @petrelharp - most of the stuff is implemented I think, except for the ``samples`` argument (which is trivial). The threading stuff is mostly not profiled, so I don't know how well it performs in practise (the basic "split the genome by tree" one works well though). I'm making threading stuff reasonably general with the goal of applying it to the stats API later - all we'd need in principle is a way of limiting the general stats calculations to a contiguous slice of the genome, and it would then mostly apply directly.

Questions:
- This implementation is coming up with a different answer when we have trees with multiple roots. Do we special-case to a slower implementation that finds the local root when the MRCA is null?
- Should we call this ``branch_divergence_matrix`` for clarity, or should we also think about implementing the "site" version of this also (not sure how much work that is though)?

jeromekelleher:
I think this is more-or-less done now @petrelharp - the only thing I'm not sure about is the
semantics for multiroot trees. Shouldn't the divergence between disconnected subtrees be
infinity, rather than the sum of the local distances to root?

I guess that would cause problems for tree sequences that contain missing data though (say
a snipped out centromere), as the divergence matrix would just be infinity everywhere then.

jeromekelleher:
Tests failing because of #2745

jeromekelleher:
Aha, turns out there isn't a problem at all, it's just a subtle numerical precision problem around 0.

Ready for a look now I think!

petrelharp:
Notes from the meeting now:
- `mode="site"`: we should add the `mode` argument, and if `"site"` it should do just the same thing, but measure branch lengths by counting the number of mutations on them.
- `individuals`: we'd like to obtain the indiv x indiv matrix, so should add this argument, but in a later PR? (And, note that the easy way to do this is postprocessed, by numpy; however, the indiv-by-indiv matrix is 4x smaller, so if we did push the operation down all the way we'd have a not-insignificant reduction in memory.)
- `samples` or `sample_sets`? (tldr; proposal is keep API as-is): Other stats methods use `sample_sets`; we *could* follow this, and special-case so that if sample sets is a flat list of IDs then it computes the n x n matrix between those samples. However, we think maybe this is the wrong way to go, because if the API follows the other stats like this, people will want to use this method like `sample_sets = [ts.samples(population=0), ts.samples(population=1)]`, which would be horribly inefficient if there's lots of samples per population and we're just getting the reduced matrix by numpy as mentioned above. Also, we can't think of use cases besides (a) `samples = <list of sample IDs>`, (b) `samples = <list of individuals>`, and (c) `samples = <list of large groups of samples, eg, populations>`; the proposal is that people doing (c) would use `ts.divergence( )` instead of this function, even though that does something slightly different in the `mode="site"` case. (Yes, this will lead to confusion, but confusion is inevitable.)

jeromekelleher:
I think having sample_sets, samples and individuals as mutually exclusive parameters that do what we want would be fine? Specifying a list of n singleton sets is quite annoying and ineffecient, and having another parameter to me is better than overloading sample_sets.

So, keep the API as is for now, but ultimately implement following the standard sample sets approach in C, and then add support in python for individuals and sample_sets?

jeromekelleher:
How does this look as a basic definition of the site-mode version @petrelharp? If we agree on this as what we're trying to compute, I'll have a think about how to do things a bit more efficiently (maintaining cumulative mutation counts per node should be possible so that we can use the efficient MRCA calculation, but will involve writing a full-fat incremental algorithm).

```python3
def rootward_path(tree, u, v):
    while u != v:
        yield u
        u = tree.parent(u)


def site_divergence_matrix_naive(ts, windows=None, samples=None):
    windows_specified = windows is not None
    windows = [0, ts.sequence_length] if windows is None else windows
    num_windows = len(windows) - 1
    samples = ts.samples() if samples is None else samples

    n = len(samples)
    D = np.zeros((num_windows, n, n))
    tree = tskit.Tree(ts)
    for i in range(num_windows):
        left = windows[i]
        right = windows[i + 1]
        tree.seek(left)
        # Iterate over the trees in this window
        while tree.interval.left < right and tree.index != -1:
            span_left = max(tree.interval.left, left)
            span_right = min(tree.interval.right, right)
            mutations_per_node = collections.Counter()
            for site in tree.sites():
                if span_left <= site.position < span_right:
                        mutations_per_node[mutation.node] += 1
            for j in range(n):
                u = samples[j]
                for k in range(j + 1, n):
                    v = samples[k]
                    w = tree.mrca(u, v)
                    if w != tskit.NULL:
                        wu = w
                        wv = w
                    else:
                        wu = local_root(tree, u)
                        wv = local_root(tree, v)
                    du = sum(mutations_per_node[x] for x in rootward_path(tree, u, wu))
                    dv = sum(mutations_per_node[x] for x in rootward_path(tree, v, wv))
                    # NOTE: we're just accumulating the raw mutation counts, not
                    # multiplying by span
                    D[i, j, k] += du + dv
            tree.next()
        # Fill out symmetric triangle in the matrix
        for j in range(n):
            for k in range(j + 1, n):
                D[i, k, j] = D[i, j, k]
    if not windows_specified:
        D = D[0]
    return D

```

jeromekelleher:
I pushed through the changes here for the site mode divmat @petrelharp, and I think we're almost API complete enough to go for the initial merge.

Annoyingly, the basic tree-by-tree site mode version that I have here is much slower than I thought it would be, and we need a different algorithm. In some quick testing, we currently have the branch mode divmat is at least 10X faster than the general-stat based method, whereas the site mode is at least 10X **slower**.

I don't really know why this is exactly - it might just be that doing the MRCAs by traversal really is that much slower than using Schieber-Vishkin. But it's clear we'll need to do some more work here.

What I'd suggest is that we fill out the testing here on this much, to make sure we're all happy that this really is what we want to compute, and then revisit after this PR is merged, with a better algorithm for site-mode. Are you happy with this? Is the current API a good forwards-compatible starting point (pending the change of default mode to "stat")?

python/tests/test_divmat.py
```diff
@@ -0,0 +1,992 @@
...
+                u = samples[j]
+                for k in range(j + 1, n):
+                    v = samples[k]
+                    w = sv_mrca(u, v, *tables)
+                    assert w == tree.mrca(u, v)
+                    if w != tskit.NULL:
+                        tu = ts.nodes_time[w] - ts.nodes_time[u]
+                        tv = ts.nodes_time[w] - ts.nodes_time[v]
+                    else:
+                        # Question for @petralharp - is this the right definition?
```
petrelharp, line 188: I like this definition? Advantages I can think of are (a) it matches what we'd get from the `site` stat, and (b) if you divide the tree into slices by time and compute the contribution to divergence in each time slice, they add to the whole thing?

c/tskit/trees.c
```diff
@@ -6009,3 +6038,520 @@ tsk_treeseq_kc_distance(const tsk_treeseq_t *self, const tsk_treeseq_t *other,
...
+                }
+            }
+        }
+    }
+}
+
+static void
+sv_tables_build(sv_tables_t *self, tsk_tree_t *tree)
+{
+    sv_tables_reset(self, tree);
```
petrelharp, line 6212: Wikipedia tells me that
> [Sleator](https://en.wikipedia.org/wiki/Daniel_Sleator) and[ Tarjan](https://en.wikipedia.org/wiki/Robert_Tarjan) ([1983](https://en.wikipedia.org/wiki/Lowest_common_ancestor#CITEREFSleatorTarjan1983)) proposed the dynamic LCA variant of the problem in which the data structure should be prepared to handle LCA queries intermixed with operations that change the tree (that is, rearrange the tree by adding and removing edges). This variant can be solved in O ( log ⁡ N ) O(\log N) time in the total size of the tree for all modifications and queries.

... I don't suppose you've looked into this?

petrelharp: Huh, but reading the paper (https://www.cs.cmu.edu/~sleator/papers/dynamic-trees.pdf) it seems to be O(log(n)) to find MRCAs, not O(1), which is different that SV.

petrelharp (approved):
I see no issues - LGTM!

c/tskit/trees.c
jeromekelleher: I've been wanting to get a clever CS person to delve into this literature for years - an incrementally updated version of SV would be super powerful. But it's all quite hard unfortunately...

petrelharp: Have you thought about the other O(1) method, involving mapping to RMQ (see Wikipedia)? That method seems fairly understandable and dynamic-able.

petrelharp: Hm, although we'd then need a dynamic RMQ structure... never mind.

jeromekelleher:
I think this is ready for a final look and merge. Would you mind having a quick scan please @benjeffery?

benjeffery (approved):
Looks great! 
I note that `numba` is mentioned but not used - was the intent to use it in the test suite?

jeromekelleher:
> I note that numba is mentioned but not used - was the intent to use it in the test suite?

Well spotted. It's handy to have a version of this code lying around that you can run numba on directly, but I don't think we need to bother with adding numba support here for functions in the test suite.

I've added a comment to clarify

c/tskit/trees.c
```diff
@@ -6009,3 +6038,519 @@ tsk_treeseq_kc_distance(const tsk_treeseq_t *self, const tsk_treeseq_t *other,
...
+        stat_site = true;
+    }
+    /* It's an error to specify more than one mode */
+    if (stat_site + stat_branch > 1) {
+        ret = TSK_ERR_MULTIPLE_STAT_MODES;
+        goto out;
+    }
+
+    if ((options & TSK_STAT_POLARISED) || (options & TSK_STAT_SPAN_NORMALISE)) {
+        /* TODO better error */
```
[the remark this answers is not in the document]
jeromekelleher: I've added the specific errors.

jeromekelleher:
I've tied up a few loose ends, this is ready to go now.

Also opened #2791 to track the missing support for string-window specs.

Issue #2625, opened 2022-11-02
Make tqdm(ts.variants()) work better


jeromekelleher:
It would be nice if this would pull out the overall length of the iter. Should be easy enough?

jeromekelleher:
Should be straightforward, we just have to abstract the code for the variants function into an iterator class (which returns the right len), a bit like the current TreeIterator class.

hyanwong:
Just revisiting this. The `variants()` method has `left` and `right` parameters, so it's a bit more complicated. An iterator would need to know not just ts.num_sites, but the number of sites between left and right. I suspect we would also want to change the `trees()` iterator to have `left` and `right` too (see https://github.com/tskit-dev/tskit/issues/24), so perhaps we should implement that first, change the TreeIterator to account for that, then use the same basic function to add `len` to the variants iterator.

jeromekelleher:
I don't think the ``trees`` iterator is needed for this, as we use the ``Tree`` class directly.

benjeffery:
Breakpoints are stored on the tree sequence object, and are accessible from Python, so I think that could be used to get a count of the trees in the interval for `len`

hyanwong:
> Breakpoints are stored on the tree sequence object, and are accessible from Python, so I think that could be used to get a count of the trees in the interval for `len`

Yeah, this does it, I think:

```python
ts._check_genomic_range(left, right)
breaks = self.breakpoints(as_array=True)
left_index = breaks.searchsorted([left], side="right")
right_index = breaks.searchsorted([right], side="left")
num_trees = right_index - left_index + 1
```

benjeffery:
Closing for inactivity and labelling "future", please re-open if you plan to work on this.

benjeffery:
Ah didn't notice it had a PR. That PR is in draft so hasn't been reviewed.

PR #2762, opened 2023-06-13
Generic iterator class for trees and variants

hyanwong:
## Description

Pull out the logic that iterates over trees and variants into a single base class, and allow this class to return a length.
Also allows left, right, copy to be passed to the `trees()` iterator (to match the `variants()` iterator), so that we can
iterate over a subset of the genome.

Fixes https://github.com/tskit-dev/tskit/issues/2625, https://github.com/tskit-dev/tskit/issues/24

# PR Checklist:


hyanwong:
BTW, this renders `ts.aslist()` a bit obsolete, as you can do `ts.trees(copy=True)`, in the same way you do for variants.

PR #2776, opened 2023-07-05
Add doc example of creating a matrix of all pairwise comparisons

hyanwong:
This is a common thing to want to do


jeromekelleher (commented):
Let's keep this one on ice until we have the pairwise matrix code up to speed. We won't want people to land on this (much slower) way of doing the same thing.

hyanwong:
Fair point. I guess the API would be different then.

petrelharp:
Wait darn it I was moving too fast - I thought this was an issue, so edited it to reflect a bigger thing that I wanted to do. Apologies, I"ll undo that.

PR #2785, opened 2023-07-07
Tests and documentation for genetic_relatedness_weighted

brieuclehmann:
I've created a new PR to address the remaining tasks for #1246, namely adding tests and documentation. The tests are a bit messy but I think they cover everything. I'm not sure how to write the raw summary function under the general_stat framework, so would appreciate a hand / some pointers with this @petrelharp !

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -71,6 +71,7 @@ appears beside the listed method.
     * Multi-way
         * {meth}`~TreeSequence.divergence`
         * {meth}`~TreeSequence.genetic_relatedness`
+          {meth}`~TreeSequence.genetic_relatedness_weighted`
         * {meth}`~TreeSequence.f4`
           {meth}`~TreeSequence.f3`
           {meth}`~TreeSequence.f2`
@@ -593,6 +594,12 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
   of samples.
 
+`genetic_relatedness_weighted`
+: {math}`f(w_i, w_j, x_i, x_j) = \frac{1}{2}(x_i - w_i m) (x_j - w_j m)`,
+
+  where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
+  of samples, and {math}`w_j = \sum_{k=1}^n W_kj` is the sum of the weights in the {math}`j`th column of the weight matrix.
+  
 `Y2`
 : {math}`f(x_1, x_2) = \frac{x_1 (n_2 - x_2) (n_2 - x_2 - 1)}{n_1 n_2 (n_2 - 1)}`
 
--- a/python/tests/test_tree_stats.py
+++ b/python/tests/test_tree_stats.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2022 Tskit Developers
+# Copyright (c) 2018-2023 Tskit Developers
 # Copyright (C) 2016 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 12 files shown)


python/tskit/trees.py
```diff
@@ -7650,6 +7691,51 @@ def genetic_relatedness(
...
+        r"""
+        Computes weighted genetic relatedness: if the k-th pair of indices is (i, j)
+        then the k-th column of output will be
+        :math:`\sum_{a,b} W_{ai} W_{bj} C_{ab}`,
+        where :math:`W` is the matrix of weights, and :math:`C_{ab}` is the
+        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>` between sample
+            a and sample b, summing over all pairs of samples in the tree sequence.
+
+        :param numpy.ndarray W: An array of values with one row for each sample and one
+            column for each set of weights.
```
petrelharp, line 7712:
```suggestion
        :param numpy.ndarray W: An array of values with one row for each sample node and
            one column for each set of weights.
```

petrelharp:
Well, let's see: do you mean writing the python general stat function, for tests [here](https://github.com/tskit-dev/tskit/pull/2785/files#diff-4d6ee7e1fdeb228e5a8a5fe5c37af121dc87237d0c30563eafab80282fe3ed0eR2289)? We just need to translate the C code to python, really:
```
genetic_relatedness_weighted_summary_func(tsk_size_t state_dim, const double *state,
    tsk_size_t result_dim, double *result, void *params)
{
    indexed_weight_stat_params_t args = *(indexed_weight_stat_params_t *) params;
    const double *x = state;
    tsk_id_t i, j;
    tsk_size_t k;
    double meanx, ni, nj;

    meanx = state[state_dim - 1] / args.total_weights[state_dim - 1];
    for (k = 0; k < result_dim; k++) {
        i = args.index_tuples[2 * k];
        j = args.index_tuples[2 * k + 1];
        ni = args.total_weights[i];
        nj = args.total_weights[j];
        result[k] = (x[i] - ni * meanx) * (x[j] - nj * meanx) / 2;
    }
    return 0;
}
```
... or, better, go back and look at the definition of this and see what the python code ought to be? Let's see, I actually think it should be exactly the same as what you've got there, from genetic relatedness, which is:
```
        def f(x):
            mx = np.sum(x) / n  # this will be in the last entry of `state` in the C code
            return np.array(
                [(x[i] - n[i] * mx) * (x[j] - n[j] * mx) / 2 for i, j in indexes]
```

docs/stats.md
```diff
@@ -590,6 +591,11 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
   where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
   of samples.
 
+`genetic_relatedness_weighted`
+: {math}`f(w_i, w_j, x_i, x_j) = \frac{1}{2}w_iw_j(x_i - m)(x_j - m)`,
```
petrelharp, line 595:
```suggestion
: {math}`f(w_i, w_j, x_i, x_j) = \frac{1}{2} w_i w_j (x_i - m) (x_j - m)`,
```

docs/stats.md
```diff
@@ -590,6 +591,11 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+  where {math}`m = \frac{1}{n}\sum_{k=1}^n x_k` with {math}`n` the total number
+  of samples.
```
petrelharp, line 598: Is this right? We wrote this out carefully in the theory; and looking at the C code it looks like `m` should be the average weight or something, but I've not gone back to the theory yet?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.3.0
```
petrelharp, line 3: Hm, looks like you did `pre-commit update` or something? I *think* you're not supposed to do that; to un-do you can do `git checkout main .pre-commit-config.yaml`.

python/tests/test_tree_stats.py
```diff
@@ -1010,6 +1010,43 @@ def wrapped_summary_func(x):
             self.assertArrayAlmostEqual(sigma1, sigma4)
 
 
+class kWayWeightStatsMixin(WeightStatsMixin):
```
petrelharp, line 1013: I don't think this class is used?

petrelharp:
Comments above; otherwise, things look good! Thanks!

jeromekelleher:
I'll have a look after you've addressed @petrelharp's comments @brieuclehmann, and can sort out any remaining details (if they're giving you trouble)

jeromekelleher:
I just did a bit of git surgery @brieuclehmann - I'll take it from here, thanks!

PR #2786, opened 2023-07-07
Modular incremental trees

jeromekelleher:
Try to centralise our tree positioning code by bringing the logic into ``TreePosition`` class. See #2778 for discussion.

@molpopgen - the idea here is to try and pull out all the of the code that we use in the ``tsk_tree_t`` class for tree positioning into this class, so that it can then be shared by incremental algorithms as well. We should then be able to do efficient skipping to a particular tree, which will be useful for parallel algorithms.

WIP


molpopgen:
Sounds great.  IMO, it is important to have that core logic separated from updating all those tree arrays!

jeromekelleher:
I think this is ready for review and merging now.

The idea is to add a new class ``TreePosition`` (or ``tsk_tree_position_t`` in C) which takes care of the details of keeping track of which edges you need to insert and remove from the current tree as you move along the sequence. This is already a substantial improvement over what we have now because it allows us to reuse code in incremental algorithms, and we can also now define incremental algorithms that go backwards as well as forwards.

I've implemented the basic next/prev operations here in Python and C, and used the core code to replace the bits of the library code that use the ``tsk_tree_diff_iterator``. There's lots more that can be done with this, which I'll open some issues to track.

One key thing that needs to be mentioned here is that I've incorporated quite a clean way of seeking to arbitrary locations along the sequence while touching only the edges that need to be changed in order to transform the current tree into the target tree. **However** this has raise this issue that these edges are not necessarily inserted in time-increasing order. This needs to be thought about a bit, which is why I haven't gone ahead and switched over to using this code for the ``tsk_tree_t`` class itself.

@molpopgen - any chance you could take a look and see what you think?

molpopgen:
I'll take a look, but it'll be next week.

jeromekelleher:
@duncanMR good to get your eyes on this also please

duncanMR:
@jeromekelleher I will start examining the code today. Have you found a test case where the edges aren't inserted in time-decreasing order?

jeromekelleher:
Excellent. I haven't picked out specific examples where the edges aren't time sorted in seek_forward, and that would be a great thing to do in a follow up. 

I stopped working on the random seek aspect once I realised the sortedness problem, and decided to focus on getting the basic code merged so we could start using it. So it's in a half finished state at the moment

molpopgen:
Let's see if I got this:

1. The new type book-keeps index ranges of output/input edges.  This is essentially the edge diffs for the new tree.
2. Nodes can get out of (time) order because the input/output index vectors are sorted first by position and then time, and input/output do time in different orders from one another?

jeromekelleher:
> The new type book-keeps index ranges of output/input edges. This is essentially the edge diffs for the new tree.

Yep - but without having to allocate any extra memory because we're using the existing index arrays to represent the ranges.

> Nodes can get out of (time) order because the input/output index vectors are sorted first by position and then time, and input/output do time in different orders from one another?

This is subtle. The input/output arrays are sorted by position and time, which guarantees that we insert/remove the edges in the correct order for *adjacent* trees. However, when we're seeking to *non-adjacent* trees (or randomly into the middle of the sequence, like seek_from_null) we lose this guarantee.

jeromekelleher:
Can I get a review here please @benjeffery? I want to merge this so downstream stuff can go in too

jeromekelleher:
@mergifyio rebase

python/tests/tsutil.py
```diff
@@ -1713,6 +1715,196 @@ def iterate(self):
             left = right
 
 
+FORWARD = 1
+REVERSE = -1
+
+
+@dataclasses.dataclass
+class Interval:
```
benjeffery, line 1723: Why not use the existing `Interval` class?

c/tskit/trees.c
```diff
@@ -3553,6 +3553,164 @@ tsk_treeseq_split_edges(const tsk_treeseq_t *self, double time, tsk_flags_t flag
...
+        self->in.stop = 0;
+        self->out.stop = 0;
+        self->direction = TSK_DIR_FORWARD;
+    }
+
+    if (self->direction == TSK_DIR_FORWARD) {
+        left_current_index = self->in.stop;
+        right_current_index = self->out.stop;
+    } else {
+        left_current_index = self->out.stop + 1;
```
benjeffery, line 3621: Some missing coverage here and below.

benjeffery (commented):
Very nice, just a couple of small comments.

python/tests/tsutil.py
jeromekelleher: Just simpler to keep track of what it does for this kind of prototype/test code, and not have to worry about external APIs and so on

duncanMR:
Sure, will do!

PR #2793, opened 2023-07-13
Update tskit-book-theme

benjeffery:


benjeffery:
Merging now, with the theme unpinned so we can update without touching tskit.

PR #2799, opened 2023-07-13
LS model refactoring

jeromekelleher:
Some refactoring of the LS hmm code. WIP 

based on #2786 (so has a bunch of extra commits).

### DIFF
--- a/c/tests/test_haplotype_matching.c
+++ b/c/tests/test_haplotype_matching.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2019-2022 Tskit Developers
+ * Copyright (c) 2019-2023 Tskit Developers
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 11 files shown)


jeromekelleher:
@astheeggeggs I've made some progress here on refactoring the Python tree LS HMM code, mainly to deduplicate things a bit and also to remove the need for mirroring the coords of the ts. I think it's pretty close to the point where I can implement it in C.

It would be helpful if you could take a look at the code at this point to see if you spot any howlers (just the test_haplotype_matching.py code, everthing else is incidental). I've made a start on refactoring the tests a bit as well, to move away from the current class-based approach where we internally iterate over a bunch of parameters, towards something that's much more modular, and where you can run a single example with a single set of parameters easily. It's a WIP though

jeromekelleher:
I think this is breaking on some lshmm bugs that have been fixed, so the simplest thing is to wait for version 0.0.5 (https://github.com/astheeggeggs/lshmm/issues/25)

I've added a commit to bump the requirements.

Issue #1900, opened 2021-11-10
Add TreeSequence methods for r2 LD and remove LD calculator from external C API


jeromekelleher:
#1864 updated the very old and janky LD code to be a bit easier to work with from a coding perspecive. However, the API of creating a object to do calculations with is an anti-pattern and we want to provide some direct computation of LD values. 

We should add some methods to the TreeSequence Python API for computing LD that will do the same thing and use the same underlying C code, but getting rid of the low-level C-Python LDCalculator object.

The obvious thing to do from a C perspective would be to add the methods:
```c
int tsk_treeseq_get_r2(tsk_treeseq_t *self, tsk_id_t a, tsk_id_t b, double *r2);
int tsk_treeseq_get_r2_array(tsk_treeseq_t *self, tsk_id_t a, int direction,
    tsk_size_t max_sites, double max_distance, double *r2, tsk_size_t *num_r2_values);
int tsk_treeseq_get_r2_matrix(tsk_treeseq_t *self, 
     tsk_size_t num_a_sites,  tsk_id_t *a,   tsk_size_t num_b_sites, tsk_id_t *b, double *A);
```

(The motivation for adding the ``r2_matrix`` method is that there are significant performance gains to be made from doing tree transitions in a clever way when computing many pairs of r2 values. We can't take advantage of this by doing the r2_array independently.)

That will cover all the bases in terms of replicating the existing functionality, but there's a few API design choices we should think through first, especially in light of the planned generalisations (#432)

- Should we think about windowing here?
- What about "mode", should this argument be put in? (We can fail if it's not "site" initially)
- What about "sample_sets", does this make sense here too?
- Is "r2" a good enough name or should we call it "ld_r2" maybe? Then we could also have "ld_D", "ld_r" (#405)? Or should we name it something else entirely?

Basically, it would be good if we could get the basic API  "shape"  in place for the general two-site statistics so that we can port the existing LD code, and finally deprecate the Python LDCalculator.

Any thoughts @apragsdale @petrelharp?

benjeffery:
Closing for inactivity and labelling "future", please re-open if you plan to work on this.

PR #2805, opened 2023-07-18
Two Locus Statistics C API

lkirk:
## Description
As discussed with @jeromekelleher, I've opened up this PR to add the code necessary to create a C API for two locus statistics. This PR implements a general framework for multi-allelic site statistics and summary functions for r, r2, D, D2, D', Dz, pi2 to get things started. I've described this work in a [design document](https://raw.githubusercontent.com/lkirk/ts-two-locus-proto/e26017c/design-document/main.pdf) contained in my prototype repository. This PR diverges in the main entrypoint (I discussed with @jeromekelleher, he thinks that this should have a separate entrypoint from `tsk_treeseq_general_stat`), but everything else in the document is accurate.

I've implemented tests that cover the happy paths. They ensure the correctness of multiple sample sets and the summary functions. I haven't done a deep dive into the test coverage, but I know that there is at least one important branch that is not covered, I'll have to look into that a bit more closely to understand why it's not being hit.

One thing left to implement is windows. Our current thinking for site statistics is that we'd like to have one window on the left and one window on the right so that we can cut any slice out of the correlation matrix. This leaves a few questions in terms of API design that we can get to in another PR. I think this one is large enough as is.

# PR Checklist:

- [ ] Documentation including tutorial content if appropriate. <- I'll probably cover this when I implement the python API.
- [ ] Changelogs, if there are API changes. <- I'll need a bit of help with this

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -1986,6 +1986,438 @@ test_ld_silent_mutations(void)
 }
 
 static void
+test_paper_ex_two_site(void)
+{
+    tsk_treeseq_t ts;
+    double *result;
+    tsk_size_t s, result_size;
+    int ret;
+
+    double truth_one_set[6] = { 1, 0.1111111111111111, 0.1111111111111111, 1, 1, 1 };
+    double truth_two_sets[12] = { 1, 1, 0.1111111111111111, 0.1111111111111111,
+        0.1111111111111111, 0.1111111111111111, 1, 1, 1, 1, 1, 1 };
+    double truth_three_sets[18] = { 1, 1, 0, 0.1111111111111111, 0.1111111111111111, 0,
+        0.1111111111111111, 0.1111111111111111, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
+
+    tsk_treeseq_from_text(&ts, 10, paper_ex_nodes, paper_ex_edges, NULL, paper_ex_sites,
+        paper_ex_mutations, paper_ex_individuals, NULL, 0);
+
+    tsk_size_t sample_set_sizes[3];
+    tsk_size_t num_sample_sets;
+    tsk_id_t sample_sets[ts.num_samples * 3];
+
+    // First sample set contains all of the samples
+    sample_set_sizes[0] = ts.num_samples;
+    num_sample_sets = 1;
+    for (s = 0; s < ts.num_samples; s++) {
+        sample_sets[s] = (tsk_id_t) s;
+    }
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 6);
+    assert_arrays_almost_equal(result_size * num_sample_sets, result, truth_one_set);
+    tsk_safe_free(result);
+
+    // Second sample set contains all of the samples
+    sample_set_sizes[1] = ts.num_samples;
+    num_sample_sets = 2;
+    for (s = ts.num_samples; s < ts.num_samples * 2; s++) {
+        sample_sets[s] = (tsk_id_t) s - (tsk_id_t) ts.num_samples;
+    }
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 6);
+    assert_arrays_almost_equal(result_size * num_sample_sets, result, truth_two_sets);
+    tsk_safe_free(result);
+
+    // Third sample set contains the first two samples
+    sample_set_sizes[2] = 2;
+    num_sample_sets = 3;
+    for (s = ts.num_samples * 2; s < (ts.num_samples * 3) - 2; s++) {
+        sample_sets[s] = (tsk_id_t) s - (tsk_id_t) ts.num_samples * 2;
+    }
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 6);
+    assert_arrays_almost_equal(result_size * num_sample_sets, result, truth_three_sets);
+    tsk_safe_free(result);
+
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_two_site_correlated_multiallelic(void)
+{
+    const char *nodes = "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "1   0   -1\n"
+                        "0   2   -1\n"
+                        "0   4   -1\n"
+                        "0   6   -1\n"
+                        "0   8   -1\n"
+                        "0   10  -1\n"
+                        "0   12  -1\n"
+                        "0   14  -1\n"
+                        "0   16  -1\n";
+    const char *edges = "0   20   9    0,1\n"
+                        "0   20   10   2,9\n"
+                        "0   20   11   4,5\n"
+                        "0   20   12   6,11\n"
+                        "0   20   13   7,8\n"
+                        "0   20   14   3,10\n"
+                        "0   10   15   12\n"
+                        "10  20   15   13\n"
+                        "0   10   15   14\n"
+                        "10  20   15   14\n"
+                        "10  20   16   12\n"
+                        "0   10   16   13\n"
+                        "0   10   16   15\n"
+                        "10  20   16   15\n";
+    const char *sites = "7   A\n"
+                        "13  G\n";
+    const char *mutations = "0   15  T  -1\n"
+                            "0   14  G   0\n"
+                            "1   15  T  -1\n"
+                            "1   13  C   2\n";
+
+    int ret;
+
+    tsk_treeseq_t ts;
+    double *result;
+    tsk_size_t s, result_size;
+
+    double truth_D[3]
+        = { 0.043209876543209874, -0.018518518518518517, 0.05555555555555555 };
+    double truth_D2[3]
+        = { 0.023844603634269844, 0.02384460363426984, 0.02384460363426984 };
+    double truth_r2[3] = { 1, 1, 1 };
+    double truth_D_prime[3]
+        = { 0.7777777777777777, 0.4444444444444444, 0.6666666666666666 };
+    double truth_r[3]
+        = { 0.18377223398316206, -0.12212786219416509, 0.2609542781331212 };
+    double truth_Dz[3]
+        = { 0.0033870175616860566, 0.003387017561686057, 0.003387017561686057 };
+    double truth_pi2[3]
+        = { 0.04579247743399549, 0.04579247743399549, 0.0457924774339955 };
+
+    tsk_treeseq_from_text(&ts, 20, nodes, edges, NULL, sites, mutations, NULL, NULL, 0);
+
+    tsk_size_t sample_set_sizes[1] = { ts.num_samples };
+    tsk_size_t num_sample_sets = 1;
+    tsk_id_t sample_sets[ts.num_samples];
+
+    for (s = 0; s < ts.num_samples; s++) {
+        sample_sets[s] = (tsk_id_t) s;
+    }
+
+    ret = tsk_treeseq_D(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_D2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D2);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_r2);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_D_prime(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0,
+        NULL, 0, NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D_prime);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_r(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_r);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_Dz(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_Dz);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_pi2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL,
+        0, NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_pi2);
+    tsk_safe_free(result);
+
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_two_site_uncorrelated_multiallelic(void)
+{
+    const char *nodes = "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "1   0  -1\n"
+                        "0   2  -1\n"
+                        "0   4  -1\n"
+                        "0   6  -1\n"
+                        "0   8  -1\n"
+                        "0   10 -1\n"
+                        "0   12 -1\n"
+                        "0   14 -1\n"
+                        "0   16 -1\n"
+                        "0   2  -1\n"
+                        "0   4  -1\n"
+                        "0   6  -1\n"
+                        "0   8  -1\n"
+                        "0   10 -1\n"
+                        "0   12 -1\n"
+                        "0   14 -1\n"
+                        "0   16 -1\n";
+    const char *edges = "0     10    9      0,1\n"
+                        "10    20    17     0,3\n"
+                        "0     10    10     2,9\n"
+                        "10    20    18     6,17\n"
+                        "0     10    11     3,4\n"
+                        "10    20    19     1,4\n"
+                        "0     10    12     5,11\n"
+                        "10    20    20     7,19\n"
+                        "0     10    13     6,7\n"
+                        "10    20    21     2,5\n"
+                        "0     10    14     8,13\n"
+                        "10    20    22     8,21\n"
+                        "0     10    15     10,12\n"
+                        "10    20    23     18,20\n"
+                        "0     10    16     14,15\n"
+                        "10    20    24     22,23\n";
+    const char *sites = "7   A\n"
+                        "13  G\n";
+    const char *mutations = "0   15  T  -1\n"
+                            "0   12  G   0\n"
+                            "1   23  T  -1\n"
+                            "1   20  A   2\n";
+
+    tsk_treeseq_t ts;
+
+    int ret;
+    double *result;
+    tsk_size_t result_size;
+
+    double truth_D[3] = { 0.05555555555555555, 0.0, 0.05555555555555555 };
+    double truth_D2[3] = { 0.024691358024691357, 0.0, 0.024691358024691357 };
+    double truth_r2[3] = { 1, 0, 1 };
+    double truth_D_prime[3] = { 0.6666666666666665, 0.0, 0.6666666666666665 };
+    double truth_r[3] = { 0.24999999999999997, 0.0, 0.24999999999999997 };
+    double truth_Dz[3] = { 0.0, 0.0, 0.0 };
+    double truth_pi2[3]
+        = { 0.04938271604938272, 0.04938271604938272, 0.04938271604938272 };
+
+    tsk_treeseq_from_text(&ts, 20, nodes, edges, NULL, sites, mutations, NULL, NULL, 0);
+
+    tsk_size_t sample_set_sizes[1] = { ts.num_samples };
+    tsk_size_t num_sample_sets = 1;
+    tsk_id_t sample_sets[ts.num_samples];
+
+    for (tsk_size_t s = 0; s < ts.num_samples; s++) {
+        sample_sets[s] = (tsk_id_t) s;
+    }
+
+    ret = tsk_treeseq_D(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_D2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D2);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_r2);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_D_prime(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0,
+        NULL, 0, NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_D_prime);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_r(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_r);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_Dz(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_Dz);
+    tsk_safe_free(result);
+
+    ret = tsk_treeseq_pi2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL,
+        0, NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_pi2);
+    tsk_safe_free(result);
+
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_two_site_backmutation(void)
+{
+    const char *nodes
+        = "1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n"
+          "1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n"
+          "1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n"
+          "1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n1 0  -1\n"
+          "1 0  -1\n1 0  -1\n1 0  -1\n0 2  -1\n0 4  -1\n0 6  -1\n0 8  -1\n0 10 -1\n"
+          "0 12 -1\n0 14 -1\n0 16 -1\n0 18 -1\n0 20 -1\n0 22 -1\n0 24 -1\n0 26 -1\n"
+          "0 28 -1\n0 30 -1\n0 32 -1\n0 34 -1\n0 36 -1\n0 38 -1\n0 40 -1\n0 42 -1\n"
+          "0 44 -1\n0 46 -1\n0 48 -1\n0 50 -1\n0 52 -1\n0 54 -1\n0 56 -1\n0 58 -1\n"
+          "0 60 -1\n0 62 -1\n0 64 -1\n0 66 -1\n0 68 -1\n";
+
+    const char *edges
+        = "0 10 35 0,1\n0 10 36 2,35\n0 10 37 3,36\n0 10 38 4,37\n0 10 39 5,38\n"
+          "0 10 40 6,39\n0 10 41 7,40\n0 10 42 8,41\n0 10 43 9,42\n0 10 44 10,43\n"
+          "0 10 45 11,44\n0 10 46 12,45\n0 10 47 13,46\n0 10 48 14,47\n0 10 49 15,48\n"
+          "0 10 50 16,49\n0 10 51 17,50\n0 10 52 18,51\n0 10 53 19,52\n0 10 54 20,53\n"
+          "0 10 55 21,54\n0 10 56 22,55\n0 10 57 23,56\n0 10 58 24,57\n0 10 59 25,58\n"
+          "0 10 60 26,59\n0 10 61 27,60\n0 10 62 28,61\n0 10 63 29,62\n0 10 64 30,63\n"
+          "0 10 65 31,64\n0 10 66 32,65\n0 10 67 33,66\n0 10 68 34,67\n";
+
+    const char *sites = "1    A\n"
+                        "4.5  T\n";
+
+    const char *mutations = "0  50  T  -1\n"
+                            "0  48  G   0\n"
+                            "0  46  A   1\n"
+                            "1  62  G  -1\n"
+                            "1  60  T   3\n"
+                            "1  58  A   4\n";
+
+    int ret;
+    double *result;
+    tsk_size_t result_size;
+
+    tsk_treeseq_t ts;
+    tsk_treeseq_from_text(&ts, 10, nodes, edges, NULL, sites, mutations, NULL, NULL, 0);
+
+    tsk_size_t sample_set_sizes[1] = { ts.num_samples };
+    tsk_size_t num_sample_sets = 1;
+    tsk_id_t sample_sets[ts.num_samples];
+
+    for (tsk_size_t s = 0; s < ts.num_samples; s++) {
+        sample_sets[s] = (tsk_id_t) s;
+    }
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    /* assert_arrays_almost_equal(result_size, result, truth_r2); */
+    tsk_safe_free(result);
+
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_two_locus_stat_input_errors(void)
+{
+    tsk_treeseq_t ts;
+    double *result;
+    tsk_size_t s, result_size;
+    int ret;
+
+    tsk_treeseq_from_text(&ts, 1, single_tree_ex_nodes, single_tree_ex_edges, NULL,
+        single_tree_ex_sites, single_tree_ex_mutations, NULL, NULL, 0);
+
+    tsk_size_t sample_set_sizes[1];
+    tsk_size_t num_sample_sets;
+    tsk_id_t sample_sets[ts.num_samples];
+
+    sample_set_sizes[0] = ts.num_samples;
+    num_sample_sets = 1;
+    for (s = 0; s < ts.num_samples; s++) {
+        sample_sets[s] = (tsk_id_t) s;
+    }
+
+    sample_sets[1] = 0;
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_DUPLICATE_SAMPLE);
+    sample_sets[1] = 1;
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, TSK_STAT_SITE | TSK_STAT_BRANCH, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_MULTIPLE_STAT_MODES);
+
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, TSK_STAT_BRANCH, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_UNSUPPORTED_STAT_MODE);
+
+    ret = tsk_treeseq_r2(&ts, 0, sample_set_sizes, sample_sets, 0, NULL, 0, NULL, 0,
+        &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_BAD_STATE_DIMS);
+
+    sample_set_sizes[0] = 0;
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_EMPTY_SAMPLE_SET);
+    sample_set_sizes[0] = ts.num_samples;
+
+    sample_sets[1] = 10;
+    ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL, 0,
+        NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_NODE_OUT_OF_BOUNDS);
+    sample_sets[1] = 1;
+
+    tsk_treeseq_free(&ts);
+}
+
+static void
 test_simplest_divergence_matrix(void)
 {
     const char *nodes = "1  0   0\n"
@@ -2247,6 +2679,14 @@ main(int argc, char **argv)
         { "test_caterpillar_tree_ld", test_caterpillar_tree_ld },
         { "test_ld_multi_mutations", test_ld_multi_mutations },
         { "test_ld_silent_mutations", test_ld_silent_mutations },
+
+        { "test_paper_ex_two_site", test_paper_ex_two_site },
+        { "test_two_site_correlated_multiallelic",
+            test_two_site_correlated_multiallelic },
+        { "test_two_site_uncorrelated_multiallelic",
+            test_two_site_uncorrelated_multiallelic },
+        { "test_two_site_backmutation", test_two_site_backmutation },
+        { "test_two_locus_stat_input_errors", test_two_locus_stat_input_errors },
 
         { "test_simplest_divergence_matrix", test_simplest_divergence_matrix },
         { "test_simplest_divergence_matrix_windows",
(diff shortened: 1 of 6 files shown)


c/tests/test_stats.c
```diff
@@ -1980,6 +1980,324 @@ test_ld_silent_mutations(void)
...
+    ret = tsk_treeseq_pi2(&ts, num_sample_sets, sample_set_sizes, sample_sets, 0, NULL,
+        0, NULL, 0, &result_size, &result);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_EQUAL(result_size, 3);
+    assert_arrays_almost_equal(result_size, result, truth_pi2);
+    tsk_safe_free(result);
+
+    tsk_treeseq_free(&ts);
+}
+
```
lkirk, line 2174: Here's the tree for the uncorrelated test
image

c/tskit/core.c
```diff
@@ -794,8 +794,8 @@ tsk_search_sorted(const double *restrict array, tsk_size_t size, double value)
             upper = mid;
         }
     }
-    offset = (int64_t)(array[lower] < value);
-    return (tsk_size_t)(lower + offset);
+    offset = (int64_t) (array[lower] < value);
+    return (tsk_size_t) (lower + offset);
```
lkirk, line 798: I'm not sure why clang format is adding this space, but I'm fairly certain that my editor is picking up the clang-format config.

c/tskit/trees.c
```diff
@@ -2076,6 +2076,32 @@ typedef struct {
     const tsk_id_t *index_tuples;
 } indexed_weight_stat_params_t;
```
lkirk, line 2078: I pulled this function out because it's repeated in `tsk_treeseq_two_locus_general_stat`

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+        num_windows, windows, options, result);
+out:
+    tsk_safe_free(weights);
+    return ret;
+}
+
+/***********************************
+ * Two Locus Statistics
+ ***********************************/
+
```
lkirk, line 2144: Note that the pass by ref/value is a bit different from `get_allele_weights`. My version does not have any allocation and modifies the passed array directly. I couldn't quite figure out why `get_allele_weights` was designed the way it was.

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+            num_alleles_cumsum += self->site_mutations_length[site_id] + 1;
+        }
+    }
+
+    ret = get_all_mutation_samples(
+        self, num_sample_chunks, site_offsets, num_alleles, &allele_samples);
+    if (ret != 0) {
+        goto out;
+    }
+
```
lkirk, line 2612: The result size will change when we add windows.

petrelharp:
For easy reference on github: see #432 and #1900.

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    tsk_bit_array_t *ss_row; // ss_ prefix refers to a sample set
+    tsk_bit_array_t ss_A_samples[num_sample_chunks], ss_B_samples[num_sample_chunks],
+        ss_AB_samples[num_sample_chunks], AB_samples[num_sample_chunks];
+    uint8_t polarised_val = polarised ? 1 : 0;
+
+    double weights[3 * state_dim];
+    double norm[state_dim];
+    double *hap_weight_row;
+    tsk_size_t row_len = num_alleles[site_b] * state_dim;
+    // TODO: is this stack allocation dangerous??
```
lkirk, line 2401: Beyond using stack memory like this, it seems as though windows doesn't like my use of VLAs. I'm not sure if this is an inherent issue with MSVC or an issue with the c standard definition in the build. I see `ignoring unknown option '-std=c99'` in the build logs here: https://github.com/tskit-dev/tskit/actions/runs/5592422530/job/15156184424?pr=2805#step:12:35
It does seem as though MVSC is compliant up until C17, so I'd expect there to be support for VLAs, but maybe I'm not understanding the issue.

molpopgen: AFAIK, MVSC supports thru at least C11 but **without** VLAS.  Some Googling seems to agree, at least as of 2021: https://www.reddit.com/r/C_Programming/comments/kzskyi/why_cant_use_a_variable_to_declare_an_arrays_size/

molpopgen: Another thing to consider: VLAs are legal, but rarely used in C.  They are **not** valid C++ code, and tskit may be compiled with a C++ compiler. Doing so is probably rare in practice, but we don't want to prevent it.

lkirk: Ah, good to know. Thanks @molpopgen I think I'll have to rework this a bit to remove the reliance on VLAs. Since it's in a pretty hot loop, I was thinking about pre-allocating some memory and passing it in.

molpopgen: I haven't looked in detail at your changes, but we can certainly start simple and run `perf` once tests are passing, etc., to see where improvements are likely.

c/tskit/trees.c
```diff
@@ -2922,6 +3558,275 @@ tsk_treeseq_Y1(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
...
+}
+
+static int
+r2_summary_func(tsk_size_t state_dim, const double *state,
+    tsk_size_t TSK_UNUSED(result_dim), double *result, void *params)
+{
+    sample_count_stat_params_t args = *(sample_count_stat_params_t *) params;
+    double n;
+    const double *state_row;
+    for (tsk_size_t j = 0; j < state_dim; j++) {
```
petrelharp, line 3640:
```suggestion
    tsk_bug_assert(result_dim == state_dim);
    for (tsk_size_t j = 0; j < state_dim; j++) {
```

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    }
+
+    // A future improvement could get a union of all sample sets
+    // instead of all samples
+    get_all_samples_bits(all_samples_bits, num_samples, num_sample_chunks);
+
+    tsk_bit_array_t *allele_samples_row;
+    tsk_size_t num_alleles_cumsum = 0;
+    for (tsk_size_t t = 0; t < self->num_trees; t++) {
+        for (tsk_size_t s = 0; s < self->tree_sites_length[t]; s++) {
```
petrelharp, line 2595: This could just loop over the sites - the sites in the table are in the same order as encountered when iterating over (trees; sites within trees).

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    }
+
+    ret = get_all_mutation_samples(
+        self, num_sample_chunks, site_offsets, num_alleles, &allele_samples);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // Number of pairs w/ replacement (sites)
+    *result_size = (num_sites * (1 + num_sites)) >> (tsk_size_t) 1;
```
petrelharp, line 2614: Is there a good reason to divide by two by bitshifting? It's more readable IMO to write `/ 2`.

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    for (tsk_size_t site_a = 0; site_a < num_sites; site_a++) {
+        for (tsk_size_t site_b = inner; site_b < num_sites; site_b++) {
+            ret = compute_general_two_site_stat_result(site_a, site_offsets[site_a],
+                site_b, site_offsets[site_b], num_sample_chunks, num_alleles,
+                allele_samples, state_dim, sample_bits, result_dim, f, f_params, norm_f,
+                polarised, &((*result)[result_offset]));
+            if (ret != 0) {
+                goto out;
+            }
+            result_offset += state_dim;
```
petrelharp, line 2645: This should be `result_dim`, no? (as written, `state_dim` always equals `result_dim`, but doing so doesn't make things simpler, I don't think?)

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+            }
+        }
+    }
+}
+
+typedef int norm_func_t(tsk_size_t state_dim, const double *hap_weights, tsk_size_t n_a,
+    tsk_size_t n_b, double *result, void *params);
+
+// TODO: should site_a/b be an id_t?
+static int
```
petrelharp, line 2381: This function doesn't use the entire `num_alleles` array, just those for `site_a` and `site_b`; wouldn't it be better for clarity to just pass in those? (maybe this is true for `state` also?)

petrelharp:
I'm working through this. You've done an impressive amount! There are some very interesting things, e.g. the encoding of descendant samples as bitstrings. 

- IIUC (and I might not), the `weights` are always effectively 0/1, or at least it's only using whether a particular weight is 0 or nonzero, right? This makes it a bit misleading to have the weights being `double`, as we don't use the values, and means that these aren't "general stats" but rather "sample set stats". Have I got this right? If so and if the generic function is exposed, it should probably be `tsk_treeseq_two_locus_sample_set_stat` not `general_stat`.

- It would help me to have the precise definitions of these seven statistics written down somewhere (indeed, this could be a draft for the docstrings). In particular, 

- Perhaps we're being too clever in only keeping track of `AB`, `Ab`, and `aB`? If we also kept track of `ab` and then computed `n = AB + Ab + aB + ab` for each site instead of using the total sample size passed in through `args`, then we would be able to handle missing data, something that we'd like to be able to do.

- I see you're filling out the entire (num mutations) x (num samples) array first then using it to compute each pairwise stat. Thanks to the clever binary representation, this isn't terrible, memory-wise. I suppose you thought about taking the other approach of iterating over pairs of trees - could you provide some background on your thinking on that? This is a speed-memory tradeoff; how much memory would it take to do this for a reasonably large example? How about in very large examples where we don't want the entire LD matrix, but only, say, the diagonal?

- Nice job turning up the references on how to compute multiallelic LD. However, do we really need a special `normalization`, which is adding some complexity to the code? For instance, couldn't `r2` be computed with no normalization if we just modify `r2_summary_func` to multiply the result by `pAB`? I am probably missing something, though: in particular, the "total" normalization scheme doesn't fit into this. However, I'm not convinced that "total" normalization (i.e., averaging over all haplotypes) is the way to go - has anyone suggested doing things this way? What's the rationale here?

again, nice work!

lkirk:
@petrelharp thank you for your review, I appreciate all of the comments. I'll address one at a time below:

> IIUC (and I might not), the weights are always effectively 0/1, or at least it's only using whether a particular weight is 0 or nonzero, right? This makes it a bit misleading to have the weights being double, as we don't use the values, and means that these aren't "general stats" but rather "sample set stats". Have I got this right? If so and if the generic function is exposed, it should probably be tsk_treeseq_two_locus_sample_set_stat not general_stat.

You're right, weights are always 1 or 0. I agree, this is more of a sample set stat, I didn't really think about the distinction until you brought this up, I can rename to `tsk_treeseq_two_locus_sample_set_stat`. I designed the API to be general enough to accept weights if we wanted to specify them in some other statistic. That said, we'd need to alter the implementation in `compute_general_two_site_stat_result` to properly utilize the weights. So, as it stands right now, the implementation for weights is half done and misleading in terms of data types. I'm happy to finish the job of implementing weights in a way that they'd be useful later on (especially since `compute_general_two_site_stat_result` needs some reworking anyways -- per my discussion with molpopgen). I'm also happy to simplify the API and remove the weights as doubles.

> It would help me to have the precise definitions of these seven statistics written down somewhere (indeed, this could be a draft for the docstrings). In particular,

Page 6 of my [design document](https://raw.githubusercontent.com/lkirk/ts-two-locus-proto/e26017c/design-document/main.pdf) defines these statistics. Is there another way I should be defining these to be more clear?

> Perhaps we're being too clever in only keeping track of AB, Ab, and aB? If we also kept track of ab and then computed n = AB + Ab + aB + ab for each site instead of using the total sample size passed in through args, then we would be able to handle missing data, something that we'd like to be able to do.

I see, would this be a scenario where a sample was deleted from a single tree? It's good to hear about these types of edge cases, I didn't think it was possible to do something like that.

> I see you're filling out the entire (num mutations) x (num samples) array first then using it to compute each pairwise stat. Thanks to the clever binary representation, this isn't terrible, memory-wise. I suppose you thought about taking the other approach of iterating over pairs of trees - could you provide some background on your thinking on that? This is a speed-memory tradeoff; how much memory would it take to do this for a reasonably large example? How about in very large examples where we don't want the entire LD matrix, but only, say, the diagonal?

Yes, I'm gathering all samples with a specific allele state before computing the statistics. I've organized things this way so that we only have to do one pass over the tree sequence before computing statistics. This also means one tree traversal per tree in the tree sequence, so the computational complexity should be roughly the same as single-site statistics (O(n + p (log n)^2)). I ran this code on the entire chromosome 1 for the SGDP dataset and it the alleles X samples matrix took ~175MB of memory. That's a dataset with 554 samples and 1,198,947 mutations. In total, ~750MB of memory was used before proceeding to the computation of the stats matrix. In contrast, the amount of data needed to store the upper triangle of the statistics for that dataset is about 5.7TB (1198947 * (1198947 + 1) / 2 * 8 / 1e12) -- This number has me second guessing my math, but I am referring to 718 billion pairwise statistics, each represented by a double.

That said, I think that once windows are implemented, I'd only compute the matrix for the data specified in the window, which would drop the size even more.

Sorry if that was long winded. In other words, if we want to know how much memory the allele matrix occupies, we can do the following calculation (RoundUp is a mathematica function):

```
(RoundUp[num_samples / 32]) * (num_mutations + num_sites)  / 1e6 * 4
```

I'm still not sure what we consider very large, but in my opinion, the size of this intermediate data pales in comparison to the size of the stats output.

> Nice job turning up the references on how to compute multiallelic LD. However, do we really need a special normalization, which is adding some complexity to the code? For instance, couldn't r2 be computed with no normalization if we just modify r2_summary_func to multiply the result by pAB? I am probably missing something, though: in particular, the "total" normalization scheme doesn't fit into this. However, I'm not convinced that "total" normalization (i.e., averaging over all haplotypes) is the way to go - has anyone suggested doing things this way? What's the rationale here?

We tried a variety of different norming methods for obtaining reasonable results from the various two-locus statistics. One indicator of correctness for using "total" for non-ratio statistics is that the sum of non-polarized D adds up to 0. If we norm by the haplotype frequency, and compute non-polarized D, we end up with a non-zero value. In this manner, we went through and validated our norming methods with the expected extremes of each two-locus statistic. I don't think we found any papers suggesting to normalize things this way, but it's the only method that produces results that appear correct for our handful of test cases. I can provide a more detailed notebook that outlines these results a bit better if that would help.

c/tskit/trees.c
lkirk: I didn't realize until recently that you could do truncated division with unsigned integers. It seems that this works with c99 on clang and gcc, I can change this to division so it's more readable. Here's a simple [playground](https://cplayground.com/?p=hamster-porcupine-ape) session showing my findings.

c/tskit/trees.c
lkirk: Yes, I think you're right. I can pass in pointers to the num_alleles and state array instead of the beginnings + offsets. That would definitely simplify things here.

c/tskit/trees.c
lkirk: Oh, this is good to know, I will make this change as well

c/tskit/trees.c
lkirk: Yes, you're right. I will change this.

c/tskit/trees.c
lkirk: Should all of the summary functions be making this assertion? I was a little confused about result vs state dims when looking over the codebase. My understanding is that these might not be the same if we're computing a branch vs site statistic.

lkirk (commented):
I will batch all of the suggested changes in the next few days. Thanks again for all of the feedback.

c/tskit/trees.c
petrelharp: See https://tskit.dev/tskit/docs/latest/data-model.html#valid-tree-sequence-requirements (and https://tskit.dev/tskit/docs/latest/data-model.html#table-definitions ) for ordering requirements on the tables.

petrelharp:
Excellent, thanks! Let's see:

> I'm happy to finish the job of implementing weights in a way that they'd be useful later on (especially since compute_general_two_site_stat_result needs some reworking anyways -- per my discussion with molpopgen). I'm also happy to simplify the API and remove the weights as doubles.

Well, do we have a use case that uses non-0/1 weights? If not, then we should not do the general case.

> > It would help me to have the precise definitions of these seven statistics written down somewhere (indeed, this could be a draft for the docstrings).
> 
> Page 6 of my [design document](https://raw.githubusercontent.com/lkirk/ts-two-locus-proto/e26017c/design-document/main.pdf) defines these statistics. Is there another way I should be defining these to be more clear?

Right, I saw that, and that is helpful, but it still leaves questions - it's not precise enough for someone to compute exactly the same thing from scratch (especially, someone who doesn't have an idea beforehand of what is supposed to be computed). For instance, how does one deal with multiallelic sites, especially for the statistics with ratios (eg. r)? For an example, see the [docstring for Y3](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.Y3): it says "The average density of sites at which a differs from b and c, per unit of chromosome length.". This is (or tries to be :) clear enough that it's easy and obvious how to compute the statistic in a naive way. (The definitions for your stats won't be so concise, but oh well.) 

> I see, would this be a scenario where a sample was deleted from a single tree?

This is how we represent missing data: https://tskit.dev/tskit/docs/latest/data-model.html#sec-data-model-missing-data i.e., a sample that wasn't genotyped for a stretch of chromosome (we haven't dealt with this in the usual stats API yet, but would like to).

 
> Sorry if that was long winded. In other words, if we want to know how much memory the allele matrix occupies, we can do the following calculation (RoundUp is a mathematica function):
> 
> (RoundUp[num_samples / 32]) * (num_mutations + num_sites)  / 1e6 * 4

Oh, not long-winded at all, detail is good. =)  So, I guess the downside here is that computing the LD matrix for a fixed number of sites scales (in memory) with the number of samples. Hm, so, the internal state is comparable to the (full) output when the number of samples is comparable to the number of sites (in a window). Good to know.

To be clear, I think the approach here is very clever, and I don't think you should change it; we just want to be clear about the implications. It wouldn't scale to Biobank-scale data (~1M individuals), but that's okay; it's probably faster for smaller datasets than other approaches, and moving forward we could switch off to another algorithm for big computations.

This does make me think about something else we might want in the API, though: in practice, won't people want to specify a specific set of SNPs to compute this with? For instance, just removing the singletons would reduce the size of the output substantially. Perhaps we want to pass in a "site mask" that says which sites to use?

> ... normalization... I can provide a more detailed notebook that outlines these results a bit better if that would help.

That would help - I am still fuzzy on exactly how the normalization work (at least, I need the definition?).

c/tskit/trees.c
```diff
@@ -489,7 +489,7 @@ tsk_treeseq_init(
 
     if (tsk_treeseq_get_time_units_length(self) == strlen(TSK_TIME_UNITS_UNCALIBRATED)
         && !strncmp(tsk_treeseq_get_time_units(self), TSK_TIME_UNITS_UNCALIBRATED,
-               strlen(TSK_TIME_UNITS_UNCALIBRATED))) {
```
jeromekelleher, line 492: Looks like you've got the wrong version of clang-format here. See [here](https://tskit.dev/tskit/docs/stable/development.html?highlight=clang%20format#sec-development-c-requirements)

(There's a long paper trail of us complaining that clang-format has such terrible compatibility :vomiting_face: )

c/tskit/trees.c
```diff
@@ -2076,6 +2076,32 @@ typedef struct {
...
 } indexed_weight_stat_params_t;
 
+static int
+init_sample_weights(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets, double **weights)
+{
+    int ret = 0;
+    tsk_id_t u, sample_index;
+    tsk_size_t j, k, l;
+    double *weight_row;
```
jeromekelleher, line 2086: style nit: we'd usually put a blank line between the declarations and the function code

c/tskit/trees.c
```diff
@@ -2103,22 +2126,635 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    const tsk_size_t *num_alleles, const tsk_bit_array_t *state, tsk_size_t state_dim,
+    tsk_bit_array_t *sample_sets, tsk_size_t result_dim, general_stat_func_t *f,
+    sample_count_stat_params_t *f_params, norm_func_t *norm_f, bool polarised,
+    double *result)
+{
+    int ret = 0;
+    const tsk_bit_array_t *A_samples, *B_samples;
+    tsk_size_t w_A = 0, w_B = 0, w_AB = 0;
+    tsk_bit_array_t *ss_row; // ss_ prefix refers to a sample set
+    tsk_bit_array_t ss_A_samples[num_sample_chunks], ss_B_samples[num_sample_chunks],
```
jeromekelleher, line 2393: I don't think we can allocate on the stack like this, we'll eventually overflow for large sample sizes, and it's very hard to tell in advance when it's going to happen. It's just not worth the trouble.

c/tskit/trees.c
jeromekelleher: My 10c is that allocing anything on the  stack that isn't a fixed, small size just isn't worth the hassle. It's not portable, unpredictable when it fails and doesn't give much/any perf advantages.

c/tskit/trees.c
jeromekelleher: +1 on this - any compiler worth it's salt will replace with a bitshift if there's a literal ``integer_value / 2`` in there.

jeromekelleher (commented):
This is great, well done @lkirk! 

I've made a light review pass and noted a few implementation details around the use of bitsets and memory allocations that I think will need to be changed. I've not got deep into all the details, I'll try and get my head around it a bit better in the next pass.

c/tskit/trees.c
jeromekelleher: (You might notice that we do a lot of allocing from the stack in the test suite - but we can get away with a lot of stuff there if it makes the code easier to write and read because the inputs are fixed)

lkirk (commented):
I've refactored `tsk_bit_array_t`, taken care of the VLAs, and have addressed most of the comments, which I will mark as done to clean things up.

c/tskit/trees.c
jeromekelleher: Simplest to install 6 I think, we're not going to change this in the short term. Anyway, whatever version of clang-format we'd settle on would probably be slightly different to the version you're using and you'd need to change anyway.

c/tskit/trees.c
jeromekelleher: Yeah, we do keep all the declarations at the start. The original reason was to support older C compilers, but now we're using C99 that doesn't apply any more. Now it's mostly for consistency, but there are still two good reasons for doing it this way:

1. It helps keep track of how big and complicated your function is (if you're declaring too many variables, the function is too big)
2. It helps keep a consistent pattern for memory management, and safety. All pointers are initialised to NULL before any code is run, and everything is ``safe_free'd`` at the end of the function.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,651 @@ typedef struct {
...
+        site = &tree_sites[s];
+
+        tsk_bit_array_get_row(&mut_samples, mut_samples_offset, &mut_samples_row);
+        mut_samples_offset += self->site_mutations_length[site->id];
+
+        tsk_bit_array_get_row(allele_samples, site_offsets[site->id], &out_row);
+        get_allele_samples(site, &mut_samples_row, &out_row, &(*num_alleles)[site->id]);
+    }
+out:
+    tsk_safe_free(mut_samples.data);
```
jeromekelleher, line 2320: Better to call ``tsk_bit_array_free(&mut_samples)``, so that the implementation can change

jeromekelleher (commented):
A few minor comments on memory safety. Haven't had a chance to go through fully.

c/tskit/trees.c
lkirk: Makes sense, I've gone ahead moved all declarations to the start to be consistent with this style.

c/tskit/core.c
```diff
@@ -1177,3 +1177,97 @@ tsk_avl_tree_int_ordered_nodes(const tsk_avl_tree_int_t *self, tsk_avl_node_int_
...
+bool
+tsk_bit_array_contains_id(const tsk_bit_array_t *self, const tsk_id_t node)
+{
+    if (node == TSK_NULL) {
+        return false;
+    }
+    return tsk_bit_array_contains(self, (tsk_bit_array_value_t) node);
+}
+
+void
```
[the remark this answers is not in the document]
lkirk: I've changed this function to return a `tsk_size_t` value, but I've kept the functions that modify a pointer in place, even though they don't raise an error. I think that it's best to avoid copies of the bit arrays.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,651 @@ typedef struct {
...
+
+    tsk_bug_assert(state != NULL);
+    alleles[0] = site->ancestral_state;
+    allele_lengths[0] = site->ancestral_state_length;
+    *num_alleles = 1;
+
+    for (mutation_index = 0; mutation_index < site->mutations_length; mutation_index++) {
+        mutation = site->mutations[mutation_index];
+        /* Compute the allele index for this derived state value. */
+        allele = 0;
```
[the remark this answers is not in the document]
lkirk: Indeed it would be. I copied this from `get_allele_weights` FWIW, could fix this in that function too.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,651 @@ typedef struct {
...
+        site = &tree_sites[s];
+        for (tsk_size_t m = 0; m < site->mutations_length; m++) {
+            num_mutations++;
+            if (site->mutations[m].node > top_mut_node) {
+                top_mut_node = site->mutations[m].node;
+            }
+        }
+    }
+
+    tsk_bit_array_init(&mut_samples, num_samples, num_mutations);
```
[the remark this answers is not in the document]
lkirk: I see, freeing an uninitialized pointer is undefined behavior. I've added a memset to the initialization function. However, it does seem as though `tsk_safe_free` handles this by setting the pointer to `NULL` first. Since we're initializing these arrays, I have used `free` in the `tsk_bit_array_free` function instead of `tsk_safe_free`.

c/tskit/trees.c
jeromekelleher: Thanks :+1:

jeromekelleher (commented):
See point about using ``tsk_safe_free`` - safety first when it comes to handling memory!

c/tskit/core.c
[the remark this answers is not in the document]
jeromekelleher: Agreed - I'd just use return values for simple scalars

c/tskit/trees.c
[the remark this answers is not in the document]
jeromekelleher: See point in just-made comment about additional safety concerns

c/tskit/trees.c
jeromekelleher: I'm not sure either - let's discuss tomorrow!

c/tskit/trees.c
```diff
@@ -2076,19 +2076,661 @@ typedef struct {
...
+        goto out;
+    }
+    ret = tsk_bit_array_init(
+        &node_paths, tsk_treeseq_get_num_nodes(self), num_mutations);
+    if (ret != 0) {
+        goto out;
+    }
+
+    // Size of stack is derived from tsk_tree_get_size_bound
+    stack = tsk_malloc((1 + self->num_samples + num_edges) * sizeof(*stack));
```
jeromekelleher, line 2270: This is actually a substantial over-estimate of the  max stack size - the simplest thing to do there is to the ``ts.num_nodes`` as the maximum stack size (when we don't know anything about the individual trees that it'll be used on)

lkirk:
Sorry I'm a bit late in responding to your comments.

> Well, do we have a use case that uses non-0/1 weights? If not, then we should not do the general case.

I do remember discussing something with @apragsdale, but I don't think we'd discussed anything concrete. Let's discuss when we meet, but I'm leaning towards simplifying things to counts-only for now, then expanding once we have a clear need.

> Right, I saw that, and that is helpful, but it still leaves questions - it's not precise enough for someone to compute exactly the same thing from scratch (especially, someone who doesn't have an idea beforehand of what is supposed to be computed).

I've created a small demo notebook that might help illustrate how exactly we're doing the normalization. It can be found [here](https://github.com/lkirk/ts-two-locus-proto/blob/a6b13c4/notebooks/DemonstrateCorrectness.ipynb). In addition, please see my explanation at the bottom. Hopefully this helps with understanding.

> This is how we represent missing data: https://tskit.dev/tskit/docs/latest/data-model.html#sec-data-model-missing-data i.e., a sample that wasn't genotyped for a stretch of chromosome (we haven't dealt with this in the usual stats API yet, but would like to).

The solution you've proposed seems straightforward, I'm not sure how it would tie into normalization, I'd have to do some testing to see what makes sense.

> Oh, not long-winded at all, detail is good. =) So, I guess the downside here is that computing the LD matrix for a fixed number of sites scales (in memory) with the number of samples. 

> Hm, so, the internal state is comparable to the (full) output when the number of samples is comparable to the number of sites (in a window). Good to know.

> To be clear, I think the approach here is very clever, and I don't think you should change it; we just want to be clear about the implications. It wouldn't scale to Biobank-scale data (~1M individuals), but that's okay; it's probably faster for smaller datasets than other approaches, and moving forward we could switch off to another algorithm for big computations.

I'm not exactly sure what you mean by the "full" output here. If we have a set with 1M samples/1M sites/1M mutations (meaning we're dealing w/ biallelic data)... We expect to use 250GB to store the states, but we'd still need 4TB to store the actual statistics. In any case, I think a potential solution to scaling this is in the windowing/masking. If you could compute these stats in chunks, parallelized on a cluster, that'd probably be somewhat ideal. Especially given that windowed computation wouldn't see much overhead because you could limit many of the operations to the bounds of the window/mask. An iterative algo could also work but would be much slower (I think).

> This does make me think about something else we might want in the API, though: in practice, won't people want to specify a specific set of SNPs to compute this with? For instance, just removing the singletons would reduce the size of the output substantially. Perhaps we want to pass in a "site mask" that says which sites to use?

I think that's what I was getting at with the windows part of the design (hinted in some of the code, but not actually implemented). However, the one issue with providing a window is that we'd need to return the sites back to the user once we've intersected their coordinates with the window coordinates. Otherwise the user wouldn't know which sites are making up the results matrix. (And I think a single window makes sense here, ideally we can chop out an NxM matrix from the entire results matrix, reducing the memory/computational requirements to a manageable size).

So, what you're saying about a site mask sounds more like what we want to do. By using a sites mask, the user knows what sites are going into the matrix. The problem comes down to the api, how much work do we reasonably want to put onto the user when it comes to intersecting/storing/manipulating sites. And, how do we create a sites mask that is reasonably efficient in terms of memory? (I suppose we could use bit masks to store sites in a mask as well....).

> That would help - I am still fuzzy on exactly how the normalization work (at least, I need the definition?).

The notebook is linked above, but I'll also explain with actual words:

"Total" normalization: multiply the result of each comparison between alleles by the proportion of alleles.

For example, if we are comparing an A locus with 3 alleles and a B locus with 3 alleles, we'd compute each statistic for A1-B1/A1-B2/A1-B3/A2-B1/... and we'd multiply them by 1/9

"Hap Weighted" normalization: We multiply the result of each comparison between alleles by the proportion of haplotypes for each comparison.

Using the same A/B locus as above, let's say that we are comparing 10 samples. If A1-B1 have no counts, we will multiply the stat result by 0. If A1-B2 have 5 counts, then we'll multiply the stat result by 1/2 (5/10).

lkirk:
We met over zoom and decided that the best course of action is to wrap this up (it's getting a bit messy). In order to wrap things up, we're prioritizing the steps necessary for merge. I'll list out what I'll be doing when I get back from vacation:

1. Rebase with main so that the circleci:build-32 test passes (see guide in stdpopsim)
2. Finish up test coverage, covering bit arrays (I think the line that hits the nonbinary samples case will be eliminated when I adopt api functions).
3. Use api functions for iteration + built in preorder
4. Address memory safety issues
5. Final bit array counting algo (the portable one -- it offers significant performance gains)
6. Capture the remaining discussion points/concerns in github issues.

c/tskit/trees.c
petrelharp: Well, I wasn't trying to minimize the number of variables, I was trying to make it more clear what the function uses?

lkirk:
> > "Total" normalization: multiply the result of each comparison between alleles by the proportion of alleles.
> > For example, if we are comparing an A locus with 3 alleles and a B locus with 3 alleles, we'd compute each statistic for A1-B1/A1-B2/A1-B3/A2-B1/... and we'd multiply them by 1/9
> > "Hap Weighted" normalization: We multiply the result of each comparison between alleles by the proportion of haplotypes for each comparison.
> 
> Ah, that makes sense. And, the notebook is helpful (and, a great start for test cases!).
> 
> And, okay. My feeling is that "hap weighted" or maybe "allele freq weighted" makes much more sense than "total", since "total" can be strongly affected by rare alleles (eg sequencing error). And in this case, the weighting is just a function of the frequencies, so can be computed within `f` instead of outside of it, which would remove the need for doing `norm_f` separately. But, normalization isn't adding a lot to the complexity of the code, so we can discuss this later. (Maybe you all have a use case where we need 'total' normalization?)
> 
> Want me to open an issue?

Interesting, I hadn't considered the point about rare variants. I think the main draw to using "total" is the fact that it seems to be correct for D and other statistics (demo'd a bit at the end of that notebook). In any case, please feel free to open an issue, I agree things could be simplified, I think @apragsdale might have an opinion here as well. I'll get to writing other issues when I return.

c/tskit/trees.c
lkirk: I see what you mean, so I suppose removing the offsets from within this function would simplify things. I can refactor this when I make another pass over this stuff.

lkirk: Now refactored to leave the offsets in the calling function

jeromekelleher (commented):
Some minor comments, which I didn't get a chance to finish. More later!

c/tskit/trees.c
```diff
@@ -2076,19 +2076,567 @@ typedef struct {
...
+    }
+    *out_num_alleles = num_alleles;
+out:
+    tsk_safe_free(alleles);
+    tsk_safe_free(allele_lengths);
+    return ret;
+}
+
+static int
+get_mutation_samples(const tsk_tree_t *self, const tsk_size_t *site_offsets,
```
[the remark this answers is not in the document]
lkirk: I think it'd be tough to use this function in isolation, I would consider it to be an internal function until decided otherwise. In that vain, I've changed this parameter to `tree` instead of `self`.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,567 @@ typedef struct {
...
+    tsk_bit_array_t *allele_samples, tsk_size_t **num_alleles)
+{
+    int ret = 0;
+
+    const tsk_treeseq_t *restrict ts = self->tree_sequence;
+
+    const tsk_size_t num_samples = tsk_treeseq_get_num_samples(ts);
+    const tsk_size_t num_tree_sites = ts->tree_sites_length[self->index];
```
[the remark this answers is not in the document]
lkirk: Good point, I eliminated that variable entirely.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,567 @@ typedef struct {
...
+    const tsk_flags_t *restrict flags = ts->tables->nodes.flags;
+    const tsk_id_t *restrict mut_nodes = ts->tables->mutations.node;
+
+    const tsk_site_t *restrict site;
+
+    tsk_bit_array_t mut_samples, mut_samples_row, out_row;
+    tsk_size_t num_mutations, mut_samples_offset, mut_offset;
+    tsk_size_t s, m, n;
+
+    tsk_memset(&mut_samples, 0, sizeof(mut_samples));
```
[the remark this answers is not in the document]
lkirk: Ah, I think I overcorrected a bit, I moved them before any `goto` call, but after all declarations.

c/tskit/trees.c
```diff
@@ -2076,19 +2076,567 @@ typedef struct {
...
+    tsk_id_t *nodes = tsk_malloc(tsk_tree_get_size_bound(self) * sizeof(*nodes));
+
+    if (nodes == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+
+    // Get the number of mutations
+    num_mutations = 0;
+    for (s = 0; s < num_tree_sites; s++) {
```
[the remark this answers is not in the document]
lkirk: Fixed

c/tskit/trees.c
```diff
@@ -2124,6 +2124,580 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    tsk_safe_free(alleles);
+    tsk_safe_free(allele_lengths);
+    return ret;
+}
+
+static int
+get_mutation_samples(const tsk_tree_t *tree, const tsk_size_t *site_offsets,
+    tsk_bit_array_t *allele_samples, tsk_size_t **num_alleles)
+{
+    int ret = 0;
```
jeromekelleher, line 2211: Nitpick, but not sure this whitespace within the declarations section helps understanding? Usually all declarations are done in one chunk, which makes it easier to see where the code begins (and you start thinking about memory safety and the potential state of variables after error-gotos)

c/tskit/trees.c
```diff
@@ -2124,6 +2124,580 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    // Current position minus num ancestral alleles to this point, if there are any sites
+    mut_offset = tree->sites_length
+                     ? (site_offsets[tree_sites[0].id] - (tsk_size_t) tree_sites[0].id)
+                     : 0;
+
+    // Traverse down the tree, recording all samples below each mutation.
+    // We currently perform one preorder traversal per mutation, but there
+    // might be more clever ways to do this. The preorder traversals begin at
+    // the focal mutation's node.
+    for (m = 0; m < num_mutations; m++) {
```
jeromekelleher, line 2257: I don't get why you need to use the arrays here: why not use the site and mutation objects within the tree struct?
```
for (s = 0; s < tree->sites_length; s++) {
     site = tree->sites[s];
     for (m = 0; m < site.mutations_length; m++) {
           node = site.mutations[m].node;
           // do traversal rooted at node. 
    }
}

File: c/tskit/trees.c [hunk 26]
```diff
@@ -2124,6 +2124,580 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+}
+
+static int
+get_all_mutation_samples(const tsk_treeseq_t *self, const tsk_size_t *site_offsets,
+    tsk_size_t *num_alleles, tsk_bit_array_t *allele_samples)
+{
+    int ret = 0;
+    tsk_tree_t tree;
+
+    ret = tsk_tree_init(&tree, self, TSK_TS_INIT_BUILD_INDEXES | TSK_NO_SAMPLE_COUNTS);
```
jeromekelleher, line 2452: ``TSK_TS_INIT_BUILD_INDEXES``  isn't a valid option for tree_init.

jeromekelleher: To me the ``get_mutation_samples`` function is more complicated than it needs to be, and it would be clearer to merge it into this function (as it does very little at the moment)

c/tskit/trees.c
```diff
@@ -2124,6 +2124,580 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    tsk_bit_array_t site_a_state, site_b_state;
+    tsk_size_t site_id, num_alleles_cumsum, inner, result_offset;
+    tsk_size_t site_a, site_b;
+    bool polarised = false;
+
+    const tsk_size_t num_sites = self->tables->sites.num_rows;
+    const tsk_size_t num_samples = self->num_samples;
+    const tsk_size_t max_alleles = self->tables->mutations.num_rows + num_sites;
+    tsk_size_t *restrict site_offsets = tsk_malloc(num_sites * sizeof(*site_offsets));
+    tsk_size_t *restrict num_alleles = tsk_malloc(num_sites * sizeof(*num_alleles));
```
jeromekelleher, line 2489: There's no real point in making this a restrict pointer in this function, we're not doing any tight loops on it, and just passing it on to other functions. Declaring as a regular pointer means we can call ``tsk_safe_free`` and reduce the code a bit.

c/tskit/trees.c
```diff
@@ -2124,6 +2124,580 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    // A future improvement could get a union of all sample sets
+    // instead of all samples
+    get_all_samples_bits(&all_samples_bits, num_samples);
+
+    num_alleles_cumsum = 0;
+    for (site_id = 0; site_id < num_sites; site_id++) {
+        // Initialize the allele_samples with all samples in the ancestral allele
+        tsk_bit_array_get_row(&allele_samples, num_alleles_cumsum, &allele_samples_row);
+        tsk_bit_array_add(&allele_samples_row, &all_samples_bits);
+        // Store the allele offset for each site
```
jeromekelleher, line 2517: Do we need to do this stuff with ``site_offsets``, if we're iterating over the mutations using the inbuilt structs from the tree (as suggested above)?

jeromekelleher (commented):
LGTM. I think the implementation is still overly complicated, but using the in-built mutation tracking in the tree code should simplify quite a bit. 

I think we'll want to evolve the implementation a bit, but I've created some issues to track. We're good to merge after this round of comments, IMO.

c/tskit/core.c
```diff
@@ -1177,3 +1177,101 @@ tsk_avl_tree_int_ordered_nodes(const tsk_avl_tree_int_t *self, tsk_avl_node_int_
     ordered_nodes_traverse(self->head.rlink, 0, out);
     return 0;
 }
+
+// Bit Array implementation. Allows us to store unsigned integers in a compact manner.
+// Currently implemented as an array of 32-bit unsigned integers for ease of counting.
+
+int
+tsk_bit_array_init(tsk_bit_array_t *self, tsk_size_t num_bits, tsk_size_t length)
```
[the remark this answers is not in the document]
lkirk: Yes, you're right, it's still a bit unclear. I responded with some thoughts.

c/tskit/trees.c
lkirk: These offsets were put in place to reduce the amount of times I'd have to compute the offsets and to make sure all of the code is in sync. Now that things have been simplified and I'm not using offsets in 3 places, I think it's best to get rid of the malloc'd array of offsets. We still need offsets to index into the array of allele_samples, but now it's loop counters instead of a whole malloc'd array.

c/tskit/trees.c
lkirk: Fixed.

c/tskit/trees.c
lkirk: I did a refactor of all of this as you've suggested. It seems much simpler. I think with some of the suggestions about the bit array refactor, this code will be simplified even more.

c/tskit/trees.c
lkirk: See my comment about offsets, I've cleaned things up, using the suggested loop.

c/tskit/trees.c
lkirk: You're right, I have a habit of breaking things up with whitespace, but sometimes I can go overboard. I've cleaned things up a bit here.

jeromekelleher (approved):
LGTM - let's squash n merge :tada: 

(Can you squash down to one commit please @lkirk?)

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+
+    for (mutation_index = 0; mutation_index < site->mutations_length; mutation_index++) {
+        mutation = site->mutations[mutation_index];
+        /* Compute the allele index for this derived state value. */
+        for (allele = 0; allele < num_alleles; allele++) {
+            if (mutation.derived_state_length == allele_lengths[allele]
+                && tsk_memcmp(
+                       mutation.derived_state, alleles[allele], allele_lengths[allele])
+                       == 0) {
+                break;
```
jeromekelleher, line 2165: We should be covering this, really. Do we need an example with some extra mutations or something?

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+get_all_samples_bits(tsk_bit_array_t *all_samples, tsk_size_t n)
+{
+    tsk_size_t i;
+    const tsk_bit_array_value_t all = ~((tsk_bit_array_value_t) 0);
+    const tsk_bit_array_value_t remainder_samples = n % TSK_BIT_ARRAY_NUM_BITS;
+
+    all_samples->data[all_samples->size - 1]
+        = remainder_samples ? ~(all << remainder_samples) : all;
+    for (i = 0; i < all_samples->size - 1; i++) {
+        all_samples->data[i] = all;
```
jeromekelleher, line 2247: We should definitely be covering this - need a sample size > 2?

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    j = 0;
+    for (k = 0; k < num_sample_sets; k++) {
+        tsk_bit_array_get_row(sample_sets_bits, k, &bits_row);
+        for (l = 0; l < sample_set_sizes[k]; l++) {
+            u = sample_sets[j];
+            sample_index = self->sample_index_map[u];
+            if (tsk_bit_array_contains(
+                    &bits_row, (tsk_bit_array_value_t) sample_index)) {
+                ret = TSK_ERR_DUPLICATE_SAMPLE;
+                goto out;
```
jeromekelleher, line 2550: We should be covering this

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    sample_count_stat_params_t f_params = { .sample_sets = sample_sets,
+        .num_sample_sets = num_sample_sets,
+        .sample_set_sizes = sample_set_sizes,
+        .set_indexes = set_indexes };
+
+    tsk_memset(&sample_sets_bits, 0, sizeof(sample_sets_bits));
+
+    ret = tsk_treeseq_check_sample_sets(
+        self, num_sample_sets, sample_set_sizes, sample_sets);
+    if (ret != 0) {
```
jeromekelleher, line 2586: And this - pass in some bad samples

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+        goto out;
+    }
+
+    /* If no mode is specified, we default to site mode */
+    if (!(stat_site || stat_branch)) {
+        stat_site = true;
+    }
+    /* It's an error to specify more than one mode */
+    if (stat_site + stat_branch > 1) {
```
jeromekelleher, line 2600: All these input errors should be covered

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+    }
+
+    if (state_dim < 1) {
+        ret = TSK_ERR_BAD_STATE_DIMS;
+        goto out;
+    }
+    if (result_dim < 1) {
+        ret = TSK_ERR_BAD_RESULT_DIMS;
+        goto out;
+    }
```
jeromekelleher, line 2612: Since we're not sure about how the ``windows`` thing is going to work I vote we just ``assert(left_windows == NULL && right_windows == NULL)``; for now, and delete the uncovered code.

c/tskit/trees.c
```diff
@@ -2124,6 +2124,530 @@ tsk_treeseq_sample_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_s
...
+
+    if (stat_site) {
+        // TODO: assert 1 window left/right
+        ret = tsk_treeseq_two_site_count_stat(self, state_dim, &sample_sets_bits,
+            result_dim, f, &f_params, norm_f, left_windows, right_windows, options,
+            result_size, result);
+    } else {
+        // TODO: branch statistics
+        ret = TSK_ERR_GENERIC;
```
jeromekelleher, line 2643: should be BAD_STAT_MODE, and should be covered

jeromekelleher (commented):
I think we need one more update to get test coverage @lkirk - if we merge now these cases will be forgotten and end up never being covered.

Most will be covered by provoking error conditions. The ``test_simplest_divergence_matrix`` function can be a good pattern to follow there for provoking errors.

Let's just remove the code for windows for now. I think the C code, certainly, should have a fixed set of sites provided as input which will simplify things. But let's leave that till next update.

c/tskit/trees.c
lkirk: I added a test with a couple of back-mutations, that covers it

c/tskit/trees.c
lkirk: I added a test with 35 samples, which is >32 samples, which covers it

jeromekelleher:
LGTM, happy to merge when you're satisfied with the code coverage.

lkirk:
I think I've gotten all of the interesting test cases. It's annoying I'm only able to partially hit [this](https://app.codecov.io/gh/tskit-dev/tskit/pull/2805/blob/c/tskit/trees.c#L2189) case and the one above it, but I think I'm satisfied with the outcome. Unfortunately, I'm slightly (2%) under the threshold for passing the coverage check. I think that's mostly to do with the number of out of memory errors that this code can throw. Let me know if you disagree.

lkirk:
Re-squashed and ready

jeromekelleher:
@mergifyio rebase

jeromekelleher:
I'm not sure why mergify doesn't like this, just going to merge manualy.

PR #2807, opened 2023-07-28
Correct doc instances of "raises"

hyanwong:
The correct docstring formatting was wrong in a few places


petrelharp (approved):
LGTM, good catch!

PR #2811, opened 2023-08-03
Fix 32bit CI

benjeffery:
Latest cython breaks compilation of h5py.


jeromekelleher:
Is it time to drop h5py?

benjeffery:
Ok, so after updating the 32bit docker image as it was on 3.7 and that was causing issues, `numba` now won't build as it needs `llvm`. However it needs specifically llvm version 10, which is no longer in the main ubuntu ppa.
On searching I see I should be able to get it from universe: https://pkgs.org/download/llvm-10
Will try that tomorrow.

benjeffery:
Went a bit round the houses here, but this seems to work for now.

jeromekelleher (approved):

Issue #2779, opened 2023-07-07
Improved algorithm for site-mode divergence matrix


jeromekelleher:
The divergence matrix code merged in #2736 gives a reasonably efficient way to compute the pairwise branch-mode divergence among a set of samples, but it's very poor for site mode. This is probably because we're not using the fast MRCA calculations in the internal loop.

The simplest way to do this may be to keep a count of the total number of mutations that are ancestral to each node as we to along the genome, and then we should be able to use the same framework as the branch stat calculation.

This will be complicated by keeping track of windows also though.

Doing this will need an incremental algorithm that can start and stop at a given point efficiently, so we should probably get some C code in for doing this in a generic way first, as dicussed in #2778

jeromekelleher:
Actually, I don't see how we can make this much more efficient than the naive approach currently implemented. My assertion above is wrong:

The core of the current branch mode operation is to do this:
```           
                    w = tree.mrca(u, v)                  
                    tu = ts.nodes_time[w] - ts.nodes_time[u]
                    tv = ts.nodes_time[w] - ts.nodes_time[v]                  
                    d = (tu + tv) * span
                    D[i, j, k] += d
```
for pairs of samples u, v and their indexes j, k.

Above I implied we could do roughly the same thing for the site mode if we kept track of the total number of mutations descending from each node, i.e. replacing ``ts.nodes_time`` with ``num_descending_mutations``. However this doesn't work because the num_descending_mutations for the MRCA isn't just the sum of the num_descending_mutations of the two incoming paths for the target samples u and v, it also includes all mutations for any other children (i.e., if it's a polytomy). So, to use this logic we need to keep track of which child of the MRCA u and v came in from, in which case there's nothing to be gained from using the Schieber Vishkin algorithm for computing MRCAs.

More generally, it's not clear to me that there is much to be gained from using more complex algorithms either. Fundamentally, each mutation defines a partition of the samples into the set of those that descend from the mutation and those that don't (say A and B), and we need to update the divergence matrix by adding one to each cell for the cartesian product of A and B. In principle, we can avoid incrementing-by-one each time if two or mutation mutations share the same sample partition. However, running some simple coalescent simulations (human-like) I get
```
n       #muts   singleton fraction      mean
10      5532    0.024945770065075923    9.250836120401338
100     10564   0.32677016281711474     2.04570100697134
250     12369   0.3879860942679279      1.795731707317073
500     13814   0.4302881135080353      1.6597380752132644
[...]

Issue #2781, opened 2023-07-07
Document divergence_matrix


jeromekelleher:
Following up on #2736, we need to document the function. 

Note that I left the old partially implemented version of divergence matrix as a commented out block here as it could be a useful starting point. We should delete that as part of closing this issue.

Maybe be worth doing #2780 first.

jeromekelleher:
Also add to CHANGELOG when documenting

jeromekelleher:
@petrelharp should we document this for the upcoming release? There's some nice stuff on parallelisation that would be helpful for the downstream code in genetic_releatedness_matrix which is currently undocumented. Would be good to document this once, and then just refer back to the ts.divergence_matrix documentation for reference.

jeromekelleher:
Here's the [code](https://github.com/tskit-dev/tskit/blob/f662d7792d95e3871da1cc54fbde4aed4f3ff055/python/tskit/trees.py#L8331) for reference. I think we have an old version commented out immediately above, so the main job in terms of documentation would probably be describing the parallelisation stuff (which is fairly straightforward?)

benjeffery:
@petrelharp This is the last issue for 0.6.1 - shall we leave things undocumented to get it out?

petrelharp:
This is going to involve a bit of work because to do this properly we need to introduce `mode="mutation"`, which is what the thing we currently call "site" mode actually does. Rather than rush that (and given my other homework at the moment), I would like to postpone to the next release.

jeromekelleher:
OK, let's park for now.

PR #2813, opened 2023-08-04
Better divmat site mode

jeromekelleher:
An improved mutation-by-mutation algorithm for site mode divmat. See #2779 for discussion

Benchmarks:

```
mode n       trees    muts    stats_api   divmat
site    10      4232    5532    0.00    0.00
site    100     9206    10564   0.25    0.02
site    250     10981   12369   1.85    0.08
site    500     12348   13814   7.75    0.29
site    1000    13347   14731   38.30   1.01
```

I tried running it for n=5000 also, but ran out of patience. I think it's clear that the new algorithm is a lot better for this regime anyway.

Code:
```python
def lib_divergence_matrix(ts, mode="site"):
    upper_tri = [
        (i, j) for i in range(ts.num_samples) for j in range(ts.num_samples) if j > i
    ]
    divs = ts.divergence(
        [[j] for j in ts.samples()], indexes=upper_tri, mode=mode, span_normalise=False
    )
    out = np.zeros((ts.num_samples, ts.num_samples))
    for (i, j), div in zip(upper_tri, divs):
        out[i, j] = out[j, i] = div
    return out

def compare_perf():
    seed = 1234
    for n in [10, 100, 250, 500, 1000, 5000]:
        for mode in ["site"]:
            ts = msprime.sim_ancestry(
                n,
                ploidy=1,
                population_size=10**4,
                sequence_length=1e7,
                recombination_rate=1e-8,
                random_seed=seed,
            )
            ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=seed)
            before = time.perf_counter()
            D1 = lib_divergence_matrix(ts, mode=mode)
            time_lib = time.perf_counter() - before
            before = time.perf_counter()
            D2 = ts.divergence_matrix(mode=mode)
            time_nb = time.perf_counter() - before
            # assert np.allclose(D1, D2)

            print(
                mode,
                n,
                ts.num_trees,
                ts.num_mutations,
                f"{time_lib:.2f}",
                f"{time_nb:.2f}",
                sep="\t",
            )
```

### DIFF
--- a/python/tests/test_divmat.py
+++ b/python/tests/test_divmat.py
@@ -22,8 +22,6 @@
 """
 Test cases for divergence matrix based pairwise stats
 """
-import collections
-
 import msprime
 import numpy as np
 import pytest
@@ -266,10 +264,19 @@ def stats_api_divergence_matrix(ts, windows=None, samples=None, mode="site"):
     return out
 
 
-def rootward_path(tree, u, v):
-    while u != v:
-        yield u
-        u = tree.parent(u)
+def group_alleles(genotypes, num_alleles):
+    n = genotypes.shape[0]
+    A = np.zeros(n, dtype=int)
+    offsets = np.zeros(num_alleles + 1, dtype=int)
+    k = 0
+    for a in range(num_alleles):
+        offsets[a + 1] = offsets[a]
+        for j in range(n):
+            if genotypes[j] == a:
+                offsets[a + 1] += 1
+                A[k] = j
+                k += 1
+    return A, offsets
 
 
 def site_divergence_matrix(ts, windows=None, samples=None):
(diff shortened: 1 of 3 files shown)


jeromekelleher:
I tried running this on the unified genealogy trees for chr5 q (hgdp_tgp_sgdp_high_cov_ancients_chr5_q.dated.trees). Details:
```
╔═════════════════════════╗
║TreeSequence             ║
╠═══════════════╤═════════╣
║Trees          │   235014║
╟───────────────┼─────────╢
║Sequence Length│181538261║
╟───────────────┼─────────╢
║Time Units     │  unknown║
╟───────────────┼─────────╢
║Sample Nodes   │     7524║
╟───────────────┼─────────╢
║Total Size     │  1.1 GiB║
╚═══════════════╧═════════╝
╔═══════════╤════════╤═════════╤════════════╗
║Table      │Rows    │Size     │Has Metadata║
╠═══════════╪════════╪═════════╪════════════╣
║Edges      │ 9921165│302.8 MiB│          No║
╟───────────┼────────┼─────────┼────────────╢
║Individuals│    3762│  1.1 MiB│         Yes║
╟───────────┼────────┼─────────┼────────────╢
║Migrations │       0│  8 Bytes│          No║
╟───────────┼────────┼─────────┼────────────╢
║Mutations  │11000819│388.2 MiB│          No║
╟───────────┼────────┼─────────┼────────────╢
║Nodes      │ 1217256│ 62.2 MiB│         Yes║
╟───────────┼────────┼─────────┼────────────╢
║Populations│     215│ 11.9 KiB│         Yes║
╟───────────┼────────┼─────────┼────────────╢
║Provenances│      18│ 12.6 KiB│          No║
╟───────────┼────────┼─────────┼────────────╢
║Sites      │ 4325444│337.4 MiB│         Yes║
╚═══════════╧════════╧═════════╧════════════╝
```

For one thread it needs 2.7G of RAM. Running with 8 threads needed 6G of RAM (reasonable, since we're creating 8 n*n matrices). The old stats API based code quickly ran out of memory on my machine (32G RAM).

jeromekelleher:
Ugggghhhh, looks like there's some memory access problem. Digging.

jeromekelleher:
OK, fixed the memory error, ready for review!

jeromekelleher:
And here's an experimental variant-based method. Has basically the same performance (some wrinkles to be worked out with multi-allelics)

jeromekelleher:
I've updated this to include the correct algorithm for dealing with multi-allelic sites, to give the same answer as the stats API version. This will be easy to implement in C, if we decide to make the change.

What do you think @petrelharp? Unless we can see a clear way forward for making an efficient version counting mutations on the paths, I'm not sure there's much point in creating confusion here by using the number-of-mutations version for site-mode.

jeromekelleher:
Just had a chat with @petrelharp and we decided to change the definition to exactly match the current stats API.

jeromekelleher:
I think this is basically ready to go @petrelharp (although I need to rerun the benchmarks)

python/tests/test_divmat.py
```diff
@@ -391,6 +389,8 @@ def test_single_tree_sites_per_branch(self, m):
         )
         np.testing.assert_array_equal(D1, m * D2)
 
+    # Disabling while we're experimenting with variant-based definition
+    @pytest.mark.skip()
```
petrelharp, line 393: so... we're deleting this one, then?

petrelharp:
LGTM!  Things I have noticed are pending are
- test when first window is not at 0 (at least codecov thinks this isn't)
- fix up the docstring (see the [old version](https://github.com/tskit-dev/tskit/blob/b6f98729b0924f7c71e640553bb8728eff0326b9/python/tskit/trees.py#L7819))

Also, I think we should add the `span_normalise` argument (and probably have it default to `True`, [as usual](https://tskit.dev/tskit/docs/latest/stats.html#sec-stats-span-normalise)). ([Here's](https://github.com/tskit-dev/tskit/blob/b6f98729b0924f7c71e640553bb8728eff0326b9/c/tskit/trees.c#L1827) how we do it elsewhere.) The reason is that, if I've got this right, currently the code is computing the non-span-normalized version (ie just counting mutations);l however, "divergence" usually means "number of mutations divided by sequence length", i.e., span normalized. But we'll have need for the "just count mutations" version, so rather than do error-prone post-hoc multiplying by window lengths to get it out, we should just add the argument.

jeromekelleher:
Great, thanks @petrelharp! Good spots on all counts. I think we can merge this much for now, and follow up with more PRs.

Looks like this version is a small bit slower than the last version (~20%) but I don't think that's worth worrying about.

I logged issues to track the remaining stuff I think:

- #2781
- #2818

jeromekelleher:
Manually merging because we're missing a formal "approval"

Issue #2794, opened 2023-07-13
Replace low-level tree positioning code with tsk_tree_position_t calls


jeromekelleher:
Moving to the next and previous tree in a sequence is currently implemented in [tsk_tree_advance](https://github.com/tskit-dev/tskit/blob/555913cbf830c5c680b5cb26e5f96b3c6ab17ad1/c/tskit/trees.c#L4394) and seeking to a position along the sequence (from the initial state) is done in [tsk_tree_seek_from_null](https://github.com/tskit-dev/tskit/blob/555913cbf830c5c680b5cb26e5f96b3c6ab17ad1/c/tskit/trees.c#L4600) .

We can replace this logic with calls to ``tsk_tree_position_t``, following the patterns used in #2786.

We should do some basic performance testing to check that we haven't introduced any regressions by doing this, say by checking on 

- A large msprime simulated tree sequence
- A large chromosome from the unified genealogy trees
- A SARS-Cov-2 tree sequence

Note that we didn't implement ``seek_backwards`` in #2786, as I thought this would be a good exercise for someone learning how all this stuff works. The first part of closing this issue would be a commit that implements seek_backwards in Python, and adds some more testing to ensure it's correct.

Note that this doesn't address the issues in #2792, but that can be done as a follow-up, since we're not going to make performance any worse by doing it this way.

We would continue to implement ``seek`` by iteration for the non-null case, because we're less sure about the performance implications of this, and we'd need some additional testing to verify that it's correct.

@duncanMR - are you up for this?

duncanMR:
@jeromekelleher Yes, I'd be happy to implement `seek_backwards` and do some performance tests. Just to confirm: are the unit tests for `seek_forward` mostly complete in #2786, such that I just need to translate them to the `seek_backward` case?

jeromekelleher:
Great!

Testing is almost complete I think, but we probably need a few more tests for seeking randomly along the sequence, and changing directions.

PR #2819, opened 2023-08-15
Add seek_backward function with tests

duncanMR:
# Description

Addresses the first part of #2794, to create a `seek_backward` function which can reverse down a tree sequence (or jump to any tree from the null state). Given the symmetry between `seek_forward` and `seek_backward`, I just translated all the tests from the former to the latter. All of them pass now. In both the `TreePosition` and `StatefulTree` versions of `seek_backward`, I resorted to explicitly handling the null case with if statements because I couldn't see a more elegant way that works in all cases; I may have missed a neater solution there. 

I included some commented-out print statements to match the ones currently in `seek_forward`; I can easily remove them if need be.

# PR Checklist:

### DIFF
--- a/python/tests/test_tree_positioning.py
+++ b/python/tests/test_tree_positioning.py
@@ -98,10 +98,6 @@ class StatefulTree:
         old_left, old_right = self.tree_pos.interval
         self.tree_pos.seek_forward(index)
         left, right = self.tree_pos.interval
-        # print()
-        # print("Current interval:", old_left, old_right)
-        # print("New interval:", left, right)
-        # print("index:", index, "out_range:", self.tree_pos.out_range)
         for j in range(self.tree_pos.out_range.start, self.tree_pos.out_range.stop):
             e = self.tree_pos.out_range.order[j]
             e_left = self.ts.edges_left[e]
@@ -113,16 +109,9 @@ class StatefulTree:
                 assert self.parent[c] != -1
                 self.parent[c] = -1
             assert e_left < left
-        # print("index:", index, "in_range:", self.tree_pos.in_range)
         for j in range(self.tree_pos.in_range.start, self.tree_pos.in_range.stop):
             e = self.tree_pos.in_range.order[j]
             if self.ts.edges_left[e] <= left < self.ts.edges_right[e]:
-                # print("keep", j, e, self.ts.edges_left[e], self.ts.edges_right[e])
-                # print(
-                #     "INSERT:",
-                #     self.ts.edge(e),
-                #     self.ts.nodes_time[self.ts.edges_parent[e]],
-                # )
                 c = self.ts.edges_child[e]
                 p = self.ts.edges_parent[e]
                 self.parent[c] = p
@@ -132,12 +121,34 @@ class StatefulTree:
                 # The first and last indexes in the range should always be valid
                 # for the tree.
                 assert a < j < b - 1
-                # print("skip", j, e, self.ts.edges_left[e], self.ts.edges_right[e])
 
     def seek_backward(self, index):
-        # TODO
-        while self.tree_pos.index != index:
-            self.prev()
+        old_left, old_right = self.tree_pos.interval
+        self.tree_pos.seek_backward(index)
+        left, right = self.tree_pos.interval
+        for j in range(self.tree_pos.out_range.start, self.tree_pos.out_range.stop, -1):
+            e = self.tree_pos.out_range.order[j]
+            e_right = self.ts.edges_right[e]
+            # We only need to remove an edge if it's in the current tree, which
+            # can only happen if the edge's right coord is >= the current tree's
+            # right coordinate.
+            if e_right >= old_right:
+                c = self.ts.edges_child[e]
+                assert self.parent[c] != -1
+                self.parent[c] = -1
+            assert e_right > right
+        for j in range(self.tree_pos.in_range.start, self.tree_pos.in_range.stop, -1):
+            e = self.tree_pos.in_range.order[j]
+            if self.ts.edges_right[e] >= right > self.ts.edges_left[e]:
+                c = self.ts.edges_child[e]
+                p = self.ts.edges_parent[e]
+                self.parent[c] = p
+            else:
+                a = self.tree_pos.in_range.start
+                b = self.tree_pos.in_range.stop
+                # The first and last indexes in the range should always be valid
+                # for the tree.
+                assert a > j > b + 1
 
     def iter_backward(self, index):
         while self.tree_pos.index != index:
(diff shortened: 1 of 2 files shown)


python/tests/test_tree_positioning.py
```diff
@@ -135,9 +135,43 @@ def seek_forward(self, index):
...
-        while self.tree_pos.index != index:
-            self.prev()
+        old_left, old_right = self.tree_pos.interval
+        old_index = self.tree_pos.index
+        self.tree_pos.seek_backward(index)
+        left, right = self.tree_pos.interval
+        # print("Current interval:", old_left, old_right)
+        # print("New interval:", left, right)
+        # print("index:", index, "out_range:", self.tree_pos.out_range)
+        if old_index == -1:
```
jeromekelleher, line 145: Hmmm, this doesn't smell right to me. I think we need to look at the initial conditions to think about why we *don't* need to exclude for the null tree. I guess we should really take care of this by setting the out_range to be empty, in the TreePosition.seek_forward and TreePosition.seek_backward functions, rather than having to reason about it here. As you point out, we should have no edges to remove, so there's no point in iterating over the out_range at all.

jeromekelleher (commented):
Looks good @duncanMR!

This is very subtle, but I think you've hit on a good point. The out_range should really just be empty when the seeking from NULL. Do you think you could update the forward and reverse iters to account for this (and add tests to verify)?

python/tests/test_tree_positioning.py
duncanMR: That makes sense to me, I'll give that a try now. It worked in the `seek_forward` case because it was checking if `e_left < old_right = 0` but I soon realised that it didn't work out so neatly when translated to `seek_backward`.

duncanMR: After some experiments, an approach I've found that works is as follows, from the `TreePosition.seek_forward` function as an example:

```python
j = right_current_index
self.out_range.start = j
while j < M and right_coords[right_order[j]] <= left:
    j += 1
self.out_range.stop = j

if self.index == -1:
    #No edges, so out_range should be empty
    self.out_range.start = self.out_range.stop
```

The `out_range.stop` value still needs to be updated for later iterations to work, but setting start = stop prevents `StatefulTree.seek_forward` from iterating over the `out_range`. It works similarly in the `seek_backward` case. Is that what you have in mind?

duncanMR: I just pushed a commit with these changes. The tests are rather trivial; I couldn't think of another way to check the `out_range` is empty without adding code to track if `StatefulTree.seek_x` is looping over the `out_range` or not.

jeromekelleher:
Looks good, let's review when we meet today

duncanMR:
I'm revising some of the tests for clarity. Just posting my reasoning here since it will be hard to read from the commits. We currently have two tests of this form:

```python
    def test_forward_to_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.iter_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.iter_forward(index - 1)
        tree1.assert_equal(tree2)
        tree2 = StatefulTree(self.ts())
        tree2.iter_backward(index - 1)
        tree1.assert_equal(tree2)

    def test_seek_forward_next_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.iter_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.seek_forward(index - 1)
        tree1.assert_equal(tree2)
        tree2 = StatefulTree(self.ts())
        tree2.iter_backward(index - 1)
        tree1.assert_equal(tree2)
```

- The second function is identical to the first, but instead of using `iter_forward` from null it uses `seek_forward` from null. This is redundant because we also have a test that ensures that `iter_forward` and `seek_forward` are the same when used from null. So we only need one of them.
- If the first assertion in the function is true, the second assertion is effectively testing if `iter_backward` to an index from null gives the same result as `iter_forward`. This isn't tested elsewhere and seems worthwhile, but I think it should be a separate test. 

So this is what we are left with:

```python
    def test_seek_forward_to_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.iter_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.iter_forward(index - 1)
        tree1.assert_equal(tree2)

    def test_iter_backward_matches_iter_forward(self, index):
        ts = self.ts()
        tree1 = StatefulTree(ts)
        tree1.iter_forward(index)
        tree2 = StatefulTree(ts)
        tree2.iter_backward(index)
        tree1.assert_equal(tree2)
```

As discussed with @jeromekelleher, I'm also going to make generic functions for the tests that are repeated across multiple classes, to reduce unnecessary repetition. I'm still going to make separate functions for `seek_forward` and `seek_backward` for now, though. I don't think it's worth the trouble to combine them.

duncanMR:
I'm revising some of the tests for clarity. Just posting my reasoning here since it will be hard to read from the commits. We currently have two tests of this form:

```python
    def test_forward_to_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.iter_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.iter_forward(index - 1)
        tree1.assert_equal(tree2)
        tree2 = StatefulTree(self.ts())
        tree2.iter_backward(index - 1)
        tree1.assert_equal(tree2)

    def test_seek_forward_next_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.iter_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.seek_forward(index - 1)
        tree1.assert_equal(tree2)
        tree2 = StatefulTree(self.ts())
        tree2.iter_backward(index - 1)
        tree1.assert_equal(tree2)
```

- The second function is identical to the first, but instead of using `iter_forward` from null it uses `seek_forward` from null. This is redundant because we also have a test that ensures that `iter_forward` and `seek_forward` are the same when used from null. So we only need one of them.
- If the first assertion in the function is true, the second assertion is effectively testing if `iter_backward` to an index from null gives the same result as `iter_forward`. This isn't tested elsewhere and seems worthwhile, but I think it should be a separate test. 

So this is what we are left with:

```python
    def test_seek_forward_to_prev(self, index):
        tree1 = StatefulTree(self.ts())
        tree1.seek_forward(index)
        tree1.prev()
        tree2 = StatefulTree(self.ts())
        tree2.seek_forward(index - 1)
        tree1.assert_equal(tree2)

    def test_iter_backward_matches_iter_forward(self, index):
        ts = self.ts()
        tree1 = StatefulTree(ts)
        tree1.iter_forward(index)
        tree2 = StatefulTree(ts)
        tree2.iter_backward(index)
        tree1.assert_equal(tree2)
```

As discussed with @jeromekelleher, I'm also going to make generic functions for the tests that are repeated across multiple classes, to reduce unnecessary repetition. I'm still going to make separate functions for `seek_forward` and `seek_backward` for now, though. I don't think it's worth the trouble to combine them.

jeromekelleher:
Sounds great, thanks @duncanMR!

duncanMR:
I just committed the test changes and print statement cleanups. Is it okay if I squash the commits for merging?

jeromekelleher (approved):
Yep, LGTM. Let's squash and merge

duncanMR:
I rebased and squashed, so I should be ready now.

jeromekelleher:
Can you create an issue to track adding the new ``step`` function to the C API please, so we don't forget?

PR #2820, opened 2023-08-15
Add span normalise

jeromekelleher:
Tick off a few issues with the divmat implementation.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -7784,10 +7784,11 @@ class TreeSequence:
         chunks.append(chunk)
         return chunks
 
-    def _parallelise_divmat_by_tree(self, num_threads, **kwargs):
+    def _parallelise_divmat_by_tree(self, num_threads, span_normalise, **kwargs):
         """
         No windows were specified, so we can chunk up the whole genome by
-        tree, and do a simple sum of the results.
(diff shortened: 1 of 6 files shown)


python/tskit/trees.py
```diff
@@ -7852,27 +7856,42 @@ def worker(sub_windows):
...
-        windows = [0, self.sequence_length] if windows is None else windows
-
+        windows = self.parse_windows(windows)
         mode = "site" if mode is None else mode
 
-        # NOTE: maybe we want to use a different default for num_threads here, just
-        # following the approach in GNN
+        # FIXME this logic should be merged into __run_windowed_stat if
+        # we generalise the num_threads argument to all stats.
         if num_threads <= 0:
```
petrelharp, line 7873: not `<= 1`?

petrelharp (approved):
LGTM!

python/tskit/trees.py
jeromekelleher: Subtle difference - you might want to do the work in *one* background worker thread. Zero or less means that the work without threads, but default (which simplifies debugging a lot)

jeromekelleher:
Codecov is just wrong about the coverage here, ignoring.

PR #2853, opened 2023-10-06
Add macros to LsHmm_forward_matrix and LsHmm_backward_matrix

### LINKED ISSUE #2841: Release GIL around tskit C API functions?
This arose in the context of @szhan, who is using the tskit lshmm API and wishes to run some operations in parallel.

It would be good to be able to release the GIL so operations can run in parallel if possible - I assume though that this is complicated by the fact that you don't want two operations running concurrently on the same `LsHmm` object?

szhan:
Fixes #2841

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -13254,8 +13254,10 @@ LsHmm_forward_matrix(LsHmm *self, PyObject *args)
             PyExc_ValueError, "haplotype array must have dimension (num_sites,)");
         goto out;
     }
-    err = tsk_ls_hmm_forward(self->ls_hmm, PyArray_DATA(haplotype_array),
-        compressed_matrix->compressed_matrix, TSK_NO_INIT);
+    Py_BEGIN_ALLOW_THREADS err
+        = tsk_ls_hmm_forward(self->ls_hmm, PyArray_DATA(haplotype_array),
+            compressed_matrix->compressed_matrix, TSK_NO_INIT);
+    Py_END_ALLOW_THREADS;
     if (err != 0) {
         handle_library_error(err);
         goto out;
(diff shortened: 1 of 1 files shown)


python/_tskitmodule.c
```diff
@@ -13254,9 +13254,11 @@ LsHmm_forward_matrix(LsHmm *self, PyObject *args)
...
         goto out;
     }
-    err = tsk_ls_hmm_forward(self->ls_hmm, PyArray_DATA(haplotype_array),
-        compressed_matrix->compressed_matrix, TSK_NO_INIT);
-    if (err != 0) {
+    Py_BEGIN_ALLOW_THREADS err
+        = tsk_ls_hmm_forward(self->ls_hmm, PyArray_DATA(haplotype_array),
+            compressed_matrix->compressed_matrix, TSK_NO_INIT);
+    Py_END_ALLOW_THREADS if (err != 0)
+    {
```
benjeffery, line 13261: Did the linter put it all on one line? I'd expect this:
```suggestion
    Py_BEGIN_ALLOW_THREADS
    err = tsk_ls_hmm_forward(self->ls_hmm, PyArray_DATA(haplotype_array),
            compressed_matrix->compressed_matrix, TSK_NO_INIT);
    Py_END_ALLOW_THREADS
    if (err != 0)
    {
```

benjeffery (commented):
Would be nice to add some tests for using this code in multiple threads.

python/_tskitmodule.c
szhan: I tried committing the suggested format, but it reverts to the older format.

jeromekelleher:
The linter is stupid here and doing the wrong thing. The only solution is to put in some comments to force it not to join lines that shouldn't be joined (see other usages of PyBEGIN etc)

szhan:
lint is doing it in other places like

```
    Py_BEGIN_ALLOW_THREADS err = tsk_treeseq_genealogical_nearest_neighbours(
        self->tree_sequence, PyArray_DATA(focal_array), num_focal, reference_sets,
        reference_set_size, num_reference_sets, 0, PyArray_DATA(ret_array));
    Py_END_ALLOW_THREADS if (err != 0)
    {
        handle_library_error(err);
        goto out;
    }
```

python/_tskitmodule.c
benjeffery: If you add a `;` after the macro then it will probably let you!

szhan: Works! Thanks!

jeromekelleher (changes requested):
Hmm, I'm not sure this is such a great idea actually... I can see potential problems with running multiple hmms on the same lshmm object. And if they're on different lshmm objects, well we might as well do things in subprocesses

jeromekelleher:
I think we should drop this and do your multi-threaded matching using processes instead @szhan

PR #2856, opened 2023-10-09
Remove 3.7 and update pre-commit to 3.8

### LINKED ISSUE #2597: Drop support for python 3.7
We will need to do this at some point, so opening issue to track.

benjeffery:
Fixes #2597

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -16,7 +16,6 @@ classifiers =
     Programming Language :: C
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
@@ -46,7 +45,7 @@ platforms =
 
 [options]
 packages = tskit
-python_requires = >=3.7
+python_requires = >=3.8
 include_package_data = True
 install_requires =
     jsonschema>=3.0.0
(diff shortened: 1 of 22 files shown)


python/setup.cfg
```diff
@@ -16,7 +16,6 @@ classifiers =
     Programming Language :: C
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
```
jeromekelleher, line 21: Should add 3.11 here

benjeffery: Good spot!

python/tskit/drawing.py
```diff
@@ -403,13 +403,15 @@ def draw_tree(
         warnings.warn(
             "tree_height_scale is deprecated; use time_scale instead",
             FutureWarning,
+            stacklevel=4,
```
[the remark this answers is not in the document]
benjeffery: One of the pre-commit plugins was complaining about its absence. With `stacklevel` the user gets more than one line of the stacktrace with a warning.

python/setup.cfg
benjeffery: Just checked and it was already there - just snipped off this diff.

jeromekelleher: Ah, my bad

python/tskit/drawing.py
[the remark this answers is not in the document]
benjeffery: B028: No explicit stacklevel argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

PR #2860, opened 2023-10-11
Add python 3.12 support

### LINKED ISSUE #2855: Python 3.12 support
Python 3.12 has been released. 
The good news is that tskit builds and almost all tests pass. The failing ones are those that depend on `numba` via `lshmm` as `numba` for 3.12 has not yet been released. As this will be an issue for every Python release, I suggest we make those tests optional and raise a warning instead of running the tests.

benjeffery:
Fixes #2855

### DIFF
--- a/.github/workflows/docker/shared.env
+++ b/.github/workflows/docker/shared.env
@@ -1,4 +1,5 @@
 PYTHON_VERSIONS=(
+    cp312-cp312
     cp311-cp311
     cp310-cp310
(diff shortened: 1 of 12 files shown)


benjeffery:
Looks like we can't yet even create a conda environment - let alone install anything in it. Will come back to this in a few weeks.

benjeffery:
Replaced by #2964

Issue #2830, opened 2023-08-30
Windows/Site masking for two-site statistics


lkirk:
Implement a method of masking sites for two-site statistics. Since we're interested in slicing a subset of the correlation matrix, we won't need windows for sites. Or, rather, we'll want 1 window.

The overall goal here is to slice out an NxM matrix, not just an NxN matrix. I imagine that windowing will also allow parallelization of the two-site statistics.

In our discussions on #2805, @petrelharp mentioned a site mask, which I find interesting. One issue with providing a window to the two-sites code is that we won't know which site each element of the matrix corresponds to. So, if we provide a list or mask of sites, we can know up-front which sites correspond to which elements of the matrix. Otherwise, I'd thought of returning a list of sites that fall within the window, which is less appealing I think.

We also might want some helper functions that allow us to mask sites within genomic ranges. I'd be curious to know your thoughts @petrelharp, @jeromekelleher, @apragsdale.

jeromekelleher:
An explicit ``sites`` argument sounds good to me. Do you have some proposals for what a Python API would look like?

jeromekelleher:
Here's one proposal to get things moving:
```python
ts.r2_matrix(sites=[[0, 1], [2, 3]]) 
```
Returns a 2x2 matrix R of r^2 values from the pairs of sites, like where, R[0, 0] = r2 between site 0 and 2, R[0, 1] is r2 between 0 and 3, etc.

This seems like a reasonably flexible thing to get started with?

lkirk:
I've had a couple of competing thoughts...

### Explicitly listing the sites

Advantages: we can leave one out easily (i.e. `right_sites=[4, 6, 7]`)
Disadvantages: the list is longer
```
ts.r2_matrix(left_sites=[0, 1, 2, 3, 4, 5], right_sites=[4, 5, 6, 7])
```

### Site mask

Advantages: potentially being able to utilize numpy masking and indexing facilities.
Disadvantages: need one list element per site, potentially too complex.
```
left_mask = np.zeros(ts.num_sites, dtype=bool)
right_mask = np.zeros(ts.num_sites, dtype=bool)
left_mask[1:20] = True
left_mask[4] = False
right_mask[:] = True
ts.r2_matrix(left_sites=left_mask, right_sites=right_mask)
```

### Range

Advantages: compact, easy to specify
Disadvantages: we must process every site, potentially forcing the user to delete sites or stitch together multiple matricies, skipping desired sites.

```
[...]

Issue #2834, opened 2023-08-30
Refactor tsk_bit_array_* structures


jeromekelleher:
I'm a bit confused about what the ``tsk_bit_array_`` struct really is. I thought it was a straightforward bit-set implementation, but there's the idea of ``rows`` which I'm confused by. Is it a 2D bit set? So, a list of N independent bit sets? If so, I think we should change the API to be more explicit about this, and make operations work on the rows and bits rather than using the ``get_row`` operation to get a row, and then having methods which just work on a single row (like intersect, substract, etc). 

Also, I think it would be clearer if we used set theoretic operations through out, so ``add`` -> ``union`` etc.

So, to be clear, we'd have operations like
```
tsk_bit_array_set_bit(self, row, bit)
tsk_bit_array_contains(self, row, bit)
```
etc

What do you thing @lkirk?

lkirk:
Yes, indeed these arrays are a `list of N independent bit sets`. I actually like the word "bit set" more than "bit array" as well. In addition to the refactor, maybe renaming things to `tsk_bitset_*` is a good idea as well?

I like your suggestions, they will simplify the calling code quite a bit.

benjeffery:
Closing for inactivity and labelling "future", please re-open if you plan to work on this.

PR #2863, opened 2023-10-16
Specify sites for two-site statistics

lkirk:
## Description

As discussed on #2830, I've set the foundation on the C side of things for specifying row/column sites. I've also got a bit of the cpython code that wraps this once we're happy with how the c-api looks.

This allows a user to specify an array of row sites to be compared against an array of column sites. To avoid computing redundant entries in the matrix, we implement an index that tracks items and their overlap. When we encounter entries that are redundant, we reflect their value across the diagonal of the matrix (for the valid square sub-matrix that crosses the diagonal). I've attached a diagram to the pull request that describes this in more detail.

In addition to these changes, I've cleaned up the results matrix allocation. Since the python code will be allocating and managing this memory, we'll pull out the allocation code.

I've added tests to capture all of this new functionality. I'm sure we'll need to do a final sweep for coverage once we're done with review.

# PR Checklist:

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2355,18 +2355,63 @@ out:
     return ret;
 }
 
+static void
+get_site_row_col_indices(tsk_size_t n_rows, const tsk_id_t *row_sites, tsk_size_t n_cols,
+    const tsk_id_t *col_sites, tsk_id_t *sites, tsk_size_t *n_sites, tsk_size_t *row_idx,
+    tsk_size_t *col_idx)
+{
+    tsk_size_t r = 0, c = 0, s = 0;
+
+    // Iterate rows and columns until we've exhaused one of the lists
+    while ((r < n_rows) && (c < n_cols)) {
+        if (row_sites[r] < col_sites[c]) {
+            sites[s] = row_sites[r];
+            row_idx[r] = s;
+            s++;
+            r++;
+        } else if (col_sites[c] < row_sites[r]) {
+            sites[s] = col_sites[c];
+            col_idx[c] = s;
+            s++;
+            c++;
+        } else { // row == col
+            sites[s] = row_sites[r];
+            col_idx[c] = s;
+            row_idx[r] = s;
+            s++;
+            r++;
+            c++;
+        }
+    }
+
+    // If there are any items remaining in the other list, drain it
+    while (r < n_rows) {
+        sites[s] = row_sites[r];
+        row_idx[r] = s;
+        s++;
+        r++;
+    }
+    while (c < n_cols) {
+        sites[s] = col_sites[c];
+        col_idx[c] = s;
+        s++;
+        c++;
+    }
+    *n_sites = s;
+}
+
 static int
-get_mutation_samples(
-    const tsk_treeseq_t *ts, tsk_size_t *num_alleles, tsk_bit_array_t *allele_samples)
+get_mutation_samples(const tsk_treeseq_t *ts, const tsk_id_t *sites, tsk_size_t n_sites,
+    tsk_size_t *num_alleles, tsk_bit_array_t *allele_samples)
 {
     int ret = 0;
     const tsk_flags_t *restrict flags = ts->tables->nodes.flags;
     const tsk_size_t num_samples = tsk_treeseq_get_num_samples(ts);
     const tsk_size_t *restrict site_muts_len = ts->site_mutations_length;
-    const tsk_site_t *restrict site;
+    tsk_site_t site;
     tsk_tree_t tree;
     tsk_bit_array_t all_samples_bits, mut_samples, mut_samples_row, out_row;
-    tsk_size_t max_muts_len, mut_offset, num_nodes, s, m, n;
+    tsk_size_t max_muts_len, site_offset, num_nodes, site_idx, s, m, n;
     tsk_id_t node, *nodes = NULL;
     void *tmp_nodes;
 
@@ -2374,11 +2419,12 @@ get_mutation_samples(
     tsk_memset(&all_samples_bits, 0, sizeof(all_samples_bits));
 
     max_muts_len = 0;
-    for (s = 0; s < ts->tables->sites.num_rows; s++) {
-        if (site_muts_len[s] > max_muts_len) {
-            max_muts_len = site_muts_len[s];
+    for (s = 0; s < n_sites; s++) {
+        if (site_muts_len[sites[s]] > max_muts_len) {
+            max_muts_len = site_muts_len[sites[s]];
         }
     }
+    // Allocate a bit array of size max alleles for all sites
     ret = tsk_bit_array_init(&mut_samples, num_samples, max_muts_len);
     if (ret != 0) {
         goto out;
@@ -2387,103 +2433,111 @@ get_mutation_samples(
     if (ret != 0) {
         goto out;
     }
-
+    get_all_samples_bits(&all_samples_bits, num_samples);
     ret = tsk_tree_init(&tree, ts, TSK_NO_SAMPLE_COUNTS);
     if (ret != 0) {
         goto out;
     }
 
-    // A future improvement could get a union of all sample sets
-    // instead of all samples
-    get_all_samples_bits(&all_samples_bits, num_samples);
-
-    // Traverse down each tree, recording all samples below each mutation. We perform one
-    // preorder traversal per mutation.
-    mut_offset = 0;
-    for (ret = tsk_tree_first(&tree); ret == TSK_TREE_OK; ret = tsk_tree_next(&tree)) {
+    // For each mutation within each site, perform one preorder traversal to gather
+    // the samples under each mutation's node.
+    site_offset = 0;
+    for (site_idx = 0; site_idx < n_sites; site_idx++) {
+        tsk_treeseq_get_site(ts, sites[site_idx], &site);
+        ret = tsk_tree_seek(&tree, site.position, 0);
+        if (ret != 0) {
+            goto out;
+        }
         tmp_nodes = tsk_realloc(nodes, tsk_tree_get_size_bound(&tree) * sizeof(*nodes));
         if (tmp_nodes == NULL) {
             ret = TSK_ERR_NO_MEMORY;
             goto out;
         }
         nodes = tmp_nodes;
-        for (s = 0; s < tree.sites_length; s++) {
-            site = &tree.sites[s];
-            tsk_bit_array_get_row(allele_samples, mut_offset, &out_row);
-            tsk_bit_array_add(&out_row, &all_samples_bits);
-            // Zero out results before the start of each iteration
-            tsk_memset(mut_samples.data, 0,
-                mut_samples.size * max_muts_len * sizeof(tsk_bit_array_value_t));
-            for (m = 0; m < site->mutations_length; m++) {
-                tsk_bit_array_get_row(&mut_samples, m, &mut_samples_row);
-                node = site->mutations[m].node;
-                ret = tsk_tree_preorder_from(&tree, node, nodes, &num_nodes);
-                if (ret != 0) {
-                    goto out;
-                }
-                for (n = 0; n < num_nodes; n++) {
-                    node = nodes[n];
-                    if (flags[node] & TSK_NODE_IS_SAMPLE) {
-                        tsk_bit_array_add_bit(
-                            &mut_samples_row, (tsk_bit_array_value_t) node);
-                    }
-                }
-                mut_offset++;
+
+        tsk_bit_array_get_row(allele_samples, site_offset, &out_row);
+        tsk_bit_array_add(&out_row, &all_samples_bits);
+
+        // Zero out results before the start of each iteration
+        tsk_memset(mut_samples.data, 0,
+            mut_samples.size * max_muts_len * sizeof(tsk_bit_array_value_t));
+        for (m = 0; m < site.mutations_length; m++) {
+            tsk_bit_array_get_row(&mut_samples, m, &mut_samples_row);
+            node = site.mutations[m].node;
+            ret = tsk_tree_preorder_from(&tree, node, nodes, &num_nodes);
+            if (ret != 0) {
+                goto out;
             }
-            mut_offset++; // One more for the ancestral allele
-            get_allele_samples(site, &mut_samples, &out_row, &(num_alleles[site->id]));
+            for (n = 0; n < num_nodes; n++) {
+                node = nodes[n];
+                if (flags[node] & TSK_NODE_IS_SAMPLE) {
+                    tsk_bit_array_add_bit(
+                        &mut_samples_row, (tsk_bit_array_value_t) node);
+                }
+            }
         }
+        site_offset += site.mutations_length + 1;
+        get_allele_samples(&site, &mut_samples, &out_row, &(num_alleles[site_idx]));
     }
-    // if adding code below, check ret before continuing
+// if adding code below, check ret before continuing
 out:
     tsk_safe_free(nodes);
     tsk_tree_free(&tree);
     tsk_bit_array_free(&mut_samples);
     tsk_bit_array_free(&all_samples_bits);
-    return ret;
+    return ret == TSK_TREE_OK ? 0 : ret;
 }
 
 static int
 tsk_treeseq_two_site_count_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
     const tsk_bit_array_t *sample_sets, tsk_size_t result_dim, general_stat_func_t *f,
-    sample_count_stat_params_t *f_params, norm_func_t *norm_f,
-    const double *TSK_UNUSED(left_window), const double *TSK_UNUSED(right_window),
-    tsk_flags_t options, tsk_size_t *result_size, double **result)
+    sample_count_stat_params_t *f_params, norm_func_t *norm_f, tsk_size_t n_rows,
+    const tsk_id_t *row_sites, tsk_size_t n_cols, const tsk_id_t *col_sites,
+    tsk_flags_t options, double *result)
 {
+
     int ret = 0;
-    tsk_bit_array_t allele_samples;
-    tsk_bit_array_t site_a_state, site_b_state;
-    tsk_size_t inner, result_offset, inner_offset, a_offset, b_offset;
-    tsk_size_t site_a, site_b;
+    tsk_bit_array_t allele_samples, c_state, r_state;
     bool polarised = false;
-    const tsk_size_t num_sites = self->tables->sites.num_rows;
+    tsk_id_t *sites;
+    tsk_size_t r, c, s, n_alleles, n_sites, *row_idx, *col_idx;
+    double *result_row;
     const tsk_size_t num_samples = self->num_samples;
-    const tsk_size_t max_alleles = self->tables->mutations.num_rows + num_sites;
-    tsk_size_t *num_alleles = tsk_malloc(num_sites * sizeof(*num_alleles));
-    const tsk_size_t *restrict site_muts_len = self->site_mutations_length;
+    tsk_size_t *num_alleles = NULL, *site_offsets = NULL;
+    tsk_size_t result_row_len = n_cols * result_dim;
 
     tsk_memset(&allele_samples, 0, sizeof(allele_samples));
 
-    if (num_alleles == NULL) {
+    sites = tsk_malloc(self->tables->sites.num_rows * sizeof(*sites));
+    row_idx = tsk_malloc(self->tables->sites.num_rows * sizeof(*row_idx));
+    col_idx = tsk_malloc(self->tables->sites.num_rows * sizeof(*col_idx));
+    if (sites == NULL || row_idx == NULL || col_idx == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    get_site_row_col_indices(
+        n_rows, row_sites, n_cols, col_sites, sites, &n_sites, row_idx, col_idx);
+
+    // We rely on n_sites to allocate these arrays, they're initialized to NULL for safe
+    // deallocation if the previous allocation fails
+    num_alleles = tsk_malloc(n_sites * sizeof(*num_alleles));
+    site_offsets = tsk_malloc(n_sites * sizeof(*site_offsets));
+    if (num_alleles == NULL || site_offsets == NULL) {
         ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
 
-    ret = tsk_bit_array_init(&allele_samples, num_samples, max_alleles);
+    n_alleles = 0;
+    for (s = 0; s < n_sites; s++) {
+        site_offsets[s] = n_alleles;
+        n_alleles += self->site_mutations_length[sites[s]] + 1;
+    }
+    ret = tsk_bit_array_init(&allele_samples, num_samples, n_alleles);
     if (ret != 0) {
         goto out;
     }
-    ret = get_mutation_samples(self, num_alleles, &allele_samples);
+    ret = get_mutation_samples(self, sites, n_sites, num_alleles, &allele_samples);
     if (ret != 0) {
-        goto out;
-    }
-
-    // Number of pairs w/ replacement (sites)
-    *result_size = (num_sites * (1 + num_sites)) / 2U;
-    *result = tsk_calloc(*result_size * result_dim, sizeof(**result));
-
-    if (result == NULL) {
-        ret = TSK_ERR_NO_MEMORY;
         goto out;
     }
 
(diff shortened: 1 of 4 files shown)


c/tests/testlib.h
```diff
@@ -55,6 +55,22 @@ void parse_individuals(const char *text, tsk_individual_table_t *individual_tabl
...
+        }                                                                               \
+    } while (0)
+#define ASSERT_ARRAYS_EQUAL(len, a, b)                                                  \
+    do {                                                                                \
+        tsk_size_t j;                                                                   \
+                                                                                        \
+        for (j = 0; j < len; j++) {                                                     \
+            CU_ASSERT_EQUAL(a[j], b[j]);                                                \
+        }                                                                               \
+    } while (0)
```
lkirk, line 73: I've added these because it makes failures much easier to debug. For example, with the original `assert_arrays_almost_equal` function, we see this type of thing when the assertion fails:

```
  Test: test_paper_ex_two_site ...FAILED
    1. ../tests/testlib.c:975  - CU_ASSERT_DOUBLE_EQUAL(a[j],b[j],1e-9)
```

Note the source file/line number reference. Now, with this new macro `ASSERT_ARRAYS_ALMOST_EQUAL`, we see this:

```
  Test: test_paper_ex_two_site ...FAILED                                                                                                                       
    1. ../tests/test_stats.c:2032  - CU_ASSERT_DOUBLE_EQUAL(result[j],truth_one_set[j],1e-9)
```

I now see the exact line where my assertion fails.

I see this as a quality of life improvement, but I'm happy to remove it if you think it doesn't fit here.

c/tskit/trees.c
```diff
@@ -2355,28 +2355,157 @@ compute_general_two_site_stat_result(const tsk_bit_array_t *site_a_state,
...
+    }
+    tmp = tsk_realloc(in, n * sizeof(*tmp));
+    if (tmp == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    *out = tmp;
+out:
+    return ret;
+}
```
lkirk, line 2394: This reallocation routine allows us to only store the individual sites we've specified, instead of arrays that are over-allocated to store all possible sites. This could be useful in large tree sequences.

c/tskit/trees.c
```diff
@@ -2446,79 +2598,107 @@ get_mutation_samples(
...
+        tsk_bit_array_get_row(&allele_samples, site_offsets[r_idx], &r_state);          \
+        tsk_bit_array_get_row(&allele_samples, site_offsets[c_idx], &c_state);          \
+        ret = compute_general_two_site_stat_result(&r_state, &c_state,                  \
+            num_alleles[r_idx], num_alleles[c_idx], num_samples, state_dim,             \
+            sample_sets, result_dim, f, f_params, norm_f, polarised,                    \
+            &(result_row[(c_id) *result_dim]));                                         \
+        if (ret != 0) {                                                                 \
+            goto out;                                                                   \
+        }                                                                               \
+    } while (0);
```
lkirk, line 2662: This macro simplifies the loops below, but I'll admit, it does make debugging a bit more difficult

c/tskit/trees.c
```diff
@@ -2558,14 +2738,46 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
+check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_rows)
+{
+    int ret = 0;
+    tsk_id_t id;
+    tsk_size_t i;
+
+    if (sites == NULL) {
+        goto out;
+    }
+
```
lkirk, line 2751: I'd like to add a few more error values, but I assumed that would merit some discussion.

Something like: `TSK_ERR_DUPLICATE_SITES`, and `TSK_ERR_NULL_SITE`.

lkirk:
The matrix filling is now a bit more complicated that it was initially. 
Here's a diagram of how it works: [Matrix Filling.pdf](https://github.com/tskit-dev/tskit/files/12913123/Matrix.Filling.pdf). In the case of computing an NxN matrix, this will reduce redundant computations substantially, allowing us to compute the upper triangle and reflect it to make the results matrix. It also allows us to compute a rectangular NxM matrix while retaining the benefits of avoiding redundant computations (i.e. computing r^2 for site 0/1 and 1/0).

c/tskit/trees.c
```diff
@@ -2558,14 +2738,46 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
 }
 
+static int
+check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_rows)
+{
+    int ret = 0;
+    tsk_id_t id;
+    tsk_size_t i;
+
+    if (sites == NULL) {
```
jeromekelleher, line 2748: What's the interpretation here? I would have though this would be an error

c/tskit/trees.c
```diff
@@ -2558,14 +2738,46 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
+        goto out;
+    }
+
+    for (i = 0; i < num_sites - 1; i++) {
+        id = sites[i];
```
jeromekelleher, line 2753: Can use a generic check like this:
```
if (sites[i] < 0 || sites[i] >= num_sites) {
    ret = TSK_ERR_SITE_OUT_OF_BOUNDS
    goto out

}

File: c/tskit/trees.c [hunk 7]
```diff
@@ -2558,14 +2738,46 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
+        if (id == TSK_NULL) {
+            ret = TSK_ERR_BAD_PARAM_VALUE; // TODO: new err type? TSK_BAD_SITE_VALUE?
+            goto out;
+        }
+        if (id > sites[i + 1]) {
```
jeromekelleher, line 2758: Does it matter if there are repeats? We're just specifying things to compute in the output matrix, right? I worry we're putting effort in to stop people doing "stupid" things, when this might also stop them doing clever things

c/tskit/trees.c
```diff
@@ -2558,14 +2738,46 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
+            ret = TSK_ERR_UNSORTED_SITES; // TODO: this checks no repeats, but error is
+                                          // ambiguous
+            goto out;
+        }
+    }
+    // check if the last or only value is null
+    if (num_sites != 0
```
jeromekelleher, line 2765: Why are we doing this?

c/tests/testlib.h
jeromekelleher: Yep, very nice. I'm sure I already did something equivalent, but it must not have been merged yet.

jeromekelleher (commented):
I've had a quick look and overall LGTM

Can you explain the rationale for doing the site index lookup, and what your assumptions about the input are please? I'm not quite following what you're checking for, and why.

Requiring that the row and cols are sorted seems reasonable all right.

Also, I don't follow the need for the site index and duplicate checking. Does this happen in common cases? Could we not deal with this more easily at a higher level, if so?

c/tskit/trees.c
lkirk: This is leftover from when I thought I'd specify the defaults in C. If no sites were specified and we were going to use all of the sites, we wouldn't need to validate them, we'd just want to use all available sites. However, I think I decided it'd be much simpler to specify the defaults on the python side. Do you agree? If so, I'll make this condition an error.

c/tskit/trees.c
lkirk: Repeats are going to complicate the indexing of the output array. We'd have to track the number of shared for rows and columns, we wouldn't be able to rely on their union. Also, we'd need a bit of rework when walking the trees because we assume that the sites are ascending and not repeating.

In short, I think it's possible to support repeats, but it's not currently supported. Do you think it's something we should be supporting?

c/tskit/trees.c
lkirk: We check if every item is non-null and ordered by iterating from 0->n-1. We then need to check the last item if it's null. Also, since we have checked if the sites are ordered, by checking if the last site exceeds the length of the site rows table, we can ensure we're not specifying a site that's out of bounds.

c/tskit/trees.c
lkirk: I think this is more straightforward. If we've encountered an out of bounds site, immediately fail.

Since we're checking pairs, we'll still need to handle the last site separately.

lkirk:
> Can you explain the rationale for doing the site index lookup, and what your assumptions about the input are please? I'm not quite following what you're checking for, and why.

Sorry about the input validation, there were a couple of issues with that code that made it confusing. I've cleared it up now I think. My only question is how to fail if no sites are specified. Do we want to do a `tsk_bug_assert` or provide a more informative error type?

The site index lookup is needed because we're only collecting data for a subset of sites. We first store all of the sites we want to collect data from, then use that list while walking over the trees. When we compute statistics, we index into the data we've gathered with the `_idx` suffix entries of the index struct.

The assumptions about the input sites are that they are ordered, valid sites (non-null and in-range), and non-duplicated. As I've mentioned in my comments above, we're not equipped to handle duplicates, but I could add code that allows for duplicates if that seems like a reasonable requirement.

Overall, I think that we'll specify sites in python and if they're not specified, we can pass through an array of all sites. The C code can simply error if no sites are specified, but I think it's worth validating on the C side if we intend to make this part of the public C api surface.

jeromekelleher:
Thanks for the explanations @lkirk. I'm afraid I'm still not really getting the algorithm here, and this is where I'm really missing having a straightforward Python implementation --- I find it very hard to think about actual algorithmic processes when it's obscured by all the C-stuff.

Do you have a Python prototype of this code somewhere? I think we're going to need it at some point for maintainability anyway, so maybe we could switch to creating a ``test_two_site_stats.py`` file, following the general pattern of ``test_divmat.py``? We can fill in the calls to the "lib" code here later, but we're going to need it anyway, and it would really help me to have some python code to look at and play with.

I think the basic "shape" (and API) of this is good, I just don't follow the details of the "indexing" code. From a quick glance I can see it has some memory management issues, but I want to get my head around what it does and why before jumping in to that. In general, I think we should try to develop the algorithm in Python first, test, and then implement in C.

lkirk:
I have a python prototype [here](https://github.com/lkirk/ts-two-locus-proto/blob/d84e360/two_locus_proto/site.py), but it has fallen out of sync with the c implementation. It doesn't have sample sets or the ability to subset sites. In any case, it doesn't follow the c implementation and is more of an orthogonal check I'd say. I can add these new features to it, but I'll need another day or so to get that done. In the meantime, I tried to come up with another document in an attempt to clarify the details of the indexing: 
[Indexing.pdf](https://github.com/tskit-dev/tskit/files/12992331/Indexing.pdf)

lkirk:
I've added a python test file that currently tests itself (`python/tests/test_ld_matrix.py`). It's up to date with the details of the algorithm now.

lkirk:
Hi @jeromekelleher, just checking in on this. Is there anything more I can do to clarify the inner workings of the indexing algo? The python code I've attached mirrors the indexing closely. Lines 202-226 of the python code, along with the written documentation I've attached should explain things relatively well. I'm happy to have a zoom call or provide some further explanation if needed.

jeromekelleher:
Whoops, sorry @lkirk - I've been trying to get a manuscript over the line and been focusing completely on that for a few days. I'll have a good look at this tomorrow.

jeromekelleher:
Thanks @lkirk, I think I'm starting to get the idea here... The variable names on the indexing structure aren't that obvious to me, so some descriptive comments in the Python version would be helpful.

I still don't follow why we think there'll be a lot of shared sites between the rows and the cols though. Do you have some ideas for applications where we expect this? And some quantitative ideas on how much time you'd expect to save by doing this indexing?

Premature optimisation is a real thing, and I don't know how many times I've come up with some complicated fancy idea for how to do something, only to be disappointed to see that it's actually slower than the simple, obvious method. So, I would always now implement things in the simple, obvious way first, *even if* (or especially, maybe) I think I know a better way.

lkirk:
Hi @jeromekelleher sorry for my late reply. Thank you for taking another look. I've thought about if we might be over-complicating things here. I think any way we slice it, it's going be complicated to index into gathered data, site ids, and result matrix.

I tried to write a reduced version of this code and it wasn't a whole lot simpler. The issue is that we need to maintain a union of the sites and its corresponding indices (which happen to be the same indices as the gathered data/state matrix). Then we also need to store the row/column indices into the results matrix. All said, if we eliminated shared/diff tracking, we could get rid of 4 struct members in `stat_matrix_site_indicies`, and we'd eliminate all but 1 loop in the result matrix filling (maybe this would be the real gain from simplification?).

This complexity stems from the requirement that we compute LD for arbitrary NxM matricies. If it was an NxN matrix, this would be easier.

The reason we want to track shared sites between rows and columns is the case where a user specifies a square output matrix where the row sites and column sites are identical (a very common case). We only want to compute the upper triangle and reflect to the lower triangle of the output matrix because computations across the diagonal are redundant. Most of the time is spent on intersecting sample sets between sites, so we want to avoid redundant computations as much as possible (IMO). I did a simple benchmark where I took chromosome 3 from the SGDP trees inferred by tsinfer and ran this code on:

- 10,000 sites that perfectly overlap -- forming a square matrix (`best_case`)
- 10,000 sites for the row site ids and 10,000 sites for the column site ids that are non-overlapping (`worst_case`)

```
best_case took 13.81 s
worst_case took 26.44 s
```

We can see that by tracking overlaps and reflecting we get an UP TO 2x speedup in a common usecase.

The nice thing is that we also handle edge cases where we can simply specify arbitrary lists of site ids, retaining any benefits from avoiding redundant computations.

That said, I'm happy to consider alternative approaches.

I also went ahead and tried to clean up the python prototype, renaming the indexing object attributes to (hopefully) more descriptive names. I also made a first pass at documenting the python code so it's hopefully more understandable.

python/tests/test_ld_matrix.py
```diff
@@ -0,0 +1,561 @@
...
+                for mutation in site.mutations:
+                    # if we've seen the state, use the index as the enumerated
+                    # state value, otherwise, add another state value
+                    if mutation.derived_state in states:
+                        current_state = states.index(mutation.derived_state)
+                    else:
+                        states.append(mutation.derived_state)
+                        current_state = len(states) - 1
+
+                    for sample_id in tree.samples(mutation.node):
```
jeromekelleher, line 58: I don't think this is right - you're assuming that the samples are 0 to n - 1 here, but that's not true in general. What you need to do is use the sample_index_map:
```
sample_index_map = np.zeros(ts.num_nodes, dtype=int32) - 1
for j, u in enumerate(ts.samples()):
     sample_index_map[u] = j

 # Later:
     for u in tree.samples(mutation.node):
         sample_index = sample_index_map[u]  
         mutation[site.id, sample_index] = current_state
```

```

File: python/tests/test_ld_matrix.py [hunk 10]
```diff
@@ -0,0 +1,561 @@
...
+    a 2d numpy array. We enumerate all allele states in a site, and
+    store the enumerated state for the sample.
+
+    :param ts: Tree sequence to gather data from
+    :param sites: List of sites to gather data for
+    :returns: 2d numpy array containing the final allelic state
+              for all samples and sites in the index.
+    """
+    state = np.zeros((len(sites), ts.num_samples), dtype=int)
+    current_site = 0
```
jeromekelleher, line 43: No point in tracking ``current_site`` here, just use ``site.id`` below.

jeromekelleher (commented):
Sorry I've been so slow @lkirk - this is tricky stuff and I haven't found time to get by head into it. I started getting my head back into it here, but I've gotten bogged down again. Maybe we should organise a call for next week so we can go through the code together?

python/tests/test_ld_matrix.py
lkirk: I think I want to punt this change to when I write all of the python tests for ld_matrix. I don't currently have any data to test the sample index map with, but I will create some when I make those tests.

c/tskit/trees.c
```diff
@@ -2558,14 +2708,43 @@ sample_sets_to_bit_array(const tsk_treeseq_t *self, const tsk_size_t *sample_set
...
 }
 
+static int
+check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_rows)
+{
+    int ret = 0;
+    tsk_size_t i;
+
+    if (sites == NULL || num_sites == 0) {
+        ret = TSK_ERR_BAD_SITE_POSITION; // TODO: error should be no sites?
```
lkirk, line 2718: Still considering a `TSK_ERR_NULL_SITES` error here or something of the sort.

lkirk:
Hi @jeromekelleher sorry for the delay in fixing this up. I've covered the issues that we discussed during our meeting:

1. We're no longer reallocating things
2. The malloc calls are now checked
3. I've renamed instances of haplotype matrix to haplotype counts
4. I've renamed members of the index struct to be a little more intuitive
5. No more incrementing variables during use
6. I've gotten test coverage up to 94%

I think this is about ready, please let me know if there are any other blockers that I need to address before it's ready to merge.

c/tskit/trees.h
```diff
@@ -1818,6 +1816,29 @@ bool tsk_tree_position_prev(tsk_tree_position_t *self);
 int tsk_tree_position_seek_forward(tsk_tree_position_t *self, tsk_id_t index);
 int tsk_tree_position_seek_backward(tsk_tree_position_t *self, tsk_id_t index);
 
+// These are in the header for testing purposes only
+struct stat_matrix_site_indicies {
```
jeromekelleher, line 1820: typo probably should be, ``stat_matrix_site_indices``. Elsewhere we use the ``indexes`` plural form, FYI but whatever you prefer.

jeromekelleher: I guess strictly speaking if we put this in the public header we should also put in a ``tsk_`` prefix so that we're fulfilling the contracts we specify in the docs. (Names space collisions are unlikely I agree, but we do say that all public symbols start with a ``tsk_`` prefix)

jeromekelleher: Also might as well follow the usual conventions and ``typedef`` it as ``tsk_stat_matrix_site_indexes`` (rather than as a struct)

python/tests/test_ld_matrix.py
jeromekelleher: This is still relevant

python/tests/test_ld_matrix.py
jeromekelleher: OK, can we drop a comment to this effect in here so we don't forget please?

python/tests/test_ld_matrix.py
```diff
@@ -0,0 +1,559 @@
...
+    for i in range(len(sites) - 1):
+        if sites[i] < 0 or sites[i] >= max_sites:
+            raise ValueError(f"Site out of bounds: {sites[i]}")
+        if sites[i] >= sites[i + 1]:
+            raise ValueError(f"Sites not sorted: {sites[i], sites[i + 1]}")
+    if sites[-1] < 0 or sites[-1] >= max_sites:
+        raise ValueError(f"Site out of bounds: {sites[i + 1]}")
+
+
+def get_matrix_indices(row_sites, col_sites):
```
jeromekelleher, line 176: I think this would be more naturally done as the constructor to SiteMatrixIndex, so that we call `` idx = SiteMatrixIndex(row_sites, col_sites)`` in client code. Here, I found myself going up and down the file to try and figure out what the fields were actually doing, where we should really just encapsulate everything in the SiteMatrixIndex class.

python/tests/test_ld_matrix.py
```diff
@@ -0,0 +1,559 @@
...
+    :returns: 3d numpy array containing LD for (sample_set,row_site,column_site)
+    """
+    state = np.asarray(state)
+    norm = NormMethod(norm_method)
+
+    result = np.zeros((len(sample_sets), idx.n_rows, idx.n_cols))
+
+    def compute(left_states, right_states):
+        hap_counts = np.zeros((np.max(left_states) + 1, np.max(right_states) + 1))
+        # compute the haplotype matrix
```
jeromekelleher, line 277: This is still confusing me - what's the haplotype matrix? (I'm pretty sure it's not the same as what we get when calling ``ts.haplotypes()``, e.g.). Can we rename this please?

jeromekelleher (commented):
I've gone through again and I'm afraid I still haven't managed to get my head fully around the SiteMatrixIndex. I feel like it could be substantially simpler, and I'm having to spend a lot of time understanding how it works. Unless the code is easy to follow, it won't get maintained or extended, so I do want to get this right.

This is a pity, because it's just an implementation detail, and it's blocking adding the actual functionality.

My suggestion here would be to take a backup branch of this PR, and just pull out the SiteMatrixIndex stuff from this PR, using a naive and direct algorithm instead. Let's not worry about absolute performance, and just focus on getting the functionality in and do optimisations afterwards. I think this is really just a case of deleting code, as the actual logic can be recast really easily using the "compute" primitive you have already. 

Then, once we have the functionality in we can come back to the SiteMatrixIndex logic, which should be a simple diff from your current commit to main, and easy to recover.

I know this is annoying, and I'm sorry to make this so slow. It will be much easier to understand your optimisations when they are isolated though, and I'm confident we can get them in quickly when they are pulled out by themselves.

lkirk:
@jeromekelleher I agree with you. It could definitely be simpler and it'll be a bit of technical debt looming over this code. I'll pull it out and make things simpler. I can have something by Friday for you to look at.

Along these lines... I was thinking about writing the python code from the bottom up to mirror the C code. Would that make things simpler in the long run? I'm happy to do that too.

jeromekelleher:
That sounds great, thanks @lkirk . Mirroring the python and c implementations (to a reasonable degree, no need to be too literal) would help all right

python/tests/test_ld_matrix.py
```diff
@@ -0,0 +1,834 @@
...
+from itertools import combinations_with_replacement, permutations
+from typing import Any, Callable, Generator, Self
+
+import numpy as np
+import pytest
+
+import tskit
+
+
+class BitSet:
```
lkirk, line 11: I've modeled this class after the BitSet that I'm imagining for #2834.

The only strange thing about the API is that the intersect method saves to a third length-1 array, while the rest of the methods modify the calling instance. I suppose `union` and `difference` could write out to an `out` array like `intersect`, but it would be clunky and slightly confusing to be passing the calling instance back into the method.

In any case, I'm not sure how much effort should be put into this API since technically only the two-locus code is using it for now.

lkirk:
Okay, so that took a little longer than expected. I've simplified things @jeromekelleher.

There are no longer any optimizations or fancy index tracking around the results matrix. We're now just filling the matrix from top to bottom, using the exact indices specified by the user.

The python code now closely mirrors the C code (except for the simple refactor of bit arrays that I commented on in the code). I've also annotated the python code with types (compatible with all supported versions of python) to add some further documentation.

Code coverage is up, tests are passing and merge conflicts are handled. I'm happy to hear any feedback and/or points of confusion. Hopefully this makes things more understandable.

c/tskit/trees.c
```diff
@@ -2428,10 +2480,16 @@ get_mutation_samples(
...
-            mut_offset++; // One more for the ancestral allele
-            get_allele_samples(site, &mut_samples, &out_row, &(num_alleles[site->id]));
+            site_offset += site->mutations_length + 1;
+            get_allele_samples(site, &mut_samples, &out_row, &(num_alleles[site_idx]));
+            site_idx++;
+
+            // If we've hit the last site in our index, finish up
+            if (site_idx >= n_sites) {
+                ret = 0;
+                goto out;
```
jeromekelleher, line 2491: We only use gotos for "exception handling", so it would be good to change the flow a bit here. I'm also not sure that you're correctly iterating over the trees (see comment above). How about this?

```c
while (site_idx < num_sites) {
    ret = tsk_tree_seek(&tree, sites_position[sites[site_idx]], 0);
   if (ret < 0) {
        goto out;
   }
    tsk_bug_assert(ret == TSK_TREE_OK);
    for (s = 0; s < tree.sites_length && site_idx < num_sites; s++) {
           site = &tree.sites[s];
           site_idx++;
        
          // do stuff
     } 

}
```

c/tskit/trees.c
```diff
@@ -2387,20 +2433,20 @@ get_mutation_samples(
...
 
     // Traverse down each tree, recording all samples below each mutation. We perform one
-    // preorder traversal per mutation.
-    mut_offset = 0;
-    for (ret = tsk_tree_first(&tree); ret == TSK_TREE_OK; ret = tsk_tree_next(&tree)) {
+    // preorder traversal per mutation. We start at the tree containing the first site of
+    // interest. We rely on our sites bounds checking to access the first site position
+    site_offset = 0;
+    site_idx = 0;
+    for (ret = tsk_tree_seek(&tree, ts->tables->sites.position[sites[0]], 0);
```
jeromekelleher, line 2448: Is this correct? Won't you seek to the first tree over and over doing this? See below for suggested restructuring of flow control.

c/tskit/trees.c
jeromekelleher: Let's file a follow-up issue on this one - additional error cases for 2 site stats.

jeromekelleher (commented):
Looks great! A couple of small issues, but looks good to merge then. Thanks for your patience @lkirk!

c/tskit/trees.c
lkirk: You're right about that. I didn't realize that this would cause the loop to wrap around until I messed with it in isolation. See my detailed comment below.

c/tskit/trees.c
lkirk: Ah, I see what you're saying. After a few tests, I see what's going on. The issue I have in your proposed solution is that we'd also be calling realloc for every site. From what I understand, realloc is not specified in the C standard (GNU returns the same memory when realloc is called with the same size, but I'm not sure about other libc implementations -- see [here](https://www.gnu.org/software/libc/manual/html_node/Changing-Block-Size.html) in the portability notes).

I've proposed a solution (my most recent commit) where we seek to the first tree, then keep seeking until we run out of sites or trees.

What do you think about this? I can implement what you've proposed if you think that's the better, more straightforward route.

jeromekelleher: Generally, things have to be correct *first* and then fast afterwards.  You make them correct by doing things simply first. Then you profile, and if it's not fast enough, you optimise **based on this profiling**. 

The realloc thing you identify is a small detail (which is trivially worked around, if it really *is* a problem). You haven't profiled this, so you don't know if it is a problem. That is  not sufficient reason to make the code so much more complicated.

I do understand the impulse to optimise, but you're going about it the wrong way. I've found this classic [worse is better](https://dreamsongs.com/RiseOfWorseIsBetter.html) essay very illuminating, and has convinced me of the importance of simplicity of implementation, almost above all other considerations.

lkirk: Thank you for passing that along, I've heard of "worse is better" colloquially, but I haven't seen this essay before. I agree with it and like the ideas, I just need to translate them into my software design.

I've adopted your suggestions and things look a bit simpler. `tsk_tree_seek` always returns 0 unless there's an error, so there's no need to make any further asserts as far as I can tell.

I also updated the python code to reflect the change.

lkirk:
Squashed and ready. I'll make that issue for the error types today. I also talked with @apragsdale and he suggests we make an issue for tracking the matrix optimizations if we ever decide to remove redundant computations. Does that sound reasonable?

jeromekelleher:
Sounds good to me. Let's get the Python-C plumbing setup up next and so we can start running this stuff!

PR #2871, opened 2023-11-14
Fix draw_svg warning for mutations above unplotted nodes

### LINKED ISSUE #2870: Warning for `draw_svg` plot uses `logging` module instead of `warnings`
I came across a minor issue with the plotting code: the `draw_svg` warning for mutations above unplotted nodes uses the `logging` module instead of the `warnings` module, that we'd usually use for user warnings. I discovered this because I couldn't find a way to suppress it. The warning is [currently this](https://github.com/tskit-dev/tskit/blob/main/python/tskit/drawing.py#L1367C1-L1371C14):

```python
if len(unplotted) > 0:
     logging.warning(
          f"Mutations {unplotted} are above nodes which are not present in the "
          "displayed tree, so are not plotted on the topology."
     )
```

If we were to use `warnings` instead, I'm guessing we'd do something like this? I'm not sure what stack level is appropriate here.

```python
if len(unplotted) > 0:
     warnings.warn(
          f"Mutations {unplotted} are above nodes which are not present in the "
          "displayed tree, so are not plotted on the topology.",
          UserWarning,
          stacklevel=2
     )
```

duncanMR:
## Description

Fixes #2870.
I changed the warning module and modified all the tests that raised that warning to catch it correctly. I made the stack level 2 simply because that seemed to be common with `UserWarning` instances elsewhere in the package; I'm not sure how one determines that here.

# PR Checklist:

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2022 Tskit Developers
+# Copyright (c) 2018-2023 Tskit Developers
 # Copyright (C) 2017 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


benjeffery (commented):
Generally in the test we're using pytest.warns as it's a bit cleaner and lets you check the message: https://happytest.readthedocs.io/en/latest/warnings/#warns

jeromekelleher:
Yep lgtm

benjeffery (approved):
Thanks @duncanMR!

PR #2867, opened 2023-11-01
Add tree_pos to tsk_tree_t struct

duncanMR:
## Description

This PR aims to address the first part of #2795 by adding the `tree_position` class to the `tsk_tree_t` struct. I've done a first draft of this and it passes all the unit tests without any memory issues detected by Valgrind. I'm not confident that the memory management is correct yet, though!

# PR Checklist:

- [ ] Tests that fully cover new/changed functionality.
- [ ] Documentation including tutorial content if appropriate.
- [ ] Changelogs, if there are API changes.

Issue #2795, opened 2023-07-13
Replace calls to tsk_tree_position_t with the internal tree version


jeromekelleher:
#2794 outlines the process of using the ``tks_tree_position_t`` class to handle the logic of moving between trees. To do this we'll embed a ``tsk_tree_position_t`` into the ``tsk_tree_t`` struct. A nice consequence of this (I think) is that we'll then have access (withing the library at least) to the edges that were inserted and removed in order to perform the last tree transformation, and this will remove the need to have a separate ``tree_position_t`` iterator for the places we're currently using them (since #2786):

- kc_distance algorithm
- haplotype matching

PR #2874, opened 2023-12-04
Add tree_pos to tsk_tree_t struct (updated)

duncanMR:
## Description

This PR aims to address the first part of https://github.com/tskit-dev/tskit/issues/2795 by adding the tree_position class to the tsk_tree_t struct. This is a fresh branch based on changes in #2867 to fix a CI issue.

# PR Checklist:


duncanMR:
@jeromekelleher Unfortunately the CI issues remain on the fresh branch. Is there anyway to force update the package cache?

jeromekelleher:
@benjeffery what'st the easiest way of dealing with this?

duncanMR:
Ben suggested bumping the version number in the cache config, which I did in the last commit, but for some reason it seems to have not refreshed the cache.

benjeffery:
Bumping the number will certainly have flushed the cache. Think there is a deeper issue here - checking it now.

duncanMR:
Ah, I realised that using Sync Fork on GitHub causes a merge. Will not use it again. Hopefully CI works now, just need to fix the coverage issue

benjeffery:
> Ah, I realised that using Sync Fork on GitHub causes a merge. Will not use it again. Hopefully CI works now, just need to fix the coverage issue

If you need to rebase to main and the PR is other wise complete you can use "_mergifyio rebase" replacing the _ with a @.

duncanMR:
I think I've covered all the cases where `tsk_tree_position_t` was used instead of `tsk_tree_t`. I haven't figured out how to resolve the codecov warning: the tests implicitly check if `tsk_tree_position_init` returns a nonzero value within `tsk_tree_init`, and I've added checks that all the initialised values are correct.

c/tskit/trees.c
```diff
@@ -7007,10 +7002,10 @@ tsk_treeseq_kc_distance(const tsk_treeseq_t *self, const tsk_treeseq_t *other,
     if (ret != 0) {
         goto out;
     }
-    tsk_tree_position_next(&tree_pos[0]);
-    tsk_bug_assert(tree_pos[0].index == 0);
+    tsk_tree_position_next(&trees[0].tree_pos);
```
jeromekelleher, line 7005: Ah no, that's not the point - we need to use the *internal* calls to tree_pos_next. Calling externally like this should break all sorts of things.

jeromekelleher (commented):
I think you're doing this the wrong way around. You need to update the tree class to use the ``tree_pos_t`` first, *then* alter external code to use it directly.

jeromekelleher:
I would drop the last two commits for now @duncanMR, and focus on changing tree methods to use the tree_pos

duncanMR:
I've moved over all the tree positioning functions to use `tree_pos`; for now, I've made `seek_from_null` just call `tsk_tree_next` repeatedly for now, until we implement `seek_forward`. All tests are passing except for 21 in the class `tests/test_highlevel.py::TestSeekDirection`. I've narrowed down the cause: the `left_child` and `right_sib` arrays are calculated incorrectly when calling `prev` from the null state:

```python
from tests import tsutil
from tests import test_tree_positioning
ts = tsutil.all_trees_ts(3)
tree_state = test_tree_positioning.StatefulTree(ts)
tree_state.prev()
print(tree_state)
```
```
parent: [3, 3, 4, 4, -1]
position:
	index: 3
	interval: Interval(left=3.0, right=4.0)
	direction: -1
	in_range: EdgeRange(start=8, stop=4, order=array([0, 5, 2, 6, 4, 8, 7, 3, 1], dtype=int32))
	out_range: EdgeRange(start=8, stop=8, order=array([0, 2, 4, 5, 8, 1, 6, 3, 7], dtype=int32))
```
So we will insert edges 1 ($3 \rightarrow 0$), 3 ($3 \rightarrow 1$), 7 ($4 \rightarrow 2$) and 8 ($4 \rightarrow 3$) in that order.
Because edge 7 is inserted first, we set `left_child(4) = 2` and `right_sib[2] = 3`. This is incorrect, because it is supposed to be `left_child(4) = 3` and `right_sib(3)=2`, as we can see from the tree plot:

image

I'm still trying to figure out how the system for node ordering works. I think we either need to change the `in_range` calculation, or we have to revise how `left_child` and `right_sib` are updated, but I'm not sure yet.

jeromekelleher:
Hmm, the ordering of children for a parent is documented as arbitrary, and we can change it. If that's the only problem then I wouldn't consider it a bug.

Is it just those tests expect a particular ordering?

duncanMR:
Yes, it seems that `test_highlevel.py::TestSeekDirection` is the only test class that checks the child and sib arrays directly with the `assert_trees_identical` function. The tests fail because the child and sib arrays are calculated differently when reaching the last tree in a sequence with `prev` or `next`. I get that the ordering is arbitrary, but I would have thought we'd need it to be independent of the tree positioning direction?

jeromekelleher:
> I get that the ordering is arbitrary, but I would have thought we'd need it to be independent of the tree positioning direction?

It's always been conditioned on the precise order in which edges are inserted and removed. For example, you will  slightly different ``left_sib`` arrays when you seek to the same tree in the left and right directions. There's no avoiding this if we want to keep it efficient for large polytomies.

So, just to clarify, the arrays are correct here, but they are ordered slightly differently to the currently library code? I'm OK with that, if that's the case.

duncanMR:
> > I get that the ordering is arbitrary, but I would have thought we'd need it to be independent of the tree positioning direction?
> 
> It's always been conditioned on the precise order in which edges are inserted and removed. For example, you will slightly different `left_sib` arrays when you seek to the same tree in the left and right directions. There's no avoiding this if we want to keep it efficient for large polytomies.
> 
> So, just to clarify, the arrays are correct here, but they are ordered slightly differently to the currently library code? I'm OK with that, if that's the case.

I see. Yes, the `right_sib` and `left_child` have just been permuted at one position; here, Tree 2 is what the current library code outputs and Tree 1 is the result of calling `prev` from the null tree, which used to give the same result prior to my changes:

```python
Tree 1 left_child: [-1 -1 -1  0  2  4]
Tree 2 left_child: [-1 -1 -1  0  3  4]
Tree 1 left_sib:   [-1  0 -1  2 -1 -1]
Tree 2 left_sib:   [-1  0  3 -1 -1 -1]
```

The trees are identical otherwise. Should I just remove the checks for whether the left/right child/sib arrays are identical for this test?

jeromekelleher:
Yes for now can just comment out those tests so we can see how the rest of the suite is doing (would come up with a better solution before merging)

duncanMR:
I've got `seek_from_null` working using `seek_forward` and `seek_backward`. I just need to add some tests now and check that the performance hasn't regressed.

duncanMR:
I had some trouble replicating Kevin Thornton's performance plots for `seek_from_null`, since my timing data for each seek has higher variance and I don't have his benchmarking code to compare with. Strangely, the times I get for the new `tree_pos` implementation have tended to have higher variance; sometimes, a single seek takes an order of magnitude longer than normal. Overall, it seems like the new implementation is comparable in performance to the old method, but slightly slower on average due to the occasionally long seek times:

```python
import stdpopsim
import tskit
import time
import pandas as pd

species = stdpopsim.get_species("HomSap")
model = species.get_demographic_model("OutOfAfrica_3G09")
contig = species.get_contig("chr1", length_multiplier=0.2)
samples = {"YRI": 0, "CHB": 0, "CEU": 2e4}
engine = stdpopsim.get_engine("msprime")
test_ts = engine.simulate(model, contig, samples, seed=1)

def seek_from_null_sweep(ts, method):
    data = []

    for n in range(0, ts.num_trees - 1, 500):
        tree = tskit.Tree(ts)
        before = time.perf_counter()
        tree.seek(n)
        data.append([n, duration, method])

    # Convert the list to a DataFrame
    df = pd.DataFrame(data, columns=['Index', 'Time', 'Method'])
    return df

#lib_null_df = seek_from_null_sweep(test_ts, method="lib")
tree_pos_null_df = seek_from_null_sweep(test_ts, method="tree_pos")
```

image

We can't determine a clear difference between the methods from a single run, but if I repeatedly time how long it takes to seek from null to various points along the genome, the difference between the two implementations does seem significant in some cases. Here, I seeked to 1000 indices in the HGDP and msprime simulated trees, and only 10 indices in the sc2ts trees. This how I timed it:

```python
def test_seek_from_null(ts, num_runs, num_indices):
    times = np.zeros(num_runs)
    tree = tskit.Tree(ts)
    indices = np.linspace(0, ts.num_trees - 1, num_indices, dtype=int)
    for run in range(num_runs):
        before = time.perf_counter()
        for n in indices:
            tree.clear()
            tree.seek(n)
        times[run] = time.perf_counter() - before
    return np.mean(times), np.std(times)
```

Here are the results, including the tests for other tree positioning methods (the times are averaged over 30 runs for `seek_from_null` and 50 runs for the other tests):

image

@jeromekelleher Do you think this warrants further investigation / profiling? My test setup is far from ideal: background processes and the complexity of WSL + Jupyter + Windows causes significant variation in the timings I calculate. In my first experiments, the performance of the new method was actually worse, but I couldn't replicate that later on.

molpopgen:
I don't totally remember the details but my benchmarks were probably all done using a prototype written using tskit-rust and are now lost to history.

jeromekelleher:
Interesting - what's the percentage difference in the average ``seek_from_null`` times? I guess we should run it through ``perf`` or something to see where the time is being spent in the low-level code.

c/tskit/trees.c
```diff
@@ -5485,164 +5635,73 @@ tsk_tree_insert_edge(tsk_tree_t *self, tsk_id_t p, tsk_id_t c, tsk_id_t edge_id)
...
-    if (self->index == -1) {
-        ret = tsk_tree_first(self);
-    } else if (self->index < num_trees - 1) {
-        ret = tsk_tree_advance(self, TSK_DIR_FORWARD, tables->edges.right,
-            tables->indexes.edge_removal_order, &self->right_index, tables->edges.left,
-            tables->indexes.edge_insertion_order, &self->left_index);
+    valid = tsk_tree_position_next(&self->tree_pos);
+
+    if (valid) {
+        for (j = self->tree_pos.out.start; j != self->tree_pos.out.stop; j++) {
```
jeromekelleher, line 5694: I think these pointer references might be adding up. Try something like
```
tree_pos = self->tree_pos;
for (j = tree_pos.out.start; j != tree_pos.out.stop; j++) {
...
}```

**Review by jeromekelleher — COMMENTED** [review]
There's an obvious pointer dereferencing that could be leading to the perf difference. Let's avoid using ``self->`` within loops.

File: c/tskit/trees.c (hunk 2 above)
Comment (duncanMR) [reply c/tskit/trees.c:5694]: Looks like this fixes the problem. I'm just running a better perftest script now to make sure

**Comment by duncanMR** [comment]
I wrote a better performance test that calculates mean seek times for each of a subset of indices

```python
def test_seek_from_null(ts, ts_name, num_runs, num_indices, method):
    times_matrix = np.zeros((num_runs, num_indices))
    indices = np.linspace(0, ts.num_trees - 1, num_indices, dtype=int)
    
    for run in range(num_runs):
        for i, n in enumerate(indices):
            tree = tskit.Tree(ts)
            before = time.perf_counter()
            tree.seek(n)
            after = time.perf_counter()
            times_matrix[run, i] = (after - before) * 1000000
    
    mean_times = np.mean(times_matrix, axis=0)
    std_times = np.std(times_matrix, axis=0)

    # Create a DataFrame to return
    results_df = pd.DataFrame({
        'index': indices,
        'mean': mean_times,
        'sd': std_times,
        'method': method,
        'ts': ts_name
    })

    return results_df
```
I computed the seek times for three TS, starting with a huge simulated ts:

image

The new algorithm performs substantially better toward the end of the sequence (results are averaged over 200 runs)

image

The variance in seek times is extremely high in the HGDP case for both methods, but the implementations perform within measurement error for each other

image

I had to reduce the number of runs to 50 for sc2ts because it takes so long. It's another clear win for the new implementation

image

I've updated the LSHMM and KC distance code. I removed the `tsk_tree_position_step` function because it is actually useless (we never want to update `tsk_tree_position` separately to the tree now). I think it's ready for review now. I just need to rebase

duncanMR:
@Mergifyio rebase

c/tskit/trees.c
```diff
@@ -5673,111 +5735,57 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
-    const double *restrict breakpoints = treeseq->breakpoints;
-    const tsk_id_t *restrict insertion = tables->indexes.edge_insertion_order;
-    const tsk_id_t *restrict removal = tables->indexes.edge_removal_order;
-
-    // NOTE: it may be better to get the
-    // index first and then ask if we are
-    // searching in the first or last 1/2
-    // of trees.
-    j = -1;
+    double interval_left = self->interval.left;
```
jeromekelleher, line 5747: You're not actually using these values that ``interval_left`` and ``interval_right`` are being initialised to. Better to just declare,
```c
double interval_left, interval_right;
```

c/tskit/trees.c
```diff
@@ -5673,111 +5735,57 @@ tsk_tree_position_in_interval(const tsk_tree_t *self, double x)
...
-                break;
-            }
-            if (x >= edge_left[e] && x < edge_right[e]) {
-                p = edge_parent[e];
-                c = edge_child[e];
-                tsk_tree_insert_edge(self, p, c, e);
+        ret = tsk_tree_position_seek_forward(&self->tree_pos, index);
+        if (ret != 0) {
+            goto out;
+        }
```
jeromekelleher, line 5764: It took me a couple of minutes to realise that because we're seeking from the null tree there are no edges to remove. It's worth adding a comment to that effect here I think.

jeromekelleher (commented):
Fantastic. I've gone through this carefully and just spotted one detail. Ready to merge after that's fixed.

Can you keep the scripts etc that you use here handy, as it would be nice to do a tskit .dev news item with these perf numbers when we release this. Once you've implemented the seek_forward/backward functionality I think we'll want to do a joint Python and C API release, as the perf improvements are significant.

duncanMR:
> Fantastic. I've gone through this carefully and just spotted one detail. Ready to merge after that's fixed.
> 
> Can you keep the scripts etc that you use here handy, as it would be nice to do a tskit .dev news item with these perf numbers when we release this. Once you've implemented the seek_forward/backward functionality I think we'll want to do a joint Python and C API release, as the perf improvements are significant.

Thanks for the feedback; I've corrected both issues and rebased. I'll definitely keep my scripts handy; I'll be using them to assess the new `seek` functionality soon.

jeromekelleher (approved):

PR #2880, opened 2023-12-13
Add c/CMakeLists.txt

lukashuebner:
I've been using this `CMakeLists.txt` using the following code in my code for a while now:

```
add_subdirectory("extern/tskit/c" "extern/tskit")
target_link_libraries(${MY_TARGET} INTERFACE tskit) # ${MY_TARGET} is header-only, thus INTERFACE
```

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -1,0 +1,68 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(
+    tskit
+    DESCRIPTION "Succinct tree sequences"
+    LANGUAGES C
+)
+
+# -- warning flags --
+list(
+    APPEND
+    TSKIT_C_FLAGS
+    "-Wall"
+    "-Wextra"
+    "-Werror"
+    "-Wpedantic"
+    "-W"
+    "-Wmissing-prototypes"
+    "-Wstrict-prototypes"
+    "-Wconversion"
+    "-Wshadow"
+    "-Wpointer-arith"
+    "-Wcast-align"
+    "-Wcast-qual"
+    "-Wwrite-strings"
+    "-Wnested-externs"
+    "-fshort-enums"
+    "-fno-common"
+)
+
+# -- kastore --
+add_library(kastore STATIC)
+target_compile_features(kastore PRIVATE c_std_99)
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Thanks @lukashuebner, this will be really helpful for CMake users!

I wonder if it belongs in the tsk-tbuild-example repo though? It would naturally fit into a ``CMake``subdirectory

https://github.com/tskit-dev/tskit-build-examples

lukashuebner:
Imho, the right place would be in this repo. Otherwise the user would need to maintain a list of `tskit`'s source files, dependencies and build settings themself, which would violate the encapsulation principle. If you prefer the tskit-build-examples repo, though, I can create a PR there.

jeromekelleher:
I see your point @lukashuebner, but I don't really want to get into maintaining multiple build-systems within tskit. We've chosen meson here as the libraries build system, and the specific point of the tskit-build-examples repo is to give examples of how to build applications using tskit using different build systems. As the dev docs say:

> There are many different build systems and approaches to compiling code, and so it’s not possible to give definitive documentation on how tskit should be included in downstream projects. Please see the build examples repo for some examples of how to incorporate tskit into different project structures and build systems.

So, it would be great if you could add this to the build examples for other CMake users!

lukashuebner:
See https://github.com/tskit-dev/tskit-build-examples/pull/7

jeromekelleher:
Great, thanks.

Out of interest, did you realise that we had this build-example repo? How was your experience of getting CMake running, and did you find the dev-docs useful?

lukashuebner:
I created the `CMakeLists.txt` a few months back, so I'm not 100 % sure, but I think I didn't know the repo with the build examples existed. Getting CMake running was pretty straightforward though – your `meson.build` file is easy enough to read and translate to CMake. (I never used Meson and would consider myself an average CMake user.)

jeromekelleher:
Maybe we need to signpost this a bit more prominently, say in the README.md in the c directory? 

Do you build your applications "in tree" here in the tskit source, or do you then install the headers and lib files to your file system somewhere?

benjeffery:
I'm assuming we're not merging this as it went to the example repo - please reopen if that isn't the case.

PR #2900, opened 2024-02-08
Testing CI

benjeffery:
Having some local failures on 3.10 - checking other versions on CI


benjeffery:
Looks like an dep update, and the CI versions are ok. Will need to do a dep update as it has been a while.

PR #2901, opened 2024-02-08
Error when the VCF will have a zero position site

### LINKED ISSUE #2838: `write_vcf` returning position 0
hey team-

working with @mufernando on some stdpopsim coding we stumbled upon a bug with `write_vcf`. using the default `position_transform` value, `write_vcf` can return position `0` which isn't allowed under the [vcf spec](https://samtools.github.io/hts-specs/VCFv4.2.pdf) i believe. e.g., this tree sequence

```
print(ts.tables.sites)
╔═════╤════════╤═══════════════╤════════╗
║id   │position│ancestral_state│metadata║
╠═════╪════════╪═══════════════╪════════╣
║0    │       0│              C│        ║
║1    │     534│              C│        ║
║2    │     991│              T│        ║
║3    │    1188│              G│        ║
║4    │    1281│              C│        ║
║5    │    1301│              A│        ║
```
will return the following vcf

```
##fileformat=VCFv4.2
##source=tskit 0.5.5
##FILTER=<ID=PASS,Description="All filters passed">
##contig=<ID=1,length=5000000>
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	tsk_0	tsk_1	tsk_2	tsk_3	tsk_4	tsk_5	tsk_6	tsk_7	tsk_8	tsk_9	tsk_10	tsk_11	tsk_12	tsk_13	tsk_14	tsk_15	tsk_16	tsk_17	tsk_18	tsk_19
1	0	0	C	G	.	PASS	.	GT	0|0	0|0	0|0	0|0	0|0	0|0	0|0	1|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0
1	534	1	C	G	.	PASS	.	GT	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|0	0|1	0|0	0|0	0|0	0|0	0|0
1
.
.
.
```
i'm guessing we should not have been using the default position_transform here but instead legacy? either way i think this is a bug

benjeffery:
Fixes #2838

Note that this is a fairly breaking change that we should think about, given that the default is for msprime output to require the new flag to `write_vcf`.

### DIFF
--- a/python/tskit/vcf.py
+++ b/python/tskit/vcf.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# Copyright (c) 2019-2022 Tskit Developers
+# Copyright (c) 2019-2024 Tskit Developers
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -63,6 +63,7 @@ class VcfWriter:
         site_mask,
         sample_mask,
         isolated_as_missing,
+        allow_position_zero,
     ):
         self.tree_sequence = tree_sequence
         self.contig_id = contig_id
@@ -110,6 +111,20 @@ class VcfWriter:
             if not callable(sample_mask):
                 sample_mask = np.array(sample_mask, dtype=bool)
                 self.sample_mask = lambda _: sample_mask
+
+        # The VCF spec does not allow for positions to be 0, so we error if one of the
+        # transformed positions is 0 and allow_position_zero is False.
+        if not allow_position_zero and np.any(
+            self.transformed_positions[~site_mask] == 0
+        ):
+            raise ValueError(
+                "A variant position of 0 was found in the VCF output, this is not "
+                "fully compliant with the VCF spec. If you still wish to write the VCF "
+                'please use the "allow_position_zero" argument to write_vcf. '
+                "Alternatively, you can increment all the positions by one using "
+                '"position_transform = lambda x: 1 + x" or coerce the zero to one with '
+                '"position_transform = lambda x: np.fmax(1, x)"'
+            )
 
     def __make_sample_mapping(self, ploidy, individuals):
         """
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,6 +2,13 @@
 [0.5.7] - 2023-XX-XX
 --------------------
 
+**Breaking Changes**
+
+- The VCF writing methods (`ts.write_vcf`, `ts.as_vcf`) now error if a site with
+  position zero is encountered. The VCF spec does not allow zero position sites.
+  Suppress this error with the `allow_position_zero` argument.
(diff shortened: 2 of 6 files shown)


molpopgen:
One thought experiment is: what if some app using tskit is already generating all site positions on 1 <= x <  seq length + 1?

benjeffery:
> One thought experiment is: what if some app using tskit is already generating all site positions on 1 <= x < seq length + 1?

That wouldn't be a valid tree sequence as all positions have to be less than sequence length.

molpopgen:
> That wouldn't be a valid tree sequence as all positions have to be less than sequence length.

Oops -- it would be if I'd written it correctly (w/o the +1).  But my point should have been: we have no firm requirement that the minimum position actually used is zero.

benjeffery:
I'm not sure I get what you mean - you can have a tree sequence with no sites? Maybe you mean we have no firm specification for how the reference sequence in the tree sequence maps onto the position field. We don't even have a requirement that the ref seq length is equal to  `sequence_length-1`.

molpopgen:
> I'm not sure I get what you mean - you can have a tree sequence with no sites? Maybe you mean we have no firm specification for how the reference sequence in the tree sequence maps onto the position field. We don't even have a requirement that the ref seq length is equal to `sequence_length-1`.

Imagine that someone only considers positions from [10, seqlen).  Their "genome" starts at position 10, not 0.  That is a valid tree sequence and they can choose their seqlen so that the max allowed site position matches whatever they have in mind, say 100.  So they are modeling a gene segment from positions [10, 100] for some reason using a table collection with seqlen of 101.

This is a valid use of the API.  What would this PR do to this use case?

python/tskit/vcf.py
```diff
@@ -111,6 +112,20 @@ def __init__(
                 sample_mask = np.array(sample_mask, dtype=bool)
                 self.sample_mask = lambda _: sample_mask
 
+        # The VCF spec does not allow for positions to be 0, so we error if one of the
+        # transformed positions is 0 and allow_position_zero is False.
```
jeromekelleher, line 116: Sufficient to check first site, I think?

jeromekelleher (commented):
Generally looks good to me. I hope the noise in our test suite isn't a general indication...

jeromekelleher:
This is just checking to see if the first position is zero, nothing else. The seqlen stuff was a digression on the thread

python/tskit/vcf.py
benjeffery: I think we need to check them all, as the transform function could have shuffled the order.

petrelharp:
I love it (as much as is possible for a weird VCF hack). Nice solution.

mufernando:
Just pinging this because this issue tripped me up again!

Issue #2829, opened 2023-08-30
Python interface for 2-locus statistics


lkirk:
Implement a python interface to the 2-locus statistics so that we may begin more rigorous testing.

lkirk:
After talking with @apragsdale, I think that the LD interface will differ from the other stats interfaces. My understanding is that tskit typically groups stats by name and allows parameters for a given mode to be passed through a single entrypoint `ts.stat(params, mode="site/branch/node")`. Because the parameters for each mode differ so drastically, we propose that we instead group the stats by mode in the API instead of by stat name. We've sketched out what this might look like:

LD Matrix
Only acts on sites, no branch stats.
```python
ts.ld_matrix(summary_func='r2', sites=[[0, 1, 2], [2, 3]], sample_sets=[...])
output_shape = [n_sites, m_sites, n_sample_sets]
```

LD scores
Mode specifies whether or not we're comparing sites x branches or sites x sites
```python
ts.ld_scores(summary_func='r2', sites=[0, 2, 3], max_dist=100.5, sample_sets=[...], mode="site/branch")
output_shape = [n_sites, n_sample_sets]
```

LD Decay
Summarize LD decay for branches or sites. The distance bins will be applied within a specific window.
```python
ts.ld_decay(summary_func='r2', windows=[...], distance_bins=[...], mode="site/branch")
output_shape = [n_windows, n_distance_bins, n_sample_sets]
```

@petrelharp @jeromekelleher what are your thoughts about organizing things in this way?

apragsdale:
To expand on the motivation here a bit: I don't think that the LD matrix vs scores vs decay represent different modes, in the way that branch/site/node are different modes to compute statistics (like diversity).

But rather, matrix/scores/decay are different two-locus summaries of the data with return values that have different shapes and interpretations, but the same shape and interpretation within functions for different summary functions (like r^2, r, or D^2).

These different functions can (sometimes) be used with either site or branch mode (except for the LD matrix, for which only "site" makes sense). Within each of these larger LD functions, you may wish to compute the LD matrix, for example, using many different summary functions (`'r2'`, `'D2'`, `'r'`, `'D'`, `'r*r'` across populations, `'D*D'`, and many others), so it seemed to make a lot of sense to move the summary functions inside these broader matrix/scores/decay functions as an argument, to keep the API more organized.

[...]

PR #2902, opened 2024-02-09
Python LD Matrix API

lkirk:
## Description

As described in #2829, I've implemented the python API for the ld_matrix code. It should cover everything we described for the sites mode (barring the one comment I recently left on that issue). The test coverage handles odd edge cases (in the form of the mutated topology examples), the happy path, as well as input validation.

Do I need to create documentation before this can go out or can I handle it in a subsequent PR?

# PR Checklist:

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2668,21 +2668,17 @@ tsk_treeseq_two_locus_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sampl
         ret = TSK_ERR_MULTIPLE_STAT_MODES;
         goto out;
     }
-    if (state_dim < 1) {
-        ret = TSK_ERR_BAD_STATE_DIMS;
-        goto out;
-    }
     // TODO: impossible until we implement branch/windows
     // if (result_dim < 1) {
     //     ret = TSK_ERR_BAD_RESULT_DIMS;
     //     goto out;
     // }
-
     ret = tsk_treeseq_check_sample_sets(
         self, num_sample_sets, sample_set_sizes, sample_sets);
     if (ret != 0) {
         goto out;
     }
+    tsk_bug_assert(state_dim > 0);
     ret = sample_sets_to_bit_array(
         self, sample_set_sizes, sample_sets, num_sample_sets, &sample_sets_bits);
     if (ret != 0) {
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -9861,6 +9861,130 @@ out:
 }
 
 static PyObject *
+TreeSequence_ld_matrix(TreeSequence *self, PyObject *args, PyObject *kwds,
+    two_locus_count_stat_method *method)
+{
+    PyObject *ret = NULL;
+    static char *kwlist[]
+        = { "sample_set_sizes", "sample_sets", "row_sites", "col_sites", "mode", NULL };
+
+    PyObject *row_sites = NULL;
+    PyObject *col_sites = NULL;
+    PyObject *sample_set_sizes = NULL;
+    PyObject *sample_sets = NULL;
+    PyArrayObject *sample_set_sizes_array = NULL;
+    PyArrayObject *sample_sets_array = NULL;
+    PyArrayObject *row_sites_array = NULL;
+    PyArrayObject *col_sites_array = NULL;
+    PyArrayObject *result_matrix = NULL;
+    npy_intp result_shape[3];
+    char *mode = NULL;
+    tsk_size_t num_sample_sets;
+    tsk_flags_t options = 0;
+    int err;
+
+    if (TreeSequence_check_state(self) != 0) {
+        goto out;
+    }
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OOOOs", kwlist, &sample_set_sizes,
+            &sample_sets, &row_sites, &col_sites, &mode)) {
+        goto out;
+    }
+    if (parse_stats_mode(mode, &options) != 0) {
+        goto out;
+    }
+    if (parse_sample_sets(sample_set_sizes, &sample_set_sizes_array, sample_sets,
+            &sample_sets_array, &num_sample_sets)
+        != 0) {
+        goto out;
+    }
+    row_sites_array = (PyArrayObject *) PyArray_FROMANY(
+        row_sites, NPY_INT32, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (row_sites_array == NULL) {
+        goto out;
+    }
+    col_sites_array = (PyArrayObject *) PyArray_FROMANY(
+        col_sites, NPY_INT32, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (col_sites_array == NULL) {
+        goto out;
+    }
+
+    result_shape[0] = PyArray_DIM(row_sites_array, 0);
+    result_shape[1] = PyArray_DIM(col_sites_array, 0);
+    result_shape[2] = num_sample_sets;
+    result_matrix = (PyArrayObject *) PyArray_ZEROS(3, result_shape, NPY_FLOAT64, 0);
+    if (result_matrix == NULL) {
+        goto out;
+    }
+
+    // clang-format off
+    Py_BEGIN_ALLOW_THREADS
+    err = method(self->tree_sequence, num_sample_sets,
+        PyArray_DATA(sample_set_sizes_array), PyArray_DATA(sample_sets_array),
+        result_shape[0], PyArray_DATA(row_sites_array), result_shape[1],
+        PyArray_DATA(col_sites_array), options, PyArray_DATA(result_matrix));
+    Py_END_ALLOW_THREADS
+        // clang-format on
+
+        if (err != 0)
+    {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = (PyObject *) result_matrix;
+    result_matrix = NULL;
+out:
+    Py_XDECREF(row_sites_array);
+    Py_XDECREF(col_sites_array);
+    Py_XDECREF(sample_set_sizes_array);
+    Py_XDECREF(sample_sets_array);
+    Py_XDECREF(result_matrix);
+    return ret;
+}
+
+static PyObject *
+TreeSequence_D_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_D);
+}
+
+static PyObject *
+TreeSequence_D2_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_D2);
+}
+
+static PyObject *
+TreeSequence_r2_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_r2);
+}
+
+static PyObject *
+TreeSequence_D_prime_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_D_prime);
+}
+
+static PyObject *
+TreeSequence_r_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_r);
+}
+
+static PyObject *
+TreeSequence_Dz_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_Dz);
+}
+
+static PyObject *
+TreeSequence_pi2_matrix(TreeSequence *self, PyObject *args, PyObject *kwds)
+{
+    return TreeSequence_ld_matrix(self, args, kwds, tsk_treeseq_pi2);
+}
+
+static PyObject *
 TreeSequence_get_num_mutations(TreeSequence *self)
 {
     PyObject *ret = NULL;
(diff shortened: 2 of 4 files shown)


lkirk:
I'm a little bit confused by the code coverage annotations in the `_tskitmodule.c` file, I'm pretty sure that there are some false positives. I can do a more thorough check, I was wondering if it's a known issue.

I went ahead and made an initial cut of documentation so that this python code at least has some barebones documentation. I can add demos and examples in a few days or in a subsequent PR.

I've attached a [pdf render](https://github.com/tskit-dev/tskit/files/14236436/python.pdf) of the docs I've created so far. The `ld_matrix` docstring makes it into the TreeSequence methods. I've also added a section inside of `multi-site statistics`.

jeromekelleher (commented):
Generally looks good, although I haven't dug deep into the Python C side of things.

I'm not sure I agree with the plumbing setup though: I think we should have functions methods called ``D_matrix``, D2_matrix`` etc rather than have the stat specified as a string (although that may be a useful sugar to add later).

For the _tskitmodule.c coverage, you need to add some basic tests to the ``test_lowlevel.py`` file which stresses the interface. The rest of the suite doesn't count towards coverage of the C code.

For testing, let's stay miles away from ``test_tree_stats.py``, it's horrifying.  Please use test_divmat.py as an example for how we're hoping to structure the tests from now on.

This is quite a large PR again, it really would help to try and keep things to smaller so I can give feedback earier in the development cycle. Or, if the changes are necessarily large, opening a draft PR at an early stage to I can comment on earlier versions.

docs/stats.md
```diff
@@ -655,6 +655,95 @@ framework which has the same calling conventions as the single site stats,
...
+
+  This method assigns equal weight to each of the possible haplotypes for a
+  given pair of sites.
+
+Generally speaking, we use the `hap_weighted` method of normalisation for
+unpolarised ratio statistics, such as {math}`r^2` and {math}`D^{\prime}`.
+We use `total_weighted` to normalise the rest.
+
+The two-locus statistics all take frequencies of haplotype configurations as
+input. The function signature below ({math}`f(f_a, f_b, f_{ab})`) describes a
```
apragsdale, line 696: Small thing on style: you use "f" for both function and allele/haplotype frequencies. Above, you use capital F for the summary function - could that be used here, to help with readability?

apragsdale (commented):
Small comment.

Agreed that smaller PRs would make it quite a bit more bite-sized.

About the method calls - do you envision splitting out each summary function to its own method call, but keeping the `ld_matrix(..., stat="r")` high-level interface? There are relatively few stats implemented right now, maybe just half a dozen, but there are many two-locus statistics that user could realistically want, especially as we start to think about multi-population stats. The single LD matrix function at the high level would keep things compact and clean, perhaps?

lkirk:
> I'm not sure I agree with the plumbing setup though: I think we should have functions methods called `D_matrix`, D2_matrix`` etc rather than have the stat specified as a string (although that may be a useful sugar to add later).

I see how you've suggested how to structure things, I think I agree, but I'm wondering if we'd expose all of the matrix methods outside of `_ll_tree_sequence` (in `trees.py`) or if we would wrap them all in `ld_matrix`. My sense from the discussion on #2829 is that we'd have one unified LD interface, since as @apragsdale points out, there are many more statistics to implement.

> For the _tskitmodule.c coverage, you need to add some basic tests to the `test_lowlevel.py` file which stresses the interface. The rest of the suite doesn't count towards coverage of the C code.

I see, I can put my input validation tests there to get the test coverage up.

> For testing, let's stay miles away from `test_tree_stats.py`, it's horrifying. Please use test_divmat.py as an example for how we're hoping to structure the tests from now on.

I remember you saying this. I'm sorry that I ended up importing it, I got focused on test cases and the various edge cases and got lost in the weeds on that one. I think I should just delete what I have and put the python tests into another PR altogether.

> This is quite a large PR again, it really would help to try and keep things to smaller so I can give feedback earier in the development cycle. Or, if the changes are necessarily large, opening a draft PR at an early stage to I can comment on earlier versions.

This is something I'm still trying to improve on. That's why I reached out on slack to gauge how to split things up. Can I propose that I remove all documentation and high-level python testing, focusing specifically on code that is present in `_tskitmodule.c`, `trees.py`, and the minimal things that will go into `test_lowlevel.py` and leave it at that?

In fact, I could just close this PR and open a smaller one if that'd be cleaner.

After we agree on this layer, I'll open up a PR for extensive python testing and documentation separately. Does that sound like more manageable chunks?

jeromekelleher:
> About the method calls - do you envision splitting out each summary function to its own method call, but keeping the ld_matrix(..., stat="r") high-level interface? There are relatively few stats implemented right now, maybe just half a dozen,

You're right @apragsdale - I think it would be nice to have a ``ld_matrix(..., stat="x")`` function. I'm just saying we should implement the dispatch layer in Python rather than C. A golden rule here is to not do anything complicated in C unless you have to.

jeromekelleher:
My suggestion here to move things along is to implement the minimal changes for ``_tskitmodule.c`` with corresponding tests in ``test_lowlevel.py``, which has a 1-1 mapping from C API functions to low-level Python-C functions. There's a bit of boilderplate involved, but in my experience it's very stable boilerplate - you do it once, make sure it's tested and then forget about it.

How does that sound?

lkirk:
> My suggestion here to move things along is to implement the minimal changes for `_tskitmodule.c` with corresponding tests in `test_lowlevel.py`, which has a 1-1 mapping from C API functions to low-level Python-C functions. There's a bit of boilderplate involved, but in my experience it's very stable boilerplate - you do it once, make sure it's tested and then forget about it.
> 
> How does that sound?

That sounds good. I'll rework this PR into what you've suggested.

c/tskit/trees.c
```diff
@@ -2668,21 +2668,20 @@ tsk_treeseq_two_locus_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sampl
...
-
     ret = tsk_treeseq_check_sample_sets(
         self, num_sample_sets, sample_set_sizes, sample_sets);
     if (ret != 0) {
         goto out;
     }
+    if (state_dim < 1) {
+        ret = TSK_ERR_BAD_STATE_DIMS;
+        goto out;
+    }
```
lkirk, line 2684: I reordered this to avoid a cryptic error about state dims. If a user passes in empty sample sets, they should get an error about empty sample sets instead of a "BAD_STATE_DIM" error

lkirk: In fact, I'm not even sure how this could be triggered, since the state dim is just `num_sample_sets`. I think this could be removed, what do you think?

lkirk:
I've simplified things and added the per-stat matrix methods to the `TreeSequence` object

jeromekelleher (commented):
Excellent, thanks @lkirk. I think we need to toughen up the input validation for the row and col arguments (I suspect it is possible to segfault the current code, which should never be possible).

Other than that, looks perfect :+1:

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,110 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+    if (parse_stats_mode(mode, &options) != 0) {
+        goto out;
+    }
+    if (parse_sample_sets(sample_set_sizes, &sample_set_sizes_array, sample_sets,
+            &sample_sets_array, &num_sample_sets)
+        != 0) {
+        goto out;
+    }
+
+    result_shape[0] = PyArray_DIM(row_sites, 0);
```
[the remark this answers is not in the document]
lkirk: You're right, it wasn't safe. I was controlling the input data in the outer python layer, but it makes sense to prevent segfaulting at this layer too. I've added an `asarray` call, which helpfully makes assertions about casting.

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,110 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+    result_shape[1] = PyArray_DIM(col_sites, 0);
+    result_shape[2] = num_sample_sets;
+    result_matrix = (PyArrayObject *) PyArray_ZEROS(3, result_shape, NPY_FLOAT64, 0);
+    if (result_matrix == NULL) {
+        goto out;
+    }
+
```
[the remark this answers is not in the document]
lkirk: `PyArray_DATA` is called inside of every usage of `Py_BEGIN_ALLOW_THREADS`, so I think it's probably fine. I skimmed through the numpy array documentation and couldn't find any mention of the GIL in the context of `PyArray_DATA`.

python/tests/test_lowlevel.py
```diff
@@ -1512,6 +1512,81 @@ def test_extend_edges_bad_args(self):
...
+
+        mode = "site"
+        sample_sets = ts.get_samples()
+        sample_set_sizes = np.array([len(sample_sets)], dtype=np.uint32)
+        row_sites = np.arange(ts.get_num_sites(), dtype=np.int32)
+        col_sites = row_sites
+
+        # happy path
+        a = stat_method(sample_set_sizes, sample_sets, row_sites, col_sites, mode)
+        assert a.shape == (10, 10, 1)
```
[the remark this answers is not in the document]
lkirk: I've added some more test cases that have bad types. I've got a test case with a string and with a `uint32` (which we cannot cast to `int32`). Is that what you were talking about?

Now we've got a test with lists as well as numpy arrays

lkirk:
Okay, I think that addresses your concerns. The one thing I haven't been able to do (mostly because I'm not sure how) is to leak check this code. Is that even possible? I don't see any CI tests for something like that.

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,128 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+        if (err != 0)
+    {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = (PyObject *) result_matrix;
+    result_matrix = NULL;
+out:
+    Py_XDECREF(sample_set_sizes_array);
```
jeromekelleher, line 9937: Need to Py_XDECREF(col_sites_array) and row_sites_array too. Currently the code will leak memory.

It's a good idea to call this in a loop and watch the output of ``top``. Something like
```python
for _ in range(100000):
     ts._ll_tree_sequence.D_matrix(...)
```
and you'll see the memory usage grow.

Unfortunately there's no better way to detect memory leaks in this layer that I know of.

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,128 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+        goto out;
+    }
+    row_sites_array = (PyArrayObject *) PyArray_FROMANY(
+        row_sites, NPY_INT32, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (row_sites_array == NULL) {
+        goto out;
+    }
+    col_sites_array = (PyArrayObject *) PyArray_FROMANY(
+        col_sites, NPY_INT32, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (col_sites_array == NULL) {
```
jeromekelleher, line 9908: Not testing this case - need to pass in some bad arguments to ``col_sites``, like ``col_sites = ["asdf"]`` or something.

python/tests/test_lowlevel.py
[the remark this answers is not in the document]
jeromekelleher: Yep, looks pretty good. Just need to add some tests for ``col_sites`` now too

jeromekelleher (commented):
Looks good. Need to address the memory leak, and cover the ``col_sites`` argument and then we're done I think

jeromekelleher:
> The one thing I haven't been able to do (mostly because I'm not sure how) is to leak check this code. Is that even possible?

Yeah, this is a pain. Running in a loop and watching ``top`` is the best way I've found. Memray might do it now, though, I haven't tried it for this purpose.

lkirk:
Cleaned up memory leaks, added tests for column sites. Good catch on that memory leak, I have tried memray in the past, but hadn't thought to run it multiple times to accumulate a noticeable amount of leaking memory. That seemed to do the trick.

Here's a plot of memory over time *before* I fixed the leak:
image
And *after* the leak was fixed:
image

So, it seems like this isn't leaking memory anymore. Less satisfying than using valgrind, but I think I'm convinced.

I've preemptively squashed down the commits, but let me know if there's anything else to fix.

jeromekelleher (approved):
Superb - nice work @lkirk, this Python C API stuff is tricky and you nailed it:+1:

jeromekelleher:
Great to see this in.

Next step is to get the top-level Python interface in. I suggest we get the bare-bones in, without documentation first, so that we can be reasonably relaxed about the parameters etc, and then spend a period evaluating it performance and usablity-wise. Once we're happy, then we do the final documentation.

If we document prematurely, it prevents us pushing out quick bugfix releases on tskit until we're 100% happy with the final API. Better to not document initially, and evaluate.

Hopefully you can cherry-pick out the commits from your earlier branch to do this?

Sound good?

jeromekelleher:
Also great to know that memray works for this now, much better than staring at ``top``! Maybe we should add a note about this to the Python C API dev docs?

lkirk:
> Also great to know that memray works for this now, much better than staring at `top`! Maybe we should add a note about this to the Python C API dev docs?

Sure, I'll draft up a PR for that.

lkirk:
> Great to see this in.
> 
> Next step is to get the top-level Python interface in. I suggest we get the bare-bones in, without documentation first, so that we can be reasonably relaxed about the parameters etc, and then spend a period evaluating it performance and usablity-wise. Once we're happy, then we do the final documentation.
> 
> If we document prematurely, it prevents us pushing out quick bugfix releases on tskit until we're 100% happy with the final API. Better to not document initially, and evaluate.
> 
> Hopefully you can cherry-pick out the commits from your earlier branch to do this?
> 
> Sound good?

Absolutely, I'll move the previous commits to a new branch and open a PR. I think something that was a bit confusing to me when making contributions is the PR checklist, it sort of implies that everything should come as a package. Docs, tests, along with the implementation. It seems, at least for things that are a bit more involved, we want more of a piecemeal approach. I suppose I really should have been reading [this](https://tskit.dev/tskit/docs/stable/development.html#best-practices-for-development) in more detail :/. I see what you mean about documentation blocking releases: once documentation goes up, it's public.

PR #2906, opened 2024-02-19
Two locus user-facing python ld matrix API

lkirk:
## Description

This is an implementation of the user-facing python API for `ld_matrix` as discussed in #2829

I've added some tests that cover most edge cases. I think I'd like to add some tests that store literal matricies, but I wanted to get eyes on this before I added those final tests.

# PR Checklist:

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -2471,8 +2471,8 @@ get_mutation_samples(const tsk_treeseq_t *ts, const tsk_id_t *sites, tsk_size_t 
             for (n = 0; n < num_nodes; n++) {
                 node = nodes[n];
                 if (flags[node] & TSK_NODE_IS_SAMPLE) {
-                    tsk_bit_array_add_bit(
-                        &mut_samples_row, (tsk_bit_array_value_t) node);
+                    tsk_bit_array_add_bit(&mut_samples_row,
+                        (tsk_bit_array_value_t) ts->sample_index_map[node]);
                 }
             }
         }
@@ -2612,9 +2612,8 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
     int ret = 0;
     tsk_size_t i;
 
-    if (sites == NULL || num_sites == 0) {
-        ret = TSK_ERR_BAD_SITE_POSITION; // TODO: error should be no sites?
-        goto out;
+    if (num_sites == 0) {
+        return ret; // No need to verify sites if there aren't any
     }
 
     for (i = 0; i < num_sites - 1; i++) {
@@ -3638,9 +3637,10 @@ D_prime_summary_func(tsk_size_t state_dim, const double *state,
         double p_B = p_AB + p_aB;
 
         double D = p_AB - (p_A * p_B);
-        if (D >= 0) {
+        result[j] = 0;
+        if (D > 0) {
             result[j] = D / TSK_MIN(p_A * (1 - p_B), (1 - p_A) * p_B);
-        } else {
+        } else if (D < 0) {
             result[j] = D / TSK_MIN(p_A * p_B, (1 - p_A) * (1 - p_B));
         }
     }
--- a/python/tests/test_ld_matrix.py
+++ b/python/tests/test_ld_matrix.py
@@ -1,3 +1,27 @@
+# MIT License
+#
+# Copyright (c) 2023-2024 Tskit Developers
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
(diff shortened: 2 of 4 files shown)


c/tskit/trees.c
```diff
@@ -3638,9 +3637,10 @@ D_prime_summary_func(tsk_size_t state_dim, const double *state,
         double p_B = p_AB + p_aB;
 
         double D = p_AB - (p_A * p_B);
-        if (D >= 0) {
```
lkirk, line 3641: The purpose of this change is to avoid dividing by zero. When D is zero, the denominator becomes zero.

python/tests/test_ld_matrix.py
```diff
@@ -699,6 +735,210 @@ def r2_summary_func(
...
+) -> None:
+    """Summary function for the D statistic. We first compute the proportion of
+    AB, A, and B haplotypes, then we compute the D statistic, storing the outputs
+    in the result vector, one entry per sample set.
+
+    :param state_dim: Number of sample sets.
+    :param state: Counts of 3 haplotype configurations for each sample set.
+    :param result: Vector of length state_dim to store the results in.
+    :param params: Parameters for the summary function.
+    """
```
lkirk, line 749: Are these docstrings too much? I'm not sure how much they're really adding to the understanding of this code.

c/tskit/trees.c
jeromekelleher: Should the result be NaN then? We should probably mark this as a follow-up issue, these things can be tricky to nail down.

python/tests/test_ld_matrix.py
```diff
@@ -569,7 +598,7 @@ def two_site_count_stat(
 
 def sample_sets_to_bit_array(
     ts: tskit.TreeSequence, sample_sets: List[List[int]]
-) -> Tuple[np.ndarray, BitSet]:
+) -> Tuple[np.ndarray, np.ndarray, BitSet]:
```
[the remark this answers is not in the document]
jeromekelleher: Not worth getting rid of the typehints now you've gone to the trouble of adding them. I'd just favour simplicity and readability over mypy-correctness

python/tests/test_ld_matrix.py
jeromekelleher: Could just document one of the summary funcs, and leave the rest clear? It is helpful documenting the first one here

python/tests/test_ld_matrix.py
```diff
@@ -699,6 +735,210 @@ def r2_summary_func(
...
+        p_Ab = state[1, k] / n
+        p_aB = state[2, k] / n
+
+        p_A = p_AB + p_Ab
+        p_B = p_AB + p_aB
+
+        D = p_AB - (p_A * p_B)
+        denom = p_A * p_B * (1 - p_A) * (1 - p_B)
+
+        if denom == 0 and D == 0:
```
jeromekelleher, line 836: See point above about NaN results. We found for the other stats API that it was easier to just let the division by zero happen and return NaN where the stat is undefined. Using a capturewarnings stops annoying warnings happening.

python/tskit/trees.py
```diff
@@ -7567,6 +7567,65 @@ def __one_way_sample_set_stat(
...
+        sample_sets,
+        sites=None,
+        mode=None,
+    ):
+        if sample_sets is None:
+            sample_sets = self.samples()
+        if sites is not None and any(
+            not hasattr(a, "__getitem__") or isinstance(a, str) for a in sites
+        ):
+            raise ValueError("sites must be a list of lists, tuples, or ndarrays")
```
jeromekelleher, line 7582: Need to test this

jeromekelleher (commented):
LGTM - just needs a bit of extra testing on the input value checking.

We should open an issue to track dealing with undefined values,  if you disagree with letting the division by zero happen and returning NaN

c/tskit/trees.c
apragsdale: That's a good question. Should D' (and other ratio statistics like r and r^2) return zero if there is no variation, or NaN? I had thought those results should be zero at first, but now I'm not so sure... We could tackle this in a follow-up issue, like you say.

apragsdale (commented):
Thanks @lkirk! Great to see this moving along!

lkirk:
re: docstrings, there's already one present in the code for `r2`, so I left that one in place but got rid of all of the ones in this PR.

Input validation testing is done, all testable branches are covered.

I'll open an issue for further NaN discussions.

lkirk:
I'm a little confused by the results of the coverage suite. It's annotating lines on the diff, but the patch coverage is reported as 100%.

I've opened #2907 to further investigate the NaN behavior.

Squashed down preemptively, but let me know if there's anything else needed here.

python/tests/test_ld_matrix.py
```diff
@@ -798,36 +993,100 @@ def get_all_site_partitions(n):
...
+
+@pytest.mark.parametrize(
+    "ts",
+    [
+        ts
+        for ts in get_example_tree_sequences()
+        if ts.id not in {"no_samples", "empty_ts"}
+    ],
+)
+@pytest.mark.parametrize("stat", SUMMARY_FUNCS.keys())
```
jeromekelleher, line 1067: Nice test! I didn't think of the trick of filtering by ID for the outliers :+1:

jeromekelleher (approved):
Thanks @lkirk, looks great. CodeCov is clearly lying, so let's merge rather than try to figure out what it's confused about.

PR #2908, opened 2024-02-21
Dividing by zero should be NaN

### LINKED ISSUE #2907: Should 2-locus ratio statistics return NaN on 0/0?
When we encounter situations in $r$, $r^2$, and $D^{\prime}$ where we have $\frac{0}{0}$, should we return $0$ or NaN? @apragsdale brings up a good point: What is the limit behavior as we approach $\frac{0}{0}$?

Let's follow up on this to ensure correctness of the API.

lkirk:
Allow C and Python portions of the code to divide by zero natively, relying on their respective runtime implementations. Remove code that forces the two-locus statistics to return zero when we encounter 0/0 situations. In testing, it was ensured that some test cases return NaN values, which we diff appropriately.

Fixes #2907

# PR Checklist:


jeromekelleher (approved):
Nice :thumbsup:

Ready to merge after squash

lkirk:
Squashed and ready.

PR #2909, opened 2024-02-22
Add note about creating a reverse node map

hyanwong:
I'm always wanting to do this and forgetting the most efficient method (for instance, https://github.com/tskit-dev/tskit/discussions/2524#discussioncomment-3757002 does so via an inefficient sort). I think adding this code snippet should be generally useful. Since it's in the docs for the TableCollection method, I don't worry about bloating the documentation, which is mostly contained in the TreeSequence version.

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# Copyright (c) 2018-2023 Tskit Developers
+# Copyright (c) 2018-2024 Tskit Developers
 # Copyright (c) 2017 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


python/tskit/tables.py
```diff
@@ -3377,6 +3377,18 @@ def simplify(
         requirements to specify a valid tree sequence (but the resulting tables
         will).
 
+        .. note::
+            To invert the returned ``node_map``, that is, to obtain a reverse
+            mapping from the node ID in the output table to the node ID in
```
jeromekelleher, line 3382: typo here, one of these should be "input table"

python/tskit/tables.py
```diff
@@ -3377,6 +3377,18 @@ def simplify(
...
+            the output table, you can use::
+
+                rev_map = np.zeros_like(node_map, shape=simplified_ts.num_nodes)
+                kept = node_map >= 0
```
jeromekelleher, line 3386: Slightly more obvious as ``node_map != tskit.NULL``?

jeromekelleher (commented):
Nice, that's a really handy snippet.

hyanwong:
> Nice, that's a really handy snippet.

Thanks ❤️  Fixed my typo & changed your suggestion. Should be OK to merge.

jeromekelleher (approved):

Issue #2792, opened 2023-07-12
Characterise propagation up the tree performance with seeking


jeromekelleher:
#2786 added the new  ``tsk_tree_position_t`` class which takes care of tracking the edges that need to go in and out of a tree in order to transform it into a target tree. This works really well for the basic ``next`` and ``prev`` operations, and there is an implementation of ``seek_forward``, which gives a straightforward way of transforming a tree into any other tree in the sequence, touching the minimum number of edges. It's worth looking at some of the key bits of code here:

```python
       # The range of edges we need consider for removal starts
        # at the current right index and ends at the first edge
        # where the right coordinate is equal to the new tree's
        # left coordinate.
        j = right_current_index
        self.out_range.start = j
        # TODO This could be done with binary search
        while j < M and right_coords[right_order[j]] <= left:
            j += 1
        self.out_range.stop = j

        # The range of edges we need to consider for the new tree
        # must have right coordinate > left
        j = left_current_index
        while j < M and right_coords[left_order[j]] <= left:
            j += 1
        self.in_range.start = j
        # TODO this could be done with a binary search
        while j < M and left_coords[left_order[j]] <= left:
            j += 1
        self.in_range.stop = j

        self.interval.left = left
        self.interval.right = breakpoints[index + 1]
```
Here we fill in the ``in_range`` and ``out_range`` ranges of indexes into the left and right edge sortings which must be examined. We can be sure that we don't need to look at any edges outside of these ranges. Then, on the client side, we do something like this:
```python
       old_left, old_right = self.tree_pos.interval
        self.tree_pos.seek_forward(index)
        left, right = self.tree_pos.interval
        for j in range(self.tree_pos.out_range.start, self.tree_pos.out_range.stop):
            e = self.tree_pos.out_range.order[j]
            e_left = self.ts.edges_left[e]
            # We only need to remove an edge if it's in the current tree, which
            # can only happen if the edge's left coord is < the current tree's
            # right coordinate.
            if e_left < old_right:
                c = self.ts.edges_child[e]
                assert self.parent[c] != -1
                self.parent[c] = -1
            assert e_left < left
        for j in range(self.tree_pos.in_range.start, self.tree_pos.in_range.stop):
            e = self.tree_pos.in_range.order[j]
            if self.ts.edges_left[e] <= left < self.ts.edges_right[e]:
                c = self.ts.edges_child[e]
                p = self.ts.edges_parent[e]
                self.parent[c] = p
```

[...]

PR #2911, opened 2024-02-26
Implement efficient seeking from non-null trees using tree_pos

duncanMR:
## Description

Continuing from #2874, we want to finish moving over the tree-positioning code to use `tree_pos` efficiently. At the moment, `tsk.tree_seek` will either call `tsk_tree_seek_from_null` or `tsk_tree_seek_linear` depending on whether we are starting from the null tree or not. `seek_linear` repeatedly calls `next` or `prev` until it reaches the given position, with the direction being determined by which would cover the shortest distance.

As a first pass, I've implemented `tsk_tree_seek_forward` and `tsk_tree_seek_backward` and I've incorporated them into `tsk_tree_seek_linear`. We will need to revise some of the `test_highlevel.py` seek tests, because the direction we choose to seek along is different to the old approach in some cases. For example, we now seek forward to go from the first to the last tree in a sequence.

Curiously, my implementation passes all the C tests with no memory issues detected by Valgrind, and it also passes all the `test_highlevel.py` and `test_tree_positioning` tests except for the ones dependent on seeking direction. However, it has caused chaos with other Python tests, causing failures and segfaults in `test_stats.py` and `test_divmat.py` among others. The failing/crashing tests seem to be primarily be associated with LD calculations and divergence. 

I'm currently trying to determine whether the problems are due to an error in my implemention (most likely) or the subtle problems with the time ordering of inserted edges, discussed in #2792.

# PR Checklist:

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -6532,6 +6532,90 @@ out:
     return ret;
 }
 
+static int TSK_WARN_UNUSED
+tsk_tree_seek_forward(tsk_tree_t *self, tsk_id_t index)
+{
+    int ret = 0;
+    tsk_table_collection_t *tables = self->tree_sequence->tables;
+    const tsk_id_t *restrict edge_parent = tables->edges.parent;
+    const tsk_id_t *restrict edge_child = tables->edges.child;
+    const double *restrict edge_left = tables->edges.left;
+    const double *restrict edge_right = tables->edges.right;
+    double interval_left, e_left;
+    const double old_right = self->interval.right;
+    tsk_id_t j, e;
+    tsk_tree_position_t tree_pos;
+
+    ret = tsk_tree_position_seek_forward(&self->tree_pos, index);
+    if (ret != 0) {
+        goto out;
+    }
+    tree_pos = self->tree_pos;
+    interval_left = tree_pos.interval.left;
+
+    for (j = tree_pos.out.start; j != tree_pos.out.stop; j++) {
+        e = tree_pos.out.order[j];
+        e_left = edge_left[e];
+        if (e_left <= old_right) {
+            tsk_bug_assert(edge_parent[e] != TSK_NULL);
+            tsk_tree_remove_edge(self, edge_parent[e], edge_child[e]);
+        }
+        tsk_bug_assert(e_left < interval_left);
+    }
+
+    for (j = tree_pos.in.start; j != tree_pos.in.stop; j++) {
+        e = tree_pos.in.order[j];
+        if (edge_left[e] <= interval_left && interval_left < edge_right[e]) {
+            tsk_tree_insert_edge(self, edge_parent[e], edge_child[e], e);
+        }
+    }
+    tsk_tree_update_index_and_interval(self);
+out:
+    return ret;
+}
+
+static int TSK_WARN_UNUSED
+tsk_tree_seek_backward(tsk_tree_t *self, tsk_id_t index)
+{
+    int ret = 0;
+    tsk_table_collection_t *tables = self->tree_sequence->tables;
+    const tsk_id_t *restrict edge_parent = tables->edges.parent;
+    const tsk_id_t *restrict edge_child = tables->edges.child;
+    const double *restrict edge_left = tables->edges.left;
+    const double *restrict edge_right = tables->edges.right;
+    double interval_right, e_right;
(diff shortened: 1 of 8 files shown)


jeromekelleher:
That is interesting. I can see the LD calculator causing problems as that does a lot of seeking backward and forward (was the original motivation for bidirectional seeking). I don't see anything obvious wrong, but it must be something to do with sample counts.

One thing that we can at least make progress on is that I think we should keep the option of seeking linearly. So, we make a new option to ``seek``, ``TSK_SEEK_LINEAR`` which always uses the linear seek algorithm. We'll want to expose this to Python also, so there'll be a bit of plumbing involved. 

We should specify this option in the ld_calculator, as that is definitely somewhere that linear seeking makes sense.

duncanMR:
> That is interesting. I can see the LD calculator causing problems as that does a lot of seeking backward and forward (was the original motivation for bidirectional seeking). I don't see anything obvious wrong, but it must be something to do with sample counts.
> 
> One thing that we can at least make progress on is that I think we should keep the option of seeking linearly. So, we make a new option to `seek`, `TSK_SEEK_LINEAR` which always uses the linear seek algorithm. We'll want to expose this to Python also, so there'll be a bit of plumbing involved.
> 
> We should specify this option in the ld_calculator, as that is definitely somewhere that linear seeking makes sense.

Okay; that makes sense to me. I'll revert `seek_linear` to its previous form and move the seek algorithm that uses `tree_pos` to a new function. I'll make the `tree_pos` method the default for `seek`, then see exactly which methods need to be changed back to linear in order for the tests to pass. Shall we call the new `tree_pos` method `tsk_tree_seek_nonlinear`?

jeromekelleher:
seek_skip might be a bit more descriptive?

jeromekelleher:
Something to bear in mind here is that sample_lists might not be compatible with this when seeking around randomly because of the non-linear order that edges can get inserted. A straightforward approach to getting this merged then could be to simply require that TSK_SEEK_LINEAR is always used with sample lists.

jeromekelleher:
I'm going to take a quick look at this to see what's happening

c/tskit/trees.c
```diff
@@ -6556,9 +6560,9 @@ tsk_tree_seek_forward(tsk_tree_t *self, tsk_id_t index)
     for (j = tree_pos.out.start; j != tree_pos.out.stop; j++) {
         e = tree_pos.out.order[j];
         e_left = edge_left[e];
-        if (e_left <= old_right) {
+        if (e_left < old_right) {
```
jeromekelleher, line 6561: I think this was a lot of the problem @duncanMR - took me a while to track it down!

Duncan-JR: Oh wow, sorry I missed that!

benjeffery: Nice find!

benjeffery:
> I guess a reasonable way forward might be to merge this much and file an issue to track what the performance differences are on a few different things and then decide whether we want to make it the default or not based on that.

Sounds good to me.

c/tskit/trees.c
```diff
@@ -6549,7 +6635,7 @@ tsk_tree_seek_index(tsk_tree_t *self, tsk_id_t tree, tsk_flags_t options)
 }
 
 static int TSK_WARN_UNUSED
-tsk_tree_seek_linear(tsk_tree_t *self, double x, tsk_flags_t TSK_UNUSED(options))
```
benjeffery, line 6552: I don't understand why the unused flags are being removed? Won't ever be part of the public API?

c/tskit/trees.h
```diff
@@ -1375,12 +1382,22 @@ we will have ``position < tree.interval.right``.
 
 Seeking to a position currently covered by the tree is
 a constant time operation.
+
+Seeking to a position from a non-null tree use a linear time
```
benjeffery:
```suggestion
Seeking to a position from a non-null tree uses a linear time
```

benjeffery (approved):
LGTM! Couple of small things

c/tskit/trees.c
jeromekelleher: It makes more sense when you look at where we call tsk_tree_seek_skip or tsk_tree_seek_linear below by looking at ``options``. Doesn't really matter ultimately as it's a private function.

Issue #2904, opened 2024-02-11
Simpler, more efficient interface for pair coalescence rates


nspope:
I'm thinking that the `CoalescenceTimeDistribution` class would be best moved to a separate package (where it can be numba-fied), and that instead we'll want to put an efficient algorithm for calculating pair coalescence rates within time windows into tskit.

The only computationally intensive part of these statistics -- that would need to go into the C library -- is calculating the number of pairs that coalesce in a given node, where each pair-node combination is weighted by its span along the sequence. 

Here's a particularly simple edge-based algorithm for doing this, that could be easily extended to include cross-coalescence between sample subsets, windows along the genome, etc:

```python
def pair_coalescence_counts(ts):
    """
    Calculate weighted # of pair coalescence events per node using incremental updates across trees
    """

    edges_child = ts.edges_child
    edges_parent = ts.edges_parent

    nodes_parent = np.full(ts.num_nodes, tskit.NULL)
    sample_counts = np.zeros(ts.num_nodes)
    coalescing_pairs = np.zeros(ts.num_nodes)

    for c in ts.samples():
        sample_counts[c] = 1

    for ed in ts.edge_diffs():
        edges_in = [e.id for e in ed.edges_in]
        edges_out = [e.id for e in ed.edges_out]

        right_span = ts.sequence_length - ed.interval.left

        for e in edges_out:
            c, p = edges_child[e], edges_parent[e]
            nodes_parent[c] = tskit.NULL
            update = sample_counts[c]
            while p != tskit.NULL:
                coalescing_pairs[p] -= update * (sample_counts[p] - sample_counts[c]) * right_span
                c, p = p, nodes_parent[p]
            p = edges_parent[e]
            while p != tskit.NULL:
                sample_counts[p] -= update
                p = nodes_parent[p]

        for e in edges_in:
            c, p = edges_child[e], edges_parent[e]
            nodes_parent[c] = p
            update = sample_counts[c]
            while p != tskit.NULL:
                sample_counts[p] += update
                p = nodes_parent[p]
            p = edges_parent[e]
            while p != tskit.NULL:
                coalescing_pairs[p] += update * (sample_counts[p] - sample_counts[c]) * right_span
                c, p = p, nodes_parent[p]

    return coalescing_pairs
```

A crude correctness check is to verify the relationship between coalescing pair counts and branch diversity,

```python
ts = msprime.sim_ancestry(
    samples=1000, ploidy=1, population_size=1e4, 
[...]

PR #2915, opened 2024-04-01
API for counting coalescing pairs

nspope:
Core algorithm, tests and API that partially address #2904


nspope:
@jeromekelleher and @petrelharp this is ready for a look whenever you have time. 

Shall I move the core algorithm into the C library in this PR or in a followup?

python/tests/test_coalrate.py
```diff
@@ -0,0 +1,357 @@
...
+        n = 10
+        model = msprime.BetaCoalescent(alpha=1.5)  # polytomies
+        tables = msprime.sim_ancestry(
+            samples=n,
+            recombination_rate=1e-8,
+            sequence_length=5e5,
+            population_size=1e4,
+            random_seed=1024,
+            model=model,
+        ).dump_tables()
```
petrelharp, line 160: perhaps an assert that this example has more than one tree?

python/tests/test_coalrate.py
```diff
@@ -0,0 +1,357 @@
...
+        t = self.example_ts().dump_tables()
+        ss0 = np.flatnonzero(t.nodes.population == 0)
+        ss1 = np.flatnonzero(t.nodes.population == 1)
+        remove_0 = np.logical_and(t.edges.left > 2e5, t.edges.right <= 3e5)
+        remove_0 = np.logical_and(remove_0, np.in1d(t.edges.child, ss0))
+        remove_1 = np.logical_and(
+            t.edges.left > 8e5, t.edges.right <= t.sequence_length
+        )
+        remove_1 = np.logical_and(remove_1, np.in1d(t.edges.child, ss1))
+        remove = np.logical_and(remove_0, remove_1)
```
petrelharp, line 227:
```suggestion
        remove = np.logical_and(remove_0, remove_1)
        assert np.sum(remove) > 0
```

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
-        )
+        The number of coalescing pairs may be calculated within or between the
+        non-overlapping lists of samples contained in `sample_sets`. In the
+        latter case, pairs are counted if they have exactly one member in each
+        of two sample sets. If `sample_sets` is omitted, a single group containing
+        all samples is assumed.
+
+        The argument `indexes` may be used to specify which pairs of sample
+        sets to compute the statistic between, and in what order. If omitted,
+        all possible pairs of sample sets are output in lexicographic order
```
petrelharp, line 9312: Hm, this is somewhat different than [in other stats](https://tskit.dev/tskit/docs/latest/stats.html#output-dimensions) - if we followed that, then if there were two sample sets and `indexes=None` that'd be the same as `indexes=(0,1)`; while if there weren't two sample sets then `indexes=None` would produce an error.

petrelharp: A nice thing about making people input the indexes explicitly rather than letting them rely on `indexes=None` is then they can't get mixed up about what order the output is coming in (since the output is not labeled).

nspope: that'd be good to match. In which case, if there's one sample set and indexes=None, then I think that should be equivalent to indexes=(0,0)?

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
+                inside = sample_counts[c]
+                while p != tskit.NULL:
+                    outside = sample_counts[p] - sample_counts[c]
+                    for i, (j, k) in enumerate(indexes):
+                        weight = inside[j] * outside[k] + inside[k] * outside[j]
+                        coalescing_pairs[w, p, i] -= weight * remainder
+                    c, p = p, nodes_parent[p]
+                p = edges_parent[e]
+                while p != tskit.NULL:
+                    sample_counts[p] -= inside
```
petrelharp, line 9393: I think this can move up three lines, and eliminate this second while loop?

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
+                while p != tskit.NULL:
+                    sample_counts[p] += inside
+                    p = nodes_parent[p]
+                p = edges_parent[e]
+                while p != tskit.NULL:
+                    outside = sample_counts[p] - sample_counts[c]
+                    for i, (j, k) in enumerate(indexes):
+                        weight = inside[j] * outside[k] + inside[k] * outside[j]
+                        coalescing_pairs[w, p, i] += weight * remainder
+                    c, p = p, nodes_parent[p]
```
petrelharp, line 9412:
```suggestion
                while p != tskit.NULL:
                    outside = sample_counts[p]
                    for i, (j, k) in enumerate(indexes):
                        weight = inside[j] * outside[k] + inside[k] * outside[j]
                        coalescing_pairs[w, p, i] += weight * remainder
                    sample_counts[p] += inside
                    c, p = p, nodes_parent[p]
```

petrelharp:
The `remainder` method is very clever. However, it does involve a lot of adding & subtracting of close numbers: is floating point error liable to be an issue? If there are `n` samples and a sequence length `L`, then a node present for O(1) sequence length will have their coalescing pairs computed as `n^2 * (L + O(1)) - n^2 * L`; I suppose this is only a problem if `1 / L` is machine epsilon? So, not a worry?

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
+            while w < num_windows and windows[w + 1] <= position:  # flush window
+                remainder = sequence_length - windows[w + 1]
+                for c, p in enumerate(nodes_parent):
+                    if p == tskit.NULL:
+                        continue
+                    inside = sample_counts[c]
+                    outside = sample_counts[p] - sample_counts[c]
+                    for i, (j, k) in enumerate(indexes):
+                        weight = inside[j] * outside[k] + inside[k] * outside[j]
+                        coalescing_pairs[w, p, i] -= weight * remainder / 2
```
petrelharp, line 9430: why is this divided by 2?

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
+                        if w + 1 < num_windows:
+                            coalescing_pairs[w + 1, p, i] += weight * remainder / 2
+                w += 1
+
+        for i, (j, k) in enumerate(indexes):
+            coalescing_pairs[:, :, i] /= 1 + int(j == k)
```
petrelharp, line 9436:
```suggestion
        for i, (j, k) in enumerate(indexes):
            if j == k:
                coalescing_pairs[:, :, i] /= 2
```
Easier to read...

python/tskit/trees.py
```diff
@@ -9291,25 +9290,156 @@ def ibd_segments(
...
+
+        if span_normalise:
+            for w, s in enumerate(np.diff(windows)):
+                coalescing_pairs[w] /= s
+
+        return coalescing_pairs.squeeze()
```
petrelharp, line 9442: I think we only want to squeeze dimensions for which the corresponding argument was `None`? We could use the method for this over in `stats.py`?

petrelharp:
This looks great - I've made some comments; and perhaps it needs another test case for which some samples might be parents to other samples? (e.g., produced by the nonWF simulator in the test suite, I think?)

python/tests/test_coalrate.py
```diff
@@ -0,0 +1,357 @@
...
+    and right child. For higher arities, the count is summed over all possible
+    pairs of children.
+    """
+    output = np.zeros(ts.num_nodes)
+    for t in ts.trees():
+        sample_counts = np.zeros((ts.num_nodes, 2), dtype=np.int32)
+        for n in t.postorder():
+            samples = list(t.samples(n))
+            sample_counts[n, 0] = np.intersect1d(samples, sample_set_0).size
+            sample_counts[n, 1] = np.intersect1d(samples, sample_set_1).size
```
jeromekelleher, line 49: Could do this in a single postorder pass I think? Just for clarity of what the algorithm is doing.

jeromekelleher (commented):
Generally looks good, I'll have a closer look when @petrelharp's comments are addressed.

I'd suggest getting this much merged first, then following up with a draft PR that implements the algorithm in Python (in the test file) using the TreePosition code. Once we're happy with that algorithm,translating to C should be straightforward.

python/tskit/trees.py
nspope: This won't work, unfortunately -- the first while loop needs to use `sample_counts[p]` before it's modified, for all `p` on the path from the first child to the root. Only after these are updated can `sample_counts` be modified.

nspope:
> I suppose this is only a problem if 1 / L is machine epsilon

Right -- if you've got a node with span `S` in a sequence of length `L`, you'll lose bits if `S / L < machine epsilon`. In practice this seems like this won't happen, and in principle can be identified. What do you think, @jeromekelleher?

jeromekelleher:
Seems pretty unlikely to me

python/tskit/trees.py
nspope: won't work for same reason as above

petrelharp (approved):

petrelharp:
We discussed changing the name, but I've forgotten to what? And, to be clear, i think Jerome's suggesting modifying this python algorithm, not adding a new one.

nspope:
This is ready for another look ...
- Addressed @petrelharp's comments -- algorithm had to be modified slightly to work when internal nodes are flagged as samples.
- Changed name to 'pair_coalescence_counts'. The idea is this'll have a twin, 'pair_coalescence_rates' that invokes 'pair_coalescence_counts' under the hood and converts counts to rates within time windows or quantile intervals.
- Added a 'time_discretisation' argument that maps nodes to time intervals. When there's a ton of nodes / windows / sample sets the output array may get too big to store in memory. In which case this lets one get a more compact summary that can be used to e.g. calculate coalescence rates. By default this is a string "nodes"; otherwise an array of breakpoints ending at np.inf.

nspope:
> i think Jerome's suggesting modifying this python algorithm, not adding a new one.

But in a followup PR, right?

python/tskit/trees.py
nspope: it's because this accumulates weights from all `children * (children - 1)` ordered pairs of children for anode -- and instead we want to accumulate weights from all `children * (children - 1) / 2` unordered pairs. Both orderings of a pair produce the same weight, so we divide by two

python/tskit/trees.py
```diff
@@ -9291,25 +9290,203 @@ def ibd_segments(
...
-        *,
         sample_sets=None,
-        weight_func=None,
-        window_breaks=None,
-        blocks_per_window=None,
-        span_normalise=False,
+        indexes=None,
+        windows=None,
+        span_normalise=True,
+        time_discretisation="nodes",
```
jeromekelleher, line 9299: Maybe ``time_bins`` or ``time_windows``?

python/tskit/trees.py
```diff
@@ -9291,25 +9290,203 @@ def ibd_segments(
...
+            raise ValueError("Window breaks must be strictly increasing")
+
+        if isinstance(time_discretisation, str) and time_discretisation == "nodes":
+            nodes_map = np.arange(self.num_nodes)
+            output_size = self.num_nodes
+        else:
+            if not isinstance(time_discretisation, np.ndarray):
+                raise ValueError("Time discretisation must be an array of breakpoints")
+            if not (
+                time_discretisation[0] == 0.0 and time_discretisation[-1] == np.inf
```
jeromekelleher, line 9376: Time doesn't have to start at zero - we can have negative times.

jeromekelleher (commented):
LGTM - some minor points on naming the time windows argument and how we deal with negative times. If you like you can drop this argument from the initial version and make an issue to track?

jeromekelleher:
Any idea why codecov isn't working here @benjeffery? It's making the diff impossible to read here, which is worse than useless.

nspope:
> If you like you can drop this argument from the initial version and make an issue to track?

Thanks @jeromekelleher, it'd be great to sort this API out now as it's very close.

re: naming, @petrelharp brought up that the stats methods will eventually have a `time_windows` argument, but that the behavior will be a bit different because `pair_coalescence_counts` needs a (default) time discretisation option of `"nodes"` as well as an option to pass window breakpoints. So we could either:
1. Name this argument something other than `time_windows` -- I think `time_bins` is fine (definitely better than `time_discretisation`)
2. Call this argument `time_windows` and document that it accepts an additional string option that won't be accepted by other stats methods

Either is fine by me. Do you have a preference @petrelharp?

benjeffery:
> Any idea why codecov isn't working here @benjeffery? It's making the diff impossible to read here, which is worse than useless.

I'm seeing a "Commit YAML is invalid" error at codecov - checking it out.

benjeffery:
@mergifyio rebase

benjeffery:
@nspope I've rebased here to fix CI so you won't be able to push changes without a reset. Not sure how much of a git ninja you are so let me know if you need to push changes.

petrelharp:
> I think having this method accept an additional argument to time_windows here would be totally fine. It would be more confusing to use a different name just to avoid that minor bit of inconsistency I think.

Sounds good to me!

nspope:
Thanks @benjeffery! Looks like codecov is failing lwt-tests and python-c-tests -- I don't think that's due to this PR?

nspope:
Thanks @jeromekelleher -- I've modified so that `time_windows` argument accepts either `"nodes"` or a sorted array of breakpoints. Nodes that fall outside of `time_windows` aren't counted in the output.

nspope:
Darn, it looks like codecov is still mangling the diff, though this is rebased onto 9e1bf0 ?

benjeffery:
So the comment in https://github.com/tskit-dev/tskit/pull/2915#issuecomment-2029229828 is correct - only one line is missing coverage. However, previous comments made by codecov are not removed :(
I think the only way to clear those would be to open a new PR.

nspope:
Hmm, now the post-test Codecov upload is erroring out:

> [2024-04-10T04:01:34.167Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

I'll go ahead and open a new PR ...

benjeffery:
> Hmm, now the post-test Codecov upload is erroring out:
> 
> > [2024-04-10T04:01:34.167Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
> 
> I'll go ahead and open a new PR ...

Yeah, we sometimes get that one. I spent a while investigating it and it seems to be transient on codecov's end.

nspope:
@benjeffery it may be related to v3 codecov-actions deprecation, see [comment](https://github.com/tskit-dev/tskit/pull/2920#issuecomment-2048117786). Happy to wait and see if it resolves itself though

nspope:
Well-- bumping to codecov-actions to v4 got rid of the upload errors, but the coverage reports don't seem to be making their way here.

nspope:
@benjeffery Sorry for the hassle ... but any chance you could clone and force-push this to re-trigger codecov? To see if the issue might be related to this PR coming from a fork? It seems like the coverage report is working fine in #2924

benjeffery:
@mergifyio rebase

benjeffery:
Codecov issue now seems fixed here - it has removed all it's comments about uncovered lines.

jeromekelleher:
Are we good to merge then @nspope?

nspope:
Yes! Thanks for the fix @benjeffery

PR #2920, opened 2024-04-10
Add `pair_coalescence_counts` method

nspope:
Reprise of #2915

PR #2922, opened 2024-04-10
Test codecov

benjeffery:


jeromekelleher:
Hmm, this really doesn't seem to be working well. I'm wondering if codecov is really helping here, or we should just ditch it and try something else. It'll be a lot of hassle switching though, as it 's quite complicated with all the different C, Python etc tests.

benjeffery:
> Hmm, this really doesn't seem to be working well. I'm wondering if codecov is really helping here, or we should just ditch it and try something else. It'll be a lot of hassle switching though, as it 's quite complicated with all the different C, Python etc tests.

It seems to have made the report here ok - I'll add another PR with some additional lines to see that it detects coverage correctly.

nspope:
The problems we're seeing may be because of #2915 coming from a fork? The issues linked to [here](https://github.com/tskit-dev/tskit/pull/2920#issuecomment-2048087845) are largely concerned with this case.

benjeffery:
> The problems we're seeing may be because of #2915 coming from a fork? The issues linked to [here](https://github.com/tskit-dev/tskit/pull/2920#issuecomment-2048087845) are largely concerned with this case.

All my PRs come from a fork too.

jeromekelleher:
Is there something we can do to unplug progress here? Would we be better off turning off codecov for the moment so that in-flight PRs can be viewed and merged?

nspope:
If it's only affecting my PRs, then it must be something on my end. I'm not sure what to do, though -- if I check coverage locally, it's 100% on the diff (and, codecov agreed with that before it started bugging out a few days ago).

petrelharp:
One possible issue that I see is that #2922 (in which the current `main/`'s HEAD commit was authored)  didn't pass codecov, and in fact not all the coverage files got uploaded for it, so when comparing to it things are wonky:
Screenshot from 2024-04-14 21-13-54
Screenshot from 2024-04-14 21-13-36

This I guess can't be the direct source of these woes, however, as it happened after?

petrelharp:
I'm trying in #2927, fyi

petrelharp:
Nope - it's broken for me, too.

Note that to get codecov to actually run I needed to make a new commit (seems codecov was cacheing the commit hash, since it didn't re-run even though the tests did).

Note also that what it's using for `base` looks not right:
Screenshot from 2024-04-14 22-48-18
I'm currently unsure what all this means, though.

benjeffery:
> Is there something we can do to unplug progress here? Would we be better off turning off codecov for the moment so that in-flight PRs can be viewed and merged?

Note that codecov has never been a requirement for mergify to merge PRs. This still needs fixing, but it doesn't prevent a merge/

PR #2928, opened 2024-04-15
Set explict paths for codecov

benjeffery:
I think codecov is confused by multiple coverage reports for the python code. Here I'm attempting to be explict about which paths apply to which tests.


jeromekelleher (approved):

benjeffery:
Interestingly - now most of the uploads generate errors. I think these are due to source file paths not being found, although codecov gives no more detail. Will try to fix these, then we should have a good setup.

Screenshot at 2024-04-15 17-14-19

PR #2930, opened 2024-04-17
Fix codecov paths

benjeffery:


benjeffery:
I finally understand why we've been getting so much odd codecov behaviour, I think! Some reports were not being used as the paths were relative to a subdirectory. Hopefully this fixes that.

PR #2932, opened 2024-04-18
Low level pair coalescence counts

nspope:
Low level extension of #2915

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -8145,3 +8145,308 @@ out:
     tsk_table_collection_free(&tables);
     return ret;
 }
+
+/* ======================================================== *
+ * Pair coalescence
+ * ======================================================== */
+
+static int
+check_node_output_map(const tsk_size_t num_nodes, const tsk_size_t num_outputs,
+    const tsk_id_t *node_output_map)
+{
+    int ret = 0;
+    tsk_id_t max_index, index;
+    tsk_size_t i;
+
+    max_index = TSK_NULL;
+    for (i = 0; i < num_nodes; i++) {
+        index = node_output_map[i];
+        if (index < TSK_NULL) {
+            ret = TSK_ERR_BAD_NODE_OUTPUT_MAP;
+            goto out;
+        }
+        if (index > max_index) {
+            max_index = index;
+        }
+    }
+    if (num_outputs < 1 || (tsk_id_t) num_outputs != max_index + 1) {
+        ret = TSK_ERR_BAD_NODE_OUTPUT_MAP_DIM;
+        goto out;
+    }
+out:
+    return ret;
+}
+
+static inline void
+pair_coalescence_count(tsk_size_t num_set_indexes, const tsk_id_t *set_indexes,
+    tsk_size_t num_sample_sets, const double *parent_count, const double *child_count,
+    const double *parent_state, const double *inside, double *outside, double *result)
+{
+    tsk_size_t i;
+    tsk_id_t j, k;
+    for (i = 0; i < num_sample_sets; i++) {
+        outside[i] = parent_count[i] - child_count[i] - parent_state[i];
+    }
+    for (i = 0; i < num_set_indexes; i++) {
+        j = set_indexes[2 * i];
+        k = set_indexes[2 * i + 1];
+        result[i] = outside[j] * inside[k];
+        if (j != k) {
+            result[i] += outside[k] * inside[j];
+        }
+    }
+}
+
+int
+tsk_treeseq_pair_coalescence_stat(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
+    tsk_size_t num_set_indexes, const tsk_id_t *set_indexes, tsk_size_t num_windows,
+    const double *windows, tsk_size_t num_outputs, const tsk_id_t *node_output_map,
+    tsk_flags_t options, double *result)
+{
+    int ret = 0;
+    double left, right, remaining_span, window_span;
+    tsk_id_t e, p, c, u, v, w, i, j;
+    tsk_size_t num_samples;
+    tsk_tree_position_t tree_pos;
+    const tsk_table_collection_t *tables = self->tables;
+    const tsk_size_t num_nodes = tables->nodes.num_rows;
+    const double sequence_length = tables->sequence_length;
+
+    /* buffers */
+    bool *visited = NULL;
+    tsk_id_t *nodes_sample_set = NULL;
+    tsk_id_t *nodes_parent = NULL;
+    double *nodes_sample = NULL;
+    double *sample_count = NULL;
+    double *coalescing_pairs = NULL;
+    double *nodes_weight = NULL;
+    double *outside = NULL;
+    double *count = NULL;
+
+    /* row pointers */
+    double *inside = NULL;
+    double *weight = NULL;
+    double *output = NULL;
+    double *above = NULL;
+    double *below = NULL;
+    double *state = NULL;
+    double *pairs = NULL;
+
+    tsk_memset(&tree_pos, 0, sizeof(tree_pos));
+
+    /* check inputs */
+    ret = tsk_treeseq_check_windows(self, num_windows, windows, options);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = check_set_indexes(num_sample_sets, 2 * num_set_indexes, set_indexes);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = tsk_treeseq_check_sample_sets(
+        self, num_sample_sets, sample_set_sizes, sample_sets);
+    if (ret != 0) {
+        goto out;
+    }
+    ret = check_node_output_map(num_nodes, num_outputs, node_output_map);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* map nodes to sample sets */
+    nodes_sample_set = tsk_malloc(num_nodes * sizeof(*nodes_sample_set));
+    if (nodes_sample_set == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    ret = get_sample_set_index_map(self, num_sample_sets, sample_set_sizes, sample_sets,
+        &num_samples, nodes_sample_set);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* initialize internal state */
+    visited = tsk_malloc(num_nodes * sizeof(*visited));
+    nodes_parent = tsk_malloc(num_nodes * sizeof(*nodes_parent));
+    nodes_sample = tsk_calloc(num_nodes * num_sample_sets, sizeof(*nodes_sample));
+    sample_count = tsk_malloc(num_nodes * num_sample_sets * sizeof(*sample_count));
+    outside = tsk_malloc(num_sample_sets * sizeof(*outside));
+    count = tsk_malloc(num_set_indexes * sizeof(*count));
+    coalescing_pairs
+        = tsk_calloc(num_outputs * num_set_indexes, sizeof(*coalescing_pairs));
+    nodes_weight = tsk_malloc(num_outputs * num_set_indexes * sizeof(*nodes_weight));
+    if (nodes_parent == NULL || nodes_sample == NULL || sample_count == NULL
+        || coalescing_pairs == NULL || nodes_weight == NULL || outside == NULL
+        || count == NULL || visited == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    for (c = 0; c < (tsk_id_t) num_nodes; c++) {
+        i = nodes_sample_set[c];
+        if (i != TSK_NULL) {
+            state = GET_2D_ROW(nodes_sample, num_sample_sets, c);
+            state[i] = 1.0;
+        }
+        nodes_parent[c] = TSK_NULL;
+        visited[c] = false;
+    }
+    tsk_memcpy(
+        sample_count, nodes_sample, num_nodes * num_sample_sets * sizeof(*sample_count));
+
+    ret = tsk_tree_position_init(&tree_pos, self, 0);
+    if (ret != 0) {
+        goto out;
+    }
+
+    w = 0;
+    while (true) {
+        tsk_tree_position_next(&tree_pos);
+        if (tree_pos.index == TSK_NULL) {
+            break;
+        }
+
+        left = tree_pos.interval.left;
+        right = tree_pos.interval.right;
+        remaining_span = sequence_length - left;
+
+        for (u = tree_pos.out.start; u != tree_pos.out.stop; u++) {
+            e = tree_pos.out.order[u];
+            p = tables->edges.parent[e];
+            c = tables->edges.child[e];
+            nodes_parent[c] = TSK_NULL;
+            inside = GET_2D_ROW(sample_count, num_sample_sets, c);
+            while (p != TSK_NULL) {
+                v = node_output_map[p];
+                if (v != TSK_NULL) {
+                    above = GET_2D_ROW(sample_count, num_sample_sets, p);
+                    below = GET_2D_ROW(sample_count, num_sample_sets, c);
+                    state = GET_2D_ROW(nodes_sample, num_sample_sets, p);
+                    pairs = GET_2D_ROW(coalescing_pairs, num_set_indexes, v);
+                    pair_coalescence_count(num_set_indexes, set_indexes, num_sample_sets,
+                        above, below, state, inside, outside, count);
+                    for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+                        pairs[i] -= count[i] * remaining_span;
+                    }
+                }
+                c = p;
+                p = nodes_parent[c];
+            }
+            p = tables->edges.parent[e];
+            while (p != TSK_NULL) {
+                above = GET_2D_ROW(sample_count, num_sample_sets, p);
+                for (i = 0; i < (tsk_id_t) num_sample_sets; i++) {
+                    above[i] -= inside[i];
+                }
+                p = nodes_parent[p];
+            }
+        }
+
+        for (u = tree_pos.in.start; u != tree_pos.in.stop; u++) {
+            e = tree_pos.in.order[u];
+            p = tables->edges.parent[e];
+            c = tables->edges.child[e];
+            nodes_parent[c] = p;
+            inside = GET_2D_ROW(sample_count, num_sample_sets, c);
+            while (p != TSK_NULL) {
+                above = GET_2D_ROW(sample_count, num_sample_sets, p);
+                for (i = 0; i < (tsk_id_t) num_sample_sets; i++) {
+                    above[i] += inside[i];
+                }
+                p = nodes_parent[p];
+            }
+            p = tables->edges.parent[e];
+            while (p != TSK_NULL) {
+                v = node_output_map[p];
+                if (v != TSK_NULL) {
+                    above = GET_2D_ROW(sample_count, num_sample_sets, p);
+                    below = GET_2D_ROW(sample_count, num_sample_sets, c);
+                    state = GET_2D_ROW(nodes_sample, num_sample_sets, p);
+                    pairs = GET_2D_ROW(coalescing_pairs, num_set_indexes, v);
+                    pair_coalescence_count(num_set_indexes, set_indexes, num_sample_sets,
+                        above, below, state, inside, outside, count);
+                    for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+                        pairs[i] += count[i] * remaining_span;
+                    }
+                }
+                c = p;
+                p = nodes_parent[c];
+            }
+        }
+
+        /* flush windows */
+        while (w < (tsk_id_t) num_windows && windows[w + 1] <= right) {
+            remaining_span = sequence_length - windows[w + 1];
+            tsk_memcpy(nodes_weight, coalescing_pairs,
+                num_outputs * num_set_indexes * sizeof(*nodes_weight));
+            tsk_memset(coalescing_pairs, 0,
+                num_outputs * num_set_indexes * sizeof(*coalescing_pairs));
+            for (j = 0; j < (tsk_id_t) num_samples; j++) { /* traverse subtree */
+                c = sample_sets[j];
+                p = nodes_parent[c];
+                while (!visited[c] && p != TSK_NULL) {
+                    v = node_output_map[p];
+                    if (v != TSK_NULL) {
+                        above = GET_2D_ROW(sample_count, num_sample_sets, p);
+                        below = GET_2D_ROW(sample_count, num_sample_sets, c);
+                        state = GET_2D_ROW(nodes_sample, num_sample_sets, p);
+                        pairs = GET_2D_ROW(coalescing_pairs, num_set_indexes, v);
+                        weight = GET_2D_ROW(nodes_weight, num_set_indexes, v);
+                        pair_coalescence_count(num_set_indexes, set_indexes,
+                            num_sample_sets, above, below, state, below, outside, count);
+                        for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+                            pairs[i] += count[i] * remaining_span / 2;
+                            weight[i] -= count[i] * remaining_span / 2;
+                        }
+                    }
+                    visited[c] = true;
+                    c = p;
+                    p = nodes_parent[c];
+                }
+            }
+            for (j = 0; j < (tsk_id_t) num_samples; j++) {
+                c = sample_sets[j];
+                p = nodes_parent[c];
+                while (visited[c] && p != TSK_NULL) {
+                    visited[c] = false;
+                    c = p;
+                    p = nodes_parent[c];
+                }
+            }
+            if (options & TSK_STAT_SPAN_NORMALISE) {
+                window_span = windows[w + 1] - windows[w];
+                for (v = 0; v < (tsk_id_t) num_outputs; v++) {
+                    weight = GET_2D_ROW(nodes_weight, num_set_indexes, v);
+                    for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+                        weight[i] /= window_span;
+                    }
+                }
+            }
+            // TODO:
+            // for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+            //     reduce(i, col, nodes_weight, &result[w * row_dim * col_dim])
+            // };
+            for (v = 0; v < (tsk_id_t) num_outputs; v++) {
+                output = GET_3D_ROW(
+                    result, num_outputs, num_set_indexes, (tsk_size_t) w, v);
+                weight = GET_2D_ROW(nodes_weight, num_set_indexes, v);
+                for (i = 0; i < (tsk_id_t) num_set_indexes; i++) {
+                    output[i] = weight[i];
+                }
+            }
+            w += 1;
+        }
+    }
+out:
+    tsk_tree_position_free(&tree_pos);
+    tsk_safe_free(nodes_sample_set);
+    tsk_safe_free(nodes_parent);
+    tsk_safe_free(nodes_sample);
+    tsk_safe_free(sample_count);
+    tsk_safe_free(coalescing_pairs);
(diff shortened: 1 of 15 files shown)


nspope:
I'd like to generalize this algorithm slightly:
- Currently, the output is either a `num_windows` by `num_nodes` array (which is very large), or a `num_windows` by `num_time_windows` array where the counts are summed within time windows.
- Conceptually, the "nodes" output gives the empirical distribution of pair coalescence times in windows across the gnome. That is, for each window we have a vector of RVs (node times) and a vector of weights (pair coalescence counts).
- From this distributional viewpoint, there's lots of useful things that be calculated: the CDF, quantiles, moments, coalescence rates, etc. (of which the "sum in time windows" option of the current implementation is one special case)

So, I think it'd be useful to take the current algorithm, and have it apply a summary function at the end of each window. This would let us calculate any useful summary statistic without having to create a potentially humongous windows by nodes array as an intermediate.

jeromekelleher (commented):
Looks good to me - do you want to add the summary func stuff now before we start porting to C? Probably a good idea, if you want to do this in the short term.

nspope:
The docs build is failing with numpy 2 issues on `import msprime`, after a rebase-- is this expected?

jeromekelleher:
This is probably a package cache issue @benjeffery?

nspope:
I bumped the cache version to no avail-- however, it looks like `requirements/CI-docs/requirements.txt` has msprime pinned to 1.3.1?

nspope:
Yup, pinning msprime to the latest version in requirements/* seems to have done the trick.

jeromekelleher (approved):
This all looks great to me @nspope - the algorithm seems very clean.

Regarding Python callbacks for C code, although there's some existing infrastructure for doing this, I'd leave this bit until last. The Python C API stuff is very tricky and there are footguns everywhere. Also, performance will be terrible - there's no way you can use this approach for anything other than prototyping. The summary func gets called in the deepest inner loops, and getting the Python interpreter involved there just doesn't work performance-wise.

nspope:
Hmm, tests are passing on Ubuntu but evidently not on MacOS or Windows. There's no error message, and the location looks stochastic. Not sure how to debug this as I only have Linux locally (on which everything runs fine).

jeromekelleher:
I would strongly advise writing some simple C tests that exercise all reachable code paths, and running it through valgrind. This is a much more effective way of tracking down these kinds of bugs. I can help with this, if it's not clear how to start?

nspope:
Makes sense @jeromekelleher, I can take a crack at it. Just so I'm understanding the scope, is the goal to recreate the entire test suite from python in `c/tests` (correctness tests, etc)? Or something more minimal?

jeromekelleher:
Something minimal. You're just looking to exercise all code paths so that valgrind can see them.

nspope:
That did the trick ... is there a good way to trigger OOM errors after the mallocs, to get those lines covered?

nspope:
The only substantial coverage gap is in `_tskitmodule.c`: https://app.codecov.io/gh/tskit-dev/tskit/pull/2932/blob/python/_tskitmodule.c#L9924 where it looks like all execution paths ends in the check,
```c
    if (TreeSequence_check_state(self) != 0) {
        goto out;
    }
```
however, this can't be the *only* execution path, because otherwise `ll_tree_sequence.pair_coalescence_counts` wouldn't return anything and none of the python test suite would pass.

Is there something else needed to get the CPython parts covered?

nspope:
Ah, nevermind-- I see that coverage for CPython is getting read off of `test_lowlevel.py`

jeromekelleher:
> is there a good way to trigger OOM errors after the mallocs, to get patterns like

We've been ignoring these here, as they are very tricky to trigger. We did it in the kastore C tests by running with a patched malloc which fails after n allocs, but it's a lot of work for a pretty marginal benefit.

c/tskit/trees.c
```diff
@@ -1232,6 +1232,26 @@ tsk_treeseq_check_windows(const tsk_treeseq_t *self, tsk_size_t num_windows,
...
 
+static int
+check_time_windows(tsk_size_t num_time_windows, const double *time_windows)
+{
+    int ret = TSK_ERR_BAD_TIME_WINDOWS;
+    tsk_size_t j;
+
+    if (num_time_windows == 1) {
+        ret = TSK_ERR_BAD_NUM_TIME_WINDOWS;
+        goto out;
```
jeromekelleher, line 1243: We would want to test this in C. These are error conditions that you can imagine, e.g., we don't correctly clean memory up after. So, it's important to provoke them when running in valgrind.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
     tsk_table_collection_free(&tables);
     return ret;
 }
+
+/* ======================================================== *
+ * Pair coalescence
+ * ======================================================== */
+
+typedef struct {
+    /* TODO: better to include the implm from haplotype_matching.c, rather than
```
jeromekelleher, line 8174: Just duplicate it, but note that the same struct is defined here.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    const double *time_windows, tsk_id_t *result)
+{
+    int ret = 0;
+    const tsk_table_collection_t *tables = self->tables;
+    const tsk_size_t num_nodes = tables->nodes.num_rows;
+    tsk_node_argsort_t *nodes_time = NULL;
+    tsk_size_t i, j;
+
+    if (*num_time_windows == 0) { /* keep nodes unpooled */
+        for (i = 0; i != num_nodes; i++) {
```
jeromekelleher, line 8203: I guess it's the same thing, but ``for (i = 0; i < num_nodes; i++)`` is more idiomatic.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+        goto out;
+    }
+    for (i = 0; i != num_nodes; i++) {
+        result[i] = TSK_NULL; /* nodes outside windows */
+        nodes_time[i].value = tables->nodes.time[i];
+        nodes_time[i].index = i;
+    }
+    qsort(nodes_time, (size_t) num_nodes, sizeof(*nodes_time), cmp_node_argsort);
+
+    for (i = 0, j = 0; i != *num_time_windows; i++) {
```
jeromekelleher, line 8222: The comma operator is a surprising thing, and best avoided IMO. I would generally aim for C code to be more obvious than terse:
```
j = 0;
for (i = 0; i < *num_time_windows; i++) {
...
```

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
+    tsk_size_t num_set_indexes, const tsk_id_t *set_indexes, tsk_size_t num_windows,
+    const double *windows, tsk_size_t num_time_windows, const double *time_windows,
+    tsk_flags_t options, double *result)
+{
+    int ret = 0;
+    double left, right, remaining_span, window_span;
+    tsk_id_t e, p, c, u, v, n, w, i, j, k, row, col, inp;
+    tsk_size_t total_samples;
+    int weight;
```
jeromekelleher, line 8246: Plain 'int' values are tricky, as they can be anything from 16 bit to 64 bit depending on platform. It saves headaches to be explicit about the types, if there's any chance at all of overflow. So here - what's the maximum weight value? I'd err on the side of caution and declare things as int64, if there's any possibility of overflow from int32.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    tsk_tree_position_t tree_pos;
+    const tsk_table_collection_t *tables = self->tables;
+    const tsk_size_t num_nodes = tables->nodes.num_rows;
+    const double sequence_length = tables->sequence_length;
+    tsk_id_t *nodes_sample_set = NULL;
+    tsk_id_t *nodes_parent = NULL;
+    tsk_id_t *nodes_time_window = NULL;
+    int *nodes_sample = NULL;
```
jeromekelleher, line 8254: See comment above about using 'int' values. These are ``tsk_id_t`` values, right?

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+        goto out;
+    }
+    ret = get_sample_set_index_map(self, num_sample_sets, sample_set_sizes, sample_sets,
+        &total_samples, nodes_sample_set);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* initialize internal state */
+    inp = (tsk_id_t) num_sample_sets;
```
jeromekelleher, line 8294: Why not use num_sample_sets in the malloc here rather than ``inp``? Saves a cast, and it's easier to read.

jeromekelleher: Likewise for later memsets etc.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    col = (tsk_id_t) num_set_indexes;
+    nodes_parent = tsk_malloc(num_nodes * sizeof(*nodes_parent));
+    nodes_sample = tsk_malloc(num_nodes * (tsk_size_t) inp * sizeof(*nodes_sample));
+    sample_count = tsk_malloc(num_nodes * (tsk_size_t) inp * sizeof(*sample_count));
+    if (nodes_parent == NULL || nodes_sample == NULL || sample_count == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    tsk_memset(nodes_sample, 0, num_nodes * (tsk_size_t) inp * sizeof(*nodes_sample));
```
jeromekelleher, line 8303: Can avoid this by calling ``tsk_calloc`` instead of tsk_malloc above.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    }
+    ret = get_time_windows_index_map(
+        self, &num_time_windows, time_windows, nodes_time_window);
+    if (ret != 0) {
+        goto out;
+    }
+
+    /* initialize outputs */
+    row = (tsk_id_t) num_time_windows;
+    inside = tsk_malloc((tsk_size_t) col * sizeof(inside));
```
jeromekelleher, line 8329: Whoops, this needs to be ``sizeof(*inside)``. This will cause obscure failures on different platforms (ask me how I know!)

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    outside = tsk_malloc((tsk_size_t) col * sizeof(outside));
```
jeromekelleher, line 8330: ditto

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+    if (inside == NULL || outside == NULL) {
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    coalescing_pairs
```
jeromekelleher, line 8335: Ditto. Best check all your sizeof()s

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+        = tsk_malloc((tsk_size_t) row * (tsk_size_t) col * sizeof(coalescing_pairs));
+    nodes_weight
+        = tsk_malloc((tsk_size_t) row * (tsk_size_t) col * sizeof(nodes_weight));
+    if (coalescing_pairs == NULL || nodes_weight == NULL) {
```
jeromekelleher, line 8339: I think you can merge together some of these malloc/check things. I think it's fine to malloc up to (say) 10 things at once and check them all for NULL just after. We need to review these manually anyway, and it saves some noise in the coverage reports.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+        ret = TSK_ERR_NO_MEMORY;
+        goto out;
+    }
+    tsk_memset(coalescing_pairs, 0,
```
jeromekelleher, line 8343: Use ``calloc`` in general rather than malloc + memset(0) - saves a function call, and is potentially more efficient.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+        (tsk_size_t) row * (tsk_size_t) col * sizeof(*coalescing_pairs));
+
+    tsk_memset(&tree_pos, 0, sizeof(tree_pos));
```
jeromekelleher, line 8346: This needs to be done *before* any potentially failing operation. The reason for zeroing out is to make sure that the ``tree_pos_free`` operation is safe in all cases. So, it must be done before any ``goto out`` calls can happen.

c/tskit/trees.c
jeromekelleher: Ah - I see where you got this from. The tree_pos inner loops are a special case because they can go in either direction, so we use != stop.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+            p = tables->edges.parent[e];
+            c = tables->edges.child[e];
+            nodes_parent[c] = TSK_NULL;
+            for (i = 0; i != inp; i++) { /* samples beneath child */
+                inside[i] = sample_count[c * inp + i];
+            }
+            while (p != TSK_NULL) {
+                v = nodes_time_window[p];
+                if (v != TSK_NULL) {
+                    for (i = 0; i != inp; i++) { /* samples beneath sibs */
```
jeromekelleher, line 8373: I would rather use ``num_sample_sets`` here (or whatever it is) with a cast if necessary. It's easier to read, and there's no run-time cost to having a cast. The casts are really just there to keep the compiler happy.

c/tskit/trees.c
```diff
@@ -8145,3 +8165,339 @@ tsk_treeseq_extend_edges(const tsk_treeseq_t *self, int max_iter,
...
+                        outside[i] = sample_count[p * inp + i]
```
jeromekelleher, line 8374: I think there a macro for doing this GET_2D_ROW (and a GET_3D_ROW) function, may as well use these?

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,150 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
     return ret;
 }
 
+/* TODO:
+ * in the default case (no time windows) we calculate statistics at the level of nodes.
```
jeromekelleher, line 9864: Why don't we do this in Python instead? It should be easy to create a set of time-windows that correspond to the node times using numpy, right? In general, we should only do stuff in C if we really have to, particularly if it involves twisty logic around default input parameters.

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,150 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+    PyArrayObject *time_windows_array = NULL;
+    npy_intp *shape;
+
+    time_windows_array = (PyArrayObject *) PyArray_FROMANY(
+        time_windows, NPY_FLOAT64, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (time_windows_array == NULL) {
+        goto out;
+    }
+    shape = PyArray_DIMS(time_windows_array);
+    if (shape[0] == 1) { /* allow zero length array */
```
jeromekelleher, line 9882: All these error cases should be provoked in ``test_lowlevel.py``. This really is critical, as it's super easy to shoot yourself in the foot in the Python C API. A small mistake handling an error condition can lead to a really obscure core dump minutes later in totally unrelated parts of the program, which can take weeks to track down. (I have a war stories on this - really!)

python/_tskitmodule.c
```diff
@@ -9860,6 +9860,150 @@ TreeSequence_divergence_matrix(TreeSequence *self, PyObject *args, PyObject *kwd
...
+        goto out;
+    }
+    if (span_normalise) {
+        options |= TSK_STAT_SPAN_NORMALISE;
+    }
+
+    tsk_size_t num_nodes = tsk_treeseq_get_num_nodes(self->tree_sequence);
+    npy_intp dims[3];
+    dims[0] = num_windows;
+    dims[1] = num_time_windows > 0 ? num_time_windows : num_nodes;
```
jeromekelleher, line 9979: Yeah, let's insist on passing in explicit time windows. There's nothing to be gained (you malloc the memory later anyway) so let's simplify where we can.

python/tests/test_lowlevel.py
```diff
@@ -1691,6 +1691,119 @@ def test_divergence_matrix(self):
         with pytest.raises(_tskit.LibraryError, match="UNSUPPORTED_STAT_MODE"):
             ts.divergence_matrix([0, 1], sizes, ids, mode="node")
 
+    def test_pair_coalescence_counts(self):
```
jeromekelleher, line 1694: We want small tests that do one thing here. (Older tests in tskit were quite sprawling - we're trying to move to better testing).

I would make a new test class here, like
```
class TestPairCoalescecnceErrors:
    def run(sample_sets=None, sample_set_sizes=None, indexes=None, time_windows=None)
         ts = # some reasonable example
         # generate some reansonable defaults for unspecified parameters
        return ts.pair_coalescence_counts(...)
   
@pytest.mark.parametrize("node", [-1, -2, 10**6)
   def test_bad_node_in_sample_sets(self, node):
         with pytest.raises(...)
               self.run(sample_sets=[node])
```

It's a bit tedious to set up, but it's way more useful in terms of finding problems and fixing them.

jeromekelleher (commented):
Looks great! There's a lot going on in there, and I've made a few comments style-wise to make things a bit more idiomatic (mostly just change loop conditions to <). There's a few bugs in there that'll only show up in error conditions though, so we do need some more testing of those.

I think a separate C test case where you take a simple case, and then try to provoke all the input errors around time windows etc would be the best approach.

For the Python C API, we have to be super thorough here and make sure that any non-memory allocation failure cases are covered in test_low_level.py. I know it's a lot of boiler plate, but this is what's involved I'm afraid.

We also want to make sure that the reference counting stuff is working properly. The best way I've found of doing this is to run the code in a tight loop, and watch ``top``.

Can you do something like this and check please?
```
ts = # some intermediate size example, so that loops are quick
for _ in range(1000000):
      ts.pair_coalescence_counts(...)
```

The memory usage should stabilise after 30 seconds or so, and then remain constant. Then, try commenting out some of the XDECREFs, and see what happens to memory usage.

c/tskit/trees.c
nspope: The reason I'd settled on int rather than tsk_id_t (int64) for these, is that there's a compiler warning/error due to precision loss when multiplying int64 weight with double below. I think if I explicitly cast from int64 to double I can avoid the warnings-- or are there issues with this?

python/_tskitmodule.c
nspope: The time windows thing is tricky. We want to support 3 cases:
1. Per-node output, in which case nodes with the same timepoint are separate rows
2. Per-timepoint output, in which case nodes with the same timepoint are pooled into the same row (and row order is sorted by time)
3. Per-window output, in which case nodes are pooled to windows.

(2) and (3) can be achieved by passing in the right numpy array. However, (1) needs special handling, which is what the `num_time_windows == 0` branch handles.

nspope:
Thanks @jeromekelleher! All is clear to me except how to handle the special case where time windows are nodes (which is an important one). I left some comments above describing the issue. Briefly, I'm detecting the "nodes" case via a zero-length time windows array. This can't be replaced by an array of unique times in the tree sequence, because we want to preserve node identity even if its age is not unique, and also the order of nodes in the output array. 

I suppose an alternative (to using a zero-length time windows array) is to use `options` to store a flag which tells us we are in `nodes` mode (like is done for the general stats). What do you think is the cleanest option here?

c/tskit/trees.c
jeromekelleher: Cast int64 to double then, that's definitely safe in the ranges we'll be dealing with

python/_tskitmodule.c
jeromekelleher: Hmm, ok. Maybe we should have an explicit mode option then or something?

nspope: That seems like a good idea-- I'll deal with the other issues first, then we can work out the right way to handle "nodes" mode.

jeromekelleher: Yeah, let's get a version with required time windows merged, and figure out the other two cases in a follow up

nspope:
Getting another random windows crash in the CI, but valgrind isn't showing any issues. I traced it back to commit 6287b84 which switched from using `int` to `double`, and switched to using the row access macro.

jeromekelleher:
I really would prefer if we could back out of doing the three different options here and focus on the simpler case of just a fixed set of required time windows first. There's a lot of complexity here already, and I want to make sure we catch all of the boring C level stuff before we have to think about that extra level of trickiness.

So, can we leave that bit for a follow up, and focus on getting 100% test coverage on the bits that should be covered? That means more tests in C, and breaking up the Python C tests like I suggested.

nspope:
So I should pytest.mark.skip any tests using the "nodes" output for the time being? These are the bulk of the tests, which is why I tried to support that mode.

I'm working on extending the test_lowlevel case to get the CPython covered, but I think all the C code paths are now covered aside from OOM.

nspope:
Here's memory consumption over time:



looks to me like it's stabilizing -- is this what you were after @jeromekelleher ?

Code here:
<details>
<summary>monitor mem usage with ps</summary>

```bash
#!/usr/bin/env bash

echo '
import tskit
import numpy as np
import msprime
ts = msprime.sim_ancestry(
  1000, sequence_length=1e5, recombination_rate=1e-8, population_size=1e4
)
time_windows = np.array([0, np.max(ts.nodes_time)/2, np.inf])
for _ in range(100000):
  ts.pair_coalescence_counts(time_windows=time_windows)
' >del_memory_monitor.py

rm -f del_memory_monitor.log
python del_memory_monitor.py &
for i in {1..1000}; do
  ps -eo cmd,%mem,vsz,rsz | grep "python del_memory_monitor" >> del_memory_monitor.log
  sleep 0.0001
done 
rm del_memory_monitor.py

echo '
import numpy as np
vsz, rsz = np.loadtxt("del_memory_monitor.log", usecols=[2,3]).T
step = np.arange(vsz.size) * 0.0001
plt.plot(step, vsz / np.max(vsz), "-", label="VSZ")
plt.plot(step, rsz / np.max(rsz), "-", label="RSZ")
plt.ylabel("Memory usage / max(memory usage)")
plt.xlabel("Wall time (relative)")
plt.xscale("log")
plt.legend()
plt.savefig("del_memory_monitor.png")
' | python
rm del_memory_monitor.log

```
</details>

jeromekelleher:
Nice - did you try it with some decrefs commented out?

nspope:
this is with results_array XDECREF commented out. Looks qualitatively similar ... ?

jeromekelleher:
Hmm, you need bigger examples then. You should be leaking memory heavily there.

nspope:
Ah, the output array was too small. 

Here's with XDECREF commented out:


Here's with XDECREF:


looks sane to me.

<detail>
<summary>memory monitoring code for real</summary>

```bash
#!/usr/bin/env bash

echo '
import tskit
import numpy as np
import msprime
ts = msprime.sim_ancestry(
  100, sequence_length=1e6, recombination_rate=1e-8, population_size=1e4
)
time_windows = np.append(np.unique(ts.nodes_time), np.inf)
windows = np.linspace(0, ts.sequence_length, 100)
sample_sets = list(ts.samples())
sample_set_sizes = [ts.num_samples]
indexes = [(0, 0)]
for _ in range(100000):
    ts.ll_tree_sequence.pair_coalescence_counts(
        time_windows=time_windows, 
        windows=windows, 
        indexes=indexes, 
        sample_sets=sample_sets,
        sample_set_sizes=sample_set_sizes,
        span_normalise=False,
    )
' >del_memory_monitor.py

python del_memory_monitor.py &>del_memory_monitor.out &
sleep 1
for i in {1..1000}; do
  ps -eo cmd,%mem,vsz,rsz | grep "python del_memory_monitor"
  sleep 0.1
done >del_memory_monitor.log
echo `wc -l del_memory_monitor.log`

wait

echo '
import numpy as np
vsz, rsz = np.loadtxt("del_memory_monitor.log", usecols=[2,3]).T
step = np.arange(vsz.size) * 0.1
plt.plot(step, vsz / np.max(vsz), "-", label="VSZ")
plt.plot(step, rsz / np.max(rsz), "-", label="RSZ")
plt.ylabel("Memory usage (divided by max)")
plt.xlabel("Wall time (relative to start)")
plt.legend()
plt.savefig("del_memory_monitor.png")
' | python
```
</details>

nspope:
Maybe the thing to do with time windows is change the input in `ts.ll_pair_coalescence_counts` from `time_windows` to `node_bin`, where `node_bin` maps nodes to bins (which could be time bins, or any other delineation). Then the Python API can internally do the binning, which allows all three cases of time windows (nodes, unique time points, time intervals) to be passed to the low-level routine.

jeromekelleher:
Nice, that sounds like a good idea. Do you want to push forward with that, or merge this much and iterate? (I haven't reviewed again, will need to take another good look)

nspope:
I gave it a shot, but it looks like I'm getting stochastic crashes on Windows. Totally perplexed as to why, because the semantics are very similar to the last commit, which worked.

(It could be because I had skipped a bunch of tests in the last commit?)

nspope:
ugh, figured it out -- a very obscure issue. If you have `an_uint64 = 1 + (tsk_size_t) an_int32` then there's UB on windows. Presumably this has something to do with type promotion. May be related to the infamous `(uint64 + 1) == double` behavior in numpy (https://github.com/numpy/numpy/issues/15084).

nspope:
Anyway, I think this is ready for another look @jeromekelleher

nspope:
No, I spoke to soon -- after restarting the tests, I'm still seeing the crash. Do you have any suggestions on how to debug here @jeromekelleher ?

nspope:
Working now-- bad malloc (d'oh). This is ready for another look whenever you have time @jeromekelleher .

Here's mem usage once more, with the `time_windows -> nodes_output_map` change:
<p>


</p>

c/tskit/trees.c
```diff
@@ -8252,10 +8252,10 @@ tsk_treeseq_pair_coalescence_stat(const tsk_treeseq_t *self, tsk_size_t num_samp
     nodes_parent = tsk_malloc(num_nodes * sizeof(*nodes_parent));
     nodes_sample = tsk_calloc(num_nodes * num_sample_sets, sizeof(*nodes_sample));
     sample_count = tsk_malloc(num_nodes * num_sample_sets * sizeof(*sample_count));
+    outside = tsk_malloc(num_sample_sets * sizeof(*outside));
```
petrelharp, line 8274: d'oh!

nspope: yup :-)

jeromekelleher (commented):
LGTM! Just needs a couple of more tests on the low-level input, and we're good to merge I think

nspope:
Done!

It occurs to me that the order of dimensions here -- (windows, time_windows, indexes) -- should be consistent with the (forthcoming) time windowed stats. @petrelharp what order did you guys settle on?

jeromekelleher (approved):
Superb stuff @nspope - there's a lot of things to get up to speed on here and you've done a great job :+1:

petrelharp:
>  (windows, time_windows, indexes) 

Same!

PR #2934, opened 2024-04-23
Two locus branch stats python prototype

lkirk:
I've re-opened a PR to avoid some issues with build caching. Please see the original discussion on #2912.

Currently, this algorithm creates a matrix of LD, performing a pairwise comparison of all trees in the tree sequence.

This implementation lacks windows/positions, sample sets and polarisation. The outputs of the code produce results in units of branch length, needing to be multiplied by mu^2 or divided by product of the total branch length of the two trees.

This algorithm works by keeping a running sum of the statistic between two trees, updating each time we encounter a branch addition or removal. The tricky part is that we have to remove or add LD contributed by samples that already existed or that will remain under a given node after the addition/removal of branches.

We include a validation against the original formulation of this problem, by including an implementation that was described in McVean
2002. The original formulation computing the covariance of tMRCAs of 2, 3, and 4 samples of individuals on the trees in question. This implementation has several limitations 1) it is very slow and 2) it does not work for trees that are decapitated, as certain samples do not have MRCAs.

### DIFF
--- a/python/tests/test_ld_matrix.py
+++ b/python/tests/test_ld_matrix.py
@@ -24,8 +24,12 @@ Test cases for two-locus statistics
 """
 import contextlib
 import io
+from copy import deepcopy
+from dataclasses import dataclass
+from itertools import combinations
 from itertools import combinations_with_replacement
 from itertools import permutations
+from itertools import product
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -38,6 +42,7 @@ import numpy as np
 import pytest
 
 import tskit
+from tests import tsutil
 from tests.test_highlevel import get_example_tree_sequences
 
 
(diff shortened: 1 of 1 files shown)


lkirk:
@jeromekelleher I'm still seeing issues with the build cache. Is there anything I can do to help here? I'm not in a huge rush for what it's worth, just wanted to close this out.

jeromekelleher:
@benjeffery any advice here?

benjeffery:
> @benjeffery any advice here?

I'm in progress with this at #2935

jeromekelleher:
We've pushed through a few changes in the CI setup, can you rebase now please @lkirk? Hopefully things will go a bit more smoothly now.

lkirk:
Thank you @benjeffery for fixing that!

@jeromekelleher This is ready now, I chopped many more slow tests from this. I realized that there are only 2 cores being used on the test servers and I hadn't anticipated how the slow tests would compound in runtime. As written, things were taking ~1.5 hours and I've adjusted things so now only important tests are considered, adding about 2 minutes to the test runtime.

Does this seem reasonable?

jeromekelleher:
2 minutes is still quite a bit. You could just use the naive version on a handful of small cases to verify things are lining up as expected, and defer doing tests on the full range of examples until the C version is in place?

benjeffery:
Any slow tests should be marked with `@pytest.mark.slow`.

jeromekelleher (approved):

benjeffery:
@mergifyio rebase

PR #2935, opened 2024-04-24
Fix osx tests

benjeffery:


benjeffery:
Well this was fun. The base osx image has changed, so conda paths have moved, libraries that we need were missing, and some tests failed due to tiny numerical differences.

PR #2938, opened 2024-04-25
extend path algorithm (extend_edgesV2)

hfr1tz3:
New extension (pun intended) to the `extend_edges` algorithm now called `extend_paths` (subject to change). We noticed that certain examples within unary regions would not be extended simply with extend edges. We propose this algorithm to handle examples like the following:
For edge path $a \to b \to c$ in tree $T_1$ and edge $a \to d \to c$ in $T_2$, assuming $d\notin T_1$ and $b\notin T_2$ and $t(b)>t(d)$ we should have extended path $a\to b \to d \to c$ in tree $T_2$. 
Unit tests still need to be built, and the algorithm needs to be cleaned. 

We hypothesize that using this alongside extend edges could be the optimal strategy for minimizing edges in tree sequences.


petrelharp:
Proposal for the basis for a test:
PXL_20240508_000130292

petrelharp:
I've added two things:
1. a simple check based on the previous comment for whether a ts is no longer extendable; so if our algorithm terminates before max_iter, then that test should agree; and
2. an implementation of extend paths that just rebuilds the whole tree sequence after every tree, so is a lot simpler. (I called it "naive", but it's not actually simple enough to be 'naive'; I had to deal with quite a few fiddly things for it.) It's real slow, so can't be used on anything at all big.

Next step: compare it to the non-naive version.

petrelharp:
This is hopefully ready for you to look at, @nspope!

One question is that currently `max_iter` is required; but, should we provide a default? Should we do a warning if it hasn't terminated (probably)?

hfr1tz3:
> One question is that currently `max_iter` is required; but, should we provide a default? Should we do a warning if it hasn't terminated (probably)?
I think that makes sense. I haven't checked if we each reach the max_iter value for the long sequences we look at, but I feel like it terminates before that value in most cases.

petrelharp:
Okay - now it's ready for reals, @nspope. The 'partial coverage' things that codecov flags are cases where we're like
``` 
if (A && B) { ... }
```
but then not all four combinations of `A` and `B` ever get seen. This let me identify one place that could be simplified, but I've looked at the others and I don't think any of the others can. They *are* getting hit by the python tests, and constructing C tests for all of them would be a Big Pain.

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+{
+    int ret = 0;
+    tsk_id_t tj;
+    tsk_size_t num_nodes = tsk_treeseq_get_num_nodes(ts);
+
+    tsk_memset(self, 0, sizeof(haplotype_extender_t));
+
+    self->ts = ts;
+    self->edges = edges;
+    tsk_edge_table_copy(&ts->tables->edges, self->edges, TSK_NO_INIT);
```
nspope, line 7929: Shouldn't there be a return code check here?

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
     }
 
-out:
-    tsk_tree_free(&tree);
+    ret = haplotype_extender_extend_paths(&haplotype_extender);
+    if (ret != 0) {
+        goto out;
+    }
 
+    ret = haplotype_extender_free(&haplotype_extender);
```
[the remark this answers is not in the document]
nspope: also, is a return code check actually needed for `haplotype_extender_free` here?

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+    if (self->direction == TSK_DIR_FORWARD) {
+        valid = tsk_tree_position_next(&tree_pos);
+    } else {
+        valid = tsk_tree_position_prev(&tree_pos);
+    }
+
+    while (valid) {
+        left = tree_pos.interval.left;
+        right = tree_pos.interval.right;
+        haplotype_extender_next_tree(self, &tree_pos);
```
nspope, line 8288: Need a return code check here?

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+        max_new_edges = 0;
+        next_max_new_edges = INFINITY;
+        while (max_new_edges < INFINITY) {
+            for (ex_in = self->edges_in_head; ex_in != NULL; ex_in = ex_in->next) {
+                e_in = ex_in->edge;
+                c = self->edges->child[e_in];
+                if (self->last_degree[c] > 0) {
+                    ne = haplotype_extender_mergeable(self, c);
+                    if (ne <= max_new_edges) {
+                        haplotype_extender_merge_paths(self, c, left, right);
```
nspope, line 8298: Return check needed here?

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+        old_parent = TSK_NULL;
+    }
+    if (new_parent != old_parent) {
+        if (self->parent_out[child] == new_parent) {
+            // if our new edge is in edges_out, it should be extended
+            e_out = self->last_nodes_edge[child];
+            self->far_side[e_out] = there;
+            edge_list_set_extended(&self->edges_out_head, e_out);
+        } else {
+            e_out = tsk_edge_table_add_row(
```
nspope, line 8120: `e_out` could be an error code here (negative) -- so need a check?

nspope (commented):
LGTM, caught a few places where return codes should be checked

c/tskit/trees.c
```diff
@@ -7814,254 +7835,541 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+            && self->edges->right[e] == self->edges->left[e + 1]) {
+            self->edges->right[e] = self->edges->right[e + 1];
+            self->edges->left[e + 1] = self->edges->right[e + 1];
+        }
+    }
+    for (e = 0; e < (tsk_id_t) num_edges; e++) {
+        keep[e] = self->edges->left[e] < self->edges->right[e];
     }
+    ret = tsk_edge_table_keep_rows(self->edges, keep, 0, NULL);
+out:
```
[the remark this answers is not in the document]
petrelharp: huh, I wonder why valgrind didn't catch that?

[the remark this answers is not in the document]
petrelharp: Oh I see
```
int
tsk_tree_position_free(tsk_tree_position_t *TSK_UNUSED(self))
{
    return 0;
}
```

c/tskit/trees.c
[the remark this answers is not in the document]
petrelharp: yep - right on both counts

petrelharp:
Okay - @jeromekelleher, I think this is ready for a quick look (and then we can merge it! :tada:)

c/tskit/trees.c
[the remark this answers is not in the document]
nspope: ah -- but prolly good to have in there in case the free stops being a placeholder

petrelharp:
Wait two more things:
- provide some kind of feedback for `max_iter`
- one remaining uncovered code bit

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+            }
+        }
+        self->next_nodes_edge[child] = e_out;
+        self->next_degree[child] += 1;
+        self->next_degree[new_parent] += 1;
+        self->parent_out[child] = TSK_NULL;
+        if (old_edge != TSK_NULL) {
+            for (ex_in = self->edges_in_head; ex_in != NULL; ex_in = ex_in->next) {
+                e_in = ex_in->edge;
+                if (e_in == old_edge && (ex_in->extended == 0)) {
```
petrelharp, line 8151: we just need a C test for which `ex_in->extended != 0` here

petrelharp: ah ha - turns out this was not needed

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+        } else {
+            break;
+        }
     }
-    edges_in_head = NULL;
-    edges_in_tail = NULL;
-    edges_out_head = NULL;
-    edges_out_tail = NULL;
-    e_out = 0; // only to avoid an 'maybe uninitialized' compile warning
+    if ((num_extended == 0) || (p_in != p_out) || (p_in == TSK_NULL)) {
```
petrelharp, line 8213: we don't have a C test where `p_in != p_out` but `num_extended == 0` (this happens in python though)

petrelharp:
Okay - now I've legit hit everything we can with tests (and simplified some things along the way). @jeromekelleher - could you just throw a quick eyeball at this?

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
     const tsk_size_t num_nodes = tables->nodes.num_rows;
-    const tsk_size_t num_edges = tables->edges.num_rows;
-    tsk_id_t *degree = tsk_calloc(num_nodes, sizeof(*degree));
-    tsk_id_t *out_parent = tsk_malloc(num_nodes * sizeof(*out_parent));
-    tsk_bool_t *keep = tsk_calloc(num_edges, sizeof(*keep));
-    bool *not_sample = tsk_malloc(num_nodes * sizeof(*not_sample));
-
-    tsk_memset(&edge_list_heap, 0, sizeof(edge_list_heap));
-    tsk_memset(&tree_pos, 0, sizeof(tree_pos));
+    double *sites_position = tables->sites.position;
```
jeromekelleher, line 7863: These should probably be ``const``?

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+        self->near_side = self->edges->right;
+        self->far_side = self->edges->left;
+    }
+
+    self->edges_in_head = NULL;
+    self->edges_in_tail = NULL;
+    self->edges_out_head = NULL;
+    self->edges_out_tail = NULL;
+
+    tsk_memset(&self->edge_list_heap, 0, sizeof(self->edge_list_heap));
```
jeromekelleher, line 7948: You don't need to memset the edge_list_heap to 0 here as its already been done above when you zero'd the top level struct.

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
     }
 
-    if (forwards) {
-        near_side = edges->left;
-        far_side = edges->right;
+out:
+    return ret;
+}
+
+// static void
```
jeromekelleher, line 7982: I tend to make the ``print_state`` a part of the public interface so that it can be tested and it stays up to date. I guess the struct here is not exposed in the header so there's no way to access it. We worked around this for simplify by adding a ``TSK_DEBUG`` option [here](https://github.com/tskit-dev/tskit/blob/024e0425139efcf8b6c30229699910a5757a71fa/c/tskit/tables.c#L12106)

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+                goto out;
+            }
+            /* pointers to left/right might have changed! */
+            if (self->direction == TSK_DIR_FORWARD) {
+                self->near_side = self->edges->left;
+                self->far_side = self->edges->right;
+            } else {
+                self->near_side = self->edges->right;
+                self->far_side = self->edges->left;
+            }
```
jeromekelleher, line 8137: Maybe allocate the new edge list node here and pass to ``edge_list_append_entry`` rather than passing around the ``edge_list_heap``? It's not much extra code, and will save you some a few unreachable lines. It seemed odd to me on reading to pass the allocator in rather than the actual pointer.

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+    // returns the number of new edges needed
+    // if the paths in parent_in and parent_out
+    // up through nodes that aren't in the other tree
+    // end at the same place and don't have conflicting times;
+    // otherwise, return infinity
+    tsk_id_t p_in, p_out, child;
+    float num_new_edges; // needs to be float so we can have infinity
+    int num_extended;
+    double t_in, t_out;
+    bool climb_in, climb_out;
```
jeromekelleher, line 8180: Lotta pointer chasing here, I would do
```
const double *nodes_time =  self->ts->tables->nodes.time;
```
Faster, and less typing!

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+    haplotype_extender_t *self, tsk_id_t c, double left, double right)
+{
+    int ret = 0;
+    tsk_id_t p_in, p_out, child;
+    double t_in, t_out;
+    bool climb_in, climb_out;
+
+    p_out = self->parent_out[c];
+    p_in = self->parent_in[c];
+    t_out = self->ts->tables->nodes.time[p_out];
```
jeromekelleher, line 8230: Same here - you don't want to be chasing 3 levels of pointers here in a loop. Grab a const pointer at the start of the function and the compiler will be able to do a *way* better job.

c/tskit/trees.c
```diff
@@ -7814,75 +7835,455 @@ remove_unextended(edge_list_t **head, edge_list_t **tail)
...
+{
+    int ret = 0;
+    bool valid;
+    double left, right;
+    float ne, max_new_edges, next_max_new_edges;
+    tsk_tree_position_t tree_pos;
+    edge_list_t *ex_in;
+    tsk_id_t e_in, c, e;
+    tsk_size_t num_edges;
+    tsk_bool_t *keep;
```
jeromekelleher, line 8278: Need to initialise ``keep`` to NULL

jeromekelleher (commented):
LGTM. Very nice, I like the new design. I spotted a few small problems and a couple of suggestions on layout.

If you're looking for performance, I think there's a modest gain on the table from taking pointers to stuff within a function rather than lots of ``self->x->y`` within loops.   I wouldn't bother for now though, except in places where it'll also improve readability.

petrelharp:
Thanks for the pointer poitners! I think this is ready to merge (assuming tests pass).

petrelharp:
Argh, yes - I just missed one cast. Ready now, hopefully!

PR #2945, opened 2024-05-03
Drop metadata schemas on tables.

### LINKED ISSUE #2944: Add XTable.drop_metadata
It's useful to drop metadata on tables. The only question really is how to deal with schemas, how about:

```python
def drop_metadata(self, keep_schema=False):
     """
     Drops all metadata in this table. By default, the schema is also cleared, except if ``keep_schema`` is True.
     """
```

jeromekelleher:
Closes #2944

Mostly done, just needs a quick test on drop_schemas

Some incidental Black-update noise also


benjeffery:
Merging and rebasing on #2946 will fix the noise.

benjeffery (approved):
Nice, LGTM!

PR #2948, opened 2024-05-09
time windows in statistics

petrelharp:
Here @tforest and I are starting in on adding time windows to statistics. We're starting with what was sketched out in #683, and will explain things in more detail here when we're farther along (ignore this for now).


petrelharp:
Note: it is not clear how to do this for `site` statistics, since the site stat is of the form
$$ \sum_a f(w_a) $$
where the sum is over alleles, and $w_a$ is the weight of all samples with allele $a$;
however, it is *mutations* that have times, not alleles.

The proposal will probably be to compute a site stat that sums over mutations, not alleles, but we'll start with `branch` stats only for now.

petrelharp:
Next step:
- do the AFS first, since it's less tangled up

Also maybe:
- allow `ts.decapitate( )` to take `inf` as an argument (that does nothing) ?

andrewkern:
a small nudge here that i mentioned to @petrelharp in passing-- it would be great to have an _expectation_ from theory as to what time stratified quantities like the SFS should be under the (standard, neutral) coalescent

tforest:
Some thoughts after working on time windows.

After these edits the moment the output of, let's say, the AFS is a still 2D array of windows, same for time_windows, when using either of them individually. However, when using windows and time_windows at the same time, the output is a 3D array, with the following shape: [num_windows][num_time_windows][sample_size]. When windows or time_windows are None, associated dimensions are dropped accordingly. 
As there is now two types of windows, it will become ambiguous that the historical "windows" parameter is in fact corresponding specifically to genomic spanning windows. We did not renamed it for now though, as it would break previous behavior.

Some ideas:

- Add new benchmarks for summary stats to see if the implemented features are optimized both in terms of computational space and time complexity.
- Add some plots for summary stats to observe how time windows impact them.

petrelharp:
A note on the potential confusion between `windows` and `time_windows` - often one endpoint of the `time_windows` will be Inf, so if we make sure we produce an informative error if the `windows` aren't finite, we'll help people avoid the mistake.

python/tests/test_tree_stats.py
```diff
@@ -144,39 +144,93 @@ def windowed_tree_stat(ts, stat, windows, span_normalise=True):
...
-    for tree in ts.trees():
-        x = np.zeros((ts.num_nodes, k))
-        x[ts.samples()] = w
-        for u in tree.nodes(order="postorder"):
-            for v in tree.children(u):
-                x[u] += x[v]
-        if polarised:
-            s = sum(tree.branch_length(u) * f(x[u]) for u in tree.nodes())
+    sigma = np.zeros((ts.num_trees, tw, m))
+    for j, upper_time in enumerate(time_windows[1:]):
```
petrelharp, line 163: This assumes that `time_windows[0]` is `0`, I think.

petrelharp: This could be fixed by setting `sigma` not to 0 but to -1 times the value calculated from `ts.decapitate(time_windows[0])`.

python/tests/test_tree_stats.py
```diff
@@ -3519,47 +3631,84 @@ def naive_site_allele_frequency_spectrum(
...
-
-            # Advance the trees
-            more = [tree.next() for tree in trees]
-            assert len(set(more)) == 1
-            if not more[0]:
-                break
-        if span_normalise:
-            S /= end - begin
-        out[j, :] = S
+        for k, upper_time in enumerate(time_windows[1:]):
```
petrelharp, line 3659: see comment above - I think this assumes `time_windows[0] == 0`

python/tests/test_tree_stats.py
```diff
@@ -144,39 +144,93 @@ def windowed_tree_stat(ts, stat, windows, span_normalise=True):
...
         if span_normalise:
             for j, tree in enumerate(ts.trees()):
                 sigma[j] /= tree.span
-        return sigma
+        out = sigma
     else:
-        return windowed_tree_stat(ts, sigma, windows, span_normalise=span_normalise)
+        out = windowed_tree_stat(ts, sigma, windows, span_normalise=span_normalise)
+    if drop_time_windows:
+        # beware: this assumes the first dimension is windows
```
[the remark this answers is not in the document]
petrelharp: hm but perhaps replaced by
```
assert len(out.shape) == 3
```

python/tests/test_tree_stats.py
```diff
@@ -3519,47 +3631,84 @@ def naive_site_allele_frequency_spectrum(
...
+                more = [tree.next() for tree in trees]
+                assert len(set(more)) == 1
+                if not more[0]:
+                    break
+            if span_normalise:
+                S /= end - begin
+            out[j, k, :] = S
+
+    if drop_time_windows:
+        # beware: this assumes the first dimension is windows
```
petrelharp, line 3692:
```suggestion
        assert len(out.shape) == 3
```

python/tests/test_tree_stats.py
```diff
@@ -3519,47 +3631,84 @@ def naive_site_allele_frequency_spectrum(
...
+        assert out.shape[1] == 1
+        out = out[:, 0]
+    elif drop_windows:
+        # drop windows dim if only using time windows
```
petrelharp, line 3696:
```suggestion
```
we might drop the windows if we are *not* using time windows also

python/tests/test_tree_stats.py
```diff
@@ -3519,47 +3631,84 @@ def naive_site_allele_frequency_spectrum(
...
+        out = out[0]
+        # assert out.shape[0] == 1
```
petrelharp, line 3698: This should still be okay, no?
```suggestion
        assert out.shape[0] == 1
        out = out[0]
```

python/tests/test_tree_stats.py
```diff
@@ -3519,47 +3631,84 @@ def naive_site_allele_frequency_spectrum(
...
+    # Warning: when using Windows and TimeWindows,
+    # the output has three dimensions
```
petrelharp, line 3700: can delete this

python/tests/test_tree_stats.py
```diff
@@ -3604,23 +3759,29 @@ def branch_allele_frequency_spectrum(
...
 
-    def update_result(window_index, u, right):
-        if 0 < count[u, -1] < ts.num_samples:
-            x = (right - last_update[u]) * branch_length[u]
-            c = count[u, :num_sample_sets]
-            if not polarised:
-                c = fold(c, out_dim)
-            index = tuple([window_index] + list(c))
-            result[index] += x
+    def update_result(window_index, u, right, time_windows):
```
petrelharp, line 3762: `time_windows` isn't being changed by this function
```suggestion
    def update_result(window_index, u, right):
```

python/tests/test_tree_stats.py
```diff
@@ -3604,23 +3759,29 @@ def branch_allele_frequency_spectrum(
...
+        for k_tw, _ in enumerate(time_windows[:-1]):
+            if 0 < count[u, -1] < ts.num_samples:
+                # interval between child and parent inside the window
+                t_v = branch_length[u] + time[u]
+                tw_branch_length = min(time_windows[k_tw + 1], t_v) - max(
+                    time_windows[0], time[u]
```
petrelharp, line 3768: Whoops?
```suggestion
                    time_windows[k_tw], time[u]
```

python/tests/test_tree_stats.py
```diff
@@ -3604,23 +3759,29 @@ def branch_allele_frequency_spectrum(
...
-            c = count[u, :num_sample_sets]
-            if not polarised:
-                c = fold(c, out_dim)
-            index = tuple([window_index] + list(c))
-            result[index] += x
+    def update_result(window_index, u, right, time_windows):
+        for k_tw, _ in enumerate(time_windows[:-1]):
+            if 0 < count[u, -1] < ts.num_samples:
+                # interval between child and parent inside the window
+                t_v = branch_length[u] + time[u]
```
petrelharp, line 3766: Here we're losing the advantage of cacheing `branch_length[u]`; this might as well be `time[v]` (with `v` passed in also).

Alternatives:
1. let `branch_length` be a (num nodes x num time windows) array instead of just a vector, so that we'd have
```
tw_branch_length = branch_length[u, k_tw]
```
2. Do the calculation [down a few lines](https://github.com/tskit-dev/tskit/pull/2948/files#diff-4d6ee7e1fdeb228e5a8a5fe5c37af121dc87237d0c30563eafab80282fe3ed0eR3781), something like this (this is not right):
```
u = edge.child
v = edge.parent
t_c = time[u]
t_p = time[v]
time_window_index = 0
while t_p < time_windows[time_window_index + 1]:
    while v != -1:
        tw_branch_length = min(time_windows[k_tw + 1], t_p) - max(time_windows[k_tw], t_c)
        update_result(window_index, time_window_index, v, t_left, tw_branch_length)
        count[v] -= count[u]
        t_c = t_p
        v = parent[v]
        t_p = time[v]
    time_window_index += 1
```
The advantage to this is that computation isn't increased by a factor of (num time windows). The disadvantage might be that the code is harder to understand?

python/tests/test_tree_stats.py
```diff
@@ -3651,14 +3812,23 @@ def update_result(window_index, u, right):
...
+                update_result(window_index, u, w_right, time_windows)
             window_index += 1
         tree_index += 1
 
     assert window_index == windows.shape[0] - 1
     if span_normalise:
         for j in range(num_windows):
             result[j] /= windows[j + 1] - windows[j]
+
+    if drop_time_windows:
```
petrelharp, line 3824: see suggestions above

petrelharp (commented):
Looks good! One question about a possible refactor, and suggesting moving the "general stat" stuff to a different PR.

benjeffery:
I've added this work to the next release milestone. Hoping to get a release out in a week or two, if that is too ambitious for this let me know.

petrelharp:
Probably too ambitious, but we might have something in by then.

python/tskit/trees.py
```diff
@@ -7637,6 +7637,7 @@ def parse_windows(self, windows):
         # Note: need to make sure windows is a string or we try to compare the
         # target with a numpy array elementwise.
         if windows is None:
+            # initiate default spanning windows
```
petrelharp, line 7640:
```suggestion
```

petrelharp (commented):
This looks great! Some suggestions, mostly minor; let's discuss getting the tests in there.

c/tskit/trees.c
```diff
@@ -3484,10 +3513,25 @@ tsk_treeseq_site_allele_frequency_spectrum(const tsk_treeseq_t *self,
     return ret;
 }
 
+#define MAX(a,b) ((a) > (b) ? (a) : (b))
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
```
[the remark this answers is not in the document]
tforest: YES! Actually. Thanks!

c/tskit/trees.c
```diff
@@ -1232,6 +1232,35 @@ tsk_treeseq_check_windows(const tsk_treeseq_t *self, tsk_size_t num_windows,
...
+
+    if (num_windows < 1) {
+        ret = TSK_ERR_BAD_NUM_WINDOWS;
+        goto out;
+    }
+
+    if (windows[0] < 0) {
+	goto out;
+    }
+    if (windows[num_windows] > INFINITY) {
```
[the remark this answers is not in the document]
tforest: Correct, you're right, it would be difficult to initialize something larger than _INFINITY_! I'm removing this.

c/tskit/trees.c
```diff
@@ -3497,24 +3541,32 @@ tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double righ
     tsk_size_t *coordinate = tsk_malloc(num_sample_sets * sizeof(*coordinate));
```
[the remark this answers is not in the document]
tforest: HA! IDK, I'll let it there for now, but yeah maybe!

c/tskit/trees.c
```diff
@@ -3497,24 +3541,32 @@ tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double righ
...
-        afs = result + afs_size * window_index;
-        for (k = 0; k < num_sample_sets; k++) {
-            coordinate[k] = (tsk_size_t) count_row[k];
-        }
-        if (!polarised) {
-            fold(coordinate, result_dims, num_sample_sets);
-        }
-        increment_nd_array_value(afs, num_sample_sets, result_dims, coordinate, x);
+    if (parent[u] != -1){
+	t_v = time[parent[u]];
```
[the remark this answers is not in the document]
tforest: There's no t_u variable, do you think that's necessary? I'm using time[u] here.

c/tskit/trees.c
```diff
@@ -3525,12 +3577,12 @@ tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double righ
...
     tsk_size_t num_sample_sets, double *counts, tsk_size_t num_windows,
-    const double *windows, const tsk_size_t *result_dims, tsk_flags_t options,
-    double *result)
+    tsk_size_t num_time_windows, const double *windows, const double *time_windows,
+    const tsk_size_t *result_dims, tsk_flags_t options, double *result)
 {
     int ret = 0;
     tsk_id_t u, v;
-    tsk_size_t window_index;
+    tsk_size_t window_index, time_window_index;
```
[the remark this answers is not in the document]
tforest: Correct. So the allocation should be done locally too (in the update function), I presume.

python/_tskitmodule.c
```diff
@@ -9077,7 +9077,7 @@ parse_windows(
     npy_intp *shape;
 
     windows_array = (PyArrayObject *) PyArray_FROMANY(
-        windows, NPY_FLOAT64, 1, 1, NPY_ARRAY_IN_ARRAY);
+				      windows, NPY_FLOAT64, 1, 1, NPY_ARRAY_IN_ARRAY);
```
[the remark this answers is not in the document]
tforest: Yes, or probably an indentation that I made by mistake. I put it back how it was before.

c/tests/test_stats.c
```diff
@@ -1258,23 +1270,24 @@ verify_afs(tsk_treeseq_t *ts)
     sample_set_sizes[0] = n - 2;
     sample_set_sizes[1] = 2;
     ret = tsk_treeseq_allele_frequency_spectrum(
-	ts, 2, sample_set_sizes, samples, 0, NULL, 0, NULL, 0, result);
+        ts, 2, sample_set_sizes, samples, 0, NULL, 0, NULL, 0, result);
```
petrelharp, line 1273: uh-oh, are these tabs?

c/tskit/trees.c
```diff
@@ -3755,8 +3744,12 @@ tsk_treeseq_allele_frequency_spectrum(const tsk_treeseq_t *self,
         num_time_windows = 1;
         time_windows = default_time_windows;
     } else {
-        ret = tsk_treeseq_check_time_windows(
-            num_time_windows, time_windows);
+        if (stat_site
+            && tsk_memcmp(time_windows, default_time_windows, sizeof(double)) != 0) {
```
petrelharp, line 3748: Hm, this is a bit awkward - what if instead we used `num_time_windows=0` to mean "default/no time windows"?

petrelharp: Oh, wait - we're already *in* the `else` clause where we know that `time_windows != NULL`. So I don't think we need to check this at all - just throw the error?
```suggestion
            ) {
```

petrelharp: So, if someone explicitly specifies `time_windows = [0, np.inf], mode="node"` then they'll get the error, but that's okay - the error message says "you can't specify time windows", not "time windows must be [0, Inf)" (I think that's what it says anyhow).

c/tskit/trees.c
```diff
@@ -3535,9 +3535,9 @@ tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double righ
                 if (!polarised) {
                     fold(coordinate, result_dims, num_sample_sets);
                 }
-                tw_branch_length
-                    = fabs(TSK_MIN(time_windows[time_window_index + 1], t_v)
-                           - TSK_MAX(time_windows[time_window_index], time[u]));
+                tw_branch_length = TSK_MAX(
+                    0.0, TSK_MIN(time_windows[time_window_index + 1], t_v)
+                             - TSK_MAX(time_windows[time_window_index], time[u]));
```
petrelharp, line 3540: whoops, good catch

petrelharp:
Closing in favor of #3266.

PR #2952, opened 2024-05-23
Switch to ruff linting

benjeffery:
For numpy2 checks etc.

### DIFF
--- a/python/tests/test_divmat.py
+++ b/python/tests/test_divmat.py
@@ -22,6 +22,7 @@
 """
 Test cases for divergence matrix based pairwise stats
 """
+
 import array
 import collections
 import functools
@@ -212,14 +213,8 @@ def branch_divergence_matrix(ts, sample_sets=None, windows=None, span_normalise=
                                 tu = ts.nodes_time[w] - ts.nodes_time[u]
                                 tv = ts.nodes_time[w] - ts.nodes_time[v]
                             else:
-                                tu = (
-                                    ts.nodes_time[local_root(tree, u)]
-                                    - ts.nodes_time[u]
-                                )
-                                tv = (
-                                    ts.nodes_time[local_root(tree, v)]
-                                    - ts.nodes_time[v]
-                                )
+                                tu = ts.nodes_time[local_root(tree, u)] - ts.nodes_time[u]
+                                tv = ts.nodes_time[local_root(tree, v)] - ts.nodes_time[v]
                             d = (tu + tv) * span
                             D[i, j, k] += d
             tree.next()
@@ -303,7 +298,7 @@ def stats_api_matrix_method(
     # contiguous, so that we just look at specific sections of the genome.
     drop = []
     if windows[0] != 0:
-        windows = [0] + windows
+        windows = [0, *windows]
         drop.append(0)
     if windows[-1] != ts.sequence_length:
         windows.append(ts.sequence_length)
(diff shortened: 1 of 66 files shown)


benjeffery:
Well this was a lot noisier than I was expecting, although it did show up a couple of actual errors.

jeromekelleher:
Ruff is quite opinionated about getting people to use f strings etc isn't it?

What were the errors?

LGTM though, happy to switch.

Given the noisiness we may want to try and flush a few open PRs through first.

benjeffery:
If we don't want to enforce f-strings we can switch that off - they do generally read better though.

The errors were:
A test where a subsequent `pytest.raises` was nested inside the first, such that the second didn't run.
An ibd test that was missing an `assert` so the value was evaluated, but not checked.

python/tests/test_divmat.py
```diff
@@ -725,19 +716,20 @@ def test_all_trees_interval(self, interval, mode, span_normalise):
...
+            list(range(27)),
+            [5, 7, 9, 20],
+            [5.1, 5.2, 5.3, 5.5, 6],
+            [5.1, 5.2, 6.5],
         ],
     )
     @pytest.mark.parametrize("mode", DIVMAT_MODES)
     @pytest.mark.parametrize("span_normalise", [True, False])
     def test_all_trees_windows(self, windows, mode, span_normalise):
+        print(windows)
```
hyanwong, line 732: stray print?

hyanwong (approved):
I only had a brief look through, but apart from a stray `print`, everything LGTM. Setting the line length to 90 rather than 89 has made a few lines more readable too.

Is it worth running a perf test, as there have been a few changes involving changing lists to tuples (I assume this should make things a smidge faster, although not noticeably)

python/tskit/trees.py
```diff
@@ -8717,11 +8686,7 @@ def tjd_func(sample_set_sizes, flattened, **kwargs):
             g = np.array([np.sum(1 / np.arange(1, nn) ** 2) for nn in n])
             with np.errstate(invalid="ignore", divide="ignore"):
                 a = (n + 1) / (3 * (n - 1) * h) - 1 / h**2
-                b = (
-                    2 * (n**2 + n + 3) / (9 * n * (n - 1))
-                    - (n + 2) / (h * n)
-                    + g / h**2
-                )
+                b = 2 * (n**2 + n + 3) / (9 * n * (n - 1)) - (n + 2) / (h * n) + g / h**2
```
petrelharp, line 8689: I don't really like this change - the previous version was written out on multiple lines so it's easier to parse by eye. However, most of these many lines -> one line changes I do like, so maybe it's okay?

python/tskit/metadata.py
```diff
@@ -641,8 +638,8 @@ def __init__(self, schema: Mapping[str, Any] | None) -> None:
             except KeyError:
                 raise exceptions.MetadataSchemaValidationError(
                     f"Unrecognised metadata codec '{schema['codec']}'. "
-                    f"Valid options are {str(list(codec_registry.keys()))}."
-                )
+                    f"Valid options are {list(codec_registry.keys())!s}."
+                ) from None
```
petrelharp, line 642: This `raise from None` thing is going to be a bit opaque to new people.

petrelharp:
I'm happy with whatevery you all think, but I did find two places I wished it was less opinionated (see comments).

python/tskit/trees.py
hyanwong: In this one case you could #noqa the line, I guess?

python/tskit/metadata.py
jeromekelleher: I fail to see the value of that one too

molpopgen: It makes the output  easier to read when exceptions occur, although developers may prefer to see the entire context chain of raised errors.

hyanwong:
I've been using ruff and it's much quicker than our previous linter, which makes development a little slicker and more palatable, so I would be happy with switching over.

benjeffery:
Still planning to do this - waiting till the number of in-flight PRs is a bit lower.

benjeffery:
I've just tried to bring this up-to-date and while doing so noticed quite a few new changes that weren't in this original PR as `ruff` had added new checks. As such I don't think we want to switch as our usual PR's moving forward will have lint changes in, `black` has been relatively stable. I'm open to `ruff` once/if things have settled down in a year or two, but right now I think it will be too noisey.

PR #2964, opened 2024-06-20
Compile on numpy2

jeromekelleher:
Part of supporting numpy >= 2. Following the ABI [advice](https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-abi-handling) from numpy, we should be able to compile against numpy version >= 2, and also have a runtime dependency < 2. Hopefully this is all we need to do initially to ship compatible wheels and conda packages (which are also broken)


jeromekelleher:
Looks like this is mostly working now. Outstanding issues:

- Some tests are failing on most recent numpy 1.x version and also on numpy 2.x. 
- We need numpy 2.x compatible kastore distributions, so parking this until that's ready.

jeromekelleher:
Shall I pick this up or is it simpler to start again @benjeffery? The kastore conda-forge packages should be built and available in the next few hours, so that should unblock us here.

benjeffery:
> Shall I pick this up

I'm working on this now!

python/setup.py
```diff
@@ -1,22 +1,18 @@
...
 from setuptools import setup
 from setuptools.command.build_ext import build_ext
 
-
 IS_WINDOWS = platform.system() == "Windows"
 
 
-# Obscure magic required to allow numpy be used as a 'setup_requires'.
-# Based on https://stackoverflow.com/questions/19919905
 class local_build_ext(build_ext):
```
jeromekelleher, line 11: Do we need this at all any more?

python/tests/test_ld_matrix.py
```diff
@@ -1034,7 +1034,7 @@ def pi2_unbiased(
         w_Ab = state[1, k]
         w_aB = state[2, k]
         w_ab = n - (w_AB + w_Ab + w_aB)
-        with suppress_division_by_zero_warning():
+        with np.errstate(over="ignore", divide="ignore", invalid="ignore"):
```
jeromekelleher, line 1037: Wouldn't it be better to update the original function here rather than duplicating?

jeromekelleher (commented):
Actual code changes look small, great!

python/setup.py
benjeffery: Good spot, removed in 03e40911

python/tests/test_ld_matrix.py
benjeffery: Fixed in a8cb721b

python/setup.py
benjeffery: Hmm, removing this worked locally, but not on CI.

benjeffery:
All seems to be working - I'll squash and merge.

benjeffery:
Realised we were doing wheels for 3.12, but not testing on it. Fixing that now

jeromekelleher (commented):
LGTM! What's the strategy for pushing an updated release - cherry pick this back onto last release tag?

benjeffery:
> LGTM! What's the strategy for pushing an updated release - cherry pick this back onto last release tag?

Seems the simplest approach.

PR #2966, opened 2024-06-20
Fix tests for lshmm 0.0.8

### LINKED ISSUE #2962: Fixup tests for lshmm 0.0.6
Recent lshmm release breaks some tests in tskit test suite: https://github.com/astheeggeggs/lshmm/releases/tag/v0.0.6

Would you mind look at this please @szhan

szhan:
## Description

Update tests and requirements to use lshmm 0.0.7. Fixes #2962

# PR Checklist:

- [x] Update tests using lshmm.
- [x] Update development requirements.

### DIFF
--- a/python/tests/test_genotype_matching.py
+++ b/python/tests/test_genotype_matching.py
@@ -4,6 +4,7 @@ import itertools
 import lshmm as ls
 import msprime
 import numpy as np
+import pytest
 
 import tskit
 
@@ -1257,6 +1258,7 @@ class LSBase:
 
     # Define a bunch of very small tree-sequences for testing a collection of
     # parameters on
+    @pytest.mark.skip(reason="No plans to implement diploid LS HMM yet.")
     def test_simple_n_10_no_recombination(self):
         ts = msprime.simulate(
             10, recombination_rate=0, mutation_rate=0.5, random_seed=42
@@ -1264,11 +1266,13 @@ class LSBase:
         assert ts.num_sites > 3
         self.verify(ts)
 
+    @pytest.mark.skip(reason="No plans to implement diploid LS HMM yet.")
     def test_simple_n_6(self):
         ts = msprime.simulate(6, recombination_rate=2, mutation_rate=7, random_seed=42)
         assert ts.num_sites > 5
         self.verify(ts)
 
+    @pytest.mark.skip(reason="No plans to implement diploid LS HMM yet.")
     def test_simple_n_8_high_recombination(self):
         ts = msprime.simulate(8, recombination_rate=20, mutation_rate=5, random_seed=42)
         assert ts.num_trees > 15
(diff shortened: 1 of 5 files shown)


python/tests/test_genotype_matching.py
```diff
@@ -1345,12 +1345,17 @@ def verify(self, ts):
             self.assertAllClose(ll_tree, ll_mirror_tree_dict)
 
             # Ensure that the decoded matrices are the same
+            flipped_G_check = np.flip(G_check, axis=0)
+            flipped_s = np.flip(s, axis=1)
+            num_alleles = ls.core.get_num_alleles(flipped_G_check, flipped_s)
```
jeromekelleher, line 1350: This seems a bit backwards, requiring the client code to call into a private method in lshmm. How about we make ``num_alleles`` an optional argument in ``lshmm``, which if not provided calls ``get_num_alleles`` on the reference panel and query string?

jeromekelleher: Also the reason this is failing is because you've not imported ``lshmm.core``

jeromekelleher (commented):
Thanks for the updates @szhan, this is very helpful.

I do wonder if we should push out another release of lshmm in which ``num_alleles`` is an optional keyword-only arg? This is quite user-unfriendly at the moment.

python/tests/test_genotype_matching.py
szhan: I'm planning on getting rid of that `num_alleles` argument altogether in the next release lshmm. It is done this way right now because of how input data are being handled in the diploid case.

jeromekelleher: Ok, want to fixup here so by importing lshmm.core?

szhan: I think I'll get the next release of lshmm ASAP today.

szhan: Hopefully, `0.0.7` will take care of it. https://github.com/astheeggeggs/lshmm/releases/tag/v0.0.7

jeromekelleher: Great - I think you probably need to update the version pins in the various ``requirements`` files. Do a search for lshmm in there

szhan: Updated tests and pins.

szhan:
The requirements that the first value of recomb. prob. array being zero and all sites in the input data must be variable are removed.

szhan:
The tests seem to be caused by not excluding `MISSING` when counting the number of alleles to scale mutation rates.

jeromekelleher:
I think we can skip the failing diploid - I'm not sure how soon the diploid model will actually be implemented, so let's not worry about it. Just mark the failing tests as skip.

jeromekelleher:
I don't know either to be honest. Let's not get bogged down in it if it's just on the diploid side, there's no corresponding C code (and no plan to implement)

szhan:
I think the tests for the diploid case are failing because the fact that number of alleles can vary across sites is not being accounted for when computing the emission probability matrix. It should be fine if all the sites were biallelic, but some sites are probably invariant from the simulations.

jeromekelleher (approved):

PR #2975, opened 2024-08-06
Use reduction function in `pair_coalescence_stat`

nspope:
Followup to #2932

- Use a reduction function in `pair_coalescence_stat`, and change some naming conventions to be less confusing
- The reduction function is applied when each window is flushed, and is of the form `f(weights, values, params)`
- Because node weights may be aggregated into bins prior to reduction, `values` (which are node ages) also need to be aggregated. This is done by taking the weighted average of ages in each bin, which reduces to node ages if each node is in its own bin.
- There's some matrix transposition done to ensure that the inputs to the reduction function are contiguous in memory, which makes for nicer semantics

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -319,9 +319,9 @@ verify_pair_coalescence_counts(tsk_treeseq_t *ts, tsk_flags_t options)
     tsk_id_t sample_sets[n];
     tsk_size_t sample_set_sizes[P];
     tsk_id_t index_tuples[2 * I];
-    tsk_id_t node_output_map[N];
+    tsk_id_t node_bin_map[N];
     tsk_size_t dim = T * N * I;
-    double C1[dim];
+    double C[dim];
     tsk_size_t i, j, k;
 
     for (i = 0; i < n; i++) {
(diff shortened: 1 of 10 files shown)


nspope:
Because of the transposition, the output dimensions are now `(genomic windows, sample set indexes, time windows)`, which seems like a nicer layout to me because it's like adding an extra dimension onto e.g. divergence (which is the mean of the pair coalescence distribution). But, this doesn't agree with the time windowed stats draft #2948, so would be good to sort out now (@petrelharp).

nspope:
More numpy 2 issues in the CI:
```
tests/test_file_format.py:34: in <module>
    import h5py
../../.local/lib/python3.10/site-packages/h5py/__init__.py:25: in <module>
    from . import _errors
h5py/_errors.pyx:1: in init h5py._errors
    ???
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```
I'm guessing this is because the circleci workflow is using h5py 3.9.0, when we need 3.11.0 for numpy 2 compatibility. Why this didn't get hit yesterday on #2932 is mysterious ...

nspope:
Ready for a look whenever @jeromekelleher . Not sure where I'm going wrong with "codecov/project"?

jeromekelleher (approved):
This is very pretty, great stuff! I just have one comment about division by zero.

Regarding the order of dimensions, I don't really have an opinion here and happy to go with what you and @petrelharp think. I'm happy to merge this and keep plugging forward while we figure it out.

nspope:
Let's get this in then-- thanks!

PR #2979, opened 2024-08-29
Create C api for two-locus branch stats

lkirk:
Adds a C implementation of two-locus branch statistics. It mirrors the python implementation except where we iterate over edge differences and collect them for updating the stat. We use tree_seek_index to seek to arbitrary positions and tsk_tree_next to move from tree to tree. The tricky part was getting backwards iteration correct. All tests agree with the python prototype.

In addition, I had to fix a bug in the python implementation where node ids were being added to our TreeState object instead of sample ids (encoded in the sample index map). The python tests have also been updated to remove the slow naive version (after validating that it agrees with the python and c implementation -- on test cases where the runtime was reasonable). Python tests have been trimmed for runtime.

The CPython code has been updated to parse positions in addition to sites.

I also found the need to clean up some of the bounds checking code to return reasonable error messages to the user (also updated in the two-locus site stats).

Finally, I added a few unbiased statistics for use in validating this code.

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -808,7 +808,7 @@ out:
 
 /* Stats functions */
 
-#define GET_2D_ROW(array, row_len, row) (array + (((size_t)(row_len)) * (size_t) row))
+#define GET_2D_ROW(array, row_len, row) (array + (((size_t)(row_len)) * (size_t)(row)))
 
 static inline double *
 GET_3D_ROW(double *base, tsk_size_t num_nodes, tsk_size_t output_dim,
@@ -2621,9 +2621,12 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
             ret = TSK_ERR_SITE_OUT_OF_BOUNDS;
             goto out;
         }
-        if (sites[i] >= sites[i + 1]) {
-            // TODO: this checks no repeats, but error is ambiguous
-            ret = TSK_ERR_UNSORTED_SITES;
+        if (sites[i] > sites[i + 1]) {
+            ret = TSK_ERR_STAT_UNSORTED_SITES;
+            goto out;
+        }
+        if (sites[i] == sites[i + 1]) {
+            ret = TSK_ERR_STAT_DUPLICATE_SITES;
             goto out;
         }
     }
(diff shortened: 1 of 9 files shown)


lkirk:
Okay, this is ready for review now, had to clean up code coverage and a few lingering memory edge cases. I'm not sure how much further I can push the code coverage.

c/tskit/trees.c
```diff
@@ -808,7 +808,7 @@ tsk_treeseq_get_individuals_time(const tsk_treeseq_t *self, double *output)
 
 /* Stats functions */
 
-#define GET_2D_ROW(array, row_len, row) (array + (((size_t)(row_len)) * (size_t) row))
+#define GET_2D_ROW(array, row_len, row) (array + (((size_t)(row_len)) * (size_t)(row)))
```
lkirk, line 811: This macro has bitten me a few times now. I thought I'd make it a bit safer by adding parentheses so that expressions don't have strange behavior when passed in. For instance, when passing in an offset (ie n+4), the operator precedence of the multiplication makes the final 2d row offset incorrect. I don't foresee any issues with an extra set of parentheses, but macros can be tricky.

jeromekelleher: Ouch - good update and sorry for my macro parameter sloppiness! Feel free to change to an inline function if you like, I don't think it makes much difference in practise.

python/_tskitmodule.c
```diff
@@ -10135,24 +10135,71 @@ TreeSequence_pair_coalescence_quantiles(
...
+    if (positions == Py_None) {
+        array = (PyArrayObject *) TreeSequence_get_breakpoints(self);
+        if (array == NULL) {
+            goto out;
+        }
+        *out_dim = PyArray_DIM(array, 0) - 1; // NB the last element must be truncated
+    } else {
+        array = (PyArrayObject *) PyArray_FROMANY(
+            positions, NPY_FLOAT64, 1, 1, NPY_ARRAY_IN_ARRAY);
+        if (array == NULL) {
```
jeromekelleher, line 10177: Should cover this in test_lowlevel.py by passing in an bad array, say one with the wrong or the wrong type.

c/tskit/trees.c
```diff
@@ -2636,12 +2639,568 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
...
+    tsk_size_t state_dim, tsk_size_t result_dim, int sign, general_stat_func_t *f,
+    sample_count_stat_params_t *f_params, double *result)
+{
+    int ret = 0;
+    double a_len, b_len;
+    double *restrict B_branch_len = B_state->branch_len;
+    // TODO: is this early return okay?
+    b_len = B_branch_len[c] * sign;
+    if (b_len == 0) {
+        return ret;
```
jeromekelleher, line 2920: Should test this early return - but it's not clear what it's needed for?  How do you get a branch length of zero, and do you really need to defend against it here?

c/tskit/trees.c
```diff
@@ -2636,12 +2639,568 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
...
+    sample_count_stat_params_t *f_params, norm_func_t *TSK_UNUSED(norm_f),
+    tsk_size_t n_rows, const double *row_positions, tsk_size_t n_cols,
+    const double *col_positions, tsk_flags_t TSK_UNUSED(options), double *result)
+{
+    int ret = 0;
+    int r, c;
+    tsk_id_t *row_indexes, *col_indexes;
+    tsk_size_t i, j, k, row, col, *row_repeats, *col_repeats;
+    tsk_bit_array_t node_samples;
+    iter_state l_state, r_state;
```
jeromekelleher, line 3112: Whoops, need to initialise ``result_tmp=NULL`` or we'll have undefined behaviour on error conditions. Can you double check that all of the variables you are mallocing are initialised to NULL please?

c/tskit/trees.c
```diff
@@ -2636,12 +2639,568 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
...
+    double *result_tmp, *result_row;
+    const tsk_size_t num_nodes = self->tables->nodes.num_rows;
+
+    tsk_memset(&node_samples, 0, sizeof(node_samples));
+    tsk_memset(&l_state, 0, sizeof(l_state));
+    tsk_memset(&r_state, 0, sizeof(r_state));
+    if ((result_tmp = tsk_malloc(result_dim * sizeof(*result_tmp))) == NULL) {
```
jeromekelleher, line 3119: I think I'd rather have an extra line and slightly simpler code in these cases,even if it is a bit tedious, like
```c
result_tmp = tsk_malloc(result_dim * sizeof(*result_tmp));
if (result_tmp == NULL) {
    err...
}
```
It's easy enough the make mistakes without trying to do two things at once.

jeromekelleher: Ditto the next few blocks of code here

jeromekelleher (commented):
Generally LGTM. I spotted a few minor things which would be good to clear up.

Have you checked this for memory leaks with the run-in-a-loop trick? There's a few XDECREFs I'm not 100% about, would be good to check, please.

python/_tskitmodule.c
lkirk: I missed that case, I've added some tests that cover it

c/tskit/trees.c
lkirk: You're right, it's not actually something we would run into. I wasn't able to trigger this conditional. I've removed it.

c/tskit/trees.c
lkirk: Ah right, I've fixed all of these now I think.

c/tskit/trees.c
lkirk: Fixed

lkirk:
Hi @jeromekelleher thank you for taking a look.

> Have you checked this for memory leaks with the run-in-a-loop trick? There's a few XDECREFs I'm not 100% about, would be good to check, please.

I did the run-in-a-loop trick and could not detect any memory leaks. Which `XDECREF`s were you referring to?

c/tskit/trees.c
```diff
@@ -2636,12 +2639,573 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
...
+    tsk_size_t n_edges_in;
+} iter_state;
+
+static int
+iter_state_init(iter_state *self, const tsk_treeseq_t *ts, tsk_size_t state_dim)
+{
+    int ret = 0;
+    const tsk_size_t num_nodes = ts->tables->nodes.num_rows;
+
+    self->tree = tsk_malloc(sizeof(*self->tree));
```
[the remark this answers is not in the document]
jeromekelleher: Hmm, we also need to memset the struct itself to ensure all the pointers are zero. Here's how I would do it:
```
memset(self, 0, sizeof(*self));

ret = tsk_tree_init(&self->tree, ts, TSK_NO_SAMPLE_COUNTS);
```

By using an embedded tree struct (rather than a pointer like you currently have) we avoid having to do the malloc, and (more importantly) making sure that the memory is zero'd out before any possible error conditions can occur.

c/tskit/trees.c
```diff
@@ -2636,12 +2639,573 @@ check_sites(const tsk_id_t *sites, tsk_size_t num_sites, tsk_size_t num_site_row
...
+    int r, c;
+    tsk_id_t *row_indexes = NULL, *col_indexes = NULL;
+    tsk_size_t i, j, k, row, col, *row_repeats = NULL, *col_repeats = NULL;
+    tsk_bit_array_t node_samples;
+    iter_state l_state, r_state;
+    double *result_tmp = NULL, *result_row;
+    const tsk_size_t num_nodes = self->tables->nodes.num_rows;
+
+    tsk_memset(&node_samples, 0, sizeof(node_samples));
+    tsk_memset(&l_state, 0, sizeof(l_state));
```
jeromekelleher, line 3116: Ah ok, you're doing the memset here. No harm to redo in the iter_state class, just in case stuff gets refactored later and forgotten about. Up to you.

jeromekelleher (commented):
LGTM - spotted one definite safety problem, good to merge after resolving that.

c/tskit/trees.c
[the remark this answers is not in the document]
lkirk: I like the direction of using an embedded pointer. I've incorporated that change.

I think the one thing I'm on the fence about is the use of `memset` inside of `iter_state_init`. I'm not sure it provides much benefit (please correct me if I'm wrong) because we will still need to do the `memset` of `l_state` and `r_state` in the calling function (`tsk_treeseq_two_branch_count_stat`). This is because if we fail to initialize `l_state`, `r_state` must be zero'd at the time of freeing, which is done in the calling function. A refactor would still need to preserve this behavior unless the initialization of `l_state` and `r_state` were somehow decomposed (which I don't see a strong case for because they're both needed in the subsequent for loops).

I'd be happy to add it in for safety, I just don't see what it adds exactly.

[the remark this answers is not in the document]
jeromekelleher: Sure, it's not that important. I wrote the above before seeing the calling code

lkirk:
Everything is squashed and ready.

jeromekelleher (approved):

Issue #2986, opened 2024-09-12
Clarification on pairwise coalescent rates


Luker121:
Hi,

I have a .trees file with 37 individuals from 4 populations generated in SLiM, which I have recapitated, simplified, and added mutations to. Now, I want to calculate the pairwise coalescent rate both between and within populations, and obtain the mean and variance for each population pair.

Here is a simplified version of my current code:

```
for tree in tree_seq.trees():
        tree_start, tree_end = tree.interval
        overlap_span = min(tree_end, window_end) - max(tree_start, window_start)
        if overlap_span <= 0:
            continue

        for ind1 in pop1:
            for ind2 in pop2:
                if ind1 != ind2 or pop1 != pop2:
                    tmrca = tree.tmrca(ind1, ind2)
                    if tmrca > 0:
                        pairwise_tmrcas.append(tmrca)
                        weights.append(overlap_span)
                        total_weights += overlap_span

    if pairwise_tmrcas and total_weights > 0:
        weighted_mean_tmrca = np.sum(np.array(pairwise_tmrcas) * np.array(weights)) / np.sum(weights)
        pairwise_rate = 1 / weighted_mean_tmrca if weighted_mean_tmrca > 0 else np.nan

        pairwise_rates = [1 / tmrca for tmrca in pairwise_tmrcas if tmrca > 0]  # Avoid division by zero
        variance_rate = np.var(pairwise_rates)
        q5_rate = np.percentile(pairwise_rates, 5)
        q95_rate = np.percentile(pairwise_rates, 95)

        return {
            'mean': pairwise_rate,
            'variance': variance_rate,
            'q5': q5_rate,
            'q95': q95_rate
        }
```

In the loop, I iterate over each tree and individual in the respective population(s), compute the TMRCA using tree.tmrca(ind1, ind2), and weight the values by tree span. Then, I calculate the inverse of the weighted TMRCAs and scale by 2Ne (since the individuals are diploid).

My question is: is there a more efficient way to achieve this, or is there a built-in function to extract pairwise coalescent rate values directly from .trees files? Or is there a function to extract coalescent rates in general? Because I couldn't find any of this for more than one diploid individual.

petrelharp:
Hello! Good question; the short answer is: it's a work in (active) progress; see for instance https://github.com/tskit-dev/tskit/pull/2985 .

[...]

PR #2989, opened 2024-09-19
Catch case when windows do not span entire sequence for pair coalescence stats

nspope:
Fixes issue reported here: https://github.com/tskit-dev/tskit/issues/2986#issuecomment-2361134470

### DIFF
--- a/python/tests/test_lowlevel.py
+++ b/python/tests/test_lowlevel.py
@@ -4402,8 +4402,8 @@ class TestPairCoalescenceCountsErrors:
     def test_c_tsk_err_bad_windows(self):
         ts = self.example_ts()
         L = ts.get_sequence_length()
-        with pytest.raises(_tskit.LibraryError, match="BAD_WINDOWS"):
-            self.pair_coalescence_counts(ts, windows=[-1.0, L])
+        with pytest.raises(_tskit.LibraryError, match="TSK_ERR_BAD_WINDOWS"):
+            self.pair_coalescence_counts(ts, windows=[1.0, L])
 
     def test_c_tsk_err_bad_node_bin_map(self):
         ts = self.example_ts()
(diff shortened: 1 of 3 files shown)


nspope:
The issue was not passing `TSK_REQUIRE_FULL_SPAN` to `tsk_treeseq_check_windows` -- and I noticed that `tsk_treeseq_divergence_matrix` also doesn't do this here: https://github.com/tskit-dev/tskit/blob/dcee409933ce826dc1d672d4dc0b267e0bcc2d4b/c/tskit/trees.c#L8463

which may be a bug?

Also I wonder if it'd be better to move this option to the header rather than defining inline?

https://github.com/tskit-dev/tskit/blob/dcee409933ce826dc1d672d4dc0b267e0bcc2d4b/c/tskit/trees.c#L1194

nspope:
This is ready @jeromekelleher

python/tests/test_lowlevel.py
```diff
@@ -4402,8 +4402,8 @@ def test_c_tsk_err_node_out_of_bounds(self, bad_node):
     def test_c_tsk_err_bad_windows(self):
         ts = self.example_ts()
         L = ts.get_sequence_length()
-        with pytest.raises(_tskit.LibraryError, match="BAD_WINDOWS"):
-            self.pair_coalescence_counts(ts, windows=[-1.0, L])
+        with pytest.raises(tskit.LibraryError, match="TSK_ERR_BAD_WINDOWS"):
```
petrelharp, line 4405: should all these be `_tskit.LibraryError`?

nspope: what's the distinction?

jeromekelleher: I don't think it matters, it's probably the same exception. The _tskit module is the low-level C Python module, which defines LibraryError. Somewhere in Python we're importing this exception from _tskit, which then becomes tskit.LibraryError also.

So I guess Peter's right, it's better to directly catch the exception that we know the C code will throw.

jeromekelleher (commented):
Probably better to be consistent with other code on where the exception is coming from all right.

python/tests/test_lowlevel.py
nspope: shall I go ahead and make this change across the board in `tests/test_lowlevel.py`, then? there's a number of places (outside of the pair coalescence stuff) where `tskit.LibraryError` is checked for

nspope:
Done-- there's a number of other places in test_lowlevel.py where `tskit.LibraryError` is checked for rather than `_tskit.LibraryError`, but I left those alone here.

jeromekelleher (approved):

PR #2995, opened 2024-09-24
Clarify that path_length is a count, and that distance_between exists

hyanwong:
I find "path length" a bit confusingly named. Hopefully this addition/change in the docstring will clarify things.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -2976,7 +2976,7 @@ class Tree:
 
     def path_length(self, u, v):
         """
-        Returns a count of the number of edges between two nodes in this tree.
+        Returns the number of edges on the path in this tree between the two nodes.
         If the two nodes have a most recent common ancestor, then this is defined as
         ``tree.depth(u) + tree.depth(v) - 2 * tree.depth(tree.mrca(u, v))``. If the nodes
         do not have an MRCA (i.e., they are in disconnected subtrees) the path length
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -2970,13 +2970,16 @@ def rf_distance(self, other):
 
     def path_length(self, u, v):
         """
-        Returns the path length between two nodes
-        (i.e., the number of edges between two nodes in this tree).
+        Returns a count of the number of edges between two nodes in this tree.
```
jeromekelleher, line 2973: The "a count" is redundant here, right (saying it implies that there's more than one way to count the number of edges traversed). I think saying "path" somewhere in the description is important, as it's talking about the path traversed when walking the tree from one node to the other. Just saying "between" could be confused with counting all edges in the subtree rooted at the MRCA or something.

hyanwong: At the moment it says "Returns the path length between two nodes", which I immediately interpret as a branch length. I am trying hard to point out that this is an integer count, not (really) a length in the normal tree-parlance.

I guess you could say "Counts the number of edges when walking the path between two nodes in the tree"?

petrelharp:
```suggestion
        Returns the number of edges on the path in this tree between the two nodes.
```

hyanwong: LGTM

jeromekelleher (approved):

benjeffery:
@mergifyio rebase

PR #3002, opened 2024-09-25
Add ability to plot annotations (e.g. genes) along the x-axis

### LINKED ISSUE #3000: Ability to add tick positions and annotations to the X axis in draw_svg()
Quite often it's helpful to show e.g. genes along the genome on the X axis. Here's a hacked example:
Screenshot 2024-09-25 at 13 22 02

The code to hack this into an existing _tskit_ SVG plot is below, but it involves parsing the SVG string for the position of the X axis, and then doing some calibrations. This seems very messy. It might be nicer to pass in a list x_ticks (like we do for y_ticks), and also allow a dictionary of x_regions, of the form `{"ORF1a": (266, 13468), ...}`

```python
def new_draw_svg(*args, x_regions=None, **kwargs)
        if x_regions is None:
            xz_regions={}
        svg = ts.draw_svg(*args, **kwargs)
        if len(x_regions) > 0:
            assert svg.startswith("<svg")
            header = svg[:svg.find(">") + 1]
            footer = "</svg>"

            # Find SVG positions of the X axis
            m = re.search(r'class="x-axis".*?class="ax-line" x1="([\d\.]+)" x2="([\d\.]+)" y1="([\d\.]+)"', svg)
            assert m is not None
            x1, x2, y1 = float(m.group(1)), float(m.group(2)), float(m.group(3))
            xdiff = x2 - x1
            x_box_svg = '<rect fill="yellow" stroke="black" x="{x}" width="{w}" y="{y}" height="{h}" />'
            x_name_svg = '<text text-anchor="middle" alignment-baseline="hanging" x="{x}" y="{y}">{name}</text>'
            x_scale = xdiff / ts.sequence_length
            x_boxes = [
                x_box_svg.format(
                    x=x1 + p1 * x_scale,
                    w=(p2-p1) * x_scale,
                    y=y1,
                    h=20)  # height of the box: hardcoded for now to match font height
                for p1, p2 in x_regions.values()
            ]
            x_names = [
                x_name_svg.format(x=x1 + (p[0] + p[1])/2 * x_scale, y=y1+2, name=name)
                for name, p in x_regions.items()
            ]
            # add the new SVG to the old
            svg = (header + "".join(x_boxes) +  "".join(x_names) + footer) + svg
            # Now wrap both in another SVG
            svg = header + svg + footer
        return tskit.drawing.SVGString(svg)
```

hyanwong:
## Description

Fixes #3000 

# PR Checklist:


hyanwong:
I picked yellow region boxes for no particular reason, other than we are unlikely to use that colour for anything else (yellow symbols are generally impossible to see). I should probably add a tutorial example if this addition is thought to be useful.

python/tskit/drawing.py
```diff
@@ -791,82 +795,96 @@ def draw_x_axis(
...
-            if tick_length_upper is None:
-                tick_length_upper = tick_length_lower
-            y = rnd(self.plotbox.max_y - self.x_axis_offset)
-            dash_locs = np.concatenate(
-                (
-                    [self.plotbox.left],
-                    self.x_transform(alternate_dash_positions),
-                    [self.plotbox.right],
+        if not self.x_axis:
+            return
```
jeromekelleher, line 799: Needs a test

python/tskit/drawing.py
```diff
@@ -791,82 +795,96 @@ def draw_x_axis(
...
+        if len(x_regions) > 0:
+            regions_group = x_axis.add(dwg.g(class_="x-regions"))
+            for i, ((left, right), label) in enumerate(x_regions.items()):
```
jeromekelleher, line 802: I think you need some error checking on the region tuples here  - just 
```0 <= left < right <= ts.sequence_length`` is fine.

jeromekelleher (commented):
Nice, I like it.

What happens when you have overlapping regions (which will happen, I'd imagine)?

python/tskit/drawing.py
hyanwong: Yep, that seems sensible. I'll add

hyanwong:
> Nice, I like it.
> 
> What happens when you have overlapping regions (which will happen, I'd imagine)?

We just plot overlapping boxes. I guess that's fine? If people want to distinguish them they can colour them using transparency. Actually, I think it would be good to make the boxes semi-opaque, so people notice. 50% opacity example below: this LGTM, so I'll change the default style.

Screenshot 2024-09-26 at 11 34 34

jeromekelleher (approved):
LGTM!

petrelharp:
This is excellent!

PR #3003, opened 2024-09-26
allow `genetic_relatedness_vector` to do windows that don't span the whole genome

petrelharp:
Pretty straightforward; and this allows parallelization across windows.

*Edit:* I've added into this the ability to pass in a different set of nodes (added here since it's on top of these changes).

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -9908,9 +9908,10 @@ typedef struct {
     const double *weights;
     tsk_size_t num_windows;
     const double *windows;
+    tsk_size_t num_focal_nodes;
+    const tsk_id_t *focal_nodes;
     tsk_flags_t options;
     double *result;
-    /* tree */
     tsk_tree_position_t tree_pos;
     double position;
     tsk_size_t num_nodes;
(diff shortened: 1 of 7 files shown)


c/tskit/trees.c
```diff
@@ -9977,6 +9984,20 @@ tsk_matvec_calculator_init(tsk_matvec_calculator_t *self, const tsk_treeseq_t *t
     tsk_memset(result, 0, num_windows * num_samples * num_weights * sizeof(*result));
     tsk_memset(self->parent, TSK_NULL, num_nodes * sizeof(*self->parent));
 
+    ret = tsk_tree_position_init(&self->tree_pos, ts, 0);
+    if (ret != 0) {
+        goto out;
```
petrelharp, line 9989: this function always returns 0 currently

c/tskit/trees.c
```diff
@@ -9977,6 +9984,20 @@ tsk_matvec_calculator_init(tsk_matvec_calculator_t *self, const tsk_treeseq_t *t
...
+    }
+    /* seek to the first window */
+    index = (tsk_id_t) tsk_search_sorted(breakpoints, num_trees + 1, windows[0]);
+    if (breakpoints[index] > windows[0]) {
+        index--;
+    }
+    ret = tsk_tree_position_seek_forward(&self->tree_pos, index);
+    if (ret != 0) {
+        goto out;
```
petrelharp, line 9998: same here

petrelharp:
I think this is ready to go - the only uncovered lines are not cover-able.

petrelharp:
I've also put in the code here to let this work with arbitrary nodes. To see those changes separately look at the commits.

petrelharp:
This looks good to go! (One test will need to be re-run but it failed for other reasons.)

jeromekelleher (approved):
LGTM! I haven't gone through the fine details, but generally look good.

PR #3004, opened 2024-09-27
Add changelog for 2870

### LINKED ISSUE #2870: Warning for `draw_svg` plot uses `logging` module instead of `warnings`

benjeffery:
Fixes #2870

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -30,6 +30,9 @@
 - Limit output HTML when a tree sequence is displayed that has a large amount of metadata.
   (:user:`benjeffery`, :pr:`2999`)
 
+- Fix warning in `draw_svg` to use correct warnings module.
+  (:user:`duncanMR`, :issue:`2870`, :pr:`2871`)
+
 **Features**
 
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):

hyanwong:
Duncan says this is correct, so merge when ready.

PR #3005, opened 2024-09-28
Unifying stat method desc and linking to math summary

gregorgorjanc:
I was working through the stats tutorial at https://tskit.dev/tutorials/analysing_tree_sequences.html (incoming PR) and thought that linking stats/math summaries (https://tskit.dev/tskit/docs/stable/stats.html#summary-functions) to method docs would be very useful.

I also followed a style from the diversity method doc (what is the stat, then links to sample_sets, windows, ...).

@petrelharp can you have a look at this PR?

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -7515,7 +7515,6 @@ class TreeSequence:
         as sample sets will give ``f`` an argument of length two, giving the number
         of samples in ``A`` and ``B`` below the node in question. So, if we define
 
-
         .. code-block:: python
 
             def f(x):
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -8246,6 +8249,9 @@ def genetic_relatedness(
         """
         Computes genetic relatedness between (and within) pairs of
         sets of nodes from ``sample_sets``.
+        Please see the :ref:`summary functions <sec_stats_summary_functions>`
```
petrelharp, line 8252:
```suggestion
        See the :ref:`summary functions <sec_stats_summary_functions>`
```
For consistency.

python/tskit/trees.py
```diff
@@ -8478,8 +8484,12 @@ def genetic_relatedness_weighted(
         centre=True,
     ):
         r"""
-        Computes weighted genetic relatedness. If the :math:`k` th pair of indices
-        is (i, j) then the :math:`k` th column of output will be
+        Computes weighted genetic relatedness.
+        Please see the :ref:`summary functions <sec_stats_summary_functions>`
+        section on the exact definition of the calculated statistic.
```
petrelharp, line 8489: I don't actually think the "summary function" is a good place to refer people to here. That is a definition, but it only makes sense if someone understands How It All Works. The approach we've taken in other statistics is to provide a more plain (but equivalent) definition. I think here the thing to do is to just refer to the docs for `genetic_relatedness`, because this function's notion of genetic relatedness is the same as that one.

petrelharp:
In general I like the rearrangments, but I don't think we should add "please see summary functions for a precise definition". Instead, we should make what's said there in the documentation precise enough, if it isn't already? I am pretty sure that what's there for `diversity`, for instance, *is* precise - did you find it ambiguous? The 'summary function' definition is nice and concrete once you understand it, but most readers won't be on board with this. Some other way of referring to the summary functions could be helpful but more like "if you want the computational details, see XXX"?

gregorgorjanc:
@petrelharp I followed your suggestions in the last commit.

I thought adding the link to computational/mathematical definition would be useful to provide concrete/mathematical detail of what is calculated. This makes things concrete, but since it's only a link, it does not overwhelm a user (I agree that simple descriptive text should be preferred). Also, you have written these math details for some stats, so adding these links exposes/leverages that part of the doc (most users will only look at function docs and won't go over the whole stats doc to find math detail).

Happy to iterate.

PR #3007, opened 2024-10-03
Add title attributes to SVG symbols

hyanwong:
So we can reveal node and mutation details on SVG mouseover


jeromekelleher (approved):
LGTM

Any svg version weirdness to worry about?

hyanwong:
I don't think this should cause any problems. It's only an extra attribute. It works fine on the browsers I have used.

I forgot to document the parameters, however, so will add that now.

PR #3008, opened 2024-10-03
randomized svd draft

hanbin973:
## Description

A draft of randomized principal component analysis (PCA) using the `TreeSequence.genetic_relatedness_vector`. The implementation contains `spicy.sparse` which should eventually be removed. 
This part of the code is only used when collapsing a `#sample * #sample` GRM into a `#individual * #individual` matrix. 
Therefore, it will not be difficult to replace with pure numpy.

The API was partially taken from [scikit-learn](https://scikit-learn.org/dev/modules/generated/sklearn.decomposition.PCA.html).

To add some details, `iterated_power` is the number of power iterations in the range finder in the randomized algorithm. The error of SVD decreases exponentially as a function of this number.
The effect of power iteration is profound when the eigen spectrum of the matrix decays slowly, which seems to be the case of tree sequence GRMs in my experience.

`indices` specifies the individuals to be included in the PCA, although decreasing the number of individuals does not meaningfully reduce the amount of computation.

### DIFF
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -738,6 +738,7 @@ class TestRelatednessVector:
         )
         np.testing.assert_array_almost_equal(D1, D2)
 
+
 def pca(ts, windows, centre, samples=None, individuals=None, time_windows=None):
     assert samples is None or individuals is None
     if samples is None:
@@ -749,15 +750,21 @@ def pca(ts, windows, centre, samples=None, individuals=None, time_windows=None):
     if drop_dimension:
         windows = [0, ts.sequence_length]
     if time_windows is None:
-        Sigma = relatedness_matrix(ts=ts, windows=windows, centre=False)[:, ii, :][:, :, ii]
+        Sigma = relatedness_matrix(ts=ts, windows=windows, centre=False)[:, ii, :][
+            :, :, ii
+        ]
     else:
         assert time_windows[0] < time_windows[1]
         ts_low, ts_high = (
             ts.decapitate(time_windows[0]),
-            ts.decapitate(time_windows[1])
+            ts.decapitate(time_windows[1]),
         )
-        Sigma_low = relatedness_matrix(ts=ts_low, windows=windows, centre=False)[:, ii, :][:, :, ii]
-        Sigma_high = relatedness_matrix(ts=ts_high, windows=windows, centre=False)[:, ii, :][:, :, ii]
+        Sigma_low = relatedness_matrix(ts=ts_low, windows=windows, centre=False)[
+            :, ii, :
+        ][:, :, ii]
+        Sigma_high = relatedness_matrix(ts=ts_high, windows=windows, centre=False)[
+            :, ii, :
+        ][:, :, ii]
         Sigma = Sigma_high - Sigma_low
     if individuals is not None:
         ni = len(individuals)
@@ -775,7 +782,8 @@ def pca(ts, windows, centre, samples=None, individuals=None, time_windows=None):
     if drop_dimension:
         U = U[0]
         S = S[0]
-    return U, S   
+    return U, S
+
 
 def allclose_up_to_sign(x, y, **kwargs):
     # check if two vectors are the same up to sign
@@ -1105,17 +1113,14 @@ class TestPCA:
             random_seed=12345,
         )
         samples = [3, 0, 2, 5, 6, 15, 12, 17, 7, 9, 11]
-        time_low, time_high = (
-            ts.nodes_time.max() / 4, 
-            ts.nodes_time.max() / 2
-        )
+        time_low, time_high = (ts.nodes_time.max() / 4, ts.nodes_time.max() / 2)
         self.verify_pca(
             ts,
             num_windows=num_windows,
             num_components=5,
             centre=centre,
             samples=samples,
-            time_windows=[time_low,time_high],
+            time_windows=[time_low, time_high],
         )
 
     @pytest.mark.parametrize("centre", (True, False))
@@ -1181,15 +1186,12 @@ class TestPCA:
             random_seed=12345,
         )
         individuals = [3, 0, 2, 5, 6, 15, 12, 11, 7, 17]
-        time_low, time_high = (
-            ts.nodes_time.max() / 4, 
-            ts.nodes_time.max() / 2
-        )
+        time_low, time_high = (ts.nodes_time.max() / 4, ts.nodes_time.max() / 2)
         self.verify_pca(
             ts,
             num_windows=num_windows,
             num_components=5,
             centre=centre,
             individuals=individuals,
-            time_windows=[time_low,time_high],
+            time_windows=[time_low, time_high],
         )
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2024 Tskit Developers
+# Copyright (c) 2025 Tskit Developers
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -460,7 +460,7 @@ def check_relatedness_vector(
     return R
 
 
-class TestExamples:
+class TestRelatednessVector:
 
     def test_bad_weights(self):
         n = 5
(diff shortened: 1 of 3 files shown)


hanbin973:
@petrelharp Here's the code.

python/tskit/trees.py
```diff
@@ -8592,6 +8593,156 @@ def genetic_relatedness_vector(
         )
         return out
 
+    def pca(
+            self,
+            n_components: int = 10,
+            iterated_power: int = 3,
+            n_oversamples: int = 10,
+            indices: np.ndarray = None,
```
petrelharp, line 8601: Let's call this argument `individuals`?

python/tskit/trees.py
```diff
@@ -8592,6 +8593,156 @@ def genetic_relatedness_vector(
...
+                    )
+                                
+            assert cols.size == arr.shape[0], "Dimension mismatch"
+            # centering
+            x = arr - arr.mean(axis=0) if centre else arr # centering within index in rows
+            x = individual_idx_sparray(ts.num_individuals, cols).dot(x)
+            x = sample_individual_sparray(ts).dot(x)
+            x = ts.genetic_relatedness_vector(W=x, windows=windows, mode="branch", centre=False)
+            x = sample_individual_sparray(ts).T.dot(x)
+            x = individual_idx_sparray(ts.num_individuals, rows).T.dot(x)
```
petrelharp, line 8715: I think this assumes that all individuals' nodes are samples.  Note that we can use the `nodes` argument to `genetic_relatedness_vector` to get an arbitrary list of (possibly non-sample) nodes; why not just use that?

So, I think we can do something like this:
```
ij = np.vstack([[n, k] for k, i in enumerate(individuals) for n in self.individual(i).nodes])
sample_list = ij[:, 0]
indiv_index = ij[:, 1]
x = ts.genetic_relatedness_vector(W=x, ..., nodes=sample_list)
x = np.bincount(indiv_index, x)
```

python/tskit/trees.py
```diff
@@ -8592,6 +8593,156 @@ def genetic_relatedness_vector(
...
+            x = x - x.mean(axis=0) if centre else x # centering within index in cols
+
+            return x
+
+
+        if indices is None: indices = np.array([i.id for i in self.individuals()])
+        if random_state is None: random_state = np.random.default_rng()
+        
+        def _G(x):
```
petrelharp, line 8724: perhaps we can just roll `_genetic_relatedness_vector` into here?

python/tskit/trees.py
```diff
@@ -8592,6 +8593,156 @@ def genetic_relatedness_vector(
...
+            return _genetic_relatedness_vector(
+                    self.ts,
+                    x,
+                    indices,
+                    indices,
+                    centre,
+                    windows
+                    )
+
+        U, D, _ = _rand_svd(
```
petrelharp, line 8734: Here `_` should be equal to `U`, I'm guessing? Since the matrix is symmetric?

petrelharp:
This looks great! Very elegant. I think probably we ought to include a `samples` argument, though? For consistency, but also since the tree sequence represents phased data, and so it's actually informative to look at the PCs of maternally- and paternally-inherited chromosomes separately.

So, how about the signature is like
```
def pca(samples=None, individuals=None, ...)
```
and:
- the default is equivalent to `samples=ts.samples(), individuals=None`
- you can't have both `samples` and `individuals` specified
- if `individuals` is a list of individual IDs then it does as in the code currently
- otherwise, is just skips the "sum over individuals" step

Note that we could be getting PCs for non-sample nodes (since individual's nodes need not be samples); I haven't thought through whether the values you get are correct or informative. My guess is that maybe they are? But we need a "user beware" note for this?

python/tskit/trees.py
hanbin973:
```
x = ts.genetic_relatedness_vector(W=x, ..., nodes=sample_list)
```
should slightly be
```
x = ts.genetic_relatedness_vector(W=x[indiv_index], ..., nodes=sample_list)
```
to expand the array of individuals to array of nodes, I think?

python/tskit/trees.py
hanbin973: Yes. U=V for SVD on a symmetric matrix.

python/tskit/trees.py
hanbin973: I left `_G` to handle the new `samples` option.

python/tskit/trees.py
petrelharp: hah, yes - good catch!

python/tskit/trees.py
petrelharp: just making sure.

python/tskit/trees.py
```diff
@@ -8684,56 +8684,33 @@ def _genetic_relatedness_vector(
...
+            samples, sample_individuals = ij[:,0], ij[:,1] # sample node index, individual of those nodes
             x = arr - arr.mean(axis=0) if centre else arr # centering within index in rows
-            x = individual_idx_sparray(ts.num_individuals, cols).dot(x)
-            x = sample_individual_sparray(ts).dot(x)
-            x = ts.genetic_relatedness_vector(W=x, windows=windows, mode="branch", centre=False)
-            x = sample_individual_sparray(ts).T.dot(x)
-            x = individual_idx_sparray(ts.num_individuals, rows).T.dot(x)
+            x = self.genetic_relatedness_vector(W=x[sample_individuals], windows=windows, mode="branch", centre=False, nodes=samples)
+            bincount_fn = lambda w: np.bincount(sample_individuals, w)
+            x = np.apply_along_axis(bincount_fn, axis=0, arr=x) # I think it should be axis=1, but axis=0 gives the correct values why?
```
hanbin973, line 8693: The matvec is sometimes GRM * matrix, so `x` is often a matrix than a vector. `np.bincount` only works for 1-dimensional weights, so I used `np.apply_along_axis` and `lambda` to vectorize `np.bincount`.

hanbin973: The comment I left after # looks like mostly a convention issue in that function. When `axis=0`, the columns are separately retrieved from the array. When `axis=1`, the rows are retrieved.

petrelharp: Agree, that seems confusing but maybe makes sense after all?

python/tskit/trees.py
```diff
@@ -8592,6 +8592,132 @@ def genetic_relatedness_vector(
...
+    def pca(
+            self,
+            n_components: int = 10,
+            iterated_power: int = 3,
+            n_oversamples: int = 10,
+            samples: np.ndarray = None,
+            individuals: np.ndarray = None,
+            centre: bool = True,
+            windows: list = None,
+            random_state: np.random.Generator = None,
```
petrelharp, line 8604: usually we just pass in a `seed`, any objections to doing that, instead?

petrelharp:
Ah, sorry - one more thing - does this work with `windows`? (It looks like not?)

I think the way to do the windows would be something like
```
drop_windows = windows is None
if drop_windows:
    windows = [0, self.sequence_length]

# then do stuff; with these windows genetic_relatedness will always return an array where the first dimension is "window";
# so you can operate on each slice separately

if drop_windows:
    # get rid of the first dimension in the output
```
Basically - get it to work in the case where `windows` are specified (ie not None) and then we can get it to have the right behavior.

hanbin973:
A simple test case for the `windows` feature.
```
demography = msprime.Demography()
demography.add_population(name="A", initial_size=5_000)
demography.add_population(name="B", initial_size=5_000)
demography.add_population(name="C", initial_size=1_000)
demography.add_population_split(time=1000, derived=["A", "B"], ancestral="C")
ts = msprime.sim_ancestry(
    samples={"A": 500, "B": 500},
    sequence_length=1e6,
    recombination_rate=3e-8,
    demography=demography, 
    random_seed=12)
seq_length = ts.sequence_length

U, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[0, seq_length/2, seq_length])
U0, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[0, seq_length/2])
U1, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[seq_length/2, seq_length])

idx = 0 # idx is the idx-th principal component
# correlation instead of allclose because PCA is rotation symmetric
np.corrcoef(U[0][:,idx], U0[:,idx]), np.corrcoef(U[1][:,idx], U1[:,idx])
```

Because of the randomness of the algo, the correlation is not exactly 1, although it's nearly 1 like 0.99995623-ish.

python/tskit/trees.py
hanbin973: I changed the option from `random_state` to `random_seed` following msprime.

hanbin973:
I just noticed that `centre` doesn't work with `nodes` option. The new commit fixed this problem.

hanbin973:
Check results for two windows.
```
demography = msprime.Demography()
demography.add_population(name="A", initial_size=5_000)
demography.add_population(name="B", initial_size=5_000)
demography.add_population(name="C", initial_size=1_000)
demography.add_population_split(time=1000, derived=["A", "B"], ancestral="C")
seq_length =1e6
ts = msprime.sim_ancestry(
    samples={"A": 500, "B": 500},
    sequence_length=seq_length,
    recombination_rate=3e-8,
    demography=demography, 
    random_seed=12)

# for individuals
U, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[0, seq_length/2, seq_length])
U0, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[0, seq_length/2])
U1, _ = ts.pca(individuals=np.asarray([i.id for i in ts.individuals()]), iterated_power=5, random_seed=1, windows=[seq_length/2, seq_length])

idx = 0 # idx is the idx-th principal component
# correlation instead of allclose because PCA is rotation symmetric
np.corrcoef(U[0][:,idx], U0[0][:,idx]), np.corrcoef(U[1][:,idx], U1[0][:,idx])

# for nodes
U, _ = ts.pca(iterated_power=5, random_seed=1, windows=[0, seq_length/2, seq_length])
U0, _ = ts.pca(iterated_power=5, random_seed=1, windows=[0, seq_length/2])
U1, _ = ts.pca(iterated_power=5, random_seed=1, windows=[seq_length/2, seq_length])

idx = 0 # idx is the idx-th principal component
# correlation instead of allclose because PCA is rotation symmetric
np.corrcoef(U[0][:,idx], U0[0][:,idx]), np.corrcoef(U[1][:,idx], U1[0][:,idx])
```

python/tskit/trees.py
```diff
@@ -8593,138 +8593,188 @@ def genetic_relatedness_vector(
         return out
 
     def pca(
-            self,
```
petrelharp: I rearranged these to better match other methods (e.g., `windows` always come first, so I had it first after `n_components`)

python/tskit/trees.py
```diff
@@ -8593,138 +8593,188 @@ def genetic_relatedness_vector(
...
-            centre: bool = True,
-            windows: list = None,
-            random_seed: int = None,
-            ) -> (np.ndarray, np.ndarray):
-        """
-        Run randomized singular value decomposition (rSVD) to obtain principal components.
-        API partially adopted from `scikit-learn`: 
-        https://scikit-learn.org/dev/modules/generated/sklearn.decomposition.PCA.html
+        self,
+        n_components: int = 10,
```
petrelharp, line 8597: perhaps we should not have a default?

python/tskit/trees.py
```diff
@@ -8593,138 +8593,188 @@ def genetic_relatedness_vector(
...
+        if n_components > dim:
+            raise ValueError(
+                "Number of components must be less than or equal to "
+                "the number of samples (or individuals, if specified)."
+            )
+
+        random_state = np.random.default_rng(random_seed)
+
         def _rand_pow_range_finder(
-                operator: Callable,
```
petrelharp: linting complains about `Callable` for some reason

python/tskit/trees.py
```diff
@@ -8593,138 +8593,188 @@ def genetic_relatedness_vector(
...
-                    )
+                operator, operator_dim, num_vectors, depth, num_vectors, rng
+            )
             C = operator(Q).T
             U_hat, D, V = np.linalg.svd(C, full_matrices=False)
             U = Q @ U_hat
-            return U[:,:rank], D[:rank], V[:rank]
+            return U[:, :rank], D[:rank], V[:rank]
 
         def _genetic_relatedness_vector_individual(
```
petrelharp, line 8703: these changes are all automatic linting

python/tskit/trees.py
```diff
@@ -8593,138 +8593,188 @@ def genetic_relatedness_vector(
...
+            def bincount_fn(w):
+                np.bincount(sample_individuals, w)
+
             x = np.apply_along_axis(bincount_fn, axis=0, arr=x)
-            x = x - x.mean(axis=0) if centre else x # centering within index in cols
+            x = x - x.mean(axis=0) if centre else x  # centering within index in cols
 
             return x
 
         def _genetic_relatedness_vector_node(
```
petrelharp, line 8738: same: automatic linting

petrelharp:
Okay; here I've made a good start at the tests.  I think everything is working fine; the tests are not passing because (I think) of numerical tolerance. I could just set the numerical tolerance to like 1e-4 and they'd pass, but I think this is flagging a bigger issue - how do we tell we're getting good answers? I ask because currently the tests pass (at default tolerances) for small numbers of samples but not for 30 samples; if I increase `iterated_power` then they do pass; so the current defaults seem okay for 30 samples, but I worry that at 10,000 samples they might be very bad. If `u, s` is one of the output vector, value pairs, can we compare `s * u` to `W @ u` and translate the result to some measure of tolerance? Not very good options here that I can think of are
1. add a `tolerance` argument and we iterate until this tolerance is reached
2. return an estimate of error; perhaps we should return some kind of structured output that contains information about convergence as well?

We'd like this to be not a can of worms; I think our goal is to have something that is good enough, and fowards-compatible for an improved method in the future.

Notes:
- For testing, I use `np.linalg.svd`, and (of course) had to mess around so that we'd correctly deal with indeterminancy of  sign and also indeterminancy of ordering when singular values are the same.
- Perhaps we should call it `num_components` rather than `n_components` to match elsewhere?
- I did some rearranging of how the `operator` in `_rand_svd` was defined to avoid lint complaining (but was unsuccessful; so inserted the `NOQA`)
- Some of the changes were due to automatic linting; apologies for the noise there.

TODO: 
- add tests that include `samples` and `individuals` arguments
- test on some weirder tree sequences (that's easy)

jeromekelleher:
Just a quick note that I'd be very much in favour of returning a dataclass here rather than a tuple so that the option of returning more information about convergence etc is open.

hanbin973:
There's an adaptive rangefinder algorithm described in Halko et al. (https://arxiv.org/pdf/0909.4061, Algo 4.2). I don't see it implemented in scikit-learn (https://scikit-learn.org/dev/modules/generated/sklearn.decomposition.TruncatedSVD.html#sklearn.decomposition.TruncatedSVD). 
Halko et al. also has an error estimation formula with an overhead similar to the main routine.

I like Jerome's idea to return a class instead of the result. There's an intermediate matrix `Q` (approximate range of the original matrix) that takes up most of the computation, and storing it is useful in many cases. For example, one can add additional power iterations to `Q` without doing it from scratch to improve precision.

hanbin973:
Not a classobject yet but added random_sketch to the input/output.

jeromekelleher:
Re the result object, I'd imagined something like
```python
@dataclasses.dataclass
class PcaResult:
    descriptive_name1: np.ndarray # Or whatever type hints we can get to work
    descriptive_name2...
```

hanbin973:
Now, `pca()` returns a dataclass of the following
```
@dataclass
class PCAResult:
    U: np.ndarray
    D: np.ndarray
    Q: np.ndarray
    E: np.ndarray
```

`U` and `D` are as before. Q is the range sketch matrix that is used as the approximate orthonormal basis of the GRM. It is also the most and the only expensive part of the algorithm that involves GRM*matrix operations. E is the error bounds for the singular values. Both Q and E will have different values for each windows if present. 

A user can continuously improve their estimate through Q. `pca` now has a `range_sketch: np.ndarray = None` option that accepts Q from the previous found of the pca. This can be done like
```
pca_result = ts.pca( ... )
pca_result_round_2 = ts.pca( ..., range_sketch = pca_result.Q, ...)
```

If the first round did `q` power iterations and the second round did `p` additional power iterations, the result of the second round has total `q+p` iterations. By adding additional power iterations in successive rounds, one can improve the accuracy without running the whole process from scratch.

python/tskit/trees.py
```diff
@@ -8637,8 +8637,11 @@ def pca(
             be automatically generated. Valid random seeds must be between 1 and
             :math:`2^32 − 1`.
         :param np.ndarray range_sketch: Sketch matrix for each window. Default is None.
-        :return: A tuple (U, D, Q) of ndarrays, with the principal component loadings in U
-            and the principal values in D. Q is the range sketch array for each window.
+        :return: A class object with attributes U, D, Q and E.
```
jeromekelleher, line 8640: I know this is probably heresy, but could we put names on these variables? Like:
```
@dataclasses.dataclass
class PcaResult:
     """
     The result of a call to TreeSequence.pca() capturing the output values and algorithm convergence details.
     """
     loadings: np.ndarray
     """
     The principle component loadings.
     """
     values: np.ndarray
     """
     The principle component values.
     """
     range_sketch: np.ndarray
     """
     The range sketch. See XXX for details?
     """
     error_bound: np.ndarray
     """
     ...
     """
```
Then, the return would be 
```
:return: An instance of :class:`PcaResult` encapsulating the principle components, loadings and algorithm
    converence details.
```

(A "class object" is something specific, which this isn't)

jeromekelleher (commented):
This looks great, but I would suggest we break the nested functions out to the module level rather than embedding them in the TreeSequence class. The function is currently too long, and it's not clear what needs to be embedded within the function because it's using the namespace, vs what's in there just because. It would be nice to be able to  test the bits of this individually, and putting them at the module level will make that possible.

Certainly the return class should be defined at the module level and added to the Sphinx documentation so that it can be linked to.

python/tskit/trees.py
hanbin973: Sounds good. Where should I place the functions and the new class?

jeromekelleher: Put them at the end of trees.py for now. We should probably make a new module for some of this statsy stuff, but let's not bother for now.

petrelharp: Note: the functions should not be public, I would think. (the class, yes)

python/tskit/trees.py
```diff
@@ -10387,3 +10291,134 @@ def write_ms(
...
+    U = Q @ U_hat
+
+    error_factor = np.power(
+            1 + 4 * np.sqrt(2 * operator_dim / (rank - 1)), 
+            1 / (2 * depth + 1)
+            )
+    error_bound = D[-1] * (1 + error_factor)
+    return U[:, :rank], D[:rank], V[:rank], Q, error_bound
+
+def _genetic_relatedness_vector_individual(
```
jeromekelleher, line 10346: These are best seen as methods of the TreeSequence class because they have the first argument as a tree sequence. I'd refactor as 
```
    def _genetic_relatedness_vector_individual(self, arr, indices, mode...):
          ...
    def _genetic_relatedness_vector_node(self, ...)
         ...
```

jeromekelleher (commented):
Minor nitpick about code organisation!

python/tskit/trees.py
jeromekelleher: Essentially maintenance, yes, but also freedom to change things in the future

python/tskit/trees.py
hanbin973: They are now tree sequence methods. SVD functions were moved into the `pca()` function. Only the class definition is left outside.

hanbin973:
It now has a time-resolved feature. You can select branches within the lower and the upper time limits. It is based on decapitate.

petrelharp:
NICE!

python/tskit/trees.py
```diff
@@ -8592,6 +8592,252 @@ def genetic_relatedness_vector(
...
+            :meth:`genetic_relatedness_vector
+            <.TreeSequence.genetic_relatedness_vector>`)
+        :param bool centre: Centre the genetic relatedness matrix.
+        :param int iterated_power: Number of power iteration of range finder.
+        :param int num_oversamples: Number of additional test vectors.
+        :param int random_seed: The random seed. If this is None, a random seed will
+            be automatically generated. Valid random seeds must be between 1 and
+            :math:`2^32 − 1`.
+        :param np.ndarray range_sketch: Sketch matrix for each window. Default is None.
+        :return: A class object with attributes U, D, Q and E.
```
petrelharp, line 8696:
```suggestion
        Performs principal component analysis (PCA) for a given set of samples or
        individuals. The principal components are the eigenvectors of the genetic
        relatedness matrix, which are obtained by a randomized singular value
        decomposition (rSVD) algorithm.
        
        Concretely, if :math:`M` is the matrix of genetic relatedness values, with
        :math:`M_{ij}` the output of
        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`
        between sample :math:`i` and sample :math:`j`, then by default this returns
        the top `num_components` eigenvectors of :math:`M`, so that
        `output.factor[i,k]` is the position of sample `i` on the `k`th PC.
        If `samples` or `individuals` are provided, then this does the same thing,
        except with :math:`M_{ij}` either the relatedness between `samples[i]`
        and `samples[j]` or the nodes of `individuals[i]` and `individuals[j]`,
        respectively.
        
        The parameters `centre` and `mode` are passed to
        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`;
        if `windows` are provided then PCA is carried out separately in each window.
        If `time_windows` are provided, then genetic relatedness is measured using only
        ancestral material within the given time window (see
        :meth:`decapitate <.TreeSequence.decapitate>` for how this is defined).
                
        So that the method scales to large tree sequences, the underlying method
        relies on a randomized SVD algorithm. Larger values of `iterated_power` and
        `num_oversamples` should produce better approximations to the true eigenvalues,
        at the cost of greater compute times and/or memory usage. The method relies on
        constructing `range_sketch`, a low-dimensional approximation to the range
        of :math:`M`, so that the result of a previous call to `pca()` may be passed in.
        To check for convergence, compare
        `pc1 = ts.pca()` and `pc2 = ts.pca(range_sketch=pc1.range_sketch)`; the difference
        between `pc1.factors` and `pc2.factors` provides a diagnostic of the convergence
        of the algorithm (i.e., if they are close then it has likely converged).
        Algorithms are based on Algorithms 8 and 9 in Martinsson and Tropp,
        https://arxiv.org/pdf/2002.01387 .
        
        :param int num_components: Number of principal components to return.
        :param list windows: An increasing list of breakpoints between the windows
            to compute the statistic in (default: the entire genome).
        :param np.ndarray samples: Samples to perform PCA with (default: all samples).
        :param np.ndarray individuals: Individuals to perform PCA with. Cannot specify
            both `samples` and `individuals`.
        :param np.ndarray time_windows: The time interval on which to apply PCA:
            currently, this must be either None (default, covers all time) or a single interval.
        :param str mode: A string giving the "type" of relatedness to be computed
            (defaults to "branch"; see
            :meth:`genetic_relatedness_vector
            <.TreeSequence.genetic_relatedness_vector>`).
        :param bool centre: Centre the genetic relatedness matrix.
        :param int iterated_power: Number of power iteration of range finder.
        :param int num_oversamples: Number of additional test vectors.
        :param int random_seed: The random seed. If this is None, a random seed will
            be automatically generated. Valid random seeds are between 1 and
            :math:`2^32 − 1`.
        :param np.ndarray range_sketch: Sketch matrix for each window. Default is None.
        :return: A :class:`.PCAResult` object, containing estimated principal components,
            eigenvalues, and other information.
```
To be consistent with the API developed by @tforest I think we should call it `time_windows` not `time_window`; we can just require that there's only one window for the moment. (This way we won't have to change the API in the future.)

python/tskit/trees.py
```diff
@@ -10171,3 +10417,29 @@ def write_ms(
...
+class PCAResult:
+    """
+    The result of a call to TreeSequence.pca() capturing the output values
+    and algorithm convergence details.
+
+
+    """
+    factors: np.ndarray
+    """
+    The principal component factors. It is an orthogonal matrix.
```
petrelharp, line 10431:
```suggestion
    The principal component factors. Columns are orthogonal, with one entry per sample
    or individual (see :meth:`pca <.TreeSequence.pca>`).
```

python/tskit/trees.py
```diff
@@ -10171,3 +10417,29 @@ def write_ms(
...
+    """
+    eigen_values: np.ndarray
+    """
+    Eigenvalues of the genetic relatedness matrix.
+    """
+    range_sketch: np.ndarray
+    """
+    Range sketch matrix. Can be used as an input for .pca() call with range_sketch option
+    to further improve precision..
```
petrelharp, line 10440:
```suggestion
    Range sketch matrix. Can be used as an input for :meth:`pca <.TreeSequence.pca>` option
    to further improve precision.
```

python/tskit/trees.py
```diff
@@ -10171,3 +10417,29 @@ def write_ms(
...
+    """
+    error_bound: np.ndarray
+    """
+    Error bounds for the eigenvalues.
```
petrelharp, line 10444:
```suggestion
    An estimate of the error bounds for the eigenvalues (experimental).
```

petrelharp:
I made a pass through the docs. We need to add `time_windows` to the tests still, and see what's going on with the CI.

python/tskit/trees.py
```diff
@@ -8592,6 +8592,276 @@ def genetic_relatedness_vector(
...
+            [
+                [n, k]
+                for k, i in enumerate(indices)
+                for n in self.individual(i).nodes
+            ]
+        )
+        samples, sample_individuals = (
+            ij[:, 0],
+            ij[:, 1],
+        )  # sample node index, individual of those nodes
```
jeromekelleher, line 8629: Putting comments at the end of lines is causing them to get broken by Black. Better to put the comments on the line immediately above.

python/tskit/trees.py
```diff
@@ -10171,3 +10441,30 @@ def write_ms(
...
+    and algorithm convergence details.
+
+
+    """
+    factors: np.ndarray
+    """
+    The principal component factors. Columns are orthogonal, with one entry per sample
+    or individual (see :meth:`pca <.TreeSequence.pca>`).
+    """
+    eigen_values: np.ndarray
```
jeromekelleher, line 10458: eigenvalues is one word, isn't it?

jeromekelleher (commented):
Looks good to me. I think we need to tidy up the lint and get tests passing next so we can see how coverage is doing?

python/tests/test_relatedness_vector.py
```diff
@@ -737,3 +737,147 @@ def test_disconnected_non_sample_topology(self, centre):
...
+            with pytest.raises(tskit.LibraryError, match="TSK_ERR_BAD_WINDOWS"):
+                ts.pca(n_components=2, windows=bad_w)
+
+    def test_bad_num_components(self):
+        ts = msprime.sim_ancestry(
+            3,
+            ploidy=2,
+            sequence_length=10,
+            random_seed=123,
+        )
```
petrelharp, line 837: maybe a test for `n_components=0` and `-1` also?

python/tests/test_relatedness_vector.py
```diff
@@ -737,3 +737,147 @@ def test_disconnected_non_sample_topology(self, centre):
...
+
+def allclose_up_to_sign(x, y, **kwargs):
+    # check if two vectors are the same up to sign
+    x_const = np.isclose(np.std(x), 0)
+    y_const = np.isclose(np.std(y), 0)
+    if x_const or y_const:
+        if np.allclose(x, 0):
+            r = 1.0
+        else:
+            r = np.mean(x / y)
```
petrelharp, line 762: This is not right, as here we want `r` to be +/-1, I think?

petrelharp:
It looks like the things to do here are:
- get the tests working (right now they fail with `FAILED tests/test_relatedness_vector.py::TestPCA::test_bad_windows - TypeError: pca() got an unexpected keyword argument 'n_components'`
- either remove for now the `time_windows` argument or write tests for it
- write tests that exercise the `individuals` argument (or remove it)
- write a test that uses `range_sketch`
- write tests that exercise `iterated_power` and `num_oversamples`: probably, just something that checks whether setting these to bigger numbers still gets us (nearly) the same answer

jeromekelleher:
Bumping this one - we want to get ts.pca implemented and released as soon as we can. What's left to do here @hanbin973? Can we help with anything to get it over the line?

hanbin973:
> Bumping this one - we want to get ts.pca implemented and released as soon as we can. What's left to do here @hanbin973? Can we help with anything to get it over the line?

It's just the test codes that are missing. We have to make the tests pass. I told @petrelharp that I will work on it, but well it didn't go as planned :(

jeromekelleher:
I don't think this works if we use it on a tree sequence where the samples aren't 0,...,n. Trying this out on the SARS-CoV-2 data, I got:
```
/tmp/ipykernel_620501/3297030094.py in ?()
----> 1 pca = ts.pca(4, samples=ts.samples())

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(self, num_components, windows, samples, individuals, time_window, mode, centre, iterated_power, num_oversamples, random_seed, range_sketch)
   8856                 else:
   8857                     low = _f_low(arr=x, indices=indices, mode=mode, centre=centre, windows=this_window)
   8858                     return high - low
   8859 
-> 8860             U[i], D[i], _, Q[i], E[i] = _rand_svd(
   8861                 operator=_G,
   8862                 operator_dim=dim,
   8863                 rank=num_components,

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(operator, operator_dim, rank, depth, num_vectors, rng, range_sketch)
   8803             """
   8804             Algorithm 8 in https://arxiv.org/pdf/2002.01387
   8805             """
   8806             assert num_vectors >= rank > 0
-> 8807             Q = _rand_pow_range_finder(
   8808                 operator, operator_dim, num_vectors, depth, num_vectors, rng, range_sketch
   8809             )
   8810             C = operator(Q).T

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(operator, operator_dim, rank, depth, num_vectors, rng, range_sketch)
   8786             else:
   8787                 Q = range_sketch
   8788             for _ in range(depth):
   8789                 Q = np.linalg.qr(Q).Q
-> 8790                 Q = operator(Q)
   8791             Q = np.linalg.qr(Q).Q
   8792             return Q[:, :rank]

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(x)
   8852             def _G(x):
-> 8853                 high = _f_high(arr=x, indices=indices, mode=mode, centre=centre, windows=this_window)
   8854                 if time_window is None:
   8855                     return high
   8856                 else:

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(self, arr, indices, mode, centre, windows)
   8610         centre: bool = True,
   8611         windows = None,
   8612         ) -> np.ndarray:
   8613         x = arr - arr.mean(axis=0) if centre else arr
-> 8614         x = self._expand_indices(x, indices)
   8615         x = self.genetic_relatedness_vector(
   8616             W=x, windows=windows, mode=mode, centre=False, nodes=indices,
   8617         )[0]

~/.local/lib/python3.10/site-packages/tskit/trees.py in ?(self, x, indices)
   8596         x: np.ndarray,
   8597         indices: np.ndarray
   8598         ) -> np.ndarray:
   8599         y = np.zeros((self.num_samples, x.shape[1]))
-> 8600         y[indices] = x
   8601 
   8602         return y

IndexError: index 2482157 is out of bounds for axis 0 with size 2482157
[6]:
```
but if I first simplify (ts = ts.simplify()) so that the samples are 0 to n it runs fine. Just needs a simple test for this case I'd imageine.

python/tskit/trees.py
```diff
@@ -8591,6 +8591,288 @@ def genetic_relatedness_vector(
...
+        :param bool centre: Centre the genetic relatedness matrix.
+        :param int iterated_power: Number of power iteration of range finder.
+        :param int num_oversamples: Number of additional test vectors.
+        :param int random_seed: The random seed. If this is None, a random seed will
+            be automatically generated. Valid random seeds are between 1 and
+            :math:`2^32 − 1`.
+        :param np.ndarray range_sketch: Sketch matrix for each window. Default is None.
+        :return: A :class:`.PCAResult` object, containing estimated principal components,
+            eigenvalues, and other information.
+            The principal component loadings are in U
```
jeromekelleher, line 8732: This last bit is out of date now.

python/tskit/trees.py
```diff
@@ -10171,3 +10453,30 @@ def write_ms(
...
+
+@dataclass
+class PCAResult:
+    """
+    The result of a call to TreeSequence.pca() capturing the output values
+    and algorithm convergence details.
+
+
+    """
+    factors: np.ndarray
```
jeromekelleher, line 10465: We need to define the dimensions here. Currently the first dimension is samples/individuals and the second is the num_componnents. Is there a strong reason for doing it this way? I expected it to be ``(num_components, num_samples)`` as you usually want to access all the values for a given component together?

jeromekelleher: Aha, it's following scikit learn. That's an excellent guide to follow - let's just do what scikit learn does and make our API as compatible with them as possible? We should document this as a stated goal.

jeromekelleher (commented):
Few minor comments on return value interface

jeromekelleher:
I successfully ran this on a 2.5M sample SARS-CoV-2 ARG. Took about 30 seconds, and seemed to converge OK (at least from my cursory check using the recommended approach). A summary of the results is here: https://github.com/jeromekelleher/sc2ts-paper/issues/372

petrelharp:
Okay - I've put in a bunch of the testing code for correct arguments, etcetera. Still TODO:

- [ ] testing for the error bounds
- [ ] figure out why `individuals` is not passing tests (see below)

One issue I've turned up along the way is that the `individuals` mode I think was not agreeing with our definition of `genetic_relatedness`: as described in the relatedness paper, relatedness between two sample sets is the *average* of the relatedness between pairwise combinations of samples. This is just a factor of 4 for diploids - but that's a discrepancy that will confuse people - and actually makes a difference for mixed ploidy (eg sex chromosomes). I think I've sorted this out correctly.

Have we convinced ourselves that the default `iterated_power` is a good value? I have a bit of a hard time figuring out what "good" means - like, what tolerance do we need these eigenvectors to?

I also refactored the code to randomly generate `range_sketch` outside of the loop over windows, because this makes (a) the testing-for-correct-inputs code up top cleaner, and (b) the `_rand_pow_range_finder` and `_rand_svd` functions simpler. This has a memory downside of having to have `range_sketch` for all windows, not just one -- but, this is something we're returning anyhow, so is not a big deal. 

I added an `individuals` argument to the testing implementation, but this isn't matching; I've stared at it a bunch and don't know why. Oddly, `ts.pca( samples=x)` and `ts.pca( individuals=x)` match for haploid individuals, but these don't match to the testing code.

petrelharp:
Also - I wonder if a better name for `iterated_power` would be `power_iterations`?

Oh, I see that [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) says `iterated_power` (and also `n_components` instead of `num_components`). Hm. I think that being compatible with them is good, but copying their bad API choices is not requred?  (Well, `n_components` is a fine API, but it isn't consistent with the rest of our API?, which is `num_X` for things?)

python/tskit/trees.py
```diff
@@ -8779,17 +8804,16 @@ def _rand_pow_range_finder(
             """
             Algorithm 9 in https://arxiv.org/pdf/2002.01387
             """
-            assert num_vectors >= rank > 0, "num_vectors should be larger than rank"
+            assert num_vectors >= rank > 0, "num_vectors should not be smaller than rank"
```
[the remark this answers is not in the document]
hanbin973: Good point. geq/leq are tricky.

python/tskit/trees.py
hanbin973: I'm always confused by what's a loading and what's a loading score (or factor). U is should be the score. Will work on it.

jeromekelleher:
I'd vote for compatible with scikit second, compatible with our other APIs first.

petrelharp:
The failure is that the two answers are Definitely Not the Same. Clicking on the "Tests / ..." link above, or running locally, we get
```
FAILED tests/test_relatedness_vector.py::TestPCA::test_individuals[1-0-True] - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-08

Mismatched elements: 5 / 5 (100%)
Max absolute difference among violations: 5733.50077407
Max relative difference among violations: 0.60127089
 ACTUAL: array([3802.135878, 3157.318008, 2779.806231, 2273.337602, 1637.086583])
 DESIRED: array([9535.636652, 5611.718265, 4424.8184  , 4081.358172, 2913.365737])
```
To run this locally, do
```
python -m pytest tests/test_relatedness_vector.py::TestPCA::test_individuals[1-0-True] 
```

I hear what you're saying about comparing `U.T @ U`, but I think we do want to actually test the individual eigenvectors, not just that the low dimensional approx is right. (Or, maybe I don't understand your suggestion?) The testing code is robust to ordering of eigen-things and sign switches; however, if two eigenvalues are very close, then you're right, we could have problems with nonidentifiabilty that way.  Please do have a look at the [testing code](https://github.com/tskit-dev/tskit/pull/3008/files#diff-2ba767a58b99304f147dd1efecdf4e2f9d35c49c061fab21b2104b8c63ae02a9R786), though, to see if you have a better suggestion?

But - I don't think that's the problem - inserting a `print` into the testing code shows that these are the eigenvalues:
```
[3837.56623916, 3364.17632569, 2843.38691979, 2419.60371452, 655.31546225]
```

petrelharp:
Good catch!
```
ndows,)
-        U, D = pca(ts=ts, windows=windows, samples=samples, centre=centre)
+        U, D = pca(ts=ts, windows=windows, centre=centre, samples=samples, individuals=individuals)
```
And - hah - I guess I missed the `...` in the output saying that they're of different lengths.

hanbin973:
Now all the tests are passing in my local machine. The error was due to a minor omission of the individuals option in the `verify_pca()` function.

petrelharp:
Hey, that's great! I need to scrutinize codecov still, but there's one not totally trivial linting issue:
```
python/tskit/trees.py:8886:24: B023 Function definition does not bind loop variable '_f_high'.
python/tskit/trees.py:8888:29: B023 Function definition does not bind loop variable 'indices'.
python/tskit/trees.py:8891:29: B023 Function definition does not bind loop variable 'this_window'.
python/tskit/trees.py:8896:27: B023 Function definition does not bind loop variable '_f_low'.
python/tskit/trees.py:8898:33: B023 Function definition does not bind loop variable 'indices'.
python/tskit/trees.py:8901:33: B023 Function definition does not bind loop variable 'this_window'.
```

(I get this by running the pre-commit hooks; see the developer docs for how to set this up; in particular by
```
pre-commit run --files tskit/trees.py tests/test_relatedness_vector.py 
```

Would you like to have a look at how to deal with that? (I haven't figured out what it means yet.)

hanbin973:
The linting error appears when a global variable inside a function that is defined within a loop is the loop variable. I can't immediately think of a case where this causes a problem, but the point is that the use of the loop variable inside the function should be explicit. An easy hack is to add the loop variable to the function argument and set it as the default value. This is adapted from : https://stackoverflow.com/questions/54947434/function-definition-in-a-for-loop-using-local-variable

hanbin973:
I don't know why docs build is failing to detect the new PCAResult class.

jeromekelleher:
@benjeffery any thoughts on why the docs build is failing? If we squash the commits down to a reasonable number (5, say, or must 1 may be easier) then I think we can log an issue to track beefing up the test coverage a bit, and merge.

benjeffery:
Docs build is failing as there is a link to the `PCAResult` class, but that class is not included in the docs.

hanbin973:
I can't tell how great it is to write code under the guidance of expert maintainers. I added the `.rst` block for `PCAResult` and now the error is gone. I'm not sure if I've put the stuff in the right section though.

petrelharp:
Something funny is going on with codecov: I think the missing line numbers are out of sync with the code. I'd like to at least look at the coverage before merging - let me just squash down the commits, that might help.

jeromekelleher:
> I know, right? I have learned so much from Jerome & Ben.

Likewise Peter! :heart:

hanbin973:
Great. Tests cover all parts of the code and are passing.

jeromekelleher:
Happy to merge when you are @petrelharp

petrelharp:
Let's merge, but there's been no testing so far of the "error bound"; I'd like to get at least a little scrutiny on that before releasing.

hanbin973:
I'll look at it over night.

PR #3010, opened 2024-10-07
Allow multi-sample tips for Tree.draw_svg()

hyanwong:
As discussed with @benjeffery, these are induced by passing a different (postorder) list as the `order` parameter. This additional use of "order" is deliberately not documented. Here's an example:

Screenshot 2024-10-07 at 15 28 01

The stored SVGs have changed because as part of the refactoring I properly rounded some floating point values, so now e.g. positions like `7.0` are stored as simply `7` in the SVG, for clarity and efficiency.

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -175,6 +175,10 @@ def check_order(order):
         "minlex": "minlex_postorder",
         "tree": "postorder",
     }
+    # Silently accept a tree traversal order as a valid order, so we can
+    # call this check twice if necessary
+    if order in traversal_orders.values():
+        return order
     if order not in traversal_orders:
         raise ValueError(
             f"Unknown display order '{order}'. "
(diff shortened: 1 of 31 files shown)


hyanwong:
Note that this should allow us to e.g. concatenate clades containing all non-tracked nodes. I will follow up with another PR to deal with large untracked polytomies.

python/tskit/drawing.py
```diff
@@ -1686,12 +1697,21 @@ def assign_x_coordinates(self):
...
+            if is_tip:
                 leaf_x += 1
                 node_x_coord[u] = leaf_x
             else:
-                child_coords = [node_x_coord[c] for c in self.tree.children(u)]
+                child_coords = [
+                    node_x_coord[c] for c in self.tree.children(u) if c in node_x_coord
+                ]
+                if len(child_coords) == 0:
+                    raise ValueError(
```
jeromekelleher, line 1712: Should try to cover this if possible

jeromekelleher (commented):
LGTM, just need to cover a missed error condition if poss

jeromekelleher (approved):

PR #3012, opened 2024-10-08
Add "pack_polytomy" ability to draw_svg()

hyanwong:
This allows us to visualise trees with large polytomies (such as from tsinfer or sc2ts) by using the "tracked_samples" functionality of a constructed `tskit.Tree`. Specifically, we pack all lineages in a polytomy that containing entirely "untracked" nodes, to allow us to focus on only the tracked nodes.

The `drawing._postorder_tracked_minlex_traversal()` function provides a way to create a postorder traversal that puts all the non tracked-node lineages on the right hand side of the polytomy, which is required for packing. At the moment I have kept this as a private function, and the new (undocumented) functionality is triggered using e.g. 

```python
import os
import numpy as np
import tskit
import tszip

ts_dir = "../data"
filename = os.path.join(ts_dir, "biotite_nj_long_arg_v7_clustloc-mrm_2-rw_10-mgs_10-2020-10-29")
ts = tszip.decompress(filename + ".ts.tsz")

mut_labels = {
    m.id: f"{ts.site(m.site).ancestral_state if m.parent == tskit.NULL else ts.mutation(m.parent).derived_state}{int(ts.site(m.site).position)}{m.derived_state}"
    for m in ts.mutations()
}

pango = "B.1.240"
tracked_nodes = [u for u in ts.samples() if ts.node(u).metadata.get("Viridian_pangolin") == pango] # could also use ti.pango_lineage_samples[pango]
tree = ts.first(tracked_samples=tracked_nodes)
order = list(tskit.drawing._postorder_tracked_minlex_traversal(tree, collapse_tracked=True))
print(len(order), f"nodes in subtree. {len(tracked_nodes)} nodes in cyan are {pango}")
tree.draw_svg(
    time_scale="rank",
    y_axis=True,
    order=order,
    canvas_size=(650, 600),
    size=(600, 600),
    node_labels={u: ts.node(u).metadata.get("Viridian_pangolin", "") for u in order if u not in tracked_nodes},
    mutation_labels=mut_labels,
    all_edge_mutations=True,
    symbol_size=4,
    pack_untracked_polytomies=True,
    style=(
        "".join(f".n{u} > .sym {{fill: cyan}}" for u in tracked_nodes + [39]) +
        ".lab.summary {font-size: 9px}" + 
        ".polytomy {font-size: 10px}" +
        ".mut .lab {font-size: 10px}"
        ".plotbox {transform: translateX(40px)}"
        ".y-axis .lab {font-size: 12px}"
    ),
)
```
Screenshot 2024-10-08 at 12 58 23


jeromekelleher (approved):

hyanwong:
> Let's not worry too much about final API and get something in there for now that we can start playing with.

Yep, 100% agree, which is why I've left it all undocumented

hyanwong:
After discussion with Jerome, it seemed useful to be able to collapse a subtree if (say) > 99% of the samples within it were tracked samples.

It would be relatively easy to change the `collapse_tracked` parameter to expect a proportion (e.g. 0.99), specifying the collapsing threshold. By default, `None` would mean "never collapse" and `1` (or `True`) would mean only collapse if all samples under a node are tracked samples.

hyanwong:
Cool, done. Thanks.

PR #3013, opened 2024-10-08
Correct C changelog for extend_haplotypes

benjeffery:
Reviewing before release I spotted that the method had changed name without the changelog being updated. Hopefully I have this right @petrelharp?


petrelharp:
Good catch, thanks!

PR #3015, opened 2024-10-08
Draw svg title

hyanwong:
I've always wanted to be able to add a title to our SVG plots. Now that we are outputting them e.g. from sc2ts, I think this is vital to keep track of what we are plotting:

Screenshot 2024-10-08 at 23 04 10

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -86,6 +86,8 @@
   be summarised as a dotted line (:user:`hyanwong`, :issue:`3011` :pr:`3010`, :pr:`3012`)
   
 
+- Added a ``title`` parameter to ``.draw_svg()`` methods (:user:`hyanwong`, :pr:`3015`)
+
 --------------------
 [0.5.8] - 2024-06-27
 --------------------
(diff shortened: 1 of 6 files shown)


benjeffery (approved):
LGTM!

jeromekelleher (approved):

python/tskit/intervals.py
```diff
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2023 Tskit Developers
+# Copyright (c) 2024 Tskit Developers
```
jeromekelleher, line 3: Not sure why this file got touched?

jeromekelleher (commented):
Looks like you need to drop intervals.py from the PR?

python/tskit/intervals.py
hyanwong: It didn't, but one of the CI checks for some reason was complaining about in in a previous PR. Happy to remove

hyanwong: I just removed the commit that did the year updates for extra untouched files that were previously complained about.

PR #3016, opened 2024-10-09
Add resources to provenance

benjeffery:
## Description

Add optional `resources` to provenances along with a helper method to populate the field. This is intended for long-running processed such as tsinfer, where measuring the resources used across distributed pipelines is awkward and easier to do in the actual process that makes the tree sequence.

# PR Checklist:

### DIFF
--- a/python/tests/test_provenance.py
+++ b/python/tests/test_provenance.py
@@ -43,9 +43,6 @@ import tskit
 import tskit.provenance as provenance
 
 
-_start_time = time.time()
-
-
 def get_provenance(
     software_name="x",
     software_version="y",
@@ -231,11 +228,12 @@ class TestGetResources:
 
     def test_get_resources_values(self):
         delta = 0.1
-        resources = provenance.get_resources(time.time() - delta)
+        t = time.time()
+        resources = provenance.get_resources(t - delta)
         assert isinstance(resources["elapsed_time"], float)
         assert isinstance(resources["user_time"], float)
         assert isinstance(resources["sys_time"], float)
-        assert resources["elapsed_time"] >= delta
+        assert resources["elapsed_time"] >= delta - 0.001
         assert resources["user_time"] > 0
         assert resources["sys_time"] > 0
         if resource is not None:
@@ -272,7 +270,7 @@ class TestGetSchema:
     def test_form(self):
         s = provenance.get_schema()
         assert s["schema"] == "http://json-schema.org/draft-07/schema#"
-        assert s["version"] == "1.0.0"
+        assert s["version"] == "1.1.0"
 
 
 class TestTreeSeqEditMethods:
--- a/python/tests/test_provenance.py
+++ b/python/tests/test_provenance.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2020 Tskit Developers
+# Copyright (c) 2018-2024 Tskit Developers
 # Copyright (C) 2018 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -26,6 +26,14 @@ Tests for the provenance information attached to tree sequences.
 import json
 import os
 import platform
+import sys
+import time
+
+try:
+    import resource
+except ImportError:
+    resource = None  # resource.getrusage absent on windows
+
 
 import msprime
 import pytest
@@ -33,6 +41,9 @@ import pytest
 import _tskit
 import tskit
 import tskit.provenance as provenance
+
+
+_start_time = time.time()
 
 
 def get_provenance(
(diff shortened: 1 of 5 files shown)


benjeffery:
@jeromekelleher I'm not sure how library methods such as `simplify` that could be run as part of a larger process are going to record their `max_mem`.

python/tskit/provenance.py
```diff
@@ -72,6 +80,31 @@ def get_environment(extra_libs=None, include_tskit=True):
     return env
 
 
+def get_resources():
```
jeromekelleher, line 83: Do we want to do this here? I feel like this will get misused pretty badly as people won't realise that time is relative to when tskit got imported. It's only really appropriate in cases where the time is essentially the full life of the interpreter.

python/tskit/provenance.schema.json
```diff
@@ -45,6 +45,28 @@
...
+        },
+        "user_time": {
+          "description": "User time used in seconds.",
+          "type": "number"
+        },
+        "sys_time": {
+          "description": "System time used in seconds.",
+          "type": "number"
+        },
+        "max_mem": {
```
jeromekelleher, line 65: Let's make this ``max_memory`` here, I've commited this in sc2ts and the brevity doesn't really help.

python/tskit/util.py
```diff
@@ -216,7 +216,8 @@ def pack_arrays(list_of_lists, dtype=np.float64):
     """
     Packs the specified list of numeric lists into a flattened numpy array
     of the specified dtype with corresponding offsets. See
-    :ref:`sec_encoding_ragged_columns` for details of this encoding of columns
+    :ref:`sec_encoding_ragged_columns` for detThis information
```
jeromekelleher, line 219: Some collatoral damage?

jeromekelleher (commented):
I don't know if we want to get into the business of generating this in tskit, although I agree it would be nice to not have to redo it in libraries.

I guess we could require the ``start_time`` as a parameter, which would force users to generate a sensible one?

I think this is something that we don't expect to be recorded much, except in cases where there's a *lot* of resources involved. Msprime, for example, wouldn't bother with this.

benjeffery:
Yes, it was only when I went to add it to existing tskit methods I realised this was the wrong approach. I don't think adding it to all tskit methods is useful. 

I still think the method is useful though, how about I add the `start_time` argument and leave it undocumented?

jeromekelleher:
Sgtm

benjeffery:
@jeromekelleher Ok! Should be good to go.

python/tests/test_provenance.py
```diff
@@ -35,6 +43,9 @@
 import tskit.provenance as provenance
 
 
+_start_time = time.time()
```
jeromekelleher, line 46: Global value probably doesn't help with testing

python/tests/test_provenance.py
```diff
@@ -178,6 +220,33 @@ def test_libraries(self):
...
+    def test_used_resources_keys(self):
+        resources = provenance.get_resources(_start_time)
+        assert "elapsed_time" in resources
+        assert "user_time" in resources
+        assert "sys_time" in resources
+        if resource is not None:
+            assert "max_memory" in resources
+
+    def test_used_resources_values(self):
+        resources = provenance.get_resources(_start_time)
```
jeromekelleher, line 233: Something less fragile would be ``time.time() - delta``, and then test that elapsed time is >= delta.

python/tests/test_provenance.py
```diff
@@ -178,6 +220,33 @@ def test_libraries(self):
...
+        assert isinstance(resources["elapsed_time"], float)
+        assert isinstance(resources["user_time"], float)
+        assert isinstance(resources["sys_time"], float)
+        assert resources["elapsed_time"] > 0.0001
+        assert resources["user_time"] > 0.0001
```
jeromekelleher, line 238: Just do > 0 here, there will surely be cases there this fails in CI or whatever

python/tskit/provenance.py
```diff
@@ -27,6 +27,13 @@
 import json
 import os.path
 import platform
+import sys
+import time
+
+try:
+    import resource
+except ImportError:
```
jeromekelleher, line 35: I think it's just that the module doesn't exist on Win right? Otherwise check wouldn't work.

python/tskit/provenance.py
```diff
@@ -72,6 +79,24 @@ def get_environment(extra_libs=None, include_tskit=True):
...
+def get_resources(start_time):
+    # Returns a dict describing the resources used by the current process
+    times = os.times()
+    ret = {
+        "elapsed_time": time.time() - start_time,
+        "user_time": times.user + times.children_user,
+        "sys_time": times.system + times.children_system,
+    }
+    if resource is not None:
+        # Don't report max memory on Windows. We could do this using the psutil lib, via
```
jeromekelleher, line 91: This comment isn't accurate for tskit - I don't think we'd want a dependency on psutil.

python/tskit/provenance.py
```diff
@@ -72,6 +79,24 @@ def get_environment(extra_libs=None, include_tskit=True):
...
+        # psutil.Process(os.getpid()).get_ext_memory_info().peak_wset if demand exists
+        ret["max_memory"] = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+        if sys.platform != "darwin":
+            ret["max_memory"] *= 1024  # Linux, freeBSD et al reports in KB, not bytes
```
jeromekelleher, line 95: Should be specific it's KiB not KB (10**3)

jeromekelleher (approved):
Spotted a few more things, but LGTM then.

We should get some wider input on this before adding to schema i guess, maybe do a shout out on Slack?

benjeffery:
Comments addressed in 06a41328

benjeffery:
@jeromekelleher feedback on slack was good, think we can merge now

jeromekelleher:
Let's do it. This is a really useful addition, great idea!

benjeffery:
@mergifyio rebase

PR #3018, opened 2024-10-11
Comma sep display numbers

### LINKED ISSUE #3017: Values as comma-separated thousands
It would improve readability for large values if we could use commas to separate thousands, for example when displaying the various tables or the summary table.

benjeffery:
Fixes #3017


jeromekelleher:
Definitely worth a changelog entry

PR #3021, opened 2024-10-16
Implement tree.ancestors

### LINKED ISSUE #2706: Tree.ancestors(u) method
As I've been implementing tree traversal methods, I've commonly wanted to go up the chain of parents for a node. I think there's a good argument for a `tree.parents(u)` (or `tree.ancestors?`) method which returns an iterator which then stops when the root is reached. This is a pretty fundamental tree operation and handy to avoid messing with all that `tskit.NULL` stuff for newbies:

```python
def parents(self, u):
    while self.parent(u) != tskit.NULL:
        u = self.parent(u)
        yield u
```

Thoughts? Worth the extra API burden?

hyanwong:
Fixes #2706

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -551,6 +551,7 @@ Iterator access
     .. autosummary::
 
       Tree.nodes
+      Tree.ancestors
 
 Array access
     .. autosummary::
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
LGTM

benjeffery:
@mergifyio rebase

PR #3022, opened 2024-10-16
Use more recent Sphinx

hyanwong:
I seem to need this to get the docs compiling on my Mac laptop, which seems to require Sphinx >= 5. Is there any reason we pin to 4.5 specifically? 7.4.7 is the latest in the v7 series, and v8 is too bleeding edge to work smoothly.

### DIFF
--- a/python/requirements/development.txt
+++ b/python/requirements/development.txt
@@ -28,7 +28,7 @@ pytest
 pytest-cov
 pytest-xdist
 si-prefix
-sphinx==4.5
+sphinx==7.4.7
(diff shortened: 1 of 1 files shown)


benjeffery:
For the docs CI we don't specify - the deps set it implicitly at 7.4.7, so yes this should be updated to that.

benjeffery (approved):

benjeffery:
@mergifyio rebase

benjeffery:
@mergifyio rebase

PR #3023, opened 2024-10-16
python3 -> python in Makefile

hyanwong:
My installation doesn't bother defining python3 any more. I think we can assume python is v3 now

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -71,11 +71,11 @@ an overview of how to contribute a new feature to `tskit`.
 ### Requirements
 
 To develop the Python code you will need a working C compiler and a
-development installation of Python (>= 3.8). On Debian/Ubuntu we can install these
+development installation of Python (>= 3.9). On Debian/Ubuntu we can install these
 with:
 
 ```bash
-$ sudo apt install python3-dev build-essential doxygen
+$ sudo apt install python3-dev python-is-python3 build-essential doxygen
 ```
 
 Python packages required for development are listed in `python/requirements/development.txt`.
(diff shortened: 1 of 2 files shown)


benjeffery:
Yes, we're at the awkward point where some have removed `python3` and some still have 2.x. I think we break on more people's machines with `python3` now, so happy to change.

benjeffery (approved):

jeromekelleher:
sorry I'm going to -1 this direct change as it breaks my setup, and I suspect will for Peter and a few others who regularly work with the C and are on Debian based Linux distros

Is there some other way to do this? Or can we just leave it for a bit?

jeromekelleher (commented):
See point about breaking stuff

hyanwong:
Oh, sorry. I didn't realise there were still python2 installations out there. I guess just leave it until Debian links python -> python3 rather than python2?

Alternatively, `sudo apt update && sudo apt install python-is-python3`, according to https://unix.stackexchange.com/questions/609855/how-to-make-python-an-alias-of-python3-systemwide-on-debian?

jeromekelleher:
Ah, hold on:
```
sudo apt install python-is-python3
```
sorts it.

Could we add ``python-is-python3`` to the requirements listed [here](https://tskit.dev/tskit/docs/stable/development.html#requirements) and elsewhere please? Maybe we could add a tip

benjeffery:
What are you running? Python 2 got taken out of ubuntu 4 years ago.

hyanwong:
> What are you running? Python 2 got taken out of ubuntu 4 years ago.

I think Debian simply doesn't define a `python` (not even for python2, which is not installed by default anyway), unless you install `python-is-python3`

jeromekelleher:
> What are you running? Python 2 got taken out of ubuntu 4 years ago.

You don't get python -> to python3 automatically, you have to install python-is-python3 (which I haven't been doing, up to now).

So, I'm happy to make the change, we just need to update the development docs so that they work.

benjeffery:
I've added a note to the docs and a friendly error message if the makefile finds python missing.

jeromekelleher:
Nice, thanks Ben

PR #3043, opened 2024-10-25
Updates to two-locus branch stats (python)

lkirk:
During the validation of the original branch LD algorithm, we realized that the LD matrix could get "poisoned" with NaN values if we attempted to make an adjustment to a node that did not contain any samples, which occurs with some frequency. Once a NaN enters our running sum, the rest of the values in the LD matrix are NaN. Another issue we're seeing is that this algorithm is rather slow. It's tough to avoid with something that is inherently quadratic, but calling the summary function twice for each parent of the added/removed edge was something we were trying to avoid.

We tore things apart and simplified the algorithm so that we no longer have to do adjustments as we're adding and removing edges. This new version removes the LD contribution from all modified nodes and adds the contribution from all nodes at the end of the routine, once we know the final state of samples under each node.

In addition, we're seeing that we call the summary function ~30% fewer times, which is a nice improvement because that's the most expensive operation. In addition, we no longer need to track child samples from the edge being added. I've done a small, microbenchmark to show this improvement: [here](https://github.com/lkirk/tskit-ld/blob/2d91a78/analysis/misc/NewBranchAlgoBench.ipynb). We do see a runtime improvement, but I'm curious to know how this looks in the C implementation, 

We do need to iterate over the affected nodes (which need to be stored uniquely), so I've come up with an algorithm for pulling items out of a bit array. The python version contained the `get_items` function for debugging, but now it'll be used in the iteration over modified nodes. I also did a small shootout (in C) for these types of algorithms and this beats linear search by ~50%.

### DIFF
--- a/python/tests/test_ld_matrix.py
+++ b/python/tests/test_ld_matrix.py
@@ -132,15 +132,28 @@ class BitSet:
 
     def get_items(self: "BitSet", row: int) -> Generator[int, None, None]:
         """Get the items stored in the row of a bitset
+        Uses a de Bruijn sequence lookup table to determine the lowest bit set.
+        See the wikipedia article for more info: https://w.wiki/BYiF
 
         :param row: Row from the array to list from.
         :returns: A generator of integers stored in the array.
         """
+        lookup = [0, 1, 28, 2, 29, 14, 24, 3, 30, 22, 20, 15, 25, 17, 4, 8, 31, 27,
+                  13, 23, 21, 19, 16, 7, 26, 12, 18, 6, 11, 5, 10, 9]  # fmt: skip
+        m = np.uint32(125613361)
         offset = row * self.row_len
         for i in range(self.row_len):
-            for item in range(self.CHUNK_SIZE):
-                if self.data[i + offset] & (self.DTYPE(1) << item):
-                    yield item + (i * self.CHUNK_SIZE)
+            v = self.data[i + offset]
+            if v == 0:
+                continue
+            else:
+                # v & -v operations rely on integer overflow
+                with np.errstate(over="ignore"):
+                    lsb = v & -v  # isolate the least significant bit
+                    while lsb:  # while there are bits remaining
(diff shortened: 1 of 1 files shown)


python/tests/test_ld_matrix.py
```diff
@@ -132,15 +132,28 @@ def add(self: "BitSet", row: int, bit: int) -> None:
...
     def get_items(self: "BitSet", row: int) -> Generator[int, None, None]:
         """Get the items stored in the row of a bitset
+        Uses a de Bruijn sequence lookup table to determine the lowest bit set.
+        See the wikipedia article for more info: https://w.wiki/BYiF
 
         :param row: Row from the array to list from.
         :returns: A generator of integers stored in the array.
         """
+        lookup = [0, 1, 28, 2, 29, 14, 24, 3, 30, 22, 20, 15, 25, 17, 4, 8, 31, 27,
+                  13, 23, 21, 19, 16, 7, 26, 12, 18, 6, 11, 5, 10, 9]  # fmt: skip
```
lkirk, line 142: This is a minimum perfect hash to obtain the set bit, given the least significant bit in the chunk

lkirk, line 142: This is a minimum perfect hash to obtain the set bit, given the least significant bit in the chunk

jeromekelleher (commented):
Updates look pretty straightforward here @lkirk, so seem worth the effort.

I think it's important to accept the limitations of the quadratic nature of the calculation  though, and I'd imagine we'd be well into diminishing returns after this batch of optimisations.

lkirk:
@jeromekelleher Agreed, thanks for taking a look. Commits are squashed and checks are passing. Anything else you need on my end?

jeromekelleher (approved):

jeromekelleher:
Nope, happy to merge if you are

lkirk:
Yes, this is ready for merge

PR #3050, opened 2024-11-06
Fix SVG pack_untracked_polytomies bug

### LINKED ISSUE #3049: (Undocumented) pack_untracked_polytomies option not correctly identifying branches to pack
Fortunately we haven't documented this option yet:

```python
import tskit
tree = tskit.Tree.generate_balanced(20, arity=4, tracked_samples=np.arange(2, 8))
tree.draw_svg(size=(1000, 400), pack_untracked_polytomies=True)
```

gives 

Screenshot 2024-11-06 at 15 55 30

But it shouldn't be showing all those collapsed nodes. Instead, I think this should look like this:

Screenshot 2024-11-06 at 15 52 17

hyanwong:
Fixes #3049

# PR Checklist:

### DIFF
--- a/python/tests/test_drawing.py
+++ b/python/tests/test_drawing.py
@@ -2602,6 +2602,19 @@ class TestDrawSvg(TestDrawSvgBase):
             )
             assert "Ticks {10: '10'} lie outside the plotted axis" in caplog.text
 
+    def test_polytomy_collapsing(self):
+        tree = tskit.Tree.generate_balanced(
+            20, arity=4, tracked_samples=np.arange(2, 8)
+        )
+        svg = tree.draw_svg(pack_untracked_polytomies=True)
+        # Should have one collapsed node (untracked samples 8 and 9)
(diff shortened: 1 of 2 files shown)


python/CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 [0.6.1] - 2024-XX-XX
 --------------------
 
+**Bugfixes**
+
+- Fix to ``Tree.draw_svg()`` with the (undocumented) experimental
+  ``pack_untracked_polytomies=True`` option. (:user:`hyanwong`,
```
jeromekelleher, line 8: Probably doesn't need to go in changelog as it's undocumented (we're sort of documenting it otherwise)

jeromekelleher (approved):

python/CHANGELOG.rst
hyanwong: Fine, I wasn't sure about that one either. Now removed.

PR #3058, opened 2024-11-19
Correct output dimension for `ts.pair_coalescence_counts` with extra time windows

### LINKED ISSUE #3046: `ts.pair_coalescence_counts` truncates output array dimension to last time window with a node
If `ts.pair_coalescence_counts` is provided with time windows, the second dimension of the output array should have size `time_windows.size - 1` even if there are no nodes beyond one of the breakpoints. Current, the time windows are truncated to the breakpoint above the last node in the tree sequence.

nspope:
Fixes #3046

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,6 +2,12 @@
 [0.6.1] - 2024-XX-XX
 --------------------
 
+**Bufixes**
+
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):

nspope:
mergify issues again:

> nspope does not have write access to the forked repository.

PR #3059, opened 2024-11-20
Normalize pair coalescence counts/rates using non-missing span

### LINKED ISSUE #3053: `pair_coalescence_rate` doesn't account for empty regions (e.g. flanks)
You might hope that the span-normalised coalescence rate would be the same for a given tree, even if the tree sequences has empty regions at the start or end, or even in the middle. However, this test fails:

```python
ts = tskit.Tree.generate_comb(10, span=10).tree_sequence
time_intervals = np.logspace(0, np.log10(ts.max_time), 4)
time_intervals = np.concatenate(([0], time_intervals, [np.inf]))
ts_with_missing = ts.delete_intervals([[0,2],[6,8]])
assert np.allclose(
    ts.pair_coalescence_rates(time_intervals),
    ts_with_missing.pair_coalescence_rates(time_intervals),
)
```

It passes once we trim down:

```python
ts_trimmed = ts.delete_intervals([[0,2],[6,10]]).trim()
assert np.allclose(
    ts.pair_coalescence_rates(time_intervals),
    ts_trimmed.pair_coalescence_rates(time_intervals),
)
```

@nspope says
> I see what the problem is. Basically, I'm assuming weights sum to 1 so that we can calculate rates even when there are multiple roots (e.g. decapitated trees). That won't be the case if we're incorrectly normalising, however

nspope:
Fixes #3053

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -316,12 +316,15 @@ verify_pair_coalescence_counts(tsk_treeseq_t *ts, tsk_flags_t options)
     const double *breakpoints = tsk_treeseq_get_breakpoints(ts);
     const tsk_size_t P = 2;
     const tsk_size_t I = P * (P + 1) / 2;
+    const tsk_size_t B = 8;
     tsk_id_t sample_sets[n];
     tsk_size_t sample_set_sizes[P];
     tsk_id_t index_tuples[2 * I];
     tsk_id_t node_bin_map[N];
     tsk_size_t dim = T * N * I;
     double C[dim];
+    double C_B[T * B * I];
+    double C_Nh[T * (N / 2) * I];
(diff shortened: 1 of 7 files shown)


jeromekelleher (approved):
Nice, great use of the ``num_edges`` count :+1:

nspope:
The changelog is updated, and I added some tests checking the behavior of the statistics on empty windows. In particular, `pair_coalescence_counts` will return zeros (rather than NaNs) if `span_normalise=True` and a window has no nonmissing span; the other two functions will return NaNs.

jeromekelleher (approved):
LGTM. Great to see these awkward corner cases getting hit so soon, this stuff is definitely getting used!

PR #3067, opened 2024-12-05
Clarify mutation order in Site() object

hyanwong:
This is useful to know without having to dive into the order requirements docs. I'm often looking this up to find the inherited state at a node.

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -370,7 +370,11 @@ class Site(util.Dataclass):
     mutations: np.ndarray
     """
     The list of mutations at this site. Mutations within a site are returned in the
-    order they are specified in the underlying :class:`MutationTable`.
+
+    order they are specified in the underlying :class:`MutationTable`. For canonical
+    (i.e., valid) tables, this means ancestral mutations precede their descendants, so
+    older mutations (as defined by the canonical mutation ordering; see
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Note that there are are loopholes in the mutation data requirements which means that it's possible for this to not actually be true (which is a bug, but not one we can easily resolve now)

hyanwong:
Right, I know that the sorting process doesn't necessarily enforce parent/child correctness (and this is noted in the docs) but they also say:

> when there are multiple mutations per site, mutations should be ordered by decreasing time, if known, and parent mutations must occur before their children..
> Violations of these sorting requirements are detected at load time.

I guess violations of mutation parent order are not (yet?) detected at load time, so the doc wording should be changed to point out this bug?

jeromekelleher:
It's a bit tricky. Maybe you could put in a link to the definitions instead of explaining, so at least it's all in one place?

hyanwong:
Maybe we just leave this open until it's (eventually) fixed? It does my head in trying to figure out from the rather involved mutation sorting requirements that the most recent mutations for a site are (should be) at the end of the list. I feel that just needs to be stated simply somewhere, for the non-technical reader.

benjeffery (approved):

PR #3090, opened 2025-02-28
Add fixed-length arrays in metadata

### LINKED ISSUE #3088: Add fixed-length array support to struct metadata
@bhaller has a use case for this (discussed on the slack) would look something like:

`{"type": "array", {"items": {"type":"number", "binaryFormat":"B"}}, "length":"5"}`

This would also decode well with numpy structured arrays.

benjeffery:
Fixes #3088 

This turned out to be very straightforward so I went ahead and finished it.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -532,6 +532,17 @@ types above. `L` is the default. As an example:
 
 Will result in an array of 2 byte integers, prepended by a single-byte array-length.
 
+For arrays with a known fixed size, you can specify the `length` property instead:
+```
+{"type": "array", "length": 3, "items": {"type":"number", "binaryFormat":"i"}}
+```
+This creates a fixed-length array of exactly 3 integers, without storing the array length in the encoded data. 
+Fixed-length arrays are more space-efficient since they don't need to store the length prefix.
+
+When using fixed-length arrays:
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
LGTM!

benjeffery:
@bhaller - can you check this does what you wanted?

bhaller:
It looks great!  The only comment I'd have is that none of the unit tests appears to test with type `B`; they're all with `i` and `d`, it looks like.  I imagine that if it works for `i` and `d` then it works for `B` and others, so perhaps that isn't really worth testing; but `B` is what I'll be using.

I'm quite hesitant to try it out myself by trying to install this PR on my machine; in general, I live in constant fear of the possibility of breaking my Python configuration, since I have so little understanding of that.  :-O  I wouldn't know how to install from this PR anyway, so that would be a whole project for me.  :->  But I reviewed the diffs carefully, and they look great apart from my one comment above.

Thanks!

benjeffery:
Yeah, impossible to test with all the combinations given that it supports arrays of objects!

I really recommend virtualenvs. I haven't had a python installation snafu in over a decade.

bhaller:
Maybe someday I'll figure that out.  :->

PR #3091, opened 2025-02-28
Add `numpy_dtype` method to metadata codecs.

benjeffery:
This will need other infrastructure around it, but this function gives a numpy structured dtype for `struct`  metadata schemas, allowing fast, copy-free decoding of metadata buffers!

Stacked on #3090  - see https://github.com/tskit-dev/tskit/commit/6ea9b7e8a26fd9b9ef7a5ab6d308a8b204e90d14 for the clean diff.

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -103,7 +103,7 @@ class AbstractMetadataCodec(metaclass=abc.ABCMeta):
     def decode(self, encoded: bytes) -> Any:
         raise NotImplementedError  # pragma: no cover
 
-    def numpy_dtype(self) -> Any:
+    def numpy_dtype(self, schema) -> Any:
         raise NotImplementedError
 
 
@@ -675,6 +675,7 @@ class StructCodec(AbstractMetadataCodec):
     def numpy_dtype(self, schema):
         # Mapping from struct format characters to NumPy dtype strings
         # Note: All are little-endian as enforced by the struct codec
+        # This means they will be the standard size across platforms
         FORMAT_TO_DTYPE = {
             # Boolean
             "?": "?",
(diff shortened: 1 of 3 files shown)


python/tskit/metadata.py
```diff
@@ -603,6 +672,93 @@ def decode(self, encoded: bytes) -> Any:
         # Set by __init__
         pass  # pragma: nocover
 
+    def numpy_dtype(self, schema):
+        # Mapping from struct format characters to NumPy dtype strings
+        # Note: All are little-endian as enforced by the struct codec
```
jeromekelleher, line 677: So given that we're forcing little endian, these all have the "standard size". We should note that.

jeromekelleher (commented):
Not grokked the details, but looks like a great start!

python/tskit/metadata.py
benjeffery: Not sure I get what you mean here - they would be the same size if they were big-endian?

jeromekelleher: Format characters (https://github.com/tskit-dev/tskit/pull/3091#discussion_r1976098928) by default are platform dependent, so ``i`` means a C int, which can be anything from 2 to 8 bytes depending on your platform. By explicitly specifying "<i" struct uses the "standard size" so it's guaranteed to be 4 bytes.

benjeffery:
I've added a commit that adds `TreeSequence.individuals_metadata` as an example of how the high-level API would work.

jeromekelleher:
Looks great! Let's tidy up remaining bits of test coverage and merge, then follow up with getters for the various other metadata arrays.

benjeffery:
I've removed the individuals metadata method and will do that in a separate commit. This PR is now just the `numpy_dtype` and `structured_array_from_buffer` methods on `MetadataSchema`. 

We talked about string encoding last week. The `StructCodec` allows specification of the string encoding from any on the list at https://docs.python.org/3.12/library/codecs.html#standard-encodings. Numpy however only supports no encoding ("S" dtype) or `utf-32` ("U" dtype), so we would need to go with "S" which ends up with `bytes` objects when you do, e.g. `array["field"][0]`.

We could run `numpy.char.decode` on the field in the structured array and then re-insert it - but that does mean copying and re-shuffling the structured array, which seems like it should be left as a choice for the user.

benjeffery:
@jeromekelleher modulo the string stuff this is ready to go

jeromekelleher:
OK, shall we leave strings out of this and log an issue for follow-up?

jeromekelleher (approved):

PR #3095, opened 2025-03-06
Initial definition for TSK_TRACE_ERRORS

jeromekelleher:
This is a proposal to address the discussion in #3094.

Essentially, we add a function-like macro ``tsk_trace_error``, and everywhere we normally do
```c
ret = TSK_ERR_X;
goto out;
```
we instead do
```c
ret = tsk_trace_error(TSK_ERR_X);
goto out;
```

We can then define ``tsk_trace_error`` to be a no-op for production code, and define a function which emits some error traces. For the changes I've made here, we get this on stderr:
```
tskit-trace-error: -701='Sequence length must be > 0. (TSK_ERR_BAD_SEQUENCE_LENGTH)' at line 11000 in ../tskit/tables.c
```

## Implementation notes

- tsk_trace_error must be a macro, otherwise we can't use __FILE__ and __LINE__ to get locations
- This structure has no change on control flow, which is essential
- It should be easy enough to sed the changes to implement this for the vast majority of the library.

I can push this through and document if folks like it.

### DIFF
--- a/c/CHANGELOG.rst
+++ b/c/CHANGELOG.rst
@@ -2,7 +2,11 @@
 [1.1.4] - 2024-XX-XX
 --------------------
 
+**Changes**
 
+- Added the TSK_TRACE_ERRORS macro to enable tracing of errors in the C library.
+  This is useful for debugging as errors will print to stderr when set.
+  (:user:`jeromekelleher`, :pr:`3095`).
 
 --------------------
 [1.1.3] - 2024-10-16
--- a/c/meson.build
+++ b/c/meson.build
@@ -3,6 +3,11 @@ project('tskit', ['c', 'cpp'],
     default_options: ['c_std=c99', 'cpp_std=c++11']
 )
 
+debug_c_args = []
+if get_option('buildtype').startswith('debug')
+    debug_c_args = ['-DTSK_TRACE_ERRORS']
+endif
+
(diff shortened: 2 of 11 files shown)


benjeffery:
Seems like a solid idea to me, should be able to hit most of the changepoints with a regex!

jeromekelleher:
OK cool, I'll fill this out as I get a chance.

benjeffery:
Did you want this for 0.6.1? I'm happy to finish it up.

jeromekelleher:
Would be great if you could pick this up, thanks @benjeffery. I think the code changes are done, it's just a case of documenting what's going on in the developer docs and telling devs how to turn on the flag at compile time.

benjeffery:
@jeromekelleher I've added some detail to the docs and added some `meson` config that turns on the flag if `--buildtype=debug` is set.

jeromekelleher:
That's great, thanks. I think we do need to document the actual setting of ``-DTSK_TRACE_ERRORS`` though, as users of the C API aren't necessarily using meson.

benjeffery:
Ok added! Needs a squash, I'll do that if it all looks good.

jeromekelleher (commented):
LGTM! I can't approve as it's my PR. Squash away

benjeffery (approved):

PR #3098, opened 2025-03-07
Structured array highlevel accessors

benjeffery:


python/_tskitmodule.c
```diff
@@ -10760,7 +10760,7 @@ TreeSequence_get_individuals_metadata_offset(TreeSequence *self, void *closure)
     }
     individuals = self->tree_sequence->tables->individuals;
     ret = TreeSequence_make_array(
-        self, individuals.num_rows + 1, NPY_UINT32, individuals.metadata_offset);
+        self, individuals.num_rows + 1, NPY_UINT64, individuals.metadata_offset);
```
jeromekelleher, line 10763: Hmm, did the tests pass before because we didn't look at the contents of the offset arrays?

python/tskit/trees.py
```diff
@@ -5788,6 +5801,52 @@ def individuals_flags(self):
...
+        The returned dtype will depend on the metadata schema used. Only a subset
+        of `struct` metadata schemas are supported.
+        See :ref:`sec_structured_array_metadata` for more information.
+        """
+        return self.table_metadata_schemas.individual.structured_array_from_buffer(
+            self._individuals_metadata
+        )
+
+    @property
+    def individuals_metadata_offset(self):
```
jeromekelleher, line 5818: I'm not sure we want to offer this array in the high-level API. It may just be confusing. I would pull this out of the current PR, as it's orthogonal

jeromekelleher (commented):
Looks good

python/_tskitmodule.c
benjeffery: Yeah, the low-level tests don't check the contents, just the semantics and lifetime/memory concerns.

benjeffery:
@jeromekelleher one more thought here - we could return the raw metadata bytes if the codec doesn't support structuring?

python/tskit/trees.py
benjeffery: Offset accessors removed in https://github.com/tskit-dev/tskit/pull/3098/commits/b74b824a2a20a245542bf7600e5ae2a82f73bbeb

jeromekelleher:
Let's not - maybe we'll think of a better way of doing things in future and having it error out now leaves those options open.

jeromekelleher (approved):
LGTM. Would be good to have a perf check just to be sure before release, but happy to merge this.

PR #3099, opened 2025-03-12
Updated branch algorithm for two-locus stats

lkirk:
I apologize for the lapse between this PR and the initial python prototype update (#3043). We wanted to do some validation before submitting this. These changes follow along with the aforementioned PR and will be the last of optimization/refinement of the branch method.

A brief description of these changes, as a refresher...

These updates remove redundant adjustments to the running sum of our statistic. They also make the computation more resiliant to being "poisoned" by NaN values (which happens when we make an adjustment to a node with no samples -- a frequent occurrence).

In essence, we defer the removal of the LD contribution from removed samples and the addition of the LD contribution from added samples until all edges have been added to the tree (during tree transitions).


jeromekelleher (approved):
LGTM!

Issue #3100, opened 2025-03-13
test error bound in PCA and provide guidance on number of iterations


petrelharp:
We've merged #3008, but aren't actually testing whether the returned `error_bound` is doing something sensible. This will be a bit tricky as it's stochastic, so some preliminary experimentation should point us at the right way to do the testing. For instance, maybe it's true that the max diff between the true eigenvectors and ones returned by `.pca` is no more than (in the tests) `5 * error_bound`. This doesn't catch any problems where the error bound is too *big*, though.

So, I think we want two separate things:
- some benchmarking, to check the error bound is doing the right thing
- some simple unit tests that checks the right thing is being computed, at least roughly, without testing that the statistical properties of that thing are as advertised.

Also, over in #3008 we suggested renaming `iterated_power` to `power_iterations`, or even `num_iterations`.

hanbin973:
https://github.com/tskit-dev/tskit/pull/3102 addresses the issue. I wrote the test code, spotted a bug, fixed it, and now everything is running okay.

PR #3102, opened 2025-03-14
error bound test code

hanbin973:
## Description

This addresses the Issue https://github.com/tskit-dev/tskit/issues/3100.
The test is passing in my local machine. 

I'm just capturing the equations from https://arxiv.org/pdf/0909.4061
The error bound returned by the program is the spectral norm (or the operator norm) of the following 
image

Q is the `range_sketch` matrix and the randomized SVD output is the full SVD of `QQ*A` by construction.
Hence, the error we are computing is the reconstruction error: The difference between the GRM and the `USV*` where the latter is the SVD we obtained from our implementation.

The bound we return is
image
Although the statement is true only on average, the authors show in simulations that the bound is pretty conservative, hence holding in practice with high probability.

I also spotted a bug in the original form that returned an extremely small error due to picking the wrong eigenvalue.

### DIFF
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -1220,7 +1220,7 @@ class TestPCA:
         self.verify_error_est(
             ts,
             num_windows=num_windows,
-            num_components=5,
+            num_components=4,
             centre=centre,
             samples=samples,
             time_windows=[time_low, time_high],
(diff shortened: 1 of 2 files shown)


petrelharp:
I *love* all the details. This is excellent. I'm having a look.

hanbin973:
I suspect two things. The estimate uses the k+1-th singular value estimate to make the bound. This number doesn't seem to be precise enough when it's close the size of the matrix, as we're using small matrices for testing. A simple fix was to use only 4 components instead of 5 and it's passing in my local mac + github test. 

I think it's usually less than 10 power iterations in most applications. When I was playing with the Quebec data some time ago, anything above 4 usually was accurate enough.

petrelharp:
Awesome! That makes sense. I'm going to merge this then have a go at updating the docs in a separate PR.

PR #3103, opened 2025-03-15
docs and argument adjustments to pca()

petrelharp:
I've made some minor changes to the docs for `ts.pca()`.

This includes documenting what `error_bound` means - but, thinking about it, I'm now not sure how helpful that actually is? As noted in #3102, it's giving a bound on the spectral norm of `M - U U^* M`, where `M` is the GRM and `U` is the range sketch, i.e., our approx to the subspace spanned by the first `num_components` eigenvectors of `M`. The spectral norm is just the largest singular value. When we get the top 20 PCs in a 1e5 x 1e5 matrix, we don't expect our 20-dimensional approximation to be super close to the GRM. But, does this somehow tell us how close the results are to the true PCs?

Oh, and I've also changed `iterated_power` to `num_iterations`, diverging from `scikit-learn` (since they have terrible notation IMO); feel free to say I should change it back.

### DIFF
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -1010,14 +1010,14 @@ class TestPCA:
                 num_oversamples=2,
                 range_sketch=np.zeros((ts.num_samples, 4)),
             )
-        with pytest.raises(ValueError, match="iterated_power should be"):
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM!

Can you have a quick look please @hanbin973 ?

PR #3104, opened 2025-03-19
error bound tests

petrelharp:
Here's a more precise description of what (we think) the error bound implies for precision, and additional tests.

I was having trouble getting this to pass, but then realized it was because the `time_window` in the test was making it so the matrix was low rank in some genomic windows. So, I removed the time windowing (we don't need to test that part here, as I don't see how it'd interact with the random approximation besides in trivial ways like this).

Also note that the error_bound is labeled "experimental" - which is good, as this is not a rigorous benchmarking.

### DIFF
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -838,11 +838,12 @@ def assert_errors_bound(pca_res, D, U, w=None):
     # and there must be at least one b_i that is big (since sum_i b_i^2 = 1 - |delta|^2).
     # More concretely, let m = min_i |lambda - L_i|^2,
     # so that
-    #  epsilon > \sum_i (lambda - L_i)^2 b_i^2 + lambda^2 |delta|^2
+    #  epsilon^2 > \sum_i (lambda - L_i)^2 b_i^2 + lambda^2 |delta|^2
     #   >= m * \sum_i b_i^2 + lambda^2 |delta|^2
     #   = m * (1-|delta|^2) + lambda^2 |delta|^2.
-    # Hence, min_i |lambda-L_i|^2 = m < (epsilon - lambda^2  |delta|^2) / (1- |delta|^2).
-    # In summary: epsilon should be the bound on error in eigenvalues,
+    # Hence,
+    # min_i |lambda-L_i|^2 = m < (epsilon^2 - lambda^2  |delta|^2) / (1- |delta|^2).
+    # In summary: roughly, epsilon should be the bound on error in eigenvalues,
     # and epsilon / sigma[k+1] the L2 bound for eigenvectors
     # Below, the 'roughly/should be' translates into the factor of 5.
 
--- a/python/tests/test_relatedness_vector.py
+++ b/python/tests/test_relatedness_vector.py
@@ -818,6 +818,50 @@ def assert_pcs_equal(U, D, U_full, D_full, rtol=1e-5, atol=1e-8):
         assert found_it, f"{k}th singular vector {u} not found in {U_full}."
 
 
+def assert_errors_bound(pca_res, D, U, w=None):
+    # Bounds on the error are from equation 1.11 in https://arxiv.org/pdf/0909.4061 -
+    # this gives a bound on reconstruction error (i.e., spectral norm between the GRM
+    # and the low-diml approx). But since the spectral norm is
+    # |X| = sup_v |Xv|^2/|v|^2,
+    # this implies bounds on singular values and vectors also:
+    # If G v = lambda v, and we've got estimated singular vectors U and values diag(L),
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM. @hanbin973 can you have a look please?

python/tests/test_relatedness_vector.py
```diff
@@ -818,6 +818,50 @@ def assert_pcs_equal(U, D, U_full, D_full, rtol=1e-5, atol=1e-8):
         assert found_it, f"{k}th singular vector {u} not found in {U_full}."
 
 
+def assert_errors_bound(pca_res, D, U, w=None):
+    # Bounds on the error are from equation 1.11 in https://arxiv.org/pdf/0909.4061 -
+    # this gives a bound on reconstruction error (i.e., spectral norm between the GRM
+    # and the low-diml approx). But since the spectral norm is
+    # |X| = sup_v |Xv|^2/|v|^2,
```
hanbin973, line 825: The correct definition is `|X| = sup_v |Xv|_2 / |v|_2`.

petrelharp: oh good! I got it in my head that it was squared, which seemed wrong.

python/tests/test_relatedness_vector.py
```diff
@@ -818,6 +818,50 @@ def assert_pcs_equal(U, D, U_full, D_full, rtol=1e-5, atol=1e-8):
...
+    # then let v = \sum_i b_i u_i + delta be the projection of v into U,
+    # and we have that
+    #  |lambda v - U L U* v|^2
+    #   = \sum_i b_i (lambda - L_i)^2 + lambda^2 |delta|^2
+    #   < \epsilon   (where epsilon is the spectral norm bound error_bound)
+    # so
+    #  |delta| < sqrt(epsilon / lambda)
+    # since this is the amount by which the eigenvector v isn't hit by the columns of U.
+    # Then also for each i that if b_i is not small then
+    #  |lambda - L_i| < sqrt(epsilon)
```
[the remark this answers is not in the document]
hanbin973: The RHS is `epsilon + |delta|^2 * (epsilon - \lambda^2) / (1-|delta|^2)` and assuming that `lambda` is suitably larger than `epsilon` which is roughly `lambda_{k+1}` (k is the rank of the approximation, the formula in eq 1.11 of the paper),  the residual term `|delta|^2 * (epsilon - \lambda^2) / (1-|delta|^2)` is negative, proving the inequality. This somewhat explains why we can't trust the l-th component that is too close to the k-th component: the singular value of the l-th component should be reasonably larger than the k+1-th component for the bound to hold.

python/tests/test_relatedness_vector.py
```diff
@@ -818,6 +818,50 @@ def assert_pcs_equal(U, D, U_full, D_full, rtol=1e-5, atol=1e-8):
...
+    # Bounds on the error are from equation 1.11 in https://arxiv.org/pdf/0909.4061 -
+    # this gives a bound on reconstruction error (i.e., spectral norm between the GRM
+    # and the low-diml approx). But since the spectral norm is
+    # |X| = sup_v |Xv|^2/|v|^2,
+    # this implies bounds on singular values and vectors also:
+    # If G v = lambda v, and we've got estimated singular vectors U and values diag(L),
+    # then let v = \sum_i b_i u_i + delta be the projection of v into U,
+    # and we have that
+    #  |lambda v - U L U* v|^2
+    #   = \sum_i b_i (lambda - L_i)^2 + lambda^2 |delta|^2
```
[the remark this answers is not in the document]
petrelharp: whoops - typo!

python/tests/test_relatedness_vector.py
```diff
@@ -820,25 +820,32 @@ def assert_pcs_equal(U, D, U_full, D_full, rtol=1e-5, atol=1e-8):
...
+    #  |lambda - L_i| < epsilon
     # and there must be at least one b_i that is big (since sum_i b_i^2 = 1 - |delta|^2).
-    # In summary: sqrt(epsilon) should be the bound on error in eigenvalues,
-    # and sqrt(epsilon) / sigma[k+1] the L2 bound for eigenvectors
+    # More concretely, let m = min_i |lambda - L_i|^2,
+    # so that
+    #  epsilon > \sum_i (lambda - L_i)^2 b_i^2 + lambda^2 |delta|^2
+    #   >= m * \sum_i b_i^2 + lambda^2 |delta|^2
+    #   = m * (1-|delta|^2) + lambda^2 |delta|^2.
+    # Hence, min_i |lambda-L_i|^2 = m < (epsilon - lambda^2  |delta|^2) / (1- |delta|^2).
```
hanbin973, line 844: L841 and 844's epsilon both needs a square. Everything else looks okay.

PR #3114, opened 2025-03-23
Doc syntax correction for embedded backtick

hyanwong:
I think you need to have a `\ ` after a backtick within a string oc characters


benjeffery:
I think they might need to be ` `` ` ?

hyanwong:
Hmm, maybe that's right. I have a load of stuff built-up in my tskit branches, so didn't have time to check this worked. Sorry If I got the syntax wrong: do please adjust as necessary.

benjeffery:
Fixed up!

PR #3115, opened 2025-03-24
Remove svgwrite dep

### LINKED ISSUE #1795: Remove svgwrite as a dependency, do svg xml manipulation directly
It's debateable whether svgwrite is doing much for us at this point, and the API has some limitations in what it allows us to do (see #1289).

Since what we're doing svg-wise is quite simple, it should be possible to replace the svgwrite functionality with a home-grown xml emitting tool fairly simply.

If we're not currently testing that our output is valid SVG (leaning on svgwrite ensuring this) we should add this to our unit tests. This should be a pretty localised change though, and would change the test suite very little.

benjeffery:
Fixes #1795

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2024 Tskit Developers
+# Copyright (c) 2018-2025 Tskit Developers
 # Copyright (c) 2015-2017 University of Oxford
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -30,13 +30,13 @@ import math
 import numbers
 import operator
 import warnings
+import xml
 from dataclasses import dataclass
 from typing import List
 from typing import Mapping
 from typing import Union
 
 import numpy as np
-import svgwrite
 
 import tskit
 import tskit.util as util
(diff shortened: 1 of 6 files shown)


hyanwong:
Woohoo! Seems like this almost worked like magic then?

python/tests/data/svg/ts_y_axis_regular.svg
```diff
@@ -1,172 +1,172 @@
...
+		<style type="text/css">.background path {fill: #808080; fill-opacity: 0}.background path:nth-child(odd) {fill-opacity: .1}.x-regions rect {fill: yellow; stroke: black; opacity: 0.5}.axes {font-size: 14px}.x-axis .tick .lab {font-weight: bold; dominant-baseline: hanging}.axes, .tree {font-size: 14px; text-anchor: middle}.axes line, .edge {stroke: black; fill: none}.axes .ax-skip {stroke-dasharray: 4}.y-axis .grid {stroke: #FAFAFA}.node &gt; .sym {fill: black; stroke: none}.site &gt; .sym {stroke: black}.mut text {fill: red; font-style: italic}.mut.extra text {fill: hotpink}.mut line {fill: none; stroke: none}.mut .sym {fill: none; stroke: red}.mut.extra .sym {stroke: hotpink}.node .mut .sym {stroke-width: 1.5px}.tree text, .tree-sequence text {dominant-baseline: central}.plotbox .lab.lft {text-anchor: end}.plotbox .lab.rgt {text-anchor: start}.polytomy line {stroke: black; stroke-dasharray: 1px, 1px}.polytomy text {paint-order:stroke;stroke-width:0.3em;stroke:white}</style>
```
hyanwong, line 4: It appears that `>` has been wrongly replaced by `&gt;`. We need to make sure the CSS is not XML escaped (this is one of the reasons for putting the CDATA tags in)

python/tests/data/svg/ts_y_axis_regular.svg
```diff
@@ -1,172 +1,172 @@
...
 	</defs>
 	<g class="tree-sequence">
 		<g class="background">
-			<path d="M56.8,0 l184.64,0 l0,138.2 l-129.477,25 l0,5 l-55.1634,0 l0,-5 l0,-25 l0,-138.2z"/>
```
hyanwong, line 8: For some reason, the `rnd` command (or whatever it is that means we don't have too high a degree of precision e.g. near 0) is not working here. The PR replaces `l0,-25` with `l4.26326e-14,-25`:

```
<g class="background">
    <path d="M56.8,0 l184.64,0 l0,138.2 l-129.477,25 l0,5 l-55.1634,0 l0,-5 l4.26326e-14,-25 l0,-138.2z"/>
```

We are also seeing `l0,25` -> l0.0,25 (superfluous `.0`)

python/tskit/drawing.py
```diff
@@ -1285,16 +1423,18 @@ def __init__(
...
+                container,
+                "g",
+                {
+                    "class": subplot.svg_class,
+                    "transform": f"translate({rnd(subplot_x)} {y})",
+                },
             )
-            for svg_items in subplot.root_groups.values():
-                svg_subplot.add(svg_items)
+            for _, svg_items in subplot.root_groups.items():
```
hyanwong, line 1436: This seems a meaningless change. The original was better.

hyanwong:
Mostly seems fine, apart from the XML escaping of the stylesheets, and a few minor niggles.

hyanwong:
There's a workaround for the CDATA issue in the top answer at https://stackoverflow.com/questions/174890/how-to-output-cdata-using-elementtree

That should work for writing to a file, but I'm not sure about writing to a string.

python/tests/data/svg/ts_y_axis_regular.svg
benjeffery: Yeah as I said "Done apart from CDATA handling which seems to be a pain with ElementTree"

python/tskit/drawing.py
benjeffery: Yeah, weird!

python/tests/data/svg/ts_y_axis_regular.svg
benjeffery: I think it is cleaner to subclass `ElementTree` and override `_write`. I'll code it up.

python/tskit/drawing.py
```diff
@@ -679,33 +684,90 @@ def __init__(
...
         self.plotbox = Plotbox(size)
         self.root_groups = {}
         self.svg_class = svg_class
         self.timescaling = None
         self.root_svg_attributes = root_svg_attributes
-        self.dwg_base = dwg.add(dwg.g(class_=svg_class))
-        self.drawing = dwg
+
+        # Create SVG document
+        svg_attrs = {
```
jeromekelleher, line 703: This is quite low-level directly working with SVG. How about we add a new class "SvgWriter" or something that abstracts things a bit, and gives us the flexibility to switch backend?

jeromekelleher (commented):
I'm happy with the general idea of removing svgwrite as it will make things easier to maintain. I would like a less noisy transition though - can we basically implement the subset of the SVGwrite API as a first pass, and see how things go from there?

python/tskit/drawing.py
benjeffery: I've had a quick stab in the last commit, seems promising! Just a few details to fix.

benjeffery:
Ok! That was a great idea @jeromekelleher - this PR now removes `svgwrite` with zero changes to test code or the test comparison data. (Except come weird numbers? maybe a unicode thing)

hyanwong:
What are the weird numbers?

hyanwong:
> What are the weird numbers?

Ah, perhaps that's me using unicode bold numbers in the "polytomy summary" labels. I imagine that we should be allowing UTF-8 though? Indeed that is required, ISTR, if we don't want to specify an `<xml ...>` tag above the `<svg>` tag when writing to a file, so we probably *do* want to make sure that unicode characters work OK.

benjeffery:
Ok I'll check the actual byte codes so that we are byte perfect with the existing weird numbers.

benjeffery:
It was actually the line order, which is now fixed. Waiting for coverage checks then will merge.

jeromekelleher:
Superb. Can we log an issue to track removing this from the conda package too somewhere on upgrade so we don't forget.

benjeffery:
https://github.com/tskit-dev/tskit/issues/3117

PR #3119, opened 2025-03-26
Fix code for numpy 1.23.5

### LINKED ISSUE #3118: New pca code requires newer numpy
In `pyproject.toml` we specify numpy>=1.23.5 but this gives:

```python
  File "/home/benj/projects/tskit/testenv/lib/python3.10/site-packages/tskit/trees.py", line 8983, in _rand_pow_range_finder
    Q = np.linalg.qr(Q).Q
AttributeError: 'tuple' object has no attribute 'Q'
```

Upgrading to 1.25.0 resolves the error so I suggest we update the minimum in pyproject.toml (and the conda build!)

benjeffery:
Fixes #3118

### DIFF
--- a/python/requirements/development.txt
+++ b/python/requirements/development.txt
@@ -28,7 +28,6 @@ pytest
 pytest-cov
 pytest-xdist
 si-prefix
-sphinx==8.2.3
 sphinx-argparse
 sphinx-autodoc-typehints>=1.18.3
 sphinx-issues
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I think it would be good to continue targeting the oldest version of numpy that is compatible with numpy 2, if we can. See alternative suggestion for solving #3118

benjeffery:
Great idea, switched to that.

jeromekelleher (approved):

PR #3126, opened 2025-03-31
Make wheel build on OSX universal and test on both intel and ARM.

benjeffery:


benjeffery:
Lint fail here is spurious - seems to happen if I push to `test` (which is needed for wheel testing) before PR'ing on another branch.

PR #3140, opened 2025-04-14
Freeze metadata

### LINKED ISSUE #993: Metadata can appear to be set directly, but doesn't work: unfriendly to the user
I see that it looks like I can set metadata properties after creating them, but it doesn't actually update the metadata. I think it should either work, or we should raise an error if possible:

```
tables.metadata_schema = tskit.MetadataSchema({"codec":"json"})
tables.metadata = {'foo': {}}
tables.metadata['foo']['bar']=20
ts = tables.tree_sequence()
print(ts.metadata['foo']) # doesn't return 20
```

benjeffery:
Fixes #993

Here's a suggestion for how frozen metadata could work - note that this will have a performance impact as we have to traverse the metadata structure. There is probably a way to avoid the traversal with `object_hook` in JSON and changing the decoder functions in struct. I'll do that if we think this is the right way to go.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,10 +2,17 @@
 [0.6.2] - 2025-04-01
 --------------------
 
+
 **Bugfixes**
 
 - Meatdata.schema was returning a modified schema, this is fixed to return a copy of
   the original schema instead (:user:`benjeffery`, :issue:`3129`, :pr:`3130`)
+
+**Breaking Changes**
+
+- To avoid confusion, metadata returned from `.metadata` accessors is now a `FrozenDict`
+  that does not allow mutation of its contents. This is because mutating the object doesn't
+  update the underlying metadata. (:user:`benjeffery`, :issue:`993`, :pr:`3140`)
 
 --------------------
 [0.6.1] - 2025-03-31
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -2515,3 +2515,186 @@ class TestStructuredArrays:
 
         dtype = metadata.MetadataSchema(schema).numpy_dtype()
         assert dtype.names == ("id", "name", "age")
+
+
+class TestFrozenMetadata:
+    def test_json_simple_frozen(self):
+        schema = metadata.MetadataSchema({"codec": "json"})
+        data = {"a": 1, "b": 2, "c": [1, 2, 3], "d": {"x": 10, "y": 20}}
+        encoded = schema.validate_and_encode_row(data)
+        decoded = schema.decode_row(encoded)
+
+        # Check that we can't modify the top-level dict
+        with pytest.raises(TypeError):
+            decoded["e"] = 5
+
+        with pytest.raises(TypeError):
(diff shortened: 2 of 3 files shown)


petrelharp:
Just thinking this through. Currently the nice way to modify metadata is like
```
md = tables.metadata
md['foo'] = 123
tables.metadata = md
```
IIUC, with this change, this won't work - it'll need an additional thing to convert `md` to a regular Dict? Just an observation, not sure what to do about it.

I guess we should also measure the performance implications.

jeromekelleher:
Hmm. I leaning towards marking #993 as a wontfix and living with it as a known wart. Changes like this *will* break code, probably quite a lot of it. As it's not fixing an outright bug, and more an attempt to prevent people from making mistakes (or helping them find their mistakes quicker) it's not obvious that the good outweighs the harm here. For example, I can see a lot of older code getting broken by the change, resulting in people pinning to older versions of tskit rather than bothering with a fix.

jeromekelleher:
(Thanks for working it up @benjeffery, it's really helpful to have a concrete proposal to discuss!)

jeromekelleher:
A large and loud warning about this behaviour at the top of the metadata section, perhaps along with a warning on every accessor function (via an rst macro thingy) would be a better solution all round at this point I think)

hyanwong:
What about a user warning if you deliberately attempt to set metadata directly? Or is that too complicated?

benjeffery:
> Hmm. I'm leaning towards marking #993 as a wontfix and living with it as a known wart.

I completely agree, having worked it up and seeing how much unrelated test breakage we had. I'd be happy with some prominent docs warnings. 

I also don't think the behaviour is too unexpected, but maybe I'm too close to the code to simulate what it is like for an outsider.

petrelharp:
> I also don't think the behaviour is too unexpected, but maybe I'm too close to the code to simulate what it is like for an outsider.

I think it's pretty mystifying that assigning does nothing; however, people will figure this out reasonably quickly (and actually modifying metadata is a semi-advanced thing). But given there doesn't seem to be a reasonable way to fix it, a bigger issue IMO is that the way to actually modify metadata is relatively obscure. I'd propose we provide an example of modifying existing (top-level and per-table-entry) metadata (with a header, so it's easily findable) in [the docs](https://tskit.dev/tskit/docs/latest/metadata.html).

hyanwong:
Because it's specifically a python thing, it's documented at https://tskit.dev/tutorials/metadata.html#modifying-metadata-and-schemas, I think.

petrelharp:
> Because it's specifically a python thing, it's documented at https://tskit.dev/tutorials/metadata.html#modifying-metadata-and-schemas, I think.

Ah, right - and, there's a link up top from [the other Metadata docs](https://tskit.dev/tskit/docs/latest/metadata.html). Still, it'd be good to have not just "see here" but maybe a warning that in-place modification wont' work. (Also, those docs don't have the simplest use cases of "change an existing value in top-level metadata" or "change/add a value in each row".)

benjeffery:
Closing for #3156

PR #3145, opened 2025-04-26
import xml.dom.minidom; closes #3144

### LINKED ISSUE #3144: `ts.draw_svg(path=...)` fails without `import xml.dom`
The problem: create a new venv (or whatevs) and install (only) `tskit`; then run
```
import tskit
ts = tskit.Tree.generate_star(5).tree_sequence
ts.draw_svg(path='test.svg')
```
This results in
```
  File "<python-input-9>", line 1, in <module>
    ts.draw_svg(path='test.svg')
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/peter/debug/tskit/plot_xml/venv/lib/python3.13/site-packages/tskit/trees.py", line 7500, in draw_svg
    draw.drawing.saveas(path, pretty=True)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/home/peter/debug/tskit/plot_xml/venv/lib/python3.13/site-packages/tskit/drawing.py", line 245, in saveas
    f.write(self.tostring(pretty=pretty))
            ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/peter/debug/tskit/plot_xml/venv/lib/python3.13/site-packages/tskit/drawing.py", line 240, in tostring
    return xml.dom.minidom.parseString(self.root.tostring()).toprettyxml()
           ^^^^^^^
AttributeError: module 'xml' has no attribute 'dom'
```
IIUC,  `xml.dom` is a submodule, so doing `import xml` does not suffice to let us use `xml.dom`; so we need to do `import xml.dom`? But, doing `import xml.dom` and then the draw_svg call results in 
```
AttributeError: module 'xml.dom' has no attribute 'minidom'
```
and so we need to do `import xml.dom.minidom`. Indeed, this is not caught by the tests, because that line appears in `tests/test_drawing.py`.

petrelharp:
Simple fix (but, I don't know how to change the tests to have caught this...)

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -4,8 +4,11 @@
 
 **Bugfixes**
 
-- Meatdata.schema was returning a modified schema, this is fixed to return a copy of
+- Metadata.schema was returning a modified schema, this is fixed to return a copy of
   the original schema instead (:user:`benjeffery`, :issue:`3129`, :pr:`3130`)
+
+- ``TreeSequence.draw_svg(path=...)`` was failing due to a missing
+  `import xml.dom.minidom` (:user:`petrelharp`, :issue:`3144`, :pr:`3145`)
 
 --------------------
 [0.6.1] - 2025-03-31
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -30,7 +30,7 @@ import math
 import numbers
 import operator
 import warnings
-import xml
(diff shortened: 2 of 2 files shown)


benjeffery (approved):
Thanks for this @petrelharp - will see if the tests can be fixed separately.

PR #3146, opened 2025-04-28
Add test for minidom

benjeffery:
Test to ensure #3144 is not repeated.


jeromekelleher:
This is such a weird once off that I'm not sure it's worth stressing about? Great if we can make this test work easily, but let's not worry about it if not.

benjeffery:
Theres a few pushes here as I wanted to check the test failed without the new fix, should be ready to go now.

jeromekelleher (approved):

PR #3153, opened 2025-05-03
Add individual nodes method

benjeffery:
Adds a method to get an array of (individuals, ploidy) node ids.

I considered adding in the case where there are no individuals and the user specifies ploidy instead for all the sample nodes, but that seems like a separate method.

I'm also not 100% sure it is an error if a non-sample node is in an individual.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,4 +1,14 @@
 --------------------
+[0.6.4] - 2025-XX-XX
+--------------------
+
+**Features**
+
+- Add ``TreeSequence.individuals_nodes`` attribute to return the nodes
+  associated with each individual as a numpy array.
+  (:user:`benjeffery`, :pr:`3153`)
+
+--------------------
 [0.6.3] - 2025-04-28
 --------------------
 
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -223,7 +223,8 @@ handle_library_error(int err)
     const char *not_kas_format_msg
         = "File not in kastore format. Either the file is corrupt or it is not a "
           "tskit tree sequence file. It may be a legacy HDF file upgradable with "
-          "`tskit upgrade` from tskit version<0.6.2 or a compressed tree sequence file that can be decompressed "
+          "`tskit upgrade` from tskit version<0.6.2 or a compressed tree sequence file "
+          "that can be decompressed "
           "with `tszip`.";
     const char *ibd_pairs_not_stored_msg
         = "Sample pairs are not stored by default "
@@ -9715,12 +9716,12 @@ TreeSequence_weighted_stat_vector_method(
     if (result_array == NULL) {
         goto out;
     }
-	Py_BEGIN_ALLOW_THREADS
-    err = method(self->tree_sequence, w_shape[1], PyArray_DATA(weights_array),
-        num_windows, PyArray_DATA(windows_array), num_focal_nodes,
-        PyArray_DATA(focal_nodes_array), PyArray_DATA(result_array), options);
-	Py_END_ALLOW_THREADS
-    if (err != 0) {
+    Py_BEGIN_ALLOW_THREADS err
+        = method(self->tree_sequence, w_shape[1], PyArray_DATA(weights_array),
+            num_windows, PyArray_DATA(windows_array), num_focal_nodes,
+            PyArray_DATA(focal_nodes_array), PyArray_DATA(result_array), options);
+    Py_END_ALLOW_THREADS if (err != 0)
+    {
         handle_library_error(err);
         goto out;
     }
@@ -10744,7 +10745,8 @@ TreeSequence_get_individuals_metadata(TreeSequence *self, void *closure)
         goto out;
     }
     individuals = self->tree_sequence->tables->individuals;
-    ret = TreeSequence_make_array(self, individuals.metadata_length, NPY_UINT8, individuals.metadata);
+    ret = TreeSequence_make_array(
+        self, individuals.metadata_length, NPY_UINT8, individuals.metadata);
 out:
     return ret;
 }
@@ -10835,7 +10837,8 @@ TreeSequence_get_nodes_metadata(TreeSequence *self, void *closure)
         goto out;
     }
     nodes = self->tree_sequence->tables->nodes;
-    ret = TreeSequence_make_array(self, nodes.metadata_length, NPY_UINT8, nodes.metadata);
+    ret = TreeSequence_make_array(
+        self, nodes.metadata_length, NPY_UINT8, nodes.metadata);
 out:
     return ret;
 }
@@ -10926,7 +10929,8 @@ TreeSequence_get_edges_metadata(TreeSequence *self, void *closure)
         goto out;
     }
     edges = self->tree_sequence->tables->edges;
-    ret = TreeSequence_make_array(self, edges.metadata_length, NPY_UINT8, edges.metadata);
+    ret = TreeSequence_make_array(
+        self, edges.metadata_length, NPY_UINT8, edges.metadata);
 out:
     return ret;
 }
@@ -10941,7 +10945,8 @@ TreeSequence_get_edges_metadata_offset(TreeSequence *self, void *closure)
         goto out;
     }
     edges = self->tree_sequence->tables->edges;
-    ret = TreeSequence_make_array(self, edges.num_rows + 1, NPY_UINT64, edges.metadata_offset);
+    ret = TreeSequence_make_array(
+        self, edges.num_rows + 1, NPY_UINT64, edges.metadata_offset);
 out:
     return ret;
 }
@@ -10971,7 +10976,8 @@ TreeSequence_get_sites_metadata(TreeSequence *self, void *closure)
         goto out;
     }
     sites = self->tree_sequence->tables->sites;
-    ret = TreeSequence_make_array(self, sites.metadata_length, NPY_UINT8, sites.metadata);
+    ret = TreeSequence_make_array(
+        self, sites.metadata_length, NPY_UINT8, sites.metadata);
 out:
     return ret;
 }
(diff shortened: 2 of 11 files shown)


python/tskit/trees.py
```diff
@@ -10476,6 +10476,45 @@ def ld_matrix(
...
+            max_ploidy = max(max_ploidy, len(ind.nodes))
+        if max_ploidy == 0:
+            raise ValueError("No nodes refer to individuals")
+
+        # Initialize output array with -1 (indicating no node)
+        result = np.full((self.num_individuals, max_ploidy), -1, dtype=np.int32)
+
+        for i in range(self.num_individuals):
+            ind = self.individual(i)
+            if len(ind.nodes) == 0:
```
jeromekelleher, line 10506: I don't think we should check for sample status here, we're just returning the actual nodes. There's no reason that an individual couldn't have mixed sample status or not be a sample at all I think.

python/tskit/trees.py
```diff
@@ -10476,6 +10476,45 @@ def ld_matrix(
             mode=mode,
         )
 
+    def individual_nodes(self):
```
jeromekelleher, line 10479: For backwards compatibility, we should provide a ``ploidy`` argument here I think which kicks in for the no-individuals case (and it's an error is this is not None when num_individuals > 0). Here, we do what the VCF code is currently doing.

jeromekelleher (commented):
LGTM - I think we should be less prescriptive though. A lot of the old insisting it was a sample idea came from the ``variants`` code, but we can support non-samples in variants now perfectly well, so there's no real reason to disallow it.

jeromekelleher:
> I considered adding in the case where there are no individuals and the user specifies ploidy instead for all the sample nodes, but that seems like a separate method.

Ah, I see, I didn't spot this. What's the interface we provide for ``bio2zarr`` then? Does it do:
```
if ts.num_individuals == 0:
    individual_nodes = ts.individuals_nodes()
else:
    individual_nodes = ts.split_samples_by_ploidy(ploidy)
```

It would be nice to have a method that does this in tskit that does this?

If we take this approach, then I think we could make the ``individuals_nodes`` an attribute, as it's a natural extension to the tabular data model.

petrelharp:
> I'm also not 100% sure it is an error if a non-sample node is in an individual.

Non-sample nodes can definitely be in individuals (from SLiM for instance).

python/tskit/trees.py
petrelharp: Really? I'm probably missing something, but that is long-ago-enough that I think it would only confuse things?

python/tskit/trees.py
```diff
@@ -10476,6 +10476,45 @@ def ld_matrix(
...
+            Values of -1 indicate unused slots for individuals with ploidy
+            less than the maximum.
+        :rtype: numpy.ndarray (dtype=np.int32)
+        """
+        if self.num_individuals == 0:
+            raise ValueError("Tree sequence has no individuals")
+
+        max_ploidy = 0
+        for i in range(self.num_individuals):
+            ind = self.individual(i)
```
petrelharp, line 10496: Just curious - why not
```suggestion
        for ind in self.individuals():
```

python/tskit/trees.py
petrelharp: Agree - that'd be weird, but we can think of use cases - what if we have sequence data for a parent and only sparse data for a child, such that we can impute one of the sets of chromosomes for the child but not the other.

jeromekelleher:
I'm leaning towards this being an array that we create on demand (in the Python C layer) like the other derived arrays. It's such a fundamental object and I think that ultimately the array representation is the most useful anyway (rather than adding lots of search-style parameters to a method). Any reason not to do this do we think?

python/tskit/trees.py
jeromekelleher: Yes, I've changed my mind on this - see point below about making ``ts.individuals_nodes`` an attribute array.

benjeffery:
> I'm leaning towards this being an array that we create on demand (in the Python C layer)

Yes, agreed - let's get the implementation agreed in python and then I'll convert.

I've just pushed a commit that removes the errors related to the sample flags. It also removes the error if there is one or more individuals that have no nodes. I've left the error if there are no individuals - but this could just return an empty array, which I'm leaning towards.

benjeffery:
I've added the commit to return an empty array in the case of no individuals as that feels more correct.

python/tests/test_highlevel.py
```diff
@@ -5491,3 +5491,91 @@ def test_error_if_no_schema(self, table_name):
...
+        tables.individuals.add_row(flags=0, location=(0, 0), metadata=b"")
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0, individual=0)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0, individual=0)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0, individual=1)
+        tables.nodes.add_row(flags=tskit.NODE_IS_SAMPLE, time=0, individual=1)
+        ts = tables.tree_sequence()
+
+        result = ts.individual_nodes
+        assert result.shape == (2, 2)
+        assert np.array_equal(result, [[0, 1], [2, 3]])
```
jeromekelleher, line 5510: may as well use ``nt.assert_array_equal``

python/tskit/trees.py
```diff
@@ -10476,6 +10476,30 @@ def ld_matrix(
...
+        """
+        Return an array of node IDs for each individual in the tree sequence.
+        Errors if the tree sequence has no individuals.
+
+        :return: Array of shape (num_individuals, max_ploidy) containing node IDs.
+            Values of -1 indicate unused slots for individuals with ploidy
+            less than the maximum.
+        :rtype: numpy.ndarray (dtype=np.int32)
+        """
+        if self.num_individuals == 0:
```
jeromekelleher, line 10490: I agree, this isn't an error, and should just return an empty array (ideally without a special case for C code)

jeromekelleher (commented):
LGTM, and let's get of the zero individuals error also

python/tests/test_highlevel.py
benjeffery: Done in 3873bcbd

python/tskit/trees.py
benjeffery: Done in 4a815ba6

benjeffery:
I'll proceed with the CPython - look like it is just a conversion of the existing ragged array.

jeromekelleher:
Great, should be straightforward hopefully.

python/CHANGELOG.rst
```diff
@@ -1,3 +1,12 @@
+--------------------
+[0.6.4] - 2025-XX-XX
+--------------------
+
+**Features**
+- Add ``TreeSequence.individual_nodes`` method to return the nodes
```
jeromekelleher, line 7:
```suggestion
- Add ``TreeSequence.individuals_nodes`` attribute to return the nodes
```

python/_tskitmodule.c
```diff
@@ -8724,6 +8724,61 @@ TreeSequence_get_individuals_time(TreeSequence *self)
...
+        goto out;
+    }
+
+    num_individuals = tsk_treeseq_get_num_individuals(self->tree_sequence);
+    
+    individual_nodes = self->tree_sequence->individual_nodes;
+    individual_nodes_length = self->tree_sequence->individual_nodes_length;
+    
+    for (tsk_id_t i = 0; i < (tsk_id_t)num_individuals; i++) {
+        tsk_size_t ploidy = individual_nodes_length[i];
```
jeromekelleher, line 8750: variable definitions in the first block by convention

python/_tskitmodule.c
```diff
@@ -8724,6 +8724,61 @@ TreeSequence_get_individuals_time(TreeSequence *self)
...
+TreeSequence_get_individuals_nodes(TreeSequence *self)
+{
+    PyObject *ret = NULL;
+    PyArrayObject *ret_array = NULL;
+    npy_intp dims[2];
+    tsk_size_t num_individuals;
+    tsk_size_t max_ploidy = 0;
+    tsk_id_t *node_mem;
+    tsk_size_t j;
+    tsk_id_t **individual_nodes;
```
jeromekelleher, line 8737: These should be const I think

jeromekelleher (commented):
LGTM, a few C tweaks needed.

Is the C clang-formatted?

benjeffery:
Comments addressed - seems my local clang broke, fixed now.

python/tskit/trees.py
petrelharp: okay phew thought I was missing something =)

jeromekelleher (commented):
Still a typo in changelog

jeromekelleher:
I think this is ready to merge, we just have to fix up the clang-formatting overshoot (probably wrong version?)

benjeffery:
Have a segfault on CI that I'm struggling to recreate locally. Will debug on CI if I can't.

Not sure about the clang overshoot - I'm running version 6.0. Will patch up by hand if I have to.

jeromekelleher:
> Not sure about the clang overshoot - I'm running version 6.0. Will patch up by hand if I have to.

If CI is happy I'm happy, quite possible that non-clang formatted stuff crept in an some point.

python/_tskitmodule.c
```diff
@@ -8724,6 +8725,59 @@ TreeSequence_get_individuals_time(TreeSequence *self)
...
+TreeSequence_get_individuals_nodes(TreeSequence *self)
+{
+    PyObject *ret = NULL;
+    PyArrayObject *ret_array = NULL;
+    npy_intp dims[2];
+    tsk_size_t ploidy;
+    tsk_size_t max_ploidy = 0;
+    tsk_id_t *node_mem;
+    tsk_size_t j;
+    const tsk_size_t num_individuals
```
jeromekelleher, line 8738: Here's your problem - you're calling tree sequence methods before ``TreeSequence_check_state``

jeromekelleher:
The segfault should be a simple, see above. CI working!

python/_tskitmodule.c
benjeffery: *facepalm* Thanks.

benjeffery:
Now stacked on #3155 which fixes the linting.

jeromekelleher (approved):

Issue #3154, opened 2025-05-06
ts.num_sites, ts.num_mutations, and ts.segregating_sites() give different values


milesandersonmn:
Hi, everyone

I noticed that when getting the number of segregating sites using the `segregating_sites()` function, the value is the same as the `num_mutations` property but higher than the `num_sites` property. I suppose this would imply that the segregating sites function is actually including sites that are fixed. I can try to modify my scripts to work around this, but just thought I would bring this to your attention.

molpopgen:
The result is also possible when some sites have > 1 mutated state.  This can occur in situations like ancestral state + 2 derived states are present, no ancestral + 3 derived states, etc., at a single site.

gregorgorjanc:
I guess the name `segregating_sites()` is maybe confusing. Looking at the docs at 

https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.segregating_sites

we have 

```
What is computed depends on mode. For a sample set A, computes:

“site”
The sum over sites of the number of alleles found in A at each site minus one, per unit of chromosome length. If all sites have at most two alleles, this is the density of sites that are polymorphic in A. To get the number of segregating minor alleles per window, pass span_normalise=False.
```

So, you are getting the number of mutations, which can be larger than the number of sites (loci/places/...) where we see mutations  (=are polymorphic, are segregating).

petrelharp:
What's being computed is (almost) "number of distinct non-ancestral alleles seen"; so, what you say could happen if some mutations are not seen in the samples, or if the same allele gets produced by more than one mutation. (I said "almost" because if actually "number of alleles minus one"; you can think of the "minus one" as removing the ancestral allele, but if the ancestral allele is no longer seen in the samples, then it's some other one that doesn't get counted). See [the msprime docs](https://tskit.dev/msprime/docs/stable/mutations.html#silent-mutations) for some more on this.
I feel like I've written more about this somewhere; googling I found [this question](https://bioinformatics.stackexchange.com/questions/22053/calculating-the-segregating-sites-via-tskit-versus-counting/23397#23397).

Perhaps a clarifying sentence or two in the docs would help; suggestions welcome?

petrelharp:
Closed in #3158

PR #3158, opened 2025-05-07
segregating_sites doc clarification

gregorgorjanc:
A proposal of improved doc for `segregating_sites()` based on discussion in https://github.com/tskit-dev/tskit/issues/3154

@petrelharp how does this look?

# PR Checklist:

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -9349,11 +9349,17 @@ class TreeSequence:
         What is computed depends on ``mode``. For a sample set ``A``, computes:
 
         "site"
-            The sum over sites of the number of alleles found in ``A`` at each site
-            minus one, per unit of chromosome length.
-            If all sites have at most two alleles, this is the density of sites
-            that are polymorphic in ``A``. To get the **number** of segregating minor
-            alleles per window, pass ``span_normalise=False``.
+            The sum over sites of [the number of alleles found in ``A`` at each site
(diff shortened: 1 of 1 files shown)


python/tskit/trees.py
```diff
@@ -9349,11 +9349,17 @@ def segregating_sites(
...
-            alleles per window, pass ``span_normalise=False``.
+            The sum over sites of [the number of alleles found in ``A`` at each site
+            minus one], per unit of chromosome length.
+            If all sites have at most two alleles in ``A``,
+            this is the density of segregating/polymorphic sites in ``A``
+            (since the "minus one" reduces the sum for ancestral allele).
+            Hence, multiple mutations per site increase the
+            returned value, but only if they segregate in ``A``.
+            If an ancestral allele does not segregate in ``A``,
+            the "minus one" will reduce the sum for another allele.
```
petrelharp, line 9360:
```suggestion
            If all sites have at most two alleles in ``A``,
            this is the density of segregating/polymorphic sites in ``A``
            (since the "minus one" reduces the sum for monoallelic sites).
            For sites with more than two alleles, the sum is increased by
            one for each additional allele segregating in ``A``.
```
How about this?

gregorgorjanc: I like it!

petrelharp (approved):
Thanks!!!

PR #3163, opened 2025-05-15
Map samples function

benjeffery:
@jeromekelleher this follows our discussion for the default case. For the case where the caller has specified individuals, only fatal errors are triggered.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -17,12 +17,27 @@
   so a set of tree sequence can be added to the right of an existing one.
   (:user:`hyanwong`, :pr:`3165`, :issue:`3164`)
 
+- Add ``TreeSequence.map_to_vcf_model`` method to return a mapping of
+  the tree sequence to the VCF model. 
+  (:user:`benjeffery`, :pr:`3163`)
 
 **Fixes**
 
 - Correct assertion message when tables are compared with metadata ignored.
   (:user:`benjeffery`, :pr:`3162`, :issue:`3161`)
   
+**Breaking changes** 
+
+- ``TreeSequence.write_vcf`` now filters non-sample nodes from individuals
+  by default, instead of raising an error. These nodes can be included using the
+  new ``include_non_sample_nodes`` argument. 
+  By default individual names (sample IDs) in VCF output are now of the form
+  ``tsk_{individual.id}`` Previously these were always 
+  ``"tsk_{j}" for j in range(num_individuals)``. This may break some downstream
+  code if individuals are specified. To fix, manually specify ``individual_names``
+  to the required pattern.
+  (:user:`benjeffery`, :pr:`3163`)
+
 --------------------
 [0.6.3] - 2025-04-28
 --------------------
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -59,6 +59,12 @@ from tskit import UNKNOWN_TIME
 LEGACY_MS_LABELS = "legacy_ms"
 
 
+@dataclass
+class VcfModelMapping:
+    individuals_nodes: np.ndarray
+    individuals_name: np.ndarray
+
+
 class CoalescenceRecord(NamedTuple):
     left: float
     right: float
@@ -6412,6 +6418,7 @@ class TreeSequence:
         sample_mask=None,
         isolated_as_missing=None,
         allow_position_zero=None,
+        include_non_sample_nodes=None,
     ):
         """
         Convert the genetic variation data in this tree sequence to Variant
@@ -6427,21 +6434,21 @@ class TreeSequence:
         :ref:`sec_export_vcf_constructing_gt` section for more details
         and examples.
 
-        If individuals that are associated with sample nodes are defined in the
+        If individuals are defined in the
         data model (see :ref:`sec_individual_table_definition`), the genotypes
-        for each of the individual's samples are combined into a phased
-        multiploid values at each site. By default, all individuals associated
-        with sample nodes are included in increasing order of individual ID.
+        for each of the individual's nodes are combined into a phased
+        multiploid values at each site. By default, all individuals are
+        included with their sample nodes, individuals with no nodes are
+        omitted. The ``include_non_sample_nodes`` argument can be used to
+        included non-sample nodes in the output VCF.
 
         Subsets or permutations of the sample individuals may be specified
-        using the ``individuals`` argument. It is an error to specify any
-        individuals that are not associated with any nodes, or whose
-        nodes are not all samples.
+        using the ``individuals`` argument.
 
         Mixed-sample individuals (e.g., those associated with one node
         that is a sample and another that is not) in the data model will
-        result in an error by default. However, such individuals can be
-        excluded using the ``individuals`` argument.
+        only have the sample nodes output by default. However, non-sample
+        nodes can be included using the ``include_non_sample_nodes`` argument.
 
         If there are no individuals in the tree sequence,
         synthetic individuals are created by combining adjacent samples, and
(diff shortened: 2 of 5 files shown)


python/tskit/trees.py
```diff
@@ -10520,6 +10520,56 @@ def sample_nodes_by_ploidy(self, ploidy):
         sample_node_ids = sample_node_ids.reshape((num_samples_per_individual, ploidy))
         return sample_node_ids
 
+    def map_samples_to_vcf(self, individuals=None):
+        """
+        Returns a list of lists of node IDs, where each sublist contains the
+        sample nodes associated with the same individual.
+        If `individuals` is not specified (the default) nodes for all individuals where
+        all of their nodes are samples are returned. If `individuals` is specified,
+        only the nodes for the specified individuals are returned.
```
jeromekelleher, line 10529: This looks about right. We also want to return the individual names though, so I'd imagined it would be something like
```
@dataclasses.dataclass
class VcfModelMapping:
    individuals_nodes: np.ndarray
    individuals_name: np.ndarray

def map_to_vcf_model(self, individuals=None, ploidy=None):
     """
     Returns a mapping of the tskit data model to VCF "samples".
...
    return VcfModelMapping(individuals_names, individuals_names)
```

jeromekelleher (commented):
Looks like the right logic, but I think we probably want to be explicit about what we're doing and provide the full mapping into the VCF data model in one function. Ploidy is annoying, but we'd just end up doing something awkward in the VCF code to support, I think. At least this way the full model mapping is one in one place.

benjeffery:
I've added the ploidy as suggested.

Where do you see the individual names coming from? There's no standard in metadata - so I assume there will need to be an argument to specify where it is coming from?

jeromekelleher:
By default we'll just generate ``tsk_0`` etc, but we could have a ``name_key`` or something argument that we'd use to extract from metadata.

How would we use this to round-trip VCF samples IDs from tsinfer? That's got to be a thing we can do.

benjeffery:
I've added new separate tests - once we're happy I will integrate with existing VCF code and update those tests and docs.

jeromekelleher (commented):
Yep, I think this is it.

jeromekelleher:
I've done a deep-dive on this over in #3166, good to get your thoughts @benjeffery

benjeffery:
Added docs and changelog

jeromekelleher:
What do you think of the idea about a ``samples_nodes_only=True`` parameter here @benjeffery? That would help to avoid the guesswork about when to include individuals nodes or not? So, the semantics now becomes something like this:

```
if individuals is None:
      individuals = np.arange(ts.num_individuals)
individual_nodes = []
for ind_id in individuals:
      ind = ts.individuals()
      nodes = []
      for node in ind.nodes:
            if sample_nodes_only:
                 if ts.is_sample(node):
                     nodes.append(node)
            else:
                     nodes.append(node)

individual_nodes = np.array(individual_nodes)
```
Note that this does allow for zero-ploid individuals, but I think that's actually fine and can make sense sometimes. This will work fine in VCF Zarr. 

In the tskit VCF code we can raise an error if there's any zero-ploid individuals, because outputting zero ploid VCF is probably an error.

Does this make sense? Sorry for all the back and forth, this is a very tricky issue and it'll save a lot of pain in the long run if we can get it right now.

benjeffery:
I've added a commit here with the logic that we talked about - I've used `include_non_sample_nodes` as the name of the parameter instead of `sample_nodes_only` as I think turning this "on" should feel like you are doing something unusual.

The code is so much simpler that this feels like the right move. There is one question I'm not sure of.

If an individual has nodes `[1,2]` but 1 is not a sample, should this function return `[-1,2]` or `[2,-1]`. It feels like the first is more "correct", but relies on callers to filter sensibly and not stop at the first `-1`.

jeromekelleher:
> If an individual has nodes [1,2] but 1 is not a sample, should this function return [-1,2] or [2,-1]. It feels like the first is more "correct", but relies on callers to filter sensibly and not stop at the first -1.

Good question. I guess it comes down to whether we regard the ordering of "ploids" in VCF as significant. Given that we're always going to put dimension padding at the end, I guess we should return [2, -1] in this case (since we're explicitly mapping into the VCF model).

python/tskit/trees.py
```diff
@@ -10520,6 +10527,112 @@ def sample_nodes_by_ploidy(self, ploidy):
...
+        else:
+            if individuals is None:
+                individuals = np.arange(self.num_individuals, dtype=np.int32)
+            if min(individuals) < 0 or max(individuals) >= self.num_individuals:
+                raise ValueError("Invalid individual ID")
+
+            individuals_nodes = self.individuals_nodes[individuals]
+            if not include_non_sample_nodes:
+                individuals_nodes[
+                    np.logical_not(
```
jeromekelleher, line 10613: This is a bit of a corner case, so I think we can just do a loop to make our lives a bit easier:
```
non_sample_nodes =  np.logical_not(self.nodes_flags[individuals_nodes] & tskit.NODE_IS_SAMPLE)
if np.any(non_sample_nodes) and not include_non_sample_nodes:
    individuals_nodes[non_sample_nodes] = -1
    for index in np.where(non_sample_nodes):
        # do the switcharoo to make sure -1 is at the end
```

jeromekelleher (commented):
Great, this is it - let's finish it up and ship it!

python/CHANGELOG.rst
```diff
@@ -12,12 +12,22 @@
...
 
 - Correct assertion message when tables are compared with metadata ignored.
   (:user:`benjeffery`, :pr:`3162`, :issue:`3161`)
   
+**Changes** 
+
+- ``TreeSequence.write_vcf`` now filters non-sample nodes from individuals
+  by default, instead of raising an error. These nodes can be included using the
+  new ``include_non_sample_nodes`` argument.
+  (:user:`benjeffery`, :pr:`3163`)
```
jeromekelleher, line 29: We need to add a breaking change here: "By default individual names are now of the form "tsk_{individual.id}", whereas previously names (i.e., sample IDs in the VCF were always "``"tsk_{j}" for j in range(num_individuals)``. This may break some downstream code if individuals are specified. To fix, manually specify ``individual_names`` to the required pattern"

python/tskit/trees.py
```diff
@@ -10520,6 +10532,129 @@ def sample_nodes_by_ploidy(self, ploidy):
...
+        Maps the sample nodes in this tree sequence to a representation suitable for
+        VCF output, using the individuals if present.
+
+        Creates a VcfModelMapping object that contains both the nodes-to-individual
+        mapping as a 2D array of (individuals, nodes) and the individual names. The
+        mapping is created by first checking if the tree sequence contains individuals.
+        If it does, the mapping is created using the individuals in the tree sequence.
+        Be default only the sample nodes of the individuals are included in the mapping,
+        unless `include_non_sample_nodes` is set to True, in which case all nodes
+        belonging to the individuals are included. Any individuals without any nodes
```
jeromekelleher, line 10553: No - I think these should be zero ploid, right?

python/tskit/trees.py
```diff
@@ -10520,6 +10532,129 @@ def sample_nodes_by_ploidy(self, ploidy):
...
+        are ignored and not included in the mapping.
+        If no individuals are present, the mapping is created using only the sample nodes
+        and the specified ploidy.
+
+        If neither `name_metadata_key` nor `individual_names` is not specified, the
+        individual names are set to "tsk_{individual_id}" for each individual.
```
jeromekelleher, line 10559: Document the naming convention for the no-individuals case above.

jeromekelleher (commented):
Looks good,  I think we're nearly there. Just two things left I think:

1. Documenting the new naming method and flagging as a potentially breaking change
2. Nailing down the zero-ploid individual corner case. We need to make sure that we are passing along individuals that have no nodes (or no sample nodes) without filtering them out (as the docs state currently). Good to test this explicitly, if not already done.

python/CHANGELOG.rst
benjeffery: Good point, fixed in b5de86fa

python/tskit/trees.py
benjeffery: Fixed in 5e3e4011

jeromekelleher (approved):
Great, let's merge

PR #3165, opened 2025-05-15
Concatenate and shift functions

### LINKED ISSUE #3164: New concatenate function to join tree sequences
A number of people (including me) have been trying to join two tree sequences together, end to end, e.g. when obtained from different chromosomes. I have messed this up (message in slack, I forgot that `keep_intervals` can also change node IDs because it automatically simplifies), and I suspect others find this tricky too.

I propose two new main TableCollection methods, `shift` and `concatenate` (based on the numpy function, which concatenates multiple things together), and possibly an additional shortcut method, provisionally called `append`, which simply concatenates two TableCollections. These can be exposed in the normal way as TreeSequence methods.

As this essentially only relies on `union`, I think most of the behavioural edge cases & errors should be caught in that function. The helpful addition here is to (a) shift the coordinates of the other TC to fit and (b) by default, only map the samples together.

```python
def shift(self, value=None, sequence_length=None, record_provenance=True):
    """
    Shift the coordinate system (used by edges and sites) of this TableCollection by a given
    value. Positive values shift the coordinate system to the right, negative values to the left.
    The sequence length of the tree sequence will be changed by `value`, unless
    `sequence_length` is given, in which case the sequence length will be set to that value.

    .. note::
        No attempt is made to check that the new coordinate system or sequence length is valid:
        if you wish to do this, use the equivalent TreeSequence.shift() method instead.
    """
    if value is None:
        value = 0
    if value != 0:
        self.edges.left += value
        self.edges.right += value
        self.sites.position += value
    if sequence_length is None:
        self.sequence_length += value
    else:
        self.sequence_length = sequence_length
    
    if record_provenance:
        parameters = {
            "command": "shift", "value": value, "sequence_length": sequence_length
        }
        self.provenances.add_row(
        )


def concatenate(self, others, *, node_mappings=None, record_provenance=True, **kwargs):
    """
    Concatenate a set of table collections, with a comparable set of samples, to the
    right of the existing table collection. This shifts the other tables
    rightwards then repeatedly calls {meth}`union`. By default, the sample nodes
    in the other node table, in numerical order, are mapped to the sample nodes in the
    existing node table, in numerical order.

    .. note::
        To add gaps between the concatenated tables, use :meth:`shift` or
        to remove gaps, use :meth:`trim` before concatenating.

    :param TableCollection other: A list of additional table collections.
    :param list node_mappings: A list of node mappings, of the same length as `others`.
        Each node mapping is an array of integers
        of the same length as the number of nodes in the equivalent table collection
        in `others`, where the k'th element gives the id of the node in the current table
        collection corresponding to node k in the other table collection (see {meth}`union`).
        If any node mapping is None, only the sample nodes between the two nodes
        tables are mapped to each other, with samples taken in numerical order.
        If None (default), ``node_mappings`` is treated as a list of None values.
    :param bool record_provenance: If True (default), record details of this call to
            ``concatenate`` in the returned tree sequence's provenance information
            (Default: True).
    :param kwargs: Additional keyword arguments to pass to {meth}`union`.


    returns: a new tree sequence
    """
    if node_mappings is None:
        node_mappings = [None] * len(others)
    if len(node_mappings) != len(others):
        raise ValueError("`node_mappings` must have the same length as `others`")
    samples = np.where(self.nodes.flags & tskit.NODE_IS_SAMPLE)[0]
    for other, node_mapping in zip(others, node_mappings):
        other_samples = np.where(other.nodes.flags & tskit.NODE_IS_SAMPLE)[0]
        if len(other_samples) != len(samples):
            raise ValueError("each `other` must have the same number of samples as `self`")
        other.shift(self.sequence_length, record_provenance=False)
        self.sequence_length += other.sequence_length
        if node_mapping is None:
            node_mapping = np.full(other.nodes.num_rows, tskit.NULL, dtype=np.int32)
            node_mapping[other_samples] = samples
        self.union(other, node_mapping=node_mapping, **kwargs)
    if record_provenance:
        parameters = {
            "command": "concatenate",
            "TODO": "add concatenate parameters"  # tricky because each other TC has provenances
        }
        self.provenances.add_row(
        )

def append(self, other, node_mapping=None, **kwargs, record_provenance=True):
    """
    Append another table collections to the right of the existing table collection.
    This is a shortcut for ``.concatenate([other], sample_maps=[sample_map], ...)``
    intended for adding a single table collection to another.
    """
    self.concatenate(
        [other],
        node_mappings=[node_mapping],
        record_provenance=False,
        **kwargs
    )
    if record_provenance:
        parameters = {
            "command": "append",
            "TODO": "add append parameters"  # tricky because the other TC has provenances
        }
        self.provenances.add_row(
        )
```

I think it's helpful to have a function that doesn't require the user to wrap the `[other]` in a list, but rather than have a separate `append` method, we could allow concatenate to take *either* a list of Table Collections, or a single table collection? Or we could take all non-keyword arguments to be additional TableCollections.

hyanwong:
## Description

Fixes #3164 

# PR Checklist:

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -7074,6 +7074,182 @@ class TestTrim(unittest.TestCase):
             ts.trim()
 
 
+class TestShift:
+    """
+    Test the shift functionality
+    """
+
+    @pytest.mark.parametrize("shift", [-0.5, 0, 0.5])
+    def test_shift(self, shift):
+        ts = tskit.Tree.generate_comb(2, span=2).tree_sequence
+        tables = ts.dump_tables()
+        tables.delete_intervals([[0, 1]], simplify=False)
+        tables.sites.add_row(1.5, "A")
+        ts = tables.tree_sequence()
+        ts = ts.shift(shift)
+        assert ts.sequence_length == 2 + shift
+        assert np.min(ts.tables.edges.left) == 1 + shift
+        assert np.max(ts.tables.edges.right) == 2 + shift
+        assert np.all(ts.tables.sites.position == 1.5 + shift)
+        assert len(list(ts.trees())) == ts.num_trees
+
+    def test_sequence_length(self):
+        ts = tskit.Tree.generate_comb(2).tree_sequence
+        ts = ts.shift(1, sequence_length=3)
+        assert ts.sequence_length == 3
+        ts = ts.shift(-1, sequence_length=1)
+        assert ts.sequence_length == 1
+
+    def test_empty(self):
+        empty_ts = tskit.TableCollection(1.0).tree_sequence()
+        empty_ts = empty_ts.shift(1)
+        assert empty_ts.sequence_length == 2
+        empty_ts = empty_ts.shift(-1.5)
+        assert empty_ts.sequence_length == 0.5
+        assert empty_ts.num_nodes == 0
+
+    def test_migrations(self):
+        tables = tskit.Tree.generate_comb(2, span=2).tree_sequence.dump_tables()
+        tables.populations.add_row()
+        tables.migrations.add_row(0, 1, 0, 0, 0, 0)
+        ts = tables.tree_sequence().shift(10)
+        assert np.all(ts.tables.migrations.left == 10)
+        assert np.all(ts.tables.migrations.right == 11)
+
+    def test_provenance(self):
+        ts = tskit.Tree.generate_comb(2).tree_sequence
+        ts = ts.shift(1, record_provenance=False)
+        params = json.loads(ts.provenance(-1).record)["parameters"]
+        assert params["command"] != "shift"
+        ts = ts.shift(1, sequence_length=9)
+        params = json.loads(ts.provenance(-1).record)["parameters"]
+        assert params["command"] == "shift"
+        assert params["value"] == 1
+        assert params["sequence_length"] == 9
+
+    def test_too_negative(self):
+        ts = tskit.Tree.generate_comb(2).tree_sequence
+        with pytest.raises(tskit.LibraryError, match="TSK_ERR_BAD_SEQUENCE_LENGTH"):
+            ts.shift(-1)
+
+    def test_bad_seq_len(self):
+        ts = tskit.Tree.generate_comb(2).tree_sequence
+        with pytest.raises(
+            tskit.LibraryError, match="TSK_ERR_RIGHT_GREATER_SEQ_LENGTH"
+        ):
+            ts.shift(1, sequence_length=1)
+
+
+class TestConcatenate:
+    def test_simple(self):
+        ts1 = tskit.Tree.generate_comb(5, span=2).tree_sequence
+        ts2 = tskit.Tree.generate_balanced(5, arity=3, span=3).tree_sequence
+        assert ts1.num_samples == ts2.num_samples
+        assert ts1.num_nodes != ts2.num_nodes
+        joint_ts = ts1.concatenate(ts2)
+        assert joint_ts.num_nodes == ts1.num_nodes + ts2.num_nodes - 5
+        assert joint_ts.sequence_length == ts1.sequence_length + ts2.sequence_length
+        assert joint_ts.num_samples == ts1.num_samples
+        ts3 = joint_ts.delete_intervals([[2, 5]]).rtrim()
+        # Have to simplify here, to remove the redundant nodes
+        assert ts3.equals(ts1.simplify(), ignore_provenance=True)
+        ts4 = joint_ts.delete_intervals([[0, 2]]).ltrim()
+        assert ts4.equals(ts2.simplify(), ignore_provenance=True)
+
+    def test_multiple(self):
+        np.random.seed(42)
+        ts3 = [
+            tskit.Tree.generate_comb(5, span=2).tree_sequence,
+            tskit.Tree.generate_balanced(5, arity=3, span=3).tree_sequence,
+            tskit.Tree.generate_star(5, span=5).tree_sequence,
+        ]
+        for i in range(1, len(ts3)):
+            # shuffle the sample nodes so they don't have the same IDs
+            ts3[i] = ts3[i].subset(np.random.permutation(ts3[i].num_nodes))
+        assert not np.all(ts3[0].samples() == ts3[1].samples())
+        assert not np.all(ts3[0].samples() == ts3[2].samples())
+        assert not np.all(ts3[1].samples() == ts3[2].samples())
+        ts = ts3[0].concatenate(*ts3[1:])
+        assert ts.sequence_length == sum([t.sequence_length for t in ts3])
+        assert ts.num_nodes - ts.num_samples == sum(
+            [t.num_nodes - t.num_samples for t in ts3]
+        )
+        assert np.all(ts.samples() == ts3[0].samples())
+
+    def test_empty(self):
+        empty_ts = tskit.TableCollection(10).tree_sequence()
+        ts = empty_ts.concatenate(empty_ts, empty_ts, empty_ts)
+        assert ts.num_nodes == 0
+        assert ts.sequence_length == 40
+
+    def test_samples_at_end(self):
+        ts1 = tskit.Tree.generate_comb(5, span=2).tree_sequence
+        ts2 = tskit.Tree.generate_balanced(5, arity=3, span=3).tree_sequence
+        # reverse the node order
+        ts1 = ts1.subset(np.arange(ts1.num_nodes)[::-1])
+        assert ts1.num_samples == ts2.num_samples
+        assert np.all(ts1.samples() != ts2.samples())
+        joint_ts = ts1.concatenate(ts2)
+        assert joint_ts.num_samples == ts1.num_samples
+        assert np.all(joint_ts.samples() == ts1.samples())
+
+    def test_internal_samples(self):
+        tables = tskit.Tree.generate_comb(4, span=2).tree_sequence.dump_tables()
+        nodes_flags = tables.nodes.flags
+        nodes_flags[:] = tskit.NODE_IS_SAMPLE
+        nodes_flags[-1] = 0  # Only root is not a sample
+        tables.nodes.flags = nodes_flags
+        ts = tables.tree_sequence()
+        joint_ts = ts.concatenate(ts)
+        assert joint_ts.num_samples == ts.num_samples
+        assert joint_ts.num_nodes == ts.num_nodes + 1
+        assert joint_ts.sequence_length == ts.sequence_length * 2
+
+    def test_some_shared_samples(self):
+        ts1 = tskit.Tree.generate_comb(4, span=2).tree_sequence
+        ts2 = tskit.Tree.generate_balanced(8, arity=3, span=3).tree_sequence
+        shared = np.full(ts2.num_nodes, tskit.NULL)
+        shared[0] = 1
+        shared[1] = 0
+        joint_ts = ts1.concatenate(ts2, node_mappings=[shared])
+        assert joint_ts.sequence_length == ts1.sequence_length + ts2.sequence_length
+        assert joint_ts.num_samples == ts1.num_samples + ts2.num_samples - 2
+        assert joint_ts.num_nodes == ts1.num_nodes + ts2.num_nodes - 2
+
+    def test_provenance(self):
+        ts = tskit.Tree.generate_comb(2).tree_sequence
+        ts = ts.concatenate(ts, record_provenance=False)
+        params = json.loads(ts.provenance(-1).record)["parameters"]
+        assert params["command"] != "concatenate"
+
+        ts = ts.concatenate(ts)
+        params = json.loads(ts.provenance(-1).record)["parameters"]
+        assert params["command"] == "concatenate"
+
+    def test_unequal_samples(self):
+        ts1 = tskit.Tree.generate_comb(5, span=2).tree_sequence
+        ts2 = tskit.Tree.generate_balanced(4, arity=3, span=3).tree_sequence
+        with pytest.raises(ValueError, match="must have the same number of samples"):
+            ts1.concatenate(ts2)
+
+    @pytest.mark.skip(
+        reason="union bug: https://github.com/tskit-dev/tskit/issues/3168"
+    )
+    def test_duplicate_ts(self):
+        ts1 = tskit.Tree.generate_comb(3, span=4).tree_sequence
+        ts = ts1.keep_intervals([[0, 1]]).trim()  # a quarter of the original
+        nm = np.arange(ts.num_nodes)  # all nodes identical
+        ts2 = ts.concatenate(ts, ts, ts, node_mappings=[nm] * 3, add_populations=False)
+        ts2 = ts2.simplify()  # squash the edges
+        assert ts1.equals(ts2, ignore_provenance=True)
+
+    def test_node_mappings_bad_len(self):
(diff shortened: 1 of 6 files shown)


python/tskit/tables.py
```diff
@@ -3991,6 +3991,93 @@ def trim(self, record_provenance=True):
...
+            }
+            self.provenances.add_row(
+                record=json.dumps(provenance.get_provenance_dict(parameters))
+            )
+
+    def concatenate(
+        self, other, *, node_mapping=None, record_provenance=True, **kwargs
+    ):
+        """
+        Concatenate another table collection to the right of this one. This shifts the
```
petrelharp, line 4030:
```suggestion
        Concatenate another table collection to the right of this one. This {meth}`shift`s the
```

python/tskit/tables.py
```diff
@@ -3991,6 +3991,93 @@ def trim(self, record_provenance=True):
...
+        other table coordinate rightwards, then repeatedly calls {meth}`union` with
```
petrelharp, line 4031:
```suggestion
        other table coordinate rightwards, then calls {meth}`union` with
```

python/tskit/tables.py
```diff
@@ -3991,6 +3991,93 @@ def trim(self, record_provenance=True):
...
+        the provided ``node_mapping``. If no node mapping is give, the two table
```
petrelharp, line 4032:
```suggestion
        the provided ``node_mapping``. If no node mapping is given, the two table
```

python/tskit/tables.py
```diff
@@ -3991,6 +3991,93 @@ def trim(self, record_provenance=True):
...
+        collections must have the same number of samples, and those are treated
+        (in numerical order) as shared between the two table collections.
+        This is identical to :meth:`TreeSequence.concatenate` but
+        acts *in place* to alter the data in this :class:`TableCollection`.
+
+        .. note::
+            To add gaps between the concatenated tables, use :meth:`shift` or
+            to remove gaps, use :meth:`trim` before concatenating.
```
petrelharp, line 4040:
```suggestion
            To add gaps between the concatenated tables, use :meth:`shift` before
            concatenating; to remove gaps, use :meth:`trim`.
```

python/tskit/tables.py
```diff
@@ -3991,6 +3991,93 @@ def trim(self, record_provenance=True):
...
+            other_samples = np.where(other.nodes.flags & tskit.NODE_IS_SAMPLE)[0]
+            if len(other_samples) != len(samples):
+                raise ValueError(
+                    "each `other` must have the same number of samples as `self`"
+                )
+            node_mapping = np.full(other.nodes.num_rows, tskit.NULL, dtype=np.int32)
+            node_mapping[other_samples] = samples
+        other.shift(self.sequence_length, record_provenance=False)
+        self.sequence_length = other.sequence_length
+        # NB: should we use other defaults for add_populations & check_shared_equality?
```
petrelharp, line 4065: I think you do need to do `check_shared_equality=False`, and add a test where one sample is parent to another sample in one of the tree sequences.

python/tskit/tables.py
hyanwong: Nicer! thanks

python/tskit/tables.py
hyanwong: Ah yes, good point.

I think we are fine to `check_shared_equality=False`, as the method will fail when converting to a tree sequence otherwise, anyway? Ant there shouldn't be any shared topology since the trees are non-overlapping along the genome.

hyanwong: You were right - it fails when I test with internal samples. Test added, and check_shared_equality set to False

hyanwong:
Thanks for the review @petrelharp . I tacked on another quick commit to add that ref to subset in the union docs.

python/tskit/tables.py
```diff
@@ -3991,6 +3991,96 @@ def trim(self, record_provenance=True):
...
+        if node_mapping is None:
+            samples = np.where(self.nodes.flags & tskit.NODE_IS_SAMPLE)[0]
+            other_samples = np.where(other.nodes.flags & tskit.NODE_IS_SAMPLE)[0]
+            if len(other_samples) != len(samples):
+                raise ValueError(
+                    "each `other` must have the same number of samples as `self`"
+                )
+            node_mapping = np.full(other.nodes.num_rows, tskit.NULL, dtype=np.int32)
+            node_mapping[other_samples] = samples
+        other.shift(self.sequence_length, record_provenance=False)
```
jeromekelleher, line 4065: Ah, I hadn't clocked that this formulation would required that we alter the parameter. That's pretty nasty and surprising, and I don't thinks we should do it.

python/tskit/trees.py
```diff
@@ -7060,6 +7060,72 @@ def trim(self, record_provenance=True):
...
+            record_provenance=record_provenance,
+        )
+        try:
+            ts = tables.tree_sequence()
+        except ValueError as e:
+            raise ValueError("Cannot shift due to bad coordinate values") from e
+        return ts
+
+    def concatenate(
+        self, other, *, node_mapping=None, record_provenance=True, **kwargs
```
jeromekelleher, line 7094: Let's change the signature here to
def 
```python
def concatenate(self, *args, *, node_mapping=None, record_provenance=True, **kwargs):
      tables = self.dump_tables() 
      # Union will probably invalidate the indexes anyway, but good to be explicit here.
      tables.drop_index()
     
      for other_ts in args:
            other_tables = other.dump_tables()
            other_tables.shift(tables.sequence_length, record_provenance=False)
            tables.union(other, stuff) 
    return tables.tree_sequence()
```
this way we are guaranteed not to change the input parameters and should be able to make more efficient later in terms of avoiding sorts and so on later if we need to.

python/tskit/tables.py
```diff
@@ -3991,6 +3991,96 @@ def trim(self, record_provenance=True):
...
+        :param sequence_length: The new sequence length of the tree sequence. If
+            ``None`` (default) add `value` to the sequence length.
+        """
+        self.edges.left += value
+        self.edges.right += value
+        self.sites.position += value
+        if sequence_length is None:
+            self.sequence_length += value
+        else:
+            self.sequence_length = sequence_length
```
jeromekelleher, line 4014: Hmm, this should invalidate the indexes but it doesn't seem to have triggered any issues ontesting... We should add a ``self.drop_indexes()`` at the start to be explicit that this is happening.

python/tests/test_topology.py
```diff
@@ -7074,6 +7074,136 @@ def test_failure_with_migrations(self):
...
+        ts = tskit.Tree.generate_comb(2, span=2).tree_sequence
+        tables = ts.dump_tables()
+        tables.delete_intervals([[0, 1]], simplify=False)
+        tables.sites.add_row(1.5, "A")
+        ts = tables.tree_sequence()
+        ts = ts.shift(shift)
+        assert ts.sequence_length == 2 + shift
+        assert np.min(ts.tables.edges.left) == 1 + shift
+        assert np.max(ts.tables.edges.right) == 2 + shift
+        assert np.all(ts.tables.sites.position == 1.5 + shift)
```
jeromekelleher, line 7093: Can you add a test here that checks that the indexes are correctly recomputed after shifting. Iterating over the trees should do it, like ``assert len(list(ts.trees()) == ts.num_trees``

jeromekelleher (commented):
There's a few tricky subtleties not handled here.

1. We should definitely not alter a parameter table collection - so I don't think we need TableCollection.concatenate and can just have TreeSequence.concatenate(other1, other2,....) (I realise that I asked you to focus on the pairwise approach, but I hadn't realised that we were updating all the tablecollections in order to implement)
2. I'm not clear how ``TableCollection.shift()`` is interacting with the indexes. This is important.

python/tskit/tables.py
hyanwong: Oh, that's a very good point. I wonder if we can avoid that somehow. E.g. could we add the other to the *left* of self? That way we only need to shift the coordinate system of `self` every time.

Alternatively, we could temporarily shift the coordinates of `self` so that they are negative, and end at zero, then union with `other`, then shift the whole lot rightwards again. I think that avoids modifying `other`, but at the expense of calling `shift` twice rather than once.

jeromekelleher: Let's not bother and just do it for tree sequences instead.

hyanwong: I just checked. We can't `union` a table collection with negative positions. But we could add to the left easily enough, which doesn't alter `other`. The naming would probably need changing. E.g. the TableCollection method could be `prepend`, and then we could easily use that in a TreeSequence.concatenate() function:

```python
def concatenate(self, other):
    tables = other.dump_tables()
    tables.prepend(self.tables)
    return tables.tree_sequence()
```

python/tskit/trees.py
hyanwong: I wondered about that, but you might need a different node_mapping for each of the args. It's find if you only concatenate on the samples (and all of them have the same number of samples), but not if you have an explicit mapping, I think?

python/tskit/tables.py
jeromekelleher: Please, let's not bother - the TreeSequenced version is all we need and has the required semantics.

python/tskit/trees.py
hyanwong: The node mapping(s) have to match the number of nodes in each of the `others` (which are likely to be different). So I think it looks like this:


```python
    def concatenate(
        self, others, *, node_mappings=None, record_provenance=True, **kwargs
    ):
        """
        Concatenate a set of tree sequences to the right of this one, by repeatedly
        calling {meth}`union` with ``check_shared_equality=False`` and (optionally)
        a node mapping for each of the ``others`` If any node mapping is ``None``
        only map the sample nodes between the input tree sequence and this one,
        based on the numerical order of sample node IDs.

        .. note::
            To add gaps between the concatenated tables, use :meth:`shift` or
            to remove gaps, use :meth:`trim` before concatenating.

        :param list others: A list of other tree sequences to add to the right
            of this one.
        :param Union[list, None] node_mappings: An list of node mappings for each
            tree sequence in ``others``. Each should either be an array of integers
            of the same length as the number of nodes in ``other`` (see :meth:`union`
            for details), or ``None``. If ``None``, only sample nodes are mapped to
            each other. Default: ``None``, treated as ``[None] * len(others)``.
        :param bool record_provenance: If True (default), record details of this call to
            ``concatenate`` in the returned tree sequence's provenance information
            (Default: True).
        :param \\**kwargs: Additional keyword arguments to pass to :meth:`union`,
            e.g. ``add_populations``.
        """
        if node_mappings is None:
            node_mappings = [None] * len(others)

        samples = self.samples()
        tables = self.dump_tables()
        for node_mapping, other in zip(node_mappings, others):
            if node_mapping is None:
                other_samples = other.samples()
                if len(other_samples) != len(samples):
                    raise ValueError(
                        "each `other` must have the same number of samples as `self`"
                    )
                node_mapping = np.full(other.num_nodes, tskit.NULL, dtype=np.int32)
                node_mapping[other_samples] = samples
            other_tables = other.dump_tables()
            other_tables.shift(self.sequence_length, record_provenance=False)
            tables.sequence_length = other_tables.sequence_length
            # NB: should we use a different default for add_populations?
            tables.union(
                other_tables,
                node_mapping=node_mapping,
                check_shared_equality=False,  # Needed as checks fail with internal samples
                record_provenance=False,
                **kwargs,
            )
        if record_provenance:
            parameters = {
                "command": "concatenate",
                "TODO": "add concatenate parameters",  # tricky as both have provenances
            }
            tables.provenances.add_row(
            )

        return tables.tree_sequence()
```

hyanwong: I had to `import json` and `import tskit.provenance as provenance` at the top of trees.py, which implies this is the only function of this sort that doesn't have a TableCollection equivalent. But I guess that's fine.

hyanwong: (Oh, and I use the convention of passing in a list to match the numpy `concatenate` method, and so that it's easy to say that `node_mappings` has to be of the same length as the `others` list. We could also allow both to be a single item however, or use `*args` and a keep a list for the potential `node_mappings`.

python/tests/test_topology.py
jeromekelleher: I don't see the test?

petrelharp:
> We should definitely not alter a parameter table collection

Sorry what did you mean there, @jeromekelleher?

petrelharp:
And - I don't see why `shift`ing should change the indices?

jeromekelleher:
>> We should definitely not alter a parameter table collection

> Sorry what did you mean there, @jeromekelleher?

The previous version called .shift on the parameter TableCollection, which modifies it. So, calling ``.concat`` twice on the same input would give different answers, e.g., which would be pretty nasty

jeromekelleher:
> And - I don't see why shifting should change the indexes?

I guess it shouldn't alter the relative order of the edges all right - but I'm not clear whether the current code will automatically invalidate the indexes, or if they are just left as is and that's OK?

python/tskit/tables.py
```diff
@@ -3991,6 +3991,39 @@ def trim(self, record_provenance=True):
...
+        .. note::
+            No attempt is made to check that the new coordinate system or sequence length
+            is valid: if you wish to do this, use {meth}`TreeSequence.shift` instead.
+
+        :param value: The amount by which to shift the coordinate system.
+        :param sequence_length: The new sequence length of the tree sequence. If
+            ``None`` (default) add `value` to the sequence length.
+        """
+        self.drop_index()
+        self.edges.left += value
```
jeromekelleher, line 4009: What about migrations? Probably simplest to just implement this, as testing for a  ValueError is about as much effort.

python/tskit/trees.py
```diff
@@ -7060,6 +7062,108 @@ def trim(self, record_provenance=True):
...
+        :raises ValueError: If the new coordinate system is invalid (e.g., if
+            shifting the coordinate system results in negative coordinates).
+        """
+        tables = self.dump_tables()
+        tables.shift(
+            value=value,
+            sequence_length=sequence_length,
+            record_provenance=record_provenance,
+        )
+        try:
```
jeromekelleher, line 7089: Looks like this isn't covered - and  is this try/catch/rereaise really needed?

jeromekelleher (commented):
Looks good - missing an edge case on migrations though.

python/tskit/tables.py
hyanwong: Good catch. I completely forgot that migrations have a left and right too. I will implement and test

python/tskit/trees.py
hyanwong: It'll be caught anyway I guess, so I'll remove it.

hyanwong:
Addressed comments and squashed. Should be mergeable.

python/tskit/trees.py
```diff
@@ -7060,6 +7062,104 @@ def trim(self, record_provenance=True):
...
+            call to ``concatenate`` in the returned tree sequence's provenance
+            information (Default: True).
+        :param bool add_populations: If True (default), nodes new to ``self`` will
+            be assigned new population IDs (see :meth:`union`)
+        """
+        if node_mappings is None:
+            node_mappings = [None] * len(args)
+        if add_populations is None:
+            add_populations = True
+        if len(node_mappings) != len(args):
```
jeromekelleher, line 7123: Codecov says this is still not coverted, I think it's right?

jeromekelleher (commented):
Looks good to merge, once we've covered the remaining missing bit

jeromekelleher (approved):

PR #3174, opened 2025-05-23
Include positions in VCF mapping

### LINKED ISSUE #3173: Add coordinate mapping to the VcfModelMap
The new  ``map_vcf_model`` is great for downstream users of the data as it tells us how individuals map to "samples" under various scenarios. Another quirky thing about tskit (from VCF's perspective) is the allowance for 0 and continuous coordinates. We deal with this in tskit ``write_vcf`` by various tricks,  and it would be nice if we didn't have to do all that again in bio2zarr.

Is there any reason we can't put the coordinate mapping logic into the ``map_vcf_model`` function as well? We're returning a data class, so there's no problem in adding another array in there (which would presumably be an int64 array of the mapped coordinates).

Any thoughts @benjeffery?

benjeffery:
Fixes #3173

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,4 +1,13 @@
 --------------------
+[0.6.5] - 2025-0X-XX
+--------------------
+
+**Features**
+
+- ``TreeSequence.map_to_vcf_model`` now also returns the transformed positions and
+  contig length. (:user:`benjeffery`, :pr:`XXXX`, :issue:`3173`)
+
+--------------------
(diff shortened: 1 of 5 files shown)


jeromekelleher (commented):
Hmm, that is nice and straightforward, very nice. Let's do it.

I don't think we need to push out the release immediately though, it's not urgent for bio2zarr as there's a lot of stuff we want to push through to get a release out ASAP

jeromekelleher (approved):

PR #3176, opened 2025-05-28
Fix bug in `ts.pair_coalescence_counts` when a window breakpoint falls within missing interval

### LINKED ISSUE #3175: Incorrect span normalisation for `ts.pair_coalescence_counts` with missing data and windows
There's a bug in how *internal* missing data is tallied by:
```python
ts = msprime.sim_ancestry(samples=1, discrete_genome=False)                                                                              
ts = ts.delete_intervals(np.array([[0.3, 0.6]]))
counts = ts.pair_coalescence_counts(
    windows=np.array([0.0, 0.31, 1.0]),
    time_windows=np.array([0.0, np.inf]),
    span_normalise=True,
).flatten()
np.testing.assert_array_almost_equal(counts, np.ones(2))  # <<-- fails
#assert np.all(counts > 0)  # <<-- also fails
```
this issue doesn't occur when the missing data is all at the flanks.

nspope:
Fixes #3175 via a faulty interval intersection in the C implementation, and adds a test that would fail on main

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -9586,8 +9586,9 @@ tsk_treeseq_pair_coalescence_stat(const tsk_treeseq_t *self, tsk_size_t num_samp
                 window_span = windows[w + 1] - windows[w] - missing_span;
                 missing_span = 0.0;
                 if (num_edges == 0) {
+                    /* missing interval, so remove overcounted missing span */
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
Nice catch, LGTM

PR #3179, opened 2025-05-28
Two-way two-locus statistics (python proto)

### LINKED ISSUE #2832: Multipopulation two-site statistics
We'd like to implement two-population versions of the statistics we've implemented thus far: $D$, $D^{\prime}$, $D^{2}$, $D_{z}$, $\pi_{2}$, $r$, $r^{2}$

This means that we'll strictly require two sample sets, comparing the A locus from sample set 1 and the B locus from sample set 2.

There is a lot of [prior art](https://github.com/tskit-dev/tskit/blob/main/c/tskit/trees.c#L3132) and the way we've structured two-locus statistics, this should be relatively straightforward.

Are there any other stats we want to implement @apragsdale?

lkirk:
This PR implements two-way LD statistics, specified between sample sets. During the development of this functionality, a number of issues with the designation of state_dims/result_dims were discovered. These have been resolved, testing clean for existing code and providing the proper behavior for this new code.

The mechanism by which users will specify a multi-population (or two-way) statistic is by providing the `index` argument. This helps us avoid creating another `ld_matrix` method for the TreeSequence object.

In other words, for a one-way statistic, a user would specify:
```python
ts.ld_matrix(stat="D2", sample_sets=[ss1, ss2])
```
Which would output a 3D ndarray containing one LD matrix per sample set.

```python
ts.ld_matrix(stat="D2", sample_sets=[ss1, ss2], indexes=[(0, 1)])
```
Which would output a 2D ndarray containing one LD matrix for the index pair. This would use our `D2_ij_summary_func`, instead of the `D2_summary_func`. Finally, if a user provided
```python
ts.ld_matrix(stat="D2", sample_sets=[ss1, ss2], indexes=[(0, 1), (1, 1)])
```
We would output a 3D ndarray containing one LD matrix _per_ index pair provided.

Since these are two-way statistics, the indexes must be length 2. We plan on enabling users to implement k-way via a "general_stat" api. We did not implement anything more than two-way statistics here because of the combinatoric explosion of logic required for indexes > 2.

I added some basic tests to demonstrate that things were working properly. If we compute two-way statistics on identical sample sets, they should be equal to the one-way statistics. Unfortunately, this does not apply to unbiased statistics, which I've validated manually.

I've also cleaned up the docstrings a bit and fixed a bug with the D_prime statistic, which should not be weighted by haplotype frequency.

Fixes #2832


python/tests/test_ld_matrix.py
```diff
@@ -810,6 +844,30 @@ def two_locus_count_stat(
...
+    if indexes is not None:
+        indexes = tskit.util.safe_np_int_cast(indexes, np.int32)
+        idx_lens = {len(i) for i in indexes}
+        if idx_lens != {2}:
+            raise ValueError(
+                f"Sample set indexes must be length 2, lengths: {idx_lens}"
+            )
+
+        check_sample_stat_inputs(num_sample_sets, 2, result_dim, indexes)
+        # Unbiased multipopulation statistics require disjoint sample sets because
```
apragsdale, line 858: Do we want to be this restrictive about the specification of sample sets? Perhaps this could be something that is simply documented, that the construction of the unbiased cross-population D^2 stat assumes sample sets do not overlap, and not raise an error.

apragsdale (commented):
Thanks @lkirk.

> If we compute two-way statistics on identical sample sets, they should be equal to the one-way statistics. Unfortunately, this does not apply to unbiased statistics, which I've validated manually.

This is ok (even expected) for the unbiased statistics, since the formulation assumes sampling without replacement, but having overlapping sample sets would essentially copy those repeated sample nodes. I don't think we should worry about it.

python/tests/test_ld_matrix.py
lkirk: After discussing this offline, it makes sense to remove this check. It will maintain consistency with other multi-way statistics (such as divergence) to warn in the documentation, but let the user decide the sample sets.

jeromekelleher (commented):
I've had a look through and the code all LGTM. The changes seem sensible, but I'd need to spend a lot longer on it to really digest the details and understand the finer points.

I'm going to ping @petrelharp here for a review as he's much better at this sort of stuff

python/tests/test_ld_matrix.py
```diff
@@ -1128,22 +1344,63 @@ def d2_unbiased(
...
+        raise ValueError(
+            "Insufficient number of sample sets: "
+            f"num_sample_sets: {num_sample_sets} tuple_size: {tuple_size}"
+        )
+    if num_index_tuples < 1:
+        raise ValueError(f"Insufficient number of index tuples: {num_index_tuples}")
+    check_set_indexes(num_sample_sets, tuple_size * num_index_tuples, index_tuples)
+
+
+def ld_matrix(
```
petrelharp, line 1399: this is the proposed API

jeromekelleher (approved):
I'm happy to merge this as a step in the right direction - are you ok to merge @petrelharp ?

PR #3180, opened 2025-05-29
union doc clarification: closes #3168

### LINKED ISSUE #3168: Unexpected `union` behaviour: edges dropped if tables have all identical nodes
I think this should produce a table with edges spanning 0..2, but instead, the `union`ed TS is missing 0..1

```python
import tskit
import numpy as np

ts = tskit.Tree.generate_comb(2, span=2).tree_sequence
# Cut it up
ts1 = ts.keep_intervals([[0, 1]])
ts2 = ts.delete_intervals([[0, 1]])
node_mapping = np.arange(ts.num_nodes)
# Paste back together
ts_both = ts2.union(ts1, node_mapping=node_mapping, check_shared_equality=False)
ts_edges_squashed = ts_both.simplify()
print(ts_both.draw_text())
assert ts.equals(ts_edges_squashed, ignore_provenance=True)  # fails
```

Here's what the result of `union` looks like

```
1.00┊     ┊  2  ┊
    ┊     ┊ ┏┻┓ ┊
0.00┊ 0 1 ┊ 0 1 ┊
    0     1     2
```

petrelharp:
Along the way I noticed another confusing thing that might happen: if you try to union two contradictory bits you get an error, so I've clarified that also. (In retrospect, not so confusing...)

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -7404,7 +7404,8 @@ class TreeSequence:
         By default, populations of newly added nodes are assumed to be new
         populations, and added to the population table as well. This can be
         thought of as a "node-wise" union: for instance, it can not be used
-        to add new edges between nodes already in ``self``.
+        to add new edges between two nodes already in ``self`` or new mutations
+        above nodes already in ``self``.
 
         If the resulting tree sequence is invalid (for instance, a node is
         specified to have two distinct parents on the same interval),


petrelharp:
To be clear - the situation you're talking about with mutations would also fail the `check_shared_overlap` check, right?

jeromekelleher (approved):

PR #3183, opened 2025-05-29
Create a new merge() function, and use that for concatenate

### LINKED ISSUE #3181: span-wise union
As discussed in #3168, something we'd like to do is squash together tree sequences describing different bits of the genome. For instance: say I've got simulations of 3L and 3R and would like to make a tree sequence of the entire chromosome 3. One possible not-very-good name: `ts1.union_span(ts2)`?  (Or: `merge`?)

Possible functionality:
1. the span covered by all edges in ts1 and in ts2 do not overlap (no overlap)
2. any edge in ts2 describes a parent for a node in ts2 that has no parent on that span (no node-wise overlap)
3. ts1 and ts2 can overlap but they must be identical where they do

So, where `ts.union` was a node-wise union, this would be an edge-wise union. Case (3) is the most general, but also definitely the most annoying to code up and I am not aware of any use cases. Case (1) would be easy and efficient: simply loop through edges in each in insertion order and check for any overlap; @hyanwong suggested special-casing this one anyhow. Case (2) would require building the trees, I think, so I'm not sure if it'd be much easier than (3). 

So I think I'm proposing doing (1), and am looking for a name? But suggestions of use cases for (2) or (3) welcome?

hyanwong:
## Description

It appears as if `union` wasn't really the right underlying function to use for `concatenate`, e.g. it doesn't deal as expected with sites above samples, so I had to roll my own `merge` utility.

Fixes #3181 

# PR Checklist:

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -4263,7 +4263,8 @@ class TableCollection(metadata.MetadataProvider):
                     i >= self.populations.num_rows
                     or self.populations[i] != other.populations[i]
                 ):
-                    raise ValueError("Non-matching populations")
+                    raise ValueError(f"Non-matching populations:\n   self: {self.populations[i]}"
+                                                  f"\n   self: {other.populations[i]}")
         individual_map = {}
         population_map = {}
         for new_node in np.where(node_map == tskit.NULL)[0]:


hyanwong:
p.s. I set `add_populations=False` in the `concatenate` method, overriding the default, as I expect concatenate to be used e.g. on multiple chromosomes with the same population definitions, and it would be confusing to create new populations for each chromosome, I feel.

Defaulting to `add_populations=True` for the `merge` method seems OK though, as this is more of a low-level thing, and matches `union`.

hyanwong:
> Might you be able to write down an explicit list of what exactly gets added, in the style of union?

Good idea. I'll do that, and add some WF-with-overlapping-gens type tests (I have hacked a few with internal samples in there already, but I can add one or two more)

hyanwong:
Added the following list, and a couple of WF tests for `concatenate`

```
Items that will be ported from ``other`` into ``self`` (and given new IDs) are:

1. All edges in ``other``
2. All migrations in ``other``
3. All mutations in ``other``
4. Sites whose positions are new to ``self``
5. Individuals whose nodes are new to ``self``.
6. If ``add_populations=True``, populations whose nodes are new to ``self``
```

hyanwong:
> We still need to agree on a different name than "merge" though

Here's a suggestion for reducing confusion. I implement a 'union' like part which skips merging edges if they are identical, then we rename the existing `union` to `union_nodes`, and call this one `union_edges`. We keep the name `union` as a deprecated (but permanently maintained) alias to `union_nodes` (we could or could not document it, I suppose)

python/tskit/tables.py
```diff
@@ -4196,6 +4199,142 @@ def union(
...
+                nodes_pop = nodes_pop[node_map != tskit.NULL]
+            for i in np.unique(nodes_pop[nodes_pop != tskit.NULL]):
+                if (
+                    i >= self.populations.num_rows
+                    or self.populations[i] != other.populations[i]
+                ):
+                    raise ValueError("Non-matching populations")
+        individual_map = {}
+        population_map = {}
+        for new_node in np.where(node_map == tskit.NULL)[0]:
```
petrelharp, line 4269: Perhaps we should check that the nodes in the node map are identical? So something like
```
for oj, sj in enumerate(node_map):
    if sj != tskit.NULL:
        if self.node(sj) != other.node(oj):
            raise ValueError("Nodes do not match...")
    else:
          # continue as below
```
```

File: python/tskit/tables.py [hunk 2]
```diff
@@ -4196,6 +4199,142 @@ def union(
...
+                            "One of the tree sequences to concatenate has a "
+                            "population not present in the existing tree sequence"
+                        )
+            node_map[new_node] = self.nodes.append(node.replace(**params))
+
+        for e in other.edges:
+            self.edges.append(
+                e.replace(child=node_map[e.child], parent=node_map[e.parent])
+            )
+        site_map = {}
```
petrelharp, line 4297: You could save some trouble here by doing
```
ns = self.sites.num_rows
self.sites.append_columns(**other.sites.asdict())
for mut in other.mutations:
    self.mutations.append(
        mut.replace(
            node=node_map[mut.node],
            site=mut.site + ns,
            parent=tskit.NULL
        )
# then after the self.sort()
self.deduplicate_sites()
```

petrelharp:
> Items that will be ported from ``other`` into ``self`` (and given new IDs) are:
> 
> 1. All edges in ``other``
> 2. All migrations in ``other``
> 3. All mutations in ``other``
> 4. Sites whose positions are new to ``self``
> 5. Individuals whose nodes are new to ``self``.
> 6. If ``add_populations=True``, populations whose nodes are new to ``self`

This list is missing the new nodes thenselves? (those for which `node_mapping` is NULL); otherwise, looks good!

python/tskit/tables.py
hyanwong: Aha, I forgot about deduplicate_sites, thanks.

While I'm at it, is there a way of figuring out quickly if edges already exist in the tree sequence? There's no deduplicate_edges functionality, or similar, right? And sorting edges with duplicates will simply error out.

petrelharp:
*Note:* I found another current use of `union`: https://tskit.dev/pyslim/docs/latest/vignette_continuing.html#continuing-the-simulation; this one could be done with either `union` or `merge`.

hyanwong:
> How does this sound?

I like the idea of not creating more confusion by similarly named functions, so something like that SGTM.

I guess the current function could be used as a python implementation in the test suite if we want. How about the checking of e.g. populations against each other (which includes metadata). I suppose this would be a good thing for the current `union` too?

hyanwong:
> Want me to have a go at the C version?

I would *love* it if you have time, thanks so much.

I'm happy to modify my python tests to check it works as intended.

hyanwong:
I converted this to draft, as @petrelharp 's suggestion is different, but we should use the tests from this one.

PR #3197, opened 2025-06-07
Make SVG standard widths and heights available in the class definition

hyanwong:


hyanwong:
I just tacked on another code-duplication removing nit, taking repetitive code out of trees.py to make a standard `draw()` function in the drawing module

PR #3201, opened 2025-06-08
Allow Y axis on RHS of SVG plot

hyanwong:
This is a bit trivial, but it was the easiest way to make nice tanglegram plots for sc2ts, in which trees are rotated in different directions (see below). So I thought as I had done the work, I might as well see if it is sensible to include in tskit? Otherwise I'll just keep it on my own branch.

Note that this allows `y_axis=True` or `y_axis="right"`, which is a bit yucky. No tests either, in case it seems a silly idea:

Screenshot 2025-06-08 at 18 34 50

### DIFF
--- a/python/tskit/drawing.py
+++ b/python/tskit/drawing.py
@@ -412,6 +412,19 @@ def check_x_lim(x_lim, max_x):
     return x_lim
 
 
+def check_y_axis(y_axis):
+    """
+    Checks the specified y_axis is valid and sets default if None.
+    """
+    if y_axis is None:
+        y_axis = False
+    if y_axis is True:
+        y_axis = "left"
+    if y_axis not in ["left", "right", False]:
+        raise ValueError(f"Unknown y_axis specification: '{y_axis}'.")
+    return y_axis
+
+
 def create_tick_labels(tick_values, decimal_places=2):
     """
     If tick_values are numeric, round the labels to X decimal_places, but do not print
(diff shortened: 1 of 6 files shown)


benjeffery:
Seems useful - if you don't like to overloaded arg then adding another arg that is `y_axis_right` would be possible. I'm ok with the overloaded arg in the "softer" parts of the API like this. I would change the code to only allow the strings "left" and "right" as currently a misspelling gives you odd behavior.

hyanwong:
I'm happy to overload the argument, and just allow "left" and "right".

hyanwong:
Should we default to having a Y axis be shown? I think there was an issue about this.

hyanwong:
OK, added the tests now (hard to test the actual layout, so I added it to two existing svg mockups)

python/tskit/drawing.py
```diff
@@ -1050,7 +1064,11 @@ def set_spacing(self, top=0, left=0, bottom=0, right=0):
         if self.x_axis:
             bottom += self.x_axis_offset
         if self.y_axis:
```
jeromekelleher: This would be more explicit as 
```
if self.y_axis == "left":
    left = self.y_axis_offset
elif self.y_axis == "right":
    right = self.y_axis_offset
```
The outer Falsey comparison doesn't add anything and just makes the logic less clear

hyanwong: Changed, and logic for `True` added above

PR #3202, opened 2025-06-09
Prevent iterating over a TopologyCounter

### LINKED ISSUE #1462: iterating over count_topologies method never stops
I would expect the `Tree.count_topologies()` method to return the Counter objects only for the valid sample sets, but iterating over a `tskit.TopologyCounter` seems never to end:

```
c = tskit.Tree.generate_star(3).count_topologies([[0],[1],[2]])
for a in c: # Infinite loop
    print(a)
```

Am I misunderstanding something @daniel-goldstein ?

benjeffery:
Fixes #1462


benjeffery:
@hyanwong You're closer to this than I am - this seems like it should fix #1462. yes?

jeromekelleher:
Let's take the simple path here and raise an error rather than try to implement the right thing

PR #3209, opened 2025-06-09
Make random-binary non-recursive

### LINKED ISSUE #2626: Remove recursion from generate_random_binary
A trivial thing, but this hits python's recursion depth limit:

```
tskit.Tree.generate_random_binary(1_000_000)
```

At some point in the distant future we might want to recode `generate_random_binary` not to use recursion.

hyanwong:
Fixes #2626

### DIFF
--- a/python/tskit/combinatorics.py
+++ b/python/tskit/combinatorics.py
@@ -150,23 +150,43 @@ class TreeNode:
         while root.parent is not None:
             root = root.parent
 
-        # Canonicalise the order of the children within a node. This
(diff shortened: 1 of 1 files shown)


jeromekelleher:
A rushed, vibe coded update to a subtle algorithm that has not been statistically validated is not helpful. What's the point of this? Increasing the stack depth is easily done, and for all we know this update completely destroys the statistical properties. I don't actually care if it doesn't, the point is that we all have better things to do.

PR #3210, opened 2025-06-10
Raise an error when messing with coords if there is a reference_sequence

### LINKED ISSUE #2091: How do reference sequences deal with trimming etc.
At the moment, I'm pretty sure that operations which change the coordinate system, primarily `trim`, `rtrim`, `ltrim`, and `shift` don't touch the reference sequence at all. I guess eventually they should (although how this happens when the ref seq is a URL, I have no idea). Alternatively, we should document that the TS after trimming may not have the reference sequence aligned correctly, so `alignments` will create bad data.

I though it worth opening up an issue just to keep track of this. I guess another issue is how to ensure that the reference sequence is of the same length as the genome. ISTR there was some discussion of this, but I can't find the issue on brief glance.

(NB probably worth dumping this issue in the reference sequence project)

hyanwong:
Fixes #2091

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -7074,6 +7074,18 @@ class TestTrim(unittest.TestCase):
         with pytest.raises(ValueError):
             ts.trim()
 
+    def test_reference_sequence(self):
+        # Test that we fail if there is a reference sequence
+        tables = tskit.TableCollection(3.0)
+        tables.reference_sequence.data = "ABC"
+        ts = tables.tree_sequence()
+        with pytest.raises(ValueError, match="reference sequence"):
+            ts.ltrim()
+        with pytest.raises(ValueError, match="reference sequence"):
+            ts.rtrim()
+        with pytest.raises(ValueError, match="reference sequence"):
+            ts.trim()
+
 
 class TestShift:
     """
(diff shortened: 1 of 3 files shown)


hyanwong:
I added an extra commit to this as I was documenting the error returns, as it was easier to make the trim() commands do the right thing with migrations than figure out how to word the description of when the error message was triggered!

python/tskit/tables.py
```diff
@@ -3910,6 +3907,8 @@ def ltrim(self, record_provenance=True):
 
         :param bool record_provenance: If ``True``, add details of this operation
             to the provenance table in this TableCollection. (Default: ``True``).
+        :raises ValueError: If the TableCollection has no edges, has a reference sequence
+            or has migrations that extend into the trimmed regions.
```
benjeffery: This line is no longer true?

python/tests/test_topology.py
```diff
@@ -7059,19 +7059,41 @@ def test_trims_no_effect(self):
         trimmed_ts = ts.trim(record_provenance=False)
         assert ts == trimmed_ts
 
-    def test_failure_with_migrations(self):
+    def test_migration_trimming(self):
         # All trim functions fail if migrations extend further than rightmost or
```
benjeffery: This comment is now orphaned.

benjeffery (commented):
Couple of comments, needs a changelog.

jeromekelleher:
These tests for migrations don't look very comprehensive to me - can we switch back to raising an error if we're not going to put the effort in to doing things properly? It is much better to not implement something (particularly if it's very unlikely to be used) that to do it badly and introduce bugs for others to trace down and fix later.

hyanwong:
OK, reverted to first commit, added changelog.

PR #3212, opened 2025-06-10
Check mutation parents on tree sequence init

benjeffery:
This PR makes tskit require that the mutation parents are correct - i.e. present and consistent with the tree topology. This can usually be done by calling `compute_mutation_parents` if the parents are not set already.

### DIFF
--- a/c/tskit/tables.c
+++ b/c/tskit/tables.c
@@ -10917,11 +10917,159 @@ out:
     return ret;
 }
 
+static int TSK_WARN_UNUSED
+tsk_table_collection_compute_mutation_parents_to_array(
+    const tsk_table_collection_t *self, tsk_id_t *mutation_parent)
+{
+    int ret = 0;
+    const tsk_id_t *I, *O;
+    const tsk_edge_table_t edges = self->edges;
+    const tsk_node_table_t nodes = self->nodes;
+    const tsk_site_table_t sites = self->sites;
+    const tsk_mutation_table_t mutations = self->mutations;
+    const tsk_id_t M = (tsk_id_t) edges.num_rows;
+    tsk_id_t tj, tk;
+    tsk_id_t *parent = NULL;
+    tsk_id_t *bottom_mutation = NULL;
+    tsk_id_t u;
+    double left, right;
+    tsk_id_t site;
+    /* Using unsigned values here avoids potentially undefined behaviour */
+    tsk_size_t j, mutation, first_mutation;
+
+    parent = tsk_malloc(nodes.num_rows * sizeof(*parent));
+    bottom_mutation = tsk_malloc(nodes.num_rows * sizeof(*bottom_mutation));
+    if (parent == NULL || bottom_mutation == NULL) {
+        ret = tsk_trace_error(TSK_ERR_NO_MEMORY);
+        goto out;
+    }
+    tsk_memset(parent, 0xff, nodes.num_rows * sizeof(*parent));
+    tsk_memset(bottom_mutation, 0xff, nodes.num_rows * sizeof(*bottom_mutation));
+    tsk_memset(mutation_parent, 0xff, self->mutations.num_rows * sizeof(tsk_id_t));
+
+    I = self->indexes.edge_insertion_order;
+    O = self->indexes.edge_removal_order;
+    tj = 0;
+    tk = 0;
+    site = 0;
+    mutation = 0;
+    left = 0;
+    while (tj < M || left < self->sequence_length) {
+        while (tk < M && edges.right[O[tk]] == left) {
+            parent[edges.child[O[tk]]] = TSK_NULL;
+            tk++;
+        }
+        while (tj < M && edges.left[I[tj]] == left) {
+            parent[edges.child[I[tj]]] = edges.parent[I[tj]];
+            tj++;
+        }
+        right = self->sequence_length;
+        if (tj < M) {
+            right = TSK_MIN(right, edges.left[I[tj]]);
+        }
+        if (tk < M) {
+            right = TSK_MIN(right, edges.right[O[tk]]);
+        }
+
+        /* Tree is now ready. We look at each site on this tree in turn */
+        while (site < (tsk_id_t) sites.num_rows && sites.position[site] < right) {
+            /* Create a mapping from mutations to nodes. If we see more than one
+             * mutation at a node, the previously seen one must be the parent
+             * of the current since we assume they are in order. */
+            first_mutation = mutation;
+            while (mutation < mutations.num_rows && mutations.site[mutation] == site) {
+                u = mutations.node[mutation];
+                if (bottom_mutation[u] != TSK_NULL) {
+                    mutation_parent[mutation] = bottom_mutation[u];
+                }
+                bottom_mutation[u] = (tsk_id_t) mutation;
+                mutation++;
+            }
+            /* Make the common case of 1 mutation fast */
+            if (mutation > first_mutation + 1) {
+                /* If we have more than one mutation, compute the parent for each
+                 * one by traversing up the tree until we find a node that has a
+                 * mutation. */
+                for (j = first_mutation; j < mutation; j++) {
+                    if (mutation_parent[j] == TSK_NULL) {
+                        u = parent[mutations.node[j]];
+                        while (u != TSK_NULL && bottom_mutation[u] == TSK_NULL) {
+                            u = parent[u];
+                        }
+                        if (u != TSK_NULL) {
+                            mutation_parent[j] = bottom_mutation[u];
+                        }
+                    }
+                }
+            }
+            /* Reset the mapping for the next site */
+            for (j = first_mutation; j < mutation; j++) {
+                u = mutations.node[j];
+                bottom_mutation[u] = TSK_NULL;
+                /* Check that we haven't violated the sortedness property */
+                if (mutation_parent[j] > (tsk_id_t) j) {
+                    ret = tsk_trace_error(TSK_ERR_MUTATION_PARENT_AFTER_CHILD);
+                    goto out;
+                }
+            }
+            site++;
+        }
+        /* Move on to the next tree */
+        left = right;
+    }
+
+out:
+    tsk_safe_free(parent);
+    tsk_safe_free(bottom_mutation);
+    return ret;
+}
+
+static int TSK_WARN_UNUSED
+tsk_table_collection_check_mutation_parents(const tsk_table_collection_t *self)
+{
+    int ret = 0;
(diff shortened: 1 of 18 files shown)


benjeffery:
So for a tree sequence with 4M total nodes, 3M mutations the `main` `tskit.load` time is 0.7s. With this PR that rises to 1.2s.

I think this is good enough for us to use for now, it can always be improved later with a non-breaking change.

I'm going to check all the other places that integrity is checked to see if they need this new check flag, then I'll add `TSK_COMPUTE_MUTATION_PARENTS` to `tsk_treeseq_init` and start fixing up the tests. I also need to add tests for the new code.

jeromekelleher:
I think this is a good path. We can whittle the time down another bit later, as you say.

benjeffery:
Open questions that occur as I'm fixing up tests:

 - Should we sort after computing parents inside `tsk_treeseq_init`? One test fails as it specifies mutations, then when the parents are assigned the sort order is incorrect. I lean towards checking the order and sorting if needed.

- `tree_diff_iter` checks integrity, but does it need to check mutation parents. It doesn't use them, so doesn't need to strictly, but anyone writing something using it should be checking if using the parents.

c/tskit/tables.h
```diff
@@ -4619,6 +4624,8 @@ int tsk_table_collection_deduplicate_sites(
     tsk_table_collection_t *tables, tsk_flags_t options);
 int tsk_table_collection_compute_mutation_parents(
     tsk_table_collection_t *self, tsk_flags_t options);
+int tsk_table_collection_compute_mutation_parents_no_integrity_check(
```
jeromekelleher: Why don't we support the TSK_NO_CHECK_INTEGRITY flag on tsk_table_collection_compute_mutation_parents rather than exposing this function in the public API?

benjeffery: Because it takes a `tsk_table_collection_t *` and `tsk_table_collection_check_integrity` has a `const tsk_table_collection_t *`.

benjeffery: But I guess a solution would be to keep `tsk_table_collection_compute_mutation_parents_no_integrity_check` as `static` and add `TSK_NO_CHECK_INTEGRITY` to `tsk_table_collection_compute_mutation_parents` that `tsk_treeseq_init` could use as it has a non-const reference?

benjeffery:
> * Should we sort after computing parents inside `tsk_treeseq_init`?

Hmm, this is more complicated - `compute_mutations_parents` does checks whuile it is working, e.g. line 11081:

```C
/* Reset the mapping for the next site */
            for (j = first_mutation; j < mutation; j++) {
                u = mutations.node[j];
                bottom_mutation[u] = TSK_NULL;
                /* Check that we haven't violated the sortedness property */
                if (mutation_parent[j] > (tsk_id_t) j) {
                    printf("mutation_parent[%d] = %d, j = %d\n",
                           (int) j, (int) mutation_parent[j], (int) j);
                    ret = tsk_trace_error(TSK_ERR_MUTATION_PARENT_AFTER_CHILD);
                    goto out;
                }
            }
```
This is being triggered in python tests where we have set up some mutations, in the wrong order, then when we got o compute the parents, this triggers.

c/tskit/tables.h
benjeffery: I've made this change in 094f301c - seems a lot cleaner, thanks.

c/tskit/tables.c
```diff
@@ -10986,11 +10986,159 @@ tsk_table_collection_check_index_integrity(const tsk_table_collection_t *self)
...
+static int TSK_WARN_UNUSED
+tsk_table_collection_check_mutation_parents(const tsk_table_collection_t *self)
+{
+    int ret = 0;
+    tsk_mutation_table_t mutations = self->mutations;
+    tsk_id_t *new_parents = NULL;
+    tsk_size_t j;
+
+    if (mutations.num_rows == 0) {
+        goto out;
```
jeromekelleher: I think ``return ret`` is better here, reserving ``goto out`` for error handling.

c/tskit/tables.c
```diff
@@ -12346,117 +12502,32 @@ tsk_table_collection_deduplicate_sites(
...
-        if (tj < M) {
-            right = TSK_MIN(right, edges.left[I[tj]]);
-        }
-        if (tk < M) {
-            right = TSK_MIN(right, edges.right[O[tk]]);
+    if (!(options & TSK_NO_CHECK_INTEGRITY)) {
+        /* Set the mutation parent to TSK_NULL so that we don't error
+           in the integrity check on data we're about to overwite.
+        */
+        tsk_memset(
```
jeromekelleher: I don't think we need to do this, and we should't update the data if we error. TSK_CHECK_TREES doesn't imply TSK_CHECK_MUTATION_PARENTS

c/tskit/trees.c
```diff
@@ -458,7 +458,26 @@ tsk_treeseq_init(
...
+            ret = (int) num_trees;
+            goto out;
+        }
+
+        ret = tsk_table_collection_compute_mutation_parents(
+            self->tables, TSK_NO_CHECK_INTEGRITY);
+        if (ret != 0) {
+            goto out;
+        }
+    }
```
jeromekelleher: Missing an "else" here, we're running check_integrity twice for no reason

c/tskit/trees.h
```diff
@@ -71,6 +71,11 @@ when the tree sequence is initialised. Indexes are required for a valid
 tree sequence, and are not built by default for performance reasons.
 */
 #define TSK_TS_INIT_BUILD_INDEXES (1 << 0)
+/**
+If specified, mutation parents in the table collection will be overwritten
+with those computed from the topology when the tree sequence is initialised.
+*/
+#define TSK_TS_INIT_COMPUTE_MUTATION_PARENTS (1 << 1)
```
jeromekelleher, line 78: Does this get automatically incorporated into the docs for the treeseq_init function - I can't remember!

jeromekelleher (commented):
Looks good. Spotted a few thing along the way. 

I think the thing to do now is to get C API users to try this out and see what breaks.

c/tskit/tables.c
benjeffery: Fixed in d8de7931

c/tskit/tables.c
benjeffery: Good point about not changing if we error. Here I was thinking about the fact that `TSK_CHECK_TREES` implies `TSK_CHECK_MUTATION_ORDERING` which checks order based on the existing parent column.

Fixed in 829c9ce6

c/tskit/trees.c
benjeffery: Gah! Good spot. Fixed in 026fff34

c/tskit/trees.h
benjeffery: Yep, just checked and it is there

benjeffery:
So _actually_ not performing a double check makes this work musch faster.

We're now at 0.7s for `main` and 0.9s for this branch!

jeromekelleher:
> We're now at 0.7s for main and 0.9s for this branch!

Excellent - with #3218 we'll get this down to something negligible.

Looks like there's a few error conditions to be hit in the C tests? (Or is this just codecov not updating quickly enough or something?)

benjeffery:
Ah, I'd covered the functionality in Python, but not in C. Fixed now, the only uncovered lines are memory errors.

jeromekelleher:
> @jeromekelleher what's the timeframe for needing feedback on this? Is there urgency here?

No major hurry, we can leave this sit for a few weeks.

molpopgen:
One question:

```
- Add ``TSK_CHECK_MUTATION_PARENTS`` option to ``tsk_table_collection_check_integrity``
  to check that mutation parents are consistent with the tree sequence topology.
  This option implies ``TSK_CHECK_TREES``.
  (:user:`benjeffery`, :issue:`2729`, :issue:`2732`, :pr:`3212`).
```

What does "implies" mean in this context?  Does it mean that if "check mut parents" is 1, then the internal code will also check trees?  Or does it mean that if "check mut parents" is 1, then "check trees MUST ALSO be 1, and therefore the flags in a bitfield are no longer additive?  (I'd have an issue with the latter from a design perspective.)

benjeffery:
> Does it mean that if "check mut parents" is 1, then the internal code will also check trees

It means this.

benjeffery:
That's correct, setting all the parents to `TSK_NULL` will fire the error - but only if there is topology and mutations inconsistent with this.

YunDeng98:
@benjeffery so the converted tree sequence will now need to have the parent annotated for each mutation?

hyanwong:
Perhaps this isn't the right issue to raise this, but I seem to be able to add and delete mutations so as to create a tree sequence in which `compute_mutation_parents()` fails, even after sorting. I wonder if after merging this PR we need to add the correct behaviour to `tables.sort()` to allow this? The main issue is what order to use for mutations that occur on the same node at the same site. We can use time, but in the case of identical times (or if times are unknown) we can presumably use the current mutation order?

```python
import msprime
import tskit
import numpy as np

ts = msprime.sim_mutations(msprime.sim_ancestry(10, sequence_length=100, random_seed=1, recombination_rate=0.01), rate=1, random_seed=1)
print("Simulated", ts.num_mutations, "mutations")

# Add some random mutations, and delete some others
tables = ts.dump_tables()
tables.mutations.time = np.full_like(tables.mutations.time, tskit.UNKNOWN_TIME)
np.random.seed(10)
for s in ts.sites():
    tables.mutations.add_row(site=s.id, node=np.random.randint(ts.num_nodes), derived_state="A")
keep = np.ones(tables.mutations.num_rows, dtype=bool)
keep[0:100] = False
tables.mutations.replace_with(tables.mutations[keep])
# Zap all the parent IDs
tables.mutations.parent = np.full_like(tables.mutations.parent, tskit.NULL)
assert np.all(tables.mutations.parent == tskit.NULL)
tables.sort()
tables.build_index()
tables.compute_mutation_parents()
tables.tree_sequence()  # Fails
```

jeromekelleher:
This seems like an important point @hyanwong, we should keep a separate issue to track

hyanwong:
> This seems like an important point @hyanwong, we should keep a separate issue to track

Great. Issue opened at https://github.com/tskit-dev/tskit/issues/3253

petrelharp:
On my list to do this...

petrelharp:
Belatedly, I have tested this with SLiM, and nothing catastrophic seems to happen.

jeromekelleher (approved):

benjeffery:
> @benjeffery fully fixed now, the mutation functionalities now work with SINGER conversion too!

Great! Thanks for checking it out. I'll merge now.

PR #3221, opened 2025-06-15
Remove tsk_diff_iter_t

### LINKED ISSUE #2797: Remove tsk_diff_iter_t
After #2786 we don't actually use tsk_diff_iter_t at all in the library. As it's not part of the public API (and it causes some annoying problems internally, e.g. [here](https://github.com/tskit-dev/tskit/blob/f7ba5489ae9fa7bede54ad856f181c85f8759f6e/c/tskit/trees.c#L447)) I would vote for removing it.

The only reason not to as far as I can see is that the Rust API currently uses it. The best outcome I think would be for the Rust API to start using the ``tsk_tree_position_t`` class as a replacement, which is probably ok?

The main problem with this I think is that the ``tsk_tree_position_t`` class only supports working from a ``tsk_treeseq_t``, not a ``tsk_table_collection_t``. The main reason for this is that it's useful to have the ``breakpoints`` array around. While we can do without the breakpoints for the simple ``prev`` and ``next`` operations, it would be much more awkward to do the general ``seek`` operations without it. So, I'd be inclined to keep things simple, and just have the ``tsk_tree_position_t`` depend on the tree sequence rather than table collection, but I guess we can imagine a way of just supported in ``next`` and ``prev`` if we don't have the ``breakpoints`` array.

Of course we can also just keep the ``tsk_diff_iter_t`` class.

What do you think @molpopgen?

benjeffery:
Fixes #2797


jeromekelleher (commented):
I don't think we ever documented this, so I'm happy with ditching it.

@molpopgen is the only person I can think of that this would potentially affect. Are you happy for us to ditch this? The functionality is replaced by the much more efficient and flexible tree_iter stuff.

benjeffery:
@molpopgen Suggested removing at https://github.com/tskit-dev/tskit/issues/2797#issuecomment-1634652759

Issue #3135, opened 2025-04-07
New tskit.jit module


jeromekelleher:
We've talked about having a numba support for tskit in various places, so let's add a new optional module "tskit.jit" which provides basic utilities for numba. This would be an optional dependency, and have an entirely separate namespace. E.g

```python
import tskit.jit as tsjit
import numba as nb

@nb.njit
def algorithm(tree_pos)
  while tree_pos.next():
        for j in range(tree_pos.out_range[0], tree_pos.out_range[1]):
            e = tree_pos.edge_removal_order[j]
            # Do something with out edge ID e
        for j in range(tree_pos.in_range[0], tree_pos.in_range[1]):
            e = tree_pos.edge_insertion_order[j]
            # Do something with in edge ID e

tree_pos = tsjit.TreePosition(ts)
algorithm(tree_pos)
```

This should enable a lot of downstream algorithms and reduce code duplication.

jeromekelleher:
So, to be clear, this would not be installed by default and would only be included in the install if we did ``pip install tskit[jit]`` or something similar.

gregorgorjanc:
Is this so you don't depend on numba by default?

jeromekelleher:
Yep. It's a tricky dependency, not needed by downstream things like msprime.

benjeffery:
See also #2778

benjeffery:
Fixed in several PRs e.g. #3225, #3256, #3255

PR #3225, opened 2025-06-18
Initial numba module

benjeffery:
Part of #3135

### DIFF
--- a/python/tests/test_jit.py
+++ b/python/tests/test_jit.py
@@ -1,0 +1,36 @@
+import itertools
+import sys
+from unittest.mock import patch
+
+import pytest
+
+import tests.tsutil as tsutil
+
+
+def test_numba_import_error():
+    # Mock numba as not available
+    with patch.dict(sys.modules, {"numba": None}):
+        with pytest.raises(ImportError, match="pip install numba"):
+            import tskit.jit.numba  # noqa: F401
+
+
+@pytest.mark.numba
+@pytest.mark.parametrize("ts", tsutil.get_example_tree_sequences())
+def test_correct_trees_forward(ts):
+    import tskit.jit.numba as jit_numba
+
+    numba_ts = jit_numba.numba_tree_sequence(ts)
+    in_index = ts.indexes_edge_insertion_order
+    out_index = ts.indexes_edge_removal_order
+    for numba_edge_diff, edge_diff in itertools.zip_longest(
+        numba_ts.edge_diffs(), ts.edge_diffs()
+    ):
+        assert edge_diff.interval == numba_edge_diff.interval
+        for edge_in_index, edge in itertools.zip_longest(
+            range(*numba_edge_diff.edges_in_index_range), edge_diff.edges_in
+        ):
+            assert edge.id == in_index[edge_in_index]
+        for edge_out_index, edge in itertools.zip_longest(
+            range(*numba_edge_diff.edges_out_index_range), edge_diff.edges_out
+        ):
+            assert edge.id == out_index[edge_out_index]
--- a/python/tskit/jit/numba.py
+++ b/python/tskit/jit/numba.py
@@ -1,0 +1,83 @@
+from dataclasses import dataclass
+
+
+try:
+    import numba
+except ImportError:
+    raise ImportError(
+        "Numba is not installed. Please install it with `pip install numba` "
+        "or `conda install numba` to use the tskit.jit.numba module."
+    )
+
+
+# Decorator that makes a jited dataclass by removing certain methods
+# that are not compatible with Numba's JIT compilation.
+def jitdataclass(cls):
+    dc_cls = dataclass(cls, eq=False, match_args=False)
+    del dc_cls.__dataclass_params__
+    del dc_cls.__dataclass_fields__
+    del dc_cls.__repr__
+    del dc_cls.__replace__
+    return numba.experimental.jitclass(dc_cls)
+
+
+@jitdataclass
+class NumbaEdgeDiff:
+    interval: numba.types.UniTuple(numba.float64, 2)
+    edges_in_index_range: numba.types.UniTuple(numba.int32, 2)
+    edges_out_index_range: numba.types.UniTuple(numba.int32, 2)
+
+
+@jitdataclass
+class NumbaTreeSequence:
+    num_edges: numba.int64
+    sequence_length: numba.float64
+    edges_left: numba.float64[:]
+    edges_right: numba.float64[:]
+    indexes_edge_insertion_order: numba.int32[:]
+    indexes_edge_removal_order: numba.int32[:]
+
+    def edge_diffs(self, include_terminal=False):
+        left = 0.0
+        j = 0
+        k = 0
+        edges_left = self.edges_left
+        edges_right = self.edges_right
+        in_order = self.indexes_edge_insertion_order
+        out_order = self.indexes_edge_removal_order
+
+        while j < self.num_edges or left < self.sequence_length:
+            in_start = j
+            out_start = k
+
+            while k < self.num_edges and edges_right[out_order[k]] == left:
+                k += 1
+            while j < self.num_edges and edges_left[in_order[j]] == left:
+                j += 1
+            in_end = j
+            out_end = k
+
+            right = self.sequence_length
+            if j < self.num_edges:
+                right = min(right, edges_left[in_order[j]])
+            if k < self.num_edges:
+                right = min(right, edges_right[out_order[k]])
+
+            yield NumbaEdgeDiff((left, right), (in_start, in_end), (out_start, out_end))
+
+            left = right
+
+        # Handle remaining edges that haven't been processed
+        if include_terminal:
+            yield NumbaEdgeDiff((left, right), (j, j), (k, self.num_edges))
+
+
+def numba_tree_sequence(ts):
+    return NumbaTreeSequence(
+        num_edges=ts.num_edges,
+        sequence_length=ts.sequence_length,
+        edges_left=ts.edges_left,
+        edges_right=ts.edges_right,
+        indexes_edge_insertion_order=ts.indexes_edge_insertion_order,
+        indexes_edge_removal_order=ts.indexes_edge_removal_order,
+    )
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,6 +126,14 @@ jobs:
             python -m pytest -x --cov=tskit --cov-report=xml --cov-branch -n2 --durations=20 tests
           fi
 
+      - name: Run numba tests
+        working-directory: python
+        run: |
+          source ~/.profile
+          conda activate anaconda-client-env
+          pip install numba
+          python -m pytest -x --only-numba-tests --cov=tskit.numba --cov-report=xml --cov-branch -n2 --durations=20 tests
+  
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5.4.0
         with:
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -64,6 +64,12 @@ def pytest_addoption(parser):
         default=False,
         help="To help debugging, draw lines around the plotboxes in SVG output files",
     )
+    parser.addoption(
+        "--only-numba-tests",
+        action="store_true",
+        default=False,
+        help="Only run tests marked with @pytest.mark.numba",
+    )
 
 
 def pytest_configure(config):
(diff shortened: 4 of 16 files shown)


benjeffery:
I've done quite a bit of `numba` investigation and found a way to use dataclasses in numba code. This seems to come at very little performance cost compared to tuples and is a lot nicer. Using a generator also seems to work fine!

benjeffery:
CI Fixed.

Here's some benchmarking with the "coalescent_nodes" method from #2778 on a TS with 12M edges:

Using `ts.edge_diffs`: 23.3s
Calculating edge diffs and coalescent nodes in a single `numba.njit` function: 0.085s
Using the classes here, calculating coalescent nodes in separate client `numba.njit` function: 0.093s

jeromekelleher:
Shall we move the first commit into its own PR? It's cluttering up this one and making it hard to see the real changes.

benjeffery:
> just provide direct access to the edges out and edges in

That's how this code works,
Your code [here](https://github.com/jeromekelleher/sc2ts/blob/7758245c3dc537aeec3b7cd6282241b65f8843dd/sc2ts/jit.py#L212C1-L221C30):

```python
    while tree_pos.next():
        for j in range(tree_pos.out_range[0], tree_pos.out_range[1]):
            e = tree_pos.edge_removal_order[j]
            c = edges_child[e]
            p = edges_parent[e]
            parent[c] = -1
            u = p
            while u != -1:
                num_samples[u] -= num_samples[c]
                u = parent[u]
```
becomes
```python
    for tree_pos in numba_ts.edge_diffs():
        for j in range(*tree_pos.edges_out_index_range):
            e = numba_ts.indexes_edge_removal_order[j]
            c = edges_child[e]
            p = edges_parent[e]
            parent[c] = -1
            u = p
            while u != -1:
                num_samples[u] -= num_samples[c]
                u = parent[u]
```

It is still compiled, and 30% faster!

benjeffery:
> How is it faster then?

Mutating numpy arrays to maintain the state involves the following:
1. Creating a temporary list (build_list).
2. Performing bounds checks for the slice.
3. Copying the data from the list into the array's memory.

Whereas yielding lightweight immuatable objects is much more amenable to numba optimisation. We might be able to get the same gains by using native objects for the state rather than numpy arrays if you are set against iteration.

benjeffery:
New code looks just as fast, proceeding to add some more tests. Will merge this then before doign docs.

hyanwong:
Re docs, eventually we probably want a "high performance" tutorial with some of this stuff, but I can have a stab at that after 1.0. There's some comments here, which might be useful, however: https://github.com/tskit-dev/tutorials/issues/150#issuecomment-980012717

docs/numba.md
```diff
@@ -0,0 +1,150 @@
...
+    samples=50000,
+    sequence_length=100000,
+    recombination_rate=0.1,
+    random_seed=42
+)
+numba_ts = tskit_numba.numba_tree_sequence(ts)
+print(type(numba_ts))
+```
+
+## Tree Traversal
```
hyanwong: I normally think of "tree traversal" as iterating through the tree structure itself. Do you mean "Iterating through trees" here? I can't see any pre/postorder traversal code here.

benjeffery: Good point, I was avoiding the word "iteration" not not confuse it with a Python iterator - but I'll change it back as this is more confusing!

hyanwong: Maybe just "moving between trees"?

python/tskit/jit/numba.py
```diff
@@ -191,199 +49,416 @@ def __init__(self, start, stop, order):
...
-        self.ts = ts
-        self.index = -1
-        self.direction = 0
-        self.interval = (0, 0)
-        self.in_range = NumbaEdgeRange(0, 0, np.zeros(0, dtype=numba.int32))
-        self.out_range = NumbaEdgeRange(0, 0, np.zeros(0, dtype=numba.int32))
-
-    def set_null(self):
+# We cache these as otherwise we'll do a numba re-compile
+# on every new call to `numba_tree_sequence`
```
jeromekelleher: Can you add some implementation notes here please on why the class is declared within the cached function? Is it because of the UnicodeCharSeq  type needing to be parameterised before it can be compiled?

benjeffery: Yes, we need to know the string length to use before compiling the class. I'm working on the diversity calculation, once I know this implemenation works for that I'll add some notes.

python/tests/test_jit.py
```diff
@@ -133,6 +133,91 @@ def coalescent_nodes_python(ts):
...
+    if ts.num_samples < 1:
+        pytest.skip(
+            "Tree sequence must have at least one sample for diversity calculation"
+        )
+
+    import tskit.jit.numba as jit_numba
+
+    NODE_IS_SAMPLE = tskit.NODE_IS_SAMPLE
+
+    @numba.njit
```
jeromekelleher, line 171: Why are importing the jit module within the test functions here and defining the algorithm? I think we can assume that developers have numba installed, and there's pytest ways of skipping the module for CI?

jeromekelleher (commented):
Looks great. Not obvious to me why we're setting up the tests like this though.

python/tests/test_jit.py
benjeffery: Further up the test file we test importing the `tskit.jit.numba` module while it is mocked out, which needs the module to not be imported. I could try and find another way to do that which doesn't require all the local imports?

jeromekelleher: Maybe do that test in its own module so? The local import stuff will confuse people in to thinking it's necessary and the LLMs might also get this idea from the example code.

benjeffery: I've fixed by making the imported module look absent in the test that requires it to be, and have removed the function-local imports.

python/tskit/jit/numba.py
```diff
@@ -0,0 +1,797 @@
...
+        Examples
+        --------
+        >>> tree_index = numba_ts.tree_index()
+        >>> while tree_index.next():
+        ...     # Process current tree at tree_index.index
+        ...     print(f"Tree {tree_index.index}: {tree_index.interval}")
+        """
+        return NumbaTreeIndex(self)
+
+    def diversity(self):
```
jeromekelleher: I don't think this belongs here - ``diversity`` is an example calculation that provide that **uses** the jit wrapper, it's not an API we provide. The jit wrapper only provides functionality for people to write their own algorithms.

python/tskit/jit/numba.py
```diff
@@ -0,0 +1,797 @@
...
+                    tree_index.interval[1] - tree_index.interval[0]
+                )
+
+            return result / self.sequence_length
+
+    if os.environ.get("NUMBA_DISABLE_JIT"):
+        tree_index_spec = []
+    else:
+        tree_index_spec = [
+            ("ts", NumbaTreeSequence.class_type.instance_type),
```
jeromekelleher: Does this class (NumbaTreeIndex) also depend on the size of alleles? Can we shift it outside of the cache?

python/tskit/jit/numba.py
```diff
@@ -0,0 +1,797 @@
...
+                while k >= 0 and mutations_site[k] >= self.site_range[0]:
+                    k -= 1
+                self.mutation_range = (k + 1, old_mutation_left)
+
+            return self.index != -1
+
+    return NumbaTreeSequence
+
+
+def numba_tree_sequence(ts):
```
jeromekelleher: API-wise, I wonder if we're overnaming things here with ``tskit.jit.numba.numba_tree_sequence()`` - it's getting almost Java-esque! How about something like
```python
nts = tskit.numba.jitwrap(ts) # Returns a NumbaTreeSequence
```

"jitwrap" is descriptive and feels like it's something that could in principle be extended to other JIT engines

tree_index = jts.tree_index()

python/tests/test_jit.py
```diff
@@ -0,0 +1,278 @@
...
+            "Tree sequence must have at least one sample for diversity calculation"
+        )
+
+    numba_ts = jit_numba.numba_tree_sequence(ts)
+    diversity_numba = numba_ts.diversity()
+    diversity_python = ts.diversity(mode="branch")
+
+    assert diversity_numba == pytest.approx(diversity_python, rel=1e-5)
+
+
```
jeromekelleher, line 252: Should do this as a parametrize over the examples as well

python/tests/test_jit.py
```diff
@@ -0,0 +1,278 @@
...
+    np.testing.assert_array_equal(
+        numba_ts.indexes_edge_insertion_order, ts.indexes_edge_insertion_order
+    )
+    np.testing.assert_array_equal(
+        numba_ts.indexes_edge_removal_order, ts.indexes_edge_removal_order
+    )
+    assert numba_ts.indexes_edge_insertion_order.dtype == np.int32
+    assert numba_ts.indexes_edge_removal_order.dtype == np.int32
+    assert numba_ts.breakpoints.dtype == np.float64
+    np.testing.assert_array_equal(numba_ts.breakpoints, ts.breakpoints(as_array=True))
```
jeromekelleher: We're not testing the string fields here - what about sites_ancestral_state etc?

python/tests/test_jit.py
```diff
@@ -0,0 +1,278 @@
...
+
+def test_numba_tree_sequence_properties(ts_fixture):
+    ts = ts_fixture
+
+    numba_ts = jit_numba.numba_tree_sequence(ts)
+
+    assert numba_ts.num_trees == ts.num_trees
+    assert numba_ts.num_edges == ts.num_edges
+    assert numba_ts.sequence_length == ts.sequence_length
+    np.testing.assert_array_equal(numba_ts.edges_left, ts.edges_left)
```
jeromekelleher: usually use ``import numpy.testing as nt; nt.assert_array_equal()```

docs/numba.md
```diff
@@ -0,0 +1,151 @@
...
+```{code-cell} python
+t = time.time()
+python_num_edges = []
+for tree in ts.trees():
+    python_num_edges.append(tree.num_edges)
+print(f"Normal Time taken: {time.time() - t:.4f} seconds")
+
+assert jit_num_edges == python_num_edges, "JIT and normal results do not match!"
+```
+
```
jeromekelleher, line 136: I think we want to include the ``diversity`` calculation here in the documentation - we want to have a real example that people can use as a starting point.

jeromekelleher (commented):
Some more comments

python/tskit/jit/numba.py
benjeffery: Yes, sorry forgot to move it back as it was easier to develop in the class.

Moved to tests and docs in https://github.com/tskit-dev/tskit/pull/3225/commits/a17260684236a6a466c0c5a3c0150ef348578719

python/tskit/jit/numba.py
benjeffery: This class has to hold a reference to the tree sequence, so numba has to be told the concrete type of NumbaTreeSequence, which we only know once allele lengths are known.

python/tskit/jit/numba.py
benjeffery: Renamed to jitwrap in https://github.com/tskit-dev/tskit/pull/3225/commits/de1d80413d141314c62c244706889107de58acaf

python/tests/test_jit.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3225/commits/a92e60768263bd9e9bf977e9acfc3a920b633d01

python/tests/test_jit.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3225/commits/29759d04f7c85ba13941c9f2eae355aecd8282b7

python/tests/test_jit.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3225/commits/6686ad0130df56bf2931d9843e99d307e79e01b0

docs/numba.md
benjeffery: Done in https://github.com/tskit-dev/tskit/pull/3225/commits/a17260684236a6a466c0c5a3c0150ef348578719

benjeffery:
> I'm still uncomfortable with the code duplication

I've figured out a solution to this in https://github.com/tskit-dev/tskit/pull/3225/commits/41f59274a19f6a52f4317dc2468f4f36974dc9b9

I've addressed the rest of the comments

jeromekelleher (approved):
Fantastic! Let's this much in and build on it.

Issue #2632, opened 2022-11-08
Numpy access to ancestral and derived state columns in TreeSequence


jeromekelleher:
It would be really nice if we could access the ancestral and derived states as numpy string arrays. In the common single-allele case, we can easily do this:
```python
ancestral_state = tables.sites.ancestral_state.view("S1")
Out[64]: array([b'T', b'G', b'A', ..., b'T', b'T', b'A'], dtype='|S1')
```
However, this is a byte array and "not recommended". I guess we should try to return the values as proper numpy ``str`` types?

Can easily do this but it makes a copy (I assume?)
```python
derived_state.astype(str)
Out[63]:
array(['T', 'G', 'A', ..., 'T', 'T', 'A'], dtype='<U1')
```

I haven't managed to make a direct view of the memory that gets treated as a unicode string.

As a first pass we could just error out of the states aren't all single chars (ascii chars even), and I guess we could just copy the data in a way that's numpy string friendly when we have variable length alleles? I guess it would be good to know if this is possible before painting ourselves into a corner.

Related: #2631

benjeffery:
` tables.sites.ancestral_state.view("S1").astype(str, copy=False)` works for me, and I think avoids the copy?

jeromekelleher:
No, this doesn't work unfortunately - I think ``copy=False`` is a request that numpy does its best to fulfil, when possible. I just tried it out there and the OWNDATA flag is false when you do ``.astype(str, copy=False)``.

I don't get why the view ``dtype=U1`` doesn't work though - aren't these UTF-8 encoded single byte strings?

benjeffery:
> I don't get why the view `dtype=U1` doesn't work though - aren't these UTF-8 encoded single byte strings?

`numpy.dtype("U1").itemsize` is `4` as there has to be room for the largest, 4-byte unicode character.

`numpy.dtype("S1").itemsize` is `1`. So I don't think we can get to a unicode dtype without a copy or some low-level hacking.

jeromekelleher:
Note that the current plan for this is to make a numpy stringdype in C

PR #3228, opened 2025-06-19
WIP - decode ragged string columns

benjeffery:
A start on exposing the ragged columns as arrays #2632

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -10774,6 +10774,118 @@ TreeSequence_make_array(TreeSequence *self, tsk_size_t size, int dtype, void *da
     return make_owned_array((PyObject *) self, size, dtype, data);
 }
 
+PyObject *
+TreeSequence_decode_ragged_string_column(
+    TreeSequence *self, tsk_size_t num_rows, const char *data, const tsk_size_t *offset)
+{
+    PyObject *ret = NULL;
+    PyArrayObject *ret_array = NULL;
+    npy_intp dims[1];
+    char *string_data = NULL;
+    tsk_size_t i;
+    tsk_size_t max_length;
+    tsk_size_t len;
+    tsk_size_t start;
+    tsk_size_t end;
+    char *dest;
+    bool all_length_one;
+
+    dims[0] = (npy_intp) num_rows;
+
+    /* If all strings are length one we can back the array with tskit memory */
+    all_length_one = true;
+    for (i = 0; i < num_rows; i++) {
+        len = offset[i + 1] - offset[i];
+        if (len != 1) {
+            all_length_one = false;
+            break;
+        }
+    }
+
+    if (all_length_one && num_rows > 0) {
+        ret_array = (PyArrayObject *) PyArray_New(&PyArray_Type, /* subtype */
+            1,                                                   /* nd */
+            dims,                                                /* dims */
+            NPY_STRING,                                          /* type_num */
+            NULL,                                                /* strides */
+            (void *) data,                                       /* data */
+            1,   /* itemsize = 1 for S1 */
+            0,   /* flags */
+            NULL /* obj */
+        );
+        if (ret_array == NULL) {
+            goto out;
+        }
+
+        if (PyArray_SetBaseObject(ret_array, (PyObject *) self) != 0) {
+            goto out;
+        }
+        Py_INCREF(self);
+
+    } else {
+        /* We have to pad the strings to make a rectangular array */
+        max_length = 0;
+        for (i = 0; i < num_rows; i++) {
+            len = offset[i + 1] - offset[i];
+            if (len > max_length) {
+                max_length = len;
+            }
+        }
+
+        /* Ensure at least S1 */
+        if (max_length == 0) {
+            max_length = 1;
+        }
+
+        if (num_rows > 0) {
+            tsk_size_t total_size = num_rows * max_length;
+            string_data = PyMem_Calloc(total_size, sizeof(char));
+            if (string_data == NULL) {
+                PyErr_NoMemory();
+                goto out;
+            }
+
+            for (i = 0; i < num_rows; i++) {
+                start = offset[i];
+                end = offset[i + 1];
+                len = end - start;
+
+                dest = string_data + (i * max_length);
+                if (len > 0) {
+                    memcpy(dest, data + start, len);
+                }
+                /* PyMem_Calloc zero-fills so we don't need to pad */
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


benjeffery:
Would appreciate a check over the tricky Python C API stuff here.

jeromekelleher:
I think this entails something like this:
```
allocator = NpyString_acquire_allocator(...)
for (j = 0; j < num_records; j++)
    NpyString_pack(allocator, packed_string, data[offset[j]], offset[j+ 1] - offset[j])
return packed_string
```

I haven't used this API yet though, so haven't seen good examples of it in practise. Maybe that would be a good first step - let's track down an idiomatic example of it in use that we can copy?

benjeffery:
I searched github and couldn't find a single example of how to do this, but it wasn't too hard to figure out. 
We're now pinning to numpy>2.

python/_tskitmodule.c
```diff
@@ -10774,6 +10775,62 @@ TreeSequence_make_array(TreeSequence *self, tsk_size_t size, int dtype, void *da
...
+        pack_result = NpyString_pack(allocator,
+            (npy_packed_static_string
+                    *) (array_data + (i * ((PyArray_Descr *) string_dtype)->elsize)),
+            data + offset[i], offset[i + 1] - offset[i]);
+        if (pack_result == -1) {
+            goto out;
+        }
+    }
+
+    if (PyArray_SetBaseObject((PyArrayObject *) array, (PyObject *) self) != 0) {
```
jeromekelleher: Why are we setting the base object and increffing self here? I thought the allocator was just creating a new copy of the string data so there's no need for this?

python/_tskitmodule.c
```diff
@@ -10774,6 +10775,62 @@ TreeSequence_make_array(TreeSequence *self, tsk_size_t size, int dtype, void *da
...
+    PyArray_CLEARFLAGS((PyArrayObject *) array, NPY_ARRAY_WRITEABLE);
+
+    ret = array;
+    array = NULL;
+
+out:
+    if (allocator != NULL) {
+        NpyString_release_allocator(allocator);
+    }
+    Py_XDECREF(array);
```
jeromekelleher, line 10865: I think we probably need to decref string_dtype also?

jeromekelleher (commented):
Cool!

Will it be possible to conditionally compile this if we're on numpy 2? I had hoped we could do this, as keeping numpy 1 support for a while longer would be good.

python/_tskitmodule.c
benjeffery: Good spot - forgot to remove it!

python/_tskitmodule.c
benjeffery: I originally had that, and it segfaults with it. The docs don't state the type of reference `PyArray_DescrFromType` returns.

benjeffery:
How's this approach for conditional compilation? 
Will probably need a separate CI test, as my attempt here to add it to the existing one has hit issues with python versions.

python/_tskitmodule.c
```diff
@@ -24,15 +24,21 @@
...
+#include <numpy/numpyconfig.h>
+
+#if defined(NPY_2_0_API_VERSION) && NPY_API_VERSION >= NPY_2_0_API_VERSION
+#define NPY_NO_DEPRECATED_API NPY_2_0_API_VERSION
+#undef NPY_FEATURE_VERSION
+#define NPY_FEATURE_VERSION NPY_2_0_API_VERSION
+#define HAVE_NUMPY_2 1
+#else
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
-#define TSK_BUG_ASSERT_MESSAGE                                                          \
```
jeromekelleher, line 28: We still want this message don't we?

jeromekelleher (commented):
Looks pretty clean, yeah. There's a bunch of subtleties though and I'd have to check the code out and poke at it to really understand

python/_tskitmodule.c
benjeffery: Whoops, fixed.

benjeffery:
Some more details about the new string type:

1- Main array is struct of (length, pointer)
2 - Short string optimisation stores string directly in the place of the pointer
3 - Arena allocator means that strings are contiguous on the heap.

python/CHANGELOG.rst
```diff
@@ -19,6 +19,11 @@
 - Add ``TreeSequence.mutations_edge`` which returns the edge ID for each mutation's
```
jeromekelleher, line 19: Let's add a big warning here that we're moving to numpy 2.0 by default, along with the errors you might get and what to do. For example, what happens when you have older msprime, kastore etc lying around that are compiled againsd numpy 1.0?

python/tskit/trees.py
```diff
@@ -5952,6 +5954,26 @@ def sites_position(self):
         """
         return self._sites_position
 
+    @property
+    def sites_ancestral_state(self):
+        """
+        The ``ancestral_state`` column in the
+        :ref:`sec_site_table_definition` as a numpy array (dtype=StringDtype).
+        """
+        if self._sites_ancestral_state is None:
```
jeromekelleher, line 5967: This feels overly complicated. Why not:
```
if not not _tskit.HAS_NUMPY_2:
    raise RuntimeError(...)
if self._sites_ancestral_state is None:
     self._sites_ancestral_state = ...
return self._sites_ancestral_state
```
It doesn't look like there's any additional information in the ``raise`` path, and you may as well just let the error happen. Is the performance cost of looking up `` _tskit.HAS_NUMPY_2`` a concern or something?

jeromekelleher (commented):
I've had a close look at this, and gone through the numpy 2.0 code to see how it works. I'm happy this is correct and to merge (modulo on minor nit). I added a few tests also.

python/tskit/trees.py
benjeffery: My thinking her was not to mask errors that weren't due to numpy 1, but your way is a cleaner way of doing that so happy to switch.

jeromekelleher: Let's switch so

benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3228/commits/c7ae977974ea1a3bf8fc295a0dbd985fa21e764a

python/CHANGELOG.rst
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3228/commits/13a08a0bd9df82b01ffb09918fe5d35b90d84c21

jeromekelleher (approved):

PR #3235, opened 2025-06-26
Fix genetic_relatedness to allow single sample set with self-comparisons

### LINKED ISSUE #3055: `genetic_relatedness` thinks it needs 2 samples sets but is valid with only one
Self-comparisons are valid for this statistic, and so this should work:
```
>> ts.genetic_relatedness([[0]], indexes=[(0,0)])
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.12/site-packages/tskit/trees.py", line 8403, in genetic_relatedness
    out = self.__k_way_sample_set_stat(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.12/site-packages/tskit/trees.py", line 7843, in __k_way_sample_set_stat
    stat = self.__run_windowed_stat(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.12/site-packages/tskit/trees.py", line 7664, in __run_windowed_stat
    stat = method(*args, **kwargs, windows=windows)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_tskit.LibraryError: Insufficient sample sets provided. (TSK_ERR_INSUFFICIENT_SAMPLE_SETS)
```

andrewkern:
## Description

This PR fixes issue #3055 where `genetic_relatedness` would fail when computing self-comparisons with a single sample set. As @petrelharp pointed out there is no reason this shouldn't be allowed for certain statistics, including `genetic_relatedness`

### The Problem

Previously, calling `genetic_relatedness` with a single sample set and indexes referring to that set would raise an error:

```python
ts.genetic_relatedness([[0]], indexes=[(0,0)])
# TSK_ERR_INSUFFICIENT_SAMPLE_SETS: Insufficient sample sets provided
```

This occurred because the underlying C API validates that at least k=2 sample sets are provided for k-way statistics, even when the indexes only reference a subset of those sets.

### This PR

The fix adds an `allow_self_comparisons` parameter to the internal `__k_way_sample_set_stat` method, which is only set to `True` for `genetic_relatedness`. When enabled:

1. The method validates that all indexes reference valid sample sets
2. If fewer than k sample sets are provided, it pads the list with dummy sample sets to satisfy the C API requirement
3. The dummy sets are never accessed during computation since only the sets referenced by the indexes are used

This ensures that:
- Only `genetic_relatedness` behavior is changed
- Other statistics (like Fst) continue to enforce the minimum sample set requirement
- The C API contract is satisfied without modifying the C code

Further, this PR sets the stage for allowing for other appropriate k-way stats to be computed with self comparisons by setting a flag. 

## Testing

Added comprehensive tests for all three computation modes (site, branch, node) to verify:
- Single sample set with self-comparison works correctly
- Multiple samples within a single set work correctly  
- The fix doesn't affect other statistics

All existing tests continue to pass, but @petrelharp this would benefit from you specifically looking over what I've done to be sure you're okay with the way I've implemented `allow_self_comparisons`

Fixes #3055

# PR Checklist:

- [] Changelogs, if there are API changes.

I'm not sure if I should touch the changelogs here?

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -4515,7 +4515,7 @@ check_sample_stat_inputs(tsk_size_t num_sample_sets, tsk_size_t tuple_size,
 {
     int ret = 0;
 
-    if (num_sample_sets < tuple_size) {
+    if (num_sample_sets < 1) {
         ret = tsk_trace_error(TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
         goto out;
     }
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -881,9 +881,6 @@ verify_two_way_stat_func_errors(
     ret = method(ts, 0, sample_set_sizes, samples, 1, set_indexes, 0, NULL,
         options | TSK_STAT_SITE, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
-    ret = method(ts, 1, sample_set_sizes, samples, 1, set_indexes, 0, NULL,
-        options | TSK_STAT_SITE, &result);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
 
     ret = method(ts, 2, sample_set_sizes, samples, 0, set_indexes, 0, NULL,
         options | TSK_STAT_SITE, &result);
@@ -938,12 +935,6 @@ verify_three_way_stat_func_errors(tsk_treeseq_t *ts, general_sample_stat_method 
     ret = method(ts, 0, sample_set_sizes, samples, 1, set_indexes, 0, NULL,
         TSK_STAT_SITE, &result);
     CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
-    ret = method(ts, 1, sample_set_sizes, samples, 1, set_indexes, 0, NULL,
-        TSK_STAT_SITE, &result);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
-    ret = method(ts, 2, sample_set_sizes, samples, 1, set_indexes, 0, NULL,
-        TSK_STAT_SITE, &result);
-    CU_ASSERT_EQUAL_FATAL(ret, TSK_ERR_INSUFFICIENT_SAMPLE_SETS);
 
     ret = method(ts, 3, sample_set_sizes, samples, 0, set_indexes, 0, NULL,
         TSK_STAT_SITE, &result);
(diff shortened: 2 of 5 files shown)


petrelharp:
Let's see. First, it's quite easy to [modify the checking code that produces the error](https://github.com/tskit-dev/tskit/blob/9821725e4706f61baa005e3def6fad354083973c/c/tskit/trees.c#L4513). I went ahead and did that (see this commit: 57ad49f7046885).

However, one *can* compute 'self' comparisons for other statistics, for instance, this works:
```
ts.divergence([[0], [1]], indexes=[(0, 0)])
```
or this
```
ts.f4([[0],[1],[2],[3]], indexes=[(0,0,0,0)])
```
These are perfectly valid things to compute, just the thought was they are probably not telling someone what they actually want to know. So, if the intention behind the restriction for "at least k sample sets for a k-way stat" was to keep people from computing things they probably didn't want to compute, then we're not actually doing that. We could require indexes be unique, but I don't think we should actually do that - for instance, `f4(..., indexes=[(0,0,1,1)])` is actually a version of `f2( )` that people compute sometimes (I think, or maybe it's `(0, 1, 0, 1)`). So, in the next commit I've just removed the requirement that there be at least `k` sample sets - instead, requiring there be at least 1 sample set in all cases.

How's that sound, @andrewkern?

c/tskit/trees.c
```diff
@@ -4624,6 +4624,7 @@ tsk_treeseq_genetic_relatedness(const tsk_treeseq_t *self, tsk_size_t num_sample
     const double *windows, tsk_flags_t options, double *result)
 {
     int ret = 0;
+    // we want to allow self comparisons
```
andrewkern: maybe expand this comment to explain the `2` below? also, maybe we want the comment to be C-style rather than C++, i.e. `/* */

c/tskit/trees.c
```diff
@@ -4515,7 +4515,7 @@ check_sample_stat_inputs(tsk_size_t num_sample_sets, tsk_size_t tuple_size,
 {
     int ret = 0;
 
-    if (num_sample_sets < tuple_size) {
+    if (num_sample_sets < 1) {
```
andrewkern, line 4518: maybe add a comment here to explain the check?

petrelharp: It seems pretty self-explanatory to me? Like, how can you compute a statistic with only one sample set? (The comment would be something like `// check there is at least one sample set`, but we like to avoid this sort of comment that just says the same thing as the code?)

c/tskit/trees.c
petrelharp: oh whoops, this is outdated, since the `2` is the tuple size; there's nothing in this code now that explicitly allows self comparisons. I'll remove it.
```suggestion
```

PR #3243, opened 2025-07-25
C and Python API for two-way two-locus stats

lkirk:
This PR implements the C and Python API for computing two-way two-locus statistics. The algorithm is identical to the python version, except during testing I uncovered a small issue with normalisation. We need to handle the case where sample sets are of different sizes. The fix for this was to average the normalisation factor for each sample set.

Test coverage has been added to cover C, low-level python and some high-level tests.

### DIFF
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -2637,12 +2637,15 @@ test_paper_ex_two_site(void)
 
     tsk_treeseq_from_text(&ts, 10, paper_ex_nodes, paper_ex_edges, NULL, paper_ex_sites,
         paper_ex_mutations, paper_ex_individuals, NULL, 0);
+    double truth_three_index_tuples[27] = { 1, 1, NAN, 0.1111111111111111,
+        0.1111111111111111, NAN, 0.1111111111111111, 0.1111111111111111, NAN,
+        0.1111111111111111, 0.1111111111111111, NAN, 1, 1, 1, 1, 1, 1,
+        0.1111111111111111, 0.1111111111111111, NAN, 1, 1, 1, 1, 1, 1 };
 
-    tsk_size_t sample_set_sizes[3];
-    tsk_id_t sample_sets[ts.num_samples * 3];
+    tsk_size_t sample_set_sizes[3], num_index_tuples;
+    tsk_id_t sample_sets[ts.num_samples * 3], index_tuples[2 * 3] = { 0, 1, 0, 0, 0, 2 };
     tsk_size_t num_sites = ts.tables->sites.num_rows;
-    tsk_id_t *row_sites = tsk_malloc(num_sites * sizeof(*row_sites));
-    tsk_id_t *col_sites = tsk_malloc(num_sites * sizeof(*col_sites));
+    tsk_id_t *sites = tsk_malloc(num_sites * sizeof(*sites));
 
     // First sample set contains all of the samples
     sample_set_sizes[0] = ts.num_samples;
@@ -2651,14 +2654,13 @@ test_paper_ex_two_site(void)
         sample_sets[s] = (tsk_id_t) s;
     }
     for (s = 0; s < num_sites; s++) {
-        row_sites[s] = (tsk_id_t) s;
-        col_sites[s] = (tsk_id_t) s;
+        sites[s] = (tsk_id_t) s;
     }
 
     result_size = num_sites * num_sites;
     tsk_memset(result, 0, sizeof(*result) * result_size * num_sample_sets);
     ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, num_sites,
-        row_sites, NULL, num_sites, col_sites, NULL, 0, result);
+        sites, NULL, num_sites, sites, NULL, 0, result);
 
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     assert_arrays_almost_equal(result_size * num_sample_sets, result, truth_one_set);
@@ -2672,7 +2674,7 @@ test_paper_ex_two_site(void)
 
     tsk_memset(result, 0, sizeof(*result) * result_size * num_sample_sets);
     ret = tsk_treeseq_r2(&ts, num_sample_sets, sample_set_sizes, sample_sets, num_sites,
-        row_sites, NULL, num_sites, col_sites, NULL, 0, result);
+        sites, NULL, num_sites, sites, NULL, 0, result);
 
     CU_ASSERT_EQUAL_FATAL(ret, 0);
     assert_arrays_almost_equal(result_size * num_sample_sets, result, truth_two_sets);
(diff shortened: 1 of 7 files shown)


python/tests/test_ld_matrix.py
```diff
@@ -244,6 +244,41 @@ def norm_hap_weighted(
...
+    set_indexes = params["set_indexes"]
+
+    for k in range(result_dim):
+        i = set_indexes[k][0]
+        j = set_indexes[k][1]
+        ni = sample_set_sizes[i]
+        nj = sample_set_sizes[j]
+        wAB_i = hap_weights[0, i]
+        wAB_j = hap_weights[0, j]
+        result[k] = (wAB_i / ni / 2) + (wAB_j / nj / 2)
```
apragsdale: Based on our discussion earlier this week, I think it's more natural to norm by the total weights and sample sizes, rather than the average. (So: `(wAB_i + w_AB_j) / (ni + nj)`)

c/tskit/trees.c
```diff
@@ -4781,6 +4802,201 @@ tsk_treeseq_f2(const tsk_treeseq_t *self, tsk_size_t num_sample_sets,
...
+        state_row = GET_2D_ROW(state, 3, j);
+        p_AB = state_row[0] / n;
+        p_Ab = state_row[1] / n;
+        p_aB = state_row[2] / n;
+        p_A = p_AB + p_Ab;
+        p_B = p_AB + p_aB;
+        D_j = p_AB - (p_A * p_B);
+        denom_j = sqrt(p_A * p_B * (1 - p_A) * (1 - p_B));
+
+        result[k] = (D_i * D_j) / (denom_i * denom_j);
```
apragsdale, line 4973: This will spuriously produce a NAN if the allele is not variable in one of the two sample sets. From our discussion earlier this week, I think we can calculate `result = (D_i * D_j) / denom`, where `denom = pA * (1 - pA) * pB * (1 - pB)`, and `pA` is `(pA_i + pA_j) / (n_i + n_j)` and likewise for `pB`.

apragsdale (commented):
Hi @lkirk - thanks for starting this PR. I have a few suggestions for the multi-pop stat calculations and the normalization here.

python/tests/test_ld_matrix.py
lkirk: I remember our conversation about this, I've incorporated the change.

c/tskit/trees.c
lkirk: I'm not sure we can combine the allele frequencies this way. Consider this example:

```python
>>> Ai = 0.3; Bi = 0.5; Aj = 0.7; Bj = 0.2; ni = 37; nj = 39
>>> A = (Ai + Aj) / (ni + nj)
>>> B = (Bi + Bj) / (ni + nj)
>>> A * (1-A) * B * (1-B)
0.00011849496867350617
>>> (Ai*(1-Ai)*Bi*(1-Bi)) * (Aj*(1-Aj)*Bj*(1-Bj))
0.0017640000000000006
```

python/tests/test_ld_matrix.py
```diff
@@ -1001,18 +1035,18 @@ def r2_ij_summary_func(
         i = set_indexes[k][0]
         j = set_indexes[k][1]
         n = sample_set_sizes[i]
-        p_AB = state[0, k] / n
-        p_Ab = state[1, k] / n
-        p_aB = state[2, k] / n
+        p_AB = state[0, i] / n
+        p_Ab = state[1, i] / n
+        p_aB = state[2, i] / n
```
lkirk: This fixes an error in the original implementation. The previous tests were rather simple and did not catch the issue.

c/tests/test_stats.c
```diff
@@ -2788,35 +2824,70 @@ test_paper_ex_two_branch(void)
...
+
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    assert_arrays_almost_equal_nan(
+        result_size * num_index_tuples, result, truth_three_index_tuples);
+
     tsk_treeseq_free(&ts);
-    tsk_safe_free(row_positions);
-    tsk_safe_free(col_positions);
+    tsk_safe_free(positions);
+    tsk_safe_free(positions);
```
jeromekelleher: Double free here

python/_tskitmodule.c
```diff
@@ -10659,6 +10659,138 @@ TreeSequence_Dz_unbiased_matrix(TreeSequence *self, PyObject *args, PyObject *kw
...
 
+static PyObject *
+TreeSequence_k_way_ld_matrix(TreeSequence *self, PyObject *args, PyObject *kwds,
+    npy_intp tuple_size, k_way_two_locus_count_stat_method *method)
+{
+    PyObject *ret = NULL;
+    static char *kwlist[] = { "sample_set_sizes", "sample_sets", "indexes", "row_sites",
+        "col_sites", "row_positions", "column_positions", "mode", NULL };
+
+    PyObject *row_sites = NULL, *col_sites = NULL, *row_positions = NULL,
```
jeromekelleher: Can you declare these variables separately one per line please, like
```
 PyObject *row_sites = NULL;
 PyObject *col_sites = NULL;
// etc
```
It's how it's done in the rest of the code and makes it easier for mass search and replace etc later on.

jeromekelleher: It's a way of saying "watch out for this variable, it has tricky state" vs just standard declarations.

c/tskit/trees.c
apragsdale: Sorry, I miswrote that. I meant `pA` is `(nA_i + nA_j) / (n_i + n_j)`

lkirk: Okay, I added this change. I also added some test coverage for unequal sample set sizes, so that we can see the results of this change and how the stats produced vary -- it can be quite a bit in some cases.

c/tests/test_stats.c
lkirk: I'm surprised valgrind didn't catch this, fixed.

python/_tskitmodule.c
lkirk: I fixed this in the ld_matrix method above as well, it didn't follow that style either.

lkirk:
Rebased and squashed.

c/tests/test_stats.c
jeromekelleher: I think it's safe given how tsk_safe_free works, but no point in having it in there

jeromekelleher:
I think we're just waiting for your thumbs-up on the stats here @petrelharp - I'm happy with the implementation

lkirk:
@jeromekelleher: @apragsdale and I have decided to defer the changes to the summary function so that we don't hold up this PR. I've reverted the stat to its original formulation, which is the more standard representation, though it is more sensitive to missing data.

I'll give an update on slack tomorrow about the overall status of the two-locus work.

PR #3246, opened 2025-08-05
Fix metadata schemas on assert_equals

### LINKED ISSUE #3244: here are node tables that are not `equals( )` and yet have no detectable differences
This tree sequence:  [ts1.zip](https://github.com/user-attachments/files/21539566/ts1.zip) and this one: [ts2.zip](https://github.com/user-attachments/files/21539568/ts2.zip) (note, these are not tszipped despite the suffix)  have node tables that if you do
`tsa.tables.nodes.assert_equals(tb.tables.nodes)` then you get:
```
  File "<python-input-28>", line 1, in <module>
    ta.assert_equals(tb)
    ~~~~~~~~~~~~~~~~^^^^
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.13/site-packages/tskit/tables.py", line 465, in assert_equals
    raise AssertionError(
    ^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
    )  # pragma: no cover
    ^
AssertionError: Tables differ in an undetected way - this is a bug, please report an issue on gitub
```
Indeed, examining their `flags`, `population`, `individual`, `time`, `metadata`, and `metadata_offset` vectors and their `metadata_schema` individually says they're all identical. The way [this error comes up](https://github.com/tskit-dev/tskit/blob/c23aafcfc3bd49bc41955a3f509f50d0fbf4b840/python/tskit/tables.py#L2412)  is if the lower-level tree sequence `equals` method returns `False` but then we can't find the difference in python.

The problem seems to be in metadata, since doing
```
>>> tsa.tables.assert_equals(tsb.tables, ignore_metadata=True)
```
does not produce this error.

benjeffery:
Fixes #3244

### DIFF
--- a/python/tests/test_tables.py
+++ b/python/tests/test_tables.py
@@ -1401,6 +1401,21 @@ class AssertEqualsMixin:
                 table_5row.assert_equals(table2)
             table_5row.assert_equals(table2, ignore_metadata=True)
 
+    def test_degenerate_metadata_schema(self, table_5row):
+        if hasattr(table_5row, "metadata_schema"):
+            table_5row.clear()
+            table1 = table_5row.copy()
+            table2 = table_5row.copy()
+            table1.ll_table.metadata_schema = (
+                '{"codec": "json", "properties": '
+                '{"A": {"type": "integer"}, "B": {"type": "number"}}}'
+            )
+            table2.ll_table.metadata_schema = (
+                '{"codec": "json", "properties": '
+                '{"B": {"type": "number"}, "A": {"type": "integer"}}}'
+            )
+            table1.assert_equals(table2)
+
     def test_row_changes(self, table_5row, test_rows):
         for column_name in test_rows[0].keys():
             table2 = self.table_class()
@@ -4216,6 +4231,17 @@ class TestTableCollectionAssertEquals:
         ):
             t1.assert_equals(t2)
         t1.assert_equals(t2, ignore_reference_sequence=True)
+
+    def test_degenerate_metadata_schema(self, t1, t2):
+        t1._ll_object.metadata_schema = (
+            '{"codec": "json", "properties": '
+            '{"A": {"type": "integer"}, "B": {"type": "number"}}}'
+        )
+        t2._ll_object.metadata_schema = (
+            '{"codec": "json", "properties": '
+            '{"B": {"type": "number"}, "A": {"type": "integer"}}}'
+        )
+        t1.assert_equals(t2)
 
 
 class TestTableCollectionMethodSignatures:
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -462,9 +462,18 @@ class BaseTable:
                 f"other={other.num_rows}"
             )
 
+        # We can reach this point if the metadata schemas byte representations
+        # differ when the decoded schema is the same
+        if (
+            not ignore_metadata
+            and self.ll_table.metadata_schema != other.ll_table.metadata_schema
+            and self.metadata_schema == other.metadata_schema
+        ):
+            return
+
         raise AssertionError(
             "Tables differ in an undetected way - "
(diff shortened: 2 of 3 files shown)


python/tests/test_tables.py
```diff
@@ -1401,6 +1401,21 @@ def test_metadata_schema(self, table_5row):
                 table_5row.assert_equals(table2)
             table_5row.assert_equals(table2, ignore_metadata=True)
 
+    def test_degenerate_metadata_schema(self, table_5row):
+        if hasattr(table_5row, "metadata_schema"):
+            table_5row.clear()
```
jeromekelleher, line 1406: Not sure we should be mutating the global fixture here? Better to copy first then clear?

Also, why are we checking if the fixture has a metadata_schema?

python/tests/test_tables.py
```diff
@@ -4217,6 +4232,17 @@ def test_ignore_reference_sequence(self, t1, t2):
             t1.assert_equals(t2)
         t1.assert_equals(t2, ignore_reference_sequence=True)
 
+    def test_degenerate_metadata_schema(self, t1, t2):
+        t1._ll_object.metadata_schema = (
```
jeromekelleher, line 4236: Likewise on mutating parameters?

jeromekelleher (commented):
Logic makes sense, but tests seem dodgy?

python/tests/test_tables.py
benjeffery: The default scope for a fixture is `function` so each test gets a fresh instance. 
We're checking for the schema as this test suite is run against all tables, which includes `ProvenanceTable`.

python/tests/test_tables.py
benjeffery: Again, the instance is per-test.

jeromekelleher (approved):
LGTM, thanks for clarifications!

python/tskit/tables.py
```diff
@@ -462,9 +462,17 @@ def assert_equals(self, other, *, ignore_metadata=False):
                 f"other={other.num_rows}"
             )
 
+        # We can reach this point if the metadata schemas byte representations
+        # differ when the decoded schema is the same
+        if (
+            not ignore_metadata
+            and self.ll_table.metadata_schema != other.ll_table.metadata_schema
```
petrelharp, line 469: I know it has this above but just to make it totally clear that this is exactly what we're testing for here
```suggestion
            and self.metadata_schema == other.metadata_schema
            and self.ll_table.metadata_schema != other.ll_table.metadata_schema
```

python/tskit/tables.py
```diff
@@ -3277,9 +3285,17 @@ def assert_equals(
                 other.provenances, ignore_timestamps=ignore_timestamps
             )
 
+        # We can reach this point if the metadata schemas byte representations
+        # differ when the decoded schema is the same
+        if (
+            not ignore_metadata
+            and self._ll_object.metadata_schema != other._ll_object.metadata_schema
```
petrelharp, line 3293:
```suggestion
            and self.metadata_schema == other.metadata_schema
            and self._ll_object.metadata_schema != other._ll_object.metadata_schema
```

petrelharp (commented):
Hm, interesting - I was going to go the more pedantic route of having `assert_equals` fail if the bytes were different. I guess this is better; however, it means that the C function and the python function do different things, so that should be documented on the C side?

*Edit:* However, if you agree that #3249 is a bug, then probably we should go the other route and make the python agree with the C code.

python/tskit/tables.py
benjeffery: Done!

python/tskit/tables.py
benjeffery: Done!

PR #3250, opened 2025-08-07
Testing for numpy errors

benjeffery:


benjeffery:
All seems to work here, errors over at #3225 seem to be when measuring coverage in the jit module.

PR #3256, opened 2025-08-15
Initial ARG traversal indexes for numba

benjeffery:
I've followed the initial suggestions in https://github.com/tskit-dev/tskit/discussions/2869

I'm not sure about the part in descendant span where a node is considered to inherit from itself for the whole sequence, but as this is just an example I'm sure we can remove the TODO and leave as is?

### DIFF
--- a/python/tskit/jit/numba.py
+++ b/python/tskit/jit/numba.py
@@ -32,7 +32,7 @@ class NumbaEdgeRange:
     Represents a range of edges during tree traversal.
 
     This class encapsulates information about a contiguous range of edges
-    that are either being removed or added to step from one tree to another
+    that are either being removed or added to step from one tree to another.
     The ``start`` and ``stop`` indices, when applied to the order array,
     define the ids of edges to process.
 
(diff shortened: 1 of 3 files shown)


jeromekelleher:
It's straightforward to do if someone wants it, but we're not planning to do it now.

python/tskit/jit/numba.py
```diff
@@ -417,24 +432,98 @@ def __init__(
...
         """
-        # This method will be overriden when the concrete JIT class NumbaTreeIndex
+        # This method will be overriden when the concrete JIT class TreeIndex
         # is defined in `jitwrap`.
-        return NumbaTreeIndex(self)  # pragma: no cover
+        return TreeIndex(self)  # pragma: no cover
+
+    def child_index(self):
+        """
+        Create child index array for finding child edges of nodes.
```
jeromekelleher: Add a statement on the computational complexity of this task, like

This operation involves a linear pass through the edge table and  therefore requires O(num_edges) time.

python/tskit/jit/numba.py
```diff
@@ -417,24 +432,98 @@ def __init__(
...
+
+        if last_parent != -1:
+            child_range[last_parent, 1] = self.num_edges
+
+        return child_range
+
+    def parent_index(self):
+        """
+        Create a :class:`ParentIndex` for finding parent edges of nodes.
+
```
jeromekelleher, line 482: Add a statement on complexity.

This operation involves sorting the edge data by child ID and left coordinate, and therefore requires O(E log E) time, there E is the number of edges.

benjeffery:
@jeromekelleher Notes on time complexity added.

I realised if we dropped the left sort from `parent_index` we can do it in `O(E + N)` time by doing a counting sort.

Issue #3253, opened 2025-08-11
Currently `tables.sort()` does not always sort mutations into the required order


hyanwong:
Taken from https://github.com/tskit-dev/tskit/pull/3212#issuecomment-3172729936

Currently I can add and delete mutations so as to create a tree sequence in which `compute_mutation_parents()` fails, even after sorting (code below). Given the upcoming requirement for mutations to be sorted correctly to make a valid tree sequence, I think we need to fix `tables.sort()` to also sort mutations into a valid order? The tricky part is, I think, the order to use for mutations that occur on the same node at the same site. We can use time, but in the case of identical times (or if times are unknown) we can presumably use the existing mutation order?

```python
import msprime
import tskit
import numpy as np

ts = msprime.sim_mutations(msprime.sim_ancestry(10, sequence_length=100, random_seed=1, recombination_rate=0.01), rate=1, random_seed=1)
print("Simulated", ts.num_mutations, "mutations")

# Add some random mutations, and delete some others
tables = ts.dump_tables()
tables.mutations.time = np.full_like(tables.mutations.time, tskit.UNKNOWN_TIME)
np.random.seed(10)
for s in ts.sites():
    tables.mutations.add_row(site=s.id, node=np.random.randint(ts.num_nodes), derived_state="A")
keep = np.ones(tables.mutations.num_rows, dtype=bool)
keep[0:100] = False
tables.mutations.replace_with(tables.mutations[keep])
# Zap all the parent IDs
tables.mutations.parent = np.full_like(tables.mutations.parent, tskit.NULL)
assert np.all(tables.mutations.parent == tskit.NULL)
tables.sort()
tables.build_index()
tables.compute_mutation_parents()
tables.tree_sequence()  # Fails
```

petrelharp:
> We can use time, but in the case of identical times (or if times are unknown) we can presumably use the existing mutation order?

I thought this is what we already did, in fact. I'm quite surprised this is failing.

hyanwong:
It says somewhere in the docs "The [TableCollection.sort()](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TableCollection.sort) method ensures that mutations are sorted according site ID, but does not at present enforce that mutations occur after their parent mutations."

PR #3257, opened 2025-08-19
Fix mutation ordering

benjeffery:
Addressing #3253 

This PR changes the mutation sort order from `site->time->ID` to `site->time->node_time->num_descendants->ID`

This is required as if mutations have unknown times, and no parent info, we need to use the node time to have parents before children. For mutations at the same node, we assume the existing order is parents before children. We need to use num descendants to use parent info if it is present for mutations at the same node.

I've updated a few of the breaking Python tests to give a flavour of the impact. This PR still needs tests for the new order, and the rest of the broken tests fixing, but thought this was a good point for feedback.

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -8905,6 +8905,7 @@ static void
 test_sort_tables_errors(void)
 {
     int ret;
+    tsk_id_t ret_id;
     tsk_treeseq_t ts;
     tsk_table_collection_t tables;
     tsk_bookmark_t pos;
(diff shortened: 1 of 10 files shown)


jeromekelleher:
What's the difference now between sort_mutations and sort_canonical? Would a simpler way of thinking about this be to say that we're changing things by making ``sort_mutations`` simply call ``sort_canonical``? Presumably we thought hard about what ``sort_canonical`` had to do, so we're just swapping in that definition now as a stricter ordering. I'd imagine we only kept the original sorting to avoid breakage at the time.

jeromekelleher:
Let's just use sort_canonical for now, and see how we get on with test suite breakage? I don't see much point in having another stricter sort if we're already changing the semantics.

benjeffery:
I think this is ready for final review. I've added @hyanwong's initial issue failing code as a test.

petrelharp:
Hm - we should benchmark the change? - recall that `sort` is a bottleneck for SLiM for instance. To be clear, this is only making `sort` match `canonical` for mutations, not other things, right?

Re: the extend haplotypes - thanks for digging into that! For reference, here's the error:
```
=========================== short test summary info ============================
FAILED tests/test_extend_haplotypes.py::TestExamples::test_suite_examples_defaults[back_mutations] - AssertionError: MutationTable row 1222 differs:
self.node=15 other.node=9
self.metadata=b'1186' other.metadata=b'1188'
```
The error is fine, though - it's because the tables end up in different sorted order, since the node below a mutation can change after extending haplotypes, and hence the mutation's position in the table after the new sorting.

benjeffery:
On already sorted tables (French-Canadian dataset) this branch adds about 40% to the sort time. I can do some more digging with shuffled tables if you think it is worth it.

benjeffery:
Rebased and squashed - I think we just need your green tick here @jeromekelleher

jeromekelleher (approved):

PR #3266, opened 2025-09-09
Time windows in AFS

tforest:
## Description

This is a WIP and replaces #2948 as a more specific PR regarding time-windows in AFS.

_In fine_, users will be able to use ts.allele_frequency_spectrum(time_windows=tw), with **_tw_** being a list of any non-overlapping time intervals in {0, inf}.

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -520,7 +520,7 @@ tsk_strerror_internal(int err)
                   "(TSK_ERR_BAD_SAMPLE_PAIR_TIMES)";
             break;
         case TSK_ERR_BAD_TIME_WINDOWS:
-            ret = "Time windows must be strictly increasing and end at infinity. "
+            ret = "Time windows must be strictly increasing. "
                   "(TSK_ERR_BAD_TIME_WINDOWS)";
             break;
         case TSK_ERR_BAD_NODE_TIME_WINDOW:
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -3491,6 +3491,7 @@ out:
 static int TSK_WARN_UNUSED
 tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double right,
     const double *restrict branch_length, double *restrict last_update,
+    const double *restrict time, tsk_id_t *restrict parent, const double *time_windows,
     const double *counts, tsk_size_t num_sample_sets, tsk_size_t window_index,
     const tsk_size_t *result_dims, tsk_flags_t options, double *result)
 {
@@ -3503,7 +3504,7 @@ tsk_treeseq_update_branch_afs(const tsk_treeseq_t *self, tsk_id_t u, double righ
     const double *count_row = GET_2D_ROW(counts, num_sample_sets + 1, u);
     double x = (right - last_update[u]) * branch_length[u];
     const tsk_size_t all_samples = (tsk_size_t) count_row[num_sample_sets];
-
+    // time, parent, time_windows (list),
     if (coordinate == NULL) {
         ret = tsk_trace_error(TSK_ERR_NO_MEMORY);
         goto out;
@@ -3529,8 +3530,8 @@ out:
 static int
 tsk_treeseq_branch_allele_frequency_spectrum(const tsk_treeseq_t *self,
     tsk_size_t num_sample_sets, double *counts, tsk_size_t num_windows,
-    const double *windows, const tsk_size_t *result_dims, tsk_flags_t options,
-    double *result)
+    tsk_size_t num_time_windows, const double *windows, const double *time_windows,
+    const tsk_size_t *result_dims, tsk_flags_t options, double *result)
 {
     int ret = 0;
     tsk_id_t u, v;
@@ -3576,15 +3577,15 @@ tsk_treeseq_branch_allele_frequency_spectrum(const tsk_treeseq_t *self,
             u = edge_child[h];
             v = edge_parent[h];
             ret = tsk_treeseq_update_branch_afs(self, u, t_left, branch_length,
-                last_update, counts, num_sample_sets, window_index, result_dims, options,
-                result);
+                last_update, node_time, parent, time_windows, counts, num_sample_sets,
+                window_index, result_dims, options, result);
             if (ret != 0) {
                 goto out;
             }
             while (v != TSK_NULL) {
                 ret = tsk_treeseq_update_branch_afs(self, v, t_left, branch_length,
-                    last_update, counts, num_sample_sets, window_index, result_dims,
-                    options, result);
+                    last_update, node_time, parent, time_windows, counts,
+                    num_sample_sets, window_index, result_dims, options, result);
                 if (ret != 0) {
                     goto out;
                 }
@@ -3604,8 +3605,8 @@ tsk_treeseq_branch_allele_frequency_spectrum(const tsk_treeseq_t *self,
             branch_length[u] = node_time[v] - node_time[u];
             while (v != TSK_NULL) {
                 ret = tsk_treeseq_update_branch_afs(self, v, t_left, branch_length,
-                    last_update, counts, num_sample_sets, window_index, result_dims,
-                    options, result);
+                    last_update, node_time, parent, time_windows, counts,
+                    num_sample_sets, window_index, result_dims, options, result);
                 if (ret != 0) {
                     goto out;
                 }
@@ -3628,8 +3629,8 @@ tsk_treeseq_branch_allele_frequency_spectrum(const tsk_treeseq_t *self,
             for (u = 0; u < (tsk_id_t) num_nodes; u++) {
                 tsk_bug_assert(last_update[u] < w_right);
                 ret = tsk_treeseq_update_branch_afs(self, u, w_right, branch_length,
-                    last_update, counts, num_sample_sets, window_index, result_dims,
-                    options, result);
+                    last_update, node_time, parent, time_windows, counts,
+                    num_sample_sets, window_index, result_dims, options, result);
                 if (ret != 0) {
                     goto out;
                 }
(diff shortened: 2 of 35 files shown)


petrelharp:
There's one line with partial coverage that I've confirmed using `tsk_bug_assert( )`s is being covered by the python tests:
Screenshot From 2025-09-16 12-56-51

I've added another C test that might satisfy this? But this is fine.

petrelharp:
The last thing to do is to put in some tests to compare versions of the method (eg decap-based and not) for a few different time windows.

python/tests/test_util.py
```diff
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2018-2024 Tskit Developers
+# Copyright (c) 2018-2025 Tskit Developers
```
jeromekelleher: Just to note here that the Copyright notice is meant to only be updated when the file itself is changed. So, the copyright years should (strictly speaking) be just those in which significant changes were made to the file.

Not that it matters in practise and there's no need to back out of these changes if you don't want to, but I just wanted to clarify my understanding of the copyright notices.

tforest: Oh yeah that makes sense, thank you for the note on that @jeromekelleher!

c/tests/testlib.c
```diff
@@ -136,20 +138,21 @@ const char *nonbinary_ex_sites = "1  0\n"
...
+0.20┊   ┃     7 ┊   ┃       ┊   7       ┊         ┊
+    ┊   ┃     ┃ ┊   ┃       ┊ ┏━┻━━┓    ┊         ┊
+0.17┊   6     ┃ ┊   6       ┊ ┃    ┃    ┊         ┊
+    ┊ ┏━┻━┓   ┃ ┊ ┏━┻━━┓    ┊ ┃    ┃    ┊         ┊
+0.09┊ ┃   5   ┃ ┊ ┃    5    ┊ ┃    5    ┊         ┊
+    ┊ ┃ ┏━╋━┓ ┃ ┊ ┃ ┏━━╋━━┓ ┊ ┃ ┏━━╋━━┓ ┊         ┊
+0.07┊ ┃ ┃ ┃ ┃ ┃ ┊ ┃ ┃  4  ┃ ┊ ┃ ┃  4  ┃ ┊         ┊
+    ┊ ┃ ┃ ┃ ┃ ┃ ┊ ┃ ┃ ┏┻┓ ┃ ┊ ┃ ┃ ┏┻┓ ┃ ┊         ┊
+0.00┊ 0 1 3 9 2 ┊ 0 1 2 3 9 ┊ 0 1 2 3 9 ┊ 0 1 2 3 ┊
+    0           2           7          10        100
```
petrelharp, line 155: I've added a non-sample leaf to this example, which we needed to get better coverage.

c/tskit/core.c
```diff
@@ -526,9 +526,13 @@ tsk_strerror_internal(int err)
                   "(TSK_ERR_BAD_SAMPLE_PAIR_TIMES)";
             break;
         case TSK_ERR_BAD_TIME_WINDOWS:
-            ret = "Time windows must be strictly increasing and end at infinity. "
+            ret = "Time windows must start at zero and be strictly increasing. "
                   "(TSK_ERR_BAD_TIME_WINDOWS)";
             break;
+        case TSK_ERR_BAD_TIME_WINDOWS_END:
```
petrelharp, line 532: we needed to disambiguate the various notions of "bad" as different stats have different requirements now

python/tests/test_tree_stats.py
```diff
@@ -40,8 +42,30 @@
 import tskit
 import tskit.exceptions as exceptions
 
+
 np.random.seed(5)
 
+# Notes for refactoring:
```
petrelharp, line 48: These are notes about what happens in this file, in particular for a near-future me in refactoring the tests here.

python/tests/test_tree_stats.py
```diff
@@ -4252,13 +4470,29 @@ def verify_sample_sets(self, ts, sample_sets, windows):
...
             self.assertArrayAlmostEqual(sfs1, sfs3)
+            assert np.all(sfs3 >= 0)
 
 
 class TestBranchAlleleFrequencySpectrum(
     TestAlleleFrequencySpectrum, TopologyExamplesMixin
 ):
     mode = "branch"
 
+    def generate_params(self, ts):
```
petrelharp, line 4490: Here is the main code testing the new time windows; this gets called on all the example tree sequences.

petrelharp: (Note that it also determines which combinations of windows and sample sets get tested.)

petrelharp:
Ah one more thing - perhaps @tforest can benchmark this to see if we've affected performance for non-time-windowed AFS by a lot?

c/tskit/trees.c
```diff
@@ -1256,6 +1257,37 @@ tsk_treeseq_check_windows(const tsk_treeseq_t *self, tsk_size_t num_windows,
...
+    // which is required for some time window functions.
+    int ret = TSK_ERR_BAD_TIME_WINDOWS;
+    tsk_size_t j;
+
+    if (num_windows < 1) {
+        ret = TSK_ERR_BAD_TIME_WINDOWS_DIM;
+        goto out;
+    }
+
+    if (windows[0] < 0.0) {
```
jeromekelleher: Could combine these two into one ``windows[0] <= 0.0`` condition?

c/tskit/trees.c
```diff
@@ -23,6 +23,7 @@
  * SOFTWARE.
  */
 
+#include "tskit/core.h"
```
jeromekelleher: tskit includes should go after the std library ones

jeromekelleher: Anyway, isn't this already included already by trees.h ?

c/tskit/trees.c
```diff
@@ -3716,13 +3749,28 @@ tsk_treeseq_allele_frequency_spectrum(const tsk_treeseq_t *self,
...
+    if (time_windows == NULL) {
+        num_time_windows = 1;
+        time_windows = default_time_windows;
+    } else {
+        ret = tsk_treeseq_check_time_windows(num_time_windows, time_windows);
+        if (ret != 0) {
+            goto out;
+        }
+        if (stat_site
+            && tsk_memcmp(time_windows, default_time_windows, 2 * sizeof(const double))
```
jeromekelleher: Took me a while to parse this - so it's saying that only the default time windows are supported for STAT_SITE, yeah? Might be easier to read and less fragile to say
```
// Site mode does not support time windows
if (stat_site && !(time_windows[0] == 0 && isinf(time_windows[1]))
```

python/_tskitmodule.c
```diff
@@ -9576,24 +9578,28 @@ TreeSequence_allele_frequency_spectrum(
     if (parse_windows(windows, &windows_array, &num_windows) != 0) {
         goto out;
     }
-
-    shape = PyMem_Malloc((num_sample_sets + 1) * sizeof(*shape));
+    if (parse_windows(time_windows, &time_windows_array, &num_time_windows) != 0) {
+        goto out;
+    }
+    shape = PyMem_Malloc((num_sample_sets + 1 + 1) * sizeof(*shape));
```
jeromekelleher: The ``+ 1 + 1``s are confusing me - maybe just use + 2 and clarify why with a comment?

python/tskit/trees.py
```diff
@@ -7993,27 +8013,40 @@ def __one_way_sample_set_stat(
             # of integers then drop the dimension
             if len(sample_sets.shape) == 1:
                 sample_sets = [sample_sets]
-                drop_dimension = True
-
+                if ll_method.__name__ != "allele_frequency_spectrum":
```
jeromekelleher: I'm assuming this is temporary?

c/tskit/trees.c
petrelharp: why so we could

petrelharp: or rather just `!=`

petrelharp: Hm, also I was just remembering that in fact times can be negative, so being `< 0` is not necessarily an error

c/tskit/trees.c
petrelharp: ah that would be MUCH better, thanks. I had thought there was some reason we couldn't do that, but it seems okay.

python/tskit/trees.py
petrelharp: *this* is not temporary (so maybe there's a better way to do it) - this is because of https://github.com/tskit-dev/tskit/pull/3266#issuecomment-3289645790

jeromekelleher: Ah, I see we're doing this control flow based on introspecting the method name in a few places. This does smell quite badly I'm afraid

Can we refactor this by adding some parameters to ``__one_way_sample_set_stat`` which default to the standard behaviour, but we set to "False" when called from the top-level ``allele_frequency_spectrum`` function?

petrelharp:
Not sure what you mean @jeromekelleher - the only `0` in that expression is an index?

petrelharp:
(whoops - missed the stuff in the comment above that clarified the problem is that the argument to `isinf( )` is double but it's wanting a float)

tforest:
Yes! I'm trying something @petrelharp :)

petrelharp:
I think the problem is the argument: for Windows it looks like [isinf requires a float](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/isinf?view=msvc-170), not a double?

tforest:
That was it, yes. :+1:

petrelharp:
Nice. I'm squashing & rebasing.

jeromekelleher:
I do think we should figure out a different way of dealing with dimensions etc than by introspecting the function pointer argument --- this is a really fragile and opaque way of doing control flow. Fine for this to be logged as an issue and done in a follow-up, but it it will bite eventually if we don't fix it.

jeromekelleher (approved):

PR #3272, opened 2025-09-17
Add C coverage docs

benjeffery:
@petrelharp asked how to locally work with the C coverage, instead of that getting lost on slack I thought  I'd just add it to the docs.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -737,6 +737,39 @@ and IO errors are difficult to simulate in C.) Code coverage statistics are
 automatically tracked using [CodeCov](https://codecov.io/gh/tskit-dev/tskit/).
 
 
+### Viewing coverage reports
+
+To generate and view coverage reports for the C tests locally:
+
+Compile with coverage enabled:
+   ```bash
+   $ cd c
+   $ meson build -D b_coverage=true
+   $ ninja -C build
+   ```
+
+Run the tests:
+   ```bash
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -737,6 +737,38 @@ and IO errors are difficult to simulate in C.) Code coverage statistics are
 automatically tracked using [CodeCov](https://codecov.io/gh/tskit-dev/tskit/).
 
 
+### Viewing coverage reports
+
+To generate and view coverage reports for the C tests locally:
+
+Compile with coverage enabled:
+   ```bash
+   $ meson build-gcc c -D b_coverage=true
```
jeromekelleher: The -gcc bit here will likely confuse people (who may not be using GCC). Just use ``meson build -D b_coverage=tru ``.

Also simpler to say "do this in the c directory``?

Also I think meson raises a warning if you use it like this now, isn't there a new recommened way of calling it?

benjeffery: Fixed. I don't get any errors from `meson build -D ....`

jeromekelleher (approved):

petrelharp:
Excellent. Out of curiosity, how does this get run by CodeCov? I.e., is there a config file somewhere where we have these commands (or something), or is this just the automatic stuff codecov does when we ask it for coverage on the `c/` directory?

PR #3277, opened 2025-09-22
Mutation inherited state

### LINKED ISSUE #2631: Mutation's "inherited" state
Getting the state that a mutation inherited is fiddly right now, and has been annoying me quite bit recently. We should add a method to the Python object at least to get it, but it would also be nice to have access to it in numpy for reasoning efficiently about reversions and so on.

Does ``inherited_state`` feel like the right term here? 

The numpy stuff would depend on us making some decisions about vectorised access to the string columns which I guess I should open a seperate issue on.

```python
tables = ts.tables
assert np.all(tables.mutations.derived_state_offset == np.arange(ts.num_mutations + 1))
derived_state = tables.mutations.derived_state.view("S1")
assert np.all(tables.sites.ancestral_state_offset == np.arange(ts.num_sites + 1))
ancestral_state = tables.sites.ancestral_state.view("S1")
# Compute the "inherited state" for each mutation
inherited_state = ancestral_state[ts.mutations_site]
mutations_with_parent = ts.mutations_parent != -1
parent = ts.mutations_parent[mutations_with_parent]
assert np.all(parent >= 0)
inherited_state[mutations_with_parent] = derived_state[parent]

# Check
for mut in ts.mutations():
    state0 = ts.site(mut.site).ancestral_state
    if mut.parent != -1:
        state0 = ts.mutation(mut.parent).derived_state
    assert state0 == inherited_state[mut.id].decode()
```

benjeffery:
Fixes #2631

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -97,6 +97,10 @@ for more information.
   TreeSequence.mutations_node
   TreeSequence.mutations_parent
   TreeSequence.mutations_time
+  TreeSequence.mutations_derived_state
+  TreeSequence.mutations_metadata
+  TreeSequence.mutations_edge
+  TreeSequence.mutations_inherited_state
   TreeSequence.migrations_left
   TreeSequence.migrations_right
   TreeSequence.migrations_right
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1218,6 +1218,7 @@ class TestTreeSequence(HighLevelTestCase):
     def test_mutations(self, ts):
         self.verify_mutations(ts)
 
+    @pytest.mark.skipif(not _tskit.HAS_NUMPY_2, reason="Requires NumPy 2.0 or higher")
     @pytest.mark.parametrize("ts", tsutil.get_example_tree_sequences())
     def test_mutation_inherited_state_property(self, ts):
         inherited_states = ts.mutations_inherited_state


python/tskit/trees.py
```diff
@@ -6320,6 +6364,7 @@ def mutation(self, id_):
             time=time,
             edge=edge,
             metadata_decoder=self.table_metadata_schemas.mutation.decode_row,
+            tree_sequence=self,
```
jeromekelleher: Hmm - what does this mean for serialisation?

benjeffery: I copied the pattern from `Individual`, which we've been using since https://github.com/tskit-dev/tskit/commit/8a8e61e2ae7bae5ef54ff6dafcecab11d7e3962d in 2022. These can still be pickled as the reference is followed, and the TS pickled too. If you are pickling a collection of these, `pickle` is smart enough to only store the ts once.

jeromekelleher: I would find it surprising that  the whole ts gets pickled if I just wanted to pass a simple mutation though. It it that bad if we just store an extra attribute on the object?

I'm not sure the individual thing was such a good idea looking back - we don't have to fix it now, but I don't want to reinforce a questionable decision there

benjeffery: Yeah, I get what you're saying. Let me think about the options and I'll make a proposal here.

benjeffery:
@jeromekelleher Hows this? We pre-compute, delaying any errors from that to when the property is actually accessed. I need to add a couple more tests for coverage but I'm pretty happy with this.

c/tskit/trees.c
```diff
@@ -311,6 +318,26 @@ tsk_treeseq_init_trees(tsk_treeseq_t *self)
...
+                if (mutation_parent[mutation_id] == TSK_NULL) {
+                    /* No parent: inherited state is the site's ancestral state */
+                    mutation->inherited_state
+                        = sites_ancestral_state + sites_ancestral_state_offset[site_id];
+                    mutation->inherited_state_length
+                        = sites_ancestral_state_offset[site_id + 1]
+                          - sites_ancestral_state_offset[site_id];
+                } else {
+                    /* Has parent: inherited state is parent's derived state */
+                    tsk_id_t parent_id = mutation_parent[mutation_id];
```
jeromekelleher: Push variable declaration to the top of the function for consistency

PR #3283, opened 2025-09-24
extend `union` to do "disjoint union"

petrelharp:
Here's the start at what we discussed in #3183. Might you have a go at putting in the python tests, @hyanwong?

Something I haven't done that we said we might in the other PR is require that if `all_edges` or `all_mutations` are True then `check_shared_overlap` is False. I don't think we actually need to require this (since in some cases all three might be true and it's fine!) but we should probably say in the docstring that the user probably *wants* this to be the case. (I just don't remember why that is right now.)

Still TODO:
- [ ] python tests
- [ ] CHANGELOG
- [ ] say something about checking shared overlap in the docs

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -11241,6 +11241,107 @@ test_table_collection_union(void)
 }
 
 static void
+test_table_collection_disjoint_union(void)
+{
+    int ret;
+    tsk_id_t ret_id;
+    tsk_table_collection_t tables;
+    tsk_table_collection_t tables1;
+    tsk_table_collection_t tables2;
+    tsk_table_collection_t tables12;
+    tsk_id_t node_mapping[4];
+
+    tsk_memset(node_mapping, 0xff, sizeof(node_mapping));
+
+    ret = tsk_table_collection_init(&tables1, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables1.sequence_length = 2;
+
+    // set up nodes, which will be shared
+    // flags, time, pop, ind, metadata, metadata_length
+    ret_id = tsk_node_table_add_row(
+        &tables1.nodes, TSK_NODE_IS_SAMPLE, 0.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(
+        &tables1.nodes, TSK_NODE_IS_SAMPLE, 0.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(&tables1.nodes, 0, 0.5, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(&tables1.nodes, 0, 1.5, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret = tsk_table_collection_copy(&tables1, &tables2, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // for tables1:
+    // on [0, 1] we have 0, 1 inherit from 2
+    // left, right, parent, child, metadata, metadata_length
+    ret_id = tsk_edge_table_add_row(&tables1.edges, 0.0, 1.0, 2, 0, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_edge_table_add_row(&tables1.edges, 0.0, 1.0, 2, 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_site_table_add_row(&tables1.sites, 0.4, "T", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_mutation_table_add_row(
+        &tables1.mutations, ret_id, 0, TSK_NULL, TSK_UNKNOWN_TIME, NULL, 0, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret = tsk_table_collection_build_index(&tables1, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_sort(&tables1, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // all this goes in tables12 so far
+    ret = tsk_table_collection_copy(&tables1, &tables12, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // for tables2; and need to add it to tables12 also:
+    // on [1, 2] we have 0, 1 inherit from 3
+    // left, right, parent, child, metadata, metadata_length
+    ret_id = tsk_edge_table_add_row(&tables2.edges, 1.0, 2.0, 3, 0, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_edge_table_add_row(&tables2.edges, 1.0, 2.0, 3, 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_site_table_add_row(&tables2.sites, 1.4, "A", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_mutation_table_add_row(
+        &tables2.mutations, ret_id, 0, TSK_NULL, TSK_UNKNOWN_TIME, "T", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret = tsk_table_collection_build_index(&tables2, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    ret = tsk_table_collection_sort(&tables2, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    // also tables12
+    ret_id = tsk_edge_table_add_row(&tables12.edges, 1.0, 2.0, 3, 0, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_edge_table_add_row(&tables12.edges, 1.0, 2.0, 3, 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_site_table_add_row(&tables12.sites, 1.4, "A", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_mutation_table_add_row(
+        &tables12.mutations, ret_id, 1, TSK_NULL, TSK_UNKNOWN_TIME, "T", 1, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret = tsk_table_collection_build_index(&tables12, 0);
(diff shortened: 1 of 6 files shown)


petrelharp:
Whoops - *now* this should be ready to write tests for.

hyanwong:
I made a PR to your PR, @petrelharp - the only thing I can see that might be a problem is that we might want to include empty sites when unioning one TS with another. If so, we could either:

1. Include empty sites when `all_mutations=True`
2. Have an additional parameter: e.g. `all_sites`
3. Have a single parameter `all` which does `all_mutations`, `all_edges` and `all_sites` (is there a call for separating edges and sites?)

I guess (2) gives the most flexibility for the user?

petrelharp:
Let's keep the convo here, and lmk if you don't have permissions to push here.

That all looks good; I think we just need some additional checks for correctness of the union (a lot of the checks are just that the number of things is right).

I guess I don't think we should have a separate `all_sites` argument, we should just have `all_mutations` also include all sites. Otherwise the behavior is complicated.

It's tempting to have a `disjoint` argument that does `all_mutations=True`, `all_edges=True`, and `check_shared_overlap=False`. I'm not sure, since it's adding another argument, but I think I like it since it points users towards the use case. (And if they use this with two that have shared overlap they'll get duplicated edges and hence an error, which is good.)

c/tskit/tables.h
```diff
@@ -858,11 +858,21 @@ equality of the subsets.
...
  */
 #define TSK_UNION_NO_ADD_POP (1 << 1)
+/**
+By default, union only adds only edges adjacent to a newly added node;
+this option adds all edges.
+ */
+#define TSK_UNION_ALL_EDGES (1 << 2)
+/**
+By default, union only adds only mutations on newly added edges;
+this option adds all mutations.
```
petrelharp:
```suggestion
By default, union only adds only mutations on newly added edges, and
sites for those mutations; this option adds all mutations and all sites.
```

python/tskit/tables.py
```diff
@@ -3953,6 +3955,10 @@ def union(
             should be the index of the equivalent node in ``self``, or
             ``tskit.NULL`` if the node is not present in ``self`` (in which case it
             will be added to self).
+        :param bool all_edges: If True, then all edges in ``other`` are added
+            to ``self``. Must have ``check_shared_equality=False``.
```
petrelharp:
```suggestion
            to ``self``.
```

python/tskit/tables.py
```diff
@@ -3953,6 +3955,10 @@ def union(
...
+        :param bool all_mutations: If True, then all mutations in ``other`` are added
+            to ``self``. Must have ``check_shared_equality=False``.
```
petrelharp:
```suggestion
            to ``self``.
```

python/tskit/tables.py
```diff
@@ -4334,6 +4334,8 @@ def union(
         self,
         other,
         node_mapping,
+        all_edges=False,
```
benjeffery: Placing these at the start of kwargs has broken compatibility with old code that might have been using them positionally as we had no `*`.

python/tskit/trees.py
```diff
@@ -7198,15 +7198,16 @@ def concatenate(
...
-        calling :meth:`~TreeSequence.union` with an (optional)
-        node mapping for each of the ``others``. If any node mapping is ``None``
-        only map the sample nodes between the input tree sequence and this one,
-        based on the numerical order of sample node IDs.
+        Concatenate a set of tree sequences to the right of this one, by shifting
+        their coordinate systems and repeatedly calling :meth:`~TreeSequence.union`.
+        By default, the samples in current and input tree sequences are assumed to
+        refer to the same nodes, and are matched based on the numerical order of
+        sample node IDs. This can be changed by providing explicit ``node_mappings``
+        for each input tree sequence.
```
petrelharp: I think it's confusing to refer to union straight up, because it's not using the default behavior of union (in fact it's quite different!). So anyone who goes and looks at the union docs will be very confused (especially because this docstring doesn't say it's doing all_mutations and all_edges). How about this:
```suggestion
        Concatenate a set of tree sequences to the right of this one, by shifting
        their coordinate systems and adding all edges, sites, mutations, and
        any additional nodes, individuals, or populations needed for these.
        Concretely, to concatenate an ``other`` tree sequence to ``self``, the value
        of ``self.sequence_length`` is added to all genomic coordinates in ``other``,
        and then the concatenated tree sequence  will contain all edges, sites, and
        mutations in both. Which nodes in ``other`` are treated as "new", and hence
        added as well, is controlled by ``node_mapping``. Any individuals to which
        new nodes belong are added as well.
        
        The method uses :meth:`.shift` followed by :meth:`.union`, with
        ``all_mutations=True``, ``all_edges=True``, and ``check_shared_equality=False``.
        
        By default, the samples in current and input tree sequences are assumed to
        refer to the same nodes, and are matched based on the numerical order of
        sample node IDs; all other nodes are assumed to be new. This can be
        changed by providing explicit ``node_mappings`` for each input tree sequence
        (see below).
```

python/tskit/trees.py
```diff
@@ -7513,6 +7518,26 @@ def union(
         nodes are in entirely new populations, then you must set up the
         population table first, and then union with ``add_populations=False``.
 
+        This method makes sense if the "shared" portions of the tree sequences
+        are equal; the option ``check_shared_equality`` performs a consistency
+        check that this is true. If this check is disabled, it is very easy to
+        produce nonsensical results via subtle inconsistencies.
+
+        The behavior above can be changed by ``all_edges`` and ``all_mutations``.
+        If ``all_edges`` is True, then all edges in ``other` are added to
```
benjeffery: Missing backtick

python/_tskitmodule.c
```diff
@@ -4347,15 +4347,18 @@ TableCollection_union(TableCollection *self, PyObject *args, PyObject *kwds)
     npy_intp *shape;
     tsk_flags_t options = 0;
     int check_shared = true;
+    int all_edges = false;
+    int all_mutations = false;
     int add_populations = true;
-    static char *kwlist[] = { "other", "other_node_mapping", "check_shared_equality",
-        "add_populations", NULL };
+    static char *kwlist[] = { "other", "other_node_mapping", "all_edges",
```
benjeffery: Although it is internal code, best practice is to add them at the end. It's no different but avoid breaking if anyone is using it.

c/tests/test_tables.c
```diff
@@ -11240,6 +11240,107 @@ test_table_collection_union(void)
...
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    // now disjoint union-ing tables1 and tables2 should get tables12
+    ret = tsk_table_collection_copy(&tables1, &tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    node_mapping[0] = 0;
+    node_mapping[1] = 1;
+    node_mapping[2] = 2;
+    node_mapping[3] = 3;
+    ret = tsk_table_collection_union(&tables, &tables2, node_mapping,
```
benjeffery, line 11334: Should this not check the output?

python/tests/test_topology.py
```diff
@@ -7215,18 +7215,54 @@ def test_reference_sequence(self):
...
+            assert site1.metadata == site2.metadata
+            assert site1.ancestral_state == site2.ancestral_state
+            assert len(site1.mutations) == len(site2.mutations)
+            for m1, m2 in zip(site1.mutations, site2.mutations):
+                assert m1.metadata == m2.metadata
+                assert m1.derived_state == m2.derived_state
+                assert m1.time == m2.time
+        ns_nodes = np.where(ts.tables.nodes.flags & tskit.NODE_IS_SAMPLE == 0)[0]
+        assert len(ns_nodes) > 0
+        for u1, u2 in zip(ns_nodes[: len(ns_nodes)], ns_nodes[len(ns_nodes) :]):
```
benjeffery: I think the second argument is an empty slice, so this loop won't run. It is wise to always use the `strict` argument to detect these kind of things.

python/tskit/trees.py
```diff
@@ -7513,6 +7518,26 @@ def union(
...
+        ``self``, instead of only edges adjacent to added nodes. If
+        ``all_mutations`` is True, then similarly all mutations in ``other``
+        are added (not just those on added nodes); furthermore, all sites
+        at positions without a site already present are added to ``self``.
+        The intended use case for these options is a "disjoint" union,
+        where for instance the two tree sequences contain information about
+        disjoint segments of the genome (see :meth:`.concatenate`).
+        For some such applications it may be necessary to set
+        ``check_shared_equality=False``: for instance, if `other` has an
+        an identical copy of the node table but no edges, then
```
benjeffery: Double "an", and single backticks.

c/tskit/tables.h
```diff
@@ -858,11 +858,21 @@ equality of the subsets.
...
 #define TSK_UNION_NO_CHECK_SHARED (1 << 0)
 /**
- By default, all nodes new to ``self`` are assigned new populations. If this
+By default, all nodes new to ``self`` are assigned new populations. If this
 option is specified, nodes that are added to ``self`` will retain the
 population IDs they have in ``other``.
  */
 #define TSK_UNION_NO_ADD_POP (1 << 1)
+/**
+By default, union only adds only edges adjacent to a newly added node;
```
benjeffery: Double "only"

python/tskit/tables.py
hyanwong: Oh, good point. Easy enough to put a `*` in now, right?

petrelharp: I'm moving them to the end, but if it's not too rude then we should add a `*`.

benjeffery: Yeah, add a star, but just before your new kwargs, to not break existing positional usage.

python/tests/test_topology.py
```diff
@@ -7215,18 +7215,54 @@ def test_reference_sequence(self):
...
+            assert len(site1.mutations) == len(site2.mutations)
+            for m1, m2 in zip(site1.mutations, site2.mutations):
+                assert m1.metadata == m2.metadata
+                assert m1.derived_state == m2.derived_state
+                assert m1.time == m2.time
+        ns_nodes = np.where(ts.tables.nodes.flags & tskit.NODE_IS_SAMPLE == 0)[0]
+        assert len(ns_nodes) > 0
+        for u in ns_nodes:
+            node1 = joint_ts.node(u)
+            node2 = joint_ts.node(u)
```
benjeffery, line 7257: I'm still confused here, aren't node1 and node2 the same object?

petrelharp: LOL yes sorry, will fix

petrelharp:
Okay - perhaps someone (@hyanwong?) should have a careful read of the docstrings and then this is good? The failing coverage is because the diff is not many lines and there's a "we can't cover this" line in that diff, so the percent covered on the diff is small.

python/tskit/trees.py
```diff
@@ -7198,15 +7198,28 @@ def concatenate(
...
-        only map the sample nodes between the input tree sequence and this one,
-        based on the numerical order of sample node IDs.
+        Concatenate a set of tree sequences to the right of this one, by shifting
+        their coordinate systems and adding all edges, sites, mutations, and
+        any additional nodes, individuals, or populations needed for these.
+        Concretely, to concatenate an ``other`` tree sequence to ``self``, the value
+        of ``self.sequence_length`` is added to all genomic coordinates in ``other``,
+        and then the concatenated tree sequence  will contain all edges, sites, and
+        mutations in both. Which nodes in ``other`` are treated as "new", and hence
+        added as well, is controlled by ``node_mapping``. Any individuals to which
```
benjeffery: node_mappings rather than node_mapping

python/tskit/trees.py
benjeffery: "An list"

python/tskit/trees.py
```diff
@@ -7524,6 +7562,10 @@ def union(
         :param TableCollection other: Another table collection.
```
benjeffery: Another tree sequence?

PR #3288, opened 2025-09-28
Make TreeSequence.tables return zero-copy ImmutableTableCollection

### LINKED ISSUE #760: Table collection returned by ts.tables not read-only.

benjeffery:
Stacked on #3287 (See https://github.com/tskit-dev/tskit/commit/3b69da9492c360a43280ddf575992c8f9f2e7643 for this PR's diff)
Fixes #760

It was bothering me that we were releasing 1.0 without taking the chance to fix `TreeSequence.tables`. It's a breaking change (although one that has been warned about) so should go before 1.0.

I then realised there was a "quick" way to do this by having an `ImmutableTableCollection` that is solely backed by the low-level `_tskit.TreeSequence` class. Access to all table data is zero copy, and using that interface guarantees that we are not mutating the tree sequence.

One complicating factor:
 - For compatibility, we have to repack ragged string arrays, as the low-level returns `StringDType`. We could make (e.g.)`TableCollection.sites.ancestral_state` return a `StringDType` instead, but that would be quite a nasty breaking change. The other option is to add another low-level array access method for the data in the numpy 1 way. This would be better as currently `numpy<2` has to return a normal `TableCollection`. (Writing this makes it seem the obvious way now)


jeromekelleher:
Very nice, clever idea to use the low-level tree sequence to back and guantee immutability.

I guess the high-level question is here, why do we need this now that we have immutable array-level access on the TreeSequence? I had forgotten about this because in a way we don't really need it any more. It feels like a lot of code and complexity (and code breakage) to access functionality that we already have in a slightly different way.

benjeffery:
While it's true you can get the arrays, they don't have the table semantics of iteration over row objects, subsetting, equality etc, `ts.tables` as currently implemented is a bit of a footgun unless you've read the docs closely, and this fixes that for existing code. I agree it's a bit much code - I think I can get that down a bit.

jeromekelleher:
> I think I can make this a non-breaking change. If a mutating method is called, we emit a deprecation warning, replace the ImmutableTC with a normal TableCollection then call the mutator on that.

Eeesh - I can see that having intended consequences... Let's get some feedback

molpopgen:
I think this is a good way to go.  Instead of a deprecation warning, I'd do a hard break and raise errors when applying mutable operations.  (Or, remove the mutable ops from the API of the immutable type.)

python/tests/test_immutable_table_collection.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def get_mutable_and_immutable(ts):
+    mutable = ts.dump_tables()
+    immutable = ts.tables
+    assert isinstance(immutable, tskit.tables.ImmutableTableCollection)
+    return mutable, immutable
+
+
+@pytest.mark.parametrize("ts", tsutil.get_example_tree_sequences())
```
jeromekelleher, line 25: This is a great test pattern - I didn't realise you could parametrize a class like this 👍

python/tskit/tables.py
```diff
@@ -377,6 +363,24 @@ def nbytes(self) -> int:
...
+
+        if self is other:
+            return True
+
+        if not isinstance(other, BaseTable) or self.table_name != other.table_name:
+            return False
+
+        # If can only use mutable fast path if both tables are mutable
+        if self.mutable and not other.mutable:
+            self, other = other, self
```
jeromekelleher: Swapping the self reference is a bit of a footgun. How about 
```
base = self
if self.mutable and not other.mutable:
    base = other
    other = self
return base._...
```
Also comment above doesn't quite parse

python/tskit/tables.py
```diff
@@ -468,6 +495,192 @@ def __eq__(self, other):
...
+                    else:
+                        formatted_values.append(str(value))
+                rows.append(formatted_values)
+        return headers, rows
+
+
+def _assert_table_collections_equal(tc1, tc2, **kwargs):
+    # This is shared between TableCollection and ImmutableTableCollection,
+    # could go in a base class, but there's not much else in common
+    ignore_metadata = kwargs.get("ignore_metadata", False)
```
jeromekelleher: Not clear why you're using kwargs here rather than just having standard function signature?

python/tskit/tables.py
```diff
@@ -653,98 +870,232 @@ def copy(self):
...
+        "packset_derived_state",
+        "packset_record",
+        "packset_timestamp",
+        "squash",
+    }
+
+    # These are set by subclasses.
+    table_name = None
+    mutable_class = None
+    mutable = False
```
jeromekelleher: mutable isn't set by subclass, push up above comment?

python/tskit/tables.py
```diff
@@ -653,98 +870,232 @@ def copy(self):
...
+            # Optimization: directly assign array properties when no row_indices
+            # This avoids the `if self._row_indices is None` check on every
+            # property access
+            for column_name in self.column_names:
+                if not column_name.endswith("_schema"):
+                    # Get array from low-level tree sequence
+                    array = getattr(self._llts, f"{self.table_name}_{column_name}")
+                    if column_name == "metadata":
+                        array = np.asarray(array)
+                        if array.dtype == np.uint8:
```
jeromekelleher: Common here on why we need this

python/tskit/tables.py
```diff
@@ -653,98 +870,232 @@ def copy(self):
...
-    def _text_header_and_rows(self, limit=None):
-        display_columns = [
-            col for col in self.column_names if not col.endswith("_offset")
-        ]
-        headers = ("id",) + tuple(display_columns)
+    def _fast_equals(self, other, *, ignore_metadata=False, ignore_timestamps=False):
+        if self.num_rows != other.num_rows:
+            return False
+        if (
+            not ignore_metadata
```
jeromekelleher, line 984: This stanza is repeated a few times - worth having a simple helper function to implement?

python/tskit/tables.py
```diff
@@ -653,98 +870,232 @@ def copy(self):
...
-            if np.issubdtype(arr.dtype, np.floating):
-                float_columns[col] = 0 if self._columns_all_integer(col) else 8
+        # If we are a subset with row_indices, we need to provide indexed access
+        # to array properties.
+        # we don't get to this code if we're a full table as we set the arrays
+        # directly in __init__.
+        if name in self.column_names:
+            # Get the full array and index it
+            full_array = getattr(self._llts, f"{self.table_name}_{name}")
+            # We have to view as int8 here to maintain compatibility with
```
jeromekelleher: We did this above somewhere too - can we abstract to a helper function?

python/tskit/tables.py
```diff
@@ -653,98 +870,232 @@ def copy(self):
...
+
+            # Handle ragged arrays specially as we have to rebuild to give a view.
+            offset_column_name = f"{name}_offset"
+            if offset_column_name in self.column_names:
+                offset_array = getattr(
+                    self._llts, f"{self.table_name}_{offset_column_name}"
+                )
+                indexed_offsets = offset_array[self._row_indices]
+                next_offsets = offset_array[self._row_indices + 1]
+                pieces = [
```
jeromekelleher: It's not obvious that we're incurring this Python loop from the external API, but I guess it's OK because it only happens if we actually access ragged columns? But (maybe I'm jumping ahead here) if we use row-level access, then won't *all* columns get read?

python/tskit/trees.py
```diff
@@ -4321,21 +4322,23 @@ def tables_dict(self):
...
-        :return: A :class:`TableCollection` containing all a copy of the
-            tables underlying this tree sequence.
-        :rtype: TableCollection
+        Note that if tkskit was built with Numpy 1, this method acts as
+        :meth:`.dump_tables` and returns a mutable TableCollection.
+
+        :return: An immutable view of the TableCollection underlying this tree sequence.
         """
-        return self.dump_tables()
+        if not _tskit.HAS_NUMPY_2:
```
jeromekelleher, line 4335: Maybe we should emit a warning here? This will be potentially surprising

benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3288/commits/f929c6d2913c6d37d9f79a972d37822493503830

python/tskit/tables.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3288/commits/d00201b2167e799bd744345a3775fb663f4e70b1

python/tskit/tables.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3288/commits/35b25acb1656f9f5fab73ca05ec13a1e0fd0d46c

python/tskit/tables.py
benjeffery: Fixed in https://github.com/tskit-dev/tskit/pull/3288/commits/8dfdc73bc7c0aa6715a6a16b02700b127bce2734

python/tskit/tables.py

python/tskit/tables.py
benjeffery: It's only twice that I can see?

python/tskit/tables.py
benjeffery: I've done a load of perf work - see below.

python/tskit/trees.py
```diff
@@ -4321,21 +4322,29 @@ def tables_dict(self):
...
-            **view** in the future. Thus, if the tables will subsequently be
-            updated, please use the :meth:`.dump_tables` method instead as
-            this will always return a new copy of the TableCollection.
+        This view shares the same data as the TreeSequence (zero-copy).
+        Use :meth:`.dump_tables` for a modifiable copy.
 
-        :return: A :class:`TableCollection` containing all a copy of the
-            tables underlying this tree sequence.
-        :rtype: TableCollection
+        Note that if tkskit was built with Numpy 1, this method acts as
```
```suggestion
        Note that if tskit was built with Numpy 1, this method acts as
```

python/tskit/tables.py
```diff
@@ -653,98 +874,316 @@ def copy(self):
...
+            # TableCollection methods use the LWT code, which is stuck returning
+            # int8 for compatibility see https://github.com/tskit-dev/tskit/issues/3284
+            if name == "metadata":
+                full_array = full_array.view(np.int8)
+
+            is_offset = name.endswith("_offset")
+            is_ragged = f"{name}_offset" in self.column_names
+
+            if self._row_indices is None and self._row_slice is None:
+                object.__setattr__(self, name, full_array)
```
jeromekelleher: Hold on, why are we calling setattr here? Seems pretty weird for a getattr to have a sideeffect?

python/tskit/tables.py
```diff
@@ -653,98 +874,316 @@ def copy(self):
...
-                            formatted_values.append(", ".join(f"{p:,}" for p in value))
-                        else:
-                            formatted_values.append(", ".join(map(str, value)))
-                    elif col in float_columns:
-                        dp = float_columns[col]
-                        formatted_values.append(f"{value:,.{dp}f}")
-                    elif isinstance(value, (int, np.integer)):
-                        formatted_values.append(f"{value:,}")
+                indices = self._row_indices
+                if is_offset:
```
jeromekelleher: There's a *lot* of logic in this method. Can we break down these different cases to call smaller helper functions? Like, here we'd call ``result = self._get_offset_array_by_indices(indices)``. I don't like having lots of complex logic nested inside in conditionals.

I think we should aim to have the overall getattr function implemented in ~10 lines of code

python/tskit/tables.py
jeromekelleher: This should be documented clearly at a minimum

benjeffery: We store the calculated column back on `self` so that next time it gets picked up as a normal attribute. Also to calculate `ragged` we have to calculate `ragged_offset`. We might as well store the result of calculating `ragged_offset` rather than re-compute when the user needs it. I'll add a note.

python/tskit/tables.py
benjeffery: How's https://github.com/tskit-dev/tskit/pull/3288/commits/8e858dc65d2b84c0dbfd32332d5f61f709036bc4 ?

python/tskit/tables.py
```diff
@@ -1073,6 +1020,55 @@ def __getattr__(self, name):
...
+            indexed_offsets = offset_array[indices]
+            next_offsets = offset_array[indices + 1]
+            lengths = next_offsets - indexed_offsets
+            lengths64 = lengths.astype(np.int64, copy=False)
+            total = int(lengths64.sum())
+            if total == 0:
+                result = np.empty(0, dtype=full_array.dtype)
+            else:
+                result = full_array[
+                    indexed_offsets.astype(np.int64, copy=False)[
```
jeromekelleher: Can we break this one into its own function, and add a few simple unit tests? It's just not obvious to me what this is doing and giving it a name will help.

Like
```
result = full_array[get_offsets_col_from_lengths(indexed_offsets, lengths64)]
```

python/tskit/tables.py
```diff
@@ -1073,6 +1020,55 @@ def __getattr__(self, name):
...
+        start = row_slice.start
+        stop = row_slice.stop
+        if is_offset:
+            return full_array[start : stop + 1]
+        if is_ragged:
+            offset_array = getattr(self._llts, f"{self.table_name}_{name}_offset")
+            return full_array[offset_array[start] : offset_array[stop]]
+        return full_array[row_slice]
+
+    def _select_column(self, full_array, name, is_offset, is_ragged):
```
jeromekelleher, line 1056: There isn't any shared code between these two clauses, so how about
```
def _select_column_ragged(...)

def _select_column_offset(...)

File: python/tskit/tables.py [hunk 15]
```diff
@@ -988,76 +989,22 @@ def _fast_equals(self, other, *, ignore_metadata=False, ignore_timestamps=False)
...
-                            lengths, dtype=offset_array.dtype, out=offsets_result[1:]
-                        )
-                    object.__setattr__(self, f"{name}_offset", offsets_result)
+                result = result.view(np.int8)
+            if not (self._row_indices is None and self._row_slice is None):
+                is_offset = name.endswith("_offset")
+                is_ragged = f"{name}_offset" in self._set_column_names
+                if self._row_indices is None:
+                    result = self._slice_column(result, name, is_offset, is_ragged)
                 else:
```
jeromekelleher, line 1022: Control flow is clearer this way?

```pytohn
if is_ragged:
     result = self._select_column_ragged(result, name)
elif is_offset:
     result = self._select_column_offset(result, name)
else:
    result = result[indices]
```

hyanwong:
Amazing work. Thanks so much Ben.

Issue #3290, opened 2025-09-29
Benchmarking the Two-Locus Framework and LD Calculator


lkirk:
For the purposes of assessing how the two-locus framework performs relative to the LD Calculator, I've performed a series of benchmarks. In our initial benchmarks, the two-locus framework was rather slow relative to the LD Calculator (up to 10x slower). We went ahead and improved some of the low-hanging inefficiencies in the code and now I'm happy to report that the two-locus framework is within 2x the speed of the LD Calculator, and in some cases faster. I have created a document that attempts to thoroughly detail the changes made and the reasoning behind them: [benchmarks.pdf](https://github.com/user-attachments/files/22733686/benchmarks.pdf). For brevity, I've listed some of the highlights below for a high-level overview.

Here is a brief description of the changes that we made to increase performance:
1. **Base case**: The original code that is currently in tskit ([link](https://github.com/lkirk/tskit/commit/9821725e4706f61baa005e3def6fad354083973c)).
2. **Malloc out of hot loop**: Preallocate a structure of arrays used for
   temporary calculations instead of allocating and freeing for every pair of
   sites ([link](https://github.com/lkirk/tskit/pull/12/commits/462b6d936e088b5b72e4e2197a2201ee7d134670)).
3. **Refactor Bit Arrays**: Refactor bit array interface, removing the need for
   temporary arrays in many cases. All functions now take a row index as a
   parameter ([link](https://github.com/lkirk/tskit/pull/13/commits/8971aa9ef1d0c8bc422034bc1fb152746dfd24f3)).
4. **Precompute Allele Counts and Biallelic Summary Function**: Store precomputed
   bit arrays for each sample set and allele and the count of samples for each
   allele. Introduce a biallelic summary function that avoids multiple redundant
   computations, leaving the original normalized summary function for
   multiallelic sites ([link](https://github.com/lkirk/tskit/pull/14/commits/eb568405b371b8314e9ad6d4a1f928169b38b9eb)).

Python and C tests are all passing for each of these patches.

We benchmark each change (in our benchmarks, each change is layered on the next) on a set of tree sequences with the following parameters:
|parameter|value|
|---|---|
|r|1e-8|
|$N_e$|1e4|
|L|2e6|

With `ploidy=1` and sample sizes ranging from $10^2$ - $10^5$. We sample 15 replicates to obtain these results:

[...]

PR #3291, opened 2025-10-02
Optimize two-locus site operations

lkirk:
This PR is a combination of three separate modifications. They are described below and in #3290. Fixes (#3290).

* Two-locus malloc optimizations

This revision moves all malloc operations out of the hot loop in two-locus statistics, instead providing pre-allocated regions of memory that the two-locus framework will use to perform work. Instead of simply passing each pre-allocated array into each function call, we introduce a simple structure called `two_locus_work_t`, which stores the statistical results, and provides temporary arrays for storing the normalisation constants. Setup and teardown methods for this work structure are provided. Python and C tests are passing and valgrind reports no errors.

* Refactor bit array api, rename to bitset.

As discussed in #2834, this patch renames tsk_bit_array_t to tsk_bitset_t. Philosophically, we treat these as sets and not arrays, performing intersections, unions, and membership tests. Therefore, it makes sense to alter the API to use set theoretic vocabulary, describing the intent more precisely. Fundamentally, the bitset structure is a list of N independent bitsets. Each operation on two sets must select the row on which to operate. The tsk_bitset_t originally tracked `len` only, which was N, the number of sets. For convenience, we also track the `row_len`, which is the number of unsigned integers per row. If we multiply `row_len` by `TSK_BITSET_BITS`, we get the number of bits that each set (or row) in the list of bitsets will hold.

We had also discussed each set theoretic operation accepting a row index instead of a pointer to a row within the bitset object. Now, each operation accepts a row index for each bitset structure passed into the function. This simplifies the consumption of this API considerably, removing the need of storing and tracking many intermediate temporary array pointers. We also see some performance improvements from this cleanup. For DRY purposes, I've created a private macro, `BITSET_DATA_ROW`, which abstracts away the pointer arithmetic for selecting a row out of the list of sets. Because of these changes, `tsk_bit_array_get_row` is no longer needed and has been removed from the API.

This change does not change the size of the "chunk", which is the unsigned integer storing bits. It remains a 32 bit unsigned integer, which is most performant for bit counting (popcount). I've streamlined the macros used to determine which integer in the row will be used to store a particular bit. Everything now revolves around the TSK_BITSET_BITS macro, which is simply 32 and bitshift operations have been converted to unsigned integer division.

Testing has been refactored to reflect these changes, removing tests that operate on a specific rows. Tests in c and python are passing and valgrind shows no errors.

Fixes (#2834).

* Precompute A/B Counts and Biallelic Summary Func

Precompute A/B counts for each sample set. We were previously computing them redundantly each for each site pair in our results matrix. The precomputation happens in a function called `get_mutation_sample_sets`, which takes our list of sets (`tsk_bitset_t`) for each mutation and intersects the samples with a particular mutation with the sample sets passed in by the user. The result is an expanded list of sets with one set per mutation per sample set. During this operation, we compute the number of samples containing the given allele for each mutation, avoiding the need to perform redundant count operations on the data.

In addition to precomputation, we add a non-normalized version of `compute_general_two_site_stat_result` for situations where we're computing stats from biallelic loci. We dispatch the computation of the result based on the number of alleles in the two loci we're comparing. If the number of alleles in both loci is 2, then we simply perform an LD computation on the derived alleles for the two loci. As a result, we remove the need to compute a matrix of LD values, then take a weighted sum. This is much more efficient and means that we only run the full multiallelic LD routine on sites that are multiallelic.

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -1260,16 +1260,16 @@ tsk_avl_tree_int_ordered_nodes(const tsk_avl_tree_int_t *self, tsk_avl_node_int_
 }
 
 // Bit Array implementation. Allows us to store unsigned integers in a compact manner.
-// Currently implemented as an array of 32-bit unsigned integers for ease of counting.
+// Currently implemented as an array of 32-bit unsigned integers.
 
 int
-tsk_bit_array_init(tsk_bit_array_t *self, tsk_size_t num_bits, tsk_size_t length)
+tsk_bitset_init(tsk_bitset_t *self, tsk_size_t num_bits, tsk_size_t length)
 {
     int ret = 0;
 
-    self->size = (num_bits >> TSK_BIT_ARRAY_CHUNK)
-                 + (num_bits % TSK_BIT_ARRAY_NUM_BITS ? 1 : 0);
-    self->data = tsk_calloc(self->size * length, sizeof(*self->data));
+    self->row_len = (num_bits / TSK_BITSET_BITS) + (num_bits % TSK_BITSET_BITS ? 1 : 0);
+    self->len = length;
+    self->data = tsk_calloc(self->row_len * length, sizeof(*self->data));
     if (self->data == NULL) {
         ret = tsk_trace_error(TSK_ERR_NO_MEMORY);
         goto out;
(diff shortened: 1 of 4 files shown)


c/tskit/core.h
```diff
@@ -1104,29 +1104,31 @@ FILE *tsk_get_debug_stream(void);
 
 /* Bit Array functionality */
 
-typedef uint32_t tsk_bit_array_value_t;
+// define a 32-bit chunk size for our bitsets.
+// this means we'll be able to hold 32 distinct items in each 32 bit uint
+#define TSK_BITSET_BITS (tsk_size_t) 32
```
jeromekelleher:
```suggestion
#define TSK_BITSET_BITS ((tsk_size_t) 32)
```
Extra parentheses avoid unexpected casts when using the macro

c/tskit/trees.c
```diff
@@ -2589,43 +2601,92 @@ get_mutation_samples(const tsk_treeseq_t *ts, const tsk_id_t *sites, tsk_size_t
...
     tsk_safe_free(nodes);
     tsk_tree_free(&tree);
-    tsk_bit_array_free(&mut_samples);
-    tsk_bit_array_free(&all_samples_bits);
+    tsk_bitset_free(&mut_samples);
+    tsk_bitset_free(&all_samples_bits);
     return ret == TSK_TREE_OK ? 0 : ret;
 }
 
+static int
```
jeromekelleher: Can you give a quick comment here defining what this function does, and what the memory management cycle is? So, it looks like if it succeeds, the caller is responsible for freeing the allele_sample_sets and allele_sample_set_counts?

c/tskit/trees.c
```diff
@@ -2589,43 +2601,92 @@ get_mutation_samples(const tsk_treeseq_t *ts, const tsk_id_t *sites, tsk_size_t
...
+                sample = (tsk_bitset_val_t) sample_index_map[sample_sets[k + ss_off]];
+                if (tsk_bitset_contains(allele_samples, i, sample)) {
+                    tsk_bitset_set_bit(allele_sample_sets, j + i * num_sample_sets, k);
+                    (*allele_sample_set_counts)[j + i * num_sample_sets]++;
+                }
+            }
+            ss_off += sample_set_sizes[j];
+        }
+    }
+out:
```
jeromekelleher: I think there's memory leak here if allele_sample_set_counts succeeds but tsk_bitset_init fails. I think to be watertight and for simplicity we should free both these in error conditions.
```c
out:
     if (ret != 0) {
           tsk_safe_free(*allele_sample_set_counts);
           tsk_bitset_free(...)
     }
```

c/tskit/trees.c
```diff
@@ -2589,43 +2601,92 @@ get_mutation_samples(const tsk_treeseq_t *ts, const tsk_id_t *sites, tsk_size_t
...
-    tsk_bit_array_free(&all_samples_bits);
+    tsk_bitset_free(&mut_samples);
+    tsk_bitset_free(&all_samples_bits);
     return ret == TSK_TREE_OK ? 0 : ret;
 }
 
+static int
+get_mutation_sample_sets(const tsk_bitset_t *allele_samples, tsk_size_t num_sample_sets,
+    const tsk_size_t *sample_set_sizes, const tsk_id_t *sample_sets,
+    const tsk_id_t *sample_index_map, tsk_size_t *max_ss_size,
```
jeromekelleher: Generally I find it easier to follow what's a return parameter and keep track of what the function is doing if they are explicitly named as such, like ``ret_max_ss_size``, ``ret_allele_sample_set_counts`` etc. Then,  to avoid fiddly dereferencing, I'd do something like
```c
tsk_size_t max_ss_size = 0;

for (i = 0; i < num_sample_sets; i++) {
     max_ss_size = TSK_MAX(sample_set_sizes[i], max_ss_size);
}
....

// Just before successful function exit assign the return parameters
*ret_max_ss_size = max_ss_size;
out:
```

c/tskit/trees.c
```diff
@@ -2644,35 +2705,57 @@ tsk_treeseq_two_site_count_stat(const tsk_treeseq_t *self, tsk_size_t state_dim,
...
-
     n_alleles = 0;
-    for (s = 0; s < n_sites; s++) {
-        site_offsets[s] = n_alleles;
-        n_alleles += self->site_mutations_length[sites[s]] + 1;
+    max_alleles = 0;
+    for (i = 0; i < n_sites; i++) {
+        site_offsets[i] = n_alleles * num_sample_sets;
+        n_alleles += self->site_mutations_length[sites[i]] + 1;
+        if (self->site_mutations_length[sites[i]] > max_alleles) {
```
jeromekelleher: TSK_MAX is a handy shortcut for this

c/tskit/trees.c
```diff
@@ -3289,7 +3359,42 @@ tsk_treeseq_two_branch_count_stat(const tsk_treeseq_t *self, tsk_size_t state_di
...
+    tsk_size_t num_samples)
+{
+    int ret;
+    tsk_size_t j, k, l;
+    tsk_id_t u, sample_index;
+    tsk_bitset_t tmp;
+
+    tsk_memset(&tmp, 0, sizeof(tmp));
+    ret = tsk_bitset_init(&tmp, num_samples, 1);
+    if (ret != 0) {
```
jeromekelleher, line 3364: Use ``goto out`` for consistency (in case we ever add something above it which needs cleanup)

c/tskit/core.c
```diff
@@ -1278,96 +1278,111 @@ tsk_bit_array_init(tsk_bit_array_t *self, tsk_size_t num_bits, tsk_size_t length
     return ret;
 }
 
-void
-tsk_bit_array_get_row(const tsk_bit_array_t *self, tsk_size_t row, tsk_bit_array_t *out)
-{
-    out->size = self->size;
-    out->data = self->data + (row * self->size);
-}
+#define BITSET_DATA_ROW(bs, row) (bs)->data + (row) * (bs)->row_len
```
jeromekelleher: Again, safer to wrap full expression of the macro in parantheses as unexpected things can happen if used with different precedence operators.

jeromekelleher (commented):
LGTM! Overall looks like a nice cleanup with simpler code. I haven't grokked every detail, but assuming our test coverage is good I think it's fairly safe.

A few minor details pointed out, happy to merge after that.

c/tskit/core.c
lkirk: Agreed, fixed.

c/tskit/core.h
lkirk: Good catch, fixed

c/tskit/trees.c
lkirk: Thanks, this was a typo, fixed.

c/tskit/trees.c
lkirk: I've fixed this in a few place where I noticed it. I'm now using TSK_MAX any time I'm computing the max over an array.

c/tskit/trees.c
lkirk: In response to this and the next two comments:

This was an oversight on my part. All of these allocations are freed by the caller. I think if the caller is going to use the allocated memory by passing it to other functions and freeing it, the caller should go ahead and malloc it too.

I went ahead and moved these allocations into the calling function (`tsk_treeseq_two_site_count_stat`). To me it seems much more straightforward now. We have to do some simple computations, like computing the size of the largest sample set and the total number of alleles. All of this has been localized to one block in `tsk_treeseq_two_site_count_stat` with comments stating the data dependencies.

Since there are no longer any allocations in this function and `max_ss_size` is computed in the calling function, there is no longer an issue with indirection and this function can be `static void` because there is no failure path.

As to what this function does, Section 3.2.4 of my [benchmarks document](https://github.com/user-attachments/files/22733686/benchmarks.pdf) should hopefully provide some clarity. I've also added a code comment to further clarify for future readers.

c/tskit/trees.c
lkirk: This is no longer needed after refactoring this function, but I agree, this is a lot cleaner.

lkirk (commented):
I've cleaned things up a bit, please have another look. I think the memory ownership of `get_mutation_sample_sets` makes a lot more sense now. The caller allocates all memory and `get_mutation_sample_sets` fills the inputs with data.

c/tskit/core.c
```diff
@@ -1278,96 +1278,111 @@ tsk_bit_array_init(tsk_bit_array_t *self, tsk_size_t num_bits, tsk_size_t length
...
+    v = v - ((v >> 1) & 0x55555555);
+    v = (v & 0x33333333) + ((v >> 2) & 0x33333333);
+    return (((v + (v >> 4)) & 0xF0F0F0F) * 0x1010101) >> 24;
+}
+
+tsk_size_t
+tsk_bitset_count(const tsk_bitset_t *self, tsk_size_t row)
+{
+    tsk_size_t i = 0;
+    uint32_t count = 0;
```
benjeffery: Nit-picky, but strictly speaking, with `_TSK_BIG_TABLES` and enough samples this would silently give the wrong count?

lkirk: Good point, and I'm even casting to a `tsk_size_t` at the end. I was trying to avoid 32->64 bit casting in this loop, but I doubt it'll make much difference in the end. I've switched to `tsk_size_t`. If I understand correctly, the change will address the issue you're concerned with?

benjeffery: LGTM!

PR #3292, opened 2025-10-08
Complete string methods and reduce dupe in CPython code

benjeffery:
Extracted from the immutable work. When adding the new string accessors, I also reduced duplication in the CPython code a bit.
The diff looks worse than this is as it removes so much code that things don't line up anymore.
The basic gist is that we define a macro for all the common table methods, and for all the array access methods.

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -1037,72 +1037,390 @@ out:
 }
 
 /*===================================================================
+ * Table Macros
+ *===================================================================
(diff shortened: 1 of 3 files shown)


jeromekelleher:
This is a bit more macro magic than I'm comfortable with, but I do see the advantages. As it's such a big diff I think we need to do some memory integrity checks. I forget how we do this here - can you run them please?

PR #3294, opened 2025-10-14
Add isolated_as_missing to ts.alignements

### LINKED ISSUE #3293: Enable ``alignments`` on sc2ts ARG internal nodes
The current implementation of ``alignments`` and (ultimately) ``variants`` is too strict about preventing people from using it on internal nodes. We have to do a lot of faddling around like [this](https://github.com/jeromekelleher/sc2ts-paper/blob/main/arg_postprocessing/scripts/get_all_nodes_fasta.py) to get it t work, and even then it still doesn't work on the distributed version because it has a disconnected node (which we're not interested in).

This is core functionality for sc2ts users, so we want to make this possible. Even if it's just a flag saying "do this anyway" that's better than the current status.

Related:

- #1896 
- #1908

Can you have a look and see what the options are please @benjeffery ?

benjeffery:
Fixes #3293 

This seemed to be quite straight-forward? Maybe when we added `isolated_as_missing` we just didn't apply it to `alignements`?

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,5 +1,5 @@
 ----------------------
-[1.0.0b1] - 2025-09-24
+[1.0.0b2] - 2025-XX-XX
 ----------------------
 
 **Breaking Changes**
@@ -20,6 +20,11 @@
 
 - ``draw_svg()`` methods now associate tree branches with edge IDs
   (:user:`hyanwong`, :pr:`3193`, :issue:`557`)
+
+- ``TreeSequence.alignments`` now accepts ``isolated_as_missing=False`` so that
+  alignments can be emitted for non-sample nodes (e.g., internal ARG nodes) when
+  missing data are imputed to the ancestral state. Full missing-data support
+  remains tracked in :issue:`1896`. (:user:`benjeffery`, :issue:`3293`)
 
 - ``draw_svg()`` methods now allow the y-axis to be placed on the right-hand side
   using ``y_axis="right"`` (:user:`hyanwong`, :pr:`3201`)
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -5596,6 +5596,7 @@ class TreeSequence:
         samples=None,
         left=None,
         right=None,
+        isolated_as_missing=None,
     ):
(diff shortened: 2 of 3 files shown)


python/tskit/trees.py
```diff
@@ -5679,6 +5682,11 @@ def alignments(
             (default) alignments start at 0.
         :param int right: Alignments will stop before this genomic position. If ``None``
             (default) alignments will continue until the end of the tree sequence.
+        :param bool isolated_as_missing: If True (default), isolated samples without
+            mutations are treated as missing data and this method raises an error
+            when any are detected. If False, missing data are imputed with the
```
jeromekelleher, line 5691:
```suggestion
            when any are detected. If False, missing data is imputed with the
```

python/CHANGELOG.rst
```diff
@@ -21,6 +21,10 @@
 - ``draw_svg()`` methods now associate tree branches with edge IDs
   (:user:`hyanwong`, :pr:`3193`, :issue:`557`)
 
+- ``TreeSequence.alignments`` now accepts ``isolated_as_missing=False`` so that
+  alignments can be emitted for non-sample nodes (e.g., internal ARG nodes) when
+  missing data are imputed to the ancestral state. (:user:`benjeffery`, :issue:`3293`)
```
```suggestion
  missing data is imputed to the ancestral state. (:user:`benjeffery`, :issue:`3293`)
```

python/tskit/trees.py
```diff
@@ -5730,7 +5740,9 @@ def alignments(
         # incorrectly if have a sample isolated over the region (a, b],
         # and if we have sites at each position from a to b, and at
         # each site there is a mutation over the isolated sample.
-        if any(tree._has_isolated_samples() for tree in self.trees()):
+        if isolated_as_missing and any(
+            tree._has_isolated_samples() for tree in self.trees()
+        ):
             raise ValueError(
                 "Missing data not currently supported in alignments; see "
```
jeromekelleher, line 5753: Let's update this to suggest using ``isolated_as_missing``, but note some caveats?

benjeffery:
I've updated to add more detail in the docs and error.

PR #3300, opened 2025-10-21
enable span_normalise for genetic_relatedness_vector

petrelharp:
We just forgot to do anything with this argument at the C level. Whoops.

This is straightforward... except that because previously it was doing `span_normalise=False`, and the default is now `True` in agreement with other methods (see [`genetic_relatedness`](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness); [`genetic_relatedness_matrix`](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness_matrix)), so the **output will change** for anyone who was using this. The solution is to set `span_normalise=False` (which previously did nothing).

### DIFF
--- a/c/tskit/trees.c
+++ b/c/tskit/trees.c
@@ -10810,9 +10810,6 @@ tsk_treeseq_genetic_relatedness_vector(const tsk_treeseq_t *self, tsk_size_t num
         tsk_matvec_calculator_print_state(&calc, tsk_get_debug_stream());
     }
     ret = tsk_matvec_calculator_run(&calc);
-    if (ret != 0) {
-        goto out;
-    }
 out:
     tsk_matvec_calculator_free(&calc);
     return ret;


jeromekelleher (approved):
LGTM!

python/CHANGELOG.rst
```diff
@@ -59,6 +59,11 @@
   (ensuring that parent mutations occur before children), then node, then
   their original order in the tables. (:user:`benjeffery`, :pr:`3257`, :issue:`3253`)
 
+- Fix bug in ``TreeSequence.genetic_relatedness_vector`` that previously ignored
```
benjeffery, line 62: This should go in breaking changes?

jeromekelleher: I think it's legitimately a bugfix because the behaviour we're breaking was a clear bug (and the code is pretty fresh).

benjeffery:
Whoops looks like this needed a squash - I will do some surgery on `main` to fix.

PR #3301, opened 2025-10-22
Remove guard against internal samples when decoding genotypes

benjeffery:
@jeromekelleher This seems to be the most straightforward way of supporting internal nodes with missingness - if we have a specific set of nodes we traverse to check which are internal and missing.

### DIFF
--- a/c/tskit/genotypes.c
+++ b/c/tskit/genotypes.c
@@ -134,6 +134,7 @@ tsk_variant_init(tsk_variant_t *self, const tsk_treeseq_t *tree_sequence,
     tsk_size_t num_samples_alloc;
 
     tsk_memset(self, 0, sizeof(tsk_variant_t));
+    self->missingmess_cache_tree_index = TSK_NULL;
 
     /* Set site id to NULL to indicate the variant is not decoded */
     self->site.id = TSK_NULL;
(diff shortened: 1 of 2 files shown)


c/tskit/genotypes.c
```diff
@@ -218,15 +211,21 @@ tsk_variant_init(tsk_variant_t *self, const tsk_treeseq_t *tree_sequence,
...
     if (self->alt_samples != NULL) {
         self->traversal_stack = tsk_malloc(
             tsk_treeseq_get_num_nodes(tree_sequence) * sizeof(*self->traversal_stack));
         if (self->traversal_stack == NULL) {
             ret = tsk_trace_error(TSK_ERR_NO_MEMORY);
             goto out;
         }
+        self->sample_is_present
+            = tsk_malloc(num_samples_alloc * sizeof(*self->sample_is_present));
+        if (self->sample_is_present == NULL) {
```
jeromekelleher: usual practise is to bundle multiple  mem error checks together

c/tskit/genotypes.c
```diff
@@ -425,13 +425,48 @@ tsk_variant_mark_missing(tsk_variant_t *self)
...
+                    genotypes[sample_index] = TSK_MISSING_DATA;
+                    num_missing++;
+                }
+            }
+        }
+    } else {
+        tsk_memset(present, 0, self->num_samples * sizeof(*present));
+
+        for (root = left_child[N]; root != TSK_NULL; root = right_sib[root]) {
+            stack[++stack_top] = root;
```
jeromekelleher: We don't use return value of increment operators, separate into two statements. Please use the existing patterns for these kinds of traversals so I don't have to think about it.

jeromekelleher (commented):
I'm not really getting this, so I think we need to move up to Python first.

c/tskit/genotypes.c
benjeffery: Fixed

c/tskit/genotypes.c
benjeffery: Fixed

c/tskit/genotypes.c
```diff
@@ -416,6 +413,55 @@ tsk_variant_update_genotypes_traversal(
...
+            stack[stack_top] = v;
+        }
+    }
+
+    for (j = 0; j < self->num_samples; j++) {
+        u = self->samples[j];
+        present[j]
+            = present[j]
+              && !((flags[u] & TSK_NODE_IS_SAMPLE) != 0
+                     && self->tree.parent[u] == TSK_NULL && left_child[u] == TSK_NULL);
```
jeromekelleher, line 459: get a restrict pointer for self->parent also

c/tskit/genotypes.c
```diff
@@ -536,6 +593,10 @@ tsk_variant_decode(
      * mutations directly over samples should not be missing */
     num_missing = 0;
     if (!impute_missing) {
+        if (self->alt_samples != NULL
+            && self->missingmess_cache_tree_index != self->tree.index) {
+            tsk_variant_update_missing_cache(self);
```
jeromekelleher, line 598: Nice

c/tskit/genotypes.c
```diff
@@ -416,6 +413,55 @@ tsk_variant_update_genotypes_traversal(
...
+    tsk_id_t root, u, v, sample_index;
+    tsk_size_t j;
+    int stack_top;
+
+    tsk_bug_assert(self->alt_samples != NULL);
+    tsk_bug_assert(self->tree.index != TSK_NULL);
+
+    tsk_memset(present, 0, self->num_samples * sizeof(*present));
+
+    stack_top = -1;
```
jeromekelleher, line 435: This is probably a bit dumb, but why do we need to do the traversal? Can't we just do the ``num_samples`` loop below, check each sample to see if it's isolated?

benjeffery:
I think I've finally understood this now. I was getting hung up on roots and how they fit into the missingness classification but a root can never be a non-sample node without parents or children (i.e. isolated) as roots have to be reachable up the tree from samples. I think this means we don't need the traversal at all. I'm taking time to nail the test cases first, but wanted to check that my understanding is correct!

jeromekelleher:
I think you're right, yeah.

PR #3306, opened 2025-10-24
Add json+struct codec

benjeffery:
Add a metadata codec that supports JSON and a binary blob.

### DIFF
--- a/c/tskit/core.c
+++ b/c/tskit/core.c
@@ -32,10 +32,10 @@
 #include <kastore.h>
 #include <tskit/core.h>
 
-#define UUID_NUM_BYTES 16
+#define UUID_NUM_BYTES              16
 #define TSK_JSON_BINARY_HEADER_SIZE 21
 
-static const uint8_t TSK_JSON_BINARY_MAGIC[4] = { 'J', 'B', 'L', 'B' };
+static const uint8_t _tsk_json_binary_magic[4] = { 'J', 'B', 'L', 'B' };
 
 #if defined(_WIN32)
 
@@ -143,7 +143,7 @@ out:
 
 int
 tsk_json_struct_metadata_get_blob(const char *metadata, tsk_size_t metadata_length,
-    const char **json, tsk_size_t *json_length, const uint8_t **blob,
+    const char **json, tsk_size_t *json_length, const char **blob,
     tsk_size_t *blob_length)
 {
     int ret;
@@ -153,7 +153,7 @@ tsk_json_struct_metadata_get_blob(const char *metadata, tsk_size_t metadata_leng
     uint64_t header_and_json_length;
     uint64_t total_length;
     const uint8_t *bytes;
-    const uint8_t *blob_start;
+    const char *blob_start;
     const char *json_start;
 
     if (metadata == NULL || json == NULL || json_length == NULL || blob == NULL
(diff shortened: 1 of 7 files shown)


petrelharp:
This is the right way.

bhaller:
> Sorry, school holidays here so had some leave.

:->

> I'm not sure about this, it's been a long standing principle that the C code doesn't touch the contents of metadata or have anything to do with codecs. I can see a small helper that just extracts the binary buffer being helpful though. I've added that in the last commit.

This looks good to me, thanks.  I can see the desire for the C code not to get into metadata/codec stuff, but this seems like a good exception to that principle to carve out, otherwise every client of this codec would have to reinvent this (potentially bug-prone) wheel.  I think this is the right design.

As far as I can tell there is no dependence here on the JSON string being null-terminated; is that correct?  Looks like after calling `tsk_json_binary_metadata_get_blob()` the length of the JSON can just be determined by the caller based upon the offset to, or the length of, the binary blob.

And since the header length and magic bytes and such are in `core.c` and thus private, it looks like the intent is for the caller to blindly call `tsk_json_binary_metadata_get_blob()` without knowing whether the metadata is in fact compliant, and use the return value to determine compliance and raise the appropriate user-visible error?  Which seems right, just checking.

I'm happy with this, thanks!

benjeffery:
> As far as I can tell there is no dependence here on the JSON string being null-terminated; is that correct?

Yes, that's right as we have fixed lengths. Makes me think the helper should return pointers/lengths to both parts so that API users can grab and decode the JSON?

jeromekelleher:
This generally looks good to me. One question is should we provide a "setter" function in C also if we're providing a getter?

benjeffery:
Ok, I've added the layer that let's python read the binary. I think this is now complete?

benjeffery:
> Is it OK for you to have this PR sitting open for some weeks until that happens?

Yep, I'm on leave for a week now anyway. I finished this up before I left as was told it was becoming urgent for you.

bhaller:
> Is this actually ready to review and hopefully merge @benjeffery ? I ask because I think @bhaller is going to be moving to get the new SLiM version compatible with tskit, so I guess a released tskit version with this functionality would be very useful?

Yes!  Sorry, the previous work on my big SLiM project took quite a while, so I didn't get to trying this out nearly as soon as I thought I would.  But I am now optimistic that I am ready to start building on top of this feature in SLiM more or less now.  :->  I thought this had been merged and released a while ago, actually!  I'm not really sure how to build my own tskit from a feature branch like this, being rather clueless about both GitHub and Python, so yes, it'd be great if this came out in a release.  Thanks for noticing this, @hyanwong!

bhaller:
@jeromekelleher and @petrelharp I just requested a review (probably you, Jerome, but I thought Peter would want to be in the loop here also since he will use this in pyslim).  Note I have a handful of suggestions above that have not yet been implemented; I don't know whether they're good ideas or not.  I rolled my own alignment-padding for the binary chunk for now, but it does seem to me like something that ought to be built in.  Ditto for the private data like the magic bytes; I've just copied that stuff into SLiM for now.  Anyhow, see what you think.  I've got SLiM compiling with new code that uses this PR, but I haven't tested that at all since I haven't attempted to build a tskit with this PR included.  :-O  Thanks!

bhaller:
@benjeffery Just a ping on this.  It would be helpful to get this finalized, merged, and released.  How are things on your end?  Thanks!

c/tskit/core.c
```diff
@@ -33,6 +33,9 @@
 #include <tskit/core.h>
 
 #define UUID_NUM_BYTES 16
+#define TSK_JSON_BINARY_HEADER_SIZE 21
+
+static const uint8_t TSK_JSON_BINARY_MAGIC[4] = { 'J', 'B', 'L', 'B' };
```
jeromekelleher: Reserve caps for macros  - use something like ``_tsk_json_binary_magic`` here.

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+    const char *json_start;
+
+    if (metadata == NULL || json == NULL || json_length == NULL || blob == NULL
+        || blob_length == NULL) {
+        ret = tsk_trace_error(TSK_ERR_BAD_PARAM_VALUE);
+        goto out;
+    }
+    bytes = (const uint8_t *) metadata;
+    if (metadata_length < TSK_JSON_BINARY_HEADER_SIZE) {
+        ret = tsk_trace_error(TSK_ERR_FILE_FORMAT);
```
jeromekelleher: I think we should define a specific error here for JSON blob decoding. File format is too generic.

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+        goto out;
+    }
+    if (memcmp(bytes, TSK_JSON_BINARY_MAGIC, sizeof(TSK_JSON_BINARY_MAGIC)) != 0) {
+        ret = tsk_trace_error(TSK_ERR_FILE_FORMAT);
```
jeromekelleher: See last comment

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+        goto out;
+    }
+    version = bytes[4];
+    if (version != 1) {
+        ret = tsk_trace_error(TSK_ERR_FILE_VERSION_TOO_NEW);
```
jeromekelleher: I think we can just return the same ``TSK_ERR_JSON_BLOB_FORMAT`` in this case for now. Pretty unlikely there'll be multiple versions.

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+        goto out;
+    }
+    json_length_u64 = tsk_load_u64_le(bytes + 5);
+    binary_length_u64 = tsk_load_u64_le(bytes + 13);
+    if (json_length_u64 > UINT64_MAX - (uint64_t) TSK_JSON_BINARY_HEADER_SIZE) {
```
jeromekelleher, line 180: Wait, what's this testing for?

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+        ret = tsk_trace_error(TSK_ERR_FILE_FORMAT);
+        goto out;
+    }
+    header_and_json_length = (uint64_t) TSK_JSON_BINARY_HEADER_SIZE + json_length_u64;
+    if (binary_length_u64 > UINT64_MAX - header_and_json_length) {
```
jeromekelleher, line 185: It's also not obvious what we're guarding against here - overflow of the binary length?

c/tskit/core.c
```diff
@@ -121,6 +140,67 @@ tsk_generate_uuid(char *dest, int TSK_UNUSED(flags))
...
+        ret = tsk_trace_error(TSK_ERR_FILE_FORMAT);
+        goto out;
+    }
+    total_length = header_and_json_length + binary_length_u64;
+    if ((uint64_t) metadata_length < total_length) {
```
jeromekelleher, line 190: I'm finding this quite hard to follow also

Issue #3303, opened 2025-10-24
`compute_mutation_parents` errors when existing mutation parents don't make sense


petrelharp:
Here's a MWE:
```
t = tskit.TableCollection(sequence_length=1.0)

t.nodes.add_row(time=0)
t.sites.add_row(position=0.0, ancestral_state='A')
t.mutations.add_row(node=0,site=0, derived_state='C', parent=0)

t.compute_mutation_parents()
```
This errors with
```
Traceback (most recent call last):
  File "/home/peter/projects/tskit-dev/pyslim/test.py", line 9, in <module>
    t.compute_mutation_parents()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.13/site-packages/tskit/tables.py", line 3864, in compute_mutation_parents
    self._ll_tables.compute_mutation_parents()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
_tskit.LibraryError: Parent mutation refers to itself. (TSK_ERR_MUTATION_PARENT_EQUAL)
```

Similar errors occur if the `parent` points at a mutation at a different site, etcetera. This is new behavior - previously, things like [this unit test](https://github.com/tskit-dev/pyslim/blob/main/tests/test_tree_sequence.py#L925) ran fine; now they error. (That test goes through and throws out every other mutation, so the resulting `parent` references no longer make sense.)

Since `compute_mutation_parents` does not need to look at the `parents` array, I don't think it makes sense to check integrity of what's there already? Huh, I would have *thought* the solution would be just not adding `TSK_CHECK_MUTATION_PARENTS` to [this line](https://github.com/tskit-dev/tskit/pull/3212/files#diff-a7113e19292b94afe45f78ce230588105799e862a5f6f15d1af1ba81816f6e76R12442), but it is in fact not there, so I'm not sure why this is being checked?

Huh: I see in the docstring for `compute_mutation_parents` that
> This will produce an error if mutations         are not sorted (i.e., if a mutation appears before its mutation parent)        *unless* the two mutations occur on the same branch, and have unknown times        in which case there is no way to detect the error.

Perhaps the problem is something to do with #3257. I will look into it more.

[...]

PR #3307, opened 2025-10-24
don't check mutation parents if we haven't asked to check them

petrelharp:
This was an attempt at closing #3303. HOWEVER, I'm pretty sure it's WRONG, because it makes `check_integrity` *only* check mutation parent stuff if we've asked to. But we do need to usually check mutation parents.

So... I guess we need a `TSK_NO_CHECK_MUTATION_PARENTS` flag? Also?  (And, deprecate `TSK_CHECK_MUTATION_PARENTS`?)


petrelharp:
Closing following discussion in #3303.

PR #3309, opened 2025-11-03
Tolerate invalid parents when computing mutation parents

### LINKED ISSUE #3303: `compute_mutation_parents` errors when existing mutation parents don't make sense

benjeffery:
Fixes #3303

### DIFF
--- a/c/tests/test_tables.c
+++ b/c/tests/test_tables.c
@@ -10776,6 +10776,85 @@ test_check_integrity_bad_mutation_parent_topology(void)
 }
 
 static void
+test_table_collection_compute_mutation_parents_tolerates_invalid_input(void)
+{
+    int ret;
+    tsk_id_t ret_id;
+    tsk_table_collection_t tables;
+    tsk_id_t site;
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1.0;
+
+    ret_id = tsk_node_table_add_row(&tables.nodes, 0, 1.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_node_table_add_row(
+        &tables.nodes, TSK_NODE_IS_SAMPLE, 0.0, TSK_NULL, TSK_NULL, NULL, 0);
+    CU_ASSERT_FATAL(ret_id >= 0);
+    ret_id = tsk_edge_table_add_row(&tables.edges, 0.0, 1.0, 0, 1, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret_id, 0);
+    site = tsk_site_table_add_row(&tables.sites, 0.0, "A", 1, NULL, 0);
+    CU_ASSERT_FATAL(site >= 0);
+    ret_id = tsk_mutation_table_add_row(
+        &tables.mutations, site, 1, TSK_NULL, TSK_UNKNOWN_TIME, "C", 1, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret_id, 0);
+
+    ret = tsk_table_collection_build_index(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.mutations.parent[0] = 42;
+
+    ret = tsk_table_collection_compute_mutation_parents(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    CU_ASSERT_FATAL(tables.mutations.parent[0] == TSK_NULL);
+
+    tsk_table_collection_free(&tables);
+}
+
+static void
+test_table_collection_compute_mutation_parents_restores_on_error(void)
+{
+    int ret;
+    tsk_id_t ret_id;
+    tsk_table_collection_t tables;
+    tsk_id_t site;
+
+    ret = tsk_table_collection_init(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.sequence_length = 1.0;
+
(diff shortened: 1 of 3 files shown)


c/tskit/tables.c
```diff
@@ -12436,22 +12436,48 @@ tsk_table_collection_compute_mutation_parents(
...
         }
     }
 
     ret = tsk_table_collection_compute_mutation_parents_to_array(
         self, self->mutations.parent);
     if (ret != 0) {
-        goto out;
+        goto restore;
+    }
+
```
jeromekelleher, line 12472: I think it's simpler if we just have one exit point and ``goto out`` on error. We've safely gated anyway with the ``restore_parents`` flag, so I think its clear.

c/tests/test_tables.c
```diff
@@ -10775,6 +10775,41 @@ test_check_integrity_bad_mutation_parent_topology(void)
...
+    site = tsk_site_table_add_row(&tables.sites, 0.0, "A", 1, NULL, 0);
+    CU_ASSERT_FATAL(site >= 0);
+    ret = tsk_mutation_table_add_row(
+        &tables.mutations, site, 1, TSK_NULL, TSK_UNKNOWN_TIME, "C", 1, NULL, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+    ret = tsk_table_collection_build_index(&tables, 0);
+    CU_ASSERT_EQUAL_FATAL(ret, 0);
+    tables.mutations.parent[0] = 42;
+
```
jeromekelleher, line 10806: Need to check the restore behaviour on error also

python/tests/test_tables.py
```diff
@@ -3540,6 +3540,18 @@ def test_table_references(self):
...
+        parent = tables.nodes.add_row(time=1.0)
+        child = tables.nodes.add_row(time=0, flags=tskit.NODE_IS_SAMPLE)
+        tables.edges.add_row(left=0.0, right=1.0, parent=parent, child=child)
+        site = tables.sites.add_row(position=0.0, ancestral_state="A")
+        tables.mutations.add_row(site=site, node=child, derived_state="C")
+        tables.build_index()
+        tables.mutations.parent[:] = 42
+        tables.compute_mutation_parents()
+        assert tables.mutations.parent[0] == tskit.NULL
+
```
jeromekelleher, line 3554: Another test checking the parent data is preserved on error

c/tskit/tables.c
petrelharp: agree - currently it reads as if an error in `compute_mutation_parents_to_array` might try to restore even if there's no backup (because of `TSK_NO_CHECK_INTEGRITY`). it won't because of the flag, but still.

python/tests/test_tables.py
petrelharp: if possible we should check the data is preserved on error even if `TSK_NO_CHECK_INTEGRITY` is on

c/tests/test_tables.c
benjeffery: https://github.com/tskit-dev/tskit/pull/3309/commits/193af2205c1252fcd41e1b02627b475787c674ea

c/tskit/tables.c
benjeffery: https://github.com/tskit-dev/tskit/pull/3309/commits/f3836169292797e32053adaf0f81b33d56905610

python/tests/test_tables.py
benjeffery: https://github.com/tskit-dev/tskit/pull/3309/commits/a3e12f791a705030556d94a23fe0cb3147617c46

PR #3312, opened 2025-11-04
Add link-ancestors to TreeSequence and ImmutableTableCollection

### LINKED ISSUE #3311: `link_ancestors` probably should work on immutable tables, but does not?
Currently
```
>>> ts = tskit.Tree.generate_comb(2).tree_sequence
>>> ts.link_ancestors()
  File "<python-input-6>", line 1, in <module>
    ts.link_ancestors()
    ^^^^^^^^^^^^^^^^^
AttributeError: 'TreeSequence' object has no attribute 'link_ancestors'
>>> ts.tables.link_ancestors()
  File "<python-input-7>", line 1, in <module>
    ts.tables.link_ancestors()
    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/projects/tskit-dev/tsvenv/lib/python3.13/site-packages/tskit/tables.py", line 4817, in __getattr__
    raise ImmutableTableError(
    ...<2 lines>...
    )
tskit.exceptions.ImmutableTableError: Cannot call link_ancestors() on immutable table collection. Use TreeSequence.dump_tables() for mutable copy.
```
but the docs for `link_ancestors` says
```
Returns an [EdgeTable](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.EdgeTable) instance describing a subset of the genealogical relationships between the nodes in samples and ancestors.
...
```
and mentions nothing about modifying the tables that are passed in. I have not verified that it doesn't modify the tables.

So, I think this error is in error?

benjeffery:
Fixes #3311 
As this method doesn't mutate tables - shouldn't it be on `TreeSequence` and `ImmutableTC`?

### DIFF
--- a/python/tests/test_topology.py
+++ b/python/tests/test_topology.py
@@ -4901,6 +4901,10 @@ class TestMapToAncestors:
         if compare_lib:
             lib_result = ts.dump_tables().link_ancestors(samples, ancestors)
             assert ancestor_table == lib_result
+            ts_result = ts.link_ancestors(samples, ancestors)
+            assert ancestor_table == ts_result
+            tables_result = ts.tables.link_ancestors(samples, ancestors)
+            assert ancestor_table == tables_result
         return ancestor_table
 
     def test_deprecated_name(self):
@@ -4914,6 +4918,10 @@ class TestMapToAncestors:
         tss = s.link_ancestors()
         lib_result = ts.dump_tables().map_ancestors(samples, ancestors)
         assert tss == lib_result
+        ts_result = ts.link_ancestors(samples, ancestors)
+        assert tss == ts_result
+        immutable_result = ts.tables.map_ancestors(samples, ancestors)
+        assert tss == immutable_result
         assert list(tss.parent) == [8, 8, 8, 8, 8]
         assert list(tss.child) == [0, 1, 2, 3, 4]
         assert all(tss.left) == 0
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -4780,6 +4780,21 @@ class ImmutableTableCollection(metadata.MetadataProvider):
             ]
         )
 
+    def link_ancestors(self, samples, ancestors):
+        """
+        See :meth:`TableCollection.link_ancestors`.
+        """
+        samples = util.safe_np_int_cast(samples, np.int32)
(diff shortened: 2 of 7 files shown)


petrelharp:
I think the motivation for not making this a TS method earlier was that the API seemed not totally mature. But maybe that ship has sailed.

python/tskit/tables.py
```diff
@@ -4780,6 +4780,21 @@ def __str__(self):
...
+        See :meth:`TableCollection.link_ancestors`.
+        """
+        samples = util.safe_np_int_cast(samples, np.int32)
+        ancestors = util.safe_np_int_cast(ancestors, np.int32)
+        ll_edge_table = self._llts.link_ancestors(samples, ancestors)
+        return EdgeTable(ll_table=ll_edge_table)
+
+    def map_ancestors(self, *args, **kwargs):
+        """
+        Deprecated alias for :meth:`link_ancestors`.
```
petrelharp, line 4794: maybe this is our opportunity to remove `map_ancestors`?

petrelharp: although that would doubtless break someone's code somewhere...

python/tests/test_topology.py
```diff
@@ -4901,6 +4901,11 @@ def do_map(self, ts, ancestors, samples=None, compare_lib=True):
         if compare_lib:
             lib_result = ts.dump_tables().link_ancestors(samples, ancestors)
             assert ancestor_table == lib_result
+            ts_result = ts.link_ancestors(samples, ancestors)
+            assert ancestor_table == ts_result
+            if _tskit.HAS_NUMPY_2:
```
jeromekelleher: I don't think the HAS_NUMPY_2 thing is needed, should work either way right?

python/tskit/tables.py
jeromekelleher: Let's not bother - it's not doing any harm and no point in breaking code if we don't have to

jeromekelleher (commented):
LGTM. I think we need some test coverage, including some basic tests that just check that it runs and gives a meaningful result  on a simple tree sequence (or over the example tree sequences).

Also need test coverage on the low-level meth obvs

python/tests/test_topology.py
benjeffery: Fixed

Issue #3299, opened 2025-10-16
'mutation' mode for statistics


petrelharp:
Currently we have [three modes](https://tskit.dev/tskit/docs/latest/stats.html#sec-stats-mode) for statistics. Quoting from there:

> These three examples can all be answered in the same way with the tree sequence: first, draw all the paths from one genome to the other through the tree sequence (back up to their common ancestor and back down in each marginal tree). Then, (site) count the number of mutations falling on the paths, (branch) measure the length of the paths, or (node) count how often the path goes through each node. 

However, that's a bit wrong - what's written for `site` mode is only true under an infinite-sites model of mutations (ie, not more than one mutation per site). The reason is that `site` mode is defined so that you get a result that is equal to what you'd get from the genome sequences themselves; and hence:
1. you have to use identity-by-state, and so can't know if two identical alleles came from different mutations
2. you can't see extra mutations that may have happened along the way, and so you only know "are these identical or not", rather than "how many mutations separate these two" at a given site.

However, given inferred tree sequences, what we'd arguably rather compute is the more sorts of statistics that simply "count the number of mutations falling on the paths", even if these mutations happen on the same site. This is closer to what we can more easily do theory about, etcetera. Furthermore, there's things we want to do that we *can't* do with `site` mode, including:
1. time windows: `site` mode is about allelic state, and without a 1-to-1 mapping between alleles and mutations, alleles don't have times
2. relatedness matrix operations: several of the efficient algorithms for relatedness matrices don't work if we want to use identity-by-state instead of identity-by-descent: `genetic_relatedness_matrix` and `genetic_relatedness_vector`. The reason the algorithms don't work is because we're relying on the tree to tell us who's similar to who, and for instance assuming that if nodes `a` and `b` differ by three mutations, then those three mutations also differ between everyone in the subtrees below `a` and `b`. I see no way to get around that while keeping the algorithms in their current form at all.

So, I'd like to add a new `mode` for statistics, `mode="mutation"`. Implementation is conceptually straightforward: we want to do the same thing as `mode="branch"` except the area of the branch (span multiplied by time interval) is replaced by number of mutations on the branch. One choice to be made here regards time windows: for a given time window, do we

  a. count only mutations with `time` that fall in that window, or
  b. just "spread out the mass of" mutations uniformly on the edge

[...]

PR #3314, opened 2025-11-08
Add 'mutation' mode

petrelharp:
**Edit:** more detail here. Context:
- `genetic_relatedness_matrix` is built on top of `divergence_matrix`
- `divergence_matrix` was undocumented
- `divergence_matrix`, and hence `genetic_relatedness_matrix`, uses the Scheiber-Vishkin algorithm, which has to do with TMRCAs; so, it is not suitable to `site` mode, and instead computes `mutation` mode
- however, when implemented, there was not a `mutation` mode, and `site` mode is very similar, so previously `genetic_relatedness_matrix` had `site` and `branch` mode, and a caveat in the docs saying "this isn't really quite site mode in the same sense as elsewhere in the docs"

So, here I'm
- making `genetic_relatedness_matrix` and `divergence_matrix` take "mutation" mode, and *not* site mode
- documenting `divergence_matrix`
- making other statistics mode *not* take "mutation" mode
- adding a warning to `genetic_relatedness_matrix` with `site` mode saying "you should use mutation mode" (it still throws an error)
- I've not added a warning or anything to `divergence_matrix` because it was previously undocumented.

This has very few changes, because it's just a re-naming of the previous behavior. In other words, the mechanism has not changed, just what it's called. I couldn't help but do a little bit of refactoring of some tests along the way, apologies.

This *is* a breaking change, since now `ts.genetic_relatedness_matrix(..., mode="site")` will now throw an error. Note however that  `ts.genetic_relatedness_matrix(...)` does not change, because previously the default was `mode="site"` and now it is `mode="mutation"`, which is the same thing.  So, we could allow `mode="site"` to keep working, but deprecated, but this isn't right, because it is not computing the same thing as `genetic_relatedness(..., mode="site")`; it will be computing the same thing as `genetic_relatedness(..., mode="mutation")` once the latter is implemented.

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -214,7 +214,7 @@ site
 branch
 : How long since these genomes' common ancestor?
 
-branch
+mutation
 : How many mutations have occurred since these genomes' common ancestor?
 
 node


petrelharp:
Note that the docs are a bit light on what the mode means generally, but that's intentional, since currently it's only used in these two functions.

hyanwong:
I agree with @gregorgorjanc

petrelharp:
I've added an informative warning saying to change "site" to "mutation" now, which removes any hesitation I had about the breaking change. (It still errors, though.)

docs/stats.md
```diff
@@ -204,29 +204,43 @@ statistics API in use.
...
 Roughly speaking, these answer the following sorts of question:
 
 site
-: How many mutations differentiate these two genomes?
+: At how many sites do these two genomes differ?
 
 branch
 : How long since these genomes' common ancestor?
 
+branch
```
```suggestion
mutation
```

jeromekelleher (commented):
I think this is a good idea, but I've gotten lost in what's actually being done here. Why are we updating all the code for divergence matrix, if it's not mentioned in the changelog? The changes made here feel a lot bigger than what's being discussed, but I've forgotten where we actually landed with divmat in the end.

petrelharp:
The connection is that genetic_relatedness_matrix uses divergence _matrix under the hood, so this is actually affecting both. My mistake missing it from the changelog.

benjeffery:
I also can't see any tests (I might have missed them) that aren't `discrete_genome=False` and actually have multiple mutations per site.

petrelharp:
See updated explanation above.

jeromekelleher (approved):
Great! That all makes sense, thanks.

I've logged #3320 to track documenting threading parameters.

benjeffery:
Sorry @petrelharp, I'm still confused. In #3299 you say that the existing site mode is only a valid mutation mode for a tree sequence where there is only one mutation per site (i.e. infinite sites model). If we are are renaming the existing site mode to be a true mutation mode, don't we need to make it a valid mutation mode for all tree sequences?

petrelharp:
Dang, good catch - looks like you're right, @benjeffery. I'm not sure how that happened (but suspect it was my fault). I'll just verify (probably with a test) that this is what's going on, and if so, put up a new PR with just a corrected docstring (and documentation for `divergence_matrix`).

petrelharp:
Talking through this out loud here; I want to check that the disagreement between `genetic_relatedness_matrix` and `genetic_relatedness` is not a bug.  (**tldr; no, not a bug, things are complicated**)

In site mode, [relatedness is](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness) 
> Frequency of pairwise allelic matches in the window between two sample sets

while [divergence is](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.divergence)
>  the density of sites at which the two carry different alleles

That sure *sounds* like relatedness should equal 1 - divergence. However, that is not true, even when you use `centre=False, proportion=False`; for instance:
```
sets = [[k] for k in mts.samples()]
indexes = [(i,j) for i in range(mts.num_samples-1) for j in range(i+1, mts.num_samples)]
  mts.genetic_relatedness(sets, indexes=indexes, centre=False, span_normalise=True, proportion=False, polarised=False)
# array([0.016, 0.051, 0.045, 0.036, 0.015, 0.051])
mts.divergence(sets, indexes=indexes, span_normalise=True)
# array([0.059, 0.024, 0.03 , 0.039, 0.06 , 0.024])
```
Well, this makes sense because `genetic_relatedness` isn't counting agreement at monomorphic sites, as is convention.
Also consider this rather ominous mention of "it's complicated" in the docstring for `genetic_relatedness`:
> Note: The default for this statistic - unlike most other statistics - is polarised=True. Using the default value centre=True, setting polarised=False will only multiply the result by a factor of two for branch-mode, or site-mode if all sites are biallelic. (With multiallelic sites the difference is more complicated.)

Consider a site with three alleles with respective frequencies p1, p2, p3 and q1, q2, q3 in the two sample sets (note q != 1-p). Consulting the [summary functions](https://tskit.dev/tskit/docs/latest/stats.html#summary-functions), the contribution from the site
- to `genetic_relatedness` with `polarised=False, proportion=False` is `p1 q1 + p2 q2 + p3 q3`;
- to `divergence` is `p1 (1-q1) + p2 (1-q2) + p3 (1-q3) `.

So, in terms of these per-site contributions with these options,`divergence + genetic_relatedness = p1+p2+p3 = 1` . So we see that
```
ts.divergence(..., span_normalise=False) + ts.genetic_relatedness(..., span_normalise=False, centre=False, proportion=False, polarised=False) == ts.num_sites
```
Let's call this relationship `D(i,j) + R(i,j) = S`; therefore, as in the centred version of relatedness,
```
R(i,j) - R(i,t) - R(s,j) + R(s,t) = - D(i,j) + D(i,t) + D(s,j) - D(s,t) .
```
This is just like the relationship we use in [`genetic_relatedness_matrix`](https://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness_matrix), but *differing by a factor of 2*.
HOWEVER, nothing is wrong here - that's with `polarised=False` for `genetic_relatedness`! And, that's not the default, for good reason. 

So, is there a nice relationship between divergence and *polarised* relatedness? The difference between `polarised=True` and `False` for `genetic_relatedness` is the term of `pA qA`, where those are the ancestral allele frequencies. And that's not a term that's simply related to everything else.  I suppose the relationship we get with `P(i,j) = (polarised relatedness)` would then be `D(i,j) + P(i,j) = (total number of sites at which both i and j are not ancestral)`; since the latter quantity depends on `i` and `j`, there's no nice cancellation in the centred version.

**Conclusion:** there is *not* a nice general relationship between `divergence` and `genetic_relatedness` with `mode="site"` and (the default) `polarised=True` for `genetic_relatedness`.  What we've got (the nice relationship for bi-allelic sites) is the best we can do.

benjeffery:
Thanks @petrelharp! Glad I wasn't going mad, although it felt like it! 🤪Do we need to fix/add anything here before 1.0?

PR #3325, opened 2025-11-13
Allow duplicate nodes in variant decoding

### LINKED ISSUE #3323: Allow duplicate nodes in tsk_variant_init
One of the requirements of the new alignments code is that duplicate samples are allowed. As this code will use the existing genotype decoding machinery, we'll need to relax that requirement in the existing code.

benjeffery:
Fixes #3323


jeromekelleher:
I don't think it's worth doing, no. I tried before and it's too complicated with not *that* big a gain in flexibility. We've basically painted ourselves into a corner here I think and should just live with it.

benjeffery:
One option is to allow duplicate nodes at the `alignments` level by just copying there after decoding.

Issue #3319, opened 2025-11-11
Alignments C API


benjeffery:
#3317 Added missingness support for alignments via a pure-Python implementation. Implement a C API for this and use it. Something like
`tsk_tree_sequence_alignments(tsk_tree_sequence_t *self, const char *ref_seq, tsk_size_t ref_seq_length, tsk_id_t *nodes, double* left, double *right, char **alignments)`? Although we don't have any existing C API that returns what could be a huge chunk of memory like this. We could consider an iterator like the `variants` API, decoding a sensible amount of nodes at a time for efficiency.

jeromekelleher:
Iterator seems unnecessary to me. User can easily spilt over the input sets of nodes, right?

One thing we should do that's different to existing APIs is *not* require that the nodes are unique. Turns out there's lots of good reasons for duplicate nodes and its a faff working around it.

Left and right are just single input numbers, so don't need to be pointers?

PR #3326, opened 2025-11-14
Alignments using C API

benjeffery:
~Stacked on #3319~

### DIFF
--- a/python/_tskitmodule.c
+++ b/python/_tskitmodule.c
@@ -6146,18 +6146,21 @@ TreeSequence_decode_alignments(TreeSequence *self, PyObject *args, PyObject *kwd
     num_nodes = (tsk_size_t) PyArray_DIM(nodes_array, 0);
     nodes = PyArray_DATA(nodes_array);
 
-    Py_ssize_t total
-        = (Py_ssize_t)((size_t) num_nodes * (size_t)((tsk_size_t)(right - left)));
-    buf_obj = PyBytes_FromStringAndSize(NULL, total);
+    buf_obj = PyBytes_FromStringAndSize(
+        NULL, (Py_ssize_t)(num_nodes * (tsk_size_t)(right - left)));
     if (buf_obj == NULL) {
         goto out;
     }
     buf = PyBytes_AS_STRING(buf_obj);
 
-    Py_BEGIN_ALLOW_THREADS err = tsk_treeseq_decode_alignments(self->tree_sequence,
+    // clang-format off
+    Py_BEGIN_ALLOW_THREADS 
+    err = tsk_treeseq_decode_alignments(self->tree_sequence,
         ref_seq, (tsk_size_t) ref_len, nodes, num_nodes, left, right, missing_char, buf,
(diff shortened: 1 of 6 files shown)


python/tskit/trees.py
```diff
@@ -5751,45 +5756,34 @@ def alignments(
...
+        flat = self._ll_tree_sequence.decode_alignments(
+            ref_bytes,
+            sample_ids,
+            int(interval.left),
+            int(interval.right),
+            missing_data_character,
+            bool(isolated_as_missing),
+        )
+
+        flat_arr = np.frombuffer(
```
jeromekelleher: Can you test that this is definitely the same underlying buffer as the low-level version? A copy would be bad here.

python/_tskitmodule.c
```diff
@@ -6079,6 +6079,99 @@ TreeSequence_get_individuals_nodes(TreeSequence *self)
...
+
+    nodes_array = (PyArrayObject *) PyArray_FROMANY(
+        py_nodes, NPY_INT32, 1, 1, NPY_ARRAY_IN_ARRAY);
+    if (nodes_array == NULL) {
+        goto out;
+    }
+    num_nodes = (tsk_size_t) PyArray_DIM(nodes_array, 0);
+    nodes = PyArray_DATA(nodes_array);
+
+    Py_ssize_t total
```
jeromekelleher: This is a bit convoluted cast-wise, what about
```
total = (Py_ssize_t)(num_nodes  * (tsk_size_t)(right - left)
```
standard crib about declarations at the top.

python/_tskitmodule.c
```diff
@@ -6079,6 +6079,99 @@ TreeSequence_get_individuals_nodes(TreeSequence *self)
...
+        = (Py_ssize_t)((size_t) num_nodes * (size_t)((tsk_size_t)(right - left)));
+    buf_obj = PyBytes_FromStringAndSize(NULL, total);
+    if (buf_obj == NULL) {
+        goto out;
+    }
+    buf = PyBytes_AS_STRING(buf_obj);
+
+    Py_BEGIN_ALLOW_THREADS err = tsk_treeseq_decode_alignments(self->tree_sequence,
```
jeromekelleher: You have to turn off clang format locally to prevent it messing things up around Py_BEGIN_ALLOW_THREADS

jeromekelleher:
Also iterator pattern would allow us do things like specify a memory budget,which would be quite handy

benjeffery:
> This last point is decisive for me - something like
> 
> ```
> for a in ts.alignments(max_mem="1G"):
>      # do something with a
> ```
> 
> would be super handy, especially if it double-buffered and decoded alignments in a background thread while the foreground thread was feeding the iterator.
> 
> We **do not** need to do this for the 1.0 release though!

OMG I love this.

jeromekelleher:
LGTM -  I think we just need to verify the memory-tightness (standard loop on something medium sized) to check the Python-C bit and we're done.

PR #3329, opened 2025-11-20
Documentation/Consistency Review

benjeffery:
I thought it wise before 1.0 to do a full (automated) review that documentation lines up with the code we're running. Several hours of token crunching and 45min of human review/improvement resulted in these commits, which are best viewed as separate diffs as it's a bit of a grab-bag. I think these are all worth merging though - even if they are things we weren't planning to get to.

That there wasn't more than this to find is very reassuring, but then a lot of effort has been put in to ensure tskit's quality.

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -554,7 +554,7 @@ will always fail. Use `tskit.is_unknown_time` to detect unknown values.
 
 #### Migration requirements
 
-Given a valid set of nodes and edges, the requirements for a value set of
+Given a valid set of nodes and edges, the requirements for a valid set of
 migrations are:
 
 - `left` and `right` must be finite values that lie within the tree sequence
--- a/docs/export.md
+++ b/docs/export.md
@@ -382,19 +382,34 @@ id and individual id, and two underscores will throw an error.
 ### Modifying coordinates
 
 Tree sequence site positions can be floating point values, whereas VCF
-requires discrete, 1-based integers. The ``position_transform`` argument
-controls how tskit maps coordinates into VCF. Translating non-integral
+requires positive integers. The ``position_transform`` argument
+controls how tskit maps coordinates into VCF. Translating non-integer
 positions necessarily loses precision; by default we round to the nearest
-integer, so multiple sites may share the same output position. Because VCF
-parsers differ, we do **not** automatically shift from tskit's 0-based
-convention to VCF's 1-based convention.
+integer, so multiple sites may share the same output position. 
+Furthermore, tskit's coordinate system starts at zero,
+whereas the VCF standard requires positions to be positive,
+and so a site at position 0 is not valid in the VCF standard.
+Because VCF parsers differ, we do **not** do anything to account for this.
+
+The simplest resolution of this discrepancy in convention between tskit and VCF
+positions is deal with any site at position 0 as a special case (for instance,
+by discarding or ignoring it).
+A different interpretation of this difference between tskit's position
+and VCF's POS field
+is that they are different coordinate systems: tskit coordinates are
+"distance to the right of the left end of the chromosome",
+while VCF coordinates are "which number site, counting from the left end
+of the chromosome and starting at one".
+Under this interpretation, the solution is to supply an explicit
+``position_transform`` that adds 1 to the coordinate when outputting
+to VCF (or, using the ``"legacy"`` option described below). However, this can
+easily lead to off-by-one errors converting between the coordinate systems,
+so should only be used if you really are using 0-based coordinates for your
+tree sequence.
 
 :::{warning}
-Most VCF tools expect POS and contig coordinates to be 1-based. If your tree
-sequence already uses discrete, 0-based integer positions, leaving the default
-transform will produce off-by-one coordinates in the VCF. Supply an explicit
-``position_transform`` (for example, add 1 after rounding) or use the
-``"legacy"`` option to shift to 1-based coordinates.
+Most VCF tools cannot deal with a POS value of 0. If your tree
+sequence contains a site with position 0, this will likely cause an error.
 :::
 
 Internally, the coordinates used in the VCF output are obtained by applying
@@ -415,8 +430,7 @@ transformed position is 0, :meth:`TreeSequence.write_vcf` will raise a
 - mask the offending sites using the ``site_mask`` argument; or
 - choose a ``position_transform`` that maps 0 to a valid positive position.
 
-For example, to shift all coordinates by 1 to make them strictly
-1-based, we could define:
+For example, to shift all coordinates by 1, we could define:
 
 ```{code-cell}
 def one_based_positions(positions):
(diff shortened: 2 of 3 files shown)


docs/stats.md
```diff
@@ -35,11 +35,14 @@ Please see the {ref}`tutorial <tutorials:sec_tutorial_stats>` for examples of th
...
-handled correctly by site statistics defined here, as we always
-assign missing data to be equal to the ancestral state. Later
-versions will add this behaviour as an option and will account
-for the presence of missing data by default.
+Site statistics defined here currently treat :ref:`missing data<sec_data_model_missing_data>`
+in the same way as in earlier versions of tskit: when computing site-based
+statistics, isolated samples without mutations directly above them are treated
+as carrying the ancestral allele rather than as missing. Future versions of
+tskit may expose options to treat missing data differently in statistics; for
+now, if you need explicit control over how missing data are handled you should
```
```suggestion
now, if you need explicit control over how missing data is handled you should
```

benjeffery: :D

benjeffery: 95bad615

benjeffery:
IBD extracted to https://github.com/tskit-dev/tskit/pull/3330

docs/data-model.md
```diff
@@ -443,7 +451,7 @@ has no effect on nodes.
 Given a valid set of nodes and a sequence length {math}`L`, the simple
 requirements for each edge are:
 
-- We must have {math}`0 \leq` `left` {math}`<` `right` {math}`\leq L`;
+- We must have finite coordinates with {math}`0 \leq` `left` {math}`<` `right` {math}`\leq L`;
```
petrelharp, line 454: This is redundant, but it's fine; this is supposed to be careful and specific. (you can't have a non-finite value `x` for which `0 <= x < L` with `L` finite)

docs/data-model.md
```diff
@@ -549,16 +557,30 @@ will always fail. Use `tskit.is_unknown_time` to detect unknown values.
 Given a valid set of nodes and edges, the requirements for a value set of
```
petrelharp:
```suggestion
Given a valid set of nodes and edges, the requirements for a valid set of
```

docs/export.md
```diff
@@ -369,31 +381,53 @@ id and individual id, and two underscores will throw an error.
...
-contains positions at non-integral locations there is an information loss
-incurred by translating to VCF. By default, we round the site positions
-to the nearest integer, such that there may be several sites with the
-same integer position in the output. The second difference between VCF
-and tskit is that VCF is defined to be a 1-based coordinate system, whereas
-tskit uses 0-based. However, how coordinates are transformed depends
-on the VCF parser, and so we do **not** account for this change in
-coordinate system by default.
+Tree sequence site positions can be floating point values, whereas VCF
+requires discrete, 1-based integers. The ``position_transform`` argument
```
petrelharp:
```suggestion
requires positive integers. The ``position_transform`` argument
```

docs/export.md
```diff
@@ -369,31 +381,53 @@ id and individual id, and two underscores will throw an error.
...
+controls how tskit maps coordinates into VCF. Translating non-integral
```
petrelharp:
```suggestion
controls how tskit maps coordinates into VCF. Translating non-integer
```

docs/export.md
```diff
@@ -369,31 +381,53 @@ id and individual id, and two underscores will throw an error.
...
+positions necessarily loses precision; by default we round to the nearest
+integer, so multiple sites may share the same output position. Because VCF
+parsers differ, we do **not** automatically shift from tskit's 0-based
+convention to VCF's 1-based convention.
```
petrelharp:
```suggestion
integer, so multiple sites may share the same output position. 
Furthermore, tskit's coordinate system starts at zero,
whereas the VCF standard requires positions to be positive,
and so a site at position 0 is not valid in the VCF standard.
Because VCF parsers differ, we do **not** do anything to account for this.
```

docs/export.md
```diff
@@ -369,31 +381,53 @@ id and individual id, and two underscores will throw an error.
...
+
+:::{warning}
+Most VCF tools expect POS and contig coordinates to be 1-based. If your tree
+sequence already uses discrete, 0-based integer positions, leaving the default
+transform will produce off-by-one coordinates in the VCF. Supply an explicit
+``position_transform`` (for example, add 1 after rounding) or use the
+``"legacy"`` option to shift to 1-based coordinates.
+:::
```
petrelharp, line 413: I am not sure about how this is written. I think there are different interpretations of what's going on here:
1. "0-based / 1-based" - if I'm interpreting correctly, this means that tskit coordinates are "distance to the right of the left end of the chromosome" and VCF is "which number site, counting from the left end of the chromosome".
2. "valid values" - tskit allows a site at coordinate 0, VCF does not

Under interpretation (1), the coordinates are indeed off by one. However, I think that the simpler interpretation (2) should be the preferred interpretation, because then *they're in the same coordinate system* and there's no possibility to get mixed up with off-by-one errors.

Here's a rather pedantic suggestion for clarifying the situation.
```suggestion
The simplest resolution of this discrepancy in convention between tskit and VCF
positions is deal with any site at position 0 as a special case (for instance,
by discarding or ignoring it).
A different interpretation of this difference between tskit's position
and VCF's POS field
is that they are different coordinate systems: tskit coordinates are
"distance to the right of the left end of the chromosome",
while VCF coordinates are "which number site, counting from the left end
of the chromosome and starting at one".
Under this interpretation, the solution is to supply an explicit
``position_transform`` that adds 1 to the coordinate when outputting
to VCF (or, using the ``"legacy"`` option described below). However, this can
easily lead to off-by-one errors converting between the coordinate systems,
so should only be used if you really are using 0-based coordinates for your
tree sequence.

:::{warning}
Most VCF tools cannot deal with a POS value of 0. If your tree
sequence contains a site with position 0, this will likely cause an error.
:::
```

docs/export.md
```diff
@@ -369,31 +381,53 @@ id and individual id, and two underscores will throw an error.
...
+The VCF specification does not allow positions to be 0. By default, if any
+transformed position is 0, :meth:`TreeSequence.write_vcf` will raise a
+:class:`ValueError`. If you wish to retain these records you can either:
+
+- set ``allow_position_zero=True`` to write such sites anyway;
+- mask the offending sites using the ``site_mask`` argument; or
+- choose a ``position_transform`` that maps 0 to a valid positive position.
+
+For example, to shift all coordinates by 1 to make them strictly
+1-based, we could define:
```
petrelharp:
```suggestion
For example, to shift all coordinates by 1, we could define:
```
I'm removing "1-based" because I don't think that term is unambiguous.

docs/file-formats.md
```diff
@@ -234,8 +238,13 @@ ts.dump_text(sites=sys.stdout)
 The mutation text format must contain the columns `site`,
 `node` and `derived_state`. The `time`, `parent` and `metadata` columns
 may also be optionally present (but `parent` must be specified if
-more than one mutation occurs at the same site). If `time` is absent
-`UNKNOWN_TIME` will be used to fill the column. See the
+more than one mutation occurs at the same site). If the `time` column
+is absent, the mutation times in the resulting tree sequence are set to
+{data}`tskit.UNKNOWN_TIME`. Unknown mutation times written out by
```
petrelharp:
```suggestion
{data}`tskit.UNKNOWN_TIME`, which is a numeric value that behaves like NaN.
Unknown mutation times written out by
```

PR #3330, opened 2025-11-20
Improve IBD documentation and IdentitySegments docs

benjeffery:
Extracted from #3329

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3810,7 +3810,8 @@ class TableCollection(metadata.MetadataProvider):
         """
         Sorts the individual table in place, so that parents come before children,
         and the parent column is remapped as required. Node references to individuals
-        are also updated.
+        are also updated. This is a stricter order than is required for a valid tree
+        sequence.
         """
         self._ll_tables.sort_individuals()
         # TODO add provenance
@@ -3819,9 +3820,11 @@ class TableCollection(metadata.MetadataProvider):
         """
         This puts the tables in *canonical* form, imposing a stricter order on the
         tables than :ref:`required <sec_valid_tree_sequence_requirements>` for
-        a valid tree sequence. In particular, the individual
-        and population tables are sorted by the first node that refers to each
-        (see :meth:`TreeSequence.subset`). Then, the remaining tables are sorted
+        a valid tree sequence. In particular, the population table is sorted to
+        place populations with the lowest node IDs first, and the individual table
+        is sorted firstly as in :meth:`.sort_individuals` and secondarily
+        by the lowest ID of the nodes that refer to each individual
+        (see :meth:`TreeSequence.subset`). The remaining tables are sorted
         as in :meth:`.sort`, with the modification that mutations are sorted by
         site, then time (if known), then the mutation's node's time, then number
         of descendant mutations (ensuring that parent mutations occur before
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -7513,11 +7513,15 @@ class TreeSequence:
         the ancestry of these nodes - for that, see :meth:`.simplify`.
 
         This has the side effect that it may change the order of the nodes,
-        individuals, populations, and migrations in the tree sequence: the nodes
-        in the new tree sequence will be in the order provided in ``nodes``, and
-        both individuals and populations will be ordered by the earliest retained
-        node that refers to them. (However, ``reorder_populations`` may be set to
-        False to keep the population table unchanged.)
+        populations, individuals, and migrations in the tree sequence. Nodes
+        in the new tree sequence will be in the order provided in ``nodes``.
+        Populations will be ordered in ascending order of the lowest ID of
+        the nodes that refer to them. Individuals will be not only ordered
+        so that :attr:`~Individual.parents` come before children (see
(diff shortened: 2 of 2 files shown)


docs/ibd.md
```diff
@@ -77,7 +73,20 @@ coordinate ``[left, right)`` and ancestral node ``a`` iff the most
 recent common ancestor of the segment ``[left, right)`` in nodes ``u``
 and ``v`` is ``a``, and the segment has been inherited along the same
 genealogical path (ie. it has not been broken by recombination). The
-segments returned are the longest possible ones.
+segments returned are the longest possible ones: for a fixed pair
+``(u, v)`` we follow the ancestral paths from ``u`` and ``v`` up the
+trees and merge together adjacent genomic intervals whenever both the
+MRCA ``a`` and the full ancestral paths from ``u`` and ``v`` to ``a``
+are identical.
```
petrelharp: TODO: double-check this is true. I suspect this might need some nuance at least.

petrelharp: None of the things I checked that might be "merged" are merged - even if I split an edge, that also splits the IBD segment.

petrelharp:
```suggestion
definition of a "genealogical path" used here is
the sequence of edges, rather than nodes.
So, for instance, if ``u`` inherits a segment ``[x, z)`` from ``a``,
but that inheritance is represented by two edges,
one spanning ``[x, y)`` and the other spanning ``[y, z)``,
then this represents two genealogical paths,
and any IBD segements would be split at ``y``.
In other words, the method assumes that the end
of an edge represents a recombination,
an assumption that may not reflect how the tree sequence
is used -- see
[Fritze et al](https://doi.org/10.1093/genetics/iyaf198)
for more discussion.
```

docs/ibd.md
```diff
@@ -77,7 +73,20 @@ coordinate ``[left, right)`` and ancestral node ``a`` iff the most
...
+
+This definition is purely genealogical: it depends only on the tree
+sequence topology and node times, and does not inspect allelic
+states or mutations. In particular, if we compute the MRCA of ``(u, v)``
+in each tree along the sequence, then (up to the additional refinement
+by genealogical path) the IBD segments are obtained by merging together
+adjacent MRCA intervals that share the same ancestor and paths to that
```
petrelharp:
```suggestion
by genealogical path) the IBD segments are those
that share the same ancestor and paths to that
```

docs/ibd.md
```diff
@@ -168,19 +185,26 @@ By default we get the IBD segments between all pairs of
...
+Here we have restricted attention to the samples with node IDs 0 and 2,
+so only the pair ``(0, 2)`` appears in the result. In general:
+
+- ``within`` should be a one-dimensional array-like of node IDs
+  (typically sample nodes). All unordered pairs from this set are
+  considered.
+- If ``within`` is omitted (the default), all nodes flagged as samples
+  in the node table are used.
+- Passing an empty list, e.g. ``within=[]``, is allowed and simply
+  yields a result with zero pairs and zero segments.
```
petrelharp: Unnecessary detail.
```suggestion
```

docs/ibd.md
```diff
@@ -190,6 +214,19 @@ segments = ts.ibd_segments(between=[[0,1], [2]], store_pairs=True)
...
 
+In this example we have two sample sets, ``[0, 1]`` and ``[2]``, so the
+identity segments are computed only for pairs in which one sample lies
+in the first set and the other lies in the second. More generally:
+
+- ``between`` should be a list of non-overlapping lists of node IDs.
+- All pairs ``(u, v)`` are considered such that ``u`` and ``v`` belong
+  to different sample sets.
+- Empty sample sets are permitted (e.g., ``between=[[0, 1], []]``) and
+  simply do not contribute any pairs.
```
petrelharp:
```suggestion
```

docs/ibd.md
```diff
@@ -200,6 +237,51 @@ more details.
 The ``max_time`` and ``min_span`` arguments allow us to constrain the
 segments that we consider.
 
-```{eval-rst}
-.. todo:: Add examples for these arguments.
+The ``max_time`` argument specifies an upper bound on the time of the
+common ancestor node: only IBD segments whose MRCA node has a time
+no greater than ``max_time`` are returned. The time is measured in
+the same units as the node times in the tree sequence (e.g., generations).
```
petrelharp: What else would it be measured in?
```suggestion
no greater than ``max_time`` are returned.
```

docs/ibd.md
```diff
@@ -200,6 +237,51 @@ more details.
...
+
+The ``min_span`` argument filters by genomic length: only segments with
+span strictly greater than ``min_span`` are included. This threshold is
+measured in the same units as the ``sequence_length`` (for example,
+base pairs).
```
petrelharp:
```suggestion
span strictly greater than ``min_span`` are included.
```

petrelharp:
I made the changes above and did some more here. I think it could of course be better but is good enough to go unless someone sees typos/errors.

petrelharp (commented):
These are all changes I implemented in my commit, but some of these comments have explanation so I'm putting them in still.

hyanwong (approved):
I read it all through and LGTM

PR #3332, opened 2025-11-21
Make the README and docs intro more "ARG"y

hyanwong:
Moving to an "Era of the ARG" phrasing where the tree sequence is described as a format for storing genetic genealogies / ARGs.

Now that we have the ARG interface (for going up and down the ARG), perhaps we don't want to say that the encoding is focussed on storing the correlated trees? Although I guess it is in the name.


gregorgorjanc:
I find the term "genealogy" very confusing as it can mean just about anything (pedigree, ancestry, local tree, ...) to different people. The point that you have to say "genetic genealogy" suggests that there is a "non-genetic" genealogy too. I know we discussed this before and we will not agree - the challenge is that different communities use different words:(

hyanwong:
Good point. I just switched it to "genetic history", which I think is self-explanatory.

PR #3340, opened 2025-11-26
Update glossary.md

hyanwong:
Reference ARGs in the glossary when defining a "tree sequence". I hope this clarifies rather than adds to any confusion.

We can address the "what is a tree sequence" stuff in the tutorials too.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -57,10 +57,12 @@ tree sequence
 node
 : Any point in a tree can be associated with a particular genome
   in a particular ancestor, called a "node".  Since each node represents a
-  specific genome it has a unique `time`, thought of as its birth time. Any
-  branching point in a tree must be associated with a node; that node's time
-  determines the height of the branching point.
-  See {ref}`sec_nodes_or_individuals` for discussion of what a "node" is.
+  specific genome it has a unique `time`, thought of as its birth time. Nodes
+  may or may not correspond to branching points, either in a local
+  {ref}`tree <sec_data_model_definitions_tree>` or in the whole graph.
+  However a branching point must always be associated with a node.
+  See {ref}`sec_nodes_or_individuals` for discussion of what a "node"
+  represents.
 
 (sec_data_model_definitions_individual)=
 


docs/glossary.md
```diff
@@ -36,11 +36,16 @@ tree
...
-  encoding of a sequence of correlated trees, such as one encounters looking
-  at the gene trees along a genome. A tree sequence efficiently captures the
-  structure shared by adjacent trees, (essentially) storing only what differs
-  between them.
+: A "succinct tree sequence" (or tree sequence, for brevity) is an object
+  that stores the genetic ancestry and mutational history of a set of
+  aligned DNA sequences. The name reflects the idea that a common
+  way to treat genetic ancestry is as a sequence of correlated
+  "trees" along the genome; a tree sequence provides an efficient
+  way to store differences between these trees. Technically, ancestry
```
gregorgorjanc:
```suggestion
  way to store shared branches between these trees. Technically, ancestry
```

The edge table is storing what is shared, not what is different across local trees. I know this is nitpicking, but thinking in terms of differences I have read in many places really confused me when I started to read about what is tree sequence. The left-right algorithms work with differences, I believe.

hyanwong:
Hi @gregorgorjanc - I have reworded slightly to take account of your point, and also added a glossary item for `position`. Could you have a look through the changes and check they now make sense to you? I have also put in some links between the words in the definitions, as I think that is useful.

docs/glossary.md
```diff
@@ -30,25 +30,36 @@ Here are some definitions of some key ideas encountered in this documentation.
...
 (sec_data_model_definitions_node)=
 
 node
-: Each branching point in each tree is associated with a particular genome
+: Any point in a tree can be associated with a particular genome
   in a particular ancestor, called a "node".  Since each node represents a
-  specific genome it has a unique `time`, thought of as its birth time,
-  which determines the height of any branching points it is associated with.
+  specific genome it has a unique `time`, thought of as its birth time. Any
+  branching point in a tree much be associated with a node; that node's time
```
gregorgorjanc:
```suggestion
  branching point in a tree must be associated with a node; that node's time
```

hyanwong:
Excellent, thanks for spotting that (and accepted)

docs/glossary.md
```diff
@@ -30,25 +30,36 @@ Here are some definitions of some key ideas encountered in this documentation.
...
+  aligned DNA sequences or genomes. The name reflects the idea that a common
+  way to treat genetic ancestry is as a sequence of correlated
+  {ref}`trees <sec_data_model_definitions_tree>` at different chromosomal
+  {ref}`positions <sec_data_model_definitions_position>`.
+  Branches that are shared between these trees are efficiently stored as a
+  single {ref}`edge <sec_data_model_definitions_edge>`, and adjacent trees
+  may differ by only a few such edges. These edges connect
+  {ref}`nodes <sec_data_model_definitions_node>` (genomes) in
+  the tree sequence, forming a structure which is technically known as a
+  network or graph. Graphs of this sort are sometimes called ancestral
```
jeromekelleher:
```suggestion
  the tree sequence, forming  a
  network or graph. Graphs of this sort are sometimes called ancestral
```

If someone knows what a tree is they know what a graph is, "technically" is unnecessary here. If you really don't think it's obvious then link to a wikipedia page.

docs/glossary.md
```diff
@@ -30,25 +30,36 @@ Here are some definitions of some key ideas encountered in this documentation.
...
 (sec_data_model_definitions_node)=
 
 node
-: Each branching point in each tree is associated with a particular genome
+: Any point in a tree can be associated with a particular genome
   in a particular ancestor, called a "node".  Since each node represents a
-  specific genome it has a unique `time`, thought of as its birth time,
-  which determines the height of any branching points it is associated with.
+  specific genome it has a unique `time`, thought of as its birth time. Any
+  branching point in a tree must be associated with a node; that node's time
```
jeromekelleher: I would delete the last sentence here, "height" is ambiguous and the "branch point must be associated with a node" implies that the two things are different.

jeromekelleher: It could be worth clarifying that nodes may or may not correspond to branching points

hyanwong: Yeah, I was trying not to delete material, but I agree that is doesn't add anything

PR #3343, opened 2025-11-28
Show metadata size

### LINKED ISSUE #2637: Metadata size in tables
With tree sequences from real data, some of the metadata columns are likely to take up quite a lot of space (e.g. most of the storage). It would be helpful to know this when looking at a tree sequence e.g. with `tskit info`.

At the moment, we just report "Has Metadata: No/Yes" for each table. Might it be more informative to give "Metadata: None/XXX bytes" instead? In the html output we could simply put some sort of null marker (e.g. ∅, or just left blank) versus the byte size, rather than ✅.

We could either report the size of the `metadata` column, or the size of the `metadata` + `metadata_offset` columns. I guess the first makes a little more sense.

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/2637 - I remembered this issue when writing some recent tutorial material.

I went for the absolute size rather than percentage, and simply appended the codec type, rather than having a separate column, as it seems simpler:

I will add tests if this seems like the right format:

Screenshot 2025-11-28 at 21 06 21

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -1991,8 +1991,25 @@ class TestTreeSequence(HighLevelTestCase):
         assert len(html) > 5000
         assert f"<tr><td>Trees</td><td>{ts.num_trees:,}</td></tr>" in html
         assert f"<tr><td>Time Units</td><td>{ts.time_units}</td></tr>" in html
-        for table in ts.tables.table_name_map:
-            assert f"<td>{table.capitalize()}</td>" in html
+        codecs = collections.defaultdict(int)
+        for table_name, table in ts.tables.table_name_map.items():
+            assert f"<td>{table_name.capitalize()}</td>" in html
+            if hasattr(table, "metadata_schema"):
+                schema = table.metadata_schema.schema
+                codec = schema["codec"] if schema else "raw"
+                codecs[codec] += 1
+        assert "<td>Metadata</td>" in html
+        assert "<th>Metadata</th>" in html
+        assert "<th>Metadata size</th>" in html
+        num_tables_with_metadata = 0
+        for codec, count in codecs.items():
+            assert html.count(f">{codec}</td>") == count
+            num_tables_with_metadata += count
+        # Only one table (provenances) has no metadata
+        assert num_tables_with_metadata == len(ts.tables.table_name_map) - 1
+        # All metadata tables should show the percentage metadata size
+        assert html.count("%)</td>") == num_tables_with_metadata
+
         if ts.num_provenances > 0:
             assert (
                 f"<td>{json.loads(ts.provenance(0).record)['software']['name']}</td>"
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I think it's confusing to omit the bytes value when it's empty, just say "0 Bytes". I think it's probably easier to understand if we have two columns all right.

python/tskit/util.py
```diff
@@ -534,17 +534,29 @@ def html_table(rows, *, header):
     """
 
 
+def metadata_codec(table):
+    if hasattr(table, "metadata_schema"):
+        schema = table.metadata_schema.schema
+        return "raw" if schema is None else schema.get("codec", "unknown")
```
petrelharp, line 540: Looks like the tests don't check if this value is correct? (eg, no examples of "unknown" codec)

python/tests/test_highlevel.py
```diff
@@ -1991,8 +1991,25 @@ def test_html_repr(self, ts):
...
         assert f"<tr><td>Time Units</td><td>{ts.time_units}</td></tr>" in html
-        for table in ts.tables.table_name_map:
-            assert f"<td>{table.capitalize()}</td>" in html
+        codecs = collections.defaultdict(int)
+        for table_name, table in ts.tables.table_name_map.items():
+            assert f"<td>{table_name.capitalize()}</td>" in html
+            if hasattr(table, "metadata_schema"):
+                schema = table.metadata_schema.schema
+                codec = schema["codec"] if schema else "raw"
+                codecs[codec] += 1
```
petrelharp, line 2000: Here could check if it shows 0% for tables without metadata? And that the percent is <= 100%?

python/tskit/util.py
jeromekelleher: This is a good reason why logic like this should be spelled out longhand rather than aiming for maximally concise. Much easier to track test coverage.

python/tests/test_highlevel.py
```diff
@@ -2027,8 +2044,21 @@ def test_str(self, ts):
         assert len(s) > 999
         assert re.search(rf"║Trees *│ *{ts.num_trees}║", s)
         assert re.search(rf"║Time Units *│ *{ts.time_units}║", s)
-        for table in ts.tables.table_name_map:
-            assert re.search(rf"║{table.capitalize()} *│", s)
+        codecs = collections.defaultdict(int)
```
jeromekelleher, line 2047: I'm very much negative on this style of testing now - any complex logic in a test case is to be avoided.

I suggested testing these functions as follows

```
class TestMetadataCodecStr:
     def test_plain_table_collection(self):
             s = util.metadata_codec(tskit.TableCollection(1))
             assert s == "raw"

```

This is much more useful long term as well as being easier to write and review.

jeromekelleher (commented):
Can you test following the modern modular approach please? If you're writing complicated logic in test code you're doing it wrong, as a rule of thumb.

I agree we should update existing tests minimally to capture the change in format, but do the testing of the actual conversion in standalone unit tests which are as stupidly simple as possible.

python/tskit/util.py
hyanwong: I'm not even sure that we can have a schema with no codec, but as this is display code, I thought it better not to fail here, and I didn't really want to dive into understanding what was and wasn't acceptable as a schema. Bu I guess I should just allow this to fail if there's no codec then.

jeromekelleher: If it can't be tested then it shouldn't be in the code. You need to take responsibility here for understanding what you're putting in to the code - this is what being a maintainer entails.

python/tests/test_highlevel.py
```diff
@@ -1991,8 +1991,25 @@ def test_html_repr(self, ts):
...
         assert f"<tr><td>Trees</td><td>{ts.num_trees:,}</td></tr>" in html
         assert f"<tr><td>Time Units</td><td>{ts.time_units}</td></tr>" in html
-        for table in ts.tables.table_name_map:
-            assert f"<td>{table.capitalize()}</td>" in html
+        codecs = collections.defaultdict(int)
+        for table_name, table in ts.tables.table_name_map.items():
+            assert f"<td>{table_name.capitalize()}</td>" in html
+            if hasattr(table, "metadata_schema"):
+                schema = table.metadata_schema.schema
+                codec = schema["codec"] if schema else "raw"
```
benjeffery, line 1999: I prefer "bytes" here. "raw" implies something special?

python/tskit/util.py
benjeffery, line 540: You can't have a schema without a codec. 
As a general point of taste - defensive code is almost always indicating a lack of testing or some other smell. _Unless_ it is to do with interfacing to the real world or data you don't control. Here we control everything so defensiveness is not wanted.

python/tskit/util.py
```diff
@@ -534,17 +534,29 @@ def html_table(rows, *, header):
...
+    return ""
+
+
+def metadata_size(table):
+    if hasattr(table, "metadata"):
+        frac = len(table.metadata) / table.nbytes
```
benjeffery, line 546: I'm _pretty_ sure that table.nbytes can't be zero. Can you try to create one that is? If you can't then this is fine.

PR #3344, opened 2025-12-01
Raise error if pop passed to samples is not an integer

hyanwong:
## Description

This has bugged me for ages: you can do `ts.samples(population="AFR")` without raising an error, although the `population` param is required to be an integer (so that snippet will always return an empty array)

I thought using `int` was the simplest way to check, as it avoids `isinstance(population, string)` or equivalent, and will deal with arrays etc, but it does allow `ts.samples(population="0")`, which maybe we don't want?

# PR Checklist:

### DIFF
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -992,6 +992,17 @@ class TestNumpySamples:
             ]
         assert total == ts.num_samples
 
+    @pytest.mark.parametrize("pop", ["string", "", "0", np.arange(2), 0.0, 0.5, np.nan])
+    def test_bad_samples(self, pop):
+        ts = tskit.Tree.generate_balanced(4).tree_sequence
+        with pytest.raises(ValueError, match="must be an integer ID"):
+            ts.samples(population=pop)
+
+    @pytest.mark.parametrize("pop", [0, np.int32(0), np.int64(0), np.uint32(0)])
+    def test_good_samples(self, pop):
+        ts = msprime.sim_ancestry(2)
+        assert np.array_equiv(ts.samples(population=pop), ts.samples())
+
     @pytest.mark.parametrize("time", [0, 0.1, 1 / 3, 1 / 4, 5 / 7])
     def test_samples_time(self, time):
         ts = self.get_tree_sequence(num_demes=2, n=20, times=[time, 0.2, 1, 15])
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -6533,6 +6533,9 @@ class TreeSequence:
         samples = self._ll_tree_sequence.get_samples()
         keep = np.full(shape=samples.shape, fill_value=True)
         if population is not None:
+            if not isinstance(population, numbers.Integral):
+                raise ValueError("`population` must be an integer ID")
+            population = int(population)
             sample_population = self.nodes_population[samples]
             keep = np.logical_and(keep, sample_population == population)
         if time is not None:
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -1,4 +1,14 @@
 --------------------
+[1.0.x] - YYYY-MM-DD
+--------------------
+
+**Bugfixes**
+
+- ``ts.samples(population=...)`` now raises a ``ValueError`` if the population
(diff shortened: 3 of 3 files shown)


python/tskit/trees.py
```diff
@@ -6534,7 +6534,10 @@ def samples(self, population=None, *, population_id=None, time=None):
         keep = np.full(shape=samples.shape, fill_value=True)
         if population is not None:
             sample_population = self.nodes_population[samples]
-            keep = np.logical_and(keep, sample_population == population)
```
jeromekelleher, line 6537: Something like
```
if not isinstance(identifier, numbers.Integral):
    raise ValueError(...)
population = int(population)
```
is better because we don't want to allow ``population="0"``

Doing more than one thing in a line *and* depending on exception handling is a bad idea, because the ValueError could have come from np.logical_and also, and you've now thrown away the real exception.

python/tests/test_highlevel.py
```diff
@@ -992,6 +992,11 @@ def test_samples(self):
             ]
         assert total == ts.num_samples
 
+    def test_bad_samples(self):
```
jeromekelleher: Needs more testing. Do a parametrize here with a bunch of different bad types.

python/tests/test_highlevel.py
```diff
@@ -992,6 +992,11 @@ def test_samples(self):
...
+        ts = tskit.Tree.generate_balanced(4).tree_sequence
+        with pytest.raises(ValueError, match="must be a numeric ID"):
+            ts.samples(population="string")
```
jeromekelleher: Also need to test for good types. Need to test - python integers, other types of Python number that will pass, numpy integers

python/tskit/trees.py
hyanwong: Fair point. I had thought that explicitly testing using `isinstance` is not considered terribly Pythonic, but more than happy to go that route.

python/tests/test_highlevel.py
hyanwong: Changed to test

bad: `["string", "", np.arange(2), 0.5, np.nan]`

good: `[0, np.int32(0), np.int64(0), np.uint32(0)]`  (not sure what other python numbers to test, TBH)

petrelharp: bad: `"0", 0.0` (I think)

python/tskit/trees.py
```diff
@@ -6533,6 +6533,9 @@ def samples(self, population=None, *, population_id=None, time=None):
         samples = self._ll_tree_sequence.get_samples()
         keep = np.full(shape=samples.shape, fill_value=True)
         if population is not None:
+            if not isinstance(population, numbers.Integral):
+                raise ValueError("`population` must be a numeric ID")
```
```suggestion
                raise ValueError("`population` must be a numeric (integer) ID")
```

python/tskit/trees.py
```diff
@@ -6533,6 +6533,9 @@ def samples(self, population=None, *, population_id=None, time=None):
...
+            population = int(population)
```
petrelharp, line 6538: I don't think you want this:
```
>>> int(0.5)
0
```

jeromekelleher: I agree - but I think numbers.Integral catches this?

hyanwong: I can take it out. I just left it in there from Jerome's snippet, but I think it's probably not needed.

jeromekelleher: I think Peter is saying "we don't want to allow population=0.5" so we should test for it

hyanwong: I think I added that to the tests already (and `population=0.5` fails as expected). So it's just if we want to leave in the `population = int(population)` line in there?

jeromekelleher: Well it catches some other stuff, potentially. It doesn't do any harm, and if anything weird happens during conversion that line is highlighted

hyanwong:
I'm waiting for this to merge so I can rebase #3345 . What's the etiquette here: if there is one approval, can the OP (me) click the merge button, or should I get someone else to do it? Do we require two approvals (I think not?)

Issue #1697, opened 2021-09-13
specify samples by population name


petrelharp:
It'd be nice to do e.g.
```
ts.samples("popA")
```
where `"popA"` is the *name* of a population, rather than the index.

jeromekelleher:
I'm totally +1 for this. I think we can blur the line a bit here with the data/metadata distinction by saying it's data if we use it in the **C** library. Also, if we do decide we want to add a ``name`` field to the population at some point (which wouldn't be a terrible idea) we can easily imagine dealing with that here by saying the ``name`` field takes precedence, and if this is null we check for a metadata key called ``name``.

benjeffery:
Yes, I'm happy with this, and with the idea that the C API defines what is metadata/data. Especially as we start to standardise metadata, I see it that there are three tiers:
 - data: used by C API and in tables
 - standard metadata: used by the Python API
 - user metadata: not used by tskit

hyanwong:
Also not that you "can" do `ts.samples(population="A")` and it returns an empty array, which I think should be counted as a bug.

benjeffery:
Some more up-to-date notes as this issue got posted as dupe in #2294.

This is very close to https://github.com/tskit-dev/tskit/issues/2239 where there is a proposal for the `TreeSequence` row-accessors to get similar functionality.
Personally I would prefer an explicit `ts.samples(name='BIG')` to `ts.samples('BIG')`. One option would be to make this the equivalent of #2239 which would be `ts.samples(metadata=("name", "BIG"))`

For the syntactically sweeter options `name` is promoted to "a metadata key used by the Python API" which means the standardization mentioned in https://github.com/tskit-dev/tskit/issues/1051 is relevant.

hyanwong:
> Personally I would prefer an explicit `ts.samples(name='BIG')` to `ts.samples('BIG')`

If so, I suggest that `ts.samples('BIG')` should raise an error.

hyanwong:
I just hit this "bug" again, when I tried to use `ts.samples(population='BIG')`. Note that In @benjeffery's suggestion, it's not clear that the 'BIG' is the name of a population (not, e.g an individual). It so happens that the first argument of `ts.samples` is `population`, but that seems a bit obscure to me. Perhaps

`ts.samples(population_name='big')`

or allow any argument to take a metadata dictionary:

```python
ts.samples(population={'name': 'big')  # if dict, assume we look in the metadata
```

hyanwong:
[...]

PR #3345, opened 2025-12-02
Metadata samples

hyanwong:
While fixing #3344 I thought I might as well address this long standing issue (which is still open). This allows 

```python
samples = ts.samples(population={"name": "pop_0"})
# or shorten to
samples = ts.samples({"name": "pop_0"})
```

It doesn't promote "name" (or any other metadata key) to any special status, but simply matches the passed in dictionary against the metadata for each population, which seems neater to me than having to define special metadata fields. So you can equally do:

```python
samples = ts.samples(population={"description": "Population zero"})
```

(Fixes https://github.com/tskit-dev/tskit/issues/1697)

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -2,6 +2,11 @@
 [1.0.x] - YYYY-MM-DD
 --------------------
 
+**Features**
+
+- ``ts.samples(population=...)`` now accepts dictionaries to filter samples
+  by population metadata. (:user:`hyanwong`, :issue:`1697` :pr:`3345`)
+
 **Bugfixes**
 
 - ``ts.samples(population=...)`` now raises a ``ValueError`` if the population
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -998,10 +998,90 @@ class TestNumpySamples:
         with pytest.raises(ValueError, match="must be an integer ID"):
             ts.samples(population=pop)
 
-    @pytest.mark.parametrize("pop", [0, np.int32(0), np.int64(0), np.uint32(0)])
+    @pytest.mark.parametrize(
+        "pop", [0, np.int32(0), np.int64(0), np.uint32(0), {"name": "pop_0"}, {}]
+    )
     def test_good_samples(self, pop):
         ts = msprime.sim_ancestry(2)
+        assert ts.num_populations == 1
         assert np.array_equiv(ts.samples(population=pop), ts.samples())
+
+    @pytest.mark.parametrize(
+        "pop",
+        [
+            {"name": "nonexistent"},
+            {"name": "pop_0", "description": "nonexistent"},
+            {"name": "pop_0", "nonexistent": ""},
+        ],
+    )
+    def test_samples_metadata_no_selected(self, pop):
+        ts = msprime.sim_ancestry(2)
+        with pytest.raises(
+            ValueError, match="No populations match the specified metadata"
+        ):
+            ts.samples(population=pop)
+
+    @pytest.mark.parametrize("pop", [{"name": "pop_0"}, {}])
+    def test_samples_metadata_nopop(self, pop):
+        ts = tskit.Tree.generate_balanced(4).tree_sequence
+        assert ts.num_populations == 0
+        with pytest.raises(
+            ValueError, match="No populations match the specified metadata"
+        ):
+            ts.samples(population=pop)
+
+    def test_samples_metadata_multipop(self):
+        demography = msprime.Demography()
+        demography.add_population(name="A", initial_size=10_000)
+        demography.add_population(name="B", initial_size=5_000)
+        demography.add_population(name="C", initial_size=1_000)
+        demography.add_population_split(time=1000, derived=["A", "B"], ancestral="C")
+        samples = {"A": 1, "B": 1}
+        ts = msprime.sim_ancestry(samples, demography=demography, random_seed=12)
+        with pytest.raises(ValueError, match=r"populations \(\[0, 1, 2\]\) match"):
+            ts.samples(population={"description": ""})
+
+    @pytest.mark.parametrize(
+        "pop_param",
+        [
+            {"name": "B"},
+            {"name": "B", "description": "A&B"},
+            {"description": "A&B", "+": "B⊕C"},
+        ],
+    )
+    def test_samples_metadata_onepop(self, pop_param):
+        demography = msprime.Demography()
+        N = 100
+        demography.add_population(name="A", description="A&B", initial_size=N)
+        demography.add_population(
+            name="B", description="A&B", extra_metadata={"+": "B⊕C"}, initial_size=N
+        )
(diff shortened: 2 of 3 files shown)


python/CHANGELOG.rst
```diff
@@ -1,3 +1,13 @@
+--------------------
+[1.0.x] - YYYY-MM-DD
+--------------------
+
+**Bugfixes**
+
+- ``ts.samples(population=...)`` now raises a ``ValueError`` if the population
+  ID is e.g. a population name, rather than silently returning no samples.
+  (:user:`hyanwong`, :pr:`3344`)
+
```
petrelharp, line 15: you probably want something in the "Features" section also

python/tskit/trees.py
```diff
@@ -6515,8 +6515,11 @@ def samples(self, population=None, *, population_id=None, time=None):
         of values of the form `(min_time, max_time)`, only return sample IDs
         whose node time `t` is in this interval such that `min_time <= t < max_time`.
```
petrelharp, line 6528: Maybe a longer explanation of exactly how the 'dict' option works, like:
"
The `population` can be specified either by an integer (in which case this is the population ID)
or a dictionary matching information in the population metadata. If it is a dictionary,
it should contain key-value pair(s) that match the metadata of the desired population;
for instance, ``population={'name': 'abc'}`` will select the population that has a 'name' of 'abc'
in metadata. There should be exactly one population that has matching key-value pair(s);
if not, an error is raised.
"

python/tskit/trees.py
```diff
@@ -6533,6 +6536,24 @@ def samples(self, population=None, *, population_id=None, time=None):
...
+                        pops.append(pop.id)
+                if len(pops) == 0:
+                    raise ValueError("No populations match the specified metadata")
+                if len(pops) > 1:
+                    raise ValueError(
+                        f"Multiple populations ({pops}) match the specified metadata"
+                    )
+                population = pops[0]
+            if not isinstance(population, numbers.Integral):
+                raise ValueError("`population` must be an integer ID")
```
```suggestion
                raise ValueError("`population` must be an integer ID or a dictionary")
```

hyanwong:
Great, thanks Peter - all helpful comments which I'll address now that this seems like the way to go.

python/tskit/trees.py
hyanwong: Good spot!

python/tests/test_highlevel.py
```diff
@@ -992,6 +992,97 @@ def test_samples(self):
...
+        demography.add_population(name="A", description="A&B", initial_size=N)
+        demography.add_population(
+            name="B", description="A&B", extra_metadata={"+": "B⊕C"}, initial_size=N
+        )
+        demography.add_population(name="C", extra_metadata={"+": "B⊕C"}, initial_size=N)
+        demography.add_population_split(time=1000, derived=["A", "B"], ancestral="C")
+        ts = msprime.sim_ancestry(
+            samples={"A": 1, "B": 1}, demography=demography, random_seed=12
+        )
+        samp = ts.samples(population={"name": "B"})
```
petrelharp: Whoops - I think you aren't using your parameterized argument.
```suggestion
        samp = ts.samples(population=pop)
```

python/tskit/trees.py
```diff
@@ -6514,15 +6514,32 @@ def samples(self, population=None, *, population_id=None, time=None):
...
         :param int population_id: Deprecated alias for ``population``.
         :param float,tuple time: The time or time interval of interest. If
             None, do not filter samples by time.
         :return: A numpy array of the node IDs for the samples of interest,
             listed in numerical order.
         :rtype: numpy.ndarray (dtype=np.int32)
+        :raises ValueError: If population or time is specified incorrectly.
+        :raises ValueError: If multiple or no populations match the specified metadata.
+        :raises ValueError: If a dictionary is specified to select a population
+            but the tree sequence population metadata_schema is the null schema.
```
```suggestion
            but population metadata entries are not dictionaries. 
```

python/tskit/trees.py
```diff
@@ -6514,15 +6514,32 @@ def samples(self, population=None, *, population_id=None, time=None):
...
         :param int population_id: Deprecated alias for ``population``.
         :param float,tuple time: The time or time interval of interest. If
             None, do not filter samples by time.
         :return: A numpy array of the node IDs for the samples of interest,
             listed in numerical order.
         :rtype: numpy.ndarray (dtype=np.int32)
+        :raises ValueError: If population or time is specified incorrectly.
+        :raises ValueError: If multiple or no populations match the specified metadata.
+        :raises ValueError: If a dictionary is specified to select a population
+            but population metadata entries are not dictionaries.
```
```suggestion
            but population metadata entries are not dictionaries.
```

petrelharp:
Yep; I just had another look. I think we're good. I've just rebased.

jeromekelleher:
LGTM also. You need to formally "approve" the PR during a review to unblock merging @petrelharp. When you click the "merge when ready" button then github will merge once CI passes.

python/tskit/trees.py
```diff
@@ -6533,8 +6550,24 @@ def samples(self, population=None, *, population_id=None, time=None):
...
         keep = np.full(shape=samples.shape, fill_value=True)
         if population is not None:
+            if isinstance(population, dict):
+                # look for the key names in the population metadata: we don't expect
+                # there to be many populations, so a simple loop is fine.
+                pops = []
+                for pop in self.populations():
+                    if not isinstance(pop.metadata, dict):
+                        raise ValueError("Population metadata is not a dictionary")
+                    if set(population.items()).issubset(pop.metadata.items()):
```
jeromekelleher, line 6560: We're trying to be too concise and "clever" here. Clever code that tries to do too much at once is bad. Here's a way to do it that's clearer, more obvious and less clever.
```python
# If the query keys are a subset of the population metadata keys, we try to compare. 
# Note that *all* keys must match.
pops = []
if set(population.keys()).issubset(pop.metadata.keys()):
       for key, query_value in population.items():
              # This requires the values are comparable, which should work for nested dictionaries
              # and so on. 
              if query_value != pop[key]:
                      break
        else:
                pops.append(pop.id)
```

Note that we've ended up with some quite tricky logic which needs to be tested now. So, I think we should separate this out into it's own function that can be unit tested. Something like
```
def match_metadata(table, query):
     """
     Return the row IDs of the specified table that match the specified query dictionary. All   
     rows matching *all* key-value pairs will be returned.
      """
      # implementation above

class TestMatchMetadata:
       # test all the quirky combos of stuff here.
````

petrelharp:
Thanks, @jeromekelleher. My proposal is at this point to put this aside (and close it), as it's distracting from more urgent stuff. I think good idea @hyanwong - it seemed like low-hanging fruit - but there were some hidden cans of worms. We could get this through to the end, but probably that's not (currently) worth it because we really want a differently named function that lets us solve this use case also:

> Re real-world examples, the more useful thing would be to match e.g. population={'superpopulation': 'AFR'}, which we can't do using the PR here anyway, as we require matching with only one population.

PR #3348, opened 2025-12-03
Document migrations as a legacy feature, and tidy those docs

### LINKED ISSUE #1157: Clarify Migrations
I think the documentation of migrations could use a bit more detail here: [migration table section of the docs](https://tskit.readthedocs.io/en/latest/data-model.html#sec-migration-table-definition). Specifically, the docs read:

> Migrations are performed by individual ancestors, but most likely not by an individual whose genome is tracked as a node (as in a discrete-deme model they are unlikely to be both a migrant and a most recent common ancestor). So, tskit records when a segment of ancestry has moved between populations. 
and
> The node column records the ID of the node that was associated with the ancestry segment in question at the time of the migration event. 

I think a clearer way of explaining this is that **migrations chart how nodes are associated with a population**. Specifically, the migration node refers to the _child node of the edge where the migration occurred, so a migration from source population _x_ to destination population _y_ at a given time, node, and left/right coordinate means that at the edge (or edges) denoted by the node and coordinates, ancestors younger than the time of the migration on the relevant edges belong to population _x_ and the older ancestors along the relevant edges belong to population _y_ (or at least until an intervening migration occurs). Furthermore, all the older ancestors of the parent node of the edge which exist between the left and right coordinates also belong to population _y_ (until they are affected by an older migration) and all descendants of this edge belong to population _x_ over the left/right coordinates (until they are affected by a more recent migration).
Here's an example of when this is important: if you wanted to know which tracts of ancestry (note this does not necessarily correspond to the haplotypes since it doesn't depend on variant sites) in modern samples are the result of a historic migration, we would look at the migration node and the marginal trees existing between the left and right coordinates, and then find the relevant leaf nodes. This will give the "ancestry segments" carried by samples which are the result of migrations in the absence of intervening migrations. Note that these left/right segments do _not_ always correspond to the breakpoints between edges, which I found surprising at first.

If I have all that right, then I think we should clarify a few thing: (1) migrations explain how ancestral nodes how/why ancestral nodes have a population, (2) that the migration node is the child node of the edge where the migration occured, (3) that, barring multiple migrations on an edge, the child node of the edge belongs to source population and the parent node belongs to the destination population

hyanwong:
Also fixes #1157

As per discussion below:

> Jerome: I regard the record_migrations option in msprime to be a legacy feature that has been superseded by the [additional nodes](https://tskit.dev/msprime/docs/stable/ancestry.html#recording-more-information) feature, which provides the same information as nodes explicitly marking migrations and edges, and is much simpler.

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -295,14 +295,35 @@ required for a valid set of mutations.
 
 #### Migration Table
 
+:::{note}
+Encoding migration in the migrations table is a legacy approach
+associated with older versions of `msprime`; recording movement between
+populations in the migration table is entirely optional, even when related
+nodes are assigned to different populations.
+:::
+
+:::{warning}
+The migration table may be entirely removed from the `tskit` data model
+in the future. Meanwhile, a number of `tskit` functions, such as
+{meth}`~TreeSequence.simplify()` will raise an error if data exists in
+the migrations table. 
+:::
+
+:::{seealso}
+The {ref}`msprime:sec_ancestry_record_migrations`
+sections and the associated discussion of
+{ref}`msprime:sec_demography_migration` in the `msprime` documentation.
+:::
+
 In simulations, trees can be thought of as spread across space, and it is
 helpful for inferring demographic history to record this history.
-Migrations are performed by individual ancestors, but most likely not by an
+Migrations are performed by individual ancestors, but might not be tagged by an
 individual whose genome is tracked as a `node` (as in a discrete-deme model they are
 unlikely to be both a migrant and a most recent common ancestor).  So,
-`tskit` records when a segment of ancestry has moved between
+`tskit` can record separately when a segment of ancestry has moved between
 populations. This table is not required, even if different nodes come from
 different populations.
+
 
 | Column     | Type     | Description                                            |
 | :--------- | -------- | -----------------------------------------------------: |
@@ -316,18 +337,23 @@ different populations.
 
 
 The `left` and `right` columns are floating point values defining the
-half-open segment of genome affected. The `source` and `dest` columns
-record the IDs of the respective populations. The `node` column records the
-ID of the node that was associated with the ancestry segment in question
-at the time of the migration event. The `time` column is holds floating
-point values recording the time of the event.
+half-open segment of genome affected (these need not exactly correspond to
+breakpoints between edges). The `source` and `dest` columns record the IDs of
+the respective populations (note that by `msprime` convention, "source" and
+"destination" are defined in reverse time, see
+{ref}`msprime:sec_demography_direction_of_time`.). The `time` column
+holds floating point values recording the time of the event, with migrations
+assumed to occur instantaneously. The `node` column records the ID of the child
+node of the migrating segment; in consequence the population ID of the `node` will
+match the `src` ID (unless sequential migrations affect the same `node`, in which
+case it will match the `src` value of the youngest of those migrations).
 
 The `metadata` column provides a location for client code to store
 information about each migration. See the {ref}`sec_metadata_definition` section for
 more details on how metadata columns should be used.
 
 See the {ref}`sec_migration_requirements` section for details on the properties
-required for a valid set of mutations.
+required for a valid set of migrations.
 
 
 (sec_population_table_definition)=


docs/data-model.md
```diff
@@ -316,18 +325,22 @@ different populations.
 
 
 The `left` and `right` columns are floating point values defining the
-half-open segment of genome affected. The `source` and `dest` columns
-record the IDs of the respective populations. The `node` column records the
-ID of the node that was associated with the ancestry segment in question
-at the time of the migration event. The `time` column is holds floating
-point values recording the time of the event.
+half-open segment of genome affected (these need not exactly correspond to
+breakpoints between edges). The `source` column records the ID of the
```
jeromekelleher: This is the wrong way around AFAIK, msprime records them backwards in time.

docs/data-model.md
```diff
@@ -298,11 +298,20 @@ required for a valid set of mutations.
 In simulations, trees can be thought of as spread across space, and it is
 helpful for inferring demographic history to record this history.
 Migrations are performed by individual ancestors, but most likely not by an
-individual whose genome is tracked as a `node` (as in a discrete-deme model they are
-unlikely to be both a migrant and a most recent common ancestor).  So,
-`tskit` records when a segment of ancestry has moved between
-populations. This table is not required, even if different nodes come from
-different populations.
+individual whose genome is tracked as a `tskit` node (as in a discrete-deme
+model they are unlikely to be both a migrant and a most recent common ancestor).
```
andrewkern: this is a bit confusing to me-- i get what you mean, but had to read it twice-- guessing that someone with less experience than me would also be confused. i wonder if there is a way that this could be rephrased?

Maybe "While individuals migrate during the simulations, the ancestral chunks of those individuals is unlikely to be captured as a tskit node, because it would be the rare case in which a particular chunk would be both a migrant and an common ancestor"

docs/data-model.md
hyanwong: Good point. I should have checked. `src` is the population of the child node in msprime. ISTR confusion about this before.

docs/data-model.md
hyanwong: Although we are marking this stuff a legacy, I think it is worth rephrasing in @andrewkern 's vein, So I changed it to 

> The migrations table can be used to record when a segment of ancestry moved between populations, even if the migrating genome was not captured as a tskit node (for example, because a node is unlikely to be both a migrant and a most recent common ancestor of a set of samples).

docs/data-model.md
```diff
@@ -295,14 +295,41 @@ required for a valid set of mutations.
...
-`tskit` records when a segment of ancestry has moved between
-populations. This table is not required, even if different nodes come from
-different populations.
+helpful for inferring demographic history to record this history. The
+migrations table can be used to record when a segment of ancestry
+moved between populations, even if the migrating genome was not
+captured as a tskit node (for example, because a node is unlikely to be
+both a migrant and a most recent common ancestor of a set of samples).
+
+Specifically, migrations chart how edges are associated with a population.
```
jeromekelleher: I don't think it helps to think about edges here, and this is just confusing. There's no point in trying to explain this if we don't expect/want people to use it. 

Why are we wasting time here on a feature that we don't want people to use, and expect to remove from the data model? Who is this documentation for?

Just mark this bit of the data model as something we're likely to remove, point to the relevant msprime documentation and move on.

hyanwong: I was mainly trying to close https://github.com/tskit-dev/tskit/issues/1157 - I stole the wording from @petrelharp there, as it appears as if they were in agreement about that (and I used some of Wilder's text too). But if we think the current text is not likely to be actually wrong or confusing, I guess we can just leave it?

jeromekelleher: Let's just leave it - nobody uses this, and we don't want people to use it, so minimal docs are fine.

hyanwong: Reinstated what we previously said in that part of the text. Should be OK to merge.

docs/data-model.md
```diff
@@ -295,15 +295,36 @@ required for a valid set of mutations.
 
 #### Migration Table
 
+:::{note}
+Encoding migration in the migrations table is an legacy approach
```
jeromekelleher:
```suggestion
Encoding migration in the migrations table is a legacy approach
```

PR #3355, opened 2025-12-24
Clarify individual table sorting requirements before subset

hyanwong:
This confused me, and was the reason that I couldn't canonicalise node 0 such that it pointed to individual 0 (see below and message on slack):

```python
import tskit
print("Tskit version:", tskit.__version__)
ts = tskit.load("canonicalize_issue.trees")
print("Node in orig ts:", ts.node(0))
tables = ts.dump_tables()
tables.canonicalise()
print("Node in canonicalised ts (which I am trying to force with individual id=0):", tables.nodes[0])
ts = tables.tree_sequence()
print("Actual individual 0 in canonicalised has the following nodes:", ts.individual(0).nodes)
```

```python
Tskit version: 1.0.0
Node in orig ts: Node(id=0, flags=1, time=0.0, population=0, individual=7, metadata=b'')
Node in canonicalised ts (which I am trying to force with individual id=0): NodeTableRow(flags=1, time=0.0, population=0, individual=7, metadata=b'')
Actual individual 0 in canonicalised has the following nodes: [46 47]
```

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3807,7 +3807,8 @@ class TableCollection(metadata.MetadataProvider):
         """
         Sorts the individual table in place, so that parents come before children,
         and the parent column is remapped as required. Node references to individuals
-        are also updated.
+        are also updated. This is a stricter order than is required for a valid tree
+        sequence.
         """
         self._ll_tables.sort_individuals()
         # TODO add provenance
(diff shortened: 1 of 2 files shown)


petrelharp:
Mind pointing to the part(s) of the code from which we know this is accurate?

petrelharp:
Well, individuals have a `parent` attribute, and this is being used to count the number of descendants (as measured within the individual-parent relationships only): see [here](https://github.com/tskit-dev/tskit/blob/03ad862aa5830134d840ba150b4005ad7a5a7def/c/tskit/tables.c#L7185).

python/tskit/tables.py
```diff
@@ -3816,9 +3816,13 @@ def canonicalise(self, remove_unreferenced=None):
...
         tables than :ref:`required <sec_valid_tree_sequence_requirements>` for
-        a valid tree sequence. In particular, the individual
-        and population tables are sorted by the first node that refers to each
-        (see :meth:`TreeSequence.subset`). Then, the remaining tables are sorted
+        a valid tree sequence. In particular, the population table is sorted by
+        the lowest ID of the nodes that refer to each population, and the
+        individual table is sorted to ensure that not only are the standard
+        :ref:`sorting requirements<sec_individual_requirements>` met but
+        also secondarily sorted by the lowest ID of the nodes
+        that refer to each individual (see :meth:`TreeSequence.subset`).
```
petrelharp: There are no requirements on the individual table, according to that link:
Screenshot From 2026-01-04 17-14-11

petrelharp: This should just say something like "first on number of descendants, second on referring node ID"? Ah - also when this says "sorted by" it should specify ascending or descending.

And - is "parent before child" (or vice versa?) actually a requirement that should be said in the sec_individual_requirements?

hyanwong: Ah, good spot. I thought that was a requirement, but a bit of testing shows that it is perfectly possibly to create tree sequences with individual children before parents. Sorry for the misunderstanding.

PR #3359, opened 2026-01-04
Clarify non-span-normalised pca

### LINKED ISSUE #3358: Clarify the GRM used in ts.pca
It appears to me as if the eigenvalues reported by ts.pca are those from the centred GRM used with `ts.genetic_relatedness_matrix(..., span_normalize=False)` - this is equivalent to using `ts.genetic_relatedness(..., proportion=False, span_normalize=False)` (and with polarized=True, the default). Because `span_normalize=False` is not the default, I think it would be useful to clarify this in the `pca` docstring, perhaps in the second para. It would probably also be useful to say that the eigenvectors are conventionally normalised, so that their length is 1 and mean is 0

Here's a quick demo that it really is the non-span-normalised version:

```python
import msprime
import numpy as np

ts = msprime.sim_ancestry(5, sequence_length=200)

np_pca = np.linalg.eig(ts.genetic_relatedness_matrix(mode="branch", span_normalise=False))
print("Numpy eigen analysis")
print(np_pca.eigenvalues[:3])
print(f"sumsq = {np.sum(np_pca.eigenvectors[:,0] ** 2):.5f}",
      f"mean = {np.mean(np_pca.eigenvectors[:,0]):.5f}") 

print("\nTs PCA analysis")
ts_pca = ts.pca(3, centre=True, mode="branch")
print(ts_pca.eigenvalues)
print(f"sumsq = {np.sum(ts_pca.factors[:,0] ** 2):.5f}",
      f"mean = {np.mean(ts_pca.factors[:,0]):.5f}") 
```

Resulting e.g. in 

```
Numpy eigen analysis
[2477.15094239  303.32534931  312.51979951]
sumsq = 1.00000 mean = 0.00000

Ts PCA analysis
[2477.15094239  312.51979951  303.32534931]
sumsq = 1.00000 mean = 0.00000
```

hyanwong:
Fixes #3358

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -9299,23 +9299,28 @@ class TreeSequence:
         eigenvectors of the genetic relatedness matrix, which are obtained by a
         randomized singular value decomposition (rSVD) algorithm.
 
-        Concretely, if :math:`M` is the matrix of genetic relatedness values, with
-        :math:`M_{ij}` the output of
-        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`
-        between sample :math:`i` and sample :math:`j`, then by default this returns
-        the top ``num_components`` eigenvectors of :math:`M`, so that
+        Concretely, take :math:`M` as the matrix of non-span-normalised
+        branch-based genetic relatedness values, for instance obtained by
+        setting :math:`M_{ij}` to be the :meth:`~.TreeSequence.genetic_relatedness`
+        between sample :math:`i` and sample :math:`j` with ``mode="branch"``,
+        ``proportion=False`` and ``span_normalise=False``. Then by default this
+        returns the top ``num_components`` eigenvectors of :math:`M`, so that
         ``output.factors[i,k]`` is the position of sample `i` on the `k` th PC.
-        If ``samples`` or ``individuals`` are provided, then this does the same thing,
-        except with :math:`M_{ij}` either the relatedness between ``samples[i]``
-        and ``samples[j]`` or the nodes of ``individuals[i]`` and ``individuals[j]``,
-        respectively.
+        If ``samples`` or ``individuals`` are provided, then this does the same
+        thing, except with :math:`M_{ij}` either the relatedness between
+        ``samples[i]`` and ``samples[j]`` or the average relatedness between the
+        nodes of ``individuals[i]`` and ``individuals[j]``, respectively.
+        Factors are normalized to have L2 norm 1, i.e.,
+        ``output.factors[:,k] ** 2).sum() == 1)`` for any ``k``.
 
         The parameters ``centre`` and ``mode`` are passed to
-        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`;
-        if ``windows`` are provided then PCA is carried out separately in each window.
-        If ``time_windows`` is provided, then genetic relatedness is measured using only
-        ancestral material within the given time window (see
-        :meth:`decapitate <.TreeSequence.decapitate>` for how this is defined).
+        :meth:`~.TreeSequence.genetic_relatedness`: the default ``centre=True`` results
+        in factors whose elements sum to zero; ``mode`` currently only supports the
+        ``"branch"`` setting. If ``windows`` are provided then PCA is carried out
+        separately in each genomic window. If ``time_windows`` is provided, then genetic
+        relatedness is measured using only ancestral material within the given time
+        window (see :meth:`decapitate <.TreeSequence.decapitate>` for how this is
+        defined).
 
         So that the method scales to large tree sequences, the underlying method
         relies on a randomized SVD algorithm, using


python/tskit/trees.py
```diff
@@ -9299,20 +9299,28 @@ def pca(
...
-        Concretely, if :math:`M` is the matrix of genetic relatedness values, with
-        :math:`M_{ij}` the output of
-        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`
-        between sample :math:`i` and sample :math:`j`, then by default this returns
-        the top ``num_components`` eigenvectors of :math:`M`, so that
-        ``output.factors[i,k]`` is the position of sample `i` on the `k` th PC.
-        If ``samples`` or ``individuals`` are provided, then this does the same thing,
+        Concretely, take :math:`M` as the matrix of non-span-normalised
+        branch-based genetic relatedness values, for instance obtained from
+        :meth:`~.TreeSequence.genetic_relatedness_matrix` with
```
petrelharp: I don't think you want to say this; note from [the docs](http://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness_matrix) for `genetic_relatedness_matrix`:
> Warning: in some cases, this does not compute exactly the same thing as [genetic_relatedness()](http://tskit.dev/tskit/docs/latest/python-api.html#tskit.TreeSequence.genetic_relatedness): see below for more details. 

Perhaps just stick to the "ijth element is as from `genetic_relatedness`" phrasing as before?

python/tskit/trees.py
```diff
@@ -9299,20 +9299,28 @@ def pca(
...
+        ``mode="branch"`` and ``span_normalise=False`` or equivalently by
+        setting :math:`M_{ij}` to be the result of calculating
+        :meth:`~.TreeSequence.genetic_relatedness`
+        between sample :math:`i` and sample :math:`j` with ``mode="branch"``,
+        ``proportion=False`` and ``span_normalise=False``. Then by default this
+        returns the top ``num_components`` eigenvectors of :math:`M`, so that
```
petrelharp, line 9307: This is not actually true, since by default (`centre=True`) it is the eigenvectors of the "centred" matrix PMP.

python/tskit/trees.py
```diff
@@ -9299,20 +9299,28 @@ def pca(
...
+        ``output.factors[i,k]`` is the position of sample `i` on the `k` th PC,
+        with eigenvectors normalised such that their lengths are unity (i.e.
+        ``output.factors[:,k] ** 2).sum() == 1``). If ``samples`` or
```
petrelharp: This is a good thing to note; how about just stick this further down without referring to eigenvectors (eg "Factors are normalized to have L2 norm 1, i.e., ``output.factors[:,k] ** 2).sum() == 1``)"?

python/tskit/trees.py
hyanwong: Hmm, but `TreeSequence.genetic_relatedness` defaults to `centre=True`, so I think my definition does use the centred matrix?

python/tskit/trees.py
hyanwong: Oh yes, good catch. In this case it must be the same because .pca only uses `mode="branch"`, when the two are identical, but as you say, maybe best just to mention one method?

python/tskit/trees.py
hyanwong: Sure.

python/tskit/trees.py
petrelharp: ah good point, but yes; we do plan to implement other modes.

python/tskit/trees.py
petrelharp: hah, sorry, right, never mind.

hyanwong:
Comments resolved, ready to merge if you tick as "approved", @petrelharp. Thanks!

python/tskit/trees.py
```diff
@@ -9299,23 +9299,28 @@ def pca(
         eigenvectors of the genetic relatedness matrix, which are obtained by a
         randomized singular value decomposition (rSVD) algorithm.
 
-        Concretely, if :math:`M` is the matrix of genetic relatedness values, with
-        :math:`M_{ij}` the output of
-        :meth:`genetic_relatedness <.TreeSequence.genetic_relatedness>`
-        between sample :math:`i` and sample :math:`j`, then by default this returns
-        the top ``num_components`` eigenvectors of :math:`M`, so that
+        Concretely, take :math:`M` as the matrix of non-span-normalised
+        branch-based genetic relatedness values, for instance obtained by
```
petrelharp, line 9303:
```suggestion
        genetic relatedness values, for instance obtained by
```

python/tskit/trees.py
```diff
@@ -9299,23 +9299,28 @@ def pca(
...
+        setting :math:`M_{ij}` to be the :meth:`~.TreeSequence.genetic_relatedness`
+        between sample :math:`i` and sample :math:`j` with ``mode="branch"``,
```
petrelharp, line 9305:
```suggestion
        between sample :math:`i` and sample :math:`j` with the specified ``mode``,
```

python/tskit/trees.py
```diff
@@ -9299,23 +9299,28 @@ def pca(
...
         ``output.factors[i,k]`` is the position of sample `i` on the `k` th PC.
-        If ``samples`` or ``individuals`` are provided, then this does the same thing,
-        except with :math:`M_{ij}` either the relatedness between ``samples[i]``
-        and ``samples[j]`` or the nodes of ``individuals[i]`` and ``individuals[j]``,
-        respectively.
+        If ``samples`` or ``individuals`` are provided, then this does the same
+        thing, except with :math:`M_{ij}` either the relatedness between
+        ``samples[i]`` and ``samples[j]`` or the average relatedness between the
+        nodes of ``individuals[i]`` and ``individuals[j]``, respectively.
+        Factors are normalized to have L2 norm 1, i.e.,
```
petrelharp, line 9313:
```suggestion
        Factors are normalized to have norm 1, i.e.,
```

hyanwong:
Darn it, sorry. Thanks for the follow-up PR.

PR #3366, opened 2026-01-16
python/tskit/util.py: Add a file name to FileFormatError

### LINKED ISSUE #2467: Output name of file (if possible) when tskit.load() fails
@savitakartik had an issue that she had an array of filenames of tree sequences, and was trying to load them using `tskit.load`. A bug in her script meant that some of the files weren't tree sequences, but e.g. python scripts. She got the error message

```
FileFormatError: File not in kastore format....
```

It took a while to realise the bug. It would have been obvious if the error message had printed out the filename too, e.g.

```
FileFormatError: File `my_script.py` not in kastore format....
```

Obviously not possibly is e.g. loading from a stream, but if the ts is a file on disk, this would be useful info in the error output, I think.

alxsimon:
Closes #2467
Uses the .add_note method of exception, introduced in Python 3.11. Also hardcodes the file name in
the raised from exceptions for HDF5 or zip.
Until 3.10 is EOF, needs the sys module.

## Description

Would fix #2467

I'll tidy up (changelog and linting) if you think this change is ok.

# PR Checklist:


python/tskit/util.py
```diff
@@ -878,6 +879,10 @@ def raise_known_file_format_errors(open_file, existing_exception):
     Sniffs the file for pk-zip or hdf header bytes, then raises an exception
     if these are detected, if not raises the existing exception.
     """
+    if (sys.version_info[0] >= 3) & (sys.version_info[1] > 10):
```
jeromekelleher:
```suggestion
    if sys.version_info[0] >= 3 and sys.version_info[1] > 10:
```
The "&" style and paratheses is only needed for numpy array comparison.

jeromekelleher (commented):
LGTM. Can we add a unit test please, checking that the filename goes into the exception as expected, and validating what happens when we use something that doesn't have a name? (Although, it's not obvious to me how you could have a seekable stream that doesn't have a name, but we need to do due diligence here.)

You can pytest.skip for Python < 310.

@benjeffery shall we just ditch Python 3.10 altogether?

python/tskit/util.py
alxsimon: thanks, changed

alxsimon:
> LGTM. Can we add a unit test please, checking that the filename goes into the exception as expected, and validating what happens when we use something that doesn't have a name? (Although, it's not obvious to me how you could have a seekable stream that doesn't have a name, but we need to do due diligence here.)
> 
> You can pytest.skip for Python < 310.
> 
> @benjeffery shall we just ditch Python 3.10 altogether?

Thanks for taking a look, I've trying modifying the existing tests in test_utils.py.
I'm not sure if you'd prefer some dedicated tests that don't touch the existing ones? Especially if we consider keeping some skip for python 3.11 (which I guess would skip each test with the modification for this version of Python).

I could use some pointers on how to test the stream side of things, where should it go and how do you generally do this?

jeromekelleher:
I've had a look at this and unfortunately it's really not worth bothering with until we drop Python 3.10 (which will be later this year when it's EOL'd by Python), as it's hard to emulate ``add_note`` and we'd have to add a whole bunch of conditionals in the testing code. I'll add a note to the issue for dropping 3.10 to pick this PR up again.

PR #3390, opened 2026-02-12
Update ci uv+py311

### LINKED ISSUE #3368: Drop Python 3.10
At some point this year (2026) we'll drop support for Python 3.10. Once we've done this we can pick up PR #3366 which uses a 3.10 feature.

### LINKED ISSUE #3383: Update lowest Python in CI to 3.11
And update package pins.

jeromekelleher:
Drop Python 3.10 and migrate to using a single uv lockfile to fully pin dependencies on CI/CD. This should make maintenance easier because running ``uv lock`` periodically should do the trick and it should prevent CI from breaking due to transitive dependencies.

Closes #3383 
Closes #3368

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,34 +1,61 @@
-codecov:
-  require_ci_to_pass: false
+# ignore:
+#   - "lib/tests/"
+#   - "lib/dev-tools/"
+#   - "lib/avl.c"
+#   - "lib/subprojects/**/*"
+
 comment:
-  layout: "diff, flags, files"
-fixes:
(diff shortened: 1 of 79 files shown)


jeromekelleher:
Mostly working OK. Sticking points:

- Docs - this is quite complicated
- circleCI. Would be nice if we could port this all over to GitHub actions
- Something weird on windows 3.14.
- 3.14 is generally much too slow because of (I think) pins on Zarr < 3. 

I'm going to switch to tszip now as it's a simpler project and generally a blocker on progress now.

PR #3353, opened 2025-12-16
LD matrix documentation

lkirk:
This PR provides some documentation for the ld_matrix method so that we can consider it a finalized, public API method.

I've tried to provide a sufficient level of detail without going too deep into the method. I also provide some simple demos with the provided example tree sequences. I attempt to provide everything a user would need in the docstring for the `ld_matrix` method, I'm not sure if it's too much detail.

I didn't touch the existing `Todo` note in the `Multi site statistics` section of the documentation, I wasn't exactly sure how we wanted to handle that. To my knowledge, the `LdCalculator` still exists and it would be useful to frame this method in context with that one.

cc: @petrelharp @apragsdale

Issue #2816, opened 2023-08-11
normalization for two-locus, multiallelic stats?


petrelharp:
Over at #2805, @lkirk has implemented various two-locus stats, e.g. r^2, D, etcetera. The strategy for computing these is to sum something over all pairs of alleles (one allele from each at the two loci). As currently implemented, there is a "normalization" function so that the value for a given pair of loci is
```
  \sum_i \sum_j F_{ij} W_{ij}
```
where `F_{ij}` is the summary function calculated for the pair of alleles, and `W_{ij}` is a weighting factor (the normalization).

@lkirk has proposed several weightings:
1. uniform weighting (aka "total"); so `W_{ij} = 1/(number of pairs of alleles)`
2. product of frequencies; so `W_{ij} = p_i p_j`, where `p_i` is the frequency of allele i
3. haplotype frequency; so W_{ij} = p_{ij}`, where `p_{ij}` is the frequency of the combination of alleles `i` and `j`.

To this let me add one more:
4. unweighted; so `W_{ij} = 1`.

I suspect we don't actually need the weights at all. For either (2) or (3), we can just incorporate the weight into the summary function (and, this is how the one-locus stats work). Uniform weighting seems like it has undesireable properties - for instance, adding a single new allele as a result of genotyping error could make the resulting value change by quite a lot. However, @lkirk's reports (see [this notebook](https://github.com/lkirk/ts-two-locus-proto/blob/a6b13c4/notebooks/DemonstrateCorrectness.ipynb)) that using uniform weighting gets the right answer for some statistics, while other ones do not. I haven't dug down into what's going on, so am filing the issue for us to think about later.

I guess my first question will be: in the example where we needed uniform weighting to get the right answer, can we just change the summary function so that "unweighted" gets the right answer?

And: @lkirk, please correct me if I've got some of this wrong!

lkirk:
Thanks @petrelharp all of this is accurate. I agree that we can reduce some complexity by incorporating normalization directly into the summary functions. I'd love to figure out a way to get around uniform weighting, we should discuss this further in detail.

benjeffery:
Closing for inactivity and labelling "future", please re-open if you plan to work on this.

PR #3416, opened 2026-03-06
Documentation for LD matrix methods

apragsdale:
See #3353.

This largely pulls material from an existing, open PR to complete minimal documentation for the LD matrix methods currently available in tskit. I've made edits for clarity from the original PR, and removed some material that is possibly confusing or more information than needed for a user.

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -684,14 +684,394 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
   and {math}`v_j` is the covariance of the trait with the j-th covariate.
 
 
+(sec_stats_multi_site)=
+
 ## Multi site statistics
 
-:::{todo}
-Document statistics that use information about correlation between sites, such as
-LdCalculator (and perhaps reference {ref}`sec_identity`). Note that if we have a general
-framework which has the same calling conventions as the single site stats,
-we can rework the sections above.
+(sec_stats_two_locus)=
+
+### Two-locus statistics
+
+The {meth}`~TreeSequence.ld_matrix` method provides an interface to
+a collection of two-locus statistics with predefined summary functions (see
+{ref}`sec_stats_two_locus_summary_functions`).
+The LD matrix method differs from other
+statistics methods in that it provides a unified API with an argument to
+specify different two-locus summaries of the data. It otherwise behaves
+similarly to most other functions with respect to `sample_sets` and `indexes`.
+
+Two-locus statistics can be computed using two {ref}`modes <sec_stats_mode>`,
+either `site` or `branch`, and these should be interpreted in the same way as
+these modes in the single-site statistics. That is, the `site` mode computes LD
+over observed alleles at pairs of sites, while the `branch` model computes
+expected LD conditioned on pairs of trees.
+
+(sec_stats_two_locus_site)=
+
+#### Site mode
+
+The `"site"` mode computes two-locus statistics summarized over alleles between
+all pairs of specified sites. The default behavior, leaving `sites`
+unspecified, will compute a matrix for all pairs of sites, with one row and
+column for each site in the tree sequence (i.e., an {math}`n \times n` matrix
+where {math}`n` is the number of sites in the tree sequence). We can also
+restrict the output to a subset of sites, either by specifying a single vector
+of site indexes for both rows and columns or a pair of vectors for the row
+sites and column sites separately.
+
+The following computes a matrix of the {math}`r^2` measure of linkage
+disequilibrium (LD) computed pairwise between the first 4 sites in the tree
+sequence among all samples. The `sites` must be given as a list of lists, and
+with a single list of sites specified, we obtain a symmetric square matrix. 
+
+```{code-cell} ipython3
+ld = ts.ld_matrix(sites=[[0, 1, 2, 3]])
+print(ld)
+```
+
+If a list of two lists of site indexes is provided, these specify the row and
+column sites. For instance, here we specify 2 rows and 3 columns, which
+computes a subset of the matrix shown above.
+
+```{code-cell} ipython3
+ld = ts.ld_matrix(sites=[[1, 2], [1, 2, 3]])
+print(ld)
+```
+
+#### Computational details
+
+Because we allow for two-locus statistics to be computed for multi-allelic
+data, we need to be able to combine statistical results from each pair of
+alleles into one summary for a pair of sites. This does not affect biallelic
+data (and so this section can be skipped on first reading).
+We use two implementations for
+combining results from multiple alleles: `hap_weighted` and `total_weighted`.
+These are statistic-specific and not chosen by the user, with choices motivated
+by [Zhao (2007)](https://doi.org/10.1017/S0016672307008634).
+
+Briefly, consider a pair of sites with {math}`n` alleles at the first locus and
+{math}`m` alleles at the second. (Whether this includes the ancestral allele
+depends on whether the statistic is polarised.) Write {math}`f_{ij}` as the
+statistic computed for focal alleles {math}`A_i` and {math}`B_j`. Then the
+weighting schemes are defined as:
+
+- `hap_weighted`: {math}`\sum_{i=1}^{n}\sum_{j=1}^{m}p(A_{i}B_{j})f_{ij}`,
+  where {math}`p(A_{i}B_{j})` is the frequency of haplotype {math}`A_{i}B_{j}`.
+  This method was first introduced in [Karlin
+  (1981)](https://doi.org/10.1111/j.1469-1809.1981.tb00308.x) and reviewed in
+  [Zhao (2007)](https://doi.org/10.1017/S0016672307008634).
+
+- `total_weighted`: {math}`\frac{1}{n m}\sum_{i=1}^{n}\sum_{j=1}^{m}f_{ij}`.
+  This method assigns equal weight to each of the possible pairs of focal
+  alleles at the two sites, taking the arithmetic mean of statistics over
+  focal haplotypes.
+
+Out of all of the available summary functions, only {math}`r^2` uses
+`hap_weighted` normalisation, with the remainder using uniform weighting
+(`total_weighted`).
+
+Within this framework, statistics may be either polarised or unpolarised. For
+statistics that are polarised, we compute statistic values for pairs of derived
+alleles. (For this purpose, the "derived" alleles at a site are all alleles
+except that stored as the ``ancestral_state`` for the site.) Unpolarised
+statistics compute statistics over all pairs of alleles, derived and ancestral.
+In either case, the result is averaged over these values, using one of the
+weighting scheme (described below for each statistics). The option for
+polarisation is not exposed to the user, and we list which statistics are
+polarised below.
+
+(sec_stats_two_locus_branch)=
+
+#### Branch mode
+
+The `"branch"` mode computes expected two-locus statistics between pairs of
+trees, conditioned on the marginal topologies and branch lengths of those
+trees. The trees for which we compute statistics are specified by positions,
+and for a pair of positions we consider all possible haplotypes that could be
+generated by a single mutation occurring on each of the two trees.
+
+For two trees, one with {math}`n` branches and the other with {math}`m`
+branches, there are {math}`nm` possible pairs of branches that may carry the
+pair of mutations. For each pair, we compute the two-locus statistic, and then
+sum these values weighted by the product of the two branch lengths. Given that
+the two mutations occur, this accounts for the relative probability that the
+two mutations fall on any pair of branches.
+
+In other words, imagine we place two mutations uniformly, one on each tree, and
+then compute the statistic. The branch mode computes the expected value of the
+statistic over this process, multiplied by the product of the total branch
+lengths of each tree. This weighting accounts for mutational opportunity, so that
+the sum of the branch-mode statistic over all positions in a genomic region,
+multiplied by a mutation rate,  is equal to the expected sum of the two-locus site
+statistic over all mutations falling in that region under an infinite-sites model.
+
+The time complexity of this method is quadratic in the number of samples,
+due to the pairwise comparisons of branches from each pair of trees.
+By default, this method computes
+a symmetric matrix for all pairs of trees, with rows and columns representing
+each tree in the tree sequence. Similar to the site method, we can restrict the
+output to a subset of trees, either by specifying a vector of positions or
+a pair of vectors for row and column positions separately. To select a specific
+tree, the specified positions must land in the tree span (`[start, end)`).
+
+In the following, we compute a matrix of expected {math}`r^2` within and
+between the first 4 trees in the tree sequence. The tree breakpoints are
+a convenient way to specify those first four trees.
+
+```{code-cell} ipython3
+ld = ts.ld_matrix(
+    mode="branch",
+    positions=[ts.breakpoints(as_array=True)[0:4]]
+)
+print(ld)
+```
+
+We note that these values are quite large: as described above, the statistic is
+scaled by the product of the total branch lengths of each pair of trees. To
+compute the expected {math}`r^2` value for a pair of mutations that each land
+uniformly on the pair of trees, we can divide by the product of the total
+branch lengths:
+
+```{code-cell} ipython3
+total_branch_lengths = [tree.total_branch_length for tree in ts.trees()]
+prod_branch_lengths = np.outer(total_branch_lengths, total_branch_lengths)
+print(ld / prod_branch_lengths[0:4, 0:4])
+```
+
+To compute the average {math}`r^2` for a uniformly chosen pair of mutations, we also
+weight by tree span:
+
+```{code-cell} ipython3
+tree_spans = np.array([t.span for t in ts.trees()])
+total_opportunity = np.sum(tree_spans * total_branch_lengths)
+all_ld = ts.ld_matrix(mode="branch")
+mean_ld = np.sum(all_ld * np.outer(tree_spans, tree_spans)) / total_opportunity ** 2
+print("mean infinite-sites LD:", mean_ld)
+```
+
+As with the `"site"` mode above, we can specify the row and column trees
+separately.
+
+```{code-cell} ipython3
+breakpoints = ts.breakpoints(as_array=True)
+ld = ts.ld_matrix(
+    mode="branch",
+    positions=[breakpoints[[0]], breakpoints[0:4]]
+)
+print(ld)
+```
+
+(sec_stats_two_locus_sample_sets)=
+
+#### Sample Sets
+
+Without specifying `sample_sets` or `indexes`, the `ld_matrix()` method
+computes statistics over a single sample set that includes all samples in the
+tree sequence. The API allows for the specification of a subset or multiple
+subsets of samples, so that a separate LD matrix can be computed for each. If
+`sample_sets` is specified as a single list of samples, then a single LD matrix
+is returned. A list of lists of samples will return a 3D array containing an LD
+matrix for each list of samples.
+
+Some LD statistics can be computed between sample sets (two-way statistics are
+specified below), in which case `indexes` must be specified that reference the
+indexes of the `sample_sets`, which must be a list of lists of sample nodes.
+This results in an LD matrix computed for each list of indexes. The statistics
+are selected in the same way (with the `stat` argument), and these are limited
+to a handful of statistics (see
+{ref}`sec_stats_two_locus_summary_functions_two_way`). The dimension-dropping
+rules for the result follow the rest of the tskit stats API in that a single
+list or tuple will produce a single two-dimensional matrix, while a list of
+these will produce a three-dimensional array, with the first dimension of
+length equal to the length of the list.
+
+For example, to compute the {math}`r^2` LD matrix over a subset of samples in
+the tree sequence (such as sample nodes 0 through 7), we would specify the
+samples as follows:
+
+```{code-cell} ipython3
+ts = msprime.sim_ancestry(
+    20,
+    population_size=10000,
+    sequence_length=1000,
+    recombination_rate=2e-8,
+    random_seed=12)
+ts = msprime.sim_mutations(ts, rate=2e-8, random_seed=12)
+
+ld = ts.ld_matrix(mode="site", sample_sets=range(8))
+print(ld)
+```
+
+We would get the following dimensions with the specified
+`sample_sets` and `indexes` arguments.
+
+```
+# one-way
+ts.ld_matrix(sample_sets=None) # -> 2 dimensions
+ts.ld_matrix(sample_sets=[0, 1, 2, 3]) # -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3]]) # -> 3 dimensions
+# two-way
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=(0, 1)) # -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=[(0, 1)]) # -> 3 dimensions
+```
+
+#### Why are there `nan` values in the LD matrix?
+
+For some statistics, it is possible to observe `nan` entries in the LD matrix,
+which can be surprising and may numerically impact downstream analyses. A `nan`
+entry occurs if the denominator of a ratio statistic (including {math}`r` and
+{math}`r^2`) is zero, indicating that one or both of the alleles in the pair is
+fixed or absent in the given sample set(s). This can happen for
+a number of reasons:
+
+- Some mutation models allow for reversible mutations, so a back mutation at
+  a site can result in a single allele despite multiple mutations in the
+  history of the sample.
+- LD is computed for a subsample of individuals, and some sites are not
+  variable among the sample nodes in the subsample.
+- A mutation exists above the root of the local tree, so that all samples carry
+  the mutation, and one or more sites are not variable.
+
+The `branch` mode will also return `nan` values for ratio statistics if there
+are branches in either tree on which a mutation would not result in
+a polymorphism within a sample set.
+
+:::{warning}
+This means there are two common situations in which many or all LD values will be `nan`.
+These are:
+
+1. A branch-mode ratio statistic computed on less than the full set of samples
+    will always be `nan`, since part of the trees are ancestral to none of the samples.
+2. A site-mode ratio statistic will be `nan` at any sites at which there are alleles found
+    in the entire set of samples that are not seen in the provided sample set.
+
+This behavior **may change in the future**,
+because possibly more natural behavior not currently implemented
+would be to ignore the branches/alleles not ancestral
+to any of the provided samples.
 :::
+
+(sec_stats_two_locus_sample_one_way_stats)=
+
+#### One-way Statistics
+
+One-way statistics are summaries of two loci in a single sample set, using
+a triple of haplotype counts {math}`\{n_{AB}, n_{Ab}, n_{aB}\}` and the size of
+the sample set {math}`n`, where the capitalized and lowercase letters in our
+notation represent alternate alleles.
+
+(sec_stats_two_locus_sample_two_way_stats)=
+
+#### Two-way Statistics
+
+Two-way statistics are summaries of haplotype counts between two sample sets,
+which operate on the three haplotype counts (as in one-way stats, above)
+computed from each sample set, indexed by `(i, j)`. These statistics take on
+a different meaning from their one-way counterparts. For instance `stat="D2"`
+over a pair of sample sets computes {math}`D_i D_j`, which is the product of
+the covariance measure of LD within each sample set and is related to the
+covariance of {math}`D` between sample sets.
+
+Only a subset of our summary functions are two-way statistics (see
+{ref}`sec_two_locus_summary_functions_two_way`). Note that the unbiased two-way
+statistics expect non-overlapping sample sets (see [Ragsdale and Gravel
+(2020)](https://doi.org/10.1093/molbev/msz265)), and we do not make any
+assertions about the sample sets and assume that `i` and `j` represent disjoint
+sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
+
+(sec_stats_two_locus_summary_functions)=
+
+#### Summary Functions
+
+(sec_stats_two_locus_summary_functions_one_way)=
+
+##### One-way
+
+The two-locus summary functions all take haplotype counts and sample set size
+as input. Each of our summary functions has the signature
+{math}`f(n_{AB}, n_{Ab}, n_{aB}, n)`, converting to haplotype frequencies
+{math}`\{p_{AB}, p_{Ab}, p_{aB}\}` by dividing by {math}`n`. Below,
+{math}`n_{ab} = n - n_{AB} - n_{Ab} - n_{aB}`, {math}`n_A = n_{AB} + n_{Ab}`
+and {math}`n_B = n_{AB} + n_{aB}`, with frequencies {math}`p` found by dividing
+by {math}`n`.
+
+Our convention is to use {math}`A,B` to denote derived alleles, and {math}`a,b`
+ancestral alleles (or other alleles, if the site is multi-allelic). For
+polarised statistics, we average statistics over all non-ancestral alleles. For
+unpolarised statistics, the labeling is arbitrary as we average over all
+alleles (derived and ancestral).
+
+`D`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = p_{AB}p_{ab} - p_{Ab}p_{aB} \, (=p_{AB} - p_A p_B)`
+
+  This statistic is polarised, as the unpolarised result, which averages over
+  allele labelings, is zero. Uses the `total` weighting method.
+
+`D_prime`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = \frac{D}{D_{\max}}`,
+
+  where {math}`D_{\max} = \begin{cases}
+            \min\{p_A (1-p_B), p_B (1-p_B)\} & \textrm{if }D>=0 \\
+            \min\{p_A p_B, (1-p_B) (1-p_B)\} & \textrm{if }D<0
+        \end{cases}`
+
+  and {math}`D` is defined above. Polarised, `total` weighted.
+
+`D2`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = D^2`
+
+  and {math}`D` is defined above. Unpolarised, `total` weighted.
+
+`Dz`
+:  {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = D (1 - 2 p_A) (1 - 2 p_B)`,
+
+  where {math}`D` is defined above. Unpolarised, `total` weighted.
+
+`pi2`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = p_A (1-p_A) p_B (1-p_B)`
+
+  Unpolarised, `total` weighted.
+
+`r`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = \frac{D}{\sqrt{p_A (1-p_A) p_B (1-p_B)}}`,
+
+  where {math}`D` is defined above. Polarised, `total` weighted.
+
+`r2`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = \frac{D^{2}}{p_A (1-p_A) p_B (1-p_B))}`,
+
+  where {math}`D` is defined above. Unpolarised, `haplotype` weighted.
+
+Unbiased two-locus statistics from the Hill-Robertson (1968) system are
+computed from haplotype counts. Definitions of these unbiased estimators can
+be found in [Ragsdale and Gravel
+(2020)](https://doi.org/10.1093/molbev/msz265). They require at least 4 samples
+to be valid and are specified as `stat="D2_unbiased"`, `"Dz_unbiased"`, or
+`"pi2_unbiased"`.
+
+(sec_two_locus_summary_functions_two_way)=
+
+(sec_stats_two_locus_summary_functions_two_way)=
+
+##### Two-way
+
+Two-way statistics are indexed by sample sets {math}`i, j` and compute values
+using haplotype counts within pairs of sample sets.
+
+`D2`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = D_i D_j`,
+
+  where {math}`D_i` denotes {math}`D` computed within sample set {math}`i`,
+  and {math}`D` is defined above. Unpolarised, `total` weighted.
+
+`r2`
+: {math}`f(n_{AB}, n_{Ab}, n_{aB}, n) = r_i r_j`,
+
+  where {math}`r_i` denotes {math}`r` computed within sample set {math}`i`,
+  and {math}`r` is defined above. Unpolarised, `haplotype` weighted.
+
+And `D2_unbiased`, which can be found in [Ragsdale and Gravel
+(2020)](https://doi.org/10.1093/molbev/msz265).
 
 
 (sec_stats_notes)=
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -10930,12 +10930,106 @@ class TreeSequence:
     def ld_matrix(
         self,
         sample_sets=None,
-        sites=None,
-        positions=None,
         mode="site",
         stat="r2",
+        sites=None,
+        positions=None,
         indexes=None,
     ):
+        r"""
+
+        Returns a matrix of the specified two-locus statistic (default
+        :math:`r^2`) computed from sample allelic states or branch lengths.
+        The resulting linkage disequilibrium (LD) matrix represents either the
+        two-locus statistic as computed between all pairs of specified
+        ``sites`` (``"site"`` mode, producing a
+        ``len(sites)``-by-``len(sites)`` sized matrix), or as computed from the
+        branch structures at marginal trees between pairs of trees at all
+        specified ``positions`` (``"branch"`` mode, producing a
+        ``len(positions)``-by-``len(positions)`` sized matrix).
+
+        The sites considered for ``"site"`` mode defaults to all sites (which may
+        result in a very large matrix!), but can be restricted using
+        the ``sites`` argument. Sites must be passed as a list of lists,
+        specifying the ``[row_sites, col_sites]``, resulting in a
+        rectangular matrix, or by specifying a single list of ``[sites]``, in
+        which a square matrix will be produced (see
+        :ref:`sec_stats_two_locus_site` for examples). Here, ``sites``,
+        ``row_sites``, and ``col_sites`` are each lists of site indexes.
+
+        Similarly, in the ``"branch"`` mode, the ``positions`` argument specifies
+        genomic coordinates at which the expectation for the two-locus statistic
+        is computed, given the local tree structure.
+        (See :ref:`sec_stats_two_locus_branch` for explanation of in what sense
+        this is an expectation.) This defaults to computing
+        the LD for each pair of distinct trees (this is equivalent to passing in
+        the leftmost coordinates of each tree's span, since intervals are closed on
+        the left and open on the right). Similar to the site mode, a nested list
+        of row and column positions can be specified separately (resulting in a
+        rectangular matrix) or a single list of a specified positions results
+        in a square matrix (see :ref:`sec_stats_two_locus_branch` for
+        examples). Like ``sites``, the ``positions`` must be specified as a list
+        of lists.
+
+        Some LD statistics are defined for both within a single set of samples
+        and for two sample sets. If the ``indexes`` argument is specified, then
+        ``indexes`` specifies the indexes of the sample sets in the
+        ``sample_sets`` list between which to compute LD. For instance, this
+        results in a 3D array whose ``[k,:,:]``-th slice contains LD values
+        between ``sample_sets[i]`` and ``sample_sets[j]``, where ``(i, j)`` is
+        the ``k``-th element of ``indexes``.
+
+        For more on how the ``indexes`` and ``sample_sets`` interact with the
+        output dimensions, see the :ref:`sec_stats_two_locus_sample_sets`
+        section. Statistics are defined in the
+        :ref:`sec_stats_two_locus_summary_functions_two_way` section.
+
+        **Available Stats** (use ``Stat Name`` in the ``stat`` keyword
+        argument). Statistics marked as "multi sample set" allow
+        (but do not require) computation from two sample sets
+        via the ``indexes`` argument.
+
+        ======================= ========== ================ ==============
+        Stat                     Polarised Multi Sample Set Stat Name
+        ======================= ========== ================ ==============
+        :math:`r^2`              n          y               "r2"
+        :math:`r`                y          n               "r"
+        :math:`D^2`              n          y               "D2"
+        :math:`D`                y          n               "D"
+        :math:`D'`               y          n               "D_prime"
+        :math:`D_z`              n          n               "Dz"
+        :math:`\pi_2`            n          n               "pi2"
+        :math:`\widehat{D^2}`    n          y               "D2_unbiased"
+        :math:`\widehat{D_z}`    n          n               "Dz_unbiased"
+        :math:`\widehat{\pi_2}`  n          n               "pi2_unbiased"
+        ======================= ========== ================ ==============
+
+        :param list sample_sets: A list, or a list of lists of sample node IDs,
+            specifying the groups of nodes to compute the statistic with. Defaults
+            to all samples.
+        :param str mode: A string giving the "type" of the statistic to be
+            computed. Defaults to "site", can be "site" or "branch".
+        :param str stat: A string giving the selected two-locus statistic to
+            compute. Defaults to "r2".
+        :param list sites: A list of lists of sites over which to compute an
+            LD matrix. Can be specified as a list of lists to control the row
+            and column sites. Only available in "site" mode. Specify as
+            ``[row_sites, col_sites]`` or ``[all_sites]``.
+            Defaults to all sites.
+        :param list positions: A list of lists of genomic positions where
+            expected LD is computed based on tree topologies and branch
+            lengths. Only applicable in "branch" mode. Specify as a list of
+            two lists to control the row and column positions, as
+            ``[row_positions, col_positions]``, or ``[all_positions]``.
+            Defaults to the leftmost coordinates of all trees and computes
+            LD between all pairs of trees.
+        :param list indexes: A list of 2-tuples or a single 2-tuple, specifying
+            the indexes of two sample sets over which to compute a two-way LD
+            statistic. Only :math:`r^2`, :math:`D^2`, and :math:`\widehat{D^2}`
+            are implemented for two-way statistics.
+        :return: A 2D or 3D array of LD matrices.
+        :rtype: numpy.ndarray
+        """
         one_way_stats = {
             "D": self._ll_tree_sequence.D_matrix,
             "D2": self._ll_tree_sequence.D2_matrix,
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -7,6 +7,12 @@ In development
 - Add ``json+struct`` metadata codec that allows storing binary data using a struct
   schema alongside JSON metadata. (:user:`benjeffery`, :pr:`3306`)
 
+**Features**
+
+- Add ``TreeSequence.ld_matrix`` stats method and documentation, for computing
+  two-locus statistics in site and branch mode.
+  (:user:`lkirk`, :user:`apragsdale`, :pr:`3416`) 
+
 --------------------
 [1.0.2] - 2026-03-06
 --------------------


docs/stats.md
```diff
@@ -684,8 +684,253 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+The {meth}`~TreeSequence.ld_matrix` method provides an interface to
+a collection of two-locus statistics with predefined summary functions (see
+{ref}`sec_stats_two_locus_summary_functions`) and `site` and `branch`
+{ref}`modes <sec_stats_mode>`. The LD matrix method differs from other
+statistics methods in that it provides a unified API with an argument to
+specify different two-locus summaries of the data. It otherwise behaves
+similarly to most other functions with respect to `sample_sets` and `indexes`.
+
+Two-locus statistics can be computed using two modes, either `sites` or
```
```suggestion
Two-locus statistics can be computed using two modes, either `site` or
```

docs/stats.md
```diff
@@ -684,8 +684,253 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+`branch`, and these should be interpreted in the same way as these modes in the
+single-site statistics. Site statistics allow for multi-allelic data, while
+branch statistics assume an infinite sites model. Within this framework, we
```
jeromekelleher: Hmm - that's a bit of an over simplification regarding infinite sites I think. I think we can just delete this sentence without loss of information at this point?

docs/stats.md
```diff
@@ -684,8 +684,253 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+branch lengths of the trees at a pair of positions. To perform this
+computation, we consider the counts of all possible two-locus haplotypes that
+may be generated by possible mutations on each pair of branches between the two
+trees, which is summarized using an LD statistic. We then sum each pairwise
+comparison weighted by the product of the two branch lengths above the nodes on
+each tree.
+
+Similar to the single-site statistics computed in `branch` mode, this results
+in a statistic that is proportional to the expected statistic under an infinite
+sites model (which mutation rate 1), conditioned on the pair of trees.
```
```suggestion
sites model (with mutation rate 1), conditioned on the pair of trees.
```

docs/stats.md
```diff
@@ -684,8 +684,253 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D_i * D_j`
+
+Where {math}`D` is defined above.
+
+`r2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{(p_{AB_i} - (p_{A_i}  p_{B_i})) (p_{AB_j} - (p_{A_j}  p_{B_j}))}{\sqrt{p_{A_i} (1 - p_{A_i}) p_{B_i} (1 - p_{B_i})}\sqrt{p_{A_j} (1 - p_{A_j}) p_{B_j} (1 - p_{B_j})}}`
+
+And `D2_unbiased`, which can be found in [Ragsdale and Gravel
+(2020)](https://doi.org/10.1093/molbev/msz265).
+
```
jeromekelleher, line 1075: I think we can delete the "todo" section at the bottom here, it'll just confuse people now.

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+        In the site mode, the sites under consideration can be restricted using
+        the ``sites`` argument. Sites can be passed as a list of lists,
+        specifying the ``[[row_sites], [col_sites]]``, resulting in a
+        rectangular matrix, or by specifying a single list of ``[sites]``, in
+        which a square matrix will be produced (see
+        :ref:`sec_stats_two_locus_site` for examples).
+
+        Similarly, in the branch mode, the ``positions`` argument specifies
+        loci for which the expectation for the two-locus statistic is computed
+        over pairs of trees at those positions. LD statis are computed between
```
jeromekelleher:
```suggestion
        over pairs of trees at those positions. LD stats are computed between
```

docs/stats.md
apragsdale: Agreed - deleted.

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+but do not expose it to the user, opting to provide statistics that are
+polarised where appropriate.
+
+(sec_stats_two_locus_site)=
+
+#### Site
+
+The `site` mode computes two-locus statistics summarized over alleles between
+all pairs of specified sites. The default behavior, leaving `sites`
+unspecified, this method will compute a matrix for all pairs of sites, with
```
petrelharp:
```suggestion
unspecified, will compute a matrix for all pairs of sites, with
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+rows and columns representing each site in the tree sequence (i.e., an n×n
```
petrelharp:
```suggestion
one row and column for each site in the tree sequence (i.e., an n x n
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+```{code-cell} ipython3
+ld = ts.ld_matrix(sites=[[0, 1], [1, 2, 3]])
+print(ld)
+```
+
+Because we implement two-locus statistics for multi-allelic data, we require
+a method for combining the statistical results from each pair of alleles into
+one summary for a pair of sites. There two methods for combining results from
+multiple alleles, `hap_weighted` and `total_weighted`, which are
+statistic-specific and not chosen by the user:
```
petrelharp: missing description of hap_weighted and total_weighted here?

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+(sec_stats_two_locus_branch)=
+
+#### Branch
+
+The `branch` mode computes two-locus statistics between pairs of trees. The
+resulting statistic is a summary depending on marginal tree topologies and the
+branch lengths of the trees at a pair of positions. To perform this
+computation, we consider the counts of all possible two-locus haplotypes that
+may be generated by possible mutations on each pair of branches between the two
```
petrelharp:
```suggestion
could be generated by mutations on each pair of branches between the two
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+trees, which is summarized using an LD statistic. We then sum each pairwise
+comparison weighted by the product of the two branch lengths above the nodes on
+each tree.
+
+Similar to the single-site statistics computed in `branch` mode, this results
+in a statistic that is proportional to the expected statistic under an infinite
+sites model (with mutation rate 1), conditioned on the pair of trees.
+
+The time complexity of this method is quadratic, due to the pairwise
+comparisons of branches from a pair of trees. By default, this method computes
```
petrelharp:
```suggestion
The time complexity of this method is quadratic in the number of samples,
due to the pairwise comparisons of branches from each pair of trees.
By default, this method computes
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+branch lengths of the trees at a pair of positions. To perform this
+computation, we consider the counts of all possible two-locus haplotypes that
+may be generated by possible mutations on each pair of branches between the two
+trees, which is summarized using an LD statistic. We then sum each pairwise
+comparison weighted by the product of the two branch lengths above the nodes on
+each tree.
+
+Similar to the single-site statistics computed in `branch` mode, this results
+in a statistic that is proportional to the expected statistic under an infinite
+sites model (with mutation rate 1), conditioned on the pair of trees.
```
petrelharp: This isn't quite right. At first I was going to just suggest the following:
```
in a statistic that is proportional to the expected statistic under an infinite
sites model. More precisely, suppose we place mutations on each tree as
a Poisson process of rate 1 per unit of branch length, compute
the statistic for each pair of mutations (one on each tree), and add these
up; the `branch` mode is the expected value of the resulting sum.
(Note that unlike the single-site
statistics, these mutations are placed on the trees at a given specific
pair of positions, and so the span of the trees do not affect the result.)
```

However, quick experimenting with `ts.ld_matrix()` to check if this was right pointed out that values can easily be `nan` and so an expected value across mutations is not well-defined. I think we described what this is in the original issue on this?

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+comparisons of branches from a pair of trees. By default, this method computes
+a symmetric matrix for all pairs of trees, with rows and columns representing
+each tree in the tree sequence. Similar to the site method, we can restrict the
+output to a subset of trees, either by specifying a vector of positions or
+a pair of vectors for row and column positions separately. To select a specific
+tree, the specified positions must land in the tree span (`[start, end)`).
+
+In the following, we compute a matrix of expected {math}`r^2` within and
+between the first 4 trees in the tree sequence. The tree breakpoints are
+a convenient way to specify those first four trees tree.
```
petrelharp:
```suggestion
a convenient way to specify those first four trees.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+function call.
+
+To compute a two-way two-locus statistic, the `index` argument must be
+provided. The statistics are selected in the same way (with the `stat`
+argument), but we provide a restricted set of two-way statistics (see
+{ref}`sec_stats_two_locus_summary_functions_two_way`). The dimension-dropping
+rules for the result follow the rest of the tskit stats API in that scalar
+indexes will produce a single two-dimensional matrix, while list of indexes
+will produce a three-dimensional array, with the outer dimension of length
+equal to the length of the list of indexes.
```
petrelharp:
```suggestion
rules for the result follow the rest of the tskit stats API in that a single list
or tuple will produce a single two-dimensional matrix, while list of these
will produce a three-dimensional array, with the outer dimension of length
equal to the length of the list.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = p_{ab} - p_{a}p_{b}`
+
+  This statistic is polarised, as the unpolarised result, which averages over
+  allele labelings, is zero. Uses the `total` normalisation method.
+
+`D_prime`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{D_{\max}}`
+
+  Where {math}```
+        D_{\max} = \begin{cases}
```
petrelharp:
```suggestion
: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{D_{\max}}`,

  where {math}```
        D_{\max} = \begin{cases}
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+            \min\{p_A (1-p_B), p_B (1-p_B)\} & \textrm{if }D>=0 \\
+            \min\{p_A p_B, (1-p_B) (1-p_B)\} & \textrm{otherwise}
+        \end{cases}```
+
+  and {math}`D` is defined above.
+
+`D2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = (p_{ab} - p_{a} p_{b})^2`
+
+  Unpolarised, total weighted.
```
petrelharp:
```suggestion
  Unpolarised, `total` weighted.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+`Dz`
+:  {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D (1 - 2 p_{a})(1 - 2p_{b})`
+
+  Where {math}`D` is defined above. Unpolarised, total weighted.
```
petrelharp:
```suggestion
:  {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D (1 - 2 p_{a})(1 - 2p_{b})`,

  where {math}`D` is defined above. Unpolarised, `total` weighted.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+`D_prime`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{D_{\max}}`
+
+  Where {math}```
+        D_{\max} = \begin{cases}
+            \min\{p_A (1-p_B), p_B (1-p_B)\} & \textrm{if }D>=0 \\
+            \min\{p_A p_B, (1-p_B) (1-p_B)\} & \textrm{otherwise}
+        \end{cases}```
+
+  and {math}`D` is defined above.
```
petrelharp:
```suggestion
  and {math}`D` is defined above. Polarised, `total` weighted.
```

docs/stats.md
petrelharp: (the "where" is part of the same sentence as the equation, not a new sentence, so it's not capitalized)

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+`pi2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = p_{a}p_{b}(1-p_{a})(1-p_{b})`
+
+  Unpolarised, total weighted.
+
+`r`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{\sqrt{p_{a}p_{b}(1-p_{a})(1-p_{b})}}`
+
+  Where {math}`D` is defined above. Polarised, total weighted.
+
```
petrelharp:
```suggestion
: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{\sqrt{p_{a}p_{b}(1-p_{a})(1-p_{b})}}`,

  where {math}`D` is defined above. Polarised, `total` weighted.

```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+`Dz`
+:  {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D (1 - 2 p_{a})(1 - 2p_{b})`
+
+  Where {math}`D` is defined above. Unpolarised, total weighted.
+
+`pi2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = p_{a}p_{b}(1-p_{a})(1-p_{b})`
+
+  Unpolarised, total weighted.
```
petrelharp:
```suggestion
  Unpolarised, `total` weighted.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+`r`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D}{\sqrt{p_{a}p_{b}(1-p_{a})(1-p_{b})}}`
+
+  Where {math}`D` is defined above. Polarised, total weighted.
+
+`r2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D^{2}}{p_{a}p_{b}(1-p_{a})(1-p_{b})}`
+
+  Where {math}`D` is defined above. Unpolarised, haplotype weighted.
```
petrelharp:
```suggestion
: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = \frac{D^{2}}{p_{a}p_{b}(1-p_{a})(1-p_{b})}`,

  where {math}`D` is defined above. Unpolarised, `haplotype` weighted.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+Unbiased two-locus statistics from the Hill-Robertson (1968) system are
+computed from haplotype counts. Derivations for these unbiased estimators can
+be found in [Ragsdale and Gravel
+(2020)](https://doi.org/10.1093/molbev/msz265). They require at least 4 samples
```
petrelharp:
```suggestion
computed from haplotype counts. Definitions of these unbiased estimators can
be found in [Ragsdale and Gravel
(2020)](https://doi.org/10.1093/molbev/msz265). They require at least 4 samples
```
Also derivations, but it's more relevant that the defintiions are there?

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+to be valid and are specified as `stat=D2_unbiased`, `Dz_unbiased`, or
+`pi2_unbiased`.
```
petrelharp:
```suggestion
to be valid and are specified as `stat="D2_unbiased"`, `"Dz_unbiased"`, or
`"pi2_unbiased"`.
```

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+##### Two-way
+
+Two-way statistics are indexed by sample sets {math}`i, j` and compute values
+using haplotype counts within pairs of sample sets.
+
+`D2`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D_i * D_j`
+
+Where {math}`D` is defined above.
```
petrelharp:
```suggestion
: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = D_i * D_j`,

where {math}`D` is defined above.
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+            the groups of nodes to compute the statistic with. Defaults to all
+            samples grouped by population.
+        :param str mode: A string giving the "type" of the statistic to be
+            computed. Defaults to "site", can be "site" or "branch".
+        :param str stat: A string giving the selected two-locus statistic to
+            compute. Defaults to "r2".
+        :param list sites: A list of sites over which to compute LD. Can be
+            specified as a list of lists to control the row and column sites.
+            Only applicable in site mode. Specify as
+            ``[[row_sites], [col_sites]]`` or ``[all_sites]``.
```
```suggestion
            ``[[row_sites], [col_sites]]`` or ``[all_sites]``.
            Defaults to all sites.
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+        :param list positions: A list of genomic positions where expected LD is
+            computed. Only applicable in branch mode. Can be specified as a list
+            of lists to control the row and column positions. Specify as
+            ``[[row_positions], [col_positions]]`` or ``[all_positions]``.
```
```suggestion
            ``[[row_positions], [col_positions]]`` or ``[all_positions]``.
            Defaults to the leftmost coordinates of all trees.
```

petrelharp: (Is that right?)

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+        specifying the ``[[row_sites], [col_sites]]``, resulting in a
+        rectangular matrix, or by specifying a single list of ``[sites]``, in
+        which a square matrix will be produced (see
+        :ref:`sec_stats_two_locus_site` for examples).
+
+        Similarly, in the branch mode, the ``positions`` argument specifies
+        loci for which the expectation for the two-locus statistic is computed
+        over pairs of trees at those positions. LD stats are computed between
+        trees whose ``[start, end)`` contains the given position (such that
+        repeats of trees are possible). Similar to the site mode, a nested list
```
```suggestion
        Similarly, in the branch mode, the ``positions`` argument specifies
        genomic coordinates at which the expectation for the two-locus statistic
        is computed, given the local tree structure. This defaults to computing
        the LD for each pair of distinct trees, which is equivalent to passing in
        the leftmost coordinates of each tree's span (since intervals are closed on
        the left and open on the right). Similar to the site mode, a nested list
```

petrelharp: We'll need to adjust this given the "expectation" discussion above.

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+        of row and column positions can be specified separately (resulting in a
+        rectangular matrix) or a single list of a specified positions results
+        in a square matrix (see :ref:`sec_stats_two_locus_branch` for
+        examples).
+
+        Some LD statistics are defined for two sample sets instead of within a
+        single set of samples. If the ``indexes`` argument is specified, at
+        least two sample sets must also be specified. ``indexes`` specifies the
+        sample set indexes between which to compute LD.
```
```suggestion
        indexes of the sample sets in the ``sample_sets`` list
        between which to compute LD.
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,89 @@ def impute_unknown_mutations_time(
...
+        ======================= ========== ================ ==============
+        Stat                     Polarised Multi Sample Set Stat Name
+        ======================= ========== ================ ==============
+        :math:`r^2`              n          y               "r2"
+        :math:`r`                y          n               "r"
+        :math:`D^2`              n          y               "D2"
+        :math:`D`                y          n               "D"
+        :math:`D'`               y          n               "D_prime"
+        :math:`D_z`              n          n               "Dz"
+        :math:`\pi2`             n          n               "pi2"
```
```suggestion
        :math:`\pi_2`             n          n               "pi2"
```

petrelharp (commented):
See suggested edits (and feel free to say I'm wrong on any).

petrelharp:
This is *great*, thanks! Remaining things:

- This should explain more precisely how branch mode is computed and what its interpretation is as an expected value.
- How's it work with multiallelic sites? What exactly are the weighting schemes?
- How's polarization work?

Background:

- weighting: see #2816 
- I still can't find the original issue where we laid out the conceptual framework for this - do you know where that is, @apragsdale?

petrelharp:
Okay, just checking the details of branch mode:

- the result [is calculated in](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L3332) `compute_two_tree_branch_stat` without further weighting
- and then the work is done by `compute_two_tree_branch_state_update`, which [weights f(wAb,waB,wAB) by the product of the two branches in question](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L3135)

Okay, so: for a given pair of branches (on the same or different trees), the sample counts `wAb, waB, wAB` are the same for any pair of mutation falling on those branches; so the expected contribution to average LD determined by the summary funciton `f` per unit sequence length of this pair of branches is `f(w_{Ab}, w_{aB}, w_{AB}, n)` multiplied by the lengths of the two branches.

So, for computation, I think we should say very clearly somewhere that branch mode works by summing over all pairs of branches the value of f( ) multiplied by the lengths of the two branches.

docs/stats.md
petrelharp: See more discussion on this below.

petrelharp:
As for interpretation (ie in what sense is it the expected value):

First off, I'm confused about a dumb silly thing. If I simulate a short tree sequence and compute `ld_matrix` (with r2 and site mode) I get a bunch of nans. Why? (This is confusing because [the denominator for r2](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L4372) is `p_A * p_B * (1 - p_A) * (1 - p_B)`; and msprime only generates mutations with 0 < p < 1.) 

The `nan` thing makes me think I'm missing some basic and important thing. But if my understanding above on how it is computed is correct here are some equivalent options for what branch mode means:

1. Put down two Poisson processes of mutations on each tree, and compute the stat between each pair of mutations; add those up across pairs of mutations; then branch mode is the expected value of this sum.
2. Pick two mutations uniformly, one on each tree; compute the stat; then branch mode is the expected stat multiplied by the total branch lengths of each tree.
3. Compute the average stat across all mutations using a large number of infinite-sites mutations at rate mu per bp. The result (for large mu) is mu multiplied by the sum across pairs of trees of the branch stat for those trees multiplied by the spans of the two trees.

petrelharp:
For the multiallelic question, the computations happen [here](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L2464) and [here](https://github.com/tskit-dev/tskit/blob/main/c/tskit/trees.c#L2411); based on [where they're called from](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L2731) I'm assuming that the first function is just a quicker special-purpose version of the second one for both biallelic sites.

(ps I'm assuming you know this all, but I forget everything and it's nice to go look at the details)

So, it is summing [over all pairs of alleles](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L2433) but then [normalised](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L2454).

docs/stats.md
```diff
@@ -684,14 +684,251 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+The two-locus summary functions all take haplotype counts and sample set size as
+input. Each of our summary functions has the signature
+{math}`f(w_{Ab}, w_{aB}, w_{AB}, n)`, converting to haplotype frequencies
+{math}`\{p_{Ab}, p_{aB}, p_{AB}\}` where appropriate by dividing by {math}`n`.
+
+`D`
+: {math}`f(w_{Ab}, w_{aB}, w_{AB}, n) = p_{ab} - p_{a}p_{b}`
+
+  This statistic is polarised, as the unpolarised result, which averages over
+  allele labelings, is zero. Uses the `total` normalisation method.
```
petrelharp: below this is called "total weighted" not "normalised"; just checking?

petrelharp:
I think we just need a few more paragraphs, documenting exactly what's being calculated, as above. Can you take a stab at this @apragsdale?

apragsdale:
> I think we just need a few more paragraphs, documenting exactly what's being calculated, as above. Can you take a stab at this @apragsdale?

Hi @petrelharp - yes, of course! I'll tackle these today and this weekend.

Thank you for such detailed comments. With some earlier rounds of revision on these docs, I had gone back and forth with including too much vs too little detail, and probably ended up falling on the too-little side of it.

I also agree the nan calculations are confusing, and I remember @lkirk and I having discussions about that last year. I think we had a good explanation/work-around for it. I'll dig up those notes, because it will be important to document as well. I'm also returning to this from many months away from the code, so it will be good to remind myself of all of it as well, and make sure it's documented properly.

petrelharp:
I hope it's easy to ressurrect some of those earlier versions, then?

> I also agree the nan calculations are confusing,

I'm not sure if they are? I imagine it's just "0/0 = nan", but then I'm confused why that happens?

docs/stats.md
apragsdale: I had earlier removed a long description of these to streamline the docs, but agree that I should add a brief explanation of each back in.

apragsdale:
> First off, I'm confused about a dumb silly thing. If I simulate a short tree sequence and compute `ld_matrix` (with r2 and site mode) I get a bunch of nans. Why? (This is confusing because [the denominator for r2](https://github.com/tskit-dev/tskit/blob/17bacdaa6b240b69fb15434a0470f17848f861b0/c/tskit/trees.c#L4372) is `p_A * p_B * (1 - p_A) * (1 - p_B)`; and msprime only generates mutations with 0 < p < 1.)

> I'm not sure if they are? I imagine it's just "0/0 = nan", but then I'm confused why that happens?

Peter- does this occur in site mode, as you say above? We shouldn't get nans if msprime is only generating mutations with 0 < p < 1, and if we compute r^2 over all samples. However, if we specify a subsample as our sample set, the method computes LD at all sites still, and for some of those sites a mutation may not be found in the subsamples. Then p could equal 0 or 1. In this case, you'd get nans for any pair of mutations including one or two of such sites.

But I want to make sure that we are thinking about the same scenario that generates a nan.

```
In [23]: ts = msprime.sim_ancestry(10, population_size=1e4, recombination_rate=1e-8, sequence_length=1e4, random_seed=1)

In [24]: ts = msprime.sim_mutations(ts, rate=1e-8, random_seed=1)

In [28]: ts.ld_matrix()
Out[28]: 
array([[1.        , 0.00277008, 0.00277008, 0.00277008],
       [0.00277008, 1.        , 0.00277008, 0.00277008],
       [0.00277008, 0.00277008, 1.        , 0.00277008],
       [0.00277008, 0.00277008, 0.00277008, 1.        ]])

In [36]: ts.ld_matrix(sample_sets=range(12))
Out[36]: 
array([[1.        ,        nan,        nan, 0.00826446],
       [       nan,        nan,        nan,        nan],
       [       nan,        nan,        nan,        nan],
       [0.00826446,        nan,        nan, 1.        ]])

```

Edit: In any case, this behavior should be documented.

petrelharp:
Here's what I did:
```
>>> ts = msprime.sim_ancestry(3, sequence_length=10, recombination_rate=0.1, random_seed=123)
>>> mts = msprime.sim_mutations(ts, rate=0.1, random_seed=456)
>>> mts.ld_matrix()
array([[1.   , 0.44 ,   nan, 0.44 , 0.44 , 0.125, 0.44 ],
       [0.44 , 1.   ,   nan, 1.   , 1.   , 0.1  , 1.   ],
       [  nan,   nan,   nan,   nan,   nan,   nan,   nan],
       [0.44 , 1.   ,   nan, 1.   , 1.   , 0.1  , 1.   ],
       [0.44 , 1.   ,   nan, 1.   , 1.   , 0.1  , 1.   ],
       [0.125, 0.1  ,   nan, 0.1  , 0.1  , 1.   , 0.1  ],
       [0.44 , 1.   ,   nan, 1.   , 1.   , 0.1  , 1.   ]])
>>> mts.num_sites
7
>>> mts.genotype_matrix()
array([[1, 0, 0, 0, 2, 0],
       [0, 1, 1, 1, 1, 1],
       [0, 0, 0, 0, 0, 0],
       [0, 1, 1, 1, 1, 1],
       [0, 1, 1, 1, 1, 1],
       [0, 0, 1, 0, 1, 0],
       [1, 0, 0, 0, 0, 0]], dtype=int32)
>>> mts.genotype_matrix().sum(axis=1)
array([3, 5, 0, 5, 5, 2, 1])
```

Oh! I see: "msprime does not simulate sites where mutations that aren't polymorphic" but it *does* simulate situations where backmutations can create monomorphic sites despite there being mutations:
```
>>> mts.site(2)
Site(id=2, position=3.0, ancestral_state='A', mutations=[
 Mutation(id=3, site=2, node=0, derived_state='T', parent=-1, metadata=b'', time=1.549923340261233, edge=21, inherited_state='A'), 
 Mutation(id=4, site=2, node=0, derived_state='A', parent=3, metadata=b'', time=1.0122678066613313, edge=21, inherited_state='T')], metadata=b'')
```

Okay, that's all cleared up. I ran into this so quick because I had an unrealistic mutation rate. However, this would be a good thing to mention in the section on multiple alleles.

apragsdale:
Ah, yes - that makes sense. I think I'll add a note about that as well then. Do you think the example that I've added to the docs about subsampling and nan values is overkill, or helpful?

petrelharp:
> Ah, yes - that makes sense. I think I'll add a note about that as well then. Do you think the example that I've added to the docs about subsampling and nan values is overkill, or helpful?

You mean [this bit](https://github.com/tskit-dev/tskit/pull/3416/changes#diff-b01b1c97dca975c3d5b4b674ae04dbc2c1c56305aeaa5c4ca17621d12aaa49b2R854)? I don't think this level of detail is overkill - any careful detail we add now will save us hours in the future tracking down some confused user's queries - but I do have some comments on that section:

- Perhaps separate this out into a sub(sub)^n-section on missing values? It could go at the end? That would help with concerns about this being too much detail.
- I am hesitant to suggest "just simplify to make your problems go away" since you introduce a BUNCH of annoying issues by simplifying: for instance, the columns of the result correspond to the sites in the *simplified* ts not the original, so if you're not careful matching them up to where on the genome you're looking at is annoying. I would tend to just say "hey this is one reason you get nans" and not recommend the simplification fix.
- This is a different reason for getting nans than what I encountered; probably the reason I got nans should also be mentioned? Like "you'll get nans if any sites are monomorphic in the sample. here are two reasons you'd get this from msprime" (but recall that monomorphic sites are totally normal in real data).

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        :math:`r^2`) computed from sample allelic states or branch lengths.
+        The resulting linkage disequilibrium (LD) matrix represents either the
+        two-locus statistic as computed between all pairs of specified
+        ``sites`` ("site" mode, producing a num_sites-by-num_sites sized
+        matrix), or as computed from the branch structures at marginal trees
+        between pairs of trees at all specified ``positions`` ("branch" mode,
+        producing a num_positions-by-num_positions sized matrix).
+
+        In the site mode, the sites under consideration can be restricted using
+        the ``sites`` argument. Sites can be passed as a list of lists,
```
```suggestion
        The sites considered for "site" mode defaults to all sites (which may
        result in a very large matrix!), but can be restricted using
        the ``sites`` argument. Sites can be passed as a list of lists,
```
Saying "can be restricted" raises the question "restricted from what?"

python/tskit/trees.py
petrelharp: I still like this suggestion and don't think it needs adjusting?

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        of row and column positions can be specified separately (resulting in a
+        rectangular matrix) or a single list of a specified positions results
+        in a square matrix (see :ref:`sec_stats_two_locus_branch` for
+        examples).
+
+        Some LD statistics are defined for two sample sets instead of within a
+        single set of samples. If the ``indexes`` argument is specified, at
+        least two sample sets must also be specified. ``indexes`` specifies the
+        indexes of the sample sets in the ``sample_sets`` list
+        between which to compute LD.
```
petrelharp: Let's add something like "For instance, <example> results in an array of shape (a,b,c) whose [k,:,:]-th slice contains LD values between ``sample_sets[i]`` and ``sample_sets[j]``." (I'm not making a suggestion since I haven't worked out what exactly this should say.)

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        ======================= ========== ================ ==============
+        Stat                     Polarised Multi Sample Set Stat Name
+        ======================= ========== ================ ==============
+        :math:`r^2`              n          y               "r2"
+        :math:`r`                y          n               "r"
+        :math:`D^2`              n          y               "D2"
+        :math:`D`                y          n               "D"
+        :math:`D'`               y          n               "D_prime"
+        :math:`D_z`              n          n               "Dz"
+        :math:`\pi_2`             n          n               "pi2"
```
```suggestion
        :math:`\pi_2`            n          n               "pi2"
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+
+        Some LD statistics are defined for two sample sets instead of within a
+        single set of samples. If the ``indexes`` argument is specified, at
+        least two sample sets must also be specified. ``indexes`` specifies the
+        indexes of the sample sets in the ``sample_sets`` list
+        between which to compute LD.
+
+        For more on how the ``indexes`` and ``sample_sets`` interact with the
+        output dimensions, see the :ref:`sec_stats_two_locus_sample_sets`
+        section.
```
petrelharp: Let's refer to the documentation,  and maybe also the paper where these are defined?

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        :math:`D`                y          n               "D"
+        :math:`D'`               y          n               "D_prime"
+        :math:`D_z`              n          n               "Dz"
+        :math:`\pi_2`             n          n               "pi2"
+        :math:`\widehat{D^2}`    n          y               "D2_unbiased"
+        :math:`\widehat{D_z}`    n          n               "Dz_unbiased"
+        :math:`\widehat{\pi_2}`  n          n               "pi2_unbiased"
+        ======================= ========== ================ ==============
+
+        :param list sample_sets: A list of lists of sample Node IDs, specifying
```
```suggestion
        :param list sample_sets: A list, or a list of lists of sample node IDs, specifying
```
(note that just applying this suggestion will make linting complain about line lengths)

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        least two sample sets must also be specified. ``indexes`` specifies the
+        indexes of the sample sets in the ``sample_sets`` list
+        between which to compute LD.
+
+        For more on how the ``indexes`` and ``sample_sets`` interact with the
+        output dimensions, see the :ref:`sec_stats_two_locus_sample_sets`
+        section.
+
+        **Available Stats** (use ``Stat Name`` in the ``stat`` keyword
+        argument).
```
```suggestion
        argument). Statistics marked as "multi sample set" allow
        (but do not require) computation from two sample sets
        via the ``indexes`` argument. 
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        :math:`D_z`              n          n               "Dz"
+        :math:`\pi_2`             n          n               "pi2"
+        :math:`\widehat{D^2}`    n          y               "D2_unbiased"
+        :math:`\widehat{D_z}`    n          n               "Dz_unbiased"
+        :math:`\widehat{\pi_2}`  n          n               "pi2_unbiased"
+        ======================= ========== ================ ==============
+
+        :param list sample_sets: A list of lists of sample Node IDs, specifying
+            the groups of nodes to compute the statistic with. Defaults to all
+            samples grouped by population.
```
```suggestion
            the groups of nodes to compute the statistic with. Defaults to all
            samples.
```
I don't think this [groups by population](https://github.com/tskit-dev/tskit/pull/3426/changes#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8R8285). (And, I wouldn't recommend making it do that.)

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+similarly to most other functions with respect to `sample_sets` and `indexes`.
+
+Two-locus statistics can be computed using two modes, either `site` or
+`branch`, and these should be interpreted in the same way as these modes in the
+single-site statistics. Within this framework, statistics may be either
+polarised or unpolarised. For statistics that are polarised, we compute
+statistic values for pairs of derived alleles. Unpolarised statistics compute
+statistics over all possible alleles, derived and ancestral, and the result in
+averaged over statistics computed for all pairs of alleles (see weighting
+schemes below). The option for polarisation is not exposed to the user.
```
petrelharp:
```suggestion
statistic values for pairs of derived alleles. Unpolarised statistics compute
statistics over all pairs of alleles, derived and ancestral. In either case,
the result is averaged over these values, using a weighting
scheme descrbed below. The option for polarisation is not exposed to the user.
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+computes a subset of the matrix shown above.
+
+```{code-cell} ipython3
+ld = ts.ld_matrix(sites=[[0, 1], [1, 2, 3]])
+print(ld)
+```
+
+Because we allow for two-locus statistics to be computed for multi-allelic
+data, we need to be able to combine statistical results from each pair of
+alleles into one summary for a pair of sites. We use two implementations for
```
petrelharp:
```suggestion
alleles into one summary for a pair of sites. This does not affect biallelic
data (and so this section can be skipped on first reading).
We use two implementations for
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+combining results from multiple alleles: `hap_weighted` and `total_weighted`.
+These are statistic-specific and not chosen by the user.
+
+Briefly, consider a pair of sites with {math}`n` alleles at the first locus and
+{math}`m` alleles at the second. Write {math}`f_{i,j}` as the statistic
```
petrelharp:
```suggestion
{math}`m` alleles at the second. (Whether this includes the ancestral allele
depends on whether the statistic is polarised.) Write {math}`f_{i,j}` as the statistic
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+computed for focal alleles {math}`A_i` and {math}`B_j`, with haplotype weights
+{math}`(A_i B_j, A_i b_j, a_i B_j)`, where {math}`a_i` and {math}`b_j` are the
+collection of alleles that are not the focal alleles {math}`A_i` or
+{math}`B_j`, respectively. Then the weighting schemes are defined as:
```
petrelharp:
```suggestion
computed for focal alleles {math}`A_i` and {math}`B_j`.
Then the weighting schemes are defined as:
```
This bit didn't make sense, and I think wasn't used anyhow.

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+  [Zhao (2007)](https://doi.org/10.1017/S0016672307008634).
+
+- `total_weighted`: {math}`\frac{1}{n m}\sum_{i=1}^{n}\sum_{j=1}^{m}f_{ij}`.
+  This method assigns equal weight to each of the possible pairs of focal
+  alleles at the two sites, taking the arithmetic mean of statistics over
+  focal haplotypes.
+
+Out of all of the available summary functions, only {math}`r^2` uses
+`hap_weighted` normalisation, with the remainder using uniform weighting
+(`total_weighted`).
```
petrelharp, line 771: Can we say "see Ragsdale (XXXX) for explanation of this choice" or something?

docs/stats.md
petrelharp: I'm wondering if this note about polarisation should be down below with the "weighting" schemes. Since isn't polarisation just chosen so that "people get what they expect"? So perhaps bundle those two bits into a "Computational details" section?

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+(sec_stats_two_locus_branch)=
+
+#### Branch
+
+The `branch` mode computes expected two-locus statistics between pairs of
+trees, conditioned on the marginal topologies and branch lengths of those
+trees. The trees for which we compute statistics are specified by positions,
+and for a pair of positions we consider all possible haplotypes that could be
+generated by a single mutation occurring at the two trees.
```
petrelharp:
```suggestion
generated by a single mutation occurring on each of the two trees.
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+For two trees, one with {math}`n` branches and the other with {math}`m`
+branches, there are {math}`nm` possible pairs of branches that may carry the
+pair of mutations. For each pair, we compute the two-locus statistic, and then
+sum these values weighted by the product of the two branch lengths. Given the
+two mutations occur, this accounts for the relative probability that the two
+mutations fall on any pair of branches.
+
```
petrelharp, line 799: What do you think about inserting one of the [precise interpretations](https://github.com/tskit-dev/tskit/pull/3416#issuecomment-4060713700) of what's computed? This is still very vague ("accounts for the relative probability").

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+print(ld)
+```
+
+(sec_stats_two_locus_sample_sets)=
+
+#### Sample Sets
+
+The two-locus API provides a mechanism by which to subset the samples under
+consideration, providing the ability to compute a separate LD matrix for each
+sample set or an LD matrix between sample sets. Output dimensions are handled
```
petrelharp: This isn't quite right, since separate LD matrices are produced for each sample set by default even for non-two-way statistics.

petrelharp:
```suggestion
consideration, providing the ability to compute a LD matrix between distinct sample sets.
Output dimensions are handled
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+in the same manner as the rest of the stats API (see
+{ref}`sec_stats_output_dimensions`). The two-locus API differs from the rest of
+the stats API in that we handle one-way and two-way statistics in the same
+function call.
+
+To compute a two-way two-locus statistic, the `index` argument must be
```
petrelharp:
```suggestion
To compute a two-way two-locus statistic, the `indexes` argument must be
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+provided. The statistics are selected in the same way (with the `stat`
+argument), but we provide a restricted set of two-way statistics (see
+{ref}`sec_stats_two_locus_summary_functions_two_way`). The dimension-dropping
+rules for the result follow the rest of the tskit stats API in that a single list
+or tuple will produce a single two-dimensional matrix, while list of these
+will produce a three-dimensional array, with the outer dimension of length
```
petrelharp:
```suggestion
will produce a three-dimensional array, with the first dimension of length
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+```{code-cell} ipython3
+breakpoints = ts.breakpoints(as_array=True)
+ld = ts.ld_matrix(mode="branch", positions=[breakpoints[[0]], breakpoints[0:4]])
+print(ld)
+```
+
+(sec_stats_two_locus_sample_sets)=
+
+#### Sample Sets
+
```
petrelharp, line 867: I think we need to first insert a paragraph explaining what happens without `indexes`: i.e., a list of samples gets you a matrix; and a list-of-lists of samples gets you a 3D array. Then say "okay and for a two-way thing you need a list-of-list of indexes, ...".

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+ts.ld_matrix(sample_sets=[0, 1, 2, 3]) -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3]]) -> 3 dimensions
+# two-way
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=(0, 1)) -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=[(0, 1)]) -> 3 dimensions
+```
+
+#### Why are there nan values in the LD matrix?
+
+For some statistics, it is possible to observe nan entries in the LD matrix,
```
petrelharp:
```suggestion
For some statistics, it is possible to observe `nan` entries in the LD matrix,
```

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+which can be surprising or numerically impact downstream analyses. A nan entry
```
petrelharp:
```suggestion
which can be surprising or numerically impact downstream analyses. A `nan` entry
```

docs/stats.md
petrelharp: Another important thing to mention: which branches? For instance, if there are branches above the root, or not ancestral to any samples, are those included? What about branches ancestral to samples but not to any of the samples in `sample_sets`? I assumed it would be "only branches on which a mutation would produce a polymorphic site", but I'm noticing that I can get `nan`s from branch stats, so maybe not?
```
mts.ld_matrix(sample_sets, stat='r2', indexes=(0,1), mode='branch')
array([[nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan]])
```

python/tskit/trees.py
```diff
@@ -10930,12 +10930,92 @@ def impute_unknown_mutations_time(
...
+        repeats of trees are possible). Similar to the site mode, a nested list
+        of row and column positions can be specified separately (resulting in a
+        rectangular matrix) or a single list of a specified positions results
+        in a square matrix (see :ref:`sec_stats_two_locus_branch` for
+        examples).
+
+        Some LD statistics are defined for two sample sets instead of within a
+        single set of samples. If the ``indexes`` argument is specified, at
+        least two sample sets must also be specified. ``indexes`` specifies the
+        indexes of the sample sets in the ``sample_sets`` list
```
```suggestion
        Some LD statistics are defined for two sample sets as well as within a
        single set of samples. If the ``indexes`` argument is specified, then
        these specify the indexes of the sample sets in the ``sample_sets`` list
```
Empirically this is not true: this works (at least using code in #3426): `ts.ld_matrix([[0,1]], stat='r2', indexes=(0,0))`

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+monomorphic. This can happen for a number of reasons:
+
+- The mutation models allows for reversible mutations, so a back mutation at
+  a site has resulted in a single allele despite multiple mutations in the
+  history of the sample.
+- LD is computed for a subsample of individuals, and some sites are not
+  variable among the sample nodes in the subsample.
+- A mutation exists above the root of the local tree, so that all samples carry
+  the mutation, and one or more sites are not variable.
+
```
petrelharp, line 934: perhaps address also `nan`s in branch stats here

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+(sec_stats_two_locus_sample_two_way_stats)=
+
+##### Two-way Statistics
+
+Two-way statistics are summaries of haplotype counts between two sample sets,
+which operate on the three haplotype counts (as in one-way stats, above)
+computed from each sample set, indexed by `(i, j)`. These statistics take on
+a different meaning from their one-way counterparts. For instance {math}`D_i
+D_j` can be thought of as the covariance between two loci when computed for
```
petrelharp: Isn't this called [`D^2`](https://github.com/tskit-dev/tskit/pull/3416/changes#diff-b01b1c97dca975c3d5b4b674ae04dbc2c1c56305aeaa5c4ca17621d12aaa49b2R981) (also see [here](https://github.com/tskit-dev/tskit/pull/3416/changes#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8R10985)), not `D_i D_j`?

docs/stats.md
```diff
@@ -684,14 +684,310 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+(sec_stats_two_locus_summary_functions)=
+
+#### Summary Functions
+
+(sec_stats_two_locus_summary_functions_one_way)=
+
+##### One-way
+
+The two-locus summary functions all take haplotype counts and sample set size as
+input. Each of our summary functions has the signature
```
petrelharp: All of these functions are in terms of `p`, not `w`, right? So why introduce `w` at all here?

petrelharp:
Okay! I had another careful pass through - apologies for the great many comments, but I think this stuff is neat and I want to make sure we write down exactly what's being computed. (And, there's a bunch of nitpicks.)

apragsdale:
Thanks @petrelharp - I _think_ I've addressed all/most of your comments here!

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+Because we allow for two-locus statistics to be computed for multi-allelic
+data, we need to be able to combine statistical results from each pair of
+alleles into one summary for a pair of sites. This does not affect biallelic
+data (and so this section can be skipped on first reading).
+We use two implementations for
+combining results from multiple alleles: `hap_weighted` and `total_weighted`.
+These are statistic-specific and not chosen by the user, with choices motivated
+by [Zhao (2007)](https://doi.org/10.1017/S0016672307008634) with the goal
+of return statistics as expected by the user.
```
petrelharp:
```suggestion
by [Zhao (2007)](https://doi.org/10.1017/S0016672307008634).
```
At least this should say "returning" instead of "return", but the whole phrase is kinda mysterious, so I vote to delete.

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+  focal haplotypes.
+
+Out of all of the available summary functions, only {math}`r^2` uses
+`hap_weighted` normalisation, with the remainder using uniform weighting
+(`total_weighted`).
+
+Within this framework, statistics may be either
+polarised or unpolarised. For statistics that are polarised, we compute
+statistic values for pairs of derived alleles. Unpolarised statistics compute
+statistics over all pairs of alleles, derived and ancestral. In either case,
```
petrelharp:
```suggestion
statistic values for pairs of derived alleles. (For this purpose, the "derived" alleles
at a site are all alleles except that stored as the ``ancestral_state`` for the site.)
Unpolarised statistics compute
statistics over all pairs of alleles, derived and ancestral. In either case,
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+the result is averaged over these values, using a weighting
+scheme described below. The option for polarisation is not exposed to the user.
+Instead, we implement statistics that are polarised where appropriate.
```
petrelharp:
```suggestion
scheme described below. The option for polarisation is not exposed to the user,
and list which statistics are polarised below.
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+pair of mutations. For each pair, we compute the two-locus statistic, and then
+sum these values weighted by the product of the two branch lengths. Given the
+two mutations occur, this accounts for the relative probability that the two
+mutations fall on any pair of branches.
+
+In other words, imagine we place two mutations uniformly, one on each tree, and
+then compute the statistic; the branch mode computes the expected value of the
+statistic over this process, multiplied by the product of the total branch
+lengths of each tree. In some cases, this process will compute statistics over
+branches that do not result in a polymorphism in the sample, in which case
```
petrelharp:
```suggestion
lengths of each tree. This weighting accounts for mutational opportunity, so that
the sum of the branch-mode statistic over all positions in a genomic region,
multiplied by a mutation rate,  is equal to the expected sum of the two-locus site
statistic over all mutations falling in that region under an infinite-sites model.

If some branches are ancestral to none or all of the samples, then a mutation on
these branches does not result in a polymorphism in the sample, and so the
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+ratio statistics ({math}`r`, {math}`r^2`)  will return `nan` values. See
+further discussion of `nan` values below.
```
petrelharp: Actually, I think this bit on `nan` should just be moved to the section below.

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+    recombination_rate=2e-8,
+    random_seed=12)
+ts = msprime.sim_mutations(ts, rate=2e-8, random_seed=12)
+
+ld = ts.ld_matrix(mode=“site”, sample_sets=range(8))
+print(ld)
+```
+
+For concreteness, we would expect the following dimensions with the specified
+`sample_sets` and `indexes` arguments.
```
petrelharp:
```suggestion
We would get the following dimensions with the specified
`sample_sets` and `indexes` arguments.
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+```
+# one-way
+ts.ld_matrix(sample_sets=None) -> 2 dimensions
+ts.ld_matrix(sample_sets=[0, 1, 2, 3]) -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3]]) -> 3 dimensions
+# two-way
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=(0, 1)) -> 2 dimensions
+ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=[(0, 1)]) -> 3 dimensions
```
petrelharp:
```suggestion
ts.ld_matrix(sample_sets=None) # -> 2 dimensions
ts.ld_matrix(sample_sets=[0, 1, 2, 3]) # -> 2 dimensions
ts.ld_matrix(sample_sets=[[0, 1, 2, 3]]) # -> 3 dimensions
# two-way
ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=(0, 1)) # -> 2 dimensions
ts.ld_matrix(sample_sets=[[0, 1, 2, 3], [4, 5, 6, 7]], indexes=[(0, 1)]) # -> 3 dimensions
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+```
+
+#### Why are there `nan` values in the LD matrix?
+
+For some statistics, it is possible to observe `nan` entries in the LD matrix,
+which can be surprising or numerically impact downstream analyses. A `nan` entry
+may occur when computing a ratio statistic (such as {math}`r` or {math}`r^2`)
+with a denominator of zero, indicating that one or both sites in the pair are
+monomorphic. This can happen for a number of reasons:
```
petrelharp:
```suggestion
occurs if the denominator of a ratio statistic (such as {math}`r` or {math}`r^2`)
is zero, indicating that one or both of the alleles in the pair is fixed or absent in the
sample set under consideration. This can happen for a number of reasons:
```

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+  a site has resulted in a single allele despite multiple mutations in the
+  history of the sample.
+- LD is computed for a subsample of individuals, and some sites are not
+  variable among the sample nodes in the subsample.
+- A mutation exists above the root of the local tree, so that all samples carry
+  the mutation, and one or more sites are not variable.
+
+The `branch` mode will also return `nan` values if there are branches in either
+tree on which a mutation would not result in a polymorphism within a sample
+set.
```
petrelharp: I think we would like some more discussion of this, but we can push tihs to a different PR: see #3441.

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+notation represent alternate alleles.
+
+(sec_stats_two_locus_sample_two_way_stats)=
+
+##### Two-way Statistics
+
+Two-way statistics are summaries of haplotype counts between two sample sets,
+which operate on the three haplotype counts (as in one-way stats, above)
+computed from each sample set, indexed by `(i, j)`. These statistics take on
+a different meaning from their one-way counterparts. For instance {math}`D2`
```
petrelharp:
```suggestion
a different meaning from their one-way counterparts. For instance `stat="D2"`
```
(right?)

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+#### Summary Functions
+
+(sec_stats_two_locus_summary_functions_one_way)=
+
+##### One-way
+
+The two-locus summary functions all take haplotype counts and sample set size
+as input. Each of our summary functions has the signature
+{math}`f(n_{Ab}, n_{aB}, n_{AB}, n)`, converting to haplotype frequencies
+{math}`\{p_{Ab}, p_{aB}, p_{AB}\}` where appropriate by dividing by {math}`n`.
```
petrelharp: Which alleles are `A` and which are `a`? My guess is that for polarised statistics, `A` is non-ancestral, while for unpolarised it doesn't matter because we sum over all *unordered* pairs (but we should say so).

docs/stats.md
```diff
@@ -684,14 +684,335 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+
+`D_prime`
+: {math}`f(n_{Ab}, n_{aB}, n_{AB}, n) = \frac{D}{D_{\max}}`,
+
+  where {math}
+            \min\{p_A (1-p_B), p_B (1-p_B)\} & \textrm{if }D>=0 \\
+            \min\{p_A p_B, (1-p_B) (1-p_B)\} & \textrm{otherwise}
+        \end{cases}```
+
+  and {math}`D` is defined above. Polarised, `total` weighted.
```
petrelharp, line 1018: what is Dmax?

petrelharp: Oh, I see - this bit got messed up somehow, I think it's missing the `Dmax = \begin{cases} ...`

docs/stats.md
petrelharp: hm, looking at how the stats mostly use `pa` not `pA` I wonder if `a` is derived?

petrelharp:
```suggestion
{math}`\{p_{Ab}, p_{aB}, p_{AB}\}` by dividing by {math}`n`.
```

petrelharp: > we sum over all unordered pairs 

I mean *average* over all unordered pairs, using the weighting scheme.

python/tskit/trees.py
```diff
@@ -10930,12 +10930,99 @@ def impute_unknown_mutations_time(
...
+        result in a very large matrix!), but can be restricted using
+        the ``sites`` argument. Sites can be passed as a list of lists,
+        specifying the ``[[row_sites], [col_sites]]``, resulting in a
+        rectangular matrix, or by specifying a single list of ``[sites]``, in
+        which a square matrix will be produced (see
+        :ref:`sec_stats_two_locus_site` for examples).
+
+        Similarly, in the branch mode, the ``positions`` argument specifies
+        genomic coordinates at which the expectation for the two-locus statistic
+        is computed, given the local tree structure. This defaults to computing
```
```suggestion
        is computed, given the local tree structure.
        (See :ref:`sec_stats_two_locus_branch` for explanation of in what sense
        this is an expectation.) This defaults to computing
```

docs/stats.md
apragsdale: This information was also repeated in that section below, so I've moved/condensed it.

docs/stats.md
apragsdale: I found and fixed additional issues with this section of the docs, especially the `pa` vs `pA` mixup. I also added a note about the labels `A,B` and `a,b`.

apragsdale:
A close reading also found a few inconsistencies in the docstring. I hope the small fixes I made improve some of the clarity.

I've also rebased here. Let me know if I should also go ahead and squash, if that's needed.

docs/stats.md
```diff
@@ -684,14 +684,368 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+column for each site in the tree sequence (i.e., an {math}`n \times n` matrix
+where {math}`n` is the number of sites in the tree sequence). We can also
+restrict the output to a subset of sites, either by specifying a single vector
+of site indexes for both rows and columns or a pair of vectors for the row
+sites and column sites separately.
+
+The following computes a matrix of the {math}`r^2` measure of linkage
+disequilibrium (LD) computed pairwise between the first 4 sites in the tree
+sequence among all samples. The `sites` must be given as a list of lists, and
+with a single list of sites specified, we obtain a symmetric square matrix.
```
petrelharp, line 725:
```suggestion
sequence among all samples. The `sites` must be given as either a list
or a list of lists of sample IDs.
With a single list of sites specified, we obtain a symmetric square matrix. 
```

docs/stats.md
```diff
@@ -684,14 +684,368 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+sequence among all samples. The `sites` must be given as a list of lists, and
+with a single list of sites specified, we obtain a symmetric square matrix. 
+
+```{code-cell} ipython3
+ld = ts.ld_matrix(sites=[[0, 1, 2, 3]])
+print(ld)
+```
+
+The following demonstrates how we can specify the row and column sites
+independently of each other. We're specifying 2 rows and 3 columns, which
```
petrelharp:
```suggestion
If a list of two lists of sample IDs is provided, these specify the row and column sites.
For instance, here we're specifying 2 rows and 3 columns, which
```

docs/stats.md
```diff
@@ -684,14 +684,368 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+alleles into one summary for a pair of sites. This does not affect biallelic
+data (and so this section can be skipped on first reading).
+We use two implementations for
+combining results from multiple alleles: `hap_weighted` and `total_weighted`.
+These are statistic-specific and not chosen by the user, with choices motivated
+by [Zhao (2007)](https://doi.org/10.1017/S0016672307008634).
+
+Briefly, consider a pair of sites with {math}`n` alleles at the first locus and
+{math}`m` alleles at the second. (Whether this includes the ancestral allele
+depends on whether the statistic is polarised.) Write {math}`f_{i,j}` as the
```
petrelharp:
```suggestion
depends on whether the statistic is polarised.) Write {math}`f_{ij}` as the
```

docs/stats.md
```diff
@@ -684,14 +684,368 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+compute the expected {math}`r^2` value for a pair of mutations that each land
+uniformly on the pair of trees, we can divide by the product of the total
+branch lengths:
+
+```{code-cell} ipython3
+total_branch_lengths = [tree.total_branch_length for tree in ts.trees()]
+prod_branch_lengths = np.outer(total_branch_lengths, total_branch_lengths)
+print(ld / prod_branch_lengths[0:4, 0:4])
+```
+
```
petrelharp, line 840: I think this is right, and I think could help make things concrete?
```suggestion

To compute the average {math}`r^2` for a uniformly chosen pair of mutations, we also
weight by tree span:
```{code-cell}
tree_spans = np.array([t.span for t in ts.trees()])
total_opportunity = np.sum(tree_spans * total_branch_lengths)
all_ld = ts.ld_matrix(mode="branch")
mean_ld = np.sum( all_ld * np.outer(tree_spans, tree_spans)) / total_opportunity**2
print("mean infinite-sites LD:", mean_ld)
```

```

File: docs/stats.md [hunk 71]
```diff
@@ -684,14 +684,368 @@ and boolean expressions (e.g., {math}`(x > 0)`) are interpreted as 0/1.
...
+(sec_stats_two_locus_sample_one_way_stats)=
+
+#### One-way Statistics
+
+One-way statistics are summaries of two loci in a single sample set, using
+a triple of haplotype counts {math}`\{n_{AB}, n_{Ab}, n_{aB}\}` and the size of
+the sample set {math}`n`, where the capitalized and lowercase letters in our
+notation represent alternate alleles.
+
```
petrelharp:
```suggestion
::::{warning}
This means there are two common situations in which many or all LD values will be nan.
These are:

1. A branch-mode ratio statistic computed on less than the full set of samples
    will always be nan, since part of the trees are ancestral to none of the samples.
2. A site-mode ratio statistic will be nan at any sites at which there are alleles found
    in the entire set of samples that are not seen in the provided sample set.

This behavior **may change in the future**,
because possibly more natural behavior not currently implemented
would be to ignore the branches/alleles not ancestral
to any of the provided samples.
```

docs/stats.md
apragsdale: With a unnested list for `sites`, such as `ts.ld_matrix(sites=[0, 1, 2, 3])`, I get `ValueError: Sites must be a list of lists, tuples, or ndarrays`

docs/stats.md
apragsdale: I think you mean if a "list of two lists of site indexes is provided," but I see how this writing can be improved.

docs/stats.md
petrelharp: Whoops, my mistake! Sorry.

docs/stats.md
petrelharp: You're right; as it's written is actually better. Never mind.

PR #3426, opened 2026-03-10
Two locus general matrix

lkirk:
## Description

This is the last of the required components for the LD matrix methods. I wanted feedback on the API before I add documentation, but this method is complete and tested. The final things to do are to leak check the cpython code and add some documentation.

This feature enables a user to implement their own two-locus count statistic in python, similar to `ts.sample_count_stat`. User functions take two arguments, the first is a matrix of haplotype counts and the second is a vector of sample set sizes. For instance, this is how we would implement `D` with this api:

```python
def D(X, n):
    pAB, pAb, paB = X / n
    pA = pAb + pAB
    pB = paB + pAB
    return pAB - (pA * pB)
```

Since this API supports multiallelic sites, the user can also pass a normalisation function to control how the data is normalised across multiple alleles. The normalisation function is only run when computing over multiallelic sites. I've set the default to be $1/(n_A n_B)$, which is simply the arithmetic mean of the alleles in a given pair of sites. This will suffice in the majority of cases (the only outlier is $r^2$, for which there is already a python API). We also support computing statistics between sample sets.

The user would use the above summary function like this:

```
ts.two_locus_count_stat(ts.samples(), D, 1)
```

Where `1` specifies the length of the output array, we always require 1 dimension -- same as the `ts.sample_count_stat` function.

# PR Checklist:


python/_tskitmodule.c
```diff
@@ -7946,6 +7946,290 @@ parse_positions(TreeSequence *self, PyObject *positions, npy_intp *out_dim)
     return array;
 }
 
+typedef struct {
+    PyArrayObject *sample_set_sizes;
+    PyObject *summary_func;
+    PyObject *norm_func;
+} two_locus_general_stat_params;
+
```
lkirk, line 7956: I see that the CPython code coverage is split from the rest of the python tests. I can't test this directly without a small multiallelic test case in `test_python_c.py`.

c/tskit/trees.c
```diff
@@ -3455,13 +3451,30 @@ tsk_treeseq_two_locus_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sampl
             goto out;
         }
         ret = tsk_treeseq_two_branch_count_stat(self, state_dim, num_sample_sets,
-            sample_set_sizes, sample_sets, result_dim, f, &f_params, norm_f, out_rows,
+            sample_set_sizes, sample_sets, result_dim, f, f_params, norm_f, out_rows,
             row_positions, out_cols, col_positions, options, result);
     }
 out:
     return ret;
 }
```
lkirk, line 3468: This function now serves as an inner wrapper. The the general stat accepts the summary function params so that the CPython code can pass them directly. All of the specialized stats functions call this function.

apragsdale:
Thank you for opening this, @lkirk. I'm excited to see this implemented! Do we need to do any testing to demonstrate that there are no memory leaks or anything like that, which wouldn't be included in the test suite?

I know it could be found in the tests, but it may be helpful to spell out here how the API would work for a two- or more-way stat? Would it be:
```
ts.two_locus_count_stat([sample_list_1, sample_list_2], two_way_stat_func, 2)
```

lkirk:
Yes, this needs leak checking and documentation, which I can add. I mostly wanted to make sure the user interface made sense first.

> I know it could be found in the tests, but it may be helpful to spell out here how the API would work for a two- or more-way stat? Would it be:
> ```
> ts.two_locus_count_stat([sample_list_1, sample_list_2], two_way_stat_func, 2)
> ```

The `result_dim` argument to the function gives the dimensions of the summary function output. The output is required to be 1D, so `result_dim` really tells us the length of the returned vector. In most cases, it'll be 1. See [this](https://github.com/tskit-dev/tskit/pull/3426/changes#diff-e19c98d45ae6a04849bb68ff2db40970977ac1b13c07ed4b44cc884beed69075R2645) line in the tests. So, if your summary function returned 1 value for a pair of sites, then the function call will look like this:

```
ts.two_locus_count_stat([sample_list_1, sample_list_2], two_way_stat_func, 1)
```

`two_locus_count_stat` was designed to work like [`sample_count_stat`](https://tskit.dev/tskit/docs/stable/python-api.html#tskit.TreeSequence.sample_count_stat)

petrelharp:
Hi, @lkirk! This looks pretty straightforward. To have a careful opinion about the API, I think I need to see a reasonably careful docstring? For instance, what exactly are the arguments to `f` and `norm_f`? I can probably figure it out by tracing through the code, but it'll be less error-prone if you write it down. I'll have a look through for other issues, but it will be much easier to have the description in words of what exactly it's trying to do.

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2398,256 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+            axis=0,
+        )
+
+
+@pytest.mark.parametrize(
+    "ts,stat",
+    [
+        (
+            ts := [
+                p for p in get_example_tree_sequences() if p.id == "n=100_m=32_rho=0.5"
```
petrelharp: Does this runs the risk of not being run at all if there is no such example tree sequence?

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2398,256 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+        "D_prime",
+        "Dz",
+        "pi2",
+        "D2_unbiased",
+        "Dz_unbiased",
+        "pi2_unbiased",
+    ],
+)
+def test_general_one_way_two_locus_stat_multiallelic(stat):
+    (ts,) = {t.id: t for t in get_example_tree_sequences()}["all_fields"].values
```
petrelharp: insert some things like `assert ts.num_sites > 0` and probably something asserting that this one is actually multiallelic here

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2398,256 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+        (ts, "r"),
+        (ts, "D_prime"),
+        (ts, "Dz"),
+        (ts, "pi2"),
+        (ts, "D2_unbiased"),
+        (ts, "Dz_unbiased"),
+        (ts, "pi2_unbiased"),
+    ],
+)
+def test_general_two_locus_site_stat(ts, stat):
```
petrelharp: I think insert some `assert`s here that `ts` has the required properties for being a good test.

lkirk:
Hi @petrelharp, thanks for taking a look. I didn't offer much of a description before, does this help?

## Summary Function
In the `sample_count_stat` api, there are 3 required parameters:
1. `sample_sets`: List of lists of node ids.
1. `f`: A summary function that takes a one-dimensional array of length equal to the number of sample sets and returns a one-dimensional array.
1. `output_dim`: The length of the summary function's return value.

The `two_locus_count_stat` function (see function signature below) takes the same basic inputs except that `f` takes a matrix of haplotype counts and a vector of sample set sizes as input.

I've been writing summary functions with the signature `f(X, n)`

Where X is a matrix whose rows correspond to sample sets and columns correspond to haplotype counts ($w_{Ab}, w_{aB}, w_{AB}$). Here's a sample of what that looks like:
```
        sample_set 1, sample_set 2
w_Ab  [[9.            9.]
w_aB   [0.            0.]
w_AB   [0.            0.]]
```

Why lay the data out this way? Because numpy is row-major, iteration over the arrays gives us rows. That makes it easy to select the haplotype counts in one line:
```
AB, Ab, aB = X
```
Note: under the hood, the data is still laid out in an optimal way (at least as far as I can tell -- see this [note](https://github.com/lkirk/tskit/pull/16/files#diff-76eaf4cb701579960d1e1f888222aa140f64f1b4622c28d6907d9b78c6c63871R7903-R7907))

In addition, the sample set sizes is shaped so that we can use it to normalize the counts in one go (`n` is a vector of sample set sizes)
```
pAB, pAb, paB = X / n
```

Finally, since we're no longer controlling the summary functions ourselves, the user has control over polarisation (see `polarised` parameter).

## Normalisation
Perhaps the most clunky thing about this api is that the user will need a normalisation function for multiallelic data (norm_f parameter). Its function signature is the same as the code internal to `ld_matrix()`: `f(X, n, nA, nB) -> [out_dims]` where `X` and `n` are the same, `nA` and `nB` are the number of `A` and `B` alleles, respectively. Its output dimensions should be the same as our summary function (we also validate this just like we validate the output dims of our summary function).

The default normalization function of `two_locus_count_stat` is the `total` (or uniform) normalization, we average all results into one. Here's how I have defined it in the function signature

```
lambda X, n, nA, nB: np.expand_dims(1 / (nA * nB), axis=0),
```

I would have loved to be able to pass `np.mean` as a summary function instead, but we do need haplotype counts to norm $r^2$ and this keeps our original ld_matrix implementation untouched (we use the same exact machinery for this method, so we have to follow the C api).

python/_tskitmodule.c
petrelharp: Yep, it looks that way. Can you set one up? You could do that by either having a small tree sequence with high enough mutation rate there are lots of multiple hits (we do use `msprime` in that file) or explicitly constructing the tables.

c/tskit/trees.c
```diff
@@ -3455,13 +3451,30 @@ tsk_treeseq_two_locus_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sampl
             goto out;
         }
         ret = tsk_treeseq_two_branch_count_stat(self, state_dim, num_sample_sets,
-            sample_set_sizes, sample_sets, result_dim, f, &f_params, norm_f, out_rows,
+            sample_set_sizes, sample_sets, result_dim, f, f_params, norm_f, out_rows,
             row_positions, out_cols, col_positions, options, result);
     }
 out:
```
petrelharp, line 3465: I don't know why codecov says this line is not covered; do you? (sticking some prints or asserts in here can help track that down)

Image

petrelharp:
I've looked at the code pretty carefully, and things look good besides the comments. Nice work!

python/tests/test_ld_matrix.py
lkirk: I don't think so, I'm selecting the index `[0]` out of this, which will fail for trying to select the 0th element out of an empty list. ~It might be better to use the `pytest_params=False` now that I look at `get_example_tree_sequences` again. I could avoid the `.value[0]`~ (Just looked closer, I can't do this because I'm selecting these by `id`).

python/tests/test_ld_matrix.py
lkirk: Good call, I will add this. These test tree sequences are shared by many tests and the docstring for this tree doesn't explicitly state that it has multiallelic sites, so it's good to assert in case it changes.

python/_tskitmodule.c
lkirk: I was thinking of something like this:

```python
[ins] In [1]: import msprime

[ins] In [2]: ts = msprime.sim_ancestry(2, recombination_rate=0.1, sequence_length=100, random_seed=23)
         ...: ts = msprime.sim_mutations(ts, rate=0.1, discrete_genome=True, random_seed=23)
         ...: print(f"allele counts: {set(len(s.mutations) for s in ts.sites())}")
allele counts: {1, 2, 3}

[ins] In [3]: ts.num_samples, ts.num_trees, ts.num_sites, ts.num_edges
Out[3]: (4, 47, 46, 137)
```

Which gives us a small multiallelic test case. But, yes I think it's critical to get coverage here, it's a reasonable test to make and it's impacting code coverage.

python/tests/test_ld_matrix.py
lkirk: For this test, I simply chose a tree sequence that was "sufficiently complex". I think by definition this test case should have the required properties: 100 samples, sequence length of 32, recombination rate of 0.5 -- giving 32 trees. (tests are generated [here](https://github.com/tskit-dev/tskit/blob/main/python/tests/tsutil.py#L2479), for the sake of discussion). Mutation rate is hardcoded in that function (0.1).

I'm not opposed to creating some `assert`s here, but it's not clear to me what I'd be guarding against (code changes to the example tree sequences?). Or is the goal of the `assert`s to be more explicit about the assumptions of the test?

c/tskit/trees.c
lkirk: Judging from the coverage of the rest of that function (every goto being hit), my best guess is that the compiler is optimizing that label out. In fact, I could just change all of those `goto ret` to `return ret` and remove the `out:` label since we're not doing any sort of cleanup at the end of this function.

lkirk:
Hi @petrelharp, I appreciate you taking a look. I've responded to everything with how I plan to resolve the comment. I did have one clarifying question [here](https://github.com/tskit-dev/tskit/pull/3426#discussion_r2935585700). I will clean this up tomorrow.

python/_tskitmodule.c
petrelharp: perfect. Note that `discrete_genome=True` by default (but no harm as it makes it obvious).

petrelharp:
Hi, @lkirk - I went ahead and made a few adjustments to how the examples are chosen for testing; if you see something wrong or disagree, feel free to revert. Rationale:

- Instead of listing "don't run these tests since they're slow" it's now "do run these tests" because otherwise if someone adds a new slow example to `get_example_tree_sequence` it won't slow things down mysteriously
- Instead of pulling that one test case out I just call the function that generates it. This makes it a lot more clear what's going on.

python/tests/test_ld_matrix.py
petrelharp: Both of those things? As you say: making it clear what is needed for this to be a good test, checking those are in fact satisfied, and guarding against future changes to the example-generating code.

petrelharp:
Hm, I also changed the tests that pulled out "all_fields" to explicitly call that function. Now I see that [we have a fixture](https://github.com/tskit-dev/tskit/blob/main/python/tests/conftest.py#L119) defined for that one. Do you mind removing that function call and using the fixture instead? So that just means doing
```
def test_general_one_way_two_locus_stat_multiallelic(stat, ts_fixture):
```
and then using `ts` instead of `ts_fixture` in the code (or renaming it).

I'd do this but don't want to risk making conflicting edits.

c/tskit/trees.c
petrelharp: Don't do that; it's nice to have that pattern in the code.

petrelharp: Oh! I see why this is complaining here. It's because right before this line is `if (A) { ... } else if (B) { ... }`, but there are no tests in which both A and B are false. And in fact it's impossible for this to happen. So the "else if ( )" should be an "else"; and if you want to be extra paranoid stick a `tsk_bug_assert(B)` in the code at the top of the `else { }`.

Every time I think "oh no codecov has this one wrong" it turns out that it's got a good point.

jeromekelleher:
I'm happy to look over this at some stage - please ping me when you'd like input!

c/tskit/trees.c
lkirk: I ended up removing the branch altogether and added `tsk_bug_assert(mode_branch)`

python/tskit/trees.py
```diff
@@ -10927,6 +10931,39 @@ def impute_unknown_mutations_time(
...
+        mode="site",
+    ):
+        row_sites, col_sites = self.parse_sites(sites)
+        row_positions, col_positions = self.parse_positions(positions)
+        _, sample_sets, sample_set_sizes = self.__convert_sample_sets(sample_sets)
+        result = self._ll_tree_sequence.two_locus_count_stat(
+            sample_set_sizes,
+            sample_sets,
+            f,
+            norm_f or (lambda X, n, nA, nB: 1 / (nA * nB)[np.newaxis,]),
```
petrelharp: Here let's make this more explicit (and hence easier to read): earlier do
```
if norm_f is None:
     norm_f = lambda X, n, nA, nB: 1 / (nA * nB)[np.newaxis,]
```

petrelharp: Also note that in the description you wrote
```
lambda X, n, nA, nB: np.expand_dims(1 / (nA * nB), axis=0),
```
which looks more idiomatic, but I don't know?

petrelharp:
Ah, just noticed your expanded description. So, to answer your question: the API looks good. I may have some other comments now, but will wait for you to ping me when it's ready to read carefuly (including the docstring).

lkirk:
thanks @petrelharp, glad to know it seems reasonable. I'm currently running down a bug that appears at first glance to be macos specific (it shook out with my updates). I'm hoping to be done with this in the next few hours, I'll summarize the changes and ping you when ready. For now, you might see a few more spurious pushes until I figure out what's going on. I don't have any way to test this locally on a mac.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,39 @@ def impute_unknown_mutations_time(
...
+            result_dim,
+            polarised,
+            row_sites,
+            col_sites,
+            row_positions,
+            col_positions,
+            mode,
+        )
+        if result_dim == 1:  # drop dimension
```
petrelharp: I don't think this dropping dimension depending on the output of `f` is consistent with how `ts.sample_count_stat` works: for instance,
```
>>> import tskit, msprime
>>> ts = msprime.sim_ancestry(2, sequence_length=10, recombination_rate=0.1)
>>> mts = msprime.sim_mutations(ts, rate=0.1)
>>> mts.sample_count_stat([mts.samples()], lambda x: [2, 3], strict=False, output_dim=2)
array([3.4, 5.1])
>>> mts.sample_count_stat([mts.samples()], lambda x: [2, 3], strict=False, output_dim=2, windows=[0, 3, 10])
array([[2.66666667, 4.        ],
       [3.71428571, 5.57142857]])
>>> mts.sample_count_stat([mts.samples()], lambda x: [2], strict=False, output_dim=1, windows=[0, 3, 10])
array([[2.66666667],
       [3.71428571]])
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,39 @@ def impute_unknown_mutations_time(
...
+            return result.reshape(result.shape[:2])
+        # Orient the data so that the first dimension is the sample set so that
+        # we get one LD matrix per sample set.
+        return result.swapaxes(0, 2).swapaxes(1, 2)
```
petrelharp: I'm confused about this: since the summary function takes in all the sample sets at once, it doesn't have any notion of "one LD matrix per sample set"?

python/tskit/trees.py
petrelharp: But correct me if I'm wrong!

python/tskit/trees.py
petrelharp: (and if this is necessary, note that `np.moveaxis` would do this in a single operation)

python/tskit/trees.py
petrelharp: I could be wrong, but it looks to me like to keep consistency, we *would* want
```
ts.two_locus_count_stat([0, 1, 2, 3])
```
to be the same thing as
```
ts.two_locus_count_stat([[0, 1, 2, 3]])  # <- note now a 2d array
```
except that the first case has dimension dropped.

*However* this makes no sense here because the output shape is decoupled from the sample sets, thanks to passing through the summary function. (Right?)

python/tests/test_python_c.py
```diff
@@ -2050,7 +2050,12 @@ def norm_func(X, n, nA, nB):
         a = ts_multi.two_locus_count_stat(
             ss_sizes, ss, stat_func, norm_func, 1, True, None, None, None, None, "site"
         )
-        assert a.shape == (56, 56, 1)  # ts has 56 sites
+        import platform
```
petrelharp: Well, this seems like an annoying thing to track down (sympathies), but I don't think it's the right way forward:  this strategy is setting us up for debugging the same issue if anything in msprime or the underlying rng libraries changes in the future. Instead, how about change these shapes to things like `ts.num_sites`? I see this applies to some [earlier lines as well](https://github.com/tskit-dev/tskit/pull/3426/changes/bd0a1a5683c4d8505a0ec26c8fcbf89da32c3dad#diff-448bade8e1c3d219382da2d803f14f0e61072541e940ee08f10ff486b49c7343R2039). That has the extra benefit that it's more obvious what's being tested as well.

lkirk:
@petrelharp, thank you for taking such a close look. I'll summarize what was done since your last review.

## Testing
The tests have been cleaned up substantially. Your comment about pytest fixtures gave me the idea to put a few tree sequences in their own fixtures, where I assert the properties directly in the fixture itself. `ts_100_samp_with_sites_fixture` has the same properties as the `n=100_m=32_rho=0.5` example tree sequence, but its definition is now localized to `test_ld_matrix.py`, so there's less risk of changes to the example tree sequences affecting the `general_` tests.

I've moved away from hard-coding the number of expected dimensions in the output (python C tests). I wasn't intending on keeping the platform code in the test, but I wanted a single clean commit showing exactly the (bug?) I was running into. My understanding is that most PRNGs are portable across architectures, so it was rather surprising to me that the same exact `msprime` simulation produces different tree sequences (my best guess without digging into msprime is that there may be some floating point jitter in rejection sampling? -- unless my assumptions about PRNGs are incorrect). All this being said, do you think it's appropriate to open an issue in `msprime`?

In `test_python_c.py`, I've split the biallelic and multiallelic test cases to isolate the testing of the normalization function. It's also nice to have *one* tree sequence defined and *one* set of happy-path inputs.

The python C test coverage seems to be at a reasonable level at this point.

## Dimensions
I've re-reworked some of the dimension handling code for normalization. The changes you observed before were an attempt to remove the need for `expand_dims` in a lot of places -- adopting the philosophy of propagating the dimensions through to the end. I did a couple of microbenchmarks (happy to share these if you want) to simply decide the right thing to do here. It turns out that `[value]` is both the fastest and most readable. I think putting vectorized operations and/or fancy indexing operations in the hot loop adds an appreciable amount of overhead. I worked to get these summary functions looking as idiomatic as possible so that they can be used as a reference. There are definitely small changes that can be made to make them faster, but I wanted to balance readability. This balance is perhaps more critical here than in other places because computational cost adds up quickly in the two-locus methods.

## Documentation
I've drafted up a docstring for this method, but I'll want to reference a few sections defined in #3416. I can rebase this branch from that one, just to get the definitions I need. If I do that, we'll want #3416 merged before this PR.

python/tskit/trees.py
lkirk: I've opted for
```
[1 / (nA * nB)] * result_dim
```
I mention in another comment that this is more readable and performant. I had to add the `* result_dim` to match the output dimension in cases where we have multiple sample sets for a one-way statistic.

python/tskit/trees.py
lkirk: For data locality when computing stats over sample sets, we want the sample set to be the last dimension. However, for usability @apragsdale and I thought it'd be better to have the sample set as the first dimension in the output. This makes it possible to access individual LD matrices with `result[0]`, etc... compared to `result[:,:,0]`. So, moving the axis will be necessary. I haven't seen `np.moveaxis` before, the LD matrix also use `swapaxes` twice and would benefit from this change as well.

In summary, we now move `(m, m, k)` to `(k, m, m)` with `np.moveaxis`, where `m` is the number of sites/trees and `k` is the number of ~sample sets~ output dimensions.

python/tests/test_python_c.py
lkirk: I switched to `ts.get_num_sites()` and `ts.get_num_trees()` (the lowlevel tree sequence analogs to `.num_sites` and `.num_trees`).

python/tskit/trees.py
lkirk: Correct. The sample set dimensions will not have an impact on output dimensions. The output dimension is explicitly specified by the user. I removed dimension dropping after looking at the output dimensions of the `sample_count_stat` function.

python/tskit/trees.py
lkirk: But now that I look more closely, say, at diversity

```
[nav] In [6]: ts.diversity(sample_sets=[A, B], windows="sites", mode="site").shape
Out[6]: (73, 2)
```
I realize that this decision was a deviation from other stats, where sample set is the *last* dimension.

lkirk:
@jeromekelleher I think this is ready for a look. I don't anticipate any more changes to the CPython layer, we're still discussing a couple of points, but I think your feedback would be useful at this point.

python/tests/test_python_c.py
```diff
@@ -1987,6 +2004,239 @@ def test_ld_matrix_multipop(self, stat_method_name):
...
+        a = method(ss_sizes, ss, stat_func, norm_func, 1, True, *branch_args)
+        assert a.shape == (ts.get_num_trees(), ts.get_num_trees(), 1)
+        # happy path - sample sets as lists are also valid
+        site_list_args = row_sites_list, col_sites_list, None, None, "site"
+        branch_list_args = None, None, row_pos_list, col_pos_list, "branch"
+        a = method(ss_sizes, ss, stat_func, norm_func, 1, True, *site_list_args)
+        assert a.shape == (ts.get_num_sites(), ts.get_num_sites(), 1)
+        a = method(ss_sizes, ss, stat_func, norm_func, 1, True, *branch_list_args)
+        assert a.shape == (ts.get_num_trees(), ts.get_num_trees(), 1)
+        # happy path - default array filling
```
petrelharp:
```suggestion
        # happy path - default values for site and position lists
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+        norm_f=None,
+        polarised=False,
+        sites=None,
+        positions=None,
+        mode="site",
+    ):
+        """
+        Compute two-locus statistics with a user-defined python function that
+        operates on haplotype counts. TODO: reference modes in two-locus docs.
+        On each pair of sites or trees, the summary function is provided with
```
petrelharp: I think you need to be more specific about the summary function: not just "is provided with" but say what, exactly, is the signature. (e.g., "must accept two arguments: the first, ``X``, is ...")

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+        ``X``, a matrix with shape (3, k) and ``n``, a vector with shape (k,),
+        where k is the number of sample sets provided. ``X`` is a read-only
+        matrix whose rows contain haplotype counts per sample set (counts of AB,
+        Ab, aB) and ``n`` is a vector of sample set sizes.
```
petrelharp: Say what AB, Ab, and aB are? Or perhaps we can refer to #3416 here?

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+
+        .. note::
+            Because we are operating on very small matrices/vectors, vectorised
+            operations are often times slower than operations on scalars. Simply
+            returning ``[value]`` can be faster than returning
+            ``value[np.newaxis,]`` or ``np.expand_dims(value, 0)``.
```
petrelharp: I don't understand the significance of this? Perhaps be a bit more explicit, if it's a big deal, or omit this?

petrelharp: (what I don't get is why you'd be returning these things)

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+
+        What follows is an example of computing ``D`` from a tree sequence. Many
+        more examples can be found in the test suite
+        ``test_ld_matrix.py::GeneralStatsFuncs``. Let's begin with our summary
```
petrelharp: Perhaps link to the `ld_matrix` docs for ``D``? And good idea, but I don't think we want to refer to the test suite in the docstring: if you'd like to leave this pointer in, perhaps put a comment in the function body?

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+        returning a numpy array with length equal to the number of sample sets.
+
+        .. code-block:: python
+            def D(X, n):
+                pAB, pAb, paB = X / n
+                pA = pAb + pAB
+                pB = paB + pAB
+                return pAB - (pA * pB)
+
+        ``norm_f`` is a normalisation function used to combine all computed
```
petrelharp: I think we need to give the precise requirements for the norm function here.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+        statistics for multiallelic allele pairs (TODO: see two-locus
+        docs). Biallelic sites do not require any normalisation (in fact, the
+        normalisation function is never called for biallelic sites). If one of
+        either site A or site B is multiallelic, then the normalisation function
+        will be called. The default normalisation function is identical to
+        ``total_norm`` shown in the example below. ``hap_norm`` is required for
+        normalising :math:`r^2`. Both of these examples return a numpy array
+        with length equal to the number of sample sets (for one-way stats).
```
petrelharp: It's not clear what "for one-way stats" means here.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+
+        :param list sample_sets: A list of lists of Node IDs, specifying the
+            groups of nodes to compute the statistic with.
+        :param f: A function that takes two arguments - a two-dimensional array
+            with shape (3, k) and a one-dimensional array with shape (k, ) where
+            k is the number of sample sets.
+        :param int result_dim: The length of ``f`` and ``norm_f``'s return value.
+        :param norm_f: A function that takes four arguments - the first two are
+            the same as ``f``, the second two are scalars representing the
+            number of A and B alleles, respectively.
```
```suggestion
            number of A and B alleles, respectively. If ``None``, then
            defaults to the "total" normalisation described above.
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,111 @@ def impute_unknown_mutations_time(
...
+            polarised,
+            row_sites,
+            col_sites,
+            row_positions,
+            col_positions,
+            mode,
+        )
+        # Orient the data so that the first dimension is the sample set so that
+        # we get one LD matrix per result dimension
+        return np.moveaxis(result, -1, 0)
```
petrelharp, line 11061: maybe we should do it this way also above at [this line](https://github.com/tskit-dev/tskit/pull/3426/changes#diff-ecbb802115b51a593926a3015f4ff2ec444fa4dc4915af74f4959abc60c84df8R8306)? (one call to `moveaxis` instead of two)

and, the comment here is not right yet: says "first dimension is the sample set"

python/_tskitmodule.c
```diff
@@ -7946,6 +7946,290 @@ parse_positions(TreeSequence *self, PyObject *positions, npy_intp *out_dim)
...
+    if (err == TSK_PYTHON_CALLBACK_ERROR) {
+        goto out;
+    } else if (err != 0) {
+        handle_library_error(err);
+        goto out;
+    }
+    ret = (PyObject *) result_matrix;
+    result_matrix = NULL;
+out:
+    Py_XDECREF(summary_func);
```
petrelharp, line 8215: Hm, you're decref'ing `summary_func` but not `norm_func`; that seems wrong. Have you checked for memory leaks in this code yet? (I need to look up how to do that...)

c/tskit/trees.c
petrelharp: Ah, that works. But - and sorry to be nitpicky (about so many things) but how about this goes inside an `else { }`? I think that's a lot more clear: for instance, the "site mode" and "branch mode" paths look symmetric, then.

c/tskit/trees.c
```diff
@@ -3441,27 +3437,45 @@ tsk_treeseq_two_locus_count_stat(const tsk_treeseq_t *self, tsk_size_t num_sampl
             goto out;
         }
         ret = tsk_treeseq_two_site_count_stat(self, state_dim, num_sample_sets,
-            sample_set_sizes, sample_sets, result_dim, f, &f_params, norm_f, out_rows,
+            sample_set_sizes, sample_sets, result_dim, f, f_params, norm_f, out_rows,
             row_sites, out_cols, col_sites, options, result);
-    } else if (stat_branch) {
```
petrelharp, line 3446: so proposal was to just change this to
```
} else {
  tsk_bug_assert(stat_branch);
```

petrelharp:
Great, I'll have a careful look. Let's see:

- Should there be a test that `X` is actually read-only?
- I think you've not got a test that running on a ts with no sites in site mode gives the correct thing (and doesn't error)?
- I *think* I see why we need both `tsk_treeseq_two_locus_count_stat` and `tsk_treeseq_two_locus_count_general_stat`; I think some explanatory comments about the difference would be very helpful here.

> do you think it's appropriate to open an issue in msprime?

No, this is a known thing. Maybe the default RNG is different on macOS? But, not something we're intending to go after.

> It turns out that [value] is both the fastest and most readable. 

Sorry, I'm being slow. What's the context here? 

I'm not seeing any other potential issues? The python-c is hard stuff to wade through; good job getting that working. I won't pretend I understand the ramifications of all that.

petrelharp:
I think the failing docs build is just because you need a linebreak after `.. code-block:: python`?

python/tskit/trees.py
petrelharp: ah. we definitely want this to match the other stats. I had not done that double-checking, but had assumed you had.

python/tskit/trees.py
petrelharp: hm, #3416 doesn't currently explain what these are

python/_tskitmodule.c
lkirk: I made a pass for memory leaks and found one, I've revised the summary and norm function methods above so that they don't perform intermediate transposes (which were leaking memory -- I believe they're allocating memory for new dimensions, not the underlying data).

I performed some leak checking with/without Py_XDECREF(norm_func) and observed a leak. Fixed in my recent update.

I'll make a comment on the main thread about memory leak checking.

python/tskit/trees.py
lkirk: Perhaps this note is better suited for long-form documentation instead of in the docstring -- I've removed it. I'm mostly trying to guide people to "best practices" that I've observed while testing this code.

There's two concerns wrapped up into one here:

1. Vectorized numpy array operations in the hot loop on very small vectors/matrices are more expensive than performing scalar operations. The examples in `test_ld_matrix.py` mix these a bit for the sake of readability. But, it can be good to know where potential bottlenecks in summary functions will be. Perhaps this is a better fit for the documentation.
2. The summary functions must return a 1D array with length=`result_dim`, these are just different ways of constructing the output array. It turns out that the best way is to simply construct the output array by wrapping it in a python list (i.e. `[value]`), over using `np.expand_dims(value, 0)`.

python/tskit/trees.py
lkirk: I removed the reference to `GeneralStatsFuncs`. I do think people should look at the `GeneralStatsFuncs` class to get an idea for how to implement summary functions. They're tested and implement all of the C summary functions. I suppose I can directly address this in the documentation that will accompany this code.

python/tskit/trees.py
lkirk: Good point, I've written a more general description that relies on the mapping from sample_set->stat->output dim instead of referring to one-way stats, which aren't a good tool for explaining what I mean here.

I've made a change that expresses this more directly I think.

python/tskit/trees.py
lkirk: I typically call these "Haplotype Counts" with `A/B` as the derived and `a/b` as the ancestral. Though the ancestral/derived states don't matter in the cases of biallelic sites and branches (biallelic as implemented -- a single node bifurcates "ancestral" and "derived" or vice versa).

The reason that this distinction doesn't matter too much when we're only considering 2 alleles is that LD statistics are symmetric. Swapping ancestral/derived states doesn't change the outcome (I recall testing this a while ago when I was working on normalisation).

That's all to say that `A/B` and `a/b` really just represent 2 states (or two alleles). When combined with `n` (the sample set size), haplotype counts `AB`, `Ab`, and `aB` comprise all of the information we can gather about the linkage of two sites/trees. I'm not exactly sure how much of this to spell out in #3416 or in this docstring.

c/tskit/trees.c
lkirk: Fixed, mentioned in comment below.

c/tskit/trees.c
lkirk: I agree, this does look better. I've fixed it.

python/tests/test_python_c.py
lkirk: Incorporated into my changes.

python/tskit/trees.py
lkirk: I changed the language to be more precise.

python/tskit/trees.py
lkirk: incorporated this change.

python/tskit/trees.py
lkirk: I clarified the requirements for summary functions and norm functions in my recent change.

lkirk (commented):
@petrelharp Sorry this update took so long. 
## Updates

Overall, I think I've gotten to every issue that was raised except for the `np.swapaxes` call at the end of `two_locus_count_stat` in `trees.py`. I'll make a separate commit for this, I think this will touch many small things.

### Docstring
I made some improvements to the docstring, hopefully this is a bit closer to what you had in mind.

### Leak Checking the Python C Code
In my initial memory leak test, I discovered a memory leak which lead to discovering a small bug in the branch stats (we weren't propagating the return status of the summary function -- it's fixed now).

The memory leak that I found was caused by performing a transpose operation before calling the summary function, that created a new container for the data, leaving a dangling reference to the old one. It looks like we're leak free after fixing this change (I also fixed the branch stat bug). I removed the intermediate transpose operation by wrapping the C data with a "pre-transposed" numpy array. I create a new array with `PyArray_New` so that I can provide the correct dimensions and stride, instead of relying on `PyArray_SimpleNewFromData`, which doesn't properly set the stride if the dimensions are swapped.

I'll include some selected results from leak checking here, but a more thorough write up can be found in this doc [two-locus-general-leak-check.pdf](https://github.com/user-attachments/files/26169423/two-locus-general-leak-check.pdf), along with the script I used [leak-check](https://github.com/user-attachments/files/26169433/leak-check.py).

The following plots show resident set size for the process (`rss`) and the size of the python heap (`heap`) for sampled time points. I'm running `two_locus_count_stat` 10,000 times in a for-loop, recording `heap` and `rss` as time progresses (using `memray`). The document I attached also contains a couple of examples where I induce a memory leak in `two_locus_count_stat` to show that we can indeed detect large and small memory leaks with this method.

#### Site Mode
I checked for memory leaks in the side mode with trees that contained biallelic and multiallelic sites. I also produced an error in the summary function to ensure that we're cleaning up properly on error.

Image

#### Branch Mode
I checked for memory leaks in the branch mode with a much smaller tree, I also included a test case that contains the same summary function error.

Image

python/tskit/trees.py
lkirk: This was a conscious decision with @apragsdale, we thought it was a bit more convenient to access single LD matrices with `result[0]` instead of `result[:,:,0]`, but it sounds like we should change to the latter (last dimension is sample set). I've chatted with @apragsdale about this and he sees the logic in this change. Since we use the C statistics to test the general stats, if I make the change for our general stat, I'll need to do it across the board. I _think_ it'll only affect a few tests that explicitly test output dimensions.

petrelharp: This does sound like a pain, and given you've thought about this hard, I could be convinced otherwise!

petrelharp:
Let me know when you're done here, and I'll take another pass through! Thanks for the careful attention.

petrelharp:
> I think I see why we need both tsk_treeseq_two_locus_count_stat and tsk_treeseq_two_locus_count_general_stat; I think some explanatory comments about the difference would be very helpful here.

I've come back to this question, since it seems confusing to have in the public C API two functions which are almost the same. Would it be reasonable to just have `tsk_treeseq_two_locus_count_general_stat` public and keep `tsk_treeseq_two_locus_count_stat` private? The additional stuff done in `tsk_treeseq_two_locus_count_stat` is very small, and people could go look at it to see what to do in their own code if this is not obvious. What do you think?

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2396,391 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+    @staticmethod
+    def D2_ij(X, n):
+        pAB, pAb, paB = X / n
+        pA = pAb + pAB
+        pB = paB + pAB
+        return np.prod(pAB - (pA * pB), keepdims=True)
+
+    @staticmethod
+    def D2_ij_unbiased(X, n):
+        """The identity of the sample sets is up to the user."""
```
petrelharp: is there a reason for this comment? looks like this is just a hanger-on from wherever this was copied from?

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2396,391 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+            (Ab[0] * aB[0] - AB[0] * ab[0])
+            * (Ab[1] * aB[1] - AB[1] * ab[1])
+            / (n[0] * (n[0] - 1) * n[1] * (n[1] - 1))
+        ]
+
+    @staticmethod
+    def D2_ii_ij_jj_unbiased(X, n):
+        """
+        Multiple stats can be computed from the same data. The identity of the
+        sample sets is up to the user. This function assumes two sample sets.
```
petrelharp: same here: remove this comment?

python/tests/test_ld_matrix.py
```diff
@@ -2398,3 +2396,391 @@ def test_multipopulation_r2_varying_unequal_set_sizes(genotypes, sample_sets, ex
...
+        """
+        is_multiallelic = max({len(s.mutations) for s in ts.sites()}) > 1
+        match (stat, mode, is_multiallelic):
+            case ("r2", "site", True):
+                return cls.hap_norm
+            case ("r2_ij", "site", True):
+                return cls.k_way_hap_norm
+            case (_, "branch", _):  # branch stats do not need a norm func
+                return cls.assert_no_norm_func
+            case (_, _, False):  # biallelic sites do not need a norm func
```
petrelharp:
```suggestion
            case (_, _, False):  # biallelic sites should not use the norm func
```
I think this is actually the point here

lkirk:
> > I think I see why we need both tsk_treeseq_two_locus_count_stat and tsk_treeseq_two_locus_count_general_stat; I think some explanatory comments about the difference would be very helpful here.
> 
> I've come back to this question, since it seems confusing to have in the public C API two functions which are almost the same. Would it be reasonable to just have `tsk_treeseq_two_locus_count_general_stat` public and keep `tsk_treeseq_two_locus_count_stat` private? The additional stuff done in `tsk_treeseq_two_locus_count_stat` is very small, and people could go look at it to see what to do in their own code if this is not obvious. What do you think?

I think I see what you mean now. I didn't quite realize that this was a public vs. private distinction. I created `tsk_treeseq_two_locus_count_stat` to avoid having to create the function parameters for every summary function, essentially avoiding the need to write
```
    sample_count_stat_params_t f_params = { .sample_sets = sample_sets,
        .num_sample_sets = num_sample_sets,
        .sample_set_sizes = sample_set_sizes,
        .set_indexes = set_indexes };
```
for every single summary function (can be error prone, makes reading a hassle).

That said, I don't think we actually need to expose `tsk_treeseq_two_locus_count_stat` in the abi. I can remove this from the `trees.h`.

python/tests/test_ld_matrix.py
lkirk: Same here, I can delete. It's just a bit of exposition so that the reader can learn about writing these types of functions.

python/tests/test_ld_matrix.py
lkirk: No, this is simply to add a bit of documentation to these functions, in case people use them as an example. I can delete if that's preferable.

petrelharp: What I meant to say was that this comment isn't providing any useful context for me. So either delete it, or if it's providing something useful that I'm missing then say that.

python/tests/test_ld_matrix.py
petrelharp: As above, I think delete, as written it's kinda mysterious to me.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        On each pair of sites or trees, the summary function is called with
+        haplotype counts for all provided sample sets. The summary function
+        (``f``) must accept two parameters: ``X``, a matrix with shape (3, k)
+        and ``n``, a vector with shape (k,), where k is the number of sample
+        sets provided. ``X`` is a read-only matrix whose rows contain haplotype
+        counts (AB, Ab, aB) per sample set and ``n`` is a vector of sample set
+        sizes. ``f`` must return a list of results with length ``result_dim``.
+
+        What follows is an example of computing ``D`` from a tree sequence
+        (TODO: cite two-locus docs for more details). We convert counts to
```
petrelharp: that PR is merged; we can cite it now (yay!)

petrelharp: perhaps @apragsdale could stick this in?

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        proportions, then compute ``D``, returning a numpy array with length
+        equal to the number of ``result_dim``.
```
```suggestion
        equal to the number of sample sets.
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        result_dim,
+        norm_f=None,
+        polarised=False,
+        sites=None,
+        positions=None,
+        mode="site",
+    ):
+        """
+        Compute two-locus statistics with a user-defined python function that
+        operates on haplotype counts. TODO: reference modes in two-locus docs.
```
apragsdale:
```suggestion
        operates on haplotype counts. Statistics can be computed in ``site`` mode (see :ref:`sec_stats_two_locus_site`) or ``branch`` mode (see :ref:`sec_stats_two_locus_branch`). 
```

apragsdale: This probably fails the line length,

python/tskit/trees.py
apragsdale:
```suggestion
        The result will be equivalent to using ``ts.ld_matrix(stat="D")``
        (see :ref:`sec_stats_two_locus` for usage of the built-in LD matrix calculation,
        and :ref:`sec_stats_two_locus_summary_functions` for available statistics).
        We convert counts to
```

apragsdale: Something along these lines would be helpful here.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+
+            def D(X, n):
+                pAB, pAb, paB = X / n
+                pA = pAb + pAB
+                pB = paB + pAB
+                return pAB - (pA * pB)
+
+        The summary function is called for each pair of sites or trees,
+        producing results that must be combined when multiallelic sites are
+        present (``site`` mode only), summary function results must
```
apragsdale:
```suggestion
        present (``site`` mode only), so summary function results
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        need to be normalised in order to be aggragated for all pairs of alleles
+        between both sites. Branch statistics and biallelic sites do not require
+        any normalisation, ``norm_f`` is only called if one of the two sites
```
apragsdale:
```suggestion
        any normalisation, and ``norm_f`` is only called if one of the two sites
```

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        under consideration is multiallelic. TODO: reference two-locus docs for
+        further information about normalisation. ``norm_f`` is a normalisation
```
apragsdale:
```suggestion
        under consideration is multiallelic. See :ref:`sec_stats_two_locus_computational_details` for
        further information about normalisation. ``norm_f`` is a normalisation
```

apragsdale: The section label for this reference would need to be added in the stats.md documentation, before the "Computational details" section.

python/tskit/trees.py
```diff
@@ -10927,6 +10931,118 @@ def impute_unknown_mutations_time(
...
+        function that must accept four parameters: ``X`` and ``n`` are the same
+        inputs that ``f`` accepts, along with ``nA`` and ``nB``, which hold the
+        count of ``A`` alleles and ``B`` alleles. For example, if ``A`` is
+        biallelic and ``B`` is triallelic, ``nA=2`` and ``nB=3``. ``f`` must
+        return a list of results with length ``result_dim``. The default
+        normalisation function is identical to ``total_norm`` shown in the
+        example below. ``hap_norm`` is required for normalising
+        :math:`r^2`. Both of these examples return a numpy array with length
+        equal to the number of ``result_dim``.
```
apragsdale:
```suggestion
        equal to the number of sample sets.
```

apragsdale (commented):
A few suggestions, including some links to docs.

python/tskit/trees.py
lkirk: Added the section and its reference

lkirk:
I think things are looking good on this from my end. I've fixed up the docstring, addressing all comments. As mentioned on slack, I think it might be better to open up a separate PR for a longer documentation section, does this sound reasonable?

python/tskit/trees.py
petrelharp, line 10998: these are missing the `return` (and so return None)

python/tskit/trees.py
```diff
@@ -10968,21 +10973,21 @@ def D(X, n):
 
         The summary function is called for each pair of sites or trees,
```
petrelharp, line 10974: Let's say exactly what is computed here. How about this?
```
For a given pair of sites (using ``mode="site"``), this summary function is called
for each pair of alleles. When the summary function is called for a given pair of
alleles, the "haplotype counts" (AB, Ab, aB) in a row of X count the focal allele
and all non-focal alleles. For instance, if the focal allele at the first site is G and
the focal allele at the second site is T, then "Ab" counts the number of
two-locus genotypes that have a G at the first site and do not have a T at the
second site. The alleles that are considered include the ancestral state
only if ``polarised=False``. So, if ``polarised=True`` or either site has more than two
alleles, then there will be more than one pair of alleles, and the results from
applying the summary function to each pair must be combined. This is done
by adding them up, weighted by the value of ``norm_f``.

Concretely, if the alleles at the two sites are indexed by :math:`i` and
:math:`j` respectively, the value of ``f`` computed using alleles :math:`i`
and :math:`j` is :math:`f_{ij}`, and the value of ``norm_f`` computed
using the two alleles is :math:`w_{ij}`, then the statistic computed for the
two sites is :math:`\sum_{ij} f_{ij} w_{ij}`.

For ``mode="branch"``, the summary function is called for each pair of branches
in the two trees, with haplotype counts that would be produced by a mutation on
each of those branches, as described at :ref:`sec_stats_two_locus_summary_functions`.
```

petrelharp (commented):
I think this is good, pending the doc suggestions.

python/_tskitmodule.c
```diff
@@ -7940,12 +7941,289 @@ parse_positions(TreeSequence *self, PyObject *positions, npy_intp *out_dim)
...
             goto out;
         }
+        PyArray_CLEARFLAGS(array, NPY_ARRAY_WRITEABLE);
         *out_dim = PyArray_DIM(array, 0);
     }
 out:
     return array;
 }
 
+typedef struct {
```
jeromekelleher, line 7951: There's a *lot* of complexity here, which would take me quite a bit of time to get my head around and be sure it's working properly. Before I do that, can we consider whether we really **need** to expose this interface to Python? My sense is that the general single stat API isn't really used that much - performance is unavoidably horrible. So, is there much to be gained from this?

I'm happy to dig deeper here, I just wanted to check if there's any real gain from this complexity.

petrelharp: I think the answer to this is "currently unclear". So perhaps we put this off until someone has a demonstrated need for it?

apragsdale: I also agree that it is "currently unclear." One difference between the single-site and the two-site stats methods is that there are quite a few more single-site statistics supported within tskit, so you don't need to go the route of the general stat approach. We have a hanful of more commonly used two-locus statistics implemented, but there are fewer. For example, there is no two-locus sampling distribution stat (analogous to the one-site SFS).

apragsdale: > We have a hanful of more commonly used two-locus statistics implemented, but there are fewer. For example, there is no two-locus sampling distribution stat (analogous to the one-site SFS).

Thinking about this more - I suppose this would argue for focusing on supporting those handful of other statistics that would be more commonly used, rather than a general stat framework that is going to be prohibitively slow on larger tree sequences.

petrelharp: I think that's the state of things? There's enough work remaining on this that we should probably shelve it unless someone has a use case? But it can be easily revived, and if you want to argue otherwise now then we could brainstorm how to make it easier to review/merge.

lkirk: I tend to agree with the points made about the utility of this method. On one hand I think it's quite nice for prototyping new stats and learning about the inner workings of the two-locus methods, but overall as an analysis tool the performance is prohibitively slow (especially for branch statistics which have a quadratic penalty for each summary function call). I have some ideas about how native python functions might be used in a fast way, but I think that can be moved to a separate analysis package.

> There's enough work remaining on this that we should probably shelve it unless someone has a use case?

I'm a little bit unsure what remains other than tweaks to documentation, I am pretty sure we hammered out all of the low level details. That's my understanding anyway.

lkirk:
Given the discussion above, should this be closed?

python/_tskitmodule.c
petrelharp: FWIW, there's apparently substantial work in making sure that the python memory management stuff (in `_tskitmodule.py` is totally solid (this from Jerome). I know you've done some careful work on this, and so I think if we did have some solid use cases then we'd figure out how to get that work done in a reasonable way.

petrelharp:
I think so. Thanks for your understanding on this! I know you put a ton of work in, and this will be useful in the future (e.g., if someone want to pull out this branch and compile it, they can use it.)

lkirk:
Not a problem, I appreciate all the work you did to review this @petrelharp.

python/_tskitmodule.c
jeromekelleher: There's a lot of new stuff in the Python C API here @lkirk, and going through this carefully is where the work comes in. There's a *lot* of footguns lying around, and there's no shortcut to working through every detail I'm afraid. Most of the code in _tskitmodule is following standard patterns which I worked out carefully over the years, so it's easy enough to verify. New patterns need careful attention.

I lost several weeks once chasing down a bug that crept in through the Python C API (an off by one bounds error, writing a \0 into the Python heap unexpectedly), where I ended up suspecting compiler bugs and all sorts before I finally tracked it down. Memory bugs in the Python stack are super insidious, because they may not show up for hours after first happening and are often materialise in completely unrelated places.

So - sorry I don't have time to go through this now (given the perceived utility, ultimately). I appreciate you've put loads of work in here, so if you really want to get it in, I will make some time.

lkirk: @jeromekelleher I understand, this part of the stack has always been tricky to work with, especially because there aren't a lot of great static analysis tools for ensuring correctness. I agree with the decision to close, that's not an issue. I almost responded to Peter's last [comment](https://github.com/tskit-dev/tskit/pull/3426#discussion_r3126762357) yesterday, I think my [comment](https://github.com/tskit-dev/tskit/pull/3426#discussion_r3126473117) sounded much too confident. It wasn't a call for review, I was wondering if there was more work to be done if someone were to pick it up again. It sounds like the answer is: "more review and testing + some documentation". Sorry if I sounded dismissive, I know this isn't easy stuff to review and the PR has grown quite large.

jeromekelleher: No worries @lkirk!

PR #3427, opened 2026-03-12
remove consts from tsk_json_struct_metadata_get_blob (closes #3425)

### LINKED ISSUE #3425: `json+struct` codec should take pointers to pointers to **non-const** buffers
Making a new issue for this at the request of @jeromekelleher since #3306 was merged and closed.  Tagging @petrelharp since he'll be a client of this API too, in pyslim.

Copy-paste from #3306:

> In the declaration:
> 
> `tsk_json_struct_metadata_get_blob(const char *metadata, tsk_size_t metadata_length, const char **json, tsk_size_t *json_length, const uint8_t **blob, tsk_size_t *blob_length)`
> can `const uint8_t **blob` be `uint8_t **blob` without the `const`? SLiM sometimes actually munges the incoming mutation metadata (remapping population indices if the user has asked for that, prior to the main metadata-parsing code working with the data). I could of course make a copy of the binary blob, but since it is expected to be quite large I'd prefer not to, so as to not push the memory high-water mark unnecessarily. It doesn't seem like tskit has any special interest in keeping that metadata const, right? I'd have to cast away the `const` to modify it, though, which according to the C++ standard is "undefined behavior" although I imagine it would probably work in practice.
> 
> For consistency the JSON data should perhaps also not be `const`, but I don't care about that one way or the other; SLiM doesn't need to modify that, and if it did, I could just copy the data since that is small.
> 
> (If you want the mainstream API to be `const` for safety, perhaps you could provide a non-const version of it for those who need it?)

In followup discussion, @jeromekelleher wrote:

> These have been changes to const char **blob, for consistency. The const is necessary here for us to compile and signals that the function doesn't alter the buffers.

I replied (edited/expanded a bit):

> It's a bit annoying to require a caller to pass a pointer to const just because the function itself doesn't alter the buffers. If it was just a question of a `const char *` parameter and the user passing in a `char *` to the function, that would be fine; the compiler would automatically accept the `char *` for the parameter, rather than requiring the caller to provide a `const char *`. But because of the extra level of indirection, a `const char **` parameter does not automatically accept a `char **` parameter, requiring the caller to do some faffing about with casts and such. I would argue that it is not standard practice to put the `const` there – when passing a pointer to a pointer to data, it is already assumed that the function is not supposed to muck around inside the data that is ultimately pointed to, and that guarantee is not typically made with `const` in APIs (and in many cases _cannot_ be made, in C/C++ syntax, such as guaranteeing that a function won't modify stuff that is pointed to by pointers contained within a `const` structure that it is passed in; C and C++ don't have any concept of "deep" constness). So I don't think you ought to have that `const` there. But it's not the end of the world; I thought initially that dealing with this API issue would require SLiM to cast away `const`, producing undefined behavior, but that is not the case.  It's just annoying and non-standard (IMHO).

(I'd question the decision to use `char` instead of `uint8_t`, though; they are not the same thing, and I think the semantics of `uint8_t` are more appropriate – for the binary data, at least.)

petrelharp:
As discussed in #3425, it is maybe helpful to remove the `const` from some of the arguments to [`tsk_json_struct_metadata_get_blob`](https://github.com/tskit-dev/tskit/blob/1835ea3bbb51e4b5dff698dcae43c917873c73bc/c/tskit/core.c#L145).

My superficial understanding of the issue is as follows: suppose you have a function that accepts a pointer, say `*blob`, and doesn't modify the stuff that pointer points at. Then it's fine to define your function to accept `const foo *blob` but then pass it in a non-const `foo *blob` from somewhere else, since `const` is supposed to mean "we don't change this *here*" not "we don't change this *anywhere*".  **However** if it takes a pointer-to-a-pointer, say `const foo **blob` then [there are ways](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion) that you'd actually be able to modify the contents being ultimately pointed at without, like, noticing. So it is *not* cool to pass in `foo **blob`.

So, I agree - seems like we should indeed remove the `const`.

According to [this page](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion) the "most common solution" is to declare it like `const foo* const *blob`; however, I *think* this means you can't change the pointer `*blob` either and we do indeed need to do that in this function. And I tried it and it doesn't seem to work.

So here I've removed the `const` from `**blob`. This meant I also needed to remove `const` from `**json` *and* `*metadata`, since these are all pointing at the same chunk of memory.

### DIFF
--- a/c/tests/test_core.c
+++ b/c/tests/test_core.c
@@ -101,9 +101,9 @@ test_json_struct_metadata_get_blob(void)
 {
     int ret;
     char metadata[128];
-    const char *json;
+    char *json;
     tsk_size_t json_buffer_length;
-    const char *blob;
+    char *blob;
     tsk_size_t blob_length;
     uint8_t *bytes;
     tsk_size_t metadata_length;
@@ -111,9 +111,9 @@ test_json_struct_metadata_get_blob(void)
     size_t json_length;
     size_t payload_length;
     size_t total_length;
-    const char json_payload[] = "{\"a\":1}";
-    const uint8_t binary_payload[] = { 0x01, 0x02, 0x03, 0x04 };
-    const uint8_t empty_payload[] = { 0 };
+    char json_payload[] = "{\"a\":1}";
+    uint8_t binary_payload[] = { 0x01, 0x02, 0x03, 0x04 };
+    uint8_t empty_payload[] = { 0 };
 
     bytes = (uint8_t *) metadata;
     header_length = 4 + 1 + 8 + 8;
@@ -135,7 +135,7 @@ test_json_struct_metadata_get_blob(void)
     ret = tsk_json_struct_metadata_get_blob(
         metadata, metadata_length, &json, &json_buffer_length, &blob, &blob_length);
     CU_ASSERT_EQUAL(ret, 0);
-    CU_ASSERT_PTR_EQUAL(json, (const char *) bytes + header_length);
+    CU_ASSERT_PTR_EQUAL(json, (char *) bytes + header_length);
     CU_ASSERT_EQUAL(json_buffer_length, (tsk_size_t) json_length);
     if (json_length > 0) {
         CU_ASSERT_EQUAL(memcmp(json, json_payload, json_length), 0);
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):

bhaller:
> I suspect maaaaybe github's automated stuff dropped the ball? Trying again; will manually merge if not.

The auto-merge seems to have failed again.  Maybe there's a time limit for all the checks to complete, and if they don't finish within that time limit the auto-merge gets cancelled?  In which case that time limit needs to get extended somehow/somewhere, otherwise the auto-merge is seeming pretty useless.  :->  Anyhow, manual merge for now, I guess?

PR #3437, opened 2026-03-13
Json struct alignment

### LINKED ISSUE #3423: `json+struct` codec should provide a memory alignment guarantee for the binary component
Making a new issue for this at the request of @jeromekelleher since #3306 was merged and closed.  Tagging @petrelharp since he'll be a client of this API too, in pyslim.

Copy-paste from #3306:

> One more thing that occurs to me @benjeffery: it would perhaps be nice to provide a memory alignment guarantee for the binary portion (assuming that the metadata chunk as a whole is at an aligned address, which is guaranteed by malloc). On my end, when writing out the metadata, I could of course insert null bytes in between the JSON and the binary to align the binary portion; but this gets tricky because those null bytes would have to be described in the binary schema, and the number of null bytes is variable since the length of the JSON string is variable, so it makes everything a bit of a mess. (I guess SLiM would have eight different possible binary schemas that it uses, with from 0 to 7 null bytes at the start of the binary data as needed, eww.) I could put spaces at the end of the JSON string instead to produce alignment, but I'm worried that that would make the JSON string non-canonical and make tskit produce an error, so I think (?) that solution is out.
> 
> Having an alignment guarantee (I'd suggest 8-byte) would be nice because SLiM will be writing and reading lots of data directly to/from the binary blob, so if it's unaligned it just makes for additional headaches and lower performance in the read/write code. If this alignment guarantee were built into the format expected by the json+struct codec, it would smooth the path. Thoughts?

Addendum now:

I've implemented my own alignment bytes for the binary blob in SLiM, and it was indeed a pain in the tuchus.  It requires eight different versions of the binary schema, specifying the number of alignment bytes used as being from 0 to 7 (to produce 8-byte alignment).  Lots of extra faffing about at both write and read time, which pyslim and anybody else who cares will have to duplicate.  I strongly recommend that automatic 8-byte alignment be added to this codec so clients don't have to worry about all this.

### LINKED ISSUE #3424: `json+struct` codec should check for a length problem with `!=` not `<`
Making a new issue for this at the request of @jeromekelleher since #3306 was merged and closed.  Tagging @petrelharp since he'll be a client of this API too, in pyslim.

Copy-paste from #3306:

> Right now you have:
> 
>     if ((uint64_t) metadata_length < total_length) {
>         ret = tsk_trace_error(TSK_ERR_FILE_FORMAT);
>         goto out;
>     }
> 
> Can I suggest that the test should be != rather than <? Seems like it ought to match exactly if the metadata is well-formed; and I had a subtle bug just now that had me tearing my hair out that would've been caught with a != test here. Is there a reason that you want to allow a mismatch here?

In followup discussion, @jeromekelleher wrote:

> I think there's a good reason to be lenient here in allowing the length of the input buffer to be greater than the sum of the two segments - you might want to add padding for alignment. So, I think the current version is good.

I replied:

> Hmm – this doesn't seem useful for alignment, since the length difference goes at the end...? For alignment, you want to have extra bytes inserted in between the JSON and the binary, which is not presently supported. (I suggested adding alignment bytes above, in fact; that would be quite a useful thing to provide in this PR!) As far as I can see, all that < does for the client is introduce the possibility of bugs. (I'm handling alignment myself in SLiM now, in fact, and the use of < here was in no way helpful for doing that; all it did was hide a bug from me.)

To expand on that a bit: any bytes added for alignment purposes have to be part of the binary blob itself, in the current design.  There's no way for alignment bytes to be inserted in between the JSON and the binary, except for the client to put those bytes at the beginning of the binary blob themselves – in which case the two lengths should still be equal.  It seems to me that it is never useful or correct for `metadata_length > total_length` to be true in the code above, so the correct test is with `!=`, not `<`.  I have, in fact, implemented alignment bytes in SLiM, and the two lengths are still equal. But perhaps I'm missing the point somewhere?

petrelharp:
*Edit:* This is now ready. It
- closes #3423, by (a) adding the padding bytes to the C function (`tsk_json_struct_metadata_get_blob`)
- also closes #3424, [here](https://github.com/tskit-dev/tskit/pull/3437/changes#diff-9156844acd1a9912f6b2696dc3b4b078fe5c134b483f94505338d1bad052251fR194)

This is pretty straightforward. However, I got confused along the way, basically about "why aren't the C and python code being tested against each other"; and by the end I decided that it was all okay. Keep reading if you want the explanation.

Examining #3306 I see that the python code does not use the C function (`tsk_json_struct_metadata_get_blob`) at all, so we've got no good current way to test these against each other. Options that I can think of are:

1. remove the C code entirely
2. write an entirely new class of tests that reads in python-produced tree sequence files and then parses them in C and compares to what is expected
3. write a python method that uses the C function
4. not test that the C getter and the python getter do the same thing

However, (2) and (3) are ugly. (2) would break the nice clean one-way C->python arrangment we have. And (3) would leave us with two ways of doing the same thing. This leaves us with (1) or (4).

The motivation for putting the C code in here was that it's very annoying to write and it would help client code a lot to have it available. However, the setter is arguably more useful than the getter (since one does not actually decode the metadata in C), so for (4) it seems we'd still want to additionally write the setter. So, I'm leaning towards (1), out of laziness. We could perhaps document somewhere the example setter/getter code (copied from SLiM and here respectively), to make this easy for other client code, which was the goal.

*Edit:* see below for the resolution.

### DIFF
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -294,7 +294,8 @@ class JSONStructCodec(AbstractMetadataCodec):
         header = self._HDR.pack(
             self.MAGIC, self.VERSION, len(json_bytes), len(blob_bytes)
         )
-        return header + json_bytes + blob_bytes
+        padding_bytes = bytes((-(len(header) + len(json_bytes))) % 8)
+        return header + json_bytes + padding_bytes + blob_bytes
 
     def decode(self, encoded: bytes) -> Any:
         if len(encoded) >= self._HDR.size and encoded[:4] == self.MAGIC:
@@ -302,12 +303,16 @@ class JSONStructCodec(AbstractMetadataCodec):
             if version != self.VERSION:
                 raise ValueError("Unsupported json+struct version")
             start = self._HDR.size
-            if jlen > len(encoded) - start or blen > len(encoded) - start - jlen:
+            padding_length = (-(start + jlen)) % 8
+            if (
+                jlen > len(encoded) - start
+                or blen > len(encoded) - start - jlen - padding_length
+            ):
                 raise ValueError(
                     "Invalid json+struct payload: declared lengths exceed buffer size"
                 )
             json_bytes = encoded[start : start + jlen]
-            blob_bytes = encoded[start + jlen : start + jlen + blen]
+            blob_bytes = encoded[start + jlen : start + jlen + blen + padding_length]
             json_data = self.json_codec.decode(json_bytes)
             struct_data = self.struct_codec.decode(blob_bytes)
             overlap = set(json_data).intersection(struct_data)
--- a/python/tskit/metadata.py
+++ b/python/tskit/metadata.py
@@ -201,6 +201,12 @@ class JSONStructCodec(AbstractMetadataCodec):
     The codec expects a metadata schema with separate ``json`` and ``struct``
     subschemas and produces a single dict containing the union of the keys from
     those subschemas after decoding.
+
+    The structure of the encoded metadata is as follows: first, a fixed-size
+    header that contains the number of bytes for both ``json`` and ``struct``
+    portions; next, the json; then a variable number of zeroed padding bytes
+    that brings the length to a multiple of 8 for alignment; and finally
+    the struct-encoded binary portion.
     """
 
     MAGIC = b"JBLB"
--- a/c/tests/test_core.c
+++ b/c/tests/test_core.c
@@ -109,6 +109,7 @@ test_json_struct_metadata_get_blob(void)
     tsk_size_t metadata_length;
     size_t header_length;
     size_t json_length;
+    size_t padding_length;
     size_t payload_length;
     size_t total_length;
     char json_payload[] = "{\"a\":1}";
@@ -118,8 +119,9 @@ test_json_struct_metadata_get_blob(void)
     bytes = (uint8_t *) metadata;
     header_length = 4 + 1 + 8 + 8;
     json_length = strlen(json_payload);
+    padding_length = (8 - ((header_length + json_length) & 0x07)) % 8;
     payload_length = sizeof(binary_payload);
-    total_length = header_length + json_length + payload_length;
+    total_length = header_length + json_length + padding_length + payload_length;
     CU_ASSERT_FATAL(total_length <= sizeof(metadata));
     memset(metadata, 0, sizeof(metadata));
     bytes[0] = 'J';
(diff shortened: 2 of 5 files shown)


petrelharp:
Update: @bhaller tells me that

> The problem with having a setter function is that it would require the client to prepare the JSON is a buffer, prepare the binary in a buffer, and then pass pointers to those buffers in to the setter function, which would set up the final aggregate buffer and copy the JSON and binary in.  That is a bad design for performance reasons – it copies all the metadata (which might be quite large), wasting time but even more importantly causing the high-water memory usage for the process to double.  I suppose tskit could provide a setter anyway, for simplicity for clients that don’t care about those performance issues; but SLiM wouldn’t use it, probably.  If having the setter would make your life easier for the testing, I’d have no objection to adding it as long as there is no assertion that clients have to use the setter; the format used by the codec should remain public, and clients should remain free to write their data themselves.

petrelharp:
After some more discussion with @bhaller:

*Why do we want support for JSON+struct?* We want to have support for JSON+struct metadata because it is *extremely useful*. The reason it is extremely useful is that sometimes client code will want to store large amounts of information that doesn't fit in the standard table format for some reason, and this is a very clever (thanks @benjeffery) way to do that, that is way more seamless than other alternatives.

*Why have support for it at all in C?* Other metadata parsing stuff is not in C at all, for good reason (we don't want to write or import a JSON parser or a struct parser, etcetera). However, the json and struct codecs are pretty straightforward for client C code to do themselves; this codec is annoying enough to set up (special header with magic bytes, etc) that it does make sense to provide a bit of help.

*Well then, why don't we have a setter in C, then?* I think that #3306 intended to provide a setter, but this fell through the cracks. So, I think it would be a good thing to have for the same reason as in the previous paragraph. *However*, I can't currently justify the time to actually impmlement and test it. (If we *did* provide a setter, then (with the note by @bhaller about memory requirements above in mind) we might want to have an API return a pointer to the binary metadata portion, so that the client code could fill it in, rather than passing in the pre-assembled binary portion, thus requiring an extra copy.)

So, I'm leaning towards option (4), "leave things the way they are". This is not perfect because (a) no testing that the C and python do the same thing, and (b) no setter in C; however, what we have is useful and well-tested, and furthermore, if the C and python were in disagreement we'd find out in obvious ways immediately.

bhaller:
Sounds good, with the caveat that "leave things the way they are" still needs the alignment padding bytes to be added in the python code as we talked about, yes?

petrelharp:
Okay - this is ready to go. Could @jeromekelleher or @benjeffery take a (hopefully quick) look at this?

python/tskit/metadata.py
```diff
@@ -294,20 +300,25 @@ def encode(self, obj: Any) -> bytes:
...
+            if (
+                jlen > len(encoded) - start
+                or blen > len(encoded) - start - jlen - padding_length
+            ):
                 raise ValueError(
                     "Invalid json+struct payload: declared lengths exceed buffer size"
                 )
             json_bytes = encoded[start : start + jlen]
-            blob_bytes = encoded[start + jlen : start + jlen + blen]
+            blob_bytes = encoded[start + jlen : start + jlen + blen + padding_length]
```
benjeffery, line 321: Don't you need the padding at the start? I think the test isn't catching this as its payload is 27, so it's padding is zero?

petrelharp:
Closing this in favor of #3439 and #3440.

PR #3439, opened 2026-03-20
Json struct c docs

petrelharp:
I think that removing the json+struct C code from tskit itself and providing an example in [the C docs](https://tskit.dev/tskit/docs/latest/c-api.html#examples) is probably the way to go. @bhaller provided the C code, modified from the method and tests in #3437.

Currently this has the additional docs that are also in #3440.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -527,7 +527,7 @@ of `B`, `H`, `I`, `L` or `Q` which have the same meaning as in the numeric
 types above. `L` is the default. As an example:
 
 ```
-{"type": "array", {"items": {"type":"number", "binaryFormat":"h"}}, "arrayLengthFormat":"B"}
+{"type": "array", "items": {"type":"number", "binaryFormat":"h"}, "arrayLengthFormat":"B"}
 ```
 
 Will result in an array of 2 byte integers, prepended by a single-byte array-length.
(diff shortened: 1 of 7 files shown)


docs/metadata.md
```diff
@@ -527,7 +527,7 @@ of `B`, `H`, `I`, `L` or `Q` which have the same meaning as in the numeric
 types above. `L` is the default. As an example:
 
 ```
-{"type": "array", {"items": {"type":"number", "binaryFormat":"h"}}, "arrayLengthFormat":"B"}
```
bhaller, line 530: This was a doc bug, I take it?  I'd fix it outside of this PR, since it is unrelated to the work here, right?

docs/metadata.md
```diff
@@ -555,6 +555,73 @@ As a special case under the `struct` codec, the top-level type of metadata can b
...
+and then an array of integers in binary (using the ``struct`` codec).
+If the array of integers is large, this could result in
+much better performance.
+
+
+#### Binary representation
+
+The underlying structure of the JSON+struct codec is as follows.
+(If you're not writing out data in this format,
+you don't need to worry about this.)
```
bhaller: This statement is no longer true; even if you're only reading in data in this format, you now have to worry about it, on the C side at least.  This text should point to the example as a demonstration of how to handle this format for reading and writing.

docs/metadata.md
```diff
@@ -555,6 +555,73 @@ As a special case under the `struct` codec, the top-level type of metadata can b
...
+(1) some magic bytes;
+(2) a version number;
+(3) the length of the JSON in bytes;
+(4) the length of the binary (struct) data in bytes;
+(5) the JSON data;
+(6) zero-ed "padding" bytes to bring the start of the binary section
+ into 8-byte alignment;
+(7) the binary data.
+The structure of the binary data is specified using the "struct" portion
+of the metadata schema.
```
bhaller: The above needs to be much more specific, since tskit is not doing any of this work for C clients any more.  The exactly binary structure of the json+struct codec needs to be documented here.  "Some magic bytes" should be "four magic bytes, the ASCII characters 'J', 'B', 'L', and 'B'", "a version number" should be "a 1-byte (uint8_t) version number, currently 1", "the length of the JSON in bytes" should be "a 64-bit (uint64_t) length in bytes for the JSON data, in little-endian format", and so forth.  The information given should be sufficient that someone could write the example code themselves based upon the specification given here.  This specification is, in effect, part of tskit's public API, and needs to be treated as such.

petrelharp: The stuff about `struct` is just a bit further up in this document: https://tskit.dev/tskit/docs/latest/metadata.html#struct
I'll link to that.

docs/metadata.md
petrelharp: yep

petrelharp: And sure, but it's so minor it's not worth the bother of a whole different commit/pr/etc.

bhaller: sure, I do much worse every day :->  wasn't sure how strict the tskit folks like to be about this sort of thing  :->

petrelharp:
Codecov is complaining because the example C file is not being hit. AFAICT this is expected, and perhaps we should change [this line](https://github.com/tskit-dev/tskit/blob/main/codecov.yml#L28) to ask codecov to just cover `c/tskit/` instead of `c/`?

bhaller (approved):
LGTM!

petrelharp:
Indeed, none of the `examples/` code is hit by codecov (as we'd expect, just verifying):
Screenshot From 2026-04-14 09-29-29

petrelharp:
Merging this since it's been reviewed carefully by Ben H, after consulting with Jerome. I need to double-check the change to `.codecov.yml` hasn't messed something up; will do that in a separate PR.

petrelharp:
I've checked in #3446 that the change to .codecov.yml didn't break codecov's c coverage.

PR #3440, opened 2026-03-20
json+struct: padding bytes and no C code

petrelharp:
This differs from #3437 in that:

- it removes all C code from tskit
- it adds some documentation for the json+struct codec

My intention is that we deal with the python and C code separately, so this supercedes #3437. My proposal for the C side of things is in #3439.


python/tskit/metadata.py
```diff
@@ -294,20 +300,25 @@ def encode(self, obj: Any) -> bytes:
...
+            if (
+                jlen > len(encoded) - start
+                or blen > len(encoded) - start - jlen - padding_length
+            ):
                 raise ValueError(
                     "Invalid json+struct payload: declared lengths exceed buffer size"
                 )
             json_bytes = encoded[start : start + jlen]
-            blob_bytes = encoded[start + jlen : start + jlen + blen]
+            blob_bytes = encoded[start + jlen : start + jlen + blen + padding_length]
```
petrelharp: @benjeffery says in the other PR:
> Don't you need the padding at the start? I think the test isn't catching this as its payload is 27, so it's padding is zero?

petrelharp: lol yes - in my defense, I was planning to go adjust that test to include metadata of different lengths, so presumably I would have got this

c/tskit/core.c
```diff
@@ -32,10 +32,7 @@
 #include <kastore.h>
 #include <tskit/core.h>
 
-#define UUID_NUM_BYTES              16
-#define TSK_JSON_BINARY_HEADER_SIZE 21
-
-static const uint8_t _tsk_json_binary_magic[4] = { 'J', 'B', 'L', 'B' };
+#define UUID_NUM_BYTES 16
```
bhaller, line 35: The diff here is just a bit messy, I don't care but pointing it out in case someone else cares.  :->

c/tskit/core.h
```diff
@@ -1136,10 +1116,11 @@ int tsk_generate_uuid(char *dest, int flags);
 @brief Extract the binary payload from ``json+struct`` encoded metadata.
 
 @rst
-Metadata produced by the JSONStructCodec consists of a fixed-size
-header followed by canonical JSON bytes and an optional binary payload. This helper
-validates the framing, returning pointers to the embedded JSON and binary sections
-without copying.
+Metadata produced by the JSONStructCodec consists of a fixed-size header followed
+by canonical JSON bytes, a variable number of padding bytes (which should be zero)
```
bhaller: could be zero, not should be zero

bhaller: and I don't think it should say "canonical" here -- whether the JSON is "canonical" or not is entirely up to the client using the codec for their own metadata, and should not be required by tskit, right?

c/tskit/core.h
```diff
@@ -1136,10 +1116,11 @@ int tsk_generate_uuid(char *dest, int flags);
...
+to bring the length to a multiple of 8 bytes for alignment, and finally an optional
```
bhaller: rather than "optional" you should say "(which could be of length zero)"; if you're using this codec the binary data is not "optional", in the sense that it needs a schema and needs to conform to that schema; but it might be zero-length

docs/metadata.md
```diff
@@ -527,7 +527,7 @@ of `B`, `H`, `I`, `L` or `Q` which have the same meaning as in the numeric
 types above. `L` is the default. As an example:
 
 ```
-{"type": "array", {"items": {"type":"number", "binaryFormat":"h"}}, "arrayLengthFormat":"B"}
+{"type": "array", "items": {"type":"number", "binaryFormat":"h"}, "arrayLengthFormat":"B"}
```
bhaller: I think you had this diff in your other PR too?  And, as there, if this is fixing a doc bug it should probably be done separately, not in this PR, to avoid confusion.

docs/metadata.md
```diff
@@ -555,6 +555,73 @@ As a special case under the `struct` codec, the top-level type of metadata can b
 union of `object` and `null`. Set `"type": ["object", "null"]`. Properties should
 be defined as normal, and will be ignored if the metadata is `None`.
 
+(sec_metadata_codecs_jsonstruct)=
```
bhaller: ah, all of these diffs are also in your other PR, in fact, right?  see the other PR for my comments on the remainder of this file.

c/tskit/core.h
petrelharp: whoops - this stuff should have been deleted! sorry

docs/metadata.md
petrelharp: yes, that's right - I'll move this over there

python/tskit/metadata.py
petrelharp: fixed

c/tskit/core.c
bhaller: The diff here now shows that `#define UUID_NUM_BYTES 16`, which is odd since all other C code is being removed; nothing now uses this define, right?

c/tskit/core.h
bhaller: These comments still appear to apply to the diffs as shown in GitHub now.  Am I reading the wrong version of this PR somehow?

c/tskit/core.h
bhaller: Ditto.

bhaller (commented):
stopping for now since I suspect I'm somehow looking at the wrong diffs

c/tskit/core.c
petrelharp: apparently it's used elsewhere:
```
$ grep UUID_NUM_BYTES tskit/*
tskit/core.c:#define UUID_NUM_BYTES 16
tskit/core.c:    if (!CryptGenRandom(hCryptProv, (DWORD) UUID_NUM_BYTES, buf)) {
tskit/core.c:    if (fread(buf, UUID_NUM_BYTES, 1, f) != 1) {
tskit/core.c:    uint8_t buf[UUID_NUM_BYTES];
```

python/tskit/metadata.py
```diff
@@ -294,20 +300,27 @@ def encode(self, obj: Any) -> bytes:
         header = self._HDR.pack(
             self.MAGIC, self.VERSION, len(json_bytes), len(blob_bytes)
         )
-        return header + json_bytes + blob_bytes
+        padding_bytes = bytes((-(len(header) + len(json_bytes))) % 8)
```
bhaller: this zeroes the bytes, I guess?

python/tskit/metadata.py
```diff
@@ -294,20 +300,27 @@ def encode(self, obj: Any) -> bytes:
...
+        return header + json_bytes + padding_bytes + blob_bytes
 
     def decode(self, encoded: bytes) -> Any:
         if len(encoded) >= self._HDR.size and encoded[:4] == self.MAGIC:
             _, version, jlen, blen = self._HDR.unpack_from(encoded)
```
bhaller, line 308: the syntax is really cryptic here; I guess the `_` says "skip the first bit, it's the magic and we already checked it so we don't need it in a variable"?

python/tskit/metadata.py
```diff
@@ -294,20 +300,27 @@ def encode(self, obj: Any) -> bytes:
...
             if version != self.VERSION:
                 raise ValueError("Unsupported json+struct version")
             start = self._HDR.size
-            if jlen > len(encoded) - start or blen > len(encoded) - start - jlen:
+            padding_length = (-(start + jlen)) % 8
```
bhaller: hmm, this would be dangerous in C/C++, which is not super clear on what you should expect from a modulo done on a negative number; it used to be implementation-defined, so it is best to avoid it.  is this 100% safe and well-defined in Python?  even if so, wouldn't it be conceptually simpler (for those who don't have the Python standard memorized) to calculate it without this assumption?

python/tskit/metadata.py
```diff
@@ -294,20 +300,27 @@ def encode(self, obj: Any) -> bytes:
...
+            if (
+                jlen > len(encoded) - start
+                or blen > len(encoded) - start - jlen - padding_length
+            ):
                 raise ValueError(
                     "Invalid json+struct payload: declared lengths exceed buffer size"
                 )
             json_bytes = encoded[start : start + jlen]
```
bhaller, line 321: This is surprising to me; if jlen is, say, 5, that means it is 5 bytes, so I'd expect to want to extract bytes from start to start + jlen - 1.  Is the sequence operator here really defined to extract up to, but not including, the last index in the sequence??  Or is this a bug?  Or is my brain misfiring?

python/tskit/metadata.py
```diff
@@ -294,20 +300,27 @@ def encode(self, obj: Any) -> bytes:
...
-            blob_bytes = encoded[start + jlen : start + jlen + blen]
+            blob_bytes = encoded[
+                start + jlen + padding_length : start + jlen + padding_length + blen
```
bhaller, line 323: same comment here

python/tskit/metadata.py
bhaller: I mean, in R:

> 0:5
[1] 0 1 2 3 4 5

So either you're extracting six bytes (bug), or Python is somehow leaving off the last byte here to be "helpful"...??

bhaller (commented):
next review

python/tskit/metadata.py
petrelharp: Yes:
```
>>> bytes(10)
b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```
and
```
class bytes(object)
 |  bytes(int) -> bytes object of size given by the parameter initialized with null bytes
```

python/tskit/metadata.py
petrelharp: Right; this is nice python syntax once you're used to it; it's "tuple unpacking": for instance, if `x` is a list (or equivalent) of length 3 then
```
a, b, c = x
```
sets `a`, `b`, and `c` to the entries of `x`. We don't need the first value that is unpacked here, so by convention in python when there's something that you need to give a name to but aren't going to use you call that thing `_`. (The other common case for this is `for _ in abc: ...`.)

petrelharp: And it'd be cleaner if `_HDR.unpack_from()` didn't return that first bit at all, but `_HDR` is a `struct.Struct( )` object, so we don't have control over what it returns.

python/tskit/metadata.py
petrelharp: Wow, really? Come on, C, how to do modulo on a negative number is an unambiguous thing. I was happy with this because it seemed cleaner and more obviously right. To me, I guess.

python/tskit/metadata.py
petrelharp: > Is the sequence operator here really defined to extract up to, but not including, the last index in the sequence??

Yep! That's how python works, for better or worse.

python/tskit/metadata.py
petrelharp: thanks for checking!

python/tskit/metadata.py
petrelharp: (I changed it tho)

python/tests/test_metadata.py
```diff
@@ -662,14 +664,59 @@ def test_round_trip_with_struct_and_json(self):
...
+        return tskit.MetadataSchema(schema)
+
+    def test_round_trip_with_struct_and_json(self):
+        schema = dict(self.schema_with_binary(2).schema)
+        # adds a 'xyz' property to the schema, which stores an array of numbers
+        schema["struct"]["properties"]["xyz"] = {
+            "type": "array",
+            "arrayLengthFormat": "H",
+            "items": {"type": "number", "binaryFormat": "i"},
+        }
```
bhaller: OK, so if I'm now understanding, the binary struct schema now ends up with two parts: a series of 4-byte integers that are set up by schema_with_binary(), the number of which is determined by the `num_blobs` parameter, which is passed as 2 above; AND a variable-length array of, again, 4-byte integers, set up by the code here.  If that's what you want, it's fine, but I continue to find it really confusing that you make the first part of the schema in that separate function above, and then here you convert it into a dictionary (for some reason), add another chunk to it, and then convert it back to a schema.  What is the point of these complicated shenanigans?  If you want `xyz` in the schema, why not add it in schema_with_binary() above?  If the intent is that chunks will be added in, why not leave it as a dictionary and return that, and just convert it to a schema at the end?

I guess it seems like the motivation is to share the code in schema_with_binary() between two places; but it's only a couple of lines of code shared, and it makes it really hard to follow what the final schema is that we're trying to build.  (Schemas are already really confusing since their structure is complex and the keys they use are so generically named.)

My confusion here is doubtless stemming from my lack of familiarity with Python idioms, as well as my lack of facility working with schemas.  I'm just struggling to try to figure out whether there is some concrete motivation to converting back and forth between schemas and dictionaries, and what is _really_ going on in this code (since it is often the hidden "magic" that leads to bugs).  Sorry for my cluelessness, I realize my struggles in this review are probably annoying.  To you this code seems super-simple, even trivial, I'm sure; to me, since it is in a foreign language, it requires a fair bit of thought, and when I have questions it isn't clear to me whether those questions are dumb or maybe important...

python/tests/test_metadata.py
```diff
@@ -662,14 +664,59 @@ def test_round_trip_with_struct_and_json(self):
             },
             "struct": {
                 "type": "object",
-                "properties": {"blob": {"type": "integer", "binaryFormat": "i"}},
+                "properties": {},
             },
         }
+        for j in range(num_blobs):
```
bhaller: above you wrote "labeled b0, ... bX"; so here, is X num_blobs, or num_blobs-1?  and it does start at b0, not b1?  I now don't trust Python to produce the number of iterations that I would expect, since there was that other super-weird thing it did with sequences that we talked about before...

python/tests/test_metadata.py
bhaller, line 701: OK, so this is the specs for the JSON portion here, with a string named label and a number named count

python/tests/test_metadata.py
```diff
@@ -662,14 +664,59 @@ def test_round_trip_with_struct_and_json(self):
...
+            "arrayLengthFormat": "H",
+            "items": {"type": "number", "binaryFormat": "i"},
+        }
         ms = tskit.MetadataSchema(schema)
-        row = {"label": "alpha", "count": 7, "blob": 5}
-        encoded = ms.validate_and_encode_row(row)
-        out = ms.decode_row(encoded)
-        assert out == row
+        for v in [[], [0, 2, 12], [5] * 1000]:
+            row = {"label": "abcdef xyz", "count": 7, "b0": 123, "b1": 0, "xyz": v}
```
bhaller: OK, trying to puzzle out what is going on here.  I'm not sure why the variable name here is "row"; this loop encodes and then decodes three separate `json+struct` codec chunks, right?  And I'm trying to figure out what those chunks contain exactly.  "label" and "count" are the json; I'd expect them to be separated from the binary parts (since the binary schema might also have a key named "label", for example), but here they aren't.  And what is going on with xyz?  Python is magically producing the correct binary for arrays of different sizes?  Is [5]*1000 an array of length 1000 with each entry containing a 5, or something else?

Gotta go for now, my flight is boarding.  I really think it would be MUCH easier for both of us if you could get this reviewed by someone who knows Python, but I'll return to it when I get to Atlanta.

petrelharp: > I really think it would be MUCH easier for both of us if you could get this reviewed by someone who knows Python

Yes, except it'd be MUCH harder for them since they wouldn't know what's going on with the codec. Thanks for giving it a go.

python/tests/test_metadata.py
petrelharp: >  What is the point of these complicated shenanigans?

Well, the main reason this is separated as separate function is because of the wanting to verify "how many bytes are used to store the ints". I've now changed the code to write out the schema separately in `test_round_trip_with_struct_and_json`, as you suggest, so there's no messing around with schema->dict->schema.

python/tests/test_metadata.py
petrelharp: X-1; I've updated this comment to be a bit more clear.

python/tests/test_metadata.py
petrelharp: > I'm not sure why the variable name here is "row"

Note the name of the functions are `validate_and_encode_row` and `decode_row`; this is because there's one metadata entry per row in the tables. (This is meant for top-level metadata, but that's still what the functions are called...)

petrelharp: > I'd expect them to be separated from the binary parts

Nope, that's the beauty of this codec, it just encodes some stuff in JSON and other things in binary! I'll make this more clear over in the docs in #3439.

petrelharp: >  Python is magically producing the correct binary for arrays of different sizes? Is [5]*1000 an array of length 1000 with each entry containing a 5?

Yes, and yes!

petrelharp:
elsewhere @bhaller points out that
> OK; but since you're adding 4-byte ints, you’re not really testing the different padding byte counts very effectively.  You’re only testing the 0 and 4 alignment cases well, which seems vulnerable to bugs since it’s just “none” and “half”.

I don't think having the length of the *binary* bit be divisible by 4 is a problem, since the number of padding bytes is determined by the length of the JSON (which comes first).  I'll add a few more string lengths to make sure we hit everything.

python/tests/test_metadata.py
petrelharp: see docs in https://github.com/tskit-dev/tskit/pull/3439/changes with clarifying note on this confusing point

python/tests/test_metadata.py
```diff
@@ -657,19 +657,98 @@ def test_round_trip_with_struct_and_json(self):
                 "properties": {
                     "label": {"type": "string"},
                     "count": {"type": "number"},
+                    "stuff": {
+                        "type": "array",
+                        "arrayLengthFormat": "h",
```
bhaller: Is `"h"` legal?  Tskit doc says:

> The format used for the length of the array can be chosen with arrayLengthFormat which must be one of B, H, I, L or Q which have the same meaning as in the numeric types above.

bhaller: Ah, I guess that's for `struct`... for JSON is there even an `arrayLengthFormat` key?  I'm getting confused again, sigh.

python/tests/test_metadata.py
```diff
@@ -657,19 +657,98 @@ def test_round_trip_with_struct_and_json(self):
...
-        row = {"label": "alpha", "count": 7, "blob": 5}
-        encoded = ms.validate_and_encode_row(row)
-        out = ms.decode_row(encoded)
-        assert out == row
+        for v in [[], [0, 2, 12], [5] * 1000]:
+            row = {
+                "label": "abcdef xyz",
+                "count": 7,
+                "b0": 123,
+                "b1": 0,
```
bhaller, line 683: b0 and b1 don't seem to be in the schema above...?  doesn't `ms.validate_and_encode_row` error?

bhaller: I see `schema_with_binary()` is now below (thanks, the diffs are much easier to read now!), and it defines `b0` and `b1` (for num_binary_ints >= 2, thanks for the rename of that parameter also); but I don't see it being called from here, so `b0` and `b1` don't exist here, do they?  Or am I blind?  I do like that the code here assembles the schema all by itself, without calling out to that function and then modifying the schema (thanks for that change too); but now it seems like the schema used here is incomplete...?

python/tests/test_metadata.py
```diff
@@ -657,19 +657,98 @@ def test_round_trip_with_struct_and_json(self):
...
+        # To verify this, we (somewhat pedantically) let:
+        # X = (bytes to encode the json without any binary)
+        # Y = (bytes to encode the same json with k ints in binary)
+        # and then it should be that
+        # (a) Y-X is equal to k * (bytes per int)
+        # which implies that the binary portion starts after X bytes; so
+        # (b) X should be divisible by 8.
+        ms = self.schema_with_binary(k)
+        ms0 = self.schema_with_binary(0)
+        bytes_per_blob = len(struct.pack("i", 0))
```
bhaller: excise the word "blob" here as well please; `bytes_per_binary_int`?

python/tests/test_metadata.py
```diff
@@ -657,19 +657,98 @@ def test_round_trip_with_struct_and_json(self):
...
+    @pytest.mark.parametrize("k", (0, 1, 5, 1001))
+    def test_byte_alignment(self, k):
+        # We want to test whether the binary portion begins byte-aligned.
+        # To verify this, we (somewhat pedantically) let:
+        # X = (bytes to encode the json without any binary)
+        # Y = (bytes to encode the same json with k ints in binary)
+        # and then it should be that
+        # (a) Y-X is equal to k * (bytes per int)
+        # which implies that the binary portion starts after X bytes; so
+        # (b) X should be divisible by 8.
```
bhaller: I read this through a couple of times, and didn't really understand it until I looked at the code below.  The key misunderstanding that confused me: (a), as you say, "should" be true, but has nothing to do with the assertion that byte-alignment is working correctly.  (b) is what we are actually setting out to test, and _must_ be true given the previous statements if byte-alignment is working correctly.  The way that (a) and (b) are presented as being kind of of equal footing confused me somehow.  I dunno, I'm easily confused these days, it seems like.  I think it'd be clearer if you set out: X = (bytes to encode the json without any binary); (2) if byte-alignment is working correctly, then X must be divisible by 8, since the padding is emitted whether there is any binary content after the JSON or not.  So we can simply test X for divisibility; but to make sure that this padding doesn't change when binary bytes are provided in addition to JSON, we have to do the following logic...

I dunno, I think that would make more sense to my brain.  :->  Anyhow, this test makes sense to me and I give it a thumbs-up.

python/tests/test_metadata.py
petrelharp: yes

python/tests/test_metadata.py
petrelharp: Oh good question! Experimenting, it seems like no; even the "items" is not needed. Removing.

Certain things about metadata could be, ahem, better documented, but getting too far into that would be out of scope (as I don't understand it all well enough to write those docs).

python/tests/test_metadata.py
petrelharp: Hah, you're right. Whoops. I think what's happening here is that the JSON codec is by default permissive (you can stick anything in there, which is what we want for top-level metadata), so anything not hit by `struct` will go to binary. Adding those to the binary bit.

bhaller (approved):
LGTM, thanks for the revisions.  Note the two comments that maybe you missed, but they are trivial so there's no need to hold things up for them.  :->

python/tests/test_metadata.py
petrelharp: revised this comment

petrelharp:
Merging since it's been carefully reviewed by Ben H, and by me-after-waiting-two-weeks, which is sorta like a different person. (And, I consulted with Jerome.)

PR #3442, opened 2026-03-24
Support `node_labels` in `write_nexus` via `TRANSLATE`

kaathewisegit:
## Description

Adds a `node_labels` parameter to `write_nexus` which allows renaming samples.  The trees still use the fast Newick serializer, the remapping is done with a TRANSLATE directive.

Resolves #3435

# PR Checklist:

- [?] Tests that fully cover new/changed functionality.

### DIFF
--- a/python/CHANGELOG.rst
+++ b/python/CHANGELOG.rst
@@ -11,7 +11,8 @@ In development
 
 - Add ``TreeSequence.ld_matrix`` stats method and documentation, for computing
   two-locus statistics in site and branch mode.
-  (:user:`lkirk`, :user:`apragsdale`, :pr:`3416`) 
+  (:user:`lkirk`, :user:`apragsdale`, :pr:`3416`)
+- Add `node_labels` parameter to `write_nexus`. (:user:`kaathewisegit`, :pr:`3442`)
 
 --------------------
 [1.0.2] - 2026-03-06
--- a/python/tests/test_phylo_formats.py
+++ b/python/tests/test_phylo_formats.py
@@ -26,6 +26,7 @@ Tests for phylogenetics export functions, newick, nexus, FASTA etc.
 
 import functools
 import io
+import random
 import textwrap
 
 import dendropy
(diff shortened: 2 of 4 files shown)


python/CHANGELOG.rst
```diff
@@ -7,6 +7,8 @@ In development
 - Add ``json+struct`` metadata codec that allows storing binary data using a struct
   schema alongside JSON metadata. (:user:`benjeffery`, :pr:`3306`)
 
+- Add `node_labels` parameter to `write_nexus`.
```
```suggestion
- Add `node_labels` parameter to `write_nexus`. (:user:`kaathewisegit`, :pr:`3442`)
```

python/tskit/text_formats.py
```diff
@@ -166,6 +173,11 @@ def write_nexus(
     include_trees = True if include_trees is None else include_trees
     if include_trees:
         print("BEGIN TREES;", file=out)
+
+        if node_labels:
```
jeromekelleher:
```suggestion
        if node_labels is not None:
```

Local style prefers explicit check against none instead of truthy.

python/tests/test_phylo_formats.py
```diff
@@ -474,6 +474,68 @@ def test_as_nexus_precision_1(self):
...
+            """
+        )
+        assert expected == ts.as_nexus(
+            include_alignments=False,
+            node_labels=labels,
+        )
+
+    def test_as_nexus_labels_empty(self):
+        ts = self.tree().tree_sequence
+        labels = {}
```
jeromekelleher: I  think in this case we should get a key error not have it treated as None. This will happen with the explicit check for None I suggested above.

jeromekelleher (commented):
Looks great, thanks @kaathewisegit! A few minor bits of input above.

It would be nice to verify that a third-party tool like dendropy or BioPython recogises this - would it be possible to check this?

kaathewisegit:
I've changed `write_nexus`  to raise errors if `node_labels` is empty or if it contains keys above `ts.num_samples`.

I also tried adding tests which use BioPython (generate a random ts and see if its NEXUS output is parseable), but it turns out that Bio.Nexus only accepts integer ids for TRANSLATE, so `n{i}` get rejected as illegal.

https://github.com/biopython/biopython/blob/d76a5581ca3317ec850d72450c458ec4e34e2583/Bio/Nexus/Nexus.py#L1113

python/tskit/text_formats.py
```diff
@@ -134,7 +135,21 @@ def write_nexus(
     print("#NEXUS", file=out)
     print("BEGIN TAXA;", file=out)
     print("", f"DIMENSIONS NTAX={ts.num_samples};", sep=indent, file=out)
-    taxlabels = " ".join(f"n{u}" for u in ts.samples())
+
+    if node_labels is not None:
+        if len(node_labels) == 0:
+            raise ValueError("`node_labels` cannot be empty")
+
+        if max(node_labels) >= ts.num_samples:
```
jeromekelleher: This isn't right semantically - sample IDs can be anything, not just 0...n.

python/tskit/text_formats.py
```diff
@@ -134,7 +135,21 @@ def write_nexus(
...
+            raise ValueError(
+                "Can only remap samples, got node with id {max(node_labels)}"
+            )
+
+        taxlabels = " ".join(
```
jeromekelleher, line 140: I can see two ways of doing this: strict or lenient. Here's the strict way:
```
if node_labels is not None:
    taxlabels = " ".join(node_labels[u] for u in ts.samples() # raises a KeyError if a sample isn't in node_labels
else:
      taxlabels = " ".join(f"n{u}" for u in ts.samples())
```
This is the lenient one:
```
if node_labels is not None:
    taxlabels = " ".join(node_labels[u]  if u in node_labels else f"n{u}" for u in ts.samples() 
else:
      taxlabels = " ".join(f"n{u}" for u in ts.samples())
```
So the lenient one just remaps to the standard mapping if a key is missing. I can see pros and cons in both, but I think if we do the lenient version we may as well go full lenient and not bother trying to catch edge cases (which might be valid for someone, who knows).

The strict version will raise a KeyError which will hopefully be interpretable enough, and we don't need to do anything else.

I'm happy to go either direction, whatever you think.

jeromekelleher (commented):
Some comments on the implementation.

It seems like dendropy should support TRANSLATE, so should be easy enough to use that to verify? Just one example is fine, we just want to be sure that some other software can read this.

python/tskit/text_formats.py
kaathewisegit: All of the other methods/options are lenient, so I think it makes sense to keep it that way.  I removed the checks and replaced the current logic with Yours

*(a stricter version would probably need to check the names too, which is tricky, because I'm unsure what the rules are for various NEXUS parsers)*

python/tskit/text_formats.py
kaathewisegit: Yeah, I didn't realize this was possible, I've removed this check as per the other comment

kaathewisegit:
I've switched to a lenient translation table construction.  And added dendropy parsing tests for all tree sequences from get_example_tree_sequences (except multiple-root and empty ones, as newick serialization doesn't work for those).

python/tests/test_phylo_formats.py
```diff
@@ -334,6 +335,85 @@ def test_nexus_no_trees_or_alignments(self):
...
+                return
+
+        labels = {}
+        samples = ts.samples()
+        k = random.randint(1, len(samples))
+        for node in random.sample(list(samples), k):
+            labels[node] = f"new_node_which_was_{node}"
+
+        nexus = ts.as_nexus(include_alignments=False, node_labels=labels)
+        print(nexus)
```
jeromekelleher: Stray print

python/tests/test_phylo_formats.py
```diff
@@ -334,6 +335,85 @@ def test_nexus_no_trees_or_alignments(self):
...
+
+    @pytest.mark.parametrize("ts", get_example_tree_sequences())
+    def test_parseable(self, ts):
+        for tree in ts.trees():
+            if not tree.has_single_root:
+                return
+
+        labels = {}
+        samples = ts.samples()
+        k = random.randint(1, len(samples))
```
jeromekelleher, line 417: I don't think the random does much here, I'd just do something like 
```
u = ts.samples()[0]
labels[u] = f"new_node_which_was_{u}"
```

And then after just check that ``labels[u]`` is recognised as a taxon ID by dendropy?

python/tskit/trees.py
```diff
@@ -6896,6 +6897,10 @@ def write_nexus(
         :param str missing_data_character: As for the :meth:`.alignments` method,
             but defaults to "?".
         :param bool isolated_as_missing: As for the :meth:`.alignments` method.
+        :param node_labels: A map of type `{index: name}`.  Samples present in
```
jeromekelleher:
```suggestion
        :param node_labels: A map of type `{node_id: name}`.  Samples present in
```

python/tskit/trees.py
```diff
@@ -6896,6 +6897,10 @@ def write_nexus(
...
+            the map will have the given name instead of `n{index}`.  Note that
```
jeromekelleher:
```suggestion
            the map will have the given name instead of `n{node_id}`.  Note that
```

Index is ambigous here, better to be clear

jeromekelleher (commented):
Great! Looks like it's almost ready to go in.

python/tskit/trees.py
kaathewisegit: fixed

python/tests/test_phylo_formats.py
kaathewisegit: The idea was to make sure that the parsers support translations given in arbitrary orders, because I wasn't 100% sure that'd be supported for non-integer labels.

> The TRANSLATE statement maps arbitrary labels in the tree specification to valid taxon names. If the arbitrary labels are integers, they are mapped onto the valid taxon names as dictated by the TRANSLATE command without any consideration of the order of the taxa in the matrix

I've added additional asserts which actually check the labels.  This surfaced a bug because dendropy doesn't support samples being internal nodes.  There are 3 tree sequences like that in `get_example_tree_sequences` (`all_nodes_samples`, `internal_nodes_samples`, `mixed_internal_leaf_samples`), so I filtered them out.

jeromekelleher (approved):
Looks great - thanks again @kaathewisegit!

PR #3446, opened 2026-04-14
remove test code to check codecov DO NOT MERGE

petrelharp:
I've just updated `.codecov.yml` in #3439, and so I'm checking that this still is codecov-ing correctly.


petrelharp:
Seems to be working! 🎉

Issue #3445, opened 2026-04-11
two locus stats documentation improvements


petrelharp:
The docs for [summary functions](https://tskit.dev/tskit/docs/latest/stats.html#sec-stats-two-locus-summary-functions) for two-locus stats:
<img width="1480" height="456" alt="Image" src="https://github.com/user-attachments/assets/8b3df38c-9b1e-4854-8a66-0e0a754a04a1" />
is not quite right, I think. The wrong thing is that for unpolarised stats, sometimes A is the ancestral allele (IIUC). A better phrase might be "focal allele". This is a situation where I recommend just writing a for reals definition instead of trying to, like, describe it in sentences. Then give a description. Something like:

> Suppose that at the first site there are alleles (a1, a2, ..., an), while at the second site there are alleles (b1, b2, ..., bn).
> For each pair of alleles ai and bj, define Xij = (n(ai,bj), n(ai,~bj), n(~ai,bj)),
> where n(ai,bj) is the number of two-locus genotypes in the sample set that carry both alleles ai and bj;
> n(ai,~bj) is the number of two-locus genotypes in the sample set that carry allele ai and do not carry bj;
> and similarly n(ai,~bj) is the number of two-locus genotypes in the sample set that do not carry allele ai but do carry bj.
> (We refer to these counts as "AB", "Ab", and "aB".) Then...

This needs to mention polarization (that's why this docstring was incorrect, after all). And there's something I don't understand about polarization. Experimenting, I'm not being able to predict how many times the summary function gets called with and without polarization. So, I could be totally wrong about this?

apragsdale:
Thanks for bringing this up, Peter. I'll have a close look and open a PR to address it.

petrelharp:
Closed in #3447

PR #3447, opened 2026-04-15
More carefully define two-locus haplotypes in docs

apragsdale:
As pointed out in #3445, there is room for improvement in defining the haplotype counts that are passed to the summary functions in the two-locus statistics methods. This more carefully defines those, following advice from @petrelharp, and clarifies the number of times summary functions are called for a given number of alleles at the left and right loci.

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -988,18 +988,46 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
 ##### One-way
 
 The two-locus summary functions all take haplotype counts and sample set size
-as input. Each of our summary functions has the signature
+as input. Suppose that at the first site there are alleles
+{math}`(a_1, a_2, ...)`, and at the second site there are alleles
+{math}`(b_1, b_2, ...)`. For a pair of focal alleles {math}`a_i` and
+{math}`b_j`, we define two-locus counts
+{math}`(n(a_i,b_j), n(a_i,\sim b_j), n(\sim a_i, b_j))`, where
+{math}`n(a_i,b_j)` is the number of two-locus haplotypes in the sample set that
+carry both alleles {math}`a_i` and {math}`b_j`,
+{math}`n(a_i,\sim b_j)` is the number that carry the allele {math}`a_i`
+and do not carry the allele {math}`b_j`, and
+{math}`n(\sim a_i, b_j)` is the number that carry the allele {math}`b_j`
+and do not carry the allele {math}`a_i`. That is,
+{math}`n(\sim a_i, b_j) = \sum_{k\not=i} n(a_k, b_j)`, and
+{math}`n(a_i, \sim b_j) = \sum_{l\not=j} n(a_i, b_l)`.
+
+We informally refer to focal alleles as {math}`A,B` and the above sets of
+haplotypes as {math}`(AB, Ab, aB)`, so that {math}`Ab` refers to the set
+of all haplotypes {math}`(a_i, \sim b_j)` and {math}`aB` refers to
+{math}`(\sim a_i, b_j)`.
+Their counts are labeled similarly: {math}`n_{AB} = n(A,B)`,
+{math}`n_{Ab} = n(A, \sim B)`, and {math}`n_{aB} = n(\sim A, B)`.
+Then each of our summary functions has the signature
 {math}`f(n_{AB}, n_{Ab}, n_{aB}, n)`, converting to haplotype frequencies
-{math}`\{p_{AB}, p_{Ab}, p_{aB}\}` by dividing by {math}`n`. Below,
+{math}`\{p_{AB}, p_{Ab}, p_{aB}\}` by dividing by the number {math}`n` of
+samples in the sample set. Then
 {math}`n_{ab} = n - n_{AB} - n_{Ab} - n_{aB}`, {math}`n_A = n_{AB} + n_{Ab}`
 and {math}`n_B = n_{AB} + n_{aB}`, with frequencies {math}`p` found by dividing
 by {math}`n`.
 
-Our convention is to use {math}`A,B` to denote derived alleles, and {math}`a,b`
-ancestral alleles (or other alleles, if the site is multi-allelic). For
-polarised statistics, we average statistics over all non-ancestral alleles. For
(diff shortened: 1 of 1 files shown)


docs/stats.md
```diff
@@ -988,18 +988,37 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
...
+{math}`(a_1, a_2, ...)`, and at the second site there are alleles
+{math}`(b_1, b_2, ...)`. For a pair of alleles {math}`a_i` and
+{math}`b_j`, we define two-locus counts
+{math}`(n(a_i,b_j), n(a_i,\sim b_j), n(\sim a_i, b_j))`, where
+{math}`n(a_i,b_j)` is the number of two-locus haplotypes in the sample set that
+carry both alleles {math}`a_i` and {math}`b_j`,
+{math}`n(a_i,\sim b_j)` is the number that carry the allele {math}`a_i`
+and do not carry the allele {math}`b_j`, and
+{math}`n(\sim a_i, b_j)` is the number that carry the allele {math}`b_j`
+and do not carry the allele {math}`a_i`. We refer to these haplotypes
```
petrelharp: Perhaps say something more concrete here? "We refer to these haplotypes" is not very precise (what haplotypes? "ai ~bj" is not a haplotype). You've set up this notation, now use it?

petrelharp: And a concrete example would help?

docs/stats.md
```diff
@@ -988,18 +988,37 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
...
 by {math}`n`.
 
-Our convention is to use {math}`A,B` to denote derived alleles, and {math}`a,b`
-ancestral alleles (or other alleles, if the site is multi-allelic). For
-polarised statistics, we average statistics over all non-ancestral alleles. For
-unpolarised statistics, the labeling is arbitrary as we average over all
-alleles (derived and ancestral).
+For polarised statistics, the focal alleles {math}`A` and {math}`B` denote
+derived alleles, as we do not compute the summary function for the ancestral
+alleles. For unpolarised statistics, we compute the summary function over all
```
petrelharp: You don't mean "for the ancestral alleles" here, since no-where have you defined what it means to compute it "for" an allele. Don't you mean something like "the ancestral allele is not used as the focal allele", where "focal allele" is something that probably should have been defined earlier?

petrelharp:
Thanks for this. Good start, needs more preciseness.

apragsdale:
Gotta run to some things today, but tried to add some preciseness here. Thanks Peter.

docs/stats.md
```diff
@@ -988,18 +988,48 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
...
+carry both alleles {math}`a_i` and {math}`b_j`,
+{math}`n(a_i,\sim b_j)` is the number that carry the allele {math}`a_i`
+and do not carry the allele {math}`b_j`, and
+{math}`n(\sim a_i, b_j)` is the number that carry the allele {math}`b_j`
+and do not carry the allele {math}`a_i`. That is,
+{math}`n(\sim a_i, b_j) = \sum_{k\not=i} n(a_k, b_j)`, and
+{math}`n(a_i, \sim b_j) = \sum_{l\not=j} n(a_i, b_l)`.
+For biallelic data, both of these sums have only a single contribution, and
+we sum over multiple haplotype counts when considering a pair of sites that
+have more than two alleles at either site.
```
petrelharp: I don't think this comment is helpful here, because there's two different sums you're talking about at once, and even a biallelic site has two alleles at each to sum over.

docs/stats.md
```diff
@@ -988,18 +988,48 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
...
+
+We informally refer to focal alleles as {math}`A,B` and the above sets of
+haplotypes as {math}`(AB, Ab, aB)`, so that {math}`Ab` refers to the set
+of all haplotypes {math}`(a_i, \sim b_j)` and {math}`aB` refers to
+{math}`(\sim a_i, b_j)`.
+Their counts are labeled {math}`(n_{AB}, n_{Ab}, n_{aB})`.
```
petrelharp: You've set up the notation above, so you might as well use it? (If it won't be used, maybe we don't need it?)
```suggestion
{math}`(\sim a_i, b_j)`.
Their counts are labeled similarly: {math}`n_{AB} = n(A,B)`,
{math}`n_{Ab} = n(A,\sim B)`, and {math}`n_{aB} = n(\sim A, B)`.
```

docs/stats.md
```diff
@@ -988,18 +988,48 @@ sets of samples (see also the note in {meth}`~TreeSequence.divergence`).
...
 
-Our convention is to use {math}`A,B` to denote derived alleles, and {math}`a,b`
-ancestral alleles (or other alleles, if the site is multi-allelic). For
-polarised statistics, we average statistics over all non-ancestral alleles. For
-unpolarised statistics, the labeling is arbitrary as we average over all
-alleles (derived and ancestral).
+For polarised statistics, the focal alleles {math}`A` and {math}`B` denote
+derived alleles, so that we do not compute the summary function with
+haplotype sets for which the focal alleles are the ancestral allele at either
+of the two loci.
```
petrelharp: Changing this to match the next bit better:
```suggestion
For polarised statistics, we compute the statistiic using all pairs of
non-ancestral alleles as focal alleles:
so, we do not compute the summary function with
haplotype counts for which the focal alleles are the ancestral allele at either
of the two loci.
```

petrelharp: "derived" is not the same as "non-ancestral", the way I interpret it: if ancestral is A and we mutate A -> C -> A then that last mutation to A is definitely a derived mutation, so you might interpret A as being both a derived and ancestral allele

petrelharp:
A few more suggestions...

docs/stats.md
apragsdale: What I meant here was that for a biallelic site, there would only be one "not a_i" or "not b_j" allele at each locus, and there are only more than one for multiallelic loci. But this was confusing, so removed it.

petrelharp (approved):
This is a helpful improvement. Thanks!

PR #3453, opened 2026-04-29
Add tables.load docstring

### LINKED ISSUE #3406: Add docstring to TableCollection.load()
When looking over the docs at https://tskit.dev/tskit/docs/stable/python-api.html I never saw `TableCollection.load()` in any of the sections. I only discovered this method when @hyanwong provided an example on Slack. I am happy to do this if we think it is a good idea.

hyanwong:
Fixes https://github.com/tskit-dev/tskit/issues/3406

### DIFF
--- a/python/tskit/tables.py
+++ b/python/tskit/tables.py
@@ -3455,6 +3455,24 @@ class TableCollection(metadata.MetadataProvider):
 
     @classmethod
     def load(cls, file_or_path, *, skip_tables=False, skip_reference_sequence=False):
+        """
+        Load a :class:`TableCollection` from a file or path, saved in the format defined
+        by :meth:`.dump`. Although the file must be in the correct format, unlike
+        :func:`tskit.load` further validation is not performed to check that the tables
+        meet the :ref:`sec_valid_tree_sequence_requirements`. For instance,
+        unsorted tables can be loaded without error and then sorted before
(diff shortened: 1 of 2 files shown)


python/tskit/tables.py
```diff
@@ -3455,6 +3455,20 @@ def __getstate__(self):
 
     @classmethod
     def load(cls, file_or_path, *, skip_tables=False, skip_reference_sequence=False):
+        """
+        Load a TableCollection from a file or path, saved in the format defined by
+        :meth:`.dump`. If the file is not in a valid format, an exception will be
+        raised.
```
petrelharp: This doesn't do any validation, right? (e.g., doing `tables.tree_sequence()` might produce an error) This would be a useful thing to point out here, since that's one of the main uses for this function (loading up tables that don't validate).

molpopgen: Correct -- no validation happens.  The relevant logic is line 4943 of _tskitmodule.c.

hyanwong: Yeah, great point. I will add that comment

petrelharp (approved):
LGTM, thanks!!

petrelharp:
I think we can merge this; sound good?

hyanwong:
Great - I just clicked the merge button.

PR #3460, opened 2026-05-15
Correctly account for isolated samples in normalisation of `pair_coalescence_*` methods

### LINKED ISSUE #3459: Incorrect normalisation in `TreeSequence.pair_coalescence_counts` (and `TreeSequence.pair_coalescence_rates`) with isolated samples
The `pair_normalise` option for `TreeSequence.pair_coalescence_counts` should result in the output being divided by the possible number of coalescing sample pairs---so that if `pair_normalise` and `span_normalise` are both set, the output will sum to one over the time dimension. However, if there are isolated samples over a portion of the tree sequence, then this won't be the case---the denominator is incorrect.

Specifically, let $c_i(x)$ be the number of sample pairs coalescing in node $i$ at position $x$. Without normalisation, `pair_coalescence_counts` returns $C_i = \int_a^b c_i(x) dx$ where the integral is over the interval $[a, b)$ (ignoring "empty" trees). Let $p(x)$ be the number of potentially coalescing sample pairs at position $x$ (that is, omitting isolated samples and pairs where one sample is a descendant of the other). The "full normalisation" should return $C_i / \int_a^b p(x) dx$. If all sample pairs are present over the entirety of the sequence, then $\int_a^b p(x) dx = {n \choose 2} (b - a)$, and the current implementation is hunky dory. But otherwise $\int_a^b p(x) dx < {n \choose 2} (b - a)$ and the normalised output is biased upwards.

A consequence is that pair coalescence rates (which rely on the normalised counts) are biased when there are isolated samples. For example, the figure below shows empirical rates estimated by `TreeSequence.pair_coalescence_rates` (colored lines that are individual simulations) around the theoretical expectation for the Zigzag model (black line). The "full" panel is when all samples are present, and the "drop" panel is when ancestry is partially removed so that the number of non-isolated samples varies dramatically across the sequence.

Image

nspope:
Fixes #3459 

There's the somewhat thorny issue of what the correct normalisation is when only one of `span_normalise` or `pair_normalise` are `True`. In this case (using the notation of the linked issue), what I've done is normalise by the "average number of pairs" $\int p(x) dx / (b - a)$ for "pair normalisation" (so the units are "sequence length per pair") -- this reduces to ${n \choose 2}$ when there's no isolated samples which agrees with the original implementation. Similarly I normalise by the "average pair span" $\int p(x) dx / {n \choose 2}$ for "span normalisation" (so that the units are "# of sample pairs") -- which simplifies to $b - a$ as per the original implementation where there are no isolated samples.

So the output (for the $i$th node or time window) is one of:
- $C_i$ (unnormalised)
- $C_i / \int p(x) dx$ (fully normalised)
- $C_i \times (b - a) / \int p(x) dx$ (pair normalised)
- $C_i \times {n \choose 2} / \int p(x) dx$ (span normalised)

The above assumes that there's no missing trees over the interval--- if there are, replace $b - a$ by "nonmissing span" (the original implementation discounted missing span as well)

### DIFF
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -10630,7 +10630,10 @@ class TreeSequence:
     ):
         """
         Calculate the number of coalescing sample pairs per node, summed over
-        trees and weighted by tree span.
+        trees and weighted by tree span. Precisely, if `c_i(x)` is the
+        number of pairs that coalesce in node `i` at position `x` (which may be
+        zero), then the output for the `i`th node is the integral of `c_i` over
+        the tree sequence (or the intervals defined by `windows`).
 
         The number of coalescing pairs may be calculated within or between the
         non-overlapping lists of samples contained in `sample_sets`. In the
@@ -10647,6 +10650,21 @@ class TreeSequence:
         The argument `time_windows` may be used to count coalescence
         events within time intervals (if an array of breakpoints is supplied)
         rather than for individual nodes (the default).
+
+        The flags `span_normalise` and `pair_normalise` control the units of
+        the output. Let `p(x)` be the number of pairs that could potentially
+        coalesce at position `x` (omitting "isolated" samples). If both
+        `span_normalise` and `pair_normalise` are true, then the output is
+        divided by the integral of `p(x)` over the sequence, will thus sum to
+        one over the "time" dimension (provided all non-isolated samples trace
+        back to a single common ancestor per tree). If only `span_normalise` is
+        set, then the output is divided by `(integral p(x) dx) / (num_samples *
+        (num_samples - 1) / 2)` (the average non-missing sequence length per
+        sample pair).  Similarly, if only `pair_normalise` is set, the output
+        is divided by `(integral p(x) dx) / (nonmissing sequence)` ( the
+        average number of sample pairs per base). The default is
+        `span_normalise` so that the units of the output are "number of sample
+        pairs".
 
         The output array has dimension `(windows, indexes, nodes)` with
         dimensions dropped when the corresponding argument is set to None.
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -10664,8 +10664,8 @@ class TreeSequence:
         sample pair).  Similarly, if only `pair_normalise` is set, the output
         is divided by `(integral p(x) dx) / (nonmissing sequence)` ( the
         average number of sample pairs per base). The default is
-        `span_normalise` so that the units of the output are "number of sample
-        pairs".
+        `span_normalise=True, pair_normalise=False`, so that the units
+        of the output are "number of sample pairs".
 
         The output array has dimension `(windows, indexes, nodes)` with
         dimensions dropped when the corresponding argument is set to None.
--- a/python/tskit/trees.py
+++ b/python/tskit/trees.py
@@ -10633,7 +10633,8 @@ class TreeSequence:
         trees and weighted by tree span. Precisely, if `c_i(x)` is the
         number of pairs that coalesce in node `i` at position `x` (which may be
         zero), then the output for the `i`th node is the integral of `c_i` over
-        the tree sequence (or the intervals defined by `windows`).
+        the tree sequence (or the intervals defined by `windows`),
+        possibly normalized as described below.
 
         The number of coalescing pairs may be calculated within or between the
         non-overlapping lists of samples contained in `sample_sets`. In the
--- a/c/tests/test_stats.c
+++ b/c/tests/test_stats.c
@@ -4030,6 +4030,23 @@ test_pair_coalescence_counts_missing(void)
         &ts, 5, missing_ex_nodes, missing_ex_edges, NULL, NULL, NULL, NULL, NULL, 0);
     verify_pair_coalescence_counts(&ts, 0);
     verify_pair_coalescence_counts(&ts, TSK_STAT_SPAN_NORMALISE);
+    verify_pair_coalescence_counts(&ts, TSK_STAT_PAIR_NORMALISE);
+    verify_pair_coalescence_counts(
+        &ts, TSK_STAT_SPAN_NORMALISE | TSK_STAT_PAIR_NORMALISE);
+    tsk_treeseq_free(&ts);
+}
+
+static void
+test_pair_coalescence_counts_internal(void)
+{
+    tsk_treeseq_t ts;
+    tsk_treeseq_from_text(&ts, 10, internal_sample_ex_nodes, internal_sample_ex_edges,
+        NULL, NULL, NULL, NULL, NULL, 0);
+    verify_pair_coalescence_counts(&ts, 0);
+    verify_pair_coalescence_counts(&ts, TSK_STAT_SPAN_NORMALISE);
+    verify_pair_coalescence_counts(&ts, TSK_STAT_PAIR_NORMALISE);
+    verify_pair_coalescence_counts(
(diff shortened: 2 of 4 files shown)


nspope:
Added an additional test into the C suite for "ancestor-descendant" sample pairs

petrelharp:
Slight correction to the above: I think that $p(x)$ here is (or should be) "potential pairs", which is "number of non-missing choose two" for a single sample set; in #3459 $p(x)$ is defined to exclude cases where one sample is a descendant of another. I don't think we should exclude those cases (since the interpretation is cleaner, for one thing), and I don't think this code does that.

python/tskit/trees.py
```diff
@@ -10630,7 +10630,10 @@ def pair_coalescence_counts(
     ):
         """
         Calculate the number of coalescing sample pairs per node, summed over
-        trees and weighted by tree span.
+        trees and weighted by tree span. Precisely, if `c_i(x)` is the
+        number of pairs that coalesce in node `i` at position `x` (which may be
+        zero), then the output for the `i`th node is the integral of `c_i` over
+        the tree sequence (or the intervals defined by `windows`).
```
```suggestion
        the tree sequence (or the intervals defined by `windows`),
        possibly normalized as described below.
```

python/tskit/trees.py
```diff
@@ -10648,6 +10651,21 @@ def pair_coalescence_counts(
...
+        set, then the output is divided by `(integral p(x) dx) / (num_samples *
+        (num_samples - 1) / 2)` (the average non-missing sequence length per
+        sample pair).  Similarly, if only `pair_normalise` is set, the output
+        is divided by `(integral p(x) dx) / (nonmissing sequence)` ( the
+        average number of sample pairs per base). The default is
+        `span_normalise` so that the units of the output are "number of sample
+        pairs".
+
         The output array has dimension `(windows, indexes, nodes)` with
         dimensions dropped when the corresponding argument is set to None.
```
petrelharp, line 10671: This should be amended to include the `time_windows` case, right?

python/tskit/trees.py
```diff
@@ -10648,6 +10651,21 @@ def pair_coalescence_counts(
...
+        divided by the integral of `p(x)` over the sequence, will thus sum to
+        one over the "time" dimension (provided all non-isolated samples trace
+        back to a single common ancestor per tree). If only `span_normalise` is
+        set, then the output is divided by `(integral p(x) dx) / (num_samples *
+        (num_samples - 1) / 2)` (the average non-missing sequence length per
+        sample pair).  Similarly, if only `pair_normalise` is set, the output
+        is divided by `(integral p(x) dx) / (nonmissing sequence)` ( the
+        average number of sample pairs per base). The default is
+        `span_normalise` so that the units of the output are "number of sample
+        pairs".
```
```suggestion
        `span_normalise=True, pair_normalise=False`, so that the units
        of the output are "number of sample pairs".
```

python/tskit/trees.py
```diff
@@ -10648,6 +10651,21 @@ def pair_coalescence_counts(
         events within time intervals (if an array of breakpoints is supplied)
         rather than for individual nodes (the default).
 
+        The flags `span_normalise` and `pair_normalise` control the units of
+        the output. Let `p(x)` be the number of pairs that could potentially
+        coalesce at position `x` (omitting "isolated" samples). If both
+        `span_normalise` and `pair_normalise` are true, then the output is
+        divided by the integral of `p(x)` over the sequence, will thus sum to
+        one over the "time" dimension (provided all non-isolated samples trace
```
petrelharp, line 10660: wait but what if the output is nodes?

python/tskit/trees.py
```diff
@@ -10648,6 +10651,21 @@ def pair_coalescence_counts(
...
+        back to a single common ancestor per tree). If only `span_normalise` is
+        set, then the output is divided by `(integral p(x) dx) / (num_samples *
+        (num_samples - 1) / 2)` (the average non-missing sequence length per
```
petrelharp, line 10663: Is this right? since according to this paragraph, this is just `span_normalise=True, pair_normalise=True` multiplied by (num_samples choose 2)?  (also btw this description doesn't work for distinct sample sets)

*Note:* I'll believe you if you say it is right; I don't remember where we got to on this.

petrelharp:
Could we record - hopefully in the documentation - concrete interpretations of what's being returned? Let's see, IIRC this is:

- `span_normalise=False, pair_normalize=False`: "find for each pair on how much of the genome do they coalesce in this node/time interval, and add that up across all pairs"
- `span_normalise=True, pair_normalize=True`: "pick a uniform (pair, location) out of all the possible ones; what's the probability that that pair at that position coalesces in this node/time interval"

I'm not sure right now what the interpretations for the others are (there may not be a tidy interpretation for the case of missing data; but there should at least be for the no-missing-data case).

petrelharp:
Gee, this code is very nice, but there's a lot to get my head around.

I haven't yet found whether we are also comparing to the naive method to check this for correctness - is that happening?

nspope:
> I haven't yet found whether we are also comparing to the naive method to check this for correctness - is that happening?

Not in the code base, no --- were you thinking some sort of statistical check / benchmark baked into the unit tests? There's worked examples in there, but no "accuracy threshold" (which seems like it'd be a brittle test).

I've tested outside (and could post a MWE to the parent issue, if that'd help, alongside the figure shown there).

petrelharp:
No, I mean a very naive "iterate over all pairs and count" method that we can compare to the efficient version (should be identical): we do have this in the testing code (`naive_pair_coalescence_counts`); but I haven't figured out if this has been updated to account for missing data?

python/tskit/trees.py
nspope: Let's see --- the denominator we're calculating here is "number of extant sample pairs summed across the contig" (that's p(x) at a particular coordinate x) then dividing it by the total number of pairs (however that is calculated, i.e. for distinct sample sets). So the output is just a scalar multiple of the "fully normalised" case.

Not sure what the question is exactly, let's chat about it in person then I can post a followup for clarity?

python/tskit/trees.py
nspope: I'm being sloppy here--- nodes are counted as the time dimension, and their values also sum to one (as time windowing is just summing node values within a window).

python/tskit/trees.py
nspope: good catch, so `nodes or time windows`

nspope:
> I don't think we should exclude those cases (since the interpretation is cleaner, for one thing), and I don't think this code does that.

The code actually is excluding these cases (that's the "upwards propagation" bit of the update) -- are you saying that we shouldn't? This'll certainly simplify the code but will bias the downstream statistics (rates) if so.

nspope:
> No, I mean a very naive "iterate over all pairs and count" method that we can compare to the efficient version (should be identical): we do have this in the testing code (naive_pair_coalescence_counts); but I haven't figured out if this has been updated to account for missing data?

aha --- yes, the naive implementation does the correct thing for missing data already. There is a comparison on simulations the ("messy_ts" test fixture), though it could possibly use extending.

nspope:
After discussing: @petrelharp and I think we should change the numerator to count "ancestor and descendant that are both samples" as a valid pair that coalesces at the "ancestor sample" node. This will simplify the denominator and in general makes for less complicated logic.

This is a change in behavior, but for a very niche use case. It doesn't impact the coalescence rates calculator, which currently requires all samples to be the same age.

PR #3475, opened 2026-08-07
metadata access counter; closes #3472

### LINKED ISSUE #3472: cache top-level metadata?
Decoding of metadata is usually [cached](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/metadata.py#L1077), using this [lazy decoder](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/metadata.py#L1096); for instance [here](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/tables.py#L65).

But top-level metadata is *not* cached. So, anytime you do `ts.metadata` (or `tables.metadata`) it decodes it anew. This is not ideal in the usual situation because if there's, say, a loop over all mutations, and in the loop someone pulls something out of top-level metadata then the metadata is parsed every time. This could be fixed by for instance assigning `md = ts.metadata` and using `md` in the loop, but this isn't intuitive because `ts.metadata` looks just like a dictionary. *Edit:* the decoding may be part of the problem, but the bigger problem is that a separate new copy of the metadata dictionary is returned every time: see [below](https://github.com/tskit-dev/tskit/issues/3472#issuecomment-5197443373).

However, the reason I'm writing this is that with SLiM wanting to store large amounts of data in top-level metadata (using the json+struct codec). So, every time anything is pulled out of top-level metadata, it decodes the entire thing. This happens even if you just want to pull a single number out of the JSON blob (say, `ts.metadata["SLiM"]["tick"]`) then the whole binary struct blob is decoded also.

We can see the effects of this: here's timing (using `python -m timeit`) of (a) `tskit.load`; (b) accessing `ts.metadata["SLiM"]["tick"]`, in JSON; (c) accessing `ts.metadata["SLiM_mutation_list"][0]["slim_time"]` (a single number pulled out of the binary blob) as a function of the number of mutations:

Image

Runtime to access a single value in metadata is increasing with the number of mutations - which is the size of the binary blob in metadata. Furthermore, it's increasing fast enough it'll be a problem. For instance, with 3M mutations, extrapolating, it'd take 10s to load the metadata (once). Furthermore, since the per-access cost of pulling something out of metadata is proportional to the number of mutations, the cost of looping over mutations and pulling something out of metadata for each is quadratic in the number of mutations. 

On the user's side the solution is to do `md = ts.metadata` and use `md` as noted above. But this doesn't help out pyslim, for which there are functions (e.g., like `ts.slim_time( )`) that need to pull something from metadata every time you call the function. We also can't use the existing [`lazy_decode`](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/metadata.py#L1096) to wrap the class definition because `TreeSequence` and `TableCollection` aren't dataclasses.

(Tangentially, I'm confused about how the usual cacheing works: it seems to me that the usual use case is things like `for mut in ts.mutations(); x = mut.metadata['foo']` for which caching would actually slow things down? But I haven't gone back to investigate the original implementation; I think there was benchmarking.)

petrelharp:
As lengthily discussed in #3472, if top-level metadata is large then doing `ts.metadata` a lot of times can be verrrrry slow, because of both the decoding and making a copy of the data (in roughly equal measure). Seeing no straightforward way around this, I'm proposing this stopgap: if someone does `TreeSequence.metadata` or `TableCollection.metadata` 20 times then they get a Warning suggesting they do something else. The warning only occurs once.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -153,6 +153,70 @@ time *units* of a tree sequence should be stored in the
 metadata. For example, we could set `tables.time_units = "generations"`.
 :::
 
+
+(sec_metadata_top_level_pitfalls)=
+
+### Pitfalls related to metadata
+
+Briefly, the pitfalls are:
+
+1. editing metadata directly can silently do nothing, and
+2. making repeated calls to top-level metadata will make your code very slow
+    if that metadata is large.
+
+One thing that it is important to know about metadata is that metadata access methods
+return a *copy* of the *decoded* information.
+So, for instance, each time you call
+```{code-cell}
+tables.metadata
+```
+(or ``ts.metadata``), it gives you a dictionary containing a *new copy* of the
+underlying information. So, this means that you cannot edit metadata
+by direct assignment. For instance, trying to change the metadata like this
+silently does nothing:
+```{code-cell}
+tables.metadata["taxonomy"]["subspecies"] = "lyrata"
+print(tables.metadata["taxonomy"]["subspecies"])
+```
+This is true for all types of metadata:
+for instance, editing `ts.mutation(0).metadata` will not change the metadata
+of the underlying mutation.
+See {ref}`sec_tutorial_metadata` for examples of modifying metadata in tables.
+To edit top-level metadata, first copy out the metadata, edit *that*,
+and then put it back:
+```{code-cell}
+md = tables.metadata
+md["taxonomy"]["species"] = "lyrata"
+tables.metadata = md
+```
+
+This fact about metadata can have important consequences for performance.
+For instance, if we'd like to convert all mutation times to generations,
+we should **definitely not** run
+``[mut.time / ts.metadata["generation_time"] for mut in ts.mutations()]``.
+This is because a new copy of the top-level metadata will be created
+for *every* mutation in the tree sequence.
+This may not be a big deal for some tree sequences,
+but this will take prohibitively long for tree sequences
+that have a large amount of information in top-level metadata
+(for instance, those produced by SLiM v6).
+Instead, it is good practice to make a copy of the top-level metadata,
+and use that copy henceforth, like so:
+```{code-cell}
+:tags: ["skip-execution"]
+top_md = ts.metadata
+mut_times = [
+    mut.time / top_md["generation_time"] for mut in ts.mutations()
+]
+```
+
+Because of this, tskit will produce a Warning if the top-level metadata
+is large (greater than 100Kb) and is accessed many times (more than 20 times).
+This behavior can be changed, by setting
+{data}`tskit.METADATA_ACCESS_WARNING_COUNT`
+or {data}`tskit.METADATA_ACCESS_WARNING_SIZE`.
+
+
 (sec_metadata_examples_reference_sequence)=
 
 ### Reference sequence
@@ -284,8 +348,9 @@ must be encoded and decoded. The C API does not do this, but the Python API will
 use the schema to decode the metadata to Python objects.
 The encoding for doing this is specified in the top-level schema property `codec`.
 Currently the Python API supports the `json` codec which encodes metadata as
-[JSON](https://www.json.org/json-en.html), and the `struct` codec which encodes
-metadata in an efficient schema-defined binary format using {func}`python:struct.pack` .
+[JSON](https://www.json.org/json-en.html), the `struct` codec which encodes
+metadata in an efficient schema-defined binary format using {func}`python:struct.pack`,
+and the `json+struct` codec which is a combination of the two.
 
 (sec_metadata_codecs_json)=
 
@@ -435,6 +500,12 @@ The supported numeric and boolean types are:
   - float64
   - 8
 ```
+
+In addition to the `binaryFormat` encoding given in the table above,
+the `type` key must also be set to the appropriate value.
+For boolean values the `type` should be `boolean` (not bool);
+for integer binary formats it should be `integer` (not number),
+and for floating-point binary formats it should be `number`.
 
 When attempting to pack a non-integer using any of the integer conversion
 codes, if the non-integer has a `__index__` method then that method is
@@ -602,7 +673,7 @@ into 8-byte alignment; and
 (7) the binary data.
 The JSON data is encoded as ASCII, without a null terminating byte,
 and the format of the binary data is specified using the "struct" portion
-of the metadata schema, described :ref:`above <sec_metadata_codecs_struct>`.
+of the metadata schema, described {ref}`above <sec_metadata_codecs_struct>`.
 
 (sec_metadata_schema_examples)=
 
--- a/python/tests/test_highlevel.py
+++ b/python/tests/test_highlevel.py
@@ -42,7 +42,7 @@ import tempfile
 import unittest
 import uuid as _uuid
 import warnings
-from xml.etree import ElementTree
+from html.parser import HTMLParser
 
 import kastore
 import msprime
@@ -1988,7 +1988,7 @@ class TestTreeSequence(HighLevelTestCase):
     def test_html_repr(self, ts):
         html = ts._repr_html_()
         # Parse to check valid
-        ElementTree.fromstring(html)
+        HTMLParser().feed(html)
         assert len(html) > 5000
         assert f"<tr><td>Trees</td><td>{ts.num_trees:,}</td></tr>" in html
         assert f"<tr><td>Time Units</td><td>{ts.time_units}</td></tr>" in html
@@ -3059,19 +3059,24 @@ class TestTreeSequenceMetadata:
         tc = tskit.TableCollection(1)
         ts = tc.tree_sequence()
         assert ts.metadata == b""
+        assert ts.metadata_size == 0
         tc.metadata_schema = self.metadata_schema
         data = {
             "table": "tree-sequence",
             "string_prop": "stringy",
             "num_prop": 42,
         }
+        data_enc = self.metadata_schema.validate_and_encode_row(data)
         tc.metadata = data
         ts = tc.tree_sequence()
         assert ts.metadata == data
+        assert ts.metadata_size == len(data_enc)
         with pytest.raises(AttributeError):
             ts.metadata = {"should": "fail"}
         with pytest.raises(AttributeError):
             del ts.metadata
+        with pytest.raises(AttributeError):
+            ts.metadata_size = 0
 
     def test_tree_sequence_time_units(self):
         tc = tskit.TableCollection(1)
@@ -3712,7 +3717,7 @@ class TestTree(HighLevelTestCase):
     def test_html_repr(self, ts_fixture):
         html = ts_fixture.first()._repr_html_()
         # Parse to check valid
-        ElementTree.fromstring(html)
+        HTMLParser().feed(html)
         assert len(html) > 1900
         assert "<tr><td>Total Branch Length</td><td>" in html
 
--- a/python/tests/test_metadata.py
+++ b/python/tests/test_metadata.py
@@ -2775,3 +2775,117 @@ class TestStructuredArrays:
 
         dtype = metadata.MetadataSchema(schema).numpy_dtype()
         assert dtype.names == ("id", "name", "age")
+
+
+class TestMetadataAccessCounter:
+    """
+    Test for top-level metadata multiple access warnings.
+    """
+
+    big_size = tskit.METADATA_ACCESS_WARNING_SIZE
+
+    def get_example(self, n, what):
+        # note this cannot be a fixture because the _metadata_access_counter
+        # will then persist across tests
+        t = tskit.TableCollection(sequence_length=1)
+        schema = metadata.MetadataSchema(
+            {
+                "codec": "json+struct",
+                "json": {
+                    "codec": "json",
+                    "type": "object",
+                    "properties": {
+                        "a": {"type": "string"},
+                    },
+                },
+                "struct": {
+                    "codec": "struct",
+                    "type": "object",
+                    "properties": {
+                        "x": {
+                            "type": "array",
+                            "arrayLengthFormat": "Q",
+                            "items": {"type": "integer", "binaryFormat": "q"},
+                        }
+                    },
+                },
+            }
+        )
+        md = {"a": "bcde", "x": list(range(n))}
+        t.metadata_schema = schema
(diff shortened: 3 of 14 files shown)


python/tskit/tables.py
```diff
@@ -3196,6 +3198,28 @@ def has_reference_sequence(self):
         """
         return bool(self._ll_tables.has_reference_sequence())
 
+    @property
+    def metadata(self):
+        if self._metadata_access_counter is not None:
+            self._metadata_access_counter += 1
+            if self._metadata_access_counter > 20:
```
nspope: Rather than hardcoding 20 in multiple places, how about defining METADATA_WARNING_COUNT in `__init__.py` alongside the other constants, and importing it for use here (and in the tests)?

python/tests/test_metadata.py
```diff
@@ -2775,3 +2775,40 @@ def test_explicit_ordering(self):
...
+        md = {"a": "bcde", "x": [1, 2, 3, 4]}
+        t.metadata_schema = schema
+        t.metadata = md
+        return t
+
+    def check_warns(self, t):
+        # should warn after 20 times and no more after that
+        md = t.metadata
+        for _ in range(19):
+            print(_, t._metadata_access_counter)
```
nspope: stray print?

python/tskit/tables.py
```diff
@@ -4664,6 +4689,22 @@ def metadata_schema(self):
 
     @property
     def metadata(self):
+        if self._metadata_access_counter is not None:
+            # can't set things directly on this immutable class
+            builtins.object.__setattr__(
```
nspope: Do you need builtins here? `object` exists without importing, so I think you can just do `object.__setattr__` here (and elsewhere) and remove the import. (That's done elsewhere, for example with the `_initialised` member of the same class)

nspope (approved):
Looks fine by me, some minor suggestions

nspope:
Here's one thing which isn't wrong per se but is a bit messy: checking equality may raise the counters of the involved tree sequences/tables, to a degree that varies depending on what's compared. For example,

```python
import msprime
import tskit

ts = msprime.sim_ancestry(3, sequence_length=1, random_seed=1024)
tables = ts.dump_tables()
tables.metadata_schema = tskit.MetadataSchema.permissive_json()
tables.metadata = {"a": "bcde", "x": [1, 2, 3, 4]}
ts = tables.tree_sequence()

# mutable vs mutable, same metadata (+0, +0)
t1, t2 = ts.dump_tables(), ts.dump_tables()
t1.assert_equals(t2)
print(t1._metadata_access_counter, t2._metadata_access_counter)  # (0, 0)

# mutable vs mutable, diff metadata (+2, +2)
t3, t4 = ts.dump_tables(), ts.dump_tables()
t4.metadata = {"a": "zzz"}
try:
    t3.assert_equals(t4)
except AssertionError:
    pass
print(t3._metadata_access_counter, t4._metadata_access_counter)  # (2, 2)

# immutable vs immutable, same metadata (+1, +1)
t5, t6 = ts.tables, ts.dump_tables().tree_sequence().tables
t5.assert_equals(t6)
print(t5._metadata_access_counter, t6._metadata_access_counter) # (1, 1)
```

I'm not sure there's an easy way to avoid this though (and it's just cosmetic, would just add some unavoidable noise when comparing things in a loop for instance)

petrelharp:
Good point about comparisons. I don't think I'm worried about it, though?

Thanks for the input! Will get on it.

python/tskit/tables.py
petrelharp: Hm: I did it that way because of [this bit](https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/tskit/metadata.py#L48). Looks like you're right, though, don't need it.

python/tskit/__init__.py
```diff
@@ -70,6 +70,8 @@
     "provenances",
 ]
 
+#: Threshhold for warning about multiple top-level metadata use
+METADATA_WARNING_COUNT = 20
```
jeromekelleher: As this is part of the public API, I think we should give it a more descriptive name and document how to disable it, like
```
#: Threshhold for warning about multiple top-level metadata use. To disable, set to a large number
# e.g. ``tskit.METADATA_ACCESS_WARNING_THRESHOLD = 2**32``.
METADATA_ACCESS_WARNING_THRESHOLD = 20
```

We should add a short description in the docs somewhere also so that it's findable.

python/tests/test_metadata.py
```diff
@@ -2775,3 +2775,39 @@ def test_explicit_ordering(self):
...
+            assert md == x
+            x["a"] = "this should have no effect"
+
+    def test_warns_tables(self):
+        t = self.get_example()
+        self.check_warns(t)
+
+    def test_warns_tree_sequence(self):
+        t = self.get_example().tree_sequence()
+        self.check_warns(t)
```
jeromekelleher, line 2875: Don't we need test_warnts_immutable_tables also?

Like
```
ts = self.get_example().tree_sequence()
self.check_warns(ts.tables)
```

Also we should test that setting the module variable works so that people can control this. There'll be pytest functionality for this (i.e., setting the module var temporarily)

jeromekelleher (commented):
I agree this is a good least-bad option here given the dependence on the return value being a mutable copy in downstream code. I think we should make the interface "official" though and make it flexible, as there will be cases where this is really annoying and we want to turn it off.

petrelharp:
I've set `METADATA_ACCESS_WARNING_SIZE` to 200kb. First I set it to 1Mb, but the tests were annoyingly slow. Here's some anecdata, showing number of bytes and time it takes to run the test that decodes the metadata 25 times:
```
n             time
1000       0.01s
10000     0.09s
100000   0.92s
1000000 9.2s
```
So, if we set it to 1Mb, then here we're running 12 tests that would each take 9s (in parallel, but still). However, if 200Kb is too small then we should up that. However, I think we're taking 2s per decode, so we're not warning until the user has spent 40s decoding that could be cut to 2s, so I think this is okay.

Note that I'm using the `json+struct` codec, but using just `json` is no faster.

Still TODO: documentation.

petrelharp:
Docs done. This should be ready to go.

petrelharp:
Yep - I've had a careful read through, and I think we're good.

petrelharp:
Here's screenshots of the docs built locally, for convenience:
Screenshot From 2026-08-11 16-35-40
Screenshot From 2026-08-11 16-36-12
Screenshot From 2026-08-11 16-36-23

docs/metadata.md
```diff
@@ -153,6 +153,69 @@ time *units* of a tree sequence should be stored in the
...
+underlying information. So, this means that you cannot edit metadata
+by direct assignment. For instance, trying to change the metadata like this
+silently does nothing:
+```{code-cell}
+tables.metadata["taxonomy"]["subspecies"] = "lyrata"
+print(tables.metadata["taxonomy"]["subspecies"])
+```
+This is true for all types of metadata:
+for instance, editing `ts.mutation(0).metadata` will not change the metadata
+of the underlying mutation.
```
nspope, line 183: As the following sentence deals with top-level metadata specifically, this seems like a good place to point to instructions on how to replace the table-specific metadata (e.g. "Replacing metadata *within* a table (such as a `MutationTable`) is described in XXXX. To edit top level metadata ...").

docs/metadata.md
```diff
@@ -153,6 +153,69 @@ time *units* of a tree sequence should be stored in the
...
+:tags: ["skip-execution"]
+top_md = ts.metadata
+mut_times = [
+    mut.time / top_md["generation_time"] for mut in ts.mutations()
+]
+```
+
+Because of this, tskit will produce a Warning if the top-level metadata
+is large (greater than 100Kb) and is accessed many times (more than 20 times).
+This behavior can be changed, using
```
nspope:
```suggestion
This behavior can be changed, by setting

**Review by nspope — APPROVED** [review]
Docs look good to me, a couple minor nits

File: python/tskit/tables.py [hunk 8]
```diff
@@ -3058,6 +3058,10 @@ def __init__(self, sequence_length=0, *, ll_tables=None):
         self._mutations = MutationTable(ll_table=self._ll_tables.mutations)
         self._populations = PopulationTable(ll_table=self._ll_tables.populations)
         self._provenances = ProvenanceTable(ll_table=self._ll_tables.provenances)
+        if len(self._ll_tables.metadata) > tskit.METADATA_ACCESS_WARNING_SIZE:
```
jeromekelleher: This seems confusing to me and a potential tripwire for people - why not just check the size and the counter at run time? Ah, I guess it's to avoid creating a copy of ``self._ll_tables.metadata`` to get the size. I guess this could be expensive and add some time to the constructor when the metadata is very large.

I think the right way to do this is to add a new method ``TableCollection_get_metadata_size`` around here:
https://github.com/tskit-dev/tskit/blob/12b196a43f5ff8b3e20c5f7b1014fb88e0767ea9/python/_tskitmodule.c#L4052

python/tskit/tables.py
```diff
@@ -3196,6 +3200,32 @@ def has_reference_sequence(self):
         """
         return bool(self._ll_tables.has_reference_sequence())
 
+    @property
+    def metadata(self):
+        if self._metadata_access_counter is not None:
```
jeromekelleher: Then do
```
if self._ll_tables.metadata_size > METADATA_ACCESS_WARNING_SIZE:
    self._metadata_access_counter += 1
```

This is more efficient as it avoids creating another copy of large metadata at instantiation time, and it's more straightforward semantically as you can set the thresholds any time you want.

python/tskit/tables.py
```diff
@@ -3196,6 +3200,32 @@ def has_reference_sequence(self):
...
+            self._metadata_access_counter += 1
+            if self._metadata_access_counter > tskit.METADATA_ACCESS_WARNING_THRESHHOLD:
```
jeromekelleher: Simplest to do
```suggestion
            if self._metadata_access_counter == tskit.METADATA_ACCESS_WARNING_THRESHHOLD:
```
if you want "warn once and once only" semantics?

python/tskit/tables.py
```diff
@@ -3196,6 +3200,32 @@ def has_reference_sequence(self):
...
+                    "If metadata is large, this can slow scripts down considerably. "
+                    "Instead, assign metadata to an object and use that, e.g.: "
+                    "tables_metadata = tables.metadata",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self._metadata_access_counter = None
+        return self.metadata_schema.decode_row(self.metadata_bytes)
+
+    @metadata.setter
```
jeromekelleher, line 3212: Don't need this if we're getting the size dynamically from the low-level code, right?

python/tskit/tables.py
```diff
@@ -4664,6 +4698,22 @@ def metadata_schema(self):
 
     @property
     def metadata(self):
+        if self._metadata_access_counter is not None:
```
jeromekelleher: I feel like there's a strong chance for the semantics to drift here if we have three copies of this code. Could we do something like this in metadata.py

```
def metadata_access_warning(obj, name, guidance):
    if obj.metadata_size > tskit.METADATA_ACCESS_WARNING_SIZE:
           # Some users of this function are marked immutable
           object.__setattr__(obj, "_metadata_access_counter", obj._metadata_access_counter + 1)
           if obj._metadata_access_counter == tskit.METADATA_ACCESS_WARNING_THRESHHOLD:
                    warnings.warn(
                                "It looks like you're making repeated calls to "
                                f"<{name}>.metadata. "
                                 # etc
                       )
```
This requires implementing ``metadata_size`` across the three clients, but I think that's a useful and worthwhile thing anyway. Then each call site is a simple call to this function, and we're sure the semantics will stay the same over time.

jeromekelleher (commented):
I think this is a good approach, but the implementation is a bit complicated. Implementing a cheap ``metadata_size`` property here is straightforward, and avoids jumping through hoops.

petrelharp:
Done! It is certainly cleaner. And hah now I see you only suggested adding `metadata_size` to the low-level code, not the high-level code, but well I did it anyways.

python/tests/test_highlevel.py
```diff
@@ -1988,7 +1988,7 @@ def test_load_tables(self, ts):
     def test_html_repr(self, ts):
         html = ts._repr_html_()
         # Parse to check valid
-        ElementTree.fromstring(html)
+        HTMLParser().feed(html)
```
petrelharp, line 1991: This was needed since I added top-level metadata to the example, which produced a `<details open>` tag, and the default XMLParser doesn't deal with attributes in tags (the "open").

petrelharp:
Okay, this is I think finally ready to go!

python/tskit/__init__.py
```diff
@@ -70,6 +70,13 @@
     "provenances",
 ]
 
+#: Threshhold for warning about multiple top-level metadata use.
+#: A warning will be thrown if top-level metadata is larger than "size" bytes and is
+#: accessed more that "threshhold" times.  To disable, set threshhold to a large number,
+#: for instance: ``tskit.METADATA_ACCESS_WARNING_THRESHOLD = 2**32``.
+METADATA_ACCESS_WARNING_THRESHHOLD = 20
```
jeromekelleher: Sorry, I missed this the last time. I think we should change this to ``METADATA_ACCESS_WARNING_COUNT`` for symmetry with ``METADATA_ACCESS_WARNING_SIZE`` (they are both thresholds). This is backtracking a bit on what I said before, I know.

In any case, there's a typo so we'll need to change it (double "H" in threshold)

jeromekelleher (approved):
Happy to merge after the typo is fixed/variable is renamed.

python/tskit/__init__.py
petrelharp: sounds good! and I have now learned more about the word "threshold"! also, tangentially, that it looks like merriam-webster has a five-year cyclical list for "word of the day": 
Screenshot From 2026-08-13 11-27-22

petrelharp:
Done!

PR #3481, opened 2026-09-03
acknowledge funding and update citations

petrelharp:
We'll probably be updating the citations again soon, but I figured it'd be a good time to do this.

### DIFF
--- a/docs/citation.md
+++ b/docs/citation.md
@@ -2,7 +2,25 @@
 
 # Citing tskit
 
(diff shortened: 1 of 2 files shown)


hyanwong (approved):
LGTM

docs/citation.md
```diff
@@ -2,7 +2,25 @@
 
 # Citing tskit
 
-If you use `tskit` in your work, we recommend citing the [2024 ARG Genetics paper](<https://doi.org/10.1093/genetics/iyae100>) and the [2016 msprime PLOS Computational Biology paper](<http://dx.doi.org/10.1371/journal.pcbi.1004842>):
+If you use `tskit` in your work, we recommend citing the 20206 "tskit paper":
```
jeromekelleher: Typo here. It'll probably end up being 2027 at this point, so simpler to leave out so it doesn't become wrong later?

petrelharp: hah good catch. my brain just refused to see the typo.

PR #3482, opened 2026-09-13
note on debugging bad tables

petrelharp:
This is a thing that I've needed to use a few times and seems worth writing down. It's clearly for developers of tree sequence methods, so it's in the "Developers" section. I don't know that it needs its own page but it didn't really fit elsewhere that I could see.


jeromekelleher (approved):

hyanwong (approved):
LGTM

